PR #2647, opened 2023-12-15
Single sync status button

### LINKED ISSUE #2522: Rework sync screen
## What went wrong? 😲
 
The sync screen is a little hard to follow, and when I've watched people use it, and heard feedback, they always click both buttons when they need to sync manually.

Screenshot 2023-11-21 at 2 51 47 PM

This shows to me that the screen is not clearly explaining what the intention is. 
If users are always syncing both the site and the user status, then we should just combine the two in the manual sync process. If there's a good reason to have them separated, and I think there is, then keep them separate but make it clearer.

## Expected behaviour 🤔
Some suggestions:

- add explanatory text below the heading to say what type of data is synchronised and what clicking the button will do
- make the user synchronisation section separate in some way
  - add tabs, or
  - an accordion, or
  - change colours of the user sync, and/or
  - add an icon for the user
- change the text on the button and/or the icon so that the user sync is distinct

mark-prins:
Fixes #2522

# 👩🏻‍💻 What does this PR do? 
 
Rather than bring the user sync into the manual sync process, I've just unified the api calls from the front end.
The sync process is just too different to the user update process, it doesn't seem sensible to combine the two. Running a separate api call is working though - and it means if the user sync has an error, the sync continues.

I've just had to add a little extra messaging, as there is no sync status step for the user. 

https://github.com/msupply-foundation/open-msupply/assets/9192912/2c7cbd2d-e9fa-4408-bf2d-c7760d41b36c

Tested with and without a suitable central server running.

## 📃 Documentation

- [ ] Update screenshots


andreievg (approved):
Looks good thanks, i've tried updating user password, and getting `Unknown error` and a toast with Login Error (which i think is expected). 
I think since ErrorWithDetails is not used, useUpdateUser can be simplified slightly, to not include details

mark-prins:
Good catch! thanks @andreievg 
have simplified `useUpdateUser` and retested after changing the password.

PR #2650, opened 2023-12-18
Change columns to use location code

### LINKED ISSUE #2641: location should be location code in stocktake and anywhere else that shows location

currently location shows full location name such as when doing a stocktake. users dont want to see the full location name because they can go into location to view full name. the shelves are also labelled with the location code (not full location name) so finding something is easier when looking for the code. 

Screenshot 2023-12-14 at 6 44 55 PM

old msupply  also uses location code not the full name

see scrfeenshot

mark-prins:
Fixes #2641

# 👩🏻‍💻 What does this PR do? 
 
Have changed from location name to location code.

To support that, I've also updated the location search, used when assigning a location:

image

Also noticed that a number of columns are too narrow, so have pushed the width out on those.

Screenshot 2023-12-18 at 12 09 41 PM

Added the column selection option for stocktake detail view, though unasked.

Check that the following usages of location have been changed to display the location code:

- Inbound Shipment
- Inbound Shipment Edit
- Outbound Shipment
- Outbound ShipmentEdit
- Stocktake
- Stocktake edit
- Temperature Breach
- Temperature Log
- Sensor
- Prescriptions
- PrescriptionLineEdit

## 📃 Documentation

- [ ] Possibly screenshots 🤷


andreievg:
@mark-prins, do you think it's a good idea to make a PR to main to just do the requested things in the issue (add column for code, and make columns in stocktake selectable by user) ? And then another refactor PR to develop ?
I think it would be safer (especially if PNG needs this while they are deploying)

mark-prins:
> do you think it's a good idea to make a PR to main to just do the requested things in the issue

I should have expected that
I thought that the change was safe enough, but will break it out and make another PR and another issue.

client/packages/inventory/src/Stocktake/DetailView/ContentArea/ContentArea.tsx
```diff
@@ -112,6 +112,7 @@ export const ContentArea: FC<ContentAreaProps> = ({
             buttonText={t('button.add-item')}
           />
         }
+        enableColumnSelection
```
mark-prins, line 115: not requested in the PR, have added as it has been discussed

client/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -120,33 +120,33 @@ const getColumnLookup = <T extends RecordWithId>(): Record<
...
-    width: 100,
+    width: 130,
     sortable: false,
   },
   deliveredDatetime: {
     label: 'label.delivered',
     key: 'deliveredDatetime',
     format: ColumnFormat.Date,
-    width: 100,
+    width: 130,
```
mark-prins, line 149: I retained the column width changes, they're bug fixes rather than refactors

andreievg (approved):
Thanks @mark-prins !

PR #2652, opened 2023-12-18
Add column helpers

### LINKED ISSUE #2651: Simplify column accessor and sort calls

On implementing #2641 I found that the column generation code contained a lot of duplicated code. I refactored and was then encouraged to make a new PR for the refactor in this [comment](https://github.com/msupply-foundation/open-msupply/pull/2650#issuecomment-1859531162)

Here's the refactor

Create helper functions to replace the boilerplate which is used to access row properties: when grouping is used, this requires a lot of duplicated logic.

mark-prins:
Fixes #2651

# 👩🏻‍💻 What does this PR do? 
 
As part of this, I was concerned at the amount of boilerplate used in column definitions, so I've added a couple of helpers, which has simplified the column code.

Check that the data tables in these areas are still displaying correctly:

- Stocktake : Detail view
- Inbound Shipment : Detail view
- Outbound Shipment : Detail view
- Prescriptions : Detail view

There is a lot more that could be done to simplify the column generation. This change will at least reduce duplication of logic and makes field selection much easier.

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
+++ b/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
@@ -1,69 +1,117 @@
 import { useTranslation } from '@common/intl';
+import { RecordWithId } from '@common/types';
 import { ArrayUtils } from '@common/utils';
 
-type DefaultValues = {
-  single?: string;
-  multiple?: string;
+export type ColumnProperty = {
+  path: string[];
+  default?: string;
 };
-
-type BaseProps<T> = {
-  row: T | { lines: T[] };
-  defaults?: DefaultValues;
-};
+// export type ColumnProperty<T, K> = {
+//   path: (keyof T | K | keyof K[keyof K])[];
+//   default?: string;
+// };
 
 export const useColumnUtils = () => {
   const t = useTranslation();
 
-  const getColumnEntityProperty = <T extends object>({
-    row,
-    entity,
-    key,
-    defaults,
-  }: BaseProps<T> & { entity: keyof T; key: string }) => {
-    if ('lines' in row) {
-      const entities = row.lines.flatMap((line: T) => {
-        const obj = line[entity];
-        return !!obj ? [obj] : [];
-      });
-      if (
-        entities.length !== 0 &&
-        entities[0] !== undefined &&
-        typeof entities[0] === 'object' &&
-        key in entities[0]
-      ) {
-        return ArrayUtils.ifTheSameElseDefault(
-          entities,
-          key as keyof NonNullable<T[keyof T]>,
-          defaults?.multiple ?? t('multiple')
-        );
+  /**
+   * This method tries to extract a single common field value from the input data, using a list of given object paths.
+   *
+   * Data is only extracted from the first matching path, (a path matches if path[0] exists in the input data)
+   *
+   * If the path contains one or more arrays multiple field values are collected.
+   * If all field values are the same this common field value is returned.
+   * If field values differ the default value is returned instead.
+   *
+   * If a path points to a single field the matching field value is returned. If there is no match the default is returned.
+   */
+  //  [
+  //   {
+  //     path: ['lines', 'location', 'code'],
+  //     default: 'multiple',
+  //   },
+  //   {
+  //     path: ['location', 'code'],
+  //     default: 'none',
+  //   },
+  // ]
+
+  const getValue = <T extends object, K extends keyof T>(
+    row: T,
+    property: ColumnProperty
+  ): unknown => {
+    // const isPropOfT = (path: keyof T | K ): path is K => {
+    //   return path in row;
+    // };
+
+    if (property.path.length === 0) {
+      return undefined;
+    }
+
+    const path = property.path[0] as keyof T;
+    if (path === undefined || path === null) return undefined;
+    if (!(path in row)) return undefined;
+    if (property.path.length < 2) return undefined;
+
+    const key = property.path[1] as keyof T;
+    const isObjectProperty = property.path.length > 2;
+
+    if (Array.isArray(row[path])) {
+      if (isObjectProperty) {
+        const propertyKey = property.path[2] as keyof K[keyof K];
+        const arr = (row[path] as T[]).flatMap((line: T) => {
+          const obj = line[key];
+          return !!obj ? [obj] : [];
+        });
+        const isValid =
+          arr.length === 0 ||
+          arr[0] === undefined ||
+          typeof arr[0] !== 'object' ||
+          !(key in arr[0]);
+
+        return isValid
+          ? ArrayUtils.ifTheSameElseDefault(
+              arr,
+              propertyKey as keyof NonNullable<T[keyof T]>,
+              property.default === undefined ? t('multiple') : property.default
+            )
+          : '';
       } else {
-        return '';
+        return ArrayUtils.ifTheSameElseDefault(
+          row[path] as T[],
+          key,
+          property.default === undefined ? t('multiple') : property.default
+        );
       }
     } else {
-      return (
-        row[entity]?.[key as keyof NonNullable<T[keyof T]>] ??
-        defaults?.single ??
-        ''
-      );
+      if (isObjectProperty) {
+        return (
+          row[path]?.[key as keyof NonNullable<T[keyof T]>] ??
+          property.default ??
+          ''
+        );
+      } else {
+        return row[key] ?? property.default;
+      }
     }
   };
 
-  const getColumnProperty = <T extends object>({
-    row,
-    key,
-    defaults,
-  }: BaseProps<T> & { key: keyof T }) => {
-    if ('lines' in row) {
-      const { lines } = row;
-      return ArrayUtils.ifTheSameElseDefault(
-        lines,
-        key,
-        defaults?.multiple ?? t('multiple')
-      );
-    } else {
-      return row[key] ?? defaults?.single;
-    }
+  const getColumnProperty = <T extends RecordWithId>(
+    row: T,
+    props: ColumnProperty[]
+  ) => {
+    return props.reduce(
+      (result, prop) => result ?? getValue(row, prop),
+      undefined as unknown
+    );
   };
 
-  return { getColumnProperty, getColumnEntityProperty };
+  const getColumnPropertyAsString = <T extends RecordWithId>(
+    row: T,
+    props: ColumnProperty[]
+  ) => {
+    return String(getColumnProperty(row, props) ?? '');
+  };
+
+  return { getColumnProperty, getColumnPropertyAsString };
 };
--- a/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
+++ b/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
@@ -2,14 +2,14 @@ import { useTranslation } from '@common/intl';
 import { RecordWithId } from '@common/types';
 import { ArrayUtils } from '@common/utils';
 
-export type ColumnProperty = {
-  path: string[];
-  default?: string;
-};
-// export type ColumnProperty<T, K> = {
-//   path: (keyof T | K | keyof K[keyof K])[];
+// export type ColumnProperty = {
+//   path: string[];
 //   default?: string;
 // };
+type ColumnProperty<T> = {
+  path: T[];
+  default?: string;
+};
 
 export const useColumnUtils = () => {
   const t = useTranslation();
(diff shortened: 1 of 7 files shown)


clemens-msupply (commented):
I found it hard to understand what the utils methods are doing because there are so many cases to consider, e.g. `lines`, `entity`, `key`. While docs comments on the utils methods would help a bit I am not sure if it would still be clear enough. In only understood the methods after comparing the changes in the code where they are used (in old code I found it quite straight forward to understand). For this reason I wonder if the following generalization of the problem could actually simplify it (pseudo code):

```ts
  /**
   * This method tries to extract a single common field value from the input data, using a list of given object paths.
   *
   * Data is only extracted from the first matching path...
   *
   * If the path contains one or more arrays multiple field values are collected.
   * If all field values are the same this common field value is returned.
   * If field values differ the default value is returned instead.
   *
   * If a path points to a single field the matching field value is returned. If there is no match the default is returned.
   */
  getCommonColumnProperty(data, [
    {
      path: ['lines', 'location', 'code'],
      default: 'multiple',
    },
    {
      path: ['location', 'code'],
      default: 'none',
    },
  ]);
```
This method would unify `getColumnEntityProperty` and `getColumnProperty`

mark-prins:
Classy! Thanks @clemens-msupply, great suggestion

andreievg:
I am a bit worried that we are removing type safety with this refactor, and I think we are trying to compensate this with complexity in helpers. 

I had a little play around, and this example is type safe and uses one helper method, not sure if it deals with all use case: https://github.com/msupply-foundation/open-msupply/compare/develop...develop(Column-definition)

I also agree with @clemens-msupply that things seemed to have looked ok previously ¯\_(ツ)_/¯. Btw I think the underlying issue perhaps half doing the grouping, i.e. putting in lines in 'lines' field, rather then aggregating the column when item/batch are toggled, I think the later would be more performant and we would have just one type for the column (always the fragment rather then fragment and lines of fragments), also having the grouping logic in one place might be a good idea), not suggesting a refactor of that now, was just trying to have a deeper look (could be wrong on that one too).

Another thing i was wondering, do we need default for multiples where all values are undefined ? (this is automatically and empty string now)

mark-prins:
> I am a bit worried that we are removing type safety with this refactor, and I think we are trying to compensate this with complexity in helpers.

Can you provide an example of the type safety being lost? The solution I have was type safe for all the scenarios I have.

Yes, there is more complexity in the helpers, but I am happy with that approach. I also prefer putting in an additional layer of the helper ( creating the 'string' version ) rather than pushing the logic up to the caller, as in your example. It is duplication of caller logic that I am trying to avoid.

> that things seemed to have looked ok previously

The refactor has removed a lot of code, and that's a helpful indicator. It means there was a lot of duplicated logic, if you are trying to create columns, it isn't just column definitions that you needed to specify, there was logic in the grouping. Yes, we shouldn't have to group on the front end, yes, the API should have an option to request groups, and we should have better typing of the response. For now though, I just wanted to consolidate the logic; when we find a bug in the approach, I don't want to have to search and replace the fix. I'd like to add a test and fix the problem.

> Another thing i was wondering, do we need default for multiples where all values are undefined ?

I wondered about that - the existing usage has the undefined mapping to empty string, and so I've carried that through. In other words, no - we don't have a use case for a different default value.

In summary - I'm happy with the solution I've done and the approach taken. As always, I appreciate the feedback though! 🙏

client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+      return undefined;
+    }
+
+    const path = property.path[0] as keyof T;
+    if (path === undefined || path === null) return undefined;
+    if (!(path in row)) return undefined;
+    const key = property.path[1] as keyof T;
+
+    if (Array.isArray(row[path])) {
+      if (property.path.length < 2) return undefined;
```
clemens-msupply, line 38: Nit: currently only path.length < 3 is supported, you could check for this and print a warning?

client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (property.path.length === 0) {
+      return undefined;
+    }
+
+    const path = property.path[0] as keyof T;
+    if (path === undefined || path === null) return undefined;
+    if (!(path in row)) return undefined;
+    const key = property.path[1] as keyof T;
+
+    if (Array.isArray(row[path])) {
```
clemens-msupply, line 37:
```suggestion
    if (Array.isArray(row[path])) {
         // For example, `path: ['lines', ...]`
```

client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (property.path.length < 2) return undefined;
+      const isObjectProperty = property.path.length > 2;
+      if (isObjectProperty) {
+        const propertyKey = property.path[2];
```
clemens-msupply, line 41:
```suggestion
         // For example, `path: ['lines', 'item', 'id']`
        const propertyKey = property.path[2];
```

client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        return isValid
+          ? ArrayUtils.ifTheSameElseDefault(
+              arr,
+              propertyKey as keyof NonNullable<T[keyof T]>,
+              property.default === undefined ? t('multiple') : property.default
+            )
+          : '';
+      } else {
+        return ArrayUtils.ifTheSameElseDefault(
```
clemens-msupply, line 60:
```suggestion
        // For example, `path: ['lines', 'numberOfPacks']`
        return ArrayUtils.ifTheSameElseDefault(
```

client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+          row[path] as T[],
+          key,
+          property.default === undefined ? t('multiple') : property.default
+        );
+      }
+    } else {
+      if (property.path.length < 1) return undefined;
```
clemens-msupply, line 67:
```suggestion
      // For example, `path: ['batch']` or `path: ['item', 'id']`
      if (property.path.length < 1) return undefined;
```

clemens-msupply (approved):
Looks good too me, thanks!

I added some code comment suggestions to hopefully make it easier ot see what case is currently handled.

andreievg:
> Can you provide an example of the type safety being lost? The solution I have was type safe for all the scenarios I have.

Oops yeah you are right it's still type safe

> In summary - I'm happy with the solution I've done and the approach taken. As always, I appreciate the feedback though! 

Although the type safety makes me more easy about the change, I am still concerned about the complexity of this abstraction, I think while it's fresh it mights seem ok, but going back to it in the future to make changes would require quite a cognitive/time load. 

>  I also prefer putting in an additional layer of the helper ( creating the 'string' version ) rather than pushing the logic up to the caller, as in your example

Sorry I am struggling to understand (see consumption vs helper size/complexity)

```ts
 getSortValue: row =>
            getColumnPropertyAsString(row, [
              { path: ['lines', 'location', 'code'] },
              { path: ['location', 'code'], default: '' },
            ]),
          accessor: ({ rowData }) =>
            getColumnProperty(rowData, [
              { path: ['lines', 'location', 'code'] },
              { path: ['location', 'code'] },
            ]),
```


**helpers**
 <details>
 
 ```ts
 type ColumnProperty<T> = {
  path: T[];
  default?: string;
};

export const useColumnUtils = () => {
  const t = useTranslation();

  /**
   * This method tries to extract a single common field value from the input data, using a list of given object paths.
   *
   * Data is only extracted from the first matching path, (a path matches if path[0] exists in the input data)
   *
   * If the path contains one or more arrays multiple field values are collected.
   * If all field values are the same this common field value is returned.
   * If field values differ the default value is returned instead.
   *
   * If a path points to a single field the matching field value is returned. If there is no match the default is returned.
   */
  const getValue = <T extends object>(
    row: T,
    property: ColumnProperty<KeysOfUnion<T>>
  ): unknown => {
    if (property.path.length === 0) {
      console.warn('ColumnUtils.getValue: no path provided');
      return undefined;
    }

    if (property.path.length > 3) {
      console.warn('ColumnUtils.getValue: too many path elements provided');
      return undefined;
    }

    const path = property.path[0] as keyof T;
    if (path === undefined || path === null) return undefined;
    if (!(path in row)) return undefined;
    const key = property.path[1] as keyof T;

    // e.g. when grouping by item: `path: ['lines', ...]`
    if (Array.isArray(row[path])) {
Member
@[clemens-msupply](https://github.com/clemens-msupply) clemens-msupply 3 days ago
Suggested change
    if (Array.isArray(row[path])) {
    if (Array.isArray(row[path])) {
         // For example, `path: ['lines', ...]`
@andreievg	Reply...
      if (property.path.length < 2) return undefined;
Member
@clemens-msupply clemens-msupply 3 days ago
Nit: currently only path.length < 3 is supported, you could check for this and print a warning?

@andreievg	Reply...

      const isObjectProperty = property.path.length > 2;
      if (isObjectProperty) {
        // For example, `path: ['lines', 'item', 'id']`
        // to access `item.id` where `lines: []{ item: { id: string }}`
        const propertyKey = property.path[2];
Member
@clemens-msupply clemens-msupply 3 days ago
Suggested change
        const propertyKey = property.path[2];
         // For example, `path: ['lines', 'item', 'id']`
        const propertyKey = property.path[2];
@andreievg	Reply...
        const arr = (row[path] as T[]).flatMap((line: T) => {
          const obj = line[key];
          return !!obj ? [obj] : [];
        });
        const isValid =
          arr.length === 0 ||
          arr[0] === undefined ||
          typeof arr[0] !== 'object' ||
          !(key in arr[0]);

        return isValid
          ? ArrayUtils.ifTheSameElseDefault(
              arr,
              propertyKey as keyof NonNullable<T[keyof T]>,
              property.default === undefined ? t('multiple') : property.default
            )
          : '';
      }
      // For example, `path: ['lines', 'numberOfPacks']`
      // where `lines: []{ numberOfPacks: number }}`
      return ArrayUtils.ifTheSameElseDefault(
        row[path] as T[],
        key,
        property.default === undefined ? t('multiple') : property.default
      );
    }

    // First property isn't an array, e.g. `path: ['batch']` or `path: ['item', 'id']`

    // return early if not enough items in array
    if (property.path.length < 1) return undefined;

    const isObjectProperty = property.path.length > 1;
    if (isObjectProperty) {
      return (
        row[path]?.[key as keyof NonNullable<T[keyof T]>] ??
        property.default ??
        ''
      );
    }

    // default return, just access the property
    return row[path] ?? property.default;
  };

  type KeysOfUnion<T extends object> = T extends T ? NestedKeyOf<T> : never;

  type NestedKeyOf<T> = {
    // eslint-disable-next-line @typescript-eslint/ban-types
    [Key in keyof T & string]: T[Key] extends Function
      ? never
      : T[Key] extends object
      ? `${Key}` | NestedKeyOf<T[Key]>
      : `${Key}`;
  }[keyof T & string];

  const getColumnProperty = <T extends RecordWithId>(
    row: T,
    props: ColumnProperty<KeysOfUnion<T>>[]
  ) => {
    return props.reduce(
      (result, prop) => result ?? getValue(row, prop),
      undefined as unknown
    );
  };

  const getColumnPropertyAsString = <T extends RecordWithId>(
    row: T,
    props: ColumnProperty<KeysOfUnion<T>>[]
  ) => {
    return String(getColumnProperty(row, props) ?? '');
  };

  return { getColumnProperty, getColumnPropertyAsString };
};
 ```
 
 </details>

vs

```ts
          getSortValue: row =>
            getColumnDefinition(row, row => row.location?.code ?? ''),
          accessor: ({ rowData: row }) => {
            getColumnDefinition(row, row => row.location?.code ?? '');
          },

```


**helpers**
 <details>
 
 ```ts
 const useColumnDefinition = () => {
  const t = useTranslation();

  const getColumnDefinition = <T extends object>({
    row: row: T | { lines: T[] },
    getValue:  getValue: (_: T) => string | number,
    {
    defaultMultiple = t('multiple'),
    defaultSingle = '',
  }: {
    defaultMultiple?: string;
    defaultSingle?: string;
  }) => {
    if ('lines' in row) {
      const { lines } = row;
      const values = lines.flatMap(row => {
        const value = getValue(row);
        return !!value ? [{ value }] : [];
      });

      return ArrayUtils.ifTheSameElseDefault(values, 'value', defaultMultiple);
    } else {
      return getValue(row) ?? defaultSingle;
    }
  };
  return getColumnDefinition;
};
 ```
 
 </details>

andreievg:
@mark-prins sorry for being so particular, and it looks like we are spending quite a lot of time on this so not expecting any more action. I did want to express my stance on this, thus wrote that last comment.

I think it would be prudent to invest some time in thinking a process for refactoring in the future (how do we agree on it, how much time do we spend on review, how do we identify the benefits vs time costs), keen to discuss this more in process dev chat.

Btw, I am also conscious of this while doing the sync refactor work atm, which I haven't really communicated that well to the wider team. I think I'll be feeling a bit hypocritical in light of this discussion, when that PR comes through.

PR #2653, opened 2023-12-18
1849 encounter deletion

### LINKED ISSUE #1849: Add Encounter delete status (or option to delete an encounter
Feature request: be able to delete an encounter.

Soft-delete an encounter
- [x] introduce new encounter status
- [x] hide deleted encounters in list api calls (front end)
- [x] when deleted encounter is saved removed all program events for this encounter
- [x] add delete button in encounter view
   - [x] confirmation question
   - [x] only show deleted status in dropdown if encounter is deleted
   - [x] while just deleted encounter is still open user can undeleted by changing the status
- [ ] ~option to view deleted encounter, e.g. to undelete~ out of scope because requires more UI changes
- [x] update dashboard reports to filter out deleted encounters
- [ ] build in reports should filter out deleted encounters
- [x] mark underlying document as deleted (e.g. to make it easier for dashboard reports)
- [x] remove existing document deletion service and graphql code (doesn't look like it will be needed)

clemens-msupply:
Fixes #1849 

# 👩🏻‍💻 What does this PR do? 
 

Besides the points in the issue, I rearranged toolbar slightly putting the status into the top right corner. Still not 100% but a bit better?

image

The build in reports don't need changing because the backend endpoint doesn't returns deleted encounters.

Go to an encounter and delete it. The encounter shouldn't be accessible anymore. Associated events with this a encounter should be deleted, this is harder to test manually but there is already a test case for it in the code.


server/graphql/programs/src/lib.rs
```diff
@@ -251,24 +245,6 @@ impl ProgramsMutations {
...
     }
 
-    async fn delete_document(
-        &self,
-        ctx: &Context<'_>,
-        store_id: String,
-        input: DeleteDocumentInput,
-    ) -> Result<DeleteDocumentResponse> {
-        delete_document(ctx, store_id, input)
-    }
```
clemens-msupply, line 261: Removing unused code...

server/graphql/programs/src/queries/encounter.rs
```diff
@@ -7,14 +7,17 @@ use graphql_core::{
...
 use service::auth::{Resource, ResourceAccessRequest};
 
 #[derive(Union)]
 pub enum EncounterResponse {
     Response(EncounterConnector),
 }
 
+/// Returns a list of encounters.
+///
+/// Deleted encounters are excluded from the returned list.
```
clemens-msupply, line 20: I think this is the expected behaviour (?) We could later have a filter flag to include deleted encounter (see note below)

server/service/src/programs/encounter/encounter_updated.rs
```diff
@@ -1,13 +1,70 @@
...
+    program_event::{ProgramEventService, ProgramEventServiceTrait},
+    update_program_document::{update_program_events, UpdateProgramDocumentError},
+};
+
+use super::{
+    encounter_schema::{self},
+    validate_misc::ValidatedSchemaEncounter,
+};
+
+pub(crate) fn update_encounter_row_and_events(
```
clemens-msupply, line 18: I refactored the following code into this helper method...

server/service/src/programs/encounter/insert.rs
```diff
@@ -138,7 +150,7 @@ fn generate(
         r#type: input.r#type.clone(),
         data: input.data,
         form_schema_id: Some(input.schema_id),
-        status: DocumentStatus::Active,
+        status,
```
clemens-msupply, line 153: I also set the document status to deleted. This comes in handy for the dashboard reports because then we only need to filter out the deleted document once instead of for every report...

mark-prins:
Testing note: I was getting an error on deleting, and had to manually update my JSON schema, adding `DELETED` to the `EncounterStatus` enum

One issue: after deleting, I get this prompt:

image

Is that fixed with the other PR? If not - can we remove it, I haven't changed anything. just loaded the encounter and deleted it.

client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -183,6 +185,29 @@ export const DetailView: FC = () => {
...
+      saveData();
+      navigate(
+        RouteBuilder.create(AppRoute.Dispensary)
+          .addPart(AppRoute.Patients)
+          .addPart(encounter?.patient?.id ?? '')
+          .addQuery({
+            tab: PatientTabValue.Encounters,
+          })
+          .build()
+      );
```
mark-prins, line 207: Currently, if there's an error, you are navigated away from the encounter page. It would be better to stay on the page, and only navigate if the delete succeeds:

```suggestion
      saveData().then(result => {
        if (!result) return;
        navigate(
          RouteBuilder.create(AppRoute.Dispensary)
            .addPart(AppRoute.Patients)
            .addPart(encounter?.patient?.id ?? '')
            .addQuery({
              tab: PatientTabValue.Encounters,
            })
            .build()
        );
      });
```

mark-prins (changes requested):
Overall, that's working well 🙏 
There's a few minor usability issues which I think should be tidied up before merging:
 
* Only navigate away from encounter on success
* Add tooltip to the delete icon button
* Could not find encounter error message??
* The `isDirty` prompt when trying to delete

And a couple of queries:
* Navigate to the encounter list view
* Change the success message

client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -183,6 +185,29 @@ export const DetailView: FC = () => {
...
+  };
+  useEffect(() => {
+    if (!deleteRequest) return;
+    if (
+      (data as Record<string, JsonData>)['status'] ===
+      EncounterNodeStatus.Deleted
+    ) {
+      saveData();
+      navigate(
+        RouteBuilder.create(AppRoute.Dispensary)
```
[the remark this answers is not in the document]
clemens-msupply: Think the most common use case is that you lookup a patient and then create or schedule an encounter for this patient.  You then realized you made a mistake and delete this encounter. Think then its more natural to go back to the patient's encounter view and not the global encounter view.

server/graphql/programs/src/queries/encounter.rs
clemens-msupply: Can you give an example please?

Currently, the DELETED option is in the status filter but it is overwritten when used (could remove it from the graphql definition...)

I would have to look into the different status filter options like `equalTo`, `equalAny`, etc. to see if a user requested the DELETED encounter and only if they weren't requested filter out the DELETED encounters. Is this what you mean?

client/packages/system/src/Encounter/DetailView/DetailView.tsx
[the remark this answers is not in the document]
clemens-msupply: how about navigate(-1)?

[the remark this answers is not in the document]
mark-prins: good point - I'm coming from a testing angle, which isn't the normal use case.
`navigate(-1)` should cover it though! good thought

server/graphql/programs/src/queries/encounter.rs
mark-prins: I was meaning that we could add a `Deleted` option to this filter:

image

though, as you say, it would require the DELETED status to be used in the query. 
No need to do anything, it was just a thought, of how we could implement the 'show me the list of deleted encounters'

mark-prins (approved):
👍  working beautifully! 🙏

clemens-msupply:
Thanks, I think there is some user feedback missing while the encounter is deleting so I tried to keep the modal open with a spinner for as long as the deletion is ongoing. However, I had some modal rerender issue which I tried to fix all morning. I gave up on it now... However, I fixed some infinit render in the encounter toolbar on the way...

PR #2662, opened 2023-12-27
2561 name merging on sync pull create name link record for name

### LINKED ISSUE #2561: Name merging: on sync pull create name_link record for name

During initialisation or regular synchronisation, `name` records will be inserted into omSupply as we integrate records. But as we roll out `name_link` throughout the code base, we'll need `name_link` records to be generated for each `name`.

Side affect of during insert of an `item`, create an `name_link` with `name_link.id == name_link.item_id == name.id`

### Moneyworks Jobcode 🧰

OMS:MERGEl

roxy-dao:
Fixes #2561 

# 👩🏻‍💻 What does this PR do? 
- Creates name_links for names at repo layer
- Fix postgres tests that are also broken in develop 👀 

- Run tests


server/repository/src/db_diesel/program_requisition/program_supplier.rs
```diff
@@ -300,11 +300,12 @@ mod test {
             .upsert_one(&name_store_join3)
             .unwrap();
 
-        let result = repo.query(&store3.id, filter);
+        let mut result = repo.query(&store3.id, filter).unwrap();
+        result.sort_by(|a, b| a.supplier.name_row.id.cmp(&b.supplier.name_row.id));
```
roxy-dao, line 304: Query doesn't sort programs, so need to sort for postgres tests to pass

server/repository/src/tests.rs
```diff
@@ -287,27 +287,17 @@ mod repository_test {
...
+        RequisitionLineRowRepository, RequisitionRepository, RequisitionRowRepository,
+        StockLineFilter, StockLineRepository, StockLineRowRepository, StorageConnection,
+        StoreRowRepository, UserAccountRowRepository,
     };
     use crate::{DateFilter, EqualFilter, StringFilter};
     use chrono::Duration;
     use diesel::{sql_query, sql_types::Text, RunQueryDsl};
     use util::inline_edit;
 
-    async fn insert_name_and_link(name: &NameRow, connection: &StorageConnection) -> () {
```
roxy-dao, line 301: Removed since new helper function in name_row does this anyway

server/service/src/sync/test/mod.rs
```diff
@@ -337,9 +337,6 @@ pub(crate) async fn check_records_against_database(
             UserPermission => {
                 check_delete_record_by_id!(UserPermissionRowRepository, con, id)
             }
-            Name => {
```
roxy-dao, line 340: Removing sync deletes for names since names should only be soft deleted... will need to implement soft deletes for name in both om and mSupply....

server/repository/src/tests.rs
Chris-Petty: Good call! I should probably do the same for `insert_item_and_link`

server/service/src/sync/test/mod.rs
Chris-Petty: mSupply is likely just going to continue doing hard deletes, too much would break (there isn't simply a service layer we can tweak) - though through the merge functionality (otherwise very rarely through the heavily restricted "delete" button). I was going to suggest that the central change log simply not update logs of `name` to `"delete"`, but then it might cause strife for the dashboard. Perhaps omSupply should just ignore deletes for `name`? @andreievg whaddya reckon?

Chris-Petty (approved):

Issue #923, opened 2022-12-08
Fatal error


mark-prins:
Not sure what happened - the dj demo site was not running, with only this hint in the log

```
2022-12-08 13:04:22.414718 ERROR omsupply_service::omsupply_service - panic occurred PanicInfo { payload: Any { .. }, message: Some(invalid time), location: Location { file: "C:\\Users\\Administrator\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\chrono-0.4.19\\src\\naive\\time.rs", line: 412, col: 9 }, can_unwind: true }  windows\src/windows.rs:66
```

have uploaded the postgres db and the datafile to [data share](https://drive.google.com/drive/folders/1tzCrTy65oZ_dm8SB2kbZX1N9q7YPCTw1?usp=share_link)

have stopped mSupply on the demo server, so that the demo runs. otherwise, it crashes on start up.

mark-prins:
So the issue is that `translate_sync_record` used naive_time to deserialise which panics if the time value supplied is invalid. This then takes the server down.

The problem record is an invoice, id `3cee66c2-359c-4d1f-9630-38f5136a7110` which shows an `entry_time` value of 2963575472 in the data in the sync_buffer. have edited the sync_buffer and set the value to 0 which has resolved the sync issue. 

Question is - how did that value get in there?

craigdrown:
https://docs.rs/chrono/0.2.18/chrono/naive/time/struct.NaiveTime.html#method.from_hms_opt instead?

PR #2670, opened 2024-01-03
886 add integration sync status progress

### LINKED ISSUE #886: Sync status integration progress
image

The last step here of `integration` could use a progress. The other steps say `n/total` records, integration surely could too! In my big data testing this is the slowest part... though perhaps I'm still getting loads of errors 😁.

fergie-nz:
Fixes #886 

# 👩🏻‍💻 What does this PR do? 
Adds logging and visual display for progress of integration step of syncing.

Note it only works on the initialising sync step - not syncing of already initialised setups.

This is because existing sync integration is run in a transaction.

You can test this by running branch [#886-Add](https://github.com/msupply-foundation/open-msupply/tree/886-testing-branch)
This adds an artificial delay to integration to simulate the visual logging progress in the front end.

The tests aren't working on this branch.
They also aren't working on develop for sync (for me)

I'm not sure if I am exacerbating things, or if sync testing requires specific setup.


server/service/src/sync/synchroniser.rs
```diff
@@ -183,16 +183,16 @@ impl Synchroniser {
...
         logger.done_step(SyncStep::PullRemote)?;
 
         // INTEGRATE RECORDS
         logger.start_step(SyncStep::Integrate)?;
-        //
-        let (upserts, deletes) =
-            integrate_and_translate_sync_buffer(&ctx.connection, is_initialised)
-                .map_err(SyncError::IntegrationError)?;
-        info!("Upsert Integration result: {:?}", upserts);
-        info!("Delete Integration result: {:?}", deletes);
```
fergie-nz, line 195: I assume these info flags are no longer necessary with logging added :)

mark-prins (changes requested):
Looking great thanks!

much nicer having progress - thanks for wading through the back end code to sort this one out..

I have a fix for the test, and a suggestion.
Also had a look at the linked issue (#923) - that still crashes the server and I'd hoped that we could just show an integration error. Unfortunately `serde_json::from_str` does not return an error if one of the field translations panics, the server will simply die. I went down the rabbit hole for a bit and stopped.

Also wanted to have a look at getting the stats when running in operational mode, but have spent long enough!

mark-prins:
if it's helpful - my changes are in [this branch](https://github.com/msupply-foundation/open-msupply/compare/886-testing-branch...886-testing-branch(integration))

mark-prins (approved):
👍  thanks for changes! looks great

andreievg (commented):
Looks like i haven't quite submitted a review I did yesterday, a few comments. I think we should really just be doing progress for integration only and only for upserts

server/service/src/sync/synchroniser.rs
```diff
@@ -235,20 +236,61 @@ pub fn integrate_and_translate_sync_buffer(
     > {
         let translators = all_translators();
         let table_order = pull_integration_order(&translators);
+        let step_progress = SyncStepProgress::Integrate;
 
         let sync_buffer = SyncBuffer::new(connection);
         let translation_and_integration = TranslationAndIntegration::new(connection, &sync_buffer);
         // Translate and integrate upserts (ordered by referential database constraints)
         let upsert_sync_buffer_records =
             sync_buffer.get_ordered_sync_buffer_records(SyncBufferAction::Upsert, &table_order)?;
```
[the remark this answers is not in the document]
andreievg: @fergie-nz 

As you've mentioned in telegram, this is actually not trivial:

https://github.com/msupply-foundation/open-msupply/compare/886-add-integration-sync-status-progress...886-add-integration-sync-status-progress(Example-of-option-mut)

server/service/src/sync/translation_and_integration.rs
```diff
@@ -160,7 +140,35 @@ impl<'a> TranslationAndIntegration<'a> {
                     );
                 }
             }
+
+            if true == initialisation {
```
Chris-Petty, line 144: Nit: initialisation is a boolean, ya don't need to compare it!

Nit within a nit: https://en.wikipedia.org/wiki/Yoda_conditions. That said there isn't really a disadvantage it just reads a bit funny in English, and arguably there _are_ advantages as stated on wikipedia! I didn't know WordPress actually states it as a convention!

andreievg (changes requested):
Just wanted to flag a point from my last comment as change request rather then nit. Thanks @Chris-Petty for mentioning that it's unclear if it's just a comment.

Basically as per the telegram chat, I would be happy to approve if this is implemented:
https://github.com/msupply-foundation/open-msupply/compare/886-add-integration-sync-status-progress...886-add-integration-sync-status-progress(Example-of-option-mut)(Simplified)

In summary:
* Only record integration progress for upserts during initialisation (don't record it during normal operations or for delete)
* Only show integration progress during initialisation (in UI), I haven't actually checked this bit

andreievg (approved):
Looks good, thanks @fergie-nz

PR #2676, opened 2024-01-08
Warn on dependencies for hooks

### LINKED ISSUE #2441: Set: 'react-hooks/exhaustive-deps': 'warn'
Make 'react-hooks/exhaustive-deps' a warning. Tried it out for some time now and find it very helpful.

Its actually relatively hard to validate some dependencies it complains about and thus would be good to have them not missing in the first place.

Surprisingly it helped more in detecting re-rendering issues rather than issue of not rendering correctly. For example, when fixing some of the warnings it results in infinite rendering loop which then points to some underlying issues of some other code which was hidden before.
It also highlights how some js functions are recreated constantly but then never been used in the hooks because the hooks don't depend on them...

`client/eslintrc.js`:

```diff
@@ -49,7 +49,7 @@ module.exports = {
       'always',
       { markers: ['#', '/'], exceptions: ['-'] },
     ],
-    'react-hooks/exhaustive-deps': 'off',
+    'react-hooks/exhaustive-deps': 'warn',
     '@typescript-eslint/no-explicit-any': 'warn',
   },
   ignorePatterns: ['**/operations.generated.ts', '**/types/schema.ts'],
```

mark-prins:
Fixes #2441 

# 👩🏻‍💻 What does this PR do? 
 
That's odd, I thought that I'd already done this 🤷 

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -49,7 +49,7 @@ module.exports = {
       'always',
       { markers: ['#', '/'], exceptions: ['-'] },
     ],
-    'react-hooks/exhaustive-deps': 'off',
+    'react-hooks/exhaustive-deps': 'warn',
     '@typescript-eslint/no-explicit-any': 'warn',
     'no-console': ['error', { allow: ['info', 'warn', 'error'] }],
   },


Chris-Petty:
Haven't tested but looks innocuous!

Chris-Petty (approved):

PR #2689, opened 2024-01-09
Rework the changelog_deduped view

### LINKED ISSUE #2688: Pushing remote records to the central server is slow
Pushing many remote records to the central server takes a long time and on a tablet it even renders the whole UI unusable.

The changelog_deduped view has been identified to be very slow to query. In a first step, this view needs to be optimized.

clemens-msupply:
Fixes #2688 

# 👩🏻‍💻 What does this PR do? 
Replaces the existing view:

```sql
CREATE VIEW changelog_deduped AS
            SELECT t1.cursor,
                t1.table_name,
                t1.record_id,
                t1.row_action,
                t1.name_id,
                t1.store_id,
                t1.is_sync_update
            FROM changelog t1
            WHERE t1.cursor = (SELECT max(t2.cursor) 
                            from changelog t2
                            where t2.record_id = t1.record_id)
            ORDER BY t1.cursor;
```

with 

```sql
CREATE VIEW changelog_deduped AS
                SELECT c.cursor,
                  c.table_name,
                  c.record_id,
                  c.row_action,
                  c.name_id,
                  c.store_id,
                  c.is_sync_update
                FROM (
                  SELECT record_id, MAX(cursor) AS max_cursor
                  FROM changelog
                  GROUP BY record_id
                ) grouped
                INNER JOIN changelog c
                  ON c.record_id = grouped.record_id AND c.cursor = grouped.max_cursor
                ORDER BY c.cursor;
```

The first view query was a problem for both, sqlite and postgres. Now the query seems to be fast so I haven't experimented with putting another index on record_id...

The problem was worse when there where lots of unsynced changelog entries. In the code there is always a clause like `WHERE cursor > 1000`. This speeds up the old query when you caught up with the changelog but didn't help much when there are lots of newer entries in the changelog... Now a full query of the view is fast but I haven't tested how far this scales up, e.g. 1 year of unsynced sensor data.

Check that both SQL queries give identicale results. (There is a test for the view which passes but a manual logic check would be good as well)

If you have a database with many changelog entries you can run both queries, e.g. in sqlite browser, and see the query time differences.

I used the programs import script to generate lots of remote records. Might be tricky to share this datafile but can do if needed, e.g. if you don't have an way to generate lots of sensor events...

### DIFF
--- a/server/repository/src/migrations/v1_06_00/changelog_deduped.rs
+++ b/server/repository/src/migrations/v1_06_00/changelog_deduped.rs
@@ -1,0 +1,32 @@
+use crate::{migrations::*, StorageConnection};
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    sql!(
+        connection,
+        r#"
+            DROP VIEW changelog_deduped;
+
+            -- View of the changelog that only contains the most recent changes to a row, i.e. previous row
+            -- edits are removed.
+            -- Note, an insert + delete will show up as an orphaned delete.
+            CREATE VIEW changelog_deduped AS
+                SELECT c.cursor,
+                  c.table_name,
+                  c.record_id,
+                  c.row_action,
+                  c.name_id,
+                  c.store_id,
+                  c.is_sync_update
+                FROM (
+                  SELECT record_id, MAX(cursor) AS max_cursor
+                  FROM changelog
+                  GROUP BY record_id
+                ) grouped
+                INNER JOIN changelog c
+                  ON c.record_id = grouped.record_id AND c.cursor = grouped.max_cursor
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
🙏  In my postgres cold chain database (128k changelog records) the original query takes +4s and regularly locks up, I now cannot even run an execution plan on it.

the new version is running at about 1.8s - more importantly it isn't locking up, so is a lot more reliable!

If however, I add 

```
CREATE INDEX changelog_record_id_idx ON changelog (record_id);
```

then the new query runs in at 232ms!
Would you mind adding the index too?

clemens-msupply:
Thanks! I also tried with a larger dataset in Postgres and the index improves the query time.

image

PR #2691, opened 2024-01-10
2608 add cancel buttons to modals

### LINKED ISSUE #2608: Cannot cancel the window while choosing supplier name when creating internal order
## What went wrong? 😲
 
On the tablet when we want to create a new order it shows a window to choose a supplier. If I did not find the name or if I want to cancel there is no way to do that. And every time we have to end up with creating dummy internal orders with the wrong supplier names. Which will cause many mistakes and internal orders to pile up on the list.
But on a desktop, you can close it by clicking outside the window.

## Expected behaviour 🤔
Should be able to close the window.

## How to Reproduce 🔨

1. In tablet to replenishment.
2. Create order.
3. Now you get the window with the supplier list.
4. Here you cannot close the window until you choose at least one supplier.

## Your environment 🌱

 - Version: 
 - Platform:
   - [ ] Android (tablet)

fergie-nz:
Fixes #2608

# 👩🏻‍💻 What does this PR do? 
Enables background click or tap closing for Internal Orders modal.

A number of other modals are closed by clicking outside (such as inbound shipment modal). This matches this pattern.

An alternative option would be to add a cancel button - though this may be more confusing? I'm not sure.

You can test this by going to
- Replenishment
- Internal Orders
- Create new
- Clicking outside (or tapping on table) to close the modal

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/ListView/CreateRequisitionModal.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/ListView/CreateRequisitionModal.tsx
@@ -1,6 +1,7 @@
 import React, { FC } from 'react';
 import {
   BasicSpinner,
+  DialogButton,
   ModalTabs,
   useDialog,
   useTranslation,
@@ -80,6 +81,7 @@ export const CreateRequisitionModal: FC<CreateRequisitionModalProps> = ({
       width={500}
       slideAnimation={false}
       title={t('label.new-requisition')}
+      cancelButton={<DialogButton variant="cancel" onClick={onClose} />}
     >
       <InnerComponent />
     </Modal>
--- a/client/packages/requisitions/src/RequestRequisition/ListView/CreateRequisitionModal.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/ListView/CreateRequisitionModal.tsx
@@ -1,7 +1,6 @@
 import React, { FC } from 'react';
 import {
   BasicSpinner,
-  DialogButton,
   ModalTabs,
   useDialog,
   useTranslation,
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
We originally had the background click-away enabled and there were complaints - leading to the current situation where you get stuck in modals! It turns out that people were prone to mis-clicking and closing modals accidentally. This is mostly a problem in the complex modals, such as editing lines in a shipment. Given that this is a simple modal and you can get back to the state you are in, I think the click-away is ok.

I do however worry that the affordance isn't there - you don't know that you can click outside the modal until you've done it. What would be great is having a button somewhere. It's worth checking to see how a cancel button would look ( I think a centred cancel would be ok ) and if that's too intrusive then a small (x) icon in top right, which is an icon button that also closes?

( I also thought that esc wasn't enabled and I use that to close modals too.. but that was just because the input field had focus )

fergie-nz:
> We originally had the background click-away enabled and there were complaints - leading to the current situation where you get stuck in modals! It turns out that people were prone to mis-clicking and closing modals accidentally. This is mostly a problem in the complex modals, such as editing lines in a shipment. Given that this is a simple modal and you can get back to the state you are in, I think the click-away is ok.
> 
> I do however worry that the affordance isn't there - you don't know that you can click outside the modal until you've done it. What would be great is having a button somewhere. It's worth checking to see how a cancel button would look ( I think a centred cancel would be ok ) and if that's too intrusive then a small (x) icon in top right, which is an icon button that also closes?
> 
> ( I also thought that esc wasn't enabled and I use that to close modals too.. but that was just because the input field had focus )

Nice, thankyou Mark.

I did have a play with adding a button, but had noticed that the modal for supplier selection (in the Inbound Shipment menu) uses a similar click background to close, and is used in a number of places.

Do you think it is worth adding a button in all cases?

I assumed that if the lack of button was causing confusion, it would have been flagged as an issue as the supplier selection modal is used in more places, but maybe worth refactoring them all to have a button.

mark-prins:
I'd like to have a consistent cancel button on all modals - just worried about the space it takes up and if it makes things look cluttered. will consider that one and will add an issue - thanks!

PR #2693, opened 2024-01-10
added comment column

### LINKED ISSUE #2678: Show breach comments in the breaches table
Coming from #2627.
We'd like to add another column to the table

mark-prins:
Fixes #2678

# 👩🏻‍💻 What does this PR do? 
 
Displays the comment column.
I thought perhaps to rename the column to 'Acknowledgement comment' but that felt clumsy, and then the name is also different to the modal where the field is called 'comment'. So it's just comment.

Best to piggy back on #2692 and test this after testing that.

## 📃 Documentation

It adds a new column, worth checking to see if this is visible in documentation screenshots and if the list of columns is in the docs, in which case add this column too please.

### DIFF
--- a/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/TemperatureBreachList.tsx
+++ b/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/TemperatureBreachList.tsx
@@ -182,6 +182,7 @@ const ListView: FC = () => {
         },
         sortable: false,
       },
+      ['comment', { sortable: false }],
     ],
     { onChangeSortBy: updateSortQuery, sortBy },
     [sortBy]


adamdewey:
@mark-prins - one more thing I was just thinking about, is it possible to store/display which user did the acknowledgement? 

That would save people trying to figure out who wrote the comment..

mark-prins:
> @mark-prins - one more thing I was just thinking about, is it possible to store/display which user did the acknowledgement?

the easy way would be to add the name to the comment prior to storing the comment - that way we don't need to make another modification to mSupply and the name is also visible in the comment column in the list view.

A more correct way to do this would be to store the user id in the breach table alongside the comment, that requires adding a new field to the mSupply table, so that the user can sync to central and be used when reinitialising a site. We'd also have to add a new column to the list view, one which is only really valid when the breach was acknowledged manually in omSupply.

adamdewey:
Coool.. hmm, I can't really think of much need for actually being able to do much with the user name data except for someone that wants to know 'Who wrote this? Who do I speak to about it?' when reviewing a comment. 

So the easy approach might be ok for now?? (he says with shifty eyes, looking out for the future Adam coming back through a Rick and Morty portal to get him later)

clemens-msupply (approved):
Work great! Haven't looked at the docs though

PR #2694, opened 2024-01-10
Move tab logic from useUrlQuery to DetailTabs

### LINKED ISSUE #2431: Remove "tab" specifc code from the `common`  `useUrlQuery` method
In https://github.com/openmsupply/open-msupply/pull/2412 there was a change to reset the tab url queries when changing a tab.

In `packages/common/src/hooks/useUrlQuery/useUrlQuery.ts` there is a block:

```ts
     else if (key === 'tab') {
        newQueryObject[key] = String(value);
        Object.entries(newQueryObject).forEach(([k, v]) => {
          if (v !== value) delete newQueryObject[k];
        });
      } 
```

However, as a `common` hook it shouldn't really make assumptions about application specific conventions about how tabs are handled but just manage the query params.

Possible options:
1) Remove this code and handle the url reset when changing a tab, e.g. where `<TabContext value={currentTab}>` is used
2) Remember which tab had which url queries params (this functionality was also needed for the patient programs tabs where currently the query params are lost when changing tabs)
  For example, we could have a new hook that persist the tab url params in a `zustand` store or in the local storage.

Note, why is resetting the url params working for the patient tabs?

clemens-msupply:
Fixes #2431 

# 👩🏻‍💻 What does this PR do? 
- Removes the tab specific code from useUrlQuery and moves it to DetailTabs (tab query is set when changing the tab)
- Remembers query parameters for each tab (only for as long as being on the tab view)

For the second point, is this actually something we want to have or should I revert to the old behaviour where all filter params are reset when changing between tabs?

Best to be tested with a cold chain dataset (Monitoring tabs)
https://github.com/msupply-foundation/open-msupply/files/13303292/oms_cold_chain.zip

- Set filter for each tab and move back and forth between tabs

### DIFF
--- a/client/packages/common/src/hooks/useUrlQuery/useUrlQuery.ts
+++ b/client/packages/common/src/hooks/useUrlQuery/useUrlQuery.ts
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { useSearchParams } from 'react-router-dom';
 
 export interface UrlQueryObject {
@@ -51,7 +52,7 @@ interface useUrlQueryProps {
   skipParse?: string[];
 }
 
-export const useUrlQuery = ({ skipParse = [] }: useUrlQueryProps = {}) => {
+export const useUrlQuery = ({ skipParse }: useUrlQueryProps = {}) => {
   const [searchParams, setSearchParams] = useSearchParams();
 
   const updateQuery = (values: UrlQueryObject, overwrite = false) => {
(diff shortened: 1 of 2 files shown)


client/packages/common/src/hooks/useUrlQuery/useUrlQuery.ts
```diff
@@ -89,11 +85,14 @@ export const useUrlQuery = ({ skipParse = [] }: useUrlQueryProps = {}) => {
...
-    updateQuery,
-    parseRangeString,
-  };
+  return useMemo(
+    () => ({
+      urlQuery: parseSearchParams(searchParams, skipParse ?? []),
+      updateQuery,
+      parseRangeString,
+    }),
+    [searchParams, skipParse]
```
clemens-msupply, line 94: This useMemo was needed to avoid a render loop in the DetailTabs (useEffect that depends on the urlQuery)...

CarlosNZ (approved):
Yeah, that all looks good and works nicely.

> For the second point, is this actually something we want to have or should I revert to the old behaviour where all filter params are reset when changing between tabs?

I think it makes sense to keep them. If we come across a use case where it's preferable to reset them, then we can add an additional property to `tabDefinitions` to specify that they should be reset. But not worth doing until we encounter such a case (which may well be never) :)

PR #2697, opened 2024-01-11
Copy fridge tag file if rename failed

### LINKED ISSUE #2696: Moving fridge tag upload file fails when temp and app_data dir are on different devices/partitions
## What went wrong? 😲

From: https://doc.rust-lang.org/std/fs/fn.rename.html

```
Rename a file or directory to a new name, replacing the original file if to already exists.

This will not work if the new name is on a different mount point.
```

This was the case for my setup ("Invalid cross-device link")

Solution: detect failure and try to copy the file instead.

clemens-msupply:
Fixes #2696 

# 👩🏻‍💻 What does this PR do? 
Handles (valid) rename error case.

Have the feeling there must be a method or crate for it but couldn't find it and it was easy to implement...

Fridge tag upload should still work when upload temp dir and app_data directory are on the same mountpoint/device.

### DIFF
--- a/server/server/src/upload_fridge_tag.rs
+++ b/server/server/src/upload_fridge_tag.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use actix_multipart::form::{tempfile::TempFile, MultipartForm};
 use actix_web::{
     post,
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
⭐  very nice!

PR #2701, opened 2024-01-12
2681 merge program event patient

### LINKED ISSUE #2681: Name merging: program_event.patient_id

- [ ] Add `program_event.patient_link_id` and migrate with `program_event.patient_link_id = program_event.patient_id`
- [ ] Remove `program_event.patient_id`
- [ ] Update any views that relied on relations of `program_event.patient_id`
- [ ] Update `program_event` at the repository layer to return related `name` records based on `program_event.patient_link_id` rather than `program_event.patient_id`

### Moneyworks Jobcode 🧰

OMS:MERGE

clemens-msupply:
Fixes #2681

# 👩🏻‍💻 What does this PR do? 
Uses name_link for program events


server/service/src/programs/patient/patient_updated.rs
```diff
@@ -49,12 +48,6 @@ pub fn update_patient_row(
     let name_upsert = patient_to_name_row(store_id, update_timestamp, patient, existing_name)?;
     name_repo.upsert_one(&name_upsert)?;
 
-    let name_link_row = NameLinkRow {
-        id: name_upsert.id.clone(),
-        name_id: name_upsert.id.clone(),
-    };
-    NameLinkRowRepository::new(con).upsert_one(&name_link_row)?;
```
clemens-msupply, line 56: I sneaked this in from:
https://github.com/msupply-foundation/open-msupply/commit/fea8cefaabd5b0d50d78b5bf01863d8520f56dc3#r136857975

server/repository/src/db_diesel/program_event.rs
```diff
@@ -192,15 +222,103 @@ impl<'a> ProgramEventRepository<'a> {
...
+    use crate::{
+        mock::{context_program_a, mock_patient, mock_patient_b, MockDataInserts},
+        test_db::setup_all,
+        EqualFilter, Pagination, ProgramEventFilter, ProgramEventRepository, ProgramEventRow,
+        ProgramEventRowRepository,
+    };
+
+    #[actix_rt::test]
+    async fn program_event_delete() {
+        let (_, connection, _, _) = setup_all("program_event_delete", MockDataInserts::all()).await;
```
clemens-msupply, line 266: I added this test because the delete filter had to change a bit...

server/repository/src/migrations/v1_06_00/program_event_patient_link_id.rs
```diff
@@ -0,0 +1,38 @@
...
+        ALTER TABLE program_event
+        ADD COLUMN patient_link_id TEXT NOT NULL DEFAULT 'temp_for_migration' REFERENCES name_link(id); -- Can't have NOT NULL without a default... no sqlite PRAGMA for turning constraints off!
+
+        UPDATE program_event
+        SET patient_link_id = patient_id;
+
+        PRAGMA foreign_keys = ON;
+     "#,
+    )?;
+
```
Chris-Petty, line 36: Not dropping the old `patient_id` and indexing `patient_link_id`?

server/service/src/programs/program_event.rs
```diff
@@ -104,6 +104,7 @@ fn remove_event_stack(
     //      5---->20--|
     let mut current_end_datetime = longest.active_end_datetime;
     for mut current in previous_stack {
+        let current = &mut current.program_event_row;
```
Chris-Petty, line 107: Nit/syntax Q: Could you destructure this in the for loop var?

Chris-Petty (changes requested):

server/graphql/types/src/types/program/program_event.rs
```diff
@@ -64,18 +64,28 @@ impl ProgramEventSortInput {
...
 impl ProgramEventNode {
-    pub async fn patient_id(&self) -> &Option<String> {
-        &self.row.patient_id
+    pub async fn patient_id(&self) -> Option<String> {
+        self.name_row().as_ref().map(|it| it.id.clone())
     }
 
     pub async fn patient(&self, ctx: &Context<'_>) -> Result<Option<PatientNode>> {
-        let Some(patient_id) = &self.row.patient_id else {
+        let Some(patient_id) = self.name_row().as_ref().map(|it| &it.id) else {
```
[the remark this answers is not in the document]
clemens-msupply: Before it was just a `patient_id: Option<String>` and now its a `name: Option<NameRow>`. In Typescript you would do `name?.id`, i.e. check if name is Some and then get `name.id`. 

In Rust there is `Option::map()` to check if the Option is Some and if this is the case it takes the Option value and you can transform it to another value. For example, using `name.map(|it| it.id)` `Option<NameRow>` becomes `Option<String>`. You then need as_ref because you don't want to consume the NameRow in the option (because we only have a reference to the option in the first place...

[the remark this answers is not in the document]
clemens-msupply: > Also might be easier to explain in person but why does `patient_link_id` implementation require doing this in the gql layer? We are generally avoiding touching anything but the repo layer and the odd test that breaks.

Not quite sure what you mean, I avoid accessing the patient_link_id by using the NameRow which has been joined in the repo layer. Yeah might be easier to discuss in person.

Chris-Petty:
@clemens-msupply waddup

Also conflicts now if you wouldn't mind resolving

server/service/src/programs/program_event.rs
clemens-msupply: I updated the `previous_stack` Vec to match what it was before...

server/repository/src/migrations/v1_06_00/program_event_patient_link_id.rs
clemens-msupply: Thanks this was trickier than expected! For some unknown reason I used a "proper" FK constraint in sqlite which can't be deleted! see change...

Chris-Petty: Aaah yea I see, annoying! Likely no where with that many program events so not a big deal

Chris-Petty (approved):
Sweet!

PR #2702, opened 2024-01-12
Start some merging docs

clemens-msupply:
My current understanding of how merging and link tables work, so this might be a bit wrong...

### DIFF
--- a/server/repository/src/merging.md
+++ b/server/repository/src/merging.md
@@ -64,7 +64,7 @@ and all three links point to the same `"kept_name_id"`.
 
 <!-- Links are repo only -->
 
-In the code reference to the link tables should be confined to the repository layer.
+In the code base, references to the link tables should be confined to the repository layer.
 For example, querying an entity in the service layer must be done by `name_id` rather than by `name_link_id`.
 Internally, the repository looks up all `name_links` which match the `name_id` and then uses these `name_links` to do the query.
 Note, if a merge happened while the service layer still uses the id of the soft deleted entity, a query using this id will return nothing because there will be no links pointing to this old id.
--- a/server/repository/src/merging.md
+++ b/server/repository/src/merging.md
@@ -1,0 +1,76 @@
+# Merging of Entities (e.g. Names or Items)
+
+Two or more entities such as items or names may be merged after they have been created, e.g. because there are duplicates.
+When merging two entities one entity is soft deleted while the second entity is stays in the system and is not considered the "merged" entity.
+Merging multiple entities can be achieved by recursively merging two entities till all entities are merged.
+
+<!-- Links -->
+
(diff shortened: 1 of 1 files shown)


server/repository/src/merging.md
```diff
@@ -0,0 +1,76 @@
...
+NameLink {id: "old_id", name_id: "kept_name_id"}
+NameLink {id: "previously_merged", name_id: "kept_name_id"}
+NameLink {id: "kept_name_id", name_id: "kept_name_id"}
+```
+
+and all three links point to the same `"kept_name_id"`.
+
+<!-- Links are repo only -->
+
+In the code reference to the link tables should be confined to the repository layer.
```
Chris-Petty, line 67:
```suggestion
In the code base, references to the link tables should be confined to the repository layer.
```

server/repository/src/merging.md
```diff
@@ -0,0 +1,76 @@
...
+For example, querying an entity in the service layer must be done by `name_id` rather than by `name_link_id`.
+Internally, the repository looks up all `name_links` which match the `name_id` and then uses these `name_links` to do the query.
+Note, if a merge happened while the service layer still uses the id of the soft deleted entity, a query using this id will return nothing because there will be no links pointing to this old id.
+
+<!-- However, updating a name_id is done using the row.name_link_id -->
+
+When updating an entity of a data row the link fields of the repository objects are used, e.g. `row.name_link_id = "New name id"`.
+This works because current entity has a matching link entry with the same id, e.g `name_link.id = name_id`.
+While this can be fully hidden in the repository layer this would require another `upsert` data structure which contains the entity id instead of the link id.
```
Chris-Petty, line 76: I think an example here would be good, I'm having trouble conceptualising the point of this section without one and I wrote half the code heh. An outbound shipment probably the most easy case. Maybe if I'm not too unhinged:

```
For example, when creating an outbound shipment you should use `invoice.name_link_id = customer_name.id`. `customer_name.id` is guaranteed to have a corresponding `name_link_id`. For existing shipments you should **not** update `invoice.name_link_id` to the returned related name entity's `name.id`, even if it may be different. Any difference is due to a merge having occurred, and updating the `name_link_id` would undermine a design goal of potentially being able to "unmerge" entities. 
```

Chris-Petty (changes requested):
Thanks! This is an awesome start to documenting this, if not 90% there! Just requesting changes for a suggested example for the last section.

The comments defining sections could possibly be sub headers rather than hidden markers.

server/repository/src/merging.md
clemens-msupply: Thanks added this and rearranged it slightly

Chris-Petty (approved):

PR #2706, opened 2024-01-12
Change the default dates for temperature chart

### LINKED ISSUE #2700: Cold chain: Change graph to display 24 hours from current time backwards

The cold chain graph is now displaying 24 hours 🥳 

But it seems to be from the end of the day backwards. 

image

It would be more useful if we could display from the current time backwards

This would maximise the data the user could see and also give them an indication of whether they're receiving recent data or not

Display 24 hours from current time backwards.

mark-prins:
Fixes #2700 

# 👩🏻‍💻 What does this PR do? 
 
Defaults the filter dates and times for the charts, and updates the max date values for the filters.

View the chart! by default you'll see entries going back in time (for one day) from now

### DIFF
--- a/client/packages/coldchain/src/Monitoring/ListView/TemperatureLog/Toolbar.tsx
+++ b/client/packages/coldchain/src/Monitoring/ListView/TemperatureLog/Toolbar.tsx
@@ -44,7 +44,7 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
                   name: t('label.from-datetime'),
                   urlParameter: 'datetime',
                   range: 'from',
-                  maxDate: new Date(),
+                  maxDate: DateUtils.addMinutes(new Date(), -30),
                   isDefault: true,
                 },
                 {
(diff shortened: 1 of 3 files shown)


client/packages/coldchain/src/Monitoring/ListView/TemperatureLog/Toolbar.tsx
```diff
@@ -44,15 +44,15 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
                   name: t('label.from-datetime'),
                   urlParameter: 'datetime',
                   range: 'from',
-                  maxDate: new Date(),
+                  maxDate: DateUtils.addMinutes(new Date(), -30),
```
fergie-nz, line 47: I wonder if we would ever work with countries that cross multiple timezones - and they WOULD want to look at times in the 'future'

fergie-nz (approved):
Nice Mark looks great. 

I've noticed a few other funny behaviours with date selection which I'll add as other issues.

client/packages/coldchain/src/Monitoring/ListView/TemperatureLog/Toolbar.tsx
mark-prins: oh, good point! this is where being conservative could come back to bite me!

Issue #2713, opened 2024-01-14
Cold chain: Can't select both 'From' and 'To' dates in date filter


adamdewey:
## What went wrong? 😲

You can only select either a 'From' date or a 'To' date, not both: 

https://github.com/msupply-foundation/open-msupply/assets/19318937/7528f5e6-c0f3-444c-9576-22a136161dc0

## Expected behaviour 🤔

User is able to select a date range by selecting both From and To dates

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Open Cold Chain graph
2. Select a From date
3. Select a To date

## Your environment 🌱

 - Version: V1.6.00
 - Platform:
   - [x] android (tablet)
   - [ ] browser (extra points if you tell us which one)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

roxy-dao:
Closed in #2723

marthakeezy:
Tested in v1.6.02 - test is passed. 

Checked Chart & Log tabs, clicked out of date & time picker and the dates remained entered. PASS ✅

PR #2723, opened 2024-01-15
Parse dates in date filter

### LINKED ISSUE #2712: Cold chain: Touching screen causes date filters to reset
## What went wrong? 😲

Touching the screen removes any date filters - meaning you can't touch breach icons to show more details. 

https://github.com/msupply-foundation/open-msupply/assets/19318937/cb662294-cf26-40a4-bf36-6af3eb28e341

## Expected behaviour 🤔

Date filters shouldn't change unless user changes them deliberately. 

## How to Reproduce 🔨

1. Select a date filter
2. Touch anywhere on graph

## Your environment 🌱

 - Version: V1.6.00
 - Platform:
   - [x] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

mark-prins:
Fixes #2712 #2713

# 👩🏻‍💻 What does this PR do? 
 
The DateTimePickerInput was failing to parse the date - the default format of `P p` is used to both format a date object to a string and to parse the date string.

However, the parse was failing silently with the result that the input was setting the date value to `null` on a blur event, resetting the input value.

The fix, for me at least, is to provide a locale to the `parse` call.
An alternative is to set the date format at `dd/MM/yyyy HH:mm` as this works for parsing as well. However, I thought it best to use the local regional format if possible. I've tried setting my OS regional settings to something different, however this has no effect.

As per the issues, focus on a date input and then click elsewhere, triggering a blur event.
If the input is populated with a date, you should see no effect i.e. the current date value remains.

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -83,13 +83,14 @@ export const DateUtils = {
   addCurrentTime,
   getDateOrNull: (
     date?: Date | string | null,
-    format?: string
+    format?: string,
+    options?: Parameters<typeof parse>[3]
   ): Date | null => {
     if (!date) return null;
     if (date instanceof Date) return date;
     const maybeDate =
       format && typeof date === 'string'
-        ? parse(date, format, new Date())
+        ? parse(date, format, new Date(), options)
         : new Date(date);
     return isValid(maybeDate) ? maybeDate : null;
   },
--- a/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -130,7 +130,7 @@ export const DateTimePickerInput: FC<
       minDate={minDate}
       maxDate={maxDate}
       {...props}
-      value={DateUtils.getDateOrNull(props.value, undefined, dateParseOptions)}
+      value={props.value}
     />
   );
 };
--- a/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -3,7 +3,7 @@ import { DateTimePicker, DateTimePickerProps } from '@mui/x-date-pickers';
 import { BasicTextInput } from '../../TextInput/BasicTextInput';
 import { useAppTheme } from '@common/styles';
 import { StandardTextFieldProps, TextFieldProps } from '@mui/material';
-import { DateUtils, useTranslation } from '@common/intl';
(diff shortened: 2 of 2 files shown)


client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
```diff
@@ -126,7 +130,7 @@ export const DateTimePickerInput: FC<
       minDate={minDate}
       maxDate={maxDate}
       {...props}
-      value={DateUtils.getDateOrNull(props.value)}
+      value={DateUtils.getDateOrNull(props.value, undefined, dateParseOptions)}
```
CarlosNZ, line 133: Strictly speaking, you probably don't need this one, as the `value` should always be a Date object (not a string) -- it gets converted on input (onBlur), or is already a Date if using the Picker.

client/packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -83,13 +83,14 @@ export const DateUtils = {
   addCurrentTime,
   getDateOrNull: (
     date?: Date | string | null,
-    format?: string
+    format?: string,
+    options?: Parameters<typeof parse>[3]
```
CarlosNZ, line 87: Might be overkill, but I wonder if we should tighten the type definitions here to ensure that if a string date is provided, then `format` and `options` (with `locale`) must also be provided (but optional if date is `Date`).

Might make the input type ridiculously complicated though, so only if you think its worth it.

CarlosNZ (approved):
Looks good thanks. I think this is the correct way to handle it. The `onBlur` event is intended for a string input, in which case a format and locale should be required to interpret it correctly.

client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
mark-prins: yes, good point. I'll just remove the `getDateOrNull` then and see if that causes any problems..

client/packages/common/src/intl/utils/DateUtils.ts
mark-prins: the `options` is only required if you have a format which uses locale specific items (such as `P` or `p` in this case) so I wouldn't want to make that required.
I like the idea, so had a look - and can make the type of `format` depend on the `date` type, something like:

```
  getDateOrNull: <T extends Date | string | null | undefined>(
    date: T,
    format: T extends string ? string : string | undefined,
    options?: Parameters<typeof parse>[3]
  ): Date | null => {
```

but.. couldn't see how to have format as optional, only how to have the type change i.e. `string | undefined`. the format paramater still has to be specified, and that results in lots of code changes to add in `getDateOrNull(date, undefined)` which isn't great.

PR #2725, opened 2024-01-15
2636 requisition_line.item_id translation from item_link

### LINKED ISSUE #2636: Item merging: requisition_line translator push correct item_id

Same as https://github.com/msupply-foundation/open-msupply/pull/2634 but for requisition_line

In the translator use the _domain_ repo instead of the _row_ repo.

### Moneyworks Jobcode 🧰

OMS:MERGE

Chris-Petty:
Fixes #2636 

# 👩🏻‍💻 What does this PR do? 

Changes `requisition_line` translation to look up item_id from the related `ItemLinkRow` instead of using it directly (because it's really `item_link_id` at the moment).

Automated Tests

### DIFF
--- a/server/service/src/sync/translations/requisition_line.rs
+++ b/server/service/src/sync/translations/requisition_line.rs
@@ -4,7 +4,7 @@ use crate::sync::{
 };
 use chrono::NaiveDateTime;
 use repository::{
-    ChangelogRow, ChangelogTableName, ItemRowRepository, RequisitionLineRow,
+    ChangelogRow, ChangelogTableName, ItemLinkRowRepository, ItemRowRepository, RequisitionLineRow,
     RequisitionLineRowRepository, StorageConnection, SyncBufferRow,
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


server/service/src/sync/translations/requisition_line.rs
```diff
@@ -216,4 +227,45 @@ mod tests {
             assert_eq!(translation_result, record.translated_record);
         }
     }
+
+    #[actix_rt::test]
+    async fn test_requisition_line_push_merged() {
+        // The item_links_merged function will merge ALL items into item_a, so all stock_lines should have an item_id of "item_a" regardless of their original item_id.
+        let (_, connection, _, _) = setup_all(
```
clemens-msupply, line 234: Nit: looks a bit fragile too me, e.g. if you add more mock data that will break this test. But I might be wrong...

clemens-msupply (approved):
Looks good. As we discussed we should rename the column to item_link_id so that we catch these cases. Wonder if it really affect that many files.

server/service/src/sync/translations/requisition_line.rs
Chris-Petty: So long as no one removes item_a then the test should work. Is a kinda wild approach though haha.

PR #2726, opened 2024-01-15
Replace manufacturer

### LINKED ISSUE #2717: Cold chain: Change sensor type to 'mSupply RTMD'

The product that is registered with WHO PQS is our own RTMD, so it would be good to update the Sensor type here to say `mSupply RTMD`: 

image

mark-prins:
Fixes #2717

# 👩🏻‍💻 What does this PR do? 
 
The replacement to `mSupply RTMD` I've made a front end substitution - this is because the concern was a f/e display issue, and the current method of formatting (`Formatter.enumCase`) would not provide the correct result (mSupply). I thought that by using a translation, we could (a) display in a local language if required and (b) update when the PM whimsy strikes.

The serial replacement however, has been done in the GQL query so that it is handled for both the sensor detail modal and the list view:

Screenshot 2024-01-15 at 5 58 42 PM

(the list is slightly visible behind the modal)

Check the sensor list and detail views

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/coldchain/src/Sensor/ListView/ListView.tsx
+++ b/client/packages/coldchain/src/Sensor/ListView/ListView.tsx
@@ -10,6 +10,7 @@ import {
   Formatter,
   useEditModal,
   useUrlQuery,
+  SensorNodeType,
 } from '@openmsupply-client/common';
 import { useSensor, SensorFragment } from '../api';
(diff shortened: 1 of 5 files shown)


roxy-dao (approved):
LGTM~

PR #2730, opened 2024-01-15
2682 Name merging: stock_line.supplier_id

### LINKED ISSUE #2682: Name merging: stock_line.supplier_id

- [ ] Add `stock_line.supplier_link_id` and migrate with `stock_line.supplier_link_id = stock_line.supplier_id`
- [ ] Remove `stock_line.supplier_id`
- [ ] Update any views that relied on relations of `stock_line.supplier_id`
- [ ] Update `stock_line` at the repository layer to return related `name` records based on `stock_line.supplier_link_id` rather than `stock_line.supplier_id`

### Moneyworks Jobcode 🧰

OMS:MERGE

roxy-dao:
Fixes #2682

# 👩🏻‍💻 What does this PR do? 
Supplier link id for stock line

Run tests

### DIFF
--- a/server/repository/src/migrations/v1_06_00/stock_line_add_supplier_link_id.rs
+++ b/server/repository/src/migrations/v1_06_00/stock_line_add_supplier_link_id.rs
@@ -1,0 +1,44 @@
+use crate::{migrations::sql, StorageConnection};
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    #[cfg(feature = "postgres")]
+    sql!(
+        connection,
+        r#"
+        -- Adding stock_line.supplier_link_id
+        ALTER TABLE stock_line
+        ADD COLUMN supplier_link_id TEXT;
+        
+        UPDATE stock_line
+        SET supplier_link_id = supplier_id;
+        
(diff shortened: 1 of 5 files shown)


server/repository/src/migrations/v1_06_00/stock_line_add_supplier_link_id.rs
```diff
@@ -0,0 +1,44 @@
...
+        UPDATE stock_line
+        SET supplier_link_id = supplier_id;
+        PRAGMA foreign_keys = ON;
+     "#,
+    )?;
+
+    sql! {
+        connection,
+        r#"
+        DROP INDEX index_stock_line_supplier_id;
```
Chris-Petty, line 37: I tweaked this just removing the IF EXISTS, think it's straight forward enough I'll do it in the other PRs that I requested the same. Test pass!

That said, I see that at least for this one the index is created in 1_06_00 migration as well (maybe why you changed the order of where the indexing migration is in 1_06_00). Maybe we should just remove it rather than drop it straight after creating it.

Chris-Petty (approved):

server/repository/src/migrations/v1_06_00/stock_line_add_supplier_link_id.rs
roxy-dao: If the merging goes into 1.6 then we can... but looks like they'll be going to 1.7 at this rate?

PR #2738, opened 2024-01-15
Ignore acknowledge status from CCA

### LINKED ISSUE #2704: Cold chain: Breaches acknowledged in CCA bypass breach acknowledgement workflow in Open mSupply
## What went wrong? 😲

In Open mSupply the users must enter a comment before acknowledging a breach. 
This means there is a record for other users to understand what happened and what action was taken. 

However there is a way to bypass this: by acknowledging the breaches on CCA which causes breaches to be marked as acknowledged in Open mSupply, but without a comment. 

This bypasses the workflow that we want to enforce - i.e. that users need to add a comment to explain why they acknowledged the breach. 

## Expected behaviour 🤔

Breaches acknowledged in CCA should only affect CCA. 
The breach should still need to be acknowledged in Open mSupply, and a comment added as normal. 

Perhaps a clarification would be to change the terminology in CCA to 'clearing' the breach visual alert, rather than acknowledging it? 

I did muse about adding the ability to add a comment in CCA, but there would be no way to enforce the capturing of which user entered the comment. 

Additionally it wouldn't be possible for sensors such as Fridge-tags or others, so I think we should disregard this option. 

## How to Reproduce 🔨

1. Produce a breach with CCA
2. Note that the breach appears as unacknowledged in Open mSupply
3. Acknowledge the breach in CCA
4. Note that the breach appears as acknowledged in Open mSupply (without a comment)

## Your environment 🌱

 - Version: V1.6.00
 - Platform:
   - [x] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

mark-prins:
Fixes #2704

# 👩🏻‍💻 What does this PR do? 
 
Ignore the breach acknowledgement status sent by the CCA via the API

Send a breach using the API
- new breaches: unacknowledged is always true
- updated breaches: unacknowledged remains as it was
- the comment entered by a user won't be overwritten by the CCA API call either

here's my postman collection which has the login and breach creation calls:

[Cold Chain.postman_collection.json](https://github.com/msupply-foundation/open-msupply/files/13943252/Cold.Chain.postman_collection.json)

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/server/src/cold_chain/temperature_breach.rs
+++ b/server/server/src/cold_chain/temperature_breach.rs
@@ -24,7 +24,7 @@ use super::validate_request;
 #[serde(rename_all = "camelCase")]
 pub struct TemperatureBreach {
     id: String,
-    acknowledged: bool,
+    // acknowledged: bool,
     #[serde(rename = "endTimestamp")]
     end_unix_timestamp: Option<i64>,
     sensor_id: String,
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
LGTM~ Thanks Mark!

Issue #2711, opened 2024-01-14
Cold chain: Multiple breach icons displayed for a single breach


adamdewey:
## What went wrong? 😲

Multiple breach icons are sometimes shown on the graph for a singular breach event: 

https://github.com/msupply-foundation/open-msupply/assets/19318937/57983462-bce1-460d-b6f9-1513e18af407

## Expected behaviour 🤔

Only 1 breach icon should be shown for a single breach event - at the point where the breach starts. 

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Take datafile from vaccine demo server
2. Open data on a tablet (not sure if reproducible on desktop?)

## Your environment 🌱

 - Version: V1.6.00
 - Platform:
   - [x] android (tablet)
   - [ ] browser (extra points if you tell us which one)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

roxy-dao:
Closed in #2741

nisha-dangol:
Tested in v1.6.2 exe

Unlike the original, I see that there is only single breach icon showing for a single breach; so simply not replicable anymore; so moving to Done!

PR #2741, opened 2024-01-16
Show the breach temperature as the point temperature

### LINKED ISSUE #2708: Cold chain: Graph does not represent temperature data accurately - max / min temperature data points not shown
## What went wrong? 😲

Due to data averaging the graph does not accurately reflect the temperature history of the sensors. 

For example here you can see that there has been a breach, but the graph doesn't show the line ever crossing the 8 °C threshold: 

image

You also lose some detail with the data smoothing. 

## Expected behaviour 🤔

Ideally no averaging would be applied and all data points would be shown. 

For example as demonstrated by the Grafana graph of the same data here: 

image

## How to Reproduce 🔨

1. Compare Open mSupply graphs with Grafana graphs

## Your environment 🌱

 - Version: V1.6.00
 - Platform:
   - [x] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

mark-prins:
Fixes #2708

# 👩🏻‍💻 What does this PR do? 
 
Iterate over the chart data and for points that have a breach associated, use the breach max_or_min temperature instead of the averaged temperature value.

This allows breach indicators to show the 'correct' value i.e. the value which you see when viewing the breach. If there isn't a breach, then the averaging behaves as normal.

I also tried increasing the number of datapoints - going up to 300 ( 24h of data, with 5 min logging interval = 288 data points ) displays the breaches correctly using the example data given.. but is incredibly slow. too slow to use.

So, I went with option 2. This is fast enough, and has minimal impact. Looks even better when combined with the fix for #2711 as well:

image

You might find it useful to test with the omsupply_vaccine.zip sqlite file which is in the [google drive](https://drive.google.com/open?id=1e7475Z9J-j1KnferOQOrcYSSTnFp9wtk&usp=drive_fs)

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/graphql/general/src/queries/temperature_chart.rs
+++ b/server/graphql/general/src/queries/temperature_chart.rs
@@ -8,7 +8,8 @@ use graphql_core::{
     ContextExt,
 };
 use graphql_types::types::{SensorNode, TemperatureLogFilterInput};
-use repository::{TemperatureChartRow, TemperatureLogFilter};
+use repository::{StorageConnection, TemperatureChartRow, TemperatureLogFilter};
+use service::temperature_breach::query::get_max_or_min_breach_temperature;
 use service::{
     auth::{Resource, ResourceAccessRequest},
     temperature_chart::{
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
Looks good and makes sense.

Apologies Mark, as for your other PR, I can't seem to get the datafile working for testing to actually see how it looks, but it looks good.

I am wondering about 2 cases where we might need to make changes in future, or in which the chart would be misleading:

If there is the type of breach where the high or low value is sustained (but not enough to trigger a breach condition for any individual measurement), there could be a case where the max or min function would still give a value which is below or above the breach threshold despite it meeting the breach condition. So given the right timeseries averaging, this could still result in a breach warning within or close to the normal range. A more rare case but possibly still confusing.

I also wonder as a general case if we should be showing some measure of variance if we are averaging measurements for graph displays. A trend which oscillates wildly on a micro scale but is averaged out to appear static could be very misleading. I wonder if having a warning next to the graph if there is a high variance would be a good prompt for users to dig deeper. This wouldn't have to be anything too math heavy to make it user friendly - maybe a simple warning that "temperature fluctuations are high for this sensor, and risk of breach is high" or something like that.

fergie-nz:
> If there is the type of breach where the high or low value is sustained (but not enough to trigger a breach condition for any individual measurement), there could be a case where the max or min function would still give a value which is below or above the breach threshold despite it meeting the breach condition. So given the right timeseries averaging, this could still result in a breach warning within or close to the normal range. A more rare case but possibly still confusing.

I seem to remember that there were two cases for a breach:

1 being a high or low enough value to trigger an immediate breach

2 being sustained low or high above or below a tighter band will also cause a breach.

But I'm not sure, and now I wonder if number 2 is all in my head. That above comment only makes sense in the context of number 2 :)

mark-prins:
thanks - yes, the averaging of the data points could hide variations. I noticed on my domestic fridge the temperature is a lot more variable than I expected. Not an issue as long as it stays within the acceptable range though. might be good to have an indication of variability in the temperature? I like that idea.

For breaches - in the CCA there is the concept of cumulative breach, which has a different threshold to the consecutive. the issue is primarily how to display multiple thresholds on the chart.. compounded by displaying multiple sensors, which CCA doesn't have to cope with. The chart was conceived as an approximation - a quick visual indicator.. but we may have to reconsider, now that it's proving popular!

PR #2753, opened 2024-01-17
Name merging invoice.name id translator get actual name.id

### LINKED ISSUE #2745: Name merging: invoice.name_id translator get actual name.id

name merging needs to make sure that when records with FK field `name_link_id` or similar sync, they correctly send the `name_id` of the linked name as `name_id`, not `name_link_id` as `name_id`. e.g.

https://github.com/msupply-foundation/open-msupply/blob/e4de7353c761bb836288f808a6750bc26e9ae182/server/service/src/sync/translations/invoice.rs#L289

The `name_id` here is actually `name_link_id` value from the sql `invoice` table. `InvoiceRepository` does `inner join` shenanigans to get the related `item_id`.

Simplest solution is to use the domain repositories rather than the DB row repositories. e.g.

https://github.com/msupply-foundation/open-msupply/blob/e4de7353c761bb836288f808a6750bc26e9ae182/server/service/src/sync/translations/invoice.rs#L279-L280

Should use `InvoiceRepository` instead and required changes around that.

If that can't be used, it's OK to get it from `NameLinkRepository`.

Eventually we will actually rename `name_id` to `name_link_id` properly which will make this much clearer.

### Moneyworks Jobcode 🧰

OMS:MERGE

roxy-dao:
Fixes #2745

# 👩🏻‍💻 What does this PR do? 
Makes the translator use the `InvoiceRepository` rather than `InvoiceRowRepository` to get the correct name_id.

- Run tests

### DIFF
--- a/server/service/src/sync/test/merge_helpers.rs
+++ b/server/service/src/sync/test/merge_helpers.rs
@@ -1,8 +1,10 @@
 use repository::{
-    mock::{mock_name_link_from_name, MockDataCollection},
-    NameLinkRowRepository, NameRow, RepositoryError, StorageConnection,
+    mock::{mock_item_link_from_item, mock_name_link_from_name, MockDataCollection},
+    ItemLinkRowRepository, ItemRow, NameLinkRowRepository, NameRow, RepositoryError,
+    StorageConnection,
 };
 
+// Changes all name_id values to "name_a" to merge all names into "name_a".
 pub fn merge_all_name_links(
     connection: &StorageConnection,
     mock_data: &MockDataCollection,
@@ -20,6 +22,29 @@ pub fn merge_all_name_links(
         let mut name_link = mock_name_link_from_name(&name);
         name_link.name_id = "name_a".to_string();
         name_link_repo.upsert_one(&name_link)?;
+    }
+
+    Ok(())
+}
+
+// Changes all item_id values to "item_a" to merge all items into "item_a".
+pub fn merge_all_item_links(
+    connection: &StorageConnection,
+    mock_data: &MockDataCollection,
+) -> Result<(), RepositoryError> {
+    let item_link_repo = ItemLinkRowRepository::new(&connection);
+
+    let items: Vec<ItemRow> = mock_data
+        .data
+        .iter()
+        .map(|(_, mock)| mock.items.clone())
+        .flatten()
+        .collect();
+
+    for item in items {
+        let mut item_link = mock_item_link_from_item(&item);
+        item_link.item_id = "item_a".to_string();
+        item_link_repo.upsert_one(&item_link)?;
     }
 
     Ok(())
--- a/server/service/src/sync/test/merge_helpers.rs
+++ b/server/service/src/sync/test/merge_helpers.rs
@@ -14,8 +14,7 @@ pub fn merge_all_name_links(
     let names: Vec<NameRow> = mock_data
         .data
         .iter()
-        .map(|(_, mock)| mock.names.clone())
-        .flatten()
+        .flat_map(|(_, mock)| mock.names.clone())
         .collect();
 
     for name in names {
@@ -37,8 +36,7 @@ pub fn merge_all_item_links(
     let items: Vec<ItemRow> = mock_data
         .data
         .iter()
-        .map(|(_, mock)| mock.items.clone())
-        .flatten()
+        .flat_map(|(_, mock)| mock.items.clone())
         .collect();
 
     for item in items {
--- a/server/service/src/sync/test/merge_helpers.rs
+++ b/server/service/src/sync/test/merge_helpers.rs
@@ -1,0 +1,26 @@
+use repository::{
+    mock::{mock_name_link_from_name, MockDataCollection},
+    NameLinkRowRepository, NameRow, RepositoryError, StorageConnection,
+};
+
+pub fn name_links_merged(
+    connection: &StorageConnection,
+    mock_data: &MockDataCollection,
+) -> Result<(), RepositoryError> {
+    let name_link_repo = NameLinkRowRepository::new(&connection);
+
+    let names: Vec<NameRow> = mock_data
+        .data
+        .iter()
+        .map(|(_, mock)| mock.names.clone())
+        .flatten()
+        .collect();
+
+    for name in names {
+        let mut name_link = mock_name_link_from_name(&name);
+        name_link.name_id = "name_a".to_string();
+        name_link_repo.upsert_one(&name_link)?;
+    }
+
+    Ok(())
+}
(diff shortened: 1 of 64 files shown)


server/service/src/sync/test/merge_helpers.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    let names: Vec<NameRow> = mock_data
+        .data
+        .iter()
+        .map(|(_, mock)| mock.names.clone())
+        .flatten()
+        .collect();
+
+    for name in names {
+        let mut name_link = mock_name_link_from_name(&name);
```
clemens-msupply, line 20: Ups, I think I deleted `mock_name_link_from_name` after some refactoring/cleaning because it wasn't used anymore...

server/service/src/sync/test/merge_helpers.rs
```diff
@@ -0,0 +1,26 @@
+use repository::{
+    mock::{mock_name_link_from_name, MockDataCollection},
+    NameLinkRowRepository, NameRow, RepositoryError, StorageConnection,
+};
+
+pub fn name_links_merged(
```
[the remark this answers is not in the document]
clemens-msupply: Its because you need a `name_links_merged`, right? I would prefer to be consitend with item_links_merged(), i.e. add some similar option for the mock data. @Chris-Petty any preferences on which version to use? Like Roxies version because its more explicit...

However, the function name could be improved and the function could do with some docs about what it does and what it is used for...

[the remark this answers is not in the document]
Chris-Petty: Hmm I think I prefer Roxy's, in particular because it doesn't require a silly comment explaining why it isn't part of "all()": 

https://github.com/msupply-foundation/open-msupply/blob/2c21a6b2e282fce3a59eace90cb8c8dcce40407a/server/repository/src/mock/mod.rs#L290

TLDR: The above - do it Roxy's way 😆 

And because... it's more transparently imperative code. Bit of a refactor tangent but all the mock code goes into the mock data inserty thingy which has helper functions for every property that is all boiler plate, but ultimately I don't see the benefit of:

```rust
MockDataInserts::none()
                .mock_units();
                .items()
                .item_links_merged()
                .names()
                .stores()
                .locations()
                .barcodes()
                .full_invoices(),
```
over
```rust
mock_units();
mock_items();
mock_item_links_merged();
mock_names();
mock_stores();
mock_locations();
mock_barcodes();
mock_full_invoices();
`

Where each of those functions just do the upsert on their respective repo.  Ideally each in their own file with the mock data they'll insert. That way when I need to know what data is actually inserted I can go to def and it actually tells me lol. I went in so many circles the other day trying to find where the actual data of `.full_requisitions()` is defined by navigating code, I gave up and just let it run and checked the sqlite db for what data it makes instead lol.

Functionally all I see is it being used in this way, other than `MockDataInserts::all()`, which can just be a function that calls all the other mock functions. Unless it provided something like sync where it ensured things were inserted in referentially intact order all I've felt is pain IMO 😅.

Or if Repos all had a "Repo" trait that ensured `upsert_one` etc. was defined and all data type structs had a trait that ensured you can access their corresponding repo. Then MockDataCollection could generically iterate all the mock data in it (in referential order) but pretty sure this is getting overcooked for tests and really someone will one day want to force in data that is bad for testing IDK 🤪.

File: server/service/src/sync/test/merge_helpers.rs (hunk 1 above)
Comment (Chris-Petty) [reply server/service/src/sync/test/merge_helpers.rs:20]: Updated from base to get the compiler upset, v true. Request changes!

File: server/service/src/sync/test/merge_helpers.rs [hunk 3]
```diff
@@ -24,3 +26,26 @@ pub fn merge_all_name_links(
...
+    mock_data: &MockDataCollection,
+) -> Result<(), RepositoryError> {
+    let item_link_repo = ItemLinkRowRepository::new(&connection);
+
+    let items: Vec<ItemRow> = mock_data
+        .data
+        .iter()
+        .map(|(_, mock)| mock.items.clone())
+        .flatten()
+        .collect();
```
Chris-Petty, line 42: Nice!

server/service/src/sync/translations/stock_line.rs
```diff
@@ -204,20 +206,14 @@ mod tests {
...
-            MockDataInserts::none()
-                .units()
-                .items()
-                .item_links_merged()
-                .names()
-                .stores()
-                .locations()
-                .barcodes()
-                .stock_lines(),
+            MockDataInserts::all(),
```
Chris-Petty, line 211: Was there a reason we needed to go to all? Individually probably negligible on test run time, but if doing `all` for ALL tests then it might slow things down a bit?

Chris-Petty (changes requested):
ded 
image

server/service/src/sync/translations/stock_line.rs
clemens-msupply: Sqlite has a migration cache for all(). So all() in sqlite is actually faster than individual selected data... For postgres all() is slower though.

However, I think for sqlite we should update the test code to use all() even if you selected specific data (and explicitely opt out if the test doesn't work with all())...

server/service/src/sync/test/merge_helpers.rs
clemens-msupply: Or `flat_map()`?

server/service/src/sync/translations/stock_line.rs
Chris-Petty: I wonder if you could get a similar benefit from using postgres template databases? https://www.postgresql.org/docs/current/manage-ag-templatedbs.html

clemens-msupply: I tried this (there is a branch somewhere) but I only got something like a 50% speed up and `slow / 2 `  was still `slow` for which reason I didn't follow through with it. Thought having a full test without cache might be more beneficial in this case.

Think the reason is that even with a template Postgres still has to populate the internal data structs while for sqlite we just do a very fast sequential copy of the database file.

server/service/src/sync/test/merge_helpers.rs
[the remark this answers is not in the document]
clemens-msupply: Good point, also find it hard to follow what has been inserted. However, the sqlite test caching might break...

server/service/src/sync/translations/stock_line.rs
clemens-msupply: Maybe I should give this a try:
https://aiven.io/community/forum/t/faq-why-is-database-creation-from-templates-slower-for-large-databases-in-postgresql-15/421

clemens-msupply: I am in favour to keep them. If this has been done for performance reasons, I would rather make the test caching smarter...

server/service/src/sync/test/merge_helpers.rs
clemens-msupply: At the moment `mock_name_link_from_name` is only used once could you please inline it? I don't think we need a helper method for it...

Issue #2471, opened 2023-11-13
Rename omSupply -> Open mSupply


Chris-Petty:
## What went wrong? 😲

Last I heard the memo was to rename `"omSupply"` to `"Open mSupply"` as "omSupply" is a bit of an internal shorthand and not our product's name. Therefore, installer/uninstaller should probably make the change.

![Screenshot 2023-11-10 at 12 28 17](https://github.com/msupply-foundation/open-msupply/assets/7684221/f46df749-fcd7-43ab-887a-12fea88e5844)

Also (ignore I've uninstalled haha) the installation directory:

![Screenshot 2023-11-10 at 12 28 42](https://github.com/msupply-foundation/open-msupply/assets/7684221/145c0569-bcb8-43ab-aa1a-bcb093f5c7e7)

(tangentially, should configuration be cleaned up by the uninstaller? IDK what convention is for what files to leave behind but I generally don't like apps that leave rubbish behind!)

## Your environment 🌱

 - Version: 1.5.0-RC3
 - Platform: All windows installers/uninstallers

nisha-dangol:
Tested in v1.6.0-rc3 exe

I do see the word `Open mSuppply` in place of `omsupply`. So, looks like a PASS to me!
<img width="553" alt="oms2" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/960369a5-352f-413b-98bf-0d5748246031">
<img width="552" alt="oms3" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/80154fa0-adbe-4414-a521-8531f3583c98">
<img width="552" alt="oms4" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/bf00fab5-709a-43e9-a698-a2bab7968731">
<img width="556" alt="oms5" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/eef2682d-4828-413e-a760-89fc271c2be7">
<img width="549" alt="oms1" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/9b9efd81-dbc5-45c9-bbda-73bdefa658e8">
<img width="154" alt="oms-shortcut" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/06611721-7522-480c-825c-01624b7ad624">

As for the server upgrade from 1.5.4 to 1.6, it takes the previous location (`omsupply Server`); which is correct technically. So, I'll PASS this!
[...]

PR #2764, opened 2024-01-18
Detect path when running an upgrade

### LINKED ISSUE #2737: Server upgrade fails
## What went wrong? 😲
 
With the issue #2471 coming in, the upgrade process of open mSupply server has an issue.
The existing server is detected, and the upgrade runs successfully. However, the new version is installed into the folder `Open mSupply Server` and the configuration and logs remain in the previous installation folder (in my case, 'omSupply Server').

As a result, the service does not start but enters an endless starting state.

## Expected behaviour 🤔
I would expect the current folder to be used. 
Please retest this scenario too:

* Install an earlier version into a non-standard folder
* Upgrade

## How to Reproduce 🔨

1. Install a previous version
2. Upgrade to v1.6.00-rc1

## Your environment 🌱

 - Version: 
 - Platform:
   - [ ] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [x] server (windows)

mark-prins:
Fixes #2737

# 👩🏻‍💻 What does this PR do? 
 
Sets the install path to be the currently installed location, if an upgrade is detected.

- Install a previous version into a non-standard folder. 
- Install this version: you should see the upgrade process and it will detect the current folder

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/build/windows/omsupply_server.suf
+++ b/build/windows/omsupply_server.suf
@@ -4073,6 +4073,11 @@ local omSupplyServerPath = "";
 -- Check if the omSupply Server has been already installed on this machine
 if Registry.DoesKeyExist(HKEY_LOCAL_MACHINE, "Software\\omSupply Server") then
 	omSupplyServerPath = Registry.GetValue(HKEY_LOCAL_MACHINE, "Software\\omSupply Server", "Path", true);
+	local index = String.ReverseFind(omSupplyServerPath, "\\");
+	if index ~= -1 then
+		omSupplyServerPath = String.Left(omSupplyServerPath, index-1)
+		SessionVar.Set("%AppFolder%", omSupplyServerPath);
+	end
 end
 
 hasOmsupplyInstalled = File.DoesExist(omSupplyServerPath);
@@ -4404,7 +4409,7 @@ g_HandleSystemReboot();
 </SessionVar>
 <SessionVar>
 <Name>%omsupply_db_name%</Name>
-<Value>omsupply-database</Value>
+<Value>open-msupply</Value>
 <Type>10</Type>
 </SessionVar>
 <SessionVar>
--- a/build/windows/omsupply_server.suf
+++ b/build/windows/omsupply_server.suf
@@ -2990,7 +2990,7 @@ end
 <OverwriteCtrlPanel>1</OverwriteCtrlPanel>
 <OverwriteSC>1</OverwriteSC>
 <CreateCPEntry>1</CreateCPEntry>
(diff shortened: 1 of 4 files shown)


build/windows/omsupply_server.suf
```diff
@@ -4404,7 +4409,7 @@ g_HandleSystemReboot();
 </SessionVar>
 <SessionVar>
 <Name>%omsupply_db_name%</Name>
-<Value>omsupply-database</Value>
+<Value>open-msupply</Value>
```
mark-prins, line 4412: not part of the bug fix, just tidying up.

build/windows/omsupply_standalone.suf
```diff
@@ -3492,7 +3497,7 @@ g_HandleSystemReboot();
 </SessionVar>
 <SessionVar>
 <Name>%omsupply_db_name%</Name>
-<Value>open-msupply-database</Value>
+<Value>open-msupply</Value>
```
mark-prins, line 3500: made it consistent

build/windows/omsupply_server.suf
clemens-msupply: I have no idea what it's doing :) but can you rename the database name? e.g. what if it is just a version upgrade?

mark-prins:
> can you rename the database name? e.g. what if it is just a version upgrade?

The database name is only used for new installs - we take the name and patch a `local.yaml` file.. but for upgrades, it doesn't touch the config.

clemens-msupply (commented):
Works fine. I installed 1.6 Rc2 and then Rc3 and it picked my previously selected directory.

However, when uninstalling it I got two entries Rc2 + Rc3 after unistalling one of them I can't uninstall the other anymore...

image

This might be an unrelated problem though?

mark-prins:
I think unrelated.. though worth looking at since I'm working on the installer. I only saw the one entry after installing the second.. will check and see how windows determines a unique app

clemens-msupply (approved):

mark-prins:
I've found the issue with the upgrade - the unique registry key value in the uninstall settings was appending the product version to the product name, which meant that is was creating a new entry in the uninstall list 🙄 

This won't fix the issue of upgrading to 1.6.00+ however, because of #2471
The name has changed from 1.6 onwards and so will always create a new entry. after that though, the upgrade works!

You can test with an upgrade from [v1.5.04-install-test](https://jenkins.msupply.org:8443/job/omSupplyMain%20-%20Installers/202/) to [1.6.00-install-test](https://jenkins.msupply.org:8443/job/omSupplyMain%20-%20Installers/203/)

PR #2765, opened 2024-01-18
Name merging requisition.name id translator get actual name.id

### LINKED ISSUE #2746: Name merging: requisition.name_id translator get actual name.id

name merging needs to make sure that when records with FK field `name_link_id` or similar sync, they correctly send the `name_id` of the linked name as `name_id`, not `name_link_id` as `name_id`. e.g.

See #2745.

See #2745.

Eventually we will actually rename `name_id` to `name_link_id` properly which will make this much clearer.

### Moneyworks Jobcode 🧰

OMS:MERGE

roxy-dao:
Fixes #2746

# 👩🏻‍💻 What does this PR do? 
Makes the translator use the `RequisitionRepository` rather than `RequisitionRowRepository` to get the correct name_id.

- Run tests

### DIFF
--- a/server/repository/src/mock/test_requisition_service.rs
+++ b/server/repository/src/mock/test_requisition_service.rs
@@ -41,7 +41,7 @@ pub fn mock_test_requisition_service() -> MockData {
         .push(mock_request_draft_requisition_calculation_test());
     result
         .full_requisitions
-        .push(mock_new_response_requisition_for_update_test());
+        .push(mock_full_draft_response_requisition_for_update_test());
     result
         .full_requisitions
         .push(mock_new_response_requisition_test());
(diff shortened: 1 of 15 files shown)


server/repository/src/mock/test_requisition_service.rs
```diff
@@ -198,6 +201,34 @@ pub fn mock_draft_response_requisition_for_update_test_line() -> RequisitionLine
     })
 }
 
+pub fn mock_full_draft_response_requisition_for_update_test() -> FullMockRequisition {
```
roxy-dao, line 204: Kinda hacky since API doesn't allow `Response Requisitions` to have a `DRAFT` status... Created to pass `NotARequestRequisition` error... that should never happen anyways but have kept the error in to be consistent with invoice/response requisition errors...

clemens-msupply: Was this already a problem/hacky before?

server/repository/src/mock/test_requisition_service.rs
```diff
@@ -159,7 +162,7 @@ pub fn mock_draft_response_requisition_for_update_test() -> RequisitionRow {
         r.name_id = "name_a".to_owned();
         r.store_id = mock_store_a().id;
         r.r#type = RequisitionRowType::Response;
-        r.status = RequisitionRowStatus::Draft;
+        r.status = RequisitionRowStatus::New;
```
clemens-msupply, line 165: Please rename the fn name from `mock_draft_response_requisition_for_update_test` to `mock_new_response_requisition_for_update_test` + rename id string...

clemens-msupply (changes requested):

server/repository/src/mock/test_requisition_service.rs
roxy-dao: Yeah it's testing for DRAFT status, I just moved it so it doesn't fail the translator tests

clemens-msupply (approved):

PR #2777, opened 2024-01-19
Refetch notifications

### LINKED ISSUE #2762: Cold chain: Breach notification bar does not appear until user logs out and back in
## What went wrong? 😲

The breach notification bar is not appearing until the user forces some kind of refresh by e.g. logging out and back in 

https://github.com/msupply-foundation/open-msupply/assets/19318937/c716c6be-481f-48c7-9d12-d0bcd0e5e1e5

Also noticed that excursions didn't appear either: 

image

## Expected behaviour 🤔

Notification bar should appear as soon as a breach or excursion occurs. 

## How to Reproduce 🔨

1. Log into Open mSupply
2. Produce a breach on CCA
3. Observe Open mSupply

## Your environment 🌱

 - Version: V1.6.00-rc2 
 - Platform:
   - [x] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

mark-prins:
Fixes #2762

# 👩🏻‍💻 What does this PR do? 
 
Adds a regular refetch of notifications (breaches & excursions) so that if the site ends up in a state like the one in the issue (a breach is identified but no notification shown) then a refetch will happen and the notification will be shown.

Have also disabled the cache for that query - though I'm concerned that the notification query is a little slow, so I set the refetch to be five minutes. Didn't want it to be called too often.

Not sure if you'll be able to recreate the scenario in the issue - I wasn't able to.
You can check that the site will fetch notifications every 5 minutes ( reduce the polling interval if you are impatient )

## 📃 Documentation

- [ ] Worth mentioning that the notification is refetched

### DIFF
--- a/client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
+++ b/client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
@@ -6,7 +6,9 @@ import {
 } from '@openmsupply-client/common';
 import { ListParams } from '../../api';
 
-const POLLING_INTERVAL_IN_MINUTES = 5;
+const MILLISECONDS_PER_MINUTE = 60 * 1000;
+const POLLING_INTERVAL_IN_MILLISECONDS = 3 * MILLISECONDS_PER_MINUTE;
+const STALE_TIME_IN_MILLISECONDS = 1 * MILLISECONDS_PER_MINUTE;
 
 export const useTemperatureNotifications = (queryParams: ListParams) => {
   const api = useTemperatureNotificationApi();
@@ -22,8 +24,9 @@ export const useTemperatureNotifications = (queryParams: ListParams) => {
           warning(`${t('error.fetch-notifications')}: ${e.message}`)()
         ),
     {
-      cacheTime: 0,
-      refetchInterval: 1000 * 60 * POLLING_INTERVAL_IN_MINUTES,
+      cacheTime: POLLING_INTERVAL_IN_MILLISECONDS,
+      refetchInterval: POLLING_INTERVAL_IN_MILLISECONDS,
+      staleTime: STALE_TIME_IN_MILLISECONDS,
     }
   );
 };
--- a/client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
+++ b/client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
@@ -6,14 +6,24 @@ import {
 } from '@openmsupply-client/common';
 import { ListParams } from '../../api';
 
+const POLLING_INTERVAL_IN_MINUTES = 5;
+
 export const useTemperatureNotifications = (queryParams: ListParams) => {
   const api = useTemperatureNotificationApi();
   const { warning } = useNotification();
   const t = useTranslation('coldchain');
 
-  return useQuery(api.keys.paramList(queryParams), () =>
-    api.get
-      .list(queryParams)()
-      .catch(e => warning(`${t('error.fetch-notifications')}: ${e.message}`)())
+  return useQuery(
+    api.keys.paramList(queryParams),
+    () =>
+      api.get
+        .list(queryParams)()
+        .catch(e =>
+          warning(`${t('error.fetch-notifications')}: ${e.message}`)()
+        ),
+    {
+      cacheTime: 0,
+      refetchInterval: 1000 * 60 * POLLING_INTERVAL_IN_MINUTES,
+    }
   );
 };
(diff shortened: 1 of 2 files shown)


client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
```diff
@@ -6,14 +6,24 @@ import {
...
+  return useQuery(
+    api.keys.paramList(queryParams),
+    () =>
+      api.get
+        .list(queryParams)()
+        .catch(e =>
+          warning(`${t('error.fetch-notifications')}: ${e.message}`)()
+        ),
+    {
+      cacheTime: 0,
```
clemens-msupply, line 25: Did you set it 0 because the cache is not needed/not used? For example, the user never triggers this query manually because it is only run in the background on the main page? So the cache will never be used because the query is only done because of the refetchInterval?

Otherwise, it might be ok to have `refetchInterval == cacheTime` and keep `staleTime =0`. Then a user interaction would immediately get the cached query and because of staleTime=0 will query the current notifications...

client/packages/common/src/api/GqlContext.tsx
```diff
@@ -31,6 +31,12 @@ const permissionExceptions = [
   'requisitionCounts',
   'temperatureNotifications',
 ];
+
+// these queries are not considered to be part of the user's activity
+// they occur in the background and should not be used to determine
+// if the user has remained active
+const ignoredQueries = ['refreshToken', 'syncInfo', 'temperatureNotifications'];
```
clemens-msupply, line 38: thanks for the comment, helped me a lot :)

clemens-msupply (approved):
Looks good! Untested though because it looks straight forward.

client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
mark-prins: I was in two minds about that! 
The cache time 0 will disable the cache - my thinking is that we only ever want the latest info for the notification area; I rather like having it cached though, because it's an expensive query to run.

so if we set `refetchInterval == cacheTime` then the original problem will be fixed with the refetch, and the server impact will be minimised for the case when the user navigates or refreshes prior to the cache expiring. but it does mean that the user won't get the latest updates until the cache times out.. given that it's only a few minutes.. probably worth it?

clemens-msupply: There is also staleTime which defaults to 0. This means if the user triggers the query the old data is immediately displayed when the old data is within the cacheTime but because staleTime is 0 it will also start a new query. 

Hope this makes sense.
So you could tweak staleTime and cacheTime...

client/packages/coldchain/src/Monitoring/api/TemperatureNotification/hooks/document/useTemperatureNotifications.ts
```diff
@@ -22,8 +24,9 @@ export const useTemperatureNotifications = (queryParams: ListParams) => {
           warning(`${t('error.fetch-notifications')}: ${e.message}`)()
         ),
     {
-      cacheTime: 0,
-      refetchInterval: 1000 * 60 * POLLING_INTERVAL_IN_MINUTES,
+      cacheTime: POLLING_INTERVAL_IN_MILLISECONDS,
+      refetchInterval: POLLING_INTERVAL_IN_MILLISECONDS,
+      staleTime: STALE_TIME_IN_MILLISECONDS,
```
clemens-msupply, line 29: Nit: is this more readable?:

```ts
      cacheTime: 3 * MINUTE,
      refetchInterval: 3 * MINUTE,
      staleTime: 1 * MINUTE,
```

With `const MINUTE = 60 * 1000;`

mark-prins: it is more readable, yes. I prefer to have CONSTS at the top of the file as a rule, it means that if you want to change anything then you don't need to scroll down too far. having the numerical values lower in the code goes against that - although in this case, line 27 isn't that far to go!

I also prefer to have `MILLISECONDS` in the const name rather than just `MINUTE` because it isn't clear what units useQuery is expecting.

clemens-msupply: In the js world everything is in `ms` on default, isn't it? But maybe just `_MS_` is enough here?

clemens-msupply: I mean `STALE_TIME_IN_MILLISECONDS` -> `STALE_TIME_MS`

mark-prins: mostly yes - but some packages helpfully have parameters in minutes or seconds. I've changed to `_MS` to make things neater 🙏

PR #2779, opened 2024-01-19
Fix for Temperature Chart date filter

### LINKED ISSUE #2757: Cold chain: Stuck state with dates in cold chain monitoring
## What went wrong? 😲

Clicking on the date (rather than the date picker), then clicking elsewhere on the screen will cause the 'to' date to be null.
There are some other ways to cause the date to become null, but this one is the easiest.

When the 'to' date is null, the app will get stuck in loading and fail to render. Refreshing the page does not fix this. The absent closing date is kept in the URL. While loading, the user is also unable to change the selected dates back.

The only way to fix is to manually delete the url date range and then refresh, or by accessing a different route via the lefthand side app bar and then navigating back.

https://github.com/msupply-foundation/open-msupply/assets/90807420/b4b72ae5-a749-4131-aa67-2f6045015588

## Expected behaviour 🤔
Shouldn't freeze in loading state when no end date is selected

## How to Reproduce 🔨

Open omSupply with coldchain. 
Deselect the to date.

mark-prins:
Fixes #2757 

# 👩🏻‍💻 What does this PR do? 
 
When either part of the date filter was null, the date value was being set to the current datetime ( to the nearest millisecond ) which updates the filter and causes `TemperatureChart` to re-render, causing the `Toolbar` to re-render which updates the filter.

To fix, I've lifted out the to/from date extraction, rounded off the time to the nearest second and memoised the resulting date strings.

The original repro step has been fixed, sorry. This is what I've been doing:
- On the Temperature Chart, click on the from (or to) date
- click on the day part of the day, hit `del`
- the page then re-renders, you'll see lots and lots of api requests for the `temperatureChart` query

One concern, if you clear the 'from' date it defaults to a day ago, but doesn't tell you that it has. There can be no value in the date input, but that isn't treated as 'no from date' it is treated as 'a day ago'.

That's a little misleading. is it something to address, do you think?

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/coldchain/src/Monitoring/ListView/TemperatureChart/useTemperatureChartData.ts
+++ b/client/packages/coldchain/src/Monitoring/ListView/TemperatureChart/useTemperatureChartData.ts
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { useTheme } from '@common/styles';
 import { DateUtils } from '@common/intl';
 import { Sensor } from './types';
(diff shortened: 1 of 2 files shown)


client/packages/coldchain/src/Monitoring/ListView/TemperatureChart/useTemperatureChartData.ts
```diff
@@ -6,14 +7,57 @@ import {
...
 
 const MAX_DATA_POINTS = 30;
 const BREACH_RANGE = 2;
 const BREACH_MIN = 2;
 const BREACH_MAX = 8;
 
+const useFilterDates = (
+  filterDatetime: InputMaybe<DatetimeFilterInput> | undefined
+) => {
+  const now = DateUtils.setMilliseconds(new Date(), 0);
```
mark-prins, line 24: this is mostly just a move - I lifted it out so that I could experiment with memoising and decided to leave it, it's neater in a separate function

mark-prins, line 24: the `setMilliseconds` is one of the changes

client/packages/coldchain/src/Monitoring/ListView/TemperatureChart/useTemperatureChartData.ts
```diff
@@ -6,14 +7,57 @@ import {
...
+      if (fromDatetime >= toDatetime && !hasAfterOrEqualTo) {
+        fromDatetime = DateUtils.addDays(
+          new Date(toDatetime),
+          -1
+        ).toISOString();
+      }
+    }
+  }
+
+  return useMemo(
```
mark-prins, line 53: and the memoising is the other

clemens-msupply (approved):
Works fine! i.e. it keeps rendering

> One concern, if you clear the 'from' date it defaults to a day ago, but doesn't tell you that it has. There can be no value in the date input, but that isn't treated as 'no from date' it is treated as 'a day ago'.
>
>That's a little misleading. is it something to address, do you think?

Yes sounds misleading but on the other hand the graph has x-axis labels which should be clear enough? However, I can't reproduce if I clear the day it says `Date value is invalid` and it doesn't update the graph...

mark-prins:
🙏 thanks again @clemens-msupply

Issue #2733, opened 2024-01-15
Move merging migrations to 1.7


roxy-dao:
## Is your feature request related to a problem? Please describe 👀
Move merging migrations to 1.7. Should be done after every issue for merge has been completed.

### Moneyworks Jobcode 🧰
OMS:MERGE

Chris-Petty:
pls revert this commit where I removed indexing from 1.06 migration and double check drop/create indexes e4de7353c761bb836288f808a6750bc26e9ae182

Chris-Petty:
Closed in #2792

PR #2782, opened 2024-01-21
2607 drop migration changelog rows

### LINKED ISSUE #2607: Merging: utility method to drop changelogs created by migrations

In the merge changes, many records are being touched and potentially synced unnecessarily by migrations. We need to avoid this unnecessary traffic.

- Disabling triggers. This is easy for postgres, but it appears in sqlite there is no such option or PRAGMA. There is a [DB connection config that can enable/disable triggers](https://sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenabletrigger), but I'm not sure if we can manipulate this easily at runtime.
- Given above, dropping and recreating the triggers. [Perhaps it's not too hard! ](https://stackoverflow.com/a/44592846)

clemens-msupply:
Fixes #2607

# 👩🏻‍💻 What does this PR do? 
Remembers latest changelog and delete all changelog entries created during the merge migrations.

@Chris-Petty do we need to test this a bit more? or do you think it should be ok. For example, are there any cases where a migration change need to be synced?

Probably good to wrap all merge migrations into a function to group them by feature and to not accidentally put an unrelated migration in (after moving to 1.7 https://github.com/msupply-foundation/open-msupply/issues/2733)...

### DIFF
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -141,9 +141,11 @@ impl<'a> ChangelogRepository<'a> {
         Ok(result.unwrap_or(0) as u64)
     }
 
-    // Drop all change logs (for tests), can't set test flag as it's used in another crate
-    pub fn drop_all(&self) -> Result<(), RepositoryError> {
-        diesel::delete(changelog::dsl::changelog).execute(&self.connection.connection)?;
+    // Delete all change logs with cursor greater-equal cursor_ge
+    pub fn delete(&self, cursor_ge: i64) -> Result<(), RepositoryError> {
+        diesel::delete(changelog::dsl::changelog)
+            .filter(changelog::dsl::cursor.ge(cursor_ge))
+            .execute(&self.connection.connection)?;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


Chris-Petty (changes requested):
> @Chris-Petty do we need to test this a bit more? or do you think it should be ok. For example, are there any cases where a migration change need to be synced?

Hmm this seems good enough as is! Within merge I can't think of any migration that needs to be synced. 

In the past mSupply has required syncing some changes in migrations but not others. E.g.
- Generating joins (ahem link records) between tables: central server has the same IDs and could do the same migration, saving on sync traffic. (This would be valid with an actual join table too if it used a composite primary key of the 2 FKs in the join record)
- New denormalised field on invoice calculated from invoice lines: the central server may not have the latest invoice_line data and thus can't independently do the same migration, so the remote sites owning the data must sync the updates.

> Probably good to wrap all merge migrations into a function to group them by feature and to not accidentally put an unrelated migration in (after moving to 1.7 https://github.com/msupply-foundation/open-msupply/issues/2733)...

Yea good call please do it!

clemens-msupply:
> > Probably good to wrap all merge migrations into a function to group them by feature and to not accidentally put an unrelated migration in (after moving to 1.7 #2733)...
> 
> Yea good call please do it!

Ok thanks, did this but it got bigger than expected because I also decided to add a test for it. The test uses the same test data as the other merge test, i.e. I moved the test data into a helper method...

Chris-Petty (approved):
Sweet! Structure looks good though trading disorder for complexity maybe heh. Lesser of evils?

PR #2789, opened 2024-01-22
Add appending to database name

### LINKED ISSUE #817: Add `.sqlite` to generated sqlite database files
Currently when I run omSupply in dev, it creates a file called `omsupply-database`. It'd be more convenient if this was named `omsupply-database.sqlite` as then when I open it, my OS (Mac) tries to use my preferred DB app rather than a text editor as it currently does. 

Shouldn't affect pg based dbs in anyway I'd think.

fergie-nz:
Fixes #817 

# 👩🏻‍💻 What does this PR do? 
Adds .sqlite to sqlite database names if .sqlite isn't already the suffix

You can test this by running Cargo run to create a new db with different name configurations in local.yaml.

A new database should be added with .sqlite appended.

### DIFF
--- a/server/repository/src/database_settings.rs
+++ b/server/repository/src/database_settings.rs
@@ -50,6 +50,9 @@ impl DatabaseSettings {
 impl DatabaseSettings {
     pub fn connection_string(&self) -> String {
         let sqlite_suffix = |s: String| -> bool {
+            if s.len() < 8 {
+                return false;
+            }
             let split = s.char_indices().nth_back(6).unwrap().0;
             if &s[split..] == ".sqlite" {
                 return true;
--- a/server/repository/src/database_settings.rs
+++ b/server/repository/src/database_settings.rs
@@ -56,25 +56,20 @@ impl DatabaseSettings {
             self.database_name.clone()
         } else {
             // first check if database exists on disk. If it does, we will use db filename as is without appending .sqlite
-            let db_exists = self.check_db_exists();
-            match db_exists {
+            // Note, using `try_exists()` because we want to be able to store the sqlite file on a different partition or disc.
+            // If the disc is a network drive and the drive is temporarily offline it might happen that exists() returns false but doesn't
+            // say that there was a network error (not 100% if this really is how exists() work but try_exists seems safer). Then if the drive
+            // goes online, creating a new database will run as normal and return two files. This might result in data loss if
+            // data from the old file hasn't been synced yet. There are a number of feasible cases where this might occur, for example when mSupply
+            // automatically starts after a machine boots up a network drive might also be in the process of being mounted.
+            let exists = Path::new(&self.database_name.clone())
+                .try_exists()
+                .expect("Can't check existence of database file");
+            match exists {
                 true => self.database_name.clone(),
                 false => format!("{}.sqlite", self.database_name.clone()),
             }
         }
-    }
-
-    pub fn check_db_exists(&self) -> bool {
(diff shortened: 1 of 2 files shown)


server/repository/src/database_settings.rs
```diff
@@ -49,7 +49,20 @@ impl DatabaseSettings {
...
+            let split = s.char_indices().nth_back(6).unwrap().0;
+            if &s[split..] == ".sqlite" {
+                return true;
+            }
+            return false;
+        };
+        match sqlite_suffix(self.database_name.clone()) {
+            true => self.database_name.clone(),
+            false => format!("{}.sqlite", self.database_name.clone()),
+        }
```
clemens-msupply, line 65: Unfortunately this breaks existing installation, i.e. it doesn't use my existing `omsupply-database` file. I will add a comment to the issue...

clemens-msupply (changes requested):

server/repository/src/database_settings.rs
clemens-msupply: Can this be simplified to:
```rust
pub fn connection_string(&self) -> String {
        if self.database_name.ends_with(".sqlite") {
            self.database_name.clone()
        } else {
            format!("{}.sqlite", self.database_name)
        }
    }
```

roxy-dao: This is being done in sqlite feature though... so would the file extension be anything other than sqlite...?

clemens-msupply: Yes if an existing installation uses the old database name...

fergie-nz: I'm assuming that the you're after the connection_string function which still calls to check if the db exists without the suffix like:

```
pub fn connection_string(&self) -> String {
        if self.database_name.ends_with(".sqlite") {
            // just use DB if name ends in .sqlite
            self.database_name.clone()
        } else {
            // first check if database exists on disk. If it does, we will use db filename as is without appending .sqlite
            let db_exists = self.check_db_exists();
            match db_exists {
                true => self.database_name.clone(),
                false => format!("{}.sqlite", self.database_name.clone()),
            }
        }
    }
    ```

File: server/repository/src/database_settings.rs [hunk 2]
```diff
@@ -49,7 +51,30 @@ impl DatabaseSettings {
...
+            // first check if database exists on disk. If it does, we will use db filename as is without appending .sqlite
+            let db_exists = self.check_db_exists();
+            match db_exists {
+                true => self.database_name.clone(),
+                false => format!("{}.sqlite", self.database_name.clone()),
+            }
+        }
+    }
+
+    pub fn check_db_exists(&self) -> bool {
```
clemens-msupply, line 67: Bit misleading method name. It checks if the db with name `database_name` exist. But it might only exist with ".sqlite" extension. Maybe just inline this one-liner (even though comment is quite verbose) to avoid confusion.

To avoid "if nesting" you can write the main method it like:

```rust
pub fn connection_string(&self) -> String {
        if self.database_name.ends_with(".sqlite") {
            // just use DB if name ends in .sqlite
            return self.database_name.clone()
        } 
         
        // first check if database exists on disk. If it does, we will use db filename as is without appending .sqlite
        let exists = INLINE DB CHECK HERE
        match exists {
            true => self.database_name.clone(),
            false => format!("{}.sqlite", self.database_name.clone()),
        }
}
```

clemens-msupply (approved):
Thanks works as expected!

Issue #2491, opened 2023-11-16
Line inputs we cannot backspace to remove single digit value


Chris-Petty:
## What went wrong? 😲
 

https://github.com/msupply-foundation/open-msupply/assets/7684221/737bcf28-b9d9-4c35-9a7f-f375a4724be6

When editing a stock take line, data entry errors can occur. If there is a single digit `6` and you meant `7`, you cannot simply backspace to clear `6` and then type in `7`. You have to highlight `6` and replace it.

I assume this affects UI throughout mSupply but haven't tried yet.

## Expected behaviour 🤔

Backspace should clear the input. If you change focus after clearing the value it should be set to 0.

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create a stocktake
2. Add a line
3. Add a batch
4. Edit counted pack or pack size. Put in `7`.
5. Try to clear `7` with backspace and be confused

## Your environment 🌱

 - Version: 1.5.0-RC3
 - Platform: Desktop/Server + Windows Server 12

Chris-Petty:
This is almost certainly a duplicate of https://github.com/msupply-foundation/open-msupply/issues/1870

PR #2794, opened 2024-01-23
#1440 #2491 Number component refactor

### LINKED ISSUE #1440: Refactor four number input components into a single one
As mentioned in the discussion [here](https://github.com/openmsupply/open-msupply/pull/1433#discussion_r1151118288), there are four different number input components, which are almost the same:

Screenshot 2023-03-29 at 11 33 29 AM

And they were not actually consistent -- the "PositiveNumberInput" is supposed to only accept numbers 1 and above, but was actually allowing `0` if you set `min = 0` (but not anything *less* than zero), making it practically indistinguishable from the "NonNegativeNumberInput" when used this way.

This is all very confusing, and [it was suggested](https://github.com/openmsupply/open-msupply/pull/1433#discussion_r1151118288) that it would be better to just have a single component and specify the range using `min` and `max` props if we need to restrict it.

The only other difference is "Integer" or "Decimal", but that also could easily be distinguished with a prop.

CarlosNZ:
Fixes #1440 and #2491

# 👩🏻‍💻 What does this PR do? 
 

Sorry, bit of a massive one here -- the actual change to the number components is fairly small, but they're used in a lot of places!

Basically, all the number inputs can be consolidated to a single component with a just a couple of props to specify the type of input (integer/decimal, positive/negative). Similarly with the `NumberInputCell` (used in tables), which just use the main Number inputs.

The four props that can be set are:

```ts
  allowNegative?: boolean; 
  min?: number; // default 0
  max?: number;
  precision?: number; // number of decimal places, default 0 (i.e integer)

```
It's possible for the `allowNegative` and `min` props to conflict (e.g. `{ allowNegative: true, min: 10 }`). In this case, the "min" value will take precedence and `allowNegative` will be ignored.

Also addresses the problem in #2491 -- was mainly due to many number inputs not handling an `undefined` value, which is what is (temporarily) required when backspacing to nothing.

Unfortunately, will be a bit tricky to test out every use case affected here. I suggest just checking a handful and make sure they behave as expected and we can keep our eyes open for any potential bugs in the next few weeks. But I'm reasonably confident that it's working fine in most cases since there isn't really much functionality change, it's just changing the component called and tweaking the props a bit.

Have also update the Storybook components. Do we need anything more for this?


CarlosNZ:
> ## Bundle size difference
> 
> Comparing this PR to `main`
> Old size 	New size 	Diff
> 8.36 MB 	8.35 MB 	-6.65 KB (-0.08%)

Nice to have a decrease for once ;)

client/packages/invoices/src/InboundShipment/DetailView/modals/TaxEdit.tsx
```diff
@@ -17,6 +17,7 @@ interface TaxEditProps {
 export const TaxEdit = ({ disabled = false, tax, onChange }: TaxEditProps) => {
   const modalController = useToggle();
   const t = useTranslation('replenishment');
+  const [val, setVal] = useState<number | undefined>(tax);
```
CarlosNZ, line 20: I got rid of the "NonNegativeNumberInputModal" as all it was was a wrapper round InputModal with a different input field (number rather than text). However, it's just as simple to just pass the `NumericTextInput` as the `Input` prop directly. The result of that, though, is that something needed to manage the state of the tax input value, hence the added `useState` here. (This is more consistent with the other implementations, where the parent component manages the state and the input components just call the passed in `onChange` method)

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
```diff
@@ -117,9 +117,10 @@ export const OutboundLineEditForm: React.FC<OutboundLineEditFormProps> = ({
     updateIssueQuantity(allocatedQuantity);
   };
 
-  const handleIssueQuantityChange = (quantity: number) => {
+  const handleIssueQuantityChange = (quantity?: number) => {
     setIssueQuantity(quantity);
-    allocate(quantity, Number(packSizeController.selected?.value));
+    if (quantity !== undefined)
```
CarlosNZ, line 122: This is the only one that really required any special handling for the `undefined` value, and it's not much.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { NumUtils } from '@common/utils';
+export interface NumericTextInputProps
+  extends Omit<StandardTextFieldProps, 'onChange'> {
+  onChange?: (value: number | undefined) => void;
+  width?: number | string;
+  defaultValue?: number;
+  allowNegative?: boolean;
+  min?: number;
+  max?: number;
+  precision?: number;
```
mark-prins, line 13: :nit while I like how concise `precision` is, I think it requires explanation, where `decimalPlaces` or `decimalLimit` do not

client/packages/common/src/ui/components/modals/InputModal/InputModal.stories.tsx
```diff
@@ -25,27 +28,39 @@ export const BasicExample = () => {
...
+          alert(val);
+        }}
       />
     </Box>
   );
 };
 
 export const NonNegativeNumberInputModalExample = () => {
   const modalController = useToggle();
+  const [val, setVal] = useState<number>();
```
mark-prins, line 42: this can be removed - there's no point keeping it, now that the component has been deleted

client/packages/common/src/ui/components/inputs/TextInput/TextInput.stories.tsx
```diff
@@ -66,22 +63,32 @@ const NumericTemplate: Story = () => {
           <Grid item xs>
             <StyledPaper>
               <Typography>NumericTextInput</Typography>
-              <NumericTextInput />
+              <NumericTextInput
```
mark-prins, line 66: This one could have a more accurate title now. (2 decimal places, allows negatives)
It would be good to add an example showing the defaults. You could have
* default options
* allows negative
* max value set

mark-prins (changes requested):
Only half way through the files; have found some showstoppers, so will put in a comment at this point.

The number input cell is now editable - and should not be:

Screenshot 2024-01-24 at 3 14 52 PM

This is showing up in various tables e.g. outbound line edit modal is now:

Screenshot 2024-01-24 at 3 16 37 PM

and was

Screenshot 2024-01-24 at 3 16 44 PM



I can enter non-numeric input, and the intention of the `NumericTextInput` component is that it only allows numeric input e.g.

Outbound shipment line edit modal:

Screenshot 2024-01-24 at 3 15 49 PM

after entering a character into the outbound shipment edit modal pack Qty issued:

Screenshot 2024-01-24 at 3 16 26 PM

client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.tsx
```diff
@@ -204,7 +205,7 @@ export const useInboundShipmentColumns = () => {
       [
         'numberOfPacks',
         {
-          Cell: PositiveNumberCell,
+          Cell: props => <NumberInputCell {...props} min={1} />,
```
mark-prins, line 208: there are now 12 instances of `<NumberInputCell {...props} min={1} />` and only 14 usages of `NumberInputCell` in total. 

Why not just change the defaults of `NumberInputCell`? then you don't need to import React and create a new component here

client/packages/invoices/src/OutboundShipment/DetailView/modals/TaxEdit.tsx
```diff
@@ -28,12 +29,20 @@ export const TaxEdit = ({ disabled = false, tax, update }: TaxEditProps) => {
...
       {modalController.isOn && (
-        <NonNegativeNumberInputModal
-          max={100}
+        <InputModal
           isOpen={modalController.isOn}
           onClose={modalController.toggleOff}
-          onChange={value => update(NumUtils.round(value, 2))}
-          initialValue={NumUtils.round(tax, 2)}
+          Input={
+            <NumericTextInput
```
mark-prins, line 36: this tax edit modal behaves differently - it allows entering this sort of thing

Screenshot 2024-01-24 at 4 07 20 PM

which then results in this when you save and re-open

Screenshot 2024-01-24 at 4 04 40 PM

the current numeric input does not allow the invalid numbers

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
```diff
@@ -229,11 +230,12 @@ export const PrescriptionLineEditForm: React.FC<
           />
           <Grid container>
             <ModalLabel label={t('label.issue')} />
-            <NonNegativeIntegerInput
+            <NumericTextInput
               autoFocus
               value={issueQuantity}
               onChange={handleIssueQuantityChange}
-              defaultValue={0}
+              defaultValue={1}
```
mark-prins, line 237: the defaultValue doesn't show initially - when I add an item the Issue quantity is `0`
Which is good - it actually shouldn't default to `1` at that point - if it defaults to a valid number then we don't know if the user has forgotten to enter an issue quantity and cannot alert. we'd save the prescription line with a quantity of 1 which might not be correct

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -0,0 +1,60 @@
...
+export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
+  (
+    {
+      sx,
+      InputProps,
+      width = 75,
+      onChange,
+      defaultValue,
+      allowNegative,
+      min = allowNegative ? -Infinity : 0,
```
mark-prins, line 25: Rather than infinity, it would be safer to use `NumUtils.MAX_SAFE_API_INTEGER`
The infinite value, or the max value for an integer would both cause an api exception to be thrown

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -54,7 +54,7 @@ const UIComponent = (props: ControlProps) => {
     handleChange(estimatedPath, false);
   };
 
-  const onChangeAge = (newAge: number) => {
+  const onChangeAge = (newAge: number | string = '') => {
```
mark-prins, line 57: this should be `const onChangeAge = (newAge: number | undefined)` to match the `NumericTextInput` which does not have string as a possible input for the `onChange` prop.

This method then has to handle the undefined possibility

mark-prins (changes requested):
The new version of `NumericTextInput` is very neat! nice work in consolidating the various implementations and usages 😄 

There's a few instances where the behaviour differs from the current which need fixing up.
The ones I've got are
* outbound line edit modal ( can enter `40-50` as number of packs to issue; In store and Available should not be enterable; pack qty can type in characters and get NaNs  )
* outbound detail view table ( pack qty column is editable and should be text )
* outbound tax edit ( can type text which doesn't get saved but results in undefined value; can save 100-10 etc )
* inbound line edit modal ( you can type text in # packs which give NaN in Unit Qty )
* inbound detail view table ( the pack qty column should not be editable )
* inbound tax edit ( can type text which doesn't get saved but results in undefined value; can save 100-10 etc )
* Prescription detail view (pack qty should not be editable)

CarlosNZ:
> The number input cell is now editable - and should not be:

Oops, yeah -- I'd overzealously assumed these cells were all using the NumericInput, but they were actually just display-only cells. I've restored the "PositiveNumberCell" component, but renamed it to just "NumberCell" since there was nothing explicitly "positive" about it.

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
CarlosNZ: Yeah, the `defaultValue` will only show up if the `value` prop is not yet defined. And in this case, it's being set to `0` in the initial `useState`, so no need for `defaultValue` here at all. Have removed :)

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
CarlosNZ: Yeah, the `NumericTextInput` now explicitly requires a `number | undefined` type, so all these cases where the component was having to handle empty strings as well as number have now been simplified :)

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
CarlosNZ: Agree, have renamed to `decimalLimit`

client/packages/common/src/ui/components/inputs/TextInput/TextInput.stories.tsx
CarlosNZ: Done 👍🏼

CarlosNZ:
> The number input cell is now editable - and should not be:

Oops, yeah -- I'd overzealously assumed these cells were all using the NumericInput, but they were actually just display-only cells. I've restored the "PositiveNumberCell" component, but renamed it to just "NumberCell" since there was nothing explicitly "positive" about it.

> this tax edit modal behaves differently - it allows entering this sort of thing

Yup, those should be behaving nicely now with the updated NumbericTextInput with no `type="number"` attribute.

---

Okay, @mark-prins, have re-factored the `NumericTextInput` to use a standard "text" input, and included proper validation and formatting, including localisation of decimals and thousands separator. I think it's much better now.

### Features
- We've lost the stepper, but I've added a keyboard stepper 😄 (cursor-up/down), including multiplier (hold Shift)
- You shouldn't be able to enter any invalid characters anywhere:
  - "-" only allowed at start, and only if min is below 0
  - "." only allowed once, and only if decimalLimit > 0
  - "," separator allowed (but ignored internally)
  - no other characters
- Formatting as you type (including thousands separator for large numbers)
- Localised (decimal and separator character follows language selection)
- 2 additional props (one of which was previously part of standard "number" type): `step` and `multiplier` (for keyboard stepper)
- Made a few tweaks to the currency and IntlNumberUtils, see inline comments

### Questions
- For languages that use something other than "." for decimal, do you think we should continue to accept "." input as well, in case users are just used to entering that way?

### Future improvements
(i.e. things that would be nice but this is getting a little out of hand already so will save for another day)
- more formatting options (including currency, etc -- basically, anything that can be done by `Intl.NumberFormat` should be able to be specified
- integrate with currency input (they're really just variations of the same thing, especially with the aforemententioned formatting options)
- able to specify a minimum decimal places to display (e.g. show "5.0" rather than just "5")
- integrate with Mui "NumberInput" once it matures a bit more

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+  ) => {
+    const { format, parse } = useFormatNumber();
+    const {
+      options: { separator, decimal },
+    } = useCurrency();
+    const [textValue, setTextValue] = useState(format(value ?? defaultValue));
+
+    useEffect(() => {
+      setTextValue(format(value));
+    }, [value]);
```
CarlosNZ, line 47: So we manage two representations of the number: `value`, which is passed in as a prop and updated via `onChange()` is the actual number, `textValue` is the current formatted and localised string representation.

I wouldn't have used an additional `useState` value for the string representation -- we should always be able to derive it from `value`. However, we need to manage it separately, since user needs to be able to type "-" and "." even though that's not enough to update the number yet, so `textValue` will be set independently to `onChange` when this happens.

client/packages/common/src/ui/components/modals/InputModal/InputModal.stories.tsx
CarlosNZ: Have renamed to just "NumberInputModalExample"

mark-prins (changes requested):
that's looking great - so much better to have some control of the input and detect invalid characters

I have a couple of instances where handling of no value is making behaviour strange. 
Can you have a look at these two ( applies to outbound and inbound shipments )

The validation on the edit line is detecting a `0` value and warning the user.. but you get no warning if you delete what is in the issue quantity input. The modal just closes. It should warn the user like the zero does:

https://github.com/msupply-foundation/open-msupply/assets/9192912/c437877c-298f-4c27-a609-4de24784cace

You can clear the quantity for a line, and mess with the calculations:

https://github.com/msupply-foundation/open-msupply/assets/9192912/3f4b5b48-cdee-4229-87e8-844b40bb1547

mark-prins:
Sorry, just noticed something while testing a different PR. Created an inbound shipment, added an item, and I cannot enter a number of packs.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+          '& .MuiInput-input': { textAlign: 'right', width: `${width}px` },
+          ...sx,
+        }}
+        InputProps={InputProps}
+        onChange={e => {
+          const input = e.target.value
+            // Remove separators -- using split/join as .replaceAll() not
+            // supported
+            .split(separator)
+            .join('')
```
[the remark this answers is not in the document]
CarlosNZ: The reason I didn't do that was not all separator characters need to be escaped -- `.` does, but not `,` or ` `. Although I now see that an escaped comma doesn't actually hurt, it still works.

You *can* break it if the separator is a number, as `\3` represents a character code not an escaped `3`, although I can't see any reason why we'd need that. So... any chance that any escaped separator would be a problem here? Guessing not, so yeah, will change.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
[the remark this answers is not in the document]
CarlosNZ: This is one "rule" I often see that I don't really understand -- it often seems to cause more problems than it solves. In this particular case, `format` is a _static_ method returned from `useFormatNumber` and it loads before the `useEffect` is first run. We know it can't be changed, so why do we need to include it in the dependencies?

And in fact, in this case, it actually breaks the functionality -- the `useEffect` runs twice after each change, with one of them getting an `undefined` value -- the outcome is that the typed decimal input (before subsequent digits are added) gets immediately wiped out, so can't add a decimal.

Do you know what the reason for this "rule" is, as I think the dependency array should be as small as possible to limit re-renders, and putting in things that we know won't change seems wasteful?

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
```diff
@@ -10,24 +10,28 @@ export const NumberInputCell = <T extends RecordWithId>({
...
-        const newValue = e.target.value;
+      onChange={num => {
+        const newValue = num;
         setBuffer(newValue);
         updater({ ...rowData, [column.key]: Number(newValue) });
       }}
+      // Make the default min=1 as this is the typical implementation
+      // in Data Tables
+      min={1}
+      {...props}
```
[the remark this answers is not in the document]
CarlosNZ: I wanted to be able to allow any of the additional `StandardTextFieldProps` (passed through in `NumericTextInput` to be passed through if desired. But I see what you mean about the CellProps though.

Don't suppose there's a nice way to remove unwanted `CellProps` before passing any other text field props through is there, without naming every possible one?

Also, looking at what's in `CellProps`, we'd probably want to pass `isDisabled`, `isRequired` and `isError` through too, right?

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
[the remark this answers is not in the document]
mark-prins: if the function executed by useEffect is using functions or variables and those change, but aren't part of the dependencies, then the effect isn't re-run with the modified values. that's what the linter is trying to catch. and it's a valid concern.

I've found that blindly adding in all dependencies to calm the linter has caused problems though, with the effect being called more than expected. In this case it's good to be explicit and to comment why the deps array is limited. If nothing else, adding the linter option has provided motivation for thinking more carefully about dependencies. I think it's good to have the discussion

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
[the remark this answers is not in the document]
CarlosNZ: Okay, I've refactored it to this, which I think is the best I can do without unpicking several of the parents and descendants of `NumberInputCell`: 36c01899d

I've reduced the type of `NumericTextInputProps` to *only* have the numeric specific props, and created a union type within NumericTextInput to combine them with `StandardTextFieldProps`. So then the other components using `NumericTextInputProps` won't be expecting any TextField props as well.

We can lose the additional `TextInputProps` if we want, it's just their for completeness. 🤷‍♂️

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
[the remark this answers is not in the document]
CarlosNZ: Okay @mark-prins , have added comment for now: 4a7596d15

> I've found that blindly adding in all dependencies to calm the linter has caused problems though, with the effect being called more than expected.

Yes, that's exactly what was happening.

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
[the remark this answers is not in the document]
mark-prins: Nice! that looks good. Can you remove the `{...props}` though (line 62) - I'm still getting console errors.

have retested inbound shipments (quantity entry, defaulting to `0`) and outbound shipments (defaulting to `0`)

Two things left then:
- Prescriptions have the same issue, where you can clear the input and get NaNs
- the console errors

image

🙏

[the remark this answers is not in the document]
CarlosNZ: Righto, that should be sorted now: e3a409b99

Thanks for the thorough checks. Lots of implementations (and all _slightly_ different...)

mark-prins (approved):
🎉  awesome!! 
thanks for persevering with the change requests, it's a component that is used all over the place, in slightly different ways as you say. I think that we've covered it pretty well now.

PR #2796, opened 2024-01-23
Android back button

### LINKED ISSUE #2611: Android in kiosk mode back button issue
## What went wrong? 😲
 
We are facing an issue when we click back button multiple times our open msupply apps get stuck and no way to restart it so we have to restart the tablet completely… and users are very good in pressing back button multiple times. 

This case is found while tablets are in kiosk mode but in normal tablets it’s working perfectly… any advice on it is it something with kiosk or our apps?

## Expected behaviour 🤔
Back button shouldn't cause an issue. what's up with kiosk mode?

## How to Reproduce 🔨

## Your environment 🌱

 - Version: 
 - Platform:
   - [ ] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

mark-prins:
Fixes #2611

# 👩🏻‍💻 What does this PR do? 
 
Added the call to `App.exitApp` and also moved the back button handler out of `useNativeClient`. It is actually fine to leave it there, but I wanted to call the back button handler explicitly, and to call it from the host package, not the common package. Lifting it out makes the handler clearer.

Have tested that clicking 'back' when you are on these pages will close the app:
- Discovery
- Initialisation
- Login
- Android

and so, if you click back multiple times, when you hit the end of the navigation stack, the app is closed. It prevents the 'getting stuck' issue and is a more natural behaviour.

There is a thing - after the app exits when you reopen the app it is in a strange state. I've spent too long looking at this already and couldn't see a simple way to resolve that one, so it is still an issue. After a time the app simply closes; alternatively you can manually unload the app (task switcher, remove) and then it works fine.

Here's an apk file for testing: `open-msupply-1.6.00-debug.apk` in this [google drive](https://drive.google.com/open?id=1Ax4J-x4-ZXbyOzpM_VixVGQ9WxW4jn2Q&usp=drive_fs)

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
+++ b/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
@@ -1,0 +1,31 @@
+import { useEffect } from 'react';
+import { App } from '@capacitor/app';
+import { useNavigate } from '@openmsupply-client/common';
+import { Capacitor } from '@capacitor/core';
+
+export const useBackButtonHandler = ({
+  isNavigateEnabled,
+  dependencies,
+}: {
+  isNavigateEnabled: boolean;
+  dependencies?: unknown[];
+}) => {
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    if (Capacitor.isNativePlatform()) {
+      App.removeAllListeners();
+
+      App.addListener('backButton', ({ canGoBack }) => {
+        if (canGoBack && isNavigateEnabled) navigate(-1);
+        else App.exitApp();
+      });
+    }
+
+    return () => {
+      if (Capacitor.isNativePlatform()) {
+        App.removeAllListeners();
+      }
+    };
+  }, [isNavigateEnabled, navigate, dependencies]);
+};
--- a/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
+++ b/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
@@ -16,14 +16,16 @@ export const useBackButtonHandler = ({
 
       App.addListener('backButton', ({ canGoBack }) => {
         if (canGoBack && isNavigateEnabled) navigate(-1);
-        else App.exitApp();
+        else {
+          App.minimizeApp();
+        }
       });
     }
 
-    return () => {
-      if (Capacitor.isNativePlatform()) {
+    if (Capacitor.isNativePlatform()) {
+      return () => {
         App.removeAllListeners();
-      }
-    };
+      };
+    }
   }, [isNavigateEnabled, navigate]);
 };
--- a/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
+++ b/client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
@@ -5,10 +5,8 @@ import { Capacitor } from '@capacitor/core';
 
 export const useBackButtonHandler = ({
   isNavigateEnabled,
-  dependencies,
 }: {
   isNavigateEnabled: boolean;
-  dependencies?: unknown[];
 }) => {
   const navigate = useNavigate();
 
@@ -27,5 +25,5 @@ export const useBackButtonHandler = ({
         App.removeAllListeners();
       }
     };
-  }, [isNavigateEnabled, navigate, dependencies]);
+  }, [isNavigateEnabled, navigate]);
 };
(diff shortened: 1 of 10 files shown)


client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+}) => {
+  const navigate = useNavigate();
+
+  useEffect(() => {
+    if (Capacitor.isNativePlatform()) {
+      App.removeAllListeners();
+
+      App.addListener('backButton', ({ canGoBack }) => {
+        if (canGoBack && isNavigateEnabled) navigate(-1);
+        else App.exitApp();
```
mark-prins, line 19: this is the key change - exiting rather than doing nothing

client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+      });
+    }
+
+    return () => {
+      if (Capacitor.isNativePlatform()) {
+        App.removeAllListeners();
```
mark-prins, line 25: I did play around with specific handles but have had so much trouble getting this tested, I've given up and removed them all.

The `addListener` call does return a promise that holds an object with a `remove` function, so I assigned that to a ref and called that. It might still work, but as I say, I've spent long enough on this.

client/packages/common/src/hooks/useNativeClient/useNativeClient.ts
```diff
@@ -104,7 +101,7 @@ export const useNativeClient = ({
     }));
 
     nativeAPI.startServerDiscovery();
-  };
+  }, [discovery, nativeAPI]);
```
mark-prins, line 104: offroads - this is the linter telling me to do stuff

client/packages/common/src/hooks/useNativeClient/useNativeClient.ts
```diff
@@ -214,7 +199,7 @@ export const useNativeClient = ({
     if (previousServer === null) return;
 
     connectToPrevious(previousServer);
-  }, [state.previousServer, autoconnect]);
+  }, [state.previousServer, autoconnect, state, nativeAPI, connectToPrevious]);
```
mark-prins, line 202: offroads

client/packages/common/src/hooks/useNativeClient/useBackButtonHandler.ts
clemens-msupply: Nit return the clean up method only in `if (Capacitor.isNativePlatform()) { }`

Please ignore me if you think this would require further testing...

clemens-msupply (commented):
I can pretty consistently crash the app by:
- Going back on the login page
- Restarting the app -> crash in the remote server startServer method...

Is this what you are seeing as well? Tried a bit of debugging but haven't found the reason yet.

clemens-msupply:
Regarding the crash it seems to be working fine in server mode (previously I had it in client mode)

mark-prins:
@clemens-msupply thanks for reviewing!
If the app is crashing in client mode: For the server that the tablet is connecting to - have you built the front end and recompiled the server? The change is in the front end and the front end it hosted by the server which the tablet is connecting to, and not the code the tablet is running.

client/packages/host/src/components/Login/Login.tsx
```diff
@@ -52,6 +51,7 @@ export const Login = () => {
   useEffect(() => {
     setPageTitle(`${t('app.login')} | ${t('app')} `);
     LocalStorage.removeItem('/error/auth');
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
mark-prins: Confused: I had an issue where the effect was being called several times, but I'm not seeing that now. I've update with the correct dependencies thanks

clemens-msupply:
> @clemens-msupply thanks for reviewing! If the app is crashing in client mode: For the server that the tablet is connecting to - have you built the front end and recompiled the server? The change is in the front end and the front end it hosted by the server which the tablet is connecting to, and not the code the tablet is running.
> 
> Have retested in client mode and it's working for me

I just tried again. I see the same behaviour as you described. After the first backbutton close, the app startup screen shows for a while then the app just closes.

Then restarting the app works fine. But after doing it again (backbutton + restart) the app crashes (remote server crash). Next start works again...

What do you think should we investigate more? or create a new issue?

mark-prins:
I'll have another look - I did spend some time trying to find the cause and didn't find it.

mark-prins:
Thanks @clemens-msupply - I've made a change to minimize the app rather than exit, and this resolves the crash.
I've also done the `useEffect` refactor.

The crash is related to the remote_server still running - I changed to stop the server when exiting the app, and that changes behaviour but still doesn't fix. whereas the minimise works nicely.

clemens-msupply (approved):
Great! Minimizing works here (though I actually tried it when debugging last time but apparently not)

Issue #2759, opened 2024-01-17
Merging: use proper link name instead of just renaming the column


clemens-msupply:
## Is your feature request related to a problem? Please describe 👀
After the first merge of the `feature/merging` branch column renames should be removed, e.g:

```rust
    #[column_name = "name_link_id"]
    pub name_id: String,
```

The reason is that all query method exped a `name_id` but with the current code we use a `name_link_id`. Thus a query would fail for merged names/items.

Same changes should be applied for item_id/item_link_id

## Moneyworks Jobcode 🧰

OMS:MERGE

Chris-Petty:
I think I'll close this and make issues for all the other areas so we can track/split better

PR #2803, opened 2024-01-23
Rename name_link_id and clinician_link_id for the InvoiceRow

clemens-msupply:
Part of #2759

Mostly testing files where I just initialized the name_link_id column with the proper name.id.

I sometimes also just passed on existing link_id where I thought it's safe to do so, because otherwise I would have to do another DB call... Please review carefully.

Fixed some clippy warnings along the way...

### DIFF
--- a/server/service/src/invoice_line/inbound_shipment_line/insert/generate.rs
+++ b/server/service/src/invoice_line/inbound_shipment_line/insert/generate.rs
@@ -37,7 +37,7 @@ pub fn generate(
             &existing_invoice_row.store_id,
             new_line.clone(),
             false,
-            &existing_invoice_row.name_id,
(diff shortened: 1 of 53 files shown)


server/service/src/invoice_line/inbound_shipment_line/insert/generate.rs
```diff
@@ -37,7 +37,7 @@ pub fn generate(
             &existing_invoice_row.store_id,
             new_line.clone(),
             false,
-            &existing_invoice_row.name_id,
```
clemens-msupply, line 40: Used to initialised another name_link_id...

roxy-dao (approved):
Thanks for all the changes Clemens 🙏🏻

PR #2806, opened 2024-01-23
2795 create currency table and sync

### LINKED ISSUE #2795: Create `Currency` table and sync
Sync the `currency` table to support multi-currencies

- Create currency table
- Sync

roxy-dao:
Fixes #2795

# 👩🏻‍💻 What does this PR do? 
Creates the `currency` table and sync table with central. 

- Run tests
- 
Sync will probably need to be updated with new sync refactoring


server/repository/src/db_diesel/currency_row.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::repository_error::RepositoryError;
+
+use chrono::NaiveDate;
+use diesel::prelude::*;
+
+table! {
+    currency (id) {
+        id -> Text,
+        rate -> Double,
+        currency_code -> Text,
```
roxy-dao, line 12: Seems like `currency` in central saves currency codes, so have renamed it (and also column can't be same as table name, will still be something similar to currency_code even with `sql_name`)

server/repository/src/db_diesel/currency.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::{EqualFilter, Pagination, Sort};
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct Currency {
+    pub currency_row: CurrencyRow,
+}
+
+#[derive(Clone, PartialEq, Debug)]
+pub struct CurrencyFilter {
+    pub id: Option<EqualFilter<String>>,
```
mark-prins, line 21: Probably useful to have `is_home_currency` and `is_active` as filterable too
`is_active` doesn't appear to be used in mSupply afaics, but if we have the column, we would want to filter by that.
My suggestion:
(a) change the default to `true` and add to the filter, or
(b) remove the `is_active` column, or
(c) forget I said anything, keep the column, keep it defaulting to false and don't filter by it!

server/repository/src/db_diesel/currency.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn query_by_filter(
+        &self,
+        filter: CurrencyFilter,
+    ) -> Result<Vec<Currency>, RepositoryError> {
+        self.query(Pagination::all(), Some(filter), None)
+    }
+
+    pub fn query(
+        &self,
+        pagination: Pagination,
```
mark-prins, line 55: 🤔  maybe pagination isn't required for this one? there won't be very many currencies, and when we want to get them, we would want all available currencies

server/repository/src/db_diesel/currency_row.rs
mark-prins: good call! this could even be just `code`

mark-prins (approved):
👍  that's great thanks!
I have some minor suggestions to consider - that aside, it's working nicely and looks good to me

PR #2809, opened 2024-01-23
Rename name_link_id for RequisitionRow

clemens-msupply:
Part of #2759

Mostly tests again but a couple of place to look more careful at...


server/service/src/requisition/common.rs
```diff
@@ -2,16 +2,28 @@ use repository::{
...
 }
 
+pub fn check_requisition_exists(
+    connection: &StorageConnection,
+    id: &str,
+) -> Result<Option<Requisition>, RepositoryError> {
+    Ok(RequisitionRepository::new(connection)
+        .query_by_filter(RequisitionFilter::new().id(EqualFilter::equal_to(id)))?
+        .pop())
+}
```
clemens-msupply, line 25: I needed the Requisition for one case but the existing method was used quite a lot. For this reason I renamed the existing one and create a new one that returns the Requisition (containing the NameRow)

roxy-dao (approved):
Thanks Clemens~

PR #2810, opened 2024-01-24
2807 query to get currencies

### LINKED ISSUE #2807: Query to get currencies
Need query for getting all currencies and getting one

roxy-dao:
Fixes #2807

# 👩🏻‍💻 What does this PR do? 
Query to get one currency or all currencies... though if we don't allow editing for this right now, probably don't need the one currency query 👀 

- Use `graphiql` or something similar to call endpoints


server/graphql/general/src/lib.rs
```diff
@@ -295,6 +299,24 @@ impl GeneralQueries {
...
         )
     }
+
+    pub async fn currency(
+        &self,
+        ctx: &Context<'_>,
+        currency_id: String,
+    ) -> Result<CurrencyResponse> {
+        get_currency(ctx, &currency_id)
+    }
```
clemens-msupply, line 309: If we don't need it maybe not add it for now? at least not on the graphql level? (hard to remove api endpoints without breaking potential users...)

server/graphql/general/src/queries/currency.rs
```diff
@@ -0,0 +1,50 @@
...
+        Some(currency) => CurrencyResponse::Response(CurrencyNode::from_domain(currency)),
+        None => CurrencyResponse::Error(NodeError {
+            error: NodeErrorInterface::record_not_found(),
+        }),
+    };
+
+    Ok(response)
+}
+
+pub fn get_currencies(
```
clemens-msupply, line 29: nit: just call it `currencies` ?

server/graphql/types/src/types/currency.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.row().date_updated
+    }
+
+    pub async fn is_active(&self) -> bool {
+        self.row().is_active
+    }
+}
+
+#[derive(Union)]
+pub enum CurrenciesResponse {
```
clemens-msupply, line 87: These query endpoint specific types (response, sort and filter input) should be specific to the `currency`/`currencies` endpoint so they don't need to be here but can go into the query endpoint file.

What about the CurrencyNode? Do you plan to use it in other graphql crates? Otherwise, I would also keep them in the currency query file...

The types crate was just for types that need to be shared between graphql crates...

clemens-msupply (changes requested):

server/graphql/types/src/types/currency.rs
roxy-dao: Ah right. Yeah, I don't think it'll be used anywhere else. 

Thanks for the review Clemens~ all fixes can be found here [d6ff0ea](https://github.com/msupply-foundation/open-msupply/pull/2810/commits/d6ff0ea1aad359e9541083e09764b91014a4143c)

clemens-msupply (approved):
Thank you for the changes!

Issue #608, opened 2022-09-13
Sync cary over 14-09


andreievg:
# Sync Cary Over 14/09

Carry over from https://github.com/openmsupply/open-msupply/issues/156

## Sync v5 branch

- [ ] Update form dev and merge
- [ ] Make sure all TODO in diff in this [PR are summaries](https://github.com/openmsupply/open-msupply/pull/587) before merge

## Questions

* Is it possible to change visibility of name to store in mSupply withouth deleting it ?
* How does pricing work in transfers ?

## Safety

- [ ] Remote data deletes are only allowed for `foreign` records
- [ ] Owned data integrated only during initialisation ?
- [ ] Extra protection in transfer processors for possible circular updates (inifinte looping), some [comments here](https://github.com/openmsupply/open-msupply/pull/579#discussion_r969158425)

## Manual Tests

- [ ] Write up manual cases

## Edge Cases

- [ ] When name_id is changed for shipment, and it's already synced and other half created (outbound shipment in picked status and inbound shipment created), this would cause original inbound shipment to be orphaned in picked status. For name_id changes after picked status should delete and re-created outbound shipment (in service). So another mutation to change name_id (https://github.com/openmsupply/open-msupply/issues/188)
- [ ] It's possible to have stock for an item that is not in master list (shoudl be able to still issue and do stocktake for these items)
- [ ] If there are deletes for transfers in sync queue for foreign site on central server, and that foreign site is `re-initialised`, sync queue will be cleared and we loose the record of those deletes (is the same true for merge records ?)

## Improvements

- [ ] Use cursors for remote sync instead of acknowledgement (this is already implemented in mSupply), and align central and remote data shape in mSupply (also use cursor instead of id)
- [ ] Batch size to be adjustable (setting via mutation). We have more advanced batch size handling in mSupply Mobile, I am not sure if complexity justifies it (vs just increasing batch size)
- [ ] Transfer itegration tests across more then 2 sites ?
- [ ] Quite often we need to know if store and name is active on site, which requires database query, ideally this should be available in memory store (ideally). Transfers check this and invoice_line translator checks this.
- [ ] Should system user be used/displayed for transfer creation/edit ?
- [ ] It would be good if constraints are always enforced (for linked requisitions and linked invoices, extra `circular` translator that adds these, vs removing referencial constraints, does this work for deletes ?)

## Loggin/Errors

Need to have a discussion of logging and error handling some questions to answer
[...]

PR #2818, opened 2024-01-24
Central config server transition

### LINKED ISSUE #2062: omSupply Central Server Transition
# Central Server Background

We use classic mSupply central server for 
* Configurations (store, items, master lists, etc..)
* Synchronisation (central sync hub for remote site, transfer of records between sites, backup, etc..)

When omSupply development started, we decided that it was much faster to use existing mSupply central server rather then implement new omSupply central server. This option offered other benefits, like being able to use mSupply and omSupply in the same system.

# Motivation

1. We are planning to replace mSupply central server with omSupply central server at some stage
2. New configuration UI needs to be added to mSupply for new omSupply only features (this would need to be done twice when omSupply is implemented), i.e. document/program configurations, multi pack units, asset module config.
3. New omSupply only supervisor (all store) mode features may need to be added to mSupply, i.e. global program/document reports, global asset and cold chain performance dashboard/reports ?

Although our existing strategy is to do `3` in Graphana dashboard, having these indicators directly in omSupply adds an option that may be more desirable in some use cases.

# Proposal

omSupply site can act as `central configuration` and `central data hub` for a subset of <ins>new record</ins> types. It would run in parallel with exiting mSupply remote server. omSupply remote sites would sync with both mSupply and then omSupply central server

Some research was done in [this trello card](https://trello.com/c/AJk8pLkh/26-omsupply-central-parallel) (the card and the issue will be updated with details)

<ins>new record</ins> types: It should be possible to transition to old records too, but this brings it's own challenges, for now focus of this issue is for new record types. 

# Examples

### Mutli Unit Configurations

Multi unit configuration UI can be added to omSupply central, and will sync to omSupply remote sites.

### All store mode <a href="#all-store-mode" id="all-store-mode"></a>

Any mSupply instance can access data for all stores via central omSupply API. When data for all stores needs to be accessed, it can be routed to central omSupply API rather then using current remote omSupply server. Non active store data can only be accessed when there is internet connection.

This option would require more consideration (i.e. front end consumable API of central server needs to be backwards compatible)

# Challenges

* All data needs to be present in omSupply central to make decisions about "shared" (patient related) record type routing (i.e. all name_store_joins). For new central data or new remote data this should not be a problem, since routing is pretty simple.

andreievg:
Fixes #2062

# 👩🏻‍💻 What does this PR do? 

* Creates a REST endpoint for central server sync API (just pull for now)
  * Authentication of site is proxied to original mSupply central server (which requires some refactoring to make SyncSettings serialisable and sharable between omSupply instances)
* Adds a check for IS_CENTRAL_SERVER env variable (to be discuss further in carry over issue) and uses it various places
* Translation for pack variant and adds pack variant to changelog
* Adds a pull from omsupply central to existing syncrhoniser
* Reshapes sync api error and push records shape to allow for existing changelog to push record to also be used by central server (with existing translators)
* Push translator now get a enum passed through to differentiate between Legacy, OmSupplyRemote/CentralSitePush (where OmSupplyCentralSitePush is for central server sync API, when central server responds to pull request) 
* Create a common cursor control (there were a few implementations of this, decided to consolidate to one common one)

I think it's worth looking at follow up [PR with doc](https://github.com/msupply-foundation/open-msupply/pull/2820) additions to understand how this work, briefly:

omSupply central server is just another remote site, but it should be hosted (or routed) on the same machine as original mSupply central, with port + 2 of original mSupply port (this is up for contention, the simples I thought for now). IS_CENTRAL_SERVER env variable is used to control if central server functionality should be on or off.

When omSupply remote site asks omSupply central for records, it passes a cursors, which is used by omSupply central to figure out what needs to be pushed from it's change log. Existing translators are used (since push translation will be the same as what is returned by pull, when the times comes to implement remote records).

Translations between omSupply central and remote are just done with serde.

Start omSupply with port + 2 of original mSupply port, and with IS_CENTRAL_SERVER env variable (can look at [integration test in follow up PR](https://github.com/msupply-foundation/open-msupply/pull/2819), there is a command line there). Start another mSupply instance normally. Change pack variants in omSupply central, they should sync to mSupply remote (note, for now it's possible to edit pack variants in omSupply remote, but it will be disabled)

There are a few carry overs and consideration to discuss in a meeting, they are in [carry over issue](https://github.com/msupply-foundation/open-msupply/issues/2821), under "Questions" section.

I think if i was reviewing this i would look at sync_on_central files, then v6 files, then the changes in translations and renaming should make more sense. I can also go over the PR in person.

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


server/graphql/general/src/queries/sync_status.rs
```diff
@@ -61,6 +61,7 @@ pub struct FullSyncStatusNode {
     prepare_initial: Option<SyncStatusNode>,
     integration: Option<SyncStatusNode>,
     pull_central: Option<SyncStatusWithProgressNode>,
+    pull_v6: Option<SyncStatusWithProgressNode>,
```
andreievg, line 64: This is not added to front end yet, but in carry over issue

server/graphql/general/src/sync_api_error.rs
```diff
@@ -2,7 +2,7 @@ use async_graphql::*;
 use graphql_core::standard_graphql_error::StandardGraphqlError;
 use repository::SyncLogRowErrorCode;
 use service::sync::{
-    api::{SyncApiError, SyncApiErrorVariant, SyncApiV5CreatingError, SyncErrorCodeV5},
+    api::{SyncApiError, SyncApiErrorVariantV5, SyncApiV5CreatingError, SyncErrorCodeV5},
```
andreievg, line 5: Changed since it's also used in V6 now

server/postman/Open mSupply REST.postman_collection.json
```diff
@@ -0,0 +1,44 @@
+{
```
andreievg, line 1: Check out postman collection to see the shape of request/response

server/repository/src/db_diesel/key_value_store.rs
```diff
@@ -185,3 +186,25 @@ impl<'a> KeyValueStoreRepository<'a> {
         Ok(row.and_then(|row| row.value_bool))
     }
 }
+
+pub struct CursorController(KeyValueType);
```
andreievg, line 190: This was refactored and applied to places that use cursors, sorry this was done quite early on, I probably should have done after this PR, it adds a little bit of noise to the PR

server/repository/src/db_diesel/pack_variant_row.rs
```diff
@@ -18,9 +19,21 @@ table! {
...
+    Default,
+    Eq,
+    Ord,
+    PartialOrd,
+    Serialize,
+    Deserialize,
 )]
 #[table_name = "pack_variant"]
+#[serde(rename_all = "camelCase")]
 pub struct PackVariantRow {
```
andreievg, line 37: Pack variant is translated purely with serde (well apart from SyncTranlsation implementation for dependencies and to indicate which sites should push it

server/repository/src/migrations/mod.rs
```diff
@@ -89,6 +90,7 @@ pub fn migrate(
         Box::new(v1_05_00::V1_05_00),
         Box::new(v1_05_04::V1_05_04),
         Box::new(v1_06_00::V1_06_00),
+        Box::new(v1_07_00::V1_07_00),
```
andreievg, line 93: Added v1_07_00 translations to pack variant, and moved all pack_variant translations there

server/repository/src/migrations/v1_07_00/central_omsupply.rs
```diff
@@ -0,0 +1,61 @@
...
+                ALTER TYPE key_type ADD VALUE IF NOT EXISTS 'SYNC_PULL_CURSOR_V6';
+            "#
+        )?;
+    }
+
+    if cfg!(feature = "postgres") {
+        sql!(
+            connection,
+            r#"
+                ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'pack_variant';
```
andreievg, line 27: note, no deletion triggers

server/service/src/lib.rs
```diff
@@ -244,6 +244,10 @@ pub fn i32_to_u32(num: i32) -> u32 {
     num.try_into().unwrap_or(0)
 }
 
+pub fn i64_to_u64(num: i64) -> u64 {
```
andreievg, line 247: used for cursor translation when sending in API

server/service/src/sync/api/common_records.rs
```diff
@@ -9,24 +9,22 @@ fn empty_object() -> serde_json::Value {
 }
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
-pub(crate) struct CommonSyncRecordV5 {
-    #[serde(rename = "tableName")]
+#[serde(rename_all = "camelCase")]
```
andreievg, line 12: I was changing SyncAction from SyncActionV5, and decided to just use rename_all, another sidetrack sorry

server/service/src/sync/api/core.rs
```diff
@@ -1,53 +1,36 @@
...
+use reqwest::{header::HeaderMap, Client, Response, Url};
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::json;
 use thiserror::Error;
 use url::ParseError;
 
 use super::*;
 
+#[cfg(target_os = "android")]
+const APP_NAME: &'static str = "Open mSupply Mobile";
```
andreievg, line 17: Added to carry over, to make sure it's Android not mobile

server/service/src/sync/api/core.rs
```diff
@@ -59,25 +42,40 @@ pub enum SyncApiV5CreatingError {
 }
 
 impl SyncApiV5 {
-    pub fn new(
+    pub fn new_settings(
         settings: &SyncSettings,
         service_provider: &ServiceProvider,
         sync_version: u32,
-    ) -> Result<Self, SyncApiV5CreatingError> {
+    ) -> Result<SyncApiSettings, SyncApiV5CreatingError> {
```
andreievg, line 49: SyncApiSettings needs to be sent to omSupply central for authentication (for hardware id and credentials, thus it's shape was refactored and SyncApiCredentials was removed

server/service/src/sync/api/get_site_info.rs
```diff
@@ -24,7 +24,7 @@ impl SyncApiV5 {
     // Get site status
     pub(crate) async fn get_site_info(&self) -> Result<SiteInfoV5, SyncApiError> {
         let route = "/sync/v5/site";
-        let response = self.do_get_no_query(route).await?;
+        let response = self.do_get(route, &()).await?;
```
andreievg, line 27: Side track, done early on

server/service/src/sync/api_v6/core.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    })?;
+
+    Ok(result)
+}
+
+pub(crate) fn get_omsupply_central_url(sync_v5_url: &str) -> Result<Url, ParseError> {
+    let mut url = Url::parse(sync_v5_url)?;
+    // Unwrap is safe unless domain is not http or https, see port_or_know_default()
+    url.set_port(Some(url.port_or_known_default().unwrap() + PORT_OFFST))
```
andreievg, line 58: Yeah, this is hacky, I agree, but alternatives was quite a bit of work ? To be discussed in the carry over issues/question

server/service/src/sync/api_v6/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use super::{
+    api::{
+        CommonSyncRecord, ParsedError, ParsingResponseError, SyncApiError, SyncApiErrorVariantV5,
+    },
+    translations::PushSyncRecord,
+    SyncApiSettings,
+};
+
+#[derive(Deserialize, Debug, Error, Serialize)]
```
andreievg, line 21: Most types here are used in the API and are shared between omSupply central and remote, type safety should be guaranteed for the same version, but need to think about this when version changes

server/service/src/sync/central_data_synchroniser.rs
```diff
@@ -69,38 +73,14 @@ impl CentralDataSynchroniser {
...
     connection
         .transaction_sync(|con| {
             SyncBufferRowRepository::new(con).upsert_one(row)?;
-            CentralSyncPullCursor::new(con).update_cursor(cursor)
+            cursor_controller.update(con, cursor)
         })
         .map_err(|e| e.to_inner_error())
 }
-
-struct CentralSyncPullCursor<'a> {
```
andreievg, line 83: I think this is a good example why I wanted to consolidate cursor controller

server/service/src/sync/remote_data_synchroniser.rs
```diff
@@ -229,14 +240,11 @@ impl RemoteDataSynchroniser {
...
-    let value =
-        KeyValueStoreRepository::new(connection).get_i32(KeyValueType::RemoteSyncPushCursor)?;
-    let cursor = value.unwrap_or(0);
-    Ok(cursor as u64)
-}
-
-fn update_push_cursor(connection: &StorageConnection, cursor: u64) -> Result<(), RepositoryError> {
-    KeyValueStoreRepository::new(connection)
-        .set_i32(KeyValueType::RemoteSyncPushCursor, Some(cursor as i32))
+impl From<PushSyncRecord> for RemoteSyncRecordV5 {
```
andreievg, line 243: I think it's ok to use From when it's used like RemoteSyncRecordV5::from (because can navigate here with click through, rather then the trait)

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        sync_v5_settings,
+    }: SyncRequestV6,
+) -> Result<SyncBatchV6, SyncParsedErrorV6> {
+    use SyncParsedErrorV6 as Error;
+
+    if !is_central_server() {
+        return Err(Error::NotACentralServer);
+    }
+    // Check credentials again mSupply central server
+    SyncApiV5::new(sync_v5_settings)
```
andreievg, line 78: Authenticate against mSupply central

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        .map_err(|e| Error::OtherServerError(format_error(&e)))?
+        .get_site_status()
+        .await
+        .map_err(Error::from)?;
+
+    // TODO Versioning ?
```
andreievg, line 84: Added to carry over

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    let changelogs = changelog_repo.changelogs(cursor, batch_size, filter.clone())?;
+    let total_records = changelog_repo.count(cursor, filter)?;
+    let max_cursor = changelog_repo.latest_cursor()?;
+
+    let end_cursor = changelogs
+        .last()
+        .map(|log| log.cursor as u64)
+        .unwrap_or(max_cursor);
+
+    let records = translate_changelogs_to_push_records(
```
andreievg, line 101: I like how this could be used with minimum changes

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl CentralDataSynchroniserV6 {
+    pub(crate) async fn pull<'a>(
+        &self,
+        connection: &StorageConnection,
+        batch_size: u32,
+        logger: &mut SyncLogger<'a>,
+    ) -> Result<(), CentralPullErrorV6> {
+        let cursor_controller = CursorController::new(KeyValueType::SyncPullCursorV6);
+        // TODO protection from infinite loop
```
jmbrunskill, line 49: Is there an issue for this TODO?

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -0,0 +1,97 @@
...
+                    connection,
+                    &cursor_controller,
+                    &buffer_row,
+                    cursor as u64,
+                )?;
+            }
+
+            cursor_controller.update(&connection, end_cursor + 1)?;
+
+            if is_emtpy && total_records == 0 {
```
jmbrunskill, line 77:
```suggestion
            if is_empty && total_records == 0 {
```

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -0,0 +1,97 @@
...
+
+            logger.progress(SyncStepProgress::PullCentralV6, total_records)?;
+
+            let is_emtpy = records.is_empty();
+
+            for SyncRecordV6 { cursor, record } in records {
+                // Would ideally have  mapping method for this
+                let buffer_row = record.to_buffer_row()?;
+
+                insert_one_and_update_cursor(
```
jmbrunskill, line 67: It's not important for this PR, but I feel like if catch the error instead of using `?` it should be safe to update the cursor only on  error on finishing the batch. I'm not sure if updating the cursor on every record is much overhead but feels like it could slow down the processing which I think could be a bit slow at times?
I guess it's probably more important for things like TemperatureLog Row where we might have a lot of records.

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+
+pub async fn pull(
+    service_provider: &ServiceProvider,
+    SyncRequestV6 {
+        cursor,
+        batch_size,
+        sync_v5_settings,
+    }: SyncRequestV6,
+) -> Result<SyncBatchV6, SyncParsedErrorV6> {
+    use SyncParsedErrorV6 as Error;
```
jmbrunskill, line 72: Why not just  `use SyncParsedErrorV6` personally not a big fan of this re-naming pattern. I guess it makes copy and paste easier?

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    SyncApiV5::new(sync_v5_settings)
+        .map_err(|e| Error::OtherServerError(format_error(&e)))?
+        .get_site_status()
+        .await
+        .map_err(Error::from)?;
+
+    // TODO Versioning ?
+
+    let ctx = service_provider.basic_context()?;
+    // To push or not ?
```
jmbrunskill, line 87: I don't understand this comment?

server/service/src/sync/sync_on_central/mod.rs
jmbrunskill, line 101: It feels a bit weird that this function is called `Pull` but creates `Push` records.
Maybe the function could be called `pull_from_remote` that does sound a bit clunky though.
This is just me sharing my thoughts/feedback, not suggesting we need to change it...

server/service/src/sync/test/pull_and_push.rs
```diff
@@ -67,7 +68,13 @@ async fn test_sync_pull_and_push() {
     check_test_records_against_database(&connection, test_records).await;
 
     // PUSH UPSERT
-    let mut test_records = get_all_push_test_records();
+    let mut test_records = vec![
+        get_all_push_test_records(),
+        get_all_omsupply_ceantral_push_records(),
```
jmbrunskill, line 73:
```suggestion
        get_all_omsupply_central_push_records(),
```

server/service/src/sync/translations/mod.rs
```diff
@@ -208,31 +215,42 @@ impl PushTranslateResult {
...
 }
 
+pub(crate) enum PushTranslationType {
+    // When omSupply remote is pushing to og mSupply central
+    Legacy,
+    // When omSupply remote is pushing to omSupply central
+    #[allow(dead_code)]
+    OmSupplyRemoteSitePush,
+    // When omSupply ceantral is pushing to omSupply remote
+    // this is when omSupply central is responding to pull request from omSupply remote
```
jmbrunskill, line 250: Should comment be after the  `OmSupplyCentralSitePush` a bit confused if this all relates to    `OmSupplyRemoteSitePush`?

server/service/src/sync/translations/mod.rs
```diff
@@ -264,8 +282,14 @@ pub(crate) trait SyncTranslation {
     }
 
     // By default matching by change log type
-    fn match_push(&self, row: &ChangelogRow) -> bool {
-        self.change_log_type().as_ref() == Some(&row.table_name)
+    fn match_push(&self, row: &ChangelogRow, r#type: &PushTranslationType) -> bool {
+        match r#type {
+            PushTranslationType::Legacy => self.change_log_type().as_ref() == Some(&row.table_name),
+            // Have to manually specify
+            PushTranslationType::OmSupplyRemoteSitePush => unimplemented!(),
```
jmbrunskill, line 289: Is `unimplemented!()` a good idea here? Feels a bit risky, but I can see it's maybe better then returning true or false...

jmbrunskill (approved):
Looks good to me, but haven't tested myself.
Figure we can refine/test a bit more as we add new features for CCE for example.

I would be keen to see or perhaps do a test branch where we try out a feature flag in rust rather than `is_central` server, doesn't feel like the `is_central` thing is creating spaghetti code so far.

I don't have my head deep enough into the space, but feels like where remote servers could send records for sites they don't own or something checks for that kind of thing might get a bit messier, but I'm personally happy to go with what you have here. Doesn't feel like it would be too hard to migrate to using a rust feature flag if we wanted too.

server/service/src/sync/api_v6/core.rs
```diff
@@ -0,0 +1,107 @@
+use crate::sync::{api::ParsingResponseError, SyncApiSettings};
+use reqwest::{Client, Response};
+use thiserror::Error;
+use url::{ParseError, Url};
+
+use super::*;
+
+// For now port_offset of 2
+static PORT_OFFST: u16 = 2;
```
jmbrunskill, line 9: Still interested in another solution rather than port offset. Doesn't look too hard to plumb another url through IMHO.
Even if we default to the port offset, but allow it to be set in a configuration that would be a good thing.
I think having a separate port as an option will make dev's life much easier. We don't all want to have to setup Caddy or something...

If rust could be the proxy for old mSupply that would be a good solution. I'm keen for a follow up issue, and happy to experiment with options if you'd like.

clemens-msupply: If I understood correctly, we have the use-case that we need to run the remote server and the new central server on the same machine, and then we need two ports, don't we?

Think as long as we can configure the url/port we can decide later how we deploy it and what is best for a dev setup.

server/service/src/sync/translations/mod.rs
roxy-dao: Also central (:

server/service/src/sync/central_data_synchroniser_v6.rs
andreievg: Nope, this comment has been there already, i just added to [original sync carry over](https://github.com/msupply-foundation/open-msupply/issues/608), thanks

server/service/src/sync/central_data_synchroniser_v6.rs
andreievg: Thanks, cc08fbe03ce8dbc6b91a8bbcf2e97931ef784075

server/service/src/sync/central_data_synchroniser_v6.rs
andreievg: The idea here was we update sync buffer and cursor atomically (if one fails the other fails). I added to this issue: https://github.com/msupply-foundation/open-msupply-internal/issues/39#issuecomment-1940314362

server/service/src/sync/sync_on_central/mod.rs
andreievg: > I guess it makes copy and paste easier

That and less verbose down the line, especially when used in chain mapping with `map_err` (most likely the latter)

server/service/src/sync/sync_on_central/mod.rs
andreievg: Neither do i hehe, i did a self review but missed some things, thanks: dcbd771360dd4f616bbebc49acbd30c5d80634e9

server/service/src/sync/sync_on_central/mod.rs
andreievg: I was hoping the OmSupplyCentralSitePush type would make sense (plus the comments for that variant). 

I also found it confusing.

How about I rename translators and associated methods to: "from_sync_record", "to_sync_record" ?

andreievg: And of course add some comment to the "pull" method here to say it's used by central API in response to a pull records request from remote site

server/service/src/sync/api_v6/core.rs
andreievg: I think the main reason i went with port offset for now, was because of existing sites. How do we point them to the omSupply central server instance unless we caddy or port offset ? 

> I'm keen for a follow up issue, and happy to experiment with options if you'd like.

Sure although I am wondering for this feature to go through, what is the minimum that is needed now, and what is the strategy for existing sites for later ?

Use case, existing deployment, and we upgrade them to omSupply version that requires omSupply central (not the pack variant version but say the assets version). We don't do the port offset thing, or the caddy thing, how do we tell all remotes sites that server is now somewhere else ? 

(I think we would need classic mSupply central to expose API to query for omSupply central server location, and this would happen as the first step of sync)

server/repository/src/db_diesel/key_value_store.rs
```diff
@@ -185,3 +186,25 @@ impl<'a> KeyValueStoreRepository<'a> {
...
+
+impl CursorController {
+    pub fn new(cursor_type: KeyValueType) -> Self {
+        Self(cursor_type)
+    }
```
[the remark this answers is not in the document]
andreievg: Hmm, i think the first is both shorter and clearer, won't change for now, but will take note

server/service/src/sync/api_v6/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[derive(Error, Debug)]
+pub(crate) enum SyncApiErrorVariantV6 {
+    #[error("Connection problem")]
+    ConnectionError(#[from] reqwest::Error),
+    #[error("Could not parse respose")]
+    ParsedError(#[from] SyncParsedErrorV6),
+    #[error("Could not parse respose")]
+    ParsingResponseError(#[from] ParsingResponseError),
```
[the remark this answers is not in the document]
andreievg: > There is a spelling error in response

a64540d2926549d8c4586fa380fced819b0fe2c1, thanks

> The variant naming is inconsistent "Parsed" vs "Parsing"...

I thought they are different, since 'parsing' error is before it's parsed, and parsed error indicates we matched an error from api response. Not 100% happy with those variants, but not sure what's the best alternative

server/server/src/sync_on_central/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use service::{
+    service_provider::ServiceProvider,
+    sync::{
+        api_v6::{SyncPullResponseV6, SyncRequestV6},
+        sync_on_central,
+    },
+};
+
+pub fn config_sync_on_central(cfg: &mut web::ServiceConfig) {
```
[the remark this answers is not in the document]
andreievg: I was wondering how to differentiate between things that are done by central server and remote server sync to central, that's why i did sync_on_central (including folder names in various places).

server/server/src/sync_on_central/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    cfg.service(pull);
+}
+
+#[post("/sync/pull")]
```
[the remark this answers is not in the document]
andreievg: hmm, good point, I think there would be other rest API that are central specific, sounds like we should. thanks
2729e1dea977dc80f2b7a8f055cf3f378ca764f5

server/service/src/sync/api_v6/core.rs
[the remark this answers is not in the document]
andreievg: 73d8b6965f4d6a391c59e330d0efd49ee529fe02

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -0,0 +1,97 @@
...
+                records,
+            } = self.sync_api_v6.pull(cursor, batch_size).await?;
+
+            logger.progress(SyncStepProgress::PullCentralV6, total_records)?;
+
+            let is_emtpy = records.is_empty();
+
+            for SyncRecordV6 { cursor, record } in records {
+                // Would ideally have  mapping method for this
+                let buffer_row = record.to_buffer_row()?;
```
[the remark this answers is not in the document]
andreievg: I am not sure about this comment, removed: 46a019ce175d7b4ede4d0c4bb1251c25de475dea

server/service/src/sync/mod.rs
```diff
@@ -83,3 +86,15 @@ impl ActiveStoresOnSite {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SyncApiSettings {
+    pub server_url: String,
+    pub username: String,
+    pub password_sha256: String,
+    pub site_uuid: String,
+    pub app_version: String,
+    pub app_name: String,
+    pub sync_version: String,
```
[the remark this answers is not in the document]
andreievg: Thanks, yeah better place in API, ideally in some shared place like api/mod, api/v5/, api/v6, that's a bit too much to refactor right now, I think it would also mean we can move some of the errors and sync record shapes to a common place, add to [carry over](https://github.com/msupply-foundation/open-msupply/issues/2821)

`Relocate common API shapes to a shared space`

server/service/src/sync/test/pull_and_push.rs
```diff
@@ -77,14 +84,31 @@ async fn test_sync_pull_and_push() {
...
+    .flatten()
+    .collect::<Vec<PushSyncRecord>>();
+
+    // Combine and sort
     translated.sort_by(|a, b| a.record.record_id.cmp(&b.record.record_id));
     test_records.sort_by(|a, b| a.record_id.cmp(&b.record_id));
 
     // Test ids and table names
-    assert_eq!(
+    pretty_assertions::assert_eq!(
```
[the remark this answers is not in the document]
andreievg: Nope, removed: fd5ea47e32a411a24857234338b9dd7513facdad

server/util/src/lib.rs
```diff
@@ -27,3 +29,9 @@ pub use error::*;
 
 mod file;
 pub use file::*;
+
+pub fn is_central_server() -> bool {
```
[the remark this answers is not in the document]
andreievg: It was a bit of a mission passing it around, and I wasn't yet sure how this would be specified (via feature flag, env variable, a settings that syncs etc..). Also was thinking about global multi read one writer mutex for this sort of thing, added to [carry over](https://github.com/msupply-foundation/open-msupply/issues/2821) as comment

server/util/src/lib.rs
```diff
@@ -27,3 +29,9 @@ pub use error::*;
...
+    env::var("IS_CENTRAL_SERVER")
+        .map(|is_cetral_server| is_cetral_server.to_lowercase() == "true")
```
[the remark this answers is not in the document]
andreievg: 7c2b13d8022c280f4ca4118f8fea4807636724cf, thanks

server/repository/src/db_diesel/key_value_store.rs
andreievg: Agree done here: d92564a6175eefdd581de199bcfd2659738a8ca4

server/service/src/sync/api_v6/mod.rs
andreievg: Yep, I think keeping separate API version for backward compatibility of all sites would have way too many use cases to keep track of, I think we would take similar approach to what is being done for legacy mSupply central sync api, try to keep it backwards compatible, if we do break compatibility it would require all sites to update to new version. I think this would be possible if we 'sync' binary to remote sites, something that has been mentioned, maybe for RnD day

andreievg:
Thank you for your reviews

@jmbrunskill 

> I would be keen to see or perhaps do a test branch where we try out a feature flag in rust rather than is_central server, doesn't feel like the is_central thing is creating spaghetti code so far.

Hmm, as per discussion in our meeting, the advantage of feature flag would be to build seperate binary, which is usefull when app is being installed, we do have installers that can add env variable to service startup. I am still trying to figure out exactly why feature flag woudl be useful

@clemens-msupply 

> Looks good! Mostly some spelling fixes. Think its better to only read the env variable once at startup and put it into the settings service or similar....

There is a comment about this above and added to carry over

@jmbrunskill @clemens-msupply 

There were a few questions and confusions about push translations used for central and remote, I did a branch where I renamed things to try and make it easier to understand, it's this one: https://github.com/msupply-foundation/open-msupply/compare/Central-config-server-transition...Central-config-server-transition(Renaming-translations), I didn't want to add confusion to PR review by adding loads more changes, so though to keep it separate for now, if you agree will merge into this branch before merging this branch

You can comment in the commit: https://github.com/msupply-foundation/open-msupply/commit/44b072420323584b423870da3e7c7b926f7d5703

The main change is:

https://github.com/msupply-foundation/open-msupply/blob/44b072420323584b423870da3e7c7b926f7d5703/server/service/src/sync/translations/mod.rs#L243-L257

and here:

https://github.com/msupply-foundation/open-msupply/blob/44b072420323584b423870da3e7c7b926f7d5703/server/service/src/sync/translations/mod.rs#L259-L333

I still like the idea of using one translation trait for all translation operation on a particular entity (remote pull and push to legacy, omSupply central, and when omSupply central is translating records in response to pull request from remote site).

How does that one look ?

I think apart form that I dealt with all of the PR comments, thanks once again for review and re-review

jmbrunskill:
> How does that one look ?

I like this change, removes confusion about pushing/pulling

server/util/src/lib.rs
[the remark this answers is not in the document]
clemens-msupply: Was mainly thinking about alterntive setups, e.g when creating a test setup where you might not want to use an env variable. For things like that it might be useful have configurable.

Having a global variable for it is probably ok but still have the feeling its "cleaner" to pass it arround, e.g. the immutable settings object we already have seems suitable and is already available through the SettingService.

clemens-msupply (approved):
Thanks for the changes!

andreievg:
Thanks for your reviews, much appreciated, there is still some things to discuss, but they should all be captured in the carry over

PR #2819, opened 2024-01-24
Central config server transition integration tests

andreievg:
Links to #2062-3 (integration tests for omSupply central)

# 👩🏻‍💻 What does this PR do? 

The updated docs should explain what this PR achieves.

* Add om_supply_central_graphql_operations to integration TestData struct, which also required adding Default spread on existing tests. This field allows mutating data in omSupply central (which needs to be started in `APP_SERVER__DEBUG_NO_ACCESS_CONTROL` mode, thank you @clemens-msupply for this field, made this functionality so much simpler (otherwise would need to create user, login etc..)
* Creates TestData for pack variant
* Adds omSupply central config test driver, which has a helper for running graphql query and for asking omSupply central to sync. 

Running integration tests as per integration test doc

See PR with docs for further explanation of omSupply central server https://github.com/msupply-foundation/open-msupply/pull/2820

### DIFF
--- a/server/service/src/sync/test/integration/omsupply_central/mod.rs
+++ b/server/service/src/sync/test/integration/omsupply_central/mod.rs
@@ -1,0 +1,135 @@
+mod pack_variant;
+mod test;
+
+use std::time::Duration;
+
+use reqwest::Client;
+
+use crate::sync::{
+    api_v6::get_omsupply_central_url,
+    settings::SyncSettings,
+    test::{
+        check_integrated,
+        integration::{
+            central_server_configurations::{ConfigureCentralServer, SiteConfiguration},
+            init_test_context, SyncIntegrationContext,
+        },
+    },
+};
+
+use super::{GraphqlRequest, SyncRecordTester};
+
+async fn test_omsupply_central_records(identifier: &str, tester: &dyn SyncRecordTester) {
+    // util::init_logger(util::LogLevel::Info);
+    // Without re-initialisation
+    println!("test_omsupply_central_records{}_init", identifier);
+
+    let central_server_configurations = ConfigureCentralServer::from_env();
+    let SiteConfiguration {
+        new_site_properties,
+        sync_settings,
+    } = central_server_configurations
+        .create_sync_site(vec![])
+        .await
+        .expect("Problem creating sync site");
+
+    let SyncIntegrationContext {
+        connection,
+        synchroniser,
+        ..
+    } = init_test_context(&sync_settings, &identifier).await;
+
+    let steps_data = tester.test_step_data(&new_site_properties);
+
+    for (index, step_data) in steps_data.into_iter().enumerate() {
+        println!(
+            "test_omsupply_central_records_{}_step{}",
+            identifier,
+            index + 1
+        );
+
+        central_server_configurations
+            .upsert_records(step_data.central_upsert)
+            .await
+            .expect("Problem inserting central data");
+
+        // Sync omSupply central server first
+        sync_omsupply_central(&sync_settings).await;
+        // Integrate omSupply central server records via graphql
+        for graphql_operation in step_data.om_supply_central_graphql_operations {
+            graphql(&sync_settings, graphql_operation).await;
+        }
+
+        synchroniser.sync().await.unwrap();
+        check_integrated(&connection, step_data.integration_records)
+    }
+}
+
+// Helper for graphql queries
+async fn graphql(sync_settings: &SyncSettings, graphql: GraphqlRequest) -> serde_json::Value {
+    let mut url = get_omsupply_central_url(&sync_settings.url).unwrap();
+    url = url.join("graphql").unwrap();
+
+    let result = Client::new()
+        .post(url.clone())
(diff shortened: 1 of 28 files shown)


server/service/src/sync/test/integration/omsupply_central/pack_variant.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        let item_id = uuid();
+        // STEP 1 - insert
+        let item_json = json!({
+            "ID": item_id.clone(),
+            "type_of": "general",
+
+        });
+        let mut rng = rand::thread_rng();
+        let pack_size: u32 = rng.gen();
```
clemens-msupply, line 31: Any special reason why you use a random pack_size?

server/service/src/sync/test/integration/omsupply_central/test.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(test)]
+mod tests {
+    use crate::sync::test::integration::omsupply_central::{
+        pack_variant::PackVariantTester, test_omsupply_central_records,
+    };
+
+    #[actix_rt::test]
+    async fn omsupply_central_sync_pack_variant() {
```
clemens-msupply, line 8: Can/should this be renamed to also start with `integration_sync` so that it will run with all the other integration test when doing: `cargo test integration_sync`?

clemens-msupply (approved):
Works great, thank you! just some spelling and questions...

Do you plan to have something similar as for the 4D mSupply central server when running tests? For example, we create test sites on central, would we use test DBs or similar for the omSupply central server tests as well? Could it cause problems to run tests on a "prod" open mSupply central server?

I have a bit of a none standard setup, running 4D in a Windows VM. This caused some trouble running the tests, e.g.

- 4D central runs on http://192.168.178.101:2048
- omSupply central runs on http://localhost:2050

So the port offset logic didn't work for me and I had to hard code it in...

Should I create a PR to add an optional config option for SYNC_URL_OM_CENTRAL so that I can separately configure how to reach the 4D server and how to reach the omSupply central? If this new option is not set I would fallback to the current behaviour.

server/service/src/sync/test/integration/omsupply_central/pack_variant.rs
andreievg: I think that was a confusion on my part (api enforces uniqueness of pack_size/item, but we creating new item anyway), changed: 1e0109286f154fdb9e7363062519c6e209794e6a

server/service/src/sync/test/integration/omsupply_central/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+        sync_settings,
+        GraphqlRequest {
+            query: "mutation { manualSync }".to_string(),
+            ..Default::default()
+        },
+    )
+    .await;
+
+    loop {
+        // TODO max timeout ?
```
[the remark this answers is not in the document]
andreievg: Thanks, added to TODO 186d5f1cff591552add053b194c2e32097b4cb83

server/service/src/sync/test/integration/omsupply_central/test.rs
andreievg: Thanks, great point: 95de51c2afcb49393a1a388eae0d365efeffef36

andreievg:
> Works great, thank you! just some spelling and questions...

Thanks for suggestions/questions

> Do you plan to have something similar as for the 4D mSupply central server when running tests? For example, we create test sites on central, would we use test DBs or similar for the omSupply central server tests as well? Could it cause problems to run tests on a "prod" open mSupply central server?

I was hoping that the test 'driver' (the method running the test), would be able to do all of the data mutations needed for the test, the only pre-requisite is having a site on legacy central for sync_test mutations and another site on legacy central for omSupply central. And the driver uses sync_test on legacy central mutations and graphql mutations on omSupply central + sync graphql trigger for data in tests.

> I have a bit of a none standard setup, running 4D in a Windows VM. This caused some trouble running the tests, e.g.
> 
> * 4D central runs on http://192.168.178.101:2048
> * omSupply central runs on http://localhost:2050
> 
> So the port offset logic didn't work for me and I had to hard code it in...

Sorry, i didn't think of that, and the port offset things is not good, just intermediary step for now.

> Should I create a PR to add an optional config option for SYNC_URL_OM_CENTRAL so that I can separately configure how to reach the 4D server and how to reach the omSupply central? If this new option is not set I would fallback to the current behaviour.

Actually I am leaning toward this now:

* In legacy central mSupply central server configurations with configure which site is 'central' and how it can be reached (this is just stored against the 'sites' table).
* Before every sync operation, there is a call to some 'info' end point, this will double check that it's ok to sync this site and also will return back `is this site central omSupply server` and `where is central omSupply server`. That should give us enough flexibility and an easy way to inform existing sites that are updated, where to find omSupply server.

I think `SYNC_URL_OM_CENTRAL` is a good idea, but I feel like it's going to be super seeded by the above. Btw I added a comment to this comment in carry over: https://github.com/msupply-foundation/open-msupply/issues/2821,

andreievg:
Thanks for review and running the test !

PR #2824, opened 2024-01-24
Backspace too quickly and the query is incorrect

### LINKED ISSUE #2822: Location filter doesn't retains all list when Backspace is pressed
## What went wrong? 😲

https://github.com/msupply-foundation/open-msupply/assets/117138852/b8a3c682-4abd-465b-9178-aeac56d9e053

Screenshot 2024-01-24 at 15 32 12

## How to Reproduce 🔨

1. Open mSupply
2. Go to Inventory ->Locations
3. Search for existing location by name so only one location would get filtered
4. Now press long backsapce key so that all the search field clears up
5. All the locations would not pop-up on the field (BUG)

## Your environment 🌱
<v1.6>
 - Version: 
 - Platform:
   - [ ] android (tablet)
   - [x] desktop (windows)
   - [ ] desktop (macOS)
   - [x] server (windows)

mark-prins:
Fixes #2822

# 👩🏻‍💻 What does this PR do? 
 
If you enter a value into a text filter and then slowly press backspace, deleting each character in turn, then the filter works correctly.

If you do what I think is more normal, and mash backspace repeatedly until the text is cleared, then the first letter of the query remains.

This is because the url query is cleared immediately and then the debounced call fires and sets the query to be the final character in the search text. I've removed the inconsistent behaviour.

As above, try clearing search text quickly using backspace and see what happens.

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Filters/TextFilter.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/TextFilter.tsx
@@ -29,8 +29,7 @@ export const TextFilter: FC<{
 
   const handleChange = (newValue: string) => {
     setValue(newValue);
-    if (newValue === '') updateQuery({ [filterDefinition.urlParameter]: '' });
-    else debouncedOnChange(newValue);
+    debouncedOnChange(newValue);
   };
 
   return (


roxy-dao (approved):
Works perfectly thanks Mark! ❤️

PR #2825, opened 2024-01-25
Rename item_link_id for MasterListLineRow

clemens-msupply:
Part of #2759

This one introduces a proper MasterListLine domain object. This actually follows the idea of a repository more closely to make the DB layer independent of the rest of the system. Maybe we should have done this for other repos as well?

We could even make MasterListLineRow private to the repo and change all usages to MasterListLine...

### DIFF
--- a/server/graphql/general/src/queries/tests/master_lists.rs
+++ b/server/graphql/general/src/queries/tests/master_lists.rs
@@ -108,7 +108,7 @@ mod graphql {
                 json!({
                     "id": line.id,
                     "item": {
-                        "id": line.item_id
+                        "id": line.item_link_id
                     }
                 })
             })
--- a/server/repository/src/db_diesel/item.rs
+++ b/server/repository/src/db_diesel/item.rs
@@ -514,22 +514,22 @@ mod tests {
         let master_list_line_rows = vec![
             MasterListLineRow {
                 id: "id1".to_owned(),
-                item_id: "item1".to_owned(),
+                item_link_id: "item1".to_owned(),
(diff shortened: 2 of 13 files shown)


andreievg (approved):
Looks good, thanks @clemens-msupply

Chris-Petty:
Also approve! But got side tracked :)

> We could even make MasterListLineRow private to the repo and change all usages to MasterListLine...

I think this has some legs because honestly I get confused half the time which I'm meant to use where.

https://github.com/msupply-foundation/open-msupply/blob/aeb9a384d0f0b2431083ee7ce3f51a8eb5fe0ca0/server/repository/src/mock/full_master_list.rs#L62-L78

For instance would it mean for tests such as above we'd use `MasterListNameJoin` and `MasterListLine` and `item_id` and `name_id` respectively?

clemens-msupply:
> Also approve! But got side tracked :)
> 
> > We could even make MasterListLineRow private to the repo and change all usages to MasterListLine...
> 
> I think this has some legs because honestly I get confused half the time which I'm meant to use where.
> 
> https://github.com/msupply-foundation/open-msupply/blob/aeb9a384d0f0b2431083ee7ce3f51a8eb5fe0ca0/server/repository/src/mock/full_master_list.rs#L62-L78
> 
> For instance would it mean for tests such as above we'd use `MasterListNameJoin` and `MasterListLine` and `name_id` and `item_id` respectively?

Yes I think this would be nice. However, this would require some more refactoring for upserts and the mock tests...

PR #2829, opened 2024-01-25
2827 separate currency from language

### LINKED ISSUE #2827: Separate currency from language
The currency symbol is currently linked to the language, but many countries can use the same language but a different currency.

Separate currency from language while keeping the current currency shape below.
```
en: {
    symbol: '$',
    separator: ',',
    decimal: '.',
    precision: 2,
    pattern: '!#',
    negativePattern: '-!#',
    format,
  }
```

The language key should be replaced by the currency code.

roxy-dao:
Fixes #2827

# 👩🏻‍💻 What does this PR do? 
Removes `is_active` from currency... add sort by `is_home_currency` and store's `currency_code` to `UserNode` and finally separate currency from language.

Default currency: `USD/NZD`

- Run frontend tests with `yarn test`

### DIFF
--- a/client/packages/common/src/intl/currency/currency.test.tsx
+++ b/client/packages/common/src/intl/currency/currency.test.tsx
@@ -43,27 +43,21 @@ describe('currency formatting - en', () => {
 
 describe('currency formatting - fr', () => {
   it('formats a string with up to the precision number of decimal places, dropping decimal places where needed', () => {
-    const { result } = renderHookWithProvider(() =>
-      useCurrency(undefined, 'EUR')
-    );
+    const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c(1.11111111111).format();
     expect(f1).toBe('1,11 €');
   });
 
   it('formats a string with up to the precision number of decimal places, dropping decimal places where needed even with large numbers', () => {
-    const { result } = renderHookWithProvider(() =>
-      useCurrency(undefined, 'EUR')
-    );
+    const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c(111_111.11111111111).format();
     expect(f1).toBe('111.111,11 €');
   });
 
   it('does drop trailing zeroes', () => {
-    const { result } = renderHookWithProvider(() =>
-      useCurrency(undefined, 'EUR')
-    );
+    const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     // Note: Using a string to pass into c as formatters will generally
     // auto clear trailing zeroes when literal numbers.
@@ -71,17 +65,13 @@ describe('currency formatting - fr', () => {
     expect(f1).toBe('111,11 €');
   });
   it('has a minimum of two trailing zeroes, adding one if needed', () => {
-    const { result } = renderHookWithProvider(() =>
-      useCurrency(undefined, 'EUR')
-    );
+    const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c('111,1000').format();
     expect(f1).toBe('111,10 €');
   });
   it('has a minimum of two trailing zeroes, adding two if needed', () => {
-    const { result } = renderHookWithProvider(() =>
-      useCurrency(undefined, 'EUR')
-    );
+    const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c(111).format();
     expect(f1).toBe('111,00 €');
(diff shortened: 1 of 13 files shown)


server/graphql/general/src/queries/currency.rs
```diff
@@ -30,21 +31,18 @@ pub struct CurrencySortInput {
 pub struct CurrencyFilterInput {
     pub id: Option<EqualFilterStringInput>,
     pub is_home_currency: Option<bool>,
-    pub is_active: Option<bool>,
```
roxy-dao, line 33: This field isn't actually being used in central...

server/graphql/types/src/types/user.rs
```diff
@@ -52,6 +52,27 @@ impl UserStoreNode {
     pub async fn store_mode(&self) -> StoreModeNodeType {
         StoreModeNodeType::from_domain(&self.user_store.store_row.store_mode)
     }
+
+    pub async fn home_currency_code(&self, ctx: &Context<'_>) -> Result<Option<String>> {
```
roxy-dao, line 56: Added to get default currency to load throughout store.

client/packages/common/src/intl/currency/currency.test.tsx
```diff
@@ -44,52 +42,46 @@ describe('currency formatting - en', () => {
...
-    const { result } = renderHookWithProvider(useCurrency, {
-      providerProps: { locale: 'fr' },
-    });
+    const { result } = renderHookWithProvider(() =>
+      useCurrency(undefined, 'EUR')
+    );
 
-    const f2 = result.current.c(1.11111111111).format();
-    expect(f2).toBe('1,11 €');
+    const f1 = result.current.c(1.11111111111).format();
```
roxy-dao, line 50: Just because... consistency 🤷🏻

client/packages/common/src/intl/currency/currency.ts
```diff
@@ -67,80 +67,120 @@ export const format: currency.Format = (
...
+    }
+    case 'COP': {
+      return {
+        symbol: '$',
+        pattern: '# !',
+        negativePattern: '-# !',
+        format,
+      };
+    }
+    default:
```
mark-prins, line 161: How about doing this?
```suggestion
    case 'USD':
    case 'NZD':
    default:
```

and removing the first entry in the case statement?

client/packages/common/src/intl/currency/currency.ts
```diff
@@ -67,80 +67,120 @@ export const format: currency.Format = (
...
+        negativePattern: '-!#',
+        format,
+      };
+  }
 };
 
-export const useCurrency = (dp?: number) => {
-  const { currentLanguage: language } = useIntlUtils();
-  const options = currencyOptions[language];
+export const useCurrency = (dp?: number, code?: Currencies) => {
```
mark-prins, line 174: I can find no usage of `useCurrency` or `useFormatCurrency` that actually use the dp parameter, so I'm tempted to simply drop it.

if you think we should retain `dp` as a parameter it would be nicer if this was an object, then we don't have to pass `undefined` for the first param when you only want to specify the second..

```suggestion
export const useCurrency = ({dp, code}: {dp?: number, code?: Currencies}) => {
```

client/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -40,9 +40,9 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   value,
   ...restOfProps
 }) => {
-  const { c, options, language } = useCurrency();
-  const prefix = language !== 'fr' ? options.symbol : '';
-  const suffix = language === 'fr' ? options.symbol : '';
+  const { c, options, currencyCode } = useCurrency();
+  const prefix = currencyCode !== 'EUR' ? options.symbol : '';
+  const suffix = currencyCode === 'EUR' ? options.symbol : '';
```
mark-prins, line 45: I hadn't remembered the hard coded values in here!
Could you make this more flexible please - I would do this:
```suggestion
  const { c, options } = useCurrency();
  const isSymbolLast = options.pattern.endsWith('!');
  const prefix = !isSymbolLast ? options.symbol : '';
  const suffix = isSymbolLast ? options.symbol : '';
```

which defaults to the symbol being a prefix, and is using the formatting pattern from the currency options, not a hard coded value.

mark-prins (changes requested):
🌟 Looks great thanks!
have a suggested simplification for your consideration..

mark-prins (approved):
👍 Looks great - thanks!

PR #2851, opened 2024-01-29
Rename item_link_id for StockLineRow

### LINKED ISSUE #2849: Merging: Rename StockLineRow::item_id to item_link_id

After the first merge of the `feature/merging` branch column renames should be removed, e.g:

```rust
    #[column_name = "name_link_id"]
    pub name_id: String,
```

The reason is that all query method exped a `name_id` but with the current code we use a `name_link_id`. Thus a query would fail for merged names/items.

Same changes should be applied for item_id/item_link_id.

See past similar issues for example PRs.

## Moneyworks Jobcode 🧰

OMS:MERGE

clemens-msupply:
Fixes #2849

### DIFF
--- a/server/service/src/invoice_line/stock_out_line/update/mod.rs
+++ b/server/service/src/invoice_line/stock_out_line/update/mod.rs
@@ -1,6 +1,6 @@
 use repository::{
     InvoiceLine, InvoiceLineRow, InvoiceLineRowRepository, RepositoryError, StockLine,
-    StockLineRow, StockLineRowRepository,
+    StockLineRowRepository,
 };
 
 use crate::{
--- a/server/service/src/invoice_line/stock_out_line/update/mod.rs
+++ b/server/service/src/invoice_line/stock_out_line/update/mod.rs
@@ -1,6 +1,6 @@
 use repository::{
-    InvoiceLine, InvoiceLineRow, InvoiceLineRowRepository, RepositoryError, StockLineRow,
-    StockLineRowRepository,
+    InvoiceLine, InvoiceLineRow, InvoiceLineRowRepository, RepositoryError, StockLine,
+    StockLineRow, StockLineRowRepository,
 };
 
 use crate::{
@@ -68,9 +68,9 @@ pub fn update_stock_out_line(
             InvoiceLineRowRepository::new(&connection).upsert_one(&update_line)?;
 
             let stock_line_repo = StockLineRowRepository::new(&connection);
-            stock_line_repo.upsert_one(&batch_pair.main_batch)?;
+            stock_line_repo.upsert_one(&batch_pair.main_batch.stock_line_row)?;
             if let Some(previous_batch) = batch_pair.previous_batch_option {
-                stock_line_repo.upsert_one(&previous_batch)?;
+                stock_line_repo.upsert_one(&previous_batch.stock_line_row)?;
             }
 
             get_invoice_line(ctx, &update_line.id)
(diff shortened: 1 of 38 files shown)


server/service/src/stocktake/update.rs
```diff
@@ -241,15 +250,16 @@ fn generate_stock_line_update(
...
-        Some(item) => item,
-        None => {
-            return Err(UpdateStocktakeError::InternalError(format!(
-                "Can't find item {} for existing stocktake line {}!",
-                &stock_line.item_id, stocktake_line.line.id
-            )))
-        }
-    };
+    let stock_line_item =
+        match ItemRowRepository::new(connection).find_one_by_id(&stock_line_item_id)? {
```
clemens-msupply, line 254: The stocktake load the stock_line_row which doesn't has the stock_line item loaded...

Here raises the question: who load who and where to stop loading? The problem would be even worth if there are circular table references.

In domain driven design, you would define an aggregate root, e.g. stocktake or stock_line. All roots are allowed to only contain references to other roots (which avoids the circular reference problem). So far we were fine to just load what we needed, but we might have to have a look at some of these DDD ideas...

server/service/src/invoice_line/stock_out_line/update/mod.rs
```diff
@@ -85,9 +85,9 @@ pub fn update_stock_out_line(
 /// validation and updates need to apply to both batches
 pub struct BatchPair {
     /// Main batch to be updated
-    pub main_batch: StockLineRow,
+    pub main_batch: StockLine,
     /// Optional previous batch (if batch was changed)
-    pub previous_batch_option: Option<StockLineRow>,
```
clemens-msupply, line 90: Had to refactor this a bit to use the StockLine...

server/service/src/invoice_line/stock_out_line/validate.rs
```diff
@@ -43,42 +43,36 @@ pub fn check_unique_stock_line(
...
-                .map_err(|_| LocationNotFound)?;
-
-            match location {
-                Some(location) if location.on_hold => Err(LocationIsOnHold),
-                Some(_) => Ok(()),
-                None => Err(LocationNotFound),
+    match &batch.location_row {
+        Some(location) => {
+            if location.on_hold {
+                return Err(LocationIsOnHold);
```
clemens-msupply, line 73: Since the StockLine already has the location loaded I was able to get rid of the LocationNotFound error :)

clemens-msupply: This method could be refactored to just return a `bool` but I leave it for now...

server/service/src/repack/generate.rs
```diff
@@ -25,13 +25,13 @@ struct StockLineJob {
 
 pub fn generate(
     ctx: &ServiceContext,
-    stock_line: StockLineRow,
+    stock_line: StockLine,
```
Chris-Petty, line 28: Is this related to what you mentioned the other day, that StockLineRow and its ilk should be private?

Chris-Petty (approved):
Nice!

server/service/src/repack/generate.rs
clemens-msupply: The row structs are really just the Diesel structs to deserialize from the database. We used these row objects as domain objects to avoid doing a 1 to 1 mapping from Diesel object to our Domain object (StockLineRow -> StockLine). This seemed very laborious and cumbersome at the time... Now this would have come in handy though and would have saved us some time.

We should at least have used a type alias consistently `type StockLine = StockLineRow`...

So yes the Diesel structs should best be private.

Issue #2821, opened 2024-01-24
omSupply Central Server Transition Carry Over


andreievg:
Carry over issue for: https://github.com/msupply-foundation/open-msupply/issues/2062

Related PRs:

https://github.com/msupply-foundation/open-msupply/pull/2820
https://github.com/msupply-foundation/open-msupply/pull/2819
https://github.com/msupply-foundation/open-msupply/pull/2818
https://github.com/msupply-foundation/open-msupply/pull/2799
https://github.com/msupply-foundation/open-msupply/pull/2798

# Carry Over Tasks

- [ ] Versioning
- [ ] Add sync steps to front end
- [ ] item_link for merging (on pack variant)
- [ ] Organise a meeting to talk over the questions
- [ ] Example of adding new central data (similar to legacy sync example in sync docs)
- [ ] Make sure all central data is soft deleted (there are TODO about this)
- [ ] I had to add "../app_data" for test service for integration tests (since hardware id for the 'test' site is populated when omSupply central server for the test starts), and it needs to match when sites is created
- [ ] Investigate stack overflow, this happened when I tried by accident to push unknown shape to original central mSupply server, do we get infinite loop of there is some error in push to original central mSupply server ?
- [ ] I noticed invoices don't have is_sync_update, does it mean when transfer arrives we sync it back ? Or do we have another mechanism to filter out (by name_id ?) Should we just use is_sync_update to be consistent ?

# Questions

1. When we omSupply central server in production, we will need to make sure it's backed up, and backup would need to match mSupply backup ? Or do we just try to backup the data that's different to mSupply ?
2. Should we allow omSupply remote sites to run without central server for now ? (since it's just central data and worst case the configuration would be missing ?)
3. Where to put sync steps for omSupply central ?
4. How do we determine is central server ? With env variable ?
5. The idea to use change log, purely, would it scale ?
6. What do feel about branching and fragmenting logic with is central server flag ? Do we split code base
7. Mirror server for future ? (i don't think it should be needed, since omSupply central would be easily accessible via browser and wont' require much bandwidth)
8. In order for all features for omSupply central to work, more information is required, which currently lives on original central server, do we sync it ? how ?
9. In the future, what would happen with existing mSupply sites if we decide to bring existing mSupply data to omSupply central sync ? (also how do we version/migrate existing omSupply sites for this ?)
10. Is current authentication modal ok for now ? (against original mSupply server)
11. Way forward ? I think we have enough for assets btw

[...]

PR #2866, opened 2024-01-31
2502 restrict pack variant editing to central server

### LINKED ISSUE #2502: Restrict pack variant editing to omSupply central server
Restrict editing of pack variants for omSupply central server only (would still be visible though), this is to be done in service and graphql permissions. Carry over from #2309 

### Moneyworks Jobcode 🧰
OMS:NEST

andreievg:
Fixes #2502 

# 👩🏻‍💻 What does this PR do? 

* add isCentralServer graphql query (no permission check, should it have ?)
* move central server graphql mutations behind centralServer endpoint (idea is to have one place where isCentralServer is checked)
* add isCentralServer api query to front end, and useIsCentralServerOrWarning hook to help with central server checks
* disallow deleting of pack variant when user doesn't have permission (this was done for editing and for creating pack variants but not for delete)
* Silently ignore sync error to omSupply central server, this is just temporary, this point is in carry over issue, we can discuss this tomorrow as it's in the [carry over issue](https://github.com/msupply-foundation/open-msupply/issues/2821)

Start server in central server mode:
IS_CENTRAL_SERVER=true cargo run
make sure user has this permission (if not tick enable it and re-login):
Screenshot 2024-01-31 at 5 01 16 PM
- [ ] go to item and pack_variant tab, try adding a pack variant (should be able to add)
Remove above permission from user (and re-login)
- [ ] go to item and pack_variant tab, try selecting pack variant deleting it (should get permission error).
Stop server and start in remote site mode
`cargo run`
- [ ]  go to item and pack_variant tab, try deleting, adding new pack variant or editing existing one, should get an error saying this operation is only allowed on central server

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/client/packages/common/src/authentication/api/api.ts
+++ b/client/packages/common/src/authentication/api/api.ts
@@ -83,6 +83,10 @@ export const getAuthQueries = (sdk: Sdk, t: TypedTFunction<LocaleKey>) => ({
       const result = await sdk.refreshToken();
       return refreshTokenGuard(result);
     },
+    isCentralServer: async () => {
+      const result = await sdk.isCentralServer();
+      return result.isCentralServer;
+    },
     me: async (token?: string) => {
       try {
         const result = await sdk.me(
--- a/client/packages/common/src/authentication/api/hooks/index.ts
+++ b/client/packages/common/src/authentication/api/hooks/index.ts
@@ -4,3 +4,4 @@ export * from './useUserDetails';
 export * from './useRefreshToken';
 export * from './useLogin';
 export * from './useGetUserPermissions';
+export * from './useIsCentralServer';
--- a/client/packages/common/src/authentication/api/hooks/useAuthApi.ts
+++ b/client/packages/common/src/authentication/api/hooks/useAuthApi.ts
@@ -10,6 +10,7 @@ export const useAuthApi = () => {
 
   const keys = {
     me: (token: string) => ['me', token] as const,
+    isCentralServer: ['isCentralServer'] as const,
     refresh: (token: string) => ['refresh', token] as const,
   };
 
--- a/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts
+++ b/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts
@@ -5,10 +5,13 @@ import { useTranslation } from '@common/intl';
 
 export const useIsCentralServerApi = () => {
   const api = useAuthApi();
-  // api.keys.isCentralServer should guarantee that this is called just once
-  // on page load
-  const { data } = useQuery(api.keys.isCentralServer, () =>
-    api.get.isCentralServer()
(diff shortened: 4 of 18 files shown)


mark-prins:
I didn't see where you mentioned which mSupply branch to use. I've used develop and cannot find the permission

mark-prins:
@andreievg There's a fundamental bug which has come from a different PR. 
In `Host.tsx` the `<Init />` will need to be moved to be under the `<AuthProvider>`. As it stands, the store id is always `store-id` and therefore the permission check fails. I can add a new variant, but then it doesn't show me the change because the store id is incorrect

mark-prins:
Can you tweak the UI please?
The select drop down should be on the right, nearer the checkboxes. Perhaps it has to go under the 'new pack variant' button.

Also - the select doesn't make sense when there is no table, because it isn't clear that there is a table:

Screenshot 2024-02-07 at 12 17 59 PM

Can we hide it in this case?

Chris-Petty:
Testing github actions

mark-prins (changes requested):
Working as described - once I fixed the store ID problem I had. Couple of things to change, as noted.

Can you update the Host.tsx, and preferably, update the layout of the pack unit editing please?

Can you also update the comment in `Host.tsx` please? The text says 'api keys' but I think it should be 'cache keys':

```
// They will be refetched on page reload or when store is changed based on cache usePackUnits cache keys
```

and a typo in `usePackVariants`

```
// or if store is changed (based on api keys)
```

there's a 'd' missing from the end of changed 🙏

andreievg:
@mark-prins thanks for review

In this commit: 8800bd36d2473a4d81322f539e966f86a44a59c1

* Fixed comments, api -> cache, Unit -> Variatns
* Relocated <Init> to be within AuthProvider
* Boolean -> boolean
* I moved drop down to the right in pack variant view (as per screenshot), first it was under the button, and it seemed to use too much vertical space, which is very limited on tablets, so just added drop down next to button (toggled via visibility rather then conditional rendering). I couldn't figure out easy way to see if there are rows in the table, i added isEmpty to useTableState, which was changes when setRows was called, but that didn't 'hide' dropDown when the last row was deleted ¯\_(ツ)_/¯, so I left it as TODO). How to check there are no rows ? Also the drop down is visible in other views, is there an issue to change that ? (see last screenshot)
* Renamed hook, although i settled for slightly longer but more explicit naming. Thanks for the suggestions and identifying as confusion, it really was confusing (might still be)

Also added 'd' to change -> ac1bebdb104746e54800380be0a226dcf6f667d1

> doOrDie which isn't helpful but I like it

Haha I wish we could have a bit more fun with variable names like this. But it's public repo hehe

mark-prins (approved):

mark-prins:
cool - thanks for the changes @andreievg! looking good

andreievg:
@mark-prins thanks for review !

PR #2869, opened 2024-01-31
2840 link column rename encounter

### LINKED ISSUE #2840: Merging: Rename EncounterRow::patient_id to patient_link_id

After the first merge of the `feature/merging` branch column renames should be removed, e.g:

```rust
    #[column_name = "name_link_id"]
    pub name_id: String,
```

The reason is that all query method exped a `name_id` but with the current code we use a `name_link_id`. Thus a query would fail for merged names/items.

Same changes should be applied for item_id/item_link_id.

See past similar issues for example PRs.

## Moneyworks Jobcode 🧰

OMS:MERGE

clemens-msupply:
Fixes #2840 and #2843

action bump2

### DIFF
--- a/server/repository/src/db_diesel/encounter.rs
+++ b/server/repository/src/db_diesel/encounter.rs
@@ -1,20 +1,20 @@
 use super::{
-    clinician_link_row::{clinician_link, clinician_link::dsl as clinician_link_dsl},
-    document::latest_document::dsl as latest_document_dsl,
-    encounter_row::encounter::{self, dsl as encounter_dsl},
-    name_link_row::{name_link, name_link::dsl as name_link_dsl},
-    name_row::{name, name::dsl as name_dsl},
-    program_enrolment_row::program_enrolment::dsl as program_enrolment_dsl,
-    program_row::{program, program::dsl as program_dsl},
+    clinician_link_row::clinician_link,
+    clinician_row::clinician,
+    encounter_row::encounter::{self},
+    name_link_row::name_link,
+    name_row::name,
+    program_row::program,
     StorageConnection,
 };
 
 use crate::{
+    db_diesel::program_enrolment_row::program_enrolment,
     diesel_macros::{apply_date_time_filter, apply_equal_filter, apply_sort, apply_string_filter},
-    latest_document, ClinicianLinkRow, DBType, DatetimeFilter, EncounterRow, EncounterStatus,
-    EqualFilter, NameLinkRow, NameRow, Pagination, PatientFilter, PatientRepository,
-    ProgramEnrolmentFilter, ProgramEnrolmentRepository, ProgramRow, RepositoryError, Sort,
-    StringFilter,
+    latest_document, ClinicianLinkRow, ClinicianRow, DBType, DatetimeFilter, EncounterRow,
+    EncounterStatus, EqualFilter, NameLinkRow, NameRow, Pagination, PatientFilter,
+    PatientRepository, ProgramEnrolmentFilter, ProgramEnrolmentRepository, ProgramRow,
+    RepositoryError, Sort, StringFilter,
 };
 
 use diesel::{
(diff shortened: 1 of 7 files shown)


server/repository/src/db_diesel/encounter.rs
```diff
@@ -132,12 +132,24 @@ type EncounterJoin = (
...
-    Option<ClinicianLinkRow>,
+    Option<(ClinicianLinkRow, ClinicianRow)>,
 );
-pub type Encounter = (EncounterRow, ProgramRow, NameRow);
 
-fn to_domain((encounter_row, program_row, (_, name_row), _): EncounterJoin) -> Encounter {
-    (encounter_row, program_row, name_row)
+#[derive(Clone)]
+pub struct Encounter {
+    pub row: EncounterRow,
```
Chris-Petty, line 140: Nit: other repos have pattern where this would be `encounter_row`, though I don't like elsewhere having to do `encounter.encounter_row` so this is sort of nice. 

Tangent: I was reading an [interesting blog](https://blog.sigma-star.io/2024/01/people-dont-understand-oop/) on OOP yesterday and in terms of jargon I liked this bit:

> I’m sure if you have done any object-oriented programming you’ve heard something like “Don’t use public properties […]” (properties in the sense of member variables) at some point. And this is true, because public properties expose the internal state, and can potentially cause high object coupling. However, as with any dogma, it’s usually a good idea to question it. In this case the complete “guideline” is “Don’t use public properties, use getters and setters instead.”, which is completely wrong. In terms of encapsulation getters and setters are just as bad as public properties, as they do nothing to prevent object coupling. If you have a class without any methods (besides getters and setters) it doesn’t really fit our object definition. A term that has been used for this is “record”.

Now, definitely in the realms of bikeshedding, but if this "Object/struct that is just public properties, no methods" definition of "record" was a broadly accepted jargon then I'd understand `EncounterRecord` quite clearly - it's the raw dog entity, the database record... which some people would call a "row"... did I mention bikeshedding 😂?

The _real_ thing I wanted to touch on was that in DDD as I've been reading it I'd think that our domain objects should be a bundle of domain objects rather than rows? I'm not sure what the implications of that would be just thinking aloud as I understand our code and DDD better.

Chris-Petty (approved):

server/repository/src/db_diesel/encounter.rs
```diff
@@ -1,20 +1,20 @@
...
-    name_link_row::{name_link, name_link::dsl as name_link_dsl},
-    name_row::{name, name::dsl as name_dsl},
-    program_enrolment_row::program_enrolment::dsl as program_enrolment_dsl,
-    program_row::{program, program::dsl as program_dsl},
+    clinician_link_row::clinician_link,
+    clinician_row::clinician,
+    encounter_row::encounter::{self},
+    name_link_row::name_link,
+    name_row::name,
+    program_row::program,
```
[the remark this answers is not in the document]
clemens-msupply: Do you review by commit? and if yes how do you do it? I would love to do it, but I don't have a good tool to track my progress, e.g. compared to the build in github PR UI.

Being able to review by commit would also solve the stacked PR problem, e.g. you could just keep adding new commits to the same branch...

server/repository/src/db_diesel/encounter.rs
clemens-msupply: >Nit: other repos have pattern where this would be encounter_row, though I don't like elsewhere having to do encounter.encounter_row so this is sort of nice.

Yes I know, should have used encounter_row to be consistent. But I also liked `row` better and probably doesn't matter much...  Easy to refactor though.

`Record` or `Entity` might be a better name for our domain object rather than rows. We used Diesel rows because they looked exactly as our domain objects.

I think the following would be good ideas. With the merge link changes this:

`type MyEntity = MyDieselRow`

would become:

```rust
struct MyDieselRow {
   ...
   name_link_id
}

pub struct MyEntity {
  ...
  name_id
}

impl MyEntity {
   fn from_row(row: (MyDieselRow, NameRow)) -> MyEntity {...}
}
```

Where MyDieselRow is only used in the repos. If we had used a type alias for all our rows these rename changes would have been easier and would have touched only a couple of files in the repo layer...

PR #2879, opened 2024-02-01
Add option for login layout background image size

### LINKED ISSUE #2877: Specify background image size for login screen

If not using a gradient, but specifying a background image, it is useful to be able to specify the background size.

Currently:

Screenshot 2024-02-01 at 3 26 44 PM

Desired:

Screenshot 2024-02-01 at 3 32 50 PM

mark-prins:
Fixes #2877

# 👩🏻‍💻 What does this PR do? 
 

I had to add this in order to make a demo theme look nice. Just a minor tweak to allow specifying the `background-size` property on the login screen LHS

You can try with the custom theme in [this folder](https://drive.google.com/open?id=1IBM2SmyfQQdRzh341CXONX4rk7Bl6QjZ&usp=drive_fs)

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/host/src/components/Login/LoginLayout.tsx
+++ b/client/packages/host/src/components/Login/LoginLayout.tsx
@@ -40,6 +40,7 @@ export const LoginLayout = ({
         flex="1 0 50%"
         sx={{
           backgroundImage: (theme: Theme) => theme.mixins.gradient.primary,
+          backgroundSize: (theme: Theme) => theme.mixins.gradient.size,
           padding: '0 80px 7% 80px',
         }}
         display="flex"
--- a/client/packages/common/src/styles/theme.ts
+++ b/client/packages/common/src/styles/theme.ts
@@ -41,6 +41,7 @@ declare module '@mui/material/styles/createMixins' {
     gradient: {
       primary: string;
       secondary: string;
+      size?: string;
       tertiary: string;
     };
     icon: {


client/packages/host/src/components/Login/LoginLayout.tsx
```diff
@@ -40,6 +40,7 @@ export const LoginLayout = ({
         flex="1 0 50%"
         sx={{
           backgroundImage: (theme: Theme) => theme.mixins.gradient.primary,
+          backgroundSize: (theme: Theme) => theme.mixins.gradient.size,
```
CarlosNZ, line 43:
```suggestion
          backgroundSize: (theme: Theme) => theme.mixins.gradient.size,
          backgroundPosition: 'center',
          backgroundRepeat: 'no-repeat',
```

The "center" makes it look better through a bigger range of window resizes.
The "no-repeat" does nothing with your theme (if "cover" is set) but is better when the "size" property is not specified (i.e. defaulting to "contain"), so makes sense to leave it on permanently. In fact, it's arguable whether this theme might actually be better just with "no-repeat" and not changing the "backgroundSize" (this makes it never overflow the box, even with extreme aspect ratios), but that's subjective and not particularly important in most cases.

CarlosNZ (approved):
Great thanks. Just made an additional suggestion for a couple of other properties that probably help.

CarlosNZ:
Also -- it would be nice if you could toggle the "Custom Theme" and "Custom Icon" buttons off and on again without losing the current customisations. Perhaps "customTheme" enabled setting could be a separate local storage item to the theme data itself? Not huge, but might be worth changing at some point if it's easy to do.

mark-prins:
> it would be nice if you could toggle the "Custom Theme" and "Custom Icon" buttons off and on again without losing the current customisations

Yeah - and I started down that path.. but the current implementation stores only the value and the hash in the database. To add the enabled prop requires api changes so I'll leave that for a separate issue.

Issue #2668, opened 2023-12-29
Shouldn't oms allocate placeholder lines based on issue quantity?


nishaDangol-Sussol:
## What went wrong? 😲
Sometimes, despite the available packs (pack-sizes), oms discards the available stock and issues out unfavorable stocks (for the requested quantity) when chosen to Allocate placeholder lines 

## Expected behaviour 🤔
Maybe the first priority is expiry date of stocks, but after that the next priority should definitely be the availability of pack-size as per the requested quantity! 

## How to Reproduce 🔨
For this particular scenario, these are the steps what I did:
1. Create a new stocktake for an item with multiple batch of multiple pack-sizes: I create two of pack-sizes 1 & 2
2. Get a IO from other store so that you can see the requisition. For this example, I made the IO quantity to be 3
3. Supply quantity and create outbound shipement.
4. In shipment window, select that line (with requested quantity of 3) and allocate placeholder lines
5. See error: it chose the pack-size 2 and chose more-than-required quantity for this (which could have been avoided if the correct pack-size was chosen instead) 
Note: Expiry of both available stocks is same, so I'd definitely expect the pack-size 1 being chosen over the next here!

## Recordings
https://github.com/msupply-foundation/open-msupply/assets/59193490/7a055662-0665-475c-9486-217f399bcc96

## Your environment 🌱

 - Version: v1.5.4
 - Platform:
   - browser (extra points if you tell us which one)

fergie-nz:
Some notes to add to this:

Allocate logic is currently in two different places. In the front end, and also here in the back end.

This logic might be better abstracted in the front end so it can be used for both allocation use cases.

Also it is noted that best case behaviour might be different depending on a number of factors:
-Expiry date of stock
-size of order
-Simplicity of order (ie if there are stock sizes of 1,10,100,1000 available, and an order of 999 is placed, it might be easier to order one pack of 1000 rather than nine of each 1,10, and 100.)

fergie-nz:
This issue is relevant to possible refactor:

#2908

mark-prins:
Thanks - have added to the refactor of outbound shipment edit line modal

andreievg:
Linked issue: https://github.com/msupply-foundation/open-msupply/issues/3252

PR #2880, opened 2024-02-01
Outbound shipment allocation - unable to type multi digit quantities

### LINKED ISSUE #2727: Won't let me issue any quantity starting with '1' in outbound shipment in case of stocks are available in multiple pack sizes other than 1
## What went wrong? 😲
 
In outbound shipment, when there are different pack sizes (other than 1) available, weirdly it won't let me issue any quantity starting with `1`, like `101` or `11` or so.

## Expected behaviour 🤔
Regardless of what's available, it should not interfere with what the user wants to input

## How to Reproduce 🔨
1. Make sure you have stocks available of different pack-sizes, just not 1.
2. Now make a new outbound shipment and add that item
3. When you select that item, make sure you see the below window stating issue [quan] `units in packs of` .
image

4. Now try to issue any number starting with 1, like try putting '107' or '12` or so
5. See error: the '1' in your input automatically switches to other number (depending the pack size that's available I think) For me, '107' is changing to '307', '12' is changing to '33'. VERY WEIRD BUG! 

## Screenshots/Recordings
Mobile recordings: [didn't upload desktop recording since it's difficult to show which key I'm pressing in my keyboard in a single screen]

https://github.com/msupply-foundation/open-msupply/assets/59193490/926d17f0-e862-43e2-94df-5172bb5f231f

https://github.com/msupply-foundation/open-msupply/assets/59193490/f9a0880e-d55e-4bd4-8624-cb02f5aa8b6f

## Your environment 🌱

 - Version: v 1.5.03, v1.5.04
 - Platform:
   - [x] android (tablet)
   - [x] browser (extra points if you tell us which one)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [x] server (windows)

mark-prins:
Fixes #2727

# 👩🏻‍💻 What does this PR do? 
 
Debounces the issue quantity change to allow entry of multi-digit values.
currently, if you have a range of unhelpful pack sizes available, then the allocation logic will round up your quantity, overwriting the entered digits and preventing you from entering in the desired quantity.

* Have some stock for an item, which has pack sizes other than `1`
* Create an outbound shipment, add that item
* Issue a quantity for a number which contains multiple digits and starts with `1` e.g. `10` or `105` etc.

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/common/src/hooks/useDebounce/useDebouncedValueCallback.ts
+++ b/client/packages/common/src/hooks/useDebounce/useDebouncedValueCallback.ts
@@ -9,10 +9,17 @@ import { FnUtils } from '@common/utils';
 export const useDebouncedValueCallback = <T extends (...args: any[]) => any>(
   callback: T,
   depsArray: DependencyList,
-  wait = 500
+  wait = 500,
+  callbackDepsArray: DependencyList = []
 ): ((...args: Parameters<T>) => Promise<ReturnType<T>>) => {
-  const memoizedCallback = useCallback(FnUtils.debounce(callback, wait), []);
-  const debounced = useCallback(memoizedCallback, depsArray);
+  const memoizedCallback = useCallback(
+    FnUtils.debounce(callback, wait),
+    callbackDepsArray
+  );
+  const debounced = useCallback(memoizedCallback, [
+    ...depsArray,
+    memoizedCallback,
+  ]);
 
   return debounced;
 };
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
@@ -68,6 +68,7 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
   const { Modal } = useDialog({ isOpen, onClose, disableBackdrop: true });
   const [currentItem, setCurrentItem] = useBufferState(item);
   const [isAutoAllocated, setIsAutoAllocated] = useState(false);
+  const [okDisabled, setOkDisabled] = useState(false);
 
   const { mutateAsync: insertBarcode } = useOutbound.utils.barcodeInsert();
   const { status } = useOutbound.document.fields('status');
@@ -219,7 +220,7 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
       }
       okButton={
         <DialogButton
-          disabled={!currentItem}
+          disabled={!currentItem || okDisabled}
           variant="ok"
           onClick={() => handleSave(onClose)}
         />
@@ -244,6 +245,7 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
           showZeroQuantityConfirmation={showZeroQuantityConfirmation}
           hasOnHold={hasOnHold}
           hasExpired={hasExpired}
+          setOkDisabled={setOkDisabled}
         />
 
         <TableWrapper
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
@@ -246,6 +246,7 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
           hasOnHold={hasOnHold}
           hasExpired={hasExpired}
           setOkDisabled={setOkDisabled}
+          draftStockOutLines={draftStockOutLines}
         />
 
         <TableWrapper
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
@@ -13,7 +13,7 @@ import {
   Typography,
   useFormatNumber,
   useDebounceCallback,
-  FnUtils,
+  useDebouncedValueCallback,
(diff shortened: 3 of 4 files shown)


client/packages/common/src/hooks/useDebounce/useDebouncedValueCallback.ts
```diff
@@ -9,10 +9,17 @@ import { FnUtils } from '@common/utils';
...
   depsArray: DependencyList,
-  wait = 500
+  wait = 500,
+  callbackDepsArray: DependencyList = []
 ): ((...args: Parameters<T>) => Promise<ReturnType<T>>) => {
-  const memoizedCallback = useCallback(FnUtils.debounce(callback, wait), []);
-  const debounced = useCallback(memoizedCallback, depsArray);
+  const memoizedCallback = useCallback(
+    FnUtils.debounce(callback, wait),
+    callbackDepsArray
```
mark-prins, line 17: adding in dependencies for this callback - otherwise the function is memoized, and in our case, creates a captured enclosure of the `draftOutboundLines` which then gives a 'no stock available' error when allocating

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -68,6 +68,7 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
   const { Modal } = useDialog({ isOpen, onClose, disableBackdrop: true });
   const [currentItem, setCurrentItem] = useBufferState(item);
   const [isAutoAllocated, setIsAutoAllocated] = useState(false);
+  const [okDisabled, setOkDisabled] = useState(false);
```
mark-prins, line 71: disabling ok until the debounce has happened. without it, you could click ok before the allocation has happened and get 0 stock allocated

CarlosNZ:
I can't replicate the original bug as described in the issue. Do you have a data file/specific example where it's apparent?

mark-prins:
I just used the current datafile - you don't need anything too special.
You will have to add some stock lines for an item though. 
To add, create a stocktake, choose an item, add some batches and finalise.
Make sure that the pack size isn't 1  for all batches.

If you already have stock with pack size 1, edit the stock line and change the pack size.

Then, when you add an outbound shipment line for that item, if you enter 1 as the quantity, the allocation routine will bump up the value to the available pack size (as long as there isn't a pack size 1 row available).

CarlosNZ (approved):
Nice, looks good, working well.

Noticed a small, unrelated problem though:
Screenshot 2024-02-02 at 3 23 14 PM
Substitutions here are not correct.

In `utils.tsx`, the wrong property names are being used:
```ts
 if (allocatedQuantity !== requestedQuantity && allocatedQuantity > 0) {
    alerts.push({
      message: t('messages.over-allocated', {
        quantity: format(allocatedQuantity),
        issueQuantity: format(requestedQuantity),
      }),
      severity: 'warning',
    });
    return alerts;
  }
```

Need to change `quantity` and `issueQuantity` to `allocatedQuantity` and `requestedQuantity`


One other thing -- feel free to tell me in person. What's the logic in the rounding up that's going on here? It seems to round up in bigger jumps than it needs to. If I have an item with pack sizes of `2` and `35`, I can enter `2`, `4`, `35`, and `37` as valid quantities as they can be made up from the pack sizes available. However, if I enter `1` or `36`, it'll round up to `35` and `70` respectively. I would have though it should round to the smallest number that can be made up from the available pack sizes -- in which case `1` would round to `2` and `36` would round to `37`. Is this not what it's supposed to do?

client/packages/common/src/hooks/useDebounce/useDebouncedValueCallback.ts
fergie-nz: I wonder if you would ever need to differentiate between the depsArray and the callbackDepsArray? ie would you  ever need to refresh debounced while leaving the momoizedCallback stale? And if not, is it better to consolidate the two?

mark-prins: yes, I do need the differentiation. The original use-case for this hook required that the memoized function be static, while the second `useCallback` is updated based on the data which was to be provided to the first.

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
```diff
@@ -79,13 +84,16 @@ export const OutboundLineEditForm: React.FC<OutboundLineEditFormProps> = ({
...
   const unit = item?.unitName ?? t('label.unit');
 
-  const updateIssueQuantity = (quantity: number) => {
-    setIssueQuantity(
-      Math.round(
-        quantity / Math.abs(Number(packSizeController.selected?.value || 1))
-      )
-    );
-  };
+  const updateIssueQuantity = useCallback(
```
fergie-nz, line 87: It takes me a long time to internalise why a memoized function is also needed here.

I'm slightly fearful of needing to make changes here and it getting very complicated.

Maybe this is a pretty typical scenario? Maybe it is just me? I'm curious if it could be refactored somehow.

fergie-nz (approved):
That is awesome Mark, thank you for doing all of this. Works great.

fergie-nz:
> One other thing -- feel free to tell me in person. What's the logic in the rounding up that's going on here? It seems to round up in bigger jumps than it needs to. If I have an item with pack sizes of `2` and `35`, I can enter `2`, `4`, `35`, and `37` as valid quantities as they can be made up from the pack sizes available. However, if I enter `1` or `36`, it'll round up to `35` and `70` respectively. I would have though it should round to the smallest number that can be made up from the available pack sizes -- in which case `1` would round to `2` and `36` would round to `37`. Is this not what it's supposed to do?

Related to issue #2668 on allocating correctly / preference of allocating

PR #2892, opened 2024-02-05
Add some more docs about the used backend layers

### LINKED ISSUE #2858: Improve backend docs about architecture/layers/motivation

clemens-msupply:
Fixes #2858 

Add some docs about the backend layers. Question: is there a better place/dir to put these files? There are a couple of files scattered throughout the repo...

### DIFF
--- a/server/architecture.md
+++ b/server/architecture.md
@@ -3,13 +3,13 @@
 Most of the source code in this repository is split into three layers.
 These layers are the repository layer, the service layer, and the API layer.
 There are a couple of advantages to maintain these three layers.
-For example, the repository layer decouples the business logic from database details.
-Similarly, the business logic in the service layer is independent of the API layer, e.g. a single service can be used for Graphql or Rest endpoints.
+For example, the repository layer decouples the business logic from the database details.
+Similarly, the business logic in the service layer is independent of the API layer, e.g. a single service can be used for GraphQL or REST endpoints.
 This architecture is very similar to the so-called "clean architecture" in domain driven design.
 
 The three layers should not have any circular dependencies.
 For example, the repository layer can be used by the service layer or the API layer while the service layer can only be used by the API layer.
-This is enforced by putting the layers in separate Rust crates which can't have circular dependencies either.
+This is enforced by having the layers in separate Rust crates which can't have circular dependencies either.
 
 ## Repository Layer
 
@@ -65,8 +65,8 @@ The API layer can use one or more services to query or mutate data.
 While we mainly use GraphQL for our endpoints it would easily be possible to provide REST endpoints by reusing the existing services.
 
 Before doing the service call the API layer calls an auth service to verify that a user is allowed to call the current endpoint.
-Note, this auth check could be further moved into service layer.
-However, this would either make a service very specific to particular endpoint, i.e. service becomes hard to reuse, or another endpoint specific service would need to be introduced.
+Note, this auth check could be moved into service layer.
+However, this would either make a service very specific to particular endpoint, i.e. service becomes hard to reuse, or another endpoint specific auth-service would need to be introduced.
 
 # Horizontal and Vertical Layers
 
--- a/server/architecture.md
+++ b/server/architecture.md
@@ -30,7 +30,7 @@ struct MyDieselRow {
 }
 ```
 
-the domain object should only contains the information used in the business logic, i.e. the `name_id` instead of the internal `name_link_id`:
+the domain object should only contain the information used in the business logic, i.e. the `name_id` instead of the internal `name_link_id`:
 
 ```rust
 pub struct MyEntity {
--- a/server/architecture.md
+++ b/server/architecture.md
@@ -1,0 +1,77 @@
+# Layers
+
+Most of the source code in this repository is split into tree layers.
+These layers are the repository layer, the service layer, and the API layer.
+There are a couple of advantages to maintain these three layers.
+For example, the repository layer decouples the business logic from database details.
+Similarly, the business logic in the service layer is independent of the API layer, e.g. a single service can be used for Graphql or Rest endpoints.
+This architecture is very similar to the so-called "clean architecture" in domain driven design.
+
+The three layers should not have any circular dependencies.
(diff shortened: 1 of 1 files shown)


server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+For example, entities returned from the repository layer look the same for Postgres and Sqlite.
+This database abstraction should be as small as possible, e.g. instead of adding specific queries for specific use-cases, existing query method should be reused where possible.
+
+Furthermore, the repository layer keeps the service logic decoupled from database schema changes.
+For example, a table schema change should only affect the repository that uses the changed table.
+This can be achieved by mapping the data returned from the database to a "domain object" which is database independent.
+For example, rows from the mSupply tables `name` and `item` are not referenced directly by their `id` but joined via a link table to allow name and item merges.
+While the database row may look like:
+
+```rust
```
clemens-msupply, line 26: The following section is more how it should be rather than how it is implemented (from my experience with column rename for the merge changes). Let me know what you think about it.

server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+The [`ServiceProvider`](service/src/service_provider.rs) struct contains all available services.
+The `ServiceProvider` should be used in higher layer, e.g. in GraphQl endpoint to get a service instance.
+
+The `ServiceProvider` allows us to configure which services are used.
+This is mainly used for testing purposes where services can easily be swapped out for a mock service.
+For example, when testing GraphQL related logic such as data mapping a mock service can be configured to always return the required data for a specific test.
+
+## API Layer
+
+The API layer provides an interface for the server, e.g. GraphQL or REST endpoints used by the browser UI or other users.
```
roxy-dao, line 65:
```suggestion
The API layer provides an interface for the server through the GraphQL or REST endpoints used by the browser UI or other users.
```

server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+The API layer can use one or more services to query or mutate data.
+While we mainly use GraphQL for our endpoints it would easily be possible to provide REST endpoints reusing the existing services.
```
roxy-dao, line 67:
```suggestion
While we mainly use GraphQL for our endpoints it would easily be possible to provide REST endpoints by reusing the existing services.
```

server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+
+Before doing the service call the API layer calls an auth service to verify that a user is allowed to call the current endpoint.
+Note, this auth check could be further moved into service layer.
+However, this would either make a service very specific to particular endpoint, i.e. service becomes hard to reuse, or another endpoint specific service would need to be introduced.
```
roxy-dao, line 71:
```suggestion
Note, this auth check could be moved into service layer.
However, this would restrict the service to a particular endpoint, making the service hard to reuse.
```

roxy-dao (changes requested):
Awesome~! Thanks Clemens. Very informative~.

I think it's fine to keep the information where is it right now unless you want to move information about the specific layers to where they're relevant.

And just a few comments below~:

server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+This architecture is very similar to the so-called "clean architecture" in domain driven design.
+
+The three layers should not have any circular dependencies.
+For example, the repository layer can be used by the service layer or the API layer while the service layer can only be used by the API layer.
+This is enforced by putting the layers in separate Rust crates which can't have circular dependencies either.
+
+## Repository Layer
+
+The repository layer provides an abstraction over the used database.
+For example, entities returned from the repository layer look the same for Postgres and Sqlite.
```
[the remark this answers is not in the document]
clemens-msupply: Think I keep it as an example.

server/architecture.md
```diff
@@ -0,0 +1,77 @@
+# Layers
+
+Most of the source code in this repository is split into tree layers.
+These layers are the repository layer, the service layer, and the API layer.
+There are a couple of advantages to maintain these three layers.
+For example, the repository layer decouples the business logic from database details.
+Similarly, the business logic in the service layer is independent of the API layer, e.g. a single service can be used for Graphql or Rest endpoints.
```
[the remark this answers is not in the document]
clemens-msupply: I think its better to keep it as example otherwise it sounds like there are no other advantages... I use your other changes though!

server/architecture.md
```diff
@@ -0,0 +1,77 @@
...
+This architecture is very similar to the so-called "clean architecture" in domain driven design.
+
+The three layers should not have any circular dependencies.
+For example, the repository layer can be used by the service layer or the API layer while the service layer can only be used by the API layer.
+This is enforced by putting the layers in separate Rust crates which can't have circular dependencies either.
```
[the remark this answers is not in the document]
clemens-msupply: I like my examples, but took you grammar fix.

server/architecture.md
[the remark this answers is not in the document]
roxy-dao: Oh sorry. I misunderstood. Cause the first sentence says "a couple of advantages"... so it sounds like there are only 2 advantages even with the `For example`

server/architecture.md
clemens-msupply: I rephrased it a bit more...

server/architecture.md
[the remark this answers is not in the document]
roxy-dao: Can you still capitalise `GraphQL` and `REST`?

clemens-msupply:
I renamed API layer to the more common name: "controller layer"

roxy-dao (approved):
Thanks Clemens~

server/architecture.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+However, currently this pattern is only followed in a few places because for most entities the database object looks exactly like the domain object, so the database object is used as the domain object to avoid trivial object mapping.
+In this case a simple type def should be used, keeping the internal database struct private:
+
+```rust
+pub type MyEntity = MyDieselRow;
+```
+
```
clemens-msupply, line 48: TODO: add composite types from query repo

PR #2899, opened 2024-02-06
2290 sqlite db config

### LINKED ISSUE #2290: Configuration option for database path

The request was made to be able to specify the path of the sqlite database, if sqlite is being used.

In mSupply currently, a common configuration is to have the datafile on a separate drive, to allow separation from the program files, enable simpler backup configuration and to have the data stored on a drive separate to the OS, which in theory allows for faster access.

Add an optional configuration setting to the `yaml` files, so specify the path of the database. This will be ignored if postgreSQL is being used.

If set, the SQLite file is loaded from the path, if not the current setting is used: which is the path of the executing binary.

Perhaps you'd like to do #817 at the same time?

fergie-nz:
Fixes #2290

# 👩🏻‍💻 What does this PR do? 
Adds optional database path in yaml file. This allows specification of database path

Add a path or nested path to local.yaml file, then start the back end with cargo run.

You should see the sqlite database set up in the specified path.


server/repository/src/database_settings.rs
```diff
@@ -72,8 +73,19 @@ impl DatabaseSettings {
...
+                format!("{}/{}", path, self.connection_string())
+            }
+            None => self.connection_string(),
+        };
+        return result;
+    }
+
     pub fn connection_string_without_db(&self) -> String {
-        self.connection_string()
+        self.database_path()
```
jmbrunskill, line 88: Does this work with postgres?

jmbrunskill: Sorry I see it's behind the `#[cfg(all(not(feature = "postgres"), not(feature = "memory")))]` so I think it should be ok.

jmbrunskill: but you will need to add this to the postgres `get_test_db_settings`
```
error[E0063]: missing field `database_path` in initializer of `DatabaseSettings`
  --> repository/src/test_db/postgres.rs:11:5
   |
11 |     DatabaseSettings {
   |     ^^^^^^^^^^^^^^^^ missing `database_path`
```

Try this PR with `cargo test --features=postgres` and probably start up the app in postgres mode as well for good measure :)

jmbrunskill (changes requested):
`cargo test --features=postgres` isn't working, would be good to have that fixed before we merge.

server/repository/src/database_settings.rs
clemens-msupply: This `connection_string_without_db` is only used for the postgres feature right? could you please try to remove it for sqlite and memory? think it should work without it...

jmbrunskill (approved):
Looks good thanks Ferg!

PR #2907, opened 2024-02-07
2572 ok next short cuts

### LINKED ISSUE #2572: Shortcuts for "OK" and "OK & Next" buttons

image

When entering stock takes on a PC, I want to do fast data entry. Part of that is quickly pressing "OK & Next" with a shortcut, so I can save changes on this row and begin editing the next row without my hands leaving the keyboard... also in development it makes spamming in data a bit easier 😁. 

Not sure if bug or request, but same same. Allegedly some modals might already have this but I checked outbound shipment and NADA!

Should be a shortcut for OK & Next, OK and Cancel. I suggest:

- "Enter" for OK & Next
- "ctrl + enter" for OK
- "esc" Seems to already cancel modals 🙂 

Are their any modals that only have "OK", No "OK & Next"? In those cases the shortcut for OK can be "enter"

### Additional context 💌

Not sure what the desired behaviour should be... but when I press `cmd+k` to get the shortcuts menu it pops up behind the modal so I can't really see the options though I doubt it gives options for the modal context, but for the page behind the modal.

Tangent because I think it'd be neat: I noticed that the "add batch" button has a (+) to indicate the shortcut. But, doing that for all buttons will obviously look a bit ugly. It'd be COOL if you could press and hold a button combo (say "alt+ctrl") that either changes each label to include the shortcut combination OR a little tool tip 😃.

### Your environment 🌱

 - Version: dev 1.6.0 MacOS Firefox dev edition

lache-melvin:
Fixes #2572

# 👩🏻‍💻 What does this PR do? 
 

Adds enable-able key bindings to the useDialog Modal component, to allow hitting Enter/Ctrl+Enter to Ok+Next/Ok from anywhere in the modal.

Some unit tests for the handler, otherwise open some modals and hit Ctrl+enter/Ctrl+Shift+enter 😅 

Note: haven't enabled the keybindings on all the modals yet... partly because I'm still learning my way around to find them lol, but also I figured we should solve the below first.

It has come to my attention that this is actually had a needs discussion flag on it - word vomiting here but a real life convo may be easier...

Using the enter key as the main shortcut does make sense, it's the logical submit key. However I'm running into some issues given the enter key also used in other components of the Modal (e.g. selecting from a dropdown, entering multiline text...)

The ideal solution would be to wrap the modal contents in a `<form>` tag, and place the `submit` type on the button that we want to call on Enter. Then we get the default browser form behaviour - hitting enter fires the submit button's onClick handler, except when a form field that accepts Enter as an input is focussed (like the multiline inputs).

However, there's a few flaws to this plan:
- For each ok/next button handler, we would need to add `e.preventDefault()` to stop the HTTP POST default browser behaviour on submit (More refactoring, something else to remember whenever adding a new modal...)
  - (Note to self, if we go with this option, WrappedNextButton would also need to pass through the event to the onClick handler!)
- Because the okButton/nextButton props on the useDialog Modal are optional, and those ok/next buttons could be any react element, we can't really rely on the `type=submit` being present
- How might we assign `submit` to the Next button if present, otherwise to the Ok button?
  - Could be developer responsibility when they define new Modals, add type on the dialogButton that they want to fire on enter (seems like something that would be likely to forget?)
  - We could add logic to useDialog that will assign the type depending on which buttons are provided (perhaps a little hacky/magicky)
- We still need a separate event handler for the Ctrl+Enter (i.e. OK when OK+Next present) - handling the shortcuts in two different places?

An alternative would be to use Ctrl+Enter and Ctrl+Shift+Enter? Then we'd have no competition for the key combos and could just fire the onNext/onOk handlers. Not quite as intuitive, but could be resolved with training. Means we're not leaning on default browser behaviour where we could be... but might be the simplest solution?

## 📃 Documentation

Might need to update useDialog storybook once implementation decided


client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useDraftStocktakeLines.ts
```diff
@@ -13,6 +13,7 @@ export const useDraftStocktakeLines = (
   const [draftLines, setDraftLines] = useState<DraftStocktakeLine[]>([]);
 
   useEffect(() => {
+    if (!item) setDraftLines([]);
```
lache-melvin, line 16: Not directly related, but we weren't clearing the stocktake draft lines on ok+next, so the buttons were still enabled (which would submit the previous item again, causing an error)

lache-melvin:
Have gone with Ctrl+Enter and Ctrl+Shift+Enter - I think the time investment to get something good working with the Enter key is probably not worth it... though very open to opinions on that please!

Also - these are _enableable_ keybindings... but I've enabled them in most places, which is meaning another prop in a lot of places, and I don't think they're likely to interfere with anything - should I change it to disableable? Or is it best not to change the default behaviour of the common component?

Where do I find the source for the public docs to add something about the shortcuts here? https://docs.msupply.foundation/docs/introduction/tutorial/#keyboard-shortcuts

mark-prins:
We should have a quick chat to decide where to go with this one. My initial thoughts:

- Love the implementation and tests wrapped around it! 
- However, I don't like CTRL+ENTER / CTRL+SHIFT+ENTER because that isn't intuitive
- I would have the `useDialog` implement enter handling unless explicitly told not to, so that it is the default behaviour
- My preference is to add a form and use onSubmit or submit buttons (though the MUI buttons force the use of `type="button"` 

We'd also need to add a visual indicator to show which button is the default / Enter activated button.

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -163,15 +163,12 @@ export const InboundLineEdit: FC<InboundLineEditProps> = ({
...
       >
         {isLoading ? (
           <BasicSpinner messageKey="saving" />
         ) : (
-          <form
-            autoComplete={
-              'on' /* Required for previously entered batches to be remembered and suggested in future shipments */
-            }
-          >
+          <>
```
lache-melvin, line 171: Can't have a <form> within a <form>, so added optional autocomplete prop for this case

client/packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -149,24 +160,53 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
...
 
+    if (okButton) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { onClick, type, ...restOfOkButtonProps } = okButton.props;
+
+      WrappedOkButton = React.cloneElement(okButton, {
+        onClick: defaultPreventedOnClick(onClick),
+        // If the next button is not present/disabled, the ok button should be a submit button (allow firing on enter key press)
+        type:
+          !disableOkKeyBinding && (!nextButton || nextButton.props.disabled)
```
lache-melvin, line 202: curious what you think of enabling 'OK on enter' if the next button is disabled? The use case was going through existing stocktake/shipment lines, you could hit Enter to OK & Next your way through them, and then on the last item, when next wasn't available, Enter would fire the OK and close the modal

lache-melvin:
> We'd also need to add a visual indicator to show which button is the default / Enter activated button.

Do you have an idea for how to indicate this? My thought was to make the OK button outlined when there is an OK & Next button, but I'm not sure how to do this elegantly... I've found at least one place where the okButton isn't a DialogButton (it's a LoadingButton) which means I don't think we can create a new DialogButton variant to handle this UX?

client/packages/common/src/hooks/useDialog/useDialog.tsx
mark-prins: I like it. and at the same time feel vaguely that I should be concerned that it isn't consistent or something.

As a user, the behaviour you have is perfect - when running through a list of things, you would be clicking OK & Next repeatedly until then end when you click OK. so yep, all good.

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
mark-prins: yes, nicely handled

mark-prins (approved):
🎉 
that is so nice to use! we should have done this ages ago 🙏

client/packages/common/src/hooks/useDialog/useDialog.tsx
lache-melvin: > I like it. and at the same time feel vaguely that I should be concerned that it isn't consistent or something.

My sentiments exactly 😁

Issue #2814, opened 2024-01-24
Update f/e dependencies


mark-prins:
## Is your feature request related to a problem? Please describe 👀

As per title. This one was waiting for the feature branch merge

## Describe the solution you'd like 🎁
`yarn outdated --json | $(yarn bin)/format-yarn-outdated --format markdown`

# Major
| Package | Current | Wanted | Latest | Package Type | URL | CHANGELOG |
|---|---|---|---|---|---|---|
| @electron-forge/cli | 6.2.1 | 6.4.2 | 7.2.0 | @openmsupply-client/electron | devDependencies | https://github.com/electron/forge#readme |  |
| @electron-forge/maker-dmg | 6.2.1 | 6.4.2 | 7.2.0 | @openmsupply-client/electron | devDependencies | https://github.com/electron/forge#readme |  |
| @electron-forge/maker-squirrel | 6.2.1 | 6.4.2 | 7.2.0 | @openmsupply-client/electron | devDependencies | https://github.com/electron/forge#readme |  |
| @electron-forge/maker-zip | 6.2.1 | 6.4.2 | 7.2.0 | @openmsupply-client/electron | devDependencies | https://github.com/electron/forge#readme |  |
| @electron-forge/plugin-webpack | 6.2.1 | 6.4.2 | 7.2.0 | @openmsupply-client/electron | devDependencies | https://github.com/electron/forge#readme |  |
| @graphql-codegen/near-operation-file-preset | 2.5.0 | 2.5.0 | 3.0.0 | openmsupply-client | devDependencies | https://github.com/dotansimha/graphql-code-generator-community#readme |  |
| @graphql-codegen/typescript-graphql-request | 5.0.0 | 5.0.0 | 6.1.0 | openmsupply-client | devDependencies | https://github.com/dotansimha/graphql-code-generator-community#readme |  |
| @graphql-codegen/typescript-msw | 1.1.6 | 1.1.6 | 2.1.0 | openmsupply-client | devDependencies | https://github.com/dotansimha/graphql-code-generator-community#readme |  |
| @swc/cli | 0.1.62 | 0.1.65 | 0.2.0 | openmsupply-client | devDependencies | https://github.com/swc-project/cli#readme |  |
| @testing-library/jest-dom | 5.17.0 | 5.17.0 | 6.2.1 | openmsupply-client | devDependencies | https://github.com/testing-library/jest-dom#readme |  |
[...]

PR #2914, opened 2024-02-08
2812 gql codegen unrunnable code

### LINKED ISSUE #2812: GraphQL codegen generates unrunnable code
## What went wrong? 😲
 
Some changes to the operations.graphql files have resulted in 
* code that cannot run
* code that fails our linting pre-commit

**/client/packages/invoices/src/OutboundShipment/api/operations.graphql**

The fragment import path is incorrect

```
import { StockOutLineFragmentDoc } from '../../../../../../packages/invoices/src/StockOut/operations.generated';
```

**/client/packages/invoices/src/Prescriptions/api/operations.graphql**

The fragment import path is incorrect

```
import { StockOutLineFragmentDoc } from '../../../../../../packages/invoices/src/StockOut/operations.generated';
```

**/client/packages/invoices/src/StockOut/operations.generated.ts**

This operations file only has fragments; graphql codegen doesn't cope with this, and adds the `defaultWrapper` function which isn't used and fails linting.

**client/packages/requisitions/src/RequestRequisition/api/operations.generated.ts**

The fragment import path is incorrect

```
import { NameRowFragmentDoc } from '../../../../../../packages/system/src/Name/api/operations.generated';
```

**client/packages/system/src/Stock/api/operations.generated.ts**

The fragment import path is incorrect

```
import { LocationRowFragmentDoc } from '../../../../../../packages/system/src/Location/api/operations.generated';
```

## Expected behaviour 🤔
Running `yarn generate` should generate good code.

I've been unable to force the fragment imports to work correctly when importing from another file.
The closest I've got is to set the `importAllFragmentsFrom` prop (ref: [near-operations-file-preset docs](https://the-guild.dev/graphql/codegen/plugins/presets/near-operation-file-preset)) and move _all_ fragments into a single file.

This file would then also require a dummy query in it, so that the linting passes.

Other references:
* [graphql-code-generator source](https://github.com/dotansimha/graphql-code-generator/tree/master)
* [Fragment import issue](https://github.com/dotansimha/graphql-code-generator/issues/4307)
* GraphQL code generator [typescript reference](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript)

## How to Reproduce 🔨

1. Run `yarn generate`

## Your environment 🌱

 - Version: 
 - Platform:
   - [ ] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

lache-melvin:
Fixes #2812

# 👩🏻‍💻 What does this PR do? 
 

Somehow stumbled upon the fact that if you run codegen once, you get the incorrect import path for the shared fragments, but if you run it again they resolve themselves?!

Understanding _why_ this second codegen run resolves everything was becoming quite the rabbit hole... I'm uncertain how I feel about a hacky fix like this rather than actually configuring the codegen correctly, but as described in the issue, it seems that if we continue with the colocated operations files, the only other solution would be to put all the fragments in one file (which sort of defeats the define-next-to-use thing we've got going on...)

Have added a dummy query to the fragment-only file for now... the typescript-graphql-request plugin needs there to be at least one query. Will come back to this tomorrow...


mark-prins:
oho! good find - I didn't notice that running twice works.. though that is odd

lache-melvin:
I've had a bit of a play around, and it _appears_ that running codegen twice is working reliably, including when adding new fragment/using them in new places. It is a little hacky, but given it's unlikely we'll be changing our codegen setup much, I think its ok?

Same feeling for the dummy query in the fragment only file - I think the `typescript-graphql-request` adds sufficient value everywhere else, and defining it is fairly harmless. I don't think it will cause much confusion.

The other option there would be to move those fragment definitions to one of the operations files they are used in?

mark-prins (approved):
👍 Love it! no idea why it works, or, more to the point, why it doesn't in the first run. 
Works for me though, this scratches the itch.
Planning to upgrade graphql-codegen (#2814) and will retest the single run after that 😉

clemens-msupply (dismissed):
> The other option there would be to move those fragment definitions to one of the operations files they are used in?

Yes please, I think we shouldn't have fragments defined twice because it can become very confusing when using the generated code... Think fixing the root problem is a better solution.

lache-melvin:
> Yes please, I think we shouldn't have fragments defined twice because it can become very confusing when using the generated code... Think fixing the root problem is a better solution.

Ah, they're not defined twice, we just have one file (invoices/src/StockOut/operations.graphql) where there are some fragments defined, but they are only used in a different operations files (e.g. for the StockOutLine fragment, that is in OutboundShipment & Prescriptions.) I think they are defined in their own file to keep them close to the other StockOut code, where their generated TS types are used?

lache-melvin:
> Planning to upgrade graphql-codegen (#2814) and will retest the single run after that 😉

Oh this will be interesting to know if it resolves!

clemens-msupply:
> Ah, they're not defined twice, we just have one file (invoices/src/StockOut/operations.graphql) where there are some fragments defined, but they are only used in a different operations files (e.g. for the StockOutLine fragment, that is in OutboundShipment & Prescriptions.) I think they are defined in their own file to keep them close to the other StockOut code, where their generated TS types are used?

Ah ok don't mind them, even though having them defined where they are used sounds better to me..

lache-melvin:
> Ah ok don't mind them, even though having them defined where they are used sounds better to me..

Yeah I went a little back and forth on this one - define them where they are used in another graphQL query, or where the generated types are most heavily used? Will leave how it is for now, as I think it makes sense from a TS code location perspective (i.e. `import { StockOutLineFragment } from '../../StockOut';` but I'm up for discussing it further!

PR #2928, opened 2024-02-13
2921 select lines from inbound

### LINKED ISSUE #2921: Select Inbound Shipment lines to return to supplier
Epic: #2897 

 - Can select lines from inbound shipment for supplier return 
   - [x] Add option to actions dropdown
 - New return items UI (with all selected lines), for supplier return
   - [x] New modal showing selected lines, with input field for the quantity of items to return
   - [x] Mock API request for data needed for this view 
Api should really provide anything that is required for return items view, based on provided ids of lines

Image

lache-melvin:
Fixes #2921

# 👩🏻‍💻 What does this PR do? 
 

Adds first page of the return items from inbound shipment modal:
Screenshot 2024-02-13 at 12 56 17 PM

- Open an existing/create a new Inbound Shipment
- Select some of the shipment lines
- Select 'Return to supplier'
- The Return Items modal opens (populated with a dummy line)
- You can edit the quantity to return


client/packages/invoices/src/InboundShipment/api/hooks/line/useNewSupplierReturnLines.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  useNotification,
+  useQuery,
+  useTableStore,
+  useTranslation,
+} from '@openmsupply-client/common';
+import { useInboundApi } from '../utils/useInboundApi';
+import { useInboundRows } from './useInboundRows';
+
+export const useNewSupplierReturnLines = () => {
```
lache-melvin, line 10: Have added this as an inbound shipment API hook... but it may be better located with other returns-related API code?

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
```diff
@@ -49,6 +53,14 @@ export const Toolbar: FC = () => {
   ]);
   const { isGrouped, toggleIsGrouped } = useInbound.lines.rows();
   const t = useTranslation('replenishment');
+
+  const generateNewReturnLines = useInbound.lines.newReturnLines();
+
+  const onReturn = async () => {
+    const lines = await generateNewReturnLines();
+    if (lines) onReturnLines(lines);
+  };
```
lache-melvin, line 62: I'm not sure how I feel about giving the Toolbar the responsibility of querying for the new return lines, and only opening the modal if there are any.

Part of me wanted to do the querying for new return lines within the modal, but I don't like the idea of rendering the modal, only to determine there are no lines selected and having to close itself again?

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
```diff
@@ -49,6 +73,15 @@ export const Toolbar: FC = () => {
...
   const { isGrouped, toggleIsGrouped } = useInbound.lines.rows();
   const t = useTranslation('replenishment');
+
+  const selectedIds = useSelectedIds();
+  const generateNewReturnLines = useReturns.lines.newReturnLines(selectedIds);
+
+  const onReturn = async () => {
+    const lines = await generateNewReturnLines();
+    if (lines) onReturnLines(lines);
+  };
```
lache-melvin, line 83: I'm not sure how I feel about giving the Toolbar the responsibility of querying for the new return lines, and only opening the modal if there are any.

Part of me wanted to do the querying for new return lines within the modal, but I don't like the idea of rendering the modal, only to determine there are no lines selected and having to close itself again?

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -132,6 +132,7 @@
   "label.auth-status": "Approval Status",
   "label.available-packs": "Available (packs)",
   "label.available-quantity": "Available quantity: {{number}} units",
+  "label.available-quantity-for-return": "Quantity Available for Return",
```
CarlosNZ, line 135: Wonder if these should be in `distribution.json`?

client/packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -14,6 +15,7 @@
   "heading.are-you-sure": "Are you sure?",
   "heading.consumption-history": "Consumption History (monthly)",
   "heading.related-documents": "Related documents",
+  "heading.return-items": "Return Items",
```
CarlosNZ, line 18: nit: just having a cursory look at other areas, it would seem standard capitalisation might be better for these: "Return item", "Select quantity", etc. (rather than Title Case)

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
CarlosNZ: What about, instead of the Toolbar calling `onOpenReturns` directly, why not have it call a new function in the DetailView component that checks the selected lines and then opens the modal if appropriate? You could even have it so that it passes another prop `hasSelectedLines` to the Toolbar, and then have the Toolbar not even display the "Return" option unless its `true`?

Is it easy to get the selected state of all the lines within DetailView? (I would have thought so, but I'm not seeing it obviously available from a quick glance)

client/packages/invoices/src/Returns/api/api.ts
```diff
@@ -0,0 +1,14 @@
+import { Sdk } from './operations.generated';
```
CarlosNZ, line 1: I don't understand the structure of these Api folders at all? Is it the same pattern you used for HSH?

I could really do with a lesson on how it works/why it's structured like this, as I find it a massive headache to do even the simplest things.

client/packages/invoices/src/Returns/api/hooks/utils/useReturnsApi.ts
```diff
@@ -0,0 +1,17 @@
...
+    base: () => ['returns'] as const,
+    count: () => [...keys.base(), 'count'] as const,
+    detail: (id: string) => [...keys.base(), storeId, id] as const,
+    newReturns: () => [...keys.base(), storeId, 'newReturns'] as const,
+  };
+
+  const { client } = useGql();
+  const queries = getReturnsQueries(getSdk(client), storeId);
+  return { ...queries, storeId, keys };
+};
```
CarlosNZ, line 17: This is good though. I think I'll use this branch as a base for my one so I can start with something that seems accessible.

client/packages/invoices/src/Returns/modals/NewReturnItems/ReturnQuantitiesTable.tsx
```diff
@@ -0,0 +1,54 @@
+import {
+  CellProps,
+  DataTable,
+  NumberInputCell,
+  SupplierReturnLine,
+  useColumns,
+} from 'packages/common/src';
+import React from 'react';
+
+const QuantityToReturnCell = (props: CellProps<SupplierReturnLine>) => (
```
CarlosNZ, line 10: Just wondering -- why create a new component that just wraps `NumberInputCell`?

If it's just so you can include it in the column definition below, then you can just inline in the definition:
```tsx
[
      'numberOfPacksToReturn',
      { Cell: (props) =>   <NumberInputCell {...props} isRequired max={props.rowData.availableNumberOfPacks}min={1}/>, width: 100, setter: updateLine },
],
```

Maybe that's messier? I dunno.

CarlosNZ (commented):
Looking good, @lache-melvin  👍

Couple of small comments/questions. See what you think. We should probably merge this without too much more scrutiny as we'll do a thorough review once we've got a few more bits and pieces in place.

client/packages/invoices/src/Returns/api/api.ts
lache-melvin: Yeah... it sounds like it's a headache for everyone - we initially copied this pattern in HSH/Notify but refactored it out because its a lot of complexity without seeming to add much value...

Just kept it here to be consistent with the rest of the app, but I'd definitely be up for a refactor! Even one R&D day 🤔

client/packages/invoices/src/Returns/modals/NewReturnItems/ReturnQuantitiesTable.tsx
lache-melvin: Haha no I think you're right - originally had some logic in there and felt I should pull out, but now it doesn't add much value. Fixed: [68d5070](https://github.com/msupply-foundation/open-msupply/pull/2928/commits/68d507001841556a7f95870eaf9e862e130a2447)

client/packages/common/src/intl/locales/en/common.json
lache-melvin: These are column labels, which are provided from the common package, so they do need to be here, but I agree its a bit confusing... I wonder if we could distinguish them with "common.{bla}" rather than just "label."?

PR #2952, opened 2024-02-15
2950 inbound return reason

### LINKED ISSUE #2950: Add reason step to inbound return modal

Need to add a reason and a comment for why this stock has been returned. Inverse of supplier return (#2929)

- A second step in the inbound return edit modal, where I can enter a reason and a comment
- Mock mutation for saving the return

lache-melvin:
Fixes #2950 

# 👩🏻‍💻 What does this PR do? 
 

Adds reason step to the inbound return from outbound shipment modal
Screenshot 2024-02-15 at 2 49 01 PM

- Go to outbound shipment details page
- Select some outbound shipment lines
- Select process return action
- Click Ok & next
- See reason step
- Edit the reason & comment
- Ok button is enabled once you select a reason
- Mutation is called and modal closes on OK


client/packages/invoices/src/Returns/api/hooks/document/useInsertInboundReturn.ts
```diff
@@ -0,0 +1,20 @@
+import { useQueryClient, useMutation } from '@openmsupply-client/common';
+import { useReturnsApi } from '../utils/useReturnsApi';
+
+export const useInsertInboundReturn = () => {
+  const queryClient = useQueryClient();
+  // const navigate = useNavigate();
+  const api = useReturnsApi();
+  return useMutation(api.insertInboundReturn, {
```
andreievg, line 8: Do you think there is a way use the return from mutation as the new cache for the return ? Or is it cleaner to just do another query by invalidating cache ?

This comment is not just for this mutation, but other mutations that return new 'state', and actually the return from API is kind of made to avoid extra API calls

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
```diff
@@ -1,19 +1,40 @@
...
+  InboundReturnInput,
+  InboundReturnLine,
+  InboundReturnLineInput,
+  RecordPatch,
+} from '@openmsupply-client/common';
 import { useReturns } from '../../api';
 
-export const useDraftInboundReturnLines = (stockLineIds: string[]) => {
-  const [draftLines, setDraftLines] = React.useState<InboundReturnLine[]>([]);
+export type DraftInboundReturnLine = InboundReturnLine & {
```
andreievg, line 11: InboundReturnLine would have reasonId and comment (api would), maybe best to add to MockApi now ?

client/packages/invoices/src/Returns/api/hooks/document/useInsertInboundReturn.ts
mark-prins: I've wondered about this too. We could use [setQueryData](https://tanstack.com/query/v4/docs/framework/react/guides/prefetching#manually-priming-a-query) to manually insert the mutation result into the cache. that would save a round trip to the api 🤔

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
```diff
@@ -25,5 +46,26 @@ export const useDraftInboundReturnLines = (stockLineIds: string[]) => {
...
+      }
+    );
+
+    const input: InboundReturnInput = {
+      id: FnUtils.generateUUID(),
+      customerId,
+      inboundReturnLines,
+    };
+
+    // TODO: error handling here
```
andreievg, line 65: I can do a quick session tomorrow on error types Structure/Standard, and how we use them in front end (in Stocktake Reason for example), although haven't done UI where errors are on multiple pages

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
```diff
@@ -25,5 +46,26 @@ export const useDraftInboundReturnLines = (stockLineIds: string[]) => {
     });
   };
 
-  return { lines: draftLines, update };
+  const saveInboundReturn = async () => {
+    const inboundReturnLines: InboundReturnLineInput[] = draftLines.map(
```
andreievg, line 50: I would usually do below, but this is completely stylistic preference, please don't change, just an option for future if you agree:

```TS
const inboundReturnLines: InboundReturnLineInput[] = draftLines.map(
      ({ id, reasonId, numberOfPacksReturned, stockLineId, comment }) => ({
        id,
        stockLineId,
        reasonId,
        comment,
        numberOfPacksReturned,
      })
    );
```

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -7,61 +7,103 @@ import {
...
+  const { currentTab, onChangeTab } = useTabs(Tabs.Quantity);
+
+  const returnsSteps = [
+    { tab: Tabs.Quantity, label: t('label.quantity'), description: '' },
+    { tab: Tabs.Reason, label: t('label.reason'), description: '' },
+  ];
+
+  const getActiveStep = () => {
+    const step = returnsSteps.find(step => step.tab === currentTab);
+    return step ? returnsSteps.indexOf(step) : 0;
```
andreievg, line 46: findIndex ?

client/packages/invoices/src/Returns/api/hooks/document/useInsertInboundReturn.ts
lache-melvin: Oo this would be rad!

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
lache-melvin: ah yes of course TY 🙏

andreievg (changes requested):
Can you please add comment and reasonId to API, just want to make sure it's there now and we don't need to do a union for the patch type.

Also would be good to [watch this query](https://github.com/msupply-foundation/open-msupply/blob/b4c99a3ecc4537ce4ec0b10109fdb48c1dffc116/client/packages/system/src/ReturnReason/api/hooks/document/useReturnReasons.ts#L4-L11), when we have more lines (does it actually only query once?), i remember having to [do this](https://github.com/msupply-foundation/open-msupply/blob/0a47d5388a5c45cb7029d070d57cfaf598022b81/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts#L16), because there was a query every time compoenent was reloaded, even though cache key didn't chage.

Also, do you think we need to be able to go to previous step ? (don't need to solve atm, but I think need to flag it somewhere

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
lache-melvin: ooo yes so much better love it

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
lache-melvin: That would be great please!!

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
lache-melvin: will do 🚀

lache-melvin:
> Also would be good to [watch this query](https://github.com/msupply-foundation/open-msupply/blob/b4c99a3ecc4537ce4ec0b10109fdb48c1dffc116/client/packages/system/src/ReturnReason/api/hooks/document/useReturnReasons.ts#L4-L11), when we have more lines (does it actually only query once?), i remember having to [do this](https://github.com/msupply-foundation/open-msupply/blob/0a47d5388a5c45cb7029d070d57cfaf598022b81/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts#L16), because there was a query every time compoenent was reloaded, even though cache key didn't chage. 

Oo great point, yes will make a note to check this out when we implement the API side!

> Also, do you think we need to be able to go to previous step ? (don't need to solve atm, but I think need to flag it somewhere

I wondered about this - I checked the other wizard modal we have, which doesn't allow a go back, and Anisha didn't have go back in her design, so I figured let's just try this for now. Personally I would probably like to be able to click the `1` in the stepper and go back to the first page though. Maybe a question for Richard at the end? Or happy to include now :)

lache-melvin:
Comment/reasonId added to mock API!

andreievg:
Thanks for changes !

> Maybe a question for Richard at the end?

👍

andreievg (approved):

PR #2997, opened 2024-02-16
2953 supplier return list view

### LINKED ISSUE #2953: Add Supplier/Outbound return list view

We need a way to get to the supplier returns we are creating!

- Add to app drawer under replenishment
- List view of all supplier returns
- Can add the buttons (add/print/etc) but they don't need to work yet :)
- Mock the API - should just be a new kind of invoice

lache-melvin:
Fixes #2953

# 👩🏻‍💻 What does this PR do? 
 

Screenshot 2024-02-16 at 2 57 54 PM

- Go to Replenishment > Outbound Returns
- See table view
- Selecting lines + delete calls the mock delete mutation
- Export to CSV works
- Create return button opens the supplier name modal, but doesn't go any further after that yet :)

Note: I think the filters work, but we'll know once the API actually returns data based on what we pass in 😁 

Some code is commented out for now, but left for reference: handling the create button, disabling verified rows...

Probably going to be some merge conflicts with this/Carl's branch/my others... but didn't want to keep using previous as a base and creating a dependency on what PRs can be merged in first... feel like I have a pretty good sense of where the overlap will be and should be manageable though :)

...Also super indecisive PR on the outbound/supplier return naming front... started on the frontend with outbound return, but then supplier return was already in the backend so now there's a bit of both 🙈  I'll update this PR to whichever we choose when we do!


client/packages/invoices/src/InvoiceService.tsx
```diff
@@ -54,6 +59,7 @@ const InvoiceService: FC = () => {
       <Route path={prescriptionsRoute} element={<PrescriptionListView />} />
       <Route path={prescriptionRoute} element={<PrescriptionDetailView />} />
 
+      <Route path={outboundReturnsRoute} element={<OutboundReturnListView />} />
       <Route path={outboundReturnRoute} element={<ReturnsDetailView />} />
```
CarlosNZ, line 63: Should probably keep these names consistent, so make the other one same as yours?

Also -- the imports at the top are separate, can you make it a single import statement?

client/packages/invoices/src/Returns/api/hooks/document/useOutboundDeleteRows.ts
```diff
@@ -0,0 +1,53 @@
...
+      .catch(err => {
+        throw err;
+      });
+  };
+
+  const confirmAndDelete = useDeleteConfirmation({
+    selectedRows,
+    deleteAction,
+    canDelete: selectedRows.every(
+      ({ status }) => status === InvoiceNodeStatus.New // is this accurate?
```
CarlosNZ, line 40: The outbound shipment uses  `canDeleteInvoice` function -- can we use the same one?

Looks like the criteria is  `invoice.status === InvoiceNodeStatus.New ||  invoice.status === InvoiceNodeStatus.Allocated`, but I'm not sure if that's exactly what we'll have. We haven't really discussed statuses yet. Fine to leave a "TO DO" here I guess?

CarlosNZ (approved):
LGTM, couple of tiny nits.

PR #3000, opened 2024-02-16
2760 om supply should not make transfers for old invoices/requisitions

### LINKED ISSUE #2760: omSupply should not make transfers for old invoices/requisitions

When doing a deployment it is common to:

- Convert regular names (facilities) into stores which may have had outbound shipments made to them over several years
- Phase roll outs, making stores names that have already started receiving orders

When that new store/site is deployed, omSupply will currently make transfers for all past shipments. For old heavily used names that could be hundreds of transfers going back years. This is usually annoying - first step in deployment is doing a stocktake for the store, so receiving stock via inbound shipments after that will over report stock.

But, in deployment of sites, any inbound shipments that they haven't physically received yet _are_ useful after that initial stocktake as it is indeed new physical stock to enter into stock levels.

Another case this happens is that supplying stores may have unfinalised outbound shipments for the new store's name. Once finalised a transfer will be made. Again, this may be for transfers that were physically received long ago, or recent enough that they haven't received and could use the inbound shipment.

[mSupply desktop has gone through this](https://github.com/msupply-foundation/msupply/issues/8201).  

image

Double checked - Big oof:

https://github.com/msupply-foundation/open-msupply/blob/e0fa90930f13f3b21a2df5c71dd2247b46493ffe/server/repository/src/db_diesel/store_row.rs#L8-L17

Probably 1: Add `created_datetime` (and translate in sync etc.). ~~For existing stores I think just set to start of epoch? Otherwise I've always wanted an API in central server where the client can request certain tables/records resync from central server 😃 but that's a bit much maybe.~~ Maybe it's in the sync_buffer still and can simply be re-integrated to get the `created_datetime`?

After a brief discussion with @andreievg there are some further complexities in omSupply, as it isn't so simple to delete an inbound shipment. It's possible that:

1. Inbound shipment generated, linked to outbound shipment (`invoice.linked_invoice_id`)
2. This transfer record is synced back to the supplier site
3. Supplier site updates Outbound shipment linking it with FK to Inbound shipment (`invoice.linked_invoice_id`)
4. This syncs back to customer site
5. customer site now cannot delete the inbound shipment as it'd break CI -> SI FK.
(There is also a window before step 4 where the customer COULD delete it and create FK constraint errors in the sync buffer for both sites)

So... essentially set all received quantities to 0, finalise the invoice with some log message (open to suggestions!). 

### Additional context 💌

- https://github.com/msupply-foundation/msupply/issues/7695

roxy-dao:
Closes #2760

# 👩🏻‍💻 What does this PR do? 
Branch was getting a bit out of hand so have split this issue into two parts, one for the invoice and one for requisition... Sorry!

- Add `created_date` to store and `ZERO_QUANTITIES_FOR_INVOICE_CREATED_BEFORE_STORE` enum for logging and sync. 
- Don't transfer shipments that were created 30 days before the store was created.  
- Service/endpoint to batch set the quantity of lines (and other fields) to 0 if the invoice was created before store (will work for any invoice that was created before the store). This sets all fields except for the `name` and `item` related fields to None.
- Add hooks to batch set quantities (and other related fields in the line) to 0 (or None)
- Add `0 line quantity` action to Inbound Toolbar
- Translate log enum (: 

Screenshot 2024-02-16 at 15 19 34
Screenshot 2024-02-16 at 15 15 59

1. Backend
- [ ] Run tests
2. Frontend 
- [ ] Set PC time to before the store was created
- [ ] Create some Inbound Shipments with lines to test
- [ ] Go to the Detail View
- [ ] Try to 0 lines.

I would have liked the selection and endpoints to be done in invoice since it's easier to select all the lines and just 0 them but as requested have done in the detail view... There's a frontend check that forces the user to select all lines (just not a backend one) 👀 ... Feels dodgy that they don't have to 0 all lines in an invoice. 

Should you be able to zero out invoices regardless of whether it was created before store or not? 

Also might have been excessive in clearing the invoice line... 😅 

## Documentation
- Add new `Set quantity to 0` action to docs


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -354,6 +355,7 @@
   "log.changed-to": "to",
   "log.invoice-created": "Shipment created",
   "log.invoice-deleted": "Shipment deleted",
+  "log.zero-quantities-for-invoice-created-before-store": "The quantity of all lines has been set to 0 and invoice has been finalised as it was created before the store.",
```
roxy-dao, line 358: Not too sure about this message... open for suggestions. Didn't wanna make it too long 👀

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
```diff
@@ -115,6 +128,14 @@ export const Toolbar: FC = () => {
             <DropdownMenuItem IconComponent={DeleteIcon} onClick={onDelete}>
               {t('button.delete-lines')}
             </DropdownMenuItem>
+            {isInvoiceCreatedBeforeStore && (
+              <DropdownMenuItem
+                IconComponent={RewindIcon}
```
roxy-dao, line 133: Wasn't too sure about which icon to use... couldn't find the broom icon on feature and Mark suggested rewind icon so went with that 🤷🏻

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const notAllLinesSelected = selectedRows.length !== lines?.length;
+
+  const onZeroQuantities = async () => {
+    await mutateAsync(selectedRows).catch(err => {
+      throw err;
+    });
+  };
+
+  const confirmAndZeroLines = useDeleteConfirmation({
```
roxy-dao, line 36: Reusing the useDeleteConfirmation modal since it works for this... but should probably rename. Can't think of a name for it right now... will get back to it

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
```diff
@@ -0,0 +1,52 @@
...
+    selectedRows,
+    deleteAction: onZeroQuantities,
+    canDelete: !isDisabled && !notAllLinesSelected,
```
roxy-dao, line 39: Not sure if user has to select all lines to 0 or not... but seemed to make sense to me since the invoice gets finalised 👀

server/repository/src/db_diesel/invoice_line.rs
```diff
@@ -91,19 +91,7 @@ pub struct InvoiceLineFilter {
...
-            r#type: None,
-            item_id: None,
-            location_id: None,
-            requisition_id: None,
-            number_of_packs: None,
-            invoice_type: None,
-            invoice_status: None,
-            stock_line_id: None,
-        }
+        InvoiceLineFilter::default()
```
roxy-dao, line 94: Sidetracked...

server/repository/src/migrations/types.rs
```diff
@@ -1,16 +1,16 @@
 #[cfg(feature = "postgres")]
-pub(crate) const DATE: &'static str = "DATE";
+pub(crate) const DATE: &str = "DATE";
```
roxy-dao, line 2: Another sidetrack.... str is static already..

server/service/src/invoice_line/inbound_shipment_line/zero_line_quantity/generate.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    GenerateResult {
+        invoice_row,
+        new_line,
+        stock_line_id: existing_invoice_line.stock_line_id,
+    }
+}
+
+fn generate_line(line: InvoiceLineRow) -> InvoiceLineRow {
+    InvoiceLineRow {
```
roxy-dao, line 36: Overkill None'ing?

server/service/src/processors/transfer/shipment/create_inbound_shipment.rs
```diff
@@ -66,6 +67,19 @@ impl ShipmentTransferProcessor for CreateInboundShipmentProcessor {
         if linked_shipment.is_some() {
             return Ok(None);
         }
+        // 5.
+        let store = StoreRowRepository::new(connection)
```
roxy-dao, line 71: Actual place where transfers aren't created if invoice created datetime is less than store created date

server/service/src/sync/translations/store.rs
```diff
@@ -65,7 +69,7 @@ impl SyncTranslation for StoreTranslation {
         }
 
         // ignore stores without name
-        if data.name_id == "" {
+        if data.name_id.is_empty() {
```
roxy-dao, line 72: Another sidetrack....

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -37,6 +37,7 @@
   "button.sync-now": "Sync now!",
   "button.try-again": "Try again",
   "button.view": "View",
+  "button.zero-line-quantity": "0 line quantity",
```
clemens-msupply, line 40: Doesn't say much what it does or read like a command... Maybe something like "Set quantities to 0"?

server/service/src/sync/translations/store.rs
```diff
@@ -26,6 +30,9 @@ pub struct LegacyStoreRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     logo: Option<String>,
     store_mode: LegacyStoreMode,
+    #[serde(deserialize_with = "zero_date_as_option")]
+    #[serde(serialize_with = "date_option_to_isostring")]
+    pub created_date: Option<NaiveDate>,
```
clemens-msupply, line 35: How does this work? I don't have a created_date column in mSupply is it stored somewhere else in 4D?

server/service/src/processors/transfer/shipment/create_inbound_shipment.rs
```diff
@@ -66,6 +67,19 @@ impl ShipmentTransferProcessor for CreateInboundShipmentProcessor {
...
+            .find_one_by_id(&record_for_processing.other_party_store_id)?
+            .ok_or(RepositoryError::NotFound)?;
+
+        let store_created_datetime = store
+            .created_date
```
clemens-msupply, line 76: This might be nicer, e.g. you don't need to wrap `invoice_created_datetime` in a Some():

```rust
if let Some(created_date) = store.created_date {
   ...
}
```

clemens-msupply (commented):
Before continuing with the review, what is the purpose of setting the quanties of the invoice line to 0? 

And when does this happen? e.g. is this for inbound shipments that have been transferred by mSupply?

server/service/src/sync/translations/store.rs
roxy-dao: Screenshot 2024-03-08 at 07 45 08
It is stored in the store table and indicates when the store was created

roxy-dao:
> Before continuing with the review, what is the purpose of setting the quantites of the invoice line to 0?
> 
> And when does this happen? e.g. is this for inbound shipments that have been transferred by mSupply?

To prevent hundreds of records from being made in omSupply when desktop sites migrate to omSupply. When deploying, the store first does a stocktake, so if they have stock in an inbound shipment that has been transferred from a really old outbound shipment, the stocktake will show that they are overstocked.

server/service/src/sync/translations/store.rs
clemens-msupply: thanks seeing it now. This what I was seeing before in the 4D UI: 
image
But if you scroll down there is another item! :)

So in 4D its really a nullable value. Wonder if we should make it non-nullable in omSupply, i.e. map NULL to the date at `0`.

We should also already use a `create_datetime` in omSupply. Don't see a reason why we should not also record the time value in the future (when creating a store in omSupply central server) and then we already have the correct type in omSupply. @Chris-Petty I guess it only stores a date because there is not datetime DB type in 4D?

roxy-dao: Normally with those there's a created_time field in the table as well on 4d... Is time really important when creating the store? 🤔  Can add it in ig, but usually on 4d there's a created time field too 🤔

clemens-msupply: time would be useful for debugging but I guess there is also a log entry for it?

server/graphql/invoice_line/src/mutations/inbound_shipment_line/line/zero_line_quantity.rs
```diff
@@ -0,0 +1,139 @@
...
+
+#[derive(Interface)]
+#[graphql(name = "ZeroInboundShipmentLineQuantityErrorInterface")]
+#[graphql(field(name = "description", type = "&str"))]
+pub enum ZeroInboundShipmentLineQuantityErrorInterface {
+    RecordNotFound(RecordNotFound),
+    ForeignKeyError(ForeignKeyError),
+    CannotEditInvoice(CannotEditInvoice),
+    BatchIsReserved(BatchIsReserved),
+    InvoiceWasCreatedAfterStore(InvoiceWasCreatedAfterStore),
```
clemens-msupply: Do we need these? e.g.:

- RecordNotFound(RecordNotFound) -> BadUserInput/ Internal error?
- ForeignKeyError(ForeignKeyError), -> internal error
- CannotEditInvoice(CannotEditInvoice), -> BadUserInput? (what is the reason here?)
- BatchIsReserved(BatchIsReserved), -> BadUserInput
- InvoiceWasCreatedAfterStore(InvoiceWasCreatedAfterStore), -> BadUserInput

client/packages/invoices/src/Prescriptions/api/operations.graphql
```diff
@@ -50,6 +50,15 @@ fragment PrescriptionRow on InvoiceNode {
...
   }
+
+  currency {
+    id
+    code
+    rate
+    isHomeCurrency
+  }
+
+  currencyRate
```
clemens-msupply, line 61: Is this an unrelated change?

client/packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -28,6 +28,7 @@
   "info.automatic-shipment-no-edit": "You are unable to edit details until the status is confirmed as Delivered.",
   "info.manual-shipment": "This shipment was created manually. The delivery status will not be automatically updated.",
   "label.add-batch": "Add batch",
+  "label.cant-zero-quantity-disabled": "All lines must be selected and status must be New",
```
clemens-msupply: Is just "label.cant-zero-quantity"?

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
clemens-msupply: Where does the invoice gets finalized? does the user has to do it manually, or am I missing it in the code?

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  useTableStore,
+  useTranslation,
+  useDeleteConfirmation,
+} from '@openmsupply-client/common';
+import { useIsInboundDisabled } from '../utils/useIsInboundDisabled';
+import { useInboundRows } from './useInboundRows';
+import { useZeroInboundLineQuantity } from './useZeroInboundLineQuantity';
+
+export const useZeroInboundLinesQuantity = (): (() => void) => {
```
clemens-msupply, line 10: This hook feels out of place here because it also contains the UI part and the very specific part of reading the data from the table. You should either create a hook that is UI independent, e.g. where you pass in the table data as a parameter, or move this file closer to where it is used (or put the hook into Toolbar.tsx?).

Might be wrong but only saw hooks that directly map to an api in this hooks directory... For example, found it confusing to have a useZeroInboundLineQuantity and a useZeroInboundLinesQuantity hook where the former actually zeros multiple lines and the second does UI stuff..

clemens-msupply (commented):
Wonder if we need a special endpoint to set the quantity to zero? Would it work to just use the existing mutations? This would make this PR quite a bit smaller...

server/repository/src/db_diesel/activity_log_row.rs
```diff
@@ -58,6 +58,7 @@ pub enum ActivityLogType {
     PrescriptionStatusPicked,
     PrescriptionStatusVerified,
     SensorLocationChanged,
+    ZeroQuantitiesForInvoiceCreatedBeforeStore,
```
clemens-msupply, line 61: The "CreatedBeforeStore" part should be removed here and in the migration... (if we decide that we actually need a new endpoint, see other comment...)

server/service/src/invoice_line/inbound_shipment_line/zero_line_quantity/generate.rs
roxy-dao: I mean.. if you haven't received it or it is not valid anymore, is it wrong? Since a transfer would send batch info and other things as well...

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
roxy-dao: Oh this was changed sorry forgot to delete

roxy-dao:
> Wonder if we need a special endpoint to set the quantity to zero? Would it work to just use the existing mutations? This would make this PR quite a bit smaller...

Yup, this has changed a lot from what I understood originally, so can just do an update and send 0 for all the inputs (:

client/packages/invoices/src/Prescriptions/api/operations.graphql
roxy-dao: Ah yes sorry. Mark also fixed this in one of his PRs. The currency changes broke generation

server/service/src/invoice_line/inbound_shipment_line/zero_line_quantity/generate.rs
clemens-msupply: But the information might still be valuable, i.e. what has been sent etc. And then you just said you have received 0 of these. I guess that's why mSupply has a "send" and a "received" field...

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
roxy-dao: Was just keeping things consistent since that's how it's done in the rest of the project~

roxy-dao:
Screenshot 2024-03-11 at 08 36 35
I've changed the logging message...

server/graphql/general/src/queries/tests/master_lists.rs
```diff
@@ -67,15 +67,6 @@ mod graphql {
...
                   description
-                  lines {
-                      nodes {
-                          id
-                          item {
-                              id
-                          }
-                      }
-                      totalCount
-                  }
```
clemens-msupply, line 78: What was the reason to remove this?

roxy-dao: That was a change in the master list line pagination PR, but missed it since action tests didn't fail 👀

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
clemens-msupply: Just `useConfirmationDialog`? You would have to rename canDelete and I wouldn't fall back to the delete messages but make the message mandatory input parameters.

Maybe have another wrapper `useDeleteConfirmation` that calls `useConfirmationDialog` with the correct messages?

roxy-dao: Yeah, I was thinking of doing that but in a carry-over issue... touches quite a few place 🤔  Is that alright?

clemens-msupply (approved):
Looks and works good. Before merging please check with @Chris-Petty about the store created_date/created_datetime question.

Note, do we actually need a log message for this action? what makes it different to a normal action that changes the number of packs?

client/packages/invoices/src/InboundShipment/api/hooks/line/useZeroInboundLinesQuantity.ts
clemens-msupply: If you keep a compatible version of `useDeleteConfirmation` that just wraps the new more general `useConfirmationDialog` hook you can keep the changes to a minimum.

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
Chris-Petty: image

Can't find any feather one but there is broom sort of... I think I like rewind more though.

server/service/src/sync/translations/store.rs
Chris-Petty: Yea I think you both have it - 4D doesn't have date/time, and given time doesn't matter so much we didn't bother adding a field for it. So I think `Option(NaiveDate)` seems as good as it gets - would love for it to not be Option though...

It's not really `Null` in mSupply, I'm pretty sure it's the "blank" date value , literally `0000-00-00`... make of that what you will lol. Might as well be Null. 
image
(The last checkbox will do this, it's on for the majority of fields in mSupply except `om_` fields I think.)

That said, when this was added there was a migration code that tried to guess the created date based on logs or invoices:
https://github.com/msupply-foundation/msupply/blob/157aad17c99775ac720276d3e03b7b56b79ebb3d/Project/Sources/Methods/startup_update_version.4dm#L4397-L4418
Admittedly it might still fail and leave the blank value if there is no transacts nor ANY log (could be a datafile from 15 years before logging and the store was was never used??)... maybe failing those two it'd been better to use the earliest `store.created_date` from other stores... could still add a new migration and guarantee it 😆.

@clemens-msupply haha yea the sizing thing hiding fields is super dumb in 4D 😆. You can resize everything by pressing `ctrl+a` to select all tables in the schema viewer, then the cogs at the top has an auto resize. If only it reran every time a field was added... 
image

Chris-Petty: That said I don't mind using `NaiveDatetime` if it makes the logic checking `transact.created_datetime < (store.created_date(time)- 1month)` easier

server/service/src/processors/transfer/shipment/create_inbound_shipment.rs
```diff
@@ -66,6 +67,18 @@ impl ShipmentTransferProcessor for CreateInboundShipmentProcessor {
...
             return Ok(None);
         }
+        // 5.
+        let store = StoreRowRepository::new(connection)
+            .find_one_by_id(&record_for_processing.other_party_store_id)?
+            .ok_or(RepositoryError::NotFound)?;
+
+        if let Some(created_date) = store.created_date {
+            let store_created_datetime = NaiveDateTime::new(created_date, NaiveTime::from_hms_opt(0, 0, 0).unwrap_or_default());
+            let invoice_created_datetime = outbound_shipment.invoice_row.created_datetime - Duration::days(30);
```
Chris-Petty, line 77: When I saw this earlier I wasn't sure but now writing a late comment - shouldn't you subtract the duration from the `store_created_datetime`? 

```
store_created_datetime = 2024-03-12
invoice_created_datetime = 2024-03-20
invoice_created_datetime = invoice_created_datetime - 30days // ~2024-02-20

invoice_created_date < store_created_datetime == True // We're not making transfers for invoices after the store was created? The goal was only make invoices if `store_created_date - invoice_created_date > -30days`, not `store_created_date - invoice_created_date > +30days`

```

It's late and I might be cooked though but I didn't want to get this browser tab lost again lol

roxy-dao: You are right!

server/service/src/sync/translations/store.rs
clemens-msupply: I think we should at least use `pub created_date: NaiveDate` in open mSupply then right? and map "Null" or "0000-00-00" to the earliest available date? But this would also mean to reintegrate all stores to get the correct created_date...

roxy-dao: Can you create a carry over issue then if that's what we should do

Issue #1912, opened 2023-06-14
Need a good way to alert users about why they can't use certain functionality (eg. bc of permissions)


nisha-dangol:
## What went wrong? 😲
In latest `v1.1.15` apk, the `Confirm sent` button is disabled and thus the internal order can't be send anymore! 
Let know if I'm missing any new updates/changes in the flow! (since it was working well in `v1.1.14`) 🤷

## Expected behaviour 🤔
The `sent` button should not be disabled

## How to Reproduce 🔨
Steps to reproduce the behaviour:
1. Get latest v1.1.15 apk
2. Create a new internal order and add some lines
3. See error: the `Confirm sent` button is disabled
![Screenshot_20230614_174101_openmsupply-client](https://github.com/openmsupply/open-msupply/assets/59193490/ecdc627f-0b36-4ba7-a0f4-eacd3b447c8f)

## Your environment 🌱

 - Version: v1.1.15 apk
 - Platform:
   - android (tablet)
   - server (windows)
 
 ## Solution:
 
"Anisha says: Button enabled and toast"

DhanyaHerath:
@roxy-dao @mark-prins do we have a preferred pattern for things that are prevented because of permission?
Two possible options are: disable the button but with a tooltip explaining why, or keep enabled and have an alert explaining why. 

I think adding at least a tooltip could be helpful here.

mark-prins:
Already has a tooltip 😉 

I don't like the alert on an enabled button, it means I have to click to dismiss the alert. The disabled button is a good visual cue that it isn't allowed. That's personal preference though I guess.

roxy-dao:
🤔 I noticed people don't actually use the tooltips at all....

mark-prins:
So..? Do you think an alert would be more effective?

What about a toast? that'd be a compromise - the button is still enabled, an alert shows, and I don't have to click to get rid of the alert

The tooltip isn't very informative, now that I check it 😄 
![Screenshot 2023-06-15 at 3 11 42 PM](https://github.com/openmsupply/open-msupply/assets/9192912/aaa19394-8b23-4427-9046-6cc9c079228a)

DhanyaHerath:
Hahha yeah I was trying to find definitive articles telling me what was the best way. I started out neutral between "disabled and tooltip" and "enabled and alert" and then I started leaning towards enabled: 
https://uxpsychology.substack.com/p/hidden-vs-disabled-states
[...]

PR #3001, opened 2024-02-16
2751 ability to export sqlite database on android

### LINKED ISSUE #2751: Ability to export sqlite database on Android

For issues raised in production, it would be much easier (and less time consuming) for problem replication, if we were able to run the database snapshot locally in our dev environment

A way to export sqlite database for Open mSupply Mobile.

**Encryption ?**

Yes if possible, can specify password when zipping ?

**How to access this export on remote sites ?**

Maybe good to consult with support team @npendergrast @kahn-msupply. Would it be relatively simple to download data from any particular accessible folder with MDM ?

In production cannot access app data folder, so can't think of alternatives

### Additional context 💌

There is identical issue in: https://github.com/msupply-foundation/msupply-cold-chain/issues/285, I think same technique applies for both

jmbrunskill:
Fixes #2751

# 👩🏻‍💻 What does this PR do? 
Allows users with Server Administration Access to download the SQLite database, on Android or via the web.
On Android the file is zipped, making it smaller to share. 74MB zipped down to 14MB so seems like a decent saving there.

Download button shows as disabled if the server is running postgres.

image

Would be good to have testing on Android and on Desktop.
I think there will be a bug on android it running as a client and the server is in SQLite mode. I don't think that's too big an issue though.

Download performance (not android) is very slow. Maybe an actix bug or needs buffering or something? Could be a useful follow up issue unless someone has an idea how to improve it.

I'm using the refresh token cookie to validate the download request via the API.
I feel like it might be nicer to have the actual access token in a cookie to validate things like this.
We could also separate out the authentication logic into a middleware which would reduce the code duplication (e.g. cookie extraction logic is now in two places) - See: https://github.com/msupply-foundation/open-msupply/issues/3002

## 📃 Documentation

Documentation on how to export the database on android would be useful.


client/packages/android/app/src/main/java/org/openmsupply/client/NativeApi.java
```diff
@@ -349,20 +350,21 @@ public void onResolveFailed(NsdServiceInfo serviceInfo, int errorCode) {
...
         }
         try {
-                for (Enumeration<NetworkInterface> en = NetworkInterface.getNetworkInterfaces(); en.hasMoreElements(); ) {
-                    NetworkInterface ni = en.nextElement();
-                    for (Enumeration<InetAddress> enumIpAddr = ni.getInetAddresses(); enumIpAddr.hasMoreElements(); ) {
-                        InetAddress inetAddress = enumIpAddr.nextElement();
-                        if (!inetAddress.isLoopbackAddress() && !inetAddress.isLinkLocalAddress() && inetAddress.isSiteLocalAddress()) {
-                            return inetAddress.getHostAddress();
-                        }
+            for (Enumeration<NetworkInterface> en = NetworkInterface.getNetworkInterfaces(); en.hasMoreElements();) {
```
jmbrunskill, line 358: This is just reformatting from my editor not sure why...

server/server/src/cold_chain/temperature_log.rs
```diff
@@ -7,7 +7,6 @@ use anyhow::Context;
 use chrono::NaiveDateTime;
 use log::error;
 use mime_guess::mime;
-use openssl::error;
```
jmbrunskill, line 10: This unused import was annoying me, but is side track...

server/server/src/support/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        &format!("{}{}", URL_PATH, "/database"),
+        web::get().to(get_database),
+    );
+    cfg.route(
+        &format!("{}{}", URL_PATH, "/vacuum"),
+        web::post().to(vacuum_database),
+    );
+}
+
+fn validate_request(
```
mark-prins, line 32: any idea why I get this?

image

it looks like it is used..?

server/server/src/cold_chain/temperature_log.rs
mark-prins: thanks - I was going to take that out too!

server/server/src/support/database.rs
```diff
@@ -0,0 +1,71 @@
...
+pub async fn vacuum_database(service_provider: Data<ServiceProvider>) -> HttpResponse {
+    // Vacuum the database first
+    let result = service_provider.connection_manager.execute("VACUUM");
+    match result {
+        Ok(_) => HttpResponse::Ok().body("Vacuumed database successfully"),
+        Err(e) => {
+            HttpResponse::InternalServerError().body(format!("Error vacuuming database: {:#?}", e))
+        }
+    }
+}
```
mark-prins, line 71: For the handlers, I like the config macro approach as I think it reads easier. Perhaps that's partly because I'm usually running postgres, which means that the code I'm executing is always the faded out version.
It is less code overall and means that there's less likelihood of missing conditions out.

```suggestion
pub async fn vacuum_database(service_provider: Data<ServiceProvider>) -> HttpResponse {
    if cfg!(feature = "postgres") {
        return HttpResponse::InternalServerError().body("Postgres Databases vacuum not supported");
    }

    // Vacuum the database first
    let result = service_provider.connection_manager.execute("VACUUM");
    match result {
        Ok(_) => HttpResponse::Ok().body("Vacuumed database successfully"),
        Err(e) => {
            HttpResponse::InternalServerError().body(format!("Error vacuuming database: {:#?}", e))
        }
    }
}
```

client/packages/host/src/Admin/ServerSettings.tsx
```diff
@@ -79,6 +88,34 @@ export const ServerSettings = () => {
           </>
         }
       />
+
+      <Setting
+        title={t('label.download-database')}
+        component={
+          <>
+            <LoadingButton
+              disabled={databaseSettings?.databaseType !== DatabaseType.SqLite}
```
mark-prins, line 97: If the button is going to show, it would be helpful to know why the button is disabled for me.
Either helper text below the `Download Database` heading, or a tooltip, or the preferred option of the PM/design team (according to #1912) - the button is enabled but alerts when you click it. 

That latter option is something I think none of the developers want, btw. Some help text?

server/server/src/support/mod.rs
mark-prins: oh. let me answer my own question.

```
pub async fn get_database(
```

if you change to using an if statement instead, then that will fix the problem.
I'm running postgres and get the warning, because the pg version of `get_database` does not use the validate fn.

server/server/src/support/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    request: HttpRequest,
+    service_provider: &ServiceProvider,
+    auth_data: &AuthData,
+) -> Result<ValidatedUser, AuthError> {
+    let service_context = service_provider
+        .basic_context()
+        .map_err(|err| AuthError::Denied(AuthDeniedKind::NotAuthenticated(err.to_string())))?;
+
+    // We use the refresh token to get the user's access token here, as the actual access token isn't easily to pass as a header in a download link
```
mark-prins, line 41:
```suggestion
    // We use the refresh token to get the user's access token here, as the actual access token isn't easily passed as a header in a download link
```

mark-prins (approved):
👍 Nice!
Nothing major but a couple of small things that I'd like changed, as noted.
Have tested and it's working nicely! one weirdness is that when you are running in client mode against a SQLite server, the downloaded database is the locally running one, i.e. it's empty.

I think that could be misleading - perhaps something to add to the documentation? It might still be useful to have the file downloadable? If it isn't, we could just disable in that scenario.

client/packages/host/src/Admin/ServerSettings.tsx
jmbrunskill: Definitely a tricky area with differing opinions on how to handle it. I did try to put in a title but doesn't show for disabled buttons. Another option is to take out the check entirely, then you get a backend error if you are on postgres. I nearly did that originally. I personally like tooltips but they're not so great on android...

Anyway, I've found I can get tooltips working by adding span element in, I think that will do for now?

mark-prins: yep - that's good for me, thanks!

PR #3004, opened 2024-02-16
2998 inbound returns list view

### LINKED ISSUE #2998: Add customer returns list view

Inverse of #2953

Need a way to get to the customer/inbound returns that we've created:

-  Add to app drawer under distribution
- List view of all customer returns
- Can add the buttons (add/export/etc) but they don't need to work yet :)
- Mock the API - should just be a new kind of invoice

lache-melvin:
Fixes #2998

# 👩🏻‍💻 What does this PR do? 
 

Adds the customer returns list view

Screenshot 2024-02-16 at 4 41 54 PM

- Go to Distribution > Inbound Returns
- See table view
- Selecting lines + delete calls the mock delete mutation
- Export to CSV works
- Create return button opens the customer search modal, but doesn't go any further after that yet :)

Note: Same as supplier returns list PR, I think the filters work, but we'll know once the API actually returns data based on what we pass in 😁


server/graphql/invoice/src/lib.rs
```diff
@@ -7,7 +7,9 @@ pub mod invoice_queries;
 use self::invoice_queries::*;
 
 pub mod mutations;
-use self::mutations::{inbound_shipment, outbound_shipment, prescription, supplier_return};
+use self::mutations::{
+    customer_return, inbound_shipment, outbound_shipment, prescription, supplier_return,
```
CarlosNZ, line 11: Is this an example of the back-end renaming you mentioned you would do?

server/graphql/invoice/src/mutations/customer_return/delete.rs
```diff
@@ -0,0 +1,40 @@
+use async_graphql::*;
```
CarlosNZ, line 1: Presumably planning to rename to inbound_return?

CarlosNZ (approved):
Cool. I seemed to be getting some of the changes from your last PR mixed in here as well,  but I think you re-updated it from the base while I was looking at it, so all good.

One or two tiny comments

client/packages/invoices/src/Returns/InboundListView/ListView.tsx
```diff
@@ -0,0 +1,99 @@
...
+  NothingHere,
+  useToggle,
+  useUrlQueryParams,
+} from '@openmsupply-client/common';
+import { getStatusTranslator } from '../../utils';
+import { Toolbar } from './Toolbar';
+import { AppBarButtons } from './AppBarButtons';
+import { InboundReturnRowFragment, useReturns } from '../api';
+
+const InboundReturnListViewComponent: FC = () => {
```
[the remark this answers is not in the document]
lache-melvin: Was following the convention from inbound/outbound shipment, and I think other areas of the codebase - this component is wrapped in a `TableProvider` at the bottom of the file and then exported as `InboundReturnListView`, so I think the `Component` suffix just distinguishes it as the component to be wrapped before export.

server/graphql/invoice/src/lib.rs
lache-melvin: Yes has been renamed in a later PR 🙏

client/packages/invoices/src/Returns/InboundListView/ListView.tsx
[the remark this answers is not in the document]
CarlosNZ: Oh I see. I compared it against Inbound Shipment -> ListView which names it somewhat differently, but I see it's not consistent throughout. As you were...

PR #3015, opened 2024-02-18
3008 add outbound return from list

### LINKED ISSUE #3008: Add supplier return from the list view

I need to be able to create manual supplier/outbound returns.

From the Outbound Returns view, I want to click a button to create a new Outbound Return.

This should work very similarly to creating an inbound shipment.

lache-melvin:
Fixes #3008 

_Using #2997 as a base for now..._

# 👩🏻‍💻 What does this PR do? 
 

All this was already in the list view, but now connects the insert mutation on supplier selection and redirects to the resulting details view:
Screenshot 2024-02-19 at 12 28 31 PM

- On the list view, click `New Return`
- Supplier selection modal appears
- When you select a supplier, the mutation is called, and you are redirected to the detail page


lache-melvin:
Wow accidentally merged this into another branch and apparently reverting is hard 😅  This PR is now #3024

PR #3024, opened 2024-02-19
3008 create outbound return from list view

### LINKED ISSUE #3008: Add supplier return from the list view

lache-melvin:
Fixes #3008 

# 👩🏻‍💻 What does this PR do? 
 

All this was already in the list view, but now connects the insert mutation on supplier selection and redirects to the resulting details view:
Screenshot 2024-02-19 at 12 28 31 PM

- On the outbound return list view, click `New Return`
- Supplier selection modal appears
- When you select a supplier, the mutation is called, and you are redirected to the detail page

### DIFF
--- a/client/packages/invoices/src/Returns/api/hooks/document/useInsertSupplierReturn.ts
+++ b/client/packages/invoices/src/Returns/api/hooks/document/useInsertSupplierReturn.ts
@@ -1,19 +1,23 @@
-import { useQueryClient, useMutation } from '@openmsupply-client/common';
+import {
+  useQueryClient,
+  useMutation,
+  useNavigate,
+  RouteBuilder,
+} from '@openmsupply-client/common';
 import { useReturnsApi } from '../utils/useReturnsApi';
+import { AppRoute } from 'packages/config/src';
 
 export const useInsertSupplierReturn = () => {
   const queryClient = useQueryClient();
-  // const navigate = useNavigate();
+  const navigate = useNavigate();
   const api = useReturnsApi();
   return useMutation(api.insertSupplierReturn, {
-    onSuccess: () => {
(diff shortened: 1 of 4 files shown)


client/packages/invoices/src/Returns/api/hooks/document/useInsertSupplierReturn.ts
```diff
@@ -1,19 +1,23 @@
...
-      //   .addPart(AppRoute.InboundShipment)
-      //   .addPart(String(invoiceNumber))
-      //   .build();
-      // navigate(route, { replace: true });
+    onSuccess: invoiceNumber => {
+      const route = RouteBuilder.create(AppRoute.Replenishment)
+        .addPart(AppRoute.OutboundReturn)
+        .addPart(String(invoiceNumber))
+        .build();
+      navigate(route, { replace: true });
```
andreievg, line 20: This kind of weird to me, that the api hook navigates, I guess this was just following existing pattern ?

andreievg (approved):
Looks good, thank you

client/packages/invoices/src/Returns/api/hooks/document/useInsertSupplierReturn.ts
lache-melvin: Yeah this is what we've had elsewhere - happy to try changing it if we thing it's hiding/doing too much? But because we can call this insert both from the outbound shipment detail view and the returns list view, I found myself not minding the redirect logic only needing to be in one place...

PR #3026, opened 2024-02-20
2857 suggest next encounter date

### LINKED ISSUE #2857: Suggest next encounter date using date information from previous encounter
An encounter already can have a config in the document registry. Currently, this config is used for program events. This config can be extended with information about scheduling the next encounter. For example, (details to be worked out):

```ts
type Config = {
  /** Config for extracting the next encounter date from existing data.
    * If set the UI can suggest to schedule the next encounter on the date the data field refers to.
    */
  nextEncounter?: {
    /** The type of the config. Event means getting the next date from a program event */
    type: "Event"
    /**
     * For example, `arvMedicationEndOfSupply`
     */
    event: string,
    /**
      * Optional description which can be used in the UI to indicate where the suggested date is comming from.
      * If not set label is derived from field name.
      */
    label?: string
  } | {
     type: "Field",...
  },
  events: Event[]
}
```

If the dateField is specified this could be put into the date picker:

image

Or alternatively just have a line with the label and date in the "create encounter" modal, e.g.:

image

clemens-msupply:
Fixes #2857 

# 👩🏻‍💻 What does this PR do? 
- Introduces a new EncounterNode property to fetch `suggested_next_encounter` information.
- When creating an encounter, autoselects the type of the latest encounter.
- Also sets the suggested next encounter date if provided
- Mark the suggested day in the date picker

image

Please use the latest schema from the schema repo to test this change.
- Create a patient with an HIVCareEncounter and some dispensed pills
- Check the end of supply date in this encounter
- Save as visited
- Create new encounter
- The modal should have HIV Care Encounter preselected
- Furthermore the end of supply date should be selected
- Open calendar and hover the marked day
- Suggested dates in the past should be ignored


client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
```diff
@@ -33,6 +34,47 @@ import {
...
+  const { highlightedDays = [], day, outsideCurrentMonth, ...other } = props;
+
+  const matchingDay = highlightedDays.find(it => {
+    const date = it.datetime;
+    return (
+      day.getFullYear() === date.getFullYear() &&
+      day.getMonth() === date.getMonth() &&
+      day.getDate() === date.getDate()
+    );
+  });
```
mark-prins, line 56: adding in the date-fns `isSameDay` function to DateUtils will allow the simpler:
```suggestion
  const matchingDay = highlightedDays.find(it =>
    DateUtils.isSameDay(it.datetime, day)
  );
```

client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
```diff
@@ -33,6 +34,47 @@ import {
...
+  const isSelected = !props.outsideCurrentMonth && !!matchingDay;
+  return (
+    <Badge
+      key={props.day.toString()}
+      overlap="circular"
+      badgeContent={
+        isSelected ? (
+          <Tooltip title={matchingDay?.label ?? ''}>
+            <span>⌛</span>
```
mark-prins, line 65: I would have thought that an alert indicator is more appropriate?
```suggestion
            <span>⚠️</span>
```

or a _stop_ or _empty_ sort of thing.

mark-prins (approved):
👍  working nicely!

It did take me ages to get the configuration set up and working though - if you could provide a database to test with it would make getting started with PR review much easier. 

I've had trouble running the program schemas update in the past so have avoided it, which is part of the problem. Might need some help getting started next time there's a program PR

clemens-msupply:
Thanks for the review. I did some more testing and fixed the following two issues:

- Don't suggest date if there is already an encounter at the suggested
day.
- Allow editing the startDatetime if the date is suggested (previously it was fixed to the suggested day)

Setting up the schema should be relatively easy now with the 4d foot-runner script but yes having a 4d test datefile would be even easier...

PR #2357, opened 2023-10-12
2310 multi unit item] add api, service and repository

roxy-dao:
Fixes #2310

# 👩🏻‍💻 What does this PR do? 
- Create `pack_unit` table.
- Logic to calculate `total_number_of_packs` or `total_number_of_lines` based on store's stock on hand to find the most used variant based on highest total.
- Graphql endpoint to get all pack units.

# 🧪 How has/should this change been tested? 
- [ ] Run tests? (Probably not the best test 😭 )

## 💌 Any notes for the reviewer?
Not sure if we need translations at this point since omSupply central hasn't been configured yet? Can create another issue for it ig

PR #3027, opened 2024-02-20
2857 program event clean up

clemens-msupply:
Not directly part of #2857

# 👩🏻‍💻 What does this PR do? 
Some small refactoring that I noticed while working on #2357. In other places the allowed_ctx permission check is done in the service, so I did the same for program events. Previously, it made it hard to validate that permissions are taken into account correctly.


jmbrunskill (approved):
Looks good to me Clemens, not setup to test it so only looked at the code so far...

PR #3039, opened 2024-02-21
3038 map to stock line ids

### LINKED ISSUE #3038: fix shipment line ids that should be stock line ids
The selected outbound/inbound shipment line ids need to be mapped to stock line ids to be passed to the generateReturnLine queries!

lache-melvin:
Fixes #3038

# 👩🏻‍💻 What does this PR do? 
 

In previous PRs I had conflated inbound/outbound shipment line ids and stock line ids. This PR corrects that, mapping to the selected stock line ids to be passed to the generate return line queries!

It felt easiest/cleanest to do this mapping in the useSelectedIds hooks - just return the stockLinesId of each selected line, rather than the selected line ids, and then having to map at a later point. Does the code read/flow ok to you?

It all still compiles/runs 🎉

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
@@ -119,7 +119,6 @@ export const Toolbar: FC<{
           </Box>
           <DropdownMenu label={t('label.actions')}>
             <DropdownMenuItem
-              disabled={shipment?.status === InvoiceNodeStatus.New}
               IconComponent={ArrowLeftIcon}
               onClick={() => onReturnLines(selectedStockLineIds)}
             >
--- a/client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
@@ -13,28 +13,9 @@ import {
   InvoiceNodeStatus,
   Alert,
   ArrowLeftIcon,
-  useTableStore,
 } from '@openmsupply-client/common';
 import { SupplierSearchInput } from '@openmsupply-client/system';
 import { InboundRowFragment, useInbound } from '../api';
-
-const useSelectedIds = () => {
-  const { items, lines } = useInbound.lines.rows();
-
-  const selectedIds =
-    useTableStore(state => {
-      const { isGrouped } = state;
-
-      return isGrouped
-        ? items
-            ?.filter(({ id }) => state.rowState[id]?.isSelected)
-            .map(({ lines }) => lines.flat())
-            .flat()
-        : lines?.filter(({ id }) => state.rowState[id]?.isSelected);
-    })?.map(({ id }) => id) || [];
-
-  return selectedIds;
-};
 
 const InboundInfoPanel = ({
   shipment,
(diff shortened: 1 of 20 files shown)


client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
```diff
@@ -138,8 +119,9 @@ export const Toolbar: FC<{
           </Box>
           <DropdownMenu label={t('label.actions')}>
             <DropdownMenuItem
+              disabled={shipment?.status === InvoiceNodeStatus.New}
```
lache-melvin, line 122: Doesn't look like InboundShipmentLines have a stock line until they are changed to delivered status... do you think this requires a tooltip explanation or just disabling is fine?

client/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
```diff
@@ -105,12 +107,24 @@ export const Toolbar: FC<{
...
+                !selectedStockLineIds.length
+                  ? t('message.placeholder-lines-cannot-be-returned')
+                  : ''
+              }
             >
-              {t('button.return-lines')}
-            </DropdownMenuItem>
+              <span>
+                <DropdownMenuItem
+                  disabled={!selectedStockLineIds.length}
```
lache-melvin, line 120: Outbound shipment placeholder lines also don't have stock lines, so can't be selected to return - I did put a tooltip here to explain as it didn't seem obvious to me, but curious whether you think it is needed?

lache-melvin: Probably also shouldn't be able to return from OutboundShipment until after it is shipped? But I think we were procrastinating putting in some of this logic... so I was just trying to handle the cases where you select rows but don't end up with any connected stock lines!

client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
andreievg: I think our current strategy with UI/UX is to avoid disabling, and to show a warning, would prefer this tbh.

client/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -36,7 +36,7 @@ export const DetailView: FC = () => {
     onOpen: onOpenReturns,
     onClose: onCloseReturns,
     isOpen: returnsIsOpen,
-    entity: selectedOutboundShipmentLineIds,
+    entity: stockLineIds,
```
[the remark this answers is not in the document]
andreievg: Back track on this comment, stockLineIds would work just as well, and it's consistent with outboundReturn

client/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
andreievg: I am kind of moving away from thinking tooltips, since we can't really do it for mobile (and that's actually our primary focus, at least that's where the most deployment will be).

I think we would have warning when this is clicked, when status is not 'Shipped' or 'Delivered' or 'Verified', saying can only return outbound that was shipped, cannot have placeholder lines in those statuses, so we wouldn't need this warning

andreievg:
@lache-melvin, thanks mostly looks good, I am keen to approve but wondering about the warning messages, vs tooltips and vs disable

lache-melvin:
> @lache-melvin, thanks mostly looks good, I am keen to approve but wondering about the warning messages, vs tooltips and vs disable

Thanks Andrei - have updated to the warnings, that feels cleaner! And more consistent :)

PR #3041, opened 2024-02-21
3022 add items to existing outbound return

### LINKED ISSUE #3022: Add items to an existing outbound return

Similar to an inbound shipment, we want to be able to add items manually to an outbound return. We should be able to select from existing batches the stock we wish to return to a supplier.

lache-melvin:
Fixes #3022

# 👩🏻‍💻 What does this PR do? 
 

Adds an item selector to the outbound return modal, for adding new lines to an existing outbound return. The selected itemId is passed to generate return lines for available stock lines.

Screenshot 2024-03-04 at 12 11 05 PM

- From detail view, click add item button
- Select item from dropdown
- Dropdown disables, return line inputs appear for all the stock lines for that item
- Filling out and clicking OK adds those new lines to the return
- Also: the return from inbound shipment modal doesn't show the item selector!

Most of the modals have their height adjust to fit the content, but this one it feels a bit iffy for items with lots of lines, particularly the jump in size as you switch to the second table to enter the return reason. Thoughts on this? Also should the cancel/ok buttons be fixed/always visible at the bottom of the modal?


client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
```diff
@@ -0,0 +1,58 @@
...
+  ];
+
+  const getActiveStep = () => {
+    const step = returnsSteps.findIndex(step => step.tab === currentTab);
+    return step === -1 ? 0 : step;
+  };
+
+  return (
+    <TabContext value={currentTab}>
+      <WizardStepper activeStep={getActiveStep()} steps={returnsSteps} />
```
lache-melvin, line 43: Shall I make a new issue to add some kind of linking feature to this stepper? So you could click on previous steps to go back to them?

client/packages/invoices/src/Returns/modals/OutboundReturn/OutboundReturn.tsx
```diff
@@ -85,21 +74,20 @@ export const OutboundReturnEditModal = ({
...
-              />
-            </TabPanel>
-            <TabPanel value={Tabs.Reason}>
-              <ReturnReasonsTable
-                lines={lines}
-                updateLine={line => update(line)}
-              />
-            </TabPanel>
-          </TabContext>
+          {returnId && (
```
lache-melvin, line 77: There's a couple of conditionals which I don't love (hide/show the item selector, use the insert or update mutation...) - but otherwise basically identical between creating a return from inbound shipment lines, and adding an item to an existing return. Keen for feedback here!

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <TabPanel value={Tabs.Quantity}>
+        <QuantityToReturnTable
+          lines={lines}
+          updateLine={line => update(line)}
+        />
+      </TabPanel>
+      <TabPanel value={Tabs.Reason}>
+        <ReturnReasonsTable
+          lines={lines.filter(l => l.numberOfPacksToReturn > 0)}
```
lache-melvin, line 52: Filtering here so only need to add reasons/comments to the lines we're actually returning... does the UX feel alright to you?

client/packages/invoices/src/Returns/modals/OutboundReturn/ItemSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+          />
+        </Grid>
+      </ModalRow>
+      {item && (
+        <ModalRow margin={3}>
+          <ModalLabel label={t('label.unit')} justifyContent="flex-end" />
+          <BasicTextInput
+            disabled
+            sx={{ width: 150, marginBottom: '10px' }}
+            value={item.unitName ?? ''}
```
CarlosNZ, line 62: You shouldn't need the fallback `?? ''` since the item must be defined here (I tried removing it and it seemed fine)

client/packages/invoices/src/Returns/modals/OutboundReturn/ItemSelector.tsx
```diff
@@ -0,0 +1,68 @@
...
+                ? undefined
+                : item =>
+                    item.availableStockOnHand !== 0 &&
+                    !existingItemIds?.some(id => id === item.id)
+            }
+          />
+        </Grid>
+      </ModalRow>
+      {item && (
+        <ModalRow margin={3}>
```
CarlosNZ, line 57: Can we vertically center the label here? The one for the Item row already is, so would be nice if this one matched:
Screenshot 2024-03-05 at 11 07 16 AM

client/packages/invoices/src/Returns/modals/OutboundReturn/OutboundReturn.tsx
CarlosNZ: Seems reasonable to me 👍🏼

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
CarlosNZ: Yes, definitely, I naturally tried to go back and realised I couldn't! 

I think a "Back" button is probably enough, but we should replicate the UX of other steppers.

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
CarlosNZ: Can we disable/hide the "Okay & Next" button until some values are added? I agree, it's weird going to the second step and just seeing "Nothing here". 

I think the preferred UX is to keep the button active, but show an alert when trying to do something not allowed -- for example when you try and issue a quantity of 0 on an outbound shipment line. Maybe we should do that (can do in another issue if you want)

CarlosNZ (changes requested):
Cool, looks good. I've made a few comments -- feel free to decide what you want to address and what you want to push into another issue.

I'd like to do something about the Table column widths, too, if possible. The side-scrolling is annoying, especially when you have columns like "itemCode" taking up way more width than we need. New issue?

client/packages/invoices/src/Returns/api/hooks/utils/useReturnsApi.ts
```diff
@@ -27,7 +27,8 @@ export const useReturnsApi = () => {
       [...keys.outboundList(), params] as const,
     detail: (invoiceNumber: string) =>
       [...keys.base(), storeId, invoiceNumber] as const,
-    newReturns: () => [...keys.base(), storeId, 'newReturns'] as const,
+    newReturns: (itemId?: string) =>
+      [...keys.base(), storeId, 'newReturns', itemId] as const,
```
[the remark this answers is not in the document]
lache-melvin: Oh yes definitely, thank you!

client/packages/invoices/src/Returns/api/operations.graphql
```diff
@@ -77,11 +78,11 @@ query inboundReturns(
...
-query generateOutboundReturnLines($stockLineIds: [String!], $storeId: String!) {
-  generateOutboundReturnLines(
-    input: { stockLineIds: $stockLineIds }
-    storeId: $storeId
-  ) {
+query generateOutboundReturnLines(
+  $input: GenerateOutboundReturnLinesInput!
+  $storeId: String!
+) {
+  generateOutboundReturnLines(input: $input, storeId: $storeId) {
```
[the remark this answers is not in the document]
lache-melvin: No it doesn't create anything, but it's a weird one to name as it does a mix of fetching existing lines from the DB and generating placeholder lines to be filled out and saved... `calculate` isn't bad? Or something like make/pull?

Will put this aside to a separate issue, will be a significant backend renaming

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
lache-melvin: Except the plan was to delete return lines by setting the return quantity to 0, so it's technically a valid input? I think I'll put it aside as something to discuss w PO?

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
lache-melvin: Yeah cool - I'll make a note to do this separately too... we'd need to update our common `Modal` component to support another type of button

lache-melvin:
Thanks Carl! Have made most of the changes, and noted the rest in the epic. Also added the `onRowClick` to open the modal, but beware the lines in the modal don't yet include the existing lines! That's coming in #3154 😁

client/packages/invoices/src/Returns/api/operations.graphql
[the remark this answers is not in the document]
CarlosNZ: Okay, it's probably fine as is. I see "generate" is used in quite a few places. As long as we have a consistent meaning for what "generate" means, then it's all good 👍🏼 

So to me, from this context, "generate" should always mean something like "creates new entities in memory, deduced from other values, but hasn't been serialised to any permanent storage yet". Sound about right?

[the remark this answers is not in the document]
lache-melvin: Yeah exactly - I'll make a note to add some introspectable comments on this query so this is clarified!

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
CarlosNZ: PO? 

I think my preferred solution would be to show a warning when *all* quantities are 0 and you click OK: "You have specified no quantities for any of the available lines. If you proceed then no lines will be added to the return", or something like that.

But yeah, something to discuss as a group.

CarlosNZ (approved):
Looks good, thanks. Few little side issues to deal with at some point -- lets' clarify exactly what those are later today.

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
lache-melvin: Ah haha as in Richard the PO 😁 Oo yeah a warning like that sounds like a good idea! Just might need to have a different message based on whether we are adding or editing - will add issue :)

PR #2995, opened 2024-02-15
 2994 New program enrolment name is breaking build

roxy-dao:
Fixes #2994

# 👩🏻‍💻 What does this PR do? 
Use name_dsl instead of program_enrolment_dsl since program enrolment doesn't have patient id anymore.

# 🧪 How has/should this change been tested? 
1. Make sure program enrolment name filter is still working
- [ ] Set up programs
- [ ] Go to patients 
- [ ] Enrol a few in different programs
- [ ] Add the `program enrolment` filter 
- [ ] Try using it
2. Make sure it builds
- [ ] run `yarn build`

PR #3045, opened 2024-02-22
Create permissions for assets

mark-prins:
Fixes #2995 

# 👩🏻‍💻 What does this PR do? 
 
Create some placeholder permissions so that we can consume them when making the graphQL changes.

Just check that it looks ok

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/repository/src/migrations/v1_08_00/assets/asset_catalogue_item.rs
+++ b/server/repository/src/migrations/v1_08_00/assets/asset_catalogue_item.rs
@@ -11,7 +11,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
             asset_category_id TEXT NOT NULL REFERENCES asset_category(id),
             asset_type_id TEXT NOT NULL REFERENCES asset_type(id),
             manufacturer TEXT,
-            model TEXT NOT NULL,
+            model TEXT NOT NULL
         );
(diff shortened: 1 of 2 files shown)


server/repository/src/migrations/v1_08_00/assets/asset_catalogue_item.rs
```diff
@@ -11,7 +11,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
             asset_category_id TEXT NOT NULL REFERENCES asset_category(id),
             asset_type_id TEXT NOT NULL REFERENCES asset_type(id),
             manufacturer TEXT,
-            model TEXT NOT NULL,
+            model TEXT NOT NULL
```
mark-prins, line 14: offroad: noticed that this failed for postgres 🙄

fergie-nz (approved):
Looks great! 

Curious why we would or wouldn't add a permission to query logs?

Or would this fit under the generic "QueryLog" in reporting?

mark-prins:
> Curious why we would or wouldn't add a permission to query logs?

I thought the QueryAsset permission could be used, thinking that if you can see assets then you can see the logs too 🤷 
possibly influenced by the thought that I'm adding a lot of permissions for a simple feature..

PR #3047, opened 2024-02-22
2960 map temperature log datetimes to local

### LINKED ISSUE #2690: Fridge tag datetime as local datetime
## What went wrong? 😲

When importing fridge tag temperature logs, datetimes are saved in db as is (logs timestamps as utc).

## Expected behaviour 🤔

Would expect datetimes = localtime on the machine that is doing the import,

## How to Reproduce 🔨

Try importing this file: https://github.com/msupply-foundation/temperature-sensor/blob/main/data/FridgeTag%202L/130500109088_202206081014.txt

Find one of the logs in file and compare to UI, it won't match. UTC equivalent will match though

## Solution

Add timestamps offset to upload fridge tag endpoint. 

I've used something like this but just using server local datetime, but the client may not match the server. I am sure that chrono will have a way to convert NaiveDateTime to Datetime<Utc> given a timezone offset.

This is what was done for technet:

https://github.com/msupply-foundation/open-msupply/blob/79a9b5ee08ce8e9b29dfac0d4795255813a1456a/server/service/src/emd_driver.rs#L101-L123

fergie-nz:
Fixes #2690 

# 👩🏻‍💻 What does this PR do? 
Adds mapping from naive date time to local naive date time at the time of importing fridge tag data.

Import a fridge tag file such as this:
[fridge tag import.txt](https://github.com/msupply-foundation/open-msupply/files/14368174/fridge.tag.import.txt)
See that times are correct rather than assumed UTC.

Some issues remain with this implementation - see discussion [here](https://github.com/orgs/msupply-foundation/projects/21/views/7?pane=issue&itemId=50328498) and [here](https://github.com/msupply-foundation/temperature-sensor/pull/1)


jmbrunskill (approved):
Looks good to me Fergus

fergie-nz:
Some tests are failing which I believe are due to issues with the develop branch, however I am not certain.

I have been holding off merging until I am sure - but maybe someone else can offer more clarity?

jmbrunskill:
> Some tests are failing which I believe are due to issues with the develop branch, however I am not certain.
> 
> I have been holding off merging until I am sure - but maybe someone else can offer more clarity?

I just tried this on your branch and tests pass, let's check in again tomorrow on this...

PR #3049, opened 2024-02-22
2972 create asset tables

### LINKED ISSUE #2972: Create asset tables
Part of epic #2896 

### Moneyworks Jobcode 🧰

`OMSCC:AST`

jmbrunskill:
Fixes #2972 

# 👩🏻‍💻 What does this PR do? 
Adds the `asset` and `asset_internal_location` tables.

For MVP, we're considering leaving out locations but that would compromise MVP Goal #1
>Provide real-time visibility of functional storage capacity (used vs unused volume) of CCE at a national level with granularity down to the Service Delivery Points.
We could have a single location associated with the asset if we allowed hierarchical locations, which does exist, however I assume that's more work to bring in to the MVP than what I've currently done...

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/repository/src/db_diesel/assets/asset.rs
+++ b/server/repository/src/db_diesel/assets/asset.rs
@@ -1,0 +1,264 @@
+use super::asset_row::{
+    asset::{self, dsl as asset_dsl},
+    AssetRow,
+};
+
+use diesel::{dsl::IntoBoxed, prelude::*};
+
+use crate::{
+    diesel_macros::{
+        apply_date_filter, apply_equal_filter, apply_sort, apply_sort_no_case, apply_string_filter,
+    },
+    repository_error::RepositoryError,
+    DBType, DateFilter, DatetimeFilter, EqualFilter, Pagination, Sort, StorageConnection,
+    StringFilter,
+};
+
+type Asset = AssetRow;
+
+pub enum AssetSortField {
+    SerialNumber,
+    InstallationDate,
+    ReplacementDate,
+}
+
+pub type AssetSort = Sort<AssetSortField>;
+
+#[derive(Clone)]
+pub struct AssetFilter {
+    pub id: Option<EqualFilter<String>>,
+    pub serial_number: Option<StringFilter>,
+    pub category_id: Option<EqualFilter<String>>,
+    pub type_id: Option<EqualFilter<String>>,
+    pub catalogue_item_id: Option<EqualFilter<String>>,
+    pub installation_date: Option<DateFilter>,
+    pub replacement_date: Option<DateFilter>,
+}
+
+impl AssetFilter {
+    pub fn new() -> AssetFilter {
+        AssetFilter {
+            id: None,
+            serial_number: None,
+            category_id: None,
+            type_id: None,
+            catalogue_item_id: None,
+            installation_date: None,
+            replacement_date: None,
+        }
+    }
+
+    pub fn id(mut self, filter: EqualFilter<String>) -> Self {
+        self.id = Some(filter);
+        self
+    }
+
+    pub fn serial_number(mut self, filter: StringFilter) -> Self {
+        self.serial_number = Some(filter);
+        self
+    }
+
+    pub fn category_id(mut self, filter: EqualFilter<String>) -> Self {
+        self.category_id = Some(filter);
+        self
+    }
+
+    pub fn type_id(mut self, filter: EqualFilter<String>) -> Self {
+        self.type_id = Some(filter);
+        self
+    }
+
+    pub fn catalogue_item_id(mut self, filter: EqualFilter<String>) -> Self {
+        self.catalogue_item_id = Some(filter);
+        self
+    }
+
+    pub fn installation_date(mut self, filter: DateFilter) -> Self {
+        self.installation_date = Some(filter);
+        self
+    }
+
+    pub fn replacement_date(mut self, filter: DateFilter) -> Self {
+        self.replacement_date = Some(filter);
+        self
+    }
+}
+
+pub struct AssetRepository<'a> {
+    connection: &'a StorageConnection,
+}
+
+impl<'a> AssetRepository<'a> {
+    pub fn new(connection: &'a StorageConnection) -> Self {
+        AssetRepository { connection }
+    }
+
+    pub fn count(&self, filter: Option<AssetFilter>) -> Result<i64, RepositoryError> {
+        let query = create_filtered_query(filter);
+
+        Ok(query.count().get_result(&self.connection.connection)?)
+    }
+
+    pub fn query_one(&self, filter: AssetFilter) -> Result<Option<Asset>, RepositoryError> {
+        Ok(self.query_by_filter(filter)?.pop())
+    }
+
+    pub fn query_by_filter(&self, filter: AssetFilter) -> Result<Vec<Asset>, RepositoryError> {
+        self.query(Pagination::all(), Some(filter), None)
+    }
+
+    pub fn query(
+        &self,
+        pagination: Pagination,
+        filter: Option<AssetFilter>,
+        sort: Option<AssetSort>,
+    ) -> Result<Vec<Asset>, RepositoryError> {
+        let mut query = create_filtered_query(filter);
+
+        if let Some(sort) = sort {
+            match sort.key {
+                AssetSortField::SerialNumber => {
+                    apply_sort_no_case!(query, sort, asset_dsl::serial_number);
+                }
+                AssetSortField::InstallationDate => {
+                    apply_sort!(query, sort, asset_dsl::installation_date)
+                }
+                AssetSortField::ReplacementDate => {
+                    apply_sort!(query, sort, asset_dsl::replacement_date)
+                }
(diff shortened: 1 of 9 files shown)


server/repository/src/db_diesel/assets/asset_row.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+#[derive(Clone, Insertable, Queryable, Debug, PartialEq, AsChangeset, Eq, Default)]
+#[table_name = "asset"]
+pub struct AssetRow {
+    pub id: String,
+    pub store_id: String,
+    pub serial_number: String,
+    #[column_name = "asset_category_id"]
```
jmbrunskill, line 32: Question: Is it good to get the database as `asset_category_id` I like that it's not then confused with some other category_id in the db level, but also makes things slightly clunky like this...

server/repository/src/db_diesel/assets/asset_row.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub category_id: String,
+    #[column_name = "asset_type_id"]
+    pub type_id: String,
+    pub catalogue_item_id: Option<String>,
```
jmbrunskill, line 36: Given question above, arguably this should be `asset_catalogue_item_id` which feels too long...

server/repository/src/db_diesel/assets/asset_row.rs
```diff
@@ -0,0 +1,100 @@
...
+        let result = asset
+            .filter(id.eq(asset_id))
+            .first(&self.connection.connection)
+            .optional()?;
+        Ok(result)
+    }
+
+    pub fn delete(&self, asset_id: &str) -> Result<(), RepositoryError> {
+        diesel::update(asset.filter(id.eq(asset_id)))
+            .set(deleted_datetime.eq(Some(chrono::Utc::now().naive_utc())))
```
jmbrunskill, line 96: Implemented soft delete for assets, seems easiest to do it now rather than later...

server/repository/src/mock/mod.rs
```diff
@@ -340,6 +340,7 @@ impl MockDataInserts {
     }
 
     pub fn stores(mut self) -> Self {
+        self.names = true;
```
jmbrunskill, line 343: I've done this in other projects, I can't see any harm in adding these dependancies in?

server/repository/src/migrations/v1_08_00/assets/asset.rs
```diff
@@ -0,0 +1,47 @@
...
+            serial_number TEXT NOT NULL, 
+            asset_category_id TEXT REFERENCES asset_category (id),
+            asset_type_id TEXT REFERENCES asset_type (id),
+            asset_catalogue_item_id TEXT REFERENCES asset_catalogue_item (id),
+            installation_date {DATE},
+            replacement_date {DATE},
+            deleted_datetime {DATETIME},
+            created_datetime {DATETIME} NOT NULL,
+            modified_datetime {DATETIME} NOT NULL,
+            UNIQUE (serial_number, deleted_datetime) --If something doesn't have a serial number, one can be generated?
```
jmbrunskill, line 21: Worth verifying this comment. Feels like if we don't reliably have a serial number we'll need something to identify the asset, I guess people could just type `fridge1` as a serial number?

server/repository/src/db_diesel/assets/asset.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[actix_rt::test]
+    async fn test_asset_query_repository() {
+        // Prepare
+        let (_, storage_connection, _, _) = test_db::setup_all(
+            "test_asset_query_repository",
+            MockDataInserts::none().stores(),
+        )
+        .await;
+
+        // TODO: Replace this reference data with mock data, or inserted data from https://github.com/msupply-foundation/open-msupply/issues/3035
```
jmbrunskill, line 209: Haven't done this as since we'd doing migrations, it seems a good idea to wait...

server/repository/src/db_diesel/assets/asset.rs
```diff
@@ -0,0 +1,271 @@
...
+    ModifiedDatetime,
+}
+
+pub type AssetSort = Sort<AssetSortField>;
+
+#[derive(Clone)]
+pub struct AssetFilter {
+    pub id: Option<EqualFilter<String>>,
+    pub serial_number: Option<StringFilter>,
+    pub category_id: Option<EqualFilter<String>>,
```
mark-prins, line 31: would it need to have a `category` and `type` different to that on the `catalogue_item` though?
I'm thinking probably not - that the class, category and type are defined by the catalogue_item and are immutable

server/repository/src/mock/mod.rs
mark-prins: makes sense to have them in, thanks!

server/repository/src/migrations/v1_08_00/assets/asset.rs
mark-prins: I'm not sure that people will be populating serial number reliably; and while the number should exist and should be unique.. my experience is that this isn't very real world. I was thinking that it would be a nullable text field, which people may have filled in.

auto generating a serial would give incorrect data: we'd have the serial populated but it wouldn't be the serial of the actual asset, which could lead to confusion. better I think to have it blank.

mark-prins (commented):
Working nicely thanks!
I think we have a different purpose in mind for serial_number, so should clarify that before merging in. aside from that, all good!

server/repository/src/migrations/v1_08_00/assets/asset.rs
jmbrunskill: Agreed, I think we'll need something to identify which asset is which if there is no serial number.
I'll add a `name` field which can give a human readable way to identify it.

mark-prins (approved):
👍 As discussed - with adding in a descriptive column of some kind and allowing empty serial number this is all good thanks

server/repository/src/db_diesel/assets/asset.rs
jmbrunskill: I agree, was just thinking if we record it here it would make the filters easy.
I'll take them out and update the filters...

server/repository/src/migrations/v1_08_00/assets/asset.rs
adamdewey: Ideally we would be able to generate and assign unique asset numbers so the government can use Open mSupply as the source of truth for their asset register numbering. 

Typically I think countries would be printing out the unique asset numbers and sticking them onto the assets so that they can be tied back to the asset register. 

Serial numbers would typically be generated by the manufacturer of the asset and may or may not be present, as mentioned, so couldn't serve the same purpose as a centralised national asset number.

I'm unsure if `name` would serve the purpose either - I feel like users would populate that field with things like `Fridge in the green room` but that doesn't help if the fridge is later moved as there is no sticker on the asset saying `Fridge in the green room`. 

@richardmoizeau - thoughts?

jmbrunskill: I think we'll be generating a unique number as part of the QR Code part, which I guess would be a stick over sorts. I guess we could probably make the number available somewhere obvious in the UI? Not currently in any mockups though.

adamdewey: Ah, it's not super clear, but I think this is supposed to be the unique asset number: 

image

server/repository/src/db_diesel/assets/asset_catalogue_item.rs
```diff
@@ -154,13 +154,25 @@ fn create_filtered_query(filter: Option<AssetCatalogueItemFilter>) -> BoxedAsset
             code,
             manufacturer,
             model,
-            ..
+            category: _, // Handled in query() function
```
jmbrunskill, line 157: FYI - @mark-prins I updated this as it looks like you'd missed the new id filters.
This pattern of marking something as unused will warn you for anything new added so it won't be missed.

server/repository/src/db_diesel/assets/asset.rs
```diff
@@ -0,0 +1,287 @@
...
+            apply_equal_filter!(
+                sub_query,
+                Some(category_id),
+                asset_catalogue_item_dsl::asset_category_id
+            );
+            query = query.filter(asset_dsl::asset_catalogue_item_id.eq_any(sub_query));
+        }
+
+        if let Some(class_id) = class_id {
+            let mut sub_query = asset_catalogue_item_dsl::asset_catalogue_item
```
jmbrunskill, line 215: @mark-prins this subquery option is another approach to do what you've done previously.
Should be slightly better performance as a sub query and means you aren't mixing the location where we're handling the query filters. Downside is diesel doesn't make it easy to get right...

server/repository/src/migrations/v1_08_00/assets/asset.rs
jmbrunskill: Ok fair enough, have added a code field. Will probably use it for the QR code as well.

Issue #2897, opened 2024-02-06
Supplier/Customer Returns


DhanyaHerath:
# Design docs

https://docs.google.com/document/d/1szgYLDM-euSgpp9IdWom8K_OYJoxxENADiExBObjJKA/edit?usp=sharing

# Tasks

#### I think we should start with return from existing inbound/outbound shipments, with mock api

- [x] Can select lines from inbound shipment for outbound return (1h) #2921 
- [x] New return items UI (with all selected lines), for outbound return (8h) #2921 
    * Api should really provide anything that is required for return items view, based on provided ids of lines
- [x] A reason step in return items UI, for outbound return and create return on OK (5h) #2929 
    * Api to post stock line id, reason and comment, which in turn creates return (and returns it in API call)
- [x] Outbound return detail view (12h) #2920
    * Would be the same as invoice query ?
- [x] Can select lines from outbound shipment for inbound return (1h) #2944 
- [x] New return items UI (with all selected lines), for inbound return  (8h) #2944 
    * Api should really provide anything that is required for return items view, based on provided ids of lines
- [x] A reason step in return items UI, for inbound return and create return on OK (5h) #2950 
    * Api to post stock line id ? (or outbound line ids? or full info about batch/expiry, quantity etc.. ?) reason and comment, which in turn creates return (and returns it in API call)
- [x] Inbound return detail view (12h) #3163 
     * Would be the same as invoice query ?

#### Then list and creating new returns, and selecting items

- [x] Outbound return list view (5h) #2953
  * Api i think just new type of invoice
- [x] Inbound return list view (5h) #2998 
  * Api i think just new type of invoice
- [x] Add new outbound return from list (3h) #3008 
  * Same as outbound/inbound ?
- [x] Add new inbound return from list (3h) #3014 
  * Same as outbound/inbound ?
- [x] Add new line (by item) to outbound return (from existing batches) (8h) #3022 
  * Would prefer API to be a lot more direct here then how it's down in inbound/outbound
- [x] Add new line (by item) to inbound return (from previous outbound shipments or manually) (8h) #3189 
  * Would prefer API to be a lot more direct here then how it's down in inbound/outbound
- [x] Change statuses (2h)
- [x] Print (4h) #3190 

#### Synchronisation and transfers (creating inbound return from outbound return, in omSupply site) 

- [x] Add schema changes for outbound/inbound return record type, make sure they are syncing (3h) #3056
   * New invoice type
[...]

PR #3053, opened 2024-02-23
3043 generate outbound return lines

### LINKED ISSUE #3043: implement generateOutboundReturnLines

Look up stock line by ids union with stock line by item_id -> populate SupplierReturnLineNode
If existing return is specified, populate reason + comment + numberOfPacks from invoice_lines in the existing return

lache-melvin:
Fixes #3043

# 👩🏻‍💻 What does this PR do? 
 

Implements API for generateOutboundReturnLines - so we see real lines in the UI now 🎉 

Screenshot 2024-02-23 at 1 16 40 PM

Also implements `item_id` filter for when implement the add from item view, but haven't put the return_id piece in yet, as we have no returns to query 😁 

Unit tests added for the service layer... should I be adding from graphql layer too?

- Go to a delivered/finalised inbound shipment
- Select some lines
- Choose `return selected lines`
- See those lines populated in the returns modal


server/graphql/types/src/types/outbound_return_line.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct OutboundReturnLineNode {
+    pub return_line: OutboundReturnLine,
+}
+
+impl OutboundReturnLineNode {
+    pub fn from_domain(return_line: OutboundReturnLine) -> OutboundReturnLineNode {
+        OutboundReturnLineNode { return_line }
+    }
+
+    pub fn item_row(&self) -> &ItemRow {
```
lache-melvin, line 38: I only this pattern with `row`, but does exposing `item_row` and `stock_line_row` fit the pattern ok?

server/graphql/types/src/types/permissions.rs
```diff
@@ -33,6 +33,8 @@ pub enum UserPermission {
     OutboundShipmentMutate,
     InboundShipmentQuery,
     InboundShipmentMutate,
+    OutboundReturnQuery,
+    OutboundReturnMutate,
```
lache-melvin, line 37: As mentioned at stand up this morning - I think I've add the different permissions components in here, but there's something about needing to sync/fetch from central to update the user permissions, so they're not currently working...

We agreed to come back to permissions later, do them all in one go. Are we happy to leave these changes here or should I roll them back?

server/service/src/invoice/mod.rs
```diff
@@ -183,6 +186,17 @@ pub trait InvoiceServiceTrait: Sync + Send {
...
         batch_prescription(ctx, input)
     }
+
+    fn generate_outbound_return_lines(
+        &self,
+        ctx: &ServiceContext,
+        store_id: &str,
+        stock_line_ids: Vec<String>,
+        item_id: Option<String>,
+        return_id: Option<String>,
```
lache-melvin, line 196: Was seeing a bit of both - is there a newer/preferred pattern to pass an input struct vs many params?

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -0,0 +1,148 @@
...
+    stock_line_ids: Vec<String>,
+    item_id: Option<String>,
+    return_id: Option<String>,
+) -> Result<ListResult<OutboundReturnLine>, ListError> {
+    if stock_line_ids.is_empty() && item_id.is_none() && return_id.is_none() {
+        return Ok(ListResult {
+            count: 0,
+            rows: vec![],
+        });
+    }
```
lache-melvin, line 26: Should this instead be a validation step? Throw an error if the query inputs are empty?

Or alternatively, is there a way that I could define the filter below so that if nothing is passed in, the query returns no lines, as opposed to all the available stock lines?

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    let mut filter = StockLineFilter::new().is_available(true);
+
+    if !stock_line_ids.is_empty() {
+        filter.id = Some(EqualFilter::equal_any(
+            stock_line_ids.iter().map(String::clone).collect(),
+        ))
+    }
+    filter.item_id = item_id.map(|item_id| EqualFilter::equal_to(&item_id));
```
lache-melvin, line 35: With setting the filter up like this, if you pass in both stock line ids and an item id, we would only get back stock lines of that item... is that what you would expect of this API?

In practice, I think we'll only use one or the other - generate from stock_line_ids in inbound shipment, or generate from an item_id in the return detail view.

Other options:
- Throw an error - ask them to choose one or the other
- Make it an OR filter rather than AND

server/service/src/invoice/mod.rs
andreievg: Apart from store_id i prefer using 'input', benifits kind of small but here they are:

* less things to change if new input is added (just change the to_domain mapping)
* less chance of mapping error when dealing with the same type if destructure/restructure is used, i.e.


```rust

struct ServiceInput {
    id_for_this: String,
    id_for_that: String
}

struct GraphqlInput {
    id_for_this: String,
    id_for_that: String
}

impl GraphqlInput {
    fn to_domain(self) -> ServiceInput {
        GraphqlInput {
            id_for_this,
            id_for_that
        } = self;

        ServiceInput {
            id_for_this,
            id_for_that
        }
    }
}

```

vs

```rust
fn service(id_for_this: String, id_for_that: String) { /* */ }

// in Graphql, when mapping

service(input.id_for_that, input.id_for_this) // got this and that mixed up
```

server/graphql/types/src/types/outbound_return_line.rs
andreievg: I like it, and makes sense, thank you

server/graphql/types/src/types/permissions.rs
andreievg: This is fine, i've just added `Add permission checks, and sync them` [to epic](https://github.com/msupply-foundation/open-msupply/issues/2897)

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
andreievg: I wish we could do union inputs in grahpql (from memory I think async-grahpql actually added it, but it's not part of the standard graphql).

I think what you've done is fine, maybe a comment in graphql about this (btw async-graphql should expose comments [like this](https://github.com/msupply-foundation/open-msupply/blob/ae7135cb1ae2b6cd08f0ef3f92d558c04e021c90/server/graphql/types/src/types/invoice_query.rs#L38-L42) as introspectable comments)

andreievg: Hmm this is a bit trickier on second thought, because we need to only show available stock_lines when item is selected (don't want to list all of the stock_lines that ever existing for that item), but when stock_lines are selected we need to show even if not available (so that user doesn't question, 'why' i couldn't return the thing i selected). What makes it ever more spicy is that if item is selected and return_id is selected, if stock_lines is not available, but we already return some quantity for that stock_line in return, it should be listed. 
I think this would be a great unit test use case, and I think the logic would mean, get stock_lines selected in stock_line_ids, get stock_lines and invoice_lines from return_id (for that item), get all stock_lines for that item that are available, then consolidate stock_lines that we've just selected and add numberOfPacks, reason, comment from invoice_lines

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -0,0 +1,148 @@
...
+            .collect();
+        let item_id = None;
+        let return_id = None;
+
+        let result = service
+            .generate_outbound_return_lines(&context, store_id, stock_line_ids, item_id, return_id)
+            .unwrap();
+
+        assert_eq!(result.count, 2);
+        assert_eq!(result.rows[0].stock_line.item_row.id, "item_a");
```
andreievg, line 118: For a more complex use case (i've just added a suggestion for one in the comment, not for this PR though), we would need to sort by id i think:

https://github.com/msupply-foundation/open-msupply/blob/ae7135cb1ae2b6cd08f0ef3f92d558c04e021c90/server/service/src/item_stats.rs#L173

andreievg (approved):
Looks great thanks, nothing pressing, approving

Please not the comment about item_id and filters, with available_stock etc..

andreievg:
I've also added `generateSupplierReturnLines (item_id and return_id)` to [the epic](https://github.com/msupply-foundation/open-msupply/issues/2897)

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
lache-melvin: ooo that is spicy! yes definitely will want some tests around that one 😅 Will be fun to sink teeth into soon!

PR #3058, opened 2024-02-23
3056 schema changes for returns

### LINKED ISSUE #3056: Add schema changes for inbound/outbound returns
Add schema changes for supplier/customer return record type
- New invoice type
- New reason table (duplicate inventory adjustment reason _row)
- Return reason id on invoice_line table ?
- Comment on invoice_line table

lache-melvin:
Fixes #3056 

# 👩🏻‍💻 What does this PR do? 
 

Adds the needed schema changes for returns work.

The migrations run, and the code is compiling?

sync translation stuff left as todo!


server/repository/src/db_diesel/number_row.rs
```diff
@@ -248,68 +254,87 @@ mod number_row_mapping_test {
...
-            }
-            NumberRowType::Repack => {
-                assert!(
-                    NumberRowType::try_from(NumberRowType::Repack.to_string()).unwrap()
-                        == NumberRowType::Repack
-                )
+        for number_row_type in vec![
+            NumberRowType::Program("EXAMPLE_TEST".to_string()),
+            NumberRowType::OutboundReturn,
+            NumberRowType::InboundReturn,
```
lache-melvin, line 260: @andreievg why did you add the for loop here?

server/repository/src/migrations/v1_07_00/return_reasons.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{migrations::sql, StorageConnection};
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    sql!(
+        connection,
+        r#"CREATE TABLE return_reason (
+            id TEXT NOT NULL PRIMARY KEY,
+            is_active BOOLEAN,
+            reason TEXT NOT NULL
```
lache-melvin, line 9: What do I need to do to actually populate some return reasons? 😅

server/service/src/invoice_line/stock_out_line/update/generate.rs
```diff
@@ -120,6 +120,7 @@ fn generate_line(
         r#type,
         note: input.note,
         inventory_adjustment_reason_id: None,
+        return_reason_id: None,
```
lache-melvin, line 123: Following on from earlier - if I have a separate `update_return_reason_id` method, is it fine to always set `return_reason_id` to `None` here, and then update separately, or should I pull `return_reason_id` off the input and pass through here?

server/repository/src/db_diesel/number_row.rs
andreievg: Just want to add something here, from discussion, for future reference:

Not very strong on this, but kind of wanted this test to also check explicitly that ` "INBOUND_RETURN" => Ok(NumberRowType::InboundReturn),` was actually added, not sure if it's justified, since for test to work `NumberRowType::InboundReturn` would also need to be added to the array above. vs previous change where just adding the variant here would mean a comment above would be read and the extras added to try_from

server/repository/src/migrations/v1_07_00/return_reasons.rs
```diff
@@ -0,0 +1,25 @@
...
+    )?;
+
+    sql!(
+        connection,
+        r#"ALTER TABLE invoice_line ADD COLUMN return_reason_id TEXT REFERENCES return_reason(id);"#
+    )?;
+    sql!(
+        connection,
+        r#"
+           CREATE INDEX index_invoice_line_return_reason_id ON invoice_line (return_reason_id);
```
andreievg, line 20: nice

server/service/src/invoice_line/stock_out_line/update/generate.rs
andreievg: Yeah i think it's fine to be None here for now, and we always update_return_reason_id for return updates, which as you've mentioned will be done in the update PR, thanks.

server/repository/src/mock/test_requisition_service.rs
```diff
@@ -390,6 +390,7 @@ pub fn mock_new_response_requisition_test_invoice() -> FullMockInvoice {
                     location_id: None,
                     stock_line_id: None,
                     inventory_adjustment_reason_id: None,
+                    return_reason_id: None,
```
andreievg, line 393: Just fyi, we got really annoyed with this at some point, mock tests shouldn't need to be updated when new field is added.

Then inline_init method was created, alongside [this script](https://github.com/msupply-foundation/open-msupply/blob/ae7135cb1ae2b6cd08f0ef3f92d558c04e021c90/server/util/src/to_inline_init.js#L4), that helps turn existing declarations with optionals/defaults to something like below

See server/repository/src/mock/item.rs file in [this commit](https://github.com/msupply-foundation/open-msupply/commit/c1b0b2523dff8af44180a57014060944b0a9ea99#diff-2a6027d7a6f5937f45d7ee9e4bd96781950749f8bc1098d008813e17f4dc632f)

Anyways, we haven't quite updated all the mock data, and maybe will at some point, but the preferred way now is not to use inline_init but to:

```rust
ItemRow {
   id: "mock_a".to_string(),
   name: "mock_item_a",
  ..Default::default()
}
```

andreievg (approved):
Changes look good to me, thank you

PR #3067, opened 2024-02-25
2963 add service layer and graphql for queries of assets, catalogues and types

### LINKED ISSUE #2960: Query to return a list of classes
Part of epic #2896 

For the lookup query, no pagination is required as the list isn't expected to be too long. No need for filters either, given the expected size of the list we can use an autocompleter for the lookup and filter client-side.
A default sort of alphabetical.

For the list view though, that will be shown in a standard datatable, so would require pagination and sort & filter on the class name.

### Moneyworks Jobcode 🧰

`OMSCC:AST`

fergie-nz:
Fixes #2960 #2963 #2969 

# 👩🏻‍💻 What does this PR do? 
Adds queries for asset catalogue items, types, classes, and categories
Also adds service layer functions asking for query_catalogue_item permissions

Data for asset catalogue items (and their classes, categories, and types) is pre populated in #3065 (Thanks @jmbrunskill!) which has been merged into this branch. You can run this branch and test the queries on graphiql with auto populated data. How easeful!


server/graphql/catalogue/src/asset_catalogue_item_queries.rs
```diff
@@ -0,0 +1,173 @@
...
+            model: f.model.map(StringFilter::from),
+            r#type: f.r#type.map(EqualFilter::from),
+            type_id: f.type_id.map(EqualFilter::from),
+        }
+    }
+}
+
+pub fn asset_catalogue_items(
+    ctx: &Context<'_>,
+    store_id: String,
```
jmbrunskill, line 74: I don't think we need `store_id`?

server/graphql/catalogue/src/asset_catalogue_item_queries.rs
```diff
@@ -0,0 +1,173 @@
...
+    page: Option<PaginationInput>,
+    filter: Option<AssetCatalogueItemFilterInput>,
+    sort: Option<Vec<AssetCatalogueItemSortInput>>,
+) -> Result<AssetCatalogueItemsResponse> {
+    validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::QueryAssetCatalogueItem,
+            store_id: Some(store_id.clone()),
```
jmbrunskill, line 83:
```suggestion
            store_id: None,
```

server/graphql/catalogue/src/asset_catalogue_item_queries.rs
```diff
@@ -0,0 +1,173 @@
...
+    .map_err(StandardGraphqlError::from_list_error)?;
+
+    Ok(AssetCatalogueItemsResponse::Response(
+        AssetCatalogueItemConnector::from_domain(items),
+    ))
+}
+
+pub fn asset_catalogue_item(
+    ctx: &Context<'_>,
+    store_id: String,
```
jmbrunskill, line 103: Same again here

server/graphql/catalogue/src/asset_category_queries.rs
```diff
@@ -0,0 +1,129 @@
...
+    auth::{Resource, ResourceAccessRequest},
+    catalogue::query_category::{get_asset_categories, get_asset_category},
+};
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq)]
+#[graphql(remote = "repository::asset_category::AssetCategorySortField")]
+#[graphql(rename_items = "camelCase")]
+
+pub enum AssetCategorySortFieldInput {
+    Name,
```
jmbrunskill, line 25: Might want some other options, but can always be added later...

server/graphql/catalogue/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "graphql_asset_catalogue"
```
mark-prins, line 2: perhaps the folder should be `asset_catalogue`?

server/graphql/catalogue/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+        sort: Option<Vec<AssetCategorySortInput>>,
+    ) -> Result<AssetCategoriesResponse> {
+        asset_categories(ctx, store_id, page, filter, sort)
+    }
+
+    pub async fn asset_category(
+        &self,
+        ctx: &Context<'_>,
+        store_id: String,
+        #[graphql(desc = "id of the class")] id: String,
```
jmbrunskill, line 79:
```suggestion
        #[graphql(desc = "id of the category")] id: String,
```
Although I think we could just skip these description? not critical...

server/graphql/catalogue/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+        sort: Option<Vec<AssetTypeSortInput>>,
+    ) -> Result<AssetTypesResponse> {
+        asset_types(ctx, store_id, page, filter, sort)
+    }
+
+    pub async fn asset_type(
+        &self,
+        ctx: &Context<'_>,
+        store_id: String,
+        #[graphql(desc = "id of the class")] id: String,
```
jmbrunskill, line 101:
```suggestion
        #[graphql(desc = "id of the type")] id: String,
```

mark-prins (commented):
I see that @jmbrunskill is reviewing.. I'll step away..

server/graphql/lib.rs
```diff
@@ -11,6 +11,7 @@ use async_graphql::http::{playground_source, GraphQLPlaygroundConfig};
 use async_graphql::{EmptyMutation, EmptySubscription, Schema};
 use async_graphql::{MergedObject, Response};
 use async_graphql_actix_web::{GraphQLRequest, GraphQLResponse};
+use graphql_asset_catalogue::AssetCatalogueItemQueries;
```
jmbrunskill, line 14:
```suggestion
use graphql_asset_catalogue::AssetCatalogueQueries;
```
I thinik `Item` is too specific?

server/graphql/types/src/types/asset_catalogue_item.rs
```diff
@@ -0,0 +1,84 @@
...
+    nodes: Vec<AssetCatalogueItemNode>,
+}
+
+#[Object]
+impl AssetCatalogueItemNode {
+    pub async fn id(&self) -> &str {
+        &self.row().id
+    }
+
+    pub async fn asset_category_id(&self) -> &str {
```
jmbrunskill, line 26: We might want loaders for `asset_category` etc...
@mark-prins do you need this? Good idea to do it in another PR/Issue though.

server/graphql/types/src/types/asset_catalogue_item.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub async fn code(&self) -> &str {
+        &self.row().code
+    }
+    pub async fn manufacturer(&self) -> Option<String> {
+        self.row().manufacturer.as_ref().map(|it| it.clone())
+    }
+    pub async fn model(&self) -> &str {
+        &self.row().model
+    }
+    pub async fn asset_type_id(&self) -> &str {
```
jmbrunskill, line 43: Might be nice to have this next to the other `asset_` fields? I thought it was missing at first...

server/graphql/catalogue/src/asset_catalogue_item_queries.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct AssetCatalogueItemFilterInput {
+    pub id: Option<EqualFilterStringInput>,
+    pub category: Option<StringFilterInput>,
+    pub category_id: Option<EqualFilterStringInput>,
+    pub class: Option<StringFilterInput>,
+    pub class_id: Option<EqualFilterStringInput>,
+    pub code: Option<StringFilterInput>,
+    pub manufacturer: Option<StringFilterInput>,
+    pub model: Option<StringFilterInput>,
+    pub r#type: Option<EqualFilterStringInput>,
```
jmbrunskill, line 51:
```suggestion
    pub r#type: Option<StringFilterInput>,
```
I think the idea here is to be able search for a partial type name, eg. `Freezer` which would find `Fridge/Freezer` as well as `Freezer` etc...

jmbrunskill: `type_id` is the exact match one

server/repository/src/db_diesel/assets/asset_catalogue_item.rs
```diff
@@ -31,7 +31,7 @@ pub struct AssetCatalogueItemFilter {
     pub code: Option<StringFilter>,
     pub manufacturer: Option<StringFilter>,
     pub model: Option<StringFilter>,
-    pub r#type: Option<StringFilter>,
+    pub r#type: Option<EqualFilter<String>>,
```
jmbrunskill, line 34: I think we should put this back too...
```suggestion
    pub r#type: Option<StringFilter>,
```

server/repository/src/db_diesel/assets/asset_category.rs
```diff
@@ -11,7 +11,11 @@ use crate::{
     DBType, EqualFilter, Pagination, Sort, StorageConnection, StringFilter,
 };
 
-type AssetCategory = AssetCategoryRow;
+#[derive(PartialEq, Debug)]
+
+pub struct AssetCategory {
+    pub asset_category_row: AssetCategoryRow,
```
jmbrunskill, line 17: Is this to make it easier to copy and paste? Doesn't seem necessary to me...

server/repository/src/db_diesel/assets/asset_type.rs
```diff
@@ -22,7 +26,7 @@ pub type AssetTypeSort = Sort<AssetTypeSortField>;
 #[derive(Clone)]
 pub struct AssetTypeFilter {
     pub id: Option<EqualFilter<String>>,
-    pub name: Option<StringFilter>,
```
jmbrunskill, line 25: Can't this stay as a `StringFilter`?

server/service/src/catalogue/validate_category.rs
jmbrunskill, line 1: I'd be tempted to remove these empty files, until we start allowing editing of this reference data

jmbrunskill (changes requested):
Looking for @fergie-nz 
Only issue seems to be some `StringFilter`s that got changed to `EqualFilter`s
Can you review? Doesn't seem right to me...

server/graphql/catalogue/Cargo.toml
jmbrunskill: I agree though :)

server/service/src/catalogue/query_category.rs
```diff
@@ -0,0 +1,35 @@
...
+    },
+    EqualFilter, PaginationOption, RepositoryError, StorageConnection,
+};
+
+use crate::{get_default_pagination, i64_to_u32, ListError, ListResult};
+
+pub const MAX_LIMIT: u32 = 1000;
+pub const MIN_LIMIT: u32 = 1;
+
+pub fn get_asset_categories(
```
jmbrunskill, line 13: It's a bit weird for me that we don't have a service provider and this doesn't take a service context.
I've had a discussion with andrei about it, we think it's probably ok and he thinks it has been done before.
Was there a particular thing you've copied the pattern from, or is this a new innovation?

jmbrunskill: We're all for making development faster and removing roadblocks if they're not helping!

fergie-nz: Good point!

I saw this in service/sr/temperature_log/mod.rs

The get_temperature_logs takes the connection rather than the context.
I saw that I was only using the connection anyway, so I just passed that value down from the graphql layer instead.

I also import the service layer functions directly rather than using the entire service provider.

This is just an arbitrary choice. I don't really understand the pros and cons of doing it this way.

Shall I refactor to the standard pattern?

jmbrunskill: I think it's ok for now, although maybe it's worth discussing a little bit more. 
The main reasons for the service provider are things like, being able to over-ride functionality for testing (which we very rarely do), often the service layer is where we put in things like extra filters (for example to add a store_id) or to do extra validation logic. If these things aren't needed it does feel like extra overhead. I'll discuss briefly today again before merging I think if we don't follow the pattern we should write up the option and some rules about when to do which pattern maybe?

jmbrunskill: He also mentioned that we want to have tests for our service layer, there's not too much to test here, but having some tests, maybe checking against some of the prepopulated data would be good.
Especially for the `catalogue_item_queries` for `r#type`,  `class`, and `category` filters as they're relying on lookups from different tables so have a bit of extra logic in them...

server/graphql/types/src/types/asset_type.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[Object]
+impl AssetTypeNode {
+    pub async fn id(&self) -> &str {
+        &self.row().id
+    }
+    pub async fn name(&self) -> &str {
+        &self.row().name
+    }
+    pub async fn catetory_id(&self) -> &str {
```
mark-prins, line 29:
```suggestion
    pub async fn category_id(&self) -> &str {
```

server/repository/src/db_diesel/assets/asset_catalogue_item.rs
```diff
@@ -173,6 +139,29 @@ fn create_filtered_query(filter: Option<AssetCatalogueItemFilter>) -> BoxedAsset
...
+            apply_string_filter!(sub_query, Some(r#type_filter), asset_type_dsl::name);
+            query = query.filter(asset_catalogue_item_dsl::asset_type_id.eq_any(sub_query));
+        }
+        if let Some(category_filter) = category {
+            let mut sub_query = asset_category_dsl::asset_category
+                .select(asset_category_dsl::id)
+                .into_boxed();
+            apply_string_filter!(sub_query, Some(category_filter), asset_category_dsl::name);
+            query = query.filter(asset_catalogue_item_dsl::asset_category_id.eq_any(sub_query));
+        }
```
fergie-nz, line 164: @mark-prins if you were curious, here is where we moved the new filters (to within create_filtered_query)

server/service/src/catalogue/tests/query_catalogue_item.rs
```diff
@@ -0,0 +1,309 @@
...
+                None,
+                Some(
+                    AssetCatalogueItemFilter::new()
+                        .r#type(StringFilter::equal_to("Icelined Refrigerator")),
+                ),
+                None,
+            )
+            .unwrap();
+
+        assert_eq!(result.count, 151);
```
jmbrunskill, line 149:
```suggestion
        assert_eq!(result.count, 15);
```

This is what I get running the tests, and appears right from looking at the data?

jmbrunskill (approved):
Looks good to me, thanks for adding all the tests!
Just one test seems to be incorrect?

PR #3069, opened 2024-02-26
3068 remove item id from stock out line inputs

### LINKED ISSUE #3068: remove item_id from stock_out_line inputs
## The suggestion

Remove item_id from stock_out_line inputs - item_id is a required field to insert/update StockOutLines, but appears to be surplus to requirements. `stock_line_id` is already a mandatory field in those inputs, which means item must exist, and item_id is present on stock_line, which we get during validation.

Let's remove the item_id field before we do more work with stock out lines.

lache-melvin:
Fixes #3068

# 👩🏻‍💻 What does this PR do? 
 

Removes the unnecessary item_id field from the insert/update stock_out_line inputs. 

Tests are still passing... code compiles... and still appears to work 😁 

I've removed the `ItemNotFound` and `ItemDoesNotMatchStockLine` errors, as I think now these errors aren't possible... can you please check this matches your understanding? 🙏


client/packages/common/src/types/schema.ts
```diff
@@ -633,7 +633,6 @@ export type DatabaseError = DeleteLocationErrorInterface & InsertLocationErrorIn
 
 export type DatabaseSettingsNode = {
   __typename: 'DatabaseSettingsNode';
-  /** Central server url */
```
lache-melvin, line 636: Not sure how this happened... maybe codegen didn't run since merging in develop?

andreievg (approved):
Thank you, looks great, and - 260 lines of code : )

PR #3077, opened 2024-02-26
3059 insert outbound return

### LINKED ISSUE #3059: implement insertOutboundReturn
Implement the backend for inserting an Outbound Return

- Can probably use insert stock out line?
- Can use outbound shipment insert for reference
- Note number series

lache-melvin:
Fixes #3059

She's a chunky one - sorry 😭 

# 👩🏻‍💻 What does this PR do? 
 

The creating an Outbound Return from an Inbound Shipment (or from the Outbound Return list view) now actually creates a Return! 

Also hooks up the return_reasons query to the return_reasons table 😁 

Automated tests added for the service layer.

To test in the UI, you'll need to populate yourself some return reasons. From wherever you like to run your SQL:

```sql
INSERT INTO return_reason (id, reason, is_active)
VALUES 
('2153c064-7921-41f8-9716-1b91693d1346', 'Damaged', TRUE),
('c81213b1-d28e-4a44-bddc-41a49ddb4117', 'Expired', TRUE),
('f2571df2-d605-4695-8ce5-4e7242ae1ab3', 'Wrong item', TRUE),
('38fe99d1-d84a-4f95-bd29-4938f4128246', 'Other', TRUE);
```

- Create a return from the Outbound Returns list view (this will only create the invoice, invoice lines would be added from the detail view in future)
- Create a return from an Inbound Shipment
- I've added a few more fields to the query used on the details page, so you should be able to see the key data in that object!

- I haven't added graphql layer test, as I'd heard from a couple of people that we were questioning the value of it ... should I still add it here?
- There's now an additional failing test, due to not yet implementing the sync translations.


client/packages/invoices/src/Returns/modals/OutboundReturn/useDraftOutboundReturnLines.ts
```diff
@@ -45,7 +45,13 @@ export const useDraftOutboundReturnLines = (
         const { id, reasonId, numberOfPacksToReturn, stockLineId, comment } =
           line;
 
-        return { id, stockLineId, reasonId, comment, numberOfPacksToReturn };
+        return {
+          id,
+          stockLineId,
+          reasonId: reasonId ?? '', // TODO
```
lache-melvin, line 51: `reasonId` and `comment` are both optional fields on the `GenerateOutboundReturnLines` response, because they're optional on the InvoiceLine.

However, they're not really optional when it comes to inserting a new OutboundReturnLine... you need to provide a reason to return. So I made the `InsertOutboundReturnLine` input require values here...

Do we usually handle this stuff on the frontend, and pass a definitely-defined value to the backend, or still just accept Option<String> and add new validation that these options aren't None?

server/graphql/types/src/types/invoice_line.rs
```diff
@@ -158,6 +158,9 @@ impl InvoiceLineNode {
     pub async fn note(&self) -> &Option<String> {
         &self.row().note
     }
+    pub async fn return_reason_id(&self) -> &Option<String> {
+        &self.row().return_reason_id
+    }
```
lache-melvin, line 163: it's probably worth adding a loader for the return reason here? we'll display the return reasons on the detail view right?

server/service/src/invoice/outbound_return/insert/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+        invoice_line::stock_out_line::InsertStockOutLineError,
+        service_provider::ServiceProvider,
+    };
+
+    use super::InsertOutboundReturnLine;
+
+    #[actix_rt::test]
+    async fn test_insert_outbound_return_errors() {
+        fn not_visible() -> NameRow {
+            inline_init(|r: &mut NameRow| {
```
lache-melvin, line 165: Saw the comment about preferring `..Default::default()` over `inline_init` now after writing these... should I update?

server/graphql/invoice/src/mutations/outbound_return/insert.rs
```diff
@@ -49,3 +76,70 @@ pub enum InsertErrorInterface {
...
+            reason_id,
+            comment,
+        }: OutboundReturnLineInput = self;
+
+        InsertOutboundReturnLine {
+            id,
+            stock_line_id,
+            number_of_packs: number_of_packs_to_return,
+            reason_id,
+            note: comment,
```
lache-melvin, line 142: I'm questioning this mapping - it will exposed at `note` on the InvoiceLineNode after the OutboundReturn invoice is created... should we just call it `note` everywhere in code, and can call it Comment in the UI if that makes more sense?

server/service/src/invoice/outbound_return/insert/mod.rs
```diff
@@ -0,0 +1,377 @@
...
+                )
+                .map_err(|error| OutError::LineInsertError {
+                    line_id: line.id.clone(),
+                    error,
+                })?;
+
+                invoice_line_repo
+                    .update_return_reason_id(&line.id, Some(line.reason_id.clone()))
+                    .map_err(|error| OutError::LineReturnReasonUpdateError {
+                        line_id: line.id.clone(),
```
lache-melvin, line 100: Should there be a validation step that checks that the return reason exists? Rather than relying on the foreign key constraint error?

client/packages/invoices/src/Returns/modals/OutboundReturn/useDraftOutboundReturnLines.ts
andreievg: Would it be simpler for now to not enforce and have as optional ?

client/packages/invoices/src/Returns/api/operations.graphql
```diff
@@ -120,6 +120,16 @@ query invoiceByNumber($invoiceNumber: Int!, $storeId: String!) {
       lines {
         nodes {
           id
+          returnReasonId
```
andreievg, line 123: hmm, just a side note, i was looking at other invoiceByNumber queries, and they look pretty identical, would be great to perhaps consolidate in the future, also they are called like outboundByNumber or inboundByNumber, but now we have outboundReturn and outboundShipment, and this one is called invoiceByNumber (i think good to rename to outboundReturnByNumber at some point and potentially outboundShipment by number)

client/packages/invoices/src/Returns/modals/OutboundReturn/useDraftOutboundReturnLines.ts
andreievg: In front and and back end ?

server/graphql/invoice/src/mutations/outbound_return/insert.rs
```diff
@@ -34,12 +42,31 @@ pub enum InsertResponse {
...
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let service_context = service_provider.context(store_id.to_string(), user.user_id)?;
+
+    let result = service_provider
+        .invoice_service
+        .insert_outbound_return(&service_context, input.to_domain());
+
+    match result {
```
andreievg, line 62: Sometimes verboseness of rust makes it a bit harder to digest the logic, 

I like to return result here and then Ok(result), rather then Ok() in each match. This is stylistic preference though, nothing prescriptive

server/graphql/invoice/src/mutations/outbound_return/insert.rs
```diff
@@ -49,3 +76,70 @@ pub enum InsertErrorInterface {
...
+            ))
+        }
+        ServiceError::OtherPartyNotASupplier => {
+            return Ok(InsertErrorInterface::OtherPartyNotASupplier(
+                OtherPartyNotASupplier,
+            ))
+        }
+
+        // Standard Graphql Errors
+        ServiceError::InvoiceAlreadyExists => BadUserInput(formatted_error),
```
andreievg, line 97: Another small one, using `|` is a bit tidier: `ServiceError::InvoiceAlreadyExists | ServiceError::OtherPartyDoesNotExist => BadUserInput(formatted_error)`, etc..

server/graphql/invoice/src/mutations/outbound_return/insert.rs
andreievg: yeah, for sure, great point !

client/packages/invoices/src/Returns/api/operations.graphql
CarlosNZ: I agree. I think consistent naming would allow us to more readily identify possible duplication.

lache-melvin: Ah yeah great point, perhaps can rename as part of the frontend work with the detail view. I think a naming review in a separate PR at the end might be a good idea, I think there are a couple of inconsistencies like that!

server/graphql/types/src/types/invoice_line.rs
andreievg: I think for now just in the next step of the line/item edit view. Loader + in detail view if we have time/budget, how does that sound ?

server/service/src/invoice/outbound_return/insert/mod.rs
andreievg: inline_init is fine still, the preference is ..Default::default() though, easier to understand and read

server/service/src/invoice/outbound_return/insert/mod.rs
andreievg: Yeah i think so, especially for the correct type of reason and is_active on a reason.

This is how I would do this one:

* In generate, output (new invoice, insert_stock_out_lines, update_line_reasons), where insert_stock_out_lines = Vec<InsertStockOutLine> and update_line_reasons = Vec<UpdateReturnLineReason>
* Add a service method that will be common for inbound/outbound return line updates, and use it here, new error type will be added UpdateReturnLineReasonError, and used similar to InsertStockOutLine.

Please let me know if that makes sense or not, I think it's a good idea to do in this PR

andreievg (changes requested):
Looks great overall, i do have one suggestion that will make follow up PR easier, and it's adding UpdateReturnLineReason helper, that's the only change request, thanks

server/graphql/types/src/types/invoice_line.rs
lache-melvin: Ah yep great 🙏

server/service/src/invoice/outbound_return/insert/generate.rs
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    let stock_out_lines = input
+        .outbound_return_lines
+        .iter()
+        .map(|line| InsertStockOutLine {
+            id: line.id.clone(),
+            invoice_id: outbound_return.id.clone(),
+            stock_line_id: line.stock_line_id.clone(),
+            number_of_packs: line.number_of_packs.clone(),
```
lache-melvin, line 64: Should we filter out lines with number of packs == 0 here??

server/service/src/invoice/outbound_return/insert/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+    },
+}
+
+type OutError = InsertOutboundReturnError;
+
+pub fn insert_outbound_return(
+    ctx: &ServiceContext,
+    input: InsertOutboundReturn,
+) -> Result<Invoice, OutError> {
+    let outbound_return: Invoice = ctx
```
andreievg, line 62: I like how this reads now, I think easier to see the overall intention at the top level, and then it's up to the reader if they want to get bogged down with more detail, in the section they are interested in

server/service/src/invoice_line/update_return_reason_id/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        )
+        .await;
+
+        let service_provider = ServiceProvider::new(connection_manager, "app_data");
+        let context = service_provider
+            .context(mock_store_b().id, "".to_string())
+            .unwrap();
+        let service = service_provider.invoice_line_service;
+
+        // LineDoesNotExist
```
andreievg, line 104: great to see these tests

andreievg (approved):
Thanks for changes @lache-melvin, looks great

PR #3080, opened 2024-02-26
Remove all om_* fields from transfered generated in mSupply

### LINKED ISSUE #2728: Mobile migration supplier invoice in outbound shipments, serial number 0
## What went wrong? 😲
 
Converting mobile site (with existing trasaction to oms-stores) to oms-site has some serious bugs right now: the SI type is changing to CI type in below instance

## Expected behaviour 🤔
Type of invoice should never change like that

## How to Reproduce 🔨

1. Have a sync setup with multiple sites: mobile and oms.
2. Create some transactions between mobile and oms (create some CI and SI) and sync. [I created some transaction to two different oms-sites]
Screenshot_20240115_125956_mSupply Mobile
Screenshot_20240115_130007_mSupply Mobile

In mobile, you may see [this similar error](https://github.com/msupply-foundation/msupply/issues/14090#issuecomment-1874916589) of Invoice No. 0 (I noticed SI inv. No. 0 in mine)
4. Now on cloud, reset hardware ID and password for that mobile-site so we'll switch it to oms.
5. In tablet (or desktop), initialise as oms site
6. Login and check the shipments list
Screenshot_20240115_130151_openmsupply-client
Screenshot_20240115_130145_openmsupply-client

7. See error: notice that SI (inv 0) is now showing up in Outbound Shipment (CI) list - HUGE BUG! 
wonder this will be the same in other instances as well - need to investigate!

## Your environment 🌱

 - Version: v1.5.03; mSupply Server: v7-12-beta11, mobile: v8.6.5
 - Platform:
   - [x] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

clemens-msupply:
Fixes #2728 

How to reproduce:
1) setup a mobile site, e.g Huff and sync with the central server (think this is needed otherwise the central server doesn't transfer invoices if the site is not old mSupply?)
2) setup an omSupply site e.g. Ravenclaw
3) send an outbound shipment from Ravenclaw to Huff (choose a unique reference to easily identify the shipment later). Set status to shipped. Sync.
4) stop the mobile site and "migrate" the Huff site to open mSupply (open it in omSupply)

Central should have a `transact` record now with type="si" and om_type="OUTBOUNT_SHIPMENT". This is because mSupply just copied the full outbound shipment for the transfer from Ravenclaw to Huff. This is a problem when integrating this record in omSupply because `om_type` is preferred over `type`, i.e. the incoming shipment get outbound status from the original outbound shipment!

# 👩🏻‍💻 What does this PR do? 

This PR detects this mismatch and removes all `om_*` from the incoming record. mSupply created the transfer and thus should not add any om_* columns

Follow steps above and then check that the Huff site shows the shipment as incoming.

### DIFF
--- a/server/service/src/sync/translations/invoice.rs
+++ b/server/service/src/sync/translations/invoice.rs
@@ -28,7 +28,7 @@ fn match_push_table(changelog: &ChangelogRow) -> bool {
     changelog.table_name == ChangelogTableName::Invoice
 }
 
-#[derive(Deserialize, Serialize, Debug)]
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
 pub enum LegacyTransactType {
     /// Supplier invoice
     #[serde(rename = "si")]
@@ -178,6 +178,36 @@ pub struct LegacyTransactRow {
     pub om_colour: Option<String>,
 }
 
+/// The mSupply central server will map outbound invoices from omSupply to "si" invoices for the
+/// receiving store.
+/// In the current version of mSupply all om_ fields get copied though.
+/// When receiving the transferred invoice on the omSupply store the "si" get translated to
+/// outbound shipments because the om_type will override to legacy type field.
+/// In other word, the inbound shipment will have an outbound type!
+/// This function detect this case and removes all om_* fields from the incoming record.
+fn sanitize_legacy_record(mut data: serde_json::Value) -> serde_json::Value {
+    let Some(Ok(om_type)) = data
+        .get("om_type")
+        .map(|value| serde_json::from_value::<InvoiceRowType>(value.clone()))
+    else {
+        return data;
+    };
+    let Some(Ok(legacy_type)) = data
+        .get("type")
+        .map(|value| serde_json::from_value::<LegacyTransactType>(value.clone()))
+    else {
+        return data;
+    };
+    if legacy_type == LegacyTransactType::Si && om_type == InvoiceRowType::OutboundShipment {
+        let Some(obj) = data.as_object_mut() else {
+            return data;
+        };
+        obj.retain(|key, _| !key.starts_with("om_"));
+    }
+
+    data
+}
+
 pub(crate) struct InvoiceTranslation {}
 impl SyncTranslation for InvoiceTranslation {
     fn pull_dependencies(&self) -> PullDependency {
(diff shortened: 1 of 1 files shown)


server/service/src/sync/translations/invoice.rs
```diff
@@ -178,6 +178,36 @@ pub struct LegacyTransactRow {
...
 
+/// The mSupply central server will map outbound invoices from omSupply to "si" invoices for the
+/// receiving store.
+/// In the current version of mSupply all om_ fields get copied though.
+/// When receiving the transferred invoice on the omSupply store the "si" get translated to
+/// outbound shipments because the om_type will override to legacy type field.
+/// In other word, the inbound shipment will have an outbound type!
+/// This function detect this case and removes all om_* fields from the incoming record.
+fn sanitize_legacy_record(mut data: serde_json::Value) -> serde_json::Value {
+    let Some(Ok(om_type)) = data
```
andreievg, line 189: Hmm, we already have deserialisation from json record to data, shouldn't we use it ? Rather then manually accessing serde fields by value ?

```rust
let data = sanitize_legay_record(serde_json::from_str(&sync_record.data)?);

// ...

fn sanitize_legay_record(data: LegacyTransactType) -> LegacyTransactType {
    if legacy_type == LegacyTransactType::Si && om_type == Some(InvoiceRowType::OutboundShipment) {
        LegacyTransacType {
            om_type: None,
            ..data
        }
    } else {
        data
    }
}
```

andreievg: Or is it because you want to sanitise all of the om_fields dynamically ?

clemens-msupply: Yes I thought about handling all fields case by case but this seemed to become quite messy and was not sure if this causes problem if we would introduce new om_* fields. For this reason I just thought I just remove all om_* fields from the raw json object and leave the rest of the code as it is. So yes its because I remove the fields dynamically.

andreievg (approved):
Thanks, looks good to me

server/service/src/sync/translations/invoice.rs
andreievg: Ok, sounds good, i agree, this dynamic wy is safer for future, and easier to see that it applies to all fields

Issue #2918, opened 2024-02-09
Currency input incorrect for price


roxy-dao:
## What went wrong? 😲
Typed in 50 for the price... but it gets converted to 5.00 when out of focus

https://github.com/msupply-foundation/open-msupply/assets/61820074/412d3193-0841-48c9-963b-920052aedd1d

Also typing 100.00 converts to 1.00

https://github.com/msupply-foundation/open-msupply/assets/61820074/2e118cb8-3de4-40ce-b17e-92791ff18719

## Expected behaviour 🤔
Should display/save the correct number

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to an 'Inbound Shipment invoice'
2. Click on 'New line'
3. Add any item
4. Go to price tab
5. Enter a 2 digit or more number without decimals

## Your environment 🌱

 - Version: 
 - Platform:
   - browser (extra points if you tell us which one)

CarlosNZ:
Can confirm that this problem is with the Currency Input component, as you can see the bug in Storybook as well.

Having a quick look, I see we're using an external package [react-currency-input-field](https://www.npmjs.com/package/react-currency-input-field) in this component -- just wondering if this is still needed? I think with the new standard "NumberInput" (#2794), we can handle the formatting ourselves (using `Intl.NumberFormat()`)? I'm happy to follow up on that when #2794 is merged if you like?

CarlosNZ:
This has also been listed here: https://github.com/msupply-foundation/open-msupply/issues/2932

My suggestion is what I wrote on that issue:

>  See if "react-currency-input-field" can be removed and just use the built-in IntlNumber.format() function to do what we want

I think if we just consolidate the currency input into the existing NumericText input (with additional formatting options), this problem will go away.

However, I won't do it in the Number input fixes, as that's more urgent and this will take a little more time, so I'll leave this issue here to fix that.

PR #3082, opened 2024-02-26
#2932 Number fixes

### LINKED ISSUE #2932: Remaining Number Input issues
EDIT: As well as the below problem, there are a few other implementations around where the logic of the parent components still needs updating to work with the new input component (from #2794).

Let's use this issue to list them all as we find them and I'll sort them out all at once. Please add to the checklist here, and put any further detail/screenshots etc in comments below. Thanks.

- [x] Problem clearing # packs (below) (something is turning it into `type="number"` from above which is odd, too)
- [x] Still have a couple of `type="number"` attributions in Programs that should be removed
- [x] [Inbound shipment pack quantity](https://github.com/msupply-foundation/open-msupply/issues/2932#issuecomment-1942859584) should just be a NumberDisplayCell (i.e. not editable)
- [x] [French currency formatting of thousands test failing](https://github.com/msupply-foundation/open-msupply/issues/2932#issuecomment-1944492789)
- [ ] See if "react-currency-input-field" can be removed and just use the built-in IntlNumber.format() function to do what we want
- [ ] [The separator for EU should be  but input always ends up adding , for the last two numbers even if you are not typing a decimal?](https://github.com/msupply-foundation/open-msupply/issues/2932#issuecomment-1963032660)
- [x] [Decimal doesn't work beyond 2 dp](https://github.com/msupply-foundation/open-msupply/issues/2932#issuecomment-1963048439)

## What went wrong? 😲
 

https://github.com/msupply-foundation/open-msupply/assets/55115239/40ac1ca7-4693-48ab-a706-834fbf2aa50d

I'm passing in props that this should be a required value with a minimum of 1, but clearing the input results in NaN, which I can then clear to nothing...

I think this is broadly an issue with the `NumberInputCell`, as I've seen this issue in at least Inbound Shipment, Stocktake and new Supplier Returns so far...

## Expected behaviour 🤔

Clearing the input sets the value to 0, or the minimum value if min>0.

## How to Reproduce 🔨

1. Go to an Inbound Shipment details page
2. Click on a line to edit
3. Clear the # packs input
4. See `NaN`
5. Clear the `NaN`
6. See nothing in the input

## Your environment 🌱

 - Version: 1.6.x (develop)
 - Platform: browser (Chrome)

CarlosNZ:
Fixes #2932 

# 👩🏻‍💻 What does this PR do? 
 

This fixes the items listed in #2932, except for two of them (unticked), but they're covered by [this issue's suggested solution](https://github.com/msupply-foundation/open-msupply/issues/2918#issuecomment-1965489335).

The main problem I discovered is that the `format` method (in NumUtils), which uses `Intl.NumberFormat`, defaults to 3 dp. So I've just made set that to default to the maximum allowed (21 dp) and we can continue using our `.round()` method to handle precision.

Tested the main cases listed in the issue. However I haven't properly tested the two Programs components affected. All I did there was remove the `type="number"` attribute, so I think it should be fine but if anyone has Programs set up on their system and can confirm it's okay, that'd be great.

### DIFF
--- a/client/packages/common/src/intl/number/IntlNumber.ts
+++ b/client/packages/common/src/intl/number/IntlNumber.ts
@@ -16,7 +16,6 @@ export const useFormatNumber = () => {
       if (value === undefined) return '';
       const locale = options?.locale ?? currentLanguage;
       return new Intl.NumberFormat(locale, {
-        // Don't round here, as most use cases have their own internal rounding
         maximumFractionDigits: 21, // maximum allowed
         ...options,
       }).format(value);
--- a/client/packages/common/src/intl/number/IntlNumber.ts
+++ b/client/packages/common/src/intl/number/IntlNumber.ts
@@ -15,7 +15,11 @@ export const useFormatNumber = () => {
     ) => {
       if (value === undefined) return '';
       const locale = options?.locale ?? currentLanguage;
-      return new Intl.NumberFormat(locale, options).format(value);
+      return new Intl.NumberFormat(locale, {
+        // Don't round here, as most use cases have their own internal rounding
+        maximumFractionDigits: 21, // maximum allowed
+        ...options,
+      }).format(value);
     },
     round: (value?: number, dp?: number): string => {
       const intl = new Intl.NumberFormat(currentLanguage, {
--- a/client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
@@ -14,7 +14,9 @@ export const NumberInputCell = <T extends RecordWithId>({
   rowIndex,
   columnIndex,
   isDisabled = false,
-  min,
+  // Make the default min=1 as this is the typical implementation
+  // in Data Tables
+  min = 1,
   max,
   decimalLimit,
   step,
@@ -44,13 +46,11 @@ export const NumberInputCell = <T extends RecordWithId>({
         ...TextInputProps?.InputProps,
       }}
       onChange={num => {
-        const newValue = num === undefined ? 0 : num;
+        const newValue = num === undefined ? min : num;
         setBuffer(newValue);
         updater({ ...rowData, [column.key]: Number(newValue) });
       }}
-      // Make the default min=1 as this is the typical implementation
-      // in Data Tables
-      min={min ?? 1}
+      min={min}
       max={max}
       decimalLimit={decimalLimit}
       step={step}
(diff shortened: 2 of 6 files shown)


client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.ts
```diff
@@ -205,7 +203,6 @@ export const useInboundShipmentColumns = () => {
       [
         'numberOfPacks',
         {
-          Cell: props => <NumberInputCell {...props} min={1} />,
           accessor: ({ rowData }) => {
```
CarlosNZ, line 206: Should have been a read-only cell, accidentally changed in last PR.

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
```diff
@@ -44,7 +44,7 @@ export const NumberInputCell = <T extends RecordWithId>({
         ...TextInputProps?.InputProps,
       }}
       onChange={num => {
-        const newValue = num;
+        const newValue = num === undefined ? 0 : num;
```
mark-prins, line 47: I would change to this - as it currently stands, you can clear the input and it gets set to `0`. then, you can press the down arrow to go lower than zero and the value changes to the min, which defaults to `1`. Setting the default to be the min feels more intuitive

```suggestion
        const newValue = num === undefined ? min : num;
```

that was along with 
line 17: `min = 1,` and move the comment from 53 to here as well
line 53: `min={min}`

client/packages/common/src/intl/number/IntlNumber.ts
```diff
@@ -15,7 +15,11 @@ export const useFormatNumber = () => {
     ) => {
       if (value === undefined) return '';
       const locale = options?.locale ?? currentLanguage;
-      return new Intl.NumberFormat(locale, options).format(value);
+      return new Intl.NumberFormat(locale, {
+        // Don't round here, as most use cases have their own internal rounding
```
mark-prins, line 19: I don't believe the comment is accurate!
Have looked for all usages of `format` and only the NumericTextInput is rounding, though not with options to the format (`setTextValue(format(constrained))`) 

The usages are in:
* NumericTextInput (this relies on downstream usage specifying dp which I think they all do)
* StockItemSearchInput
* StockItemSearchInputWithStats
* OutboundLineEditForm
* PrescriptionLineEditForm

all the others are using integers, so won't have an issue with the change.

Maybe just remove the comment. it only really makes sense in light of the change

mark-prins (approved):
👍  thanks @CarlosNZ looking good!

I have a request to change behaviour, see what you think. It's a bit odd with the clearing of a text value resetting to `0` when the min is higher than this.

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
CarlosNZ: Yes, makes sense. I'll update this.

client/packages/common/src/intl/number/IntlNumber.ts
CarlosNZ: Yeah, it made more sense when I was considering a few additional updates.

PR #3087, opened 2024-02-27
Enhance asset catalogue filtering

mark-prins:
Fixes #3083

# 👩🏻‍💻 What does this PR do? 
 
Adds the queries for categories, types and classes and uses those to populate the filters.
You can filter by all the available options now, with the types dependent on the category selected. 
Haven't implemented the classes yet, as we only have one and that's a bit silly.

Screenshot 2024-02-27 at 5 49 47 PM
Screenshot 2024-02-27 at 5 50 21 PM
Screenshot 2024-02-27 at 5 50 29 PM

# 🧪 How has/should this change been tested? 

## 📃 Documentation

- [ ] Describe the filtering

PR #3089, opened 2024-02-27
2956 asset catalogue item list view

### LINKED ISSUE #2956: Asset catalogue list view
Part of epic #2896 

List view for the asset catalogue.

1. Navigation: 
  a. Home > Menu **Catalogue** > Sub menu **Assets** > Asset catalogue list view
  b. … > Cold Chain Equipment Inventory > Detailed View > View Catalogue page
2. Only users with the right permission can view(?)/edit the catalogue
3. Headers:
  a. Filters (similar to what has been done for Cold Chain and patients)
  b. New Asset button: to add a new asset to the catalogue
  c. Export button (?)
  d. Select action dropdown list (+ list checkboxes):
    1. Delete selected lines
    2. Change status from active to inactive
4. List: see [spreadsheet](https://docs.google.com/spreadsheets/d/1l1x38KT5cdt4Zd5piMWp3FuBuiSUv4faMHP8utDJDqU/edit#gid=0&range=A1) for reference
5. Assets from the WHO PQS catalogue are available by default in the catalogue for the following categories:
  a. E003 / Refrigerators and freezers
  b. E004 / Cold boxes and vaccine carriers

Screenshot 2024-02-15 at 5 40 54 PM

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #2956 

# 👩🏻‍💻 What does this PR do? 
 
Creates the list view for the asset catalogue items

Screenshot 2024-02-27 at 5 49 47 PM

Note that the row click event has been disabled in the downstream (#3087) PR. I figure that if we aren't editing there's no need for that yet.

You should be able to view, sort and filter the catalogue list.
We need the loaders before all of the columns can be added in, so there's only code, manufacturer and model atm.

## 📃 Documentation

- [ ] An explanation of the list view


client/packages/system/src/Asset/ListView/Toolbar.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <FilterMenu
+          filters={[
+            {
+              type: 'enum',
+              name: t('label.category'),
+              urlParameter: 'category',
+              options: [
+                {
+                  label: 'Insulated Containers',
+                  value: 'aaa-bbb-ccc',
```
jmbrunskill, line 31: I assume you'll add queries to look up the proper values here? is that captured in another issue?

I'm getting a usablity issue, when I try to search in the category, type and class filters (I thought it would be an autocomplete, but it jumps to a new screen when I press `I`)
Has this been raised already?

Kapture 2024-02-28 at 17 13 58

client/packages/system/src/Asset/api/hooks/document/useAssets.ts
```diff
@@ -0,0 +1,18 @@
+import { useQuery, useUrlQueryParams } from '@openmsupply-client/common';
+import { useAssetApi } from '../utils/useAssetApi';
+
+export const useAssets = () => {
+  const { queryParams } = useUrlQueryParams({
```
jmbrunskill, line 5: Does code need to be added here?
I can't seem to search by code..

image

client/packages/system/src/Asset/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+    t('label.code'),
+    t('label.manufacturer'),
+    t('label.model'),
+  ];
+
+  const data = items.map(node => [
+    node.id,
+    node.code,
+    node.manufacturer,
+    node.model,
```
jmbrunskill, line 20: Maybe add a TODO to update to include type, class, etc after we complete https://github.com/msupply-foundation/open-msupply/issues/3103 ?

jmbrunskill (approved):
Looks good Mark,

Would be keen to see the type, class, category filters working with correct values, and the code search fixed before merging? Or at least a new issue created to follow up.

Would also be nice if trying to type `Insulated` in the filter didn't trigger navigation...

Some other more general thoughts...

To me it's annoying to have to clear filters if you select the wrong one, feels like there should be delete filter button next to each filter (not sure if that's been considered or discussed?)

I was wondering if navigation wise, it makes sense to put assets and asset catalogue under the same top level menu item. It looks like the mockup shows this approach though, I guess people shouldn't need to navigate to this screen often, would be more for updating the catalogue once that's available?

client/packages/system/src/Asset/ListView/Toolbar.tsx
mark-prins: it's not an autocomplete, just a SELECT. and yes, one of the testers raised the same issue - when the filters were created, the enum version uses select not autocomplete though

mark-prins:
> Would be keen to see the type, class, category filters working with correct values, and the code search fixed before merging? Or at least a new issue created to follow up.

they are working, along with lookups of the reference data. the second PR (#3087) addressed that. unfortunately the branch shenanigans yesterday resulted in me deleting the target of that PR, which closed it 🙄 I've made a new one: #3105 

> To me it's annoying to have to clear filters if you select the wrong one, feels like there should be delete filter button next to each filter (not sure if that's been considered or discussed?)

yep. mentioned by most people who have used the filters! One of the original implementations had a clear adornment, but it wasn't obvious whether you were clearing the text in the filter or removing the filter. it was too clunky to have both clear, and remove buttons in the filter input, so we removed the remove.
Anisha had a great suggestion of adding checkboxes beside the filter in the filters drop down which might work.

> I was wondering if navigation wise, it makes sense to put assets and asset catalogue under the same top level menu item. It looks like the mockup shows this approach though, I guess people shouldn't need to navigate to this screen often, would be more for updating the catalogue once that's available?

They shouldn't be. Asset catalogue items are under the `Catalogue` menu item and the CCE is labelled `Equipment` and under `Cold Chain`. At this stage we are only implementing cold chain equipment - and when (if) we do assets they won't be under Catalogue

PR #3091, opened 2024-02-27
3084 Currency display should be dependant on language

### LINKED ISSUE #3084: Currency display should be dependent on language
As discussed with @CarlosNZ. The currency display should be dependant on language. I.e.:
in English we would write:
$1,250.95
€1,250.95
And in French they'd write:
1 250,95$
1 250,95€
If users are using issuing in foreign currency and see a format different from their own, it could lead to confusion.

Link currency display with locale.

roxy-dao:
Fixes #3084

# 👩🏻‍💻 What does this PR do? 
Displays currency in locale format. 
Screenshot 2024-02-28 at 10 35 23
Screenshot 2024-02-28 at 10 35 42

- mSupply
- [ ] Go to Currencies 
- [ ] Add some and update rates
- [ ] Go to Stores -> Preferences
- [ ] Enable `Store: Able to issue in foreign currency`
- om
- [ ] Go to Inbound Shipment
- [ ] Create an invoice with external Supplier
- [ ] Add some lines with prices
- [ ] Click on pen next to `Foreign Currency` in `More` panel
- [ ] Change to a currency that uses a different format from Eng e.g. `EUR`
- [ ] See that the format is still in one that Eng uses instead of the one Europe uses
- [ ] Change language to `French`
- [ ] See format change

### DIFF
--- a/client/packages/common/src/intl/currency/currency.test.tsx
+++ b/client/packages/common/src/intl/currency/currency.test.tsx
@@ -41,19 +41,19 @@ describe('currency formatting - en', () => {
   });
 });
 
-describe('currency formatting - fr', () => {
+describe('currency formatting - eur for en', () => {
   it('formats a string with up to the precision number of decimal places, dropping decimal places where needed', () => {
     const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c(1.11111111111).format();
-    expect(f1).toBe('1,11 €');
+    expect(f1).toBe('1.11 €');
   });
 
   it('formats a string with up to the precision number of decimal places, dropping decimal places where needed even with large numbers', () => {
     const { result } = renderHookWithProvider(() => useCurrency('EUR'));
 
     const f1 = result.current.c(111_111.11111111111).format();
-    expect(f1).toBe('111.111,11 €');
+    expect(f1).toBe('111,111.11 €');
   });
 
   it('does drop trailing zeroes', () => {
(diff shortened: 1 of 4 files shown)


mark-prins (approved):
hm. I see now - the currency separator and decimal indicator are defined by the locale, but the format and currency symbol are tied to the currency. So it's a problem with the currency formatting changes and not this PR - in which case, I think the changes here are good (thanks!) but I'd raise another issue to improve the currency formatting.

roxy-dao:
> hm. I see now - the currency separator and decimal indicator are defined by the locale, but the format and currency symbol are tied to the currency. So it's a problem with the currency formatting changes and not this PR - in which case, I think the changes here are good (thanks!) but I'd raise another issue to improve the currency formatting.

Was trying to find a way to do this easily but seems like methods to get the symbol format are all tied to currency 😢

mark-prins:
I wondered if `getSeparatorAndDecimal` could be renamed and changed to return the pattern as well as the separator and decimal? that is passed the locale, so it would mean that all formatting would be dependent on the locale.

or this?

```

const getPatterns = (locale: string) => {
  switch (locale) {
    case 'fr-DJ':
    case 'fr':
    case 'ru':
      return { pattern: '# !', negativePattern: '-# !' };
    default:
      return { pattern: '!#', negativePattern: '-!#' };
  }
};

...

const currencyOptions = {
  en: {
    symbol: '$',
    // separator: "," decimal = "."
    ...getSeparatorAndDecimal('en'),
    ...getPatterns('en'),
    precision: 2,
    format,
  },
  fr: {
    symbol: '€',
    // separator: " " decimal = ","
    ...getSeparatorAndDecimal('fr'),
    ...getPatterns('fr'),
    precision: 2,
    format,
  },
etc
```

roxy-dao:
Yeah think it would be in a getPatterns function... didn't see anything in Intls.NumberFormat that would return pattern based on locale. Thanks Mark~

PR #3095, opened 2024-02-27
2967 Create, Update, Delete and & Query Assets

### LINKED ISSUE #2967: Asset Mutations: create, update, delete
Part of epic #2896 

### Moneyworks Jobcode 🧰

`OMSCC:AST`

jmbrunskill:
Fixes #2967, #2974

# 👩🏻‍💻 What does this PR do? 
Adds service and graphql for `Asset` entity.

Currently assets can be created without a store, but we're passing the store in the mutations as we need to check permissions against a store currently. 

Would be good to test business logic, when that's clearer.

Currently enforces that serial numbers are unique if they're added - Not sure if we want this...
Code is also considered unique - Possibly should be generated from the backend for the barcode?

Haven't created `Log` entries for updates/creates - We probably want this right?


server/graphql/asset/src/types.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    pub async fn code(&self) -> &str {
+        &self.row().code
+    }
+
+    pub async fn serial_number(&self) -> &Option<String> {
+        &self.row().serial_number
+    }
+
+    // TODO: Loaders for store, class, category, type, catalogue_item
```
jmbrunskill, line 94: @mark-prins do you need this for frontend? Should we create a new issue for it in github?

server/service/src/asset/insert.rs
```diff
@@ -0,0 +1,112 @@
...
+    if let Some(serial_number) = &input.serial_number {
+        if AssetRepository::new(connection)
+            .query_one(AssetFilter::new().serial_number(StringFilter::equal_to(serial_number)))?
+            .is_some()
+        {
+            return Err(InsertAssetError::SerialNumberAlreadyExists);
+        }
+    }
+
+    // TODO: Do we require Name, and or Code? I'm assuming user can leave them as empty strings if they want?
```
jmbrunskill, line 66: Question for the reviewer... Any other validations needed?

server/service/src/asset/update.rs
```diff
@@ -0,0 +1,139 @@
...
+                        .serial_number(StringFilter::equal_to(serial_number)),
+                )?
+                .is_some()
+            {
+                return Err(UpdateAssetError::SerialNumberAlreadyExists);
+            }
+        }
+    }
+    if let Some(store_id) = &asset_row.store_id {
+        // TODO: confirm, maybe people can just create them on central for any store
```
jmbrunskill, line 76: Question, I've seen this pattern in things I'm copying, but do we want to remove this restriction? What happens for non-store items, do we need to check we are on central?

server/graphql/asset/src/mutations/insert.rs
```diff
@@ -0,0 +1,214 @@
...
+        Err(error) => Ok(InsertAssetResponse::Error(InsertAssetError {
+            error: map_error(error)?,
+        })),
+    }
+}
+
+#[derive(InputObject)]
+pub struct InsertAssetInput {
+    pub id: String,
+    pub store_id: Option<String>,
```
fergie-nz, line 47: I am wondering if you need store_id here given it will always be provided with the user?

Or maybe you do need it because a user from a certain store might want to add an asset to another store?

I'm not sure, just found it strange seeing two iterations of it on insert_asset in graphiql:
Screenshot 2024-02-28 at 2 02 01 PM

server/service/src/asset/insert.rs
```diff
@@ -0,0 +1,112 @@
...
+        code,
+        serial_number,
+        catalogue_item_id,
+        installation_date,
+        replacement_date,
+        created_datetime: Utc::now().naive_utc(),
+        modified_datetime: Utc::now().naive_utc(),
+        deleted_datetime: None,
+    }
+}
```
fergie-nz, line 96: I think asset_category_id, asset_type_id, and asset_catalogue_item_id are missing from AssetRow type and this generate function (or somewhere else in service layer?)

I get foreign key violation errors when I attempt to insert any asset.

fergie-nz (changes requested):
Some issue with foreign id constraints but otherwise looks great!

server/graphql/asset/src/types.rs
mark-prins: Yes please!
I've just added an issue (#3103) for class, category and type. There is already a store loader - can we add `catalogue_item` to the issue?

server/service/src/asset/insert.rs
mark-prins: Agree - the user can leave them as empty strings. It would be nice to have names in the list, I don't think we need to require them however

server/service/src/asset/update.rs
mark-prins: given how we are using store as a proxy for 'where is the asset located?' I think remove the restriction. that was assets can be created centrally, and not assigned to a store. Of course we could go both ways - require store if you are on a remote site, and allow no store if on central. easier to just remove the restriction

server/service/src/asset/insert.rs
jmbrunskill: I've removed them from `AssetRow` entirely after a discussion with Mark Prins, however I see there's still included in the migration I'll fix that. The idea is that we just need the `catalogue_item_id` link and we can look up that other related data.
You may need to re-initialise you database to get rid of those columns, or delete them manually though...

server/graphql/asset/src/mutations/insert.rs
jmbrunskill: Keen for @mark-prins opinion here. I agree if feels a bit weird!
I've done this because in theory according to the design docs you can create a asset that doesn't have a store assigned. So the on that's part of the input can be null, but you still need to send you logged in store so we can check permissions. I think it's probably ok how it is but probably needs some comments to explain...

server/service/src/asset/tests/query.rs
```diff
@@ -0,0 +1,150 @@
...
+                ))),
+                Some(AssetSort {
+                    key: AssetSortField::Name,
+                    desc: Some(true),
+                }),
+            )
+            .unwrap();
+
+        assert_eq!(result.count, 2);
+        assert_eq!(result.rows[0].id, mock_asset_a().id);
```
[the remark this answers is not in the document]
jmbrunskill: Thanks, you're right already updated :)

server/graphql/asset/src/mutations/insert.rs
mark-prins: agree - weird looking but it makes some kind of sense. I'd keep it, and comment the weirdness

fergie-nz (approved):
Nice James, looks great! 🎸

PR #3096, opened 2024-02-27
3052 Cannot choose the time correctly in date-time picker

### LINKED ISSUE #3052: Cannot choose the time correctly in date-time picker
## What went wrong? 😲
No matter what time we select, it automatically changes to the current system time

## Expected behaviour 🤔
The value we select should be prioritised?

## How to Reproduce 🔨
1. Go to patient window and create an encounter
2. While choosing the visit date, select some past date and choose any random timing
3. See error: the time automatically changes back to current system time while the date is the one we selected. I think both date & time should be the one we select

## Screenshots
https://recordit.co/Wrxvw85XW2

## Additional
Found while testing #2415 

## Your environment 🌱

- Version: v1.6.2
- mSupply version: 713-00-beta3 
- Platform:
  - [ ] android (tablet)
  - [x] browser (extra points if you tell us which one)
  - [ ] desktop (windows)
  - [ ] desktop (macOS)
  - [x] server (windows)
- Database type: 
  - [ ] PostgreSQL
  - [x] SQLite3

roxy-dao:
Fixes #3052

# 👩🏻‍💻 What does this PR do? 
Make sure that the end time for encounters

- [ ] Set up programs
- [ ] Enrol a patient in a program
- [ ] Create an encounter from that program
- [ ] Change date to the past
- [ ] Try change end time
- [ ] The date shouldn't change to today

Didn't do it in this PR since this is just a bug fix, but I think there should be some validations in place for end time so that the user doesn't choose a time that is before the start time.

### DIFF
--- a/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
+++ b/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
@@ -278,7 +278,7 @@ export const Toolbar: FC<ToolbarProps> = ({
                         : undefined;
                       if (endDatetime) {
                         setEndDatetime(endDatetime);
-                        onChange({ startDatetime, endDatetime });
+                        onChange({ endDatetime });
                       }
                     }}
                   />
--- a/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
+++ b/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
@@ -117,7 +117,7 @@ export const Toolbar: FC<ToolbarProps> = ({
       ),
       value: encounter.clinician as Clinician,
     });
-  }, [encounter]);
+  }, [encounter, getLocalisedFullName]);
 
   const getDeleteConfirmation = useConfirmationModal({
     message: t('message.confirm-delete-encounter'),
--- a/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
+++ b/client/packages/system/src/Encounter/DetailView/Toolbar.tsx
@@ -286,6 +286,9 @@ export const Toolbar: FC<ToolbarProps> = ({
                 labelWidth="60px"
                 Input={
                   <TimePickerInput
+                    minTime={
+                      startDatetime ? new Date(startDatetime) : undefined
+                    }
                     value={DateUtils.getDateOrNull(endDatetime ?? null)}
                     onChange={date => {
                       const endDatetime = date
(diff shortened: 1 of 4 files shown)


clemens-msupply (commented):
The issue was talking in about the CreateEncounterModal but this change is for the toolbar for existing encounters, right?
 
However, I can't see a problem with the existing code, could you please clarify what the problem was?

client/packages/system/src/Encounter/DetailView/Toolbar.tsx
```diff
@@ -267,11 +267,18 @@ export const Toolbar: FC<ToolbarProps> = ({
...
+                                date.getHours(),
+                                date.getMinutes()
+                              )
+                            )
+                          )
                         : undefined;
                       if (endDatetime) {
                         setEndDatetime(endDatetime);
-                        onChange({ endDatetime });
+                        onChange({ startDatetime, endDatetime });
```
clemens-msupply, line 281: Is this startDatetime needed, seems to work without?

roxy-dao: Yeah not need since it's getting the correct date now

clemens-msupply (changes requested):
Sorry for adding more things to this issue, but there are some other bugs:
In the toolbar:
1) when entering the date manually and pressing tab the date input just clears
2) When having first set the endtime and then changing the date the start time and the end time are on different days

client/packages/system/src/Encounter/DetailView/Toolbar.tsx
```diff
@@ -230,9 +230,19 @@ export const Toolbar: FC<ToolbarProps> = ({
                     onChange={date => {
                       const startDatetime = DateUtils.formatRFC3339(date);
                       setStartDatetime(startDatetime);
+                      const endDt = DateUtils.getDateOrNull(endDatetime);
                       onChange({
                         startDatetime,
-                        endDatetime: endDatetime ?? undefined,
+                        endDatetime: endDatetime
+                          ? DateUtils.formatRFC3339(
```
clemens-msupply, line 237: Nit: refactor this into a `endTimeFromStartDate(startDate: string | undefined, endTime: Date)`?

Also check for endDt `endDt ? DateUtils...` then you don't need the null check later...

roxy-dao:
> Note, its still possible to set an end time smaller then the start time but I leave this for another time... Could you please review these changes?

Yes I mentioned it above.

roxy-dao:
Thanks! Changes look good to me but you'll need to approve since I can't approve my own PR XD

clemens-msupply (approved):

PR #3114, opened 2024-02-28
3081 update outbound return lines

### LINKED ISSUE #3081: implement update outbound return lines API

Implement API side of updating an outbound return - we don't currently have a UI connected to this...

This one will use combination of stock_out_line/delete + update + insert, we don't need item_id at all, please delete (when input numberOfPacks = 0 then delete line)

lache-melvin:
Fixes #3081 
# 👩🏻‍💻 What does this PR do? 
 

Adds updateOutboundReturn mutation, allowing adding/updating/deleting outboundReturnLines for an existing return.

Automated tests added for the service layer. Can test the mutation in a graphQL playground.


server/graphql/invoice/src/mutations/outbound_return/update.rs
```diff
@@ -0,0 +1,121 @@
...
+    invoice::outbound_return::update::UpdateOutboundReturnStatus,
+};
+
+use super::insert::OutboundReturnLineInput;
+
+#[derive(InputObject)]
+#[graphql(name = "UpdateOutboundReturnInput")]
+pub struct UpdateInput {
+    pub id: String,
+    // supplier_id: String,
```
lache-melvin, line 22: Will we want to be able to change who we're returning the goods to?

server/graphql/invoice/src/mutations/outbound_return/update.rs
```diff
@@ -0,0 +1,121 @@
...
+    status: Option<UpdateOutboundReturnStatusInput>,
+    outbound_return_lines: Vec<OutboundReturnLineInput>,
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum UpdateOutboundReturnStatusInput {
+    Allocated,
```
lache-melvin, line 29: Copied the same statuses as outbound shipment, but is Allocated a relevant status for an outbound return? You're returning specific stock, so the idea of a placeholder line which you then allocate seems irrelevant right?

server/graphql/invoice/src/mutations/outbound_return/update.rs
```diff
@@ -0,0 +1,121 @@
...
+        //     return Ok(UpdateErrorInterface::OtherPartyNotVisible(
+        //         OtherPartyNotVisible,
+        //     ))
+        // }
+
+        // Standard Graphql Errors
+        ServiceError::NotAnOutboundReturn
+        | ServiceError::ReturnDoesNotBelongToCurrentStore
+        | ServiceError::ReturnIsNotEditable
+        | ServiceError::ReturnDoesNotExist => BadUserInput(formatted_error),
```
lache-melvin, line 81: Are any of these actually structured errors? I've seen ones like these as both structured and standard, but going on the "can the frontend prevent this" I think mostly these would be considered standard errors?

Except perhaps `ReturnIsNotEditable`... someone else could update the status while we're editing? Oh and then perhaps same goes for `ReturnDoesNotExist` - they could have deleted it?

server/graphql/invoice/src/mutations/outbound_return/update_lines.rs
```diff
@@ -0,0 +1,96 @@
...
+        //     return Ok(UpdateErrorInterface::OtherPartyNotVisible(
+        //         OtherPartyNotVisible,
+        //     ))
+        // }
+
+        // Standard Graphql Errors
+        ServiceError::NotAnOutboundReturn
+        | ServiceError::ReturnDoesNotBelongToCurrentStore
+        | ServiceError::ReturnIsNotEditable
+        | ServiceError::ReturnDoesNotExist => BadUserInput(formatted_error),
```
lache-melvin, line 69: Are any of these actually structured errors? I've seen ones like these as both structured and standard, but going on the "can the frontend prevent this" I think mostly these would be considered standard errors?

Except perhaps `ReturnIsNotEditable`... someone else could update the status while we're editing? Oh and then perhaps same goes for `ReturnDoesNotExist` - they could have deleted it?

server/service/src/invoice/outbound_return/update_lines/validate.rs
```diff
@@ -0,0 +1,29 @@
+use repository::{InvoiceRowType, StorageConnection};
+
+use crate::invoice::{
+    check_invoice_exists, check_invoice_is_editable, check_invoice_type, check_store,
+};
+
+use super::UpdateOutboundReturnLinesError;
+
+pub fn validate(
```
lache-melvin, line 9: These checks are actually done for each line in the stockOutLine insert/update/delete validations... not sure how much value it adds to do it at the outer layer, except maybe it's nice to have a direct error, rather than these bubbling back from one of the line mutations??

server/service/src/invoice/outbound_return/update_lines/generate.rs
```diff
@@ -0,0 +1,92 @@
...
+) -> Result<GenerateResult, RepositoryError> {
+    let existing_lines = InvoiceLineRepository::new(connection).query_by_filter(
+        InvoiceLineFilter::new().invoice_id(EqualFilter::equal_to(&outbound_return_id)),
+    )?;
+    let check_already_exists = |id: &str| {
+        existing_lines
+            .iter()
+            .find(|line| line.invoice_line_row.id == id)
+            .is_some()
+    };
```
lache-melvin, line 34: Should i pull this into a separate fn? Only wanted to do the query once, rather than for every line, but this doesn't really tie in with existing patterns...

server/service/src/invoice/outbound_return/update_lines/validate.rs
mark-prins: Not seeing what you mean re running for each line. 
It's this isn't it?
```
    let outbound_return = ctx
        .connection
            validate(connection, &ctx.store_id, &input.outbound_return_id)?;
```

mark-prins (approved):
👍 LGTM!

server/service/src/invoice/outbound_return/update_lines/validate.rs
lache-melvin: Oh, just that the invoice exists/is_editable etc. checks are also run in the insert/update_stock_out_line validation methods, so we're kind of doubling up on those checks. However I think it is best to keep them in both 👍

PR #3122, opened 2024-02-28
#3011 Outbound returns data table

### LINKED ISSUE #3011: Outbound returns detail view -- Table contents

Add DataTable

CarlosNZ:
Fixes #3011

# 👩🏻‍💻 What does this PR do? 
 

Very rudimentary data table display. I only did what I could do without making new queries (which seems unnecessary?). See inline notes.

Uses mock query to fetch return invoice, displays lines as table rows

Lots of commented-out code (from Outbound Shipment) due to reduced functionality. Not sure what we'll need to keep, maybe some of it can go completely?

Toolbar coming up next?


client/packages/invoices/src/Returns/DetailView/DetailView.tsx
```diff
@@ -32,34 +32,39 @@ export const DetailView: FC = () => {
...
-  //     value: 'Log',
-  //   },
-  // ];
+  const tabs = [
+    {
+      Component: (
+        <ContentArea
+          onRowClick={onRowClick}
+          onAddItem={onAddItem}
+          rows={data?.lines?.nodes}
```
CarlosNZ, line 58: I don't really understand why Typescript is complaining about `lines` not existing -- you can clearly see it exists on the type when hovering over `data`:
Screenshot 2024-02-29 at 12 00 02 PM

client/packages/invoices/src/Returns/DetailView/columns.ts
```diff
@@ -0,0 +1,218 @@
...
+      //             .filter(({ body }) => !!body);
+      //           return noteSections.length ? noteSections : null;
+      //         } else {
+      //           return rowData.batch && rowData.note
+      //             ? { header: rowData.batch, body: rowData.note }
+      //             : null;
+      //         }
+      //       },
+      //     },
+      //   ],
```
CarlosNZ, line 82: Have commented out the columns not currently available on the mock query response.

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
CarlosNZ, line 58: This is not how it's done elsewhere. In Outbound Shipments, rows is fetched in `ContentArea` using an additional hook `useOutbound.line.rows`, but I don't really get what that hook is doing (it seems to be just re-packaging the document query in a different way, but it's really confusing). Hence I'm just passing the columns in for now.  Suggestions welcome.

client/packages/invoices/src/Returns/DetailView/columns.ts
lache-melvin: Nice one. 

The insert mutation is now merged into the feature branch, so once you pull that in, you could remove the mock API response and query this extra stuff. Maybe in the next PR?

client/packages/invoices/src/Returns/DetailView/ContentArea.tsx
```diff
@@ -0,0 +1,126 @@
...
+//   const expandoColumns = useExpansionColumns();
+
+//   if ('lines' in rowData && rowData.lines.length > 1) {
+//     return <MiniTable rows={rowData.lines} columns={expandoColumns} />;
+//   } else {
+//     return null;
+//   }
+// };
+
+// const useHighlightPlaceholderRows = (
```
lache-melvin, line 36: I think we can remove the placeholder row stuff - we're returning specific stock, so shouldn't be possible to have unallocated return rows.

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
```diff
@@ -88,7 +92,7 @@ export const DetailView: FC = () => {
           )} */}
 
           {/* <Toolbar /> */}
-          {/* <DetailTabs tabs={tabs} /> */}
+          <DetailTabs tabs={tabs} />
           {/* <Footer /> */}
           {/* <SidePanel /> */}
         </TableProvider>
```
lache-melvin, line 98: haha I can't highlight it to put a comment there, but should also change the translation strings below `shipment` -> `return`

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
```diff
@@ -76,7 +81,6 @@ export const DetailView: FC = () => {
             },
```
lache-melvin, line 81: StockOut -> OutboundReturn above?

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
lache-melvin: I think how you have it is great to get the table in for now. 

At some point we'll need to bring in grouping by item, and then I think the rows hook will be helpful. It abstracts away the logic to return all the lines, or the item ones, which would really bloat the component if done there.

lache-melvin: Ah this is due to the query actually returning either an `InvoiceNode` or a `NodeError` (it's a bit frustrating becuase VSCode hides the `NodeError` part behind a `...` 😅) 

If you take a look in the `packages/invoices/src/Returns/api/api.ts`, you'll see some of the other methods have a type check step `if (__typename === 'InvoiceNode') return ...`, otherwise throwing an error - if you chuck one of these in the `outboundReturnByNumber` method Typescript should be satisfied!

lache-melvin (approved):

lache-melvin (approved):
Awesome Carl! Have left a few small comments but great to start getting this stuff in! Yep I think Toolbar next sounds like a good idea!

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
CarlosNZ: Ah yes, I thought it would be something to do with the union type, didn't know where to put the check. Thanks 👍

lache-melvin (approved):
Looks great! Just a couple little things... merge when ready 🚀

client/packages/invoices/src/Returns/DetailView/DetailView.tsx
lache-melvin: Oh actually I'm always getting this error modal, even when the return does exist! I think it's because you have the display logic `data ? <>[table]</> : <AlertModal />`, and on the initial render data is undefined... probably since you added the type filter thing in the api file?

lache-melvin (changes requested):

CarlosNZ:
@lache-melvin Sorry, missed a couple of things on that last pass. Should be all good now.

lache-melvin (approved):

lache-melvin (approved):
Awesome thanks for changes Carl, looks great!

PR #3130, opened 2024-02-29
3121 update outbound return

### LINKED ISSUE #3121: implement updateSupplierReturn mutation
Update status, comment

Similar logic to outbound_shipment update (in terms of stock movements)

Maybe leave placeholder for other invoice fields, like colour, supplier id etc.

lache-melvin:
Fixes #3121

# 👩🏻‍💻 What does this PR do? 
 

Implements API for updating an outbound return, updating the stock line total when the return status is updated to Shipped.

Automated tests added in the service layer.

Not currently connected to the UI, mutation can be tested with a GraphQL playground.

Have left some questions about other stuff to capture?


server/graphql/invoice/src/mutations/outbound_return/update.rs
```diff
@@ -0,0 +1,114 @@
...
+use service::{
+    auth::{Resource, ResourceAccessRequest},
+    invoice::outbound_return::update::UpdateOutboundReturnStatus,
+};
+
+#[derive(InputObject)]
+#[graphql(name = "UpdateOutboundReturnInput")]
+pub struct UpdateInput {
+    pub outbound_return_id: String,
+    // supplier_id: String,
```
lache-melvin, line 20: Do we want to support being able to change the supplier we are returning to?

server/service/src/invoice/outbound_return/update/generate.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+pub fn generate(
+    connection: &StorageConnection,
+    input: UpdateOutboundReturn,
+    existing_return: InvoiceRow,
+) -> Result<GenerateResult, UpdateOutboundReturnError> {
+    let mut updated_return = existing_return.clone();
+
+    updated_return.comment = input.comment.or(existing_return.comment.clone());
```
lache-melvin, line 22: Should we also add updating `colour` here, or do that later?

server/service/src/invoice/outbound_return/update/mod.rs
```diff
@@ -0,0 +1,331 @@
...
+pub mod validate;
+use generate::generate;
+use validate::validate;
+
+use self::generate::GenerateResult;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum UpdateOutboundReturnStatus {
+    Picked,
+    Shipped,
```
lache-melvin, line 21: Have removed `Allocated` as a status here compared to outbound shipment, as when we generate the return lines, we always include a stock line id (i.e. we won't have placeholder lines with returns) - just want to double check my understanding is correct there?

lache-melvin:
Oh no sorry I just realised that I've merged another PR into this! The frontend stuff has already been reviewed, this is just for the backend 😅

server/service/src/invoice/outbound_return/update/generate.rs
jmbrunskill: Not sure about these, maybe create a follow up issue? I don't know how the on_hold status is applied for other invoices, I'd assume we'd follow a similar pattern unless returns shouldn't ever go `on_hold`?

Colour sounds interesting, is it being set anywhere else in the code base? or just a legacy mSupply thing?

jmbrunskill (approved):
Looks good to me.

My server keeps crashing due to  
server/service/src/sync/translations/invoice.rs:to_legacy_confirm_time
```
        InvoiceRowType::InboundReturn => todo!(),
        InvoiceRowType::OutboundReturn => todo!(),
```
So a bit hard to test out...

server/service/src/invoice/outbound_return/update/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+            .unwrap();
+
+        assert_eq!(result.invoice_row.status, InvoiceRowStatus::Shipped);
+        assert!(result.invoice_row.shipped_datetime.is_some());
+
+        let updated_stock_line = stock_line_row_repo.find_one_by_id(&stock_line_id).unwrap();
+
+        assert_eq!(
+            updated_stock_line.total_number_of_packs,
+            original_stock_line.total_number_of_packs // total has not changed (stock would have already been reduced New -> Picked)
```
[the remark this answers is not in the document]
lache-melvin: Ah, new -> shipped is the test above! But this comment is confusing, I'll update it 😅

server/service/src/invoice/outbound_return/update/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+impl From<RepositoryError> for UpdateOutboundReturnError {
+    fn from(error: RepositoryError) -> Self {
+        UpdateOutboundReturnError::DatabaseError(error)
+    }
+}
+
+impl UpdateOutboundReturnStatus {
+    pub fn full_status(&self) -> InvoiceRowStatus {
```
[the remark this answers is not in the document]
lache-melvin: hm yeah - I think I was following `full_status` from somewhere else, but it's not very clear... I'll change to `as_invoice_row_status`

PR #3133, opened 2024-02-29
3132 delete outbound returns

### LINKED ISSUE #3132: deleteOutboundReturnLines API implementation

Delete the return, and release reserved stock in stock lines

lache-melvin:
Fixes #3132

# 👩🏻‍💻 What does this PR do? 
 

Implements the deleteOutboundReturns mutation.

Automated tests added in the service layer.

- In the Outbound Returns list view, add some returns
- Select some of them, and select the `Delete selected lines` action
- Lines should delete

Have done this a bit different to existing patterns, with accepting an array of ids to delete, rather a batch mutation... thoughts?


server/graphql/invoice/src/mutations/outbound_return/delete.rs
```diff
@@ -30,26 +36,35 @@ pub enum DeleteResponse {
...
-    //         .invoice_service
-    //         .delete_inbound_shipment(&service_context, input.to_domain()),
-    // )
-    Ok(DeleteResponse::Response(DeletedIdsResponse(vec![
-        "deleted_id".to_string(),
-    ])))
+    map_response(
+        service_provider
+            .invoice_service
+            .delete_outbound_returns(&service_context, input.ids),
```
lache-melvin, line 55: This is a bit different than the pattern elsewhere - I'm accepting a vec of ids to delete, and then looping in the service... this felt more direct than the batchMutation deletes done elsewhere. Keen for feedback on this?

I suppose the other option would be to make this a singular `deleteOutboundReturn` mutation, and call it multiple times from the frontend?

server/graphql/invoice/src/mutations/outbound_return/delete.rs
```diff
@@ -60,3 +75,27 @@ pub enum DeleteErrorInterface {
...
+    let graphql_error = match error {
+        // Structured Errors
+        ServiceError::InvoiceDoesNotExist => {
+            return Ok(DeleteErrorInterface::RecordNotFound(RecordNotFound {}))
+        }
+        ServiceError::CannotEditFinalised => {
+            return Ok(DeleteErrorInterface::CannotEditInvoice(
+                CannotEditInvoice {},
+            ))
+        }
```
lache-melvin, line 92: Are these structured or considered standard errors now?

server/service/src/invoice/outbound_return/delete/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+};
+
+type OutError = DeleteOutboundReturnError;
+
+pub fn delete_outbound_returns(
+    ctx: &ServiceContext,
+    ids: Vec<String>,
+) -> Result<Vec<String>, OutError> {
+    ctx.connection
+        .transaction_sync(|connection| -> Result<(), OutError> {
```
lache-melvin, line 25: Because of the for loop, I had to annotate the return type here... am I missing something for Rust to infer the type or is this expected?

server/graphql/invoice/src/mutations/outbound_return/delete.rs
jmbrunskill: @mark-prins and I were discussing this too.

Batch mutation deletes feels a bit heavy going. We're planning to do the single delete mutation called multiple times (similar to what we did in notify). I think the times when we want to delete a lot of items should be very rare, so in most cases there's not going to be a big performance improvement, or even if there is, is it important?

Would be good to discuss as a team and agree on a pattern going forward though.

client/packages/invoices/src/Returns/api/hooks/document/useOutboundDeleteRows.ts
```diff
@@ -2,13 +2,13 @@ import {
...
   useTranslation,
   useQueryClient,
-  InvoiceNodeStatus,
   useDeleteConfirmation,
   useUrlQueryParams,
   useMutation,
 } from '@openmsupply-client/common';
 import { useOutbounds } from './useOutbounds';
 import { useReturnsApi } from '../utils/useReturnsApi';
+import { canDeleteOutboundReturn } from 'packages/invoices/src/utils';
```
mark-prins, line 11: Can you change the import please - importing from 'packages/' like this will break the jest tests

```suggestion
import { canDeleteOutboundReturn } from '../../../../utils';
```

server/graphql/invoice/src/mutations/outbound_return/delete.rs
jmbrunskill: If your question is is this implementation using structured errors then no, this would be `standard` errors.
If you mean should it be a structured error, I don't personally thing so, if the frontend needs to do something special to handle the error then maybe it should be structured...

server/service/src/invoice/outbound_return/delete/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+    invoice::common::get_lines_for_invoice,
+    invoice_line::stock_out_line::{
+        delete::{delete_stock_out_line, DeleteStockOutLine, DeleteStockOutLineError},
+        StockOutType,
+    },
+    service_provider::ServiceContext,
+    WithDBError,
+};
+
+type OutError = DeleteOutboundReturnError;
```
jmbrunskill, line 18: Why not just use `DeleteOutboundReturnError` directly?
I guess this saves typing but personally I prefer to see the type...

server/graphql/invoice/src/mutations/outbound_return/delete.rs
mark-prins: this would be my preference - though we've just implemented the 'loop in the front end' version in assets.
one question - what happens with errors? if one of the deletes fail, is this a transactional operation now, so the entire Vec of ids is returned as an error? I think the batch mutation returns a collection of responses, so you can have some fail and some succeed

server/graphql/invoice/src/mutations/outbound_return/delete.rs
mark-prins: I think both of those are standard 🤷

server/service/src/invoice/outbound_return/delete/mod.rs
jmbrunskill: I don't think it's because of the for loop (at least getting rid of the for loop didn't fix the error for me)
I think it's because you have several different error types returned from `?` calls in the `transaction_sync`.

If all the `?` returns are the same type it can infer the type, but it doesn't seem to infer a type that they can all reach via a `From` implementation. I suspect it might work if the `transaction_sync` function didn't wrap the returned error in it's on generic.

This also works to fix the problem (as an alertnative approach)
`        .map_err(|error: TransactionError<DeleteOutboundReturnError>| error.to_inner_error())?;`

server/service/src/invoice/outbound_return/delete/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+    NotAnOutboundReturn,
+}
+
+impl From<RepositoryError> for DeleteOutboundReturnError {
+    fn from(error: RepositoryError) -> Self {
+        DeleteOutboundReturnError::DatabaseError(error)
+    }
+}
+
+impl From<TransactionError<DeleteOutboundReturnError>> for DeleteOutboundReturnError {
```
jmbrunskill, line 83: Maybe implementing this would avoid the need to specify the return type?
```
impl From<TransactionError<DeleteOutboundReturnError>> for RepositoryError {
```
I don't think we need to do it though!

mark-prins (changes requested):
Changes look good to me! Not sure about the batch mutations question, and I expect Andrei would have a more useful opinion than me.

Couple of things to change - the returns work currently doesn't play with postgres, can you 
- merge develop into this branch to get the 1.8 migrations
- add the following to the migrations, for postgres only:
```
ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_MUTATE';
ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_QUERY';
```

when importing in front end code, if you import from 'packages/..' then the jest tests won't run. Can you just use the alias imports please 🙏  I've updated in [this branch](https://github.com/msupply-foundation/open-msupply/compare/3132-delete-outbound-return...3132-delete-outbound-return(imports))

server/service/src/invoice/outbound_return/delete/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+        service
+            .delete_outbound_returns(&context, vec![mock_outbound_return_a().id])
+            .unwrap();
+
+        //test entry has been deleted
+        assert_eq!(
+            InvoiceRowRepository::new(&connection)
+                .find_one_by_id_option(&mock_outbound_return_a().id)
+                .unwrap(),
+            None
```
jmbrunskill, line 208: Slightly tempted to suggest testing that stocklines are updated if picked as per `delete_stock_out_line`
It should be covered in the other test cases, but would be nice to catch if someone tries to `improve` this function in the future and try to say call the repository layer directly.
Probably not likely but was trying to understand what the expected behavour is here.

jmbrunskill (approved):
Looks good to me, possibly simplier to change to a single delete per request?
Let's discuss next week.

server/service/src/invoice/outbound_return/delete/mod.rs
mark-prins: this is probably the cleverest typo that I've seen!
`alertnative` actually makes sense.. this is the 'alert native' approach 😆

lache-melvin:
> the returns work currently doesn't play with postgres, can you
> 
> * merge develop into this branch to get the 1.8 migrations
> * add the following to the migrations, for postgres only:
> 
> ```
> ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_MUTATE';
> ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_QUERY';
> ```

Thanks Mark, will do this in the feature branch!

mark-prins (approved):
👍  LGTM!
thanks for the changes, looks good

PR #3135, opened 2024-03-01
Startup log message now shows central/remote and port number

jmbrunskill:
Starting OMS should show Central or Remote Server
`<server:65>:Remote server starting in Development mode on port 8082`
or
`<server:65>:Central server starting in Development mode on port 8082`


mark-prins (approved):
😍  love it! thanks - super useful

PR #3136, opened 2024-03-01
Migration fixes

### LINKED ISSUE #2942: Unable to log into omSupply after upgrading the versions
## What went wrong? 😲
 

When upgrading from version 1.1.15 to version 1.6.01, discovered we are unable to log into omSupply, and getting the following error: 

telegram-cloud-photo-size-2-5359763283399271347-y

The dev tools screenshot (me request): 

Screenshot 2024-02-14 at 5 23 23 PM

## Expected behaviour 🤔

Clients able to log into omSupply with no issues when the versions are upgraded

## How to Reproduce 🔨

1. Install an old version of omSupply (for example, v1.1.15)
2. Upgrade to a new version of omSupply (I was using v1.6.01)
3. Attempt to log into omSupply with your credentials -> see error

## Your environment 🌱

- Version: upgrading from 1.1.15 to 1.6.01
- Platform:
  - [x] android (tablet) Galaxy Tab A8
  - [ ] desktop (windows)
  - [ ] desktop (macOS)
  - [ ] server (windows)
- Database type: 
  - [ ] PostgreSQL
  - [ ] SQLite3

mark-prins:
Fixes #2942

# 👩🏻‍💻 What does this PR do? 
 
There's a couple of issues with migration and these prevent deserialising the `name` table, which causes the login problem.

The column `is_deceased` on `name` is declared as `Bool` but is nullable. If you start with a new database and initialise, then the values in this field are `0` when using sqlite. If you upgrade a database though, all values are `NULL` and this doesn't deserialise.

To fix, I've changed the new column type to be non null with a default of false.

Have also added migration script to change existing databases, even though this is probably overkill. It means that all databases will be consistent though.

Second problem is that the `clinician` and `clinician_store_join` tables were created in v1.2.00 however there is migration script in 1.1.14 which tries to modify those tables. because of this, the addition of `is_sync_update` will fail, and the column won't exist. I've added a migration to run the addition in the version 1.6 migration, but with error handling around it to allow the statement to fail quietly if the column exists ( SQLite again - we can use IF EXISTS in postgreSQL )

You'll need to get v1.1.14, create a database, initialise, switch to this branch and migrate.

Simplest way is to use this test database:

The SQLite database in `omsupply_test_1.1.14.zip` on [google drive](https://drive.google.com/open?id=1e7475Z9J-j1KnferOQOrcYSSTnFp9wtk&usp=drive_fs) is a database at that version. it's useful to start with and try migrating up to 1.6

But if you prefer trying manually, you can use `git checkout tags/v1.1.14 -b main-1.1.14` to checkout the 1.1.1.4 tag to a new branch, which you can then delete when finished with it.

An additional test for this - create a database for version 1.5.00 of the codebase, and migrate to 1.6; see if the `is_deceased` column has the default of false and is non nullable.

## 📃 Documentation

_No user facing changes_


server/repository/src/migrations/v1_06_00/is_sync_update.rs
```diff
@@ -0,0 +1,223 @@
+use crate::{migrations::sql, StorageConnection};
```
mark-prins, line 1: this is a copy from the original migration

server/repository/src/migrations/v1_06_00/name_is_deceased.rs
```diff
@@ -0,0 +1,112 @@
...
+            r#"
+            UPDATE name set is_deceased = false WHERE is_deceased is null;
+            ALTER TABLE name ALTER COLUMN is_deceased SET NOT NULL;
+            ALTER TABLE name ALTER COLUMN is_deceased SET DEFAULT false;
+          "#
+        )?;
+    }
+
+    // the same as above - but there's no alter column for SQLite
+    if cfg!(not(feature = "postgres")) {
```
mark-prins, line 16: this is the overkill version. not needed? we could just have some installs with a nullable bool and others without?

andreievg:
> There's a couple of issues with migration and these prevent deserialising the name table, which causes the login problem.

Was wondering how it causes the login problem ?

andreievg:
@mark-prins, i am a bit worried about dropping and re-creating name table. We've been adding tests for data migrations like this and I feel like it would take quite a bit time to do a full test. How about just `update name set is_deceased = false where is_deceased is null`, and maybe follow up issue to enforce constraint in database, or drop constraint in code ?

andreievg:
> Second problem is that the clinician and clinician_store_join tables were created in v1.2.00 however there is migration script in 1.1.14 which tries to modify those tables. because of this, the addition of is_sync_update will fail, and the column won't exist. I've added a migration to run the addition in the version 1.6 migration, but with error handling around it to allow the statement to fail quietly if the column exists ( SQLite again - we can use IF EXISTS in postgreSQL )

I was wondering if this is something we should be fixing in this immediate patch ? I think it's really good to get a save patch out ASAP ? 

Just a preliminary comment about the second fix, i am just trying to figure it out now, but if it wasn't there could probably approve and merge this asap

mark-prins:
> i am a bit worried about dropping and re-creating name table

Yeah, me too. it's more than strictly needed - as the translations handle the data correctly. 
An alternative is to just change the type to match the database : change `name_row` to have `is_deceased` as `Option<Bool>` I thought it was better to have is_deceased non nullable and with a default. can't see the point of having nullable on that column

andreievg:
> I was wondering if this is something we should be fixing in this immediate patch ? I think it's really good to get a save patch out ASAP ?

Had a check and tried seems to be ok, and yeah definitely the triggers were missing and is_sync_update field too

andreievg:
> An alternative is to just change the type to match the database : change name_row to have is_deceased as Option<Bool>

I tried, there stopped at 10th file change lol, and then just tried the update statement and it worked

jmbrunskill:
I tested another site and the migration looks to be working fine for me.

jmbrunskill (approved):
Looks good to me, agree it's probably best to not recreate the name table given it's a hotfix.

andreievg (approved):
Thanks for changes, i tested in a few ways, migrating from 1.1.14 and from scratch seems to work nicely now, thanks

PR #3146, opened 2024-03-03
3104 sync asset class, category,  central to remote

### LINKED ISSUE #3104: Sync Assets -> Central to Remote

We have an `asset` table now, in the asset branch, however, we currently don't sync it to remote sites or to central.
The intention is for asset to be something akin to `shared_central` data. E.g. assets can be created centrally or at the remote site.

Some assets will have no store_id assigned, these won't be sync'd anywhere until they are assigned to a store.

### Moneyworks Jobcode 🧰

OMSCC:AST

jmbrunskill:
Fixes #3104

# 👩🏻‍💻 What does this PR do? 
Central to Remote Sync for

- [x] Class
- [x] Category
- [x] Type
- [x] Catalogue Item

Sync Tests included in PR
Hard to test now as we don't actually have UI, so this will need to be tested when these features are added in future.

For my testing I used these inserts on the central server...
```sql
INSERT INTO "main"."asset_class" ("id", "name") VALUES ('wce', 'Warm Chain Equpment');
INSERT INTO "main"."asset_category" ("id", "name", "asset_class_id") VALUES ('defridge', 'Defridgerators', 'wce');
INSERT INTO "main"."asset_type" ("id", "name", "asset_category_id") VALUES ('micro', 'microwaves','defridge');
INSERT INTO "main"."asset_catalogue_item" ("id", "code", "asset_class_id", "asset_category_id", "asset_type_id", "manufacturer", "model") VALUES ('wave1', 'WAVE/001', 'wce', 'defridge', 'micro', 'Microwave1', 'MCR1');
INSERT INTO "main"."asset_catalogue_item" ("id", "code", "asset_class_id", "asset_category_id", "asset_type_id", "manufacturer", "model") VALUES ('wave2', 'WAVE/001', 'wce', 'defridge', 'micro', 'Microwave2', 'MCR2');
```

> Note: To test this you'll need to re-initialise your central and remote servers as if you've already run migration 1.8 you might not have the triggers added in this PR.

At this stage, all the asset catalogue tables will be sync'd to all sites, there's not store, or name joins to worry about.
Similar pattern to `pack_variant`

## 📃 Documentation

_No user facing changes_


server/service/src/sync/translations/asset_catalogue_item.rs
```diff
@@ -0,0 +1,108 @@
...
+        Ok(PullTranslateResult::upsert(serde_json::from_str::<
+            AssetCatalogueItemRow,
+        >(&sync_record.data)?))
+    }
+
+    fn change_log_type(&self) -> Option<ChangelogTableName> {
+        Some(ChangelogTableName::AssetCatalogueItem)
+    }
+
+    // Only translating and pushing on central server
```
andreievg, line 49: I think this comment is now wrong, Only translate when pulling from omSupply Central ?

andreievg (approved):
Looks great and easy to review. Thanks.

I wonder if we can isolate one of the tables and create an example of how to add new omSupply central sync record ?

At some point, we would need to add compatibility check (but this is only when remote starts pushing)

jmbrunskill:
> I wonder if we can isolate one of the tables and create an example of how to add new omSupply central sync record ?
> 
I thought this would be a good example? https://github.com/msupply-foundation/open-msupply/pull/3146/commits/989ed2e56f77923d3a73f033956ae1d8e68fe196

Happy with that? I'll update the docs...

PR #3148, opened 2024-03-03
Update bug_report.md

roxy-dao:
The check list makes people think that there are tasks to do in the bug reports when there aren't.


Chris-Petty (approved):
Yussssss

Issue #3030, opened 2024-02-20
Cold Chain - Battery Level on Berlinger tags message: 'Not configured'/ 'undefined'


regotaina:
## Is your feature request related to a problem? Please describe 👀

For the end user it can be confusing to read [Not configured] on the main page, but when clicking on the monitor itself, it says 'undefined'. Unless they know it won't be captured, they can think there's an error to be fixed.

@adamdewey agrees it would be nice to show the same message to have uniformity, otherwise it can be confusing.

Monitors menu:
<img width="1489" alt="Screenshot 2024-02-21 at 9 15 40 AM" src="https://github.com/msupply-foundation/msupply-cold-chain/assets/132425096/5ff54e64-cae1-4004-9afa-05232aa31cb8">

Clicking on the monitor itself:
<img width="586" alt="Screenshot 2024-02-21 at 9 19 56 AM" src="https://github.com/msupply-foundation/open-msupply/assets/132425096/8fbca12b-d711-4fd9-99c5-3d0d0273758f">

## Describe the solution you'd like 🎁

The message can be "Not available" on both pages. or "Not configurable". 

### Describe alternatives you've considered 💭

Open to other ideas.

### Additional context 💌

Are we considering getting this data later on? Or it's just not on the roadmap?
I imagine the different message may be a result from the API response.

adamdewey:
Thanks @regotaina 

I think perhaps the wording `Not available` would be better?

adamdewey:
Or, actually, to save screen real estate - maybe we should just use a hypen `-` ?

mark-prins:
maybe the same time as #3055

marthakeezy:
Tested the above - the battery level is now displayed as '-', when looking at the list of sensors available and when looking at a specific sensor. Test passed ✅

PR #3151, opened 2024-03-04
3055 & 3030 cold chain display

### LINKED ISSUE #3055: Cold Chain: Sensors information should include a degree symbol

When logging into omSupply and navigating to Cold Chain -> Sensors, we can see a column called 'Last Reading' with the latest temperature information, and it includes a Celsius symbol. However, when clicking on a specific sensor to get more information, 'Last Reading' line only displays the numbers and not the Celsius symbol. 

The formatting of the readings should be consistent, so the 'Last Reading' when expanded should have the same degree symbol as when you first see the entry. 

N/A

### Additional context 💌

screenshot-1708653648676

[
screenshot-1708653697883
](url)

### Moneyworks Jobcode 🧰

roxy-dao:
Fixes #3055 & #3030

# 👩🏻‍💻 What does this PR do? 
Updates the temperature formats to use `Intl.NumberFormat` and handle the undefined values that were being returned in sensor modal

- [ ] Check temperatures are still being formatted as `X°C`
- [ ] Go to a sensor without a battery level and see that it now says '-' instead of undefined%

### DIFF
--- a/client/packages/coldchain/src/common/utils.ts
+++ b/client/packages/coldchain/src/common/utils.ts
@@ -1,3 +1,4 @@
+import { useIntlUtils } from '@common/intl';
 import { TemperatureBreachNodeType } from '@common/types';
 
 export const parseBreachType = (
@@ -7,4 +8,13 @@ export const parseBreachType = (
   const type = breachType?.split('_')[1];
 
   return { temperature, type };
+};
+
+export const useFormatTemperature = (temperature: number) => {
+  const { currentLanguage: language } = useIntlUtils();
+  return new Intl.NumberFormat(language, {
+    style: 'unit',
+    unit: 'celsius',
+    unitDisplay: 'short',
+  }).format(temperature);
 };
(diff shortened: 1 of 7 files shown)


client/packages/coldchain/src/common/utils.ts
```diff
@@ -8,3 +9,12 @@ export const parseBreachType = (
...
   return { temperature, type };
 };
+
+export const useFormatTemperature = (temperature: number) => {
+  const { currentLanguage: language } = useIntlUtils();
+  return new Intl.NumberFormat(language, {
+    style: 'unit',
+    unit: 'celsius',
+    unitDisplay: 'short',
+  }).format(temperature);
```
mark-prins, line 19: given that all of the usages of this function round before passing in the number, I would do this:

```suggestion
  }).format(NumUtils.round(temperature, 2));
```

mark-prins (approved):
😄 lovely thanks.
One suggestion, which I've put in a [suggestion branch](https://github.com/msupply-foundation/open-msupply/compare/3055-Cold-Chain-Sensors-information-should-include-a-degree-symbol...3055-Cold-Chain-Sensors-information-should-include-a-degree-symbol(rounding)) in case that's helpful.

PR #3152, opened 2024-03-04
Remove store_id from catalogue queries

jmbrunskill:
Fixes a bug where asset queries are trying to use a store_id which doesn't exist in graphql


mark-prins (approved):
👍 LGTM!

PR #3154, opened 2024-03-04
3144 generate outbound lines existing return

### LINKED ISSUE #3144: generateOutboundReturnLines - return_id

generateOutboundReturnLines accepts a return_id, but doesn't do anything with it yet. If a return ID is supplied, we should respond with all the items for that return. If an item ID is also supplied, we should filter by that item ID.

lache-melvin:
Fixes #3144

# 👩🏻‍💻 What does this PR do? 
 

If a return_id is supplied, we can include existing return lines in the generated lines for a particular item. Use case being onRowClick on an existing return line. The modal would open, and populate with both the existing lines (to edit) and new lines for all other batches for that item.

Automated tests in the service layer. It's a little convoluted - are there cases I'm not covering in the tests?

I'll hook this up to the edit modal from #3041 when one of them gets merged...


server/service/src/invoice/outbound_return/generate_lines/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+fn invoice_line_to_new_return_line(
+    ctx: &ServiceContext,
+    store_id: &str,
+    line: &InvoiceLine,
+) -> Result<OutboundReturnLine, RepositoryError> {
+    let stock_line_id = line
+        .invoice_line_row
+        .stock_line_id
+        .as_ref()
+        .ok_or(RepositoryError::NotFound)?;
```
lache-melvin, line 126: This might be a little bit lazy version... perhaps it would be better for me to create a specific service error here? Or try doing these checks in a validate method? Just noticed that most queries don't have any of that....

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -26,62 +29,178 @@ pub fn generate_outbound_return_lines(
...
+fn invoice_line_to_new_return_line(
+    ctx: &ServiceContext,
+    store_id: &str,
+    line: &InvoiceLine,
+) -> Result<OutboundReturnLine, RepositoryError> {
+    let stock_line_id = line
+        .invoice_line_row
+        .stock_line_id
+        .as_ref()
+        .ok_or(RepositoryError::NotFound)?;
```
lache-melvin, line 126: Using RepositoryError here might be a little bit lazy version... perhaps it would be better for me to create a specific service error here? Or try doing these checks in a validate method? Just noticed that most queries don't have any of that, and this mapped easily into the ListError 🙈

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -26,62 +29,178 @@ pub fn generate_outbound_return_lines(
         return_id,
     }: GenerateOutboundReturnLinesInput,
 ) -> Result<ListResult<OutboundReturnLine>, ListError> {
```
lache-melvin, line 31: I think this whole function could be written a lot better, but I've already spent quite a while on it. How would you improve this?

server/graphql/general/src/queries/generate_outbound_return_lines.rs
lache-melvin, line 1: Since we had the mock API implementation, we've had this file under general queries... wondering if it should go in the invoice queries though?

jmbrunskill (approved):
Good work happy to merge it when you're ready, but tried to give some useful feedback :)

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -197,7 +233,10 @@ mod test {
                     return_id,
                 },
             ),
-            Err(ListError::DatabaseError(RepositoryError::NotFound))
+            Err(ListError::DatabaseError(RepositoryError::as_db_error(
```
jmbrunskill, line 236: Ooh I haven't seen this `as_db_error` before looks handy!

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -197,7 +233,10 @@ mod test {
...
+                "Invoice line has not stock line ID",
```
jmbrunskill, line 237:
```suggestion
                "Invoice line has no stock line ID",
```

server/service/src/invoice/outbound_return/generate_outbound_return_lines.rs
```diff
@@ -20,68 +24,222 @@ pub struct GenerateOutboundReturnLinesInput {
...
-    }
     match item_id {
         Some(item_id) => {
-            filter.item_id = Some(EqualFilter::equal_to(&item_id));
-            filter.is_available = Some(true);
+            let filter = StockLineFilter::new()
+                .item_id(EqualFilter::equal_to(item_id))
+                .is_available(true);
+
+            let stock_lines = stock_line_repo.query_by_filter(filter, Some(store_id.to_string()));
```
andreievg, line 87: Actually the store_id here is the filter for item visibility, do you know if stock lines are actually filtered for this store ? What if we add stock lines for another store would they be captured here ?

server/graphql/general/src/queries/generate_outbound_return_lines.rs
jmbrunskill: Sorry I think I missed this comment before, Yes I think I'd move it out of `general to `invoice` or somewhere else if there's a better place? @andreievg ?

andreievg:
Thanks @lache-melvin, I was trying to apply my original though to the logic and came up with this: https://github.com/msupply-foundation/open-msupply/compare/3144-generate-outbound-lines-existing-return...3144-generate-outbound-lines-existing-return(Suggested-changes-for-generating-outbound-return-lines)

Which goes like this:

Get existing return lines for item_id and stock_line_ids in input.

Get new array of stock_lines_ids based on input and existing return lines (based on input).

Get stock lines based on the new stock_line_ids array

Get stock lines based on item_id input (minus the new stock_line_ids array)

Combine stock_lines with return lines

Map combined stock_lines and return lines to OutboundReturnLine

I like the above approach, but I could be biased, please check and see if it makes sense to you, and apply change if you agree), why I like this approach:
* Don't have to look up stock lines again
* I really like idea of creating structure, and changing them in seperate steps, which I think this suggestion shows

> Automated tests in the service layer. It's a little convoluted - are there cases I'm not covering in the tests?

I really liked the test, they helped a lot when refactoring to the above suggestion and helps to understand the use cases.

I think it's good to add a test where input is item_id but stock belongs to more then one store (to make sure we only include stock in this store)

lache-melvin:
> Thanks @lache-melvin, I was trying to apply my original though to the logic and came up with this: [3144-generate-outbound-lines-existing-return...3144-generate-outbound-lines-existing-return(Suggested-changes-for-generating-outbound-return-lines)](https://github.com/msupply-foundation/open-msupply/compare/3144-generate-outbound-lines-existing-return...3144-generate-outbound-lines-existing-return(Suggested-changes-for-generating-outbound-return-lines))
> 
> Which goes like this:
> 
> Get existing return lines for item_id and stock_line_ids in input.
> 
> Get new array of stock_lines_ids based on input and existing return lines (based on input).
> 
> Get stock lines based on the new stock_line_ids array
> 
> Get stock lines based on item_id input (minus the new stock_line_ids array)
> 
> Combine stock_lines with return lines
> 
> Map combined stock_lines and return lines to OutboundReturnLine
> 
> I like the above approach, but I could be biased, please check and see if it makes sense to you, and apply change if you agree), why I like this approach:
> 
> * Don't have to look up stock lines again
> * I really like idea of creating structure, and changing them in seperate steps, which I think this suggestion shows
> 
> > Automated tests in the service layer. It's a little convoluted - are there cases I'm not covering in the tests?
> 
> I really liked the test, they helped a lot when refactoring to the above suggestion and helps to understand the use cases.
> 
> I think it's good to add a test where input is item_id but stock belongs to more then one store (to make sure we only include stock in this store)

I really like this, thanks Andrei. I'll merge it in 😎

PR #3158, opened 2024-03-04
3051 Doesn't select the searched entry from list in program enrolment form

### LINKED ISSUE #3051: Doesn't select the searched entry from list in program enrolment form
## How to Reproduce 🔨

1. Get programs-setup datafile like PNG Demo
2. Have some patients with common starting letters like eg: Mimi Chan, Mimi Dan, etc
3. Create new patient and enrol it into a program like HIV Care & Treatment
4. Now in the enrolment form, in Mother field, put few characters in first name such that both previous names appear in search option, eg: put Mimi in first name and see both `Mimi Chan` and `Mimi Dan` appear in the list
5. Make sure to select the second one: `Mimi Dan`
6. See error: it automatically selects the first one : `Mimi Chan`

## Expected behaviour 🤔
It should select the one we select; not take the first entry by default! 

## Screenshot/Recording
https://recordit.co/S4MBEPHXbN

## Additional Context
Found while testing #1840 

## Your environment 🌱

- Version: 1.6.2
- mSupply version: 713-00-beta3
- Platform:
  - server (windows)
- Database type: 
  - SQLite3

roxy-dao:
Fixes #3051

# 👩🏻‍💻 What does this PR do? 
Fix the patient check.

- [ ] Have programs set up
- [ ] Create two patients that have the same first name `e.g. Mimi` and a third with whatever name you want
- [ ] Enrol the third patient into `HIV Care and Treatment` program
- [ ] Go to the mother field and type `Mimi` or whatever name you have chosen
- [ ] Select the 2nd `Mimi` on the list
- [ ] Should load correctly

### DIFF
--- a/client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
+++ b/client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
@@ -53,10 +53,10 @@ export const SearchWithUserSource = (
   const getOptionLabel = (data: PatientSchema) =>
     options?.optionString
       ? formatTemplateString(options?.optionString, data)
-      : `${data['code']} - ${data['firstName']} ${data['lastName']}`;
+      : `${data['code'] ?? ''} - ${data['firstName']} ${data['lastName']}`;
 
   const handlePatientSelect = (patientId: string) => {
-    const patient = results.find(p => (p.id = patientId));
+    const patient = results.find(p => p.id === patientId);
     if (!patient) return;
     if (!options?.saveFields) {
       handleChange(path, patient);
--- a/client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
+++ b/client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
@@ -53,7 +53,9 @@ export const SearchWithUserSource = (
   const getOptionLabel = (data: PatientSchema) =>
     options?.optionString
       ? formatTemplateString(options?.optionString, data)
-      : `${data['code'] ?? ''} - ${data['firstName']} ${data['lastName']}`;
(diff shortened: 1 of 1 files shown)


client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
```diff
@@ -53,10 +53,10 @@ export const SearchWithUserSource = (
   const getOptionLabel = (data: PatientSchema) =>
     options?.optionString
       ? formatTemplateString(options?.optionString, data)
-      : `${data['code']} - ${data['firstName']} ${data['lastName']}`;
+      : `${data['code'] ?? ''} - ${data['firstName']} ${data['lastName']}`;
```
roxy-dao, line 56: Code wasn't required.... so it showed up as undefined for me 👀  probably a better way to display this undefined value.... or make code required?

jmbrunskill (approved):
Haven't tested but looks like a valid code change :)

Ok to leave testing to the testers?

client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
clemens-msupply: If there is no code we shouldn't show the extra " - ", right? e.g.

```
`${data['code'] ?? `${data['code']} - `}${data['firstName']} ${data['lastName']}`;
```

roxy-dao: Good idea~

roxy-dao:
> Haven't tested but looks like a valid code change :)
> 
> Ok to leave testing to the testers?

Yup~ Happy to leave it to testers if you think code looks good (:

PR #3168, opened 2024-03-05
Inbound shipments: long references not visible when status is "shipped" #3079

### LINKED ISSUE #3079: Inbound shipments: long references not visible when status is "shipped"
## What went wrong? 😲

1) Long references are not visible in the Inbound Shipment list view:
image
2) For shipped shipments the reference is read only and you can't read long references:
image
(Or am I choosing an unrealistic long reference here?) 

## Expected behaviour
Have a tooltip (or something touch friendly) to see references in the list view and in the detail view.

roxy-dao:
Fixes #3079

# 👩🏻‍💻 What does this PR do? 
** Ignore branch name was created for another bug and forgot to change 🙄 and renaming will close PR again **
Adds tooltip to invoice/requisition list views and detail header for comments/references since long references/comments get cut off.

- [ ] Have an invoice/requisition with a long comment/reference... maybe do a transfer? 
- [ ] Hover over to see tooltip

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
@@ -50,6 +50,9 @@ export const BasicTextInput: FC<BasicTextInputProps> = React.forwardRef(
             ...InputProps?.sx,
           },
         }}
+        inputProps={
+          props.disabled ? { style: { textOverflow: 'ellipsis' } } : {}
+        }
         {...props}
       />
       <Box width={2}>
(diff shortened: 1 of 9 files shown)


Chris-Petty:
Potentially still needs design cause tooltips are contentious for mobile users, but an improvement over overflow!

roxy-dao:
> Potentially still needs design cause tooltips are contentious for mobile users, but an improvement over overflow!

Yeah but I think create another issue for that because tooltips are used in a lot of areas and would need change overall, but have done this for now to keep consistency

mark-prins (approved):
👍  cool - that works nicely on desktop thanks.
I suspect that being able to stretch out the input and show a longer text string would be better, but that's a bit fiddly to get working with the flex layout. What you've done is fine by me!

One thought: if you add this

```
        inputProps={
          props.disabled ? { style: { textOverflow: 'ellipsis' } } : {}
        }
```

to the `BasicTextInput` then you get more of an indication that the text is truncated:

image

PR #3169, opened 2024-03-05
#3012 Detail view Footer

### LINKED ISSUE #3012: Outbound returns detail view -- Footer

Add Footer to page

CarlosNZ:
Fixes #3012

# 👩🏻‍💻 What does this PR do? 
 

Displays Footer with Statuses and buttons, although the functionality of the buttons is not yet implemented.

A lot of copy-paste from Outbound Shipments here. I've noted a couple of larger functions which it would probably good to put into a shared area to avoid large duplications -- maybe make a new issue for that?


client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    statusLog[InvoiceNodeStatus.Shipped] = invoice.shippedDatetime;
+  }
+  if (statusIdx >= 4) {
+    statusLog[InvoiceNodeStatus.Delivered] = invoice.deliveredDatetime;
+  }
+  if (statusIdx >= 5) {
+    statusLog[InvoiceNodeStatus.Verified] = invoice.verifiedDatetime;
+  }
+  return statusLog;
+};
```
CarlosNZ, line 49: This function is duplicated in multiple places. Should we pull out into the common invoices/utils folders?

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+const getNextStatusOption = (
+  status: InvoiceNodeStatus,
+  options: SplitButtonOption<InvoiceNodeStatus>[]
+): SplitButtonOption<InvoiceNodeStatus> | null => {
+  if (!status) return options[0] ?? null;
+
+  const nextStatus = getNextOutboundStatus(status);
+  const nextStatusOption = options.find(o => o.value === nextStatus);
+  return nextStatusOption || null;
+};
```
CarlosNZ, line 85: Again, can probably pull out into shared functions?

client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    [InvoiceNodeStatus.Shipped]: null,
+    [InvoiceNodeStatus.Delivered]: null,
+    [InvoiceNodeStatus.Verified]: null,
+  };
+  if (statusIdx >= 0) {
+    statusLog[InvoiceNodeStatus.New] = invoice.createdDatetime;
+  }
+  if (statusIdx >= 1) {
+    statusLog[InvoiceNodeStatus.Allocated] = invoice.allocatedDatetime;
+  }
```
lache-melvin, line 35: Can remove this step, `Allocated` isn't a status for Outbound Returns - probably also means you should create an `outboundReturnsStatuses` for use on line 21, and in the component below 👍 

Have added in the naming issue to rename all the existing `outbound` references to `outboundShipment` so stuff like this is less easy to confuse!

client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
lache-melvin: Yeah let's create a refactor issue for this! Might need a bit of rework to handle the different invoice types having different statuses, but agree it's a lot of duplication.

I wonder if the `statusIdx` pattern is out of date for frontend as well, like Andrei was saying yesterday for backend they're moving to matching on explicit types?

client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
```diff
@@ -0,0 +1,92 @@
...
+  useTranslation,
+  AppFooterPortal,
+  //   InvoiceNodeStatus,
+  useBreadcrumbs,
+  InvoiceNodeStatus,
+} from '@openmsupply-client/common';
+import { getStatusTranslator, outboundStatuses } from '../../../utils';
+import { OutboundReturnRowFragment, useReturns } from '../../api';
+
+// import { useOutbound, OutboundFragment } from '../../api';
```
lache-melvin, line 16: Can remove the commented imports :)

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
lache-melvin: Agree, this would be great to create a refactor issue to pull out 👍

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+    SplitButtonOption<InvoiceNodeStatus>,
+    SplitButtonOption<InvoiceNodeStatus>,
+  ] = [
+    {
+      value: InvoiceNodeStatus.New,
+      label: getButtonLabel(InvoiceNodeStatus.New),
+      isDisabled: true,
+    },
+    {
+      value: InvoiceNodeStatus.Allocated,
```
lache-melvin, line 32: Remove all the allocated stuff here too :)

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+const useStatusChangeButton = () => {
+  // const { lines, status, onHold, update } = useOutbound.document.fields([
+  //   'status',
+  //   'onHold',
+  //   'lines',
+  // ]);
+
+  const { success, error } = useNotification();
+  const t = useTranslation('distribution');
```
lache-melvin, line 103:
```suggestion
  const t = useTranslation('replenishment');
```
Definitely keen to reorganise these translation files, ending up with heaps of duplicated strings across them 😅

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  // TEMP until "fields" hook available:
+  const status = data?.status ?? InvoiceNodeStatus.New;
+  const update = async (_: unknown) => true;
+  const onHold = false;
+  const lines: any = [];
+
+  const hasLinesToPrune =
+    data?.status === InvoiceNodeStatus.New &&
+    (data?.lines?.nodes ?? []).some(line => line.numberOfPacks === 0);
```
lache-melvin, line 114: Shouldn't be possible for returns, we require a number of packs to be set, or we remove the line on the backend

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+      getNextStatusOption(status, options)
+    );
+
+  const onConfirmStatusChange = async () => {
+    if (!selectedOption) return null;
+    try {
+      await update({ status: selectedOption.value });
+      success(t('messages.shipment-saved'))();
+    } catch (e) {
+      error(t('messages.error-saving-shipment'))();
```
lache-melvin, line 132:
```suggestion
      success(t('messages.return-saved'))();
    } catch (e) {
      error(t('messages.error-saving-return'))();
```

client/packages/invoices/src/Returns/api/operations.graphql
```diff
@@ -6,6 +6,12 @@ fragment OutboundReturnRow on InvoiceNode {
   invoiceNumber
   colour
   createdDatetime
+  createdDatetime
+  allocatedDatetime
```
lache-melvin, line 10:
```suggestion
```

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const { success, error } = useNotification();
+  const t = useTranslation('distribution');
+  const { data } = useReturns.document.outboundReturn();
+
+  // TEMP until "fields" hook available:
+  const status = data?.status ?? InvoiceNodeStatus.New;
+  const update = async (_: unknown) => true;
+  const onHold = false;
+  const lines: any = [];
```
lache-melvin, line 110: Per use below, looks like this should be something like { nodes: something[], totalCount: number }... I know it's just dummy data but may be good to have the types line up for reference when we do hook all this together?

lache-melvin (changes requested):
Looks great Carl! Main thing just to remove the Allocated stuff :)

client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
CarlosNZ: Yup, I've done it the same way as it's currently done in InboundShipment -- the unused one is still in this object, but it's removed from the proper list of statuses: 1c8f1168a

This will make it easier for abstracting this function to a common one too.

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
CarlosNZ: https://github.com/msupply-foundation/open-msupply/issues/3203 :)

client/packages/invoices/src/Returns/DetailView/Footer/Footer.tsx
CarlosNZ: > Yeah let's create a refactor issue for this! Might need a bit of rework to handle the different invoice types having different statuses, but agree it's a lot of duplication.

https://github.com/msupply-foundation/open-msupply/issues/3203

> I wonder if the `statusIdx` pattern is out of date for frontend as well, like Andrei was saying yesterday for backend they're moving to matching on explicit types?

It's only using the number locally as a way to determine order. I would probably do this a bit differently anyway, as part of the above refactor :)

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
CarlosNZ: Yeah, there's currently a script to find and remove unused strings, we can probably extend it to find duplicates as well
Screenshot 2024-03-06 at 4 40 58 PM

client/packages/invoices/src/Returns/DetailView/Footer/StatusChangeButton.tsx
CarlosNZ: Looks like only `totalCount` is being used below, but have added something close to your suggestion for now

lache-melvin (approved):
Great thanks Carl! 🚀

PR #3170, opened 2024-03-05
3155 generate inbound return lines

### LINKED ISSUE #3155: generateInboundReturnLines API implementation

Inverse of generateOutboundReturnLines - accepts either stock lines to generate return lines for, or item id/return id to add/edit from an existing return.

lache-melvin:
Fixes #3155 

# 👩🏻‍💻 What does this PR do? 
 

Generates inbound return lines. Includes both the generating from outbound shipment line ids, or returning existing return lines for a specific itemId. 

Automated tests in the service layer.

When selecting lines from an outbound shipment for an inbound return, generated lines now show correctly.

Tried a slightly different input type, wanting to constrain that if you want to include the existing return lines for a particular item_id, you'd need to provide both of those inputs... keen for thoughts here


server/service/src/invoice/inbound_return/generate_lines.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+pub struct ExistingLinesInput {
+    pub item_id: String,
+    pub return_id: String,
+}
+
+pub struct GenerateInboundReturnLinesInput {
+    pub outbound_shipment_line_ids: Vec<String>,
+    pub existing_lines_input: Option<ExistingLinesInput>,
```
lache-melvin, line 25: Experimenting with this input... wanting to constrain:
- If you want to generate new lines, pass in outbound shipment lines to create them from
- If you want show the existing lines for a particular item, you're going to need to include both the item and return ids

Not sure how much value it ends up adding, because might be harder to handle on the frontend? Haven't got the frontend implementation of editing an existing inbound return in place yet so didn't get to test...

Also not sure about the naming, but interested in thoughts on this idea!

server/service/src/invoice/inbound_return/generate_lines.rs
```diff
@@ -0,0 +1,269 @@
...
+    let existing_return_lines =
+        if let Some(ExistingLinesInput { item_id, return_id }) = include_existing_lines {
+            get_existing_return_lines(ctx, &item_id, &return_id)?
+        } else {
+            vec![]
+        };
+
+    let return_lines: Vec<InboundReturnLine> = existing_return_lines
+        .into_iter()
+        .chain(new_return_lines.into_iter())
```
lache-melvin, line 51: I've seen chain, extend and append in different places - I'm not confident on when to pick one over the other?

mark-prins (commented):
could you merge develop into this branch? the central server migrations haven't run. I've been working through adding the changes manually, but there's too many!

Changes look fine to me - though I'll leave the rust questions up for someone else to comment on ❤️

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
```diff
@@ -23,7 +29,7 @@ export const useDraftInboundReturnLines = (
     setDraftLines(newDraftLines);
   }, [lines]);
```
andreievg, line 30: By the way I think this is pretty bad pattern (I can see that it was copied from existing functionality though).

Here on startup, we wait on lines from another hook that calls useQuery that will essentially have useEffect(() => fetch(), []), and the data is passed back through two hooks to this one, essentially creating 3 layers of indirections and event/actions, before we actually set draft lines

I think ideally we just have `useEffect(() => fetchIt.then(stDraftLines), [])`

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
```diff
@@ -2,19 +2,25 @@ import React, { useEffect } from 'react';
...
+
+  const data = useReturns.lines.generateInboundReturnLines(
+    outboundReturnLineIds
+  );
+  const lines = data?.nodes;
 
-  const lines = useReturns.lines.inboundReturnLines(stockLineIds);
   const { mutateAsync } = useReturns.document.insertInboundReturn();
 
   useEffect(() => {
```
andreievg, line 26: Btw i was never sure why we `map(seed => ({ ...seed }))`, do you know ?

server/graphql/general/src/queries/generate_inbound_return_lines.rs
```diff
@@ -1,43 +1,75 @@
...
+    pub item_id: String,
+    pub return_id: String,
+}
 
 #[derive(InputObject, Clone)]
-pub struct GenerateInboundReturnInput {
-    pub stock_line_ids: Vec<String>,
+pub struct GenerateInboundReturnLinesInput {
+    /// The ids of the outbound shipment lines to generate new return lines for
+    pub outbound_shipment_line_ids: Vec<String>,
```
andreievg, line 24: In the PR description you have: `or returning existing return lines for a specific itemId.`, that must be in the follow up pr right ?

andreievg: Scrap that, i didn't see the existing_line_input thing, thanks

server/service/src/invoice/inbound_return/generate_lines.rs
andreievg: > If you want show the existing lines for a particular item, you're going to need to include both the item and return ids

But, there will be an option to add outbound_shipment_line_id to existing return..

andreievg: I think maybe the `id_type` and `ids`, might be the way to go ?

andreievg (approved):

andreievg (approved):
Looks good thank, just that one question about the input shape, and also i think we can consolidate how inbound line is generated from invoice_line, although i tried and it looks like its more code: https://github.com/msupply-foundation/open-msupply/compare/3155-generate-inbound-return-lines...3155-generate-inbound-return-lines(use-Consolidate-generating-inbound-line), so not sure, up to you: ¯\_(ツ)_/¯

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
lache-melvin: Oh haha - I've seen it sometimes to create a new object before mutating, but in this case I think I removed some additional fields and never cleaned this up 😅

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
lache-melvin: Like this? https://github.com/msupply-foundation/open-msupply/compare/3155-generate-inbound-return-lines...3155-generate-inbound-return-lines(direct-data-call)

server/service/src/invoice/inbound_return/generate_lines.rs
lache-melvin: Yeah you might be right, I don't think there is a use case where we would supply both the outbound_shipment_line_id and the item_id? I'm kind of keen to get this in to unblock other work - perhaps we can revisit when we do the add to existing return piece?

Also just flagging here - if we want to determine whether an outbound_shipment_line_id is already in a return, we'll need to add a `linked_invoice_line_id` to InvoiceLineRow. With inbound_shipment/outbound_return we could check the stock_line_id, but with outbound_shipment/inbound_return we can't do this as new stock lines will be generated.

lache-melvin:
> Looks good thank, just that one question about the input shape, and also i think we can consolidate how inbound line is generated from invoice_line, although i tried and it looks like its more code: [3155-generate-inbound-return-lines...3155-generate-inbound-return-lines(use-Consolidate-generating-inbound-line)](https://github.com/msupply-foundation/open-msupply/compare/3155-generate-inbound-return-lines...3155-generate-inbound-return-lines(use-Consolidate-generating-inbound-line)), so not sure, up to you: ¯_(ツ)_/¯

Oo I like this! Will merge

server/service/src/invoice/inbound_return/generate_lines.rs
andreievg: > Also just flagging here - if we want to determine whether an outbound_shipment_line_id is already in a return, we'll need to add a linked_invoice_line_id to InvoiceLineRow

Yep agree

> unblock other work

Sure, would be great to see how it looks with the 'id_type' and 'id', which would translate to `enum ItemOrOutboundLineId { ItemId(String), OutboundLineId(String) }`

client/packages/invoices/src/Returns/modals/InboundReturn/useDraftInboundReturnLines.ts
andreievg: Yeah, that's much better i think, although at this point, why tanstack ¯\_(ツ)_/¯.

I made a couple of comments here: https://github.com/msupply-foundation/open-msupply/commit/1a928782a5a1f571ba74f0a0ee0b66ba1276812b

PR #3197, opened 2024-03-06
Disable the modal save button when creating an encounter

### LINKED ISSUE #3157: Multiple encounters created when we click on OK button while creating encounter repeatedly
## What went wrong? 😲
Multiple encounters created when we click on OK button while creating encounter repeatedly

## Expected behaviour 🤔
Create only one entry of encounter after a click on OK button and discard other continuous clicks? what if system hangs (not sure how often this can be) and you end up clicking multiple times, only to end up with multiple unnecessary entries

## How to Reproduce 🔨
1. Enrol a patient to a program
2. Create encounter for the patient
3. Add date and notes if any
4. Now, repeatedly click on OK button mutliple times. I clicked once, for some reason my system hanged and I ended up clicking multiple times
5. See error: the number of encounters created = no of times you clicked on that OK button

## Your environment 🌱

- Version: v1.7.0.rc1 exe
- mSupply version: v713-beta14

- Platform: browser

- Database type: sqlite

## Additional
Noticed while testing #2925

clemens-msupply:
Fixes #3157 

# 👩🏻‍💻 What does this PR do? 
Disable the modal save button when creating an encounter

I couldn't reproduce it myself. @nisha-dangol ?

### DIFF
--- a/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
+++ b/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
@@ -86,6 +86,7 @@ export const CreateEncounterModal: FC = () => {
   const navigate = useNavigate();
   const { error } = useNotification();
   const [startDateTimeError, setStartDateTimeError] = useState<string>();
+  const [isCreating, setIsCreating] = useState(false);
 
   const handleSave = useEncounter.document.upsert(
     patientId,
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
I can't reproduce either... but approving since code looks good 👀 
Thanks Clemens...

Issue #3193, opened 2024-03-05
Asset location editing


mark-prins:
Part of epic #2896 

## Is your feature request related to a problem? Please describe 👀
[Figma designs](https://www.figma.com/file/rIK2zVFUVhtbuXo0RBzOzw/0033-2305_omSupply-ColdChain-UI?type=design&node-id=2653-16424&mode=design&t=8KRUuAo7s1V0r0RM-0)

The asset detail page has the ability to edit locations which are internal to the asset. This is a new data type, and input type with the Mui Chips / tag style interface and the ability to add locations

![image](https://github.com/msupply-foundation/open-msupply/assets/9192912/74c6b7b8-f3f5-4950-a3b1-b670b755e445)

![image](https://github.com/msupply-foundation/open-msupply/assets/9192912/45b0597a-e665-4128-870b-1836d33330b7)

## Describe the solution you'd like 🎁
We need:
- [x] Service layer for `asset_internal_location`
- [x] Query and mutations for `asset_internal_location`
- [x] A loader would be helpful, then we could return locations with the main asset query
- [x] Similarly, the `updateAsset` mutation: can we provide a list of location ids for this?
- [ ] A UI control to add and remove locations, picking from a list
- [ ] Some way (?) to add a location from this control too, and once added, add to the list for that asset

### Moneyworks Jobcode 🧰

`OMSCC:AST`

adamdewey:
@mark-prins @jmbrunskill - just wondering what's meant by 'new data type' here?

Are we still planning to link standard stock storage locations in the section that says `Cold storage location` ?

The summary of the chat we had in the CCEI product Telegram chat was: 

- We would use Store to track where assets are located
- We would use Location(s) to track what is located inside assets

That's still my current understanding - just wanted to check I hadn't missed anything :)

jmbrunskill:
@adamdewey @mark-prins my intention is for the `asset_internal_location` table to just contain the link between an existing location and the asset. Because we want to allow multiple locations inside an asset, we can't just add an `internal_location_id` to the asset record, we need another table capture the links.
(It could be a json field or something, but I think this is clearer)

adamdewey:
Ah thanks @jmbrunskill - so I understand it's a new technical data type, not a new location concept - that's all good, cheers! 😅

PR #3200, opened 2024-03-06
Asset summary tab

### LINKED ISSUE #2984: Asset Detail: Summary tab
Part of epic #2896 

image

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #2984

# 👩🏻‍💻 What does this PR do? 
 
image

This connects the update functionality to the asset, including
* Save button enables only when there's a change made
* A confirmation prompt is shown when you click save
* A confirmation prompt is shown if you navigate away from an edited page
* Validation is shown for the date fields (but this doesn't affect the save button validity)
* Error toast is displayed if there was a problem
* Success toast is save was ok

#3193 has been created to handle the location editing
The Status display and update I've left, will need another issue for that one.

## 📃 Documentation

- [ ] Part of the asset editing


adamdewey:
I think we might need to include the asset code in the Asset Identification section (as well as in the breadcrumbs at the top of the page)?

Perhaps at the start of the section?

image

Just imagining that a lot of places will likely be using manual stickers with human readable asset codes to identify assets

jmbrunskill (approved):
LGTM, I notice we don't show the name anywhere on the summary page, I wonder if we should? Or maybe we don't need name and only a code? Question for Richard I guess...

We also need somewhere we can assign a store? Or maybe we should default to the current store, and only not set the store on central server?

adamdewey:
> We also need somewhere we can assign a store? Or maybe we should default to the current store, and only not set the store on central server?

@jmbrunskill , @richardmoizeau , @mark-prins 

I think this highlights that there could be 2 big models of user functionality for asset management - centralised and decentralised. 

UNICEF are very much pushing a decentralised model for CCE - they want facilities to be able to capture data about their own assets on site, and they want to make it as easy as possible for the users, to the point that they want them to scan a barcode rather than enter any manual data themselves (of course would still need option to enter manual data for assets that don't have barcodes). 

For those local workers it wouldn't make sense to be able to choose a store as they are only concerned with their own facility. 

However for a more decentralised model, where an authority figure is assigning assets to multiple facilities then they would need a mechanism to be able to select which store the asset is in. 
Technically if they have all Store permissions they could log into each store and add the assets that way, but could be a bit tedious - would think they would prefer some kind of spreadsheet import functionality at a central level that then gets synced down to individual Stores?

adamdewey:
> LGTM, I notice we don't show the name anywhere on the summary page, I wonder if we should? Or maybe we don't need name and only a code? Question for Richard I guess...

In my (limited) experience with working with assets at companies I don't recall a 'name' being associated with an asset, just a code, type and location - but it could be useful to have a notes section somewhere where some more human info could be stored? 

E.g. `Fridge with the blue sticker - behind the vaccine shelves in the lab. Fred currently using it to store insulin 03/24`

adamdewey:
Interestingly the WHO IGA tool did include a name for assets, but you can see from this sample data that no-one ever used it...

image

mark-prins:
So to summarise, this is what I'm thinking. sound ok?
- the store which an asset belongs to is the current store / when creating an asset it is assigned to the current store and the user doesn't have an option to change it
- if you are editing the asset while on the central server though, you can change the store
- you can allocate a store when importing from a spreadsheet, this allows the multi-store admin to set up a number of assets & stores easily
- the 'name' field should be renamed to 'notes' and added to the summary page (RHS perhaps, under the status). multi-line input field

richardmoizeau:
All good for me !

adamdewey:
Nice @mark-prins 

The only thing I'm wondering about - is how would this work: 

> if you are editing the asset while on the central server though, you can change the store

Will there be a special 'central asset management' Store where you log in and can see all assets? 

Because I'm just thinking - if you already have to be in a store to do the editing, what happens when you change the store of the asset, does it disappear when you refresh the page?

jmbrunskill:
> Will there be a special 'central asset management' Store where you log in and can see all assets?

I think their maybe should be a central management store, which you need to login to edit things like asset cataglogues, pack variants etc. Right not we're relying on being on the central server, but I think having a dedicated store makes things clearer for the user. I think it needs a wider discussion though. This is a great example.

adamdewey:
Yeah I agree. 

Although I do think there is a significant difference in use case between centrally managing Master Data for stock inventory / asset catalogues and centrally managing asset distribution. 

Ideally asset distribution would have a targeted vertical, in the same way that we have figured out the vertical for stock inventory distribution. 

An efficient asset distribution vertical would eliminate the need for the distribution to be centrally administered (conversely you _could_ say stock inventory distribution could be centrally administered, but we would now think that's ridiculous)

I can see there is frustration in the community about how difficult it is to manage their asset distribution - we have a chance to improve things here. 

Long story short: if (when) we have a good vertical mapped out for asset distribution it wouldn't make sense to have an editable (or even viewable) Store value in the Asset Summary / Detail pages for users operating Open mSupply inside a Store (the users would only care about their own assets which they know is in their current Store). 

I'll bring this up in the Telegram chat to discuss further!

Issue #1957, opened 2023-07-03
Improve (reduce) Bundle size


andreievg:
Improve bundle size:

From a quick view:

* Should barcode scanner be that big for electron and isn't that electron only code (should be part of native bundle? or is a lot of barcode scanning/parsing is actually in web bundle ?)
* x-date-pickers twice ?
* reduce-css-calc sounds like dev dependency ?
* I think we using full lodash again (it should be 24kb gzipped full, but it's around 27kb gzipped ?)

![Screenshot 2023-07-03 at 1 00 02 PM](https://github.com/openmsupply/open-msupply/assets/26194949/8349c997-1484-45b0-9383-475dad3dfbf1)

andreievg:
Captured in this [tracker](https://github.com/msupply-foundation/open-msupply/issues/8810), to be re-opened when scheduled

PR #3202, opened 2024-03-06
Reduce bundle size

### LINKED ISSUE #2813: Review f/e bundle size
## What went wrong? 😲
 
What's up with the bundle size? It was much lower than this. Has something blown the size out?

image

## Expected behaviour 🤔
A more reasonable size than 8MB. At the least some confidence that all of that is required.

mark-prins:
Fixes #2813 

# 👩🏻‍💻 What does this PR do? 
 
The required change is to remove common package from the module federation shared bundles. The intention was that each package could use the federated version, the unintended consequence is that weback is unable to tree shake because it doesn't know what is required or not from that bundle. The entire common package was bundled up, all 8MB of it.

Tracked down to [this commit](https://github.com/msupply-foundation/open-msupply/pull/2568/commits/c52b31970bf730e0c9e5402e5759b6a9a21418b9#diff-b142eb1f4548423b3c912cf3b5570e8d2d5a823f5eb09da7607cfb845608e130)

The PR action should show if this works or not. Otherwise you could run the bundle analyzer plugin and see what size you get

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/coldchain/src/index.ts
+++ b/client/packages/coldchain/src/index.ts
@@ -1,3 +1,3 @@
 export { ColdchainService } from './ColdchainService';
 export { ColdchainNotification } from './ColdchainNotification';
-export * from './common';
+export { parseBreachType } from './common';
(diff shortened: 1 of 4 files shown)


mark-prins:
huh? 500 KB?! that's less than I was expecting.
When running bundle analyzer the size was halved

jmbrunskill (approved):
Looks good to me, doesn't seem to have broken anything obvious

mark-prins:
Hoping that we get more of a reduction with #1957

PR #3216, opened 2024-03-07
Rename TestSync Types to better reflect how they're used

jmbrunskill:
Fixes #....

# 👩🏻‍💻 What does this PR do? 

There are/were two types used in sync testing.
`TestSyncPullRecord` and  `TestSyncPushRecord`
However with the recent changes to central open-mSupply server these names are a bit confusing and possibly miss leading?

`TestSyncPullRecord` has been renamed to `TestIncomingSyncRecord`
`TestSyncPushRecord` has been renamed to `TestOutgoingSyncRecord`

`TestIncomingSyncRecord` represents a record that's come in via a sync process (pull or push)
`TestOutgoingSyncRecord` represents a record that will be sent via sync

Tests pass

Rename suggested by @andreievg maybe there's a better naming, or more changes that should be included with this?

> Doing this as a separate PR to avoid a huge PR with logic and renaming changes.

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -2498,7 +2498,7 @@ dependencies = [
  "graphql_types",
  "repository",
  "reqwest",
- "serde 1.0.196",
+ "serde 1.0.197",
  "serde_json",
  "service",
  "thiserror",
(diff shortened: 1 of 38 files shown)


mark-prins:
the From / To wouldn't be my first choice, but agree - it's better than Pull & Push when either of those could be an incoming record!

I would use either 
* Incoming / Outgoing
* Sent / Received

mark-prins (approved):
👍 LGTM! and all sync tests passing

jmbrunskill:
I like `Incoming` / `Outgoing` I'll update to that...

PR #3223, opened 2024-03-07
Middleware to reject incoming sync connections if not central server

jmbrunskill:
# 👩🏻‍💻 What does this PR do? 
Adds a middleware to check if we are central server.
Suggested by @andreievg while working on central server sync work.

Used postman, and checked requests aren't blocked

This should reduce the need to call is_central server in other parts of the code base, however I haven't removed the exiting check.

Note old method of returning an error actually returned a `200` status code, I think this is more correct? 

## 📃 Documentation

_No user facing changes_


mark-prins (approved):
😄 Looks good to me - unsure how to test, so I've stopped at building the server to see if nothing broke. Will trust your testing on this one

jmbrunskill:
> 😄 Looks good to me - unsure how to test, so I've stopped at building the server to see if nothing broke. Will trust your testing on this one

For future testers and such...
You can test using postman and a post to a url like this: http://localhost:<CENTRAL_SERVER_PORT>/central/sync/pull
On central server you should get a sync error (or success if you post a valid request)

On remote server you should get this..

image

Probably most important thing to test is that nothing else accidentally breaks on the remote server like downloading reports or something. Pretty confident though.

PR #3228, opened 2024-03-07
Rename name

### LINKED ISSUE #3226: Rename name column to notes
Further to [this comment](https://github.com/msupply-foundation/open-msupply/pull/3200#issuecomment-1981724003)
rename the `name` field / column / property to be `notes`

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #3226

# 👩🏻‍💻 What does this PR do? 
 
Rename the `name` field to `notes`

I've removed the `name` (now `notes`) option from the filter, because I think that's best. It's still available as a filter though:

Screenshot 2024-03-07 at 6 17 54 PM

Have updated the default sort, and the default filter options.

you may need to update your database manually if using an existing database:

```
ALTER TABLE public.asset RENAME COLUMN "name" TO notes;
ALTER TABLE public.asset ALTER COLUMN notes DROP NOT NULL;
```

From here, see if the site still runs, that the notes are shown, that the export works and the update of an asset.
The create asset modal is updated with the notes label too.


jmbrunskill (approved):
Looks good to me!

PR #3232, opened 2024-03-07
Add option to make text fields required in the UI schema

### LINKED ISSUE #3126: Should there be mandatory fields for patient creation?
## What went wrong? 😲
Currently, we can create a patient with no fields at all. 

## Expected behaviour 🤔
Don't allow if some mandatory fields are not there?

## How to Reproduce 🔨
1. Create new patient
2. GIve a name and nothing else and search
3. Just OK&Next until you land on Patient info
4. There, delete even the name you gave for search
5. See error: the Create button is still enabled (I think it should have been disabled by now)
6. Save it. And voila, you've a no-name no nothing patient
image

## Your environment 🌱

- Version: v1.6.2 exe
- mSupply version: 
- Platform:
  - server (windows)
- Database type: 
  - SQLite3

## Additional
Noticed while testing #2925

clemens-msupply:
Fixes #3126

# 👩🏻‍💻 What does this PR do? 
- Adds an option to make a text field required in the UI schema.
- Use this feature to make first and last name required (the base schema still has them as optional)

Create a new patient or move to an existing patient. Then delete first and/or last name. There should be an error.

- Make both, first and last, required or just one of them?

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -17,6 +17,8 @@ const Options = z
      * Additional pattern to be matched that can be defined in ui schema
      */
     pattern: z.string().optional(),
+    /** Overwrite the schema definition, e.g. to make a non-required field required in the UI. */
+    required: z.boolean().optional(),
     /**
      * Examples for the correct pattern
      */
@@ -71,7 +73,7 @@ const usePatternValidation = (
   const { customError, setCustomError } = useJSONFormsCustomError(path, 'Text');
 
   useEffect(() => {
-    if (!pattern || !value) {
+    if (!pattern || value === undefined) {
       setCustomError(undefined);
       return;
     }
@@ -81,7 +83,7 @@ const usePatternValidation = (
     } else {
       setCustomError(undefined);
     }
-  }, [pattern, value]);
+  }, [pattern, setCustomError, value]);
   return customError;
 };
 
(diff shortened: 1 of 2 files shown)


client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -136,7 +147,6 @@ const UIComponent = (props: ControlProps) => {
         FormHelperTextProps: error
           ? { sx: { color: 'error.main' } }
           : undefined,
-        required: props.required,
```
clemens-msupply, line 139: I first added it here as well but then it shows a "required *" which we don't show anywhere else. Think just marking the text box as red is clear enough to indicate that the field is required.

client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -71,7 +73,7 @@ const usePatternValidation = (
   const { customError, setCustomError } = useJSONFormsCustomError(path, 'Text');
 
   useEffect(() => {
-    if (!pattern || !value) {
+    if (!pattern || value === undefined) {
```
clemens-msupply, line 76: This was a bug that I noticed along the way... i.e. an empty string `!string` is evaluating to true.

Note, we shouldn't really do these `!!` or `!` checks but do a proper check against `undefined` instead. This is much clearer, and you don't have to worry about truthy and falsy conversions...

mark-prins: re the `!!` or `!` sometimes this is valid. I have had cases where I deliberately want to check `null` `undefined` or `` and prefer to have the terse code. You are right though, it does require the coder to know about the js truthy and falsy conversions which have tripped people up. Is it my responsibility that other coders don't know things though? should I write longhand code just in case?

client/packages/programs/src/JsonForms/common/components/Text.tsx
mark-prins: agree - best to be consistent. would be good to have a required indicator throughout, but we do need to discuss and co-ordinate before doing that!

client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -17,6 +17,8 @@ const Options = z
      * Additional pattern to be matched that can be defined in ui schema
      */
     pattern: z.string().optional(),
+    /** Overwrite the schema definition, e.g. to make a non-required field required in the UI. */
```
mark-prins, line 20: does this mean: 'Add a required property to allow text fields to be mandatory in the UI'?
It took me some time to understand what the comment was saying.

mark-prins (approved):
Nice! The required indication is showing - the empty text field has a red border.
A question though - the form still allows me to save the empty value. Is that correct? I would have expected that the validation would prevent me from saving

client/packages/programs/src/JsonForms/common/components/Text.tsx
clemens-msupply: Personally I use it to also make clear what my indention was, e.g. to avoid wondering if I didn't thought about the truthy/falsy case or if it was on purpose. I would even do this `if (value !== undefined && value !== '')` for `value: string | undefined`, i.e. first narrow the type down to a `string` and then compare the string.

client/packages/programs/src/JsonForms/common/components/Text.tsx
mark-prins: that's a good point - having optional data be required in the UI is inconsistent. Better to just change the requirement in DefaultPatientSchema

clemens-msupply:
> Nice! The required indication is showing - the empty text field has a red border. A question though - the form still allows me to save the empty value. Is that correct? I would have expected that the validation would prevent me from saving

Thanks, no this is not correct. I fixed it using custom errors.

I always thought having this `required` option in the UI would be a nice option, but now I am not sure if this is really a needed option? I could just make the default data schema required... (see other comment)

mark-prins:
> I always thought having this required option in the UI would be a nice option, but now I am not sure if this is really a needed option? I could just make the default data schema required... (see other comment)

Oh, true! I would try making it required in the schema - sounds like a nice simple option

clemens-msupply:
Close this one in favour of: https://github.com/msupply-foundation/open-msupply/pull/3244

PR #3234, opened 2024-03-07
3094 Pricing tab in invoice lines has a confusing column selection

### LINKED ISSUE #3094: Pricing tab in invoice lines has a confusing column selection
## What went wrong? 😲

image

The calculation for cost price here isn't intuitive, because we're missing information. Reading from left to right you are lead to believe that `$0.9 * 300 = $2.70` clearly wrong. Price is per pack, and pack size in this case is 100. Users should not have to intuit that the presented column is not the one being used in the calculation and to do some head algebra to deduce the relevant fields 😁.

There is a real struggle here, as screen real estate is precious and we often want to show _a lot_ of columns. That said, I think there's room here?

## Expected behaviour 🤔

After discussing internally with several stakeholder, more columns endorsed by @Gwalior-84 
- Batch
- No. of packs
- Units per pack
- Cost price per pack
- Total (cost) price

In Mark's context sell price per pack is rarely of value. Maybe add it if not too cramp, otherwise leave off for now till requested 🙂 .

## How to Reproduce 🔨

1. View an invoice line and go to pricing tab.

## Your environment 🌱
- Version: 1.7 dev (probably 1.6 prod though)
- DB: SQLite3

roxy-dao:
Fixes #3094

# 👩🏻‍💻 What does this PR do? 
As per requested, have removed sell price from the inbound pricing tab and added pack qty and number of packs. Have shifted these around from request since it made more sense to me to see number of packs x cost price rather than pack qty x cost price.
Screenshot 2024-03-08 at 10 42 02

- [ ] Go to the pricing tab in Inbound Shipment to see changes


mark-prins (approved):
🍾  perfect, thanks!

Can I have an extra request please.. 🙏 
There's a console error in the stock item search component, 

in `getItemOptionRenderer`
can you add the key please?
```
    <Tooltip title={`${item.code} ${item.name}`} key={item.id}>
```

regotaina:
Tested on RC3. PASSED the test 👍🏾  

Screenshot 2024-05-08 at 3 37 35 PM

PR #3235, opened 2024-03-08
#3013 Outbound return SidePanel

### LINKED ISSUE #3013: Outbound returns detail view -- SidePanel

Add SidePanel to page -- contents to be determined/discussed?

CarlosNZ:
Fixes #3013

# 👩🏻‍💻 What does this PR do? 
 

Thought I'd get this in quickly before I finish for the day.

Bare bones side panel here, basically copied the elements you'd included in your one @lache-melvin :)

Side panel opens and closes -- not much else :) 

See inline


client/packages/invoices/src/Returns/DetailView/SidePanel/AdditionalInfoSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+  //   const { colour, comment, user, update } = useOutbound.document.fields([
+  //     'colour',
+  //     'comment',
+  //     'user',
+  //   ]);
+  //   const [colorBuffer, setColorBuffer] = useBufferState(colour);
+  //   const [commentBuffer, setCommentBuffer] = useBufferState(comment ?? '');
+
+  //  TO-DO: Make this work
+  const { user } = data || {};
```
CarlosNZ, line 29: I copied this from your one, but there's no `user` node on either of our ReturnByNumber queries (that I could tell). So not sure how you got the user data to be a valid type. Can probably figure out before merging, just wanted to rush this out the door for now.

lache-melvin (changes requested):
Looks great Carl :) Just the user part of the query to update, up to you whether to add the delete now or later 🚀

CarlosNZ:
@lache-melvin Righto, have updated the `user` node in the query. I see you've made an issue for delete lines mutation, so will leave that for now.

lache-melvin (approved):
Ah yep that issue is actually to delete lines within the return, not delete the return itself, but all good to get this in and do that separately!

PR #3236, opened 2024-03-08
Comment out middleware

roxy-dao:
Middleware breaking things o/


jmbrunskill (approved):
Thanks for finding, I'll looking to why this is happening and put in the real fix...

PR #3237, opened 2024-03-08
3181 insert inbound return

### LINKED ISSUE #3181: API: insertInboundReturn
Epic #2897 

- Create an abstraction for StockIn lines
- Use this abstraction for inserting inbound returns
- Create a refactor issue to use this abstraction for Inbound

lache-melvin:
Fixes #3181

# 👩🏻‍💻 What does this PR do? 
 

Adds API for inserting an inbound return.

Creates the StockIn abstraction, similar to StockOut. Uses this abstraction for inserting the inbound return. 

Automated tests added for service layer.

- From Outbound Shipment detail view (at least delivered status)
- Select lines and click the process return of selected lines action
- Set return quantities, and optionally reasons/notes
- Submit
- You'll get redirected, but the detail view isn't merged currently...
- You can query for the invoice in a graphql playground though!


client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -80,7 +78,7 @@ export const InboundReturnEditModal = ({
         }
         okButton={
           currentTab === Tabs.Reason ? (
-            <DialogButton onClick={onOk} variant="ok" disabled={!okEnabled} />
```
lache-melvin, line 83: we have reasons as optional in the backend, so have removed the constraint here 😅

client/packages/invoices/src/Returns/modals/ReturnReasonsTable.tsx
```diff
@@ -46,6 +46,7 @@ export const ReturnReasonsComponent = ({
         Cell: TextInputCell,
         width: 200,
         setter: updateLine,
+        accessor: ({ rowData }) => rowData.note ?? '',
```
lache-melvin, line 49: fixes a console warning (didn't like it being null)

client/packages/invoices/src/utils.ts
```diff
@@ -226,6 +226,7 @@ export const canReturnInboundLines = (inbound: InboundFragment): boolean =>
   inbound.status === InvoiceNodeStatus.Verified;
 
 export const canReturnOutboundLines = (outbound: OutboundFragment): boolean =>
+  outbound.status === InvoiceNodeStatus.Shipped ||
```
lache-melvin, line 229: If a manual outbound shipment, it never goes beyond shipped, but the supplier should be able to process the return of those goods!

mark-prins: likely worth adding the check in then - only if manual then allow `Shipped` to be returned. As it is, a customer can return stock that is in transit 🤔

mark-prins (approved):
had a read through and have run the tests - looks fine to me 🙏 
would be good if someone involved or someone with more rust experience could comment!

client/packages/invoices/src/utils.ts
lache-melvin: Curious for more thoughts on this one... this is for processing the return of goods from an outbound shipment - i.e. a supplier saying yes I shipped these goods but now they are being returned. I'd guess that most of the time, inbound returns would be generated from outbound returns (created by the customer), but where you did want to create one manually, would it be worth constraining that you can't do that until the customer marks that shipment as delivered?

PR #2638, opened 2023-12-14
Display current item in OS and Stocktake when you have many items in the system

mark-prins:
Fixes #2614

# 👩🏻‍💻 What does this PR do? 
 
The current item i.e. the selected value in the autocomplete, is picked from the list of available items which is fetched with the `useItemStockOnHand` hook. If you have more than 500 items available, and the current item is further than position 500 when the list is sorted by name, then that item won't be in the first set of paginated results.

This means that the item won't be shown in the disabled autocomplete component.

The fix here is to also fetch the current item by id, and add that to the list of items. This list is already unique by id, so we won't be adding a duplicate. When editing a stocktake or OS line, the selected items aren't changeable, so the input will be disabled when showing the current item. This means that the position in the list is also not important.

# 🧪 How has/should this change been tested? 

You can edit `packages/system/src/Item/Components/StockItemSearchInput/hooks.ts` and change the page size to something smaller than your available items, or you can use a large dataset, such as that linked to in the issue.

Then, create an OS or stocktake with an item that starts with a letter low down in the alphabet (V? Z?)
Refresh the page, open the edit modal again. You will see the item name.

## 📃 Documentation

_No user facing changes_

PR #3238, opened 2024-03-08
3112 fix sort direction

### LINKED ISSUE #3112: Unable to change sort direction in lists
## What went wrong? 😲
 
Click on a column header to sort, click again to change the direction and nothing happens
Not on all pages though. So far has been confirmed on:

* outbound shipment list view
* requisition list view
* internal order list view 

## Expected behaviour 🤔
Should be able to change direction

## How to Reproduce 🔨

1. Go to any table / list view
2. Click on column to sort
3. Click on the same column again

## Your environment 🌱

- Version: 
- mSupply version:

clemens-msupply:
Fixes #3112 

The bug has been introduced here:
https://github.com/msupply-foundation/open-msupply/pull/2638/files#r1426169816

The problem was that `asc` is mapped to an empty query string. This means when there was no sort direction because of that, the initial direction was always used.

# 👩🏻‍💻 What does this PR do? 
- Fixes some lint warnings (first commit)
- Fixes the sort problem

Note: the initial sort direction is not used now...

### DIFF
--- a/client/packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
+++ b/client/packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
@@ -191,7 +191,7 @@ describe('sort', () => {
       wrapper: getWrapper(),
     });
     act(() => {
-      result.current.sort.onChangeSortBy(quantityColumn);
+      result.current.sort.onChangeSortBy(quantityColumn.key, 'asc');
     });
     expect(result.current.sort.sortBy).toEqual({
       key: 'quantity',
@@ -206,7 +206,7 @@ describe('sort', () => {
       wrapper: getWrapper(),
     });
     act(() => {
-      result.current.sort.onChangeSortBy(idColumn);
+      result.current.sort.onChangeSortBy(idColumn.key, 'desc');
     });
     expect(result.current.sort.sortBy).toEqual({
       key: 'id',
@@ -215,7 +215,7 @@ describe('sort', () => {
     });
   });
 
-  it('has the correct values after triggering a few sort bys in sequence', () => {
+  it('has the correct values after sorts', () => {
     const idColumn = createColumnWithDefaults<TestSortBy>({ key: 'id' });
     const quantityColumn = createColumnWithDefaults<TestSortBy>({
       key: 'quantity',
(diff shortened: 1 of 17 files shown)


client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
```diff
@@ -70,18 +70,24 @@ export const useUrlQueryParams = ({
...
+      const sort = column.key as string;
+      if (sort !== currentSort) {
+        // change sort key
+        updateQuery({ sort, dir: '', page: '' });
+      } else {
+        // toggle sort direction
+        const dir = column.sortBy?.direction === 'desc' ? '' : 'desc';
+        updateQuery({ dir });
+      }
+    },
```
clemens-msupply, line 88: I added some comments here, are they correct?

@mark-prins I think this method should not do the toggling but just update the sort order, i.e. instead of taking a Column it should just take a SortBy. So this hook would really just update the query params for page, filter and sort, but doesn't know about columns. The toggling should IMHO happen in HeaderCell. If you agree I can put this in.

mark-prins: yes, comments correct thanks.
The sort is only called from the header cell, which is already typed with the Column: makes sense to push anything column specific down into that component and keep this function clean. 
So yes, if you don't mind - would be good to have that.

mark-prins (approved):
🎉  thanks - changes are good. would like to follow up on your suggestion re header cell too though 🙏

clemens-msupply:
@mark-prins I made change so that the toggle happens only in the HeaderCell. It touched a couple of files and also useQueryParamsStore where I did similar changes. Could you please have another look?

client/packages/invoices/src/InboundShipment/api/hooks/line/useInboundItems.ts
```diff
@@ -28,5 +28,5 @@ export const useInboundItems = () => {
       );
   const { data } = useInboundSelector(selectItems);
 
-  return { data, sortBy, onSort: onChangeSortBy };
```
clemens-msupply, line 31: I only removed it to make sure it wasn't used anywhere...

client/packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
```diff
@@ -225,17 +225,27 @@ describe('sort', () => {
...
+    expect(result.current.sort.sortBy).toEqual({
+      key: 'id',
+      isDesc: true,
+      direction: 'desc',
+    });
+
+    act(() => {
       // should be: quantity/desc
-      result.current.sort.onChangeSortBy(quantityColumn);
+      result.current.sort.onChangeSortBy(quantityColumn.key, 'desc');
```
clemens-msupply, line 248: This test didn't make too much sense anymore so I changed it a bit...

mark-prins (approved):
Looks great - thanks @clemens-msupply 

I've found one extra minor bug, would you mind adding this in too?

In `client/packages/invoices/src/OutboundShipment/api/api.ts` can you add the case for 'theirReference' as well. Currently you can click this header to sort, but the sort key used is 'status'

added at line 51:
```
      case 'theirReference': {
        return InvoiceSortFieldInput.TheirReference;
      }
```

clemens-msupply:
> added at line 51:
> 
> ```
>       case 'theirReference': {
>         return InvoiceSortFieldInput.TheirReference;
>       }
> ```

done thanks, sure we can make it typesafe somehow but maybe another time :)

PR #3240, opened 2024-03-08
Fix central_server_only middleware wrap

jmbrunskill:
# 👩🏻‍💻 What does this PR do? 
Moves the `central_server_only` middleware to be called inside a `Scope` otherwise it appears to apply more widely than I realised

See if you can login :)

## 📃 Documentation

_No user facing changes_


roxy-dao (approved):
Thanks! Working fine now... I'm not sure how to set up central server so can't test the rest, but code looks fine!

PR #3241, opened 2024-03-10
3204 add enum of reasons

### LINKED ISSUE #3204: Add enum of reasons to asset log
Hard coded list of reasons mapped to a particular status:

See spreadsheet for details:

https://docs.google.com/spreadsheets/d/1l1x38KT5cdt4Zd5piMWp3FuBuiSUv4faMHP8utDJDqU/edit#gid=2113323501

### Moneyworks Jobcode 🧰

OMSCC:AST

fergie-nz:
Fixes #3204

# 👩🏻‍💻 What does this PR do? 
Adds enums for both reason and status
Adds validation for ensuring reason matches status

Adds tests in service layer


server/service/src/asset/validate.rs
```diff
@@ -27,3 +28,40 @@ pub fn check_asset_log_exists(
 pub fn check_user_is_user(ctx: &ServiceContext, input: &InsertAssetLog) -> bool {
     return ctx.user_id == input.user_id;
 }
+
+pub fn check_reason_matches_status(status: &Option<Status>, reason: &Option<Reason>) -> bool {
+    return match status {
```
fergie-nz, line 33: This match statement felt a bit clumsy. I wondered if there was a better way to ensure reason matching status. Especially if reasons or status change or are dynamic? But I wasn't sure

mark-prins: here's my suggestion - the None cases both return true, so I think get that out of the way first. 

```
pub fn check_reason_matches_status(status: &Option<Status>, reason: &Option<Reason>) -> bool {
    let status = match status {
        Some(status) => status,
        None => return true,
    };

    let reason = match reason {
        Some(reason) => reason.to_owned(),
        None => return true,
    };

    match status {
        Status::NotInUse => {
            reason == Reason::AwaitingDecomissioning
                || reason == Reason::Stored
                || reason == Reason::OffsiteForRepairs
                || reason == Reason::AwaitingInstallation
        }

        Status::FunctioningButNeedsAttention => {
            reason == Reason::NeedsServicing || reason == Reason::MultipleTemperatureBreaches
        }
        Status::NotFunctioning => {
            reason == Reason::Unknown
                || reason == Reason::NeedsSpareParts
                || reason == Reason::LackOfPower
        }

        _ => false,
    }
}
```

mark-prins (approved):
👍  Looks good - thanks Ferg!

server/graphql/asset/src/types.rs
```diff
@@ -241,13 +267,111 @@ impl From<AssetLogFilterInput> for AssetLogFilter {
...
             log_datetime: f.log_datetime.map(DatetimeFilter::from),
             user: f.user.map(StringFilter::from),
         }
     }
 }
 
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")] // only needed to be comparable in tests
+
+pub enum ReasonInput {
```
andreievg, line 280: Is this too generic as a name for graphql type ? What if we have other reasons, or statuses for that matter

server/repository/src/db_diesel/assets/asset_log_row.rs
```diff
@@ -1,36 +1,97 @@
...
+        reason -> Nullable<crate::db_diesel::assets::asset_log_row::ReasonMapping>,
         log_datetime -> Timestamp,
     }
 }
 
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+
+pub enum Reason {
```
andreievg, line 30: Similar to names, I think the naming is too generic please reference other 'status/type' enums

server/repository/src/db_diesel/assets/asset_log_row.rs
```diff
@@ -1,36 +1,97 @@
...
+    AwaitingInstallation,
+    Stored,
+    OffsiteForRepairs,
+    AwaitingDecomissioning,
+    NeedsServicing,
+    MultipleTemperatureBreaches,
```
andreievg, line 36: These look very cold chain specific are they not customer defined/configured ?

fergie-nz: I agree it feels a bit arbitrary and customer specific. I wonder if it would be better to leave it as a configurable enum and remove the matching-reason-to-status validation. 

@richardmoizeau do you have any thoughts on if the reasons and status's in https://docs.google.com/spreadsheets/d/1l1x38KT5cdt4Zd5piMWp3FuBuiSUv4faMHP8utDJDqU/edit#gid=2113323501 would be configurable?

mark-prins (changes requested):
Can you update the `asset_logs.rs` migration please. there's an issue running the query with postgres

```
    const REASON_ENUM_TYPE: &'static str = "TEXT";
    const STATUS_ENUM_TYPE: &'static str = "TEXT";

    #[cfg(feature = "postgres")]
    const REASON_ENUM_TYPE: &'static str = "asset_reason";
    #[cfg(feature = "postgres")]
    sql!(
        connection,
        r#"
            CREATE TYPE {REASON_ENUM_TYPE} AS ENUM (
                'AwaitingInstallation',
                'Stored',
                'OffsiteForRepairs',
                'AwaitingDecomissioning',
                'NeedsServicing',
                'MultipleTemperatureBreaches',
                'Unknown',
                'NeedsSpareParts',
                'LackOfPower',
                'Functioning',
                'Decomissioned'
                );
            "#
    )?;

    #[cfg(feature = "postgres")]
    const STATUS_ENUM_TYPE: &'static str = "asset_status";
    #[cfg(feature = "postgres")]
    sql!(
        connection,
        r#"
            CREATE TYPE {STATUS_ENUM_TYPE} AS ENUM (
                'NotInUse',
                'Functioning',
                'FunctioningButNeedsAttention',
                'NotFunctioning',
                'Decomissioned'
            );
            "#
    )?;
```

server/repository/src/db_diesel/assets/asset_log_row.rs
fergie-nz: Discussed with @mark-prins and @jmbrunskill - we will add as is with option to make configurable as a future issue if required

mark-prins (approved):
The data type isn't required and creates an overhead, so ignore that comment!

PR #3244, opened 2024-03-10
3126 mandatory patient fields 2

### LINKED ISSUE #3126: Should there be mandatory fields for patient creation?

clemens-msupply:
Fixes #3126

# 👩🏻‍💻 What does this PR do? 
Sets firstName and lastName to required for the default patient schema. This also required to unset empty text fields, e.g. when deleting all text from a text input the data field is set to undefined. This is needed because an empty string passes the required test.

I tested a bit and I think this should be fine...

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -81,7 +81,7 @@ const usePatternValidation = (
     } else {
       setCustomError(undefined);
     }
-  }, [pattern, value]);
+  }, [pattern, setCustomError, value]);
   return customError;
 };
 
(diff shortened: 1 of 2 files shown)


roxy-dao (approved):
LGTM~

PR #3250, opened 2024-03-11
2926 Datepicker moves if it is to close to the edge

### LINKED ISSUE #2926: Datepicker moves if it is to close to the edge
## What went wrong? 😲
Datepicker moves if it is to close to edge, so you can't select a date...... Popping this here since might be able to fix this with: https://mui.com/x/react-date-pickers/custom-components/#popper

https://github.com/msupply-foundation/open-msupply/assets/61820074/125fc709-e47b-4ccc-9fef-8ff3eacd82c7

## How to Reproduce 🔨

## Your environment 🌱

 - Version: 
 - Platform:
   - desktop (macOS)

roxy-dao:
Fixes #2926

# 👩🏻‍💻 What does this PR do? 
Fix position of the datetime popper, so that it doesn't move when selecting dates.

https://github.com/msupply-foundation/open-msupply/assets/61820074/1b781609-9e0e-4664-ab88-8e4fbecb7a9c

- [ ] Have an item with lots of lines for stocktake (8+)?
- [ ] Create a stocktake with that item
- [ ] Click on item to open the stocktake line edit modal
- [ ] Try change the date of the last line
- [ ] The date picker shouldn't move and you should be able to set an expiry date

### DIFF
--- a/client/packages/common/src/ui/components/inputs/DatePickers/BaseDatePickerInput/BaseDatePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DatePickers/BaseDatePickerInput/BaseDatePickerInput.tsx
@@ -91,6 +91,9 @@ export const BaseDatePickerInput: FC<
             '& .MuiPickersDay-root.Mui-selected': {
               backgroundColor: `${theme.palette.secondary.main}`,
             },
+            '& .MuiPickersPopper-paper': {
+              position: 'fixed',
+            },
           },
         },
         desktopPaper: {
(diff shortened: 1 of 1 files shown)


client/packages/common/src/ui/components/inputs/DatePickers/BaseDatePickerInput/BaseDatePickerInput.tsx
```diff
@@ -91,6 +91,9 @@ export const BaseDatePickerInput: FC<
             '& .MuiPickersDay-root.Mui-selected': {
               backgroundColor: `${theme.palette.secondary.main}`,
             },
+            '& .MuiPickersPopper-paper': {
+              position: 'fixed',
+            },
```
clemens-msupply, line 96: It works but do you know why it's doing this? i.e. is it a bug in MUI?

Could you please also add a comment here about why this is needed?

clemens-msupply (approved):
Looks good thanks.

Please see the comment / add a comment before merging.

client/packages/common/src/ui/components/inputs/DatePickers/BaseDatePickerInput/BaseDatePickerInput.tsx
roxy-dao: Think that's just their default popper values

PR #3251, opened 2024-03-11
#3195 Alert when attempting to return 0 quantities

### LINKED ISSUE #3195: Warn/handle when setting all return quantities to 0
Epic #2897 

If you don't set any quantities to return...
Image

Then you get a blank return reasons step, which isn't ideal:
Image

But we delete lines by setting their quantities to 0, so this should be allowed!

When you click `OK & Next`, check if quantities are set. If none, show a confirmation modal? And on confirm, close the modal rather than going to the next step.

We should probably display something different based on whether we are adding vs editing (i.e. "no lines will be added" vs "all lines for this item will be deleted")

CarlosNZ:
Fixes #3195. Currently just for outbound return, can replicate on Inbound when the rest of that UI is in place (or is it already in another branch?)

Screenshot 2024-03-11 at 4 16 49 PM

# 👩🏻‍💻 What does this PR do? 

 

Adds an "Alert" component to appear when attempting to proceed (Ok and Next) in the returns wizard if no quantities are specified for any items. Mimics the behaviour in Outbound Shipment when clicking "Ok" without allocating any quantities, except this alert is an "error" since it won't actually let you proceed until rectified (the outbound shipment one is just a warning).

The Alert should scroll into view if it's not currently visible on screen

The Alert should disappear once some values are added to the quantities.

Tested as above.

Not 100% convinced of the wording for the message -- open to suggestions?

Would have liked to make it scroll all the way to the top of the modal, but couldn't quite get it to work.


client/packages/invoices/src/Returns/modals/OutboundReturn/OutboundReturn.tsx
```diff
@@ -54,6 +58,13 @@ export const OutboundReturnEditModal = ({
     }
   };
 
+  const handleNext = () => {
+    if (lines.every(line => line.numberOfPacksToReturn === 0)) {
+      setShowZeroQuantityAlert(true);
+      alertRef?.current?.scrollIntoView({ behavior: 'smooth' });
```
lache-melvin, line 64: this is cool!

lache-melvin (changes requested):
What you have works great, and I love the scroll thing, but we need to consider the edit use case...

The use cases for the modal:
- Creating a new outbound return from an inbound shipment. In this case, I think your alert makes sense (though I'm wondering whether the alert or the error snack would be more consistent with what we have elsewhere??)
- Adding a new item to an existing outbound return. In this case I think your alert makes sense too. Both of these cases, we wouldn't need to save/create anything.
- Editing existing outbound return lines. You click an existing line, and it opens the modal with that line, as well as new lines for all other available stock lines of that item. In this case, if you 0 out the existing line(s), we should be able to submit (zeroed out lines will be deleted on backend.) Perhaps we could do something similar to [`StockOutAlerts`](https://github.com/msupply-foundation/open-msupply/blob/develop/client/packages/invoices/src/StockOut/Components/StockOutAlerts.tsx) where we ask the user to click OK again if they intended to set all the quantities to 0?

CarlosNZ:
> What you have works great, and I love the scroll thing, but we need to consider the edit use case...
> 
> The use cases for the modal:
> 
>     * Creating a new outbound return from an inbound shipment. In this case, I think your alert makes sense (though I'm wondering whether the alert or the error snack would be more consistent with what we have elsewhere??)
> 
>     * Adding a new item to an existing outbound return. In this case I think your alert makes sense too. Both of these cases, we wouldn't need to save/create anything.
> 
>     * Editing existing outbound return lines. You click an existing line, and it opens the modal with that line, as well as new lines for all other available stock lines of that item. In this case, if you 0 out the existing line(s), we should be able to submit (zeroed out lines will be deleted on backend.) Perhaps we could do something similar to [`StockOutAlerts`](https://github.com/msupply-foundation/open-msupply/blob/develop/client/packages/invoices/src/StockOut/Components/StockOutAlerts.tsx) where we ask the user to click OK again if they intended to set all the quantities to 0?

Okay @lache-melvin, have added a conditional warning based on the modal state. See if you think that works and then I'll go and update the Inbound version with the same thing.

client/packages/invoices/src/Returns/modals/OutboundReturn/OutboundReturn.tsx
```diff
@@ -54,6 +64,27 @@ export const OutboundReturnEditModal = ({
     }
   };
 
+  const handleNext = () => {
+    if (
+      lines.some(line => line.numberOfPacksToReturn !== 0) ||
+      zeroQuantityAlert === 'warning'
```
lache-melvin, line 70: I think if we're zeroing existing lines, we should call onOk here to save and close, rather than still taking the user to the blank reasons step. 

Also could you add a comment or something here that `zeroQuantityAlert === 'warning'` means we're in edit mode, and the user has already clicked OK with zero quantities once... took me a while to figure out what this was doing 😅

client/packages/invoices/src/Returns/modals/OutboundReturn/ReturnSteps.tsx
```diff
@@ -38,13 +46,24 @@ export const ReturnSteps = ({
     return step === -1 ? 0 : step;
   };
 
+  const alertMessage =
+    zeroQuantityAlert === 'warning'
+      ? t('messages.warn-zero-return-quantity')
```
lache-melvin, line 51: NIT 😁 Might be easier as a dev to work out what was going on here if this one was something like `messages.zero-return-quantity-will-delete-lines`?

client/packages/invoices/src/Returns/OutboundDetailView/DetailView.tsx
```diff
@@ -81,13 +87,14 @@ export const OutboundReturnsDetailView: FC = () => {
...
               initialItemId={itemId}
+              modalMode={mode}
             />
           )}
 
           <Toolbar />
           <DetailTabs tabs={tabs} />
           <Footer />
-          <SidePanel />
+          {/* <SidePanel /> */}
```
lache-melvin, line 97: Why is this commented out?

lache-melvin (changes requested):
Looks great Carl, thanks for changes! I think the main change for me would be calling `onOk` when confirming zeroing of existing lines.

Were you planning to do inbound side in this PR or separate?

client/packages/invoices/src/Returns/modals/OutboundReturn/OutboundReturn.tsx
CarlosNZ: > I think if we're zeroing existing lines, we should call onOk here to save and close, rather than still taking the user to the blank reasons step.

Right, yeah, good point. Will do.

client/packages/invoices/src/Returns/OutboundDetailView/DetailView.tsx
CarlosNZ: I dunno, I think something got mixed up when pulling in from base branch at some point. Will fix.

CarlosNZ:
> Were you planning to do inbound side in this PR or separate?

I can do it in this one assuming all the elements are in place. Just wanted to get agreement on the functionality here and then I'll replicate it

CarlosNZ:
> > Were you planning to do inbound side in this PR or separate?
> 
> I can do it in this one assuming all the elements are in place. Just wanted to get agreement on the functionality here and then I'll replicate it

Okay, done the Inbounds too. Though can't fully test every scenario due to "Add" not working.

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -32,10 +37,17 @@ export const InboundReturnEditModal = ({
   outboundShipmentLineIds,
   customerId,
   onClose,
+  modalMode,
 }: InboundReturnEditModalProps) => {
-  const t = useTranslation('distribution');
+  const t = useTranslation(['distribution', 'replenishment']);
```
lache-melvin, line 42: Hm, this probably means the translation strings here should be in `common`, rather than pulling the replenishment translations into distribution?

lache-melvin (approved):
Awesome thanks Carl!

Looks like that comment you added about what `zeroQuantity === 'warning'` means got removed... I'd maybe add it back, or pull that comparison into a clearer named variable? Up to you :)

Merge whenever ready!

PR #3255, opened 2024-03-11
3050 fix height input

### LINKED ISSUE #3050: Unsetting height doesn't save at all
## What went wrong? 😲
From here: https://github.com/msupply-foundation/open-msupply/pull/2520#issuecomment-1958960229
## How to Reproduce 🔨
1. Do the minimum for programs-setup; or simply use PNG Demo data; that has programs-setup configured properly
2. Create a new patient or use any existing
3. Enrol it to a program like HIV Care & Treatment
4. Now add an encounter, like HIV Care (we need one which will require height details)
5. Add a height value and save the whole encounter window
6. Open the same encounter, this time unset the height value - I simply backspaced the value
7. Since Save button is disabled, try adding notes or changing status until Save button is enabled, and then Save
8. Reopen the same encounter and check the height value now
9. See error: it hasn't updated at all; it's still the same first value

## Your environment 🌱

- Version: v1.6.2
- mSupply version: v713-00-beta3
- Platform:
  - server (windows)
- Database type: 
  - SQLite3

clemens-msupply:
Fixes #3050

# 👩🏻‍💻 What does this PR do? 
- Enables decimal values for the JSONFroms Number input. This was not part of the issue but it wasn't possible to enter the height in meter...
- Fix the underlying problem: stripEmptyAdditions re-added deleted primitive data fields

As described in the issue, save height reload and then delete the height. The save button should be enabled now.

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Number.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Number.tsx
@@ -47,6 +47,7 @@ const UIComponent = (props: ControlProps) => {
       Input={
         <NumericTextInput
           {...inputProps}
+          decimalLimit={10}
           min={schema.minimum}
           max={schema.maximum}
         />
--- a/client/packages/programs/src/JsonForms/stripEmptyAdditions.test.ts
+++ b/client/packages/programs/src/JsonForms/stripEmptyAdditions.test.ts
@@ -1,3 +1,4 @@
+import { isEqual } from 'lodash';
 import {
   isEqualIgnoreUndefinedAndEmpty,
   stripEmptyAdditions,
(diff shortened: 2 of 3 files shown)


mark-prins (commented):
👍 Working nicely! allows clearing of the height value. 

One small issue - after clearing the value, the save status isn't updated:

image

So you can't clear the height and save, unless you change something else.
If you do change something else, the blank value isn't saved - is that correct? The height reverts to the value saved previously

clemens-msupply:
> 👍 Working nicely! allows clearing of the height value.
> 
> One small issue - after clearing the value, the save status isn't updated:
> image
> 
> So you can't clear the height and save, unless you change something else. If you do change something else, the blank value isn't saved - is that correct? The height reverts to the value saved previously

Argh no you are describing the bug I was suppose to fix! :)

Are you sure? I just tried it again it is working here! I can clear the previously saved height value; the save button then becomes visible, and the value is removed when saving.

roxy-dao:
The `Save` button is also disabled for me if I only unset height 😢

clemens-msupply:
You are right it was still broken, don't know what I tested before... I fixed it and added a test case for it.

roxy-dao (approved):
Thanks Clemens~

PR #3257, opened 2024-03-11
Equipment list to show Status

### LINKED ISSUE #3140: Asset list view part 2
Part of epic #2896 

Follow up work from #3137

* [Adam's comment](https://github.com/msupply-foundation/open-msupply/issues/2973#issuecomment-1960717319) on colours of the status indication
* Status column
* ~Manufacturer & Model~
* ~Batch deletions~
* ~Scan? Perhaps a new issue for this one?~

### LINKED ISSUE #3204: Add enum of reasons to asset log

mark-prins:
Fixes #3140 

# 👩🏻‍💻 What does this PR do? 
 
Adds a loader to get the current status of an asset and shows the status in the equipment list like this:

image

For this, I've manually created asset logs which are associated to my equipment. Alternatively, you could wait until the next PR if you want to have the UI for adding, or use the API directly.

## 📃 Documentation

_Part of the CCEI feature_

### DIFF
--- a/server/graphql/core/src/loader/asset_status.rs
+++ b/server/graphql/core/src/loader/asset_status.rs
@@ -24,18 +24,11 @@ impl Loader<String> for AssetStatusLoader {
 
         let sort = AssetLogSort {
             key: AssetLogSortField::LogDatetime,
-            desc: Some(true),
+            desc: Some(false),
         };
 
         let result = repo
-            .query(
-                Pagination {
-                    offset: 0,
-                    limit: 1,
-                },
-                Some(filter),
-                Some(sort),
-            )?
+            .query(Pagination::all(), Some(filter), Some(sort))?
             .into_iter()
             .filter(|asset_log| asset_log.status.is_some())
             .map(|asset_log| (asset_log.asset_id.clone(), asset_log.status.unwrap()))
--- a/server/graphql/core/src/loader/asset_status.rs
+++ b/server/graphql/core/src/loader/asset_status.rs
@@ -1,6 +1,6 @@
-use repository::asset_log::{AssetLogFilter, AssetLogRepository, AssetLogSort, AssetLogSortField};
+use repository::asset_log::{AssetLogFilter, AssetLogRepository};
 use repository::asset_log_row::Status;
-use repository::{EqualFilter, Pagination};
+use repository::EqualFilter;
 use repository::{RepositoryError, StorageConnectionManager};
 
 use async_graphql::dataloader::*;
@@ -19,16 +19,10 @@ impl Loader<String> for AssetStatusLoader {
     async fn load(&self, ids: &[String]) -> Result<HashMap<String, Self::Value>, Self::Error> {
         let connection = self.connection_manager.connection()?;
         let repo = AssetLogRepository::new(&connection);
-
         let filter = AssetLogFilter::new().asset_id(EqualFilter::equal_any(ids.to_owned()));
 
-        let sort = AssetLogSort {
-            key: AssetLogSortField::LogDatetime,
-            desc: Some(false),
-        };
-
         let result = repo
-            .query(Pagination::all(), Some(filter), Some(sort))?
+            .query_latest(Some(filter))?
             .into_iter()
             .filter(|asset_log| asset_log.status.is_some())
             .map(|asset_log| (asset_log.asset_id.clone(), asset_log.status.unwrap()))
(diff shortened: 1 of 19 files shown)


server/graphql/core/src/loader/asset_status.rs
```diff
@@ -0,0 +1,39 @@
...
+        let repo = AssetLogRepository::new(&connection);
+
+        let filter = AssetLogFilter::new().asset_id(EqualFilter::equal_any(ids.to_owned()));
+
+        let sort = AssetLogSort {
+            key: AssetLogSortField::LogDatetime,
+            desc: Some(false),
+        };
+
+        let result = repo
```
mark-prins, line 30: perhaps not the most efficient way to build the hash map.

jmbrunskill: I'd be tempted to create a view like `changelog_deduped` or write a raw SQL query in diesel something like this...
```
CREATE VIEW latest_asset_log AS
  SELECT al.id,
    al.asset_id,
    al.user_id,
    al.status,
    al.comment,
    al.type,
    al.reason,
    al.log_datetime
  FROM (
    SELECT asset_id, MAX(log_datetime) AS latest_log_datetime
    FROM asset_log
    GROUP BY asset_id
  ) grouped
  INNER JOIN asset_log al
    ON al.asset_id = grouped.asset_id AND al.log_datetime = grouped.latest_log_datetime;

CREATE INDEX ix_asset_log_asset_id_log_datetime ON asset_log (asset_id, log_datetime);
```

Or just run a query for each ID maybe? Could be a lot but bounded to how many we display in a page (e.g. 20?)

I think reading all each asset's log entry just to get that latest could be a potential performance issue in future if there's 1000s of status logs...

I'm also not 100% if we're guaranteed to get the right result with this implementation. Have you found any docs to suggest that `collect()` will always take the first record for each id as the hashmap entry? I've had a look but can't find anything definitive...

server/repository/src/migrations/v1_08_00/assets/asset_log.rs
```diff
@@ -8,8 +8,8 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
         CREATE TABLE asset_log (
             id TEXT NOT NULL PRIMARY KEY,
             asset_id TEXT NOT NULL REFERENCES asset(id),
-            user_id TEXT REFERENCES user_account(id),
-            status TEXT, 
+            user_id NOT NULL TEXT REFERENCES user_account(id),
+            status TEXT,
```
jmbrunskill, line 12: Should we make this not null too?
Do we want to enter a lot with no status?

jmbrunskill (commented):
Should have been a comment on an earlier change but I wonder if we want to make more fields required? Generally simplifies code if we don't have Option<> and Nullable flying around...

fergie-nz (approved):
Looks great.

I don't quite understand how the hashmap returns just the most recent log, but is working great!

server/repository/src/migrations/v1_08_00/assets/asset_log.rs
```diff
@@ -8,8 +8,8 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
         CREATE TABLE asset_log (
             id TEXT NOT NULL PRIMARY KEY,
             asset_id TEXT NOT NULL REFERENCES asset(id),
-            user_id TEXT REFERENCES user_account(id),
-            status TEXT, 
+            user_id NOT NULL TEXT REFERENCES user_account(id),
```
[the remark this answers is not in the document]
jmbrunskill: Yup that's a bug I think. Probably works in Postgres but SQLite doesn't handle it.

jmbrunskill:
> I don't quite understand how the hashmap returns just the most recent log, but is working great!
I'm pretty sure the collect() function loops through every `vec` entry and tries to insert the value into the `HashMap` this effectively creates a last entry wins effect. I can't imagine this behaviour changing but I don't like it, both from the point of view of relying on this ordering and the implementation of collect() for HashMap, and because might be slow if there's a lot of logs to trawel through...

mark-prins:
awesome! thanks for the feedback. have implemented the view and I'm much happier with that 🙏

PR #3258, opened 2024-03-11
3254 User Login fails If sqlite database is locked for writing

### LINKED ISSUE #3254: User Login fails If sqlite database is locked for writing
## What went wrong? 😲
User logins fail with message `SOMETHING I CAN"T REMEMBER`

I think this is because SQLite can return a `SQLITE_BUSY` error if we try to write to the db when a write is in progress (aka a transaction)

Our transactions are started with `BEGIN IMMEDIATE` to avoid more than one transaction starting and then getting a deadlock.
```
IMMEDIATE causes the database connection to start a new write immediately, without waiting for a write statement. The BEGIN IMMEDIATE might fail with [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) if another write transaction is already active on another database connection.
```

Maybe we can switch back to `DEFERRED` mode? or handle the login errors differently to other error, for example if we can't update the user, should we actually return an error to login? Or just log the error in the logs?
https://www.sqlite.org/lang_transaction.html

## Expected behaviour 🤔

Logins should succeed, or give a useful error messages

## How to Reproduce 🔨

1. Sync a lot of records (e.g. add a new store to you existing site)
2. While sync is running (integrating) try to login

## Your environment 🌱
- Platform: All?
- Database type: SQLite

roxy-dao:
Fixes #3254

# 👩🏻‍💻 What does this PR do? 
Improved message if db is locked on sign in... open for message suggestions
Screenshot 2024-03-12 at 12 01 21

- [ ] Have a store with a big data set
- [ ] On central -> Move that store to open mSupply site
- [ ] On open mSupply -> start sync
- [ ] Wait until integration step (4)
- [ ] Sign out
- [ ] Try sign in again
- [ ] See newly improved message

### DIFF
--- a/client/packages/common/src/api/GqlContext.tsx
+++ b/client/packages/common/src/api/GqlContext.tsx
@@ -43,6 +43,14 @@ interface ResponseError {
   extensions?: { details?: string };
 }
 
+export class StdError extends Error {
+  public stdError?: string | undefined;
+  constructor(message: string, stdError: string | undefined) {
+    super(message);
+    this.stdError = stdError;
+  }
+}
+
 const hasError = (errors: ResponseError[], error: AuthError) =>
   errors.some(({ message }: { message?: string }) => message === error);
 
@@ -68,7 +76,10 @@ const handleResponseError = (errors: ResponseError[]) => {
   const error = errors[0];
   const { extensions } = error || {};
   const { details } = extensions || {};
-  throw new Error(details || error?.message || 'Unknown error');
+  throw new StdError(
+    details || error?.message || 'Unknown error',
+    error?.message
+  );
 };
 
 const shouldIgnoreQuery = (definitionNode: DefinitionNode) => {
--- a/client/packages/common/src/api/GqlContext.tsx
+++ b/client/packages/common/src/api/GqlContext.tsx
@@ -43,7 +43,7 @@ interface ResponseError {
   extensions?: { details?: string };
 }
 
-export class StdError extends Error {
+export class GraphqlStdError extends Error {
   public stdError?: string | undefined;
   constructor(message: string, stdError: string | undefined) {
     super(message);
@@ -76,7 +76,7 @@ const handleResponseError = (errors: ResponseError[]) => {
   const error = errors[0];
   const { extensions } = error || {};
   const { details } = extensions || {};
-  throw new StdError(
+  throw new GraphqlStdError(
     details || error?.message || 'Unknown error',
     error?.message
   );
--- a/client/packages/common/src/intl/locales/en/app.json
+++ b/client/packages/common/src/intl/locales/en/app.json
@@ -42,7 +42,7 @@
   "error.sync-api-incompatible": "Sync api version is not compatible",
   "error.authentication-error": "The server has returned an error",
   "error.connection-error": "Unable to connect to server",
-  "error.database-busy": "The server experienced a problem.",
+  "error.internal-error": "The server experienced a problem.",
   "error.integration-timeout-reached": "The central server took too long integrating pushed records. Please retry later.",
   "error.invalid-url": "Invalid url",
   "error.login": "Invalid username or password",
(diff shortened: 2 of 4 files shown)


client/packages/host/src/components/Login/Login.tsx
```diff
@@ -66,6 +66,10 @@ export const Login = () => {
       );
       return `${t('error.account-blocked')} ${formattedTime}`;
     }
+    if (error.detail?.includes('TransactionError')) {
+      return t('error.database-busy');
+    }
```
clemens-msupply, line 71: We map it to an unspecified/unexpected internal error. The string is just a helper to indicate what this error was. So we shouldn't really look for a specific string in it. We can say "The server experienced a problem please retry again later" instead.

clemens-msupply (changes requested):
See comment, think we shouldn't look for a specific message in a "debug" string...

client/packages/host/src/components/Login/Login.tsx
clemens-msupply: You could display the detail in the tooltip though...

roxy-dao:
> See comment, think we shouldn't look for a specific message in a "debug" string...

The tooltip already displays the debug string XD

client/packages/host/src/components/Login/Login.tsx
clemens-msupply: I mean could you please check if its an internal error instead of looking into the detail message (which is sqlite specific)

It should be in the `error.message == "InternalError"` field if I am not mistaken

roxy-dao: It's a database error. The issue you closed is different... I mean I guess it's the same since it's also a database error but the message won't apply...

clemens-msupply: but its both an internal error isn't it: "The server experienced a problem. Please try again later" should still work (even though a migration problem can't fix itself)

roxy-dao: Makes the message so misleading with migration problem...

roxy-dao: It still didn't like searching message for UpdateUserError D:

clemens-msupply: A migration problem should hopefully not happen, and then we would need the msg in the tooltip anyway. I think this it s ok to say "The server experienced a problem".

> It still didn't like searching message for UpdateUserError D:

Can you please check for "Internal error" instead? this might need some refactor to keep the original standard error in the error object... Please, let me know if I should have another look at it how to do this.

client/packages/host/src/components/Login/Login.tsx
```diff
@@ -66,7 +66,7 @@ export const Login = () => {
       );
       return `${t('error.account-blocked')} ${formattedTime}`;
     }
-    if (error?.detail?.includes('UpdateUserError')) {
+    if (error?.stdError?.includes('Internal error')) {
```
clemens-msupply, line 69: Nit: `error?.stdError === 'Internal error'`

client/packages/common/src/api/GqlContext.tsx
```diff
@@ -43,6 +43,14 @@ interface ResponseError {
   extensions?: { details?: string };
 }
 
+export class StdError extends Error {
```
clemens-msupply, line 46: Nit: rename to `GraphqlStdError` and maybe a quick comment what the fields are...

client/packages/common/src/intl/locales/en/app.json
```diff
@@ -42,6 +42,7 @@
   "error.sync-api-incompatible": "Sync api version is not compatible",
   "error.authentication-error": "The server has returned an error",
   "error.connection-error": "Unable to connect to server",
+  "error.database-busy": "The server experienced a problem.",
```
clemens-msupply, line 45: Nit: rename this label to `error.server-internal-error`

clemens-msupply (approved):
Thanks looks good, just some nits...

PR #3259, opened 2024-03-11
3243 add asset changes logs

### LINKED ISSUE #3243: Asset Changes Logs
Asset, asset catalogue and other records don't have the standard asset logging

fergie-nz:
Fixes #3243 

# 👩🏻‍💻 What does this PR do? 
Adds activity logs for asset inserts, updates, deletes, and log inserts

No testing has been added


mark-prins (changes requested):
Changes look great thanks Ferg!

Have run into a problem with postgres though - can you add a migration to add the types required if you run this with postgres please.

I've added a file `activity_log.rs` under the assets migration folder with the following contents:

```
use crate::{migrations::sql, StorageConnection};

pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
    if cfg!(feature = "postgres") {
        sql!(
            connection,
            r#"
            ALTER TYPE activity_log_type ADD VALUE 'ASSET_CREATED';
            ALTER TYPE activity_log_type ADD VALUE 'ASSET_UPDATED';
            ALTER TYPE activity_log_type ADD VALUE 'ASSET_DELETED';
            ALTER TYPE activity_log_type ADD VALUE 'ASSET_LOG_CREATED';
            "#
        )?;
    }

    Ok(())
}
```

mark-prins:
Perhaps for the 'updated' activity you could record the 'changed from' and 'changed to' as well?
Would that be possible? I've done something that works  - not sure if it's kosher or not, but I think it's neat:

Add `Serialize` to the traits that the `AssetRow` derives:
```
#[derive(Clone, Insertable, Queryable, Debug, PartialEq, AsChangeset, Eq, Default, Serialize)]
#[table_name = "asset"]
pub struct AssetRow {
```

and add the serialized struct as the to/from:
``
 activity_log_entry(
                &ctx,
                ActivityLogType::AssetUpdated,
                Some(updated_asset_row.id.clone()),
                Some(serde_json::to_string(&asset_row).unwrap_or_default()),
                Some(serde_json::to_string(&updated_asset_row).unwrap_or_default()),
            )?;
```

note: you'll also need to clone the `asset_row` here:
```
            let updated_asset_row = generate(&ctx.store_id, input, asset_row.clone());
```

fergie-nz:
> Perhaps for the 'updated' activity you could record the 'changed from' and 'changed to' as well? Would that be possible? I've done something that works - not sure if it's kosher or not, but I think it's neat:

Super cool. I wanted to do something like this but I saw that some other update logs didn't include the before and after.

I've added to the branch, thanks Mark!

mark-prins (approved):

mark-prins:
Very nice! thanks Ferg

PR #3260, opened 2024-03-11
Return empty string instead of "multiple" if there are no lines

### LINKED ISSUE #3029: Why does the location show as [multiple] in prescription when we try to issue items having no location?
## What went wrong? 😲
 
Why does the location show as [multiple] in prescription when we try to issue items having no location?

## Expected behaviour 🤔
Should be consistent with other window like inbound; if no location, show blank instead of confusing words like [multiple]

## How to Reproduce 🔨
1. Get some item stocks, but assign no location
2. Get that item into the prescription
3. See error: the location columns shows [multiple] as if there are multiple locations. 
image
4.I've also seen this sometimes in stocktake window, especially when `Group by item` is ON.

## Your environment 🌱

- Version: v1.6.2
- mSupply version: v713-00-beta3
- Platform:
  - server (windows)
- Database type: 
  - SQLite3

clemens-msupply:
Fixes #3029

# 👩🏻‍💻 What does this PR do? 
Handles the case if there are now lines, i.e. return "" instead of "multiple"

See issue description

### DIFF
--- a/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
+++ b/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
@@ -54,10 +54,10 @@ export const useColumnUtils = () => {
           return !!obj ? [obj] : [];
         });
         const isValid =
-          arr.length === 0 ||
-          arr[0] === undefined ||
-          typeof arr[0] !== 'object' ||
-          !(key in arr[0]);
+          arr.length > 0 &&
+          (arr[0] === undefined ||
+            typeof arr[0] !== 'object' ||
+            !(key in arr[0]));
 
         return isValid
           ? ArrayUtils.ifTheSameElseDefault(
--- a/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
+++ b/client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
@@ -53,11 +53,16 @@ export const useColumnUtils = () => {
           const obj = line[key];
           return !!obj ? [obj] : [];
         });
+
+        // Extract a value from the array if:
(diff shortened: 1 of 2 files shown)


client/packages/common/src/ui/layout/tables/utils/ColumnUtils.ts
```diff
@@ -54,10 +54,10 @@ export const useColumnUtils = () => {
...
         });
         const isValid =
-          arr.length === 0 ||
-          arr[0] === undefined ||
-          typeof arr[0] !== 'object' ||
-          !(key in arr[0]);
+          arr.length > 0 &&
+          (arr[0] === undefined ||
+            typeof arr[0] !== 'object' ||
+            !(key in arr[0]));
```
clemens-msupply, line 60: @mark-prins I wasn't sure about the other `isValid` conditions and why they are there. Do you have more info on it? then I will add it as a comment...

Tests are still passing with change...

mark-prins: not sure sorry!
Would have been taken from the previous implementation.. but makes little sense reading it now 🙄 
I would do this if I was looking at it, which is pretty messy and of course I don't know if it matches the original intention

```
        const isValid =
          arr.length > 0 &&
          !!propertyKey &&
          arr[0] !== undefined &&
          typeof arr[0] === 'object' &&
          propertyKey in arr[0];
```

mark-prins (approved):
👍 LGTM! the changes are working nicely. Have no idea what that logic block was supposed to do.. it's pretty awful, sorry about that.

clemens-msupply:
Thanks Mark, I updated the check and added comment about it

PR #3264, opened 2024-03-12
1710 The internal order Hide items over the minimum toggle button hides also lines that equals the minimum

### LINKED ISSUE #1710: The internal order Hide items over the minimum toggle button hides also lines that equals the minimum
## What went wrong? 😲
 
I am not sure if this is true for all lines that strictly have stock equal to the minimum MOS but it certainly hides those that have 0 stock and a CMM of 0 which technically, I think, should be the minimum.

hide items over minimum

## Expected behaviour 🤔
I feel these should be shown because they are technically not over the the minimum. It is  debatable though. 

## How to Reproduce 🔨

1. Create an internal order using a master list/program
2. You will probably have a couple of lines of 0 stock which have also a CMM of 0 
3. Toggle the button to "Hide the lines over the minimum"

## Your environment 🌱
 - Version [e.g. 1.1.11]
 - Platform [windows desktop]

roxy-dao:
Fixes #1710

# 👩🏻‍💻 What does this PR do? 
Show items with 0 stock on hand and 0 average monthly consumption when the `Hide stock over minimum` button is pressed.
Screenshot 2024-03-13 at 09 56 40

- [ ] Have some items that have 0 stock and 0 average monthly consumption or create some
- [ ] Create an `Internal Order`
- [ ] Use the `Hide stock over minimum` button
- [ ] Items should show


mark-prins (approved):
🎉  Looks good thanks and working as advertised!

Three follow up thoughts:
* maybe destructuring `item` and `itemStats` would make that filter simpler and easier to read?
* Why is the Switch gray? what were we thinking?? I like `color="secondary"` personally, it's more obvious when selected / active
* I have 'Service charges' showing as a valid item. 

For that last one - I think that the filter was removed in order to show service items in the item list. but these should not show when adding an item to an internal order.
How about doing this in `useStockItemsWithStats`

```
export const useStockItemsWithStats = () => {
  const queryParams = {
    sortBy: { key: 'name', isDesc: false, direction: 'asc' as 'asc' | 'desc' },
    offset: 0,
    first: 5000, // TODO: remove arbitrary limit
    filterBy: { type: { equalTo: ItemNodeType.Stock } },
```

??

roxy-dao:
Also would need to refactor `add_from_master_list` mutations to only add items that are stock

mark-prins:
oh, true. that currently pulls in the service item if you are silly enough to add that to the master list.

roxy-dao:
> * maybe destructuring `item` and `itemStats` would make that filter simpler and easier to read?
> * Why is the Switch gray? what were we thinking?? I like `color="secondary"` personally, it's more obvious when selected / active
> * I have 'Service charges' showing as a valid item when adding items to the internal order. It shouldn't be available!

Done, done and done (: Moved the `Add from Master List` thing to another issue (:

PR #3270, opened 2024-03-13
3213 Selected tab underline doesn't move correctly with SidePanel open/close

### LINKED ISSUE #3213: Selected tab underline doesn't move correctly with SidePanel open/close
## What went wrong? 😲
 

Not hindering anything, just a funky lil UX moment.

Details tab is underlined: 
Screenshot 2024-03-07 at 11 13 55 AM

I open the side panel... the underline didn't follow the `Details` tab name 😢:
Screenshot 2024-03-07 at 11 14 01 AM

Close the side panel... now it moves! So it's in the wrong place AGAIN!
Screenshot 2024-03-07 at 11 14 05 AM

Without looking into it, I am slightly surprised that the tab titles move, but their underline doesn't lol. 

## Expected behaviour 🤔

Active tab remains underlined as the side panel is opened and closed.

## How to Reproduce 🔨

On a detail view (e.g. an inbound shipment), open and close the `More` side panel. See the tab underline 

## Your environment 🌱

- Version: 1.7.0

- Platform: browser (chrome)

roxy-dao:
Fixes #3213

# 👩🏻‍💻 What does this PR do? 
Fixes underline being in funky places

- [ ] Go to invoice
- [ ] Open side panel by clicking more
- [ ] Check that the underline is still in the correct place 
- [ ] Can even move tabs

### DIFF
--- a/client/packages/common/src/ui/components/navigation/Tabs/DetailTabs.tsx
+++ b/client/packages/common/src/ui/components/navigation/Tabs/DetailTabs.tsx
@@ -9,7 +9,6 @@ import {
   useDrawer,
 } from '@common/hooks';
 import { LocaleKey, useTranslation } from '@common/intl';
-import { debounce } from '@common/utils';
 import { AppBarTabsPortal } from '../../portals';
 import { DetailTab } from './DetailTab';
 import { ShortTabList, Tab } from './Tabs';
@@ -42,17 +41,19 @@ export const DetailTabs: FC<DetailTabsProps> = ({
   const currentTab = isValidTab(currentUrlTab)
     ? currentUrlTab
     : tabs[0]?.value ?? '';
+  const { showConfirmation } = useConfirmOnLeaving(false);
 
   // Inelegant hack to force the "Underline" indicator for the currently active
   // tab to re-render in the correct position when one of the side "drawers" is
   // expanded. See issue #777 for more detail.
   const { isOpen: detailPanelOpen } = useDetailPanelStore();
   const { isOpen: drawerOpen } = useDrawer();
-  const { showConfirmation } = useConfirmOnLeaving(false);
-  const handleResize = useCallback(
-    () => debounce(() => window.dispatchEvent(new Event('resize')), 100),
-    []
-  );
+  const handleResize = useCallback(() => {
+    window.dispatchEvent(new Event('resize'));
+  }, []);
+  useEffect(() => {
+    handleResize();
+  }, [detailPanelOpen, drawerOpen, handleResize]);
 
   const [tabQueryParams, setTabQueryParams] = useState<
     Record<string, UrlQueryObject>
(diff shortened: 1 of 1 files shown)


clemens-msupply:
You just removed the debounce, right? but why is this working? or why wasn't it working before?

roxy-dao:
> You just removed the debounce, right? but why is this working? or why wasn't it working before?

debounce and useCallback aren't working well together. We either use one or the other 👀  originally it was just debounced. useCallback got added sometime with no explanation...

client/packages/common/src/ui/components/navigation/Tabs/DetailTabs.tsx
```diff
@@ -42,17 +41,19 @@ export const DetailTabs: FC<DetailTabsProps> = ({
...
-  const handleResize = useCallback(
-    () => debounce(() => window.dispatchEvent(new Event('resize')), 100),
-    []
-  );
+  const handleResize = useCallback(() => {
+    window.dispatchEvent(new Event('resize'));
+  }, []);
+  useEffect(() => {
+    handleResize();
+  }, [detailPanelOpen, drawerOpen, handleResize]);
```
clemens-msupply, line 56: This doesn't work for me, the bar is not centerred correctly...

```ts
useEffect(() => {
    setTimeout(() => {
      window.dispatchEvent(new Event('resize'));
    }, 100);
  }, [detailPanelOpen, drawerOpen]);
```
This works better though. Don't ask me why...

@mark-prins where is that underline drawn? is that in MUI or are we doing it?

roxy-dao: But work the same for me... I see no difference in the underline movement

clemens-msupply: Without the debounce or the delay the event seem to be created to early, i.e. while the sidebar is still expanding/shrinking. Thus the underline stops to early...

clemens-msupply: Just a guess: there are multiple debounce calls and the time of the first debounce call + the delay is still before the animation finished, thus it doesn't work for me. My machine is slower than yours, so I can see this and you can't (?).

client/packages/common/src/ui/components/navigation/Tabs/DetailTabs.tsx
```diff
@@ -42,17 +41,21 @@ export const DetailTabs: FC<DetailTabsProps> = ({
...
   const { isOpen: detailPanelOpen } = useDetailPanelStore();
   const { isOpen: drawerOpen } = useDrawer();
-  const { showConfirmation } = useConfirmOnLeaving(false);
-  const handleResize = useCallback(
-    () => debounce(() => window.dispatchEvent(new Event('resize')), 100),
-    []
-  );
+  // const handleResize = useCallback(() => {
+  //   window.dispatchEvent(new Event('resize'));
+  // }, []);
```
clemens-msupply, line 53: Some dead code....

clemens-msupply (approved):
Looks good, just remove the dead code please

PR #3271, opened 2024-03-13
3086 setup graphl test with data   setup graphql test with data

### LINKED ISSUE #3086: setup_graphl_test_with_data -> setup_graphql_test_with_data
## The suggestion

This is an easy one (I think)
`setup_graphl_test_with_data` should be called `setup_graphql_test_with_data`

My editor show it as a spelling error which annoys me...

Not that I can see, it's only used in tests.

5min job but will create a big diff to review, don't want to do it as part of another issue which would confuse things!

roxy-dao:
Fixes #3086 

# 👩🏻‍💻 What does this PR do? 
Typo fixing


jmbrunskill (approved):
Looks great, love a good typo fix!

Commited a change for `setup_graphl_test` too.

PR #3274, opened 2024-03-13
Insert asset log (UI)

### LINKED ISSUE #3273: Insert asset log UI
Part of epic #2896 

Make this button do something useful.
image

Implement the insert asset log functionality, as per these mock ups:

image

image

The system should:
* show a modal when the button is clicked
* have a cancel button and a next button on the first step, and an Ok on the second step
* When clicking Ok, the log is inserted
* If successful, a success toast message is shown
* if there's a problem saving, an error toast is shown
* The list of reasons depends on the status selected (if not an error is thrown by the api)
* the list of types, status and reason are pre-defined by enums (the type I've taken from the spreadsheet, there wasn't a list in the doc)

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #3273

# 👩🏻‍💻 What does this PR do? 
 
Creates a modal and wires up the insert api mutation.
Here's how the modal looks:

image

With filtered reasons:

image

image

image

Try to create, check that cancel clears and closes the modal.
You'll need to check the database to see if the log is created successfully.

## 📃 Documentation

Part of the CCEI feature


jmbrunskill:
Feels weird to have reason field with no options for me.
image

@richardmoizeau what should we do here? Would it be better to have a reason like "No Reason Required?" in this case?

Also, if I'm doing a repair maintenance `Type` I feel like I should be entering a reason for the maintenance.
Like "Reason: Regular Scheduled Maintenance"

Not sure what the goals of collecting the reasons are?

jmbrunskill:
Had a bug where ok and next button is not enabled and not clear why...
Kapture 2024-03-14 at 10 14 02

jmbrunskill:
Functional Status doesn't show on the Summary Page, is that going to be fixed in a later PR?
image

richardmoizeau:
> Feels weird to have reason field with no options for me. image
> 
> @richardmoizeau what should we do here? Would it be better to have a reason like "No Reason Required?" in this case?
> 
> Also, if I'm doing a repair maintenance `Type` I feel like I should be entering a reason for the maintenance. Like "Reason: Regular Scheduled Maintenance"
> 
> Not sure what the goals of collecting the reasons are?

reasons are used to complement the functional status (eg. functioning), not the type of record (eg. maintenance / repair). So should we rename "reason" to "status reason" for clarity ?

Reasons indicate why a status is what it is so when status is = "not functioning", you can specify a standard reason why. These reasons should be customisable (same as for the stocktake/inventory adjustment reasons): user can add/remove reasons for each status. "Status" on the other hand, should not be customisable. 

I agree with you, your wording is better when no reason is required: "No Reason Required?"

richardmoizeau:
> Had a bug where ok and next button is not enabled and not clear why... Kapture 2024-03-14 at 10 14 02 [ Kapture 2024-03-14 at 10 14 02 ](https://private-user-images.githubusercontent.com/8287373/312624597-4f6843cb-16a2-4a47-b2d1-9ce0f8013055.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTAzNjQ4MDAsIm5iZiI6MTcxMDM2NDUwMCwicGF0aCI6Ii84Mjg3MzczLzMxMjYyNDU5Ny00ZjY4NDNjYi0xNmEyLTRhNDctYjJkMS05Y2UwZjgwMTMwNTUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDMxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAzMTNUMjExNTAwWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDIxNTY3MDE5YzQxNGM4YTRhZDM0ZTA4MTJjNTI4Y2JiMDM5YjA5MzcwMWM2ZTljODhkZWNkMTdmNTEwNDFhYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.Ki_M6XVDKuyx82ZYdqboVtabHmR9wuu7v_XcsUD_e0o) [ ](https://private-user-images.githubusercontent.com/8287373/312624597-4f6843cb-16a2-4a47-b2d1-9ce0f8013055.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTAzNjQ4MDAsIm5iZiI6MTcxMDM2NDUwMCwicGF0aCI6Ii84Mjg3MzczLzMxMjYyNDU5Ny00ZjY4NDNjYi0xNmEyLTRhNDctYjJkMS05Y2UwZjgwMTMwNTUuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDMxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAzMTNUMjExNTAwWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDIxNTY3MDE5YzQxNGM4YTRhZDM0ZTA4MTJjNTI4Y2JiMDM5YjA5MzcwMWM2ZTljODhkZWNkMTdmNTEwNDFhYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.Ki_M6XVDKuyx82ZYdqboVtabHmR9wuu7v_XcsUD_e0o)

the only reason I see for this is that you have only one asset in your list OR that this asset is last record of your list.

adamdewey:
Yeah also agree with Richard that the reasons should be customisable

In that case could we just grey out the reason dropdown if there are none defined?

jmbrunskill:
> Yeah also agree with Richard that the reasons should be customisable
> 
> In that case could we just grey out the reason dropdown if there are none defined?

We can probably actually just hide the whole drop down in this case?

jmbrunskill:
> These reasons should be customisable (same as for the stocktake/inventory adjustment reasons): user can add/remove reasons for each status. "Status" on the other hand, should not be customisable.

Haha, we discussed this but didn't see any evidence that customisation was required, this will need to be a new ticket/issue as we'll need to create a process for managing these reasons, new configuration screens etc. Let's chat about this part separately.

mark-prins:
> Had a bug where ok and next button is not enabled and not clear why...

yeah, had that too. couldn't see why it happened though - so I gave up.

mark-prins:
> Functional Status doesn't show on the Summary Page, is that going to be fixed in a later PR?

Forgot about it, will pop it in this one 🙏

mark-prins:
> These reasons should be customisable

hold on!! this is a huge change! that was unclear from the reqs.
can we chat about this please?

client/packages/coldchain/src/Equipment/DetailView/UploadTab.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { AssetLogPanel } from '../Components';
+
+export const UploadTab = ({ draft, value }: AssetLogPanel) => {
+  return (
+    <AssetLogPanel value={value} draft={draft}>
+      TODO: implement the upload panel
```
jmbrunskill, line 7: I wonder about things like this when commiting to develop? Ideally we should hide the tab maybe? I think it's ok at this stage, but as we get closer to a release...

client/packages/coldchain/src/Equipment/api/hooks/log/useAssetLogInsert.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const useAssetLogInsert = () => {
+  const queryClient = useQueryClient();
+  const api = useAssetApi();
+
+  return useMutation(
+    async (log: Partial<InsertAssetLogInput>) => api.insertLog(log),
+    {
+      onSettled: assetId =>
+        queryClient.invalidateQueries(api.keys.logs(assetId ?? '')),
```
jmbrunskill, line 16: Do we need to invalidate the assets as well? As they have statuses from a loader which might be incorrect now?

jmbrunskill (approved):
Looks good to me.

A couple of comments added but generally good to go I think.

jmbrunskill:
> yeah, had that too. couldn't see why it happened though - so I gave up.

I've recreated it several times, but also haven't figured out exactly why/when...

mark-prins:
Ok - have done the following:
* fixed the button enabled state
* added the log detail to the summary tab
* disabled the reason input when there are no options ( keeping elements visible but disabled is the preferred approach: 'we don't hide things' )
* cleared the reason when you change the status ( otherwise you can create invalid combinations )

bearing in mind this will have to change if we have customisable reason values, can we get this in?

client/packages/coldchain/src/Equipment/DetailView/UploadTab.tsx
mark-prins: good point. have removed the wizard and next button etc, so the modal is now valid.

client/packages/coldchain/src/Equipment/api/hooks/log/useAssetLogInsert.ts
mark-prins: good point! have changed to use the base key and invalidated. now working

PR #3278, opened 2024-03-13
3183 update inbound return lines

### LINKED ISSUE #3183: API for updateInboundReturnLines
Epic #2897 

Similar to updateOutboundReturnLines:

- Lines that are not already stored should be added to the return
- Existing lines should be updated
- Lines with number of packs set to 0 should be deleted
- Should use the StockInLine abstraction created for insert

lache-melvin:
Fixes #3183

# 👩🏻‍💻 What does this PR do? 
 

Adds API side of update inbound return lines.

Adds the StockIn update and delete abstractions.

Automated tests in the service layer. 

Not wired up to the frontend yet - will need to test graphql via a playground.

Sorry such beefy PRs, there is a painful amount of boilerplate going on 😭


server/service/src/invoice/inbound_return/update_lines/generate.rs
```diff
@@ -0,0 +1,131 @@
...
+        InvoiceLineFilter::new().invoice_id(EqualFilter::equal_to(&inbound_return_id)),
+    )?;
+    let check_already_exists = |id: &str| {
+        existing_lines
+            .iter()
+            .find(|line| line.invoice_line_row.id == id)
+            .is_some()
+    };
+
+    let lines_to_add = inbound_return_lines
```
mark-prins, line 39: would it be weird to only iterate once and return a tuple? probably not an issue since there won't be a lot of lines, it just feels repetitive to loop over the same set of data four times.

server/service/src/invoice_line/stock_in_line/delete/validate.rs
```diff
@@ -0,0 +1,45 @@
...
+    if !check_invoice_is_editable(&invoice) {
+        return Err(CannotEditFinalised);
+    }
+    if !check_batch(&line, connection)? {
+        return Err(BatchIsReserved);
+    }
+    if !check_line_belongs_to_invoice(&line, &invoice) {
+        return Err(NotThisInvoiceLine(line.invoice_id));
+    }
+    if !check_line_not_associated_with_stocktake(connection, &line.id, store_id.to_string()) {
```
mark-prins, line 40: just a 🤔  on the naming - if `!not_associated` return `LineUsed`.
could we have a fn which was `is_line_associated_with_stocktake`?

sorry, outside of scope of PR. just wondering..

server/service/src/invoice_line/stock_in_line/update/generate.rs
```diff
@@ -0,0 +1,149 @@
+use crate::{
```
mark-prins, line 1: are the `stock_in_line` and `stock_out_line` just for returns?
I was confused bc I thought we'd probably have stock in/out already. Perhaps add 'return' in the name or pop in a new folder? some sort of distinction would be helpful (for me at least!)

mark-prins (approved):
👍 Looks good to me! Just a couple of musings as I read thru, no show stoppers

have run the code but not actively tested, I'm relying on the tests for that and will have another look when the UI is in 😉

server/service/src/invoice_line/stock_in_line/update/generate.rs
lache-melvin: We have stock_out, but not stock_in yet! The plan was to create this abstraction for returns, and at a later date, refactor inbound shipments to also use the stock_in abstraction!

server/graphql/invoice/src/mutations/inbound_return/insert.rs
```diff
@@ -10,8 +10,8 @@ use graphql_types::types::InvoiceNode;
 use service::auth::{Resource, ResourceAccessRequest};
 use service::invoice::inbound_return::insert::{
     InsertInboundReturn as ServiceInput, InsertInboundReturnError as ServiceError,
-    InsertInboundReturnLine,
 };
+use service::invoice::inbound_return::InboundReturnLineInput as ServiceLineInput;
```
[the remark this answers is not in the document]
lache-melvin: Renamed as it's called the same thing for the service and graphql layers, but that is a great point, will fix, thank you! 😅

server/service/src/invoice_line/stock_in_line/delete/validate.rs
lache-melvin: This is an interesting question... I like your thinking around clearing up the double negatives, certainly more readable.

But it looks like this actually fits with the validate pattern we have: `if !(the thing we want) return Err` (for example we also have `if !check_line_does_not_exist` in a number of places) ... probably not a biggie but would be interested what the consensus on this would be 🤔

PR #3281, opened 2024-03-13
Use a new CurrencyCell to display prices < 0.01

### LINKED ISSUE #3093: Invoice shows incorrect/imprecise sell price
## What went wrong? 😲
If the actual price is < 0.01, 0.01 is shown instead of the precise value. This can be confusing. 
image

This is for migrated data.

## Expected behaviour 🤔
- [ ] Show precise value in the list 
- [ ] Show precise value in the line detail view
- [ ] The value might get to long to fit into the list column, and thus it might need some special handling...

clemens-msupply:
Fixes #3093 

# 👩🏻‍💻 What does this PR do? 
Shows price < 0.01. There was actually some more space in the column to show prices at least as small as 0.001. Thus, I just show them now, e.g. `$0.008`.

For smaller prices we probably get into some edge cases. If prices are too small I do the following with a tooltip:

image

Also fixed a bug in our currency formatter for very small numbers.

I also replaced the current CurrencyCell by a new one. Sorry @roxy-dao saw it too late that there was one already. I also solved some lint warning for the previous CurrencyCell.

Check all places with where there are prices:
- inbound shipment
- outbound shipment
- prescriptions
- @roxy-dao could you please test the case where we want to display foreign currency? I haven't tested it yet but hoped you still have a test setup for it...


roxy-dao:
Isn't it 1 > 0.0001?

client/packages/common/src/intl/currency/currency.ts
```diff
@@ -2,16 +2,27 @@ import currency from 'currency.js';
...
 
+// See: https://stackoverflow.com/questions/50673859/how-to-prevent-very-small-numbers-from-being-converted-to-scientific-notation
+// For example 4e-7 becomes "0.0000004" instead of "4e-7"
+const toNonScientificString = (value: number): string => {
+  const valueString = value.toString();
+  const decimalsPart = valueString.split('.')?.[1] ?? '';
+  const eDecimals = Number(valueString?.split('e-')?.[1]) || 0;
+  const countOfDecimals = decimalsPart.length + eDecimals;
+  return value.toFixed(countOfDecimals);
+};
```
CarlosNZ, line 13: Can't you do this with Intl.NumberFormat? 

```js
const n = 0.0000004
console.log(n) // 4e-7 
const f = new Intl.NumberFormat("en", {maximumFractionDigits: 21}).format(n)
console.log(f) // "0.0000004"
```

I've been trying to make all number formatting localisable as much as possible e.g. the NumericTextInput, so we want to avoid assuming what the decimal character should be. (So you pass the current locale into Intl.NumberFormat)

roxy-dao:
Foreign currency is displayed in inbound/outbound line edit modal if preference is turned on and supplier/customer is external.
Screenshot 2024-03-18 at 12 10 42

roxy-dao (approved):
LGTM~ Thanks Clemens~

clemens-msupply:
thanks did you tested if it shows the foreign currency correctly?

roxy-dao:
> thanks did you tested if it shows the foreign currency correctly?

Yup and code for FC looks fine too~

Chris-Petty:
@richardmoizeau @anisha-msupply Here is the current iteration of very small monetary values (<0.01) values are shown in UI if you'd like to review. I think it gives good feedback that a value is rounded. Tooltip woe on android probably persists but that's a known broader design issue.

Chris-Petty:
Just noticed that on some data transferred from OG you can have cost values that are too small for the 2dp text inputs so just shows `0.00`, but calculated total shows value `>0`

image

image

This will likely be common for stores that convert pack size to 1 especially with high value currencies like USD.

PR #3287, opened 2024-03-14
Fix asset_log test

jmbrunskill:
Context was incorrectly created causing a `FOREIGN KEY constraint failed` error on store_id.
Turns out `asset_id` was used instead of `store_id`


mark-prins (approved):
👍  thanks @jmbrunskill!

PR #3291, opened 2024-03-14
#3288 Fix for 0 value in stocktake count

### LINKED ISSUE #3288: Cannot 0 stock in stocktake
## What went wrong? 😲
The new number changes has made it impossible to 0 out stock in a stocktake.

## Expected behaviour 🤔
Should be possible to 0 stock line in a stocktake.

## How to Reproduce 🔨

1. Create a new stocktake
2. Add a stock line that has stock
3. Enter `0` in `Counted packs`
4. See that you cannot

## Your environment 🌱

- Version: 1.7
- mSupply version: 

- Platform:

- Database type:

CarlosNZ:
Fixes #3288

# 👩🏻‍💻 What does this PR do? 
 

When we refactored the NumberInputCell, we set the default `min` to `1` as that was the more common case in the implementations. However, this means we need to explicitly set `min=0` in the places that need it. 

Working for stocktakes

@roxy-dao can you think of any other places this might be an issue?

### DIFF
--- a/client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
+++ b/client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
@@ -184,7 +184,7 @@ export const BatchTable: FC<StocktakeLineEditTableProps> = ({
       getIsError: rowData =>
         errorsContext.getError(rowData)?.__typename ===
         'StockLineReducedBelowZero',
-      Cell: props => <NumberInputCell {...props} decimalLimit={2} />,
+      Cell: props => <NumberInputCell {...props} decimalLimit={2} min={0} />,
       setter: patch => {
         // If counted number of packs was changed to result in no adjustment we
         // should remove inventoryAdjustmentReason, otherwise could have a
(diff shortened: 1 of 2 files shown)


roxy-dao:
Maybe in Inbound Lines too since we're allowing users to 0 out lines (if it's connected to an outbound shipment and hasn't been received or something)

CarlosNZ:
> Maybe in Inbound Lines too since we're allowing users to 0 out lines (if it's connected to an outbound shipment and hasn't been received or something)

Hmm, I added it to Inbound lines, but now when you set it to 0, the "Ok" button becomes disabled, so there's obviously some additional logic preventing 0 values. What do you think -- put it back to 1?

roxy-dao (approved):
LGTM~ Thanks Carl /o/

roxy-dao:
> > Maybe in Inbound Lines too since we're allowing users to 0 out lines (if it's connected to an outbound shipment and hasn't been received or something)
> 
> Hmm, I added it to Inbound lines, but now when you set it to 0, the "Ok" button becomes disabled, so there's obviously some additional logic preventing 0 values. What do you think -- put it back to 1?
> 
> Screenshot 2024-03-15 at 9 04 09 AM

Ah... yeah I guess put it back to 1.. they have a 0 out action so maybe okay?

Chris-Petty:
I think a default of 1 is probably unexpected by the uninitiated for both users and developers and I reckon will be a recurring mistake for both. UX philosophy wise don't know if it's generally a good idea to have a default as 1, users should deliberately be entering a quantity other than 0.

Issue #3295, opened 2024-03-14
3193B add front end for asset location mutations


fergie-nz:
## Describe the solution you'd like 🎁
See description in #3193

fergie-nz:
Closing as completed

PR #3297, opened 2024-03-14
3193A asset location editing

### LINKED ISSUE #3193: Asset location editing

fergie-nz:
Fixes #3193 

Back end only. 

Front end in separate issue: #3295

# 👩🏻‍💻 What does this PR do? 
Adds backend for asset location mutations including asset location dedicated loader

Repo layer tests added


server/graphql/core/src/loader/asset_location.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        let locations = location_repo
+            .query_by_filter(LocationFilter::new().id(EqualFilter::equal_any(location_ids)))?;
+
+        let mut map: HashMap<String, Vec<Location>> = HashMap::new();
+        for line in locations {
+            let list = map
+                .entry(ids[0].clone())
+                .or_insert_with(|| Vec::<Location>::new());
+            list.push(line);
```
fergie-nz, line 46: I'm not quite understanding how this is working - particularly the key. I am guessing this will only work with one assetId, which is how we are using the loader right now, but I couldn't get it to work in a more flexible manner.

server/repository/src/db_diesel/assets/asset_internal_location_row.rs
```diff
@@ -22,17 +19,15 @@ pub struct AssetInternalLocationRow {
...
-    pub modified_datetime: NaiveDateTime,
 }
 
-pub struct AssetLocationRowRepository<'a> {
+pub struct AssetInternalLocationRowRepository<'a> {
     connection: &'a StorageConnection,
 }
 
-impl<'a> AssetLocationRowRepository<'a> {
+impl<'a> AssetInternalLocationRowRepository<'a> {
```
fergie-nz, line 28: Renaming for consistency with file names

server/graphql/asset/src/types.rs
```diff
@@ -153,6 +174,16 @@ impl AssetNode {
             .map(AssetCatalogueItemNode::from_domain))
     }
 
+    pub async fn locations(&self, ctx: &Context<'_>) -> Result<AssetLocationConnector> {
```
jmbrunskill, line 177:
```suggestion
    pub async fn locations(&self, ctx: &Context<'_>) -> Result<LocationConnector> {
```

I think you can re-use the LocationConnector type, don't need the AssetLocationConnector at all!

server/graphql/core/src/loader/asset_location.rs
```diff
@@ -0,0 +1,85 @@
...
+                .or_insert_with(|| Vec::<Location>::new());
+            list.push(location);
+        }
+
+        Ok(map)
+    }
+}
+
+// #[cfg(test)]
+// mod tests {
```
jmbrunskill, line 60: Couldn't get this test working? Would be good to have a test here for this if possible...

server/repository/src/db_diesel/assets/asset_internal_location.rs
```diff
@@ -0,0 +1,195 @@
...
+            id: asset_location_id.clone(),
+            asset_id: mock_asset_a().id,
+            location_id: mock_location_1().id,
+        };
+
+        let result = asset_internal_location_row_repository
+            .insert_one(&asset_location)
+            .unwrap();
+
+        println!("result: {:?}", result);
```
jmbrunskill, line 181:
```suggestion

```
We usually don't keep println's in our tests. Useful for debugging though!

server/service/src/asset/delete.rs
```diff
@@ -30,11 +33,16 @@ pub fn delete_asset(ctx: &ServiceContext, asset_id: String) -> Result<String, De
                 None,
             )?;
 
+            let _deleted_location = AssetInternalLocationRowRepository::new(&connection)
```
jmbrunskill, line 36: Does not doing this cause a problem?
Since we are soft deleting the asset, we should be able to skip this step, although this would make the locations available for other assets...

server/service/src/asset/insert.rs
```diff
@@ -38,7 +41,7 @@ pub fn insert_asset(
         .connection
         .transaction_sync(|connection| {
             validate(&input, connection)?;
```
jmbrunskill, line 43: I think we need to validate that the location isn't already used by another asset.
Ideally we should warn the user and maybe remove the location from the other asset if they really want to do it.
Not sure the best way to alert the user but for now we should at least error if they try to use a location already assigned to another asset.

server/service/src/asset/insert.rs
```diff
@@ -28,6 +30,7 @@ pub struct InsertAsset {
     pub catalogue_item_id: Option<String>,
     pub installation_date: Option<NaiveDate>,
     pub replacement_date: Option<NaiveDate>,
+    pub location_ids: Option<Vec<String>>,
```
jmbrunskill, line 33: If it doesn't cause a problem for the front end, I'd probably remove this from the `InsertAsset` struct.
We don't all people to assign locations during the creation of a new asset, so we only really need this extra logic on update. That way it's also only in one place which I like.

server/service/src/asset/update.rs
```diff
@@ -53,6 +54,11 @@ pub fn update_asset(
                 Some(serde_json::to_string(&updated_asset_row).unwrap_or_default()),
             )?;
 
+            if input.location_ids.clone().is_some() {
```
jmbrunskill, line 57: We should have tests for adding and removing locations via this service

jmbrunskill (changes requested):
Looks good Ferg.

2 Changes I'd like to see.

- [ ] Re-use LocationConnector instead of creating a new one
- [ ] Add tests for the location saving logic in the service layer.

server/service/src/asset/insert.rs
fergie-nz: Curious! I think I wasn't understanding how assets and locations worked.

Do Locations sit within Assets? Or are Assets placed at Locations?

jmbrunskill: Locations are within the asset. Location name might be `Fridge 1 - Shelf 3` or something similar.
I tried to communicate that by calling it `asset_internal_location` to differentiate it from where an asset might be.
If we do other assets say Oxygen tanks in future, they might be stored in a location, which will require a different approach.

Hope that makes sense?

fergie-nz: Yes makes sense, thanks James

server/service/src/asset/delete.rs
fergie-nz: I'm wanting to delete these locations so that when an asset is soft deleted because it will allow locations to be assigned to new assets. 

Otherwise I'm thinking I'll need to add a soft delete to asset locations also. Maybe that is a good idea?
What do you think?

server/service/src/asset/validate.rs
```diff
@@ -55,3 +57,14 @@ pub fn check_reason_matches_status(
...
     }
 }
+
+pub fn check_locations_are_assigned(
+    location_ids: Vec<String>,
+    connection: &StorageConnection,
+) -> Result<Vec<AssetInternalLocationRow>, RepositoryError> {
+    Ok(
+        AssetInternalLocationRepository::new(connection).query_by_filter(
+            AssetInternalLocationFilter::new().id(EqualFilter::equal_any(location_ids)),
```
jmbrunskill, line 67: Does this have a bug when you're trying to save an asset, and the locations are already assigned to the current asset?

I think you want something like this, to correctly check the location_ids, but also ignore any that are assigned to the asset already.
```suggestion
            AssetInternalLocationFilter::new().location_id(EqualFilter::equal_any(location_ids)).asset_id(EqualFilter::not_equal_all(asset_id)),
```

jmbrunskill (changes requested):
Looks great Fergus,

Sorry one more change!
I think the logic for your asset_location_check is not correct, can you confirm?

jmbrunskill (approved):
Ready to go, thanks Fergus!

(Just some merge conflicts to resolve)

fergie-nz:
closing as replaced by #3413

Issue #3212, opened 2024-03-06
Add pack variant components for returns (3h)


lache-melvin:
## Is your feature request related to a problem? Please describe 👀

Pack variants has been merged! The code for pack variants is in our feature branch, but not yet added to the returns UI pieces.

If I'm returning some goods, or receiving a return, I want to be able to specify the pack variant of the goods.

## Describe the solution you'd like 🎁

Replace the PackSize column with the new `PackVariantCell` - any UI tweaks required?

lache-melvin:
Ok added in the inbound return modal as was in there editing anyway, but she's a bit funky looking when we have some items with variants and some without! Need to review this:

![Image](https://github.com/msupply-foundation/open-msupply/assets/55115239/107c86a3-769d-45fd-ba5a-bfb3d4085297)

PR #3300, opened 2024-03-15
3189 add lines by item

### LINKED ISSUE #3189: Add lines by item to existing inbound return
Epic #2897

Similar to outbound return/inbound shipment, need to be able to adds lines by item to an existing return.

On the detail view, click a `New Item` button, which opens the edit modal, including an item selector. More fields in the edit modal might need to be editable?

I can then fill out return line info for that item. 

Uncertain exactly how much detail will be required here (i.e. location, pricing...) but let's just start simple with the batch info. 

Can probably add the edit onRowClick here too?

lache-melvin:
Fixes #3189

# 👩🏻‍💻 What does this PR do? 
 

Implements the add by item/edit line version of the inbound return edit modal, and updates the table cells to be inputs:
Screenshot 2024-03-15 at 2 36 23 PM

- Add a new inbound return
- From the detail view, click Add Item
- Select an item
- [ ] A first draft line should appear
- [ ] Clicking `Add batch`, or using the `+` key, should add more lines
- [ ] You can input the line info
- Ok & Next, input reason/comment, OK
- [ ] The line is added to the return
- [ ] Clicking the line re-opens the modal
- [ ] You can update fields for that line
- [ ] Setting return quantity to 0 deletes the line

### DIFF
--- a/client/packages/invoices/src/InboundShipment/api/hooks/utils/useSelectedStockLineIds.ts
+++ b/client/packages/invoices/src/InboundShipment/api/hooks/utils/useSelectedStockLineIds.ts
@@ -1,4 +1,4 @@
-import { useTableStore } from 'packages/common/src';
+import { useTableStore } from '@openmsupply-client/common';
 import { useInboundRows } from '../line/useInboundRows';
 import { isString } from 'lodash';
 
--- a/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
+++ b/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
@@ -17,7 +17,7 @@ import { AppBarButtons } from './AppBarButtons';
 import { InboundReturnLineFragment, useReturns } from '../api';
 import { AppRoute } from '@openmsupply-client/config';
 import { SidePanel } from './SidePanel/SidePanel';
-import { ActivityLogList } from 'packages/system/src';
+import { ActivityLogList } from '@openmsupply-client/system';
 import { Footer } from './Footer';
 import { InboundReturnItem } from '../../types';
 import { InboundReturnEditModal } from '../modals';
--- a/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
+++ b/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
@@ -9,6 +9,7 @@ import {
   useTranslation,
   createQueryParamsStore,
   DetailTabs,
+  useEditModal,
 } from '@openmsupply-client/common';
 import { ContentArea } from './ContentArea';
 import { Toolbar } from './Toolbar';
@@ -18,16 +19,27 @@ import { AppRoute } from '@openmsupply-client/config';
 import { SidePanel } from './SidePanel/SidePanel';
 import { ActivityLogList } from 'packages/system/src';
 import { Footer } from './Footer';
+import { InboundReturnItem } from '../../types';
+import { InboundReturnEditModal } from '../modals';
 
 export const InboundReturnDetailView: FC = () => {
   const { data, isLoading } = useReturns.document.inboundReturn();
   const t = useTranslation('distribution');
   const navigate = useNavigate();
 
-  // TODO: hook these up to modal
-  const onRowClick = () => {};
+  const {
+    onOpen,
+    onClose,
+    isOpen,
+    entity: itemId,
+    mode,
+  } = useEditModal<string>();
 
-  const onAddItem = () => {};
+  // TODO: hook these up to modal
+  const onRowClick = (row: InboundReturnLineFragment | InboundReturnItem) =>
+    onOpen(row.itemId);
+
+  const onAddItem = () => onOpen();
 
   if (isLoading) return <DetailViewSkeleton hasGroupBy={true} hasHold={true} />;
 
@@ -56,6 +68,17 @@ export const InboundReturnDetailView: FC = () => {
           })}
         >
           <AppBarButtons onAddItem={onAddItem} />
+          {isOpen && (
+            <InboundReturnEditModal
+              isOpen={isOpen}
+              onClose={onClose}
+              outboundShipmentLineIds={[]}
+              customerId={data.otherPartyId}
+              returnId={data.id}
+              // initialItemId={itemId}
+              modalMode={mode}
+            />
+          )}
 
           <Toolbar />
           <DetailTabs tabs={tabs} />
--- a/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
+++ b/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
@@ -38,13 +38,11 @@ export const InboundReturnDetailView: FC = () => {
   const onRowClick = (row: InboundReturnLineFragment | InboundReturnItem) =>
     onOpen(row.itemId);
 
-  const onAddItem = () => onOpen();
-
   if (isLoading) return <DetailViewSkeleton hasGroupBy={true} hasHold={true} />;
 
   const tabs = [
     {
-      Component: <ContentArea onRowClick={onRowClick} onAddItem={onAddItem} />,
+      Component: <ContentArea onRowClick={onRowClick} onAddItem={onOpen} />,
       value: t('label.details'),
     },
     {
@@ -66,7 +64,7 @@ export const InboundReturnDetailView: FC = () => {
             },
           })}
         >
-          <AppBarButtons onAddItem={onAddItem} />
+          <AppBarButtons onAddItem={onOpen} />
           {isOpen && (
             <InboundReturnEditModal
               isOpen={isOpen}
--- a/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
+++ b/client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
@@ -35,7 +35,6 @@ export const InboundReturnDetailView: FC = () => {
     mode,
   } = useEditModal<string>();
 
-  // TODO: hook these up to modal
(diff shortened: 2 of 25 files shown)


client/packages/invoices/src/InboundShipment/api/hooks/utils/useSelectedStockLineIds.ts
```diff
@@ -1,4 +1,4 @@
-import { useTableStore } from 'packages/common/src';
+import { useTableStore } from '@openmsupply-client/common';
```
lache-melvin, line 1: Ah sorry for this noise, cleaned up these imports (the 'packages/..' imports break the tests)

client/packages/invoices/src/Returns/modals/InboundReturn/ItemSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+            disabled={disabled}
+            currentItemId={itemId}
+            onChange={newItem => newItem && onChangeItemId(newItem.id)}
+            extraFilter={
+              disabled
+                ? undefined
+                : item =>
+                    !items?.some(existingItem => existingItem.id === item.id)
+            }
+          />
```
lache-melvin, line 47: In an attempt to not use up even more vertical real estate, I didn't include the unit here, as we have in the shipment modals... but should I include it?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
```diff
@@ -15,35 +20,72 @@ export const QuantityReturnedTableComponent = ({
     line: Partial<GeneratedInboundReturnLineNode> & { id: string }
   ) => void;
 }) => {
+  // what to do here if there are multiple items? Some might have PV and others not...
+  // const { packVariantExists } = usePackVariant(item?.id || '', null);
```
lache-melvin, line 24: From what I can see, the pack variant implementations so far take the itemId, and if there are pack variants set for that item, we set the column as a unit-variant-and-pack-size column, and if not, then the usual pack-size column. 

That would work for the add by item/edit line use cases, but when adding from outbound shipment, we can have lots of items... some might have pack variants configured, and some might not. How should we display this column then?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
```diff
@@ -15,35 +21,76 @@ export const QuantityReturnedTableComponent = ({
...
+              'numberOfPacks',
+              {
+                label: 'label.pack-quantity-issued',
+                width: 110,
+                accessor: ({ rowData }) => rowData.numberOfPacksIssued ?? '--',
+                Cell: BasicCell,
+              },
+            ],
+          ] as ColumnDescription<GeneratedInboundReturnLineNode>[])
+        : []),
```
lache-melvin, line 86: Is there a better way to do this piece, only showing the Qty issued column if the data is present?

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
```diff
@@ -9,25 +9,36 @@ import {
...
+    isOpen,
+    entity: itemId,
+    mode,
+  } = useEditModal<string>();
 
-  const onAddItem = () => {};
+  const onRowClick = (row: InboundReturnLineFragment | InboundReturnItem) =>
+    onOpen(row.itemId);
+
+  const onAddItem = () => onOpen();
```
CarlosNZ, line 41: Why do we need this as a separate function definition? Can't you just have `onAddItem={onOpen}` below? 
Or are we going to be adding more to this function?

client/packages/invoices/src/Returns/modals/InboundReturn/ItemSelector.tsx
CarlosNZ: You mean the number of available units? It's in the dropdown when selecting an item -- I wonder if we can just keep it in there once the item is selected?
Screenshot 2024-03-18 at 4 33 59 PM
(i.e. make it show up in here, like the Batch is:)
Screenshot 2024-03-18 at 4 34 34 PM

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
CarlosNZ: Good question, maybe discuss tomorrow in stand-up?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
```diff
@@ -15,35 +21,76 @@ export const QuantityReturnedTableComponent = ({
...
+                : null,
+            }),
+        },
+      ],
+      [
+        'packSize',
+        {
+          width: 100,
+          setter: updateLine,
+          Cell: PackSizeInputCell,
```
CarlosNZ, line 60: I don't think pack size would be editable here, would it?

So it would be just a "NumberCell" rather than "NumberInputCell"?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
CarlosNZ: Can't think of a better way 🤷‍♂️

CarlosNZ (commented):
Looks great -- I just had a couple of small comments, questions.

client/packages/invoices/src/Returns/modals/InboundReturn/ItemSelector.tsx
lache-melvin: Oh sorry, that was a bit unclear, I meant like we have in shipments:

Screenshot 2024-03-19 at 11 17 39 AM

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
lache-melvin: @andreievg does this come up anywhere else?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
lache-melvin: Oh I thought it would be? We're introducing new stock, it could have any pack size? Pack size is editable for Inbound Shipments, so just went with that...

client/packages/invoices/src/Returns/modals/InboundReturn/ItemSelector.tsx
andreievg: I think it's ok not to include, especially since pack variant thing would show units/pack size together

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
```diff
@@ -62,3 +109,21 @@ export const QuantityReturnedTableComponent = ({
 };
 
 export const QuantityReturnedTable = React.memo(QuantityReturnedTableComponent);
+
+// Input cells can't be defined inline, otherwise they lose focus on re-render
```
andreievg, line 113: Nice catch, had similar with pack variant

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
andreievg: Ouch, nope this is something new, I think fir right now we can just have the `label.punit-variant-and-pack-size` label regardless

 Nice one, for already finding the pack variant stuff

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
andreievg: Yep editable, i guess there is a question should it be editable if we are adding fresh return from outbound line selection, but I think it's easier to just make it editable for both for now

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
andreievg: I played around with this syntax a bit, I've never been happy with how it looks. Ideally useColumns should just skip a null. Would width = 0 work (or is that hacky). Anyways ok with what you have here for now

andreievg (approved):
Looks great thanks.

I had some thought about differentiating between the two types of modal use (adding by item or by outbound line), but I think it's best talked about in person. Not a blocker for now

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
lache-melvin: Ok cool have put that label in, and added a note in another issue [here](https://github.com/msupply-foundation/open-msupply/issues/3212#issuecomment-2008005859) to think more about how we might display this...

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
lache-melvin: Haha I mean this way feels a bit hacky to me too - you're width 0 idea feels kind of cleaner to me, but it doesn't hide the column heading (it just overlaps the next column 😅)

PR #3313, opened 2024-03-18
Use a TooltipTextCell to display batch columns

### LINKED ISSUE #3174: Outbound Shipment: can't see the whole batch in table
## What went wrong? 😲

Screenshot 2024-03-05 at 7 00 30 PM

batch number is quite a critical data to see, should always be visible entirely

## Expected behaviour 🤔

batch number is quite a critical data to see, should always be visible entirely

## How to Reproduce 🔨

Create a outbound shipment for an item with batch number that have > 4 or 5 characters. 

## Your environment 🌱

- Version: 1.6.03
- mSupply version: Desktop (Windows)

clemens-msupply:
Fixes #3174 

Without increasing the column width `TooltipTextCell` can display up to 8 char. Plus there is a tooltip now:

image

Note, I made a comment regarding padding:
https://github.com/msupply-foundation/open-msupply/issues/3174#issuecomment-2002643639
TooltipTextCell doesn't have this padding for which reason I kept the column width.

Also should we use TooltipTextCell as a default? Probably other columns suffer the same problem?

### DIFF
--- a/client/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
+++ b/client/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
@@ -3,7 +3,7 @@ import { ColumnAlign, ColumnFormat } from '../columns/types';
 import { Formatter } from '@common/utils';
 import { RecordWithId } from '@common/types';
 import { ColumnDefinition } from '../columns/types';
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
Thanks Clemens~

clemens-msupply:
@mark-prins just to keep you in the loop about my padding question...

PR #3318, opened 2024-03-18
3190 print returns

### LINKED ISSUE #3190: Print inbound/outbound returns
Epic #2897 

Implement the printing feature for both inbound and outbound returns!

The detail view should have a `Print` button in the toolbar, which prints the return.

lache-melvin:
Fixes #3190

# 👩🏻‍💻 What does this PR do? 
 

- Adds print button to outbound return detail view
- Adds new report context types for returns

I've created a bit of a dummy report template... how best to share? SQL insert statement?

Not sure how much time to spend on this right now - the basics are in place that if someone wanted to create a returns report, they could. 

However, the Invoice DefaultQuery is probably insufficient... I'd imagine you would want to include return reasons in a returns report! I think for this, we'd need to create a return_reasons loader, against the return_reason_id on the invoice resolver. Should this be done now, or wait until someone decides they want this report?

Also, how _do_ we load new reports into the central server?

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -5177,7 +5177,9 @@ export type ReportConnector = {
 export enum ReportContext {
   Asset = 'ASSET',
   Dispensary = 'DISPENSARY',
+  InboundReturn = 'INBOUND_RETURN',
   InboundShipment = 'INBOUND_SHIPMENT',
+  OutboundReturn = 'OUTBOUND_RETURN',
   OutboundShipment = 'OUTBOUND_SHIPMENT',
   Patient = 'PATIENT',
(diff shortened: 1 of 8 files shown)


mark-prins (approved):
🥇 The changes look fine - I have the same questions though!
Ideally the report would use a custom query, so there's no need to augment the default queries to include returns. Can the query use the existing returns queries though? 
Also, yes. there is no way to load the reports. This should be a central server function : I think raise a new issue for that one.

lache-melvin:
> Can the query use the existing returns queries though?

Not really, the invoice queries include a `return_reason_id` field, but we don't yet resolve that to a reason. I'll create another issue for this, I think we're going to want it in the detail view as well 👍

PR #3322, opened 2024-03-18
Print QR code labels

### LINKED ISSUE #3145: QR code generation and printing to a label printer
Looks like there's a ZPL command to print a barcode...
https://gist.github.com/metafloor/773bc61480d1d05a976184d45099ef56
https://support.zebra.com/cpws/docs/zpl/BQ_Command.pdf

mark-prins:
Fixes #3145

# 👩🏻‍💻 What does this PR do? 
 
Adds the setting to save a label printer - this is the bare minimum, you can save one per site, it's a network enabled printer only and we only support zebras.

The settings page was getting messy, so I've tidied that up with Accordions, and have moved the server version to be top right:

Screenshot 2024-03-18 at 5 09 08 PM

Screenshot 2024-03-18 at 5 09 01 PM

The print button on the equipment detail page is now 'Print QR code' and will do just that

image

There's a success & an error toast on the print event.

If there is no printer enabled - I've done a version of the disabled notification that we've talked about:

disabled-notification

I'm testing with a GK420t, the king of label printers.
The settings I'm using are

Screenshot 2024-03-18 at 6 01 18 PM

The label height affects the vertical offset, but I've done nothing more complex than that. The QR code size is affected by the data and the magnification factor, I've used settings which look ok on our labels.

Check that the QR prints and has the correct value.

## 📃 Documentation

Part of the asset feature


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -39,6 +40,7 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
       step = 1,
       multiplier = 10,
       value,
+      noFormatting = false,
```
mark-prins, line 43: adding this - without it the port number is reformatted at `9,100` which isn't what I want at all

client/packages/host/src/Admin/DisplaySettings.tsx
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
```
mark-prins, line 1: this is just moved out of `Settings` which was the original file. 
All the new settings were created as their own component, so I'm doing the same with the display settings

client/packages/host/src/Admin/LabelPrinterSettings.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+export const LabelPrinterSettings = () => {
+  const t = useTranslation('app');
+  const { error, success } = useNotification();
+  const { mutateAsync: updateSettings } =
+    useHost.settings.updateLabelPrinterSettings();
+  const { data: settings } = useHost.settings.labelPrinterSettings();
+  const [draft, setDraft] = React.useState<LabelPrinterSettings>({
+    address: '',
+    labelHeight: 290,
```
mark-prins, line 30: the defaults of our printer 🤷

server/service/src/print/jetdirect.rs
```diff
@@ -0,0 +1,89 @@
+use anyhow::Result;
```
mark-prins, line 1: from the crate `zebrasend`

jmbrunskill:
I'm getting this error

image

-> Access Denied

Which doesn't give me much to go on!

Added some logging to find

2024-03-22 12:25:00.434951000 [ERROR] <server::print:68>:No default store found for user, or default store is not active on current site

Would be good to display errors a bit better to at least have a path forward.
This seems like an unnecessary restriction? You can't print QR codes if you aren't in your default store?

server/service/src/print/jetdirect.rs
jmbrunskill: This confused me! But I get it now, you've copied this from https://github.com/fearful-symmetry/zebrasend/blob/main/src/cmd/jetdirect.rs
I think we should have a comment in the file referencing where we copied it from!

I don't think we need to implement all this stuff, especial things like guessing if we should timeout etc.
We only need the `print` method part for now right?

If we know exactly what command we're sending I think we could look for the response for that specific command rather than relying on guessing?

server/service/src/print/jetdirect.rs
```diff
@@ -0,0 +1,89 @@
...
+        if !resp.is_empty() {
+            println!("{}", resp);
+        }
+
+        Ok(())
+    }
+
+    pub fn send_string(&self, data: String, mode: Mode) -> Result<()> {
+        let mut telnet = Telnet::connect((self.addr.clone(), self.port), 512)?;
+        self.send_command_and_print(data, &mut telnet, mode, Duration::new(0, 500))
```
jmbrunskill, line 67:
```suggestion
        self.send_command_and_print(data, &mut telnet, mode, Duration::new(0, PRINTER_TIMEOUT))
```

jmbrunskill: I wonder what worst case connection times might be for our clients? is 500ms enough?

server/service/src/print/jetdirect.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+        if !resp.is_empty() {
+            println!("{}", resp);
+        }
+
+        Ok(())
+    }
+
+    pub fn send_string(&self, data: String, mode: Mode) -> Result<()> {
+        let mut telnet = Telnet::connect((self.addr.clone(), self.port), 512)?;
```
jmbrunskill, line 66:
```suggestion
        let mut telnet = Telnet::connect_timeout((self.addr.clone(), self.port), 512)?;
```

You actually need to do something like 

SocketAddr::new(IpAddr::V4(Ipv4Addr::from_str(self.addr.clone())
                                .expect("Invalid address")), self.port)

But that could be done once and stored in the `Jetdirect` struct...

This isn't essential as it does seem to timeout after 60s for me anyway

server/service/src/print/jetdirect.rs
jmbrunskill: I'd be temped to re-arrange this into something more taylored for our use case....

client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx
```diff
@@ -4,54 +4,67 @@ import {
...
-        <ReportSelector context={ReportContext.Asset} onPrint={printReport}>
-          <LoadingButton
-            variant="outlined"
-            startIcon={<PrinterIcon />}
-            isLoading={isPrinting}
-          >
-            {t('button.print')}
-          </LoadingButton>
-        </ReportSelector>
+        <ButtonWithIcon
```
jmbrunskill, line 61:
```suggestion
        <LoadingButton
```
It's easy to double click this and print twice, would be good to avoid that...

jmbrunskill (changes requested):
Awesome stuff Mark.
Not quite sure if I should approve or request changes :)

I think the idea that you have to be in your default store is a bit of an unnecessary restriction, but happy to be convinced otherwise.

Would also be keen to at least reference the JetDirect source in the code base given it a copy and paste, but think we might be better of making a structure a bit more specific to our needs? For example do you think we'll need to send SGD commands in future?

client/packages/host/src/Admin/LabelPrinterSettings.tsx
mark-prins: nice idea - will see if there's something I can add

server/service/src/print/jetdirect.rs
mark-prins: if you have a suggestion on how to restructure - happy to hear it!
for now, just thought to implement the simple POC version

server/service/src/print/jetdirect.rs
```diff
@@ -0,0 +1,89 @@
+use anyhow::Result;
+use std::io::Write;
+use std::time::Duration;
+use telnet::{Event, Telnet};
+
+pub struct Jetdirect {
+    pub addr: String,
```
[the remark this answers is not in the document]
mark-prins: thanks - changed

mark-prins:
thanks for the review @jmbrunskill!

I've added in a test button, and propagated the errors, so we now get errors showing here and when printing:

https://github.com/msupply-foundation/open-msupply/assets/9192912/49899581-e3b9-4a76-9ef7-5d850e84e4f9


> I think the idea that you have to be in your default store is a bit of an unnecessary restriction, but happy to be convinced otherwise.

That had come from the cold chain. Didn't know that it was possible to not have a default store tbh! have simply removed all of that validation - you can now print if you have a valid token in a the auth cookie. I think that printing to a label printer is pretty low risk. 

> Would also be keen to at least reference the JetDirect source in the code base given it a copy and paste, but think we might be better of making a structure a bit more specific to our needs? For example do you think we'll need to send SGD commands in future?

Done - and have used SGD to get printer info for the test code 😉

jmbrunskill (approved):
Thanks for the changes, looks good to me!

PR #3329, opened 2024-03-18
renaming

clemens-msupply:
Mark suggestions


client/packages/common/src/intl/locales/es/inventory.json
```diff
@@ -2,7 +2,6 @@
   "label.finalised": "Finalizado",
   "heading.description": "Descripción",
   "label.batch": "Lote",
-  "label.location": "Ubicación",
```
clemens-msupply, line 5: Would the translation tool removed it automatically if you would have left it in?

mark-prins: only if `yarn i18n-unused-remove` was run - we don't remove automatically afaik

PR #3334, opened 2024-03-19
Implement master_list_line try_translate_pull_delete

### LINKED ISSUE #3306: Merging two items appears to create a duplicate of kept item in master lists
## What went wrong? 😲

After merging an item into `Acetylsalicylic Acid 100mg tabs`, when I view masters lists on omSupply with the kept item it appears twice.

image

Similar to https://github.com/msupply-foundation/open-msupply/issues/3276, this is because the master_list_item_join that was to the old item still exists but it's item_link now points to the kept item.

## Expected behaviour 🤔

After 2 items are merged, the deleted item is no longer visible and the kept item appears once.

Solving this is likely easiest by deleting what is effectively a duplicate master_list_item_join

## How to Reproduce 🔨

1. Sync omSupply with a site
2. On primary server merge 2 items that are both on the same master list
3. Sync OMS, the kept item will appear twice.

## Your environment 🌱

- Version: 1.7 dev (~rc2)
- mSupply version: 7.13 dev

- Platform: mac firefox

- Database type: sqlite3

clemens-msupply:
Fixes #3306

# 👩🏻‍💻 What does this PR do? 
Adds support to delete master_list_lines. The delete request were coming from the central server but weren't handled by omSupply.

See issue on how to reproduce the bug

### DIFF
--- a/server/service/src/sync/test/mod.rs
+++ b/server/service/src/sync/test/mod.rs
@@ -345,6 +345,9 @@ pub(crate) async fn check_records_against_database(
             ProgramRequisitionSettings => {
                 check_delete_record_by_id!(ProgramRequisitionSettingsRowRepository, con, id)
             }
+            MasterListLine => {
+                check_delete_record_by_id_option!(MasterListLineRowRepository, con, id)
+            }
             MasterListNameJoin => {
                 check_delete_record_by_id_option!(MasterListNameJoinRepository, con, id)
             }
--- a/server/service/src/sync/translation_and_integration.rs
+++ b/server/service/src/sync/translation_and_integration.rs
@@ -275,6 +275,7 @@ impl PullDeleteRecord {
             ProgramRequisitionSettings => {
                 ProgramRequisitionSettingsRowRepository::new(con).delete(id)
             }
+            MasterListLine => MasterListLineRowRepository::new(con).delete(id),
             MasterListNameJoin => MasterListNameJoinRepository::new(con).delete(id),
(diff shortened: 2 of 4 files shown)


mark-prins (approved):
LGTM! 😄

Chris-Petty:
@clemens-msupply @mark-prins this approach irks me I think irrationally, it's probably fine! I am just hung up on the thinking that merge message processing should handle everything - here happens to central send deletes for a bunch of stuff on merge via change log that it never used to via queue to OG desktop/mobile. Dunno if we can guarantee master lists will always be in the change log either (and thus sending this delete), already getting complaints of some new sites having to initialise 100K records because they have a lot of big master lists.

clemens-msupply:
You mean when receiving an item MERGE record we go through the master_lists on the remote server and update the master_list_lines according to the merge record?

Disadvantage would be that we have to implement this logic twice once in mSupply central and once in omSupply (at least without a omSupply central server). However, eventually we have to implement it anyway for omSupply central.

Chris-Petty:
> You mean when receiving an item MERGE record we go through the master_lists on the remote server and update the master_list_lines according to the merge record?
> 
> Disadvantage would be that we have to implement this logic twice once in mSupply central and once in omSupply (at least without a omSupply central server). However, eventually we have to implement it anyway for omSupply central.

@clemens-msupply yea basically. Just delete the effectively duplicate join (the one that pointed to the "deleted" item).

PR #3336, opened 2024-03-19
3191 generate other returns half

### LINKED ISSUE #3191: Generate other "half" of in/outbound returns with processors

When we create a return, the inverse return should be created for the supplier/customer on the other end of the transaction.

- [x] Similar to shipments, create the opposite return for the other party using processors.

- [x] Also, we should review how we're handling/displaying manually vs automatically created inbound returns! Currently, `isManuallyCreated` code is commented out, as we don't have a `linkedShipment`... linkedReturn?

- [x] ALSO will need to add `theirReference` - not currently querying for that!

lache-melvin:
Fixes #3191

# 👩🏻‍💻 What does this PR do? 
 

Expands the inbound/outbound shipment transfer processors to also support returns.

Updates some of the inbound return detail UI now that linkedShipment is available (to be updated to linkedInvoice or something better named!)

- Create an outbound return, and set it to picked
- Switch to the store you created the return for - you should see a generated inbound return
- Switch to the outbound return, set to shipped
- Switch to inbound return - can now edit
- Deleting a picked outbound return should also delete the generated inbound
- Outbound UI doesn't have support for comments/references yet but can test this if you edit DB directly 😁 
- Create a supplier credit in OG, see that that the resulting Inbound Return displayed in omSupply

Things I need to ask about tomorrow:
- [x] How do I test this syncs between different sites?
  - [x] Return reasons are configured centrally, so the return_reason_id will map to the same reason for everyone right??
- [x] How do I test this maps correctly from OG to OMS?
- [x] I think OMS to OG looks all good, but it's also possible I'm missing something, would be good to validate with someone
- [x] Do we want the linked requisition to get passed through to the return as well? In that case, I think we'll need to store the requisition_id when we create the returns?


server/service/src/processors/transfer/shipment/create_inbound_shipment.rs
```diff
@@ -89,6 +92,12 @@ impl ShipmentTransferProcessor for CreateInboundShipmentProcessor {
             &outbound_shipment,
             record_for_processing,
             request_requisition,
+            match outbound_shipment.invoice_row.r#type {
+                InvoiceRowType::OutboundShipment => InvoiceRowType::InboundShipment,
+                InvoiceRowType::OutboundReturn => InvoiceRowType::InboundReturn,
+                // TODO: is panicking here ok or should we return Ok(None)? (This should never happen bc of check above...)
```
andreievg, line 98: Maybe something like this would help https://github.com/msupply-foundation/open-msupply/compare/3191-generate-other-half...3191-generate-other-half(Example-of-intermediary-type) ?

andreievg (commented):
Looks great, I have a small suggestion to avoid using unimplemented

andreievg (approved):

client/packages/invoices/src/Returns/InboundDetailView/Footer/Footer.tsx
```diff
@@ -65,7 +68,11 @@ export const FooterComponent: FC = () => {
           >
             <OnHoldButton />
             <StatusCrumbs
-              statuses={inboundReturnStatuses}
+              statuses={
+                isManuallyCreated
+                  ? manualInboundReturnStatuses
+                  : inboundReturnStatuses
```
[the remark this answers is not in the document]
lache-melvin: Oo - do you mean because we would only ever see the automatically created Inbound Returns once they are picked?

Yeah it was copied from Inbound Shipment, and we should probably stay consistent with that... but you're right we would never actually see it in `New` status. Maybe you want to see the date that the Outbound Return was created vs when it was picked?

server/service/src/processors/transfer/shipment/create_inbound_shipment.rs
lache-melvin: Ah yes! That's much nicer thank you :)

Issue #3002, opened 2024-02-16
Refactor - Use Middleware for Authentication


jmbrunskill:
## The suggestion

In Health Supply Hub, we use an authentication middleware in Actix, that allows graphql and non-graphql environments to share the same authentication context. This is good for requests such as excel downloads that might need the same login to persist between downloads and graphql requests.

## Example use case

In https://github.com/msupply-foundation/open-msupply/pull/3001 we've added logic that duplicates the refesh cookie logic previously only available in graphql. Moving to authentication context middleware would move this logic to the server.

https://github.com/msupply-foundation/health-supply-hub/blob/develop/backend/server/src/middleware/authentication.rs

### Why should we invest time in this?

- Risk of code change only being made to one part of the application resulting in incorrect authentication logic
- Allows us to more easily handle authentication variations such as API header vs Login Cookie in future

### Are there any risks associated with this change?

I think this is relatively low risk as the pattern has been proven in HSH, Notify and Universal Codes

### How much effort is required?

Change should be relatively simple given we have it in HSH already.
However, if we also want to add an authentication cookie to the context, it will require more time review and testing.

PR #3337, opened 2024-03-19
3164 file uploads backend

### LINKED ISSUE #3164: File Uploads - Backend

For cold chain asset management, we need to be able to upload and sync documents.
This issue is just for the uploading part, sync will be a separate issue

See KDD: https://github.com/msupply-foundation/open-msupply/blob/develop/decisions/2024-02-28_file_sync.md

### Acceptance Criteria 💌

- [x] Uploads should be available from either the central server or remote server
- [x] Uploaded file ids should be recorded in a `sync_file_reference` table
- [x] Uploaded files should be associated with record id and associated data table type e.g. (`asset`, `asset_log`, `invoice`, etc)
- [x] File sizes should be limited (to avoid too big files locking up sync later on)

### Moneyworks Jobcode 🧰

OMSCC:AST

jmbrunskill:
Fixes #3164

# 👩🏻‍💻 What does this PR do? 
 

Allows uploading files to server which will be registered for sync.

Each file is associated with a `table_name` and `record_id` which the sync process will use to determine where the files should be synced to/from.

This PR doesn't include any of the sync code, this will be a later PR.

A new endpoint is available which has this structure...

`/sync_files/{table_name}/{record_id}`

To get an uploaded file you can call
http://{YOURSERVERURL}:{YOURPORT}/sync_files/{table_name}/{record_id}?id={FILE_ID}

To upload a file you need to use a multipart upload.
In Postman this looks something like this...

image

This will return the uploaded file id.

Right now we don't have any security implemented on this.
Thinking maybe should look at https://github.com/msupply-foundation/open-msupply/issues/3002 and use a cookie for auth?

We also don't have any validation on record_id or table name. 

To test you need to try uploading a file using postman or similar.

This might work too?
```
curl --location --request POST 'http://localhost:8082/sync_files/asset/asset1_id' \
--form 'files=@"/path/to/your/file"'
```
Then use the file id returned to construct your download url as above.

If you're not re-initialising please run this..

```
            CREATE TABLE sync_file_reference (
                id TEXT NOT NULL PRIMARY KEY,
                table_name TEXT NOT NULL, -- Associated Table
                record_id TEXT NOT NULL, -- Associated record id
                file_name TEXT NOT NULL,
                created_datetime TIMESTAMP NOT NULL, -- No modified datetime, as we don't allow updates it would break sync
                deleted_datetime TIMESTAMP
            );
```

- [ ] is it ok to have no validation of table_name and/or record_id at this stage? Tried using ChangeLogTableName but there's no string -> ChangeLogTableName established. In sync translator we define a table name, but that's not easy to use in this conext, maybe a re-factor to change how this works?
- [ ] How should we do security on this?

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -79,7 +79,7 @@ pub enum ChangelogTableName {
     AssetType,
     AssetCatalogueItem,
     #[default]
-    SyncFileReferenceRow,
+    SyncFileReference,
 }
 
 #[derive(Clone, Queryable, Debug, PartialEq, Insertable)]
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -7,6 +7,7 @@ use diesel::{
     helper_types::{IntoBoxed, LeftJoin},
     prelude::*,
 };
+use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 use util::inline_init;
 
@@ -46,7 +47,7 @@ pub enum ChangelogAction {
     Delete,
 }
 
-#[derive(DbEnum, Debug, Clone, PartialEq, Eq)]
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[DbValueStyle = "snake_case"]
 pub enum ChangelogTableName {
     Number,
@@ -77,6 +78,8 @@ pub enum ChangelogTableName {
     AssetCategory,
     AssetType,
     AssetCatalogueItem,
+    #[default]
+    SyncFileReferenceRow,
 }
 
 #[derive(Clone, Queryable, Debug, PartialEq, Insertable)]
--- a/server/repository/src/db_diesel/sync_file_reference_row.rs
+++ b/server/repository/src/db_diesel/sync_file_reference_row.rs
@@ -1,6 +1,5 @@
 use super::sync_file_reference_row::sync_file_reference::dsl::*;
 
-use crate::ChangelogTableName;
 use crate::RepositoryError;
 use crate::StorageConnection;
 
@@ -11,7 +10,7 @@ use serde::{Deserialize, Serialize};
 table! {
     sync_file_reference (id) {
         id -> Text,
-        table_name -> crate::db_diesel::changelog::ChangelogTableNameMapping,
+        table_name -> Text,
         record_id -> Text,
         file_name -> Text,
         created_datetime -> Timestamp,
@@ -25,7 +24,7 @@ table! {
 #[table_name = "sync_file_reference"]
 pub struct SyncFileReferenceRow {
     pub id: String,
-    pub table_name: ChangelogTableName,
+    pub table_name: String, // Maybe should be ChangeLogTableName ? But that causes mapping problems in the api request...
     pub record_id: String,
     pub file_name: String,
     pub created_datetime: NaiveDateTime,
--- a/server/repository/src/db_diesel/sync_file_reference_row.rs
+++ b/server/repository/src/db_diesel/sync_file_reference_row.rs
@@ -13,6 +13,7 @@ table! {
         table_name -> Text,
         record_id -> Text,
         file_name -> Text,
+        mime_type -> Nullable<Text>,
         created_datetime -> Timestamp,
         deleted_datetime -> Nullable<Timestamp>,
     }
@@ -24,9 +25,10 @@ table! {
 #[table_name = "sync_file_reference"]
 pub struct SyncFileReferenceRow {
     pub id: String,
-    pub table_name: String, // Maybe should be ChangeLogTableName ? But that causes mapping problems in the api request...
+    pub table_name: String,
     pub record_id: String,
     pub file_name: String,
+    pub mime_type: Option<String>,
     pub created_datetime: NaiveDateTime,
     pub deleted_datetime: Option<NaiveDateTime>,
 }
--- a/server/repository/src/db_diesel/sync_file_reference_row.rs
+++ b/server/repository/src/db_diesel/sync_file_reference_row.rs
@@ -1,0 +1,86 @@
+use super::sync_file_reference_row::sync_file_reference::dsl::*;
+
+use crate::ChangelogTableName;
+use crate::RepositoryError;
+use crate::StorageConnection;
+
+use chrono::NaiveDateTime;
+use diesel::prelude::*;
+use serde::{Deserialize, Serialize};
+
+table! {
+    sync_file_reference (id) {
+        id -> Text,
+        table_name -> crate::db_diesel::changelog::ChangelogTableNameMapping,
+        record_id -> Text,
+        file_name -> Text,
+        created_datetime -> Timestamp,
+        deleted_datetime -> Nullable<Timestamp>,
+    }
+}
+
+#[derive(
+    Clone, Insertable, Queryable, Debug, PartialEq, AsChangeset, Eq, Default, Serialize, Deserialize,
+)]
+#[table_name = "sync_file_reference"]
+pub struct SyncFileReferenceRow {
+    pub id: String,
+    pub table_name: ChangelogTableName,
+    pub record_id: String,
+    pub file_name: String,
+    pub created_datetime: NaiveDateTime,
+    pub deleted_datetime: Option<NaiveDateTime>,
+}
+
+pub struct SyncFileReferenceRowRepository<'a> {
+    connection: &'a StorageConnection,
+}
+
+impl<'a> SyncFileReferenceRowRepository<'a> {
+    pub fn new(connection: &'a StorageConnection) -> Self {
+        SyncFileReferenceRowRepository { connection }
+    }
+
+    #[cfg(feature = "postgres")]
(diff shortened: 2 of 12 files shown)


server/repository/src/migrations/v1_08_00/sync_file_reference.rs
```diff
@@ -0,0 +1,19 @@
...
+    sql!(
+        connection,
+        r#"
+            CREATE TABLE sync_file_reference (
+                id TEXT NOT NULL PRIMARY KEY,
+                table_name TEXT NOT NULL, -- Associated Table
+                record_id TEXT NOT NULL, -- Associated record id
+                file_name TEXT NOT NULL,
+                created_datetime TIMESTAMP NOT NULL, -- No modified datetime, as we don't allow updates it would break sync
+                deleted_datetime TIMESTAMP
```
jmbrunskill, line 13: Do we want to capture other things in here?
Maybe a category for the file in case the same record has more than one type of file associated with it?
Mime type might be handy?
File description could be good too, but depends if we're going to create UI to manage these?

server/server/src/static_files.rs
```diff
@@ -70,3 +89,127 @@ async fn plugins(
...
+            table_name: table_name.clone(),
+            // content_type: file.mime_type,
+            created_datetime: chrono::Utc::now().naive_utc(),
+            deleted_datetime: None,
+            record_id: record_id.clone(),
+        });
+        match result {
+            Ok(_) => {}
+            Err(err) => {
+                // TODO: delete the files that were uploaded?
```
jmbrunskill, line 180: What should we do here?

server/service/src/static_files.rs
```diff
@@ -65,19 +138,19 @@ impl StaticFileService {
...
 
 /// Returns the file name part of the path like:
-/// `./static_file_path/{ui}_{file_name};
-fn parse_original_file_name(id: &str, file_path: &PathBuf) -> Option<String> {
+/// `./static_file_path/{uuid}_{file_name};
+fn parse_original_file_name(id: &str, file_path: &Path) -> Option<String> {
     let file_name = file_path.file_name()?.to_string_lossy();
     let name = &file_name[id.len() + 1..];
-    if name.len() == 0 {
+    if name.is_empty() {
```
jmbrunskill, line 145: Some of these are changes from HSH code based on clippy suggestions.
Bit of copying over happening here...

server/server/src/static_files.rs
```diff
@@ -1,28 +1,48 @@
...
 use service::settings::Settings;
-use service::static_files::StaticFileService;
+use service::static_files::{StaticFileCategory, StaticFileService};
 
 // this function could be located in different module
 pub fn config_static_files(cfg: &mut web::ServiceConfig) {
     cfg.service(web::resource("/files").guard(guard::Get()).to(files));
     cfg.service(plugins);
+    cfg.service(
+        web::resource("/sync_files/{table_name}/{record_id}")
```
jmbrunskill, line 31: Is `sync_files` a good path name here?

server/server/src/middleware/content_length_limit.rs
```diff
@@ -0,0 +1,98 @@
...
+use actix_web::{
+    body::EitherBody,
+    dev::{forward_ready, Service, ServiceRequest, ServiceResponse, Transform},
+    http::StatusCode,
+    Error, HttpResponse,
+};
+use futures_util::future::LocalBoxFuture;
+use futures_util::TryStreamExt;
+
+static DEFAULT_MAX_UPLOAD_SIZE_BYTES: u64 = 100 * 1024 * 1024; // 100MB
```
jmbrunskill, line 12: What should our default limit be? Should we make it configurable some how?

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -77,6 +78,8 @@ pub enum ChangelogTableName {
     AssetCategory,
     AssetType,
     AssetCatalogueItem,
+    #[default]
+    SyncFileReferenceRow,
```
mark-prins, line 82: I'm unsure why the naming differs from the others.. Does the `Row` mean anything? 
Is this just placeholder code? can't see the corresponding usage in the translations file

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+#[derive(
+    Clone, Insertable, Queryable, Debug, PartialEq, AsChangeset, Eq, Default, Serialize, Deserialize,
+)]
+#[table_name = "sync_file_reference"]
+pub struct SyncFileReferenceRow {
+    pub id: String,
+    pub table_name: String, // Maybe should be ChangeLogTableName ? But that causes mapping problems in the api request...
```
mark-prins, line 27: table_name is fine by me - I'm not concerned at having it not reference the changelog here

server/repository/src/migrations/v1_08_00/sync_file_reference.rs
mark-prins: some sort of type would be good. For the CCEI example, we need to distinguish an image and a document.
`mime_type` would be a good option - allows the consumer to make decisions and possibly categorise

server/server/src/middleware/content_length_limit.rs
mark-prins: I think yes, and 'not now' 😄 
When we have shifted admin to the central server we could have a configuration option?

mark-prins: 100MB is more than I would have allowed.. probably fine tho

server/server/src/static_files.rs
mark-prins: it's growing on me - my initial reaction: didn't love it.
Partly because 'sync' is more commonly a verb, when the intention is an adjective i.e. 'synchronised files' and I read it as 'synchronise files'
naming is tricky - I would have gone just 'files' which is already taken. 
so.. yeah? keep it as 'sync_files' 🤷

server/server/src/static_files.rs
```diff
@@ -70,3 +95,127 @@ async fn plugins(
...
+) -> Result<Vec<UploadedFile>, Error> {
+    let service = StaticFileService::new(&settings.server.base_dir)
+        .map_err(|err| InternalError::new(err, StatusCode::INTERNAL_SERVER_ERROR))?;
+
+    let mut files = Vec::new();
+
+    while let Some(mut field) = payload.try_next().await? {
+        // A multipart/form-data stream has to contain `content_disposition`
+        let content_disposition = field.content_disposition();
+        log::info!(
```
mark-prins, line 112: nice! 😄

server/server/src/static_files.rs
mark-prins: I think (hard) delete the uploaded - otherwise the user is encouraged to upload again, and then what happens? 
Ideally, the response is a list of successes and failures so the client can determine what to do.. but if we throw a server error and it's all or nothing, I think just clear the decks and reset the state to the original. ( I think the error is pretty unlikely so it's fine to delete )

server/server/src/static_files.rs
```diff
@@ -70,3 +95,127 @@ async fn plugins(
...
+
+    let files = handle_file_upload(
+        payload,
+        settings,
+        StaticFileCategory::SyncFile(table_name.clone(), record_id.clone()),
+    )
+    .await?;
+
+    let repo = SyncFileReferenceRowRepository::new(&db_connection);
+    for file in files.clone() {
```
mark-prins, line 173: the response has `name: "files"` returned. What is this? seems hard-coded, so is it helpful?

Screenshot 2024-03-20 at 11 14 42 AM

mark-prins (approved):
👍 Looks great thanks James! working well.

yes, ideally the table names would be validated against something, but it's not essential. Likewise the record ids - and possibly too we want a tidy up routine if a file is linked to a record which is actually deleted (not softly) then the file should be deleted too.

server/server/src/static_files.rs
jmbrunskill: We could actually use `files` I think? But Iiked having a separation between sync files and the report downloads, as well as keep the code simple. I guess we could probably get away with changing the other one to something else? like `/tmp_files/`?

server/server/src/middleware/content_length_limit.rs
jmbrunskill: Yeah I thought surely that's big enough, but also not too restrictive...
PDQ PDF is 285MB, but I assume we'd upload parts to different assets not just one huge doc?

server/repository/src/db_diesel/changelog/changelog.rs
jmbrunskill: It's just a mistake, thanks for noticing!

server/server/src/middleware/content_length_limit.rs
mark-prins: some PDFs are massive! let's go with 100MB then 🙏

server/server/src/static_files.rs
mark-prins: I like the separation too.. if anything, the report `/files` should change, since that it really `/reports`

server/server/src/static_files.rs
jmbrunskill: image

PR #3342, opened 2024-03-19
RnD cargo watch

lache-melvin:
# 👩🏻‍💻 What does this PR do? 
 

Adds the cargo watch script I was playing with from last RnD day to omSupply repo 😎 

Install global deps if required, then run `cargo make watch` from the `server` directory.

Make some backend file changes, the server should restart.

Current server should be available until new version has been compiled.


roxy-dao (approved):
So cool +_+

PR #3347, opened 2024-03-20
3187 implement permissions for returns

### LINKED ISSUE #3187: Add permissions for returns, and sync them
Epic #2897 

Initial permissions code was added in #3053, but this was not synced to the central server, so these changes weren't applied.

Sync the new query/mutation permissions for returns to the central server, and ensure the permission code runs correctly. 

Update the `validate_auth` calls on queries and mutations to use the return-related permissions!

From Mark:

> the returns work currently doesn't play with postgres, can you:
> 
> - merge develop into this branch to get the 1.8 migrations
> - add the following to the migrations, for postgres only:
> ```
> ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_MUTATE';
> ALTER TYPE permission_type ADD VALUE 'OUTBOUND_RETURN_QUERY';
>```

lache-melvin:
Fixes #3187

# 👩🏻‍💻 What does this PR do? 
 

Adds/wires up permissions for returns.

Initialise from central and log in.

Try to read/write some inbound and outbound returns. Enable and disable the `Return stock from customer/supplier invoices` permissions and ensure you can/can't save returns accordingly.

Once this is in, I think we're good to go merging into develop 🤓


server/graphql/types/src/types/permissions.rs
```diff
@@ -35,6 +35,8 @@ pub enum UserPermission {
     InboundShipmentMutate,
     OutboundReturnQuery,
     OutboundReturnMutate,
+    InboundReturnQuery,
+    InboundReturnMutate,
```
lache-melvin, line 39: I'd put in most of Outbound a while back, adding Inbound throughout

server/service/src/login.rs
```diff
@@ -394,17 +395,20 @@ fn permissions_to_domain(permissions: Vec<Permissions>) -> HashSet<Permission> {
...
+                output.insert(Permission::InboundShipmentMutate);
+            }
+            // returns
             Permissions::ReturnStockFromSupplierInvoices => {
                 output.insert(Permission::OutboundReturnMutate);
             }
-            Permissions::CreateSupplierInvoices => {
-                output.insert(Permission::InboundShipmentMutate);
+            Permissions::ReturnStockFromCustomerInvoices => {
+                output.insert(Permission::InboundReturnMutate);
```
lache-melvin, line 411: Per the mSupply docs, these look like the right permissions to set to check if the user can create a return. 

However in OG it also looks like creating an outbound return (supplier credit) requires the 169 `CreateSupplierCredits` permission, and creating an inbound return (customer credit) requires the 42 `CreateCustomerInvoices` permission. 

I don't quite have my head around the distinction, and whether I should be enforcing both of those OG permissions be active here before inserting the mutate permission?

(Will follow up on this after Zen day 😇)

mark-prins (changes requested):
🎉  working nicely!

I have a request though - as well as the permission denied modal, I also get these angry looking toasts:

Screenshot 2024-03-22 at 9 29 57 AM

You can check user permissions in the front end and alter behaviour before waiting for the api to complain using the `userHasPermission` fn out of `AuthContext`.

I've implemented a new approach in #3307 (see [this comment](https://github.com/msupply-foundation/open-msupply/pull/3307#issuecomment-2005106242)) for alerting, as per #1912 (I'm now thinking that I'll change to a toast. thoughts?)

Would you mind fixing (suppressing) the additional error notifications, and implementing a permission denied prompt instead please?
I've merge

lache-melvin:
> I've implemented a new approach in #3307 (see [this comment](https://github.com/msupply-foundation/open-msupply/pull/3307#issuecomment-2005106242)) for alerting, as per #1912 (I'm now thinking that I'll change to a toast. thoughts?)

Oo cool! My initial thought was that enabled with a toast is definitely consistent with what I've seen throughout the rest of the app, so that seems like the simplest option. Especially if that's still Anisha's guidance?

Interesting reading that thread though, the different preferences people have! I think I'm with you, I'd rather have the visual cue that something isn't allowed, and a way to find out why - the hover/click explanation looks good to me! If we're doing an overhaul of all the buttons, I could get on board with it, but either way I find consistency to be most important. Another thought - what would we do for the dropdown actions that aren't allowed? Still enabled and toast?


> Would you mind fixing (suppressing) the additional error notifications, and implementing a permission denied prompt instead please?
Great call, thanks, will do 👍

lache-melvin:
Was implementing with the toast what you meant @mark-prins? Or just show the permission denied popup modal before even going to the backend?

server/service/src/login.rs
lache-melvin: Resolved, we'll just stick with the `ReturnStockFrom[X]Invoices` permissions for now 🚀

andreievg: Sounds good to me, would be great to have permission mapping doc somewhere at some point (so that when permissions are missing, either the support team or super user can reference vs looking at code), I'll ask on team lead chat what is the best medium of capturing these sorts of points

andreievg: https://github.com/msupply-foundation/msupply_docs/issues/24

mark-prins:
> Or just show the permission denied popup modal before even going to the backend?

this is what I had in mind - no need to go to the backend. except.. that would mean the logic is b/e and f/e which is not ideal.

probably best that we do the api call and alert the user in some way..

mark-prins (approved):
🙏  thanks - looking good! 

My comment might be misleading now.. I was thinking to do just what you've done ( have button enabled and toast if permission denied ) and now I'm wondering if we should hit the backend rather than check permissions in the f/e. Keeps the logic in one place.

ignore that for now tho.. we can discuss and tidy later 😄

lache-melvin:
> 🙏 thanks - looking good!
> 
> My comment might be misleading now.. I was thinking to do just what you've done ( have button enabled and toast if permission denied ) and now I'm wondering if we should hit the backend rather than check permissions in the f/e. Keeps the logic in one place.
> 
> ignore that for now tho.. we can discuss and tidy later 😄

Haha sounds good, definitely a discussion I'd like to come back to! Thanks Mark 🙏

PR #3350, opened 2024-03-20
3326 qty available for return should include amount already in return

### LINKED ISSUE #3326: Quantity Available for Return doesn't include quantity already selected for return
## What went wrong? 😲
 

I add a new item to a return, the first batch has 5 items available:
Image

Let's say I selected 4 pack to return. I hit save, and then open the modal again. Available stock has been reduced, and now I only have 1 pack available to return:

Image

If I actually wanted to return all 5, I can't set that value... in fact I can only set it to 1, as once I try to edit the value, we limit the qty to return max to be what is available!

Image

## Expected behaviour 🤔

If we open the modal in edit mode, the quantity available to return should include the quantity already set aside to return!

## How to Reproduce 🔨

1. Create a new Outbound Return
2. Add a new item, select some stock to return from a stock line of that item
3. Save the line
4. Click the line to open the modal again
5. See quantity available to return is now reduced, and you can't return the full quantity as you would like!

## Your environment 🌱

- Version: 
- mSupply version: 

- Platform:

- Database type:

lache-melvin:
Fixes #3326 

# 👩🏻‍💻 What does this PR do? 
 

Includes the quantity already in the return in the available qty for return.

Adding new item:
Screenshot 2024-03-20 at 3 21 43 PM

Set value to 5:
Screenshot 2024-03-20 at 3 21 50 PM

Save. Click row to re-open modal, available to return is still 10:
Screenshot 2024-03-20 at 3 21 50 PM

### DIFF
--- a/server/graphql/types/src/types/outbound_return_line.rs
+++ b/server/graphql/types/src/types/outbound_return_line.rs
@@ -83,9 +83,7 @@ impl OutboundReturnLineNode {
     }
 
     pub async fn available_number_of_packs(&self) -> f64 {
-        // Quantity available for return should include the number of packs already in the return
-        // (Available stock is reduced as soon as it is added to a return)
-        self.stock_line_row().available_number_of_packs + self.return_line.number_of_packs
+        self.return_line.available_number_of_packs
     }
 
     pub async fn pack_size(&self) -> &i32 {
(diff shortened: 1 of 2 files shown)


andreievg (approved):
Looks great thanks

andreievg (commented):
Would it be a lot of work to add a test for this one ? Would be great to see if it's trivial

lache-melvin:
> Would it be a lot of work to add a test for this one ? Would be great to see if it's trivial

You are so right yes adding now 🚀

Issue #3017, opened 2024-02-19
Fix clippy warnings


roxy-dao:
## The suggestion
Turn on clippy and go through all the warnings to see which make sense to fix up. There is an R&D card for this but yah...

### Why should we invest time in this?
It's nice to not have warnings in the console 👀 and makes everything more consistent.

### How much effort is required?
Maybe a bit of time to go through the 100s of warnings 👀

PR #3355, opened 2024-03-20
3017.1 fix clippy warnings

roxy-dao:
Part of #3017

# 👩🏻‍💻 What does this PR do? 
Some clippy fixes... there are still a lot.........

Make sure tests are still passing and everything is working as it should

Still heaps of clippy changes to be made so this is only part 1!


server/android/src/lib.rs
```diff
@@ -1 +1,2 @@
+#![allow(clippy::all)]
```
clemens-msupply, line 1: Why did you disable them for Android?

roxy-dao: Sorry, I just disabled it for now... will enable again when I do the rest of the clippy fixes 👀

clemens-msupply: ok cool, could you please take it out of the PR though so that we don't forget?

clemens-msupply (approved):
Haven't looked into the change in detail but tests are still passing

Issue #3357, opened 2024-03-20
Discrepancies between stocktake date and created date


roxy-dao:
## What went wrong? 😲
Users currently cannot change stocktake created date in the UI, so the stocktake date is saved in the created date. However, this causes a discrepancy when displaying this in the UI since stocktake date does not have a time and cannot be localised. 

P.S I think most the date times being displayed aren't localised and are shown in naive date?  

## Expected behaviour 🤔
Stocktake date and created date should match up right now since they `should` be the same value currently. I also expect that all datetimes should be local time and not naive time.

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Version: 1.7
- mSupply version: 

- Platform: Browserr

- Database type:

roxy-dao:
OMS: created -> datetime, stockatedate -> date
OG: created -> date, stocktakedate -> datetime

Seems like both need updating? Must be a lot of discrepancy with localised times 🤔  moving to milestone 2.1.0 for now

PR #3358, opened 2024-03-20
911 show created date in the detail panel for invoice, requisition, stocktake windows

### LINKED ISSUE #911: Show created date in the detail panel for Invoice, Requisition, Stocktake windows
The created date shows in the logs, but not in the detail panel on the side. It would be great to have that show in this main window! 
Felt confusing to have something show in the listbox, but not the invoice itself (at least on first glance): 

image

roxy-dao:
Fixes #911

# 👩🏻‍💻 What does this PR do? 
Add created datetime to the side panel and also localise the stocktake list view created datetime for consistency.

- [ ] Go to these views: Outbound Shipment, Inbound Shipment, Requisition, Stocktake, Prescription
- [ ] Go to side panel
- [ ] See created datetime

I've noticed that some datetimes throughout the app might not be localised and have created an issue here to fix it... #3357. Currently the tooltip stocktake date is different from the created datetime since its not localised... probably also want to discuss if we want this enabled some time since stocktake date can differ from created datetime

### DIFF
--- a/client/packages/inventory/src/Stocktake/DetailView/SidePanel.tsx
+++ b/client/packages/inventory/src/Stocktake/DetailView/SidePanel.tsx
@@ -15,6 +15,7 @@ import {
   InfoTooltipIcon,
   DeleteIcon,
   useDeleteConfirmation,
+  useFormatDateTime,
 } from '@openmsupply-client/common';
 import { useStocktake } from '../api';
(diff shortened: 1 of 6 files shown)


mark-prins (approved):
😍  Looks good! thanks Roxy, working nicely

PR #3365, opened 2024-03-21
3361 master list

### LINKED ISSUE #3361: Catalogue -> Master Lists -> Export exports all Master Lists, not just the ones the store has visibility to
## What went wrong? 😲
 

When navigating to Catalogue -> Master Lists -> and clicking the Export button in the top right corner, you get a report of Master Lists. Even if the store has visibility into 2 Master Lists, the export seems to be showing ALL master lists. 

## Expected behaviour 🤔

The report should only be showing the Master Lists the store has visibility into, not all Master Lists. In this particular case, there should only be 2. 

## How to Reproduce 🔨

1. Go to Catalogue -> Master Lists -> click Export
2. Note in omSupply how many Master Lists you have access to
3. Open the report in Excel -> note the number of Master Lists being shown
4. See the discrepancy

## Your environment 🌱

- Version: 1.7.0

- Platform: Windows Server 2022 Desktop

- Database type: ?

roxy-dao:
Fixes #3361

# 👩🏻‍💻 What does this PR do? 
Filters master lists by store in backend since this should only be limited to the current store.

- [ ] Have 2 stores with different master lists (Maybe best to have about 4 master lists and 2 on each store?)
- [ ] Look at the `Master List` on each store (Should only show the master lists that the store uses)
- [ ] Try exporting `Master List` information. (Should be same as above)

### DIFF
--- a/client/packages/system/src/MasterList/api/api.ts
+++ b/client/packages/system/src/MasterList/api/api.ts
@@ -30,7 +30,7 @@ export const getMasterListQueries = (sdk: Sdk, storeId: string) => ({
         offset,
         key,
         desc,
-        filter: { ...filterBy, existsForStoreId: { equalTo: storeId } },
+        filter: filterBy,
         storeId,
       });
       return result?.masterLists;
--- a/server/service/src/master_list/tests/query.rs
+++ b/server/service/src/master_list/tests/query.rs
@@ -1,5 +1,6 @@
 #[cfg(test)]
 mod query {
+    use repository::mock::mock_store_a;
     use repository::{mock::MockDataInserts, test_db::setup_all, MasterListFilter};
     use repository::{EqualFilter, StringFilter};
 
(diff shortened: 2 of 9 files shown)


client/packages/system/src/MasterList/api/api.ts
```diff
@@ -30,7 +30,7 @@ export const getMasterListQueries = (sdk: Sdk, storeId: string) => ({
         offset,
         key,
         desc,
-        filter: { ...filterBy, existsForStoreId: { equalTo: storeId } },
```
roxy-dao, line 33: Doesn't need to be filtered by this anymore since backend is doing it

server/service/src/master_list/tests/query.rs
```diff
@@ -11,64 +12,52 @@ mod query {
...
                 &context,
+                &context.store_id,
                 None,
                 Some(MasterListFilter::new().exists_for_store_id(EqualFilter::equal_to("store_a"))),
                 None,
             )
             .unwrap();
         assert!(result.count >= 1);
-
-        //Test filter for non existent store finds nothing
```
roxy-dao, line 60: Test no longer needed since we're filtering out store_id by default... though do we still need the name_id filter now that we're filtering out by store_id?

mark-prins (approved):
🤩  works nicely thanks

PR #3368, opened 2024-03-21
#3239 Delete Inbound Return lines

### LINKED ISSUE #3239: Delete return lines

Select return lines to delete is in place, but doesn't currently call the mutation. These should call the update[In/Out]boundReturnLines mutation with a numberOfPacks of 0, so they are deleted by the API.

CarlosNZ:
Fixes #3239 (partial)

# 👩🏻‍💻 What does this PR do? 
 

Just implements the Inbound delete lines mutation -- I'm having a bit of trouble getting the Outbound one to work properly (cos it needs some other stuff added first), so will do in a separate PR.

Delete some lines from an Inbound return (using Toolbar action)


client/packages/invoices/src/Returns/api/hooks/line/useDeleteSelectedInboundLines.ts
```diff
@@ -53,16 +64,3 @@ export const useDeleteSelectedInboundReturnLines = (): (() => void) => {
...
-  const queryClient = useQueryClient();
-  const api = useReturnsApi();
-  const queryKey = api.keys.inboundDetail(invoiceNumber);
-
-  // TODO: Replace with actual mutation
-  // return useMutation(api.updateInboundReturnLines, {
-  return useMutation(async (_lines: InboundReturnLineFragment[]) => {}, {
-    onSettled: () => queryClient.invalidateQueries(queryKey),
-  });
-};
```
CarlosNZ, line 68: Seemed unnecessary to make *another* hook when this is just calling an existing hook (updateLines) with a specific value (number of packs = 0)

lache-melvin (approved):
Awesome, ship it!

(Sorry I've also been in the Toolbar, will need to resolve conflicts 🙃)

PR #3369, opened 2024-03-21
3316 accept more return invoice fields on update mutation

### LINKED ISSUE #3316: Make colour/onHold/comment editable for Outbound Return

Update outbound return functionality is currently very minimal, from the first thin wire implementation. 

I want to be able to update the comment, colour, and onHold status of the Outbound Return.

lache-melvin:
Fixes #3316

# 👩🏻‍💻 What does this PR do? 
 

Allows colour, onHold, theirReference, transportReference to be updated for an Outbound Return. Most of these already were already available on the updateInboundReturn mutation, but adds the theirReference field.

Uncommented components in the UI where these are used, but we do have a separate issue to flesh out the side panel, which would be where we would show the transportReference etc.

- In outbound return detail view (when new/picked)...
- Can add a reference
- Clicking onHold
- In side panel...
- Can add comment
- Can change colour
- In inbound return detail view (when new or delivered)...
- Can add a reference

Adds a debounced mutation call for updating the text fields - this was wrapped up in the `fields` hook before... thoughts on doing it this way?


client/packages/invoices/src/Returns/InboundDetailView/Toolbar.tsx
```diff
@@ -49,17 +56,14 @@ export const Toolbar: FC = () => {
...
-                  // disabled={isDisabled}
-                  // value={theirReferenceBuffer ?? ''}
-                  // onChange={event => {
-                  //   setTheirReferenceBuffer(event.target.value);
-                  //   update({ theirReference: event.target.value });
-                  // }}
+                  disabled={isDisabled}
+                  value={theirReferenceBuffer ?? ''}
+                  onChange={event => {
+                    setTheirReferenceBuffer(event.target.value);
```
andreievg, line 64: I've noticed, for the comment in AdditionalInfoSection we don't use state variable for value, and together with debouncing on a slow network (slow 3g throttle), I get this:

https://github.com/msupply-foundation/open-msupply/assets/26194949/40a010b7-0b9c-4da6-8a31-fd48395f2c40

Do you know if this is the same behaviour for other side panels (or any other inputs you've encountered, where we use value = {something from graphql} and then update = {update graphql} ?)

client/packages/invoices/src/Returns/InboundDetailView/Toolbar.tsx
```diff
@@ -19,14 +20,20 @@ import { InboundReturnFragment, useReturns } from '../api';
...
 
-  //   const isDisabled = useOutbound.utils.isDisabled();
+  const update = (data: Partial<InboundReturnFragment>) => {
+    if (!id) return;
+    debouncedMutateAsync({ id, ...data });
+  };
+
+  const { isGrouped, toggleIsGrouped } = useIsGrouped('inboundReturn');
+  const [theirReferenceBuffer, setTheirReferenceBuffer] =
+    useBufferState(theirReference);
```
andreievg, line 36: I am not a fan of buffer state. Btw I get the same issue as the comment (due to useBufferState I think).

What we really need is a useDebunceState (a hook to call update method and to hold the 'value'):
```typescript
const [theirReferenceBuffer, setTheirReferenceBuffer] = useDebounceState(theirReference, (theirReference) => { patch({theirReference})})

```

client/packages/invoices/src/Returns/OutboundDetailView/Footer/StatusChangeButton.tsx
```diff
@@ -124,15 +110,9 @@ const useStatusChangeButton = () => {
...
-      const status = getStatusUpdateInput(selectedOption.value);
-
-      if (status) {
-        await mutateAsync({
-          outboundReturnId: data?.id,
-          status,
-        });
-        success(t('messages.return-saved'))();
-      }
+      await mutateAsync({ id: data?.id, status: selectedOption.value });
```
andreievg, line 113: What does this change do ? Do you mind commenting, i can't see it in the PR description and looks like logic is changed

client/packages/invoices/src/Returns/api/hooks/document/useUpdateInboundReturn.ts
```diff
@@ -1,12 +1,26 @@
...
   const queryClient = useQueryClient();
   const api = useReturnsApi();
-  return useMutation(api.updateInboundReturn, {
+  const mutation = useMutation(api.updateInboundReturn, {
     onSuccess: () => {
       return queryClient.invalidateQueries(api.keys.base());
     },
   });
+
+  const debouncedMutateAsync = useDebounceCallback(
```
andreievg, line 19: I kind of like thise debounceMutate, I think if it also kept track of state, would be pretty useful

server/graphql/invoice/src/mutations/outbound_return/update.rs
```diff
@@ -16,10 +16,14 @@ use service::{
...
 pub struct UpdateInput {
-    pub outbound_return_id: String,
+    pub id: String,
     // supplier_id: String,
     status: Option<UpdateOutboundReturnStatusInput>,
     comment: Option<String>,
+    colour: Option<String>,
+    on_hold: Option<bool>,
+    their_reference: Option<String>,
+    transport_reference: Option<String>,
```
andreievg, line 26: is this being used anywhere ?

andreievg (commented):
Looks pretty good thanks, i have some question before approving, mainly though that bug that i noticed

client/packages/invoices/src/Returns/OutboundDetailView/Footer/StatusChangeButton.tsx
lache-melvin: Oh sorry, forgot about this one 🙈 The mapping of InvoiceNodeStatus -> UpdateOutboundReturnStatusInput, I moved it to a parser method in `api.ts`, to align with how we do it elsewhere 😅

client/packages/invoices/src/Returns/InboundDetailView/Toolbar.tsx
lache-melvin: Oh ew haha, thanks for spotting. I've not used the `fields` hook for returns, but it appears I missed making it NOT do this. Will have a quick play... I like the debouncedState all in one idea 🤔

lache-melvin: Thoughts on this version? https://github.com/msupply-foundation/open-msupply/compare/3316-allow-mutation-of-more-invoice-fields...3316-allow-mutation-of-more-invoice-fields(useEffect-buffer-state-version)

andreievg: Thanks, I think this is good for now, until we think of a pattern to do this. Just my quick 2 cents, I reckon the API abstractions should focus on easy 'consumption' rather then logical structure/partitioning. Like I would love a hook like this (just. a quick brainstorm)

```typescript
// result is a union (rather then checking if it's loading and then if it's error or data etc..
const { patch, result } = useOutboundShip("thisIsTheId");
// do the error/loading checking etc..
const { theirReference, comment } = result.shipment
// .... that field
value={theirReference}
onChange={e => {
   // patch sets buffered state and queues debounce etc..
   patch({theirReference: e.target.value})
}}
```

lache-melvin: I like this, with the data and mutation coming from the same hook - clean and easy to reason about. Would be keen to come back to it

Issue #3320, opened 2024-03-18
Re-synced inbound shipment lines have no stock line once delivered


lache-melvin:
## What went wrong? 😲

 

When a re-synced Inbound Shipment is confirmed as Delivered, new stock is introduced. However, the Inbound Shipment InvoiceLine still has `stockLine: NULL`:

![Screenshot 2024-03-18 at 5 17 53 PM](https://github.com/msupply-foundation/open-msupply/assets/55115239/0bc147b3-6674-4fb3-b56f-ecece54bb0cb)

Can confirm that the stock line is still created! Just the relationship with the invoice line is no longer being created.

I'm not sure what other issues this causes? It's needed for returns, because we return stock based on that stockLine, but we'd have until 2.0.0 to resolve that one...

## Expected behaviour 🤔

invoice_line gets stock_line_id set when the stock is introduced, so when we query the invoice line, we also get the stock line!

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create an inbound shipment, with lines, leave New status (or if using a transferred IS, Shipped)
2. Sync to central
3. Blow away omsupply db
4. Re initialise
5. (Can also skip above steps and just use an old IS with status <=Shipped, that you know was initialised from central)
6. Set the status to Delivered
7. Use dev tools to check the stockLine field of the invoiceLines... it's null!
8. Can check `View Stock` - the stock has been created

If you create a new IS, add lines, and set to Delivered, stock line is associated fine. Bug is only when using a re-initialised shipment.

## Your environment 🌱

- Version: 1.8
- mSupply version: 7.11.14

- Platform: Browser (chrome)

- Database type: SQLite3

lache-melvin:
Uh oh - okay it's not always... can anyone replicate?

roxy-dao:
Hmm, doesn't seem to be happening for me. Tried both manual inbound shipment and transferred 🤔

lache-melvin:
Now it's the only thing that's happening for me 🥲

lache-melvin:
Have updated the description - have determined this only affects re-initialised Inbound Shipments.

Chatted with Andrei, it's possible that this issue will be solved by #3484, will test once merged back into develop. 

Thinking is that in the syncing with central/re-initialising, the tax/currency state is different, which gets updated when we change statuses. That tax/currency change currently calls `upsert_one` for each line, which overwrites any other changes to the shipment lines - hence the assignment the stock line id gets lost. 

[...]

PR #3371, opened 2024-03-21
3319 show actual error if no stock lines are associated with shipment lines to return

### LINKED ISSUE #3319: Unclear error when can't return lines
## What went wrong? 😲
 

Getting a please select lines to return error when i HAVE selected lines to return
Image

## Expected behaviour 🤔

The modal opens, or a more helpful error with why you can't open the modal...

## How to Reproduce 🔨

I have a hunch this is related to depending on the optional stock line field in the selected stock lines code... I thought it was always defined once the inbound shipment was marked as delivered, but apparently not? whoopsie 🙈 Either way, the error should actually reflect the problem!

1. Go to a new or shipped inbound shipment
2. Mark as delivered
3. Select lines to return
4. See the warning (and if you inspect the shipment, stock lines haven't been assigned... maybe this is actually the bug?)

## Your environment 🌱

- Version: 
- mSupply version: 

- Platform:

- Database type:

lache-melvin:
Fixes #3319

# 👩🏻‍💻 What does this PR do? 
 

This issue came about from another intermittent issue I appear to be having (#3320) where sometimes the inbound shipment lines don't have an associated stock line, which we need to generate the return.

Anyways, we were expecting the stockLine to always be present, and when it wasn't, we'd get a `please select lines to return` error even when we had selected lines. This PR shows an actual error message :)

Screenshot 2024-03-21 at 5 48 49 PM

Somehow set up a delivered inbound return with an invoice line with no stock_line_id (maybe just set to NULL in DB?)

- select that line to return
- see error popup

In hindsight, I'm wondering if we should instead be sending the inbound shipment line ids to the generate method, and letting that query error if there is no associated stock line? But probably not worth the rework at this point...


client/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -55,17 +55,30 @@ export const DetailView: FC = () => {
...
       const selectLinesSnack = info(t('messages.select-rows-to-return'));
       selectLinesSnack();
-    } else {
-      onOpenReturns(selectedStockLineIds);
-      setReturnMode(ModalMode.Create);
+      return;
+    }
+    if (selectedLines.some(line => !line.stockLine)) {
+      console.error('No stock line associated with the selected line(s)');
+      const selectLinesSnack = error(t('error.something-wrong'));
```
lache-melvin, line 71: Keen for thoughts here... this isn't an error the user could do anything about, and it kind of confusing to explain... so just went with generic error, and slightly more detail to the console. Should I provide more/less anywhere?

andreievg: I think this would be an edge case with inbound shipments that have been created in og store.

My preference is to show to the user, we have no access to console in production, in that case I was wondering what value would console.error bring ? 

I think it's ok to show untranslated error concatenated with error.something-wrong

andreievg (approved):
Looks good thanks, approving but I think there should be a bit more info presented to the user.

> In hindsight, I'm wondering if we should instead be sending the inbound shipment line ids to the generate method, and letting that query error if there is no associated stock line? But probably not worth the rework at this point...

Agree, with both, nicely noted thanks

client/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
lache-melvin: Ah ok perfect, will do :)

PR #3375, opened 2024-03-21
3338 fix error popup when creating inbound return from outbound shipment

### LINKED ISSUE #3338: Could not get InboundReturn error when creating from OutboundShipment
## What went wrong? 😲
 

When selecting some outbound shipment lines to return, and opening the return modal, we get this error:
Image

I think something in the modal logic must be wrong, and it thinks we're in edit mode? So it's querying for the existing return data, based on the invoice ID in the URL, and then not finding one!

I'm not sure where this is happening though 👀 

Note: this isn't happening for outbound return from inbound shipment.

## Expected behaviour 🤔

Well, firstly, no error popup. But more importantly, we shouldn't be querying for a return from the OutboundShipment page!

## How to Reproduce 🔨

1. Go to a Shipped (or greater) Outbound Shipment
2. Select some lines and click return selected lines
3. See error

lache-melvin:
Fixes #3338

# 👩🏻‍💻 What does this PR do? 
 

Before changes: creating inbound return from outbound shipment showed a `Could not get inbound return` error:
image

Now we don't see an error 🎉 
Screenshot 2024-03-22 at 9 05 15 AM

We were getting this error because the AddBatch button checks whether or not the inbound return is disabled. The add batch button isn't even visible in this view, but the hook was still being called (on a return that doesn't exist), hence the error!

Refactored the `AddBatch` button to it's own component, so that the hook will only be called when the button is visible, and the return exists.

- Go to an outbound shipment that is shipped or later
- Select some lines to return
- Modal opens, you should see no error

### DIFF
--- a/client/packages/invoices/src/Returns/OutboundDetailView/AppBarButtons.tsx
+++ b/client/packages/invoices/src/Returns/OutboundDetailView/AppBarButtons.tsx
@@ -27,7 +27,7 @@ interface AppBarButtonProps {
 export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   onAddItem,
 }) => {
-  //   const isDisabled = useOutbound.utils.isDisabled();
+  const isDisabled = useReturns.utils.outboundIsDisabled();
   const { data } = useReturns.document.outboundReturn();
   const { OpenButton } = useDetailPanel();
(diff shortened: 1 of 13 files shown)


andreievg (approved):
Thanks, looks good.

PR #3379, opened 2024-03-21
3317 implement isDisabled hook for outbound returns

### LINKED ISSUE #3317: Confirm Delivered button enabled for outbound returns
## What went wrong? 😲
 

For an outbound return, you shouldn't be able to set delivered or verified. This button doesn't work (you get an invalid status error from the backend) but it should be disabled in the first place!
Image

## Expected behaviour 🤔

Button is disabled.

## How to Reproduce 🔨

1. Go to outbound return detail view
2. Update statuses until shipped
3. See enabled `Confirm Delivered` button

lache-melvin:
Fixes #3317

# 👩🏻‍💻 What does this PR do? 
 

Thought there was a bug - the confirm delivered button was enabled when it shouldn't have been. Turns out we just hadn't implemented isDisabled yet 😁 

Implements isDisabled for outbound returns, the button is removed once the return is shipped:
Screenshot 2024-03-22 at 10 02 28 AM

- Status change button is removed once the outbound return is shipped
- Other buttons are also disabled once shipped (onHold, addItem)


client/packages/invoices/src/Returns/OutboundDetailView/SidePanel/AdditionalInfoSection.tsx
```diff
@@ -16,7 +16,8 @@ import { useReturns } from '../../api';
 export const AdditionalInfoSectionComponent: FC = () => {
   const t = useTranslation('replenishment');
   const { data } = useReturns.document.outboundReturn();
-  //   const isDisabled = useOutbound.utils.isDisabled();
+  // const isDisabled = useReturns.utils.outboundIsDisabled();
```
lache-melvin, line 19: to be uncommented when other code is merged 😅

CarlosNZ (commented):
Just quickly (haven't reviewed properly yet), I can still click on a line in Outbound return and open the modal --- this is not allowed in a standard outbound shipment. Was it the intention that we can still open the modal? (If so, we should probably disabled the inputs on the modal -- they currently let you change values, and give error when clicking "OK"). I think it would be better if we just couldn't open it though, you reckon?

lache-melvin:
> Just quickly (haven't reviewed properly yet), I can still click on a line in Outbound return and open the modal --- this is not allowed in a standard outbound shipment. Was it the intention that we can still open the modal? (If so, we should probably disabled the inputs on the modal -- they currently let you change values, and give error when clicking "OK"). I think it would be better if we just couldn't open it though, you reckon?

Ah great point - we were going to maintain the ability to open the modal, because at this point, that's the only place that return reasons can be viewed. However you're right, we could disable the inputs.

lache-melvin:
@CarlosNZ what do you think of this? Before I implement for inbound as well 😁 https://github.com/msupply-foundation/open-msupply/compare/3317-outbound-return-is-disabled...3317-outbound-return-is-disabled(disable-modal-inputs-when-return-disabled)

CarlosNZ:
> @CarlosNZ what do you think of this? Before I implement for inbound as well 😁 [3317-outbound-return-is-disabled...3317-outbound-return-is-disabled(disable-modal-inputs-when-return-disabled)](https://github.com/msupply-foundation/open-msupply/compare/3317-outbound-return-is-disabled...3317-outbound-return-is-disabled(disable-modal-inputs-when-return-disabled))

Yeah, looks good.

CarlosNZ (approved):
Looks good to me :)

I notice, however, that when you click "Ok" in the line edit table, you still get a "can't edit" error. Probably best if this button just skips the `onSave()` attempt when disabled (i.e it just closes the modal), but I can add that into my PR once this is merged -- it might get messy if you try and change it here as it affects the buttons which I've modified.

lache-melvin:
> Looks good to me :)
> 
> I notice, however, that when you click "Ok" in the line edit table, you still get a "can't edit" error. Probably best if this button just skips the `onSave()` attempt when disabled (i.e it just closes the modal), but I can add that into my PR once this is merged -- it might get messy if you try and change it here as it affects the buttons which I've modified.

Oh yeah great point! Ok great, thank you!

PR #3385, opened 2024-03-22
store selection

### LINKED ISSUE #3303: UI for changing the store of items
Part of epic #2896 

We need to be able to handle how to centrally manage the assignment of store - equipment

- [ ] In central server mode only, add a column to the equipment list which shows the allocated store
- [ ] In central server mode only, add a store selector to the detail view

Do we need to do something within sync to handle store reassignment? 

If an asset is on Store A, but Central Admin changes it to Store B? It should disappear from Store A and only appear in Store B (Assuming they’re on different sync sites)

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #3303

# 👩🏻‍💻 What does this PR do? 
 
If you are on the central server, there is now a column added to the equipment list view, and an input on the summary tab of the detail view:

image

Screenshot 2024-03-22 at 12 28 16 PM

There's also the option to set a store when creating a new asset:

Screenshot 2024-03-22 at 3 14 23 PM

You can run the following to start the central server.

```IS_CENTRAL_SERVER=true APP_SERVER__PORT=2050 APP_DATABASE__DATABASE_NAME=omsupply_central cargo run --features postgres```

I then ran `yarn webpack-cli serve --env API_HOST=http://localhost:2050` from the `client/packages/host` directory in order to target the central server.

Doing so will give you the extra column, and the ability to edit the store for an asset.
You should check that running a normal remote site does not show the store option, but instead

## 📃 Documentation

Part of the assets change

### DIFF
--- a/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
@@ -71,8 +71,7 @@ const Row = ({
   children: React.ReactNode;
   label: string;
 }) => (
-  // the .MuiFormControl-root ensures that the autocomplete input is full width
-  <Box paddingTop={1.5} sx={{ '& .MuiFormControl-root': { width: '100%' } }}>
+  <Box paddingTop={1.5}>
     <InputWithLabelRow
       labelWidth="150px"
       label={label}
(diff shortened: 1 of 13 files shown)


adamdewey:
Perhaps we should move the store column to the 1st position? 

If I was a central person I think I'd want to see that first so that I know where the equipment I'm looking at is located?

jmbrunskill (approved):
Looks good to me, we can look at fixing the asset_log sync problem when we merge asset log sync I think.

PR #3411, opened 2024-03-25
#3912 Return modal navigation updates

### LINKED ISSUE #3192: Add ability to go back to previous step in a modal (5h)

There's no way to go back to the previous step here 😢 (but I entered some batch info wrong and I really wanna go back)
Image

The Modal component offers `cancel`, `ok` and `next` button options - perhaps we could add a `back` button for cases like this.

Could also add some kind of linking feature to the wizard steps, so you could click them to go back to previous steps?

CarlosNZ:
Fixes #3192

Please note, I've typed the issue number incorrectly in the branch name.

# 👩🏻‍💻 What does this PR do? 
 

Changes the display and functionality of the buttons in the Returns modals as per [this description](https://github.com/msupply-foundation/open-msupply/issues/3192#issuecomment-2011135821).

When creating a new return from shipment, should show like this at the different steps:

Screenshot 2024-03-25 at 2 03 03 PM
Screenshot 2024-03-25 at 2 03 19 PM

And for edit a single line (but only when there is a "next" item available):
Screenshot 2024-03-25 at 2 03 32 PM

Screenshot 2024-03-25 at 2 03 38 PM

Also, check that the behaviour/display of the buttons is correct when trying to advance with 0 quantities (for the warnings, etc).

Trickiest bit was adding the "Okay and next" functionality, as we didn't currently have a way for the modal to know whether there were additional items later in the list. Went for the simplest way of implementing I could, hopefully it's adequate for our purposes.

The Inbound is pretty much an exact copy/paste of the Outbound, FYI.


client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
```diff
@@ -45,6 +48,12 @@ const getButtonProps = (
         labelKey: 'button.cancel',
         variant: 'outlined',
       };
+    case 'back':
+      return {
+        icon: <ArrowLeftIcon />,
+        labelKey: 'button.back',
+        variant: 'outlined',
+      };
```
CarlosNZ, line 56: Figured that "Back" was a generic enough button to be included here in the common set. Kept "Next Step" as a "custom" button, but could add here if we thought it warranted it?

client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
```diff
@@ -29,6 +31,7 @@ interface DialogButtonProps {
   autoFocus?: boolean;
   color?: 'primary';
   type?: 'button' | 'submit' | 'reset';
+  customLabel?: string;
```
CarlosNZ, line 34: "Custom label" allows us to keep using this component when we want some buttons to say something very case-specific. Makes it easier to keep consistent when using a Custom button label alongside other "common" buttons (Ok, cancel, etc.)

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
```diff
@@ -74,6 +88,7 @@ export const InboundReturnDetailView: FC = () => {
               returnId={data.id}
               initialItemId={itemId}
               modalMode={mode}
+              loadNextItem={nextItemId ? () => onOpen(nextItemId) : undefined}
```
CarlosNZ, line 91: Made this function's existence conditional, as otherwise we'd need another prop just to tell the modal whether or not there even *is* a next item.

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -57,6 +59,10 @@ export const InboundReturnEditModal = ({
     itemId,
   });
 
+  useEffect(() => {
+    if (initialItemId) setItemId(initialItemId);
+  }, [initialItemId]);
+
```
CarlosNZ, line 65: I could avoid adding the `useEffect` if I also called `setItem` as part of `handleNextItem`, but this seems more 
"correct" to me, as one state should cause the other to update rather than updating it in two places.

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -84,22 +100,43 @@ export const InboundReturnEditModal = ({
...
+    />
+  );
+  const OkButton = <DialogButton onClick={onOk} variant="ok" />;
+  const OkAndNextButton = (
+    <DialogButton
+      onClick={handleNextItem}
+      variant="next"
+      disabled={currentTab !== Tabs.Reason}
+    />
+  );
```
CarlosNZ, line 122: Thought it was a bit easier to read to define all the possible buttons here, and render them conditionally below, rather than mixing the component definitions in with the conditional logic.

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -84,22 +100,43 @@ export const InboundReturnEditModal = ({
...
-            <DialogButton onClick={onOk} variant="ok" />
-          ) : undefined
+          currentTab === Tabs.Quantity && zeroQuantityAlert !== 'warning'
+            ? NextStepButton
+            : OkButton
+        }
+        nextButton={
+          modalMode === ModalMode.Update && loadNextItem
+            ? OkAndNextButton
+            : undefined
```
CarlosNZ, line 139: Changed the order here so match the order they get displayed by the `useDialog` modal: "Cancel" -> "Ok" -> "Next"

client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
lache-melvin: I think `Next Step` isn't really use case specific - it's pretty generic to any kind of stepper modal. However it seems like the talk is leaning away from the modal stepper situation, so happy to leave as custom for now, and can refactor later once the design decisions are clearer.

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
```diff
@@ -51,6 +51,20 @@ export const InboundReturnDetailView: FC = () => {
     },
   ];
 
+  const getNextItemId = () => {
```
lache-melvin, line 54: Is this logic that we have used in multiple places? Feels like the kind of thing that would be nice to have some unit tests around... and potentially easy to pull out/reuse?

client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
lache-melvin: Love this!

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
lache-melvin, line 139: I think leave as is for now, and see what feedback we get. Just noting:

Intuitively, I found myself wanting the Next Step button to become the OK and Next button. So I can keep my mouse in the same place to keep clicking through.

Though Carl made the good point - Next Step and Ok & Next (both having the word "next") are quite similar and generic, so anything we can do to distinguish them is possibly quite helpful.

lache-melvin (commented):
Discussed in person - while we're in here working on this button logic, would be good to implement `Ok & Next` for the `Add Item` use case... in Inbound/Outbound Shipment, the `Add Item` modal has an `Ok & Next` button, which clears the modal so you can add another item, without needing to save and then open the modal again. Would be good to be consistent on this front 🚀

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
CarlosNZ: The other implementations have a `useNextItem` hook. I had a look and the logic is basically the same as what I've done here. However, I don't like calling it a hook, since it isn't one. Also, they'd made a whole new hook for every implementation, which I also don't like.

So what I've done now is just exported `getNextItem` as a separate utility function (in OuboundReturnsDetailView) and re-used it in Inbounds. e49f779cc

We should probably have a standardised pattern for these kind of "utility" functions as to where they go, etc., but they're not particularly widespread as far as I can see (there's just millions of hooks!!).

CarlosNZ:
> Discussed in person - while we're in here working on this button logic, would be good to implement `Ok & Next` for the `Add Item` use case... in Inbound/Outbound Shipment, the `Add Item` modal has an `Ok & Next` button, which clears the modal so you can add another item, without needing to save and then open the modal again. Would be good to be consistent on this front 🚀

Have done this now, and also included the `isDisabled` check in the conditional logic. Had to make a couple of new props: `hasNextItem` and `isNewReturn`, as I could no longer use the implied conditions (from other values) I was using previously.

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
```diff
@@ -88,10 +77,18 @@ export const InboundReturnDetailView: FC = () => {
...
               modalMode={mode}
-              loadNextItem={nextItemId ? () => onOpen(nextItemId) : undefined}
+              loadNextItem={() => {
+                if (nextItemId) onOpen(nextItemId);
+                else {
+                  // Closing and re-opening forces the modal to launch with the
+                  // item selector in focus
+                  onClose();
+                  setTimeout(() => onOpen(), 50);
+                }
```
CarlosNZ, line 87: I don't love this, but I couldn't find any other way to get the new item modal to launch with the item selector opened/in focus (without adding a bunch of complexity to the ItemSelector component). I don't fully understand how they've done it in the Shipments, but this approach actually gives a less jarring visual jump than the Shipment one anyway, so 🤷‍♂️

client/packages/invoices/src/Returns/OutboundDetailView/DetailView.tsx
```diff
@@ -114,3 +111,19 @@ export const OutboundReturnsDetailView: FC = () => {
...
+  const currentItemIndex = lines.findIndex(
+    line => line.itemId === currentItemId
+  );
+  if (currentItemIndex === -1) return;
+
+  const nextItemIndex = lines.findIndex(
+    (line, index) => index > currentItemIndex && line.itemId !== currentItemId
+  );
+  return nextItemIndex === -1 ? undefined : lines[nextItemIndex]?.itemId;
+};
```
CarlosNZ, line 129: Exported and re-used in Inbounds

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -60,12 +69,13 @@ export const InboundReturnEditModal = ({
...
   useEffect(() => {
-    if (initialItemId) setItemId(initialItemId);
+    if (initialItemId === undefined) return;
+    setItemId(initialItemId === null ? undefined : initialItemId);
   }, [initialItemId]);
 
   const onOk = async () => {
     try {
-      await save();
+      !isDisabled && (await save());
```
CarlosNZ, line 78: As mentioned in your PR, this will just close the modal without attempting to save when disabled.

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -117,7 +127,7 @@ export const InboundReturnEditModal = ({
     <DialogButton
       onClick={handleNextItem}
       variant="next"
-      disabled={currentTab !== Tabs.Reason}
+      disabled={currentTab !== Tabs.Reason || (isDisabled && !hasNextItem)}
```
CarlosNZ, line 130: The "Okay and Next" button should be disabled when `isDisabled` and we've got to the end of the list, as we don't want to be able to make a new line. (this is why I needed to explicitly pass `hasNextItem` in)

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -133,11 +143,7 @@ export const InboundReturnEditModal = ({
             ? NextStepButton
             : OkButton
         }
-        nextButton={
-          modalMode === ModalMode.Update && loadNextItem
-            ? OkAndNextButton
-            : undefined
-        }
+        nextButton={!isNewReturn ? OkAndNextButton : undefined}
```
CarlosNZ, line 146: And this is why we need an explicit `isNewReturn` prop, as the ModalMode was no longer a correct indicator of whether or not to show the Ok/Next button

CarlosNZ (commented):
Okay @lache-melvin, I've implemented the discussed suggestions and accounted for the `isDisabled` state in the buttons. See inline comments for specific details.

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
lache-melvin: Interesting... the other modals use a method that returns `true` which is captured by `useDialog` to do the little slide animation, but I guess you also get that here by forcing another `onOpen`. This does feel a bit hacky, but also less magic, which I like. I'd be curious to see what other people in the team think!

client/packages/invoices/src/Returns/OutboundDetailView/DetailView.tsx
lache-melvin: Could we put this in packages/invoices/src/utils.ts?

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
lache-melvin: For consistency... the other modals disable Ok & Next when you get to the end of the list, regardless of whether the invoice is disabled. If you want to add new lines, you need to click the add but explicitly. Could we make this the same?

lache-melvin (approved):
Looking good Carl! Just the one note for me about not having Ok & Next open the blank modal after cycling through existing lines. I think that would be good to keep consistent with other parts of the app.

Feel free to merge when ready!

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
CarlosNZ: Oh, I didn't realise that. I guess that makes sense, maybe, I can see it both ways. But sure, will keep it consistent.

client/packages/invoices/src/Returns/OutboundDetailView/DetailView.tsx
CarlosNZ: Yup, good call: 5c125e06c

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
CarlosNZ: Done in 5c125e06c

client/packages/invoices/src/Returns/InboundDetailView/DetailView.tsx
CarlosNZ: Can you point to where this is done -- I'm not seeing it, sorry.

andreievg: https://github.com/msupply-foundation/open-msupply/blob/5c125e06c4036edc3ad04f5a958e01a6fed907a7/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx#L153-L165

https://github.com/msupply-foundation/open-msupply/blob/5c125e06c4036edc3ad04f5a958e01a6fed907a7/client/packages/common/src/hooks/useDialog/useDialog.tsx#L187

andreievg:
Hmm for inbound return I am not getting the next item change (the item name at the top changes but not the table content) ¯\_(ツ)_/¯

andreievg:
Existing logic uses sorted items:

https://github.com/msupply-foundation/open-msupply/blob/5c125e06c4036edc3ad04f5a958e01a6fed907a7/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx#L123-L125

https://github.com/msupply-foundation/open-msupply/blob/5c125e06c4036edc3ad04f5a958e01a6fed907a7/client/packages/invoices/src/InboundShipment/api/hooks/document/useNextItem.ts#L11

https://github.com/msupply-foundation/open-msupply/blob/5c125e06c4036edc3ad04f5a958e01a6fed907a7/client/packages/invoices/src/InboundShipment/api/hooks/line/useInboundRows.ts#L40

I think it should be ok to copy and paster if possible

client/packages/invoices/src/Returns/modals/InboundReturn/InboundReturn.tsx
```diff
@@ -84,23 +110,44 @@ export const InboundReturnEditModal = ({
...
+      }
+    />
+  );
+
   return (
     <TableProvider createStore={createTableStore}>
       <Modal
         title={t('heading.return-items')}
-        cancelButton={<DialogButton onClick={onClose} variant="cancel" />}
+        cancelButton={currentTab === Tabs.Quantity ? CancelButton : BackButton}
```
andreievg, line 142: I like this, vs conditionaly changing onClick and variant

CarlosNZ:
> Hmm for inbound return I am not getting the next item change (the item name at the top changes but not the table content) ¯_(ツ)_/¯

Ah, okay, the `useDraftInboundLines` hook wasn't updating the lines when Item changed. Have modified now. 617a83878

CarlosNZ:
> Existing logic uses sorted items:

I think I'll leave that to a follow-up issue. Currently the modal launcher is outside the table component, so getting access to the sorted rows will require a bit more work to figure out and is beyond the simple button behaviour issue we're doing here.

New issue here: #3457

PR #3416, opened 2024-03-25
3343 sync return reasons from central

### LINKED ISSUE #3343: Store return reason options centrally (6h)

We store return_reasons in omSupply DB, but not on the central server. These should be options configurable on the central server, like inventory adjustment reasons.

Add a new option type to mSupply, for return reasons. Sync this to omSupply return reasons table.

lache-melvin:
Fixes #3343

# 👩🏻‍💻 What does this PR do? 
 

Adds changes from peering session last week, adding return reason options to mSupply and syncing them.

ON MSUPPLY: You'll need this branch checked out: https://github.com/msupply-foundation/msupply/pull/14517, and add some return reasons under Preferences > Options.

Run sync - log out and in again (return reasons are cached/considered stale for 5 minutes otherwise) and voila, you have some return reasons! 🎉


andreievg (commented):
I think it's worth renaming the translation file and test data file and adding a test variant here: https://github.com/msupply-foundation/open-msupply/blob/e3c5c764454c700551b4eaa01f20b27ca180b999/server/service/src/sync/test/test_data/inventory_adjustment_reason.rs#L3

lache-melvin:
> Thanks @lache-melvin I think it's worth renaming the translation file and test data file and adding a test variant here:
> 
> https://github.com/msupply-foundation/open-msupply/blob/e3c5c764454c700551b4eaa01f20b27ca180b999/server/service/src/sync/test/test_data/inventory_adjustment_reason.rs#L3

Oh great catches thanks Andrei, done 🙏

andreievg (approved):
Thank you @lache-melvin looks great

PR #3423, opened 2024-03-25
Support multiple SQL report queries

### LINKED ISSUE #3420: Support multiple SQL queries
Follow-up from: https://github.com/msupply-foundation/open-msupply/issues/2853

While implementing the first report it became apparent that we need to be able to support doing multiple SQL queries. For example, a report might need general store information such as logo and store name + the actual report data such as a list of invoices.

Doing everything in one query is quite cumbersome. For example, you could duplicate things like logo in an extra column along every invoice row but this is not only quite a bit of extra json data but also a problem if there are now invoices in the query. Another option is to use `UNION ALL` but this needs extra "padding" columns which needs to be maintained manually.

Just doing two queries is just much simpler.

clemens-msupply:
Fixes #3420 

# 👩🏻‍💻 What does this PR do? 
It's now possible to have 0 or 1 GraphGL queries + multiple sql queries. (See issue for the motivation)

Old reports should still work. I also updated the `examples/SQLReport` report in the `open-msupply-reports` repo to give a full example.

### DIFF
--- a/server/report_builder/src/build.rs
+++ b/server/report_builder/src/build.rs
@@ -47,40 +47,50 @@ fn parse_default_query(input: &str) -> anyhow::Result<DefaultQuery> {
 fn extract_sql_entry(
     args: &BuildArgs,
     files: &mut HashMap<String, PathBuf>,
-) -> Result<Option<(String, SQLQuery)>> {
-    match (&args.query_sqlite, &args.query_postgres) {
-        (None, None) => Ok(None),
-        (Some(query_sqlite), Some(query_postgres)) => {
+) -> Result<Vec<SQLQuery>> {
+    let Some(sql_queries) = &args.query_sql else {
+        return Ok(vec![]);
+    };
+    let result: Result<Vec<_>> = sql_queries
+        .iter()
+        .map(|query| {
+            let common_query = format!("{query}.sql");
+            if let Some(file_path) = files.remove(&common_query) {
+                let query_sql = fs::read_to_string(file_path).map_err(|err| {
+                    anyhow::Error::msg(format!("Failed to load query file: {}", err))
+                })?;
+                return Ok(SQLQuery {
+                    name: query.clone(),
+                    query_sqlite: query_sql.clone(),
+                    query_postgres: query_sql.clone(),
+                });
+            }
+            let query_sqlite = format!("{query}.sqlite.sql");
             let file_path = files
-                .remove(query_sqlite)
-                .ok_or(anyhow::Error::msg("Sqlite query file does not exist"))?;
+                .remove(&query_sqlite)
+                .ok_or(anyhow::Error::msg(format!(
+                    "Sqlite query file ({query_sqlite}) does not exist"
+                )))?;
             let query_sqlite_sql = fs::read_to_string(file_path).map_err(|err| {
                 anyhow::Error::msg(format!("Failed to load Sqlite query file: {}", err))
             })?;
-            let query_postgres_sql = if query_postgres == query_sqlite {
-                query_sqlite_sql.clone()
-            } else {
-                let file_path = files
-                    .remove(query_postgres)
-                    .ok_or(anyhow::Error::msg("Postgres query file does not exist"))?;
-                fs::read_to_string(file_path).map_err(|err| {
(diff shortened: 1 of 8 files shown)


server/report_builder/src/lib.rs
```diff
@@ -39,16 +39,13 @@ pub struct BuildArgs {
...
-    /// However, query_sqlite and query_postgres can point to the same file.
-    #[clap(long)]
-    pub query_postgres: Option<String>,
+    /// SQL query name.
+    /// This argument requires that there is a either
+    /// - a single {query_sql}.sql file
+    /// - a {query_sql}.sqlite.sql file and a {query_sql}.postgres.sql file
+    /// The query result is put in the data object under `data.{query_sql}`.
+    #[clap(long, value_parser, value_delimiter = ' ')]
+    pub query_sql: Option<Vec<String>>,
```
clemens-msupply, line 48: IMHO this got simpler now by requiring how query files are named

server/report_builder/src/build.rs
```diff
@@ -47,48 +47,58 @@ fn parse_default_query(input: &str) -> anyhow::Result<DefaultQuery> {
...
-        (None, None) => Ok(None),
-        (Some(query_sqlite), Some(query_postgres)) => {
+) -> Result<Vec<SQLQuery>> {
+    let Some(sql_queries) = &args.query_sql else {
+        return Ok(vec![]);
+    };
+    let result: Result<Vec<_>> = sql_queries
+        .iter()
+        .map(|query| {
+            let common_query = format!("{query}.sql");
```
jmbrunskill, line 57: If we replace a common query with specific ones for postgres and sqlite is there a risk the common query might be left behind on disk and cause a bug?

jmbrunskill: I guess it's all in a single json record sent from central?

server/report_builder/src/lib.rs
```diff
@@ -39,16 +39,16 @@ pub struct BuildArgs {
...
-    #[clap(long)]
-    pub query_sqlite: Option<String>,
-    /// File name of the Postgres query.
-    /// If specified the query_sqlite must be provided as well.
-    /// However, query_sqlite and query_postgres can point to the same file.
-    #[clap(long)]
-    pub query_postgres: Option<String>,
+    /// SQL query name.
+    /// This argument requires that there is either
+    /// - a single {query_sql}.sql file (for both Sqlite and Postgres)
```
jmbrunskill, line 44:
```suggestion
    /// - a single {query_name}.sql file (for both Sqlite and Postgres)
```
I think that's a bit more clearer naming (IMHO) would want to change the other comments to match of course.

server/report_builder/src/lib.rs
jmbrunskill: I think this makes sense, similar pattern to notify

server/service/src/report/string_or_vec.rs
```diff
@@ -0,0 +1,34 @@
+use std::fmt;
+
+use serde::de::{self, value, Deserialize, Deserializer, SeqAccess, Visitor};
+
+pub(crate) fn string_or_vec<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
```
jmbrunskill, line 5: Might be nice to have a test for this,  looks pretty safe.

jmbrunskill (approved):
Looks good to me.

server/report_builder/src/build.rs
clemens-msupply: Yes.

If there is a common query and a sqlite/postgres file, I could add a warning/error? For example, if the user provided a myquery.sqlite.sql and a myquery.sql file the sqlite file get silently ignored right now...

server/report_builder/src/lib.rs
clemens-msupply: I used the name of the parameter, i.e. `query_sql` I put the sql in there to differenciate between the graphql queries (`query_gql` paramter). So think {query_sql} is more consitent.

server/service/src/report/string_or_vec.rs
clemens-msupply: Agree, I tested manually but should add one if there will be another change to it!

PR #3429, opened 2024-03-25
3139 import asset from spreadsheet

### LINKED ISSUE #3139: Import assets from spreadsheet
Part of expic #2896

Following on from the initial list view #2973, implement the **Import** button which will

* open a dialog where you can drag n drop a file, or
* open a file selector
* from here, it's guesswork : the requirements say only 'Import the new/updated asset register / CCE Inventory'
* there is a 'review' and 'import' step, so perhaps display the imported list in the second step?
* and the results success / fail (?) on the last step?

image

fergie-nz:
Fixes #3139 

# 👩🏻‍💻 What does this PR do? 
Adds importing cold chain assets from .csv file
Makes asset number an optional field

Adds some validations for importing assets
No front end tests added

You can test this by navigating to coldchain -> Equipment and clicking ImportAssets
An example downloadable csv file template with which to import is available.


package.json
```diff
@@ -7,5 +7,9 @@
     "start": "cd ./server && cargo run & cd ./client && yarn start-local",
     "build": "cd ./client && yarn build && cd ../server && cargo build --release",
     "test": "cd ./server && ulimit -n 1000 && cargo test && cd ../client && yarn test"
+  },
+  "dependencies": {
```
mark-prins, line 11: these shouldn't be installed at this level - they're client only packages and need to be installed in the client app only. Ideally, just in the package they are required in, which is `common` for the `react-dropzone`; and `papaparse` is already installed, and does not need to come in again.

I'm going to remove directly in the branch, as the inclusion has blown out the PR files. Hope that's ok

client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+  );
+
+  useEffect(() => {
+    fetchAsync();
+  }, [fetchAsync]);
+
+  const csvExport = async () => {
+    const csv = importEquipmentToCsv(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      bufferedEquipment.map((row: ImportRow): any => {
```
fergie-nz, line 115: I've disabled linting in a few places which I wasn't sure about including this one. 
Not sure if the data parsing can manage better typing?

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const csv = importEquipmentToCsv(
+      emptyRows.map((row: ImportRow): InsertAssetInput => {
+        return toInsertEquipmentInput(row, catalogueItemData);
+      }),
+      t
+    );
+    FileUtils.exportCSV(csv, t('filename.cce'));
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
fergie-nz, line 50: Similar disabling lint for unknown 'any' type

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+      if (
+        row[t('label.catalogue-item-code')] === undefined ||
+        row[t('label.catalogue-item-code')].trim() === ''
+      ) {
+        rowErrors.push(
+          t('error.field-must-be-specified', {
+            field: t('label.catalogue-item-code'),
+          })
+        );
+      } else if (
```
fergie-nz, line 118: Additional check that catalogue item code matches the code of the upload sheet.

Maybe this is a helpful data validation or maybe not? It is possible to create assets without a catalogue item ID, so can remove this. However maybe it is good to know if a user thinks they have a code but the code doesn't match a catalogue item?

Current behaviour is that the upload cannot go ahead if a single code doesn't match a catalogue item

client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+      label: t('label.import'),
+      description: '',
+      clickable: false,
+    },
+  ];
+
+  return (
+    <Modal
+      okButton={
+        <DialogButton
```
mark-prins, line 239: can you change this to be `NextButton` when on steps 1 & 2 please.
It would be helpful to have
- next button visible, ok button hidden on step 1&2
- next button hidden, ok visible on step 3

client/packages/coldchain/src/Equipment/ImportAsset/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+          setPagination({
+            first: pagination.first,
+            offset: pagination.first * page,
+            page: page,
+          });
+        }}
+        columns={columns}
+        data={currentEquipmentPage}
+        noDataElement={<NothingHere body={t('error.asset-not-found')} />}
+        id={''}
```
mark-prins, line 105: is the id required?

client/packages/coldchain/src/Equipment/ImportAsset/ImportTab.tsx
```diff
@@ -0,0 +1,41 @@
...
+        direction="column"
+        justifyContent="center"
+        style={{ minHeight: '50vh' }}
+      >
+        <Typography>
+          {importErrorCount ?? 0 > 0
+            ? '(' +
+              t('messages.error-generic', { count: importErrorCount }) +
+              ')'
+            : ''}
```
mark-prins, line 31: :nit the + and multiline formatting is unusual. would you mind this?
```suggestion
          {importErrorCount ?? 0 > 0
            ? `(${t('messages.error-generic', { count: importErrorCount })})`
            : ''}
```

client/packages/coldchain/src/Equipment/ImportAsset/ReviewTab.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const EquipmentReviewTab: FC<ImportPanel & EquipmentReviewTabProps> = ({
+  tab,
+  uploadedRows,
+}) => {
+  return (
+    <ImportPanel tab={tab}>
+      <ImportReviewDataTable importRows={uploadedRows} />
+    </ImportPanel>
+  );
+};
```
mark-prins, line 19: :nit this can be simpler:

```suggestion
}) => (
  <ImportPanel tab={tab}>
    <ImportReviewDataTable importRows={uploadedRows} />
  </ImportPanel>
);
```

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+  Link,
+  FnUtils,
+  FileUtils,
+  InsertAssetInput,
+} from '@openmsupply-client/common';
+import * as EquipmentImportModal from './EquipmentImportModal';
+import { ImportRow, toInsertEquipmentInput } from './EquipmentImportModal';
+import { importEquipmentToCsv } from '../utils';
+import { AssetCatalogueItemFragment } from '@openmsupply-client/system';
+interface EquipmentUploadTabProps {
```
mark-prins, line 20: :nit can I have a newline after the imports please?

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const { error } = useNotification();
+  const [isLoading, setIsLoading] = useState(false);
+  const EquipmentBuffer: EquipmentImportModal.ImportRow[] = [];
+
+  const csvExample = async () => {
+    const emptyRows: ImportRow[] = [];
+    const csv = importEquipmentToCsv(
+      emptyRows.map((row: ImportRow): InsertAssetInput => {
+        return toInsertEquipmentInput(row, catalogueItemData);
+      }),
```
mark-prins, line 44:
```suggestion
      emptyRows.map(
        (row: ImportRow): InsertAssetInput =>
          toInsertEquipmentInput(row, catalogueItemData)
      ),
```

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+      t
+    );
+    FileUtils.exportCSV(csv, t('filename.cce'));
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const csvImport = (files: any) => {
```
mark-prins, line 51: No need for the `any`
```suggestion
const csvImport = <T extends File>(files: T[]) => {
```

just change this line also:

```
    if (!csvFile?.name.endsWith('.csv')) {
```

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+          setIsLoading(false);
+          onUploadComplete();
+        },
+      });
+    }
+    error(t('messages.error-no-file-selected'));
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const processUploadedDataChunk = (data: any) => {
```
mark-prins, line 80:
```suggestion
  const processUploadedDataChunk = <T extends File>(data: ParseResult<T>) => {
```

just change the papaparse import:

```
import Papa, { ParseResult } from 'papaparse';
```

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+    if (!data.data || !Array.isArray(data.data)) {
+      // Don't think this is likely to happen...
+      console.info('data not data');
```
mark-prins, line 83: can this be removed?

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: you can create without a catalogue item through the UI.. but let's leave it off the import for now

package.json
```diff
@@ -7,5 +7,6 @@
     "start": "cd ./server && cargo run & cd ./client && yarn start-local",
     "build": "cd ./client && yarn build && cd ../server && cargo build --release",
     "test": "cd ./server && ulimit -n 1000 && cargo test && cd ../client && yarn test"
-  }
+  },
```
mark-prins, line 10: can just be reverted now I think

mark-prins (changes requested):
Working nicely thanks Ferg! There's a lot of work in there.

I've commented throughout, I think these are the key points to change 🙏 

- Change the `OK` button to be a `Next` button i.e. show progress rather than finality
- Formatting of the template link
- Additional help text to explain the export function (maybe?) and the import requirements (maybe?)
- use Next button rather than Ok button

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
fergie-nz: Do you think best to not include code at all in the import csv?

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -0,0 +1,173 @@
...
+          justifyContent="center"
+          style={{ minHeight: '75vh' }}
+        >
+          <InlineProgress variant={'indeterminate'} color={'secondary'} />
+        </Grid>
+      ) : null}
+      <Stack spacing={2}>
+        <Upload onUpload={csvImport} />
+        <Paper>
+          <Typography variant="h5" textAlign="center">
```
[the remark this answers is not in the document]
fergie-nz: Yes I love it, outstanding eye for aesthetic, thank you

client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+          variant="ok"
+          disabled={importNotReady}
+          onClick={async () => {
+            importAction();
+          }}
```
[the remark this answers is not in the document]
fergie-nz: I'm getting an error when I try this, so have left an anonymous function (without the async)

Image 26-03-24 at 2 01 PM

mark-prins:
Sorry a correction!
Have reverted the `UploadTab` type - from `T extends File` to `any`. it's a bit messy - would like to correct, run out of time!

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
fergie-nz: I've gone down a bit of a chain of fixes to explicitly handle the dynamic keys of the parsed result.

It also exposes some weird double checking for undefined behaviour from our linter...

Maybe OK? Curious to hear your thoughts. I find these type definitions a bit arcane, so I'm unsure of the best approach here.

Fixes are:

```
// introduce new interface to accomomdate dynamic keys of parsed result
interface ParsedAsset {
  id: string;
  [key: string]: string | undefined;
}
```

and

```
const processUploadedDataChunk = (data: ParseResult<ParsedAsset>) => {
```

and here double checking for undefined (linter complains with just a single check?)

 ```
 if (
        row[t('label.asset-number')] &&
        row[t('label.asset-number')]?.trim() != ''
      ) {
        importRow.assetNumber = row[t('label.asset-number')] ?? '';
```

File: client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx (hunk 5 above)
Comment (fergie-nz) [reply client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx:232]: I'm just thinking through this case. Currently I'm automatically going to the next stage on upload - but do you think it would be better to wait for the user to click the button?

Currently also uploaded data is cleared when we move back to step one - but maybe should cancel this behaviour also.

I'm a little unclear on what the best behaviour would be

Comment (fergie-nz) [reply client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx:232]: I'm wondering if we need a next button at all and in stead automatically move to the next step on upload or on import

File: client/packages/coldchain/src/Equipment/ImportAsset/ImportReviewDataTable.tsx [hunk 17]
```diff
@@ -0,0 +1,109 @@
...
+interface ImportReviewDataTableProps {
+  importRows: ImportRow[];
+}
+export const ImportReviewDataTable: FC<ImportReviewDataTableProps> = ({
+  importRows,
+}) => {
+  const t = useTranslation('coldchain');
+  const [pagination, setPagination] = useState<Pagination>({
+    page: 0,
+    first: 10,
```
mark-prins, line 23: `10` isn't a valid value for `rowsPerPage` and this causes the selection to fail. 
can change to `20` to make it work

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: The dynamic indexing of the row object is a bit problematic. it's a shame that the row props are the headings which are translatable. 

How about [this suggestion](https://github.com/msupply-foundation/open-msupply/compare/3139-import-asset-from-spreadsheet...3139-import-asset-from-spreadsheet(enum-version)) as a bit of a column number attempt?

I've used an enum to specify the column indices and extracted the row keys so that we can use those. An alternative version would be to specify the translations, like:

```
    const columns = {
      AssetNumber: t('label.asset-number'),
      CatalogueItemCode: t('label.catalogue-item-code'),
      Notes: t('label.asset-notes'),
    }

    const assetNumber = row(columns.AssetNumber);
```

which makes the code neater and doesn't rely on the ordering of the keys. The disadvantage is that you lose the ability for users to name columns as they want, as long as the columns are in the correct order they'll import ok..

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: no - keep the code, I think it's good to have.

mark-prins:
one other thing sorry!
the style of the import box isn't the same as figma:

image

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
fergie-nz: I imagine that having dynamic translations would be important given this is user facing. 

Is that a fair assumption?

I'll give your suggestion a go

mark-prins: If it means compromising on code quality, I'd drop the requirement. It's good to have, but not essential

client/packages/coldchain/src/Equipment/ImportAsset/ImportReviewDataTable.tsx
fergie-nz: I'm curious why this is? Does it need to be higher? Or an enum in the back end?

mark-prins: It's an enum - the display is a Select and if you don't provide one of the available options you'll get a console warning, and no value selected in the component.

mark-prins (approved):

mark-prins:
Thanks for all the changes Ferg! 🙏 
Looking great - and all good to merge once the conflicts are resolved

fergie-nz:
Some things to add:

Remove id from column on export
Add column to specify by store
Ensure code is displaying rather than catalogue item ID

Remove next button altogether

Import with a store column (maybe add this on a separate issue?)

Unexplained extra tall modal during import process. Maybe we can ignore this as @adamdewey mentioned this is not a commonly used process, so doesn't need to be a focus of UX.

adamdewey:
@fergie-nz - the other fields we need to be able to import is the `serial number` and `installation date`

But they should be optional, not mandatory

PR #3433, opened 2024-03-25
3299 link original shipment to return

### LINKED ISSUE #3299: Link a shipment to a return (5h)

When I create an Outbound Return from an Inbound Shipment, I want to know what shipment I created that return from. Same for Inbound Return from Outbound Shipment. 

I want to see this link in the Related Documents section of the return detail view side panel.

We have a `linked_invoice_id` on an invoice, but this is for the generated "other half" i.e. the Inbound Return for an Outbound Return. But we don't have a way to connect an Inbound Shipment to an Outbound Return.

Will need a new join to link related invoices in the same store. 

I'm not sure if this should be use case specific, or more generic/allowing many-many so we have the ability to connect more invoices together where needed in future?

A manually created return won't have a related shipment. 

Make sure that this new field is being transferred/synced!

lache-melvin:
Fixes #3299

# 👩🏻‍💻 What does this PR do? 
 
When a return is created from a shipment, store that original shipment ID so we can link to that shipment from the return's "Related Documents":
Screenshot 2024-03-26 at 12 25 56 PM

You'll need latest develop of mSupply running!

- Find an inbound shipment that was automatically created (i.e. has a link to an outbound shipment in another store)
- Create a return from some of those inbound shipment lines
- In the Outbound Return detail view, see the inbound shipment in Related Documents in the detail tab
- Make sure Outbound Return is at least Picked, so it will be transferred to the other store
- Switch to that supplying store
- Go to Inbound Returns, see the new return
- In the Inbound Return detail view side panel, you should see a related Outbound Shipment - this should be the original outbound shipment!

Manual returns have no related documents.

Note, we don't currently have any relationship from the shipment back to the return. I.e. in the Outbound Shipment, there is no "related returns". This was agreed to in a stakeholder meeting (as potentially we could have many returns from a shipment - different lines at different times...) but just flagging that here!


client/packages/invoices/src/Returns/InboundDetailView/SidePanel/RelatedDocumentsSection.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { AppRoute } from '@openmsupply-client/config';
+import { useReturns } from '../../api';
+
+export const RelatedDocumentsSectionComponent = () => {
+  const t = useTranslation('distribution');
+  const { localisedDate: d } = useFormatDateTime();
+  const { data } = useReturns.document.inboundReturn();
+  const { originalShipment } = data ?? {};
+
+  let tooltip = '';
```
andreievg, line 23: hmm, I think tooltips don't really work for mobile users (at least it's not apparent that they exist).
I wonder if we just remove this or add the 'info' icon (to suggest it's pressable). Or even include this info as the label ?

andreievg: Screenshot 2024-03-27 at 4 24 32 PM

andreievg (approved):
Looks great thanks, i can't see any issues with logic, and you are become a reall rusta by the looks of it.

> Note, we don't currently have any relationship from the shipment back to the return. I.e. in the Outbound Shipment, there is no "related returns". This was agreed to in a stakeholder meeting (as potentially we could have many returns from a shipment - different lines at different times...) but just flagging that here!

Cools, although we can find related returns for outbound/inbound shipment now that original_shipment_id is set on the return. Maybe we can capture this in todo somewhere.

I tried the whole type of id thing, here it is, I don't think we should change for now but good to have this diff to refer to, lets chat about it and see if we want to do it as part of enhancement sprint.

https://github.com/msupply-foundation/open-msupply/compare/3299-link-original-shipment-to-return...3299-link-original-shipment-to-return(Use-id-type-shipment-or-name)

client/packages/invoices/src/Returns/InboundDetailView/SidePanel/RelatedDocumentsSection.tsx
lache-melvin: Ah yeah let's just include the info in the label - was a copy paste but I like the info being more easily available, and it's not like we're short on space in that side panel at the moment! 😁

lache-melvin:
> Cools, although we can find related returns for outbound/inbound shipment now that original_shipment_id is set on the return. Maybe we can capture this in todo somewhere.

Yep just need to add the logic for it - capture the todo in code or in the backlog?

> I tried the whole type of id thing, here it is, I don't think we should change for now but good to have this diff to refer to, lets chat about it and see if we want to do it as part of enhancement sprint.
> 
> [3299-link-original-shipment-to-return...3299-link-original-shipment-to-return(Use-id-type-shipment-or-name)](https://github.com/msupply-foundation/open-msupply/compare/3299-link-original-shipment-to-return...3299-link-original-shipment-to-return(Use-id-type-shipment-or-name))

Oo nice one! Yes would be keen to riff on this together 🚀

PR #3435, opened 2024-03-25
Don't allow clearing of languages

### LINKED ISSUE #3387: Admin -> Display Settings -> Language -> Clear Selection - unclear what the function does, currently not performing any actions
## What went wrong? 😲
 

When navigating under Admin -> Display Settings -> Language -> Clear Selection, it is unclear what the expected behaviour is. Is clicking that button supposed to reset the language back to English? If so, the button is currently not performing that action. Is there a need to have a separate button to clear selection, if we can switch back to English in the same number of clicks?

This is the same for Desktop & Tablet. 

https://github.com/msupply-foundation/open-msupply/assets/153777484/a31b943a-5e8d-4906-a152-cbcd7945ec7c

## Expected behaviour 🤔

If the button is intended for resetting the language back to English, then it should reset the language back to English. 

## How to Reproduce 🔨

1. Go to Admin -> Display Settings -> Language
2. Change the language from English to any other one (for example, French)
3. Click 'Clear Selection' -> note the button does not do anything

## Your environment 🌱

- Version: 1.7.00-rc3

- Platform:Windows Server 2022 Desktop

- Database type: ?

roxy-dao:
Fixes #3387

# 👩🏻‍💻 What does this PR do? 
Don't allow users to clear language 

- [ ] Login as admin
- [ ] Go to `Admin`
- [ ] Click on language selection
- [ ] Shouldn't see `Clear Selection`

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Select/Select.tsx
+++ b/client/packages/common/src/ui/components/inputs/Select/Select.tsx
@@ -25,7 +25,7 @@ const defaultRenderOption = (option: Option) => (
 
 export const Select: FC<SelectProps> = React.forwardRef(
   (
-    { options, renderOption, sx, InputProps, clearable = true, ...props },
+    { options, renderOption, sx, InputProps, clearable = false, ...props },
     ref
   ) => {
     const t = useTranslation();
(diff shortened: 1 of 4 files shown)


mark-prins (approved):
❤️  I've seen this 'clearable' appear in a number of unwanted places now. Given that it is more standard for a Select to not be cleared, I wonder if the prop should default to false?

roxy-dao:
> ❤️ I've seen this 'clearable' appear in a number of unwanted places now. Given that it is more standard for a Select to not be cleared, I wonder if the prop should default to false?
> 
> Nah, ignore that. There's only a small number (<20) usages and they all look ok to me

Sure! I'll go through and see what can be and can't (: and add it to this PR

PR #3445, opened 2024-03-26
Feature: Returning Goods

lache-melvin:
Merging Feature: Returning Goods

# 👩🏻‍💻 What does this PR do? 
 

Merge returns work into develop (finally 🙈)

We still have some outstanding PRs/issues, but we can do these against devel

You'll need the latest develop of mSupply (i.e. 7.14) for support of both return reasons and original shipments (linking a return back to original shipment) ... both of these are actually still in PR to go into feature branch but they're COMING

Individual PRs have been reviewed within Ruru.


andreievg (approved):
I've checked all of the shared code, looks good to me, a test is failing for me due to foreign key constraint on currency-invoice constraint for key on invoice for `test_insert_inbound_return_success`. 

However i am really confused on that front and started a discussion about it in omsupply dev chat

PR #3506, opened 2024-04-02
Show error but don't block sync if OMS Central is not available

### LINKED ISSUE #3495: OG central stops syncing if OMS central fails
## What went wrong? 😲
Can't get OG to sync again if OMS central is not on and fails.

## Expected behaviour 🤔
Should still be able to sync!

## How to Reproduce 🔨

## Your environment 🌱

- Version: 
- mSupply version: 

- Platform:

- Database type:

jmbrunskill:
Fixes #3495

# 👩🏻‍💻 What does this PR do? 
Logs an error but doesn't stop sync if there's an error uploading to open-mSupply Central

Works for me :)

## 📃 Documentation

_No user facing changes_


roxy-dao (approved):
Brilliant~ thanks

PR #3507, opened 2024-04-02
3398 file sync upload to central server

### LINKED ISSUE #3398: File Sync -> Upload to central Server

As part of the file sync feature, following on from https://github.com/msupply-foundation/open-msupply/pull/3397

- [x] Files uploaded to a remote site should be uploaded to central server (in a background process)
- [ ] Some kind of upload status should be reflected in the UI
- [x] Any files not available on the remote server should be fetched from the central server on demand

> Downloading the files to sync down for the site will be in a separate PR

See: https://github.com/msupply-foundation/open-msupply/blob/develop/decisions/2024-02-28_file_sync.md

jmbrunskill:
Fixes #3398

# 👩🏻‍💻 What does this PR do? 

Implements a MVP of file sync for open-mSupply.

When a file is uploaded on a remote site, it will be automatically sent to the central server in a separate process to the main sync process.

When a remote site needs to access a sync file, it will be downloaded and then returned to the client.

This means that files will be downloaded at the time they're accessed from central. The offline process of downloading the relevant files can be worked on in a future PR if desired. It's not clear how useful this would with limited storage space on android devices 

- [x]  Files uploaded to a remote site should be uploaded to central server (in a background process)
- [x]  Any files not available on the remote server should be fetched from the central server on demand

Testing is a bit challenging.

You'll need to have 3 open-mSupply servers.

1. One `Central Server` Running on the same IP Address as mSupply Central, with a port number +2 from the mSupply one
2. You'll need a Remote server with one sync site which you can use to upload files
3. You'll need to test you can download the file on a different sync site

Files can be uploaded with a command something like this...
`curl --location --request POST 'http://localhost:8000/sync_files/SOME_RECORD_TYPE/SOME_RECORD_ID' \
--form 'files=@"/path/to/your_file.jpg"'`

To access the file you'll need the id from the response and create a url like this...
http://your_remote_site/sync_files/SOME_RECORD_TYPE/SOME_RECORD_ID?id=THE_UPLOADED_FILE_ID

> Note: Files won't be updated until sync is completed and the file_reference record exists on the central server.
Likewise if the file hasn't been uploaded to central yet, the remote site won't able to download it either.

- [ ] I've implemented a pause process for uploads, do you think it's important to chunk file uploads so that the main sync requests aren't held up for long?
- [ ] Do we need to try to validate that the Record Types/ Record ID's before accepting an upload?
- [ ] Do we need some kind of permission/authentication on uploading a file for sync (Thinking probably Yes!)
- [ ] Should we have some kind of limit on the number of attempts to upload?

## 📃 Documentation

- [ ] File sync documentation required. This might be best added as part of the sync explanations in https://docs.msupply.foundation/docs/sync/synchronisation/


server/graphql/core/src/lib.rs
```diff
@@ -13,8 +13,8 @@ use actix_web::web::Data;
 use actix_web::HttpRequest;
 use async_graphql::{Context, Request, Response};
 
+use actix_web::http::header::COOKIE;
```
jmbrunskill, line 16: It seems like changing the reqwest import meant that we need to be correct about which `header` package we use through out the code base...

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -31,6 +54,26 @@ pub struct SyncFileReferenceRow {
     pub record_id: String,
     pub file_name: String,
     pub mime_type: Option<String>,
+    #[serde(skip_serializing)]
```
jmbrunskill, line 57: All these 
```
    #[serde(skip_serializing)]
    #[serde(default)]
```
Records are local only state, so not included in the sync record.
Feels a bit weird that that logic is here in the repository layer but using serde does make this easy...

server/service/src/sync/file_sync_driver.rs
```diff
@@ -0,0 +1,196 @@
...
+pub struct FileSyncTrigger {
+    sender: Sender<FileSyncMessage>,
+}
+
+/// Used to 'drive' file sync synchronisation, it's tasks:
+/// * Expose channel for manually triggering sync
+/// * Trigger sync every SyncSettings.interval_seconds (only when initialised)
+impl FileSyncDriver {
+    pub fn init(settings: &Settings) -> (FileSyncTrigger, FileSyncDriver) {
+        // We use a single-element channel so that we can only have one trigger message processed at a time (no need to queue them up?)
```
jmbrunskill, line 40: I wonder if this could somehow cause us problems with slow connections, long integrations or something?

server/service/src/sync/file_sync_driver.rs
```diff
@@ -0,0 +1,196 @@
...
+    ///
+    ///
+    /// Operations:
+    /// * loop
+    ///    * If initialised await for  trigger OR interval sec timeout
+    ///    * If not initialised await only for start trigger
+    ///    * do sync if any of the above were triggered
+    pub async fn run(mut self, service_provider: Arc<ServiceProvider>) {
+        // Default to a paused so file sync should un-pause once the first `sync` completes
+        let mut stopped = false;
```
jmbrunskill, line 68: I've added this `stopped` state as well as `pause` as conceptually we might want to have a stop file sync button that users can use if the want to disable uploads due to network connection issues or something?

server/service/src/sync/file_synchroniser.rs
```diff
@@ -0,0 +1,221 @@
...
+        // Upload a file (In future this could be a chunk of data, instead of a whole file)
+        // Update the file record with the progress
+        // Yield to the runtime to check if we've received a pause signal
+
+        // Get any files that need to be sent to central server
+        let sync_file_repo = SyncFileReferenceRowRepository::new(&ctx.connection);
+        let files = sync_file_repo.find_all_to_upload()?;
+
+        // Pick a random file and upload it
+        // A random file is chosen in case one particular file is having problems, others should be able to continue
```
jmbrunskill, line 75: Not sure random is needed now I've added a retry delay.
Previously had a bug were if it was trying to upload a file to central but central had lost the sync record (due to re-initialising I think) This meant file sync was stuck retrying forever even though it will never work..

server/service/src/sync/file_synchroniser.rs
```diff
@@ -0,0 +1,221 @@
...
+
+                let result = self.try_uploading_file(file).await;
+
+                let bytes_uploaded = match result {
+                    Ok(bytes_uploaded) => bytes_uploaded,
+                    Err(err) => {
+                        log::error!("Error uploading file: {:#?}", err);
+
+                        // Update database to record the chunk has failed to upload
+                        if file.retries >= MAX_UPLOAD_ATTEMPTS {
```
jmbrunskill, line 93: I'm on the fence about if we want a max upload attempts or not... Feels like it if it's genuinely never going to work we should give up at somepoint but when?

Was thinking if we have a nice UI an admin could trigger a re-try upload, maybe when on a better network connection or something?

server/service/src/static_files.rs
```diff
@@ -135,6 +149,130 @@ impl StaticFileService {
...
+        let (table_name, record_id) = match category.clone() {
+            StaticFileCategory::SyncFile(table_name, record_id) => (table_name, record_id),
+            _ => {
+                return Err(anyhow::Error::msg(
+                    "Can't download file from central server, as it's not a sync file!",
+                ))
+            }
+        };
+
+        let ctx = service_provider
```
andreievg, line 175: I think `basic_context()` should do

andreievg: Also there should be a mapping from database error to anyhoe werror already

server/service/src/static_files.rs
```diff
@@ -135,6 +149,130 @@ impl StaticFileService {
...
+    ) -> anyhow::Result<Option<StaticFile>> {
+        log::info!("Downloading file from central server");
+        if is_central_server() {
+            return Err(anyhow::Error::msg(
+                "Can't download file from central server, as I am the central server!",
+            ));
+        }
+
+        let (table_name, record_id) = match category.clone() {
+            StaticFileCategory::SyncFile(table_name, record_id) => (table_name, record_id),
```
andreievg, line 167: I feel like the verboseness of rust code can really get in a way of understanding logic, i like using Let/else and ok_or, etc to try and reduce it a little bit, for example: https://github.com/msupply-foundation/open-msupply/compare/3398-file-sync-upload-to-central-server...3398-file-sync-upload-to-central-server(Less-verbose-mapping)

andreievg: Also would this method ever be called with StatiFileCateogry that is not sync_file ? Maybe sync file needs to a be a type that is stored in StaticFileCategory::SyncFile ? And we can pass that type here explicitly then construt StatiFileCategory if we need to reserver file etc..

server/service/src/static_files.rs
```diff
@@ -135,6 +149,130 @@ impl StaticFileService {
...
+            None => {
+                return Err(anyhow::Error::msg(format!(
+                    "Can't find sync file reference with id: {}",
+                    id
+                )))
+            }
+        };
+
+        let file_name = sync_file_ref.file_name.clone();
+        let download_url = format!("{}/{}/{}?id={}", base_url, table_name, record_id, id);
```
andreievg, line 207: I feel like download and upload from central is very similar to sync api v6 pattern, I wonder if we can make it stronger type by using the same pattern ?

Sorry it would mean changing the logic a bit, I'll try to have a go at showing what mean in a suggestion.

server/server/src/static_files.rs
```diff
@@ -197,10 +211,77 @@ async fn upload_sync_file(
     Ok(HttpResponse::Ok().json(files))
 }
 
+#[put("/{table_name}/{record_id}/{file_id}")]
+async fn upload_sync_file_via_sync(
```
andreievg, line 215: As per convo, I think this should be part of sync on central

andreievg (commented):
Thanks James, looks like a lot of work!

I like the implementation in general, it's a bit hard to get my head around all of this though, to try to come up with use cases to validate the implementation. I think there are a couple of things that could improve and help with this:
* Extra statuses and field for multi part, I think if we are not using multipart right now, it's a good idea to leave it out. Yes we talked about the schema design and allowing for future improvements, but I think adding some extra fields and statuses in the future (before 2.0) should be quite straight forward ?.
* Rust is pretty verbose, I did an example of error mapping that might simplify, also in some places i saw an error or ok being being mapped to another error and saved in a record, in those cases I think it might be good to do an inner method
* I feel like the communication between central and remote introduces another path, i think it might be better using existing one (sync-v6), including partitioning and direct type mapping.

I'll have a go at an example of the above, since you are probably quite busy atm trying to finish off the Asset import etc..

andreievg:
I am not sure i understand this one:

https://github.com/msupply-foundation/open-msupply/blob/7c75f5794e16c005d25addf413e4e0af5b542c2b/server/service/src/static_files.rs#L138-L143

https://github.com/msupply-foundation/open-msupply/blob/7c75f5794e16c005d25addf413e4e0af5b542c2b/server/service/src/static_files.rs#L36

Why do we have two ids ? And am I understanding it correctly, file is stored as: (sync_files/{some_id}/{another_id}_{actualfilename})

jmbrunskill:
> Why do we have two ids ? And am I understanding it correctly, file is stored as: (sync_files/{some_id}/{another_id}_{actualfilename})

It's `sync_files/{table_name}/{associated_record_id}/{file_id}_{actualfilename}`

andreievg (commented):
I created an example, which i haven't test, but it does compile and it was made to show an alternative, which I think fits closer to existing patterns (by the way this is only done for download, not upload):

* Communication between instances of omSupply should ideally 'live' in the same place (the sync layer
* Layers this communication in the same way (and uses the same auth)
* Reduces the need to pass table_name, record_id (in some places)
* Use this error, and less verbose error mapping

https://github.com/msupply-foundation/open-msupply/compare/3398-file-sync-upload-to-central-server...3398-file-sync-upload-to-central-server(Example-of-sync-layers)

See commit (https://github.com/msupply-foundation/open-msupply/commit/9d613edb70cf84debb3ec3e318fcccd0d6f37224) for futher comments (and to comment back)

A couple of other topics for discussion:

I didn't actually add the check to see if we can download file on central, when client is calling the end point, since that check exists on the end point

What would be the auth for downloading file locally ? Currently this is behind graphl protection, since graphql query will give UUID for a temporary file, which is exposed on rest without auth but UUID shouldn't be guessable and it expires.

So yeah I haven't tested, can do, I think it's quite important to keep 'sync' logic and patterns uniform and consistent (in term of partitioning and location with code code base and layering). I think the upload would need to be updated in a similar way, actaully with upload, do we allow the upload end point to also change the record ? (the file reference entry ?), does that create a sync record from central to remote (and potentially overwrite the statuses ?). I think we should only upload the file when the sync file reference record already exists on central, which means we only really need to pass 'id' (rather then table_name and record_id)

server/service/src/sync/file_sync_driver.rs
jmbrunskill: Yeah we kind of want to consume only the latest message, but not sure what's the a nice way to do that.

server/service/src/sync/file_sync_driver.rs
jmbrunskill: Yes, exactly if the user stops it we don't want to start it automatically when the sync finishes

server/service/src/sync/file_synchroniser.rs
jmbrunskill: Yeah, was trying to create an opportunity for the `pause` message to get through and stop the uploading.
We could potentially cancel the inprogress upload process somehow, but that seems difficult and possibly problematic if a file takes more time than a single sync to upload...

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -108,10 +154,32 @@ impl<'a> SyncFileReferenceRowRepository<'a> {
         diesel::update(sync_file_reference.filter(id.eq(sync_file_reference_id)))
             .set(deleted_datetime.eq(Some(chrono::Utc::now().naive_utc())))
             .execute(&self.connection.connection)?;
-        self.insert_changelog(
-            sync_file_reference_id.to_owned(),
-            ChangelogAction::Delete,
-        )?;
+        self.insert_changelog(sync_file_reference_id.to_owned(), ChangelogAction::Delete)?;
```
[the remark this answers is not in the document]
jmbrunskill: Agreed good point. Updated.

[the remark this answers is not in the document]
jmbrunskill: [33d56a5](https://github.com/msupply-foundation/open-msupply/pull/3507/commits/33d56a581e5a7b5ad7cf1b258231c2fe0d9d5a54)

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -108,10 +154,32 @@ impl<'a> SyncFileReferenceRowRepository<'a> {
...
+    pub fn find_all_to_upload(&self) -> Result<Vec<SyncFileReferenceRow>, RepositoryError> {
+        let result = sync_file_reference
+            .filter(deleted_datetime.is_null())
+            .filter(uploaded_bytes.lt(total_bytes))
+            .filter(
+                status
+                    .eq(SyncFileStatus::ReadyToUpload)
+                    .or(status.eq(SyncFileStatus::Uploading))
+                    .or(status
+                        .eq(SyncFileStatus::UploadError)
```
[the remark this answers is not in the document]
jmbrunskill: Do you mean remove the filter
` .filter(uploaded_bytes.lt(total_bytes))` ?
Yes I think we can unless we do uploading in chunks...

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -108,10 +154,32 @@ impl<'a> SyncFileReferenceRowRepository<'a> {
...
-            ChangelogAction::Delete,
-        )?;
+        self.insert_changelog(sync_file_reference_id.to_owned(), ChangelogAction::Delete)?;
+        Ok(())
+    }
+
+    pub fn find_all_to_upload(&self) -> Result<Vec<SyncFileReferenceRow>, RepositoryError> {
+        let result = sync_file_reference
+            .filter(deleted_datetime.is_null())
+            .filter(uploaded_bytes.lt(total_bytes))
```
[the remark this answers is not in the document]
jmbrunskill: Makes sense

server/service/src/sync/settings.rs
```diff
@@ -42,4 +45,11 @@ impl SyncSettings {
             && self.password_sha256 == other.password_sha256;
         !equal
     }
+
+    pub fn file_upload_base_url(&self) -> String {
+        let omsupply_central_url = get_omsupply_central_url(&self.url)
+            .unwrap_or(Url::parse("http://localhost:8000").unwrap()); // This is hacky but I think ok for now?
+
+        format!("{}sync_files", omsupply_central_url)
```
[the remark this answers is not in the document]
jmbrunskill: Yes good idea. Done

server/service/src/sync/file_synchroniser.rs
```diff
@@ -0,0 +1,221 @@
...
+    fn from(error: RepositoryError) -> Self {
+        FileSyncError::DatabaseError(error)
+    }
+}
+
+pub struct FileSynchroniser {
+    settings: SyncSettings,
+    service_provider: Arc<ServiceProvider>,
+    static_file_service: Arc<StaticFileService>,
+    client: reqwest::Client,
```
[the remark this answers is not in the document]
jmbrunskill: You're right I think it's getting recreated every time.

server/service/src/sync/file_synchroniser.rs
andreievg: > if a file takes more time than a single sync to upload...

If the file takes longer then the interval/delay between syncs you mean ?

server/repository/src/db_diesel/sync_file_reference_row.rs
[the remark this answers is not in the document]
andreievg: Yep, just to reduce the number of things to think about for now (plus I think we should just set the status as done uploading ? Rather then look up total_bytes uploaded

server/service/src/sync/file_synchroniser.rs
jmbrunskill: Yes

jmbrunskill:
@andreievg I think I've addressed most of your concerns, anything important left?

The only thing I'm aware of is Auth, have put that in  anew issue to try to get this PR closed out.
https://github.com/msupply-foundation/open-msupply/issues/3601

andreievg (approved):
Thanks for the changes so far! Approving, pending:
* Authorisation (incoming in another PR, thanks)
* `find_all_to_upload`, was wondering it that's ok or not, left a comment

server/repository/src/db_diesel/sync_file_reference_row.rs
```diff
@@ -108,10 +161,29 @@ impl<'a> SyncFileReferenceRowRepository<'a> {
...
+        self.insert_changelog(sync_file_reference_id.to_owned(), ChangelogAction::Delete)?;
+        Ok(())
+    }
+
+    pub fn find_all_to_upload(&self) -> Result<Vec<SyncFileReferenceRow>, RepositoryError> {
+        let result = sync_file_reference
+            .filter(deleted_datetime.is_null())
+            .filter(direction.eq(SyncFileDirection::Upload))
+            .filter(
+                status.eq(SyncFileStatus::New).or(status
```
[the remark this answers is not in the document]
jmbrunskill: This version lets us do the retry at logic, easily, I find this easier to reason about than `Not Done || Not Permentant`.

I wasn't sure if we wanted to process `InProgress`, However I've convinced myself that if the server is restarted it could be stuck as Inprogress so have updated.

PR #3510, opened 2024-04-02
3441 ensure sync buffer merge records don't override upserts/delete

### LINKED ISSUE #3441: Correctly insert MERGE records into the sync_buffer
Each sync record received from mSupply contains a record_id which is the id of the entity, e.g. invoice.id.
This record_id is used as a primary key in the sync_buffer table.
This means the sync_buffer only contains the latest update for an entity.

However, this can fail when an entity is created and then merged in the same integration step:
- the UPSERT event for the new entity is overwritten by the MERGE event (deleting the old entity and keeping the new entity)
- the entity is never created and thus can't be merged
- thus the entity is missing

Note, in the future this could also cause problems for link tables or the context table which also share entity ids across tables.

## Example to reproduce the problem
- Choose item A
- Create a new item B that is a duplicate of item A
- I merged item A into item B (keep the new item B) 

Afterwards I got, these new entries in the sync_buffer:

image

The new item  "7A51FD..." is not there and thus the merge fails! mSupply bug?

_Originally posted by @clemens-msupply in https://github.com/msupply-foundation/open-msupply/issues/3306#issuecomment-2005586684_

## Proposed Solution

Instead of upserting records in the sync_buffer, append them, i.e. keep the full list of sync records.
This requires to change the sync_buffer so that the record_id is not the primary/unique key anymore.

Furthermore, we need to make sure that received_datetime strictly increases following the order as received by mSupply. For example, an early sync record should have a smaller received_datetime.
This is needed to integrate records in the correct order.

# Future work
Clean up redundant records from the sync buffer.
For example:

|record_id|action||
|---|----|-|
|1| UPSERT|redundant|
|1|UPSERT| latest|
|2|UPSERT|redundant|
|2|UPSERT|needed by merge|
|2|MERGE|latest|
|3|UPSERT|redundant|
|3|MERGE|redundant|
|3|UPSERT|latest|

Also see https://github.com/msupply-foundation/msupply/issues/14513

Chris-Petty:
Fixes #3441

# 👩🏻‍💻 What does this PR do? 

Change the `sync_buffer.record_id` value to be a new uuid for merge records to avoid it overriding upserts of the target records being updated.

- Have a primary server and an OMS remote site initialised
- On the primary create 2 items, sync
- Merge one into the other
- Sync OMS - should integrate records without error and the "kept" item should be shown in item list.

### DIFF
--- a/server/service/src/sync/api/common_records.rs
+++ b/server/service/src/sync/api/common_records.rs
@@ -124,3 +124,140 @@ impl CommonSyncRecordV5 {
         }
     }
 }
+
+// tests
+#[cfg(test)]
+mod tests {
+    use repository::{mock::MockDataInserts, test_db::setup_all, SyncBufferRowRepository};
+
+    use crate::sync::translations::special::ItemMergeMessage;
+
+    use super::*;
+
+    #[test]
+    fn test_insert_to_upsert_mapping() {
+        let record = CommonSyncRecordV5 {
+            table_name: "test".to_string(),
+            record_id: "test".to_string(),
+            action: SyncActionV5::Insert,
+            data: json!({}),
+        };
+
+        let row = record.to_buffer_row().unwrap();
+        assert_eq!(row.table_name, "test");
+        assert_eq!(row.record_id, "test");
+        assert_eq!(row.action, SyncBufferAction::Upsert);
+        assert_eq!(row.data, "{}");
+    }
+
+    #[actix_rt::test]
+    async fn test_remote_sync_batch_v5_to_sync_buffer_rows() {
+        let batch = RemoteSyncBatchV5 {
+            queue_length: 0,
+            data: vec![
+                RemoteSyncRecordV5 {
+                    sync_id: "test1".to_string(),
+                    record: CommonSyncRecordV5 {
+                        table_name: "item".to_string(),
+                        record_id: "itemA".to_string(),
+                        action: SyncActionV5::Insert,
+                        data: json!({
+                            "ID": "itemA",
+                            "item_name": "itemA",
+                            "code": "itemA",
+                            "unit_ID": "",
+                            "type_of": "general",
+                            "default_pack_size": 1,
+                        }),
+                    },
+                },
+                RemoteSyncRecordV5 {
+                    sync_id: "test2".to_string(),
+                    record: CommonSyncRecordV5 {
+                        table_name: "item".to_string(),
+                        record_id: "itemA".to_string(),
+                        action: SyncActionV5::Update,
+                        data: json!({
+                            "ID": "itemA",
+                            "item_name": "itemA",
+                            "code": "itemA",
+                            "unit_ID": "",
+                            "type_of": "general",
+                            "default_pack_size": 1,
+                        }),
+                    },
+                },
+                RemoteSyncRecordV5 {
+                    sync_id: "test3".to_string(),
+                    record: CommonSyncRecordV5 {
+                        table_name: "item".to_string(),
+                        record_id: "itemB".to_string(),
+                        action: SyncActionV5::Insert,
+                        data: json!({
+                            "ID": "itemB",
+                            "item_name": "itemB",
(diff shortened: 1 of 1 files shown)


Chris-Petty:
I am worried I haven't thought about enforcing order here. It needs to be guaranteed that integration of the `upsert` is done before the integration of the `merge`... right now hoping it's coincidentally always insertion order (`upsert` comes from central, then queue is pulled with `merge`).

Silly edge case town... central server could still have a `delete` in the central, then `merge` comes from queue. This would be from:

1. Create items A, B and C
2. Promptly merge A->B->C (the server will change the change_log entries to `delete` A and B).
3. The deletes won't do anything, then the merge will "fail". 

This is fine though, the 2 items truly practically never existed. The outcome is the same - only item C exists. Perhaps the merge translator should ignore rather than letting OMS log an integration error. Not a high priority so long as sync continues.

clemens-msupply:
> I am worried I haven't thought about enforcing order here. It needs to be guaranteed that integration of the `upsert` is done before the integration of the `merge`... right now hoping it's coincidentally always insertion order (`upsert` comes from central, then queue is pulled with `merge`).

Hehe I was just about to write a complaint, but you were quicker. Is it guaranteed that the central server sends the records in the correct order? or could they already be mixed up there? if not we could store the order in which the records arrive...

Chris-Petty:
> Hehe I was just about to write a complaint, but you were quicker. Is it guaranteed that the central server sends the records in the correct order? or could they already be mixed up there? if not we could store the order in which the records arrive...

@clemens-msupply The central server doesn't guarantee it as such, but the sync v5 protocol of pulling central_change_log records before pulling queued sync messages does guarantee the order.

clemens-msupply (approved):
Works fine, thanks!

Reproduced your case from your note:
- Create item A
- Create item B
- Merge item A into B (keep B)
- Sync -> merge integration error but item B is there

image

Think we should create a follow-up issue about record order and a "proper" sync_buffer (or keep existing issue open?)

clemens-msupply:
> @clemens-msupply The central server doesn't guarantee it as such, but the sync v5 protocol of pulling central_change_log records before pulling queued sync messages does guarantee the order.

Do you mean pulling queued messaged before change log changes? maybe easier to explain in the office...

> What I haven't looked into is what order comes out of sqlite or postgres if you just query sync_buffer for table `item`. I need to check if we also query for sync_buffer created datetime.

As far as I know they are ordered by `received_datetime` which should be okish in sqlite, e.g. (2024-04-02 21:00:02.196891368). But I remember Postgres has less precision for timestamps. If we re-do the sync_buffer table it might be better to have an auto increment column to keep the exact order...

Chris-Petty:
@clemens-msupply 
image

We "Pull Central" and save sync_buffer records for `item` `upsert`/`delete` before "Pull Remote" which includes the `merge` messages. Assuming insertion order then the integration step should have them in the right order.

# BUT ignore all that, synchroniser forces the order

Upserts > Deletes > Merges

https://github.com/msupply-foundation/open-msupply/blob/90c2180eabda6ee0669919089b65dd877314d813/server/service/src/sync/synchroniser.rs#L291-L321

andreievg:
> We "Pull Central" and save sync_buffer records for item upsert/delete before "Pull Remote" which includes the merge messages. Assuming insertion order then the integration step should have them in the right order.

It's also the translator order that determines when records are integrate, so should be ok for now, I think. https://github.com/msupply-foundation/open-msupply/blob/c92d6f325016983e84480335a576773bb07f2445/server/service/src/sync/translations/mod.rs#L57-L115

Chris-Petty:
@andreievg I'm just talking for a particular table the order of merge/delete/merge actions. What you're saying is for table order - though I thought that was determined by this beaut:

https://github.com/msupply-foundation/open-msupply/blob/e5b989e43a71a1609684f765e6d15a8fc2c045f9/server/service/src/sync/translations/mod.rs#L316-L320

PR #3513, opened 2024-04-02
Correct hook implementation

### LINKED ISSUE #3277: Cold Chain: error appears when interacting with the Monitoring Graph
## What went wrong? 😲
 

Navigating under **Cold Chain -> Monitoring -> Chart** -> and interacting with the graph by moving the line indicating a timeline triggers an error 'Oops. Something's gone wrong. Try again'. 

https://github.com/msupply-foundation/open-msupply/assets/153777484/0b8cd844-b63d-440d-89e5-b33e1c96d6a8

## Expected behaviour 🤔

We are able to interact with the graph without running into any issues. 

## How to Reproduce 🔨

1. Go to Cold Chain -> Monitoring -> Chart
2. Select any date period or no filters at all
3. Press on the screen -> a line indicating a point in time should appear
4. Try moving it around -> see error

## Your environment 🌱

- omSupply Version: 1.7.00

mark-prins:
Fixes #3277

# 👩🏻‍💻 What does this PR do? 
 
The hook `useFormatTemperature` was not following a standard pattern which resulted in the 'more hooks than previous render' error ([310](https://react.dev/errors/310?invariant=310))

Have changed to be a hook function which returns a function, rather than a fn which is called directly.

1. `yarn build`
2. `cargo run`
3. `yarn electron:start` 

and then try to use the temperature chart. I imported a fridge tag to get some data, and then adjusted the date filter so that the chart showed something

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/TemperatureBreachList.tsx
+++ b/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/TemperatureBreachList.tsx
@@ -59,7 +59,7 @@ const ListView: FC = () => {
 
   const pagination = { page, first, offset };
   const t = useTranslation('coldchain');
-  const formatTemperature = useFormatTemperature;
+  const formatTemperature = useFormatTemperature();
 
   const columns = useColumns<TemperatureBreachFragment>(
     [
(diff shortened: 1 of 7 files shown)


roxy-dao (approved):
Thanks Mark~!

PR #3514, opened 2024-04-02
3464 returns clean up

### LINKED ISSUE #3464: Returns bugs/missing bits
## What went wrong? 😲
 

I'm making docs, here are some bits we missed:

BUGS:
- [x] Outbound return edit quantity field de-focuses after typing.  Hack solution is that input field component should be defined outside of the useColumns hook (we've done this for Inbound) - I think a bigger refactor issue with the tables should involve a better solution to this!
- [x] Automatically created inbound returns should show this message before delivered:
Image
and this message once delivered:
Image
but when in Picked status its showing the post-delivered message? (then showing the right message in shipped status...) _**Marking as complete, but NOT FIXED** - I actually don't think this is broken, having the different message when shipped makes sense because that's when the are ABLE to confirm as Delivered_

TODO:
- [x] Delete outbound return from side panel
- [x] Outbound return log list
- [x] Update link to documentation

MAYBE?:
- [x] List views of returns, could/should include reference & comment, like shipments do
- [x] Modals don't know pack size
  - [x] Outbound return modal doesn't have the info button on the qty columns at least clarifying that qty is in num of packs
- [x] ~~Add transfer integration tests for new returns transfer logic~~ (own issue)
  

Also:
- Currently we allow returning goods from `Delivered` Inbound Shipment... should it only be from `Verified`?
- To check: we use the `availableNumberOfPacks` from the stock line but do we actually check that the stock isn't on hold?

lache-melvin:
Fixes #3464

# 👩🏻‍💻 What does this PR do? 
 

- Adds reference and comment to returns list views:
Screenshot 2024-04-03 at 9 55 28 AM
- Adds outbound return details log tab
Screenshot 2024-04-03 at 9 57 53 AM
- Fixes outbound return modal quantity inputs losing focus after you type
- Adds pack unit column, and info tooltips explaining that the quantity is in number of packs:
Screenshot 2024-04-03 at 9 50 35 AM
- Implements `delete` from outbound return side panel
- Link to external docs points to returns docs

- [ ] Inbound return list view: There is a reference and a comment column. If any of the returns have comments/references, you can see these here
- [ ] Outbound return list view: There is a reference and a comment column. If any of the returns have comments/references, you can see these here
- [ ] On outbound return details view, can click the Log tab and see a list of logs
- [ ] On outbound return details view, open the side panel - you can delete the outbound return with the delete button at the bottom, and get redirected back to the list of outbound returns
- [ ] On inbound/outbound return pages, clicking the Docs link in the app drawer takes you to the docs `/replenishment/outbound-returns` and `/distribution/inbound-returns` accordingly. (Will currently get a not found error as docs not published until we release 2.0.0)
- [ ] Outbound return edit modal
  - [ ] Quantity to return field doesn't lose focus after you type in it
  - [ ] You can see a pack size field
  - [ ] The quantity column headers have an info icon, you can hover over this to see that quantity is in number of packs

## 📃 Documentation

- [x] Screenshots of list views
- [x] Screenshots of edit modal


andreievg (approved):
Look great thanks.

I looked into the unit name for pack variant (inbound and outbound, and did this: https://github.com/msupply-foundation/open-msupply/compare/3464-returns-clean-up...3464-returns-clean-up(Use-item-loaded-and-unit-name))

* Use loaded on inbound/outbound return line
* This required a few more changes, creating a fragment and then using this in inputs (actually that is my preferred way, sorry I missed it in PR review), technically you can extract the type from the queries, but it's a bit harder with array return types.
* I also tried itemCode and itemName from loader, but then the generic column 'itemCode' and 'itemName' would need to be redone, so refrained (actually this should come from invoice line, not item row, as it keeps historic reference to the code and name of the item, if code or name of item are changed)

I compile and yarn generated, but haven't actually tested (by the way the 'unit' in pack variant cells can be checked by seeing the actual unit name vs just 'unit')

PR #3516, opened 2024-04-03
#3308 Implement disabling of completed returns

### LINKED ISSUE #3308: Disable verified returns in the list view (1h)

For shipments, the invoices that are not editable are greyed out (but still clickable to access the detail view). It would be good to implement similar functionality for the returns.
Image

Add a hook that determines whether or not the listed returns are disabled, and adjust their display accordingly.

CarlosNZ:
Fixes #3308

~~Putting this as a "Draft" PR initially, as I'm thinking of consolidating a couple of the "isDisabled" methods in `utils.ts` -- just waiting for clarification~~ Left as is for now.

# 👩🏻‍💻 What does this PR do? 
 

Implements "disabled" rows in list view for completed returns. Have used the same criteria for disabled as the respective shipments use -- assume this is correct?

Check returns lists -- any past "shipped" should be disabled (i.e greyed out, but still clickable)

The Shipments List views made a separate for `useDisableOutboundRows` etc. I haven't done this, as all it is is a wrapper around `useEffect` -- so I've just used `useEffect` directly in the component. Seems more straightforward to me.

### DIFF
--- a/client/packages/invoices/src/Returns/InboundListView/ListView.tsx
+++ b/client/packages/invoices/src/Returns/InboundListView/ListView.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React, { FC, useEffect } from 'react';
 import {
   useNavigate,
   DataTable,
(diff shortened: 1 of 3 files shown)


lache-melvin (approved):
Looks great! Thanks Carl :)

PR #3518, opened 2024-04-03
Add program relate report views

### LINKED ISSUE #3384: Report table views for SQL Reports

Related to: https://github.com/msupply-foundation/open-msupply/issues/2853

There are a couple of problems when directly accessing the raw DB tables in the report queries:
- Database tables may change their schema over time -> reports may break
- DB tables may contain internal logic, e.g. `name_link_id`

For this reason it would be good to introduce some report database views. These views should be stable, i.e. if the underlying table changes the report view stays compatible to the existing view. Moreover, the views should resolve internal data such as `name_link_id`.

By convention, reports must only use the report DB views.

Add views for (and define the appropriate columns). These views should only contain column that are important for report. Furthermore, the views could contain joined data if appropriated.

- [x] `report_store` (include store name here?)
- [ ] `report_name` (see above)
- [x] `report_patient` (name that are patients,  include document data column?)
- [x] `report_document` (latest documents)
- [x] `report_encounter` ( include document data column?)
- [x] `report_program_enrolment` ( include document data column?)

Not immediately needed:
- [ ] `report_invoice`
- [ ] `report_item`
- [ ] `report_stock`
- [ ] `report_stocktake`

clemens-msupply:
Fixes #3384

# 👩🏻‍💻 What does this PR do? 
Adds report views needed for programs reports. I think its better to wait with the other report views till they are needed.


jmbrunskill (approved):
Looks good to me. Nothing to test at this stage?

clemens-msupply:
> Looks good to me. Nothing to test at this stage?

Not really, there are two new copies of existing reports in: `3384-programs-report-views` in the `open-msupply-reports` repo that you could run but you would need some matching test dataset...

clemens-msupply:
I will wait with the merge, there might be some changes to the views because of Postgres performance issues...

PR #3519, opened 2024-04-03
Asset downloads tab

### LINKED ISSUE #3455: Asset detail view: Downloads tab
Part of epic #2896 

Add the contents of the document tab of the CCEI details page, like so:

image

* Modify the asset graphql query to return a list of related file ids
* Add the list component
* Add the file upload component
* Add an action button next to the upload component? 

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #3455

# 👩🏻‍💻 What does this PR do? 
 
* Adds the Documents tab
* Adds a delete method for documents which removes the reference in the db and tries to delete the local file
* Shows documents on the status log list 

Screenshot 2024-04-03 at 11 41 01 AM

Screenshot 2024-04-03 at 3 40 59 PM

Try to upload a document, then view and remove. Check that the file is deleted and the reference is removed.

I had branched off Ferg's branch in order to get the upload package.

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


jmbrunskill (changes requested):
I've tested this on Android and it's not working very well.

2 Kind of critical issues I think..
- [ ] Image open, but replace the entire OMS app, so you need to close it entirely and reopen. The back button doesn't appear to work :(
- [ ] PDF files don't open (clicking the link does nothing)

Might not be too much we can do however we shouldn't show the option to download if it will never work...

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -85,6 +87,7 @@
   "error.no-service-charges": "There are no service charges assigned to this store",
   "error.no-temperature-breaches": "No temperature breaches found",
   "error.problem-saving": "An error occurred: there was a problem saving the data.",
+  "error.an-error-occurred": "An error occurred: {{message}}",
```
jmbrunskill, line 90: Might be good to let the team know about this one, to avoid accidental similar translations being created?
I like the idea of this...

server/graphql/asset/src/types.rs
```diff
@@ -172,6 +172,16 @@ impl AssetNode {
         Ok(locations)
     }
 
+    pub async fn documents(&self, ctx: &Context<'_>) -> Result<SyncFileReferenceConnector> {
```
jmbrunskill, line 175: was contemplating if documents is correct but I think it's fine...

server/graphql/core/src/loader/sync_file_reference.rs
```diff
@@ -0,0 +1,230 @@
...
+                .clone();
+            let list = map
+                .entry(asset_id)
+                .or_insert_with(|| Vec::<SyncFileReference>::new());
+            list.push(sync_file_reference);
+        }
+
+        Ok(map)
+
+        // Ok(result
```
jmbrunskill, line 52: Remove these comments?

server/graphql/core/src/loader/sync_file_reference.rs
```diff
@@ -0,0 +1,230 @@
...
+        //         .or_insert_with(|| Vec::<Location>::new());
+        //     list.push(location);
+        // }
+
+        // Ok(map)
+    }
+}
+
+// #[cfg(test)]
+// mod tests {
```
jmbrunskill, line 106: Logic's simple here, don't think we need this type of test case.

server/graphql/core/src/loader/sync_file_reference.rs
```diff
@@ -0,0 +1,230 @@
...
+            Some(SyncFileReferenceSort {
+                key: SyncFileReferenceSortField::FileName,
+                desc: Some(false),
+            }),
+        )?;
+
+        let mut map: HashMap<String, Vec<SyncFileReference>> = HashMap::new();
+
+        for sync_file_reference in sync_file_references {
+            let asset_id = sync_file_reference
```
jmbrunskill, line 40:
```suggestion
            let file_id = sync_file_reference
```

server/repository/src/db_diesel/assets/asset.rs
```diff
@@ -159,10 +159,10 @@ impl<'a> AssetRepository<'a> {
             .limit(pagination.limit as i64);
 
         // Debug diesel query
-        println!(
-           "{}",
-            diesel::debug_query::<DBType, _>(&final_query).to_string()
-        );
+        // println!(
```
jmbrunskill, line 162: Thanks! Not sure how that got in possibly me?

server/repository/src/db_diesel/sync_file_reference.rs
```diff
@@ -0,0 +1,153 @@
...
+use crate::{repository_error::RepositoryError, DBType, EqualFilter, Pagination, Sort};
+use diesel::prelude::*;
+
+#[derive(PartialEq, Debug, Clone)]
+pub struct SyncFileReference {
+    pub sync_file_reference_row: SyncFileReferenceRow,
+}
+
+#[derive(Clone, PartialEq, Debug, Default)]
+pub struct SyncFileReferenceFilter {
```
jmbrunskill, line 22: Might want file_name as some stage,  but no need to add that yet.

server/server/src/static_files.rs
```diff
@@ -30,6 +30,7 @@ pub fn config_static_files(cfg: &mut web::ServiceConfig) {
     cfg.service(
         web::resource("/sync_files/{table_name}/{record_id}")
             .route(web::get().to(sync_files))
+            .route(web::delete().to(delete_sync_file))
```
jmbrunskill, line 33: Wondering if we should change both this and the `get` to use `/sync_files/{table_name}/{record_id}/file_id` instead of `/sync_files/{table_name}/{record_id}?id={file_id}` particularly for delete, it feels a bit wrong to use a query parameter rather than route path. Not that it really matters much!

jmbrunskill (commented):
Code wise looking good. Just need to figure out what to do about Android...

jmbrunskill:
This might be a solution? https://capawesome.io/plugins/file-opener/

server/graphql/core/src/loader/sync_file_reference.rs
mark-prins: oops 🙏

server/graphql/core/src/loader/sync_file_reference.rs
mark-prins: but it's returning the `record_id` which in this case is the asset id - the `file_id` would be the `id` on the sync_file_reference table wouldn't it?

server/repository/src/db_diesel/sync_file_reference.rs
mark-prins: thanks - I suspect I've already added 'future functionality' i.e. stuff that I'm not using yet. there's plenty of potential here!

server/server/src/static_files.rs
mark-prins: yeah, I like that idea. it's nice to use the `/file_id` pattern. agree it doesn't matter too much.. but it is nicer!

mark-prins:
I've changed the sync file patch, to remove `?id=` and have put in a 'fix' for android which results in this awfulness:

uh-oh

which is what we'd also get with the capacitor plugin, though with the benefit of a larger bundle as well.

Thoughts on that one:
- copy the file from the `static_files` dir to a publicly accessible dir and launch that ( I tried, had trouble )
- download and do the same ( the download completion didn't get called for some reason )
- serve the files on http as well as https ( that sounds possible, but I've not explored )

Essentially, I think this resolves the immediate problem and isn't really workable. you can view the file ok, but it's not great experience.

server/graphql/core/src/loader/sync_file_reference.rs
jmbrunskill: Yes you are right sorry must have miss read the code.

jmbrunskill:
> 
> which is what we'd also get with the capacitor plugin, though with the benefit of a larger bundle as well.
> 
> Thoughts on that one:
> 
> * copy the file from the `static_files` dir to a publicly accessible dir and launch that ( I tried, had trouble )
> * download and do the same ( the download completion didn't get called for some reason )
> * serve the files on http as well as https ( that sounds possible, but I've not explored )
> 
> Essentially, I think this resolves the immediate problem and isn't really workable. you can view the file ok, but it's not great experience.

Hmmm, feels it could be a hard problem!
I think this solution is good to get this PR in for now, and create a follow up issue?
Maybe the `http` is the easiest solution?

Otherwise something like what I've done for sqlite download? And maybe a special case for image files that can open in a modal somehow?

jmbrunskill (approved):
Looks good to go when you're ready Mark

mark-prins:
thanks @jmbrunskill - I'll merge it in and wait for feedback.. the current version is actually not too bad. once I've said to chrome that I don't mind insecurity, it opened other files without reprompting, so actually works fine.

would be nicer if we could have an image viewer and/or the http version

PR #3539, opened 2024-04-04
3496 inventory adjustment view stock tab

### LINKED ISSUE #3496: Add new tab in stock view modal for IA and simple adjustment
- Add new tab to stock view for Inventory adjustment
- Show current quantity
- Dropdown for stock increase or decrease
- Two input fields, one for the change amount, one for new total
  - Can update either, the other one adjusts automatically
  - Decide whether its easier to limit inputs or switch the increase/decrease if the inputted difference goes the wrong way?
- Dropdown for adjustment reasons
  - Like stock take - if reasons are available, its required
- Create/use mock API
- On save... clear inputs and update current quantity? Will a toast that the adjustment was successful.
- Does not need to show list of adjustments (will be in own ledger tab)

(Do we want to include a comment?)

lache-melvin:
Fixes #3496

# 👩🏻‍💻 What does this PR do? 
 

Adds `Adjust` tab to the stock modal. Went with `Adjust` as a verb/action word, but not sure if it's too vague?
Screenshot 2024-04-04 at 4 21 23 PM

- [ ] Select an adjustment direction
- [ ] Other fields should enable
- [ ] Enter a reason
- [ ] Update `adjust by` - `new number of packs` should update accordingly
- [ ] Update `new number of packs` - `adjust by` should update accordingly
- [ ] You get a success snack and the form is reset on save

Have added comments in the PR

## 📃 Documentation

- [ ] Overall inventory adjustment feature documentation


client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+  InventoryAdjustmentReasonSearchInput,
+  usePackVariant,
+} from '../../..';
+import { InventoryAdjustmentDirectionInput } from './InventoryAdjustmentDirectionSearchInput';
+
+// TODO... use direction input in stocktake?
+const tempAsAdjustment = (direction: AdjustmentDirectionInput | null) => {
+  switch (direction) {
+    case AdjustmentDirectionInput.Addition:
+      return Adjustment.Addition;
```
lache-melvin, line 25: we already had the `Adjustment` enum from Stocktake, but only exists in the frontend. Felt like it will be simpler logic to do on both front and backend with enum that we send in the mutation (backend business logic can use this direction to determine if the reason is correct, rather than deriving direction from the delta between current and new num of packs?)

Curious for thoughts on this!

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const { asPackVariant } = usePackVariant(
+    stockLine.itemId,
+    stockLine.item.unitName ?? null
+  );
+  const packUnit = asPackVariant(stockLine.packSize);
+
+  const saveDisabled = !draft.direction || draft.adjustBy === 0;
+
+  const save = async () => {
+    // TODO: handle error if no reason selected when reasons required
```
lache-melvin, line 54: assuming we'll do similar logic here to stocktake, backend will error if reasons are defined and one should be set here?

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+          text={packUnit}
+          textProps={{ textAlign: 'end' }}
+        />
+        <StyledInputRow
+          label={t('label.direction')}
+          Input={
+            <Box display="flex" width={160}>
+              <InventoryAdjustmentDirectionInput
+                value={draft.direction}
+                onChange={direction => {
```
lache-melvin, line 74: ended up with a lot of enabled/disabled/resetting logic in these inputs... wanting to validate the UX a bit but then would be keen to add some UI tests around it

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
```diff
@@ -0,0 +1,86 @@
...
+import { getSdk } from '..';
+
+type DraftInventoryAdjustment = {
+  direction: AdjustmentDirectionInput | null;
+  reason: InventoryAdjustmentReasonRowFragment | null;
+  adjustBy: number;
+  newNumberOfPacks: number;
+};
+
+export function useInventoryAdjustment(stockLine: StockLineRowFragment) {
```
lache-melvin, line 22: Also used this as a bit of a start for new API hook ideas

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
```diff
@@ -0,0 +1,85 @@
...
+  // would usually query data here
+
+  // can then manage state (buffered too?) - can expose debounced update from here
+  const [draft, setDraft] = useState<DraftInventoryAdjustment>({
+    direction: null,
+    reason: null,
+    adjustBy: 0,
+    newNumberOfPacks: 0,
```
lache-melvin, line 30: Keen for thoughts on having this initial state be 0...

When in initial state, direction is null, so newNumberOfPacks input is disabled. 

When direction is set, newNumberOfPacks is set to stockLine.totalNumberOfPacks - this is so that new total is reset if you change the direction again.

In my ideal world, the default/initial state would be `stockLine.totalNumberOfPacks` even when the input is disabled. However, we agreed that when the user hits save, the form would reset. Not sure what the best way would be to reset the `newNumberOfPacks` field?

- Could optimistically set it to the `newNumberOfPacks` that you sent in the mutation?
- Otherwise would have to use some kind of buffered state situation after the stockLine is refetched and we have the new `totalNumberOfPacks` value?

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+            />
+          }
+        />
+        <Box sx={{ display: 'flex', justifyContent: 'end', marginTop: '14px' }}>
+          <DialogButton
+            variant="save"
+            color="primary"
+            disabled={saveDisabled}
+            onClick={save}
+          />
```
lache-melvin, line 180: Thoughts on putting the save button here? Would have preferred to replace the `Ok` modal button with `Save`, but the outer modal component doesn't have access to the tab context so that was looking more complex 👀

CarlosNZ:
Haven't looked through it thoroughly yet, just had an initial thought on UI. I was thinking that the "direction" and "amount" selectors could be more interelated, like the "Direction" almost would be the "label" of the amount input (with default selection of "Reduce by" as that's the more common case)

Also, I'm not sure we need a separate input for the "New" quantity -- why not just the quantity itself be an input, and you can either change that directly, or use the "adjust" input (and they both change to reflect the value of the other)?

So it would look something like this (excuse the crude Photoshopping):
Screenshot 2024-04-05 at 10 18 48 AM

Happy to discuss in person -- I'm open to being convinced either way, I just think there's currently more "stuff" on screen than is necessary.

lache-melvin:
> I was thinking that the "direction" and "amount" selectors could be more interelated, like the "Direction" almost would be the "label" of the amount input (with default selection of "Reduce by" as that's the more common case)
> 
> Also, I'm not sure we need a separate input for the "New" quantity -- why not just the quantity itself be an input, and you can either change that directly, or use the "adjust" input (and they both change to reflect the value of the other)?
> 
> So it would look something like this (excuse the crude Photoshopping): Screenshot 2024-04-05 at 10 18 48 AM
> 

Oo yeah nice one! Do you have any input here @richardmoizeau ?

andreievg:
> Also, I'm not sure we need a separate input for the "New" quantity -- why not just the quantity itself be an input, and you can either change that directly, or use the "adjust" input (and they both change to reflect the value of the other)?

I kind of like the new number of packs (and showing current number of packs). 

I like the "reduce quantity by" drop down vs "adjustment direction"

andreievg:
In terms of behaviour, this is what I would expect (not saying I am a typical user though).

1 - New number of packs is always enterable (even before direction is selected)
2 - When direction is changed, the adjustment number of pack should remain the same and new number of packs should change
3 - When number of packs is entered direction and adjustment number of packs should change
4 - When Adjustment number of packs is changed, new number of packs should change

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
CarlosNZ: Is this data shape dictated by the mutation requirements? Because why do we need separate "direction" and "ajustBy" values? Can't we just "adjustBy" a negative value for reductions? Also, I'm not entirely sure why `newNumberOfPacks` doesn't just get reset/initialised with the current number -- perhaps I'm missing something?

CarlosNZ (approved):
Cool, few comments, maybe best to discuss some of them in person?

richardmoizeau:
> So it would look something like this (excuse the crude Photoshopping):
> Screenshot 2024-04-05 at 10 18 48 AM

I think this is making a lot of sense from a UX point of view, well done @CarlosNZ. 

Few questions, why do we need a separate "Save" button ? I also wonder if "save" is the right wording for finalising an InvAdj. 

And also don't forget it's quite important to implement "OK & Next" button at the stock line details level. If user is filtering all his expired batch, he wants to be able to perform inventory adjustments for all of them in a row (ideally, like in OG, OK & Next would keep you on the same tab in the modal)

> I kind of like the new number of packs (and showing current number of packs).

Yes definitely nice to show kind of before/after picture.

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
andreievg: Happy with either way, I guess i would only object doing it now if it slows down this PR too much

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
andreievg: hmm, now that newNumberOfPacks doesn't have to be editable, it can be deduced based on adjustment and total number of packs right ? So doesn't have to be part of the state

If it was editable, maybe we can treat undefined = totalNumberOfPacks and like in the input `value={draft.newNumberOfPacks ?? totalNumberOfPcaks}`

client/packages/system/src/Stock/api/hooks/keys.ts
```diff
@@ -0,0 +1 @@
+export const STOCK = 'stock';
```
andreievg, line 1: nice nice, although I would make it more specific. `STOCK_LINE`

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
```diff
@@ -0,0 +1,79 @@
...
+          newNumberOfPacks,
+          stockLineId,
+          inventoryAdjustmentReasonId: reason?.id,
+        },
+      });
+    },
+    {
+      onSuccess: () =>
+        // Stock line needs to be re-fetched to refresh quantity
+        queryClient.invalidateQueries([STOCK]),
```
andreievg, line 76: maybe the list would need to be refetched here, or on modal closed ? (maybe TODO to check when we implement the mutation)

server/graphql/inventory_adjustment/mutations/insert.rs
```diff
@@ -0,0 +1,44 @@
...
+            // TODO map Permissions::EnterInventoryAdjustments
+            resource: Resource::CreateRepack,
+            store_id: Some(store_id.to_string()),
+        },
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let _service_context = service_provider.context(store_id.to_string(), user.user_id)?;
+
+    // another pattern to discuss - returning data from mutation when we don't use it?
```
andreievg, line 38: Maybe we can import some bare minimum thing ?

I would suspect the main thing that needs to be returned is 'new number of packs', but we won't really use it, and we do need some kind of return, not sure, maybe we have a specific type we can easily return, similar to some of the common structure errors

andreievg (approved):
Thanks, looks pretty good. I made some suggestions, don't want to hold this up very much, so will approve, a few things that I think should be done thought (but can be follow up PR)

* Even though new number of packs will not be editable, I think we can still do the setAdjustment method I've mentioned in one of the comments
* Remove the new number of packs editing logic
* Set one of the directions as default (actually I don't really like using existing `Adjustment` type, i think the type should really be `AdjustmentDirection' and shouldn't have a `None` and shouldn't be nullable in Draft adjustment)

client/packages/system/src/Stock/api/hooks/keys.ts
andreievg: Although maybe it doesn't make it more specific, and the point of the constants is being able to navigate to references, which would explain the context any ways

client/packages/system/src/Stock/api/hooks/useInventoryAdjustment.ts
lache-melvin: I think this actually does refetch the list... STOCK_LINE is used as the base key in useStockApi() so would refetch everything 😅

lache-melvin:
Cool, thanks for feedback team 🙏 I've implemented the quick things... 

Agree with the changes around the state management logic - will implement these in separate PR, on top of backend changes so have access to changes in the API :)

PR #3545, opened 2024-04-04
3543 Fix outbound shipment not including service charge in total

### LINKED ISSUE #3543: Problems with service-charge total: it disappears on finalising the outbound
## What went wrong? 😲
 
When adding service charges to Outbound, the Grand total (including service charges) is re-adjusted to line totals 

## Expected behaviour 🤔
values shouldn't disappear like that!
And, honestly, idk where the service charges should be allowed actually! They're allowed in both Inbounds & Outbounds? 'cuz rn, I'm seeing it in both shipments

## How to Reproduce 🔨
1. Create a outbound with some line. Better to have line totals
2. More> Add service charges - some values so that `Grand total`>`Line total`
3. Confirm the shipment to Shipped status
4. See error: the grand-total is adjusted to line-total: voila the service charges disappears!

## Your environment 🌱

- Open mSupply Version: v1.7-rc5 exe
- Legacy mSupply Central Server Version: v714-beta3

- Platform: browser(chrome), server exe

- Database type: sqlite

## Additinal
Found while [v1.7-release testing](https://github.com/msupply-foundation/open-msupply/issues/3341)

raviSussol:
Fixes #3543

# 👩🏻‍💻 What does this PR do? 
 
Include service charge amount in the grand total calculation for the outbound shipment.

1. Create a outbound with some line. Better to have line totals
2. More> Add service charges - some values so that `Grand total`>`Line total`
3. Confirm the shipment to Shipped status
4. The grand-total is adjusted to line-total

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/server/service/src/invoice/outbound_shipment/update/generate.rs
+++ b/server/service/src/invoice/outbound_shipment/update/generate.rs
@@ -146,7 +146,7 @@ fn lines_to_trim(
     }
 
     // If new invoice status is not new and previous invoice status is new
-    // add all unallocated lines to be deleted
+    // add all unallocated lines or empty lines to be deleted
 
     let mut lines = InvoiceLineRepository::new(connection).query_by_filter(
         InvoiceLineFilter::new()
@@ -157,7 +157,8 @@ fn lines_to_trim(
     let mut empty_lines = InvoiceLineRepository::new(connection).query_by_filter(
         InvoiceLineFilter::new()
             .invoice_id(EqualFilter::equal_to(&invoice.id))
-            .number_of_packs(EqualFilter::equal_to_f64(0.0)),
+            .number_of_packs(EqualFilter::equal_to_f64(0.0))
+            .r#type(InvoiceLineRowType::StockOut.equal_to()),
     )?;
 
     if lines.is_empty() && empty_lines.is_empty() {


server/service/src/invoice/outbound_shipment/update/generate.rs
```diff
@@ -157,7 +157,8 @@ fn lines_to_trim(
     let mut empty_lines = InvoiceLineRepository::new(connection).query_by_filter(
         InvoiceLineFilter::new()
             .invoice_id(EqualFilter::equal_to(&invoice.id))
-            .number_of_packs(EqualFilter::equal_to_f64(0.0)),
+            .number_of_packs(EqualFilter::equal_to_f64(0.0))
+            .r#type(InvoiceLineRowType::StockOut.equal_to()),
```
raviSussol, line 161: This would fix not including service charge calculation in the grand total but would appear as dummy line in the ci of the msupply cause of pack size, no. of packs with other price fields are being 0 (of the service charge item) in the oms. Does that mean the service charge feature only need to be in oms (not in the msupply)? What is the reason behind mapping service charge (in oms) to the service item (in the msupply) though?

In the msupply, service item has default pack size to 1 and it uses default sell price in the ci. Above all what is the expected behavior for service charge both in the oms (created from) and the ms (received after sync)?

Sorry I may not have the proper context to do any further changes apart from the current minimal change.

Chris-Petty: I agree with all your questions and conclusions, so I am happy with this fix for now!

Chris-Petty (approved):

Chris-Petty:
@raviSussol would you be happy to open the issue with your points and label with "needs architecture" ?

raviSussol:
Did here #3554

PR #3549, opened 2024-04-04
Fix two warnings by renaming getPackVariantCell

clemens-msupply:
Some warning I noticed while working on another issue. Fixes two warning by using the correct naming convention:

React Hook "usePackVariant" cannot be called inside a callback.

A function with a name starting with an uppercase letter should only be
used as a constructor.


roxy-dao (approved):
Thanks Clemens~

Issue #3498, opened 2024-04-01
Adjustment back end mutation


andreievg:
New mutation to create an inventory adjustment (To adjust existing stock quantities)

Should we use stock_in/stock_out abstractions, or just edit quantities on the stock line directly.

>  Stock out sounds like the right idea for a stock decrease, but may or may not be able to use stock in - stock in always creates a new stock line for introduced stock, but we just want to increase the amount of existing stock?

PR #3557, opened 2024-04-07
3498 pt 1 inventory adjustment permissions

lache-melvin:
Part of #3498

# 👩🏻‍💻 What does this PR do? 
 

Adds permissions mappings for inventory adjustments.

- Disable the `Enter Inventory Adjustments`, `Edit Inventory Adjustments` and `Finalise Inventory Adjustments` permissions in mSupply for your omSupply user
- Log out and log in again
- Try to do an inventory adjustment via the stock view modal
- [ ] You should get a permission denied error
- Enable one of the `Enter Inventory Adjustments` permission
-  Log out and log in again
- Try to do an inventory adjustment via the stock view modal
- [ ] It should not error 🎉 

Query permissions not yet used anywhere to test from UI...

There are 3 IA mutation related permissions in OG - `Enter Inventory Adjustments`, `Edit Inventory Adjustments` and `Finalise Inventory Adjustments`.

However, my understanding with omSupply inv adjustments is that we'll finalise them immediately - i.e. create only. I've mapped all of those 3 permissions to the one OMS `MutateInventoryAdjustment` permission - is that correct?

## 📃 Documentation

_No user facing changes_


andreievg (approved):
Looks good, thank you, not sure about the query one, maybe we don't really need it atm ? (for ledger, but ledger should be a generally viewable, i.e. ledger should balance out, and if we remove some transactions/invoices from ledger cause they are not editable, ledger will be make sense), hmm maybe a ledger permission check will include the IA query ¯\_(ツ)_/¯)

> There are 3 IA mutation related permissions in OG - Enter Inventory Adjustments, Edit Inventory Adjustments and Finalise Inventory Adjustments.

> However, my understanding with omSupply inv adjustments is that we'll finalise them immediately - i.e. create only. I've mapped all of those 3 permissions to the one OMS MutateInventoryAdjustment permission - is that correct?

I think there is a permission process in some countries for finalisation. We haven't quite considered this workflow, which will suite the 'repack' view a bit more, maybe we can pivot that way in the future if the adjustment finalisation needs to be authorised (and stocktake just won't do)

PR #3564, opened 2024-04-08
502 fuzzy search for patient

### LINKED ISSUE #502: Fuzzy patient search
Return patients:
- with similar names
- with similar name and similar DoB ?

Idea:
- Narrow down server search result by looking for patient that match the first two chars of the first/last name.
- Use fuzzy search lib to further reduce the search results

raviSussol:
Fixes #502 

# 👩🏻‍💻 What does this PR do? 
 

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2


clemens-msupply:
Thanks! I have some comments:

- Since we are not making use of DB search functionality the fuzzy search should be in a service, reusing the existing PatientRepository. (The repository should be a minimal abstraction over the DB with no/minimal logic in it)

- The PatientSearchNode already has a score field which should be populated as returned from the fuzzy-matcher.

- Return all (or only the top 10?) matches not only the first one, e.g. the matching patient for a mispelled name could be lower in the list

andreievg:
@raviSussol can you please add some test cases to the repository ? Would really need examples to try and understand how this fuzzy search works and to be able to communicate the functionality to the users.

Also @clemens-msupply @raviSussol what are the performance implications of this, considering a country with say 10million patience (since we are not paginating in the first query ?)

clemens-msupply:
> Also @clemens-msupply @raviSussol what are the performance implications of this, considering a country with say 10million patients (since we are not paginating in the first query ?). And fuzzy search on central server (omSupply, when it has all of the patient records)

Probably need to test on such a datafile with the worst case scenario. Just a very simple and rough estimate: Randomly assuming 10 out of 26 char most commonly used for the first two letters of first and last name, this would narrow it down to:
10_000_000  / 10 / 10 / 10 / 10  == 1000 patients
Fetching an average of 20 bytes for first and last name from the db gives:
1000 * 20byte = 20kb of data to be matched.
Could require dob or year of birth to narrow it down further, e.g. by factor 20-50(?)

One assumption of the issue was that we need to be able to easily implement the algo in 4D as well. However, if we have a omSupply central server we could use postgres/sqlite extension to do the matching...

raviSussol:
@andreievg 
> can you please add some test cases to the repository ? Would really need examples to try and understand how this fuzzy search works and to be able to communicate the functionality to the users.

Oh I was supposed to update as per @clemens-msupply https://github.com/msupply-foundation/open-msupply/pull/3564#issuecomment-2046161127. So I'll update the test cases after that.

> One assumption of the issue was that we need to be able to easily implement the algo in 4D as well.

Yes this is still going to be implemented in the 4d part too - (probably in the upcoming version though?).

andreievg:
Closing for now, until this work is planned for a milestone, we can reference this PR when the times comes. Thanks @raviSussol for research work

PR #3570, opened 2024-04-09
Show CCE in sensor list view

### LINKED ISSUE #3462: Add CCE column data to sensor list
## What went wrong? 😲
CCE column data not visible in sensor list

image

## Expected behaviour 🤔
Add CE column data to sensor list
This should be lookup based on location

mark-prins:
Fixes #3462 

# 👩🏻‍💻 What does this PR do? 
 
Adds a loader for assets by location, so that I can extend the sensor query and show locations.
Have used that in order to display the `assetNumber` of an asset if there is one linked via location to a sensor.

image

Have also shown the error if you try to save an asset and an error is returned by the API. in this case, if you assign a location which is already allocated to an asset, there's a problem.

To help users with that, I've added a new query which will only show available locations, preventing the problem.

Assign a location to a sensor (import fridge tag if required, to get sensors in the list)
Assign the same location to an asset, and review the sensor list.

## 📃 Documentation

screenshots to be updated. some text would be helpful


server/graphql/location/src/lib.rs
```diff
@@ -59,6 +62,60 @@ impl LocationQueries {
             locations,
         )))
     }
+
+    /// Returns a list of locations which are available to be assigned to assets
+    pub async fn asset_locations(
```
jmbrunskill, line 67:
```suggestion
    pub async fn unassigned_asset_locations(
```

Not sure what the best wording is but asset_locations sounds like it would return the assets assigned to an asset, not the ones that are not assigned.

Not a big deal but I'd make it more specific...

jmbrunskill: I've created a suggestion branch, where we can just pass a boolean into the Location Filter, instead of this specific query. 
Might be a nice way to go and more flexible for future use cases? https://github.com/msupply-foundation/open-msupply/compare/3462-add-cce-column...3462-suggestion_subquery

jmbrunskill (approved):
Looks good Mark.
Have made a suggestion, but otherwise looks good.

The first time I tried to do the location, asset,sensor mapping the Sensor view didn't update with the CCE data.

I haven't been able to replicate this though so not sure if it was an order of operations thing, or if I was just mistaken? Either way not a big deal.

client/packages/common/src/types/schema.ts
```diff
@@ -7022,6 +7033,7 @@ export enum UserPermission {
   InboundReturnQuery = 'INBOUND_RETURN_QUERY',
   InboundShipmentMutate = 'INBOUND_SHIPMENT_MUTATE',
   InboundShipmentQuery = 'INBOUND_SHIPMENT_QUERY',
+  InventoryAdjustmentMutate = 'INVENTORY_ADJUSTMENT_MUTATE',
```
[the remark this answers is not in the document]
mark-prins: yep, looks like it was added to api but `yarn generate` hadn't been run

mark-prins:
that's awesome - thanks @jmbrunskill!
I've taken the suggestion branch and removed the new location query. that's a much nicer approach

PR #3572, opened 2024-04-09
#3498 Inventory adjustment back-end

### LINKED ISSUE #3498: Adjustment back end mutation

CarlosNZ:
Fixes #3498

# 👩🏻‍💻 What does this PR do? 
 

Pair-programmed with @lache-melvin. 

There's 2 tests not passing (currently commented out) -- see inline comments.

Run tests in `server/service/src/invoice/inventory_adjustment/insert.rs`


server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,245 @@
...
+        // assert_eq!(
+        //     service.insert_inventory_adjustment(
+        //         &context,
+        //         InsertInventoryAdjustment {
+        //             stock_line_id: mock_stock_line_a().id,
+        //             ..Default::default()
+        //         }
+        //     ),
+        //     Err(ServiceError::AdjustmentReasonNotProvided)
+        // );
```
CarlosNZ, line 162: There seems to be a problem with `check_active_adjustment_reasons` that is throwing an error causing this test to fail.

server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,245 @@
...
+        //             adjustment_type:
+        //                 crate::invoice::inventory_adjustment::AdjustmentType::Reduction,
+        //             adjustment: 50.0,
+        //             ..Default::default()
+        //         }
+        //     ),
+        //     Err(ServiceError::StockLineReducedBelowZero(StockLine {
+        //         stock_line_row: mock_stock_line_a(),
+        //         ..Default::default()
+        //     }))
```
CarlosNZ, line 205: The problem I have here is that the (pre-existing) StockLineReducedBelowZero Error requires a full `StockLine`, but we only have a `StockLineRow` available. At first I tried changing the enum definition (at line 36) to take a "StockLineRow" parameter, but that causes problems with the GraphQL error (in graphql insert, line 102) which requires a  full `StockLine` in its `from_domain()` method. I was a bit wary of changing *that* as well, as the cascade other other changes resulting from that was getting a bit messy. Any tips?

server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,245 @@
...
+            .unwrap();
+
+        assert_eq!(
+            retrieved_invoice,
+            inline_edit(&retrieved_invoice, |mut u| {
+                u.id = created_invoice.invoice_row.id;
+                u.status = InvoiceRowStatus::Verified;
+                u
+            })
+        );
```
CarlosNZ, line 243: Should we also retrieve the new invoice line and check it? 

What other types of "Success" tests should we have?

lache-melvin: Yeah could check the invoice line looks as expected. Probably more important though would be to query the stock line, and check the stock levels have been adjusted accordingly.

So I would write 2 success tests, one for addition and one for reduction 👍

server/service/src/invoice/inventory_adjustment/insert.rs
lache-melvin: Ah yep, I think the easiest way would be to use the StockLineRepository to query for the full stock line, so you can pass it in here. There is an example in stocktake_line: https://github.com/msupply-foundation/open-msupply/blob/develop/server/service/src/stocktake_line/insert.rs#L575

server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,245 @@
...
+        assert_eq!(
+            service.insert_inventory_adjustment(
+                &context,
+                InsertInventoryAdjustment {
+                    stock_line_id: mock_stock_line_a().id,
+                    inventory_adjustment_reason_id: Some("invalid".to_string()),
+                    ..Default::default()
+                }
+            ),
+            Err(ServiceError::InvalidAdjustment)
```
lache-melvin, line 174: Should be `AdjustmentReasonNotValid`... would need to provide adjustment & adjustment_type for it to get past the `input.adjustment == 0.0` check though!

server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,245 @@
...
+        );
+
+        // Reduce stock below zero
```
lache-melvin, line 177:
```suggestion
        // Adjust by zero
```

server/service/src/invoice/inventory_adjustment/insert.rs
lache-melvin: I'm not seeing an error thrown? Just that this check is being skipped - which is expected behaviour (no reasons are configured in the mock data by default, and if no reasons configured, we don't expect a reason id to be provided.) Here's how I would set up for this test: https://github.com/msupply-foundation/open-msupply/compare/3498-adjustment-back-end...3498-adjustment-back-end(test-inventory-adjustment-reason-not-provided)

server/service/src/invoice/inventory_adjustment/insert.rs
lache-melvin: Probably also good to have a quick check in one of the success tests that if no reasons configured, the reason not provided error _doesn't_ throw

server/service/src/invoice/inventory_adjustment/insert.rs
CarlosNZ: Thanks! 3821e3d21

lache-melvin:
Oh yeah - posted this in Ruru channel, but probably better here - the change to use stock_in/stock_out: https://github.com/msupply-foundation/open-msupply/compare/3498-adjustment-back-end...3498-adjustment-back-end(use-stock-in-stock-out-abstractions)

It actually looks like it ends up as slightly MORE code, but I do think it's better to have the concern of creating invoice lines/introducing stock all contained in one place. Will make any logic changes easier in future. 

Unsure about the checks I've added for inventory adjustment types though - they work for the inventory adjustments we're doing here, but I think the invoice statuses might not be right if we were to use this for stocktakes?

server/service/src/invoice/inventory_adjustment/generate.rs
```diff
@@ -0,0 +1,120 @@
...
+            AdjustmentType::Reduction => NumberRowType::InventoryReduction,
+        },
+        store_id,
+    )?;
+
+    let invoice = InvoiceRow {
+        id: uuid(),
+        user_id: Some(user_id.to_string()),
+        name_link_id: inventory_adjustment_name.id,
+        r#type: InvoiceRowType::OutboundReturn,
```
lache-melvin, line 45: Whoops missed this one, should be InventoryAddition or InventoryReduction!

server/service/src/invoice/inventory_adjustment/insert.rs
```diff
@@ -0,0 +1,352 @@
...
+                }
+            ),
+            Err(ServiceError::StockLineReducedBelowZero(stock_line))
+        );
+    }
+
+    #[actix_rt::test]
+    async fn insert_inventory_adjustment_success() {
+        let (_, connection, connection_manager, _) =
+            setup_all("insert_inventory_adjustment_errors", MockDataInserts::all()).await;
```
lache-melvin, line 253:
```suggestion
            setup_all("insert_inventory_adjustment_success", MockDataInserts::all()).await;
```

server/service/src/invoice/inventory_adjustment/generate.rs
CarlosNZ: Eek -- okay, done: ac2223464

andreievg:
> Hm ok merged the tests into that branch and actually the stock in abstraction would require further changes - currently stock_in is only used to introduce stock, so it sets the total and available quantity to whatever the invoice number_of_packs is. Starting to feel like a lot of forked logic for something that should be "shared functionality"... maybe at this point best to keep as is?

Hmm, sure, but would like to consolidate two adjustment (in stocktake and here), maybe during the two week of testing/refactor, btw made a little suggestion here: https://github.com/msupply-foundation/open-msupply/compare/3498-adjustment-back-end(use-stock-in-stock-out-abstractions)...3498-adjustment-back-end(use-stock-in-stock-out-abstractions)(Use-enum-for-stock-in-and-stock-out-line)

andreievg (approved):
Thanks @CarlosNZ @lache-melvin, i just did a browse through, looks good. I ran tests, they pass

PR #3573, opened 2024-04-09
Hide make/model for non-catalogue assets

### LINKED ISSUE #3460: CCE module make/model not showing on detail view
## What went wrong? 😲
CCE module make/model is not showing on detail view, when this is a non-catalogue item

## Expected behaviour 🤔
Don't show the make & model labels in this case

## How to Reproduce 🔨

1. Go to cold chain
2. View detail of a non-catalogue item

mark-prins:
Fixes #3460

# 👩🏻‍💻 What does this PR do? 
 
What it says on the tin. Was thinking that when the sub-catalogue changes are merged we could add the catalogue to that toolbar area. For now though, this tidies up the issue.

Shouldn't show make/model for non-catalogue assets.

### DIFF
--- a/client/packages/coldchain/src/Equipment/DetailView/Toolbar.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/Toolbar.tsx
@@ -16,30 +16,32 @@ export const Toolbar: FC = () => {
 
   return (
     <AppBarContentPortal sx={{ display: 'flex', flex: 1, marginBottom: 1 }}>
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
I wonder if we want to record a make and model for these non-catalogue assets eventually? I agree this is good enough for now.

mark-prins:
yeah, we probably should. not now though, let's see if it is requested first

PR #3577, opened 2024-04-10
Fix parsing pref records

### LINKED ISSUE #3575: Parsing legacy pref object fails
This is because some pref objects have none of the preferences set. Solution: default to false.

clemens-msupply:
Fixes #3575 

Fix parsing a pref record. And then decide if the pref record should be integrated or not.

pref entry in the sync_buffer table should look like this now (instead of errors): 

image


roxy-dao (approved):
Thanks Clemens

PR #3582, opened 2024-04-10
3500 add new stock line UI

### LINKED ISSUE #3500: Add new stock line button, new modal with the stock line form, save (editable quantity), reasons
Add ability to add a new stock line via the stock page (via a positive inventory adjustment)

- `New stock line` button in the top right
- Opens modal with item selector
- Same form UI as stock view modal, but also with quantity editable
- Include reason (positive inventory adjustment reasons)

Time allowing? Or separate issue? 
- Ability to search existing stock lines for duplicates and for prices matches
  - Auto complete on batch field
  - Or when focus moves away from batch field? Search by batch, then if there are duplicates, show a warning. Link to stock list view, filtered by that batch number, so user could open the existing stock line and increase that stock value instead
  - For duplicates, probably filter only stock lines with quantity? For pricing... maybe most used pricing for that item??

lache-melvin:
Fixes #3500

# 👩🏻‍💻 What does this PR do? 
 

Adds a `New Stock` button to the Stock view page, and a modal to enter new stock line info:
Screenshot 2024-04-10 at 3 14 12 PM

- Go to Catalogue > `View Stock`
- [ ] Click on `New Stock`
- [ ] Modal opens - select an item
- [ ] Fill out stock info
- [ ] Save the form - using mock API, nothing will save yet :)

The issue also included potentially adding the ability to search existing stock lines for duplicates and for prices matches, but I think this would be best done by a PR on top of this one?

## 📃 Documentation

- [ ] The `new stock` modal
  - [ ] Probably also covering off when to use this feature to introduce stock??


client/packages/system/src/Item/Components/PackVariant/PackVariantInput.tsx
```diff
@@ -0,0 +1,122 @@
...
+  useTranslation,
+} from '@openmsupply-client/common';
+import { usePackVariant } from '../../context';
+
+const ENTER_PACK_SIZE = -1;
+
+// This field displays a packSize number input and unit pack drop down if unit
+// pack variants exist
+
+export const PackVariantInput = ({
```
lache-melvin, line 16: This is very similar to the PackVariantEntryCell, just without the table elements. I think it would be best to refactor PackVariantEntryCell to use this component underneath? But it was a little tangled, so thought cleaner just to add here, refactor separately...

client/packages/system/src/Stock/Components/StockLineForm.tsx
```diff
@@ -99,16 +98,23 @@ export const StockLineForm: FC<StockLineFormProps> = ({
...
       >
-        <TextWithLabelRow
+        <StyledInputRow
           label={t('label.num-packs')}
-          text={String(draft.totalNumberOfPacks)}
-          textProps={{ textAlign: 'end' }}
+          Input={
+            <NumericTextInput
+              autoFocus
+              disabled={!packEditable}
```
lache-melvin, line 106: Disabling these inputs when not editable... changes the stock line edit UI to look like this:

Screenshot 2024-04-10 at 3 27 59 PM

Instead of before it was like this:

Screenshot 2024-04-10 at 3 28 09 PM

I don't think it's a biggie... though also aware of making any changes that should be hidden in 2.0.0 👀

client/packages/system/src/Stock/api/hooks/useStockLine.ts
```diff
@@ -0,0 +1,92 @@
...
+    costPricePerPack: 0,
+    totalNumberOfPacks: 0,
+    availableNumberOfPacks: 0,
+    storeId: '',
+    item: {
+      __typename: 'ItemNode',
+      code: '',
+      name: '',
+    },
+    inventoryAdjustmentReason: null,
```
lache-melvin, line 35: Keen for the thoughts here...

From the perspective of our new API hook design, I liked the idea of managing input state from the same hook that calls the mutation. However, if the UI needs a slightly different shape than the API, where would we do the mapping?

In this case, the draft state shape is a StockLineRowFragment (which has some extra stuff than needed for this API call, but that's the type needed for the stock edit form) plus the whole inventoryAdjustmentReason (we have to manage the whole thing in state for the `InventoryAdjustmentReasonSearchInput`… would be nicer if we only had to manage the id, but changes would also affect stocktake so I am refraining at this point lol)

ANYWAY this state should probably be closer to the components right? I think at that point, managing/exposing state from this hook wouldn't have a whole lot of value?

client/packages/system/src/Stock/Components/StockLineForm.tsx
```diff
@@ -99,16 +98,23 @@ export const StockLineForm: FC<StockLineFormProps> = ({
...
+              width={160}
```
CarlosNZ, line 107: Should you use your constant `INPUT_WIDTH` here?

client/packages/system/src/Stock/Components/StyledInputRow.tsx
```diff
@@ -13,7 +13,7 @@ export const StyledInputRow = ({ label, Input }: InputWithLabelRowProps) => (
     sx={{
       justifyContent: 'space-between',
       '.MuiFormControl-root > .MuiInput-root, > input': {
-        width: '160px',
+        maxWidth: '160px',
```
CarlosNZ, line 16: And `INPUT_WIDTH` here?

client/packages/system/src/Stock/api/hooks/useStockLine.ts
CarlosNZ: I think it's fine how you've done it here. It would only be an issue if there was multiple usages of the same hook that required different state structures. But that isn't really the case in most situations, is it? In the event we find one, we can figure out the best way to handle it based on the specific case, perhaps? Can you think of any?

client/packages/system/src/Item/Components/PackVariant/PackVariantInput.tsx
andreievg: I see what you mean, with styles especially, right ?

client/packages/system/src/Stock/api/hooks/useStockLine.ts
```diff
@@ -0,0 +1,92 @@
...
+  }, [draft, createMutation]);
+
+  return {
+    draft,
+    setDraft,
+    create,
+  };
+}
+
+const useCreate = () => {
```
CarlosNZ, line 51: I'd be inclined to just keep this inside the main hook above -- any particular reason to split it into its own hook? Just personal preference?

(It would save having to use `useCallback` too, I think -- I'm, assuming `useMutation` has some memo-isation mechanisms internally)

CarlosNZ (commented):
Looks good. Few questions/suggestions -- feel free to discuss in person if that's easier or I'm being too vague. :)

client/packages/system/src/Stock/Components/StockLineForm.tsx
```diff
@@ -157,10 +163,17 @@ export const StockLineForm: FC<StockLineFormProps> = ({
         flexDirection="column"
         gap={1}
       >
-        <TextWithLabelRow
+        <StyledInputRow
           label={t('label.pack')}
-          text={String(packUnit)}
-          textProps={{ textAlign: 'end' }}
+          Input={
```
andreievg, line 168: Kind of like the idea of conditional component here

client/packages/system/src/Stock/ListView/AppBarButtons.tsx
```diff
@@ -9,18 +9,24 @@ import {
...
+  useEditModal,
 } from '@openmsupply-client/common';
 import { useStock } from '../api';
 import { stockLinesToCsv } from '../../utils';
+import { NewStockLineModal } from '../Components/NewStockLineModal';
 
 export const AppBarButtonsComponent = () => {
   const { success, error } = useNotification();
-  const t = useTranslation('distribution');
+  const t = useTranslation('inventory');
```
andreievg, line 22: Don't we pass t to stockLinesToCsv ?

client/packages/system/src/Stock/api/hooks/useStockLine.ts
andreievg: i was making a similar comment earlier, and tried to change as example, it ended up looking slightly worse. I think it was mainly due to storeId, and then onSuccess.

client/packages/system/src/Stock/api/hooks/useStockLine.ts
andreievg: A agree I like it, although I think we could remove things from the common type (like storeId, itemId etc..)

andreievg (commented):
Looks good overall, was wondering about the translation mismatch, haven't checked but looks like it could cause an issue

client/packages/system/src/Item/Components/PackVariant/PackVariantInput.tsx
lache-melvin: Yeah mostly styles - also use of column methods, though I imagine these could be passed in as props fairly easily...

client/packages/system/src/Stock/api/hooks/useStockLine.ts
lache-melvin: Main motivation was that if the main hook starts exposing more methods, the main API is easy to see at the top of the file. Though probably don't really need the useCallback now you mention it...

client/packages/system/src/Stock/ListView/AppBarButtons.tsx
lache-melvin: We do, but all the translations used by stockLinesToCsv are actually from the `common` translations, so the switch is fine... really would be awesome if there was a way to type these!

andreievg: Yeah i made a comment about it a few times, i actually not certain that having different files helps us

andreievg (approved):
Looks good to me now that you've clarified the localisation change

CarlosNZ (approved):
LGTM --- thanks :)

PR #3586, opened 2024-04-10
Store filter and status log entry on create

### LINKED ISSUE #3558: CCEI misc issues - from review feedback
## What went wrong? 😲
 
A couple of items noted in a review meeting:
- ~The exported csv file has the id populated~
- ~In the exported csv file, the code column has the id rather than the string value~
- ~add new asset item which is non catalogue and it doesn’t show immediately in the list~
- ~Should assets should be assigned to names not stores?~
- [x] Filter by store in the central server equipment list
- [x] Add a status log on create / import with Functioning::observation is ‘created..’
- ~Adding multiple locations fails with a duplicate id error~
- ~If you add a location that exists there's a server error - shouldn't it fail quietly?~

This one may be fixed (when adding multiple locations) please retest:
  "\"duplicate key value violates unique constraint \\\"asset_internal_location_location_id_key\\\"\"",

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #3558

# 👩🏻‍💻 What does this PR do? 
 
Most items on the original issue are already fixed, so that leaves us:

- create a status log entry when creating an asset, either manually or via the import
  - the status should be `Functional`
  - a comment should say `Created`
- add a store filter to the list view, if running on central server

In addition, I noticed that the import is incorrectly failing to match codes due to the default pagination limit of 100 records. Since we only have 159 records I've bumped the limit to 1000 which should work for a while.

I wanted to have the text `Created` be translatable, so have used an existing entry. That required correcting the key, which was 'label.entered' as I thought it might be misleading later. It has added to the file count, sorry about that.

* Create an asset: the asset should have status 'functional' with one status log that has a comment of 'created'
* Import assets: check for the status logs
* Test the store filter which 
  * shows on central server 
  * does not show on remote ( though the filter option is available and you can manually enter the store via the querystring )


client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -133,29 +135,36 @@ export const EquipmentImportModal: FC<EquipmentImportModalProps> = ({
...
+    }
+  };
 
   const importAction = async () => {
     onChangeTab(Tabs.Import);
     const numberImportRecords = bufferedEquipment?.length ?? 0;
     if (bufferedEquipment && numberImportRecords > 0) {
-      const importErrorRows: ImportRow[] = [];
+      importErrorRows.length = 0;
       // Import count can be quite large, we break this into blocks of 100 to avoid too much concurency
```
jmbrunskill, line 163: BTW I think this 100 comes from Health Supply Hub, we put a high number in here to make it faster, but I think have less outstanding concurrent requests is probably better?

The main performance issue we had (which I think we have here too) is that after each promise resolves React re-renders and re-queries the List View in the background.

There's an issue to fix this in HSH here...
https://github.com/msupply-foundation/health-supply-hub/issues/993

Here's an example how we used invalidateQueries after the import finishes in Universal Codes...

https://github.com/msupply-foundation/unified-codes/blob/59ff94c390aa55468a26d10ae2167bbfb5d2f849/frontend/common/src/ui/components/toolbars/SearchAndDeleteToolbar.tsx#L68

jmbrunskill (approved):
Looks good to me.
I thought we had discussed allowing the import to specify a status, but I think this a good starting point anyway.

Updating status via import is probably more important if we allow updating which isn't available via import right now.

client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
mark-prins: cool - thanks James. Have refactored to only invalidate after all inserts are done

PR #3588, opened 2024-04-11
#3497 Stockline Ledger tab

### LINKED ISSUE #3497: Ledger tab in stock view modal
New tab in the stock modal - Ledger

Table view of all changes to the stock quantity

(Shipments, stocktakes, inventory adjustments...) - these should all map to a consistent shape to display though :)

CarlosNZ:
Fixes #3497

# 👩🏻‍💻 What does this PR do? 
 

This is super-crude and still needs the real query, but I was unsure about a few things here and thought we could discuss.

In particular:
- I assume we'd want to paginate this, since there could be a lot of entries?
- Do we agree on these display columns?
- Do we agree on these values returned from query (`LedgerLine` type)

I'll leave this in Draft until we have the real back-end query and then I'll come back to this.

View "Ledger" tab in an inventory item modal. Change the `exampleResult` (temporary) in `useStockLedger` to see different outputs.


client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -0,0 +1,73 @@
...
+      key: 'reason',
+      label: 'label.reason',
+    },
+  ]);
+
+  return { columns };
+};
+
+const getLocalisationKey = (type: MovementType) => {
+  switch (type) {
```
andreievg, line 57: Would be great to add exhaustive match here:

https://github.com/msupply-foundation/open-msupply/blob/c9fa2d2a3b79d90cd8cd57df91b62bbf43e7fb50/client/packages/host/src/components/Sync/Sync.tsx#L89-L90

Btw even without the API implemented you can use [InvoiceNodeType](https://github.com/msupply-foundation/open-msupply/blob/c9fa2d2a3b79d90cd8cd57df91b62bbf43e7fb50/client/packages/common/src/types/schema.ts#L2726-L2735), also sounds like we need repack type too

client/packages/system/src/Stock/api/hooks/keys.ts
```diff
@@ -1 +1,2 @@
 export const STOCK_LINE = 'stock_line';
+export const LEDGER = 'ledger';
```
andreievg, line 2: Any reason to separate those keys ? We would have to remember to invalidate both ?

client/packages/system/src/Stock/Components/Ledger/LedgerForm.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { data, isLoading, isError } = useStockLedger(stockLine);
+  const { columns } = useLedgerColumns();
+
+  return (
+    <Box display="flex" sx={{ maxHeight: 300, overflowY: 'auto' }}>
+      <TableProvider createStore={createTableStore}>
+        <DataTable
+          id="stockline-ledger"
+          columns={columns}
+          data={data as any}
```
andreievg, line 28: I guess any is temporary

andreievg (commented):
Looks pretty good so far, thanks

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -0,0 +1,73 @@
...
+  const columns = useColumns<LedgerLine>([
+    {
+      key: 'datetime',
+      label: 'label.date',
+      format: ColumnFormat.Date,
+    },
+    {
+      key: 'time',
+      label: 'label.time',
+      accessor: ({ rowData }) => localisedTime(rowData.datetime),
```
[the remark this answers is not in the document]
CarlosNZ: Just copying how it's done elsewhere (Repack log for example)

CarlosNZ:
Okay, have updated:

- re-based off #3639, so uses the real back-end query and types
- table sorting
- layout improvements (see inline comments)

client/packages/common/src/intl/locales/en/app.json
```diff
@@ -111,7 +115,8 @@
   "dispensary": "Dispensary",
   "programs": "Programs",
   "contact-trace": "Contact Tracing",
-  "prescription": "Prescriptions",
+  "prescription": "Prescription",
+  "prescriptions": "Prescriptions",
```
CarlosNZ, line 119: Since we need both singular and plural forms for some of these, I've made the key and value consistent with each other, plural wise.

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -0,0 +1,108 @@
...
+  | 'INVENTORY_REDUCTION';
+
+export enum ColumnKey {
+  'DateTime' = 'datetime',
+  'Time' = 'time',
+  'Name' = 'name',
+  'Quantity' = 'quantity',
+  'Type' = 'type',
+  'Reason' = 'reason',
+}
```
CarlosNZ, line 19: Since I've used these in `useStockLedger` for mapping column keys to GraphQL sort keys, I thought I'd make it an enum to ensure legal values are enforced.

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
CarlosNZ: Done. I haven't included "REPACK" in the back-end View query -- should I?

client/packages/system/src/Stock/Components/StockLineEditModal.tsx
```diff
@@ -117,8 +126,7 @@ export const StockLineEditModal: FC<StockLineEditModalProps> = ({
 
   return (
     <Modal
-      width={700}
-      height={575}
+      sx={{ maxWidth: 'unset', minWidth: 700, minHeight: 575 }}
```
CarlosNZ, line 129: This makes the modal more flexible for the different requirements of the different tabs.

client/packages/system/src/Stock/api/hooks/keys.ts
CarlosNZ: Changed it

client/packages/system/src/Stock/Components/Ledger/LedgerForm.tsx
CarlosNZ: Gone :)

client/packages/system/src/Stock/Components/StockLineEditModal.tsx
lache-melvin: The jumpy modal feels a bit weird to me, but I think I remember you saying you played around with widening it for all tabs and that messed with the feng shui of the form tabs... and I've seen at least one other modal that changes size on different tabs so I guess all good? at least until we get some more authoritative UX input

client/packages/system/src/Stock/Components/Ledger/LedgerForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+    key: ColumnKey.DateTime,
+    direction: 'desc',
+  });
+  const { data, isLoading, isError } = useStockLedger(stockLine, sortBy);
+  const { columns } = useLedgerColumns(sortBy, (key, direction) => {
+    setSortBy({ key, direction });
+  });
+
+  return (
+    <Box display="flex" sx={{ overflowY: 'auto' }}>
```
lache-melvin, line 30: Table isn't scrolling, does it need max height? ... I think we should always be able to see bottom modal buttons?
Screenshot 2024-04-19 at 3 20 07 PM

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const useLedgerColumns = (
+  sortBy: SortBy<LedgerRowFragment>,
+  updateSort: (sort: string, dir: 'asc' | 'desc') => void
+) => {
+  const t = useTranslation('app');
```
lache-melvin, line 25: How come using `app` and not `inventory`?

client/packages/system/src/Stock/Components/StockLineEditModal.tsx
CarlosNZ: Yeah, I don't love it, but I found it preferable to having one-size-fits-all which looks decidedly weird in some tabs. Need some UI design at some point I think.

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
CarlosNZ: Okay, have added to back-end PR now

client/packages/system/src/Stock/Components/Ledger/LedgerForm.tsx
CarlosNZ: Yes: cbd8b0f14

lache-melvin (changes requested):
Looks great for SQLite, now getting this error with Postgres 😅 :

Screenshot 2024-04-30 at 4 40 08 PM

CarlosNZ:
> Looks great for SQLite, now getting this error with Postgres 😅 :
> 
> Screenshot 2024-04-30 at 4 40 08 PM

Okay, good catch. Turns out it was a problem with the "Sort" extensions -- using the `NOCASE` collation on fields that don't support it. I've fixed it in the [back-end PR](#3572) and merged back into here: d15410825

lache-melvin (approved):
Awesome thanks Carl! Great find 😅 

As mentioned in person, probably good to fix this warning:

Screenshot 2024-05-02 at 11 49 27 AM

Merge when ready!

PR #3593, opened 2024-04-11
Add infinite scroll to catalogue items autocomplete

### LINKED ISSUE #3574: Create CCE modal only shows 20 catalogue items
## What went wrong? 😲
 
The pagination is limiting the amount of results shown in the Catalogue item DD and the filtering isn't connected, meaning that if you type in the autocomplete to filter the options, it only filters within the 20 options shown.

## Expected behaviour 🤔
* Implement an infinite scroll for the autocomplete, so that more options are fetched when you reach the end of the list
* Increase the current 20 item limit
* Filter on the server

## How to Reproduce 🔨
Compare the list of catalogue items shown with the list available in the database. some will not be available for selection

1. Go to the Equipment list
2. Click on `New Asset`
3. Select a category and view the catalogue items available

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #3574

# 👩🏻‍💻 What does this PR do? 
 
Change the autocomplete used by the catalogue item selection on the create asset modal to use the infinite scroll.

To support this, I've lifted out the `usePagination` and `useFilter` that were implemented for the `StockItemSearchInput` and have added a version of the assets query which makes use of `useInfiniteQuery`.
Have also added a new filter option, to provide an 'or' query for the four fields required.

Try to create a new asset - check that the list of catalogue items is shown, with only 100 to start with. When testing you might want to change

```
const RECORDS_PER_PAGE = 100;
```

to use a lower number (25 say?) which would then trigger the refetch more often.
Check that the filter is working, and that you can still filter by code, manufacturer, model and name of the type.

Test that the StockItemSearchInput still functions (try to create a stocktake line, for example).


client/packages/coldchain/src/Equipment/ListView/CreateAssetModal.tsx
```diff
@@ -88,9 +94,16 @@ export const CreateAssetModal = ({
     useAssetData.utils.types({
       categoryId: { equalTo: draft.categoryId ?? '' },
     });
-  const { data: catalogueItemData } = useAssetData.document.list(
-    draft.categoryId ?? ''
-  );
+  const { pagination, onPageChange } = usePagination(100);
```
mark-prins, line 97: number is arbitrary - could use the default of 500 I guess.

client/packages/coldchain/src/Equipment/api/hooks/document/useAsset.ts
```diff
@@ -1,5 +1,4 @@
-import { useMutation, useParams } from '@openmsupply-client/common';
-import { useQuery } from 'react-query';
```
mark-prins, line 2: offroad - noticed that some files were importing directly from react-query which they should not!

jmbrunskill (approved):
Looks good to me!

server/repository/src/db_diesel/assets/asset_catalogue_item.rs
```diff
@@ -27,6 +29,7 @@ pub struct AssetCatalogueItemFilter {
     pub model: Option<StringFilter>,
     pub r#type: Option<StringFilter>,
     pub type_id: Option<EqualFilter<String>>,
+    pub code_manufacturer_model_type: Option<StringFilter>,
```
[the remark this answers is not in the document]
mark-prins: nice - thanks. will do

PR #3594, opened 2024-04-11
3503 mass reduce to zero

### LINKED ISSUE #3503: Mass reduce to 0 or location as action (will need modal)

lache-melvin:
Fixes #3503

# 👩🏻‍💻 What does this PR do? 
 

From stocktake, adds ability to mass reduce to 0, and mass change location:

Reason input will only show if reasons are required:
Screenshot 2024-04-11 at 5 25 44 PM

Screenshot 2024-04-11 at 5 25 51 PM

- Create a stocktake
- Add some lines
- Select some lines
- Select action "reduce lines to 0"
- [ ] Modal appears - if reasons configured, you have to enter a reason
- [ ] Ok button enables - click ok
- [ ] Selected placeholder lines (blue) are now included lines (black), the counted # packs is 0, and your reason should be applied to all of the selected lines
- Select some lines
- Select action "change location of lines"
- [ ] Modal appears - select a new location
- [ ] Click ok - all selected lines should now have their location updated

See inline comments

## 📃 Documentation

- [ ] From stocktake - how to mass reduce lines
- [ ] From stocktake - how to change location


client/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationWithChildrenModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface ConfirmationWithChildrenModalProps extends PropsWithChildren {
+  isOpen: boolean;
+  onClose: () => void;
+  onSave: () => Promise<void>;
+  title: string;
+  message: string;
+  canSave?: boolean;
+}
+
+export const ConfirmationWithChildrenModal = ({
```
lache-melvin, line 19: Added this as a new `common` component... not using the confirmation modal context/hook though, as that would require more of a refactor. Should the naming/where it lives be different so as not to confuse this component with the `useConfirmation` components?

client/packages/inventory/src/Stocktake/DetailView/ChangeLocationModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+  onCancel: () => void;
+}
+
+export const ChangeLocationConfirmationModal = ({
+  isOpen,
+  onCancel,
+}: ChangeLocationConfirmationModalProps) => {
+  const t = useTranslation('inventory');
+
+  const [location, setLocation] = useState<LocationRowFragment | null>(null);
```
lache-melvin, line 24: Would it be worth adding logic so if either only 1 line selected, or if all lines selected have same location, to show that location here instead of blank? Should it show the like `[multiple]` thing otherwise?

client/packages/inventory/src/Stocktake/DetailView/ChangeLocationModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const onChangeLocation = useStocktake.line.changeLocation();
+
+  return (
+    <ConfirmationWithChildrenModal
+      isOpen={isOpen}
+      title={t('heading.are-you-sure')}
+      message={t('messages.confirm-change-location')}
+      onClose={onCancel}
+      onSave={() => onChangeLocation(location)}
```
lache-melvin, line 34: Unsure about UX here... currently the OK button is always enabled, so if you open the modal and then immediately press OK, it would clear the location for all lines. Would probably be better if they had to explicitly select "No location" or something right?

client/packages/inventory/src/Stocktake/DetailView/ChangeLocationModal.tsx
andreievg: I like the blank tbh

client/packages/inventory/src/Stocktake/api/hooks/line/useChangeLinesLocation.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useSelectedRows } from '../utils/useSelectedRows';
+
+export const useChangeLinesLocation = () => {
+  const { saveAndMapStructuredErrors } = useSaveStocktakeLines();
+  const t = useTranslation('inventory');
+  const { error, success } = useNotification();
+
+  const selectedRows = useSelectedRows();
+
+  const onZeroQuantities = async (location: LocationRowFragment | null) => {
```
andreievg, line 13: onChangeLocations

client/packages/inventory/src/Stocktake/DetailView/Toolbar.tsx
```diff
@@ -47,6 +55,41 @@ export const Toolbar = () => {
   const setItemFilter = (itemFilter: string) =>
     updateQuery({ itemCodeOrName: itemFilter });
 
+  const {
+    isOpen: reduceIsOpen,
+    onClose: onCloseReduce,
+    onOpen: onOpenReduce,
+  } = useEditModal();
+  const {
```
andreievg, line 63: Hmm i wonder if we need to destructure these ones ?

const reduce = useEditModal();
const changeLocation = useEditModal();

// ..

reduce.isOpen // etc

client/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationWithChildrenModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+          marginTop={2}
+        >
+          <Grid item>
+            <DialogButton variant="cancel" onClick={onClose} />
+          </Grid>
+          <Grid item>
+            <DialogButton
+              variant="ok"
+              disabled={disableOk}
+              onClick={async () => onSave()}
```
andreievg, line 62: does it have to be defined async (can the consumer decide ?)

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useStocktakeLineEdit.ts
```diff
@@ -44,59 +39,9 @@ export const useStocktakeLineEdit = (
...
-      errorsContext.setError(id, error);
-    }
-
-    const errorMessages = Object.values(errorMessagesMap);
-    return {
-      errorMessages: errorMessages.length === 0 ? undefined : errorMessages,
-    };
-  };
-
   const save = async () => {
```
andreievg, line 42: Self review comment here would be great (where did it go : )

andreievg: hmm, also previously the context was a bit closer, actually it shouldn't even be context, zustand would be much better i think (that's just opinion though)

client/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationWithChildrenModal.tsx
andreievg: It feels like a layout component, maybe we can call it that ? We can also make it stateless, i kind of like that pattern ¯\_(ツ)_/¯. I think keeping it as you've done is fine too, not strong on it either way

andreievg (approved):
Thanks, this is great.

Keen to chat about layouts and composition vs generic components with state

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useStocktakeLineEdit.ts
lache-melvin: > Self review comment here would be great (where did it go : )

Ah yes sorry, thank you 🙈 

Keen to chat about this briefly when you get a sec...

client/packages/inventory/src/Stocktake/DetailView/ChangeLocationModal.tsx
andreievg: Actually we have a component like that right ? I think something would need to be selected

PR #3595, opened 2024-04-12
Update PR template

roxy-dao:
Updating PR template


lache-melvin (approved):
TY ❤️  Though tbh if the line is long my eyes are prob more likely to skip over it... could put reminder on a new line?

PR #3596, opened 2024-04-12
Add store to equipment import

### LINKED ISSUE #3459: CCEI: Add ability to import for a store in import by excel
Users want to be able to SELECT STORE when importing assets via csv from the central server

- [ ] This should be a required field when importing from central server.
- [ ] The example csv template and download errors csv etc should include a STORE column which uses store_code to map to store id (similar to code mapping to catalogue id)
- [ ] The csv from non central server should not receive a store id, and export example (and error templates) should not include the store column

### Additional context 💌
builds on issue #3139 

### Moneyworks Jobcode 🧰
OMSCC:AST

mark-prins:
Fixes #3459 

# 👩🏻‍💻 What does this PR do? 
 
If running central server, the following happens:

* Sample template has a store column
* Store code is required
* Entering a store code will lookup against all stores
* the failed export shows the store code

Screenshot 2024-04-12 at 12 33 38 PM

Have refactored the import a little to make it simpler to add the new cell.
No change to functionality if running in Remote mode.

Start the central server using `IS_CENTRAL_SERVER=true  cargo run` to test the store upload.

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
+++ b/client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
@@ -2,7 +2,12 @@ import React, { FC, useEffect, useState } from 'react';
 import { EquipmentReviewTab } from './ReviewTab';
 import { EquipmentUploadTab } from './UploadTab';
 import { EquipmentImportTab } from './ImportTab';
-import { useDialog, useNotification } from '@common/hooks';
+import {
+  QueryParamsProvider,
+  createQueryParamsStore,
+  useDialog,
+  useNotification,
+} from '@common/hooks';
 import {
   DialogButton,
   TabContext,
@@ -14,12 +19,14 @@ import {
   FileUtils,
   AssetLogStatusInput,
   FnUtils,
+  useIsCentralServerApi,
 } from '@openmsupply-client/common';
 import { useTranslation } from '@common/intl';
 import { useAssets } from '../api';
 import { importEquipmentToCsvWithErrors } from '../utils';
 import {
   AssetCatalogueItemFragment,
+  StoreRowFragment,
   useAssetData,
 } from '@openmsupply-client/system';
 import { DraftAsset } from '../types';
@@ -43,6 +50,7 @@ export type ImportRow = {
   id: string;
   notes: string;
   errorMessage: string;
+  store: StoreRowFragment | null | undefined;
 };
 
 export type LineNumber = {
@@ -64,6 +72,9 @@ export const toInsertEquipmentInput = (
   installationDate: row.installationDate,
   id: row.id,
   notes: row.notes,
+  store: row.store
+    ? { ...row.store, __typename: 'StoreNode', storeName: '' }
+    : null,
 });
 
 export const toExportEquipment = (
@@ -78,6 +89,7 @@ export const toExportEquipment = (
   notes: row.notes,
   lineNumber: index + 2,
   errorMessage: row.errorMessage,
+  store: row.store,
 });
 
 export const toUpdateEquipmentInput = (
@@ -113,9 +125,9 @@ export const EquipmentImportModal: FC<EquipmentImportModalProps> = ({
     fetchAsync,
     isLoading,
   } = useAssetData.document.listAll();
-
   const { mutateAsync: insertAssets } = useAssets.document.insert();
   const { insertLog, invalidateQueries } = useAssets.log.insert();
+  const isCentralServer = useIsCentralServerApi();
 
   const [bufferedEquipment, setBufferedEquipment] = useState<ImportRow[]>(
     () => []
@@ -130,7 +142,8 @@ export const EquipmentImportModal: FC<EquipmentImportModalProps> = ({
       bufferedEquipment.map((row: ImportRow, index: number) =>
         toExportEquipment(row, index)
       ),
-      t
+      t,
+      isCentralServer
     );
     FileUtils.exportCSV(csv, t('filename.cce-failed-uploads'));
     success(t('success'))();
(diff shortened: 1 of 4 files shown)


client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -285,15 +298,21 @@ export const EquipmentImportModal: FC<EquipmentImportModalProps> = ({
...
-            <EquipmentUploadTab
-              tab={Tabs.Upload}
-              catalogueItemData={catalogueItemData?.nodes}
-              setEquipment={setBufferedEquipment}
-              setErrorMessage={setErrorMessage}
-              onUploadComplete={() => {
-                changeTab(Tabs.Review);
-              }}
-            />
+            <QueryParamsProvider
```
mark-prins, line 301: required by the `useStore` hook

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -30,21 +39,96 @@ interface ParsedAsset {
...
+  Formatter.naiveDate(DateUtils.getDateOrNull(value));
 
-// the row object indexes are returned in column order
-// which allows us to index the keys
-const getCell = (row: ParsedAsset, index: AssetColumn) => {
-  const rowKeys = Object.keys(row);
-  const key = rowKeys[index] ?? '';
-  return row[key] ?? '';
-};
+function getImportHelpers<T, P>(
```
mark-prins, line 45: thinking that these can be lifted out when we generalise the import modal. was trying to make them generic

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -30,21 +39,96 @@ interface ParsedAsset {
...
+    formatter?: (value: string) => unknown
+  ) => {
+    const prop = t(localeKey) as keyof P;
+    const value = row[prop] ?? '';
+
+    addRequired(key, localeKey, formatter);
+
+    // check for duplicates
+    if (rows.some((r, i) => r[key] === value && index !== i)) {
+      rowErrors.push(t('error.duplicate-asset-number'));
```
fergie-nz, line 102: I wonder if this should be generic like the above error?

Something like
```
t('error.duplicate-field-error', {field: t(localeKey), })
```

fergie-nz: But probably not required until it is moved to some utils file

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -107,68 +195,42 @@ export const EquipmentUploadTab: FC<ImportPanel & EquipmentUploadTabProps> = ({
...
-      }
-      if (getCell(row, AssetColumn.SERIAL_NUMBER) !== undefined) {
-        importRow.serialNumber = getCell(row, AssetColumn.SERIAL_NUMBER);
-      }
-      if (getCell(row, AssetColumn.INSTALLATION_DATE) !== undefined) {
-        importRow.installationDate = Formatter.naiveDate(
-          DateUtils.getDateOrNull(getCell(row, AssetColumn.INSTALLATION_DATE))
+    data.data.forEach((row, index) => {
+      const { addLookup, addCell, addUnique, importRow, rowErrors } =
+        getImportHelpers(row, rows, index, t);
```
fergie-nz, line 203: I am wondering does the rows passed to getImportHelpers update after each time it has another value pushed to it? I think I haven't quite wrapped my head around what values will be updated within nested function calls, so it isn't clear to me.
Super cool!

client/packages/coldchain/src/Equipment/utils.ts
```diff
@@ -137,50 +137,67 @@ export const translateReason = (
...
-    node.serialNumber,
-    node.installationDate,
-    node.lineNumber,
-    node.errorMessage,
-  ]);
+  if (isCentralServer) {
+    fields.push(t('label.store'));
+  }
+
+  fields.push(t('label.asset-notes'));
```
fergie-nz, line 152: I wonder if all fields should be pushed rather than having asset number and catalogue-item-code in the original array?
I'm sure doesn't make a difference functionally, but makes it easier to read for me

fergie-nz (approved):
Looks good Mark,

I am learning a lot about abstracting utils functionality through reading this

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: nice! thanks - will pop that in

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: it will - the variables are passed as references, so you get the updated values. 
you're right though, it isn't clear and while it works.. it's a little hard to see what's going on!

client/packages/coldchain/src/Equipment/utils.ts
mark-prins: cool - have done. 🙏

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
fergie-nz: I think it is clear! - I think it is just me not being clear on how values are passed. Thank you for explaining

PR #3599, opened 2024-04-12
#3496(2) Adjustment tab query update

### LINKED ISSUE #3496: Add new tab in stock view modal for IA and simple adjustment

CarlosNZ:
Fixes #3496 (already closed, but this completes it)

# 👩🏻‍💻 What does this PR do? 
 

- Changes the mutation and its implementations to use the new back-end mutation endpoint, with correct input shape.
- Also uses "AdjustmentTypeInput" from schema rather than custom Enum, and removed the "None" option from it
- Fixed the pack quantity not being updated on save in the modal (see inline for explanation)

Can make adjustments from modal, with quantities correctly adjusted up or down accordingly.


client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
```diff
@@ -143,7 +131,11 @@ const getInventoryAdjustmentReasonInputColumn = (
           value={value}
           width={column.width}
           onChange={onChange}
-          adjustment={getAdjustment()}
+          adjustmentType={
+            rowData.snapshotNumberOfPacks > (rowData?.countedNumberOfPacks ?? 0)
+              ? AdjustmentTypeInput.Reduction
+              : AdjustmentTypeInput.Addition
+          }
```
CarlosNZ, line 138: Since we're don't have a "None" option anymore, will default to "Addition" except in that one specific case, so no longer need a separate method.

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
```diff
@@ -12,33 +13,32 @@ import {
   InventoryAdjustmentReasonRowFragment,
 } from '../api';
 
-export enum Adjustment {
-  Reduction,
-  Addition,
-  None,
-}
-
```
CarlosNZ, line 20: Replaced with codegen enum

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
```diff
@@ -48,7 +48,6 @@ export const InventoryAdjustmentReasonSearchInput: FC<
     <Box display="flex" flexDirection="row" width={120}>
       <Autocomplete
         autoFocus={autoFocus}
-        disabled={isDisabled}
```
CarlosNZ, line 51: Type is never not defined, so no need to disable reason selector.

client/packages/system/src/Stock/Components/StockLineEditModal.tsx
```diff
@@ -37,6 +37,10 @@ const useDraftStockLine = (
   const [stockLine, setStockLine] = useState<StockLineRowFragment>({ ...seed });
   const { mutate, isLoading } = useStock.line.update();
 
+  useEffect(() => {
+    setStockLine(seed);
+  }, [seed]);
+
```
CarlosNZ, line 43: There were two things causing the stockline state to not get updated -- this is one of them.

client/packages/system/src/Stock/ListView/ListView.tsx
```diff
@@ -179,6 +179,10 @@ const StockListComponent: FC = () => {
 
   const repackModalController = useToggle();
 
+  const stockLine = entity
+    ? data?.nodes.find(({ id }) => id === entity.id)
+    : undefined;
+
```
CarlosNZ, line 185: This is the other one -- the `entity` is stored in a useState inside the Modal hook, and only gets updated on opening the modal. So if we derive `stockLine` from `data` rather than `entity` it will change when data changes :)

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
andreievg: Hmm what if there are no reasons ?

CarlosNZ: I thought we already had a check for this upstream, but I guess not. I'll hide the row completely if Reasons are not defined.

CarlosNZ: Okay, have made it disabled when no reasons, as per discussion.

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -103,19 +97,16 @@ export const InventoryAdjustmentForm: FC<InventoryAdjustmentFormProps> = ({
...
+              onChange={value =>
                 setDraft(state => ({
                   ...state,
-                  adjustBy: adjustBy ?? 0,
-                  newNumberOfPacks:
-                    state.direction === Adjustment.Addition
-                      ? stockLine.totalNumberOfPacks + (adjustBy ?? 0)
-                      : stockLine.totalNumberOfPacks - (adjustBy ?? 0),
+                  adjustment: value ?? 0,
+                  adjustmentType: state.adjustmentType,
```
lache-melvin, line 109:
```suggestion
```
If doing `...state,` above, shouldn't need this line?

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentForm.tsx
```diff
@@ -127,7 +118,12 @@ export const InventoryAdjustmentForm: FC<InventoryAdjustmentFormProps> = ({
             <NumericTextInput
               width={INPUT_WIDTH}
               disabled={true}
-              value={draft.newNumberOfPacks}
+              value={
+                stockLine.totalNumberOfPacks +
+                (draft.adjustmentType === AdjustmentTypeInput.Addition
+                  ? draft.adjustment
+                  : -draft.adjustment)
+              }
```
lache-melvin, line 126: Ah nice doing this as display rather than managing in state 🚀

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
```diff
@@ -12,43 +13,43 @@ import {
...
+  width,
+  onChange,
+  autoFocus = false,
+  adjustmentType,
+  isError,
+}) => {
   const { data, isLoading } =
     useInventoryAdjustmentReason.document.listAllActive();
-  const isRequired = data?.totalCount !== 0;
-  const isDisabled = adjustment === Adjustment.None || !isRequired;
```
lache-melvin, line 38: Hm, if we're going to remove the `None` option, we need another mechanism to disable this input.

From stocktake, before counted packs is touched, or when snapshot packs === counted packs, reason input should be disabled. With this change, now can enter a reason:

Screenshot 2024-04-15 at 3 13 16 PM

And if you submit with the reason, you get an error... 

Perhaps can just add an `isDisabled` prop, reuse the logic from the old `getAdjustment` in StocktakeLineEditTables?

client/packages/system/src/Stock/ListView/ListView.tsx
lache-melvin: Nice find... a bit of a shame though, that `entity` should make managing the modal data easier 🙈 

I reckon let's get this in, it works for now, though Andrei mentioned an idea last week that we could bring into the API hooks, that would help us do this more cleanly 🤓

lache-melvin (changes requested):
Looks great, thanks Carl. Just the concern regarding the change to stocktake view, we should still be able to disable the input :)

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
```diff
@@ -12,43 +13,43 @@ import {
...
   const reasonFilter = (reason: InventoryAdjustmentReasonRowFragment) => {
-    if (adjustment === Adjustment.None) return false;
-    if (adjustment === Adjustment.Addition)
+    if (adjustmentType === AdjustmentTypeInput.Addition)
       return reason.type === InventoryAdjustmentReasonNodeType.Positive;
     return reason.type === InventoryAdjustmentReasonNodeType.Negative;
   };
   const reasons = (data?.nodes ?? []).filter(reasonFilter);
 
+  const isRequired = reasons.length !== 0;
```
lache-melvin, line 46: Also - nice catch doing the isRequired after the direction filter! 😅

client/packages/system/src/InventoryAdjustmentReason/Components/InventoryAdjustmentReasonSearchInput.tsx
CarlosNZ: Right, good catch. Have added the `isDisabled` prop and set it to true in StocktakeLineEditModal when the `countedNumberOfPacks` is `null` (that's the initial value). 91affa621

That enough you think?

lache-melvin: Above it looks like it also used to be disabled when `countedNumberOfPacks` === `snapshotNumberOfPacks`, and when `countThisLine` is `false`. I'd keep consistent with current logic 👍

CarlosNZ: Cool, I missed that bit above. Thanks.

d1bf85e41

lache-melvin (approved):
Awesome, thanks for changes

PR #3600, opened 2024-04-12
Central server UI KDD

mark-prins:
# 👩🏻‍💻 What does this PR do? 
 
A summary of the [central server UI discussion](https://github.com/msupply-foundation/open-msupply-internal/discussions/47#discussioncomment-9015942) as a KDD

I'm thinking option #4 - which would mean not implementing now. In the interim, what about putting in option #1, so that at least it is obvious when running a central server. I find it difficult to know when testing and a UI indicator would be helpful.


roxy-dao (approved):
Option 4 with an option 1 as interim sounds good (:

PR #3603, opened 2024-04-14
3457 returns ok next follow table sort

### LINKED ISSUE #3457: Returns: Ok & Next doesn't follow table sorting order
After sorting the Detail view table for a return, when clicking on a line, the "Ok & Next" button in the modal doesn't follow the sorted order of the lines.

lache-melvin:
Fixes #3457

# 👩🏻‍💻 What does this PR do? 
 

The OK & Next button now loops through items in the detail view, in the order that they are sorted in the table.

There were a couple of issues:

- Ok & Next was always using `lines` (not respecting when the table was grouped by item)
- The `getNextItemId` method wasn't handling the possibility of the same item appearing at multiple points in the list (it was looking for the first instance of that item, and getting the next item from there)

- Add some different items to a return
- Mess with their sort order
- [ ] When grouped by item - Ok & Next should loop through items in the same order as the table behind
- [ ] When not grouped by item - Ok & Next should loop through items in the same order as the table behind...
  - If an item appears more than once in the list, later appearances will be skipped

When not grouped by item, the `getNextItemId` method dedupes the items, then returns the next one.

The case where it isn't super intuitive is this one: 
Screenshot 2024-04-15 at 9 16 19 AM

If I open Acetylsalicylic Acid by clicking the batch `h` line, I would like the `next` to take me to Amoxicillin, but currently it would take me to Artesunate Injection (because that's the next item after the first instance of Acetylsalicylic Acid)

Managing the state of which row we clicked on/are looping through would be a bigger change - it's similar for shipments so I think ok for now?

## 📃 Documentation

_No user facing changes_


client/packages/invoices/src/utils.test.ts
```diff
@@ -0,0 +1,35 @@
+import { getNextItemId } from './utils';
```
CarlosNZ, line 1: Maybe give this file a more specific name ("invoiceUtils.test.ts"?) so we can target just this test? I ran `yarn test utils` and it ran 5 different tests with "utils" in the name/path.

CarlosNZ (approved):
Looks really good, and thanks for adding the tests.

Just one tiny comment about the test file name. 

Cheers

client/packages/invoices/src/utils.test.ts
lache-melvin: Can't we just use the jest runner extension in VSCode if you only want to run these tests?

CarlosNZ: Oh, I don't think I have that one. I normally use partial filename matching when I run specific tests.

lache-melvin: Ah its good! Similar to rust analyzer but only for the running/debugging tests part. If you search for this extension: firsttris.vscode-jest-runner

PR #3615, opened 2024-04-15
Add inventory adjustments feature flag for v2.0.0

### LINKED ISSUE #3604: Hide inventory adjustments for 2.0
Add feature flag to hide inventory adjustments in 2.0.

Similar to API_HOST env variable exposed through webpack

lache-melvin:
Fixes #3604

# 👩🏻‍💻 What does this PR do? 
 

Adds a feature flag to hide inventory adjustments features for v2.0.0.

- Start with usual `yarn start-local` 
  - [ ] no inventory adjustments features
- Start with the below script:
```sh
# yep, that many -- to pass through lerna scripts 😅
yarn start-local -- -- --env FEATURE_INVENTORY_ADJUSTMENTS=true
```
- You should now see:
  - [ ] `New stock` button at the top of the stock view
  - [ ] `Adjust` tab in the stock detail/edit modal
  - [ ] `Change location` and `Reduce lines to 0` actions in the stocktake details action dropdown

This still feels like the fastest way to a feature flag for now, but if it's something we're going to use more of, the `-- -- --env` feels super clunky. Maybe should set up .env file? Is that worth doing now?

## 📃 Documentation

_No user facing changes_

### DIFF
--- a/client/packages/host/webpack.config.js
+++ b/client/packages/host/webpack.config.js
@@ -144,6 +144,9 @@ module.exports = env => {
     plugins: [
       new ReactRefreshWebpackPlugin(),
       new webpack.DefinePlugin({
+        FEATURE_INVENTORY_ADJUSTMENTS: JSON.stringify(
+          env.FEATURE_INVENTORY_ADJUSTMENTS
+        ),
         API_HOST: JSON.stringify(env.API_HOST),
         LOCAL_PLUGINS: JSON.stringify(localPlugins()),
       }),
--- a/client/packages/host/webpack.config.js
+++ b/client/packages/host/webpack.config.js
@@ -144,9 +144,7 @@ module.exports = env => {
     plugins: [
       new ReactRefreshWebpackPlugin(),
       new webpack.DefinePlugin({
-        FEATURE_INVENTORY_ADJUSTMENTS: JSON.stringify(
-          env.FEATURE_INVENTORY_ADJUSTMENTS
-        ),
+        FEATURE_INVENTORY_ADJUSTMENTS: env.FEATURE_INVENTORY_ADJUSTMENTS,
         API_HOST: JSON.stringify(env.API_HOST),
         LOCAL_PLUGINS: JSON.stringify(localPlugins()),
       }),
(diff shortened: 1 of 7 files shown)


andreievg (approved):
Thanks, looks great, maybe we can also add update this one to suggest the `-- --` vs `cd/package/host`

https://github.com/msupply-foundation/open-msupply/blob/dba9751db791d3357454d654bbc53be91cfd11d0/client/README.md#L27

> Maybe should set up .env file? Is that worth doing now?

Not sure really, let's see how it goes ? Maybe a helper method ? `yarn start-future-features` ¯\_(ツ)_/¯

CarlosNZ:
> ```shell
> yarn start-local -- -- --env FEATURE_INVENTORY_ADJUSTMENTS=true
> ```

This is pretty ugly -- you think we should add a new script launcher to "package.json" to convert this for us? 

This works:
```json
"start-local-features": "yarn start-local -- -- --env $1",
```

And then run with `yarn start-local-features FEATURE_INVENTORY_ADJUSTMENTS=true`

client/packages/host/webpack.config.js
```diff
@@ -144,6 +144,9 @@ module.exports = env => {
     plugins: [
       new ReactRefreshWebpackPlugin(),
       new webpack.DefinePlugin({
+        FEATURE_INVENTORY_ADJUSTMENTS: JSON.stringify(
+          env.FEATURE_INVENTORY_ADJUSTMENTS
+        ),
```
CarlosNZ, line 149: Interesting, this actually does the opposite of what you want -- presumably you `JSON.stringify`-ed it to turn the string `"true"` into boolean `true`?

However, this ends up with `FEATURE_INVENTORY_ADJUSTMENTS: "true"` in `Environment` -- you just do it directly (`FEATURE_INVENTORY_ADJUSTMENTS: env.FEATURE_INVENTORY_ADJUSTMENTS,`) it makes it `true`.

I guess Webpack already handles "true" strings?

CarlosNZ (changes requested):
Cool, just a couple of small suggestions.

CarlosNZ (approved):
LGTM, thanks :)

PR #3616, opened 2024-04-15
3601 authorisation required for sync file endpoints

### LINKED ISSUE #3601: Authorisation Required for Sync File Endpoints
## What went wrong? 😲
Currently there's no login required for endpoints that allow 
- file uploads for sync files
- Downloading sync files

## Expected behaviour 🤔
Only logged in users should be able to download and upload files

(Maybe we need something more targetted? Permissions based on the table we're uploading too?

## How to Reproduce 🔨

1. Upload a file (See assets -> Documents for example)
2. Click on an uploaded document to get the link.
3. Copy link into new browser (Private Browsing Mode, or Differnet browser not logged in to OMS)
4. Should give you a permission denied error

## Proposed Solution

Could include https://github.com/msupply-foundation/open-msupply/issues/3002 to check the login cookie.
Permissions could be simply do you have a login for now...

jmbrunskill:
Fixes #3601

# 👩🏻‍💻 What does this PR do? 
Requests to upload, download files now have improved authentication

For remote_server the endpoints are just checking if the user is logged in or not.
For sync, we're using sync_v5 credentials similar to the other central sync requests.

Bit of local testing which seems ok...

Could/should apply the Logged in Function logic to import fridge tag, and sqlite file downloads.
SQLite download need to retain the permissions check too.

## 📃 Documentation
_No user facing changes_

### DIFF
--- a/server/server/src/support/mod.rs
+++ b/server/server/src/support/mod.rs
@@ -34,6 +34,8 @@ fn validate_request(
     service_provider: &ServiceProvider,
     auth_data: &AuthData,
 ) -> Result<ValidatedUser, AuthError> {
+    // TODO: Refactor to use authentication::validate_cookie_auth
+
     let service_context = service_provider
         .basic_context()
         .map_err(|err| AuthError::Denied(AuthDeniedKind::NotAuthenticated(err.to_string())))?;
--- a/server/server/src/sync_on_central/mod.rs
+++ b/server/server/src/sync_on_central/mod.rs
@@ -84,7 +84,7 @@ async fn download_file(
     request: Json<SyncDownloadFileRequestV6>,
     settings: Data<Settings>,
 ) -> actix_web::Result<impl Responder> {
-    println!("Download file requested");
+    log::info!("Sending a file via sync");
     let (file, file_description) = sync_on_central::download_file(&settings, request.into_inner())
         .await
         .map_err(ToResponseError)?;
--- a/server/server/src/sync_on_central/mod.rs
+++ b/server/server/src/sync_on_central/mod.rs
@@ -1,6 +1,6 @@
 use std::fmt::Display;
 
-use actix_multipart::Multipart;
+use actix_multipart::form::MultipartForm;
 use actix_web::{
     error::InternalError,
     http::{
@@ -11,17 +11,15 @@ use actix_web::{
     web::{self, Data, Json},
     Error, HttpRequest, HttpResponse, Responder, ResponseError,
 };
-use repository::sync_file_reference_row::SyncFileReferenceRowRepository;
 
-use crate::{central_server_only, static_files::handle_file_upload};
+use crate::central_server_only;
 use service::{
     service_provider::ServiceProvider,
     settings::Settings,
-    static_files::StaticFileCategory,
     sync::{
         api_v6::{
             SyncDownloadFileRequestV6, SyncParsedErrorV6, SyncPullRequestV6, SyncPullResponseV6,
-            SyncPushRequestV6, SyncPushResponseV6,
+            SyncPushRequestV6, SyncPushResponseV6, SyncUploadFileRequestV6,
         },
         sync_on_central,
     },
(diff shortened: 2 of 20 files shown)


server/service/src/sync/api_v6/upload_file.rs
```diff
@@ -0,0 +1,132 @@
...
+        // Read the file into memory (ideally could be a stream or something, and upload the file in chunk so we can stop quickly when sync starts/stops/pauses)
+        let mut file_bytes = Vec::new();
+        file_handle.read_to_end(&mut file_bytes).map_err(|err| {
+            log::error!("Error reading file: {:#?}", err);
+            FileSyncError::StdIoError(err)
+        })?;
+
+        // This is an example request structure, it just exists to give a compiler warning if the structure changes
+        // You'll need to modify the actual request using the reqwest::multipart::Parts below if there's a change here...
+        let _canary_request = SyncUploadFileRequestV6 {
```
jmbrunskill, line 64: Created this canary as I didn't see easy to go from an Actix multpart request to an request multipart request.
If there's a nice conversion process we could do that here...

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -180,3 +182,96 @@ pub async fn download_file(
...
+
+        let static_file = file_service.reserve_file(
+            &sanitized_filename,
+            &static_file_category,
+            Some(file_id.clone()),
+        )?;
+        let destination = Path::new(&static_file.path);
+
+        // Copy the file from the temp location to the final location
+        // TODO: Ideally these fs operations should be done in a separate such as using web::block (see handle_upload in static_files.rs)
```
jmbrunskill, line 245: I had lifetime issues doing this, so figured we'd probably get away with this in the short term. Potentially to hang the webserver if there was some kind of disk issue but that probably suggests bigger issues...

server/server/src/support/mod.rs
```diff
@@ -34,6 +34,8 @@ fn validate_request(
     service_provider: &ServiceProvider,
     auth_data: &AuthData,
 ) -> Result<ValidatedUser, AuthError> {
+    // TODO: Refactor to use authentication::validate_cookie_auth
```
andreievg, line 37: Thanks

server/service/src/sync/api_v6/mod.rs
```diff
@@ -162,3 +166,10 @@ pub struct SyncDownloadFileRequestV6 {
     pub(crate) id: String,
     pub(crate) sync_v5_settings: SyncApiSettings,
 }
+
+#[derive(MultipartForm)]
+pub struct SyncUploadFileRequestV6 {
+    pub sync_v5_settings: Json<SyncApiSettings>,
```
andreievg, line 172: This looks great !

server/server/src/sync_on_central/mod.rs
```diff
@@ -103,68 +101,33 @@ async fn download_file(
...
-            Ok(_) => {}
-            Err(err) => {
-                log::error!(
-                    "Error saving sync file reference after sync upload: {}",
-                    err
-                );
 
-                return Err(InternalError::new(err, StatusCode::INTERNAL_SERVER_ERROR).into());
+    let result = sync_on_central::upload_file(&settings, &service_provider, file_id, form).await;
+    match result {
```
andreievg, line 113: I think here we can follow the existing pattern of push and pull response ? Otherwise we would have three patterns for doing it, pull/push, download-file (I didn't know how to do multipart response, if that's even a thing, so the error is serialized in an error response), and then this one with statuses.

andreievg (approved):
Hi, thanks James. 

I approve but only the front end authorization, I have post poned a meeting to try and get this PR reviewed but didn't quite manage to review the central upload part. 

Do you mind splitting the PR into two, and maybe merging the local endpoints auth ?

server/server/src/sync_on_central/mod.rs
jmbrunskill: Do you mean I should have a return type like `SyncUploadSuccessV6`?
Not entirely sure what you mean sorry!

andreievg:
Merging this one, as there is follow up PR: https://github.com/msupply-foundation/open-msupply/pull/3653

PR #3619, opened 2024-04-15
#3618 decimal pack size kdd

### LINKED ISSUE #3618: Decimal Pack Size KDD

Read the KDD 👉 https://github.com/msupply-foundation/open-msupply/pull/3619

Chris-Petty:
Fixes #3618 

Read the [KDD](https://github.com/msupply-foundation/open-msupply/blob/3618-decimal-pack-size-kdd/decisions/2024-04-15_decimal_pack_size.md), agree, suggest changes, make yer case!

High priority! This is a long term blocker for Team Tauhou working on migrating from legacy systems.

### DIFF
--- a/decisions/2024-04-15_decimal_pack_size.md
+++ b/decisions/2024-04-15_decimal_pack_size.md
@@ -1,8 +1,8 @@
 # Decimal Pack Sizes
 
-- _Date_: 2024-04-16
-- _Deciders_: @Chris-Petty, @roxy-dao, @clemens-msupply, @mark-prins, @andreievg, @craigdrown, @louisaw123, @Gwalior-84
-- _Status_: PROPOSED
+- _Date_: 2024-04-30
+- _Deciders_: @Chris-Petty, @roxy-dao, @clemens-msupply, @mark-prins, @andreievg, @craigdrown
+- _Status_: AGREED
 - _Outcome_: Option 1
 
 ## Context
--- a/decisions/2024-04-15_decimal_pack_size.md
+++ b/decisions/2024-04-15_decimal_pack_size.md
@@ -44,7 +44,6 @@ _Pros:_
 
 - Initialisation works with no effort
 - Allows for targeted logic for file syncing (e.g rate limiting in future)
-- Allows for files to be downloaded where they're needed rather than needing to sync everywhere (via std sync process)
 - Straightforward for all reporting. What's in the invoice/requisition/PO/tender/GR is consistent with what the user entered.
 - Straightforward for integrations, no deep explanation needed.
 
--- a/decisions/2024-04-15_decimal_pack_size.md
+++ b/decisions/2024-04-15_decimal_pack_size.md
@@ -1,6 +1,6 @@
 # Decimal Pack Sizes
 
-- _Date_: 2024-02-28
+- _Date_: 2024-04-16
 - _Deciders_: @Chris-Petty, @roxy-dao, @clemens-msupply, @mark-prins, @andreievg, @craigdrown, @louisaw123, @Gwalior-84
 - _Status_: PROPOSED
 - _Outcome_: Option 1
--- a/decisions/2024-04-15_decimal_pack_size.md
+++ b/decisions/2024-04-15_decimal_pack_size.md
@@ -1,0 +1,127 @@
+# Decimal Pack Sizes
+
+- _Date_: 2024-02-28
+- _Deciders_: @Chris-Petty, @roxy-dao, @clemens-msupply, @mark-prins, @andreievg, @craigdrown, @louisaw123, @Gwalior-84
+- _Status_: PROPOSED
+- _Outcome_: Option 1
+
+## Context
+
+In Legacy mSupply it is possible, though generally not recommended, to enter decimal values for pack sizes. These are stored as a f64 value in the 4D database. Few instances of this truly exist. Inspection of some production systems report that there is one stock line in all the history of the system where a line was introduced with decimal packsize (e.g. 0.1485 packsize for a medicine with `tabs` for units), and were promptly removed with inventory adjustment. This would suggest it merely being a data entry error. At least one country has thousands of trans_lines affected likely in error (gloves in packs of 2.5, vials in packs 0.2 etc.).
+
+For sites converting from mSupply Legacy Desktop to Open mSupply, we need to ensure that the decimal values are either maintained or adjusted. Currently, Open mSupply only supports integer pack sizes, so must be a whole number or the lines will not be integrated. Open mSupply has integer pack sizes as in principle it is bad to manage discrete quantities with floating point numbers due to inaccuracy in calculations that can occur with floating point arithmetic. Handling migration of data was deferred.
+
+The proposed legitimate occurrences of this would primarily come from managing liquids, particular in systems that use Litres/L as a unit. If repacking 1L into 40x 50mL bottles, you'd need to enter this in mSupply as 40x 0.05L bottles as we do not support unit conversions.
+
+Goals:
+
+- Must allow Legacy mSupply data to be initialised into Open mSupply via sync
+- Must allow interoperability between Legacy and Open mSupply (i.e. transfers via sync)
+- Must make sense to users, keeping grounded in recording digitally what is physically present. 40 bottles on the shelf should be 40 bottles in the system, no more no less.
+
+### Additional Context
+
+- mSupply back and forth trying to make decimals calculate/present correctly
+  - [Tax rate](https://github.com/msupply-foundation/msupply/issues/10118#event-12127072697)
+  - [ledger](https://github.com/msupply-foundation/msupply/issues/10508)
+- [HSH KDD](https://github.com/msupply-foundation/health-supply-hub/blob/develop/docs/decisions/002-storing-monetary-values.md) generally deciding to use floats for monetary values.
+- [HSH issue](https://github.com/msupply-foundation/health-supply-hub/issues/637) deciding that quantity, pack_size and unit_quantity needs to be float for easiest interoperability with mSupply tenders and consistency. Several changes were added to mSupply to strongly deter uses from using decimal pack sizes in their tender lines and making them confirm it.
+- [OMS discussion](https://github.com/msupply-foundation/open-msupply/discussions/2905) for decimal values in stocktakes.
+- [Internal OMS discussion](https://github.com/msupply-foundation/open-msupply/discussions/29) for decimal values in ledger and UI.
+- [SAGE](https://www.greytrix.com/blogs/sageaccpacerp/2015/01/07/allow-fractional-quantities-in-purchase-order/) has a preference - allows 4 decimal places if on, zero if off. Permanent if switched on.
+- [SAP](https://content.cnbssoftware.com/blog/rounding-issues-sap-uom) blog post detailing that SAP allows decimals. In this case appears to be 3dp allowed. They have Unit of Measure of "case", which contains 6 units of "each" (aka a box of 6 bottles), but it can be split as sometimes is required IRL. So 1/6 = 0.167. 6\*0.167 = 1.002 though ;-). Shenanigans are had - if stock out of 1 "each" then they make sure the source keeps the remainder i.e. `1 case - 1 each = 0.834` stock left at the source, rather than `0.833`. The receiving location gets `0.166`, rather than `1.67`. This way the amount in stock at the source location is never a partial each (>0, but <1). But now the receiver location has a partial each! When they do the next sales order, they bump the quantity in the order up to account for the "fractional dust" but are careful to only do it within a threshold to make sure it doesn't appear to send/create stock that never existed...
+
+## Options
+
+### Option 1 - Support Decimal Pack Size like Legacy mSupply
+
+This is the most straight forward for migration/integration as it's 1-to-1, the fields are simply the same numeric type. Downsides include introducing another avenue of floating point inaccuracy in arithmetic. It makes sense to users, as they entered the data and it is presented in the manner they entered it.
+
+There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
+
+_Pros:_
+
+- Initialisation works with no effort
+- Allows for targeted logic for file syncing (e.g rate limiting in future)
+- Allows for files to be downloaded where they're needed rather than needing to sync everywhere (via std sync process)
+- Straightforward for all reporting. What's in the invoice/requisition/PO/tender/GR is consistent with what the user entered.
+
+_Cons:_
+
+- UI needs to handle potentially large amounts of decimal digits gracefully. e.g. `1.3333333...` to `1.33` (2dp rounding, preference) or `1 1/3` (can be presented with neat HTML, but only nice for fractions with small denominators, harder to compare one cell to the next).
+- Backend should keep full stored value to maximum accuracy. These values can be smoothed over in some cases (e.g 0.9999999999999997 can be rounded up and treated as 1 in UI, in the backend it'll be that quantity).
+- Calculations will accrue inaccuracy. For individual lines this is usually tolerable but it does accumulate! In tenders we see it in monetary values: summing the "adjust cost" of several thousand lines can result in whole dollars of inaccuracy. Governments/Procurement agencies may have policies of acceptable level of calculation inaccuracy.
+  - It is worth highlighting that number of packs and our monetary fields are all already floating point type with these same problems, so we must already handle all the above issues regardless. **So it's not really a con if we have to do it anyway.** It just makes the surface area of floating point accuracy a smidgen higher.
+- recurring numbers create headaches of who truly has the "full" amount of something. E.g
+
+### Option 2 - Round pack size up to next whole number and adjust number of packs, monetary fields etc.
+
+We convert decimal pack sizes up to the next whole number.
+
+E.g.
+
+1 pack of 1L is repacked to 40 packs of 0.05L in mSupply desktop. The latter on integration in OMS:
+
+1. pack size: 0.05L/0.05 = 1L
+2. number of packs: 40packs\*0.05 = 1pack
+3. cost/sell price: 0.5/0.05 = $10
+
+The pharmacist feedback I have received on this is that it'd be confusing to users if they had a system telling them they had 1 of when they can see there are 40 smaller vessels on the shelf. For accounting, I have not asked a finance stakeholder yet but I presume this is very very bad form in regards of adjusting the number of what was physically ordered/received and the cost of each item ordered.
+
+A potential solution for this is to use pack variants. After converting back to 1L pack size, we use a variant of "50mL bottle" that converts all the properties back in the UI layer. Pack variants don't appear to quite support this - the assumption is the base unit is a small whole number (integer). In this case we'd need a pack variant of 0.05L, which would do the reverse of the calculations above. There is additionally an issue of when and how these pack variants are generated, as they are centrally controlled data.
+
+Going to the option 1 problem statement, we also have to solve this at integration still so no avoided in anyway:
+
+> There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
+
+_Pros:_
+
+- Open mSupply can keep using integer for pack sizes, which feels good in a programming lens. Floats with their inaccuracy are genuinely not ideal!
+
+_Cons:_
+
+- Unintuitive for users, stock levels in system no longer represents physical reality.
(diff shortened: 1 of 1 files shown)


Chris-Petty:
@craigdrown more or less [says](https://github.com/msupply-foundation/open-msupply/discussions/2905#discussioncomment-8413621) "we have to do decimal because users want to":

>And https://github.com/msupply-foundation/org-issues/discussions/40 for where we got to 18 months ago.
I'm glad we have @louisaw123 ;-) if a user wants to enter a pack size of 2.5 Litres or 0.5 ml we need to let them.
My old person view is that we started with integers in old mSupply and changed to floats, and 20 years later you're deciding to relearn the lessons the hard way rather than accepting that it's like that for a reason (and I'm sure someone will find a comment from me saying "floats aren't needed for pack sizes in omSupply" ;-) ).
I can't see how you can migrate a site from mSupply OG to using integer packs without forcing them to change the units they use - software colonialism ;-)

@andreievg [please make case for pack variants approach](https://github.com/msupply-foundation/open-msupply/discussions/2905#discussioncomment-8400992). I tried to do off code but may not have done justice sorry!

CarlosNZ:
> @craigdrown more or less [says](https://github.com/msupply-foundation/open-msupply/discussions/2905#discussioncomment-8413621) "we have to do decimal because users want to":
> 
> > I'm glad we have @louisaw123 ;-) if a user wants to enter a pack size of 2.5 Litres or 0.5 ml we need to let them.

We can let them use whatever unit they want but the *base* unit can still be required to be an integer:  2,500 mL, 500µL, etc.

Then they can use pack variants to display in whatever more "convenient" unit they desire.

I'm not actually arguing for one approach over another, I'm just pointing out that requiring integer values doesn't granular precision.

(Not sure what to do if they've chosen Litres as the base unit for an item, and *later* want to define a smaller base unit, but we can probably come up with a solution)

roxy-dao:
To know the base unit we need to know the unit the item has which might not always be the case... Training to enforce unit input may be great at first but training can only go so far... Can say the same for repacks as well, we can train users to repack but they may not always repack when the split a pack...

Chris-Petty:
Adding a thought from a convo with @regotaina - if allowing decimals it might be nice to have a pref per item in the system. E.g. (Default) don't allow decimal pack size for `paracetamol 500mg` (unit of measure _tablet_) but do turn it on specifically for `paracetamol 50mg/mL` (unit of measure _Litre_).

Chris-Petty:
> > @craigdrown more or less [says](https://github.com/msupply-foundation/open-msupply/discussions/2905#discussioncomment-8413621) "we have to do decimal because users want to":
> > > I'm glad we have @louisaw123 ;-) if a user wants to enter a pack size of 2.5 Litres or 0.5 ml we need to let them.
> 
> We can let them use whatever unit they want but the _base_ unit can still be required to be an integer: 2,500 mL, 500µL, etc.
> ...
> (Not sure what to do if they've chosen Litres as the base unit for an item, and _later_ want to define a smaller base unit, but we can probably come up with a solution)

Units are defined at the item level as it has to be concrete and consistent. We can't guarantee a use won't configure items with L as their base unit. mL is a reasonable default, but not guaranteed to be small enough either. 

Big problem with changing a item's unit scale is what to do for historic data. If you change from L to mL currently you'd have to adjust all historic lines by the scale change which basically a big no no as far as I'm aware. I assume it's bad form to say a supplier actually supplied 1000x the units at a 1/1000th the price rather than what the supplier says they sent you even if _technically_ the same haha. Beyond that you could make a conversion system (option 4) so that an item can be either mL or L (or µL) and comparisons can be made (might make SQL queries tricky though, unless there must always be post processing of SQL results). If doing all that probably make a nightmare for integrations too.

louisaw123:
> > > @craigdrown more or less [says](https://github.com/msupply-foundation/open-msupply/discussions/2905#discussioncomment-8413621) "we have to do decimal because users want to":
> > > > I'm glad we have @louisaw123 ;-) if a user wants to enter a pack size of 2.5 Litres or 0.5 ml we need to let them.
> > 
> > 
> > We can let them use whatever unit they want but the _base_ unit can still be required to be an integer: 2,500 mL, 500µL, etc.
> > ...
> > (Not sure what to do if they've chosen Litres as the base unit for an item, and _later_ want to define a smaller base unit, but we can probably come up with a solution)
> 
> Units are defined at the item level as it has to be concrete and consistent. We can't guarantee a use won't configure items with L as their base unit. mL is a reasonable default, but not guaranteed to be small enough either.
> 
> Big problem with changing a item's unit scale is what to do for historic data. If you change from L to mL currently you'd have to adjust all historic lines by the scale change which basically a big no no as far as I'm aware. I assume it's bad form to say a supplier actually supplied 1000x the units at a 1/1000th the price rather than what the supplier says they sent you even if _technically_ the same haha. Beyond that you could make a conversion system (option 4) so that an item can be either mL or L (or µL) and comparisons can be made (might make SQL queries tricky though, unless there must always be post processing of SQL results). If doing all that probably make a nightmare for integrations too.

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+This is the most straight forward for migration/integration as it's 1-to-1, the fields are simply the same numeric type. Downsides include introducing another avenue of floating point inaccuracy in arithmetic. It makes sense to users, as they entered the data and it is presented in the manner they entered it.
+
+There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
+
+_Pros:_
+
+- Initialisation works with no effort
+- Allows for targeted logic for file syncing (e.g rate limiting in future)
+- Allows for files to be downloaded where they're needed rather than needing to sync everywhere (via std sync process)
```
andreievg, line 47: These two file `files`, is that copy paste artefact ?

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+_Cons:_
+
+- UI needs to handle potentially large amounts of decimal digits gracefully. e.g. `1.3333333...` to `1.33` (2dp rounding, preference) or `1 1/3` (can be presented with neat HTML, but only nice for fractions with small denominators, harder to compare one cell to the next).
+- Backend should keep full stored value to maximum accuracy. These values can be smoothed over in some cases (e.g 0.9999999999999997 can be rounded up and treated as 1 in UI, in the backend it'll be that quantity).
+- Calculations will accrue inaccuracy. For individual lines this is usually tolerable but it does accumulate! In tenders we see it in monetary values: summing the "adjust cost" of several thousand lines can result in whole dollars of inaccuracy. Governments/Procurement agencies may have policies of acceptable level of calculation inaccuracy.
+  - It is worth highlighting that number of packs and our monetary fields are all already floating point type with these same problems, so we must already handle all the above issues regardless. **So it's not really a con if we have to do it anyway.** It just makes the surface area of floating point accuracy a smidgen higher.
```
andreievg, line 56: Yes, but is it correct to say that the effect could be multiplied if both number_of_packs and pack_size are decimal ¯\_(ツ)_/¯, or at least the cognitive load to assess use cases increases ?

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Options
+
+### Option 1 - Support Decimal Pack Size like Legacy mSupply
+
+This is the most straight forward for migration/integration as it's 1-to-1, the fields are simply the same numeric type. Downsides include introducing another avenue of floating point inaccuracy in arithmetic. It makes sense to users, as they entered the data and it is presented in the manner they entered it.
+
+There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
```
andreievg, line 41: I would have though that in the context of this discussion, repeating decimal cannot happen, number of packs can be repeating decimal, but I don't see how pack size can be entered to be repeating decimal ?

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+1. pack size: 0.05L/0.05 = 1L
+2. number of packs: 40packs\*0.05 = 1pack
+3. cost/sell price: 0.5/0.05 = $10
+
+The pharmacist feedback I have received on this is that it'd be confusing to users if they had a system telling them they had 1 of when they can see there are 40 smaller vessels on the shelf. For accounting, I have not asked a finance stakeholder yet but I presume this is very very bad form in regards of adjusting the number of what was physically ordered/received and the cost of each item ordered.
+
+A potential solution for this is to use pack variants. After converting back to 1L pack size, we use a variant of "50mL bottle" that converts all the properties back in the UI layer. Pack variants don't appear to quite support this - the assumption is the base unit is a small whole number (integer). In this case we'd need a pack variant of 0.05L, which would do the reverse of the calculations above. There is additionally an issue of when and how these pack variants are generated, as they are centrally controlled data.
+
+Going to the option 1 problem statement, we also have to solve this at integration still so no avoided in anyway:
```
andreievg, line 75: I added a comment about the pack size never being repeating decimal, does that help with resolving this ?

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+> There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
+
+_Pros:_
+
+- Open mSupply can keep using integer for pack sizes, which feels good in a programming lens. Floats with their inaccuracy are genuinely not ideal!
```
andreievg, line 81: One other pro, which is something I've been thinking about lately, is reduced cognitive load due to reduction of use cases. I think that's a metric that can help assess the very objective 'simplicity'. (something is simpler if it leads to reduced set of use cases to consider)

Like in this case we can say that decimal pack size leads to a lot more use cases to think through (all of the presentation layer stuff, that's a long time in discussion, the potential of both pack size and number of packs being decimal and the effect on accuracy of the stored data).

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+_Cons:_
+
+- Unintuitive for users, stock levels in system no longer represents physical reality.
+  - Could be assuaged by making pack variants support decimals by being f64 rather than i32. Is this just shifting the problem though? We'd still need to handle when and if rounding happens in UI and calculations no matter what we do.
```
andreievg, line 86: I actually think that rounding is something that we don't need to consider, as the user wouldn't divide things that way ? If it happens it's most likely use error, I think 'decimal' system should support most division without repeating or very large decimal numbers ? (I guess you can be converting from imperial to metric ? ohh man, don't want to thing about this lol)

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,131 @@
...
+
+Decimals would be safe then. pgsql has one of these built in, sqlite does not. We'd need a rust implementation to manage these monetary values. One such rust crate is [rusty-money](https://crates.io/crates/rusty-money). It is of note that rust is a young language, and such libraries are relatively immature compared to [Java](https://docs.oracle.com/cd/E13166_01/alcs/docs51/javadoc/com/elasticpath/domain/misc/Money.html), [.NET](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-decimal) etc.
+
+(Note .NET's `Decimal` has basically the same problems as f64 still, but the errors are relatively much smaller! Is a fancy 128bit number implementation rather than f64 standard)
+
+This is too big a task. All reports and all database and business logic would have to be rewritten. OMS needed to commit to this early if at all so I think the ship has sailed, won't write anymore about it! Would have incurred a complexity tax on all development anyway, and Legacy would still be sending us f64 to deal with lol.
+
+## Decision
+
+**Option 1 - it is the simplest and the consequences are generally known.**
```
andreievg, line 119: Ok, agree, mainly because, less logic in translation = less use case. And also if we don't spend the extra time trying to account for repeating numbers (which I think have a very very small chance of happening). And also we use pack variants to help use here:

Like your previous example: `1 pack of 1L is repacked to 40 packs of 0.05L`, we create new pack variant = 0.05L, now instead of showing `0.05L` in pack size, it would say the short name "5ml bottle" x 40.  <- and this can work without repack, user can choose to repack or use the pack variant 'lens'

decisions/2024-04-15_decimal_pack_size.md
Chris-Petty: hahaha es whoops

decisions/2024-04-15_decimal_pack_size.md
Chris-Petty: > Ok, agree, mainly because, less logic in translation = less use case. And also if we don't spend the extra time trying to account for repeating numbers (which I think have a very very small chance of happening). And also we use pack variants to help use here:
> 
> Like your previous example: `1 pack of 1L is repacked to 40 packs of 0.05L`, we create new pack variant = 0.05L, now instead of showing `0.05L` in pack size, it would say the short name "5ml bottle" x 40. <- and this can work without repack, user can choose to repack or use the pack variant 'lens'

Pack variants are currently `i32`, am I misunderstanding or would we need to change that to `f64` too?

decisions/2024-04-15_decimal_pack_size.md
Chris-Petty: I agree with the philosophy, but I think I disagree that it's less cognitive load:

1. Pack size is int so OMS sync translation will munge any lines that have decimal pack size. This must also be remembered in API for any future 3rd party integration (takes a lot to explain physical 100x 750mL bottles are being recorded as 75x 1L). Support/PMs must train pack variants to be used to achieve veneer of a particular type of repack (partial pack size), rather than using the repack functionality they'd usually use for the task. What if the wrong pack variant is being displayed? What if the appropriate pack variant isn't in the system? We will still have cognitive load of f64 calcs/imprecision as all related fields for arithmetic with pack_size are f64. Do we need to make sure any such records if munged are always pushed back to the server immediately? If from a transfer, does it now mismatch the other half or do we need to record what pack size to convert _back_ to when pushing to central server? Do repack transactions with decimal values now appear to have made no repack because we undid the conversion? Who is burdened to explain this to users?

or 

2. Pack size is float so sync translation is direct. If a 3rd party system also uses f64 we're fine (or they're using a nested pack sort of thing so a complex mapping is needed regardless of solution). UX is the same as Legacy mSupply too and relatively straightforward (use repack). We do have to deal with f64 precision in calculations, which in most contexts we already have to do for several related fields (rust calculations require us coercing i32 to f64 to do math with any of those f64 fields anyway, so we really don't avoid any bullets by keeping i32 here). We already require an imprecision mitigation strategy because number of packs is f64. We'd have to apply this strategy if munging records from f64 to i32 anyway as we recalculate the other f64 fields. UI may have to round certain imprecision away which we have to address anyway given the other f64 fields, so would reuse the components.

Option 1 sounds like a lot more new and unique cognitive load introduced, while option 2 to me is a whole lot stuff we already have to consider?

decisions/2024-04-15_decimal_pack_size.md
Chris-Petty: To use the SAP example they might have 6 "each" in a "case". If you tried to repack and manage stock as such... 
image

I think somewhat contrived as I'm not aware of anyone doing that in mSupply, but technically nothing stopping you. Also clearly in the wild this sort of thing can happen given that SAP article.

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Options
+
+### Option 1 - Support Decimal Pack Size like Legacy mSupply
+
+This is the most straight forward for migration/integration as it's 1-to-1, the fields are simply the same numeric type. Downsides include introducing another avenue of floating point inaccuracy in arithmetic. It makes sense to users, as they entered the data and it is presented in the manner they entered it.
+
+There is potential for discrepancies to occur, e.g. store A repacks `1 * 1L` to `3 * 0.33L`. There are 0.01 packs of the original line... do we just zero it? If we issue all 3 to another store, they receive 0.99L. Do we round it up? What's the threshold?
```
Chris-Petty, line 41: @louisaw123 experience of systems where there is remaining stock that can't be split:

> so when we enter the repack in there would be an option for us where we would need to enter the quantity returned or quantity destroyed if there are any remaining from the pack we used -> so normally something like 0.01L remaining we would enter in as 0.01 destroyed / wastage to take it out of system and generally cos its not something we would put back on shelf

decisions/2024-04-15_decimal_pack_size.md
```diff
@@ -0,0 +1,130 @@
...
+- Doesn't logically work for all units, so feasibility is near 0 IMO.
+
+### Option 4 - Switch all f64 amount fields to a monetary safe type
+
+Decimals would be safe then. pgsql has one of these built in, sqlite does not. We'd need a rust implementation to manage these monetary values. One such rust crate is [rusty-money](https://crates.io/crates/rusty-money). It is of note that rust is a young language, and such libraries are relatively immature compared to [Java](https://docs.oracle.com/cd/E13166_01/alcs/docs51/javadoc/com/elasticpath/domain/misc/Money.html), [.NET](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-decimal) etc.
+
+(Note .NET's `Decimal` has basically the same problems as f64 still, but the errors are relatively much smaller! Is a fancy 128bit number implementation rather than f64 standard)
+
+This is too big a task. All reports and all database and business logic would have to be rewritten. OMS needed to commit to this early if at all so I think the ship has sailed, won't write anymore about it! Would have incurred a complexity tax on all development anyway, and Legacy would still be sending us f64 to deal with lol.
```
clemens-msupply, line 114: Or use a rational number types, e.g. represent 1/3. However, all mentioned disadvantages apply as well.

clemens-msupply (commented):
Agree with Option 1. You could come up with complicated config systems, e.g. to define 1 crate has 6 bottles, or define some universal base units etc. However, will this be done correctly a priori? For example, if you quickly want to hand out one bottle out of a crate you should be able to just do it, otherwise the user might decide to not put it in the system at all or handle it externally.

mark-prins (approved):
Agree with option 1. 
I can see comments from Andrei & Clemens also for #1 - anyone object?

decisions/2024-04-15_decimal_pack_size.md
andreievg: > munged

I would say **adjusted**, in my view **munged** is what they are now (with fractional pack_size), because item wasn't defined in base units correctly (base units should have been lowest dispensable unit, or something that can make sense in a pack size for lower dispensable unit), in this case in warehouse they would have pack_size = 1000 and when 750, respectively with base unit of 1ml). 

>  If from a transfer, does it now mismatch the other half or do we need to record what pack size

What do we do when we repack to 1 (all of mobile)

> Do we need to make sure any such records if munged are always pushed back to the server immediately?

Nope, but yes something to consider

> Do repack transactions with decimal values now appear to have made no repack because we undid the conversion?

They appear as split (if pack size didn't change form original)

> Who is burdened to explain this to users?

How do you explain to user that their floats don't add up ? (sorry that's an anti question, but the point is, there is still something potentially to explain, which one is easier to explain ¯\_(ツ)_/¯)

> Option 1 sounds like a lot more new and unique cognitive load introduced, while option 2 to me is a whole lot stuff we already have to consider?

I agree in a way, but I also had a point of what if both of the number of packs and pack_size are float, how do you deal with precision then (which i thought is different to "We already require an imprecision mitigation strategy")

Tbh we had similar discussion when fractional number of packs was introduced to omSupply, kind of got overpowered with the fact that mSupply supports it, and now that same argument (that we now support it in omSupply) is used for this, so have to concede to go with option 1 (even though at heart I think both number of pack and pack size should be integer and defined as a smallest dispensable unit, in the first place)

Chris-Petty:
@andreievg @clemens-msupply @jmbrunskill (I didn't want to ping you while on holiday, pinged now!) I'll hope to merge this today

clemens-msupply (approved):
Outside this KDD, I wonder if we can or should make changes to omSupply to minimize the use of decimal numbers in the future. For example, fluids are always measured in [ml] but you can specify that the item is measured in [L]. Or is this already sufficiently handled by pack variances?

decisions/2024-04-15_decimal_pack_size.md
andreievg: yes

andreievg:
@Chris-Petty I am ok with option 1. 

I would not be ok with expanding a lot of effort to support use case where fractions become repeated or very large. I think those use cases are more theoretical then practical, and if they happen and exist in a wild, we can just say we intentionally decided to relocate the budget from solving this rare use case to other areas.

PR #3623, opened 2024-04-15
Use Upsert for soft delete

### LINKED ISSUE #3621: Sync File Deletes don't sync
## What went wrong? 😲
Sync File References don't delete correctly via sync

## Expected behaviour 🤔

If a file is deleted on one site, other sites (such as central) should see the change reflected

## How to Reproduce 🔨

1. Upload a file from a remote server to an asset (for example)
2. Sync to central
3. Delete the file from Central
4. Sync
5. File still appears on the remote site under the asset

## Your environment 🌱

## Note
We might also want to delete the files from disk but that is a different issue

jmbrunskill:
Fixes #3621

# 👩🏻‍💻 What does this PR do? 
Quick bug fix for sync_file_references not being deleted property.

The better way might be to implement the Delete Trait and Delete translations.
This would also allow us to delete the files from disk.
For now though I think this is a good interim fix

Try adding a deleting files from assets with a central and remote server.
Check different combinations work as expected.

## 📃 Documentation

_No user facing changes_


mark-prins (approved):
👍 LGTM thanks James

PR #3626, opened 2024-04-16
Display catalogue item properties

### LINKED ISSUE #3481: CCEI: Display `Energy source` for CCE in category `Refrigerators and freezers`

#3474 captures the energy source per CCE (for category `Refrigerators and freezers`)

They will be an enum of `Electricity`, `Solar`, `Passive`, `Kerosene`, `Gas`

We need to display the details in the Asset Registry view for CCE in the category `Refrigerators and freezers`

Something like: 

image

mark-prins:
Fixes #3481

# 👩🏻‍💻 What does this PR do? 
 
Adds a loader for the catalogue item properties and displays on the summary tab if any are available:

Screenshot 2024-04-16 at 4 28 09 PM

If you create an asset which is from the 'Refrigerators and freezers' category, and has properties ( such as a refrigerator ) you'll see the props. otherwise there's nothing.


server/repository/src/db_diesel/assets/asset_catalogue_item_property.rs
```diff
@@ -38,8 +51,24 @@ impl<'a> AssetCatalogueItemPropertyPropertyRepository<'a> {
 
         Ok(result.into_iter().map(to_domain).collect())
     }
-}
 
+    pub fn query_property_and_value(
```
mark-prins, line 55: this is a convenience - I'm wanting to show property name (from `asset_catalog_property` table) and the assigned value for this asset (which is in the `asset_catalog_item_property` table) and rather than making two database calls, I'm doing this so that I can use the inner join.

server/server/src/logging.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(not(target_os = "android"))]
```
mark-prins, line 1: offroad: noticed that there's a warning if you build android, so moved it down.

adamdewey:
Nice @mark-prins ! 

I think the end user might not get their head around the properties belonging to the Catalogue CCE and not the Registry CCE. 

Perhaps we should just call it `Properties` instead of `Catalogue properties` ?

fergie-nz (approved):
Looks great to me.

Only thing I found slightly confusing was the naming of asset_catalogue_property and asset_catalogue_item_property, but I got there in the end.

I also wonder if the catalogue properties should also be in asset_catalogue queries?
Would there be cases where the same catalogue item sometimes has a property, and sometimes doesn't ? Or would they exist or not exist for a given catalogue_item?

adamdewey:
> I also wonder if the catalogue properties should also be in asset_catalogue queries? Would there be cases where the same catalogue item sometimes has a property, and sometimes doesn't ? Or would they exist or not exist for a given catalogue_item?

Property fields should be defined per Catalogue `Category` and should always appear for all Asset Templates (? still don't think we have a good word for this entity yet) in that category.

mark-prins:
> Perhaps we should just call it `Properties` instead of `Catalogue properties` ?

sure, can do. That's where I started, and then changed it - the properties values aren't editable for an asset, so was trying to indicate that these are reference values. Let's change to properties and see what happens 🙏

mark-prins:
> I also wonder if the catalogue properties should also be in asset_catalogue queries?

I didn't have a use case for that yet, so didn't put it in. Really, the props should show in the catalogue view, but I didn't want to stretch out the catalogue list view table, by adding a bunch more columns. We probably need a detail view to cater for this - and that was pushed out to the mythical phase 2. Can I leave it until then?

> Would there be cases where the same catalogue item sometimes has a property, and sometimes doesn't ? Or would they exist or not exist for a given catalogue_item?

As per @adamdewey's comment. should show for all items which are based on a catalogue item.

PR #3630, opened 2024-04-16
Soft delete currencies #3607

### LINKED ISSUE #3607: Soft delete currencies
Currencies can only be deleted in OG and only if they aren't in any FK. Should be rare to have a currency deleted, but would be nice to soft delete them on OMS.

roxy-dao:
Fixes #3607

# 👩🏻‍💻 What does this PR do? 
Adds an `is_active` field (not to be confused with OG's is_active field which isn't being used right now) to currency for soft deletion of currencies.

- [ ] Run tests
- [ ] Delete a currency from OG and check OMS DB to see bool value is false

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

### DIFF
--- a/server/repository/src/db_diesel/currency.rs
+++ b/server/repository/src/db_diesel/currency.rs
@@ -20,6 +20,7 @@ pub struct Currency {
 pub struct CurrencyFilter {
     pub id: Option<EqualFilter<String>>,
     pub is_home_currency: Option<bool>,
+    pub is_active: Option<bool>,
 }
 
 pub enum CurrencySortField {
@@ -92,6 +93,12 @@ fn create_filtered_query(filter: Option<CurrencyFilter>) -> BoxedLogQuery {
             Some(false) => query.filter(currency_dsl::is_home_currency.eq(false)),
             None => query,
         };
+
+        query = match filter.is_active {
+            Some(true) => query.filter(currency_dsl::is_active.eq(true)),
+            Some(false) => query.filter(currency_dsl::is_active.eq(false)),
+            None => query,
+        };
     }
     query
 }
@@ -112,6 +119,11 @@ impl CurrencyFilter {
 
     pub fn is_home_currency(mut self, filter: bool) -> Self {
         self.is_home_currency = Some(filter);
+        self
+    }
+
+    pub fn is_active(mut self, filter: bool) -> Self {
+        self.is_active = Some(filter);
         self
     }
 }
--- a/server/graphql/types/src/types/currency.rs
+++ b/server/graphql/types/src/types/currency.rs
@@ -28,6 +28,7 @@ pub struct CurrencySortInput {
 pub struct CurrencyFilterInput {
     pub id: Option<EqualFilterStringInput>,
     pub is_home_currency: Option<bool>,
+    pub is_active: Option<bool>,
 }
 
 impl CurrencyFilterInput {
@@ -35,11 +36,13 @@ impl CurrencyFilterInput {
         let CurrencyFilterInput {
             id,
             is_home_currency,
+            is_active,
         } = self;
 
         CurrencyFilter {
             id: id.map(EqualFilter::from),
             is_home_currency,
+            is_active,
         }
     }
 }
--- a/server/repository/src/db_diesel/currency_row.rs
+++ b/server/repository/src/db_diesel/currency_row.rs
@@ -13,6 +13,7 @@ table! {
         code -> Text,
         is_home_currency -> Bool,
         date_updated -> Nullable<Date>,
+        is_active -> Bool,
     }
 }
 
@@ -25,6 +26,7 @@ pub struct CurrencyRow {
     pub code: String,
     pub is_home_currency: bool,
     pub date_updated: Option<NaiveDate>,
+    pub is_active: bool,
 }
 
 pub struct CurrencyRowRepository<'a> {
@@ -67,7 +69,8 @@ impl<'a> CurrencyRowRepository<'a> {
     }
 
     pub fn delete(&self, currency_id: &str) -> Result<(), RepositoryError> {
-        diesel::delete(currency_dsl::currency.filter(currency_dsl::id.eq(currency_id)))
+        diesel::update(currency_dsl::currency.filter(currency_dsl::id.eq(currency_id)))
+            .set(currency_dsl::is_active.eq(false))
             .execute(&self.connection.connection)?;
         Ok(())
     }
--- a/server/repository/src/db_diesel/currency_row.rs
+++ b/server/repository/src/db_diesel/currency_row.rs
@@ -1,5 +1,5 @@
 use super::{currency_row::currency::dsl as currency_dsl, StorageConnection};
-use crate::Upsert;
+use crate::{Delete, Upsert};
 
 use crate::repository_error::RepositoryError;
 
@@ -73,6 +73,24 @@ impl<'a> CurrencyRowRepository<'a> {
             .set(currency_dsl::is_active.eq(false))
             .execute(&self.connection.connection)?;
         Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct CurrencyRowDelete(pub String);
+impl Delete for CurrencyRowDelete {
+    fn delete(&self, con: &StorageConnection) -> Result<(), RepositoryError> {
+        CurrencyRowRepository::new(con).delete(&self.0)
+    }
+
+    fn assert_deleted(&self, con: &StorageConnection) {
+        assert!(matches!(
+            CurrencyRowRepository::new(con).find_one_by_id(&self.0),
+            Ok(Some(CurrencyRow {
+                is_active: false,
+                ..
+            })) | Ok(None)
+        ));
     }
 }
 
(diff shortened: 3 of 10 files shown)


clemens-msupply (changes requested):
Can I propose the following approach to soft delete items? The following assumes that we don't use the soft-deleted items in omSupply but only keep them for debugging reason. For this reason we can hide the soft delete fully in the repository.

1) instead of an is_active column use a `deletion_datetime` (ideally using the same datetime as the integration_datetime)
2) In all query methods add a hardcoded filter for `deletion_datetime IS NOT NULL`

@Chris-Petty , @andreievg , @jmbrunskill any thoughts?

clemens-msupply:
There is apparently also the case where mSupply central incorrectly deletes an entity that is still used in other rows. Then yes, we need to load the soft-deleted entries when loading the historic rows but hide the soft deleted entities when creating new rows.

Note: if possible we should handle these case as well during sync, e.g. don't fail when an optional reference is missing. Recently did this for some barcode and location references that where not in mSupply anymore.

Btw. soft deletes only help as long as the remote site doesn't resync...

Chris-Petty:
> There is apparently also the case where mSupply central incorrectly deletes an entity that is still used in other rows. Then yes, we need to load the soft-deleted entries when loading the historic rows but hide the soft deleted entities when creating new rows.
> 

Yup central naively only checks its data for references to the record to delete; could be remote sites that reference the deleted record that haven't synced.

> Note: if possible we should handle these case as well during sync, e.g. don't fail when an optional reference is missing. Recently did this for some barcode and location references that where not in mSupply anymore.
> 

Yea that could be good... though it'd be pretty dire to lose the currency record if used. It's really an edge case though I believe it's very rare to add/remove currencies beyond the first configuration of a system.

> Btw. soft deletes only help as long as the remote site doesn't resync...

Ew yea nasty. For lack of a nice orchestration system, I wonder if sync v5 should allow a remote site to push back a central record upsert if it receives a "delete" that it cannot process? Feels kinda dirty, but also might work pretty easily? Mull it over 😁

server/repository/src/db_diesel/currency.rs
```diff
@@ -92,6 +93,12 @@ fn create_filtered_query(filter: Option<CurrencyFilter>) -> BoxedLogQuery {
             Some(false) => query.filter(currency_dsl::is_home_currency.eq(false)),
             None => query,
         };
+
+        query = match filter.is_active {
+            Some(true) => query.filter(currency_dsl::is_active.eq(true)),
+            Some(false) => query.filter(currency_dsl::is_active.eq(false)),
+            None => query,
+        };
```
Chris-Petty, line 101: Fun rust having 500 ways to do one thing: 

```rust
if let Some(tis_active) = filter.is_active {
    query = query.filter(currency_dsl::is_active.eq(tis_active))
}
```

(`tis_active` because for some reason I had imported the field struct `is_active` from clinician and it was colliding - LSP autocomplete throwing me curve balls lol. Anyway I like `tis` lol.)

Chris-Petty (approved):
Had some chats and as much as I kinda like datetime "deleted" fields for holding more data, I've recently wondered what the indexing impact is 🤔? Probably "negligible" given this is a table that'll have at MOST a couple hundred records for every known currency unless we start accepting crypto LOL. In most instances it'll have 1-5 records.

Ultimately we have several active/deleted/whatever fields that are boolean so consistency is nice - intuitive for mSupply devs and generally in industry too I think?

roxy-dao:
@clemens-msupply are you happy with this? Can I merge?

jmbrunskill:
> Adds an `is_active` field (not to be confused with OG's is_active field which isn't being used right now) to currency for soft deletion of currencies.

This worries me a little bit, especially with Chris' comment that `we have several active/deleted/whatever fields that are boolean so consistency is nice`.
Are we saying that there is a field in Legacy mSupply with the same name, but will hold a different value to what we have in open-mSupply? 
If so I think that's an unnecessary confusion, and I vote strongly for `is_deleted` or `deleted_datetime`. 
If we have `is_active` in legacy that is aligned with the `is_active` in OMS then I think the current naming is fine.

To me `is_active` has a subtly different meaning to `is_deleted`, for example you might want to just disable a currency temporarily, but might want to add it back in again soon. That feels like a good example of `is_active`, if you accidentally added bitcoin and want to remove it permanently then `is_deleted` or `deleted_datetime` would be more descriptive IMHO...

>Can I propose the following approach to soft delete items? The following assumes that we don't use the soft-deleted items in omSupply but only keep them for debugging reason. For this reason we can hide the soft delete fully in the repository.

> - instead of an is_active column use a deletion_datetime (ideally using the same datetime as the integration_datetime)

I'm a fan of this approach, I've been using `deleted_datetime` though instead of `deletion_datetime`? Is that naming already in use somewhere? I think being consistent is a good idea! 
`deleted_datetime` is what we've used throughout Health Supply Hub, Notify, and our recent assets work, so keen to make that the standard name if possible...

https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/migrations/v2_00_00/sync_file_reference.rs#L22

> - In all query methods add a hardcoded filter for deletion_datetime IS NOT NULL

We've taken this approach with asset_log_reasons:
https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/db_diesel/assets/asset_log_reason.rs#L140

I think it maybe could just be a default filter though, for example you might want to find deleted items for some view?
I think if you don't provide an `is_deleted` filter, the query method should only retrieve items that aren't deleted. I didn't do this for sync_files but would be easy to add here...

https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/db_diesel/sync_file_reference.rs#L103

I totally agree that should be the default rather than having to remember to filter out deleted things in every query.

roxy-dao:
I think for currencies it is good to have an `is_active` instead of `is_deleted` since the user may wish to add the currency back again at some point which will then cause multiple entries of the same currency code in the database. 

OG does have an `is_active` field for currencies which was added for Tenders if I'm not mistake. I heard it's not being used currency and I would like to hijack this field to sync it through, but that would require mSupply changes which haven't happened yet (and also need to create an issue for it), so this is the current patch for now.

clemens-msupply:
Yes good point for `is_deleted`. I would also prefer to use `deleted_datetime` think it fits in nicely with `created_datetime` at al. and since we don't really do soft deletes in mSupply it also not really inconsistent.

Chris-Petty:
> > Adds an `is_active` field (not to be confused with OG's is_active field which isn't being used right now) to currency for soft deletion of currencies.
> 
> This worries me a little bit, especially with Chris' comment that `we have several active/deleted/whatever fields that are boolean so consistency is nice`. Are we saying that there is a field in Legacy mSupply with the same name, but will hold a different value to what we have in open-mSupply? If so I think that's an unnecessary confusion, and I vote strongly for `is_deleted` or `deleted_datetime`. If we have `is_active` in legacy that is aligned with the `is_active` in OMS then I think the current naming is fine.
> 

@jmbrunskill It's unused in msupply, added as a kludge for the tender module making currencies on remote sites when syncing with rtm breaking central data ownship. Backtracked the kludge but the field once added in msupply is forever. The intent would be as roxy says, may remove a currency but want the same record if you ever reintroduced it. Must be kept if there are any FKs.

> To me `is_active` has a subtly different meaning to `is_deleted`, for example you might want to just disable a currency temporarily, but might want to add it back in again soon. That feels like a good example of `is_active`, if you accidentally added bitcoin and want to remove it permanently then `is_deleted` or `deleted_datetime` would be more descriptive IMHO...
> 

I do agree, I flagged this at some point early this year or perhaps last year that the semantics matter but I'm surrendered to msupply where you always have to check code and behaviours rather than trust a variable/field name 😅. Will be an up hill battle maintaining meaning heh. Most of our "soft deletes" are already "is_active". Fortunately changing field names in OMS is easier...

> > Can I propose the following approach to soft delete items? The following assumes that we don't use the soft-deleted items in omSupply but only keep them for debugging reason. For this reason we can hide the soft delete fully in the repository.
>   

There are cases where you'd possibly still want to get the record, for instance to show the currency code on past invoices that used a since deleted currency.

> > * instead of an is_active column use a deletion_datetime (ideally using the same datetime as the integration_datetime)
> 
> I'm a fan of this approach, I've been using `deleted_datetime` though instead of `deletion_datetime`? Is that naming already in use somewhere? I think being consistent is a good idea! `deleted_datetime` is what we've used throughout Health Supply Hub, Notify, and our recent assets work, so keen to make that the standard name if possible...
> 
> https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/migrations/v2_00_00/sync_file_reference.rs#L22
> 
> > * In all query methods add a hardcoded filter for deletion_datetime IS NOT NULL
> 
> We've taken this approach with asset_log_reasons:
> 
> https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/db_diesel/assets/asset_log_reason.rs#L140
> 
> I think it maybe could just be a default filter though, for example you might want to find deleted items for some view? I think if you don't provide an `is_deleted` filter, the query method should only retrieve items that aren't deleted. I didn't do this for sync_files but would be easy to add here...
> 
> https://github.com/msupply-foundation/open-msupply/blob/5a2444195343cabd96ef5f613acdbeadba185307/server/repository/src/db_diesel/sync_file_reference.rs#L103
> 
> I totally agree that should be the default rather than having to remember to filter out deleted things in every query.

I like filtering deletes by default 👍

clemens-msupply (approved):

PR #3633, opened 2024-04-17
Rename asset_type table to asset_catalogue_type

### LINKED ISSUE #3608: New asset_type table clashes with existing asset_type table
## What went wrong? 😲
 

There is already an asset_type table in mSupply. When trying to integrate this tables there is now a deserialization error in omSupply.

image

## Expected behaviour 🤔

As I understand it will not be a real problem, but it would be nice to not having an error in the sync_buffer. Maybe find a new table name?

fergie-nz:
Fixes #3608

# 👩🏻‍💻 What does this PR do? 
Renames asset_type to asset_catalogue_type

Name change doesn't cascade beyond repo layer. asset_catalogue_type is referred to as asset_type throughout the rest of the code base.

I had considered renaming to "asset_set" or "asset_group", but figured it would be better to have more specificity where both tables are in context.


server/repository/src/migrations/v2_00_00/assets/reference_data.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
     sql!(
         connection,
         r#"
-        CREATE TABLE asset_type (
+        CREATE TABLE asset_catalogue_type (
```
mark-prins, line 30: possibly a comment to say that there was a clash with an existing table `asset_type` and that's why this is called `asset_catalogue_type` which doesn't match the object names etc.
It's likely not going to be helpful long term, but just in case the question comes up.

mark-prins (approved):
👍 LGTM!

I'm slightly uneasy about having the table name different to the rest of the codebase.. but I'd rather have the simple name, so I think go for it.

fergie-nz:
Thanks @roxy-dao !

PR #3583, opened 2024-04-10
3118 central server config through m supply central

andreievg:
Fixes #3118

**IMPORTANT**, use [omSupply-central-server-configurations ](https://github.com/msupply-foundation/msupply/pull/14581) branch of mSupply (will update in a sec and PR to follow)

# 👩🏻‍💻 What does this PR do? 
 
* Adds global mutex containing central server configurations, please have a look at server/service/src/sync/mod.rs first
* Site info has a couple of new fields, that tell all omSupply sites if they are central server, and where it can be reached
* During sync we first call site info to figure out if current site is central and what central site url is
* Fix integration tests
* Update documentation, to explain how omSupply central site can be launched, might need to update more internal docs (to explain how omSupply central URL is exposed in omSupply remote sites, i.e. through `site_info` end point)

# 🧪 How has/should this change been tested? 

## 💌 Any notes for the reviewer?

I had a look at sync version, it's on https://github.com/msupply-foundation/msupply/blob/6609e8078cb5b88bf7086edf8128ca1c7cd88caa/Project/Sources/Methods/syncV5API.4dm#L17C1-L17C16 atm for mSupply. If mSupply changes for this work go into next mSupply release (where iMaxVersion is 4), we should be fine with compatibility, otherwise we should increment to 5 in both (omSupl and mSup)

## 📃 Documentation

_No user facing changes_

or

- [ ] Functional change 1
- [ ] Functional change 2

PR #3635, opened 2024-04-17
3294 central sync only for site

### LINKED ISSUE #3294: Improve integration process for incoming records in central sync
## The suggestion

Right now after receiving a batch of records from a remote site, we put them into the standard sync buffer.
Once the `total` record count is sent, we trigger the sync process to do the translation and integration steps.

This has 2 issues.

1. It means that every push from a remote site would trigger a central server sync, this could be a lot of unnecessary load when we have a lot of site pushing records
2. There's a risk of incorrectly integrating records, for example if a batch of 20 records has been uploaded for a remote site but there's more records that need to be sent, and the normal sync process starts, it will try to integrate these 20 records, which might result in foreign key issues.

The risks explained above are significant and we don't want to have production issues in a huge sync system in future deployments.

If done incorrectly we could make things worse :)

Sync changes always require significant effort and we might want to consider different design options such as a dedicated incoming records sync buffer? Global mutex on integration? Integration on sync trigger?

lache-melvin:
Fixes #3294

# 👩🏻‍💻 What does this PR do? 
 

Updates oms central server v6 push to only integrate records for the site it received records from, rather than triggering a full sync to OG central. 

Uses #3583 as a base, you'll need to use latest develop of mSupply to include the [configuration changes](https://github.com/msupply-foundation/msupply/pull/14581) (See those PRs for setup)

- Start UI as remote site, make a change that would sync to oms central server, e.g. add new coldchain asset
- Trigger sync
- [ ] Only buffer rows for your site get integrated 
  - [ ] See logs for oms central, full sync has not been triggered
  - [ ] Your new asset has been integrated on central - sync buffer row is integrated, asset now exists in assets table

Have added comments throughout

## 📃 Documentation

Maybe some docs to update? Will check 👀


server/server/src/sync_on_central/mod.rs
```diff
@@ -47,3 +51,13 @@ async fn push(
 
     Ok(web::Json(response))
 }
+
+#[post("/sync/site_status")]
```
lache-melvin, line 55: Maybe should be a GET request, at least it is in v5... but we had the pattern set up for passing the sync settings in the POST request body so just stuck with that?

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -126,13 +149,63 @@ pub async fn push(
...
+    if is_last_batch {
+        set_integrating(response.site_id, true);
+
+        integrate_and_translate_sync_buffer(&ctx.connection, true, None, Some(response.site_id))
+            .await
+            // TODO map to IntegrationError once implemented
+            // .map_err(Error::IntegrationError)?;
+            .map_err(|e| {
+                Error::OtherServerError(
+                    "Error integrating records: ".to_string() + e.to_string().as_str(),
```
lache-melvin, line 161: Was hoping to be able to respond with a more explicit IntegrationError here if something goes wrong in that integration step, but looks like I'd need to implement Serialize/Deserialize for the anyhow Error... which sounded complicated?

If there's actually a simple way to do that I would love to know! Otherwise went with just `OtherServerError` here...

server/service/src/sync/api_v6/mod.rs
```diff
@@ -115,6 +124,7 @@ pub struct SyncBatchV6 {
     // Including records in this batch
     pub(crate) total_records: u64,
```
lache-melvin, line 125: Haven't done anything about the total_records thing yet... not sure if it's better to rename or remove?

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -157,14 +168,25 @@ impl SynchroniserV6 {
             .map(SyncRecordV6::from)
             .collect();
 
+            if records.is_empty() {
+                // No v6 records to push, update cursor and continue to next batch
+                if let Some(last_pushed_cursor_id) = last_pushed_cursor {
+                    cursor_controller.update(connection, last_pushed_cursor_id as u64 + 1)?
+                };
```
lache-melvin, line 175: Should I put this logic in a callback? Also used below 😱

server/service/src/sync/sync_buffer.rs
```diff
@@ -62,7 +63,11 @@ impl<'a> SyncBuffer<'a> {
                 SyncBufferFilter::new()
                     .table_name(EqualFilter::equal_to(legacy_table_name))
                     .action(action.equal_to())
-                    .integration_datetime(DatetimeFilter::is_null(true)),
+                    .integration_datetime(DatetimeFilter::is_null(true))
+                    .source_site_id(match source_site_id {
+                        Some(site_id) => EqualFilter::equal_to_i32(site_id),
+                        None => EqualFilter::i32_is_null(true),
```
lache-melvin, line 69: Is there a use case for wanting to integrate all pending sync buffer rows? Because now we can only integrate EITHER for this site (null) or for a specific site id...

server/service/src/sync/sync_buffer.rs
```diff
@@ -151,20 +156,28 @@ mod test {
...
+            .get_ordered_sync_buffer_records(
+                repository::SyncBufferAction::Upsert,
+                &table_order,
+                None,
+            )
             .unwrap();
 
         assert_eq!(
-            in_referencial_order,
+            in_referential_order,
```
lache-melvin, line 168: eheh sorry little spellcheck side quest

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -1,19 +1,29 @@
...
 use repository::{ChangelogRepository, SyncBufferRowRepository};
 use util::format_error;
 
 use crate::{
     service_provider::ServiceProvider,
-    sync::{api::SyncApiV5, translations::ToSyncRecordTranslationType, CentralServerConfig},
+    sync::{
+        api::SyncApiV5,
+        api_v6::{SiteStatusCodeV6, SiteStatusV6},
+        synchroniser::integrate_and_translate_sync_buffer,
```
lache-melvin, line 11: It feels like I'm doing something with layers wrong importing the synchroniser method here to do the integration.... should it at least live in a separate file and be used in the two places??

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -75,10 +90,13 @@ pub async fn pull(
     );
     log::debug!("Sending records as central server: {:#?}", records);
 
+    let is_last_batch = total_records <= batch_size as u64;
+
     Ok(SyncBatchV6 {
         total_records,
         end_cursor,
         records,
+        is_last_batch,
```
lache-melvin, line 99: Kind of icky having to pass is_last_batch for pull, seeing as we don't use it on the other side... oh well?

server/service/src/sync/sync_on_central/mod.rs
lache-melvin: Actually also... currently the `wait_for_integration` after calling push doesn't do anything, because because I'm awaiting integration here... maybe I shouldn't be?

Before we were just triggering sync - but then we can't let the remote site know that we failed to integrate their records... is that fine? How would you do this?

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -126,13 +152,75 @@ pub async fn push(
...
+    if is_integrating {
+        sites_being_integrated.push(site_id);
+    } else {
+        sites_being_integrated.retain(|id| *id != site_id);
+    }
+}
+
+fn spawn_integration(service_provider: Arc<ServiceProvider>, site_id: i32) -> () {
+    tokio::spawn(async move {
+        let ctx = match service_provider.basic_context() {
```
lache-melvin, line 204: The compiler is sad here once I start doing anything with `StorageConnection`, as the `transaction_level` doesn't implement `Sync`... had a bit of a play but feels like I'm shooting in the dark a bit now - send help?

server/service/src/sync/central_data_synchroniser_v6.rs
andreievg: Ooh, I am afraid of indentation, but maybe in this case it would be better to do below if !records.is_empty()

server/service/src/sync/sync_buffer.rs
andreievg: I don't think there is

server/server/src/sync_on_central/mod.rs
lache-melvin: I think you can, but you're not really meant to/it's not meant to have any impact?? I was just meaning that for v5, we send the site id and other sync settings as request headers, but we have the pattern in v6 of sending them as the POST body. `site_status` sort of sounds like the kind of thing you would `get`, but I think it's better to have the consistent mechanism to send the sync settings, so will stick with the POST 🤷‍♀️

lache-melvin:
Cool ok I think this is good to go, won't merge until #3583 merged through 🚀

PR #3637, opened 2024-04-17
Import asset properties

### LINKED ISSUE #3628: Add properties to asset import
Allow the entry of properties using the import asset csv.

In the example template, append all properties which are defined in the database - wondering if name is ok? ideally a shorter text such as a code, which we don't have the database. I recommend `name` for now.

Then if a value is supplied for the property, attempt to parse to the correct type and create a property record for each.
Display an error if the supplied value cannot be parsed correctly.

### Moneyworks Jobcode 🧰

`OMSCC:AST`

mark-prins:
Fixes #3628 

# 👩🏻‍💻 What does this PR do? 
 
* Adds configured properties to the example csv
* Parses any imported property values
  * validates that int and float values are valid numbers
  * validates that the value is part of the `allowedValues` (if supplied)
  * converts true / yes to boolean
* Fixes the delete notification (was always showing an error toast)
* Updates the delete item, so that properties are deleted if they exist

Screenshot 2024-04-17 at 9 37 27 PM

This is only available on the central server. You'll need to configure a site as a central server (see Andrei's PR for details)


fergie-nz (changes requested):
Hi Mark,

There are a few things which feel like they're lacking functionality for me, though maybe for future issues.

- I'm getting no import error messages when the refrigerated volumes are failing to import - I only get errors on the solar. I'm still trying to get to the bottom of why the volumes are failing for me, but regardless I would love to see error messaging.

- There are none of the property fields in the error summary csv (which do exist in the example csv). I think this would make it harder to correct errors while retaining the originally uploaded. 

What do you think?

mark-prins:
Thanks @fergie-nz for review and retesting! is this ok to merge now?

fergie-nz (approved):
Nice!

PR #3643, opened 2024-04-18
3119 omsupply central sync steps

### LINKED ISSUE #3119: omSupply central sync step
If omSupply central is not reached we currently show an error, but allow sync to go through. For 1.7.0 i think we can hide the error or make it less obvious (technically we don't need omSpply central for 1.7.0 to work, only when pack variants are used, so don't really want to scare users with this error, or update every installation to have the omSupply central).

We also want a sync step for omSupply central pull

lache-melvin:
Fixes #3119

# 👩🏻‍💻 What does this PR do? 
 

Exposes central v6 steps on the sync page, and maps v6 errors:
Screenshot 2024-04-19 at 9 09 20 AM

Uses #3635 as a base, you'll need to use latest develop of mSupply to include the [configuration changes](https://github.com/msupply-foundation/msupply/pull/14581) (See those PRs for setup)

- In mSupply, don't add (or remove) the `This site url` for your omSupply central server
- Start remote sync
- You should see an error in the Sync UI (V6 central server not configured)
- Add the site URL
- Sync page should now show all 6 sync steps
- v6 errors also mapped:
  - stop oms central server - UI shows Unable to connect to server rather than unknown sync error...
  - (how to test parsing error?)
  - should more errors be mapped to new codes?

The totals beneath the push steps are for all changelogs still to be pushed (not filtered by v5/v6) - that filtering is done by the translators, and it's nice to leave that logic in one place.

It was discussed about changing/adding to show total SENT though - this is possibly quite helpful from a devving/troubleshooting standpoint, to know how many records were sent to which central server... not sure how we would display though (i.e. `processed 5 / 10 records, sent 2 to central v6`...)

## 📃 Documentation

- [ ] Update screenshots for synchronisation docs


client/packages/host/src/components/SyncProgress/SyncProgress.tsx
```diff
@@ -117,10 +117,12 @@ const getSteps = (
   }
 
   if (isOperational) {
+    steps.push(getStep('sync-status.push-v6', syncStatus?.pushV6));
     steps.push(getStep('sync-status.push', syncStatus?.push));
   }
 
   steps.push(getStep('sync-status.pull-central', syncStatus?.pullCentral));
+  steps.push(getStep('sync-status.pull-central-v6', syncStatus?.pullV6));
```
andreievg, line 125: We actually not going to have a "central-v6", just pull-v6 and push-v6

andreievg (approved):
This is looking great, thanks, once again easy to review:

> (how to test parsing error?)

We have unit tests somewhere for errors, not sure how to do integration tests for those, could mock http server (similar to unit tests in sync/api), combined with integration tests

> should more errors be mapped to new codes?

I think we can record any issues pps may have when this is in use, and any confusion/ambigious errors can be mapped to translated ones

andreievg (changes requested):
While doing docs, had a realisation:

* We should show an error straight away if central server is not configured (maybe create v6 as Option)
* Should only have pull and push central steps if site is remote site, lets chat about this one

lache-melvin:
@andreievg Cool have updated naming, displayed the sync steps in the right order (lol) and added `?` so sync stops on V6 errors. 

Also tried this: https://github.com/msupply-foundation/open-msupply/compare/3119-omsupply-central-sync-steps...3119-omsupply-central-sync-steps(try-create-v6-first) So can get V6 not configured error before any sync happens - is this how you would implement it?

Screenshot 2024-04-24 at 12 47 05 PM


Kind of would be cool to have a healthcheck endpoint on central server, but I guess you just get the connection error at the V6 step if you configured e.g. the wrong port?

andreievg:
> Also tried this: https://github.com/msupply-foundation/open-msupply/compare/3119-omsupply-central-sync-steps...3119-omsupply-central-sync-steps(try-create-v6-first) So can get V6 not configured error before any sync happens - is this how you would implement it?

Yep. Although i am really paranoid about making the sync method too verbose. This looks good though, and we would eventually phase it out anyways. Happy for you to merge this in too

> Kind of would be cool to have a healthcheck endpoint on central server that you could ping when you create the v6 synchroniser, but I guess its fine, you just get the connection error in the push/pull step if you configured e.g. the wrong port? Edit: Hmm could be a nice place to check for v6 sync version mismatch too??

Hmm, good point, we might have one actually, it might tell us all about the next sync (how many records incoming etc..). I think sync version similar to the 'site is integrating' should be a check on all endpoints.

andreievg (approved):
Thanks for the changes !

PR #3648, opened 2024-04-21
3502 new stocktake from expired

### LINKED ISSUE #3502: Add new stocktake 'filter' for expired in stocktake
TODO fill in (as per team discussion)

Expired within days from now (3 hours contingency for refactor, input to be stricter as OR)

CarlosNZ:
Fixes #3502 

# 👩🏻‍💻 What does this PR do?

Front-end: date-picker in New Stocktake modal to select items expiring before a given date
Back-end: add "expiringBefore" field to input filter and apply to the found stock lines

Couple of Rust-related questions -- see inline comments

# 🧪 Testing

Create a new stocktake and filter by expiry date. Check the resulting stock list in new stocktake is correct (expiry dates before selected date)


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -163,7 +163,7 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           onChange(newNum);
         }}
         onBlur={() => {
-          onChange(Number(parse(textValue ?? '')) || undefined);
+          onChange(Number(parse(textValue ?? '')) ?? undefined);
```
CarlosNZ, line 166: Not strictly part of this PR, but fixes a mistake which wipes out a `0` entry in number input. (That darned `0 == false` thing again 🙄)

server/service/src/stocktake/insert.rs
```diff
@@ -121,7 +123,19 @@ fn generate(
         Some(_) => generate_lines_with_stock(connection, store_id, &id)?,
         None => Vec::new(),
     };
-    let lines = [master_list_lines, location_lines, items_have_stock_lines].concat();
+    let expiring_items_lines = match expires_before {
+        Some(_) => {
+            generate_lines_expiring_before(connection, store_id, &id, &expires_before.unwrap())?
```
CarlosNZ, line 128: Is this an appropriate use of `unwrap()`? Is there a better way?

lache-melvin: I guess it can't panic, given you've already matched that `expires_before` is a `Some`, but we'd usually extract the inner value of a Some like this:

```rust
Some(expiry_date) => generate_lines_expiring_before(connection, store_id, &id, &expiry_date)?
```

server/graphql/stocktake/src/mutations/insert.rs
```diff
@@ -188,6 +191,7 @@ mod test {
                     location: None,
                     master_list_id: None,
                     items_have_stock: None,
+                    expiry_date: None
```
lache-melvin, line 194:
```suggestion
                    expires_before: None
```

client/packages/inventory/src/Stocktake/ListView/CreateStocktakeButton.tsx
```diff
@@ -246,6 +254,21 @@ export const CreateStocktakeButton: React.FC<{
                   }
                   label={t('label.items-with-stock')}
                 />
+                <InputWithLabelRow
```
lache-melvin, line 257: Defs minor but the other inputs appear to have a 16px box between them:

Screenshot 2024-04-22 at 12 37 08 PM

Could we either add it, or maybe better to change the parent to use gap? 🙏

lache-melvin (changes requested):
Looks sick Carl! 

The insert stocktake tests aren't compiling (just need to pass the new field) and probably good to add a new test case to check the expiry date filtering - otherwise working great 🚀

server/service/src/stocktake/insert.rs
CarlosNZ: Great, thank you: d6bc95482

🙏

CarlosNZ:
> The insert stocktake tests aren't compiling (just need to pass the new field) and probably good to add a new test case to check the expiry date filtering - otherwise working great 🚀

Cool, done: ee68a677c

server/service/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde_yaml = "0.8.24"
 schemafy = "0.6.0"
 schemafy_core = "0.6.0"
 tera = "1"
-tokio = { version = "1.17.0", features = ["macros", "sync", "time"] }
+tokio = { version = "1.17.0", features = ["macros", "sync", "time", "fs"] }
```
CarlosNZ, line 33: This is a fix for a tokio error I was getting whenever I tried to run tests. @andreievg 's suggestion.

lache-melvin (approved):
Sweeet thanks Carl!

PR #3653, opened 2024-04-22
3652 file sync pr review updates

### LINKED ISSUE #3652: File sync PR review updates
Follow up from this PR: https://github.com/msupply-foundation/open-msupply/pull/3616.

Doing in a separate issue because the PR is changing hands and because the requested/suggested changes would cause quite a bit of diff that should really be re-reviewed

* Align result of upload file with push and pull
* Simplify file upload (don't try to do to much when file is deleted)
* Consolidate file upload logic (it's in 3 places now)

andreievg:
Fixes #3652 

# 👩🏻‍💻 What does this PR do?

As per issue, mainly this PR deals with change request in #3616, to align upload API (for central <-> remote interaction) with existing sync api patterns.

* Instead of file_id being in url part, move to multipart json part
* To avoid extra condition just use single file in multipart
* Add move_temp_file to file_service and use in existing end points (fridge tag upload and upload file)
* Notice that we are leaking errors in upload fridge tag and in upload file (local), updated to log errors to console and return generic error (extracting functionality to _inner method helped with this)
* For client request to v6 api, extracted common error mapping functionality and used in upload endpoint
* Added auth to fridge_tag upload endpoint

This PR also removes the chunk functionality, to reduce logic, since it wasn't really being used. If we use it in the future either way it would require another end point for backwards compatibility, so thought it was safe to do now

# 🧪 Testing

You will need to run omSupply central and remote server

Start remote server and front end with:

yarn && yarn start-local -- -- --port 3005
cargo run

front end available on http://localhost:3005

Start omSupply central with

IS_CENTRAL_SERVER=true  APP_SERVER__PORT=2050 APP_DATABASE__DATABASE_NAME=central APP_SERVER__BASE_DIR=central_app_data cargo run
yarn start-local -- -- --env API_HOST=http://localhost:2050

(2050 in this case is + 2 of your mSupply central running on 2048)
front end available on http://localhost:3003

Initialise both sites (separate sites), make both stores use vaccine module

Add new asset on remote site, add document.
Sync this site and wait for central to receive the file
Check file is available on central

Check that file is present in central_app_data directory under static_files

Stop remote site, and wipe it's database and remove app_data directory.

Re-initialise remote site

Try to view asset document

app_data folder should now have the document


server/server/src/static_files.rs
```diff
@@ -24,15 +25,24 @@ use service::plugin::plugin_files::{PluginFileService, PluginInfo};
...
+use thiserror::Error;
+use util::format_error;
 use util::is_central_server;
-use util::sanitize_filename;
 
 use crate::authentication::validate_cookie_auth;
 use crate::middleware::limit_content_length;
 
+#[derive(Debug, MultipartForm)]
+pub(crate) struct UploadForm {
```
andreievg, line 41: this is now used for fridge tag upload and sync on central (we don't have a common place for shared things like this in server crate)

server/server/src/static_files.rs
```diff
@@ -215,7 +155,7 @@ async fn delete_sync_file(
 
 #[post("/{table_name}/{record_id}")]
 async fn upload_sync_file(
```
andreievg, line 157: I've mentioned in PR description that this is now split into two methods, easier to avoid leaking error and logging

server/server/src/static_files.rs
```diff
@@ -232,55 +172,64 @@ async fn upload_sync_file(
...
-            Ok(_) => {}
-            Err(err) => {
-                log::error!(
-                    "Error saving file reference: {} - DELETING UPLOADED FILES",
-                    err
-                );
-                // delete any files that were uploaded...
-                for file in files {
-                    // File::create is blocking operation, use threadpool
-                    web::block(|| std::fs::remove_file(file.path)).await??;
```
andreievg, line 275: This would only happen if we fail to upsert (database error) if there is database error we have a major server problem, and having a stray file shouldn't be too much of an issue

server/server/src/sync_on_central/mod.rs
```diff
@@ -99,35 +96,35 @@ async fn download_file(
     Ok(response)
 }
 
-#[put("/sync/upload_file/{file_id}")]
```
andreievg, line 102: file_id was moved to json part of the request

server/server/src/sync_on_central/mod.rs
```diff
@@ -99,35 +96,35 @@ async fn download_file(
...
+) -> actix_web::Result<impl Responder> {
+    let response = match sync_on_central::upload_file(
+        &settings,
+        &service_provider,
+        json_part.into_inner(),
+        file_part,
+    )
+    .await
+    {
+        Ok(batch) => SyncUploadFileResponseV6::Data(batch),
```
andreievg, line 125: When omSupply talk to each other, we so far used error object rather then http response codes, used the same pattern (the premise of this PR)

server/server/src/upload_fridge_tag.rs
```diff
@@ -1,34 +1,30 @@
...
+    auth_data::AuthData,
     sensor::berlinger::{read_sensor, ReadSensor},
     service_provider::ServiceProvider,
     settings::Settings,
+    static_files::{StaticFileCategory, StaticFileService},
 };
-use util::prepare_file_dir;
+use util::format_error;
 
-const TEMP_FRIDGETAG_FILE_DIR: &str = "fridge_tag";
```
andreievg, line 20: don't need special fridge_tag folder, using static file service and temporary file, which will be removed

server/server/src/upload_fridge_tag.rs
```diff
@@ -41,36 +37,27 @@ async fn upload(
     url_params: web::Query<UrlParams>,
     settings: Data<Settings>,
     service_provider: Data<ServiceProvider>,
+    auth_data: Data<AuthData>,
+    request: HttpRequest,
 ) -> HttpResponse {
-    // TODO Permissions
+    // For now, we just check that the user is authenticated
+    // In future we might want to check that the user has access to upload fridge tag
+    if let Err(_) = validate_cookie_auth(request.clone(), &auth_data) {
```
andreievg, line 45: Added auth here

server/server/src/upload_fridge_tag.rs
```diff
@@ -79,18 +66,13 @@ fn upload_fridge_tag(
...
-    let file = form.files.pop().context("Cannot find attached file")?;
-    let file_name = file.file_name.context("Filename is not specified")?;
-
-    let dir = prepare_file_dir(TEMP_FRIDGETAG_FILE_DIR, &settings.server.base_dir)?;
-
-    let new_file_path = dir.join(file_name);
+    let file_service = StaticFileService::new(&settings.server.base_dir)?;
 
-    move_file(file.file.path(), &new_file_path)?;
+    let static_file = file_service.move_temp_file(file, &StaticFileCategory::Temporary, None)?;
```
andreievg, line 71: StaticFileCategory::Temporary would mean this file will be cleaned up

server/service/src/sync/api_v6/core.rs
```diff
@@ -21,66 +21,6 @@ pub enum SyncApiV6CreatingError {
     Other(#[source] anyhow::Error),
 }
 
-async fn response_or_err(
```
andreievg, line 24: error mapping moved to mod and used here and in upload_file

server/service/src/sync/api_v6/upload_file.rs
```diff
@@ -1,131 +1,76 @@
...
-        let mut file_bytes = Vec::new();
-        file_handle.read_to_end(&mut file_bytes).map_err(|err| {
-            log::error!("Error reading file: {:#?}", err);
-            FileSyncError::StdIoError(err)
-        })?;
+        let client = Client::new();
 
-        // This is an example request structure, it just exists to give a compiler warning if the structure changes
-        // You'll need to modify the actual request using the reqwest::multipart::Parts below if there's a change here...
-        let _canary_request = SyncUploadFileRequestV6 {
```
andreievg, line 64: _canary_request removed. There is multipart to request method below with comments about mapping

server/service/src/sync/file_synchroniser.rs
```diff
@@ -11,40 +11,35 @@ use repository::{
     RepositoryError,
 };
 
-use crate::static_files::StaticFile;
+use crate::static_files::{StaticFile, StaticFileCategory};
```
andreievg, line 14: This file is refactored to use error similar to existing sync API.

Also added early return to try help with readability

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -202,76 +204,35 @@ pub async fn upload_file(
...
-    // for each uploaded file reserve a file in the static files directory, then copy the file from the temp file location
-    // Should only be 1 file, but we will loop through them all just in case we need to do some clean up..
-    for file in files {
-        let static_file_category = StaticFileCategory::SyncFile(
+    let repo = SyncFileReferenceRowRepository::new(&ctx.connection);
+    let sync_file_reference = repo
+        .find_one_by_id(&file_id)?
+        .ok_or(Error::SyncFileNotFound(file_id.clone()))?;
+
+    file_service.move_temp_file(
```
andreievg, line 221: Just using move_temp_file vs trying to receive with chunks

server/service/src/sync/file_synchroniser.rs
```diff
@@ -131,87 +126,92 @@ impl FileSynchroniser {
...
-            None => {
-                // No files to upload
+        } else {
+            // Calculate the next retry time
+
+            // if we get a 404 error it probably means the sync_file_reference hasn't been synced yet.
+            // So wait 1 minute before retrying
+            // Otherwise, do an exponential backoff
+            let retry_at = match error.source {
+                SyncApiErrorVariantV6::ParsedError(SyncParsedErrorV6::SyncFileNotFound(_)) => {
```
lache-melvin, line 187: Not a huge one, but this error is then logged by `file_sync_driver`... i.e. in remote site server logs you see:

Screenshot 2024-04-24 at 10 37 54 AM

Initially I got a bit confused (why can't the remote server find its own file??) Then realised this is actually the error from central server... Probably this will happen a lot - sync_file_reference not synced yet? Perhaps worth logging a clearer message here on remote server?

lache-melvin (approved):
Looks great to me! Code definitely feels clean and easy to follow, consistent with existing patterns. Thanks for comments throughout PR too, super helpful for context.

I haven't spent too long sitting with what you have _removed_, will be good to see James' thoughts on that!

andreievg:
Thanks for review

> I haven't spent too long sitting with what you have removed, will be good to see James' thoughts on that!

That's one area definitely worth having @jmbrunskill look through

server/service/src/sync/file_synchroniser.rs
andreievg: ooh, a little bit concerned that response Could not be parsed, i've added to a follow up issue (since a few PR's are stacked from this one)

https://github.com/msupply-foundation/open-msupply/issues/3666

server/server/src/static_files.rs
jmbrunskill: Agreed, Mark wanted this in there but I don't think it should happen.
With postgres on a separate server I guess the chance is a bit higher but I don't think it would be a major thing if there are a few extra files on disk in this situation.

server/server/src/upload_fridge_tag.rs
jmbrunskill: FYI - clean up is on file download request right now. So if someone was to upload lots of fridge tags and not download any report the clean up wouldn't be triggered but not such a big deal I think...

jmbrunskill: Should probably but in a refactor issue at some point to do file clean up in a background process.

server/service/src/static_files.rs
```diff
@@ -56,6 +65,25 @@ impl StaticFileService {
...
+        temp_file: TempFile,
+        category: &StaticFileCategory,
+        file_id: Option<String>,
+    ) -> anyhow::Result<StaticFile> {
+        let file_name = temp_file.file_name.context("Filename not provided")?;
+        let sanitized_filename = sanitize_filename(file_name);
+
+        let static_file = self.reserve_file(&sanitized_filename, &category, file_id)?;
+        let destination = Path::new(&static_file.path);
+        // Is this blocking ? If it is it a problem ?
```
jmbrunskill, line 81: Yes I think it is blocking but I don't think it will be a big deal with the filesizes we're dealing with...

server/service/src/sync/api_v6/upload_file.rs
```diff
@@ -1,131 +1,76 @@
...
         } = self;
 
-        let route = "upload_file/";
-        let url = url
-            .join(route)
-            .unwrap()
-            .join(&sync_file_reference_row.id)
-            .unwrap(); // Unwrap should be safe here, because we know that sync_file.id is a UUID
+        let route = "upload_file";
+        let url = url.join(route).unwrap(); // Unwrap should be safe here, because we know that sync_file.id is a UUID
```
jmbrunskill, line 21:
```suggestion
        let url = url.join(route).unwrap(); // Unwrap is safe here as the route is always `upload_file`
```

server/service/src/sync/file_synchroniser.rs
jmbrunskill: Totally agree with this comment, would be good to have this cleaned up.
I'm slightly tempted to just create a dummy `sync file reference` and just allow uploading without syncing first?
Similar to what we do with temperture_log/temperature_breach records that rely on each other...
Now that we're authenticating properly I think that would be ok?

jmbrunskill (approved):
Looks good to me Andrei.

I made a quick fix to get fridge tags working here:
[cd3f21a](https://github.com/msupply-foundation/open-msupply/pull/3653/commits/cd3f21acc7604f6a410df4c5acd00a5af573e109)

We might want to change from `files` to `file` in our requests as well, rather than using the serde `rename` process here?

Keen to see that missing sync file error handled in the follow up issue to make sure we don't confuse people looking in the logs...

server/server/src/upload_fridge_tag.rs
andreievg: Agree, i'll just add it to carry over to not forget to insert this. https://github.com/msupply-foundation/open-msupply/issues/3666

andreievg:
Thanks @jmbrunskill, sorry i haven't actually checked fridge tag changes (forgot, after checking upload and download)

> We might want to change from files to file in our requests as well, rather than using the serde rename process here?

I didn't actually check front end code, i thought for some reason files would automatically go into 'files' (since i had an error uploading). Do you know if it's kind of a standard that we should keep ? Added to cary over https://github.com/msupply-foundation/open-msupply/issues/3666

andreievg:
Thank you very much for review @jmbrunskill @lache-melvin

Issue #78, opened 2022-05-31
Diesel async


andreievg:
Currently we use synchronous services and synchronous diesel operations, this would be problematic if we have more than `number of core` concurrent requests (they will start being blocking). Also tokio runtime is kind of tailored for 'tasks' running on thread vs one operation running in thread. Diesel is working on making official async library (but I don't think pool connection is implemented for it yet).

There is a way to run diesel operations somewhat asynchronously (with spawn_blocking, which puts the task on a worker thread pool). An example prototype of this was made, to show how we can use async now, with also minimum changes when we do change official diesel async.

```rust

type DBConnection = PgConnection;
type ConnectionPool = Pool<ConnectionManager<DBConnection>>;
type PooledConnection = r2d2::PooledConnection<ConnectionManager<DBConnection>>;

pub struct StorageConnection {
    pub connection: Arc<Mutex<PooledConnection>>,
}

#[derive(Debug)]
pub enum TransactionError<E> {
    FailedToStart(Error),
    FailedToCommit(Error),
    FailedToRollback(Error),
    Inner(E),
}

impl StorageConnection {
    pub async fn con<T, E, F>(&self, f: F) -> Result<T, E>
    where
        F: FnOnce(&PooledConnection) -> Result<T, E> + Send + 'static,
        T: Send + 'static,
        E: Send + 'static,
    {
        let connection = self.connection.clone();
        tokio::task::spawn_blocking(move || {
            let connection = connection.lock().unwrap();
            f(&connection)
        }
        .await
        .unwrap()
    }

    pub async fn transaction<'a, T, E, F, Fut>(&self, f: F) -> Result<T, TransactionError<E>>
    where
        F: FnOnce() -> Fut,
        Fut: Future<Output = Result<T, E>>,
    {
        { // Scope. here makes sure we release mutex guard
            let connection = self.connection.lock().unwrap();
            let transaction_manager = connection.transaction_manager();
            transaction_manager
                .begin_transaction(connection.deref())
                .map_err(|error| TransactionError::FailedToStart(error))?;
        }

        let f_result = f().await;

        {  // Scope. here makes sure we release mutex guard
            let connection: &PooledConnection = &self.connection.lock().unwrap();
            let transaction_manager = connection.transaction_manager();
            match f_result {
                Ok(value) => {
                    transaction_manager
                        .commit_transaction(connection)
                        .map_err(|error| TransactionError::FailedToCommit(error))?;
                    Ok(value)
                }
[...]

PR #3654, opened 2024-04-22
Update diesel

### LINKED ISSUE #2881: Update to Diesel 2.0
As per title. Current diesel version is `1.4.7` but we want the latest version which is `2.1.4`. This will require dependancy crates to be updated as well: 
- diesel_derive-enum: `1.1.1` -> `2.1.0`
- diesel_migrations: `1.4.0` -> `2.1.0`

**These are all major updates, so lots to change and many files will be touched**

Changelog for diesel can be found [here](https://diesel.rs/changelog.html). Current progress so far has just been renaming/changing macros, etc. 

### Additional context 💌
Progress can be found in [this branch](https://github.com/msupply-foundation/open-msupply/tree/update-diesel).

clemens-msupply:
Fixes #2881

# 👩🏻‍💻 What does this PR do?

This work is based on the initial work done by @roxy-dao which already fixed a great deal of the Diesel 2.0 migration issues.

I tried to minimize the changes outside the repository layer. For example, there are further ideas how to "asyncify" services/repositories: https://github.com/msupply-foundation/open-msupply/issues/78 which can be implemented in the next iteration.

## Mut connections
Diesel 2.0 requires `mut` database connections. To solve this problem the Diesel DB connection is wrapped in a Mutex in our StorageConnection wrapper. This means there are no changes in the service layer regarding the `mut` connection. In the repo layer the connection can now be accessed like this:
```
self.connection.lock().connection()
```
instead of
```
&self.connection.connection
```
This might look a bit cumbersome, please see my inline comment for more background... Short version is, this solution was more search and replace friendly and given that there might be more refactoring coming I opted for the "easy" solution.

## Postgres enums
An annoying issue for the pg features was that enum types must now be present in the database. And furthermore, the Rust enum name must match the pg type, e.g. `ItemType` maps to `item_type` in the db. For this reason some new types have been created and some enums renamed. On the plus side Rust types and pg types are now consistent.
However, this is a bit fragile; wrong naming or missing types will result in runtime errors.
Note, there is an option to keep the existing enum Rust name, but it required to manually create the ItemTypeMapping struct which is usually done by the macro.

## Summary of changes:
- column macros like `#[column_name = "type_"]` -> `#[diesel(column_name = type_)]`
- table names macros like `#[table_name = "serialize_helper"]` become `#[diesel(table_name = serialize_helper)]`
- type macros like  `#[sql_type = "Text"]` become: `#[diesel(sql_type = Text)]`
- Postgres enum types must exist in DB
- Rust enum type name must match Postgres enum type name

# 🧪 Testing

- run tests/integration_tests
- having a good look at the enum changes would be good, especially for change to currently develop featured (e.g. if enum def matches in DB and Rust code)

(I had problem with two integration_tests, but they seem to be unrelated, still looking into it...)

# Misc
I was really tempted to avoid the Mutex since I think this could still be done. However, this will add loads of `mut` keywords everywhere. So I don't think its worth it just to save a few ns to access a (in comparison) much slower database.


server/repository/src/migrations/v2_00_00/pg_enums.rs
```diff
@@ -0,0 +1,109 @@
+use crate::migrations::*;
+
+#[cfg(not(feature = "postgres"))]
+pub(crate) fn migrate(_: &StorageConnection) -> anyhow::Result<()> {
```
clemens-msupply, line 4: This adds all the missing pg enum types

server/repository/src/db_diesel/temperature_chart/temperature_chart_row.rs
```diff
@@ -1,220 +1,391 @@
...
+        impl AppearsInFromClause<diesel::query_builder::Only<super::table>> for $column_name {
+            type Count = Once;
+        }
+        #[cfg(feature = "postgres")]
+        impl SelectableExpression<diesel::query_builder::Only<super::table>> for $column_name {}
+    };
+}
+
+#[allow(unused_imports, dead_code, unreachable_pub)]
+pub mod temperature_chart {
```
clemens-msupply, line 113: Didn't figure out how to get the previous version working. But really, I don't think there fundamental differences. I started from the generated macro and worked backwards as described in the existing readme.

server/repository/src/db_diesel/temperature_chart/temperature_chart_row.rs
```diff
@@ -1,220 +1,391 @@
...
+            type Output = diesel::expression::is_contained_in_group_by::No;
+        }
+        impl diesel::expression::IsContainedInGroupBy<BreachIds> for TemperatureLogId {
+            type Output = diesel::expression::is_contained_in_group_by::No;
+        }
+        impl diesel::expression::IsContainedInGroupBy<SensorId> for BreachIds {
+            type Output = diesel::expression::is_contained_in_group_by::No;
+        }
+        impl diesel::expression::IsContainedInGroupBy<BreachIds> for SensorId {
+            type Output = diesel::expression::is_contained_in_group_by::No;
```
clemens-msupply, line 369: All these impl could probably be generated by a macro...

server/repository/src/db_diesel/storage_connection.rs
```diff
@@ -19,16 +17,60 @@ const BEGIN_TRANSACTION_STATEMENT: &str = "BEGIN IMMEDIATE;";
...
+    }
+
     /// Current level of nested transaction.
     /// For example:
     /// 0 => no transaction
     /// 1 => in transaction
     /// 2 => 1st nested transaction
     /// 3 => 2nd nested transaction
-    pub transaction_level: Cell<i32>,
+    pub fn transaction_level<E>(&mut self) -> Result<i32, TransactionError<E>> {
```
clemens-msupply, line 43: Don't know if it is new but there was is a way to get the transaction level from the diesel transaction, using this now...

server/repository/src/db_diesel/storage_connection.rs
```diff
@@ -70,81 +112,15 @@ impl From<TransactionError<RepositoryError>> for RepositoryError {
 impl StorageConnection {
     pub fn new(connection: DBConnection) -> StorageConnection {
         StorageConnection {
-            connection,
-            transaction_level: Cell::new(0),
-        }
-    }
-    /// Executes operations in transaction. A new transaction is only started if not already in a
-    /// transaction.
-    pub async fn transaction<'a, T, E, F, Fut>(&'a self, f: F) -> Result<T, TransactionError<E>>
```
clemens-msupply, line 79: The async version was not needed, and I think wasn't working when used in async code anyway (connection is not Sync...) so I removed it...

server/repository/src/db_diesel/number_row.rs
```diff
@@ -118,8 +118,12 @@ const NUMBER_INSERT_QUERY: &str =
...
 
 // feature postgres
-// We need to use the ON CONFLICT DO NOTHING Clause for postgres just in case 2 threads insert at the same time (SQLite <on disk> does not need this as it only allows a single write transaction at a time).
-// Without this postgres will throw a unique constraint violation error and rollback the transaction, which is hard to recover from, instead we just ignore the error and check if it returned a value
+// We need to use the ON CONFLICT DO NOTHING Clause for postgres just in case 2 threads insert at
+// the same time (SQLite <on disk> does not need this as it only allows a single write transaction
+// at a time).
+// Without this postgres will throw a unique constraint violation error and rollback the
+// transaction, which is hard to recover from, instead we just ignore the error and check if it
+// returned a value.
```
clemens-msupply, line 126: Just formatting the comment to fit onto the screen...

server/repository/src/db_diesel/number_row.rs
```diff
@@ -167,14 +171,13 @@ impl<'a> NumberRowRepository<'a> {
                     .bind::<Text, _>(store_id)
                     .bind::<Text, _>(r#type.to_string());
 
-                match insert_query.get_result::<NextNumber>(&self.connection.connection) {
+                let mut guard = self.connection.lock();
+                match insert_query.get_result::<NextNumber>(guard.connection()) {
```
clemens-msupply, line 175: This is where I got a deadlock when locking in the `get_result` and later in the Err(NotFound) branch...

server/repository/src/db_diesel/storage_connection.rs
```diff
@@ -19,16 +17,56 @@ const BEGIN_TRANSACTION_STATEMENT: &str = "BEGIN IMMEDIATE;";
...
-pub struct StorageConnection {
-    pub connection: DBConnection,
-    // pub arc_connection: Arc<DBConnection>,
+/// Helper class to avoid deref_mut() calls, which would require to import DerefMut everywhere we
+/// want to use a connection.
+/// For example, with out it it would look like:
+/// let con: &mut DBConnection = LockedConnection.inner.lock().unwrap().deref_mut();
+pub struct LockedConnection<'a> {
+    raw_connection: MutexGuard<'a, DBConnection>,
+}
```
clemens-msupply, line 26: Main reason for this LockedConnection was to avoid this deref_mut() problem as described in the comment, i.e. didn't want to manually import DerefMut everywhere.

server/repository/src/db_diesel/program_requisition/program_supplier.rs
```diff
@@ -75,10 +75,16 @@ impl<'a> ProgramSupplierRepository<'a> {
...
+            name_link_dsl::name_link::all_columns()
+                .nullable()
+                .assume_not_null(),
+            name_store_join_dsl::name_store_join::all_columns()
+                .nullable()
+                .assume_not_null(),
+            store_dsl::store::all_columns().nullable().assume_not_null(),
+            program_dsl::program::all_columns()
+                .nullable()
+                .assume_not_null(),
```
clemens-msupply, line 87: Also, a notable change...

server/repository/src/db_diesel/temperature_chart/temperature_chart.rs
```diff
@@ -23,30 +34,37 @@ impl<'a> TemperatureChartRepository<'a> {
...
         temperature_log_filter: Option<TemperatureLogFilter>,
     ) -> Result<Vec<TemperatureChartRow>, RepositoryError> {
-        let mut query = TemperatureChart {
+        /*let mut query = temperature TemperatureChart {
+            intervals: intervals.clone(),
+        }*/
+        let mut query = temperature_chart::table {
             intervals: intervals.clone(),
         }
+        .group_by((temperature_chart::IntervalId, temperature_chart::SensorId))
```
clemens-msupply, line 43: I had to move the group_by here (before boxing) otherwise I couldn't get it working.

clemens-msupply, line 43: Had to move the `group_by` before the into_boxed(), otherwise I couldn't get it working...

server/cli/Cargo.toml
```diff
@@ -19,11 +19,11 @@ service = { path = "../service" }
 server = { path = "../server" }
 graphql = { path = "../graphql" }
 
-anyhow.workspace = true
+anyhow = { workspace = true }
 async-graphql = { workspace = true }
 clap = { version = "3.1.8", features = ["derive"] }
 chrono = { version = "0.4.23" }
-diesel = { version = "1.4.7", default-features = false, features = ["chrono"] }
+diesel = { version = "2.1.5", default-features = false, features = ["chrono"] }
```
jmbrunskill, line 26: I see there's a `diesel = "2.1.6"` was this a deliberate choice to go with `2.1.5`?
Would we be better to just specify `2.1` which will take the latest patch when we use cargo update?

server/repository/src/migrations/v2_00_00/assets/asset_log.rs
```diff
@@ -2,6 +2,30 @@ use crate::migrations::DATETIME;
...
+            asset_id TEXT NOT NULL REFERENCES asset(id),
+            user_id TEXT NOT NULL REFERENCES user_account(id),
+            status asset_log_status,
+            reason_id TEXT REFERENCES asset_log_reason(id),
+            comment TEXT,
+            type TEXT,
+            log_datetime {DATETIME} NOT NULL
+          );
+        "#
+    )?;
```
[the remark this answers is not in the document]
jmbrunskill: I think we probably want to do all these migration in the  `pg_enums::migrate` function and put it in to the 2.1 migration? We don't want to assume this will be in for 2.0 installs right?

server/repository/src/migrations/v2_00_00/sync_file_reference.rs
```diff
@@ -1,6 +1,42 @@
...
+            direction sync_file_direction NOT NULL,
+            status sync_file_status NOT NULL,
+            error TEXT,
+            created_datetime TIMESTAMP NOT NULL, -- No modified datetime, as we don't allow updates it would break sync
+            deleted_datetime TIMESTAMP
+        );
+        "#,
+    )?;
+
+    #[cfg(not(feature = "postgres"))]
```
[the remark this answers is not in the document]
jmbrunskill: I don't like the approach here where we have 2 table definitions, which could more easily be miss matched between sqlite and postgres, Mark has this approach which I prefer...


Define the type as a const
``` 
   const DIRECTION_ENUM_TYPE: &str = if cfg!(feature = "postgres") {
        "sync_file_direction"
    } else {
        "TEXT"
    };
```

Create the enums for postgres
```
    #[cfg(feature = "postgres")]
    sql!(
        connection,
        r#"
        CREATE TYPE sync_file_status AS ENUM (
            'NEW',
            'IN_PROGRESS',
            'ERROR',
            'DONE',
            'PERMANENT_FAILURE'
        );
        CREATE TYPE sync_file_direction AS ENUM (
            'UPLOAD',
            'DOWNLOAD'
        );#,
);
```

Then we can have a single table definition like this:
```
    sql!(
        connection,
        r#"
            CREATE TABLE sync_file_reference (
                id TEXT NOT NULL PRIMARY KEY,
                table_name TEXT NOT NULL, -- Associated Table
                record_id TEXT NOT NULL, -- Associated record id
                file_name TEXT NOT NULL,
                mime_type TEXT,
                uploaded_bytes INTEGER NOT NULL DEFAULT 0,
                downloaded_bytes INTEGER NOT NULL DEFAULT 0,
                total_bytes INTEGER NOT NULL DEFAULT 0,
                retries INTEGER NOT NULL DEFAULT 0,
                retry_at TIMESTAMP,
                direction {DIRECTION_ENUM_TYPE} NOT NULL,
                status TEXT NOT NULL,
                error TEXT,
                created_datetime TIMESTAMP NOT NULL, -- No modified datetime, as we don't allow updates it would break sync
                deleted_datetime TIMESTAMP
            );
        "#,
    )?;
```

Similar for status etc

server/repository/src/db_diesel/number_row.rs
jmbrunskill: Interesting! I wonder if there's some risks we need to be considering around this locking?

jmbrunskill (changes requested):
Looks great clemens, a tonne of work!

I support the idea to get in this change and work on true async changes later if desired.

The postgres enum thing it pretty annoying, especially as you can get run time errors, but I think it's manageable as we'll be copying the current pattern.
Personally would be interested in exploring if we can have our own macro or submitting a code changes to allow storing as TEXT in both postgres & sqlite but still mapping to enum in rust. Still potential to have a deserialisation error but don't like having to maintain the enums in SQL and in rust.

Slightly worried if we could have some unexpected concurrency issues with the locking (like you did with the number thing) hopefully we'll pick up anything like that in testing/development though.

Only change needed is I think we need to move all the migration changes into 2.1, I don't think we want to roll this out as part of 2.0?

server/repository/src/db_diesel/number_row.rs
clemens-msupply: This is a bit of an odd case. Usually you just do something like:
`let result = query.get_result::<...>(self.connection.lock().connection())`
and then the lock is released as expected. But good to keep it in mind!

server/repository/src/migrations/v2_00_00/assets/asset_log.rs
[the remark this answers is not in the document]
clemens-msupply: Yes I just realized it on the weekend as well that I was too optimistic to get it into 2.0. Will move everything over.

clemens-msupply:
> The postgres enum thing it pretty annoying, especially as you can get run time errors, but I think it's manageable as we'll be copying the current pattern. Personally would be interested in exploring if we can have our own macro or submitting a code changes to allow storing as TEXT in both postgres & sqlite but still mapping to enum in rust. Still potential to have a deserialisation error but don't like having to maintain the enums in SQL and in rust.
> 

Was also wondering if it's possible to go back to the original Diesel approach to generate the macros from the DB schema (think you are doing it in HSH?). For example, another footgun I ran into before is that a macro must have the same field order as the columns in the database otherwise diesel tries to de-serialize the wrong columns into the wrong Rust fields (would have to check if still the case in diesel 2). Possible problem is that we have sqlite + postgres and custom migrations.

jmbrunskill:
> Was also wondering if it's possible to go back to the original Diesel approach to generate the macros from the DB schema (think you are doing it in HSH?). For example, another footgun I ran into before is that a macro must have the same field order as the columns in the database otherwise diesel tries to de-serialize the wrong columns into the wrong Rust fields (would have to check if still the case in diesel 2). Possible problem is that we have sqlite + postgres and custom migrations.

HSH is SQLite only so we're using TEXT fields for the enums, and diesel doesn't care...

clemens-msupply:
> > Was also wondering if it's possible to go back to the original Diesel approach to generate the macros from the DB schema (think you are doing it in HSH?). For example, another footgun I ran into before is that a macro must have the same field order as the columns in the database otherwise diesel tries to de-serialize the wrong columns into the wrong Rust fields (would have to check if still the case in diesel 2). Possible problem is that we have sqlite + postgres and custom migrations.
> 
> HSH is SQLite only so we're using TEXT fields for the enums, and diesel doesn't care...

I meant the Diesel cli (https://diesel.rs/guides/schema-in-depth.html)

server/repository/src/migrations/v2_00_00/sync_file_reference.rs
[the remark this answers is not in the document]
clemens-msupply: Done now. 

```
git difftool -d --tool=kdiff3 update-diesel:server/repository/src/migrations..develop:server/repository/src/migrations
```

Shows now changes to 2.0.0 migrations now

jmbrunskill (approved):
Thanks Clemens!

server/repository/src/db_diesel/number_row.rs
andreievg: I hope we would have some time to make it a little bit safer, when we move to async, as per the https://github.com/msupply-foundation/open-msupply/issues/78, which would make any locking happen within closure, i think compiler would complain if you try to double lock it.

Maybe for now, we have one simple rule, to just lock at the start of a repo method (and only lock once ?). Like in this example would this work: https://github.com/msupply-foundation/open-msupply/compare/update-diesel...update-diesel(Lock-once)

clemens-msupply: Yeah for new code this would make sense. However, the most common case is probably still a single `let result = query.get_result::<...>(self.connection.lock().connection())` call and then this is IMHO good enough. I am not too worried about deadlocks in the repo methods, the `match` example was quite unusually and for other possible deadlocks it should be straight forward to detect and find them.

And as you said being able to access the raw connection only in a `con(|con| {...})` closure would solve it as well.

andreievg:
> The postgres enum thing it pretty annoying, especially as you can get run time errors, but I think it's manageable as we'll be copying the current pattern.

I am a bit worried that you can just change the name of an enum variant, and forget about migrating the data (it's a bit harder to miss if it's enum, especially if we have a test for it). I think it's quite simple to add another enum variant in postgres. Also I like the idea that database schema shows what values can be stored (if they are know at compile time), rather then pointing to code, when asked what are the expected values of a field.

> Personally would be interested in exploring if we can have our own macro or submitting a code changes to allow storing as TEXT in both postgres & sqlite but still mapping to enum in rust. Still potential to have a deserialisation error but don't like having to maintain the enums in SQL and in rust.

We could use strum to check upserts and queries.

@clemens-msupply 
> Was also wondering if it's possible to go back to the original Diesel approach to generate the macros from the DB schema (think you are doing it in HSH?). For example, another footgun I ran into before is that a macro must have the same field order as the columns in the database otherwise diesel tries to de-serialize the wrong columns into the wrong Rust fields (would have to check if still the case in diesel 2). Possible problem is that we have sqlite + postgres and custom migrations.

Chris asked why we don't do it now. I think it was mainly the enums, field/table names (using reserve rust keywords for field/table names), views (maybe this is different in 2.0, but diesel didn't export the views, although those can be in s separate file anyways) and partitioning (nice to have the struct next to dsl it refers to). There is also the fact that diesel errors are really hard to read, I can imagine that having all of the dsl in one file and having struct definition elsewhere could cause headaches when fixing compilation errors.

andreievg (approved):
Thanks @clemens-msupply, looks like massive amount of work.
I looked at most of the code, and it's looking pretty good to me.

clemens-msupply:
I created an issue to use `diesel print-schema`:
https://github.com/msupply-foundation/open-msupply/issues/3777

PR #3657, opened 2024-04-23
3644 Fix Storybook with Feature Flags/Numeric Text fix

### LINKED ISSUE #3644: Feature flag breaks storybook
After implementing the Feature flag mechanism (in #3615), storybook is broken.

When running `yarn storybook`, we get the following error:

Image

CarlosNZ:
Fixes #3644 

# 👩🏻‍💻 What does this PR do?

Allows front-end tests to run.

Also discovered a regression in NumericTextInput which caused its test to fail, which I've fixed.

There is still one test failing -- AppFooter.test.tsx, no idea what's going on here, doesn't look like anything's changed recently, will have to wait for another issue: #3658

# 🧪 Testing

Run `yarn test` -- should all pass except for the previously mentioned AppFooter test.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -144,7 +144,7 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
 
           // DON'T use the formatValue callback here, we only want to do FULL
           // formatting onBlur
-          setTextValue(format(constrained));
+          if (!noFormatting) setTextValue(format(constrained));
           onChange(constrained);
         }}
         onKeyDown={e => {
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -141,10 +141,9 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           if (Number.isNaN(parsed)) return;
 
           const constrained = constrain(parsed, decimalLimit, min, max);
-
-          // DON'T use the formatValue callback here, we only want to do FULL
-          // formatting onBlur
-          if (!noFormatting) setTextValue(format(constrained));
+          setTextValue(
+            noFormatting ? String(constrained) : format(constrained)
+          );
           onChange(constrained);
         }}
         onKeyDown={e => {
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -142,7 +142,9 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
 
           const constrained = constrain(parsed, decimalLimit, min, max);
 
-          setTextValue(formatValue(constrained));
+          // DON'T use the formatValue callback here, we only want to do FULL
+          // formatting onBlur
+          setTextValue(format(constrained));
           onChange(constrained);
         }}
         onKeyDown={e => {
--- a/client/packages/config/src/config.ts
+++ b/client/packages/config/src/config.ts
@@ -35,7 +35,10 @@ export const Environment = {
   PRINT_LABEL_TEST: `${apiHost}/print/label-test`,
 
   // Feature flags
-  FEATURE_INVENTORY_ADJUSTMENTS,
+  FEATURE_INVENTORY_ADJUSTMENTS:
+    typeof FEATURE_INVENTORY_ADJUSTMENTS === 'undefined'
+      ? false
+      : FEATURE_INVENTORY_ADJUSTMENTS,
 };
 
 export default Environment;
--- a/client/packages/common/src/ui/components/portals/AppFooter/AppFooter.test.tsx
+++ b/client/packages/common/src/ui/components/portals/AppFooter/AppFooter.test.tsx
@@ -1,13 +1,14 @@
 import React, { FC } from 'react';
 import { fireEvent, render } from '@testing-library/react';
 import { AppFooterPortal, AppFooter } from './AppFooter';
+import { TestingProvider } from '@common/utils';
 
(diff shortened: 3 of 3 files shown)


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -142,7 +142,9 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
 
           const constrained = constrain(parsed, decimalLimit, min, max);
 
-          setTextValue(formatValue(constrained));
+          // DON'T use the formatValue callback here, we only want to do FULL
+          // formatting onBlur
+          setTextValue(format(constrained));
```
CarlosNZ, line 147: Someone had changed this with the `formatValue` callback was created. Have added comment here to prevent this happening again.

client/packages/config/src/config.ts
```diff
@@ -35,7 +35,10 @@ export const Environment = {
   PRINT_LABEL_TEST: `${apiHost}/print/label-test`,
 
   // Feature flags
-  FEATURE_INVENTORY_ADJUSTMENTS,
+  FEATURE_INVENTORY_ADJUSTMENTS:
+    typeof FEATURE_INVENTORY_ADJUSTMENTS === 'undefined'
+      ? false
+      : FEATURE_INVENTORY_ADJUSTMENTS,
```
CarlosNZ, line 41: Allows a default value to be applied to the feature flag property if it is undefined (ie properly "not defined" , not just has a value of "undefined")

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
mark-prins: There's a good reason for the change!
This is [the PR](https://github.com/msupply-foundation/open-msupply/pull/3322/files#diff-db59deaccdff071a56ae20033716d4aa22e8618c6b082e9a65adf640d55541c6)

The issue is that the NumericTextInput is forcing a numeric formatting - and this is not desired when formatting a port number (e.g. `9100` which needs to be constrained to a number, but not formatted to be `9,100` which gives a different meaning)

mark-prins:
The PR is doing more than the issue requires of it - if you split this out and fix storybook, then I can approve that. The changes to the test suite are a side concern. 

To that end: the testing section should address the original issue i.e. Run `yarn storybook`

mark-prins (changes requested):
Re introduces a functionality problem with the numeric input.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
CarlosNZ: I didn't remove the `noFormatting` option -- I just prevented it from being updated every keystroke and keeping in `onBlur`. Because otherwise, when you have a fixed decimal value (`decimalMin` prop), as soon as you enter a number, it get re-formatted to the "full" version (i.e `5` becomes `5.00`), which makes adding additional characters very difficult (and fails the tests).

client/packages/config/src/config.ts
mark-prins: This allows passing of strange values to the env var, would it be more reliable to just wrap in `FEATURE_INVENTORY_ADJUSTMENTS: Boolean(FEATURE_INVENTORY_ADJUSTMENTS)` - that would always return a boolean value. 

or truthy? with `!!FEATURE_INVENTORY_ADJUSTMENTS`

otherwise, to keep the current but simplify:

```
FEATURE_INVENTORY_ADJUSTMENTS: FEATURE_INVENTORY_ADJUSTMENTS ?? false;
```

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
CarlosNZ: Okay, I think this solves both problems -- i.e. allows unformatted entry, but fixes the problem introduced by your original change: a2d122a33

client/packages/config/src/config.ts
CarlosNZ: We tried that initially, but it doesn't work. The problem is because of the `declare const FEATURE_INVENTORY_ADJUSTMENTS: boolean;` at the top -- it tells the compiler that `FEATURE_INVENTORY_ADJUSTMENTS` definitely exists. But when it doesn't (i.e. running storybook), it freaks out when referring to a non-existing value.

I just copied your solution earlier in the file (for `API_HOST`). We can check the type of a non-existing value, but any reference to the value itself doesn't work.

Another option is just to add the env value to the storybook webpack, but we didn't want to have to maintain it in two different places.

I'd like to have a more standardised way of doing this, but this is the quickest fix to the current way env vars are being accessed. Maybe we could use [dotenv](https://www.npmjs.com/package/dotenv) or something to have simpler access to environment variables, but that would require more discussion.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
mark-prins: yes, but by adding the call to `format` back in, you get this:

image

because `format` applies the formatting as I type. True, it is cleared on blur, but it is disconcerting to use the input.

Can you instead please change the line to this:

```
setTextValue(noFormatting ? String(constrained) : format(constrained));
```

the issue is that format is called with no parameters on keypress, but is called with params on blur, so there's differing behaviour.

CarlosNZ: Okay, I think you missed the last comment which fixed both, but I've changed it again to your suggestion as it's a bit more thorough (and closer to original behaviour): 85ca54352

mark-prins (approved):
👍  cool - thanks @CarlosNZ 
Looking good. I've also taken the liberty of pushing a fix for the AppFooter test - the introduction of `useTranslation` and `useAuth` hook usage requires the `TestingProvider` to be used

CarlosNZ:
> Looking good. I've also taken the liberty of pushing a fix for the AppFooter test - the introduction of `useTranslation` and `useAuth` hook usage requires the `TestingProvider` to be used

Great, thanks @mark-prins

PR #3664, opened 2024-04-23
Performance improvements

### LINKED ISSUE #3066: Performance improvements

After reviewing the site through [lighthouse](https://developers.google.com/web/tools/lighthouse/) there were some quick wins, which I've implemented in a [branch](https://github.com/msupply-foundation/open-msupply/tree/performance)

- [ ] Reduce the number of elements on the page : the example given is a refactor of the DataTable cell which reduced the stocktake detail page from 2100 to 1600 elements. Need to check if the error state would work ( I think it won't ) and replace with an error cell for the single usage of the error state
- [ ] add the viewport element and test to see what sort of impact this has
- [ ] review download size and the effect of tree shaking in webpack
- [ ] cache the static files in actix_web : the change made satisfied lighthouse, but I didn't see cached responses in the dev tools network panel and would like to test if caching is in fact working. test caching of translation files ( separate process, so I think they are ok )

A slower site? A worse score on lighthouse?

mark-prins:
Fixes #3066

# 👩🏻‍💻 What does this PR do?

Thought that it would be good to get these improvements in before they're too stale.

* Reduce the DataTable element count
* Add the viewport element
* Implement caching in actix_web
* Implement the translation caching, as per https://github.com/msupply-foundation/health-supply-hub/pull/762

It lifts the lighthouse score, hopefully improves performance too. Now looking at this for the dashboard:

Screenshot 2024-04-23 at 6 03 26 PM

Essentially checking that the site is running correctly. 

# 🧪 Testing

- [ ] You can additionally run lighthouse, 
- [ ] and check that files are cached
- [ ] Any changes to locales should be reflected when developing
- [ ] Error states are displayed correctly on datatable cells

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/server/src/serve_frontend.rs
+++ b/server/server/src/serve_frontend.rs
@@ -14,18 +14,32 @@ use rust_embed::RustEmbed;
 struct Asset;
 
 const INDEX: &'static str = "index.html";
-const CACHE_MAX_AGE: u32 = 60 * 60 * 24; // 1 day
+const CACHE_MAX_AGE: u32 = 365 * 60 * 60 * 24; // 1 year
 
 // https://github.com/pyrossh/rust-embed/blob/master/examples/actix.rs
 fn serve_frontend(path: &str) -> HttpResponse {
     if let Some(content) = Asset::get(path) {
+        let cache_control = if path == "index.html" {
+            // The index and config files can change so we don't want to cache them
+            // The other files are generally static and can be cached
+            // Technically the config.js shouldn't change either but if it did we'd want pick it up immediately.
+            header::CacheControl(vec![header::CacheDirective::NoCache])
+        } else if path.starts_with("locales/") {
+            // These are the translation json files, in the typescript code they are cached in local storage and invalidated after a yarn build
+            // So we don't want to cache them here...
+            header::CacheControl(vec![header::CacheDirective::NoCache])
+        } else {
+            // Cache everything else for 1 year
+            header::CacheControl(vec![
+                header::CacheDirective::Public,
+                header::CacheDirective::MaxAge(CACHE_MAX_AGE),
+            ])
+        };
+
         return HttpResponse::Ok()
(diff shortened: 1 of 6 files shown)


server/server/src/serve_frontend.rs
```diff
@@ -10,13 +13,33 @@ use rust_embed::RustEmbed;
...
     if let Some(content) = Asset::get(path) {
+        let cache_control = if path == "index.html" {
+            // The index and config files can change so we don't want to cache them
+            // The other files are generally static and can be cached
+            // Technically the config.js shouldn't change either but if it did we'd want pick it up immediately.
+            header::CacheControl(vec![header::CacheDirective::NoCache])
+        } else if path.starts_with("locales/") {
+            // These are the translation json files, in the typescript code they are cached in local storage and invalidated after a yarn build
+            // So we don't want to cache them here...
+            header::CacheControl(vec![header::CacheDirective::NoCache])
```
clemens-msupply, line 30: In what scenario are they changing? Aren't all frontend files bundled into the server binary and don't change at all? And when doing dev we run the frontend separately, and they are not coming from the server?

jmbrunskill: In the old code the files are cached by the user's browser, so if you're connecting via the web browser to the server, and there's a server upgrade, the browser won't re-fetch the translation files. More of a problem for Health Supply Hub that for OMS as generally we're using android or a electron but doesn't hurt to fix :)

jmbrunskill (approved):
Looks good to me, not sure if we're using the `isError` stuff, if we find an error later we can fix it then, or maybe we should just remove it?

mark-prins:
ah, I looked harder. The error on cells is used for these columns in the stocktake:
* snapshot packs - when the `SnapshotCountCurrentCountMismatch` error is returned by api
* counted packs - when the `StockLineReducedBelowZero` error is returned by api

both in the list view and the edit modal table. Have checked and they look fine if I trigger the error

PR #3671, opened 2024-04-24
Move requisition create shipment button

### LINKED ISSUE #2787: Requisitions: how to prevent users from finalising requisition when trying to create shipment

Most people wants to click on `Confirm Finalised` once they're done supplying stock cause it feels like the next logical step. However, they are finalising the requisition that is no longer editable once finalised. 

Note: even during the Cambodia Demo, our demoer choose to do that so that does not only apply to end-users :-)

- Since the user seems to want to click on the `Status` button to create a shipment, suggestion would to have a "Create Shipment" button instead. 
- By clicking on `Create Shipment`, same behaviour as the former `Create Shipment` button (top right corner) that can be removed (or moved I guess) but well no need for 2 buttons. 
- In the background (as in not in the UI), requisition status would change from `new` to `confirmed` (see notes for alternatives). 
- if requisition partially supplied, the create button should still appear as the next default "Status". The Finalised status can still be present in the DDL 
- optional: when requisition is fully supplied, then the status "Finalised" is suggested as the next logical status. 

Notes: 
- if we have to keep the format: `Confirm  [Status]` then not sure if this would work (@mark-prins ?). Suggestion: Confirm supplied ? Confirm Processed ? Confirm Confirmed (NO!) ? None of the suggestion is as good as "Create Shipment" unfortunately. 

Move the `create shipment` button next to the `Confirm Finalised` button (right or left, personal preference for left) so that user can see both options of what to do with their requisitions. Once is shipment is created against the requisition, status can change to `Supplied`, `Processed` or `Confirmed` (NO!  NO! NO!)

### Additional context 💌

Screenshot 2024-01-22 at 5 37 28 PM

### Moneyworks Jobcode 🧰

@DhanyaHerath ?

mark-prins:
Fixes #2787

# 👩🏻‍💻 What does this PR do?

Moved the button, and made it blue to be consistent with other footer buttons.

We had a discussion about this one, the resolution that I remember was that we'd move the button.

Screenshot 2024-04-24 at 5 19 11 PM

An additional thought - what about asking the user if they'd like to finalise?

Screenshot 2024-04-24 at 5 15 26 PM

If so, should we default the checkbox to unchecked? or checked?

Currently the behaviour is exactly as it was, with the button shifted.

# 🧪 Testing

- [ ] Open a response / customer / distribution Requisition
- [ ] try the button

# 📃 Documentation

  1. The screenshots of this page need updating

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
@@ -13,7 +13,6 @@ import {
   ReportSelector,
   useReport,
 } from '@openmsupply-client/system';
-import { CreateShipmentButton } from './CreateShipmentButton';
 import { SupplyRequestedQuantityButton } from './SupplyRequestedQuantityButton';
 import { useResponse } from '../../api';
 import { JsonData } from '@openmsupply-client/programs';
@@ -35,7 +34,6 @@ export const AppBarButtonsComponent = () => {
   return (
     <AppBarButtonsPortal>
       <Grid container gap={1}>
-        <CreateShipmentButton />
         <SupplyRequestedQuantityButton />
         <ReportSelector
           context={ReportContext.Requisition}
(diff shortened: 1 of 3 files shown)


fergie-nz (approved):
Looks good!

No comment regarding finalising modal as I don't think I understand users' context well enough.

mark-prins:
@DhanyaHerath - any comment on the proposed addition?

DhanyaHerath:
@mark-prins thanks for getting this in! I like the idea of a confirmation modal but I'm unclear on what this particular one is asking me because it feels like it's asking for two different things (whether I want to make an outbound shipment, and whether I want to finalise) which is hard for me to process. 

I wonder for this particular issue whether the clarity should come from when you try to finalise?
Maybe the header could be "Would you like to finalise this requisition?" 
And the text should say something like "If you finalise, you will not be able to supply this requisition further"

image

mark-prins:
thanks for feedback!
have updated the 'not-fully-supplied' message to:

PR #3680, opened 2024-04-25
3501 add new stock line mutation

### LINKED ISSUE #3501: Add new stock line mutation
TODO fill in (as per team discussion)

Use existing stock in service, but still create inventory adjustment

lache-melvin:
Fixes #3501

# 👩🏻‍💻 What does this PR do?

Adds mutation for adding a new stock line via an inventory adjustment. 

A couple of things!

Keen for any thoughts on the code location/return types... 
- We're adding a stock line, so added the mutation in the stock_line graphql crate
- However, we're adding via an inventory adjustment, which is an invoice, so in the service layer am going to the invoice service
- But then, we actually care about the resulting stock line, so am returning the stock line from the invoice service method (well, I suppose we don't care what's returned, we never do anything with it in the frontend, but this is the pattern... maybe should try returning nothing 👀)

Also, with the `stock_in_line` abstraction... interested in which pattern we prefer for invoice line fields that only apply to some invoice types?

I copied the pattern we used for return reasons, where shared fields are on the InsertStockInLine, and then a separate step to add the inventory adjustment reason id 
- Pros: keeps the StockInLine abstraction clean - it's not doing a bunch of logic on things it doesn't need to, devs don't need to pass in a million default values
- Cons: on the developer to remember the second step - I can imagine using the stock in line abstraction somewhere else and then being like WHY ISN'T THE REASON BEING SET AHHH ...also probably a bit more overhead in general?

Thoughts??

# 🧪 Testing

- [ ] Navigate to stock view
- [ ] Click the `New Stock` button (if it's not there, ensure you have the Inventory Adjustments feature enabled)
- [ ] Select an item and fill in the form, including an inventory adjustment reason
- [ ] Click Ok
- [ ] You should now be able to find your new stock line in the stock list

Automated tests also added

# 📃 Documentation


server/graphql/stock_line/src/mutations/insert.rs
```diff
@@ -26,27 +31,13 @@ pub struct InsertInput {
...
-#[derive(SimpleObject)]
-#[graphql(name = "InsertStockLineError")]
-pub struct InsertError {
-    pub error: InsertErrorInterface,
-}
-
 #[derive(Union)]
 #[graphql(name = "InsertStockLineLineResponse")]
 pub enum InsertResponse {
-    Error(InsertError),
```
lache-melvin, line 45: No structured errors here, removing the error variant of the response

server/repository/src/db_diesel/invoice_line_row.rs
```diff
@@ -121,6 +121,18 @@ impl<'a> InvoiceLineRowRepository<'a> {
         Ok(())
     }
 
+    pub fn update_inventory_adjustment_reason_id(
```
lache-melvin, line 124: Added for separate update reason step as discussed in the PR description, rather than as part of insert_stock_in_line

server/service/src/invoice/inventory_adjustment/adjust_existing_stock/insert.rs
```diff
@@ -107,7 +107,9 @@ mod test {
     use util::inline_edit;
 
     use crate::{
-        invoice::inventory_adjustment::{insert::InsertInventoryAdjustment, AdjustmentType},
+        invoice::inventory_adjustment::{
+            adjust_existing_stock::InsertInventoryAdjustment, AdjustmentType,
```
lache-melvin, line 111: Sorry about these... had to move existing "insert inventory adjustment" to `adjust_existing_stock` to distinguish from this "add new stock" adjustment

server/service/src/invoice/validate.rs
```diff
@@ -53,9 +53,11 @@ pub fn check_invoice_is_editable(invoice: &InvoiceRow) -> bool {
             InvoiceRowStatus::Delivered => false,
             InvoiceRowStatus::Verified => false,
         },
-        InvoiceRowType::InventoryAddition
-        | InvoiceRowType::InventoryReduction
-        | InvoiceRowType::Repack => false,
+        InvoiceRowType::InventoryAddition | InvoiceRowType::InventoryReduction => match status {
+            InvoiceRowStatus::New => true,
+            _ => false,
```
lache-melvin, line 58: Needed to be able to add invoice lines, then verify

server/service/src/invoice_line/stock_in_line/insert/generate.rs
```diff
@@ -22,20 +24,28 @@ pub fn generate(
...
     };
 
-    let new_batch_option = if existing_invoice_row.status != InvoiceRowStatus::New {
+    let barcode_id = generate_barcode_id(&input, connection)?;
+
+    let new_batch_option = if should_upsert_batch(&input.r#type, &existing_invoice_row) {
         let new_batch = generate_batch(
-            &existing_invoice_row.store_id,
+            // If a stock line id is included in the input, use it
+            input.stock_line_id.is_some(),
```
lache-melvin, line 39: Not 100% happy with this... I wanted some level of idempotency for the request, so wanted to pass through the stock line ID from the input rather than generating a random one. However I'm using it here as though it is an "existing" stock line ID, and we should "update" it. It's an upsert, so it works fine, but it's a bit icky 🤷‍♀️

server/service/src/invoice_line/stock_in_line/mod.rs
```diff
@@ -11,12 +12,68 @@ pub use self::update::*;
...
 }
+
+pub struct StockLineInput {
+    pub store_id: String,
+    pub on_hold: bool,
+    pub barcode_id: Option<String>,
+    pub supplier_link_id: String,
+}
+
+pub fn generate_batch(
```
lache-melvin, line 34: Added a couple of fields (on_hold, barcode_id) and realised stock_in_line was actually using the `generate_batch` from inbound_shipment_line. Didn't want to touch inbound shipment, so created a `generate_batch` for stock_in_line.

server/repository/src/db_diesel/invoice_line_row.rs
andreievg: thanks for the comment

server/service/src/invoice_line/stock_in_line/insert/generate.rs
```diff
@@ -1,41 +1,58 @@
...
     let mut new_line = match store_preferences.pack_to_one {
         true => convert_invoice_line_to_single_pack(new_line),
         false => new_line,
     };
 
-    let new_batch_option = if existing_invoice_row.status != InvoiceRowStatus::New {
+    let barcode_option = generate_barcode(&input, connection)?;
+
+    let new_batch_option = if should_upsert_batch(&input.r#type, &existing_invoice_row) {
         let new_batch = generate_batch(
```
andreievg, line 44: This would be should **insert** batch right ?

server/service/src/invoice_line/stock_in_line/mod.rs
andreievg: Did we end up having an issue to consolidate insert lines in inbound shipments and maybe in inventory adjustments in stocktake ?

server/service/src/invoice/inventory_adjustment/add_new_stock_line/validate.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use super::insert::{AddNewStockLine, AddNewStockLineError};
+
+pub fn validate(
+    connection: &StorageConnection,
+    store_id: &str,
+    input: &AddNewStockLine,
+) -> Result<(), AddNewStockLineError> {
+    use AddNewStockLineError::*;
+    if !check_store_exists(connection, store_id)? {
```
andreievg, line 16: Hmm, you probably got this one from stocktake ? 

There is auth check to make sure store exists, and there would be a conflict in DB if it doesn't exist, I dont think this test is really needed

server/service/src/invoice/inventory_adjustment/add_new_stock_line/validate.rs
```diff
@@ -0,0 +1,45 @@
...
+        return Err(InvalidStore);
+    }
+
+    if !check_stock_line_does_not_exist(&input.stock_line_id, connection)? {
```
andreievg, line 20: I got a little bit confused here. Can we not generate stock line id ?

andreievg (commented):
Thanks! In general looks pretty good. 

I wasn't sure why the stock in line id needs to be passed in from input. Otherwise I think it looks pretty good. Maybe we can have a quick discussion about this one on Thursday, also will give me a little bit of time to think through the abstraction

server/service/src/invoice_line/stock_in_line/mod.rs
lache-melvin: No issues, just haven't refactored yet... didn't want to make changes to those areas in this PR

server/service/src/invoice_line/stock_in_line/insert/generate.rs
lache-melvin: I think use cases at the moment is insert only, but for other inventory adjustments, you could be creating a new `StockIn` invoice line that updates existing stock line...

server/service/src/invoice/inventory_adjustment/add_new_stock_line/validate.rs
lache-melvin: Thanks for review! Yeah let’s chat on Thursday. 

Main reason for passing the stock line ID was to prevent getting duplicates if you send the same request many times… other option would be to instead send an invoice ID for the inventory adjustment, and generate stock line ID - this might be simpler, and it is what we've done for the other invoices... I just found myself calling the `insertStockLine` mutation and it felt like the right ID to include there would be a stock line ID?

andreievg (approved):
Thanks @lache-melvin, as discussed, good to merge. Perhaps some changes as discussed, for refactor when doing stock in line for inbound shipment

PR #3682, opened 2024-04-26
add returns integration tests

### LINKED ISSUE #3504: Integration Tests for Transfers

lache-melvin:
Fixes #3504

# 👩🏻‍💻 What does this PR do?

Adds integrations tests for returns transfers

# 🧪 Testing

- [ ] Run transfers integration tests - they should pass :)

# 📃 Documentation

  1.
  2.


server/service/src/processors/transfer/shipment/test.rs
```diff
@@ -140,25 +153,81 @@ async fn invoice_transfers() {
...
+                &item2,
+            );
+            // Setup: create shipment
+            tester.insert_request_requisition(&service_provider).await;
+            ctx.processors_trigger.await_events_processed().await;
+            tester.check_response_requisition_created(&ctx.connection);
+            tester.insert_outbound_shipment(&ctx.connection);
+            tester.update_outbound_shipment_to_picked(&service_provider);
+            ctx.processors_trigger.await_events_processed().await;
+            tester.check_inbound_shipment_created(&ctx.connection);
```
lache-melvin, line 221: This actually raised a bit of a flag - we can create an Outbound Return from an Inbound Shipment that's actually only in `Picked` status (I would assume you could create one from a manual IS in `New` status)

The UI prevents this from happening, but should the backend prevent someone returning lines from a shipment before its verified?

lache-melvin:
Also, I occasionally got an join error with `exec_concurrent` when running with Postgres - can you replicate?

server/service/src/sync/test/integration/README.md
```diff
@@ -44,36 +44,37 @@ Only data that needs to be present on central server site is a new sync site (th
 
 ## 4 `Open mSupply central server`
 
-* The open mSupply central server should have a port offset of `+2` relative to the mSupply central server port.
```
lache-melvin, line 47: woo sorry fixed a typo or two and the formatter had a field day :)

server/service/src/sync/test/integration/errors.rs
```diff
@@ -77,8 +77,10 @@ mod tests {
             .update_sync_settings(&service_context, &sync_settings)
             .unwrap();
 
+        let hardware_id = service_provider.app_data_service.get_hardware_id().unwrap();
+
         let synchroniser =
-            get_synchroniser_with_hardware_id(&connection_manager, &sync_settings, "id1");
```
lache-melvin, line 81: This looks like a change in functionality - the test was failing on the post-initialisation sync with the "id1" hardware ID, because it already had one set (i.e. the error that is supposed to happen below) 

I'm not super familiar with all of it, but this test reads to me like we were expecting there to no hardware ID set at this step, so we'd set it to id1, then change to id2 and see the error. And now it's failing, because there is already a hardware ID by the time we get to this step. 

I've fixed it by querying for the hardware ID instead, but just wanting to double check this is expected/we just forgot to update the test, or if there are any other implications?

server/service/src/sync/test/integration/remote/invoice.rs
```diff
@@ -60,10 +60,11 @@ impl SyncRecordTester for InvoiceRecordTester {
             requisition_id: None,
             linked_invoice_id: None,
             clinician_link_id: None,
-            currency_id: currency_row.id.clone(),
+            currency_id: Some(currency_row.id.clone()),
```
lache-melvin, line 63: There were a few compilation fixes needed in the integration tests 🙈

server/service/src/sync/test/integration/errors.rs
andreievg: Yep, we had issues with hardware_id when doing omSupply central integration test (since it was possible for omSupply central to be using the same site as used for configuring test data on mSupply), so we've changed to allow for this (https://github.com/msupply-foundation/open-msupply/pull/2819/files#diff-8e0beeb411e6b526997da24481dc9d6d0979116b644fa22ca67a4c2d59055ae1, under central_server_configurations.rs, and test_helplers).

Forgot to update some tests, not sure how they were passing.

I actually was suppose to fix them in a previous PR i made, Clemens found this error too, but i think i forgot to commit. Anyways, when i fixed, i just removed the first sync in that test (since the correct hardware_id would already be set during configuration of test site and test data)

server/service/src/processors/transfer/shipment/test.rs
```diff
@@ -786,31 +964,289 @@ impl ShipmentTransferTester {
...
+            .query_one(InvoiceFilter::new_match_linked_invoice_id(
+                &self.outbound_return.id,
+            ))
+            .unwrap();
+
+        assert!(inbound_return.is_some());
+        let inbound_return = inbound_return.unwrap().invoice_row;
+        self.inbound_return = Some(inbound_return.clone());
+
+        assert_eq!(inbound_return.r#type, InvoiceRowType::InboundReturn);
```
andreievg, line 1025: btw, just an option for future, i've been using inline_edit for some test like this:

```
assert_eq!(inbound_return, InvoiceRow {
  r#type: InvoiceRowType::InboundReturn,
  store_id: self.outbound_store.id,
 ..inbound_return.clone()   
})
```

With this method assert_eq would need to be change to a prettier version if error occurs, but it might be cleaner to see what is being tested, so pros and cons of both

andreievg (changes requested):
Looks great thanks, one change request before approving, and that's manually triggering processor in service

andreievg (approved):
As per our conversation, looks good to go.

https://github.com/msupply-foundation/open-msupply/compare/3504-returns-transfers-integration-tests...3504-returns-transfers-integration-tests(Extra-comment-about-manually-triggering-processor)

PR #3683, opened 2024-04-26
Add cli to build

### LINKED ISSUE #3677: Add cli to windows build

I would like to run the `refresh-dates` cli command on the demo server, and would also like to have sync enabled after doing so.

* Add the cargo command to build the binaries for posgres and sqlite versions of the cli
* include the appropriate binary into the installer
* Update the refresh dates command to allow sync enabling

It might be possible to refresh the dates in 4D instead - but given the move away from the mSupply central server it would be preferable to refresh the omSupply end.

mark-prins:
Fixes #3677

# 👩🏻‍💻 What does this PR do?

Tweaks the `refresh-dates` cli command, so that I can keep sync enabled, and copes with the program_event max end date.

Updates the installer, so that the cli tool is built and distributed with the server.

This allows me to implement a refresh dates scheduled task on the demo server

# 🧪 Testing

- [ ] Try to install the server version; there should now be a a file `omSupply-cli.exe` in the install directory.
- [ ] Running the exe with the param `--enable-sync` will work i.e. not complain about an invalid param
- [ ] If run with that param, the message about disabling sync isn't shown

# 📃 Documentation

  1. Perhaps we could mention the cli tool, but I think not
  2. This is for internal use


clemens-msupply (approved):
Looks good!

PR #3691, opened 2024-04-29
3018 invalid mapping of m supply transact.tax to oms invoice.tax

### LINKED ISSUE #3018: (Possibly) Invalid mapping of mSupply transact.tax to OMS invoice.tax
## What went wrong? 😲

https://github.com/msupply-foundation/open-msupply/blob/7e4320318f0ff74d9b9fb340515ae33099863d6c/server/service/src/sync/translations/invoice.rs#L245

I'm finding it hard to track in our code base whether `tax` is consistently the tax percentage to apply to line (and what factor representation it is in, is 10% -> 10.0, 0.1 or 1.1) or actually the tax component of the line (i.e. $30*10% = $3.0). I'm pretty sure it's generally not the last in OMS, but legacy mSupply it is the last... sync translator is treating it as the factor. I've seen some places explicitly call variables for tax `tax_percentage` which is much clearer.

If we wanted to be super rusty maybe we could make a type aliases for `type Tax = f64` (or TaxTotal) and `type TaxPercentage = f64`, along with rustdocs explaining precisely how each is meant to be used. (@clemens-msupply perhaps another reason to have domain object types separate from the diesel record types 😃, or maybe diesel wouldn't mind a straight alias of f64... )

## Expected behaviour 🤔

Tax calculations and values should always be unambiguous in the codebase. Sync translation needs investigation for inconsistency. If inconsistent, we must decide either mapping the values from legacy records with a calculation or reimplementing tax in OMS (surely the former right?).

## Solution

https://github.com/msupply-foundation/msupply/issues/10118#issuecomment-1966107074
>@Chris-Petty Oh. OMS was supposed to fix all the badness in mSupply, it wasn't supposed to leak over! Changing the field name on Open mSupply would be a REALLY good idea.

😁 

- [ ] we should rename `invoice.tax` to `invoice.tax_rate` to reduce misinterpretation.
- [ ] if not keeping `invoice.tax` (tax_amount), it needs to be translated when pushing to legacy. (Or we keep it and populate it too for a straight translation)
- [ ] Add transact.tax_rate to mSupply (migrate past data to populate, all UI invoice logic to update it correctly)
- [ ] In incoming translator save tax_rate from tax_rate if mSupply provided it. If not, calculate from `tax`. Either that or we increment the API compatibility of OMS to require a later version of Legacy Server?

## Additional context

mSupply doesn't have an `invoice_line.tax`, tax was applied at the invoice level across the total price of all lines. omSupply moved taxes to the the `invoice_line.tax` as that allows different tax rates for different products (for NZers think GST exemption being considered by govt for fresh produce, or products like cigarettes and alcohol having specific additional taxes applied with GST on top of that (IDK if it's already handled before retailers, but anyway).

## How to Reproduce 🔨

Steps to reproduce the behaviour: (UNVALIDATED as of writing)

1. Have a legacy desktop store's customer invoice with a given tax value
2. Sync the store/site to OMS
3. Check the tax hasn't essentially gone wacko

## Your environment 🌱

- Version: Probably all for the last couple years, but I'm looking at 1.7.0 `develop` code.
- mSupply version: Been the way it is for eons, but 7.13 right now.

roxy-dao:
Fixes #3018

# 👩🏻‍💻 What does this PR do?
Rename tax to tax_rate to make it clearer that it isn't the tax amount. Should be compatible with OG 7.14 which is required for V2.0 anyways... Didn't rename it in invoice_stats though cause didn't wanna recreate the view again... 👀 

Note: OG's tax field which was being used is actually the tax_amount... does this affect anyone atm? 

- [ ] Schema change: invoice and invoice line's `tax` is now `tax_rate`

# 🧪 Testing

- [ ] Make sure tests are still running
- [ ] Make sure tax can still be set in frontend

# 📃 Documentation

  1.
  2.


clemens-msupply:
We shouldn't really (never) change an existing api. Did you check if reports use it? I also don't use it in any of my scripts so it might be ok to rename it.

However, we could also just use tax_percentage in the service instead of tax and leave the api as it is. Or is there a strong reason to use tax_rate? The api mostly uses tax_percentage and where not we could add some docs comments.

> Didn't rename it in invoice_stats though cause didn't wanna recreate the view again...

I think its worth doing this though. Or does this causes to many other changes?

roxy-dao:
> We shouldn't really (never) change an existing api. Did you check if reports use it? I also don't use it in any of my scripts so it might be ok to rename it.
> 
> However, we could also just use tax_percentage in the service instead of tax and leave the api as it is. Or is there a strong reason to use tax_rate? The api mostly uses tax_percentage and where not we could add some docs comments.
> 
> > Didn't rename it in invoice_stats though cause didn't wanna recreate the view again...
> 
> I think its worth doing this though. Or does this causes to many other changes?

Yeah, but tax can be confused with tax (amount)... =/ but if we never save tax amount I guess that'll be fine? Just confusing for new devs....

can use tax_percentage... was just using tax_rate to match OG, but I can use serde to rename. If we name it tax_percentage then then view doesn't need to be recreated

clemens-msupply:
I vaguely remember that we choose tax_percentage on purpose,  @andreievg, @mark-prins? At least we did it for the update input struct where we use `percentage`. Would have to rename it to rate there as well.

roxy-dao:
> I vaguely remember that we choose tax_percentage on purpose, @andreievg, @mark-prins? At least we did it for the update input struct where we use `percentage`. Would have to rename it to rate there as well.

Have renamed it to tax_percentage (:

server/graphql/invoice_line/src/mutations/outbound_shipment_line/service_line/update.rs
```diff
@@ -23,7 +23,7 @@ pub struct UpdateInput {
     item_id: Option<String>,
     name: Option<String>,
     total_before_tax: Option<f64>,
-    tax: Option<TaxInput>,
+    tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 26: This shouldn't be renamed, should it? percentage is already in the struct...

server/graphql/invoice_line/src/mutations/outbound_shipment_line/line/update.rs
```diff
@@ -27,7 +27,7 @@ pub struct UpdateInput {
     stock_line_id: Option<String>,
     number_of_packs: Option<f64>,
     total_before_tax: Option<f64>,
-    tax: Option<TaxInput>,
+    tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 30: Think this should stay tax (see other comment)

server/graphql/invoice_line/src/mutations/inbound_shipment_line/service_line/update.rs
```diff
@@ -23,7 +23,7 @@ pub struct UpdateInput {
     item_id: Option<String>,
     name: Option<String>,
     total_before_tax: Option<f64>,
-    tax: Option<TaxInput>,
+    tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 26: Leave it as tax?

server/graphql/invoice_line/src/mutations/inbound_shipment_line/line/update.rs
```diff
@@ -31,7 +31,7 @@ pub struct UpdateInput {
     pub expiry_date: Option<NaiveDate>,
     pub number_of_packs: Option<f64>,
     pub total_before_tax: Option<f64>,
-    pub tax: Option<TaxInput>,
+    pub tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 34: Leave it as tax?

server/graphql/invoice/src/mutations/outbound_shipment/update.rs
```diff
@@ -32,7 +32,7 @@ pub struct UpdateInput {
     their_reference: Option<String>,
     transport_reference: Option<String>,
     colour: Option<String>,
-    tax: Option<TaxInput>,
+    tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 35: Leave it as tax?

server/graphql/invoice/src/mutations/inbound_shipment/update.rs
```diff
@@ -29,7 +29,7 @@ pub struct UpdateInput {
     pub comment: Option<String>,
     pub their_reference: Option<String>,
     pub colour: Option<String>,
-    pub tax: Option<TaxInput>,
+    pub tax_percentage: Option<TaxInput>,
```
clemens-msupply, line 32: Leave it as tax?

clemens-msupply (changes requested):
Just some smaller issues in the comments.

I think we should keep using `pub tax: Option<TaxInput>,` since TaxInput already contains percentage. We could even use TaxInput to update other tax related values in the future (if there are any).
There might be some renames like this in the service as well? 

So the frontend doesn't really support updating the tax? Only `client/packages/invoices/src/InboundShipment/api/api.ts` seemed to use it, right?

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundServiceLineEdit/useServiceLineColumns.tsx
```diff
@@ -18,7 +18,7 @@ import {
 } from '@openmsupply-client/system';
 import { DraftInboundLine } from './../../../../types';
 
-const TaxPercentageCell = (props: CellProps<DraftInboundLine>) => (
+const taxPercentageCell = (props: CellProps<DraftInboundLine>) => (
```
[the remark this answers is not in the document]
roxy-dao: Opps, sorry. Didn't double check mass rename 👀

clemens-msupply (approved):
Looks good thank you!

PR #3698, opened 2024-04-29
Default to local server

### LINKED ISSUE #3695: yarn start should default to local
## The suggestion

Current `yarn start` tries to connect your frontend to the demo server.
Suggest that we add `yarn start-remote` for this behaviour as most apps use `yarn start` point to local

Saves time and confusion for developers who connect to the wrong server.

Not really only affected developers

<1 hr

mark-prins:
Fixes #3695

# 👩🏻‍💻 What does this PR do?
Change to use `yarn start-local` as the default, instead of `yarn start-remote`

# 🧪 Testing

Try running `yarn start` and verify that this is hitting your local server, and not the demo server.


roxy-dao (approved):
LGTM~

PR #3693, opened 2024-04-29
Soft Delete for Asset Catalogue Items

jmbrunskill:
Fixes #3675 

# 👩🏻‍💻 What does this PR do?

Adds soft delete for asset catalogue items.
The prevents a bug with sync if the catalogue item is deleted centrally but used remotely.
See issue for reproduction steps.

## 💌 Any notes for the reviewer?

To add the new column to the database, you'll need to run this command...

`ALTER TABLE asset_catalogue_item add column  deleted_datetime TIMESTAMP;`

# 🧪 Testing

1. Import a new asset catalogue item (On central server)
2. Sync your remote site
4. Create an asset on the remote site using the new catalogue item
5. Before syncing to central, delete the catalogue item from central
6. Sync to central server
7.  Check that the asset show up in the central server

# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [X] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: .

PR #3701, opened 2024-04-29
Implement name soft deletes

### LINKED ISSUE #3555: After patient-merge in mSupply-OG, it should reflect in patient list accordingly in oms-side as well
## Expected behaviour 🤔
Don't show merged patient in the list after merge?

## How to Reproduce 🔨

1. Have a sync-setup with at least one desktop-OG-dispensary and oms-dispensary
2. In desktop-OG, create two patients: in my eg: `flareon1` & `flareon2`
3. Edit visibility so that it's visible to oms-dispensary as well and sync
4. Now in cloud, merge those two patients: I merged `flareon1` & kept `flareon2` and sync
5. In oms-dispensary, see that merged patient (`flareon1`) still appearing!
SCR-20240405-j5m

6. Create a prescription and add that merged-patient (`flareon1`)
7. See error: OtherPartyDoesNotExist toast; which is technically correct, but I think better would be to not show the merged patient at all!
SCR-20240405-j5r

## Your environment 🌱

- Open mSupply Version: v1.7.0 exe
- Legacy mSupply Central Server Version: v714-dev5

- Platform: server exe, browser (chrome)

- Database type: sqlite

clemens-msupply:
Fixes #3555

# 👩🏻‍💻 What does this PR do?
When receiving a name MERGE record, the deleted name gets soft-delete now.

# 🧪 Testing

See issue, but found this easier:
- add a dispensary store to the central server site
- create name "a1" and "a2" on the central server
- sync and see names on the remote sites
- merge "a1" and "a2"
- sync, there should only be one name visible on the remote site now


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.0.00",
+  "version": "2.1.00",
```
clemens-msupply, line 4: Could sqeeze it in 2.0.0?

server/repository/src/db_diesel/name_row.rs
```diff
@@ -212,8 +217,11 @@ impl<'a> NameRowRepository<'a> {
         Ok(())
     }
 
+    /// Soft delete a name record (set is_active to false)
     pub fn delete(&self, name_id: &str) -> Result<(), RepositoryError> {
```
clemens-msupply, line 221: Wonder if we should rename these soft delete method to `soft_delete()`

server/service/src/sync/translations/mod.rs
```diff
@@ -336,6 +338,7 @@ pub(crate) trait SyncTranslation {
         Ok(PullTranslateResult::NotMatched)
     }
 
+    /// Translate a delete record received from the central server(s)
```
clemens-msupply, line 341: Got confused about from/to and added some comments...

server/service/src/sync/translations/name.rs
```diff
@@ -280,12 +281,18 @@ impl SyncTranslation for NameTranslation {
...
         } = NameRowRepository::new(connection)
             .find_one_by_id(&changelog.record_id)?
             .ok_or(anyhow::Error::msg(format!(
                 "Name row ({}) not found",
                 changelog.record_id
             )))?;
+        if !is_active {
+            return Ok(PushTranslateResult::Ignored(
+                "Ignore pushing soft deleted name".to_string(),
+            ));
```
clemens-msupply, line 294: It got deleted in mSupply and we soft-deleted it. So I guess it correct to not push it back...

server/repository/src/db_diesel/name_row.rs
roxy-dao: Yeah, would be nice to rename to match what the function actually does

server/service/src/sync/translations/name.rs
```diff
@@ -230,12 +231,12 @@ impl SyncTranslation for NameTranslation {
                 .or(created_date.map(|date| date.and_hms_opt(0, 0, 0).unwrap())),
             date_of_death,
             custom_data_string,
+            is_active: existing_name.map_or(true, |name| name.is_active),
```
roxy-dao, line 234: OG doesn't have an `is_active` field, so this will always be true.

It does have an `is_deleted` field, but that doesn't actually get used so any merged names are deleted...

roxy-dao (changes requested):
Nice! Just a few notes down below (: I think maybe we might be able to squeeze into 2.0.0?

server/service/src/sync/translations/name.rs
clemens-msupply: Don't think its possible right now, but thought about some case where the name is soft-deleted locally and another UPSERT record is received...

However, as of https://github.com/msupply-foundation/open-msupply/pull/3693 I plan to change it to `deleted_datetime: Option<_>` which will solve it nicely as well.

roxy-dao: The issue is that the merged name is still viewable for me tho? And the is_active bool is true for the merged name

clemens-msupply:
> Nice! Just a few notes down below (: I think maybe we might be able to squeeze into 2.0.0?

Done, also renamed `mark_deleted()` and changed to `created_datetime` column

server/service/src/sync/translations/name.rs
```diff
@@ -230,12 +231,12 @@ impl SyncTranslation for NameTranslation {
                 .or(created_date.map(|date| date.and_hms_opt(0, 0, 0).unwrap())),
             date_of_death,
             custom_data_string,
+            deleted_datetime: existing_name.and_then(|name| name.deleted_datetime),
```
roxy-dao, line 234: Why is this using existing_name? If the name was merged in OG, then this wouldn't populate the deleted datetime at all....

roxy-dao: and we're not deleting patients in OMS....

clemens-msupply: Yeah you are right it's not needed.

roxy-dao (approved):
Thanks for the changes Clemens~

PR #3706, opened 2024-04-29
Add missing postgres enums

clemens-msupply:
Part of #3654

# 👩🏻‍💻 What does this PR do?

Adds missing pg enums for assets and file sync. This is much easier to add now than to ALTER the table in 2.1.0

## Testing

Please check enums match the Rust types, i.e. the variants and the Rust type should match the PG type for the Diesel 2.0 update.


jmbrunskill (approved):
Looks good thanks for getting this into 2.0

PR #3719, opened 2024-04-30
3684 pack variant feature flag

### LINKED ISSUE #3684: Feature flag for pack variant UI
The pack variant UI has now been shown to PMs and concerns have been raised at the scope of the changes and the lack of general input.

We would like to have some time to consider how to proceed - to that end, please can we keep the functionality of pack variants, while restricting UI changes to a feature flag .

Implement a feature flag for the pack variant work. This will
* Revert the column changes ( changing unit and pack size to a single column called `Pack` )
* Remove the pack variant selector component

enabling the feature flag will re-enable these items.
All the api / backend work can remain without causing any impact.

The configuration of pack variants in the central server can remain too, the impact of that is low and won't cause an issue for most users.

I don't see that we have time to rollback all the pack variant work, nor do we have time to have PMs agree on how to proceed, or give Anisha enough time to review. Not without delaying the release of 2.0.00 significantly.

lache-melvin:
Fixes #3684

# 👩🏻‍💻 What does this PR do?

Adds feature flagging of pack variant features.

As per issue... Pack Variant UI was rushed in the end, we need more time to consider if this is what we want. This PR adds a toggle for pack variant column/UI changes.

Pack variants enabled: 
Screenshot 2024-04-30 at 12 59 20 PM

Disabled:
Screenshot 2024-04-30 at 2 20 26 PM

Also when disabled, brings back the "in packs of" labels in prescription/outbound shipment edit line modal:
Screenshot 2024-04-30 at 3 18 20 PM

And the packs/units toggle in Internal Order modal:
Screenshot 2024-04-30 at 3 19 35 PM

Also comments out the `mostUsePackVariant` logic in the backend, as this was raised as a maybe performance issue (especially as it was running against all items even when no variants were defined..) - seeing as the UI will be hidden for now, have commented out so we don't need to investigate performance at this stage. 

Have added a feature flag, but this only hides the `Pack Variants` tab in the Item detail view, where a user could configure pack variants.

For the most part, I'm using a hook `usePackVariantsEnabled`, which is checking if there are any pack variants defined for any items. This should be safe in that no sites will have pack variants defined yet, so it's the same as hiding behind a feature flag. The thinking in Ruru team was that if pack variants aren't being used, the UI shouldn't change (regardless of whether pack variants is enabled as a feature). This was largely the intention of pack variant work anyways, except for the merging of the two columns. 

If that doesn't sit right, could always change the `usePackVariantsEnabled` hook to return the feature flag boolean instead 🤷‍♀️ 

Main thing I'm hoping is that this is enough to hide pack variants for this release, but it's not a complete toggle to "reverse" the changes. Am still using the pack variant default logic in most places, which looks the same as the old pack size/unit stuff. I guess just wanting to flag that if we end up deciding to abandon/back out pack variants, that's more work than just removing what's behind this flag.

# 🧪 Testing

- [ ] Start frontend without feature flag (ensure you have no existing pack variants configured)
- [ ] Hoon through them pages - should be just like old times, no mention of pack variants
- [ ] Should be `Pack Size` and `Unit` columns everywhere instead of `Pack`
- [ ] Outbound Shipment/Prescription/Internal order modals look as above
- [ ] Restart frontend with feature flag: `yarn start-local-features FEATURE_PACK_VARIANTS=true`
- [ ] Before configuring any pack variants, should still be separate columns
- [ ] (Running as central server) configure some pack variants for items
- [ ] Now you see one column `Pack`, it's a changeable dropdown in places like Item, `Unit Variant / Pack Size` input displays for that item in places like Inbound Shipment

# 📃 Documentation

  - [ ] I know returns screenshots/column descriptions were based on pack variants enabled
  - [ ] Probably other screenshots too?? Might need a sweep of anywhere that's seen updates recently 👀

### DIFF
--- a/client/packages/inventory/src/Stocktake/DetailView/ContentArea/columns.ts
+++ b/client/packages/inventory/src/Stocktake/DetailView/ContentArea/columns.ts
@@ -13,10 +13,12 @@ import {
   LocaleKey,
   useColumnUtils,
   NumberCell,
+  ColumnDescription,
 } from '@openmsupply-client/common';
 import {
   InventoryAdjustmentReasonRowFragment,
   getPackVariantCell,
+  usePackVariantsEnabled,
 } from '@openmsupply-client/system';
 import { StocktakeSummaryItem } from '../../../types';
 import { StocktakeLineFragment } from '../../api';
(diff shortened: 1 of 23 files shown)


mark-prins:
> The thinking in Ruru team was that if pack variants aren't being used, the UI shouldn't change (regardless of whether pack variants is enabled as a feature).

I'm not totally comfortable with this approach and  I'm reluctant to add another way of configuring how the system looks & behaves: 
* It's adding a new way of configuring the system: this is confusing - where do I configure the system? we have:
  * front end plugins
  * and will have back end plugins
  * store preferences 
  * and we'll likely expand the settings section to bring in general preferences
  * yaml files to configure system settings
* It means that core parts of the system look differently for different setups; this creates an overhead for documentation
* If the pack variant model is the better way to display unit/pack size - why not do this for everyone?

that said, I don't like the outbound shipment line edit modal changes and wouldn't want that to be available for non-pack-variant-users.
All a moot point anyway; we need to define how the pv should function and how it should look.

client/packages/inventory/src/Stocktake/DetailView/ContentArea/columns.ts
```diff
@@ -66,185 +68,223 @@ export const useStocktakeColumns = ({
   const t = useTranslation();
   const { getColumnPropertyAsString, getColumnProperty } = useColumnUtils();
 
-  return useColumns<StocktakeLineFragment | StocktakeSummaryItem>(
+  const packVariantsEnabled = usePackVariantsEnabled();
```
mark-prins, line 71: :nit we've used the convention of the prefix `is` for boolean variables
```suggestion
  const isPackVariantsEnabled = usePackVariantsEnabled();
```

which makes me wonder if that should be `arePackVariantsEnabled` 🤔 funnily enough, this has never come up before

( it should still be `is` - conventions are conventions )

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -108,6 +108,7 @@ export const OutboundLineEditTable: React.FC<OutboundLineEditTableProps> = ({
     }
   };
   const unit = item?.unitName ?? t('label.unit');
+  console.log(unit);
```
mark-prins, line 111: debug code?

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
```diff
@@ -107,8 +113,8 @@ export const QuantityReturnedTableComponent = ({
 export const QuantityReturnedTable = React.memo(QuantityReturnedTableComponent);
 
 // Input cells can't be defined inline, otherwise they lose focus on re-render
-// eslint-disable-next-line new-cap
 const PackUnitEntryCell =
+  // eslint-disable-next-line new-cap
```
mark-prins, line 117: yeah, this bothers me too. the `PackVariantEntryCell` isn't a component and should not be capitalised.
Needs to be renamed to be a getter fn, which it is, rather than disabling hints.

client/packages/system/src/Item/context/usePackVariant.ts
```diff
@@ -69,6 +69,12 @@ export const useRefreshPackVariant = () => {
   }, [data, setItems]);
 };
 
+export const usePackVariantsEnabled = () => {
```
mark-prins, line 72: very neat approach 😍 
Just wondering how we can switch on / off for demo / gathering feedback.. there's no system level switching, so we have to enable, then configure, then unconfigure if we are using a demo server to get feedback from PMs for example.

mark-prins (commented):
👍  awesome - thanks, working nicely!
couple of functionality questions
- Can the env var be used in production? I'll need to enable on the demo server to gather feedback, and from what I recall of using env vars in webpack, is that they are evaluated at compile time. has anyone tested the env var approach? is it valid for dev only?
- does disabling the pv feature require un-configuring all items? makes it hard to have this on a demo server. perhaps I just need two databases and switch between them 🤔 
- I'm uneasy with the amount of conditional logic this introduces.. if pack variants are the way to go, shouldn't it be available for all users? this one would be a side discussion: output of the requirements process

mark-prins (approved):
as discussed - let's get this out there! 
🙏

client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
lache-melvin: Yeah agree - thought it was a bit much of a side quest for this PR...?

mark-prins: yes, you're right. just venting uhelpfully 😉

lache-melvin:
> * Can the env var be used in production? I'll need to enable on the demo server to gather feedback, and from what I recall of using env vars in webpack, is that they are evaluated at compile time. has anyone tested the env var approach? is it valid for dev only?

Yeah they are evaluated at compile time - just need to provide the env var when you build the frontend: `yarn build -- -- --env FEATURE_PACK_VARIANTS=true`

Have updated the `useIsPackVariantsEnabled` hook to just use the feature flag, for easier enabling/disabling of the PV features for everyone's review! I think it's also a great case you made about reducing the number of ways we have to configure things 🙏

PR #3731, opened 2024-05-02
Add home button to error boundary

### LINKED ISSUE #3515: No way to go back on desktop if error
## What went wrong? 😲
My desktop error'd and now I can't go back D: and am stuck in an endless loop of trying to press try again. This will happen if any changes breaks something, so probably not an issue if everything is functioning as it should but when it doesn't....................

https://github.com/msupply-foundation/open-msupply/assets/61820074/460a70fb-c20a-42e7-a1d0-10335fc5ca54

## Expected behaviour 🤔
Should be able to go back if the page won't load

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version: 1.7

- Platform: Desktop (electron)

- Database type:

fergie-nz:
Fixes #3515 

# 👩🏻‍💻 What does this PR do?
Adds link to navigate back to home from error

# 🧪 Testing

I was a bit dissatisfied with working through how to test this because a lot of errors thrown will not be caught by our error handler. 

A future issue could be more comprehensive error catching with this provider.

@Roxy I wonder how you were able to trigger the error in this way? 

Regardless this will allow users to navigate back to home where their current url triggers the error for whatever reason.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -138,6 +138,30 @@ Standard semver rules apply:
 2. MINOR version when you add functionality in a backwards compatible manner
 3. PATCH version when you make backwards compatible bug fixes
 
+### Error handling
+
+When running the website in development mode (e.g. `yarn start`) you will see errors are displayed, not by the error boundary, but by another screen which shows additional detail.
+If you want to see the error boundary instead (for example, when testing the boundary!) you will need to disable the dev overlay:
+
+In `/packages/host/webpack.config.js`:
+
+```
+  // update the devServer entry, to add the overlay: false
+  devServer: {
+    ...
+    client: {
+      overlay: false,
+    },
+  }
+
+  // and update the ReactRefresh plugin also:
+
+  new ReactRefreshWebpackPlugin({ overlay: false }),
+
+```
+
+There is typically no need though, as you can press escape to close the overlay and see the error boundary in action.
+
 ## Queries
 
 We're using [React Query](https://react-query.tanstack.com/overview) to query the server and manage a local cache of queries.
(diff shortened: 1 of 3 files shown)


clemens-msupply (approved):
Nice! Think a tooltip showing the "home" origin url would be helpful, maybe easy to add?

For testing, I added a `throw Error();` statement in my code I got this when navigating to the affected page:

image

Nit, it doesn't help when the error happens on the dashboard page, but then there is not much you can do about it :)

Nit, wonder if there is a better name for "home"? Not sure if it is 100% clear...
You could add a tool tip displaying the origin, think this would make it clearer.

roxy-dao (changes requested):
This doesn't work for the electron app... still no way to go back

clemens-msupply:
> This doesn't work for the electron app... still no way to go back

Right, is this a dev issue or a production issue. I tried to figure out what causes it in dev but couldn't quite figure it out... The problem I am seeing is that the GenericErrorFallback is not responsive, i.e. I can't click on the home or retry button. There is a workaround though, you can go into dev tools and back and then you will see another React error screen and afterwards you can click the home button.

I just tried it out in release mode, and it works fine there.

Not sure if somebody else has an idea how to fix it but could be problem that is a bit harder to fix?

Note, even though I thought I built everything correctly, in release mode, I see this:
image

fergie-nz:
Hey @clemens-msupply , Roxy and I are struggling to get electron to throw an error.

I wonder would you be willing to show us when you get a moment? Perhaps tomorrow in office?

mark-prins:
I can break it 🤷  will show

roxy-dao (approved):
Thanks Ferg <3

mark-prins:
Just added some notes to the readme on testing the boundary

PR #3745, opened 2024-05-02
Fix onClick callback handling

### LINKED ISSUE #3584: Contact tracing: Wrong message on save
## What went wrong? 😲
When I click `Save` when adding a contact trace, it says `Deleted successfully` in green 😆 

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

clemens-msupply:
Fixes #3584 

# 👩🏻‍💻 What does this PR do?

Use onClick callback correctly. Bit tricky one because this is not obvious.

From commit message:
```
onClick has an event parameter, this parameter got silently passed
forward to the onSave method which interpreted it as a boolean.
```

Maybe a good practice it s to not pass "function pointers" around but always create a new arrow method and call the target manually? @mark-prins

### DIFF
--- a/client/packages/system/src/ContactTrace/DetailView/Footer.tsx
+++ b/client/packages/system/src/ContactTrace/DetailView/Footer.tsx
@@ -78,7 +78,7 @@ export const Footer: FC<FooterProps> = ({
               variant="outlined"
               disabled={isDisabled || isSaving}
               isLoading={isSaving}
-              onClick={onSave}
+              onClick={() => onSave()}
               startIcon={<CheckIcon />}
             >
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
Thanks Clemens! Good job in finding this (: It really isn't obvious at all...

mark-prins:
oh nice! that's a subtle bug. 
Generally I'd always pass the function pointer, rather than creating a new function on every render and usually that would not be a problem - it's the parameter that causes the bug here. Just something for people to be aware of I think

PR #3747, opened 2024-05-02
#3536 Prescription input quantity fix

### LINKED ISSUE #3536: Prescriptions: Unable to erase quantity
## What went wrong? 😲

After typing a quantity, it cannot be erased. 
Only way to get rid of the number is to select the whole field and write something on top. (Desktop app)
On tablet we cannot do that, we need to delete the whole item line.

Video available on Gdrive.

## Expected behaviour 🤔

Users should be able to erase and input a new quantity

## How to Reproduce 🔨

1. Go to 'Prescriptions'
2. Click on 'Add Item'
3. Type a quantity
4. Now try to erase the number you input. See error

## Your environment 🌱

- Open mSupply Version: 1.7 RC4
- Legacy mSupply Central Server Version: V7.13

- Platform: Desktop app and android

CarlosNZ:
Fixes #3536

# 👩🏻‍💻 What does this PR do?

Fixed issue. Allowed quantity to be erased, but it still resets to `0` rather than empty, as it causes problems to pass `undefined` to the `allocate` method.

Also... found *another* regression with the numeric input component, which I've fixed. I've also added some more tests for this so we don't keep getting these regressions.

See inline comments

# 🧪 Testing

## Prescription input
- [ ] Start a new prescription and add an item
- [ ] Enter a quantity
- [ ] Highlight and delete the quantity -- should revert to 0
- [ ] Try and click "OK" with 0 quantity entered -- should display warning

## Numeric input fixes
- Run `yarn test numeric` and confirm all tests pass
- If you feel like it, run `yarn storybook`, go to the Numeric inputs and check they still behave as expected

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
@@ -91,4 +91,48 @@ describe('Test NumericTextInput component', () => {
     fireEvent.blur(input);
     expect(input).toHaveValue('1.100');
   });
+
+  it('should handle negative number input', async () => {
+    const { getByRole } = render(<TestNumericTextInput allowNegative />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '-' } });
+    expect(input).toHaveValue('-');
+    fireEvent.change(input, { target: { value: '-5' } });
+    expect(input).toHaveValue('-5');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('-5');
+  });
+
+  it('should handle removing input', async () => {
+    const { getByRole } = render(<TestNumericTextInput />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '500' } });
+    expect(input).toHaveValue('500');
+    fireEvent.change(input, { target: { value: '' } });
+    expect(input).toHaveValue('');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('');
+  });
+
+  it('should format large numbers', async () => {
+    const { getByRole } = render(<TestNumericTextInput />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '1000' } });
+    expect(input).toHaveValue('1,000');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('1,000');
+  });
+
+  it('should not format large numbers when explicitly prevented', async () => {
+    const { getByRole } = render(<TestNumericTextInput noFormatting />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '1000' } });
+    expect(input).toHaveValue('1000');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('1000');
+  });
 });
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -55,7 +55,7 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
       (val: number | undefined) =>
         noFormatting
           ? val === undefined
-            ? undefined
+            ? ''
             : String(val)
           : format(val, { minimumFractionDigits: decimalMin }),
       [decimalMin, format, noFormatting]
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -67,14 +67,17 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
     const isFirstRender = useRef(true);
 
     const isInputIncomplete = useCallback(
-      (value: string) =>
-        new RegExp(
+      (value: string) => {
+        if (value === '-') return true;
+
+        return new RegExp(
           // Checks for a trailing `.` or a `0` (not necessarily immediately)
           // after a `.`
           `^\\d*${RegexUtils.escapeChars(
             decimal
           )}$|\\d*${RegexUtils.escapeChars(decimal)}\\d*0$`
-        ).test(value),
+        ).test(value);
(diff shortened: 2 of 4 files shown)


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
```diff
@@ -91,4 +91,48 @@ describe('Test NumericTextInput component', () => {
...
+    const { getByRole } = render(<TestNumericTextInput allowNegative />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '-' } });
+    expect(input).toHaveValue('-');
+    fireEvent.change(input, { target: { value: '-5' } });
+    expect(input).toHaveValue('-5');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('-5');
+  });
```
CarlosNZ, line 105: Makes sure we can still enter `-` on the way to typing a negative number

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
```diff
@@ -91,4 +91,48 @@ describe('Test NumericTextInput component', () => {
...
+    const { getByRole } = render(<TestNumericTextInput />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '500' } });
+    expect(input).toHaveValue('500');
+    fireEvent.change(input, { target: { value: '' } });
+    expect(input).toHaveValue('');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('');
+  });
```
CarlosNZ, line 117: Ensures we don't get `NaN` when deleting input

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
```diff
@@ -91,4 +91,48 @@ describe('Test NumericTextInput component', () => {
...
+
+  it('should not format large numbers when explicitly prevented', async () => {
+    const { getByRole } = render(<TestNumericTextInput noFormatting />);
+    const input = getByRole('textbox');
+
+    fireEvent.change(input, { target: { value: '1000' } });
+    expect(input).toHaveValue('1000');
+    fireEvent.blur(input);
+    expect(input).toHaveValue('1000');
+  });
```
CarlosNZ, line 137: Confirms formatting prop behaves as expected

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -67,14 +67,17 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
     const isFirstRender = useRef(true);
 
     const isInputIncomplete = useCallback(
-      (value: string) =>
-        new RegExp(
+      (value: string) => {
+        if (value === '-') return true;
+
```
CarlosNZ, line 72: Not sure how long this has been broken. I guess we don't use negative values much in oms.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -164,7 +167,8 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           onChange(newNum);
         }}
         onBlur={() => {
-          onChange(Number(parse(textValue ?? '')) || undefined);
+          const parsed = parse(textValue ?? '');
+          onChange(Number.isNaN(parsed) ? undefined : parsed);
```
CarlosNZ, line 171: My fix the other day (changing `||` to `??`) caused a different problem instead (showed `NaN`). Tests will now prevent this.

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
```diff
@@ -145,8 +145,8 @@ export const PrescriptionLineEditForm: React.FC<
     updateIssueQuantity(allocatedQuantity);
   };
 
-  const handleIssueQuantityChange = (quantity?: number) => {
-    if (quantity === undefined) return;
+  const handleIssueQuantityChange = (inputQuantity?: number) => {
+    const quantity = inputQuantity === undefined ? 0 : inputQuantity;
```
CarlosNZ, line 149: The actual changes that fixes the original issue. Makes it always set a value -- defaults to `0` if undefined.

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
```diff
@@ -240,7 +240,7 @@ export const PrescriptionLineEditForm: React.FC<
               autoFocus
               value={issueQuantity}
               onChange={handleIssueQuantityChange}
-              min={1}
+              min={0}
```
CarlosNZ, line 243: To be consistent with the above. We could make it default back to `1` when undefined, but then there's not point having the warning message.

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.test.tsx
lache-melvin: Not introduced in this PR, but noticed that there is a warning with this test (`A component is changing an uncontrolled input to be controlled`)

The `noFormatting` is only used in printer label settings, where a value is required, so we don't currently have a use case where we would get this warning in the app... maybe can just capture in carry over somewhere?

That printer label settings is having the same issue (can't delete the last digit) but I think it's due to the onChange implementation rather than the input... do you want to have a look now or I can create a separate issue?

lache-melvin (approved):
Awesome, thanks Carl - great to have these tests in 🔥

CarlosNZ:
> Just the note about the test/printer label settings, up to you how much to do now, would probably err on let's get this in for RC branch and create another issue?

Good spotting, thanks. It was a simple fix for both the test warning and the label printer input, so done that here: 5e0af0536

Cheers :)

lache-melvin:
Sweeeet thanks!

PR #3749, opened 2024-05-02
3475 restarting server and refreshing causes user login error and potentially blocks user 

### LINKED ISSUE #3475: Restarting server and refreshing causes User Login Error and potentially blocks user...
## What went wrong? 😲

image

## Expected behaviour 🤔

If we need to relogin then the system should tell us, or maybe it should just skip refreshing the User if the credentials are not available after a restart

## How to Reproduce 🔨

1. Login to mSupply and go to the sync page
2. Restart your server
3. Refresh the page
4. Click Sync

## Your environment 🌱

- Version: 1.7.00

roxy-dao:
Fixes #3475

# 👩🏻‍💻 What does this PR do?
1. Gives a better error message telling the user to log out if the credentials don't match/are gone
2. Update cookies with new permissions / store preferences so that they are reflected immediately instead of on refresh

# 🧪 Testing
Testing for permission/preferences updates
- [ ] Login
- [ ] Go to `Sync page
- [ ] Update some user permissions (like admin maybe since it's obvious) or store preferences
- [ ] Sync
- [ ] Should be able to do whatever perms/preferences were given

Error testing: 
- [ ] Login
- [ ] Go to `Sync` page
- [ ] Restart server
- [ ] Press sync button
- [ ] Should see error message telling user to log out

# 📃 Documentation

  1. Screenshot of new error message? Or maybe not needed idk
  3.


client/packages/common/src/authentication/AuthContext.tsx
```diff
@@ -138,6 +144,13 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
     setCookie(newCookie);
   };
 
+  const {
```
roxy-dao, line 147: Moved this to AuthContext for better access to cookie so that the expiry doesn't get overwritten or disappear!

client/packages/common/src/authentication/api/hooks/useLogin.ts
```diff
@@ -58,6 +58,27 @@ const getMostRecentCredentials = (
   return [];
 };
 
+// returns MRU store, if set
+// or the first store in the list
+export const getStore = async (
```
roxy-dao, line 63: Moved out of login function, so that store preferences could be updated when a user is updated

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -78,6 +78,7 @@
   "error.date_maxDate": "Date value is too large",
   "error.date_minDate": "Date value is too low",
   "error.general-query-error": "There was an error communicating with the server 😢",
+  "error.invalid-credentials": "Unable to fetch user details. Please re-login to update user details.",
```
roxy-dao, line 81: Honestly would prefer to have the user log out straight away if updating fails since it's easy to ignore a message... but shouldn't happen in production often... more likely to happen in development with constantly server restarts and branch changes

jmbrunskill (changes requested):
I was still able to lock myself out after your changes sadly...
Overall it looks pretty good, but I think if we don't have the user's password, we shouldn't even try to update the user.
Here's my suggestion branch.

https://github.com/msupply-foundation/open-msupply/compare/3475-Restarting-server-and-refreshing-causes-User-Login-Error-and-potentially-blocks-user-...3475_suggestion

The main thing is just skipping the user account update if the password is missing...

I also didn't seem to see the message consistently.

I think you're right if the user's credentials are wrong, we should log them out.
However, of they are just `missing` due to a server restart we should just skip the user update (maybe with a error message)

client/packages/common/src/authentication/hooks/useUpdateUserInfo.ts
```diff
@@ -0,0 +1,74 @@
...
+      try {
+        const update = await updateUser();
+
+        if (update.__typename === 'UpdateUserNode') {
+          const permissions = await getUserPermissions(
+            cookie?.token,
+            cookie?.store
+          );
+          const userDetails = await getUserDetails(cookie?.token);
+          const store = await getStore(userDetails, mostRecentCredentials);
```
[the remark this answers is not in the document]
roxy-dao: No it's just using the existing function, the default store should still be the same

jmbrunskill (approved):
Looks good, tested the user update thing and that's working too.
Left hand draw doesn't update but I think that's expected...

roxy-dao:
> Looks good, tested the user update thing and that's working too. Left hand draw doesn't update but I think that's expected...

Not expected, some of those are store prefs so will check in another issue ;x 

Thanks James!

PR #3802, opened 2024-05-06
add useKeyboardHeightAdjustment to response line edit modal

### LINKED ISSUE #3540: Editing Requisition lines on Android is difficult
## What went wrong? 😲

When focusing quantity to supply, which is default when editing a line, I cannot see the item name or code or requested quantity
Screenshot_20240404-123732

Not intuitive how to navigate (probs 'cause I'm not familiar enough to know there are buttons hidden by keyboard), end up pressing enter on keyboard which moves focus to the comment field, and now enter just adds new lines to the comment (should we be editing the comment that came from the customer anyway?)
Screenshot_20240404-123751

## Expected behaviour 🤔

Can see relevant context when editing a line

Also like elsewhere - should be a numeric keyboard for quantity to supply

## How to Reproduce 🔨

1. Edit a requisition line on android.

## Your environment 🌱

- Open mSupply Version: 1.7.00-rc5 android server
- Legacy mSupply Central Server Version: interpreted v7.13

- Platform: Android Lenovo-X505X

- Database type: sqlite3

lache-melvin:
Fixes #3540 

# 👩🏻‍💻 What does this PR do?

Adds the `useKeyboardHeightAdjustment` hook to Response Requisition modal, so the modal height adjusts when the keyboard opens:

Screenshot 2024-05-07 at 11 23 38 AM

# 🧪 Testing

- [ ] Open edit modal for a line in a response requisition on Android
- [ ] Quantity to supply should autofocus, key pad appears
- [ ] Modal should adjust in size, so you can still see item name/code/requested quantity

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
@@ -5,6 +5,7 @@ import {
   BasicSpinner,
   useBufferState,
   ModalTabs,
+  useKeyboardHeightAdjustment,
 } from '@openmsupply-client/common';
 import { ResponseLineEditForm } from './ResponseLineEditForm';
 import { useResponse, ResponseLineFragment } from '../../api';
@@ -36,6 +37,8 @@ export const ResponseLineEdit = ({
     numberOfPacksFromQuantity,
     variantsControl,
   } = usePackVariant(draft.itemId, draft.item.unitName ?? null);
+
+  const height = useKeyboardHeightAdjustment(600);
 
   const tabs = [
     {
@@ -107,7 +110,7 @@ export const ResponseLineEdit = ({
           }}
         />
       }
-      height={600}
+      height={height}
       width={1024}
     >
       {!isLoading ? (


CarlosNZ:
Doesn't seem to make a difference on Android Chrome browser. Will try from .apk now.

mark-prins:
no, it's an android only feature. not a web, even if running on android. did that make sense?
I mean the keyboardHeight package is only for the apk

lache-melvin:
> no, it's an android only feature. not a web, even if running on android. did that make sense? I mean the keyboardHeight package is only for the apk

Oo ok, do we need to consider the browser on tablet use case?

mark-prins:
Personally, I think not. we don't have clients using a browser on a tablet (ok, possibly AV as I _think_ the apk wasn't available then) - it's just too hard to find the server URL, so that scenario is only really helpful when dev testing. 

The usual scenario for clients is the android app (apk) and so that's the scenario I'm concerned about.

andreievg (approved):
Changes look good, thanks ! 

Tested, works fine, a small `jump` when keyboard opens, but it's same same as existing modals (like internal order).

andreievg:
@mark-prins is that a general product direction ? That we not really aiming for tablets use over the browser ?(I guess the only use case is cloud server for the whole country, which we not really targeting, so makes sense)

PR #3804, opened 2024-05-06
Handle app crash

### LINKED ISSUE #3740: App crash automatically after some time
## What went wrong? 😲
 
When we leave the screen for some time (for me it was half an hour) with the cold chain > equipment list is on screen
Screenshot 2024-05-02 at 2 09 02 pm

## Expected behaviour 🤔
It should just logout
## How to Reproduce 🔨

1. Go to coldchain > equipment list
2. Leave it for some time
3. See error

## Your environment 🌱

- Open mSupply Version: v2.0.00-rc2
- Legacy mSupply Central Server Version: V7-14-00

- Platform: web, mobile

- Database type:

mark-prins:
Fixes #3740 

# 👩🏻‍💻 What does this PR do?

Handle two causes of a crash: 
* When using an infinite query, undefined is returned for one or more of the pages
* The `stockCounts` query returns null or undefined

This is almost impossible to replicate - to cause the issue I've updated `useInfiniteAssets` in order to return an invalid result within the array:

```
  const result = useInfiniteQuery(api.keys.paramList(params), ({ pageParam }) =>
    api.get.list({ ...params, ...pagination, ...pageParam })
  );

  return { ...result, data: { ...result.data, pages: [undefined] } };

```

Simply deleting the auth cookie or invalidating the query in some way ( incorrect filter, for example ) is not enough! That will return null for the `data` prop and won't return invalid pages.

# 🧪 Testing

- [ ] Check that the site behaves as normal 🤷 
- [ ] The only reliable test was to log into the demo server and leave that on the equipment list view until timed out. last time I tried this however, there was no error

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/ListView/CreateAssetModal.tsx
+++ b/client/packages/coldchain/src/Equipment/ListView/CreateAssetModal.tsx
@@ -123,7 +123,7 @@ export const CreateAssetModal = ({
 
   const catalogueItems = ArrayUtils.flatMap(
     catalogueItemData?.pages,
-    page => page.nodes
+    page => page?.nodes ?? []
   );
 
   const selectedCatalogueItem = catalogueItems.find(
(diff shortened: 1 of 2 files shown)


fergie-nz (approved):
Nice Mark,

I didn't try to recreate the errors as the changes look benign and safe in any case,

Looks great!

PR #3805, opened 2024-05-07
3800 sync assets fails with duplicate asset numbers

### LINKED ISSUE #3800: Sync assets fails with duplicate asset numbers
## What went wrong? 😲
got this error while initialising the existing site

telegram-cloud-photo-size-5-6179037344607353836-y

This is because there's a changelog entry for an asset id that doesn't exist.
I think this is because sqlite can update an id field if doing an upsert with a unique constraint on another field.
Any unique constraint failure is considered an upsert, unlike postgres which should give a unique constraint error instead.

## How to Reproduce 🔨

1. Create an asset on Site A, with asset_number `ABC`
2. Create as asset on Site B, with asset_number `ABC`
3. Sync both to central server
4. Check both assets show up on Central Server (Could be different behaviour between Postgres and SQLite)

## Your environment 🌱
- Open mSupply Version: 2.0.0-RC2

- Database type: Both?

jmbrunskill:
Fixes #3800

# 👩🏻‍💻 What does this PR do?

This removes the unique constraint on asset number
See issue description to explain why this is needed.

In future, we should probably ensure we don't have unique constraints on non-id fields, as SQLite upsert logic is triggered on any unique constraint, no just on the id field. Especially with sync we need to be careful as duplicated records can come from different sites and might not go through service layer upsert logic.

# 🧪 Testing

To test this you'll need to either re-initialise OMS Central server as this is a database change...

- [ ] Central Sync server with 1 Legacy Desktop remote site and 1 OMS remote site running this PR
- [ ] Create assets on both sites with the same asset number
- [ ] Sync both sites
- [ ] Ensure both assets are available on central
- [ ] Try re-initialising the remote sites, make sure the asset records is downloaded sucessfully on initalisation.

# 📃 Documentation

### DIFF
--- a/server/repository/src/migrations/v2_00_00/assets/asset.rs
+++ b/server/repository/src/migrations/v2_00_00/assets/asset.rs
@@ -20,11 +20,11 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
             replacement_date {DATE},
             deleted_datetime {DATETIME},
             created_datetime {DATETIME} NOT NULL,
-            modified_datetime {DATETIME} NOT NULL,
-            UNIQUE (asset_number) -- Asset numbers must be unique within a site
+            modified_datetime {DATETIME} NOT NULL
         );
         CREATE INDEX asset_catalogue_item_id ON asset (asset_catalogue_item_id);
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
Thanks James - works like a charm
👍

PR #3808, opened 2024-05-07
Enable hold button for "Shipped" invoices

### LINKED ISSUE #2526: Android version not running when no network available
## What went wrong? 😲
 

"The tablet is working with WiFi or internet but not working in local host.
..while choosing server to e don’t have localhost as option at all when we connect to WiFi, 
and then we turned off WiFi and choose local host as server but it crashed"

To be investigated.

## How to Reproduce 🔨

1. Turn off wifi
2. From the login screen, click on the server list (pencil icon)
3. Select localhost
4. Boom!

## Your environment 🌱

 - Version: 
 - Platform:
   - [ ] android (tablet)
   - [ ] desktop (windows)
   - [ ] desktop (macOS)
   - [ ] server (windows)

### LINKED ISSUE #3526: Why can't we hold the shipments from stock-transfer?
## What went wrong? 😲
 
Why can't we hold the shipments from stock-transfer?

## Expected behaviour 🤔
should be allowed to hold (in the same way we can in desktop OG)

## How to Reproduce 🔨

1. Do stock transfer from one store to another
2. In receiving store, open the received Inbound
3. Try to hold that
4. See error: you simply can't
Only while creating new Inbound, you can hold it!

## Your environment 🌱

- Open mSupply Version: v1.7.0-rc4
- Legacy mSupply Central Server Version: v714-dev3

- Platform: browser (chrome), server exe

- Database type: sqlite

clemens-msupply:
Fixes #3526 (at least partially, still not taken store pref into account, please see my question in the issue)

The incoming invoice with status Shipped can now be put on hold. However, @mark-prins should it always be possible to put it on hold if it is not Verified?

### DIFF
--- a/client/packages/invoices/src/utils.ts
+++ b/client/packages/invoices/src/utils.ts
@@ -177,6 +177,7 @@ export const isOutboundDisabled = (
   );
 };
 
+/** Returns true if the inbound shipment cannot be edited */
 export const isInboundDisabled = (inbound: InboundRowFragment): boolean => {
   const isManuallyCreated = !inbound.linkedShipment?.id;
   if (isManuallyCreated) {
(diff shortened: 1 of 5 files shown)


client/packages/invoices/src/utils.ts
```diff
@@ -178,11 +179,21 @@ export const isOutboundDisabled = (
 
 export const isInboundDisabled = (inbound: InboundRowFragment): boolean => {
```
clemens-msupply, line 180: Rename to `isOnHold`? or what does `isInboundDisabled` mean? At least some docs comment is needed

mark-prins: we've used this naming elsewhere, and it doesn't necessarily mean 'on hold' so I think it is valid to keep 'isInboundDisabled'

just means that the user cannot interact / modify the invoice

mark-prins:
yes - as per the issue discussion; a shipped inbound can be put on hold. you cannot modify the other aspects though, so the 'isDisabled' function needs to be reverted to the same behaviour as previous. with the on hold functionality controlled differently

clemens-msupply:
> yes - as per the issue discussion; a shipped inbound can be put on hold. you cannot modify the other aspects though, so the 'isDisabled' function needs to be reverted to the same behaviour as previous. with the on hold functionality controlled differently

thanks, so hold can be set when the status is not `InvoiceNodeStatus.Delivered | InvoiceNodeStatus.Verified`? Or can it still be set when Delivered?

mark-prins:
good question - It should be available when delivered too. the use-case was that stock has been delivered but not put away yet, so the delivered status would matter too

clemens-msupply:
> good question - It should be available when delivered too. the use-case was that stock has been delivered but not put away yet, so the delivered status would matter too

Ok done

mark-prins (approved):
🙏  thanks @clemens-msupply looking good!

PR #3756, opened 2024-05-03
Update stock line on scan

mark-prins:
Fixes #2931

# 👩🏻‍💻 What does this PR do?

Change the expiry and barcode inputs to use a buffered input, and assign the value from the draft directly.

## 💌 Any notes for the reviewer?

# 🧪 Testing

- [ ] Open a stock line, scan a GTIN barcode
# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [x] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

PR #3809, opened 2024-05-07
Update stock line on scan

### LINKED ISSUE #2931: bar code scanning: data not displayed when after scan
## What went wrong? 😲

When scanning an item (2D data matrix) with a zebra barcode scanner from the stock line details (View Stock), none of the fields are displayed/fields are empty. You need to closed and opened the window for the data to be displayed. 

IMAGE 2024-02-14 11:09:43

## Expected behaviour 🤔

Data should be immediately displayed. 

Additional stuff: 

- [ ] `Barcode` should be renamed to Barcode/GTIN (Stock View) or a separate field for standard barcode and another one for GTIN ?
- [ ] #2947

## How to Reproduce 🔨

https://github.com/msupply-foundation/open-msupply/assets/74992958/df41c3cb-5ed6-4de9-bfdc-4ea11e627352

See above video. 

## Your environment 🌱

 - Version: 1.6.01
 - Platform:
   - desktop (macOS)

mark-prins:
Fixes #2931

# 👩🏻‍💻 What does this PR do?

Change the expiry and barcode inputs to use a buffered input, and assign the value from the draft directly.

This version has come from @roxy-dao's [comment](https://github.com/msupply-foundation/open-msupply/pull/3756#issuecomment-2097151642) - I rebased and that didn't go well, so have had to create a new branch and cherry pick instead.

# 🧪 Testing

- [ ] Open a stock line, scan a GTIN barcode
# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Stock/Components/StockLineForm.tsx
+++ b/client/packages/system/src/Stock/Components/StockLineForm.tsx
@@ -5,7 +5,6 @@ import {
   DateUtils,
   Formatter,
   TextWithLabelRow,
-  BasicTextInput,
   CurrencyInput,
   ExpiryDateInput,
   useTranslation,
@@ -18,6 +17,7 @@ import {
   Tooltip,
   useDebounceCallback,
   NumericTextInput,
+  BufferedTextInput,
 } from '@openmsupply-client/common';
 import { StockLineRowFragment } from '../api';
 import { LocationSearchInput } from '../../Location/Components/LocationSearchInput';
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks ok to me, Slightly worried after our discussions about buffered state and API queries. It looks like this is only updated when the OK button is clicked, and state won't update till the modal is re-opened, so I think we're ok (it's not auto save situation right?)

mark-prins:
true, it's not auto save, so it's not exactly an API buffering issue. agree that this is a little concerning, but it is much better than it was!

PR #3811, opened 2024-05-07
Fix warnings

clemens-msupply:
Cleanup of #2881 

# 👩🏻‍💻 What does this PR do?

Fix all warning, mostly from chrono but some that I believe existed in the develop branch before.

### DIFF
--- a/server/service/src/sync/translations/name.rs
+++ b/server/service/src/sync/translations/name.rs
@@ -149,7 +149,7 @@ impl SyncTranslation for NameTranslation {
 
     fn try_translate_from_upsert_sync_record(
         &self,
-        connection: &StorageConnection,
+        _: &StorageConnection,
         sync_record: &SyncBufferRow,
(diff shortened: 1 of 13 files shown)


server/service/src/sync/translations/name.rs
```diff
@@ -149,7 +149,7 @@ impl SyncTranslation for NameTranslation {
 
     fn try_translate_from_upsert_sync_record(
         &self,
-        connection: &StorageConnection,
+        _: &StorageConnection,
```
fergie-nz, line 152: I'm curious is there a need to keep these fields in when they aren't used?

clemens-msupply: They are defined in the SyncTranslation trait. You have to have this parameter in the impl to fulfil the "trait contract" (with the incorrect fn signature your struct doesn't impl the trait).

roxy-dao (approved):
Thanks Clemens! Nice clean console now :D

PR #3812, opened 2024-05-07
Coldchain cherry pick for 2.0.0-RC

### LINKED ISSUE #3679: Simplify Cold Chain Graphing
## The suggestion

Currently the cold chain graphs have been a significant source of bugs and confusion in open-mSupply.

I argue that a simpler graph view would serve the needs of the users better and reduce the confusion and complexity but would have some limitations.

I propose that chart should show every datapoint in the selected time range and use a dotted line to connect them.

Something similar to this...
https://master--63da8268a0da9970db6992aa.chromatic.com/?path=/docs/api-cartesian-line--docs

image

This is easy to understand, as you see all the datapoints from the system and the dotted line shows clearly that the gaps between the datapoints are interpolated not `real` data. The user can visually see if datapoints are further apart than previous ones indicating a gap, and can evaluate the severity of it.

To avoid the browser crashing with too much data, we'd need to limit the time range that the user can select.
This would mean if the user has a use case where they need to look at a long time range, they'd need to go to the dashboard to investigate (BTW: We know the dashboard can crash the browser by using too much memory when trying to draw too many datapoints on grafana too)

We've already sunk a lot of time in to the queries and graphs for cold chain data, yet bugs and confusion still abound.
See this for the latest: https://github.com/msupply-foundation/open-msupply/issues/3283

The current graphing approach isn't easy to understand for users, and developers need to keep checking data and logs to understand differences between the datapoints, the query logic, and the graphing view.

Alternative approaches with options to customise when something counts as a `gap` allowing the user to select different kinds of aggregations, etc all would result in a lot more work for us.

I think it reduces risks of mistakes and future bugs, but does potentially introduce new bugs with the updated code.

This feels like around 1 day's work to me, but how we limit the time range might be a little bit complex?
We'll need to evaluate the risks of trying to display too much data, particularly on a tablet, but I think in general people understand if something is slow when trying to look at a lot of data...

jmbrunskill:
Fixes #3679 
See https://github.com/msupply-foundation/open-msupply/pull/3770
This is a cherry pick because of branching from the wrong place :(

# 👩🏻‍💻 What does this PR do?

Updates Cold Chain Graph


andreievg:
Hi, sorry this might have a little bit of bias.

Pre requisite to the below, is that my tablet freezes with these charts. See below

I only just now caught up to the changes to the charts, and i had a few concerns, mainly because the original reason for the aggregated charts was not really considered, and the idea is:

"There are hardware and real estate constraints" for the sites that will be using open mSupply. 

Hardware constraints would mean we can't rended too many data point, and real estate constraint suggests that we can only render so much for user to be able to digest on one screen. This was the main reason with the aggregation.

As far as I was aware, the aggregated charts worked, but I only now noticed there was a change that was causing different data to be returned on every request call and potentially performance issues (calling min/max for breach for every breach id linked to every point).

Yes there are use cases where there could be gaps in data, but I think if the charts were rendered to present aggregated data it would be quite usable and indicative (there is an issue that i linked quite a few time on the cold chain app that suggests how aggregated charts can be presented). And maybe at a certain 'zoom' we can switch to by log view (maybe less then 100 data points).

Anyways the real reason about writing this up, is that I completely missed the train on commenting to the original issue. And was wondering how did we pivot so quickly to put a pretty major change like this to 2.0 ? I thought our testing load is already stretched, and this would need to get a pretty thorough test. 
I've tried importing: https://github.com/msupply-foundation/temperature-sensor/blob/main/data/FridgeTag%202L/130500109088_202206081014.txt on a samsung and looking at charts freezes my tablet, try looking back to 2021 (actually, it doesn't quite freeze the tablet, i clicked on a breach and did actually see it but after some minutes).

Form that data set, if you try one day 01-06-2022 - 02-06-2022, it's quick and responsive. It was looking pretty good when i tried the whole of June, started getting quite slow from 20-05-2022 - 09-06-2022, and very slow from 01-05-2022 - 09-06-2022 (but wasn't freezing). The freezing is only when all of the data is squashed, and it's on the large scale (i.e. 01-01-2021 - 09-06-2022)

PR #3814, opened 2024-05-07
3736 bug: OS issue expired line value disappears on click

### LINKED ISSUE #3736: OS: Issued lined disappears when expired in
## What went wrong? 😲

When issuing stock from expired stock line, after re-opening OS line edit modal, and clicking somewhere, the line will disappear

https://github.com/msupply-foundation/open-msupply/assets/26194949/9a3364a2-debc-4702-818b-8912ffdbfc8f

Video is from Kopu store on demo store: https://demo-open.msupply.org/distribution/outbound-shipment/139?sort=itemName

## Expected behaviour 🤔

Line remains unchanged when re-opening

## How to Reproduce 🔨

In OS issue some stock from expired stock line in OS line edit, press ok to exit modal, re-open modal by clicking on the line, click anywhere (even ok button) the line will disappear from view

## Your environment 🌱

V2.0 RC branch

lache-melvin:
Fixes #3736

# 👩🏻‍💻 What does this PR do?

Prevents the auto allocation from triggering when the `issue quantity` input field loses focus:

https://github.com/msupply-foundation/open-msupply/assets/55115239/a4bb2112-5171-4f66-949b-366bba05f5b4

NumericTextInput component calls the handleIssueQuantityChange on blur, which was triggering auto allocation even though we've not actually changed the value.

- Also fixes for prescriptions, as same bug was there

Somewhat curious for thoughts on whether this could/should be the NumericTextInput's responsibility, to only call the onChange handler from `onBlur` if there actually is a change? 🤷‍♀️ If I were passing an onChange handler to a component, I wouldn't expect it to fire on blur in the first place, but if it does, I would expect it to be because the change in focus has updated the value in some way?

# 🧪 Testing

- [ ] Add a line to outbound shipment, issue stock from a line that the auto allocation wouldn't choose (e.g. expired)
- [ ] Click ok
- [ ] Reopen edit modal by clicking that line
- [ ] Click anywhere to shift focus away from the issue quantity input
- [ ] Your allocations don't change

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/StockOut/utils.tsx
+++ b/client/packages/invoices/src/StockOut/utils.tsx
@@ -344,7 +344,7 @@ export const PackQuantityCell = (props: CellProps<DraftStockOutLine>) => (
     {...props}
     max={props.rowData.stockLine?.availableNumberOfPacks}
     id={getPackQuantityCellId(props.rowData.stockLine?.batch)}
-    min={1}
+    min={0}
   />
 );
 
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
@@ -147,6 +147,10 @@ export const OutboundLineEditForm: React.FC<OutboundLineEditFormProps> = ({
   );
 
   const handleIssueQuantityChange = (quantity: number | undefined) => {
+    // this method is also called onBlur... check that there actually has been a change
+    // in quantity (to prevent triggering auto allocation if only focus has moved)
+    if (quantity === issueQuantity) return;
+
     setIssueQuantity(quantity ?? 0);
     setOkDisabled(true);
     debouncedAllocate(
--- a/client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditForm.tsx
@@ -146,6 +146,10 @@ export const PrescriptionLineEditForm: React.FC<
   };
 
   const handleIssueQuantityChange = (inputQuantity?: number) => {
+    // this method is also called onBlur... check that there actually has been a change
+    // in quantity (to prevent triggering auto allocation if only focus has moved)
+    if (inputQuantity === issueQuantity) return;
+
     const quantity = inputQuantity === undefined ? 0 : inputQuantity;
     setIssueQuantity(quantity);
     allocate(quantity, Number(packSizeController.selected?.value));


client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -344,7 +344,7 @@ export const PackQuantityCell = (props: CellProps<DraftStockOutLine>) => (
     {...props}
     max={props.rowData.stockLine?.availableNumberOfPacks}
     id={getPackQuantityCellId(props.rowData.stockLine?.batch)}
-    min={1}
+    min={0}
```
lache-melvin, line 347: I KNOW side quest bad BUT:
  - such tiny change 🥺 
  - This input is only used by these two modals (Prescription and OS edit line)
  - was preventing you from being able to clear an issue quantity once you set one

lache-melvin: Ah this has actually been raised as a separate bug: #3834

andreievg:
> Somewhat curious for thoughts on whether this could/should be the NumericTextInput's responsibility, to only call the onChange handler from onBlur if there actually is a change?

I feel like there is quite a bit of logic already in the NumberTextInput. Naturally I don't expect onBlur to be calling onChange, and actually any state change 'onBlur' are difficult, because you can always click on a button straight from focus, at which point the behaviour is undefined (we've added a lot of code in Conforma to deal with this use case actually).

Agree that ideally onChange is only called when value changed. 

Another thing that confusing is that onChange will be called if defaultValue is set (with the defaultValue) when component starts ? I think if we are creating abstraction that applies to a lot of use cases, we should have in depth doc to describe the full behaviour and the edge case (vs looking at code), I think @CarlosNZ mentioned that some test cases have been added, maybe that's one way to describe the behaviour

Also I don't understand the onBlur code, it looks like if text can't be parsed, we call onChange with undefined but setText as value ¯\_(ツ)_/¯, @CarlosNZ, do you have any thoughts on this ?

```ts
const parsed = parse(textValue ?? '');
onChange(Number.isNaN(parsed) ? undefined : parsed);
setTextValue(formatValue(value));
```

andreievg:
I wonder what the balance is for something like NumberTextInput. 
I reckon instead of correcting and restricting the input, just have an error state for it:

onChange in input
Is the number ok ?
Yes
  reset ErrorState
  call onChangeProp (for consumer)
No
  set ErrorState

onBlur
Is the number ok ?
Yes
  format input

onMount
Is the number ok ?
Yes
  format input
  reset ErrorState
No 
  set ErrorState 
  
I am probably missing some requirement, but wondering, for that requirement ?

andreievg (approved):
Thanks, changes look great

CarlosNZ:
> I reckon instead of correcting and restricting the input, just have an error state for it:

The problem with that is that you don't want the user to experience an Error state UI when in the process of typing a valid number. For example, starting a negative number with "-" gives an invalid state, or a decimal point without having typed the next digits yet. 

And you should never be able to type truly inappropriate input, such as letters. 

In my view, it's better user experience to prevent bad input for things like numbers than to hammer the user with a bunch of irritating errors for every little typo.

Similarly for things like max/min. Why even let the user type a too-big number and then expect them to fix it? The component should just fix it for them (by restricting the input).

Just like when you have to enter a value in ALL CAPS -- the input should put your input in CAPS, not let you type lower case and then give you a hard time and make you change it 😆.

andreievg:
> Similarly for things like max/min. Why even let the user type a too-big number and then expect them to fix it? The component should just fix it for them (by restricting the input).

I think that might be an example where it could be hard for user to enter, or switch between max and min. As per our convo, it's good that we don't have to support that atm (negative number i think).

> Just like when you have to enter a value in ALL CAPS -- the input should put your input in CAPS, not let you type lower case and then give you a hard time and make you change it 

AGREEE

> The problem with that is that you don't want the user to experience an Error state UI when in the process of typing a valid number. For example, starting a negative number with "-" gives an invalid state, or a decimal point without having typed the next digits yet.

Sure, I can see this being bad UX. I am a bit wary of the complexity of number inputs, and actually the logic of the NumericTextInput, takes awhile to understand. But i guess extensive test suits + consumer docs can help mitigate that.

PR #3819, opened 2024-05-07
Make the supplier name website look like another, disabled text input

### LINKED ISSUE #3426: 'Website' category does not have a text box allocated under Replenishment -> Suppliers
## What went wrong? 😲
 

When clicking on a specific supplier, a window with details appears. One of the categories you can see is 'Website' - it is the only category missing a grey-ish text box next to it, so in case there is information to display, it would be within that box

photo_2024-03-26 10 36 31

## Expected behaviour 🤔

For consistency, I think 'Website' should have a box allocated to it too, even if the field is left empty like in other cases on the same screen. 

## How to Reproduce 🔨

1. Go to Replenishment -> Suppliers
2. Click on a specific Supplier to bring up a detailed view
3. Note the 'Website' category at the bottom of the screen does not have a grey text box allocated to it

## Agreed Solution 
Either: 
1. Don't show the website label and data if there is none. 
2. Indicate that no website link has been provided with `None provided`

## Your environment 🌱

- Version: 1.7.0-RC3

- Platform: Windows Server 2022 Desktop

- Database type: SQLite3

clemens-msupply:
Fixes #3426

# 👩🏻‍💻 What does this PR do?

Make the supplier name website look like another, disabled text input

I tried to make it same with and height and same look as the other text inputs.

# Testing

There are three cases to consider:
- no website
- short website
- long website

image

image

image

### DIFF
--- a/client/packages/system/src/Name/DetailModal/DetailModal.tsx
+++ b/client/packages/system/src/Name/DetailModal/DetailModal.tsx
@@ -75,14 +75,10 @@ export const DetailModal: FC<DetailModalProps> = ({ nameId }) => {
                   target="_blank"
                   rel="noopener"
                   sx={{
-                    // truncate long links
-                    overflow: 'hidden',
-                    textWrap: 'nowrap',
-                    textOverflow: 'ellipsis',
                     // Make it look like another disabled text input consistence with other text
                     // input components in this modal
-                    width: '198',
-                    height: '34.13',
+                    width: '100%',
+                    minHeight: '34.13',
                     backgroundColor: theme => theme.palette.background.toolbar,
                     borderRadius: '8px',
                     padding: '4px 8px',
--- a/client/packages/system/src/Name/DetailModal/DetailModal.tsx
+++ b/client/packages/system/src/Name/DetailModal/DetailModal.tsx
@@ -68,16 +68,30 @@ export const DetailModal: FC<DetailModalProps> = ({ nameId }) => {
             <DetailInputWithLabelRow
               label={t('label.website')}
               inputProps={{ value: data?.website, disabled: isDisabled }}
+              inputAlignment="start"
               DisabledInput={
-                <>
-                  <MuiLink
-                    href={data.website ?? undefined}
-                    target="_blank"
-                    rel="noopener"
-                  >
-                    {data.website}
-                  </MuiLink>
-                </>
+                <MuiLink
+                  href={data.website ?? undefined}
+                  target="_blank"
+                  rel="noopener"
+                  sx={{
+                    // truncate long links
+                    overflow: 'hidden',
+                    textWrap: 'nowrap',
+                    textOverflow: 'ellipsis',
+                    // Make it look like another disabled text input consistence with other text
+                    // input components in this modal
+                    width: '198',
+                    height: '34.13',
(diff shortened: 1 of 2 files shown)


client/packages/system/src/Name/DetailModal/DetailModal.tsx
```diff
@@ -68,16 +68,30 @@ export const DetailModal: FC<DetailModalProps> = ({ nameId }) => {
...
-                </>
+                <MuiLink
+                  href={data.website ?? undefined}
+                  target="_blank"
+                  rel="noopener"
+                  sx={{
+                    // truncate long links
+                    overflow: 'hidden',
+                    textWrap: 'nowrap',
+                    textOverflow: 'ellipsis',
```
roxy-dao, line 81: I think it would be nice to wrap the website and show full URL? Think it should be fine if users don't start entering some weird super long URL.... @mark-prins what do you think? 👀 
Screenshot 2024-05-08 at 14 05 43

client/packages/system/src/Name/DetailModal/DetailModal.tsx
```diff
@@ -68,16 +68,30 @@ export const DetailModal: FC<DetailModalProps> = ({ nameId }) => {
...
+                    // Make it look like another disabled text input consistence with other text
+                    // input components in this modal
+                    width: '198',
```
roxy-dao, line 84: Was there a reason for these odd numbers? 👀 Could we have them align with the other fields? 
Screenshot 2024-05-08 at 14 09 06

roxy-dao (changes requested):
Looking much better thanks Clemens!

Have two questions below (:

client/packages/system/src/Name/DetailModal/DetailModal.tsx
clemens-msupply: these numbers worked for me and were required if we would use a ellipsis to shorten the URL. But as said not happy about it. Another option would be to wrap the url. Is this what you show in the screenshot?

roxy-dao: Yup!

clemens-msupply:
This is how it looks now:

image

This also increases the width of the text input (which was limited by the default <input> size previously)

There were some changes to the DetailedInputWithLabelRow. The DetailedInputWithLabelRow component was mainly used for Jsonfroms where it still looks fine. For item details view it IMHO looks better now:

Before:
image

After:
image

client/packages/system/src/Name/DetailModal/DetailModal.tsx
```diff
@@ -68,16 +68,26 @@ export const DetailModal: FC<DetailModalProps> = ({ nameId }) => {
...
-                </>
+                <MuiLink
+                  href={data.website ?? undefined}
+                  target="_blank"
+                  rel="noopener"
+                  sx={{
+                    // Make it look like another disabled text input consistence with other text
+                    // input components in this modal
+                    width: '100%',
+                    minHeight: '34.13',
```
roxy-dao, line 81: Nice!

roxy-dao (approved):
Thanks Clemens!

It looks a lot better now :D

PR #3821, opened 2024-05-07
3771 Have to click Ok twice if permission is denied

### LINKED ISSUE #3771: Have to click `Ok` twice if permission is denied
## What went wrong? 😲
If I go to a page which requires a permission that I don't have then I have to click `Ok` twice to dismiss the error

https://github.com/msupply-foundation/open-msupply/assets/61820074/53d17602-2211-4f7b-9c67-1fbb199ac050

## Expected behaviour 🤔
Should only have to press `Ok` once

## How to Reproduce 🔨

Steps to reproduce the behaviour: Easiest way to recreate is to turn off some permission and go to the page associated with that permission

## Your environment 🌱

- Open mSupply Version: V2.0.0

- Platform:

- Database type:

roxy-dao:
Fixes #3771

# 👩🏻‍💻 What does this PR do?
Remove toggleOff since it was causing the AuthErr modal to pop up twice

# 🧪 Testing

- [ ] Remove some permission (e.g. View assets)
- [ ] Try to go to asset page
- [ ] Should get AuthErr -> click ok
- [ ] Should disappear
- [ ] Wait for timeout 
- [ ] Same as step 3 & 4

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/hooks/useEditModal/useEditModal.ts
+++ b/client/packages/common/src/hooks/useEditModal/useEditModal.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState } from 'react';
 import { useToggle } from '../useToggle';
 
 export enum ModalMode {
@@ -20,20 +20,17 @@ export const useEditModal = <T>(): EditModalState<T> => {
   const [entity, setEntity] = useState<T | null>(null);
   const [mode, setMode] = useState<ModalMode | null>(null);
 
-  const onOpen = useCallback(
-    (entity: T | null = null) => {
-      setEntity(entity);
-      setMode(entity ? ModalMode.Update : ModalMode.Create);
-      modalControl.toggleOn();
-    },
-    [modalControl]
-  );
+  const onOpen = (entity: T | null = null) => {
+    setEntity(entity);
+    setMode(entity ? ModalMode.Update : ModalMode.Create);
+    modalControl.toggleOn();
+  };
 
-  const onClose = useCallback(() => {
+  const onClose = () => {
     setMode(null);
     setEntity(null);
     modalControl.toggleOff();
-  }, [modalControl]);
+  };
 
   return {
     onOpen,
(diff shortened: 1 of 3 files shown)


client/packages/host/src/components/ErrorAlert.tsx
```diff
@@ -48,7 +48,6 @@ export const ErrorAlert = () => {
     }
 
     if (error === AuthError.PermissionDenied) {
-      toggleOff();
```
roxy-dao, line 51: Was added to prevent alert being shown twice but is making the alert show twice 😢

clemens-msupply (approved):
Thanks works fine and looks good to me.

However, the indirect cause of the problem actually lies in `useToggle`:

```ts
export const useToggle = (initial = false): ToggleState => {
  const [isOn, setToggled] = useState(initial);

  const toggleOn = () => setToggled(true);
  const toggleOff = () => setToggled(false);
  const toggle = () => setToggled(state => !state);

  return { isOn, toggle, toggleOn, toggleOff, setToggled };
};
```

This means all the toggle* functions are changing with every call (useCallback is needed to cache them).

In the ErrorAlert you do this:
```ts
useEffect(() => {
    if (!!error) toggleOn();
    return () => toggleOff();
  }, [error, toggleOff, toggleOn]);
```
i.e. you re-enable it after calling toggleOff...

So you could improve the useToggle hook...

client/packages/common/src/hooks/useEditModal/useEditModal.ts
```diff
@@ -20,20 +20,17 @@ export const useEditModal = <T>(): EditModalState<T> => {
...
-      setMode(entity ? ModalMode.Update : ModalMode.Create);
-      modalControl.toggleOn();
-    },
-    [modalControl]
-  );
+  const onOpen = (entity: T | null = null) => {
+    setEntity(entity);
+    setMode(entity ? ModalMode.Update : ModalMode.Create);
+    modalControl.toggleOn();
+  };
```
clemens-msupply, line 27: Are you sure you can remove this useCallback? this might trigger some other problems of useEditModal users...

The modalControl object is also not cached only the members... So this useCallback should probably better depend on toggleOn and not on the wrapping modalControl object...

Might miss why you did this change.

roxy-dao: Had discussion on telegram, will merge now since changes have been reverted

PR #3825, opened 2024-05-07
3485 uncounted packs should show uncounted

### LINKED ISSUE #3485: When creating a new stocktake with 'Items with stock' checked ON, the counted packs should be equal to the snapshot packs instead of 0
## What went wrong? 😲
 
When creating a new stocktake with `Items with stock` checked ON, the `counted packs` should be equal to the `snapshot packs` instead of 0.
Rn, the `counted packs` is 0 by default and the difference(`snapshot packs` - `counted packs`) still says 0, which is mathematically wrong and looks buggish to me!
image

## Expected behaviour 🤔
When creating a new stocktake with `Items with stock` checked ON, the `counted packs` should be equal to the `snapshot packs` instead of 0.

## How to Reproduce 🔨
1. HAve some stocks in your oms-store
2. Create a new stocktake with the `Items with stock` checked ON and proceed
3. See first error: despite the `counted packs` 0, the difference is still 0.
4. Now let's say you wanna reduce some stocks to 0: so actually open up one of the lines.
5. Since the `counted packs` is already 0, do nothing and just click OK, it's still blue (like unchanged,).
6. Now open that line again and re-enter the value 0 (despite it's already there) and click OK; now the line's black (updated) and this the difference also updated - I call this bad UX experience!

## Your environment 🌱

- Version: `v1.7-rc3 exe`
- mSupply version: v713

- Platform: browser(chrome), server

- Database type: sqlite

## Additional
Spotted while [v1.7 release-testing](https://github.com/msupply-foundation/open-msupply/issues/3341)

lache-melvin:
Fixes #3485

# 👩🏻‍💻 What does this PR do?

Shows `[uncounted]` in detail page, and empty input (rather than 0) in modal for stocktake lines that haven't been counted yet:

Edit: show `-` instead of `uncounted`:
image

Screenshot 2024-05-08 at 11 38 47 AM

# 🧪 Testing

- [ ] Create new stocktake for items with stock
- [ ] See Counted # packs for each line is `[uncounted]`
- [ ] Click on a line to open edit modal
- [ ] See Counted # packs for each line is empty
- [ ] Add counted packs for lines
- [ ] Once you enter counted packs, you can input as 0, but cannot go back to empty
- [ ] Click OK
- [ ] Counted packs (including 0) display correctly in the detail table

# 📃 Documentation

  1. Stocktake screenshots


client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberCell.tsx
```diff
@@ -11,10 +11,18 @@ export const NumberCell = <T extends RecordWithId>({
   column,
   rowData,
   isError,
-}: CellProps<T>) => {
+  defaultValue,
```
CarlosNZ, line 14: I wonder if we should set a default `defaultValue` of empty string? Currently it displays `0` if a default isn't defined, which seems slightly misleading.

client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberCell.tsx
```diff
@@ -11,10 +11,18 @@ export const NumberCell = <T extends RecordWithId>({
...
+}: CellProps<T> & {
+  defaultValue?: string | number;
+}) => {
   const value = column.accessor({ rowData }) as number;
   const formattedValue = useFormatNumber().round(value, 2);
 
+  const displayValue =
+    (value === undefined || value === null) && !!defaultValue
```
CarlosNZ, line 22: We need to explicitly check that `defaultValue !== undefined` here, as falsy values (0, "") return `false` for `!!`, and so setting a default of `""` displays as `0`. ;)

CarlosNZ: Although with my earlier suggestion (default `defaultValue=""`), you won't need to check `defaultValue` at all here.

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
```diff
@@ -199,7 +199,19 @@ export const BatchTable: FC<
...
+      Cell: props => {
+        const value = props.rowData.countedNumberOfPacks;
+        return (
+          <NumberInputCell
+            {...props}
+            decimalLimit={2}
+            min={0}
+            // "Formatting" changes null/undefined (uncounted) value to 0
+            // but we want uncounted to render an empty input cell
+            noFormatting={value === null || value === undefined}
```
CarlosNZ, line 211: This shouldn't be necessary with the fixes done to `NumericTextInput` that are in `develop`. (And if it's not, we should fix it in that component rather than misusing the `noFormatting` prop here -- but I'm 90% sure it is fixed)

CarlosNZ (changes requested):
Nice :) Couple of suggestions, and I don't think you'll need the `noFormatting` prop once this is in develop -- see comment.

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
CarlosNZ: I'm tempted to suggest you copy and paste the whole `NumericTextInput` component from `develop` into this branch so we can have those fixes in 2.0.0 as well? What do you think? Maybe at least try it out to confirm that it fixes this problem here?

lache-melvin: Just pulled the latest develop in (this branch is going into develop) but not resolved there... will have a play moving this into that component though!

lache-melvin: Do you this it's behaviour we would want for all NumericTextInputs? That if the value is undefined the input renders empty? Or some additional flag?

CarlosNZ: > That if the value is undefined the input renders empty?

Yes, that's the intended behaviour

lache-melvin: Oh found it! It was handling undefined, but not null 👍

CarlosNZ: Yeah, React doesn't like `null` as a value in form components -- that's what often gives that "changing from controlled to uncontrolled component" error.

CarlosNZ (approved):
Cool, looks good. Regarding that "goes to 0 on focus" issue we discussed, there's a couple of things we can do. Up to you if you want to fix it here or leave it for another issue:

- in NumericTextInput, change the `onBlur` to:  
  ```
  onBlur={() => {
          const parsed = parse(textValue ?? '');
          const val = Number.isNaN(parsed) ? undefined : parsed;
          onChange(val);
          setTextValue(formatValue(val));
        }}
  ```
  That just makes sure the "textValue" gets the new value, not the previous state value (This needs to be done at some point anyway, it doesn't really help with this particular issue, it's just a mistake that should be remedied at some point)
- In NumberInputCell, change the `onChange` to:  
  ```
      onChange={num => {
        setBuffer(num);
        updater({ ...rowData, [column.key]: num });
      }}
  ```
  This does mean that the row state management will have to be able to handle the `undefined` value, so not sure if this would affect other areas.

lache-melvin:
> Regarding that "goes to 0 on focus" issue we discussed, ...

Took a slightly different approach ([14d431d](https://github.com/msupply-foundation/open-msupply/pull/3825/commits/14d431d727fdf117669cb0a30bcb17cc17326578)) - rather than removing the min/undefined handling from the NumberInputCell, added an isDirty check in the NumericTextInput?

Doesn't solve the "I want to backspace back to empty again" issue, but that feels like a separate issue... and I think this might be nicer than pushing the undefined handling out everywhere? What do you think?

CarlosNZ (approved):
LGTM! Thanks

PR #3826, opened 2024-05-07
3801 Internal Order Page has misaligned "Filter Items" searchbox and "Actions" drop down

### LINKED ISSUE #3801: Internal Order Page has misaligned "Filter Items" searchbox and "Actions" drop down
## What went wrong? 😲

image

image

image

## Expected behaviour 🤔

Things should be aligned otherwise they look bad.

(whine: I'm gonna guess certain styles are inherited, rather than added in via composition. So things have correct padding in one context but not another...)

## How to Reproduce 🔨

1. view an internal order

## Your environment 🌱

- Open mSupply Version: 2.0.00 on demo-open.msupply.org (IDK what RC)

roxy-dao:
Fixes #3801

# 👩🏻‍💻 What does this PR do?
Forces height on grid so that the filter and actions align 👀 
Screenshot 2024-05-08 at 11 41 40

# 🧪 Testing

- [ ] Go to Internal Order
- [ ] Create a new Internal Order
- [ ] See they should be aligned on tool bar 👀 

# 📃 Documentation

  1. Maybe screenshots if they're misaligned
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
@@ -129,7 +129,7 @@ export const Toolbar: FC = () => {
         display="flex"
         gap={1}
         justifyContent="flex-end"
-        sx={{ marginTop: 2 }}
+        sx={{ marginTop: 2, height: 37 }}
       >
         <SearchBar
           placeholder={t('placeholder.filter-items')}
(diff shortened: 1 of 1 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
```diff
@@ -129,7 +129,7 @@ export const Toolbar: FC = () => {
         display="flex"
         gap={1}
         justifyContent="flex-end"
-        sx={{ marginTop: 2 }}
+        sx={{ marginTop: 2, height: 37 }}
```
Chris-Petty, line 132: I'm too rusty on react but generally speaking adding random inline heights in the UI makes me sad 😢. Why does the search bar seemingly have padding above it? Or is it the DDL has margin below it? Ideally things would just grow to the container and the container aligns them :(

roxy-dao: Have removed and done with with some flexing instead 💪🏻

Chris-Petty (approved):
m a n i f i c e n t 

image

PR #3842, opened 2024-05-09
#3561 Fix Column show/hide behaviour

### LINKED ISSUE #3561: Column hide/show not working properly
## What went wrong? 😲
Pressing on the select box causes some hidden columns to pop back up

https://github.com/msupply-foundation/open-msupply/assets/61820074/75396ffc-9b78-45a1-b3cf-2718b7748fa5

## Expected behaviour 🤔
Columns should stay hidden.

## How to Reproduce 🔨

1. Go to any table
2. Hide some columns (maybe 4-5?)
3. Click on select box
4. See columns pop back up

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #3561

# 👩🏻‍💻 What does this PR do?

This took a lot longer than I initially thought due to having to make it elegantly handle the post-load appearance of additional columns due to plugin loading delay.

I ended up abstracting all the column state logic into a separate hook (`useColumnDisplayState`) which just returns the overall display state and a `toggleColumn` method -- this consolidates a lot of the logic that was spread between the `ColumnPicker` component and its parent, `DataTable`.

In order to handle the disappearance/appearance of columns, I found that it was not enough to just have a list of display columns OR hidden columns -- we need to know the state of *both* (so we can detect *new* columns). Hence I settled on storing the current display state in an object that holds the state of every (known) column, e.g.
```js
{
  name: true,
  itemCode: false,
  expiryDate: true,
  // ...etc.
}
```

See inline comments

# 🧪 Testing

I made a build of the front-end which pulled in plugins from the server, as per @mark-prins 's suggestion, which allowed me to see the "delay" that a plugin column has in when the table loads.

I was able to replicate something close to this in dev mode with the following:
- Copy the "StockDonor" plugin folder from the [Plugins repo](https://github.com/msupply-foundation/open-msupply-plugins/tree/main/examples/StockDonor) to `client/packages/plugins`
- Restart the front end (`yarn start`)
- Go to the "View Stock" table -- you should see an additional "Donor" column in the table
- You can drag the StockDonor folder out of the plugins folder while the app is running in order to see the effect of column removal (it'll complain in the console, but it works for this purpose)

You might want to keep an eye on the localStorage while you test it out.

After clearing localStorage/@openmsupply-client/columns/hidden to start fresh, and with the StockDonor plugin active:

- [ ] Table loads with all columns showing, and all columns show in the Column picker popover (initially all ticked)
- [ ] Toggle some columns on and off and check behaviour is as expected
- [ ] Change the column state (click a sort header) and check the hidden state remains unchanged
- [ ] Refresh the page and check the correct columns are still hidden on reload
- [ ] Hide the "Donor" column, check that it remains hidden after the sort and refresh
- [ ] Remove the "StockDonor" plugin -- ensure the columns and the picker reflect the change (this one *may* require a refresh, but I think this is okay)
- [ ] Put the "StockDonor" plugin back -- it should appear in the ColumnPicker again with the same visibility status as it did previously


client/packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -131,13 +132,15 @@ const DataTableComponent = <T extends RecordWithId>({
...
-    () =>
-      columns.filter(c => displayColumns.map(({ key }) => key).includes(c.key)),
-    [displayColumns, columns]
+  const { columnDisplayState, toggleColumn } = useColumnDisplayState(
+    id,
+    columns
   );
 
+  const columnsToDisplay = React.useMemo(() => {
+    return columns.filter(c => columnDisplayState[String(c.key)] ?? true);
```
CarlosNZ, line 141: The fallback ensures that if a new column appears that isn't yet in the display state, it will appear in the table.

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
```diff
@@ -101,13 +73,15 @@ export const ColumnPicker = <T extends RecordWithId>({
           <Typography style={{ fontWeight: 700 }}>
             {t('table.show-columns')}
           </Typography>
-          {Object.values(columns)
+          {columns
```
CarlosNZ, line 76: I don't know why this was here -- `columns` is already an array so `Object.values` does nothing as far as I can tell 🤷‍♂️

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
```diff
@@ -101,13 +73,15 @@ export const ColumnPicker = <T extends RecordWithId>({
...
             .filter(c => !!c.label)
             .map(column => (
               <FormControlLabel
                 key={String(column.key)}
-                checked={isVisible(column)}
-                control={<Checkbox onClick={() => toggleColumn(column)} />}
+                checked={columnDisplayState[column.key] ?? true}
```
CarlosNZ, line 81: Again, the fallback ensures that any new columns appear as enabled initially. It doesn't actually matter that they're not explicitly defined in the display state yet -- once they get toggled for the first time, it'll be inserted.

client/packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -180,12 +183,6 @@ const DataTableComponent = <T extends RecordWithId>({
     if (page * first > total) onChangePage(0);
   }, [pagination]);
 
-  // if the columns array changes, such as when a plugin column is added
-  // reset the display columns to the full set
```
CarlosNZ, line 184: No need to reset all columns now -- the new one appears and all other columns retain their visibility status.

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
```diff
@@ -55,30 +42,15 @@ export const ColumnPicker = <T extends RecordWithId>({
...
     <>
       <Tooltip title={t('table.show-columns')}>
         <IconButton onClick={handleClick} aria-describedby={id}>
           <ColumnsIcon
             sx={{
-              color:
-                getHiddenColumns().length > 0 ? 'secondary.main' : undefined,
+              color: !columns.every(c => columnDisplayState[String(c.key)])
+                ? 'secondary.main'
+                : undefined,
```
lache-melvin, line 53: I think super minor, up to you if you want to investigate or not... but for some reason this doesn't work quite right with plugin column on initial mount:

Per video, all fields are selected, but column icon is blue. Stays blue until you uncheck a recheck the Donor column? Goes back to being blue if you refresh...

https://github.com/msupply-foundation/open-msupply/assets/55115239/1d87e946-24f8-45ba-bfff-644e0303c4f7

lache-melvin (approved):
Looks great Carl! Thanks for this, awesome to have the columns a little more robust 🔥 

Just the one issue with the column icon, assuming because the plugin columns are loaded in after... not sure if it's worth the rabbit hole trying to fix 🤷‍♀️

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
CarlosNZ: > I think super minor, up to you if you want to investigate or not... but for some reason this doesn't work quite right with plugin column on initial mount:

Good catch -- I got the other two cases where the column is missing from the state object, missed this one.

Easily fixed: 15f206c27

Thanks 🙏

lache-melvin:
Schweeeet ship it!

client/packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -131,13 +132,15 @@ const DataTableComponent = <T extends RecordWithId>({
...
-    [displayColumns, columns]
+  const { columnDisplayState, toggleColumn } = useColumnDisplayState(
+    id,
+    columns
   );
 
+  const columnsToDisplay = React.useMemo(() => {
+    return columns.filter(c => columnDisplayState[String(c.key)] ?? true);
+  }, [columns, columnDisplayState]);
+
```
mark-prins, line 143: :nit tidier without the return
```suggestion
  const columnsToDisplay = React.useMemo(
    () => columns.filter(c => columnDisplayState[String(c.key)] ?? true),
    [columns, columnDisplayState]
  );
```

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
```diff
@@ -55,30 +42,21 @@ export const ColumnPicker = <T extends RecordWithId>({
...
-                getHiddenColumns().length > 0 ? 'secondary.main' : undefined,
+              color: !columns.every(
+                c =>
+                  columnDisplayState[String(c.key)] ||
+                  // If a column has no state (undefined), it must be a new
+                  // column, so we treat it as visible
+                  columnDisplayState[String(c.key)] === undefined
+              )
+                ? 'secondary.main'
+                : undefined,
```
mark-prins, line 59: :nit this would be equivalent wouldn't it? To me it's easier to read, not requiring the negation on every. 

```suggestion
              // If a column has no state (undefined), it must be a new
              // column, so we treat it as visible
              color: columns.some(
                c => columnDisplayState[String(c.key)] === false
              )
                ? 'secondary.main'
                : undefined,
```

client/packages/common/src/ui/layout/tables/components/Header/ColumnPicker.tsx
mark-prins: odd - can't see why it would be required.

mark-prins (commented):
Nice - looks good, thanks @CarlosNZ 
Just added some nits while reading through. overall looks great

CarlosNZ:
> Just added some nits while reading through. overall looks great

Great, thanks for taking a look and the suggestions. 910a1266f

PR #3846, opened 2024-05-09
3830 report manifest file

### LINKED ISSUE #3830: Add report_builder support to include report metadata

There are a couple of new requirements:
- SQL reports need a way to assign permissions to a report #3383 
- Report `context` shouldn't be needed to be added manually when uploading a report https://github.com/msupply-foundation/open-msupply-reports/issues/67
- Report versioning #3718

All these need a way to add extra data to the report, e.g. permission need the om_context name or versioning needs the report version and a unique report id.

This can be done by adding a `manifest.json` file that can contain all additional metadata. If this file is provided the file will be automatically added to the `ReportManifest::entries` map. This already will work with the existing code, i.e. the manifest.json file is just another resource. This also means the manifest is also available in the report Tera template, e.g. you could print the version by accessing `res.manifest.version`.

However, it may still be useful to handle the `manifest.json` file in a special way. For example, the report `context` option should be mandatory.

Add cli flags for all new option. But this would result in lengthy and hard to manage options

### Moneyworks Jobcode 🧰

OMS:JAN24

clemens-msupply:
Fixes #3830 

# 👩🏻‍💻 What does this PR do?

You can now add a `manifest.json` file to your report project directory. Currently, this manifest file only contains the report context. See Readme for some more details...

I also sneaked in support for specifying the store name instead of the store id when printing a report from the cli. It can be annoying to find out the store id and it was easy to add store name support...

# 🧪 Testing

Please follow the instruction from the readme to build and print the example report (test if instructions are clear).

Check that the build json report contains the manifest data.
The printed pdf should also display the manifest data.

# 📃 Documentation
I updated the readme, but we might add it to some internal docs?


Chris-Petty:
Action: `  = note: /usr/bin/ld: final link failed: No space left on device`

Tests are passing on my machine - I just want to do a manual test of the functionality first though!

Chris-Petty (approved):
Gave a wee whirl looks good to me

PR #3847, opened 2024-05-09
3332 shipment transfer to invoice transfer

### LINKED ISSUE #3332: Update Shipment transfer to generic Invoice transfer (3h)
## The suggestion

The transfer logic before inbound/outbound shipments has been reused for returns, as the invoices are very similar.

However, now the naming doesn't make sense. Seeing a `linkedShipment` when you're actually working with a linked return is confusing. We should update this code to use a more generic name (probably invoice?) so it's applicable to both cases.

Poor naming can be such a time sink, particularly if you are new to a codebase, but also generally when you're devving. Having to hold in your mind the cases where "shipment" is a shipment and where "shipment" is a return is a waste of energy, and can lead to more mistakes.

Doing some renames in the transfer processor code should be fairly low risk. 

We do expose a `linkedShipment` field on the InvoiceNode - to prevent breaking the API, could simply _also_ expose `linkedInvoice `. Perhaps we then deprecate `linkedShipment`.

(This is starting to raise questions for me about the number of things that are optional/mean different things on different invoices, and whether we should have shipments and returns queries, rather than a generic invoice query, but that's another topic 😅)

The underlying DB field is linked_invoice_id, so we don't need to worry about whether to change that - that's a win.

I would expect this to be pretty trivial, a couple hours tops, to test it's all working? Though documentation of this process should also be updated, diagrams etc. so probably a little bit more time there 😁

lache-melvin:
Fixes #3332

# 👩🏻‍💻 What does this PR do?

Updates transfer naming from `shipment` to `invoice`, now that it is used for both shipments and returns.

Note: broken stocktake integration test due to merging `item_name` issue... investigating this separately!

# 🧪 Testing

- [ ] Still compiles
- [ ] Transfer tests pass
- [ ] Integration tests pass

# 📃 Documentation

Internal docs updated


server/service/src/processors/transfer/invoice/delete_inbound_invoice.rs
```diff
@@ -0,0 +1,91 @@
+use repository::{
```
lache-melvin, line 1: argh sorry - looks like there were a couple of files that GH didn't realise are renamed... But same as others, no changes here other name shipment -> invoice 🥲

server/service/src/sync/test/integration/central/currency.rs
```diff
@@ -21,6 +21,7 @@ impl SyncRecordTester for CurrencyTester {
             code: "NZD".to_string(),
             is_home_currency: true,
             date_updated: None,
+            is_active: true,
```
lache-melvin, line 24: Some new fields were added, missed in integration tests 😢

server/service/src/sync/test/integration/errors.rs
```diff
@@ -95,13 +95,13 @@ mod tests {
 
         assert_matches!(
             error,
-            SyncError::RemotePushError(RemotePushError::SyncApiError(SyncApiError {
+            SyncError::SyncApiError(SyncApiError {
```
lache-melvin, line 98: Synchroniser now does a `get_site_info` call to get V6 central server info, so now get the unauthorised error at that step, rather than on remote push!

server/service/src/processors/mod.rs
```diff
@@ -34,7 +34,7 @@ pub struct Processors {
 #[derive(Debug, Error)]
 enum ProcessorsError {
     #[error("Error in shipment transfer processor ({0})")]
-    ShipmentTransfer(ProcessShipmentTransfersError),
+    ShipmentTransfer(ProcessInvoiceTransfersError),
```
roxy-dao, line 37: Should the error also be renamed? 🤔

server/service/src/processors/mod.rs
```diff
@@ -84,7 +84,7 @@ impl Processors {
                         process_requisition_transfers(&service_provider).map_err(ProcessorsError::RequisitionTransfer)
                     },
                     Some(_) = shipment_transfer.recv() => {
```
roxy-dao, line 86: And here?

server/service/src/processors/mod.rs
lache-melvin: Oo yes thank you!

roxy-dao (approved):
LGTM~

Issue #3624, opened 2024-04-16
Postgres test are flaky


clemens-msupply:
At least for me the Postgres test are failing randomly.

Know test to be failing occasionally:
- [ ] processors::transfer::requisition::test::requisition_transfer
- [ ] processors::transfer::shipment::test::invoice_transfers
- [ ] processors::transfer::shipment::test::invoice_transfers_with_merged_name
- [ ] temperature_excursion::test::temperature_excursion_tests
- [x] programs::encounter::encounter_updated::encounter_document_updated_test::test_encounter_deletion (https://github.com/msupply-foundation/open-msupply/pull/3852)
- [ ] location::tests::query::query::location_service_assigned_to_asset
- [ ] others?

jmbrunskill:
@clemens-msupply  is this fixed now with this PR? https://github.com/msupply-foundation/open-msupply/pull/3669

Can you please record which tests are failing intermittently if not.

Chris-Petty:
@clemens-msupply ditto with James, this is not enough detail for anyone to go off of!

clemens-msupply:
@Chris-Petty please run `cargo test --features postgres` if you don't find any issue I can try to reproduce.

@jmbrunskill yes `test_sync_pull_and_push` was definitely one of them. For the others I can't remember (wrote the issue after the fact to investigate later) But there are always some processor tests that are failing for a while now.

jmbrunskill:
This one failed for me on one test run, worked the next...

```
---- programs::encounter::encounter_updated::encounter_document_updated_test::test_encounter_deletion stdout ----
thread 'programs::encounter::encounter_updated::encounter_document_updated_test::test_encounter_deletion' panicked at service/src/programs/encounter/encounter_updated.rs:245:9:
assertion `left == right` failed
  left: 1
 right: 0
```

jmbrunskill:
This test is now failing intermittently even on SQLite
![telegram-cloud-photo-size-1-5136538955259428027-y](https://github.com/user-attachments/assets/53dff7a8-f801-42b9-878f-eff70126d36f)

PR #3852, opened 2024-05-09
Fix encounter updated test

clemens-msupply:
Part of #3624

# 👩🏻‍💻 What does this PR do?

This was an actual bug. When encounters are deleted the events are removed correctly now.

Our test machines are actually too fast and this test only failed if there was some slowdown, i.e. when test stretches from one millisecond to the next millisecond.
I added some hard-coded timestamps now to simulate a couple of test-cases.


roxy-dao (approved):
/o/ \o\ /o/ \o\

PR #3855, opened 2024-05-10
Add tooltip to the select/unselect all column header

### LINKED ISSUE #3853: Hovering over "select all columns" shows some graphical artifact
## What went wrong? 😲
 

image

I guess this is an empty tooltip. This can probably be fixed in the tooltip component?

Seen this for
- outbound shipments
- coldchain equipment
- probably others

## Expected behaviour 🤔

Don't show the empty tooltip

## How to Reproduce 🔨

1. go to table view
2. hover over checkbox

## Your environment 🌱

- Open mSupply Version: develop branch

clemens-msupply:
Fixes #3853 

# 👩🏻‍💻 What does this PR do?

Add a tooltip to avoid an empty tooltip to show up.

First I checked if we have data for a tool tip and if not set it to `null` (`packages/common/src/ui/layout/tables/components/Header/Header.tsx`)
However, thought having a tooltip is good solution as well...

# 🧪 Testing

- Go to a table and hover over the selection header checkbox
- Tooltip should show

### DIFF
--- a/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
@@ -5,6 +5,7 @@ import { InfoOutlineIcon, SortDescIcon } from '@common/icons';
 import { RecordWithId } from '@common/types';
 import { useDebounceCallback } from '@common/hooks';
 import { useTranslation } from '@common/intl';
+import { tooltipPlacement } from '../tooltipPlacement';
 
 export const HeaderRow: FC<PropsWithChildren<{ dense?: boolean }>> = ({
   dense,
@@ -117,13 +118,6 @@ export const HeaderCell = <T extends RecordWithId>({
     child
   );
 
-  const tooltipPlacement =
-    align === 'left'
-      ? 'bottom-start'
-      : align === 'right'
-      ? 'bottom-end'
-      : 'bottom';
-
   return (
     <TableCell
       role="columnheader"
@@ -147,7 +141,7 @@ export const HeaderCell = <T extends RecordWithId>({
     >
       <Tooltip
         title={tooltip}
-        placement={tooltipPlacement}
+        placement={tooltipPlacement(align)}
         style={{ whiteSpace: 'pre-line' }}
       >
         {HeaderLabel}
--- a/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
@@ -64,19 +64,20 @@ export const HeaderCell = <T extends RecordWithId>({
 
   const columnLabel =
     column.label === '' ? '' : t(column.label, column.labelProps);
-  const tooltip = (
-    <>
-      {!!description && <div>{t(description)}</div>}
-      {sortable ? (
-        <div>
-          {t('label.click-to-sort')}
-          {` ${columnLabel}`}
-        </div>
-      ) : (
-        columnLabel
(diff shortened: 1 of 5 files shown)


roxy-dao (commented):
Nice! Works as expected but now I can't select the first box if tooltip is showing =[

clemens-msupply:
> Nice! Works as expected but now I can't select the first box if tooltip is showing =[

You mean when hovering over the header and then going down to the first row? yes noticed this as well but this is the same for the other column. You have to move away to disable the tooltip and then select the box.

Do you think this is too annoying? The checkbox is quite self-explanatory so we might not need a tooltip.

roxy-dao:
> > Nice! Works as expected but now I can't select the first box if tooltip is showing =[
> 
> You mean when hovering over the header and then going down to the first row? yes noticed this as well but this is the same for the other column. You have to move away to disable the tooltip and then select the box.
> 
> Do you think this is too annoying? The checkbox is quite self-explanatory so we might not need a tooltip.

I think it's a bit annoying yah... probably don't need a tooltip since it's common UI on many websites? 🤔 But will this be obvious to everyone? Idk

mark-prins:
The problem is that the tooltip component is receiving a Component when there is nothing to show. the Tooltip doesn't render if there is nothing to show (which it determines with a blank / null / undefined `title` prop).

To fix the underlying issue, I would do this:

in `client/packages/common/src/ui/layout/tables/components/Header/Header.tsx`

```
  const hasTooltip = !!description || !!columnLabel;

...

<Tooltip
        title={hasTooltip ? tooltip : ''}
...
```


that way, the tooltip won't show for the checkbox column if no label is provided.

Side observation - with the current fix, I get this on the stock page:

Screenshot 2024-05-10 at 5 40 31 PM

the tooltip shows in a strange place, a long way from the checkbox!

I like having the tooltip on the select all/none checkbox - so my vote is to retain it 🙏 
but would like the Header updated..

mark-prins:
Thanks @clemens-msupply looks good to me. @roxy-dao are you ok with this? do we need to do more about non-selectability of the first checkbox?

roxy-dao (approved):
Nope! Happy as it is, there is a way to get around non-selectability of first box~

Thanks Clemens!!

client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -116,6 +117,13 @@ export const HeaderCell = <T extends RecordWithId>({
     child
   );
 
+  const tooltipPlacement =
+    align === 'left'
```
roxy-dao, line 121: Ooh.. I kinda don't like this nested tenary...

clemens-msupply: if `switch` would be an expression in JS (like in Rust)...

but you are too fast already changed it with some other refactors when fixing the table cells as well ;)

clemens-msupply:
> The problem is that the tooltip component is receiving a Component when there is nothing to show. the Tooltip doesn't render if there is nothing to show (which it determines with a blank / null / undefined `title` prop).

I actually had this already fixed and then decided to not commit it since it was not used, but you are right I added it...

> Side observation - with the current fix, I get this on the stock page:
> Screenshot 2024-05-10 at 5 40 31 PM
> 
> the tooltip shows in a strange place, a long way from the checkbox!

I couldn't reproduce it for the checkbox column, but I saw the problem for other column headers and for the table cells. I am now adjusting the tooltip placement according to the column alignment.

PR #3872, opened 2024-05-13
2945 user friendly sync login errors

### LINKED ISSUE #2945: Display user friendly errors on login failures

There have been a few complaints about the unfriendly messaging that can happen when you login and initialise. 
An example:

telegram-cloud-photo-size-1-5035418318256844280-x

Also, the request was made: _"Can we make this button show the error messages?"_

Second reported error:

telegram-cloud-photo-size-1-5069329911147244978-y

- [ ] For known errors, provide a helpful message describing the error and a suggested solution.
- [ ] On the login screen, find a way to display the error that doesn't require pressing a very small info icon
- [ ] Change the message 'Sync API version is not compatible'

The errors above might be:
- the server URL or port is incorrect
- the server is offline
- you have a network issue, such as an incorrectly configured firewall
- the port you have configured mSupply to run on is the same as the port used by open mSupply

There could also be various network issues which show up on that screen:
- your SIM card has run out of credit
- you are not connected to the WIFI
- mobile data is not turned on

We cannot show all possible resolution options, so best to just summarise: "Check the server URL" or something like that and "Check your network connection" if you get a network error.

The port issue is along the lines of 'Unexpected response from the server: check that this URL is for the central server" 🤷

lache-melvin:
Fixes #2945

# 👩🏻‍💻 What does this PR do?

Bunch of minor improvements to the error details on login/initialisation:

- The whole component (the warning icon and all the text) is now the tooltip area - so long press anywhere on the error message on tablet to see more
  - Other option here would be to add an onClick popup (like we have for the print report selector?) so that it showed up on tap rather than long press for tablet... this felt like the quicker win though, let me know what you think?

- Adds a "hint" to the error details tooltip - so we can still show the error details for debugging..

- Add hint to check network connection if network error:
Screenshot 2024-05-13 at 5 08 03 PM

- Add hint to check server URL if v6 sync misconfigured:
Screenshot 2024-05-13 at 5 10 22 PM

- If your user doesn't have any access to stores on this site (it used to log you in, then an auth error modal popped up saying there was "no user data"):
Screenshot 2024-05-13 at 4 54 27 PM

- Generic error for unknown errors (used to show `Invalid username or password`, which was often incorrect):
Screenshot 2024-05-13 at 4 54 43 PM

- Add hint if sync version is incompatible - it was asked to just change the message, but I think the one-liners are nice for the main display message, and the longer info in the hint? Thoughts?
Screenshot 2024-05-13 at 5 21 54 PM

# 🧪 Testing

- [ ] Stop your server, and try to log in
  - [ ] You should see the `Unable to connect to server` error, and the hint on hover
- [ ] In mSupply (Special > Synchronisation > site setting for your oms central server) clear the OMS central server URL/set it to some fun garbage
  - [ ] You should see the `v6 central server not configured` error, and the hint on hover
- [ ] In mSupply (Admin > Edit users > Your user > Login rights) Remove `can login` rights for your user for any of the stores on your current site
  - [ ] You should see the `Unable to log in` error, and the hint on hover
- [ ] Make sync versions incompatible (either switch to older mSupply version, or just increase the number [here](https://github.com/msupply-foundation/open-msupply/blob/af04575a272cfbb44ed4cb9ba8b91e486e033f4f/server/service/src/sync/settings.rs#L4) and restart the server
  - [ ] You should see the `Sync API version is not compatible` error, and the extra detail in the hint on hover
- [ ] Honestly not sure how to trigger the unknown error, I just was changing code to return something random?
  - [ ] You should see the `The server has returned an error` message
- [ ] On any error that has a tooltip (i.e. if the info icon is there) - the tooltip pops up on hover over anywhere on the error message
  - [ ] Test this on a tablet! Is it easier to press?

# 📃 Documentation

  1. Screenshots? Not sure how in depth the errors have been covered off in the docs...


client/packages/system/src/Sync/api/api.ts
```diff
@@ -70,11 +70,23 @@ export function mapSyncError(
     [SyncErrorVariant.Unknown]: defaultKey || 'error.unknown-sync-error',
   };
 
+  const getHint = () => {
```
roxy-dao, line 73: ❤️

client/packages/common/src/authentication/api/api.ts
```diff
@@ -79,10 +79,15 @@ export const getAuthQueries = (sdk: Sdk, t: TypedTFunction<LocaleKey>) => ({
         if ('message' in error) {
           console.error(error.message);
         }
+
+        const errorMessage = error.message.includes('Network request failed')
```
andreievg, line 83: I think this could be typed, through GraphqlStdError. Not suggesting we should do it thought, just food for thought

server/service/src/login.rs
```diff
@@ -154,6 +155,14 @@ impl LoginService {
                 });
             }
         };
+
+        // Check that the logged in user has access to at least one store on the site
+        match user_service.find_user(&user_account.id) {
```
andreievg, line 160: i wish this was named better (find_user_active_on_this_site)

client/packages/common/src/ui/components/errors/ErrorWithDetails.tsx
```diff
@@ -3,32 +3,53 @@ import {
...
+          <Typography sx={{ fontWeight: 'bold', ...sx }}>{hint}</Typography>
+        )}
+
+        {!!details && <Typography sx={sx}>{details}</Typography>}
+      </Box>
+    );
+  };
+
+  return (
+    <Tooltip title={!details && !hint ? null : <TooltipContents />}>
```
andreievg, line 36: One check for !details && !hint and !!(details || hint) seem quite close, maybe local variable ?

client/packages/host/src/components/Login/Login.tsx
```diff
@@ -55,22 +55,44 @@ export const Login = () => {
...
 
       const formattedTime = customDate(
         new Date(0, 0, 0, 0, 0, 0, timeoutRemaining),
         'm:ss'
       );
-      return `${t('error.account-blocked')} ${formattedTime}`;
+      return { error: `${t('error.account-blocked')} ${formattedTime}` };
+    }
+
+    if (error.message === 'InvalidCredentials') {
```
andreievg, line 78: I was trying to suggest that we can make this logic a bit more typed by not saving __typename to string [here](https://github.com/msupply-foundation/open-msupply/blob/4b448ffce5965e1ef808fb98b4f329a80794a046/client/packages/common/src/authentication/api/api.ts#L33-L44), but then got lost in tracing through context, helpers, and useEffects : (

andreievg: Also its' a bit wiered how we add timeoutRemaining to the AuthenticationError shape, with undefined, when we know it's present with __typename === 'AccountBlocked'

andreievg (approved):
Looks great thanks !

I think we can advertise the fact that the whole error is pressable (as tooltip), after this change goes in.

No real change requests, just some comments, thanks

mark-prins:
@lache-melvin any chance of sneaking in #3881 as well?

jmbrunskill:
Love this! Great work.

server/service/src/login.rs
lache-melvin: Hm yeah, I thought that too - I think I will rename, its a bit confusing 🙏

client/packages/host/src/components/Login/Login.tsx
lache-melvin: Yeah I don't love the strings, and I found the different layers of mapping confusing... didn't want to touch too much and bloat this PR but would be nice to come back to in another refactor issue?

andreievg: Sure, i have some thoughts about doing cookie setting on back end

roxy-dao:
Don't really cover in-dept error messages in a lot of places 🤔 so putting this off for now (unless requested otherwise)

PR #3876, opened 2024-05-13
3829 inbound delivery date updated when verified

### LINKED ISSUE #3829: Inbound Shipment delivery date is updated when verified
## What went wrong? 😲

Looks like delivery date is updated when status is changed to verified (even if inbound is already set to delivered).
This can also cause ledger discrepancy.

## Expected behaviour 🤔

Delivery date is only set when status is changed to delivered (or when delivered status is skipped like when setting straight to verified)(

## How to Reproduce 🔨

This is a date time field, and most of UI shows date, so hard to spot, fastest way to spot is through ledger tab on stock line

Create inbound, add batch, set to delivered. 
See the date time in ledger view for that stock line
Wait a minute
Set inbound as verified
See the date time in ledger view for stock line change

This is also visible as ledger discrepancy in mSupply, writing test case here for future reference for testers

Make sure stock is 0 for the item you are doing this test with
Create inbound, add batch, set to delivered. 
Create outbound, add that batch and issue some (not all) of it.
Set outbound to shipped
Go back to inbound, set to verified.
In mSupply central, log into this store, find the item, go to ledger, then print, you should see negative stock levels as per below screenshot (this is why item should be 0 at the start, to make sure other batches don't interfere with the calculation):

Screenshot 2024-05-07 at 9 51 43 PM

## Your environment 🌱

- Open mSupply Version: 2.1

lache-melvin:
Fixes #3829

# 👩🏻‍💻 What does this PR do?

Issue was that the delivered_datetime was being overwritten on Delivered -> Verified.

- Uses `match` statement rather than status index to determine whether to update delivered/verified datetime.
  - This makes the status change logic more readable, and prevents the delivered_datetime being updated when setting to `Verified`

I've seen both in different places... should `delivered_datetime` be set if we go straight from `New` to `Verified`?

If not, we need to update the ledger query to handle this possibility! Currently we get no ledger line if we go straight to verified 👀 

# 🧪 Testing

- [ ] Create an inbound shipment, add a line, and set it to delivered
- [ ] In view stock, check the timestamp in the ledger tab of your newly introduced stock line
- [ ] Wait at least a minute 😅 
- [ ] Set your inbound shipment to Verified
- [ ] Check the ledger tab again - the timestamp should NOT have changed

Or can check by going to the inbound shipment detail view, look in dev tools at the `inboundByNumber` query, check the timestamps on the inbound shipment: 

Screenshot 2024-05-14 at 10 26 22 AM

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/invoice/inbound_shipment/update/mod.rs
+++ b/server/service/src/invoice/inbound_shipment/update/mod.rs
@@ -865,7 +865,10 @@ mod test {
             .unwrap();
 
         // Ensure delivered time not updated by status change to verified
-        assert_eq!(invoice.delivered_datetime, None);
+        assert_eq!(
+            invoice.delivered_datetime,
+            mock_inbound_shipment_a().delivered_datetime
+        );
 
         assert!(invoice.verified_datetime.unwrap() > now);
         assert!(invoice.verified_datetime.unwrap() < end_time);
--- a/server/service/src/invoice/inbound_shipment/update/mod.rs
+++ b/server/service/src/invoice/inbound_shipment/update/mod.rs
@@ -864,6 +864,9 @@ mod test {
             .find_one_by_id(&stock_line_id)
             .unwrap();
 
+        // Ensure delivered time not updated by status change to verified
+        assert_eq!(invoice.delivered_datetime, None);
+
         assert!(invoice.verified_datetime.unwrap() > now);
         assert!(invoice.verified_datetime.unwrap() < end_time);
         assert_eq!(
--- a/server/repository/src/mock/invoice.rs
+++ b/server/repository/src/mock/invoice.rs
@@ -246,6 +246,9 @@ pub fn mock_inbound_shipment_a() -> InvoiceRow {
             .unwrap()
             .and_hms_milli_opt(20, 30, 0, 0)
             .unwrap();
+        r.delivered_datetime = NaiveDate::from_ymd_opt(1970, 1, 3)
+            .unwrap()
+            .and_hms_milli_opt(21, 30, 0, 0);
     })
 }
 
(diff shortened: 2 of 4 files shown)


andreievg:
> I've seen both in different places... should delivered_datetime be set if we go straight from New to Verified?

Yep all statuses 'in chain' should be updated if they are 'skipped'

andreievg (changes requested):
Looks good, it's just the delivered should always be updated, in outbound we have this:  

https://github.com/msupply-foundation/open-msupply/blob/3900445c71103c330e8ad44c7de2ee99dc45a5dd/server/service/src/invoice/outbound_shipment/update/generate.rs#L173-L221

server/service/src/invoice/inbound_shipment/update/mod.rs
```diff
@@ -864,6 +864,9 @@ mod test {
             .find_one_by_id(&stock_line_id)
             .unwrap();
 
+        // Ensure delivered time not updated by status change to verified
+        assert_eq!(invoice.delivered_datetime, None);
```
andreievg, line 868: would this test fail now ?

lache-melvin: Oh good spotting - no, because the dummy data doesn't have the delivered_datetime set as it should 😅

lache-melvin: Fixed: [6162be9](https://github.com/msupply-foundation/open-msupply/pull/3876/commits/6162be968f7e9598dbf85d1777309697b3d1c840)

andreievg (approved):
Thanks, looks great

PR #3884, opened 2024-05-14
3834 clear number inputs to 0

### LINKED ISSUE #3834: Outbound shipments: Cannot erase the number of packs to be issued
## What went wrong? 😲

After inputting a number of packs to be issued, we cannot erase the number or change it to 0.

https://github.com/msupply-foundation/open-msupply/assets/132425096/15d94c54-32dc-4070-b3d7-ee05791c12a5

It happens ons Desktop app and Android

## Expected behaviour 🤔

Number should be able to be changed to 0.

## How to Reproduce 🔨

1. Go to 'Outbound shipment'
2. Click on 'new shipment'
3. Click on 'add item' and choose an item
4. Change number of packs to be issued to any number >0
5. Try to erase this number or change it to 0.
6. See error

## Your environment 🌱

- Open mSupply Version: V2.0RC3

- Platform: Desktop app.

lache-melvin:
Fixes #3834

# 👩🏻‍💻 What does this PR do?

Updates:
- Internal order requested quantity to allow for 0 quantity input
- Updates pack size input (PackVariantEntryCell) for Inbound Shipment and Inbound Return, to allow clearing the value while typing, and then reset to the min value (e.g. 1) on blur
- Allows pack quantity field to be set to 0 in inbound shipments (you can't save the modal, but at least you can clear the input in order to change e.g. 1 to 2!)

# 🧪 Testing

- [ ] Add an item
- [ ] Enter a requested quantity
- [ ] TEST: you should be able to backspace the quantity to 0
- [ ] Create an inbound shipment
- [ ] Add an item
- [ ] Enter a pack quantity
- [ ] TEST: you should be able to backspace the quantity to 0
- [ ] Backspace the pack size
- [ ] TEST: it should clear
- [ ] Click off the pack size field, so it loses focus
- [ ] TEST: it should reset to 1
- [ ] Create an inbound return
- [ ] Add an item
- [ ] Backspace the pack size
- [ ] TEST: it should clear
- [ ] Click off the pack size field, so it loses focus
- [ ] TEST: it should reset to 1

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -168,9 +168,13 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           onChange(newNum);
         }}
         onBlur={() => {
+          const defaultVal = min === 0 ? undefined : min;
+
           const parsed = parse(textValue ?? '');
-          onChange(Number.isNaN(parsed) ? undefined : parsed);
-          setTextValue(formatValue(value));
+          const val = Number.isNaN(parsed) ? defaultVal : parsed;
+
+          onChange(val);
+          setTextValue(formatValue(val));
         }}
         onFocus={e => e.target.select()}
         {...props}
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -168,10 +168,8 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           onChange(newNum);
         }}
         onBlur={() => {
-          const defaultVal = min === 0 ? undefined : min;
-
           const parsed = parse(textValue ?? '');
-          const val = Number.isNaN(parsed) ? defaultVal : parsed;
+          const val = Number.isNaN(parsed) ? defaultValue : parsed;
 
           onChange(val);
           setTextValue(formatValue(val));
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
@@ -267,10 +267,9 @@ export const RequestLineEditForm = ({
                     )}
                     width={100}
                     onChange={quantity => {
-                      if (quantity === undefined) return;
                       update({
                         requestedQuantity:
-                          quantity * currentItem.defaultPackSize,
+                          (quantity ?? 0) * currentItem.defaultPackSize,
                       });
                     }}
                   />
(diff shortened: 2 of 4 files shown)


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -168,9 +168,13 @@ export const NumericTextInput: FC<NumericTextInputProps> = React.forwardRef(
           onChange(newNum);
         }}
         onBlur={() => {
+          const defaultVal = min === 0 ? undefined : min;
```
lache-melvin, line 171: If a min value is defined, and the input is undefined, we should set the value to that min on blur

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
```diff
@@ -267,10 +267,9 @@ export const RequestLineEditForm = ({
                     )}
                     width={100}
                     onChange={quantity => {
-                      if (quantity === undefined) return;
                       update({
                         requestedQuantity:
-                          quantity * currentItem.defaultPackSize,
+                          (quantity ?? 0) * currentItem.defaultPackSize,
```
lache-melvin, line 272: Allow clearing requested quantity back to 0 for internal order

client/packages/system/src/Item/Components/PackVariant/PackVariantEntryCell.tsx
```diff
@@ -65,10 +65,16 @@ export const PackVariantEntryCell =
...
           onChange={newValue => {
-            setPackSize(newValue || 1);
-            updater({ ...rowData, [column.key]: newValue });
+            // newValue could be undefined, while the user is inputting
+            // (e.g. they clear the field to enter a new pack size)
+            // In that case, we temporarily update the column value to 0 (so we don't have any `NaN`s)
+
+            // NumericTextInput onBlur will reset the value to our min (1) if the field is empty!
+            setPackSize(newValue);
+            updater({ ...rowData, [column.key]: newValue || 0 });
```
lache-melvin, line 77: Add handling of undefined pack size while someone is entering a new pack size. Should be reset to at least 1 on blur

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
CarlosNZ: Hmm, I don't think this is the behaviour we want. If there's a minimum, you should still be able to remove your input.

Also -- this causes a problem with negative values. The default "allowNegative" sets a minimum of `-Infinity` (or largest integer), so if the `allowNegative` prop is set, then when you remove your input and leave the input, it jumps to -999,999,999! 

(We should probably add an additional test for this case)

CarlosNZ: If we need to enforce a compulsory minimum, this should be the responsibility of the calling component, IMHO.

CarlosNZ (changes requested):
I think that default min change in NumericTextInput breaks too many other things (see comments). Can we enforce compulsory mins in the calling components when necessary? 

Have a chat IRL if it's getting too messy.

lache-melvin:
Thanks Carl, as discussed in person, updated to use defaultValue 🙏

CarlosNZ (approved):
Great, working nicely now, thanks for that.

I'll get on to the explanatory comment for Numeric input

PR #3891, opened 2024-05-14
Avoid float precision problems when multiplying pack sizes

### LINKED ISSUE #3705: Decimal Pack Sizes

As decided in KDD #3619, we require supporting decimal pack sizes in all contexts to allow compatibility with Legacy mSupply migrated data. Also it is arguably a feature.

- `float` migrated in sql for all pack size fields
- Views may need redefining
- `f64` propagated through backend and gql
- Possible tweaks to frontend code.

## Sub Issues

- [x] stock_line
- [x] invoice_line (may need to do at same time as stock_line)
- [x] #3939
- [x] stock_take_line

### Moneyworks Jobcode 🧰

**OMS:MIGRT**

clemens-msupply:
Fixes #3705

# 👩🏻‍💻 What does this PR do?

Avoids some visual problems with pack_size being a float. For example, 110.4 * 29 = 3201.6000000000004 is displayed as 3201.6 now.

I didn't make the pack_size editable as a float though. Or do you think we should do it now?

# 🧪 Testing

Please have a look for other `pack_size: float` related issues in the UI that I might have missed.

To create a decimal pack size, create a stocktake in OG mSupply an edit an item to have decimal pack size. This stock line can then be sent to the omSupply site.

### DIFF
--- a/client/packages/common/src/utils/numbers/NumUtils.ts
+++ b/client/packages/common/src/utils/numbers/NumUtils.ts
@@ -19,10 +19,24 @@ export const NumUtils = {
 
     return constrain(parsed, min, max);
   },
+  /**
+   * Round a value to a given precision.
+   *
+   * For example, round(10.232, 2) gives 10.23
+   */
   round: (value: number, dp = 0): number => {
     if (dp === Infinity) return value;
     const multiplier = 10 ** dp;
     return Math.round(value * multiplier) / multiplier;
+  },
+  /**
+   * Multiplies two float numbers avoiding "float artefacts".
+   * For example: 110.4 * 29 = 3201.6000000000004
+   * while floatMultiply(110.4, 29) = 3201.6
+   */
+  floatMultiply: (left: number, right: number): number => {
+    // Use a hacky(?) correction factor of 10:
+    return (10 * left * right) / 10;
   },
   /**
    * This constant should be used for values that are potentially send to a backend API that expects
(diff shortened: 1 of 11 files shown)


client/packages/common/src/utils/numbers/NumUtils.ts
```diff
@@ -19,11 +19,25 @@ export const NumUtils = {
...
     return Math.round(value * multiplier) / multiplier;
   },
+  /**
+   * Multiplies two float numbers avoiding "float artefacts".
+   * For example: 110.4 * 29 = 3201.6000000000004
+   * while floatMultiply(110.4, 29) = 3201.6
+   */
+  floatMultiply: (left: number, right: number): number => {
+    // Use a hacky(?) correction factor of 10:
+    return (10 * left * right) / 10;
```
jmbrunskill, line 39: `(10 * 1.001 * 1000) / 10;`
`1000.9999999999998`

I think we should maybe create some test cases here, and define the parameters/ranges of data we expect this to work with?
E.g. Do we know our users never use more than 2 decimal places for example?


`(100 * 1.001 * 1000) / 100;`
`1001`

Potentially this is better?

I think there's no 100% reliable way to deal with this issue.

jmbrunskill (approved):
Would like to see some test cases, so we can easily improve the code and still check compatibility.

Maybe should do `(100 * left * right / 100)` to be a little safer (I think)

client/packages/common/src/utils/numbers/NumUtils.ts
clemens-msupply: thanks good idea, 100 is a better factor! I added some small test as well

PR #3904, opened 2024-05-15
Lock changelog table when retrieving changelogs (postgres)

### LINKED ISSUE #3624: Postgres test are flaky

clemens-msupply:
Fixes #3624

# 👩🏻‍💻 What does this PR do?

The failing postgres tests were actually caused by a real bug! This bug could result in completely ignoring entries from the change log when the cursor was advanced too early...

To investigate the failing transfer tests, I printed out which cursors have been handled by the processor and the latest cursor of the failing transfer requisition row was not handled.
Also note that the test where only failing in Postgres. Postgres default isolation level is Read Committed (Serializable in Sqlite).
The failing tests are running independent test cases in parallel. This is what could have happened:
1) Process 1 adds to the changelog (cursor 1)
2) Process 1 finishes tx (cursor 1 becomes visible)
3) Process 2 adds to the changelog (cursor 2) but takes longer than to finish the tx
5) Process 3 adds to the changelog (cursor 3)
6) Process 3 finishes tx
7) Processor reads changlog batch and gets cursors [1, 3] (Process 2 is not done yet with its tx)
8) Processor updates the latest cursor marker to 3
9) Process 2 finishes with cursor 2, i.e. this record will never be handled because the latest cursor marker has already been updated to 3!

In this solution in this PR is to lock the changlog (and indirectly the changlog_dedup view)  TABLE using an exclusive lock. This ensures that nobody reads while somebody is writing and vice versa. Thus, a changelog cursor can't be "reserved" by a tx before it can be read, e.g. by the processor...

# 🧪 Testing
Please have a good look at other changelog methods. I think the count* methods are fine, but not 100% if some code depends on it that, for example, the current count at a cursor has to stay stable... So a critical look at these methods and how they are used would be good!

- [ ] Test if postgres/sqlite test are still failing

### DIFF
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -586,3 +586,62 @@ impl RowActionType {
         inline_init(|r: &mut EqualFilter<Self>| r.equal_to = Some(self.clone()))
     }
 }
+
+#[cfg(test)]
+mod test {
+    use tokio::sync::oneshot;
+    use util::inline_init;
+
+    use crate::{
+        mock::MockDataInserts, test_db::setup_all, ChangelogRepository, ClinicianRow,
+        ClinicianRowRepository, RepositoryError, TransactionError,
+    };
+
+    /// Example from with_locked_changelog_table() comment
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_late_changelog_rows() {
+        let (_, connection, connection_manager, _) =
+            setup_all("test_late_changelog_rows", MockDataInserts::none()).await;
+
+        ClinicianRowRepository::new(&connection)
+            .upsert_one(&inline_init(|r: &mut ClinicianRow| {
+                r.id = String::from("1");
+                r.is_active = true;
+            }))
+            .unwrap();
+
+        let (sender, receiver) = oneshot::channel::<()>();
+        let manager_2 = connection_manager.clone();
+        let process_2 = tokio::spawn(async move {
+            let connection = manager_2.connection().unwrap();
+            let result: Result<(), TransactionError<RepositoryError>> = connection
+                .transaction_sync(|con| {
+                    ClinicianRowRepository::new(con)
+                        .upsert_one(&inline_init(|r: &mut ClinicianRow| {
+                            r.id = String::from("2");
+                            r.is_active = true;
+                        }))
+                        .unwrap();
+                    sender.send(()).unwrap();
+                    std::thread::sleep(core::time::Duration::from_millis(100));
+                    Ok(())
+                });
+            result
+        });
+        receiver.await.unwrap();
+        ClinicianRowRepository::new(&connection)
+            .upsert_one(&inline_init(|r: &mut ClinicianRow| {
+                r.id = String::from("3");
+                r.is_active = true;
+            }))
+            .unwrap();
+
+        let changelogs = ChangelogRepository::new(&connection)
+            .changelogs(0, 10, None)
+            .unwrap();
+        assert_eq!(changelogs.len(), 3);
+
+        // being good and awaiting the task to finish orderly and check it did run fine
+        process_2.await.unwrap().unwrap();
+    }
+}
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -471,12 +471,24 @@ fn create_filtered_outgoing_sync_query(
     query
 }
 
-/// Postgres has Read Committed isolation level, we need to prevent other tx to add rows while
-/// reading the changelogs.
-/// For example, if changelog contains [1, 3, 4, 5] while another tx is about to add a changelog
-/// row with cursor = 2.
-/// We need to wait for this changelog to be added before doing the changelogs() query, otherwise
-/// the changelog cursor will be set to 5 and the changelog 2 will be missed.
+/// Runs some DB operation with a fully locked `changelog` table.
+/// This only applies for for Postgres and does nothing for Sqlite.
+///
+/// Motivation:
+/// When querying changelog entries, ongoing transactions might continue adding changelog entries
+/// to the queried range of changelogs.
+/// This is because Postgres has Read Committed isolation level (instead of Serialized in Sqlite).
+/// However, we assume that there will be no new changelog entries in the queried range in the
+/// future, e.g. when updating the cursor position.
+///
+/// For example, a changelog may contain [1, 3, 4, 5] while another (slow) tx is about to commit a
+/// changelog row with cursor = 2.
+/// We need to wait for this changelog 2 to be added before doing the changelogs() query, otherwise
+/// we might update the latest changelog cursor to 5 and the changelog with cursor = 2 will be left
+/// unhandled when continuing from the latest cursor position.
+///
+/// Locking the changelog table will wait for ongoing writers and will prevent new writers while
+/// reading the changelog.
 fn with_locked_changelog_table<T, F>(
     connection: &StorageConnection,
     f: F,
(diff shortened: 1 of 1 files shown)


server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -464,6 +471,52 @@ fn create_filtered_outgoing_sync_query(
...
+/// However, we assume that there will be no new changelog entries in the queried range in the
+/// future, e.g. when updating the cursor position.
+///
+/// For example, a changelog may contain [1, 3, 4, 5] while another (slow) tx is about to commit a
+/// changelog row with cursor = 2.
+/// We need to wait for this changelog 2 to be added before doing the changelogs() query, otherwise
+/// we might update the latest changelog cursor to 5 and the changelog with cursor = 2 will be left
+/// unhandled when continuing from the latest cursor position.
+///
+/// Locking the changelog table will wait for ongoing writers and will prevent new writers while
```
andreievg, line 490: While there is a lock on a table, what happens to existing transactions that try to write to the table (what does `will prevent` mean) ?

jmbrunskill (approved):
I think we probably need to include the lock on `outgoing_sync_records_from_central` too right? Otherwise it's possible that we could miss syncing records that were in a transaction while the sync process runs?

I can't help but feel like there must be a better method than exclusive lock for every read but haven't thought of one yet.
It would be nicer if we had an exclusive lock on writes for example, but don't think that's feasible.
I guess processors could gain exclusive lock, but that still leaves the possibility of other things like sync that could miss records... 

Would it be possible to create a test case which forces this ordering and verifies that the solution works? Something like this one?

https://github.com/msupply-foundation/open-msupply/blob/6211db05a177440fff281575582e9caeb5f7bd26/server/repository/src/tests.rs#L1084

 I'm interested, as I think we might want to try some alternative solutions in future and would be good to have a test case to be confident other approaches work...

server/repository/src/db_diesel/changelog/changelog.rs
jmbrunskill: I think the database will wait to try to acquire the lock (until a wait timeout). It's not only writing to the table, any reads will be blocked too.

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -464,6 +471,52 @@ fn create_filtered_outgoing_sync_query(
...
+    F: FnOnce(&mut LockedConnection) -> Result<T, RepositoryError>,
+{
+    if cfg!(feature = "postgres") {
+        use diesel::connection::SimpleConnection;
+        let result = connection.transaction_sync_etc(
+            |con| {
+                let mut locked_con = con.lock();
+                locked_con
+                    .connection()
+                    .batch_execute("LOCK TABLE ONLY changelog IN ACCESS EXCLUSIVE MODE")?;
```
andreievg, line 506: A little bit unsure about impact on performance, especially once we get central omSupply server used by lots of sites, while still wanting 'atomic' updates (invoice and invoice_lines to appear in changelog in the same transaction). 

I think ideally would be great if database could set the serials at the end of transaction.

We are actually planning to relocate all trigger insertions of changelog to a manual insert_changelog, maybe that will give use more flexibility, and we would be able to do some optimisations.

server/repository/src/db_diesel/changelog/changelog.rs
clemens-msupply: As I understood other transactions will wait till the lock is released.
And the read transaction has to wait for other transactions before it can acquire the lock.

Btw. I run the transfer tests 50 times in a row without error. (However might not mean much if there is an even subtler bug, just an indication that I reasonably tested it...)

jmbrunskill: Existing transactions can keep writing their data.

andreievg (commented):
Is it possible to do a test specific to this scenario ? (should be able to with some delays and threads)

andreievg (changes requested):
I saw @jmbrunskill approve, but also request a test, sorry changing my review from comment to request changes, for the test please

server/repository/src/db_diesel/changelog/changelog.rs
andreievg: @jmbrunskill 

> Existing transactions can keep writing their data.

What do you mean ?

server/repository/src/db_diesel/changelog/changelog.rs
clemens-msupply: yes if we manage the cursor on our own we maybe could do better? However, we probably still have to lock our cursor counter for the whole tx which might not help much...

At least the table lock is currently held quite briefly because we only load a handful of changelogs at a time.

But yes some benchmark would be good. Maybe in combination with the test that James suggested then we can test it against a new method.

clemens-msupply:
> I think we probably need to include the lock on `outgoing_sync_records_from_central` too right? Otherwise it's possible that we could miss syncing records that were in a transaction while the sync process runs?

Already have the lock around outgoing_sync_records_from_central :)

server/repository/src/db_diesel/changelog/changelog.rs
jmbrunskill: I just tested it and it appears to work like this...

Let's imagine we have a transaction which need to write `invoice` and `invoice_line` to `changelog` in progress.
The transaction writes it's `invoice` changelog record, then a change log read request asks for exclusive mode (Transaction 2)

Transaction 2 will wait until transaction 1 is finished. It is able to insert it's remaining changelog records as it already holds a write lock on the table, and transaction 2 can't claim it's lock until all the other locks are cleared.

Transaction 3 which wants to start writing to change log, wouldn't be able to start until transaction 2 is finished.

jmbrunskill:
> > I think we probably need to include the lock on `outgoing_sync_records_from_central` too right? Otherwise it's possible that we could miss syncing records that were in a transaction while the sync process runs?
> 
> Already have the lock around outgoing_sync_records_from_central :)

Yes sorry saw that after I made the comment!

clemens-msupply:
I added a test case. To see it failing please run with Postgres feature and make `with_locked_changelog_table()` not lock the table...

server/repository/src/db_diesel/changelog/changelog.rs
clemens-msupply: But the processor would only block sync briefly. And blocking the processor during sync is ok, isn't it?

andreievg: Sync (integration) would also block all other UI operations, which would typically be fine, but could be problematic, like if we do sync priority, which allows for UI/UX to be opened without fully initialised (for example temperature data to sync after UI/UX is open for operations), when temperature logs are integrating, UI/UX would be blocked. I think this change should be ok for now, I do think we can do some optimisations in changelogs are inserted via repo. 

I didn't want to get too deep into this, but couldn't resist:

* Change logs are inserted straight away if not within transaction
* If within transaction insert them but also keep record of which ids (serials) have been inserted for which transaction `txid_current()`. When transaction finishes, increment id of those changelogs.

andreievg (approved):
Looks good, thanks for the test

server/repository/src/db_diesel/changelog/changelog.rs
clemens-msupply: Don't think I 100% understand what you mean with your last two points, probably better to have an offline discussion.
Think doing it manually would be quite hard (e.g. handle cases where the server may crash). IMHO the problem can better be solved tackling the long-lasting integration transaction, e.g. making it save to integrate records in batches. (just about doing a PR for integrating in batches during initialization which also brings good speedups for pg and sqlite...)

PR #3920, opened 2024-05-16
3914 prevent resyncing line records

### LINKED ISSUE #3914: Accidentally resyncing all requisition and stock take lines in v2.0.0 migration
## What went wrong? 😲

I haven't personally validated this, but looking at code...

The following two migrations add item_name to all stock take lines and requisition lines. This may make 10,000s of sync out records for modestly used OMS sites. We should avoid resyncing all these records (Legacy Central does the same migration, the records don't need to sync the update from remotes).

https://github.com/msupply-foundation/open-msupply/blob/main/server/repository/src/migrations/v2_00_00/requisition_line_add_item_name.rs

https://github.com/msupply-foundation/open-msupply/blob/main/server/repository/src/migrations/v2_00_00/stocktake_line_add_item_name.rs

## Expected behaviour 🤔

We don't resync all the line records if we don't need to.

We have a nice method used in 1.7.0 migration that can be used: `run_without_change_log_updates`

## How to Reproduce 🔨

1. Have a site with many many requisition lines or stocktake lines on <2.0.0
2. migrate to 2.0.0
3. All those lines are now bumped to be synced whoopsies

## Your environment 🌱
Hypothetically
- Open mSupply Version: v2.0.0
- Legacy mSupply Central Server Version: 7.14.4

- Platform: All

- Database type: Both

Chris-Petty:
Fixes #3914

# 👩🏻‍💻 What does this PR do?

- Makes sure the code that updates `requisition_line.item_name` and `stock_take_line.item_name` does not resync all lines accidentally.

I added tests for temperature logs too, it appears in 1.7 -> 2.0 they are not touched/resynced.

# 🧪 Testing

- [ ] Have a remote site on 1.7 50 stock take lines and 50 requisition lines (or any known amount)
- [ ] Is fully synced
- [ ] Stop connection with central server (turn off wifi, change site password on central)
- [ ] Upgrade to 2.0
- [ ] There should NOT be 100 records to sync (or whatever the sum of lines you have is)

# 📃 Documentation

### DIFF
--- a/server/repository/src/migrations/v2_00_00/mod.rs
+++ b/server/repository/src/migrations/v2_00_00/mod.rs
@@ -1,4 +1,4 @@
-use super::{version::Version, Migration};
+use super::{helpers::run_without_change_log_updates, version::Version, Migration};
 
 use crate::StorageConnection;
 
@@ -43,11 +43,15 @@ impl Migration for V2_00_00 {
         user_change_last_synced_to_optional::migrate(connection)?;
         inventory_adjustment_logtype::migrate(connection)?;
         report_views::migrate(connection)?;
-        requisition_line_add_item_name::migrate(connection)?;
+        run_without_change_log_updates(connection, || {
+            requisition_line_add_item_name::migrate(connection)
+        })?;
         stock_on_hand_add_item_name::migrate(connection)?;
         currency_add_is_active::migrate(connection)?;
         invoice_rename_tax::migrate(connection)?;
-        stocktake_line_add_item_name::migrate(connection)?;
+        run_without_change_log_updates(connection, || {
+            stocktake_line_add_item_name::migrate(connection)
+        })?;
         name_created_datetime::migrate(connection)?;
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks good to me Chris. Nice to have that test for the cursor.
I assume if you didn't use `run_without_change_log_updates` the test would fail right?

server/repository/src/migrations/v2_00_00/mod.rs
```diff
@@ -43,11 +43,15 @@ impl Migration for V2_00_00 {
         user_change_last_synced_to_optional::migrate(connection)?;
         inventory_adjustment_logtype::migrate(connection)?;
         report_views::migrate(connection)?;
-        requisition_line_add_item_name::migrate(connection)?;
+        run_without_change_log_updates(connection, || {
```
roxy-dao, line 46: Can't have one run_without_change_log and put both requisition_line and stocktake in here?

Chris-Petty: I was 50/50 on it, I considered it easier to wrap each one than look into if there was any execution ordering issues if I moved them together. It does slightly hurt me they're not together but no one will look at this code in a few months surely 😁.

Chris-Petty:
> Looks good to me Chris. Nice to have that test for the cursor. I assume if you didn't use `run_without_change_log_updates` the test would fail right?

Yup. I do backwards TTD, fix the problem, write the test, revert my fix, fail the test, revert reverting fix, pass the test lol.

Chris-Petty:
@mark-prins @andreievg what do you think about the merging of this? Should we keep clear by going with 2.0.1 (and perhaps take down artifacts for 2.0.0 to avoid use), or rewrite history and change 2.0.0?

mark-prins:
The binaries for v2.0.00 aren't on the site yet, so there's nothing to take down.
For clarity, I'd go with v2.0.01 and we make a new release

PR #3924, opened 2024-05-16
3717 om supply central backup kdd

### LINKED ISSUE #3717: omSupply central server backup KDD
KDD for different options of omSupply central server back up. Must clearly identify the problem if backups don't align with mSupply backups

andreievg:
Fixes #3717

Didn't really deal with B.3 use cases. Started thinking about how cursors on both sides can help re-integrate data from remote to central etc, but realised it's a bit of a rabbit hole, so I didn't want to get too deep into it. I did want to outline some problems that result with backup recency mismatch.

Also do I need to get into more details about the actual backup and restore operations for postgres ?


decisions/central server backup.md
```diff
@@ -0,0 +1,113 @@
...
+_Pros:_
+
+- Can reduce the chance of B.3 related to OMC recency
+- Upskilling support staff to be able to work with postgres WAL backups
+- No UI to implement
+
+_Cons:_
+
+- Would need archiving and unarchive scripts, that would also need to deal with file backup
+- Restoring is more techincal and may take longer
```
mark-prins, line 84: :nit 
```suggestion
- Restoring is more technical and may take longer
```

jmbrunskill:
> very nice! thanks @andreievg that's great. have popped in some spelling nits that stood out while reading - while it's a bit late in the day to comment sensibly, I'm leaning to option 3. I like the idea of being able to configure backup schedule within oms, and it's a straightforward option for support

I agree it's nice to build a solution that doesn't rely on mSupply.
It's worth thinking about what happens when the postgres server is on a different machine to open-mSupply though. I can see that happening. I guess at the stage,the  postgres backup wouldn't be OMS's responsibility but good to note I think. With this low level backup approach it sounds like you need to physically copy the datafiles, not like PGDump that can stream the data to a separate machine. Am I understanding correctly?

decisions/central server backup.md
```diff
@@ -0,0 +1,113 @@
...
+|   B.3.2   |        recent        |           old          |         recent        | Loss of data while restore of OMR (via sync initialisation), due to missing data for that site on OMC. Some data being pushed OMC from OMR may not be integrated due to missing relation dependencies (that were pushed previously but are now missing). Some new OMC central data may not travel to OMR since the cursor of OMR for central data would be ahead of current OMC                                             |
+|   B.3.3   |        recent        |         recent         |          old          | This would require some restore of backup on omSupply remote site, which is not something we recommend or support (if remote site data is lost, you should re-initialise)                                                                                                                                                                                                                                                   |
```
Chris-Petty, line 43: I have one key question - isn't B.3.3 essentially what we're trying to achieve? That's basically what we rely on for current MC prior to OMS2.0? But the complication I see is: don't we require both MC and OMC to backed up together at precisely the same time? Otherwise we're constantly going to end up with B.3.1 and B.3.2? At least for the cloud files anyway

Currently MC pauses all processes while it executes the backup process, including sync APIs. Would we not need to do the same with OMC at the same time?

Chris-Petty (commented):
Great write up!

decisions/central server backup.md
```diff
@@ -0,0 +1,113 @@
...
+- Upskilling support staff to be able to work with postgres WAL backups
+
+### Option 3 - Base backup via omSupply
+
+Pg base backup can be done [via low level API (sql statements)]https://www.postgresql.org/docs/current/continuous-archiving.html#BACKUP-LOWLEVEL-BASE-BACKUP by omSupply server. With configurations and scheduling built into omSupply
+
+_Pros:_
+
+- Should be reasonably simple
+- A.3 is handled automatically (using the same mechanism as mSupply)
```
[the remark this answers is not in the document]
andreievg: A.3 is just to send backup to cloud service. A.4 would be the consistency with mSupply bit. A.4 is not really guaranteed in any of the options, A.4 would be slightly easier with sync v7 to omSupply central.

andreievg:
@jmbrunskill 

> With this low level backup approach it sounds like you need to physically copy the datafiles, not like PGDump that can stream the data to a separate machine. Am I understanding correctly?

Yep, and actually B3 is a bit more involved. After further research, sounds like we have to copy the whole data directory, I originally though it would have some sql statements that result in a full backup. I will update the KDD a bit more

decisions/central server backup.md
andreievg: > Currently MC pauses all processes while it executes the backup process, including sync APIs. Would we not need to do the same with OMC at the same time?

Hmm i think to be fully consistent yeah, we would do this, and do both backups at exactly the same time. I added another Option 4 (which i already went with, oops), because it would deal with super pressing use case, where local data is either corrupted or completely lost, yes there would be inconsistencies between omSup central and mSup central, but it would be better then completely loosing all of omSup data (omSup will start to be populated starting in 2.2, with RnR forms. Also 2.2 backup also include files, which woudl be important if Asset functionality will start to be used)

I have one key question - isn't B.3.3 essentially what we're trying to achieve?

I actually think that most common use case would be when omSup would for some reason be corrupted or server would be lost. I think we already deal with B.3.3 with re-initialisation, and we shouldn't recommend backup on remote sites (beyond sync)

andreievg:
Added Option 4, which I think is simple and workable solution for now, also implemented it that solution via omSupply cli.
We can revisit this KDD when time comes to guarantee full consistency

decisions/central server backup.md
Chris-Petty: Yea opt4 better than nothing! 

>I actually think that most common use case would be when omSup would for some reason be corrupted or server would be lost. I think we already deal with B.3.3 with re-initialisation, and we shouldn't recommend backup on remote sites (beyond sync)

Heh I've been trying to figure out what I meant in May... I am on board with remotes not having a backup other than reinit - I just meant we're gunning for ensuring we have recent backups of OM and OMC together and any option that doesn't achieve that doesn't pass the sniff test. The recency question then just boils down to how close together the OM and OMC backups are, with the ultimate goal of being lockstep with each other.

Issue #976, opened 2022-12-20
Integration performance


andreievg:
Clemens questioned why it would take so long to integrate 100k records (altogether it took about 15 minutes, with majority of the time spent on integration). I tried the index branch, that didn't really make any difference. Tried removing inner transaction in `translation_and_integration.rs` and it took about 3 minutes, with integration being super quick. 

I think we don't really need transaction for initialisation integration (since general UI is not accessible until integration is done), thus we can remove it for initialisation/integration. But for normal sync, transaction is desired (for atomic sync), although we need to revisit this and do a KDD.

Btw the need for inner transaction was because Postgres would discard the whole transaction on error (vs sqlite, which would continue)

Still not sure what the root cause is it rollbacks due to db errors ? Or is it just the overhead of starting inner transactions ?

## Sugested fix

* During initialisation, don't use transaction
* Write up KDD about sync states (when should we fail sync entirely, when is it ok to ignore failures, what to do with those failures etc..)
* Update KDD for backup, as it suggest it takes a long time to re-initialise from central.

## Priority

Although this is not an immediate problem for near future deployments (new sites, small amount of data), it was noted that it would help testers quite a lot. Thus priority is elevated and milestone is set as Android Stabilisation

PR #3932, opened 2024-05-17
2790 integration speed up for Sqlite and Postgres

### LINKED ISSUE #2790: Improve integration speed (sqlite)
From https://www.sqlite.org/faq.html#q19:

>Actually, SQLite will easily do 50,000 or more [INSERT](https://www.sqlite.org/lang_insert.html) statements per second on an average desktop computer. But it will only do a few dozen transactions per second. Transaction speed is limited by the rotational speed of your disk drive. A transaction normally requires two complete rotations of the disk platter, which on a 7200RPM disk drive limits you to about 60 transactions per second.

Initially we integrate every record in a transaction to be able to catch errors for a particular record and then continue with the integration. But I don't believe this is true anymore so we don't need to have every record in a transaction.

I experimented with the following changes, batching up integration records into transactions. Integrating coldchain records is much faster. However, there seems to be an independent issue integrating many documents...

`let start_nested_transaction = false;` in `IntegrationRecords::integrate()` and:

```rust
pub(crate) fn translate_and_integrate_sync_records(
        &self,
        sync_records: Vec<SyncBufferRow>,
        translators: &Vec<Box<dyn SyncTranslation>>,
        mut logger: Option<&mut SyncLogger>,
    ) -> Result<TranslationAndIntegrationResults, RepositoryError> {
        let step_progress = SyncStepProgress::Integrate;
        let mut result = TranslationAndIntegrationResults::new();

        // Record initial progress (will be set as total progress)
        let total_to_integrate = sync_records.len();

        // Helper to make below logic less verbose
        let mut record_progress = |progress: usize| -> Result<(), RepositoryError> {
            match logger.as_mut() {
                None => Ok(()),
                Some(logger) => logger
                    .progress(step_progress.clone(), usize_to_u64(progress))
                    .map_err(SyncLoggerError::to_repository_error),
            }
        };

        let mut integration_batch = Vec::<IntegrationRecords>::new();
        for (number_of_records_integrated, sync_record) in sync_records.into_iter().enumerate() {
            let translation_result = match self.translate_sync_record(&sync_record, &translators) {
                Ok(translation_result) => translation_result,
                // Record error in sync buffer and in result, continue to next sync_record
                Err(translation_error) => {
                    self.sync_buffer
                        .record_integration_error(&sync_record, &translation_error)?;
                    result.insert_error(&sync_record.table_name);
                    warn!(
                        "{:?} {:?} {:?}",
                        translation_error, sync_record.record_id, sync_record.table_name
                    );
                    // Next sync_record
                    continue;
                }
            };

            let integration_records = match translation_result {
                Some(integration_records) => integration_records,
                // Record translator not found error in sync buffer and in result, continue to next sync_record
                None => {
                    let error = anyhow::anyhow!("Translator for record not found");
                    self.sync_buffer
                        .record_integration_error(&sync_record, &error)?;
                    result.insert_error(&sync_record.table_name);
                    warn!(
                        "{:?} {:?} {:?}",
                        error, sync_record.record_id, sync_record.table_name
                    );
                    // Next sync_record
                    continue;
                }
            };

            integration_batch.push(integration_records);
            if integration_batch.len() >= 500
                || number_of_records_integrated + 1 == total_to_integrate
            {
                self.connection
                        for integration_records in &integration_batch {
                            // Integrate
                            let integration_result = integration_records.integrate(connection);
                            match integration_result {
                                Ok(_) => {
                                    self.sync_buffer
                                        .record_successful_integration(&sync_record)?;
                                    result.insert_success(&sync_record.table_name)
                                }
                                // Record database_error in sync buffer and in result
                                Err(database_error) => {
                                    let error = anyhow::anyhow!("{:?}", database_error);
                                    self.sync_buffer
                                        .record_integration_error(&sync_record, &error)?;
                                    result.insert_error(&sync_record.table_name);
                                    warn!(
                                        "{:?} {:?} {:?}",
                                        error, sync_record.record_id, sync_record.table_name
                                    );
                                }
                            }
                        }

                        // round down to closest step size
                        record_progress(
                            (total_to_integrate - number_of_records_integrated - 1)
                                / PROGRESS_STEP_LEN
                                * PROGRESS_STEP_LEN,
                        )?;

                        Ok(())
                    })
                    .map_err(|e: TransactionError<RepositoryError>| e.to_inner_error())?;

                integration_batch.clear();
            }
        }
        assert!(integration_batch.is_empty());

        // Record final progress
        record_progress(0)?;

        Ok(result)
    }
```

TODO:
- [ ] Remove start_nested_transaction (?)
- [ ] Benchmarking and testing for sqlite and postgres

clemens-msupply:
Fixes #2790

I first followed the approach to keep integrating in a single tx for Sqlite. This gave good speedup. However, this prevents the UI from showing progress because the status query is blocked by sync transaction.

@andreievg had the idea to cache the status in memory. However, it turned out to be a bit nasty. e.g. there was also the latest successful sync request that needed to be cached + many other calls to the repo. All doable to cache this as well, but I think the code would have been quite messy; some parts would have used the cache and some wouldn't (had to look at the frontend code to figure out which requests need to be cached...

Then I investigate why safepoints in Postgres are slow when overused and found this article:
https://www.cybertec-postgresql.com/en/subtransactions-and-performance-in-postgresql/#conclusion
It suggested that you shouldn't use more than 64 subtransactions.

For this reason I went back to the integration in batches idea, using a batch size of 64 for Postgres.

# 👩🏻‍💻 What does this PR do?

Refactors code to make sync tx behaviour configurable in a central place and configure it depending on postgres/sqlite or on if it is durring the initialization.

## Behaviour Sqlite:
- During initialization:
    - Split record integration into batches of 500
    - Individual records are not executed in subtransaction (not needed because sqlite doesn't cancel the tx when there was an error)
- After initialzation:
    - One single parent tx (as before)
    - Individual records are not executed in subtransaction

## Behaviour PG:

- During initialization:
    - Split record integration into batches of 64
    - Individual records are executed in subtransaction (to be able to handle errors)
- After initialzation:
    - One single parent tx (as before)
    - Individual records are executed in subtransaction

## Some benchmark results for cold chain dataset

https://drive.google.com/file/d/15p1Hq_2jERVtXJI9MAv94etuZUT2DkDz/view?usp=sharing

I timed the last integration step:

### Sqlite:
old: 76.7s

new: 26.3s (batch size 64)
new: 7.2s (batch size 500)

### PG:
old: 111.4s

new: 23.6s (batch size 64)

To measure times I did the following in `integrate_and_translate_sync_buffer()`
```rust
    let start = std::time::SystemTime::now();
    let result = if batch_size.is_none() {
        connection
            .transaction_sync(integrate_and_translate)
            .map_err::<RepositoryError, _>(|e| e.to_inner_error())
    } else {
        integrate_and_translate(&connection)
    }?;
    println!("ELAPSED: {:?}", start.elapsed().unwrap());

```

I haven't experimented with other batch sizes other than 64 for Postgres. Not sure if we can raise it or if there is a better sweet spot...

# 🧪 Testing
- [ ] Integrate large dataset for postgres and sqlite


server/service/src/sync/translation_and_integration.rs
```diff
@@ -85,16 +85,55 @@ impl<'a> TranslationAndIntegration<'a> {
...
         // Record initial progress (will be set as total progress)
         let total_to_integrate = sync_records.len();
+        let batch_size = match batch_size {
+            Some(batch_size) => batch_size,
+            None => total_to_integrate,
+        };
+        while position < total_to_integrate {
+            let current_batch_size = std::cmp::min(position + batch_size, total_to_integrate);
+            let batch = &sync_records[position..current_batch_size];
+            connection.transaction_sync(|_| {
```
clemens-msupply, line 107: If already in a transaction this doesn't start a new tx...

andreievg:
> @andreievg had the idea to cache the status in memory. However, it turned out to be a bit nasty. e.g. there was also the latest successful sync request that needed to be cached + many other calls to the repo

@clemens-msupply, when i suggested the idea i tested in and it worked.

andreievg:
> It suggested that you shouldn't use more than 64 subtransactions.

Technically it's possible to create more then 64 upserts or deletes form 64 sync records

server/service/src/sync/translation_and_integration.rs
```diff
@@ -196,21 +239,13 @@ impl IntegrationOperation {
 
 pub(crate) fn integrate(
     connection: &StorageConnection,
+    records_in_tx: bool,
```
andreievg, line 242: it's hard to trace USE_RECORD_TX to here, can we just use start_nested_transaction in combination with feature flag here ?

andreievg:
@clemens-msupply how would we test this change ? (mainly to guarantee that it still works for the future, as I see it getting more and more difficult to understand how integration works).

I am still leaning towards the solution that we discussed in the issue.

andreievg:
@clemens-msupply 

> It suggested that you shouldn't use more than 64 subtransactions.

in the linked issues, i made example performance tests. Can you make one please to show the difference the 64 sub transactions make ?

server/service/src/sync/translation_and_integration.rs
clemens-msupply: hehe that's what I found super hard and why I changed it. For example, when doing some speed tests I tried various combinations. However, I kept getting it wrong because I jumped around between files and set the wrong flags in the wrong files (because having the long waits between runs). For this reason I moved the configs to a one single section in the code where you can configure the behaviour as described in the PR.

Regarding `start_nested_transaction`, doing batch inserts in a transaction is usually always faster. While having this subtransaction problem for Postgres, inserting records in smaller batches in Postgres is still faster (5 times faster for this particular dataset!). For this reason we should always use transactions if we want to make integration faster, i.e. there will always be a parent transaction.

clemens-msupply:
> @clemens-msupply
> 
> > It suggested that you shouldn't use more than 64 subtransactions.
> 
> in the linked issues, i made example performance tests. Can you make one please to show the difference the 64 sub transactions make ?

Yes think would be good for a follow-up issue to optimize the batch sizes. Would be interesting to know if 64 really is the best number and how fast performance deteriorate. For example, if it's a problem if there are batches of sync records that produce 128 insert records or more.

andreievg (changes requested):
I was hoping we could agree on a solution in the issue, before implementation. And I think we see the solution a bit differently, here are my thoughts:

The solution I am leaning towards is the one in this branch: https://github.com/msupply-foundation/open-msupply/compare/develop...2790-integration-speed-alternative

For initialisation, do transaction for sqlite, but not for postgres, as per the test I shared I don't see that much difference between doing upsert/deletes in transaction or without transaction for postgres, but significant difference doing a transaction for sqlite. Add cache for sync log row to show progress for sqlite (since it's in transaction during initialisation)

I did some test to compare solutions, with the data file you supplied

**Sqlite - my PC**

on develop

ELAPSED: 13.275523s

on simple transaction

ELAPSED: 5.737334s

on batched transaction

ELAPSED: 6.429973s

**Sqlite - samsung tablet**

on develop

ELAPSED: 106.366464231s

on simple transaction

ELAPSED: 7.033277572s

on batched transaction

ELAPSED: 9.71245657s

**Postgres - my PC**. Was wondering, what hardware did you run your tests on ? As per the tests that I did in the issues, I didn't see much difference between transactions and not transaction upsert/insert of postgres data (only nested transactions were bottle neck), and since initialisation already doesn't use transaction, I was wondering how you get 5x improvement for postgres.

on develop

ELAPSED: 12.701915s

on simple transaction

ELAPSED: 13.282544s

on batched transaction

ELAPSED: 13.001174s

I trust that batched solution works, but I think it complicates integration logic, and I am confused about the benefits it offers vs the other solution we discussed. 

One possible benefit is faster operational sync, but we want to do operational sync in atomic way, to avoid incomplete data (for example transfers with missing lines), so I don't think we can do batched sync integration in operational mode. Also the issue does not raise a problem with operational sync, and the changes in this PR don't apply to operational sync.

Let's have a meeting to discuss, btw the branch I shared is not quite ready, it would need a small test to check for sync log cache (but the cache does work in UI)

Also was wondering how we can improve the communication on this issue, as we've spend quite a bit of time going backwards and forwards, and I thought we agreed on a solution in theory, it's quite time consuming to be doing alternative implementations of something to show as a practical difference. Do you have a suggestion on how we can improve on this front ?

clemens-msupply:
I did some more testing with your version. I noticed there is a problem with the `is_initialised` variable which is passed in as the `is_initialising` parameter. Weirdly this gives good results for me as well, i.e. with this config for the cold chain datafile:

is_initialised: false
execute_in_transaction: true
start_nested_transaction: true

So Postgres seem to be working fine with parent and nested tx!

However, this doesn't fit together with some tests that I did earlier on another datafile which contained lots of patient data. Here, this configuration (parent + nested tx) was very slow while the batched version worked fine.

Wonder what datafile we used when we initially saw slowdowns because the cold chain datafile doesn't seem to have these problems, or something else changed?

Lets discuss when in the office.

clemens-msupply:
I added a very simple benchmark to test various parameters and their combinations:
- number of records to insert
- error/no error
- parent tx
- nested tx

On my machine I found that parent tx + nested tx gives good results for PG and Sqlite. While Sqlite seems to be a bit faster without the nested tx, I just updated the code to have the same behaviour for Postgres and Sqlite, i.e. parent + nested tx.

So I couldn't really reproduce the initial problem... (https://github.com/msupply-foundation/open-msupply/issues/976)

I also use the row cache now. Initially "cache" confused me because you load the DB row first and then replace the returned row with the cache entry (which looks wrong for a cache). So I rephrased the cache name and added some more comments to make it more clear how it all works together...

Below are my benchmark results. @andreievg could you please check the results on your machine? Have the feeling we are getting different results on our machines...

> ## Sqlite:
> With error:
> 
> Batch size: 64
> with_error: true, n: 64, parent_tx: false, nested_tx: false, Time: 8.937166ms
> with_error: true, n: 64, parent_tx: false, nested_tx: true, Time: 11.655709ms
> with_error: true, n: 64, parent_tx: true, nested_tx: false, Time: 6.277754ms
> with_error: true, n: 64, parent_tx: true, nested_tx: true, Time: 6.051856ms
> Batch size: 500
> with_error: true, n: 500, parent_tx: false, nested_tx: false, Time: 62.732543ms
> with_error: true, n: 500, parent_tx: false, nested_tx: true, Time: 68.669774ms
> with_error: true, n: 500, parent_tx: true, nested_tx: false, Time: 19.322122ms
> with_error: true, n: 500, parent_tx: true, nested_tx: true, Time: 17.365983ms
> Batch size: 5000
> with_error: true, n: 5000, parent_tx: false, nested_tx: false, Time: 592.663885ms
> with_error: true, n: 5000, parent_tx: false, nested_tx: true, Time: 764.618219ms
> with_error: true, n: 5000, parent_tx: true, nested_tx: false, Time: 181.841167ms
> with_error: true, n: 5000, parent_tx: true, nested_tx: true, Time: 140.540523ms
> 
> Without error:
> 
> Batch size: 64
> with_error: false, n: 64, parent_tx: false, nested_tx: false, Time: 483.592409ms
> with_error: false, n: 64, parent_tx: false, nested_tx: true, Time: 254.831236ms
> with_error: false, n: 64, parent_tx: true, nested_tx: false, Time: 5.974125ms
> with_error: false, n: 64, parent_tx: true, nested_tx: true, Time: 7.056245ms
> Batch size: 500
> with_error: false, n: 500, parent_tx: false, nested_tx: false, Time: 3.827160699s
> with_error: false, n: 500, parent_tx: false, nested_tx: true, Time: 1.98352835s
> with_error: false, n: 500, parent_tx: true, nested_tx: false, Time: 21.03948ms
> with_error: false, n: 500, parent_tx: true, nested_tx: true, Time: 20.337717ms
> Batch size: 5000
> with_error: false, n: 5000, parent_tx: false, nested_tx: false, Time: 56.716187198s
> with_error: false, n: 5000, parent_tx: false, nested_tx: true, Time: 59.136125105s
> with_error: false, n: 5000, parent_tx: true, nested_tx: false, Time: 246.036333ms
> with_error: false, n: 5000, parent_tx: true, nested_tx: true, Time: 371.996773ms
> 
> 
> ## Postgres
> With error:
> Warm up
> with_error: true, n: 64, parent_tx: true, nested_tx: true, Time: 14.238771ms
> Batch size: 64
> with_error: true, n: 64, parent_tx: false, nested_tx: false, Time: 17.791893ms
> with_error: true, n: 64, parent_tx: false, nested_tx: true, Time: 19.937688ms
> with_error: true, n: 64, parent_tx: true, nested_tx: false, Time: 7.699331ms
> with_error: true, n: 64, parent_tx: true, nested_tx: true, Time: 7.349395ms
> Batch size: 500
> with_error: true, n: 500, parent_tx: false, nested_tx: false, Time: 116.502237ms
> with_error: true, n: 500, parent_tx: false, nested_tx: true, Time: 135.290812ms
> with_error: true, n: 500, parent_tx: true, nested_tx: false, Time: 54.641849ms
> with_error: true, n: 500, parent_tx: true, nested_tx: true, Time: 50.341707ms
> Batch size: 5000
> with_error: true, n: 5000, parent_tx: false, nested_tx: false, Time: 1.065878781s
> with_error: true, n: 5000, parent_tx: false, nested_tx: true, Time: 1.369578877s
> with_error: true, n: 5000, parent_tx: true, nested_tx: false, Time: 502.214964ms
> with_error: true, n: 5000, parent_tx: true, nested_tx: true, Time: 870.421896ms
> Without error:
> Warm up
> with_error: false, n: 64, parent_tx: true, nested_tx: true, Time: 38.458083ms
> Batch size: 64
> with_error: false, n: 64, parent_tx: false, nested_tx: false, Time: 613.642473ms
> with_error: false, n: 64, parent_tx: false, nested_tx: true, Time: 281.115217ms
> with_error: false, n: 64, parent_tx: true, nested_tx: false, Time: 19.166011ms
> with_error: false, n: 64, parent_tx: true, nested_tx: true, Time: 32.634006ms
> Batch size: 500
> with_error: false, n: 500, parent_tx: false, nested_tx: false, Time: 4.823579473s
> with_error: false, n: 500, parent_tx: false, nested_tx: true, Time: 1.733486213s
> with_error: false, n: 500, parent_tx: true, nested_tx: false, Time: 156.990133ms
> with_error: false, n: 500, parent_tx: true, nested_tx: true, Time: 177.055719ms
> Batch size: 5000
> with_error: false, n: 5000, parent_tx: false, nested_tx: false, Time: 15.259303502s
> with_error: false, n: 5000, parent_tx: false, nested_tx: true, Time: 9.167597354s
> with_error: false, n: 5000, parent_tx: true, nested_tx: false, Time: 1.562396011s
> with_error: false, n: 5000, parent_tx: true, nested_tx: true, Time: 1.550798405s
>

andreievg:
@clemens-msupply do you mind sharing the tests you've done ?

I did some tests with errors there is a huge difference between nested transactions in sqlite, [here is the test addition diff](https://github.com/msupply-foundation/open-msupply/compare/develop(Checking-transactions-in-database)(Nested-transaction)...develop(Checking-transactions-in-database)(Nested-transaction)(Nested-transactions-with-errors))

Basically for 100k records, with errors (every second record), we go from 4 seconds (single parent transaction) to 66 seconds (nested transactions). 
In postgres, i didn't test with errors and single transaction (since a single error will cause the whole transaction to rollback), but it took 287 seconds with nested transactions.

After writting the last paragraph I also tested on develop with diesel 2 and did non transaction for postgres as the first test, and got 17seconds without any transaction and still about 280seconds for nested transaction.

Here is the latest test on develop: https://github.com/msupply-foundation/open-msupply/compare/develop...develop(Testing-integration)

I think above may explain the original issue, what do you think ?

clemens-msupply:
Good idea to only have every second record as error, it's a bit more realistic. My benchmark is in this PR as a disabled test `bench_error_performance()`.

I also found a bug in this test with the nested tx! Think the new numbers match what you were seeing.

The main speed up comes from having a parent tx though.

```
Sqlite:

With error:

Batch size: 64
with_error: true, n: 64, parent_tx: false, nested_tx: false, Time: 288.684099ms
with_error: true, n: 64, parent_tx: false, nested_tx: true, Time: 150.439219ms
with_error: true, n: 64, parent_tx: true, nested_tx: false, Time: 5.748226ms
with_error: true, n: 64, parent_tx: true, nested_tx: true, Time: 7.450794ms
Batch size: 500
with_error: true, n: 500, parent_tx: false, nested_tx: false, Time: 2.197851972s
with_error: true, n: 500, parent_tx: false, nested_tx: true, Time: 1.132146076s
with_error: true, n: 500, parent_tx: true, nested_tx: false, Time: 19.138553ms
with_error: true, n: 500, parent_tx: true, nested_tx: true, Time: 46.222733ms
Batch size: 10000
with_error: true, n: 10000, parent_tx: false, nested_tx: false, Time: 43.490857221s
with_error: true, n: 10000, parent_tx: false, nested_tx: true, Time: 23.599305956s
with_error: true, n: 10000, parent_tx: true, nested_tx: false, Time: 498.407058ms
with_error: true, n: 10000, parent_tx: true, nested_tx: true, Time: 2.602793962s

Without error:

Batch size: 64
with_error: false, n: 64, parent_tx: false, nested_tx: false, Time: 1.094032579s
with_error: false, n: 64, parent_tx: false, nested_tx: true, Time: 614.818034ms
with_error: false, n: 64, parent_tx: true, nested_tx: false, Time: 48.833931ms
with_error: false, n: 64, parent_tx: true, nested_tx: true, Time: 56.872775ms
Batch size: 500
with_error: false, n: 500, parent_tx: false, nested_tx: false, Time: 8.889464051s
with_error: false, n: 500, parent_tx: false, nested_tx: true, Time: 5.300856687s
with_error: false, n: 500, parent_tx: true, nested_tx: false, Time: 163.714898ms
with_error: false, n: 500, parent_tx: true, nested_tx: true, Time: 179.391689ms
Batch size: 10000
with_error: false, n: 10000, parent_tx: false, nested_tx: false, Time: 99.570432136s
with_error: false, n: 10000, parent_tx: false, nested_tx: true, Time: 49.272595314s
with_error: false, n: 10000, parent_tx: true, nested_tx: false, Time: 770.066036ms
with_error: false, n: 10000, parent_tx: true, nested_tx: true, Time: 1.063902542s


Postgres:

With error:
Batch size: 64
with_error: true, n: 64, parent_tx: false, nested_tx: false, Time: 20.601337ms
with_error: true, n: 64, parent_tx: false, nested_tx: true, Time: 23.41676ms
with_error: true, n: 64, parent_tx: true, nested_tx: false, Time: 9.036724ms
with_error: true, n: 64, parent_tx: true, nested_tx: true, Time: 23.867568ms
Batch size: 500
with_error: true, n: 500, parent_tx: false, nested_tx: false, Time: 153.203243ms
with_error: true, n: 500, parent_tx: false, nested_tx: true, Time: 244.125908ms
with_error: true, n: 500, parent_tx: true, nested_tx: false, Time: 65.561212ms
with_error: true, n: 500, parent_tx: true, nested_tx: true, Time: 207.617083ms
Batch size: 10000
with_error: true, n: 10000, parent_tx: false, nested_tx: false, Time: 3.045106257s
with_error: true, n: 10000, parent_tx: false, nested_tx: true, Time: 3.797585268s
with_error: true, n: 10000, parent_tx: true, nested_tx: false, Time: 1.168556298s
with_error: true, n: 10000, parent_tx: true, nested_tx: true, Time: 4.276466748s

Without error:

Batch size: 64
with_error: false, n: 64, parent_tx: false, nested_tx: false, Time: 24.248444ms
with_error: false, n: 64, parent_tx: false, nested_tx: true, Time: 25.148084ms
with_error: false, n: 64, parent_tx: true, nested_tx: false, Time: 27.052965ms
with_error: false, n: 64, parent_tx: true, nested_tx: true, Time: 28.214085ms
Batch size: 500
with_error: false, n: 500, parent_tx: false, nested_tx: false, Time: 185.510442ms
with_error: false, n: 500, parent_tx: false, nested_tx: true, Time: 208.922228ms
with_error: false, n: 500, parent_tx: true, nested_tx: false, Time: 172.570572ms
with_error: false, n: 500, parent_tx: true, nested_tx: true, Time: 194.577553ms
Batch size: 10000
with_error: false, n: 10000, parent_tx: false, nested_tx: false, Time: 3.765842179s
with_error: false, n: 10000, parent_tx: false, nested_tx: true, Time: 4.181659326s
with_error: false, n: 10000, parent_tx: true, nested_tx: false, Time: 3.301358389s
with_error: false, n: 10000, parent_tx: true, nested_tx: true, Time: 4.037525731s
```

andreievg:
@clemens-msupply thanks for the catch up, with 100k records and 5 percent error rate, i see small difference between no transaction and nested transaction in postgres, and for sqlite, 4 second (single tx) to 12 seconds (nested tx). I guess we could just have one condition to not do nested tx for sqlite ? 
https://github.com/msupply-foundation/open-msupply/compare/develop(Testing-integration)...develop(Testing-integration)(With-5-percent)

clemens-msupply:
Thanks, yes can confirm. I updated the code as follows:

Postgres: parent tx + nested tx
Sqlite: Parent tx but NO nested tx

andreievg (approved):
Thanks @clemens-msupply for this bit of performance work and for the thorough discussions we had !

clemens-msupply:
> Thanks @clemens-msupply for this bit of performance work and for the thorough discussions we had !

Thanks for the feedback and benchmarks!

PR #3937, opened 2024-05-19
add needs daily triage label to bug report issues

lache-melvin:
# 👩🏻‍💻 What does this PR do?

Adds new `needs daily triage` label to bug report issue template.

This allows the release management team to know which newly reported bugs need to be triaged each day. Especially if this responsibility is shared, will help to stop people doubling up on work.

Any hotfix bugs (and high severity?) can then be assigned immediately, while others can wait for usual weekly triage.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,13 +2,13 @@
 name: Bug report
 about: Create a report for a bug or to help us improve
 title: A brief clear title describing the bug from a user's perspective
-labels: bug, needs triage
-assignees: ''
-
+labels: bug, needs triage, needs daily triage
+assignees: ""
 ---
 
 ## What went wrong? 😲
-<!-- Provide a clear and concise description of what the bug is. Screenshots are helpful! --> 
+
+<!-- Provide a clear and concise description of what the bug is. Screenshots are helpful! -->
 
 ## Expected behaviour 🤔
 
@@ -22,7 +22,9 @@ Steps to reproduce the behaviour:
 4. See error
 
 ## Your environment 🌱
+
 <!-- e.g. 1.2.3 -->
+
 - Open mSupply Version:
 - Legacy mSupply Central Server Version:
 <!-- e.g. android, browser (extra points if you tell us which one), desktop (windows), desktop (macOS), server (windows) -->


mark-prins:
Just checking - `needs daily triage` is equivalent to `Bug + Needs triage` right? 
Is there a need to have an extra label? can you just filter by the combination of existing labels?

lache-melvin:
> Just checking - `needs daily triage` is equivalent to `Bug + Needs triage` right? Is there a need to have an extra label? can you just filter by the combination of existing labels?

I know it's a little pedantic, but in that case how do we know which bugs have been left for weekly triage vs not looked at yet?

CarlosNZ:
> I know it's a little pedantic, but in that case how do we know which bugs have been left for weekly triage vs not looked at yet?

If the next release management teams doesn't like to do it this way, they are welcome to remove this label. :)

mark-prins:
> can you just filter by the combination of existing labels?

ah, right. so the answer is 'no' 😆

roxy-dao:
Hmm, I like just using the one `needs triage` label for this, but if you're finding it easier with another label then we can try it out ig (:

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,13 +2,13 @@
 name: Bug report
 about: Create a report for a bug or to help us improve
 title: A brief clear title describing the bug from a user's perspective
-labels: bug, needs triage
-assignees: ''
-
+labels: bug, needs triage, needs daily triage
```
mark-prins, line 5: needs both labels? I would have thought that just `needs daily triage` is enough and
the `needs (weekly) triage` can be added if required - clearly I'm not understanding the intended process here!

are you expecting that all bugs would typically be triaged twice?

mark-prins (approved):
👍 It would be good to have the triage process / release management team process defined somewhere, it might remove my need to query stuff!

.github/ISSUE_TEMPLATE/bug_report.md
roxy-dao: Would need to filter both in triage sessions... or have both removed by the person triaging the bug 👀 ......

lache-melvin: > are you expecting that all bugs would typically be triaged twice?

Yep - daily triage is just to capture anything for hotfix (and maybe assign a preliminary severity to help speed up weekly triage?) Otherwise wouldn't expect release team to triage all bugs, they may not have all the context - this should still go through main weekly triage team 🙏 

I will add to release management docs!

lache-melvin: For reference! https://github.com/msupply-foundation/open-msupply/wiki/Release-process#daily-bug-triage

PR #3958, opened 2024-05-20
3111 oms central sync versioning

### LINKED ISSUE #3111: Versioning for omSupply central server
Similar to the way we do sync with mSupply central (keep backwards compatible)
Would include adding to compatibility matrix on docs page

lache-melvin:
Fixes #3111

# 👩🏻‍💻 What does this PR do?

Adds version check for V6 sync API:

Screenshot 2024-05-21 at 10 48 16 AM

Also will create a PR in the docs repo for the new version matrix!

# 🧪 Testing

- [ ] Start OMS central server and OMS remote server
- [ ] Run sync for remote site
- [ ] Sync runs successfully
- [ ] In `service/src/sync/setting.rs`, increment the v6 sync version
- [ ] Restart remote server
- [ ] Run sync for remote site
- [ ] See the V6 sync incompatible error

# 📃 Documentation

  1. Version matrix
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
@@ -268,7 +268,6 @@ export const Summary = ({ draft, onChange, locations }: SummaryProps) => {
           <Section heading={t('label.asset-properties')}>
             {Object.entries(draft.parsedProperties).map(([key, value]) => (
               <Row key={key} label={`${value}`}>
-              >
                 <BasicTextInput
                   value={draft.parsedProperties[key]}
                   disabled
--- a/server/service/src/sync/api_v6/core.rs
+++ b/server/service/src/sync/api_v6/core.rs
@@ -8,6 +8,7 @@ use super::*;
 pub(crate) struct SyncApiV6 {
     pub(crate) url: Url,
     pub(crate) sync_v5_settings: SyncApiSettings,
+    pub(crate) sync_v6_version: u32,
 }
 
 #[derive(Error, Debug)]
@@ -22,6 +23,7 @@ impl SyncApiV6 {
     pub fn new(
         url: &str,
         sync_v5_settings: &SyncApiSettings,
+        sync_v6_version: u32,
     ) -> Result<Self, SyncApiV6CreatingError> {
         let mut url = Url::parse(url)
             .map_err(|error| SyncApiV6CreatingError::CannotParseSyncUrl(url.to_string(), error))?;
@@ -31,6 +33,7 @@ impl SyncApiV6 {
         Ok(Self {
             url,
             sync_v5_settings: sync_v5_settings.clone(),
+            sync_v6_version,
         })
     }
 
@@ -43,6 +46,7 @@ impl SyncApiV6 {
         let Self {
             sync_v5_settings,
             url,
+            sync_v6_version,
         } = self;
 
         let route = "pull";
@@ -53,6 +57,7 @@ impl SyncApiV6 {
             batch_size,
             sync_v5_settings: sync_v5_settings.clone(),
             is_initialised,
+            sync_v6_version: sync_v6_version.clone(),
         };
 
         let result = Client::new().post(url.clone()).json(&request).send().await;
@@ -74,6 +79,7 @@ impl SyncApiV6 {
         let Self {
             sync_v5_settings,
             url,
+            sync_v6_version,
         } = self;
 
         let route = "push";
@@ -82,6 +88,7 @@ impl SyncApiV6 {
         let request = SyncPushRequestV6 {
             batch,
             sync_v5_settings: sync_v5_settings.clone(),
+            sync_v6_version: sync_v6_version.clone(),
         };
 
         let result = Client::new().post(url.clone()).json(&request).send().await;
@@ -103,6 +110,7 @@ impl SyncApiV6 {
         let Self {
             sync_v5_settings,
             url,
+            sync_v6_version,
         } = self;
 
         let route = "site_status";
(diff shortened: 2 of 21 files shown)


server/service/src/sync/api_v6/core.rs
```diff
@@ -8,6 +8,7 @@ use super::*;
 pub(crate) struct SyncApiV6 {
     pub(crate) url: Url,
     pub(crate) sync_v5_settings: SyncApiSettings,
+    pub(crate) sync_v6_version: u32,
```
lache-melvin, line 11: I _think_ this makes the most sense, having `sync_v6_version` separate from the `sync_v5_settings`. Let me know if you think it should be passed in/around differently though?

server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -37,13 +41,23 @@ pub async fn pull(
...
 
     if !CentralServerConfig::is_central_server() {
         return Err(Error::NotACentralServer);
     }
+
+    if !is_sync_version_compatible(sync_v6_version) {
+        return Err(Error::SyncVersionMismatch(
+            MIN_VERSION,
+            MAX_VERSION,
+            sync_v6_version,
```
lache-melvin, line 57: curious for thoughts - the repetition of this in each method makes me want to return this error from the `is_sync_version_compatible` and just use `?`, but I like explicitly reading through the method and seeing the different errors returned?

Making me wonder whether we have a `validate` method or something for shared checks like is_central_server, is_sync_version_compatible, credential check?

client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
```diff
@@ -268,7 +268,6 @@ export const Summary = ({ draft, onChange, locations }: SummaryProps) => {
           <Section heading={t('label.asset-properties')}>
             {Object.entries(draft.parsedProperties).map(([key, value]) => (
               <Row key={key} label={`${value}`}>
-              >
```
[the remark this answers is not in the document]
andreievg: Wow, how did this sneak in, probably doing a commit with pre-commits turned off right ?

server/service/src/sync/api_v6/mod.rs
```diff
@@ -157,19 +159,22 @@ pub struct SyncPullRequestV6 {
     pub(crate) batch_size: u32,
     pub(crate) sync_v5_settings: SyncApiSettings,
     pub(crate) is_initialised: bool,
+    pub(crate) sync_v6_version: u32,
```
andreievg, line 162: I think we need to set this to default to 1. Would be good to test with 2.0 omSup remote to 2.1 omSup remote (would suspect there would be serde error,  unless default is set)

server/service/src/sync/sync_on_central/mod.rs
andreievg: Yeah ! We've actually [added middleware](https://github.com/msupply-foundation/open-msupply/blob/526046cd357ac9d6c8af5aa4d02692c8e8f2c8cb/server/server/src/middleware/central_server_only.rs), which i think is to complex (or maybe too verbose is the word), would rather a common validate method

andreievg (changes requested):
Looks great thank, i do have one request, and it's the default app version, please see comment.

Thanks for noting that version matrix needs to be updated

server/service/src/sync/api_v6/core.rs
andreievg: Yep, sounds good, thanks

server/service/src/sync/api_v6/mod.rs
```diff
@@ -157,19 +159,22 @@ pub struct SyncPullRequestV6 {
...
     pub(crate) batch: SyncBatchV6,
     pub(crate) sync_v5_settings: SyncApiSettings,
+    pub(crate) sync_v6_version: u32,
 }
 
 #[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SiteStatusRequestV6 {
     pub(crate) sync_v5_settings: SyncApiSettings,
+    pub(crate) sync_v6_version: u32,
```
andreievg, line 177: It's slightly more comiplex, but I think for v7 api (when we sync to omsuplly central with omsupply central credentials), we would want something like the following (example doesn't annotate serde etc..)

```
stuct RequestV7<T> {
   common: // credentials, version etc..
   data: <T>
}

struct PushPayload {
   batch: SyncBatchv7
}

type PushV7 =  Request<PushPayload>
```

server/service/src/sync/api_v6/mod.rs
lache-melvin: Ah yeah you're right - is this the best way to implement that? Feels a bit icky? Maybe should have a sync_v6_settings struct instead so default is only set in one place??

https://github.com/msupply-foundation/open-msupply/compare/3111-oms-central-sync-versioning...3111-oms-central-sync-versioning(default-sync-version)

lache-melvin: > Ah yeah you're right - is this the best way to implement that? Feels a bit icky? Maybe should have a sync_v6_settings struct instead so default is only set in one place?? Or maybe there is a better way in general lol
> 
> [3111-oms-central-sync-versioning...3111-oms-central-sync-versioning(default-sync-version)](https://github.com/msupply-foundation/open-msupply/compare/3111-oms-central-sync-versioning...3111-oms-central-sync-versioning(default-sync-version))

andreievg (approved):
Thanks @lache-melvin 

approving with this one: https://github.com/msupply-foundation/open-msupply/compare/3111-oms-central-sync-versioning...3111-oms-central-sync-versioning(default-sync-version)

yeah might be a little yuck, but i am ok with repetition here (given that next api will be v7). Maybe default would work, will set it to 0 and that should still work right ?

lache-melvin:
> Maybe default would work, will set it to 0 and that should still work right ?

Ohhh nice, yeah the 0 is a bit better, thank you! 🙏

PR #3962, opened 2024-05-21
make min=0 the default for number input cell

### LINKED ISSUE #3950: Cannot backspace the value to 0 in tax-field of service-charges
## What went wrong? 😲
When you try to backspace the value of `tax` field under service-charges, notice that you simply can't
Also try inputting numbers like 123, it looks weird.

## Expected behaviour 🤔
value should set to 0 (not 1) on full backspace

## How to Reproduce 🔨

1. Have some service items set up
2. Create a outbound and add lines
3. Enter a service-charge
4. Add some values to `tax` field
5. Backspace it.
6. See error: it simply doesn't reset to 0 at all. Similar issue: #3834 

## Screenshots

https://github.com/msupply-foundation/open-msupply/assets/59193490/bb08e75e-1a80-4b25-bfa4-a101c2311b83

## Your environment 🌱

- Open mSupply Version: v2.0.0
- Legacy mSupply Central Server Version: v7-14-06

- Platform: browser (chrome), apk

- Database type: sqlite

lache-melvin:
Fixes #3950

# 👩🏻‍💻 What does this PR do?

Backspace to 0 now works for tax input in the two service charge modals:

https://github.com/msupply-foundation/open-msupply/assets/55115239/f383cdfa-28c2-419d-8b76-ebfd5023ff17

Removes the default min=1 on the NumberInputCell, because everywhere except the service charge tax cells, we were explicitly setting min=0, and we actually don't want a min of 1 for service charge tax either 😅 

# 🧪 Testing

- [ ] In an Inbound Shipment, add a service charge
- [ ] Enter a value in `tax` field, then backspace, you should be able to reset to 0
- [ ] In an Outbound Shipment, add a service charge
- [ ] Enter a value in `tax` field, then backspace, you should be able to reset to 0
- [ ] In inbound/outbound returns & shipments, and prescriptions, validate still can clear all inputs back to 0

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
@@ -14,9 +14,9 @@ export const NumberInputCell = <T extends RecordWithId>({
   rowIndex,
   columnIndex,
   isDisabled = false,
-  // Make the default min=1 as this is the typical implementation
+  // Make the default min=0 as this is the typical implementation
   // in Data Tables
-  min = 1,
+  min = 0,
   max,
   decimalLimit,
   step,
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -75,7 +75,6 @@ const NumberOfPacksCell: React.FC<CellProps<DraftInboundLine>> = ({
 }) => (
   <NumberInputCell
     {...props}
-    min={0}
     isRequired={rowData.numberOfPacks === 0}
     rowData={rowData}
   />
--- a/client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
+++ b/client/packages/invoices/src/Returns/modals/InboundReturn/ReturnQuantitiesTable.tsx
@@ -127,7 +127,6 @@ const NumberOfPacksReturnedInputCell: React.FC<
     {...props}
     isRequired
     max={props.rowData.numberOfPacksIssued ?? undefined}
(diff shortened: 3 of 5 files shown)


client/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NumberInputCell.tsx
```diff
@@ -14,9 +14,9 @@ export const NumberInputCell = <T extends RecordWithId>({
   rowIndex,
   columnIndex,
   isDisabled = false,
-  // Make the default min=1 as this is the typical implementation
+  // Make the default min=0 as this is the typical implementation
   // in Data Tables
-  min = 1,
+  min = 0,
```
lache-melvin, line 19: I think this is an interim solution. Only min=0 works for this component, otherwise can't clear the value when trying to change it (e.g. can't backspace `1` to change from `1` to `2`)

This is because of the `onChange` method below - it resets the input value to the `min` if it receives undefined (i.e. empty input)

We do this because we then call the `updater`, which updates row data. Most places have other columns in the row that are calculated (e.g. you're entering # packs, which affects calculated # units field), so we can't store undefined against the row value (as # packs = undefined would result in # units = NaN).

It would be nice to store a safe default (0 or min) against the row, but have the input itself be clearable, at least while in focus. Note: for reference, this is what was done in the PackVariantEntryCell, which I think works well.

However, we can't currently do this in this component, because we use buffer state instead of typical react state to store the current input value. This is done to handle potential changes to the input value made externally to the component (e.g. changing "issue in packs of" changes the number of packs elsewhere).

All this to say, I think actually having min values in this component/in tables is a bigger refactor, not really hot fix worthy. 

Unless there is a solution I'm not seeing?

andreievg:
@lache-melvin can you please rebase to main

andreievg (approved):
Looks good as far as i can tell. And thanks for explanation, not sure if i fully understand it

PR #3974, opened 2024-05-21
run_without_change_log_updates compilation fix

lache-melvin:
main was merged back into develop - no conflicts but this param wasn't caught 😅


roxy-dao (approved):
Thanks for catching this!

PR #3977, opened 2024-05-22
Disable inserting/updating inbound lines when invoice is on hold

### LINKED ISSUE #3526: Why can't we hold the shipments from stock-transfer?

clemens-msupply:
Fixes #3526

# 👩🏻‍💻 What does this PR do?

Hotfix for an (old) issues that has been revealed by fixing 3526.

- Add backend on_hold check when inserting and updating inbound lines
- Add fe check to disable UI when invoice is on hold

# 🧪 Testing
- Create inbound shipment
- set on hold and try to edit/ update lines

### DIFF
--- a/client/packages/invoices/src/utils.ts
+++ b/client/packages/invoices/src/utils.ts
@@ -129,6 +129,9 @@ export const isOutboundDisabled = (outbound: OutboundRowFragment): boolean => {
 };
 
 export const isInboundDisabled = (inbound: InboundRowFragment): boolean => {
+  if (inbound.onHold) {
+    return true;
+  }
   const isManuallyCreated = !inbound.linkedShipment?.id;
   return isManuallyCreated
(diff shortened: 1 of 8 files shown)


clemens-msupply:
After discussion with @roxy-dao and @mark-prins, we want to keep editing an invoice. On hold just prevent others from doing status changes.

PR #3997, opened 2024-05-23
2154 Linter rule for const and warning for indirect external package imports

### LINKED ISSUE #2154: Linter rule for const and warning for indirect external package imports
I keep on messing up `const` and `lets` when transitioning between rust and ts, this causes extra work during review (to correct it). Maybe we can use https://eslint.org/docs/latest/rules/prefer-const: https://github.com/openmsupply/open-msupply/pull/2149#discussion_r1309378363

We also sometimes import directly from external package rather then re-export in common package (maybe we can get a compilation warning for this somehow, since those external packages are in common), example PR comment: https://github.com/openmsupply/open-msupply/pull/2149#discussion_r1309381291

roxy-dao:
Part of #2154

# 👩🏻‍💻 What does this PR do?
Wanted to get this in since very guilty of doing this when switching from backend to frontend 👀 
Pending: Imports - not working so will look into it later when more time

# 🧪 Testing

- [ ] 

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -28,6 +28,10 @@
     "strictPropertyInitialization": true,
     "noImplicitThis": true,
     "alwaysStrict": true,
+    "prefer-const": ["error", {
+      "destructuring": "any",
+      "ignoreReadBeforeAssign": true
+    }],
 
     "noUnusedLocals": true,
     "noUnusedParameters": true,


clemens-msupply (approved):
Looks good to me previous this was only a warning, right?

roxy-dao:
> Looks good to me previous this was only a warning, right?

No it didn't show any warning previously 👀

PR #3998, opened 2024-05-24
3905 facilities list

### LINKED ISSUE #3905: GAPS: Add a page to central server which lists all facilities
Part of epic #3660 

## Acceptance criteria
* [ ] On central server only - a list of all stores and facilities in the name table (`type`)
* [ ] The list shows the name of the facility (from the name table) 
* [ ] The list shows the code of the facility
* [ ] A graphQL endpoint is available (only when viewed from central server) which lists all stores

Create a graphQL query to list all names
ListView for names

- Name, Code, Type, Booleans for isStore, isDonor, isCustomer, isPurchaser, 

New Nav Item Under New `Admin` section (`Admin` to be renamed to Settings?)

### Moneyworks Jobcode 🧰
`CCEIDEV`

lache-melvin:
Fixes #3905

# 👩🏻‍💻 What does this PR do?

Adds facilities list view when logged in as central server:

Screenshot 2024-05-24 at 12 01 56 PM

- Went with `Manage` rather than `Admin` for now, so as not to have to make changes to existing `Admin` piece for now
- Turns out there was no permissions restriction on which names you can see, so just reusing existing names query
  - Is it worth raising an issue for the fact that anyone can query patient data?
- Existing pattern is to show the home icon if the name is a store... should I stick with that or have an isStore column?

# 🧪 Testing

- Log into central server
- [ ] There is a new `Manage` section in the Nav, with a `Facilities` page
- [ ] Facilities page lists all stores and facilities
- [ ] You can search by code
- [ ] You can search by name
- Log into remote site
- [ ] There is no `Manage` nav section

# 📃 Documentation

  1. Screenshots of `Settings` in the app bar/Administration pages - rename to Settings
  2.


jmbrunskill (approved):
Looks great.
Probably should have a discussion about the icon and location for the `Manage` section but I this is good where it is for now.

jmbrunskill:
> * Went with `Manage` rather than `Admin` for now, so as not to have to make changes to existing `Admin` piece for now

Looks fine to me, although would be nice to have a different icon and maybe rename the other one settings?

> * Turns out there was no permissions restriction on which names you can see, so just reusing existing names query
>   * Is it worth raising an issue for the fact that anyone can query patient data?

Yes please!

> * Existing pattern is to show the home icon if the name is a store... should I stick with that or have an isStore column?
I like it, seems good to me.
At some point we might want to add a filter for store, donor, etc but I think it's totally fine for this PR.

mark-prins:
oh! I was going to start a discussion to get feedback on adding the management section. it's already here!
two quick comments: all the other menu items are nouns, and 'manage' is a verb and this troubles me.. to be consistent it should be 'Management' and I'm really not sure that this works.
also the icon is the same as the admin/settings and that is potentially confusing - we need an icon that says 'manage' without being a 'manager' 🤔

client/packages/system/src/Name/api/hooks/document/useNames.ts
```diff
@@ -12,7 +12,7 @@ export const useNames = (type: 'customer' | 'supplier') => {
         first: queryParams.first,
         offset: queryParams.offset,
         sortBy: queryParams.sortBy,
-        type: type === 'customer' ? 'customer' : 'supplier',
+        type: type,
```
roxy-dao, line 15: Nit: Just type~?

client/packages/common/src/intl/locales/en/manage.json
```diff
@@ -0,0 +1,3 @@
+{
+  "placeholder.enter-an-item-code-or-name": "Enter item code or name"
```
mark-prins, line 2: now that this is in two namespaces, putting in `common.json` would be better

client/packages/host/src/routers/ManageRouter.tsx
```diff
@@ -0,0 +1,23 @@
...
+  .addPart(AppRoute.Facilities)
+  .addWildCard()
+  .build();
+
+export const ManageRouter: FC = () => {
+  const gotoCustomers = useMatch(fullFacilitiesPath);
+
+  if (gotoCustomers) {
+    return <NameService />;
+  }
```
mark-prins, line 19: :nit
```suggestion
export const ManageRouter: FC = () => {
  const gotoFacilities = useMatch(fullFacilitiesPath);

  if (gotoFacilities) {
    return <NameService />;
  }
```

should that be `gotoNames` or `<FacilityService />` 🤔

client/packages/common/src/intl/locales/en/manage.json
mark-prins: although.. I see now that it is `Enter facility code or name` really.

server/repository/src/db_diesel/name.rs
```diff
@@ -45,6 +47,8 @@ pub struct NameFilter {
     pub address2: Option<StringFilter>,
     pub country: Option<StringFilter>,
     pub email: Option<StringFilter>,
+
+    pub code_or_name: Option<StringFilter>,
```
mark-prins, line 51: I can sense the request coming in to have also the boolean filters for supplier/customer/donor.

mark-prins (approved):
Great - thanks! the verdict is hazy re naming and even positioning. 
My summary - having a different icon is good. I think the sliders are not intuitive but they are better than nothing. The name can be changed later, and while I don't like 'Manage' in context, it appears to be the preferred. 

Here's [the branch](https://github.com/msupply-foundation/open-msupply/compare/3905-facilities-list...3905-facilities-list(suggestion)) which I had the icon in, in case it helps

Side note: I have some funky stuff in my datafile..

Screenshot 2024-05-27 at 3 22 16 PM

lache-melvin:
> Side note: I have some funky stuff in my datafile..
> 
> Screenshot 2024-05-27 at 3 22 16 PM

Oh... is this just from test data or should there be some more filtering??

mark-prins:
> Oh... is this just from test data or should there be some more filtering??

I was wondering if we needed filtering. it's not deliberate test data - but there are a lot of blanks in the table.. surely we don't need to anticipate that happening?

lache-melvin:
> I was wondering if we needed filtering. it's not deliberate test data - but there are a lot of blanks in the table.. surely we don't need to anticipate that happening?

Probably from integration tests right? Doesn't _seem_ like the kind of thing we should expect to need to handle in the wild? 🤷‍♀️

mark-prins:
could well be integration tests. they did populate my datafile with a lot of crud.

lache-melvin:
FYI leaving the command K settings shortcut as `a` for now despite the rename to settings... we already have a bunch of `s` related shortcuts, so I'm not sure what would be best to change it to, if at all?

adamdewey:
I can see non-Store suppliers in the list which I think shouldn't be there?

image

adamdewey:
Actually, I'll raise a separate issue for it

PR #4004, opened 2024-05-26
Move prefer const to eslintrc

jmbrunskill:
Fixes `../../tsconfig.json(31,5): error TS5023: Unknown compiler option 'prefer-const’`

# 👩🏻‍💻 What does this PR do?

It looks like this should have been an eslint rule not a compiler option?

Does this still achieve the original goal?

DEV ONLY (no docs or testing required)


roxy-dao (approved):
Thanks James! 

Sorry... ;-;

PR #4007, opened 2024-05-27
4000 ignore soft deleted master lists

### LINKED ISSUE #4000: Ignore soft deleted master lists
## What went wrong? 😲
Current integration error for soft deleted master list says "Translator not found". Soft deletes is fairly new in mSupply, so we should recreate the record in OMS with is_active = false to keep any FK links that might exist.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #4000

# 👩🏻‍💻 What does this PR do?
OMS deals with program master lists differently from OG and has a separate table for it.

Save deleted program master list records from OG since OMS program requisitions was rolled out before soft delete was implemented for master lists in OG. 

# 🧪 Testing


server/service/src/sync/translations/invoice.rs
```diff
@@ -650,8 +650,7 @@ fn invoice_status(invoice_type: &InvoiceType, data: &LegacyTransactRow) -> Optio
             LegacyTransactStatus::Fn => InvoiceStatus::Verified,
             _ => return None,
         },
-        // mSupply will alert users to finalise any repacks if they have un-finalised repacks
-        // before they migrate to Open mSupply.
+        // mSupply will delete any unfinalised repacks before migration
```
roxy-dao, line 653: Change comment since OG is now deleting unfinalised repacks instead of alerting user

server/service/src/sync/translations/master_list.rs
```diff
@@ -47,6 +49,47 @@ impl SyncTranslation for MasterListTranslation {
...
+                name: "missing_program".to_string(),
+                code: "".to_string(),
+                // program name is the master list description
+                description: program.name.clone(),
+                is_active: false,
+            };
+            return Ok(PullTranslateResult::upsert(result));
+        }
+
+        if let Some(master_list) = master_list {
```
roxy-dao, line 78: Put this in for master lists we already have that were deleted before the soft delete was implemented in OG, but maybe can just hard delete them since they aren't linked to any requisition programs?

server/service/src/sync/translations/master_list.rs
```diff
@@ -47,6 +49,35 @@ impl SyncTranslation for MasterListTranslation {
...
+                id: master_list.id.clone(),
+                name: master_list.name.clone(),
+                code: master_list.code.clone(),
+                description: master_list.description.clone(),
+                is_active: false,
+            };
+            return Ok(PullTranslateResult::upsert(result));
+        }
+
+        Ok(PullTranslateResult::Ignored(
```
roxy-dao, line 77: We did have delete for master list before... but can't remember if it was before or after soft deletes were implemented in mSupply.... should I account for this? 🤔

clemens-msupply:
@Chris-Petty are DELETE items kept in the mSupply OG sync queue indefinitely?  For example, what happens when we init a new omSupply site? Does the deleted master_list record still get synced? If not we would miss that the master_list got deleted, and we still get foreign key constraint violations.

server/service/src/sync/translations/master_list.rs
clemens-msupply: Why do you look for the program row here? Shouldn't you always soft delete the master_list row even if there isn't a program associate with it?

roxy-dao: master_list row isn't a foreign key on anything just the programs (that's also a master list but a separate one) ...if I recall correctly?

clemens-msupply: I guess it can only be deleted in mSupply when its not used anywhere?

Or only softdelete the masterlist if a program row uses it otherwise do a hard delete? Think, this would also make sense to me when reading the code without having too much background.

Btw. the comment "Recreate these records to keep the FK link." is also misleading since you are not recreating anything...

roxy-dao: Before soft deletes was implemented a few months after OMS programs, it was all hard delete. mSupply doesn't really handle foreign keys, so it isn't a problem there. 

Yeah... Can't recreate since the delete record only has ID...

server/service/src/sync/translations/master_list.rs
```diff
@@ -47,6 +49,35 @@ impl SyncTranslation for MasterListTranslation {
...
+        let master_list =
+            MasterListRowRepository::new(connection).find_one_by_id(&sync_record.record_id)?;
+
+        if let (Some(_), Some(master_list)) = (program, master_list) {
+            let result = MasterListRow {
+                id: master_list.id.clone(),
+                name: master_list.name.clone(),
+                code: master_list.code.clone(),
+                description: master_list.description.clone(),
+                is_active: false,
```
clemens-msupply, line 72: Nit: don't need to clone here and could use `..master_list` syntax to take all existing fields...

clemens-msupply (changes requested):
Here are my change requests:
- update the "Recreate..." comment
- Always soft-delete the master list (or at least hard delete the master list if it wasn't in use?) Currently, nothing is done when there is no program which seems wrong to me, i.e. an existing master_list doesn't get deleted if there isn't a program...

roxy-dao:
> Here are my change requests:
> 
> * update the "Recreate..." comment
> * Always soft-delete the master list (or at least hard delete the master list if it wasn't in use?) Currently, nothing is done when there is no program which seems wrong to me, i.e. an existing master_list doesn't get deleted if there isn't a program...

I've hard deleted them since future master lists should all be soft deleted regardless of if they are program or not 🤔

clemens-msupply (approved):
Looks good, thank you!

PR #4008, opened 2024-05-27
3656 (2) Stock page -- repack and adjustment functionality

### LINKED ISSUE #3656: Move `Stock line` modal to its own page
## The suggestion

Stock Line has had more functionality added to it and the modal is now overcrowded with tabs. 

It would be nice if clicking on a stock line will redirect the user to the stock line page instead of opening up a modal. This would also allow the repack modal to be refactored as a tab in the stock line window.

This can be done in two parts: 
1. Refactor the modal to its own page
2. Refactor `repacks` to be a tab in the `Stock line` page

Things will looks less crowded and makes everything more mobile friendly (especially when one wants to do a repack! see #3439)

Probably not, it's the same thing we do for `Items`, might have to re-train users but should be intuitive enough...

Would need quite a bit of effort and maybe some UX input

CarlosNZ:
Fixes #3656 (Part 2)

# 👩🏻‍💻 What does this PR do?

- Make the "Repack" and "Adjust" buttons functional, using existing components
- Refactor hooks for `useRepack` into the newer pattern

See inline comments. I plan to do one more PR with the remaining Stock API hooks refactored so this whole area will be in the new style. But I thought it better to make a new PR, since it's not strictly within the purview of this issue.

I tried to re-implement the plugin functionality  in the Stock Edit form (Detail View), but I don't really have a way of properly testing this.

# 🧪 Testing

- Make sure "Repack" and "Inventory Adjustment" functionality works as expected


client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
```diff
@@ -0,0 +1,148 @@
+import React, { FC } from 'react';
```
CarlosNZ, line 1: This file is just renamed from "InventoryAdjustmentForm", only minimal code changes in here.

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }
+  };
+
+  return (
+    <Modal
+      sx={{ maxWidth: 'unset', minWidth: 700, minHeight: 575 }}
+      slideAnimation={false}
+      title={t('title.stock-line-details')}
+      okButton={
+        <DialogButton variant="ok" disabled={saveDisabled} onClick={save} />
```
CarlosNZ, line 58: Do you think we need a confirmation for saving here? We didn't have one previously, so I didn't put one in, but just wondering.

client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
```diff
@@ -215,14 +191,14 @@ export const RepackModal: FC<RepackModalControlProps> = ({
             <Typography>{t('messages.no-repack-detail')}</Typography>
           </Box>
         )}
-        <Box display="flex" flexDirection="column" height={435}>
```
CarlosNZ, line 218: "height" prop causes unnecessary vertical scrollbar

client/packages/system/src/Stock/DetailView/DetailView.tsx
```diff
@@ -30,36 +34,54 @@ export const StockLineDetailView: React.FC = () => {
...
     urlQuery: { tab },
   } = useUrlQuery();
   const { success, error } = useNotification();
   const t = useTranslation('inventory');
   const { setSuffix } = useBreadcrumbs();
 
+  const repackModalController = useToggle();
+  const adjustmentModalController = useToggle();
+
+  const [hasPluginChanged, setHasPluginChanged] = useState(false);
```
CarlosNZ, line 49: So that "hasChanged" you asked about in the last PR (@lache-melvin ) -- turns out it was for the plugins. Looks like any plugins need to maintain their own "hasChanged" state in addition to the draft `isDirty`. I've renamed it to make it clearer what it belongs to.

client/packages/system/src/Stock/api/hooks/useRepack.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export const useRepackEdit = (init: Repack) => {
+  const { stockApi, storeId, queryClient } = useStockGraphQL();
+  const [repack, setRepack] = useState<Repack>({ ...init });
+  const stockLineId = repack.stockLineId ?? '';
+
+  // FETCH
+  const queryFn = async () => {
+    const result = await stockApi.repacksByStockLine({
```
lache-melvin, line 37: If this is an intended pattern, it might be nice to add some kind of context comment here... how come individual repack line queried in a separate hook, but the repacks list is queried in an edit hook? Intuitively I would have expected the opposite?

Actually wondering why they can't all be on the same hook? (Relevant calls only enabled if when their ids are provided anyway...)

client/packages/system/src/Stock/DetailView/DetailView.tsx
```diff
@@ -99,27 +114,33 @@ export const StockLineDetailView: React.FC = () => {
...
-
-  //   return () => removeEventListener(listener);
-  // }, [addEventListener, removeEventListener]);
-
   const footerProps = {
     isSaving: isUpdating,
     showSaveConfirmation,
     showCancelConfirmation,
-    isDirty,
+    disabled: isDirty && !hasPluginChanged,
```
lache-melvin, line 121:
```suggestion
    disabled: !isDirty && !hasPluginChanged,
```
Only disable if both haven't changed?

client/packages/system/src/Stock/DetailView/AppBarButtons.tsx
```diff
@@ -24,17 +27,15 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({}) => {
           label={t('button.repack')}
           Icon={<StockIcon />}
           // TO-DO: Add repack modal
```
lache-melvin, line 29: remove todos now?

lache-melvin (approved):
Awesome thanks Carl! Looks great 😁 

Made a few small comments but merge when ready!

As far as plugins go...
- I wonder if Mark has an example for a stock edit plugin?
- Are plugins part of the testers test suite? ...Are many of our clients using them?

lache-melvin:
> * I wonder if Mark has an example for a stock edit plugin?
> * Are plugins part of the testers test suite? ...Are many of our clients using them?

Apparently no one is using front end plugins atm - code _looks_ like its doing the same thing, so I think fine to leave and if/when they are used, we'll add them into test suites 🙏

mark-prins:
> I wonder if Mark has an example for a stock edit plugin?

Yes, the examples are in [this repo](https://github.com/msupply-foundation/open-msupply-plugins)

lache-melvin (changes requested):
Hi sorry realise should have put changes requested, just for:
- Save button is disabled when it should be enabled
- useRepack - list query should be separate, edit and single repack in same hook (same as you have for useStockLine)

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
CarlosNZ: Yeah, I figured opening the modal is an intentional enough act that you're not really going to do it by accident.

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+      }
+      cancelButton={<DialogButton variant="cancel" onClick={onClose} />}
+    >
+      <Box display="flex">
```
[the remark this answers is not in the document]
CarlosNZ: Ugh, this is annoying -- the form is actually centred correctly, it just looks off because all the columns are right justified. (If there was a long enough label on the left it would look "correct"). 

I've added a bit more padding on the right side only (and slightly increased the gap between columns), which helps, but it's a bit of a hack. 

68ff74c06

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+    } catch {
+      // TODO: handle error if no reason selected when reasons required
+    }
+  };
+
+  return (
+    <Modal
+      sx={{ maxWidth: 'unset', minWidth: 700, minHeight: 575 }}
+      slideAnimation={false}
+      title={t('title.stock-line-details')}
```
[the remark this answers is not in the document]
CarlosNZ: 68ff74c06

client/packages/system/src/Stock/api/hooks/useRepack.ts
CarlosNZ: Yeah, that's right, they should really be around the other way. I think it ended up like this cos I was only refactoring one use case in this PR and that's how they were organised in the implementations. 

I think I'll leave this to the [other PR](#4009) as I've already made some additional changes in there.

client/packages/system/src/Stock/DetailView/DetailView.tsx
CarlosNZ: Oops -- yes 68ff74c06

client/packages/system/src/Stock/DetailView/AppBarButtons.tsx
CarlosNZ: b44c51c74

client/packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -239,7 +239,7 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
                 <div>{slideConfig.in && children}</div>
               </Slide>
             ) : (
-              <div>{children}</div>
```
lache-melvin, line 242: Why was this removed? I'm all for removing excessive divs but pretty nervous about this one, we use this hook everywhere and wouldn't be surprised if some modal is relying on it for layout?

client/packages/system/src/Stock/Components/InventoryAdjustment/InventoryAdjustmentModal.tsx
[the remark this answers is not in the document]
lache-melvin: Argh yeah sorry about that... would an alternative solution be to narrow the modal? There aren't other tabs constraining us to this width anymore 👀

client/packages/common/src/hooks/useDialog/useDialog.tsx
CarlosNZ: I'm on a mission to hunt down and kill spurious `<div>`s. (Numerous layout bugs have been caused by them)

I'm 90% sure this does absolutely nothing (no styling, or React component), but I'll check the implementations of `useDialog` just to be sure (all the children should have their own top-most structural element).

The only way it could have any effect is if the top-most container of the `children` is an `inline` element rather than a `block` element, but if that was the case I'd rather fix that case than have redundant block containers here.

Will report back shortly.

CarlosNZ: I'll also check the "Slide" case above and remove that div if it's not doing anything.

(Sorry for polluting the PR, I just got frustrated trying to inspect the centering layout of the modal content, and came across this redundant <div> in the HTML)

lache-melvin (approved):
Cool - if confident no regressions in the `useDialog` then I'm happy to merge... if it's looking like it's touching more stuff/is taking too long I'd ask we back out and just merge the stock page so we can get that build done, save the div removal mission for a separate PR 🙏

client/packages/common/src/hooks/useDialog/useDialog.tsx
CarlosNZ: Screenshot 2024-06-04 at 3 43 58 PM
Screenshot 2024-06-04 at 3 48 41 PM

Yeah, the `DialogContent` component already renders a `div`, so we end up with `<div><div>{content}</div></div>`, which is pretty sloppy web development 😉.

*However*... the one inside the Slide component does actually seem to be doing something -- seems to embed some CSS classes on the inner div to create the slide animation, and throws and error when missing. WHich is wierd, I'm not sure why it would just target the next top-most block inside it. But in light of that, I'll leave the other div in place now for consistency. (Something to look into another day)

da2af02d5

PR #4009, opened 2024-05-27
3656 (3) Finish Stock API refactor

### LINKED ISSUE #3656: Move `Stock line` modal to its own page

CarlosNZ:
Fixes #3656

# 👩🏻‍💻 What does this PR do?

~~*Putting this in Draft initially, as I haven't really thoroughly tested it, it just doesn't give compilation errors*~~

To-DO.

# 🧪 Testing


lache-melvin (approved):
Once import is fixed looks good 🚀 

Maybe another refactor but could be nice to pull out shared draft state stuff into a reusable hook (can return the patch, setPatch, isDirty...)?

client/packages/system/src/Stock/Components/Repack/RepackEditForm.tsx
```diff
@@ -6,42 +6,41 @@ import {
...
   draft: Repack;
+  repackData?: RepackNode;
 }
 
 export const RepackEditForm: FC<RepackEditFormProps> = ({
-  invoiceId,
   onChange,
   stockLine,
   draft,
+  repackData,
```
lache-melvin, line 32: Seeing as this is just a form component, do you think we could only pass in a draft? And let the calling component decide whether the that draft is the `draft` from the edit hook or `repackData.from`?

CarlosNZ: Yeah, good idea. Will do.

CarlosNZ: b2c78a562 :)

client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
```diff
@@ -211,10 +208,21 @@ export const RepackModal: FC<RepackModalControlProps> = ({
...
+                  packSize: draft?.packSize ?? stockLine.packSize,
+                  newPackSize: draft?.newPackSize,
+                  numberOfPacks:
+                    draft?.numberOfPacks ?? repackData?.to?.numberOfPacks,
+                  newLocationId: draft?.newLocationId,
+                  availableNumberOfPacks:
+                    draft?.availableNumberOfPacks ??
+                    stockLine?.availableNumberOfPacks,
+                  locationName:
+                    draft?.locationName ?? repackData?.to?.location?.name,
```
lache-melvin, line 223: Oh hm, I didn't realise how the different inputs mapped 😅 

Also looks like data for viewing existing repacks isn't mapping quite right...
Screenshot 2024-06-07 at 4 40 51 PM

I had a play here, wonder what you think? https://github.com/msupply-foundation/open-msupply/compare/3656-(3)-finish-stock-API-refactor...3656-(3)-finish-stock-API-refactor(form-data)

lache-melvin (changes requested):
Need to fix viewing historic repacks before merging 😅

client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
CarlosNZ: That's great, thanks.

CarlosNZ:
@lache-melvin Thanks for the suggested fix, works a treat. 

Should be all okay now, I think?

lache-melvin (approved):
Awesome, thanks Carl - let's get it in! 🚀

PR #4039, opened 2024-06-05
Patch release v.2.0.3

CarlosNZ:
Bump version for patch release

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.0.02",
+  "version": "2.0.3",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
Thanks!

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.0.02",
+  "version": "2.0.3",
```
roxy-dao, line 4:
```suggestion
  "version": "2.0.03",
```

CarlosNZ: Done --but why do we do this? It's not standard semantic versioning 🤷‍♂️
Screenshot 2024-06-05 at 2 30 17 PM

roxy-dao: actually.. was just following pattern but now you mentioned it it should be .1, .2, .3 not sure why the backend tests compare 0.1 > 0.01...

PR #4041, opened 2024-06-05
Fix Postgres test warnings

clemens-msupply:
Fix Postgres test warnings


jmbrunskill (approved):
Haha was just raising a PR for the same change.

PR #4049, opened 2024-06-05
Remove unused endpoint

### LINKED ISSUE #1942: Remove unused update_document graphql endpoint
Document mutations are currently done through the mutation endpoints for encounters and enrolments.

clemens-msupply:
Fixes #1942

Checked in the import script and in the mSupply client code and this endpoint wasn't used anywhere (think it was at some point) So I removed it now.


jmbrunskill (approved):
Slightly scared about removing something when we're not sure why it was added.
The API Design is to never update a document via a mutation directly right?
Looks like it was already blocking access to patient etc via this endpoint so I think we're good to remove it as core functionality won't be broken...

clemens-msupply:
> Slightly scared about removing something when we're not sure why it was added. The API Design is to never update a document via a mutation directly right? Looks like it was already blocking access to patient etc via this endpoint so I think we're good to remove it as core functionality won't be broken...

It was initially used to mutate a document, but now all mutations are done through the patient, program_enrolment and encounter endpoints. These endpoints may trigger some extra behaviour and thus the low level `update_document` endpoint shouldn't be used. I checked the import script and the mSupply UI, and it wasn't used.

PR #4050, opened 2024-06-06
swallow keep awake errors

### LINKED ISSUE #4048: Can't type in initialisation screen on android
## What went wrong? 😲

image

Initialise form is disabled!

## How to Reproduce 🔨

- Download APK v2.1.0 (GDrive)
- Start in server mode
- You can't enter anything in the form, no info as to why

## Your environment 🌱

- Open mSupply Version: 2.1

- Platform: Android

- Database type:

Additional: Looks like this is due to KeepAlive plugin not being available on initial load. Which sounds like maybe a capacitor problem, but it works fine on 2.0?

lache-melvin:
Fixes #4048 

# 👩🏻‍💻 What does this PR do?

Swallows KeepAwake errors if they arise, so initialisation screen can finish loading.

On initial load, KeepAwake was throwing a "not available" error, and so from the initialise hook, setIsInitialising(false) was never being called.

The form is disabled until isInitialising is set to false, so the form was never being enabled.

Seeing as the KeepAwake piece is helpful, but not critical to any workflow, I think the best solution is just to swallow the error and move on.

You can now type as expected:
Screenshot 2024-06-06 at 12 17 36 PM

TBH I have no idea why keep awake is not available on initial load. It's available after refresh (i.e. I refreshed with dev tools, then everything worked fine), but you can't just refresh easily in android app.

It's working fine in main/2.0?? Wonder if there was a dependency upgrade somewhere, but I can't pinpoint right now and feel I've spent enough time on this already...

Probably worth investigating at some point? But I'd like to get this fix to the testers for Inv Adj testing 🙏 

# 🧪 Testing

- [ ] Build APK
- [ ] You can type in the inputs

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/hooks/useNativeClient/useNativeClient.ts
+++ b/client/packages/common/src/hooks/useNativeClient/useNativeClient.ts
@@ -117,16 +117,22 @@ export const useNativeClient = ({
     // Currently only supported on native platforms via capacitor
     if (!Capacitor.isNativePlatform()) return;
 
-    const result = await KeepAwake.isSupported();
-    if (result.isSupported) await KeepAwake.allowSleep();
+    try {
+      const result = await KeepAwake.isSupported();
+      if (result.isSupported) await KeepAwake.allowSleep();
+      // If KeepAwake has errors, just swallow them
+    } catch {}
   };
 
   const keepAwake = async () => {
     // Currently only supported on native platforms via capacitor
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Maybe worth creating a follow up issue so we don't just have broken `keepawake` for ever?

PR #4051, opened 2024-06-06
3913 gaps immunisation programs backend

### LINKED ISSUE #3913: GAPS: Immunisation Programs Backend
Part of epic #3662 

For GAPS we need to track multiple vaccine course under a program.
Rather than creating a new program table for this purpose, we'll reuse the existing program table.
Normally this is created automatically from a master list, and programs required a master list previously.

We'll make the `master_list` optional on a program record with this PR.

I'm assuming we can get away with not syncing these vaccine programs as they're only needed as part of GAPS at this stage.

* [x] Add a new `is_immunisation` field to the programs table.
* [x] Make master_list optional
* [x] graphQL mutation to upsert a program (marked as immunisation)
* [x] graphQL query to return is_immunisation programs
* [x] No Sync

### Moneyworks Jobcode 🧰
`CCEIDEV`

jmbrunskill:
Fixes #3913

# 👩🏻‍💻 What does this PR do?

Graphql and Backend to create and query ImmunisationPrograms.

Immunisation Programs are created as programs in open-mSupply Central server (only)
They don't have an associated master list (so this PR also removes the need for masters lists on each program)

I've created a `ProgramService` but went very specific with `insertImmunisationProgram` request in graphql.
If we start adding different kinds of programs we might want something more generic, but I don't want the API to allow creating something that doesn't like to masterlist but should...

# 🧪 Testing

- [ ] try creating and querying a program from graphql

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_01_00/program.rs
```diff
@@ -0,0 +1,39 @@
...
+            CREATE TABLE tmp_program (
+                id TEXT NOT NULL PRIMARY KEY,
+                master_list_id TEXT,
+                name TEXT NOT NULL,
+                context_id TEXT NOT NULL,
+                is_immunisation BOOLEAN NOT NULL
+            );
+            INSERT INTO tmp_program SELECT id, master_list_id, name, context_id, false FROM program;
+            DROP TABLE program;
+            CREATE TABLE program (
```
jmbrunskill, line 25: I avoided using rename, after the bug we'd seen with SQLite before.
Might be ok now that we're using diesel 2, but didn't want to risk it...

fergie-nz (approved):
Nice! looks good

PR #4052, opened 2024-06-06
Use uuid v7 instead of v4

### LINKED ISSUE #4038: Use UUIDv7 instead of v4
## The suggestion

Since uuid7 produces time sortable ids, using uuid7 as primary key should result in better DB performance. For example, it's much more likely that all stocktake lines of a newly created stocktake are in consecutive blocks on disk while with uuid v4 rows are scattered all over the disk...

Please see https://uuid7.com/ for details.

@jmbrunskill 

We already abstract over the used uuid method. So very easy to change.

clemens-msupply:
Using uuid v7 improves write and improves read performance for sequential reads.

Fixes #4038

### DIFF
--- a/client/packages/common/package.json
+++ b/client/packages/common/package.json
@@ -51,7 +51,7 @@
     "recharts": "^2.5.0",
     "stylis": "^4.0.13",
     "stylis-plugin-rtl": "^2.1.1",
-    "uuid": "^9.0.0",
+    "uuidv7": "^1.0.0",
     "zustand": "^4.3.6"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


jmbrunskill (commented):
Aren't most of our UUIDs generated from the frontend?
Does this exist for typescript? I assume it does?

clemens-msupply:
> Aren't most of our UUIDs generated from the frontend? Does this exist for typescript? I assume it does?

ah yes you are right, would have to do it on the frontend

clemens-msupply:
Haha merged 3 days ago: https://github.com/uuidjs/uuid/pull/681 but last release was 9 month ago...

clemens-msupply:
I had a look at the uuidv7 nodejs package https://www.npmjs.com/package/uuidv7. Also had a brief browse through the repo and it looks good to me...

jmbrunskill (approved):
Looks good to me...

PR #4053, opened 2024-06-06
Feature/iga

jmbrunskill:
Moves GAPS (nee IGA) features into develop.
There's a feature flag for GAPS so this should be safe...

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

# 📃 Documentation

  1.
  2.


client/package.json
```diff
@@ -5,9 +5,8 @@
   "sideEffects": false,
   "private": true,
   "scripts": {
-    "start-local-features": "yarn start-local -- -- --env $1",
```
lache-melvin, line 8: If we're removing this script, should also update readme

client/packages/host/src/components/Navigation/CatalogueNav.tsx
```diff
@@ -49,6 +52,17 @@ export const CatalogueNav: FC = () => {
               .build()}
             text={t('master-lists')}
           />
+          <AppNavLink
+            visible={
+              // isCentralServer && Environment.FEATURE_DEMOGRAPHIC_INDICATORS
```
lache-melvin, line 57: should be uncommented?

client/packages/system/src/Item/context/usePackVariant.ts
```diff
@@ -79,6 +79,10 @@ export const useIsPackVariantsEnabled = () => {
   // return Object.keys(items).length > 0;
 };
 
+export const useIsDemographicsEnabled = () => {
```
lache-melvin, line 82: why is this in usePackVariant?

lache-melvin (approved):
Yeah I think should be fine to get into develop, we've got a couple weeks to iron anything out if need be! Haven't run it - but I think the nav link should be behind a feature flag?

Do you want to consolidate this with the FEATURE_GAPS feature flag or leave that to another day?

client/package.json
jmbrunskill: Good catch, I wonder how/why that's being removed?

client/packages/system/src/Item/context/usePackVariant.ts
jmbrunskill: I've removed it, it's not used.

client/packages/host/src/components/Navigation/CatalogueNav.tsx
jmbrunskill: Done, and changed to FEATURE_GAPS

client/package.json
mark-prins: removed because the script list was getting bloated, with the addition of a `start-remote-features` as well.
I had a look, and decided that `start-local` can be adjusted to suit both modes, no params passed and feature params passed; so there's no need to have both `start-local` and `start-local-features`

jmbrunskill: Looks like it was a suggestion from Mark...

jmbrunskill: I've updated the README, and will let the team know.

PR #4081, opened 2024-06-09
Bug fix postgres tests

jmbrunskill:
# 👩🏻‍💻 What does this PR do?

Fixes postgres tests with the immunisation program work

Went with a single context for now, for all immunisations.
We could add something in the future around parent programs in context but for now keeping it simple.

# 🧪 Testing

`cargo test --features=postgres`

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/migrations/v2_01_00/activity_log.rs
+++ b/server/repository/src/migrations/v2_01_00/activity_log.rs
@@ -7,6 +7,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
             r#"
                 ALTER TYPE activity_log_type ADD VALUE 'ASSET_PROPERTY_CREATED';
                 ALTER TYPE activity_log_type ADD VALUE 'VACCINE_COURSE_CREATED';
+                ALTER TYPE activity_log_type ADD VALUE 'PROGRAM_CREATED';
             "#
         )?;
     }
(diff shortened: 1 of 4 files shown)


server/repository/src/migrations/v2_01_00/activity_log.rs
```diff
@@ -7,6 +7,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
             r#"
                 ALTER TYPE activity_log_type ADD VALUE 'ASSET_PROPERTY_CREATED';
                 ALTER TYPE activity_log_type ADD VALUE 'VACCINE_COURSE_CREATED';
+                ALTER TYPE activity_log_type ADD VALUE 'PROGRAM_CREATED';
```
clemens-msupply, line 10: Are they used anywhere? can't find any references to the string values in the code, or how does it work?

clemens-msupply (approved):
Looks good, just one question...

server/repository/src/migrations/v2_01_00/activity_log.rs
jmbrunskill: It's generated by the enum with SCREAMING_SNAKE_CASE...
Here:https://github.com/msupply-foundation/open-msupply/blob/34feb1426e4484af539f72c1adcabb73b58c151e/server/repository/src/db_diesel/activity_log_row.rs#L29
Would be nice if we could use enums without Postgres Needing this. I feel like we should be able to do a mapping to TEXT Field in diesel...

clemens-msupply: Ah yes right, should have known...

PR #4088, opened 2024-06-09
Disable centralPatientSearch query if the input params object is empty

### LINKED ISSUE #4087: When searching only by Code 2 unrelated entries are returned
## What went wrong? 😲
When searching by NUIC: "GRKIL012NE" the central server search endpoint returns many (all?) unrelated results:

image

This is because an empty search object is used to query the central server.

clemens-msupply:
Fixes #4087 

# 👩🏻‍💻 What does this PR do?
Checks if the input object is empty, i.e. has no keys like `first: undefined`.

Could have also added the check on the server but was easier to do in js...

# 🧪 Testing

Think it requires the png programs module to be configure to be able to select the NUIC.

- Search only by NUIC or now input params at all
- The returned list should not include central server results


roxy-dao (approved):
Thanks Clemens~

Haven't tested but code looks good 👍🏻

Issue #3916, opened 2024-05-16
GAPS: vaccine course front end: edit


mark-prins:
Part of epic #3662 

## Acceptance criteria

* [ ] Coverage Rate (Numeric)
* [ ] Wastage Rate (Numeric)
* [ ] Number of doses (dose names and labels not required for this PR)

## Describe the solution you'd like 🎁
The 'create immunisation program' button is available top right. Required data must be completed before OK is enabled
After creating, show a success toast ( or an error toast on failure )
After creating, the list is refreshed
There's no delete action on the wireframe, make sure that one is added and then hooked up
the spec says 'one or more vaccines' so the wireframe needs work: likely use a datatable with the delete checkbox / action
Prompt to warn people before deleting a program

![image](https://github.com/msupply-foundation/open-msupply/assets/8287373/8ec89911-31a3-4616-ba2c-0610d69845d1)

### Additional context 💌

| User story  | B.2.1 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to be able to detail the key requirements of each individual immunization program
So that… |the cold storage demand for each immunization program can be tracked independently based on their unique set of requirements

**Solution B.2.1**

The system will provide a mechanism for central users to define an unlimited number of customisable immunization programs. 
These immunization programs will form the framework for determining estimated cold storage capacity demand. The contents of the framework are detailed in the following User Stories. 

| User story  | B.2.1.1 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to be able to select which immunization programs should be included in demand calculations
So that… | I can easily run hypothetical gap analysis and compare what happens if I include a program or not in the calculations

**Solution B.2.1.1**

The system will provide a mechanism for central users to include, or not include, immunization programs when estimating cold storage capacity demand. 

| User story  | B.2.3 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to be able to specify which demographic group each immunization program targets
So that… | our forecasting takes into account the number of people that require immunizing for each immunization program 

**Solution B.2.3**

The system will allow central users to associate demographic groups with immunization programs. 
Only 1 demographic group can be linked with each immunization program. 

| User story  | B.2.4 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to be able to specify one or more vaccines which are required for each immunization program
So that… | our forecasting takes into account the type of vaccines involved for each immunization program

**Solution B.2.4**

The system will allow central users to associate one or multiple vaccines with each immunization program. 
[...]

PR #4091, opened 2024-06-10
4058 gaps UI to update and save new immunisation programs

### LINKED ISSUE #4058: GAPS: UI to update and save immunisation programs

We now have some backend and some frontend code for managing/creating immunisation programs.
However they haven't been linked together

- [ ] Can create a new immunisation program (just need to provide a name)
- [ ] Can update the name of the program.

### Moneyworks Jobcode 🧰

CCEIDEV

jmbrunskill:
Fixes #4058

# 👩🏻‍💻 What does this PR do?

Allows the user to create and update immunisation programs.

Will be extended soon to include vaccine courses in #3916 

image
image

The app bar footer looks like it could with some visual separation but this seems to be the same as the current pattern.

# 🧪 Testing

- [ ] Login to Central Server
- [ ] Go to Programs - > Immunisations
- [ ] Create an immunisation program
- [ ] Update/Edit the program name

# 📃 Documentation

  1.
  2.


client/packages/system/src/Immunisation/ProgramDetailView/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,45 @@
...
+}: ToolbarProps) => {
+  const t = useTranslation('system');
+  return (
+    <AppBarContentPortal sx={{ display: 'flex', flex: 1, marginBottom: 1 }}>
+      <Grid container>
+        <Grid item display="flex" flex={1} flexDirection="column" gap={1}>
+          <BasicTextInput
+            fullWidth
+            value={draft.name}
+            onChange={e =>
```
fergie-nz, line 31: I wonder if onBlur better here with dirty check for saving?

I also wonder if a different UI would be better so the name seems more like 'info' about the page the user is on rather than an editable field

server/service/src/program/validate.rs
```diff
@@ -18,9 +18,14 @@ pub fn check_immunisation_program_exists(
 }
 pub fn check_program_name_exists(
     name: &str,
+    id: Option<String>,
     connection: &StorageConnection,
 ) -> Result<Option<ProgramRow>, RepositoryError> {
-    let program = ProgramRepository::new(connection)
-        .query_one(ProgramFilter::new().name(StringFilter::equal_to(name)))?;
+    let mut filter = ProgramFilter::new().name(StringFilter::equal_to(name));
```
fergie-nz, line 24: This is a cool way to make validate agnostic to insert vs update checks - I want to do the same to vaccine_courses as right now they're very verbose.

fergie-nz (approved):
I have minor preferences for UI differences in name editing in ProgramDetail view but these are opinions only. Looks great!

server/service/src/program/update_immunisation.rs
```diff
@@ -0,0 +1,90 @@
...
+use repository::{
+    ActivityLogType, ProgramRow, ProgramRowRepository, RepositoryError, StorageConnection,
+};
+use util::constants::IMMUNISATION_CONTEXT_ID;
+
+#[derive(PartialEq, Debug)]
+pub enum UpdateImmunisationProgramError {
+    ImmunisationProgramDoesNotExist,
+    ImmunisationProgramNameExists,
+    CreatedRecordNotFound,
```
lache-melvin, line 16: UpdatedRecordNotFound ?

server/graphql/programs/src/mutations/immunisation/update.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[derive(SimpleObject)]
+pub struct UpdateImmunisationProgramError {
+    pub error: UpdateImmunisationProgramErrorInterface,
+}
+
+#[derive(Interface)]
+#[graphql(field(name = "description", type = "String"))]
+pub enum UpdateImmunisationProgramErrorInterface {
+    ProgramDoesNotExist(RecordAlreadyExist),
```
lache-melvin, line 41: Should be accepting `RecordNotFound` ?

client/packages/system/src/Immunisation/api/hooks/useImmunisationProgram.ts
```diff
@@ -0,0 +1,174 @@
...
+) => {
+  const { api, storeId, queryClient } = useImmunisationGraphQL();
+  const t = useTranslation('system');
+
+  const mutationFn = async ({ name }: Partial<DraftImmunisationProgram>) => {
+    if (!id) {
+      throw new Error('No ID provided to update Immunisation Program');
+    }
+    if (!name) {
+      throw new Error('No name provided to update Immunisation Program');
```
lache-melvin, line 139: I think we should more gracefully handle someone clicking save with empty input field than throwing to the error boundary 😁 Probably handle in the component, before calling update()? Also should translate these error messages!

client/packages/system/src/Immunisation/ImmunisationProgramDetailView/AppBarButtons.tsx
```diff
@@ -17,9 +17,9 @@ export const AppBarButtons = ({}: ProgramAppBarButtonsProps) => {
       <Grid container gap={1}>
         <ButtonWithIcon
           Icon={<PlusCircleIcon />}
-          label={t('button.add-new-immunisation')}
+          label={t('button.add-new-vaccine-course')}
```
lache-melvin, line 20: Looks like you moved this translation to `catalogue.json` but didn't update the useTranslation call here?

Screenshot 2024-06-11 at 12 34 27 PM

client/packages/system/src/Immunisation/ListView/ListView.tsx
```diff
@@ -26,26 +28,31 @@ const ImmunisationProgramListComponent: FC = () => {
...
       onChangeSortBy: updateSortQuery,
       sortBy,
     },
     [updateSortQuery, sortBy]
   );
 
-  // later create modal will use <InsertImmunisationProgram> type
-  const { isOpen, onClose, onOpen } = useEditModal<any>();
+  const { isOpen, onClose, onOpen } =
+    useEditModal<InsertImmunisationProgramInput>();
```
lache-melvin, line 55: As mentioned in chat, I get this the first time I click the new program button after refresh?

Screenshot 2024-06-11 at 12 26 02 PM

Removing the isOpen check below seems to fix... dunno why?

lache-melvin (approved):

lache-melvin:
Looks great @jmbrunskill ! Main concern is throwing the error on the empty name input, would be good to smooth that out before merging 🙏 

Not from this PR, but also would be nice to make the immunization spelling consistent haha:

Screenshot 2024-06-11 at 12 24 20 PM

client/packages/system/src/Immunisation/api/hooks/useImmunisationProgram.ts
jmbrunskill: This was just to make typescript happy.
Save button shouldn't be enabled without these, but yeah could happen in future I guess...

jmbrunskill: Actually you're right, I can trigger it by removing all the text in the name...
Will update.

client/packages/system/src/Immunisation/ImmunisationProgramDetailView/AppBarButtons.tsx
jmbrunskill: Good catch!

PR #4096, opened 2024-06-10
2558 name merging store

### LINKED ISSUE #2558: Name merging: store

- [x] Add `store.name_link_id` and migrate with `store.name_link_id = store.name_id`
- [x] Remove `store.name_id`
- [x] Update any views that relied on relations of `store.name_id`
- [x] Update `store` at the repository layer to return related `name` records based on `store.name_link_id` rather than `store.name_id`

### Moneyworks Jobcode 🧰

OMS:MERGE

roxy-dao:
Fixes #2558

# 👩🏻‍💻 What does this PR do?
- Migrations for store.name_link_id
- Re-create views that use store
- Point to name_id

# 🧪 Testing

- [ ] Make sure tests still run

# 📃 Documentation

  1.
  2.


server/graphql/types/src/types/store.rs
```diff
@@ -32,13 +32,16 @@ impl StoreNode {
         let loader = ctx.get_loader::<DataLoader<NameByIdLoader>>();
 
         let response_option = loader
-            .load_one(NameByIdLoaderInput::new(&store_id, &self.row().name_id))
+            .load_one(NameByIdLoaderInput::new(
+                &store_id,
+                &self.row().name_link_id,
```
clemens-msupply, line 37: We should use name_id here and not the name_link_id, e.g. from the name_link row.  Using the name_link_id could be the wrong id, e.g. using this link id might fail because the link id is not a valid name_id anymore...

The name_link_id should be hidden in the repository. Ideally the name_link_row shouldn't even leave the repo layer, e.g. you could have store domain object that contains only the resolved name_id.

We didn't do this consistently because we wanted to keep the change minimal. However, I think here it can be done?

server/repository/migrations/postgres/2021-12-10T10-00_number_table/up.sql
```diff
@@ -8,8 +8,7 @@ CREATE TYPE number_type AS ENUM (
 );
 
 -- Numbering table holding a list of typed counters
-CREATE TABLE
-  number (
```
clemens-msupply, line 12: I would not touch these migration files. Migrations are read only once they are live. Sometimes you want to check that nothing has changed in your current branch in old migrations. Doing these style fixes can add some unnecessary noise...

server/repository/src/db_diesel/store_row.rs
```diff
@@ -91,9 +89,12 @@ impl<'a> StoreRowRepository<'a> {
         Ok(result)
     }
 
-    pub fn find_one_by_name_id(&self, name_id: &str) -> Result<Option<StoreRow>, RepositoryError> {
+    pub fn find_one_by_name_id(
+        &self,
+        name_link_id: &str,
```
clemens-msupply, line 94: Maybe rename the method as well. Not sure if we want to do this though, in general we want to lookup by name_id not by name_link_id... (see other comment)

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+          code TEXT NOT NULL,
+          site_id INTEGER NOT NULL,
+          logo TEXT,
+          store_mode TEXT DEFAULT 'STORE' NOT NULL,
+          created_date TEXT
+        );
+
+        INSERT INTO store_new SELECT * FROM store;
+
+        DROP TABLE store;
```
clemens-msupply, line 52: Nit `PRAGMA foreign_keys=off;` can be moved just before `DROP TABLE store;` right? This makes it more clear where it is needed

clemens-msupply (changes requested):
We should use name_id instead of name_link_id outside the repository. Using the name_link_id might be the wrong id to use, e.g. it may point to a deleted name_id...

At least we should use the name_id from the name_link row. However, this would be the cleaner solution:

https://github.com/msupply-foundation/open-msupply/pull/2869#discussion_r1475440266

server/graphql/types/src/types/store.rs
roxy-dao: That was gonna be another issue...

server/repository/migrations/postgres/2021-12-10T10-00_number_table/up.sql
roxy-dao: Sorry, just made it easier for me to search and doesn't affect migration...

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+    sql!(
+        connection,
+        r#"
+        -- Adding store.name_link_id
+        DROP INDEX index_store_name_id_fkey;
+        ALTER TABLE store
+        ADD COLUMN name_link_id TEXT;
+        
+        UPDATE store
+        SET name_link_id = name_id;
```
clemens-msupply, line 21: Nit: this and previous statement could be just one single line....

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
clemens-msupply: The PRAGMA is around the wrong statements, isn't it? Shouldn't it be around:

```
DROP TABLE store;
ALTER TABLE store_new RENAME TO store;
```

server/repository/src/migrations/v2_01_00/mod.rs
```diff
@@ -23,6 +24,7 @@ impl Migration for V2_01_00 {
     }
 
     fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        store_add_name_link_id::migrate(connection)?;
```
clemens-msupply, line 27:
```suggestion
        // Note, this migration deletes the consumptiopn view which is recreated in decimal_pack_size
        // migration, i.e. this migration has to run first.
        store_add_name_link_id::migrate(connection)?;
```

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+    )?;
+
+    #[cfg(not(feature = "postgres"))]
+    sql!(
+        connection,
+        r#"
+        -- Drop views early
+        DROP VIEW stock_on_hand;
+        DROP VIEW store_items;
+        -- consumption is recreated later in this migration
```
clemens-msupply, line 35: Maybe say where...

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+        );
+
+        INSERT INTO store_new SELECT * FROM store;
+
+        PRAGMA foreign_keys=on;
+        DROP TABLE store;
+        ALTER TABLE store_new RENAME TO store;
+
+
+        CREATE VIEW store_items AS
```
clemens-msupply, line 57: This view is also created in the decimal_pack_size migration...

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+        SELECT i.id as item_id, sl.store_id, sl.pack_size, sl.available_number_of_packs
+        FROM
+          item i
+          LEFT JOIN item_link il ON il.item_id = i.id
+          LEFT JOIN stock_line sl ON sl.item_link_id = il.id
+          LEFT JOIN store s ON s.id = sl.store_id;
+
+        CREATE VIEW stock_on_hand AS
```
clemens-msupply, line 65: This view is also created in the decimal_pack_size migration...

server/repository/src/migrations/v2_01_00/store_add_name_link_id.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    #[cfg(not(feature = "postgres"))]
+    sql!(
+        connection,
+        r#"
+        -- Drop views early
+        DROP VIEW stock_on_hand;
+        DROP VIEW store_items;
+        -- consumption is recreated later in this migration
+        DROP VIEW IF EXISTS consumption;
```
clemens-msupply, line 36: Why "IF EXISTS"?

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -153,7 +153,7 @@ fn generate_inbound_invoice(
     r#type: InboundInvoiceType,
 ) -> Result<InvoiceRow, RepositoryError> {
     let store_id = record_for_processing.other_party_store_id.clone();
-    let name_link_id = outbound_invoice.store_row.name_id.clone();
+    let name_link_id = outbound_invoice.store_row.name_link_id.clone();
```
clemens-msupply, line 156: Nit: the name_link_id is only used once, i.e. this variable can be inlined where it is used which makes it easier to see that then name_link_id is used correctly...

However, think even better would be to use `outbound_invoice.name_row.id`. This would further reduce the useage of `name_link_id`

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -31,7 +31,7 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
 
     /// Inbound invoice will be created when all below conditions are met:
     ///
-    /// 1. Source invoice name_id is for a store that is active on current site (transfer processor driver guarantees this)
+    /// 1. Source invoice name_link_id is for a store that is active on current site (transfer processor driver guarantees this)
```
clemens-msupply, line 34: Just using "name id" would be better (since name link id is suppose to be repo layer...)

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -111,7 +111,7 @@ fn generate_response_requisition(
     record_for_processing: &RequisitionTransferProcessorRecord,
 ) -> Result<RequisitionRow, RepositoryError> {
     let store_id = record_for_processing.other_party_store_id.clone();
-    let name_link_id = request_requisition.store_row.name_id.clone();
+    let name_link_id = request_requisition.store_row.name_link_id.clone();
```
clemens-msupply, line 114: Use `request_requisiton.name_row.id` and inline please:

```rust
RequisitionRow {
  ...
  name_link_id: request_requisiton.name_row.id.clone(),
  ...
```

server/service/src/sync/translations/name_store_join.rs
```diff
@@ -77,7 +77,7 @@ impl SyncTranslation for NameStoreJoinTranslation {
         {
             // if the name_store_join is referencing itself, then exclude it
             // this is an invalid configuration which shouldn't be possible.. but is
-            if store.name_id == data.name_id {
+            if store.name_link_id == data.name_id {
```
clemens-msupply, line 80: Is this correct? if yes I can't easily see that. For this reason I suggest to use the StoreRepository and then compare `store.name_row.id === data.name_id`

clemens-msupply (changes requested):
Thanks Roxy, some more comments, mainly:

- please check the views, they are already recreated later...
- there are some more places where we can reduce the use of name_link_id by using the name_id

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
roxy-dao: Also when names are merged... doesn't the id stay as the kept name id since we're using id = name.id?

server/service/src/processors/transfer/requisition/create_response_requisition.rs
roxy-dao: again... not the correct name id...

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
clemens-msupply: Good point, you need to get the store name_id.

> Also when names are merged... doesn't the id stay as the kept name id since we're using id = name.id?

Yes, but the records could use the deleted name_link_id. It would still work, but it would be cleaner to use the name_link_id of the current name_id.

clemens-msupply (approved):
Thank you looks great! tests run fine

PR #4103, opened 2024-06-10
3862 custom name data edit on remote site

lache-melvin:
Fixes #3862

# 👩🏻‍💻 What does this PR do?

Adds ROM-COM syncing (😁) for properties of names, to allow editing of name properties from a remote site:

With `FEATURE_GAPS` enabled, you will now see `Edit` (naming?) in the footer:
Screenshot 2024-06-11 at 11 49 15 AM

Clicking will open the edit modal for your store:
Screenshot 2024-06-11 at 11 49 24 AM

## 💌 Any notes for the reviewer?

Went with another button in the footer, rather than in the store selector, because store selector only opens if you have more than one store, and most of our client sites don't!

Will create a "remove name_property migration and run a script instead" issue, but I think as part of that, maybe we should hide the edit button if no name properties configured? Or somehow otherwise conditionally render it so only available if using GAPS?

Have removed the is_sync_update/changelog triggers for `name` table, as it was conflicting with `name_oms_field` changes. Appears to be all working fine but would appreciate a double check on that!

# 🧪 Testing

(Will need to rerun migrations first!)

- [ ] Log into remote site with FEATURE_GAPS enabled
- [ ] See `Edit` button in the footer
- [ ] Modal opens, enter some data
- [ ] Click Ok
- [ ] Click edit again - the data is still there
- [ ] Sync
- [ ] Log into OMS central server
- [ ] Go to Manage > Facilities, and find the store from the remote site that you edited
- [ ] Check that the changes are there

# 📃 Documentation

- [x] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

PR #4120, opened 2024-06-11
4117 replace FEATURE_GAPS flag with vaccine module check

### LINKED ISSUE #4117: Hide GAPS work behind Vaccine Module preference

GAPS features don't make much sense without the Vaccine module enabled, so they should only be visible if the Vaccine module preference is on. 

While you're at it - remove the GAPS feature flag, we want these things visible for 2.1!

### Moneyworks Jobcode 🧰

CCEIDEV

lache-melvin:
Fixes #4117

NOTE: #4103 needs to be included in this! I didn't want to stack, but whatever gets merged first...

# 👩🏻‍💻 What does this PR do?

Uses vaccine module to enable/disable GAPS features, and removes feature flag

# 🧪 Testing

- [ ] In mSupply, disable `Vaccine module` preference
- [ ] In omSupply (central), in the Nav Drawer, you CANNOT see:
  - [ ] Manage section
  - [ ] Programs section
  - [ ] Catalogue > Demographics page
- [ ] In mSupply, enable `Vaccine module` preference
- [ ] In omSupply (central), in the Nav Drawer, you CAN see:
  - [ ] Manage section
  - [ ] Programs section
  - [ ] Catalogue > Demographics page (note, this might end up under `Manage` at some stage

# 📃 Documentation

  1.
  2.


jmbrunskill:
I wonder if Catalogue -> Assets should be behind this preference too? Not a major but could be a tidy up?
image

@adamdewey, will all stores that need to add their population, Level, etc have the vaccine module enabled? I think the answer is yes as they'll be managing their CC assets right?

Just wondering if https://github.com/msupply-foundation/open-msupply/pull/4103 (Remote store editing properties) should be behind this feature or not. Feeling like might be ok to keep it enabled, or maybe only enabled if name properties are defined?

jmbrunskill (approved):
Looks good to me.

I assume this duplicate button is in develop, as I don't see how this could be introduced here?
image

adamdewey:
> I wonder if Catalogue -> Assets should be behind this preference too? Not a major but could be a tidy up? 
I don't think it should be - as general asset management doesn't necessarily mean they will be using the CCEI module (they might just want to manage general assets in the yet-to-be-developed general assets module)
However we could potentially hide the WHO PQS Sub-catalogue to only appear when the vaccine preference is turned on.

> @adamdewey, will all stores that need to add their population, Level, etc have the vaccine module enabled? I think the answer is yes as they'll be managing their CC assets right?
At that moment that's true - but I can imagine that other functionality might want to make use of supplementary facility-related data at some point?

> Just wondering if #4103 (Remote store editing properties) should be behind this feature or not. Feeling like might be ok to keep it enabled, or maybe only enabled if name properties are defined?
Again, may be useful for other functionality that wants to use supplementary facility data at some stage?

lache-melvin:
Ok thanks team - leaving footer edit button, probably to only be displayed when properties are defined. And leaving assets always visible 🙏

PR #4134, opened 2024-06-12
Save demographic projection data

### LINKED ISSUE #4055: GAPS: Demographics - Population Growth is not saved
## How to Reproduce 🔨

1. Go to `Catalogue->Demographics`
2. Enter a Population and growth rates
3. Click save
4. Exit the page and re-open
5. See % Growth on previous year fields have been reset to their original values (before saving)

## Your environment 🌱

- Open mSupply Version: `develop`

mark-prins:
Fixes #4055

# 👩🏻‍💻 What does this PR do?
Quick change to see why the values aren't saved.. and 27 files later, it's blown out a bit.

* Now fetching projection values from the server 
* added in defaults, for the initial state when no projections exist
* wired up the update and insert for saving
* did a little tidying along the way (have noted in comments)
* fixed an issue with update API
* added error handling and success/fail notifications
* changed the database to use floats rather than int for the % values

You'll likely have to manually update the `demographic_projection` table

postgreSQL version:
```
ALTER TABLE demographic_projection ALTER COLUMN year_1 TYPE float8;
ALTER TABLE demographic_projection ALTER COLUMN year_2 TYPE float8;
ALTER TABLE demographic_projection ALTER COLUMN year_3 TYPE float8;
ALTER TABLE demographic_projection ALTER COLUMN year_4 TYPE float8;
ALTER TABLE demographic_projection ALTER COLUMN year_5 TYPE float8;
```

# 🧪 Testing

Things to check:
- [ ] that with a default state (no projections in table) works - shows the demographic data with `0` for percentages
- [ ] changing the percentage and saving works, shows success and adds a row to the table
- [ ] updating after this works (and calls the update mutation)
- [ ] refreshing the page shows the retained values

# 📃 Documentation

  1.
  2.


client/packages/system/src/IndicatorsDemographics/DetailView/AppBarButtons.tsx
```diff
@@ -1,13 +1,15 @@
...
   FnUtils,
+  Grid,
   PlusCircleIcon,
   RecordPatch,
   useTranslation,
 } from '@openmsupply-client/common';
-import { Row } from './IndicatorsDemographics';
-import { AppBarButtonsPortal, Grid } from '@openmsupply-client/common';
+
+import { Row } from '../types';
```
mark-prins, line 12: quick tidy, and moved the types into `types`

client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
```diff
@@ -83,21 +63,24 @@ const IndicatorsDemographicsComponent: FC = () => {
...
+    useDemographicData.projection.list(draft?.[0]?.baseYear ?? 2024);
 
   const { insertDemographicIndicator, invalidateQueries } =
-    useDemographicData.document.insertIndicator();
+    useDemographicData.indicator.insert();
   const { mutateAsync: updateDemographicIndicator } =
-    useDemographicData.document.updateIndicator();
+    useDemographicData.indicator.update();
+  const upsertProjection = useDemographicData.projection.upsert();
+  const getHeaderDraft = () =>
```
mark-prins, line 82: probs not required now that the hook is returning something. but can't hurt I guess.

client/packages/system/src/IndicatorsDemographics/api/api.ts
```diff
@@ -112,45 +112,20 @@ export const getDemographicIndicatorQueries = (sdk: Sdk) => ({
...
-        throw new Error('Could not find encounter');
-      }
-    },
     list: async ({
       first,
       offset,
-      sortBy,
       filterBy,
-    }: ListParams<DemographicProjectionFragment>) => {
+    }: Omit<ListParams<DemographicProjectionFragment>, 'sortBy'>) => {
```
mark-prins, line 119: the listAll and byId fns aren't needed

client/packages/system/src/IndicatorsDemographics/api/hooks/document/useDemographicProjections.ts
```diff
@@ -1,14 +1,33 @@
...
-    filters: [{ key: 'name' }, { key: 'basePopulation' }, { key: 'id' }],
-  });
+export const useDemographicProjections = (baseYear: number) => {
   const api = useDemographicsApi();
-  const filterBy = queryParams.filterBy;
-  const params = { ...queryParams, filterBy };
-  return useQuery(api.keys.paramProjectionList(params), () =>
-    api.getProjections.list(params)
-  );
+  const params = {
```
mark-prins, line 7: we aren't using urlQueryParams so I removed that and tidied this to the essentials

client/packages/system/src/IndicatorsDemographics/api/operations.graphql
```diff
@@ -56,6 +56,7 @@ query demographicProjections(
   ) {
     ... on DemographicProjectionConnector {
       nodes {
+        __typename
```
mark-prins, line 59: with this missing, the insert thought that it failed bc it is checking for a valid typename

client/packages/system/src/IndicatorsDemographics/api/operations.graphql
```diff
@@ -147,7 +140,7 @@ mutation updateDemographicProjection(
         ... on DemographicProjectionNode {
           id
         }
-        ... on InsertDemographicProjectionError {
+        ... on UpdateDemographicProjectionError {
```
mark-prins, line 143: this prevented the update from being called..

server/graphql/demographic/src/mutations/insert_projection.rs
```diff
@@ -49,11 +49,11 @@ pub fn insert_demographic_projection(
 pub struct InsertDemographicProjectionInput {
     pub id: String,
     pub base_year: i32,
-    pub year_1: Option<i32>,
-    pub year_2: Option<i32>,
-    pub year_3: Option<i32>,
-    pub year_4: Option<i32>,
-    pub year_5: Option<i32>,
+    pub year_1: Option<f64>,
```
mark-prins, line 52: we need to store floats, so I had to modify

client/packages/system/src/IndicatorsDemographics/DetailView/utils.tsx
```diff
@@ -77,3 +83,38 @@ export const calculateAcrossRow = (
...
+
+export const getDefaultHeaderData = (baseYear: number): HeaderData => ({
+  baseYear,
+  id: '',
+  '1': { id: '1', value: 1 },
+  '2': { id: '2', value: 1 },
+  '3': { id: '3', value: 1 },
+  '4': { id: '4', value: 1 },
+  '5': { id: '5', value: 1 },
+});
```
lache-melvin, line 95: This is confusing with you also returning a default projection from the query hook if the list is empty, especially because there you return 0s and here you return 1s... can we pick one? 🤷‍♀️

client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
```diff
@@ -209,7 +190,14 @@ const IndicatorsDemographicsComponent: FC = () => {
               : await updateIndicator(indicator);
           }
         })
-      ).then(() => invalidateQueries());
+      )
+        .then(async () => {
+          if (headerDraft !== undefined)
+            await upsertProjection(mapProjection(headerDraft));
```
lache-melvin, line 196: Because you're doing this inside the loop, projections aren't saved if you have no indicators... also seems unnecessary to re-upsert the projections for every indicator - maybe do before the loop?

client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
lache-melvin: This is just adding indirection/confusion for me?

client/packages/system/src/IndicatorsDemographics/api/operations.graphql
lache-melvin: These need to be added on the insert mutation as well, getting the same error!

lache-melvin (changes requested):
Thanks Mark - looks good overall, though I don't understand the move to querying for a list if we only ever want the one projection row?

However a couple of things not working... if you have no indicators:

https://github.com/msupply-foundation/open-msupply/assets/55115239/5f727c14-8d18-48ea-ae4e-6f04a52243ff

Looks like some errors aren't being surfaced? Not sure if this is related to your changes:

https://github.com/msupply-foundation/open-msupply/assets/55115239/a2f8fc19-d1d6-4111-b6d0-76816f419618

client/packages/system/src/IndicatorsDemographics/api/operations.graphql
```diff
@@ -73,18 +74,6 @@ query demographicIndicatorById($demographicIndicatorId: String!) {
   }
 }
 
-query demographicProjectionById($demographicProjectionId: String!) {
```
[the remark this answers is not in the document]
mark-prins: we don't have an id to fetch with! what we actually need to do is fetch by `baseYear` and in lieu of a fn for that, I just filter the list and take the first row. it's not ideal but is the pattern we have elsewhere. 

would be nice to have `getByBaseYear` though. I might just do that..

client/packages/system/src/IndicatorsDemographics/DetailView/utils.tsx
mark-prins: I'll remove this and see what happens. basically changed direction mid way and kept this in.

client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
mark-prins: cool - will do 🙏

client/packages/system/src/IndicatorsDemographics/api/operations.graphql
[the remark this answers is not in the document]
lache-melvin: Ohhh that makes sense 🤣 Yeah getByBaseYear would be awesome, nice not to have the `nodes[0]` everywhere 😁

mark-prins:
thanks @lache-melvin! have tidied up. can you have another look please?

client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
```diff
@@ -136,27 +102,28 @@ const IndicatorsDemographicsComponent = () => {
       ...patch,
       value: patch.value ?? 0,
     };
-    setHeaderDraft({ ...getHeaderDraft(), [patch.id]: updatedPatch });
+    setHeaderDraft({
+      ...(headerDraft as HeaderData),
```
lache-melvin, line 106: Looks like a lot of them aren't your changes, but I'm a bit worried about the number of `as Type`s that we have going on in this file... maybe something else to see if we can fix in the cleanup issue 👀

client/packages/system/src/IndicatorsDemographics/api/hooks/document/useDemographicIndicatorInsert.ts
```diff
@@ -10,13 +10,7 @@ export const useDemographicIndicatorInsert = () => {
     queryClient.invalidateQueries(api.keys.baseIndicator());
   const { mutateAsync: insertDemographicIndicator } = useMutation(
     async (demographicIndicator: DemographicIndicatorFragment) =>
-      api.insertIndicator(demographicIndicator),
-    {
-      onSettled: () => queryClient.invalidateQueries(api.keys.baseIndicator()),
```
lache-melvin, line 15: I'm guessing you removed these because any changes are already reflected in UI?

client/packages/system/src/IndicatorsDemographics/DetailView/utils.tsx
```diff
@@ -116,3 +107,30 @@ export const mapProjection = (
   year4: headerData[4].value,
   year5: headerData[5].value,
 });
+
+export const toRow = (row: {
```
lache-melvin, line 111: nit: `toDemographicIndicatorRow` ?

server/service/src/demographic/update_demographic_projection.rs
```diff
@@ -55,22 +55,15 @@ pub fn validate(
...
-    let demographioc_projection_row =
+    let demographic_projection_row =
         match check_demographic_projection_exists(&input.id, connection)? {
             Some(demographic_projection_row) => demographic_projection_row,
             None => {
                 return Err(UpdateDemographicProjectionError::DemographicProjectionDoesNotExist)
             }
         };
-    if let Some(base_year) = input.base_year {
-        if !check_base_year_unique(base_year, connection)? {
```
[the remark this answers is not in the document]
lache-melvin: This is still breaking the tests... I think should put this check back - especially now that  the `getByBaseYear` call is returning only one projection!

lache-melvin (approved):
Thanks for changes Mark! Approving with a couple of things:

- Failing backend test - put the unique check back?
- General population is only being saved once you add an indicator? I think this might have been happening before too, feel free to add to other demographics fix issue 🙏

client/packages/system/src/IndicatorsDemographics/api/hooks/document/useDemographicIndicatorInsert.ts
mark-prins: the code was running the `invalidateQueries` after all of the save actions - having it in here means that the cache is invalidated after each row is saved, and that's not necessary

server/service/src/demographic/update_demographic_projection.rs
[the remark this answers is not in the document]
mark-prins: The reason this was removed is because it prevents updates from happening!
it checks to see if `baseYear` is anywhere in the table - and of course it will be if you are updating a row and only changing the percentages. 

What it should check, is whether any rows other than this one have this base year. I'll update the validation and tests.

mark-prins:
Thanks again Laché - have fixed the test.
Yes, the gen pop doesn't save until you add an indicator - that's a more fundamental change than I want to make on this PR.. it requires changing how the data is stored, it doesn't exist as a separate row unfortunately.
and yes, the typescript is pretty loose, but I haven't updated that either.

PR #12, opened 2022-05-23
1096 fix test

clemens-msupply:
Closes #11
Closes #5

PR #4140, opened 2024-06-12
Clippy #12

roxy-dao:
Part of https://github.com/msupply-foundation/open-msupply/issues/3017

# 👩🏻‍💻 What does this PR do?
Just one more.... *fingers crossed*

Mostly just changing to_owned to to_string in here or using clone_from to share objects and avoid allocations

# 🧪 Testing

- [ ] Make sure tests are still OK

# 📃 Documentation

  1.
  2.


server/Cargo.toml
```diff
@@ -36,6 +36,9 @@ large_enum_variant = "allow"
 module_inception = "allow"
 bool_assert_comparison = "allow"
 result_large_err = "allow"
+ptr_arg = "allow"
+# TODO: Will need frontend changes too. Will be nice for readability.
+too_many_arguments = "allow"
```
clemens-msupply, line 41: ptr_arg: sounds like a nice lint error, but I guess too many places where we violate it? maybe another refactor? but then we don't usually reuse our functions so its not really needed and would be a bit of work...

too_many_arguments: wonder if we should disable this case by case? Can't be that many places where we have too many arguments?

clemens-msupply (approved):
Thanks!

server/Cargo.toml
roxy-dao: Yup, I think ptr_arg would be nice to have in eventually too~. Though it tells you to use slice as the variable to pass in then in the method has you convert that to_vec 🤔 

I think it's fine to have it globally for now...

PR #4145, opened 2024-06-12
Merge sqlite/pg upsert_one and insert_one_or_ignore code

### LINKED ISSUE #4110: Remove replace_into SQLite upsert implementation
## The suggestion

Merge the SQLite and Postgres `upsert_one` implementations - diesel 2.0 appears to include on_conflict support for sqlite!

Discovered this in [this PR](https://github.com/msupply-foundation/open-msupply/pull/4065#issuecomment-2156926282), editing some fields on the `name` table via two different DSLs, found that the SQLite `upsert` implementation caused issues doing a complete replace (REPLACE INTO) of the row, rather than updating just the fields on the specified DSL.

## Example use case

In all our row repositories, we have this code:

```rust
    #[cfg(feature = "postgres")]
    fn _upsert_one(&self, name_row: &NameRow) -> Result<(), RepositoryError> {
        diesel::insert_into(name)
            .values(name_row)
            .on_conflict(id)
            .do_update()
            .set(name_row)
            .execute(self.connection.lock().connection())?;
        Ok(())
    }

    fn _upsert_one(&self, name_row: &NameRow) -> Result<(), RepositoryError> {
        diesel::replace_into(name)
            .values(name_row)
            .execute(self.connection.lock().connection())?;
        Ok(())
    }

    pub fn upsert_one(&self, row: &NameRow) -> Result<(), RepositoryError> {
        self._upsert_one(row)?;
        self.insert_changelog(...bla...)
    }
```

Now we could just have:

```rust
    pub fn upsert_one(&self, row: &NameRow) -> Result<(), RepositoryError> {
         diesel::insert_into(name)
            .values(name_row)
            .on_conflict(id)
            .do_update()
            .set(name_row)
            .execute(self.connection.lock().connection())?;

        self.insert_changelog(...bla...)
    }
```

- In moving away from mSupply Central management, it's possible we'll end up with more situations like we had for `name` ... (an existing table, originally from mSupply Central, but we want to add new columns to it which are only stored in open mSupply Central). We would need the "insert or update on conflict" rather than "replace" implementation to do this.
- Reduces code branching between PG and SQLite, which is always a good thing
- Slightly less boilerplate in the `Repository` implementations which will be nice

I think fairly low risk - if there is any divergence in how our `upsert`s work across PG and SQLite, that's probably a bug in the first place?

An hour or two? Basically just gotta delete some code 👀

clemens-msupply:
Fixes #4110 

# 👩🏻‍💻 What does this PR do?

Merge sqlite/pg upsert_one and insert_one_or_ignore code

# 🧪 Testing

sqlite/postgres tests are still running? (currently there is an unrelated test error)

### DIFF
--- a/server/repository/src/db_diesel/clinician_row.rs
+++ b/server/repository/src/db_diesel/clinician_row.rs
@@ -62,9 +62,7 @@ fn insert_or_ignore_clinician_link(
         clinician_id: row.id.clone(),
     };
 
-    ClinicianLinkRowRepository::new(connection).insert_one_or_ignore(&clinician_link_row)?;
-
-    Ok(())
+    ClinicianLinkRowRepository::new(connection).insert_one_or_ignore(&clinician_link_row)
 }
 
 pub struct ClinicianRowRepository<'a> {
@@ -76,21 +74,12 @@ impl<'a> ClinicianRowRepository<'a> {
         ClinicianRowRepository { connection }
     }
 
-    #[cfg(feature = "postgres")]
     fn _upsert_one(&self, row: &ClinicianRow) -> Result<(), RepositoryError> {
         diesel::insert_into(clinician::dsl::clinician)
             .values(row)
             .on_conflict(clinician::dsl::id)
             .do_update()
             .set(row)
-            .execute(self.connection.lock().connection())?;
-        Ok(())
-    }
-
-    #[cfg(not(feature = "postgres"))]
-    fn _upsert_one(&self, row: &ClinicianRow) -> Result<(), RepositoryError> {
-        diesel::replace_into(clinician::dsl::clinician)
-            .values(row)
             .execute(self.connection.lock().connection())?;
         Ok(())
     }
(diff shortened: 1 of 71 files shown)


server/repository/src/db_diesel/clinician_row.rs
```diff
@@ -206,5 +198,15 @@ mod test {
 
         assert_eq!(repo.find_is_sync_update_by_id(&row.id), Ok(Some(false)));
         assert_eq!(repo.find_is_sync_update_by_id(&row2.id), Ok(Some(false)));
+
+        // Sneak in a test for insert_one_or_ignore. If another entry is inserted/upserted in
+        // insert_one_or_ignore() the clinician_id foreign key would be violated and the following
+        // would fail:
+        ClinicianLinkRowRepository::new(&connection)
```
clemens-msupply, line 205: See code comment...

lache-melvin: Seems fine here to me, thanks for adding 🙏

lache-melvin (approved):
Sweeeeet thanks Clemens! 🚀  Yeah the other test should be fixed in develop now 😅

PR #4146, opened 2024-06-13
4079 facilities list should only include stores

### LINKED ISSUE #4079: GAPS: Non-stores are showing in `Facilities`
## What went wrong? 😲

Non-store names are showing in `Facilities` whereas I think it should only be showing Stores?

image

## Expected behaviour 🤔

Only Stores show in the list.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version: Develop 2.1.0
- Legacy mSupply Central Server Version: 2.0.0

- Platform:

- Database type:

lache-melvin:
Fixes #4079

# 👩🏻‍💻 What does this PR do?

Filters the facilities list to only include stores.

Turns out the names query was overwriting the `type` to filter out patients - fixed this compose the two filters instead 😅 

There are actually two store filters on name...

1. `isStore` - this checks whether we have a store for that name, via the `name_store_join`
2. `NameType` of `STORE` - name type from mSupply (i.e. as opposed to Facility, Patient...)

Currently I'm using option 2, would you say this is correct?

In the datafile that I'm using, it doesn't seem to matter, like there aren't any names with the `NameType` of `STORE` that don't have a linked store. Or facilities that DO have a linked store. But wanted to check if that is possible?

Though the little store icon on the left hand side is derived from `isStore`... so maybe I should be using that?? 🤷‍♀️ 

# 🧪 Testing

- [ ] Go to Manage > Facilities
- [ ] The list should only contain stores (all have the house icon on the left hand side)

# 📃 Documentation

### DIFF
--- a/server/service/src/name/query.rs
+++ b/server/service/src/name/query.rs
@@ -1,6 +1,7 @@
-use repository::{
-    EqualFilter, Name, NameFilter, NameRepository, NameSort, NameType, PaginationOption,
-};
+use repository::NameRepository;
+use repository::NameType;
+use repository::PaginationOption;
+use repository::{Name, NameFilter, NameSort};
 
 use crate::{
     get_default_pagination, i64_to_u32, service_provider::ServiceContext, ListError, ListResult,
@@ -19,17 +20,9 @@ pub fn get_names(
     let pagination = get_default_pagination(pagination, MAX_LIMIT, MIN_LIMIT)?;
     let repository = NameRepository::new(&ctx.connection);
 
-    let filter = filter.unwrap_or_default();
-
-    let type_filter = match filter.r#type.clone() {
-        Some(filter_input) => EqualFilter::<NameType> {
-            not_equal_to: Some(NameType::Patient),
-            ..filter_input
-        },
-        None => NameType::Patient.not_equal_to(),
-    };
-
-    let filter = filter.r#type(type_filter);
+    let filter = filter
+        .unwrap_or_default()
+        .r#type(NameType::Patient.not_equal_to());
 
     Ok(ListResult {
         rows: repository.query(store_id, pagination, Some(filter.clone()), sort)?,
--- a/server/service/src/name/query.rs
+++ b/server/service/src/name/query.rs
@@ -1,7 +1,6 @@
-use repository::NameRepository;
-use repository::NameType;
-use repository::PaginationOption;
-use repository::{Name, NameFilter, NameSort};
+use repository::{
+    EqualFilter, Name, NameFilter, NameRepository, NameSort, NameType, PaginationOption,
+};
 
 use crate::{
     get_default_pagination, i64_to_u32, service_provider::ServiceContext, ListError, ListResult,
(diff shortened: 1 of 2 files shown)


adamdewey:
>NameType of STORE - name type from mSupply (i.e. as opposed to Facility, Patient...)

Is this still true if the Store is set to be a `Dispensary`?

server/service/src/name/query.rs
```diff
@@ -20,9 +19,17 @@ pub fn get_names(
...
 
-    let filter = filter
-        .unwrap_or_default()
-        .r#type(NameType::Patient.not_equal_to());
+    let filter = filter.unwrap_or_default();
+
+    let type_filter = match filter.r#type.clone() {
+        Some(filter_input) => EqualFilter::<NameType> {
+            not_equal_to: Some(NameType::Patient),
+            ..filter_input
```
lache-melvin, line 27: I guess this still isn't perfect, if you pass in a `ne` filter that would get overwritten. I tried to do a more comprehensive merge but it was looking unwieldy - any tips? https://github.com/msupply-foundation/open-msupply/compare/4079-facilities-list-should-only-include-stores...4079-facilities-list-should-only-include-stores(better-not-equal-filter)

lache-melvin:
> > NameType of STORE - name type from mSupply (i.e. as opposed to Facility, Patient...)
> 
> Is this still true if the Store is set to be a `Dispensary`?

Yeah I believe that's fine, becuase Dispensiary is a setting on the store itself, rather than the name, which is what we are looking at here 👀

server/service/src/name/query.rs
jmbrunskill: I'm not 100% sure of the intention here, It's basically a default filter to exclude names I think.

My suggestion is to make this more explicit in the API, but that will require some more testing for things that might want to include patients.
https://github.com/msupply-foundation/open-msupply/compare/4079-facilities-list-should-only-include-stores...4079-facilites-list-name-suggestion

I don't like this uncommented filter as the expected behaviour isn't entirely clear, is this a security protection thing? Is it a useful default? etc...

server/service/src/name/query.rs
```diff
@@ -20,9 +19,17 @@ pub fn get_names(
...
+        },
+        None => NameType::Patient.not_equal_to(),
+    };
```
jmbrunskill, line 30:
```suggestion
     let type_filter = match filter.r#type.clone() {
        Some(filter_input) => filter_input, // If we have a Name type filter, use it, trusting the caller.
        None => NameType::Patient.not_equal_to(), // Default to excluding patients from name searches
    };
```

How about just this, keeps it simple? But doesn't solve a security problem if that's the intention...

I'm not 100% if the intention of that filter was to try to prevent this service method being used for querying patients? There is a patient service which I assume we use for actual patient related things...
If so then I think we hard code that as a filter in the repository layer similar to my suggestion branch. And don't even expose the option to get patient to the graphql layer.
@andreievg do you know the intention here?

jmbrunskill (approved):
Looks good to me, the filter is a bit of a mystery so would be good to make intention clearer and align the code to that intention.

server/service/src/name/query.rs
andreievg: > 'm not 100% if the intention of that filter was to try to prevent this service method being used for querying patients?

I think so, but this should be done in the NameRepository (as overall filter, and similar in Patient repository), looks like there was a refactor to split Name repository at some point (into patient and store/facility), but looks like it was just partially done.

Ideally I would do hard coded filter in the create_filtered_query for both Patient and Name repository. Also would remove is_patient form NameFilterInput and NameNodeType::Patient

> If so then I think we hard code that as a filter in the repository layer similar to my suggestion branch. And don't even expose the option to get patient to the graphql layer.

Yep i think as you've described, but not quite as the suggested branch implementation (i.e. there is Patient repo and patient filter, this should be used in patient queries, and reserve name completely for other types)

lache-melvin: Ok awesome thanks team - this is feeling like a bigger change than the scope of the issue... I'll just use the `isStore` filter and create a separate issue for this 🙏

lache-melvin: #4161

PR #4151, opened 2024-06-13
3642 patient search results display in created date order

### LINKED ISSUE #3642: Patient search results: display in created date order

PLM in Sierra Leone paid for this to be done in mSupply mobile here https://github.com/msupply-foundation/mobile/issues/5395

We added a "Created date" column to the patient search results, default results order was newest created date at the top (but column header was clickable to reverse it if required).

The column seems very helpful in determining whether the returned patient record is the patient sitting in front of you when there are duplicate names (very common in some places).

Something like that above. Could work with a store preference, can be always on if better.

### Additional context 💌

Gary said we'd do this for free because it's small and they're good people (strategic decision will little real cost to us but lots of goodwill earned). PLM paid to have this done in mobile (now live) and will need this in Open mSupply before they roll out to more sites (planned for Q4 2024 or Q1 2025 approximately).

roxy-dao:
Fixes #3642

# 👩🏻‍💻 What does this PR do?
Adds `created date` column to patients and allow users to use this to sort patients (defaults: newest created patient -> oldest)

Was unsure where to put the column... Didn't want it to be too right since it might be hidden on smaller screens...

Should the central search screen also return newest to oldest patients...?

# 🧪 Testing

- [ ] Have some patients with different created datetimes
- [ ] Go to `Patients` page
- [ ] See that newest patients are displayed first
- [ ] Click created date sorting, should sort correctly

# 📃 Documentation

  1. Screenshots showing the whole patients page
  2.

### DIFF
--- a/client/packages/system/src/Patient/ListView/ListView.tsx
+++ b/client/packages/system/src/Patient/ListView/ListView.tsx
@@ -77,7 +77,6 @@ const PatientListComponent: FC = () => {
     {
       key: 'createdDatetime',
       label: 'label.created',
-      width: 175,
       formatter: dateString =>
         dateString ? localisedDate((dateString as string) || '') : '',
       sortable: true,
@@ -97,7 +96,6 @@ const PatientListComponent: FC = () => {
     {
       key: 'dateOfBirth',
       label: 'label.date-of-birth',
-      width: 175,
       formatter: dateString =>
         dateString ? localisedDate((dateString as string) || '') : '',
     },
--- a/client/packages/system/src/Patient/ListView/ListView.tsx
+++ b/client/packages/system/src/Patient/ListView/ListView.tsx
@@ -39,7 +39,7 @@ const PatientListComponent: FC = () => {
     filter,
     queryParams: { page, first, offset, sortBy, filterBy },
   } = useUrlQueryParams({
-    initialSort: { key: 'code', dir: 'asc' },
+    initialSort: { key: 'createdDatetime', dir: 'desc' },
     filters: [
       {
         key: 'dateOfBirth',
@@ -74,6 +74,14 @@ const PatientListComponent: FC = () => {
   const columnDefinitions: ColumnDescription<PatientRowFragment>[] = [
     { key: 'code', label: 'label.patient-id' },
     { key: 'code2', label: 'label.patient-nuic' },
+    {
+      key: 'createdDatetime',
+      label: 'label.created',
+      width: 175,
+      formatter: dateString =>
+        dateString ? localisedDate((dateString as string) || '') : '',
+      sortable: true,
+    },
     {
       key: 'firstName',
       label: 'label.first-name',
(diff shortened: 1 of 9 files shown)


server/graphql/types/src/types/user.rs
```diff
@@ -30,7 +30,7 @@ impl UserStoreNode {
     }
 
     pub async fn name_id(&self) -> &str {
-        &self.user_store.store_row.name_id
+        &self.user_store.store_row.name_link_id
```
roxy-dao, line 33: Develop was broken... probably wanna link name to store row here to get correct name

server/graphql/reports/src/printing.rs
```diff
@@ -282,7 +282,7 @@ async fn fetch_data(
     }
     let mut data = if let Some(ResolvedReportQuery::GraphQlQuery(gql)) = graphql_query.get(0) {
         let variables = query_variables(store_id, &data_id, &arguments, &sort, &gql.variables);
-        let result = fetch_graphq_data(ctx, gql, variables).await?;
+        let result = fetch_graphql_data(ctx, gql, variables).await?;
```
roxy-dao, line 285: side quest 💃🏻

client/packages/system/src/Patient/ListView/ListView.tsx
```diff
@@ -74,6 +74,14 @@ const PatientListComponent: FC = () => {
   const columnDefinitions: ColumnDescription<PatientRowFragment>[] = [
     { key: 'code', label: 'label.patient-id' },
     { key: 'code2', label: 'label.patient-nuic' },
+    {
+      key: 'createdDatetime',
+      label: 'label.created',
+      width: 175,
```
clemens-msupply, line 80: You can reduce the width here to 110. However, I also tried to reduce the width of the dob column which works with 150 which also would work for the createdDatetime and would give a consistent column width. Another option is to remove the width option for both columns, which works fine for me and maybe makes the best use of the page width?

clemens-msupply (approved):
Thanks works fine. Some suggestions:

1) I tested the dateOfBirth column with the `sortable: true,` and it works fine, so this could be added?
2) Created column is a bit out of place on its current position, maybe better put it on last position? Looks like it has been a requested feature, so people would know that its there in case they are on a very small screen? This "column overflow problem" is a general problem that we probably need to solve at some point, e.g. have  some UI to scroll/unhide offscreen columns... So maybe better to put it at a position where it best fits in?
3) column width tweaks (see inline comment)

server/graphql/reports/src/printing.rs
clemens-msupply: +100 xp

roxy-dao:
> Thanks works fine. Some suggestions:
> 
> 1. I tested the dateOfBirth column with the `sortable: true,` and it works fine, so this could be added?
> 2. Created column is a bit out of place on its current position, maybe better put it on last position? Looks like it has been a requested feature, so people would know that its there in case they are on a very small screen? This "column overflow problem" is a general problem that we probably need to solve at some point, e.g. have  some UI to scroll/unhide offscreen columns... So maybe better to put it at a position where it best fits in?
> 3. column width tweaks (see inline comment)

Spoke with @mark-prins and the options were
(a) between ID and First name or (b) right at the end after deceased.

Though it feels weird after deceased 🤔 

Deceased | Created

PR #4158, opened 2024-06-16
4094 soft delete vaccine course

### LINKED ISSUE #4094: Add delete action to vaccine course within vaccine course detail view
There's no delete action on the [wireframe in epic issue](https://github.com/msupply-foundation/open-msupply/issues/3916), make sure that one is added and then hooked up
the spec says 'one or more vaccines' so the wireframe needs work: likely use a datatable with the delete checkbox / action

- [ ] Add delete button to remove vaccine course
- [ ] Add delete mutation to back end (will need to be soft delete)
- [ ] Add delete to graphql and front end api
- [ ] Prompt to warn people before deleting a vaccine course
- [ ] Show success or fail snack on success or fail

### LINKED ISSUE #4105: Add soft delete for vaccine course items and vaccine schedule items in back end

- [ ]  Add vaccine course item soft delete in backend up to graphql
- [ ] Add vaccine schedule item soft delete in backend up to graphql

lache-melvin:
Fixes #4094

Fixes #4105

# 👩🏻‍💻 What does this PR do?

Adds select column, delete action to immunisation program detail view, to delete vaccine courses:

Screenshot 2024-06-17 at 11 40 23 AM

Adds `deleted_datetime` to vaccine courses to soft delete them.

Not soft deleting the related vaccine course items/schedules - I think this is fine? Would only ever access these via the vaccine course right?

# 🧪 Testing

- [ ] On central server, go to Programs > Immunizations
- [ ] Open the detail view of a program
- [ ] Create some vaccine courses
- [ ] Use the select column to select some courses
- [ ] Select delete action, and confirm
- [ ] Your courses are deleted

# 📃 Documentation

  1.
  2.


server/graphql/vaccine_course/src/mutations/delete_vaccine_course.rs
```diff
@@ -0,0 +1,72 @@
...
+};
+
+pub fn delete_vaccine_course(
+    ctx: &Context<'_>,
+    vaccine_course_id: &str,
+) -> Result<DeleteVaccineCourseResponse> {
+    let user = validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::MutateVaccineCourse,
```
lache-melvin, line 20: Note that this would only work once permissions PR is merged in...

server/service/src/vaccine_course/test/delete.rs
```diff
@@ -0,0 +1,50 @@
...
+        };
+
+        let _result = service
+            .insert_vaccine_course(&context, vaccine_course.clone())
+            .unwrap();
+
+        let result = service.delete_vaccine_course(&context, vaccine_course.id.clone());
+
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), vaccine_course.id);
```
jmbrunskill, line 48: Seeing as you went to all the trouble to create a test, might not hurt to do a query to check it's not found?

jmbrunskill (commented):
Looks good to me.

Permissions is ready to merge, but waiting on the merge for saving vaccine courses as I branched off there...

jmbrunskill (approved):

adamdewey:
> Not soft deleting the related vaccine course items/schedules

I think we should soft delete the doses too (not the items) for a neater solution, just in case.

lache-melvin:
> I think we should soft delete the `vaccine course doses` too (not the items) for a neater solution, just in case.

Hm, wonder if @jmbrunskill has thoughts here... we might come back to this? 

Whenever you edit the vaccine course, we delete and recreate all the doses - I guess that doesn't prevent us soft-deleting the doses if the vaccine course is soft deleted, but it feels a bit strange to have it hard deleted sometimes, and soft deleted other times 👀 

Though if we get to the point of associating patients with specific doses, we'll probably have to consider something other than the delete/recreate pattern?

> Also I may not be understanding what a `vaccine course item` is - was assuming it means the Master Data item?

Ah yeah... `Vaccine course item` is just a reference to a master list item, just for the relationship between this vaccine course and the items in it, not really an actual entity 👍

PR #4163, opened 2024-06-17
4129 fix item link for vaccine course

### LINKED ISSUE #4129: fix item_link for vaccine_course
## What went wrong? 😲

item_link_id shouldn't be used in graphql requests for vaccine_course_item

## Expected behaviour 🤔

Follow existing pattern for item_link_ids

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #4129

# 👩🏻‍💻 What does this PR do?

Adds join table to vaccine_course_item query so that it pulls item_id in stead of relying on item_link_id

Leaves the pattern of pulling item node from a loader using the item_id pulled from the database.

This adds an additional query call so I wasn't quite sure why it is the pattern used elsewhere. It saves re writing out of all of the variables in graphql types.

# 🧪 Testing

This PR doesn't change front end behaviour. 
It adds a join to the vaccine item query so that vaccine item queries are resilient to item merges.

To test this appropriately, we need to ensure item merges can take place, and we still can query the items in a vaccine course.

- [ ]  Create a new vaccine course and add an item to this course
- [ ] Merge the item into a new item 
- [ ] See this change reflected in the vaccine course (ie name is updated to that of the new name)

# 📃 Documentation

  1.
  2.


server/graphql/types/src/types/vaccine_course_item.rs
```diff
@@ -3,42 +3,44 @@ use dataloader::DataLoader;
...
 }
 
 #[Object]
 impl VaccineCourseItemNode {
     pub async fn id(&self) -> &str {
-        &self.row().id
+        &self.row().vaccine_course_item.id
     }
 
     pub async fn item(&self, ctx: &Context<'_>) -> Result<ItemNode> {
```
jmbrunskill, line 21: If it can be done easily, It would be nice to remove the loader call.
I suspect others kept the loader for the same reason you did, it was the easiest option.

jmbrunskill (approved):
Looks good, might be worth doing the loader but only if it's quick...

server/graphql/types/src/types/vaccine_course_item.rs
```diff
@@ -3,42 +3,44 @@ use dataloader::DataLoader;
...
 impl VaccineCourseItemNode {
-    pub fn from_domain(vaccine_course_item: VaccineCourseItemRow) -> VaccineCourseItemNode {
+    pub fn from_domain(vaccine_course_item: VaccineCourseItem) -> VaccineCourseItemNode {
         VaccineCourseItemNode {
             vaccine_course_item,
         }
     }
 
-    pub fn row(&self) -> &VaccineCourseItemRow {
+    pub fn row(&self) -> &VaccineCourseItem {
```
jmbrunskill, line 43: Is this case it probalby doesn't matter, but I think part of the point of the `row()` call is to avoid needing to update all the places when something like this changes.
So  you could do 
 pub fn row(&self) -> &VaccineCourseItemRow {
        &self.vaccine_course_item.vaccine_course_item
    }
pub fn item_row(&self) -> ItemRow {
        &self.vaccine_course_item.item
}

jmbrunskill:
BTW Did you check this works with an actual item merge? Can you update the testing steps in the PR to include this step to make sure it's tested properly.

server/graphql/types/src/types/vaccine_course_item.rs
fergie-nz: Quick I think! Will do now

fergie-nz:
> BTW Did you check this works with an actual item merge? Can you update the testing steps in the PR to include this step to make sure it's tested properly.

Good point. I've updated the testing documentation.

I haven't ever done an item merge - I wonder would you be willing to show me how?

fergie-nz:
> > BTW Did you check this works with an actual item merge? Can you update the testing steps in the PR to include this step to make sure it's tested properly.
> 
> Good point. I've updated the testing documentation.
> 
> I haven't ever done an item merge - I wonder would you be willing to show me how?

OK I have it but have encountered a bug with this.

If you add two items to the same course, and subsequently merge those items, you receive duplicate items (because two are saved both pointing to the same now merged item).

I will make a new issue for this as not immediately obvious how to handle. Maybe a clean up for duplicate item_ids? 

I wonder how this is handled elsewhere in code?

Issue #2882, opened 2024-02-01
Update b/e crates


roxy-dao:
## Is your feature request related to a problem? Please describe 👀
As per title. 

# Major
| Package | Current | Latest | Package Type | CHANGELOG |
|---|---|---|---|---|
| clap | 3.1.8 | 4.4.18 | report_builder, server, cli | https://github.com/clap-rs/clap/blob/master/CHANGELOG.md |  |
| uuid | 0.8 | 1.7.0 | repository | ?? |  |
| rust-embed | 6.4.0 | 8.2.0 | repository, server | https://github.com/pyrossh/rust-embed/blob/master/changelog.md |  |
| rustls-pemfile | 1.0.0 | 2.0.0 | server | ?? |  |
| jsonwebtoken | 8.0.1. | 9.2.0 | service | https://github.com/Keats/jsonwebtoken/blob/master/CHANGELOG.md |  |
| async-graphql | 3.0.35 | 7.0.1 | cli, graphql_programs, graphql_form_schema, graphql_plugin | https://github.com/async-graphql/async-graphql/blob/master/CHANGELOG.md |  |
| async-graphql-actix-web | 3.0.35 | 7.0.1 | graphql_programs, graphql_form_schema, graphql_plugin | https://github.com/async-graphql/async-graphql/blob/master/CHANGELOG.md |  |

# Minor
| Package | Current | Latest | Package Type | CHANGELOG |
|---|---|---|---|---|
| jni | 0.19.0 | 0.21.1 | android |  https://github.com/jni-rs/jni-rs/releases |  |
| rcgen | 0.9.2 | 0.12.1 | android, server | ?? |  |
| tokio | 1.17.0 | 1.35.1 | android, repository, graphql, server, service, graphql_core, cli | https://github.com/tokio-rs/tokio/blob/master/tokio/CHANGELOG.md |  | 
| libsqlite3-sys | 0.22.2 | 0.28.0 | repository | ?? |  |
| serde_yaml | 0.8.23 | 0.9.31 | report_build, service | https://rust.libhunt.com/serde-yaml-changelog |  |
| strum | 0.24 | 0.26.1 | repository, graphql_core, graphql_stocktake_line, graphql_invoice_line | https://github.com/Peternator7/strum/blob/master/CHANGELOG.md |  |
| machine-uid | 0.2.0 | 0.5.1 | server | ?? |  |
[...]

PR #4164, opened 2024-06-17
2882.1 update b/e crates

roxy-dao:
Part of #2882

# 👩🏻‍💻 What does this PR do?
Updates all the minor/patch crates. There are 5 major updates left coming up soon...

# 🧪 Testing

- [ ] Make sure tests are still running okay and everything is still working alright.

# 📃 Documentation

  1.
  2.


server/server/src/certs.rs
```diff
@@ -118,11 +118,11 @@ impl Certificates {
 
         let key_file = cert_dir.join(PRIVATE_CERT_FILE);
         let mut file = std::fs::File::create(&key_file)?;
-        file.write_all(cert.serialize_private_key_pem().as_bytes())?;
+        file.write_all(cert.cert.pem().as_bytes())?;
```
roxy-dao: Now handled by KeyPair::serialize_pem()

server/server/src/certs.rs
```diff
@@ -118,11 +118,11 @@ impl Certificates {
...
 
         let cert_file = cert_dir.join(PUBLIC_CERT_FILE);
         let mut file = std::fs::File::create(&cert_file)?;
-        let public_cert_buffer = cert.serialize_pem()?;
+        let public_cert_buffer = cert.cert.pem();
```
roxy-dao, line 126: Now serialised with Certificate::pem

server/Cargo.toml
```diff
@@ -3,28 +3,28 @@ strip = true
...
+chrono = { version = "0.4.38", features = ["serde"] }
 reqwest = { version = "0.12", features = [
   "json",
   "rustls-tls",
   "multipart",
   "blocking",
 ], default-features = false }
-serde = "1.0.126"
-serde_json = "1.0.66"
+serde = "1.0.203"
```
jmbrunskill, line 22:
```suggestion
serde = "1.0"
```
Can't we just do this, and always have the latest 1.0 version?

server/android/Cargo.toml
```diff
@@ -9,21 +9,21 @@ path = "src/lib.rs"
...
+rcgen = "0.13.1"
 repository = { path = "../repository" }
 service = { path = "../service" }
 server = { path = "../server", default-features = false, features = [
     "android",
 ] }
-tokio = { version = "1.17.0", features = ["macros"] }
-android_logger = { version = "0.13.0" }
-simple-log = { version = "1.6" }
+tokio = { version = "1.38.0", features = ["macros"] }
```
jmbrunskill, line 24: Tokio and other depenancies are used in lots of places, can we move them to be workspace?
E.g. 
`tokio = { workspace = true }`

Then we should mostly only need to update our workspace versions?
Personally I'd try to move as many as possible to this approach. Should help with version consistency across our crates...

jmbrunskill (approved):
Looks good to me,
I haven't tested as I don't want to download all the new crates on my phone data haha.

I think we should consider putting more of this into the workspace level to avoid making the same changes in multiple crates?

I'm also not sure if we gain anything when going from say `1.0` to `1.0.2` in theory we should already get the `1.0.2` version when the lock file is recalculated right? And doesn't require further manual intervention...

server/Cargo.toml
roxy-dao: Oh yes can!

server/android/Cargo.toml
roxy-dao: Can do

jmbrunskill: Thanks!

PR #4169, opened 2024-06-18
Fix asset catalogue import with properties

### LINKED ISSUE #4012: Unable to import asset catalogue items with asset properties
## What went wrong? 😲

image

## Expected behaviour 🤔

Import should work or at least return an error such as (duplicated code)

## How to Reproduce 🔨

1. Login to OMS Central Server
2. Click on Catalogue -> Assets
3. Click Import
4. Download example import
5. Try to import

## Your environment 🌱

- Open mSupply Version:2.1

jmbrunskill:
Fixes #4012

# 👩🏻‍💻 What does this PR do?

Embarassingly easy fix in the end, I switched the example output to use the the property key instead of name, which fixes the issue.

Basically papaparse doesn't correctly parse csv header rows when they contain commas.
Was relying on a simple split on `,` the error was confusing though, as it appeared to suggest it couldn't find the right closing quote in an escape sequence. 

There's an issue in papaparse for this already.
https://github.com/mholt/PapaParse/issues/1052

# 🧪 Testing

- [ ] Central OMS Server
- [ ] Go to Catalogue -> Assets Import
- [ ] Download Example File
- [ ] Update this with real data
- [ ] Import
- [ ] Check that creating an asset inherits all the correct properties

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Asset/ImportCatalogueItem/UploadTab.tsx
+++ b/client/packages/system/src/Asset/ImportCatalogueItem/UploadTab.tsx
@@ -138,7 +138,7 @@ export const AssetItemUploadTab: FC<ImportPanel & AssetItemUploadTabProps> = ({
       exampleRows,
       t,
       false, // exclude errors
-      properties ? properties.map(p => p.name) : []
+      properties ? properties.map(p => p.key) : []
     );
     FileUtils.exportCSV(csv, t('filename.asset-import-example'));
   };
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
🎉  good find!
Import works for me.. but then it was anyway. this time I used the example and that's all good.

mark-prins:
Also - I had to add a lot of `IF EXISTS` to the drops in `store_add_name_link_id` as the migrations failed due to things not existing (postgres). was hoping that it was because I'm flipping between branches too much 🤷

PR #4149, opened 2024-06-13
4057 gaps permissions

jmbrunskill:
Fixes #4057

# 👩🏻‍💻 What does this PR do?

Adds support for new mSupply Permission for Central Server admin
Fixes permission checks for GAPS features

## 💌 Any notes for the reviewer?

# 🧪 Testing

**Setup**

Install, update, or obtain this PR from mSupply for your OG Central Server - https://github.com/msupply-foundation/msupply/pull/14837

Minimum Permissions for GAPS
image
image
image
image
image

- [ ] Login to Central Server (User permissions as above)
- [ ] Check you can add and update demographics
- [ ] Check you can add and update Immunisation Programs
- [ ] Check you can add and update Assets
- [ ] Check you can update facilities (under Manage

Then check you can't do these things with the appropriate permissions removed...

# 📃 Documentation

- [X] **Part of an epic**: documentation will be completed for the feature as a whole

PR #4170, opened 2024-06-18
4162 delete immunisation programs

### LINKED ISSUE #4162: GAPS: soft delete immunization programs

It should be possible to soft delete immunisation programs from the list view

Add selection column, and dropdown to select and delete lines.

Implement deleted datetime on immunisation programs and filter only non-deleted ones to be returned.

### Moneyworks Jobcode 🧰

CCEIDEV

lache-melvin:
Fixes #4162

# 👩🏻‍💻 What does this PR do?

Adds select column, delete action to immunisation program list view, to soft delete programs:

Screenshot 2024-06-18 at 1 41 48 PM

If you don't have the edit central server data permission, app currently crashes, issue in base pr 🙈 https://github.com/msupply-foundation/open-msupply/pull/4149#issuecomment-2174710895

Usually, our soft delete pattern has just been soft-deleting the parent record, but wondering what kind of querying might be done from Grafana/future app features... Would "all vaccine courses" be queried? In which case, I should also soft delete any vaccine courses on the program?? Or can come back to that at such a time that we need it?

# 🧪 Testing

With the edit omSupply central data permission enabled:

- [ ] On central server, go to Programs > Immunizations
- [ ]  Have/create some programs
- [ ]  Use the select column to select some programs
- [ ]  Select delete action, and confirm
- [ ]  Your programs are deleted

# 📃 Documentation

  1.
  2.


jmbrunskill (approved):
Looks good to me, haven't run tests though.

adamdewey:
> Usually, our soft delete pattern has just been soft-deleting the parent record, but wondering what kind of querying might be done from Grafana/future app features... Would "all vaccine courses" be queried? In which case, I should also soft delete any vaccine courses on the program?? Or can come back to that at such a time that we need it?

Yeah this is a good question. 

A vaccine course instance would only ever belong to a single program so in my mind it would be a neater solution if we soft-delete the children when we soft-delete the parent - feels like we're leaving loose ends otherwise?

lache-melvin:
As per meeting yesterday - also soft deleting vaccine courses when a program is soft deleted 🙏

PR #4182, opened 2024-06-18
3729 Returns: if there is no 'Reason' configured on mSupply, you still can type to search

### LINKED ISSUE #3729: Returns: if there is no 'Reason' configured on mSupply, you still can type to search
## What went wrong? 😲

- Enhancement - 

After talking to @lache-melvin about returns, we noticed the 'Reason' field is editable, which can be confusing for users as they may think a new reason can be input on that window. 
However the field is editable to enable the search only. 

We think it can be confusing for users. 

telegram-cloud-photo-size-1-5015052128800058620-y

## Proposed solution 🤔

I agree with Lache's idea: it would be a nicer user experience if the reason dropdown was disabled if there were no reasons setup in mSupply.

## Your environment 🌱

- Open mSupply Version: V2.0 RC2

- Platform: Desktop app (windows)

roxy-dao:
Fixes #3729

# 👩🏻‍💻 What does this PR do?
Disables return reason if no reason is configured
Screenshot 2024-06-19 at 2 53 18 AM

# 🧪 Testing

- [ ] Have no return reasons set up
- [ ] Go to `Inbound Returns`
- [ ] Create a new return and add a line
- [ ] Put in a quantity and go to next step
- [ ] See that reason should be disabled

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
+++ b/client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
@@ -33,7 +33,7 @@ export const ReturnReasonSearchInput: FC<ReturnReasonSearchInputProps> = ({
       <Autocomplete
         fullWidth
         autoFocus={autoFocus}
-        disabled={isDisabled || data?.nodes.length === 0}
+        disabled={isDisabled || reasons.length === 0}
         clearable={false}
         value={
           value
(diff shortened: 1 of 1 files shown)


client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
```diff
@@ -33,7 +33,7 @@ export const ReturnReasonSearchInput: FC<ReturnReasonSearchInputProps> = ({
       <Autocomplete
         fullWidth
         autoFocus={autoFocus}
-        disabled={isDisabled}
+        disabled={isDisabled || data?.nodes.length === 0}
```
lache-melvin, line 36: Nit~

```suggestion
        disabled={isDisabled || reasons.length === 0}
```

lache-melvin (approved):
Yay thank youuu

PR #4193, opened 2024-06-19
Permission f/e refactor

### LINKED ISSUE #4057: GAPS: Permissions

We plan to keep permissions simple for the GAPS work.

Handle new mSupply `oms` permissions for  Can Modify Central Data
(This will provide access to Demographics, Immunisation Programs, and Vaccine Courses)

image

Name editing will be linked to Edit Customer, supplier, and manufacturer names
image

- Create OMS only permissions controlled by  OMS Central Server (probably too much work)

### Moneyworks Jobcode 🧰
CCEIDEV

mark-prins:
Fixes #4057 

# 👩🏻‍💻 What does this PR do?
Adds my request for changes on #4149 which I didn't get in on time!

The primary change is to use the disabled notification process that we have elsewhere - check permission and show toast (or popup) on click of buttons.

The secondary request was to reduce the null coalesce operator usage; on my reading it wasn't clear from the code why it was necessary (the comment explained it well however!) and I thought that with a refactor we could 
(a) make it clearer, and 
(b) remove the coalesce need

Then I noticed that lodash had been imported directly from the app, which in previous experience increase our bundle size, so I fixed that.

# 🧪 Testing

- Check that you are prompted when you don't have permission to 
  - [ ] Create a vaccine course
  - [ ] Create an immunisation program
  - [ ] Add a demographic indicator
  - [ ] Save changes to demographics

# 📃 Documentation

  1.
  2.


client/packages/common/src/authentication/hooks/index.ts
```diff
@@ -1,2 +1 @@
 export * from './usePermissionCheck';
-export * from './useCheckPermissionWithError';
```
mark-prins, line 2: this hook isn't required, I think - and could cause confusion with the existing hook

client/packages/common/src/hooks/useNativeClient/useNativeClient.ts
```diff
@@ -1,6 +1,6 @@
 import { useState, useEffect, useCallback } from 'react';
 import { useMutation } from 'react-query';
-import { uniqWith } from 'lodash';
```
mark-prins, line 3: it's ok to import when in the common package, but for consistency, and to prevent copy-paste errors later, I've updated em all.

client/packages/common/src/intl/locales/en/coldchain.json
```diff
@@ -36,6 +36,7 @@
   "error.label-printer-not-configured": "There is no label printer configured. Please see the Admin > Devices section to configure a label printer.",
   "error.manufacturer-model-unique": "An item already exists with this manufacturer and model",
   "error.no-asset-create-permission": "You do not have permission to create a new asset.",
+  "error.unable-to-create-immunisation-program": "Unable to create Immunisation Program",
```
mark-prins, line 39: there are more we could add.. but I didn't

client/packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import DOMPurify from 'dompurify';
-import { get as extractProperty } from 'lodash';
+import { extractProperty } from '@common/utils';
```
mark-prins, line 3: I renamed in common, since all usages had renamed it!

client/packages/system/src/Immunisation/ImmunisationProgramDetailView/AppBarButtons.tsx
```diff
@@ -15,22 +16,21 @@ interface ProgramAppBarButtonsProps {
 
 export const AppBarButtons = ({ onCreate }: ProgramAppBarButtonsProps) => {
   const t = useTranslation('coldchain');
-  const checkPermissionDenied = useCheckPermissionWithError(
-    UserPermission.EditCentralData
-  );
+  const { userHasPermission } = useAuthContext();
+  const showDisabledNotification = useDisabledNotificationToast();
+
+  const onClick = () => {
```
mark-prins, line 22: this is the pattern we've put elsewhere - to check permission and show a toast.

client/packages/common/src/authentication/hooks/useCallbackWithPermission.ts
```diff
@@ -0,0 +1,17 @@
+import { useAuthContext } from '../AuthContext';
+import { useDisabledNotificationToast } from '@common/components';
+import { UserPermission } from '@common/types';
+
+export const useCallbackWithPermission = (
```
mark-prins, line 5: this is possibly a little OTT - but I think worth it.
we've used this pattern in a few places now, and rather than repeating, I've consolidated into a hook

jmbrunskill: Nice, I almost wonder if we should call it `useCallbackIfPermission` or something? 
Can't think of a really nice clear name, I think it's ok how it is. 

At least it has Permission in the name rather than `useDisabledNotificationToast` which doesn't obviously relate to permissions...

jmbrunskill (approved):
Looks good to me, thanks for the updates!

client/packages/common/src/authentication/hooks/useCallbackWithPermission.ts
mark-prins: funnily enough.. `useCallbackIfPermitted` was my first name attempt and then I dithered and came up with this. Naming stuff.. the eternal problem! Agree though - having 'permission' in there is clearer than 'disabled notification' is!

PR #4227, opened 2024-06-21
Display RC version

### LINKED ISSUE #4217: Display RC version
A request from the QA team: it would be super useful to be able to see the RC version on the site.
If we install v2.1.00-RC1 and then install RC2 later, and are juggling multiple test sites, it's hard to know which site is running which version.

On the login screen, extend the version info to also show RC.
On the admin screen, also show the RC version along with the installed version.

### Additional context 💌
Do we add the version to the binary?

mark-prins:
Fixes #4217

# 👩🏻‍💻 What does this PR do?
A somewhat exploratory PR for an issue that is technically still asking for triage love. 

The problem with having the semantically versioned rc in the `package.json` is that the version is used to update the database and run migrations. 

My suggestion here is to simply strip that final section from the semver in package.json, such that `2.10.00-rc1` is displayed on the front end unaltered, but interpreted by the API as `2.1.00`

This gives the following:
_login screen_
Screenshot 2024-06-21 at 2 11 01 PM

_settings screen_
Screenshot 2024-06-21 at 2 10 53 PM

Why is this required? It allows testers to know which version they are running.

# 🧪 Testing

# 📃 Documentation

  1.
  2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.1.00",
+  "version": "2.1.00-rc1",
   "private": true,
   "scripts": {
     "start": "cd ./server && cargo run & cd ./client && yarn start-local",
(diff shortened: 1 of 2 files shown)


lache-melvin (approved):
Very nice! This will be gladly received 😁

PR #4237, opened 2024-06-21
3874 you can perform transactions on disabled oms store

### LINKED ISSUE #3874: You can perform transactions on disabled OMS store
## What went wrong? 😲
 

You can perform transactions on disabled store

Video Reference: https://drive.google.com/file/d/1ObJkJAH2OFDs9KQPUZYbtcj4E8-o2k4R/view?usp=share_link

## Expected behaviour 🤔
disable store means restrict any further stock transfers to the disabled store? 

## How to Reproduce 🔨

1. mSupply datafile having open mSupply stores
2. Make sure they are visible to msupply stores 
3. Now disable the open msupply store

Screenshot 2024-05-13 at 17 12 41

4. Sync, logout and login to open mSupply Sites
5. Now make Stock transfer to mSupply stores
6. You can make stock transfer, Even if the store is disabled 
(BUG)

## Additional
Can i call this severity high? .. since the `disabled` oms-store looks very active, not disabled at all

## Your environment 🌱

- Open mSupply Version: v2-00-rc5
- Legacy mSupply Central Server Version: v7-15-00-dev5

- Platform: standalone

- Database type: SQlite

roxy-dao:
Fixes #3874

# 👩🏻‍💻 What does this PR do?
1. Migration to add `is_disabled` to store
2. In store selector at the bottom left of the screen: Disable store + add on hold text
Screenshot 2024-06-21 at 7 28 44 PM
3. Filter out disabled store from most recent credentials 

Note for testers: Will need to delete database and reinitialise since there has been a migration change to store

# 🧪 Testing

- [ ] in mSupply: Disable store
- [ ] Sync to OMS
- [ ] See that disabled store is greyed out in Store Selector in bottom left 

# 📃 Documentation

  1. Add a note about disabled stores
  4.


client/packages/common/src/authentication/api/hooks/useLogin.ts
```diff
@@ -70,11 +70,15 @@ export const getStore = async (
     item => item.username.toLowerCase() === userDetails?.username?.toLowerCase()
   );
 
-  if (mru?.store && stores?.some(store => store.id === mru?.store?.id)) {
+  if (
```
mark-prins, line 73: I've had to change - this is possibly and edge case, but to be safe..

My most recently used store was disabled, then I logged in and was logged into it:
This is the `Opua` store after disabling, syncing and then logging out and in again:

Screenshot 2024-06-24 at 11 36 37 AM


I've filtered the store list, like so: 
```
export const getStore = async (
  userDetails?: Partial<UserNode>,
  mostRecentCredentials?: AuthenticationCredentials[]
) => {
  const defaultStore = userDetails?.defaultStore;
  const stores = userDetails?.stores?.nodes.filter(s => !s.isDisabled);
  const mru = mostRecentCredentials?.find(
    item => item.username.toLowerCase() === userDetails?.username?.toLowerCase()
  );

  if (
    mru?.store &&
    !mru.store.isDisabled &&
    stores?.some(store => store.id === mru?.store?.id && !store.isDisabled)
  ) {
    return stores.find(store => store.id === mru.store?.id) ?? mru.store;
  }

  if (!!defaultStore && !defaultStore.isDisabled) return defaultStore;

  return !!stores && stores?.length > 0 ? stores?.[0] : undefined;
};
```

in which case I think the `    !mru.store.isDisabled &&` line is no longer required. The check should be on the list of stores returned by the server, not the MRU which could be stored before the store was disabled

mark-prins (changes requested):
🎉  lovely! thanks Roxy - nice fix.
Can I request a wee change to the f/e bit please? I was able to login to a disabled store - if the MRU object does not have `isDisabled` then you can login, it needs to check the store status from the API

andreievg:
If there is only one store, and it was disabled, what would the user see when trying to log in ?

client/packages/common/src/authentication/api/hooks/useLogin.ts
roxy-dao: [aeb8239](https://github.com/msupply-foundation/open-msupply/pull/4255/commits/aeb82392984fa7543f2c875dc6352e78d9046ea4)

PR #4240, opened 2024-06-22
4085 add import store custom fields via csv

### LINKED ISSUE #4085: GAPS: Populate Store custom fields via spreadsheet import

#3862 added functionality to store supplementary custom data against Stores. 

The central user needs to be able to populate this data for all Stores via a spreadsheet import. 

| User story  | B.2.16 |
As a… | National EPI manager
I want… | to be able to bulk upload data for facilities from Excel
So that… | I don't need to update each facility manually

**Solution B.2.16**
The system will permit a bulk import of facility data to aid the central users in the scenario that the system is not distributed and therefore cannot leverage facility users to enter their own data.

### Moneyworks Jobcode 🧰

CCEIDEV

fergie-nz:
Fixes #4085

# 👩🏻‍💻 What does this PR do?

Adds importing of custom facilities properties via csv spreadsheet

This is accessible via Manage -> Facilities -> Import Properties

You will need to add name properties and properties in the database in order to test.

# 🧪 Testing

# 📃 Documentation

  1.
  2.


client/packages/system/src/Name/ListView/Facilities/ListView.tsx
```diff
@@ -9,11 +9,15 @@ import {
...
+  useToggle,
 } from '@openmsupply-client/common';
-import { useName, NameRowFragment } from '../../api';
+import { useName } from '../../api';
 import { NameRenderer } from '../../Components';
 import { Toolbar } from './Toolbar';
 import { FacilityEditModal } from './FacilityEditModal';
+import { AppBarButtons } from './AppBarButtons';
+import { PropertiesImportModal } from '../ImportProperties/PropertiesImportModal';
+import { FacilityNameRowFragment } from '../../api/operations.generated';
```
fergie-nz, line 20: Have added a new nameRowFragment which includes properties along with a new query for facilities to include properties.

I wasn't sure whether best to do this, or to add properties to all other nameRowFragment calls.

client/packages/system/src/Name/ListView/ImportProperties/ImportPanel.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { FC, PropsWithChildren } from 'react';
```
fergie-nz, line 1: Maybe this component could be shared as generic with other similar import via csv features?

fergie-nz: Which now includes this, asset custom properties, and catalogue item custom properties

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+  {
+    columnDescriptions.push({
+      key: 'errorMessage',
+      label: 'label.error-message',
+      width: 150,
+      Cell: TooltipTextCell,
+    });
+  }
+
+  const rowsWithProperties = rows.map(row => {
```
fergie-nz, line 74: Add properties to review data table.

We don't have this in other custom properties CSV import - but I figured this makes sense because this import is specifically for properties import. Also in this case we upsert changes, so helpful to see what is being removed as well as added.

However I'm not sure how it will look with many custom properties where it could extend far off the page

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
```diff
@@ -0,0 +1,199 @@
...
+  onUploadComplete: () => void;
+  properties: NamePropertyNode[] | undefined;
+}
+interface ParsedImport {
+  id: string;
+  [key: string]: string | undefined;
+}
+
+enum FacilitiesColumn {
+  CODE = 0,
```
fergie-nz, line 34: Maybe more fields helpful here? I'm not sure

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
```diff
@@ -0,0 +1,199 @@
...
+      const code = getCell(row, FacilitiesColumn.CODE);
+      if (code === undefined || code.trim() === '') {
+        rowErrors.push(
+          t('error.field-must-be-specified', {
+            field: t('label.code'),
+          })
+        );
+      }
+
+      const id = facilities?.find(facility => facility.code == code)?.id;
```
fergie-nz, line 149: Find id by matching facility code to facilities list which was called in the list view.

Assumes unique facility codes.

Maybe a more elegant way to do this, but seemed to work fine and allows id not to be required in the csv import

client/packages/system/src/Name/api/operations.graphql
```diff
@@ -11,6 +11,20 @@ fragment NameRow on NameNode {
   }
 }
 
+fragment FacilityNameRow on NameNode {
```
fergie-nz, line 14: Or nameRowWithProperties?

client/packages/system/src/utils.ts
```diff
@@ -86,7 +87,7 @@ interface ParsedRow {
 export const processProperties = <
   T extends { properties: Record<string, string> },
 >(
-  properties: AssetPropertyFragment[],
+  properties: AssetPropertyFragment[] | PropertyNode[],
```
fergie-nz, line 90: Make utils function generic to different property node

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+    }
+    onChangeTab(tabName);
+  };
+
+  const showWarnings = errorMessage.length == 0 && warningMessage.length > 0;
+
+  console.info('showWarnings', showWarnings);
+
+  const importSteps = [
+    { label: t('label.upload'), description: '', clickable: true },
```
lache-melvin, line 170: Screenshot 2024-06-25 at 9 59 33 AM

Translations for the modal aren't coming through, same for the upload error message... also title should be import facility properties rather than CCE :grin:

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+  });
+  const currentFacilitiesPage = filteredEquipment.slice(
+    pagination.offset,
+    pagination.offset + pagination.first
+  );
+
+  // console.log('currentFacilitiesPage', currentFacilitiesPage);
+
+  return (
+    <Grid flexDirection="column" display="flex" gap={0}>
```
lache-melvin, line 99: Screenshot 2024-06-25 at 10 05 29 AM

Something's up with styles here, getting overflow issues in both directions (can't scroll)

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+      case Tabs.Import:
+        setActiveStep(2);
+        break;
+    }
+    onChangeTab(tabName);
+  };
+
+  const showWarnings = errorMessage.length == 0 && warningMessage.length > 0;
+
+  console.info('showWarnings', showWarnings);
```
lache-melvin, line 167: can remove now

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+      case Tabs.Review:
+        changeTab(tabName as Tabs);
+        break;
+      case Tabs.Import:
+        // Do nothing, user can't get to the import page without clicking the import button
+        break;
+    }
+  };
+
+  const changeTab = (tabName: Tabs) => {
```
lache-melvin, line 149: Screenshot 2024-06-25 at 10 09 34 AM

Are you calling this in the right places? The tabs look like this, as though you're on the review tab, the whole way through the process 👀

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+        // Load the error rows in to the component for review
+        setErrorMessage(t('messages.import-generic'));
+        setBufferedFacilityProperties(importErrorRows);
+        setImportErrorCount(importErrorRows.length);
+        onChangeTab(Tabs.Review);
+      }
+    }
+  };
+
+  const onClickStep = (tabName: string) => {
```
lache-melvin, line 135: Clicking steps isn't changing the step for me 👀

client/packages/system/src/Name/ListView/ImportProperties/ImportPanel.tsx
lache-melvin: Yeah that's a great idea! There is a lot of boilerplate code here... perhaps a common `ImportModal` component that handles all the tabs etc, and we pass in the example csv generator, import parser...?

lache-melvin: I reckon let's make a refactor issue for it though, rather than bloat this PR :)

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+  rows,
+  properties,
+}) => {
+  const t = useTranslation();
+  const [pagination, setPagination] = useState<Pagination>({
+    page: 0,
+    first: 20,
+    offset: 0,
+  });
+  // Could filter here for only properties that are used in import
```
lache-melvin, line 31: Nah I think it's nice to see the blank columns for the things you left blank in the CSV!

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
lache-melvin: Really like being able to review the properties! I reckon while we have a defined set of properties, this is great... perhaps if we get to the time where there are loads of them, we can do as you commented above and filter columns to only those that have changed? But I don't think scrolling to review is the end of the world in this case 🤷‍♀️

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+    });
+  }
+
+  const rowsWithProperties = rows.map(row => {
+    return { ...row, ...row.properties };
+  });
+
+  const columns = useColumns<ImportRow>(columnDescriptions, {}, []);
+
+  const filteredEquipment = rowsWithProperties.filter(row => {
```
lache-melvin, line 80: filteredFacilities?

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+      (row.errorMessage && row.errorMessage.includes(searchString)) ||
+      row.id === searchString
+    );
+  });
+  const currentFacilitiesPage = filteredEquipment.slice(
+    pagination.offset,
+    pagination.offset + pagination.first
+  );
+
+  // console.log('currentFacilitiesPage', currentFacilitiesPage);
```
lache-melvin, line 96: remove

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+        onChangePage={page => {
+          setPagination({
+            first: pagination.first,
+            offset: pagination.first * page,
+            page: page,
+          });
+        }}
+        columns={columns}
+        data={currentFacilitiesPage}
+        noDataElement={<NothingHere body={t('error.asset-not-found')} />}
```
lache-melvin, line 127: no facilities?

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+                importErrorRows.push({
+                  ...facility,
+                  errorMessage: err.message,
+                });
+              });
+          })
+        ).then(() => {
+          // Update Progress Bar
+          const percentComplete =
+            10 - (remainingRecords.length / numberImportRecords) * 100.0;
```
lache-melvin, line 110:
```suggestion
            100 - (remainingRecords.length / numberImportRecords) * 100.0;
```

To make the percentage complete?

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+          activeStep={activeStep}
+          onClickStep={onClickStep}
+        ></ClickableStepper>
+        {errorMessage ? <Alert severity="error">{errorMessage}</Alert> : null}
+        <TabContext value={currentTab}>
+          <Grid container flex={1} flexDirection="column" gap={1}>
+            <Grid item display="flex">
+              <Box flex={1} flexBasis="40%"></Box>
+              <Box flex={1} flexBasis="60%"></Box>
+            </Grid>
```
lache-melvin, line 213: What are these empty boxes doing?

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
lache-melvin: I think these are the main ones you'd want

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
```diff
@@ -0,0 +1,199 @@
...
+  setErrorMessage,
+  setFacilityProperties,
+  onUploadComplete,
+  properties,
+}) => {
+  const t = useTranslation();
+  const { error } = useNotification();
+  const [isLoading, setIsLoading] = useState(false);
+  const FacilityPropertyBuffer: ImportRow[] = [];
+  // TODO filter name properties for facility properties?
```
lache-melvin, line 56: It's fine for now, as we only use name properties for facility properties... I think either remove comment or mention that its a future improvement if we start using name properties in more places?

client/packages/system/src/Name/ListView/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+  ImportRow,
+  LineNumber,
+} from './ImportProperties/PropertiesImportModal';
+
+export const importFacilitiesPropertiesToCsv = (
+  facilities: Partial<ImportRow & LineNumber>[],
+  t: TypedTFunction<LocaleKey>,
+  properties?: string[]
+) => {
+  // TODO maybe don't need the facilities[0] fallback.
```
lache-melvin, line 13: Yeah I would remove... when you call this you're always passing in an array anyway! Also, should this method be called `exportFacilit...`?

lache-melvin (changes requested):
Hey Ferg! Looking great 🔥 Main thing is that the import/export needs to use the full facility list (not paginated) - plus a few clean up things, I think from copy paste 😁

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
```diff
@@ -0,0 +1,199 @@
...
+          <InlineProgress variant={'indeterminate'} color={'secondary'} />
+        </Grid>
+      ) : null}
+      <Stack spacing={2}>
+        <Upload onUpload={csvImport} />
+        <Typography textAlign="center">
+          {t('messages.template-download-text')}
+          <Link onClick={csvExample} to={''}>
+            {t('heading.download-example')}
+          </Link>
```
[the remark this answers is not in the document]
fergie-nz: oh yes much better, thank you!

client/packages/system/src/Name/ListView/ImportProperties/UploadTab.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const FacilityPropertyBuffer: ImportRow[] = [];
+  // TODO filter name properties for facility properties?
+  const propertyNodes: PropertyNode[] | undefined = properties
+    ?.map(property => {
+      return { ...property.property };
+    })
+    .sort();
+
+  const csvExample = async () => {
+    const facilityRows: ImportRow[] = facilities
```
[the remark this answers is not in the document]
fergie-nz: I wonder if changes I've made are OK to rely on default rows per page (which is giving me 500 facilities), or if I should use a custom listAllFacilities call. Do you think worth while writing a listAll query for this case?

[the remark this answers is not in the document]
lache-melvin: Yeah go with the listAll - I'm actually not sure how many stores a typical mSupply setup might have, but we should always be able to expect that all facilities are exported/imported again!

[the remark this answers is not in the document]
fergie-nz: Spoke to Mark and he reckoned new end point in back end to grab everything! I'll make one of those

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
fergie-nz: This one is tricky and I wonder if needs a larger refactor.

It looks like the table is basing width off the parent page rather than the modal, so doesn't think it needs a scroll.

I've added a width override, but I wonder if we need a better way to handle table width when on modals

client/packages/system/src/Name/ListView/ImportProperties/PropertiesImportModal.tsx
fergie-nz: !!! they've been carried around throughout all of our import modals

lache-melvin (changes requested):

lache-melvin:
Looking good Ferg! Main concern is just the table in modal display, should be able to view this on a tablet 🙏

client/packages/system/src/Name/ListView/Facilities/AppBarButtons.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => {
+  const t = useTranslation();
+
+  return (
+    <AppBarButtonsPortal>
+      <Grid container gap={1}>
+        <ButtonWithIcon
+          Icon={<UploadIcon />}
+          label={t('button.import-properties')}
+          onClick={importModalController.toggleOn}
```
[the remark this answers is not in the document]
fergie-nz: Ah nice great point about the UX.

I like your snack idea - I'll make one of those if no properties configured

client/packages/system/src/Name/ListView/ImportProperties/ImportReviewDataTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+            offset: pagination.first * page,
+            page: page,
+          });
+        }}
+        columns={columns}
+        data={currentFacilitiesPage}
+        noDataElement={<NothingHere body={t('error.facility-not-found')} />}
+        id="facilities' properties review table"
+        overflowX="auto"
+        width={1500}
```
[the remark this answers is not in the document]
fergie-nz: Have fixed it but the manual setting of the overall modal width is still at 1600.

I wonder if this is OK or if we have a better way to handle dynamic modal widths?

It doesn't look good when I try to mock it up on a small screen

client/packages/system/src/Name/ListView/Facilities/AppBarButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+  properties: NamePropertyNode[] | undefined;
+  propertiesLoading: boolean;
+}) => {
+  const t = useTranslation();
+  const { error } = useNotification();
+
+  const handleClick = () => {
+    properties?.length && properties.length >= 0
+      ? importModalController.toggleOn()
+      : error(t('error.no-properties-to-import'))();
```
lache-melvin, line 28: minor lol but could it be info() rather than error(), so its the blue snack? no properties doesn't mean anything has gone wrong, they just aren't enabled...

lache-melvin (approved):
~~I had a play... for some reason if you set a really low height (100px) it works?? But not sure why????? Wonder if you can have any breakthroughs from [here](https://github.com/msupply-foundation/open-msupply/compare/4085-add-import-store-custom-fields-via-csv...4085-add-import-store-custom-fields-via-csv(heights-wip)#diff-6f098b955cadedc43a5880054eab764bffde8b672b5437732ab4948db20f7a31R188) haha~~


EDIT: Let's just go with a [hard coded height](https://github.com/msupply-foundation/open-msupply/compare/4085-add-import-store-custom-fields-via-csv...4085-add-import-store-custom-fields-via-csv(hard-code-height)) and get it working!

Approving for once we get in some kind of fix for the table height 😁 


Also, the store code column is super narrow, and the store name... Not sure if you made the columns narrow to try fit them all across, but I think better to accept it will need to scroll and have adequate width for them 👍 Might also be nice to truncate the supply interval one so it doesn't take up so much height? 🙏 

Screenshot 2024-07-01 at 11 05 10 AM

lache-melvin:
Oh also noting here, I think important to have good docs around this, possibly even a little something on the import tab... given that the other imports we have are for importing NEW assets etc, I think it'll be easy for people to assume they can import new stores here.

fergie-nz:
> noting here, I think important to have good docs around this, possibly even a little something on the import tab... given that the other imports we have are for importing NEW assets etc, I think it'll be easy for people to assume they can import new

Nice! OK I'll make an issue for that now

PR #4242, opened 2024-06-23
4221 cannot save blank lines

### LINKED ISSUE #4221: Demographics - can save blank lines
## What went wrong? 😲

Some strange things going on with blank lines:

- Can save the first one
- Adding more blank lines - now get `already exists` error
- When you add some blanks and some named demographics, the blanks are filtered/removed on save
- If only blanks, they remain on save (but not refresh)
- And second video, you can add multiple blank demographics if you add them in one go

https://github.com/msupply-foundation/open-msupply/assets/55115239/a39d1600-821b-41d8-b0de-90664cfe8735

https://github.com/msupply-foundation/open-msupply/assets/55115239/9b9ced7e-60e1-4a64-a73c-35c651e44300

## Expected behaviour 🤔

Blank lines are consistently filtered out on save

## How to Reproduce 🔨

1. See video

## Your environment 🌱

- Open mSupply Version: 2.1 RC1
- Legacy mSupply Central Server Version: 7.15

- Platform: Browser

- Database type: PG

fergie-nz:
Fixes #4221 

# 👩🏻‍💻 What does this PR do?

Adds another error to handle in back end for undefined input for name
Also a little refactor on front end to reduce chance of prompting errors. First won't set draft to dirty on creation of new rows, and also cleans up empty string or spaces to undefined which will throw error in the back end on save.

I kind of want to catch all errors and make better snacks for users when adding multiple rows. For example if some succeed and some fail, you get both types of snack which could be confusing. Maybe something to look at for a new issue if it is important UX.

# 🧪 Testing

- [ ] open demographics from central server
- [ ] create new row and see that saving is still disabled
- [ ] edit some part of the row but leave name as either blank or spaces
- [ ] try saving and see that error 'indicatorHasNoName' error thrown

# 📃 Documentation

  1.
  2.


mark-prins:
Hi Ferg - the issue has a milestone of 2.1 - can you rebase the PR please to target the release branch 🙏

fergie-nz:
> Hi Ferg - the issue has a milestone of 2.1 - can you rebase the PR please to target the release branch 🙏

Thanks Mark! Will do the same with 4243 also

client/packages/system/src/IndicatorsDemographics/DetailView/utils.tsx
```diff
@@ -1,18 +1,36 @@
...
+  UpdateDemographicIndicatorInput,
+} from '@common/types';
+
+export const toInsertIndicator = (
+  row: Row,
+  indexPopulation?: number
+): InsertDemographicIndicatorInput => {
+  return {
+    id: row.id,
+    name: row.name?.trim() === '' ? undefined : row.name,
```
fergie-nz, line 17: will send name as undefined if it is empty spaces, or empty string, or was undefined to begin with

client/packages/system/src/IndicatorsDemographics/api/api.ts
```diff
@@ -36,38 +36,6 @@ const Parsers = {
     };
     return fields[sortBy.key] ?? DemographicProjectionSortFieldInput.Id;
   },
-  toInsertIndicator: (
```
fergie-nz, line 39: No longer need these parsers because we generate the types as the insertInput and updateInput rather than as a fragment on create

server/service/src/demographic/insert_demographic_indicator.rs
```diff
@@ -56,8 +57,17 @@ pub fn validate(
     input: &InsertDemographicIndicator,
     connection: &StorageConnection,
 ) -> Result<(), InsertDemographicIndicatorError> {
-    if !check_year_name_combination_unique(&input.name, input.base_year, None, connection)? {
-        return Err(InsertDemographicIndicatorError::DemographicIndicatorAlreadyExistsForThisYear);
+    match &input.name {
```
fergie-nz, line 60: I don't really like combining validations because it reads confusing and isn't modular, but this seemed to work the best.

fergie-nz:
I'm a little concerned that this is still showing changes from develop after I thought those changes would have been merged in with https://github.com/msupply-foundation/open-msupply/pull/4243/files

Will investigate tomorrow morning!

mark-prins (commented):
The process could be better. I did this to begin with:
* Added a blank line and save is disabled
* entered a value for percentage, save is enabled (probably shouldn't be as the line isn't valid. seems a bit tricky tho)
* clicked save and got a success toast, and an error toast, and the new row disappears

The part that's annoying as a user (this user, anyway) is that the new line is removed. I'd like it to remain, and a single toast saying 'error', and perhaps highlight the error field 🤷 

here's a quick suggestion - which 
* shows the error only if there is a problem (not the success)
* doesn't refresh the data if there's an error
* highlights the name field
* 'translates' the server error (real dodgy)

https://github.com/msupply-foundation/open-msupply/compare/4221-cannot-save-blank-lines...4221-cannot-save-blank-lines(suggestion)

mark-prins (commented):
Looks great Ferg! happy with the changes overall - would just like to have something like my suggestion in place. and maybe prevent updating to a blank name as well?

I wouldn't worry about the stray changes too much, but maybe we could discuss tomorrow?

fergie-nz:
> The process could be better. I did this to begin with:
> 
> * Added a blank line and save is disabled
> * entered a value for percentage, save is enabled (probably shouldn't be as the line isn't valid. seems a bit tricky tho)
> * clicked save and got a success toast, and an error toast, and the new row disappears
> 
> The part that's annoying as a user (this user, anyway) is that the new line is removed. I'd like it to remain, and a single toast saying 'error', and perhaps highlight the error field 🤷
> 
> here's a quick suggestion - which
> 
> * shows the error only if there is a problem (not the success)
> * doesn't refresh the data if there's an error
> * highlights the name field
> * 'translates' the server error (real dodgy)
> 
> [4221-cannot-save-blank-lines...4221-cannot-save-blank-lines(suggestion)](https://github.com/msupply-foundation/open-msupply/compare/4221-cannot-save-blank-lines...4221-cannot-save-blank-lines(suggestion))
> 
> The result is this:
> 
>  Screen.Recording.2024-06-24.at.5.57.10.PM.mov 
> Note that there's still an issue, where you can clear an existing name and save. this blanks the name in the db 🤔

That is awesome, thanks Mark I love it.

PR #4256, opened 2024-06-24
4254 Cant login issue

### LINKED ISSUE #4254: Cant login issue
## What went wrong? 😲

Cannot log in to remote sites

## Expected behaviour 🤔

Should be able to log in

## How to Reproduce 🔨

Use this test data file: 
[simple sync site 2.4DD.zip](https://github.com/user-attachments/files/15950276/simple.sync.site.2.4DD.zip)

You can use username 'config' to do the registration etc.. (in og mSupply)

Initialise central omSupply site, with 'central'/'pass', log in with 'test'/'pass', go do admin, then configuration and 'initialise', make changes to 'test' facility data.

Now try to initialise remote site with 'test'/'pass', log in with 'tests'/'pass', should see error

## Problem and suggested fix

I think the issue is with having record_id unique in sync_buffer and om_name_fields together with name table having the same record_id (name arrives from og central, added to sync buffer, and then om_name_fields arrive from om central, and overwrite og central name in sync_buffer, then store is unable to integrate due to FK constraint).

Suggest to add table_name to record_id in sync buffer here: 

https://github.com/msupply-foundation/open-msupply/blob/1027722f6c47dd3187f6892207c300ef10b096de/server/service/src/sync/api/common_records.rs#L90-L91

## Your environment 🌱

2.1.0-RC2, both in dev env

roxy-dao:
Fixes #4254

# 👩🏻‍💻 What does this PR do?
Append table name to record id and parse correct id to translate

/o/ \o\ /o/ \o\

# 🧪 Testing

- [ ] Initialise a OMS central server 
- [ ] In `OMS Central Server` -> Settings -> Configuration -> Initialise 
- [ ] Initialise remote server
- [ ] Should be able to login

(Also make sure tests are still okay)

# 📃 Documentation

  1.
  2.


server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -100,6 +100,12 @@ pub enum ChangelogTableName {
     NameOmsFields,
 }
 
+impl fmt::Display for ChangelogTableName {
```
roxy-dao, line 103: Needed so I can stringify the enum to split it 👀

lache-melvin:
Still can't initialise/getting a bazzilion FK errors... name records still getting overwritten somewhere?

lache-melvin (approved):
🔥 🙏 🎉

server/service/src/sync/api/common_records.rs
```diff
@@ -78,12 +78,17 @@ impl CommonSyncRecord {
...
-        } else {
-            record_id
+        let record_id = match action {
+            SyncAction::Merge => {
+                // This is (likely) a temporary fix to avoid merge sync records overriding upserts on target table
+                // causing errors as the merge is applied to record that never got upserted first.
+                uuid()
+            }
+            _ if table_name == *"name_oms_fields" => {
+                // apprehend table name to record_id to avoid name override
```
lache-melvin, line 88: Though should this be `append` lol

roxy-dao: I snuck it in 😭 
spelling hard

craigdrown: @lache-melvin  - ya missed "override" - should be "overwrite"?  ;-)
While I'm here - is table_name a new field in the sync buffer? Is it indexed? If yes, lots of indexed fields in the sync buffer will make it slow to append records to, and I have feels that this is a future performance bottleneck in large systems?

jmbrunskill: @craigdrown, Indexes are on Action, Integration Time, and Integration Error (not sure what the error one is used for...)
record_id is implicitly indexed due being to Primary Key

andreievg: Added to an issue: https://github.com/msupply-foundation/open-msupply/issues/4266

PR #4257, opened 2024-06-24
4067 view stock 'cancel/close' button disabled

### LINKED ISSUE #4067: View stock: 'Cancel/Close' button disabled
## What went wrong? 😲

On 'View stock' > 'Details'  the 'Cancel/close' button is disabled, unless you try to make a change.
It can be confusing for users, you shouldn't have to make a change to then cancel and close a window.

Screenshot 2024-06-07 at 12 56 24 PM

On other parts of the system this button is always enabled. 

## Expected behaviour 🤔

The button should be enabled all the time, this way users can go back to the list without going back to the navigation bar.
This is how it already works on other parts of OMs like Outbound shipments, Inbound shipments, etc..

## Your environment 🌱

- Open mSupply Version: v2.1.0-IA-test-2
- Legacy mSupply Central Server Version: V7.16

- Platform: Desktop app

roxy-dao:
Fixes #4067

# 👩🏻‍💻 What does this PR do?
Change button to `Close`. If stock line has been edited, prompt user that they will lose all their data if they continue or redirect back to Stock page.

# 🧪 Testing

- [ ] Go to `Stock`
- [ ] Click on stock line
- [ ] Click close -> Should redirect back
- [ ] Follow step 1 & 2
- [ ] Edit stock line
- [ ] Click Close
- [ ] See prompt

# 📃 Documentation

  1. Screenshots for Stock page
  2.

### DIFF
--- a/client/packages/system/src/Stock/DetailView/Footer.tsx
+++ b/client/packages/system/src/Stock/DetailView/Footer.tsx
@@ -5,6 +5,7 @@ import {
   AppFooterPortal,
   DialogButton,
   LoadingButton,
+  useBreadcrumbs,
 } from '@openmsupply-client/common';
 import { FormInputData } from '@openmsupply-client/programs';
 
@@ -15,6 +16,7 @@ interface FooterProps {
   inputData?: FormInputData;
   showSaveConfirmation: () => void;
   showCancelConfirmation: () => void;
+  isDirty: boolean;
 }
 
 export const Footer: FC<FooterProps> = ({
@@ -23,8 +25,10 @@ export const Footer: FC<FooterProps> = ({
   inputData,
   showSaveConfirmation,
   showCancelConfirmation,
+  isDirty,
 }) => {
   const t = useTranslation();
+  const { navigateUpOne } = useBreadcrumbs();
 
   return (
     <AppFooterPortal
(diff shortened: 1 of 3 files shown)


client/packages/system/src/Stock/DetailView/Footer.tsx
```diff
@@ -44,9 +48,10 @@ export const Footer: FC<FooterProps> = ({
             marginLeft="auto"
           >
             <DialogButton
-              variant="cancel"
-              disabled={disabled}
-              onClick={() => showCancelConfirmation()}
+              variant="close"
+              onClick={() =>
+                isDirty ? showCancelConfirmation() : navigateUpOne()
```
mark-prins, line 53: with the change - can you please add the `navigateUpOne` to the `onConfirm` of the cancel confirmation modal. The button now says 'close' but when you click it, effectively nothing happens. The `resetDraft` call doesn't revert my changes and I'm still on the same page!

Side note: the resetDraft should be doing something, right? and with this change it can actually be removed - we don't have the pattern of resetting form state anywhere else in the app

```
  const onCancel = () => {
    resetDraft();
    navigateUpOne();
  };

  const showCancelConfirmation = useConfirmationModal({
    onConfirm: onCancel,
    message: t('messages.confirm-cancel-generic'),
    title: t('heading.are-you-sure'),
  });
```

mark-prins (changes requested):
Nice - thanks! Can you tweak it please before shipping, the close button confirmation doesn't 'close' when you confirm, which feels weird to me when using.

client/packages/system/src/Stock/DetailView/Footer.tsx
roxy-dao: The resetDraft resets draft 🤣

mark-prins: buut, I edited, clicked the button and my changes didn't get reset 🤷

mark-prins (approved):
LGTM 🚢

PR #4274, opened 2024-06-25
3994 Don't display foreign currency total if foreign currency id is the same as home currency

### LINKED ISSUE #3994: Don't display foreign currency total if foreign currency id is the same as home currency
## What went wrong? 😲
Foreign currency total is still being displayed even if it is the home currency

## Expected behaviour 🤔
Should only be displaying foreign currency total if the currency_id on invoice is actually a foreign currency

## How to Reproduce 🔨

1. Create some invoices in OG
2. Move store to OMS
3. Initialise or Sync
4. Go to the invoices created
5. See the foreign currency total is there even if its home currency

## Your environment 🌱

- Open mSupply Version: 2.1.0

- Platform:

- Database type:

roxy-dao:
Fixes #3994

# 👩🏻‍💻 What does this PR do?
Use '-' instead of foreign currency total in invoice if the foreign currency is home currency, also noticed that the rates were showing up as 0 in older invoices, so have handled that as well 👀 .... 
Screenshot 2024-06-26 at 8 20 34 AM

# 🧪 Testing

- [ ] Have `issue in foreign currency` store pref on
- [ ] Create an invoice for an external customer/supplier
- [ ] Add some lines with sell price
- [ ] See `-` for foreign currency total
- [ ] Change currency
- [ ] See total

# 📃 Documentation

  1. Screenshots with foreign currency 
  2.

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/SidePanel/PricingSection.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/SidePanel/PricingSection.tsx
@@ -46,6 +46,7 @@ export const PricingSectionComponent = () => {
   const { mutateAsync: updateTax } = useInbound.document.updateTax();
   const { c: foreignCurrency } = useCurrency(currency?.code as Currencies);
   const { store } = useAuthContext();
+  const isHomeCurrency = store?.homeCurrencyCode === currency?.code;
 
   const tax = PricingUtils.effectiveTax(
     pricing?.serviceTotalBeforeTax,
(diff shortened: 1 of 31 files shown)


client/packages/invoices/src/InboundShipment/DetailView/SidePanel/PricingSection.tsx
```diff
@@ -165,14 +166,16 @@ export const PricingSectionComponent = () => {
...
+          <PanelField>{currencyRate === 0 ? 1 : currencyRate}</PanelField>
         </PanelRow>
         <PanelRow>
           <PanelLabel>{t('heading.total')}</PanelLabel>
           <PanelField>
-            {foreignCurrency(
-              pricing.foreignCurrencyTotalAfterTax ?? 0
-            ).format()}
+            {isHomeCurrency
+              ? '-'
```
mark-prins, line 175: side thought: this has been used in a number of places and I wonder if we should create a global const for `UNDEFINED_STRING_VALUE` or similar?

mark-prins (approved):
🚢  lovely - thanks Roxy!

client/packages/invoices/src/InboundShipment/DetailView/SidePanel/PricingSection.tsx
roxy-dao: wasn't sure where to put it... so just created a file under common -> utils

PR #4275, opened 2024-06-25
4024 stock lines not included in search

### LINKED ISSUE #4024: Some stock lines aren't included in search results
## What went wrong? 😲

I have some `Salbutamol` stock lines, as introduced via Inbound Shipment from another store, where `Salbutamol` is visible:
Screenshot 2024-05-31 at 5 32 17 PM

But when I use the search:

Screenshot 2024-05-31 at 5 32 25 PM

Similarly... I can't issue it anywhere?

Screenshot 2024-05-31 at 5 34 38 PM

Also can't add it to a stocktake? (Unless you create the stocktake with the `Items with stock` filter - then it appears)

## Expected behaviour 🤔

I should be able to search across all my stock lines 👀 

Actually not sure what the intended behaviour is if you get sent an item that isn't visible for you? Probably wouldn't happen too often because you would get what you requested, but still...

Maybe you shouldn't receive that inbound line at all??

## How to Reproduce 🔨

1. From one store, create an outbound shipment to another store
2. Add an item to the shipment that is visible in your store, but not visible in the receiving store
3. Set to shipped
4. Switch to receiving store and set the inbound shipment to delivered
5. In view stock page: new stock line is added, for item you shouldn't be able to see
6. Search for the item - you can't see it

## Your environment 🌱

- Open mSupply Version: 2.1

- Platform:

- Database type:

## Context

For the stock list view at least, it would appear this is the problem: we apply item visibility filtering only if `item_name_or_code` filter is set? I think it should either happen all or none of the time? Though feels if we care about item visibility we shouldn't be able to receive that stock line at all...

https://github.com/msupply-foundation/open-msupply/blob/30f80f49a32beddda2b28fdddfc17643ec22118d/server/repository/src/db_diesel/stock_line.rs#L227-L248

## Agreed solution

Similar to how visibility by master lists works, another view that checks if any item lines with stock exist, and join it to item and use as filter

lache-melvin:
Fixes #4024

# 👩🏻‍💻 What does this PR do?

Makes non-visible items available in various places if you have some stock on hand for that item.

Honestly uncertain about the composition of the filters - typically the filter object is composed as an AND filter, so its a bit funky here to use it as an OR... especially because I can imagine wanting an AND version in future (give me the visible items that have stock on hand.) I tried to make it clear what was happening with comments/naming, but could I do something different?

Per [this comment](https://github.com/msupply-foundation/open-msupply/issues/4024#issuecomment-2185564264) in the issue, I'm not 100% sure where we should/shouldn't include the "invisible items with stock" throughout the app, hasn't been clarified yet, but thought I'd get the implementation in place, using/removing should be trivial. Let me know if you have thoughts on where these items should be seen?

i.e. maybe you actually shouldn't be able to issue items that are invisible to you? But you can see them in stock list/stock take? should you be able to see them in the item list (because you have them) or not because its more of a catalogue list than current state?

# 🧪 Testing

- [ ] Have two stores using two different master lists, with some different items in those master lists
- [ ] From one store, make an Outbound Shipment to the other store, including an item that is not in the master list for that other store (i.e. not visible in the other store)
- [ ] Receive that Inbound Shipment of the invisible item in the receiving store
- [x] Item List: you can now see that item
- [x] View stock: you can now see that stock line - also test that you can still see that item when you use the `item name or code` search! (this was the original bug)
- [x] Outbound shipment: you can issue the item
- [ ] Prescription: you can issue the item
- [x] Outbound return: you can return the item
- [x] Stocktake: you can add this item in your stocktakes
- [ ] internal order: you cannot request this item
- [x] Inbound return: you cannot add this item to an inbound return
- [x] Inbound shipment: you cannot add this item to an inbound shipment

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/item.rs
+++ b/server/repository/src/db_diesel/item.rs
@@ -97,6 +97,11 @@ impl ItemFilter {
         self.is_vaccine = Some(value);
         self
     }
+
+    pub fn has_stock_on_hand(mut self, value: bool) -> Self {
+        self.has_stock_on_hand = Some(value);
+        self
+    }
 }
 
 type ItemAndUnit = (ItemRow, Option<UnitRow>);
@@ -306,8 +311,9 @@ mod tests {
         test_db, EqualFilter, ItemFilter, ItemLinkRowRepository, ItemRepository, ItemRow,
         ItemRowRepository, ItemType, MasterListLineRow, MasterListLineRowRepository,
         MasterListNameJoinRepository, MasterListNameJoinRow, MasterListRow,
-        MasterListRowRepository, NameRow, NameRowRepository, Pagination, StoreRow,
-        StoreRowRepository, StringFilter, DEFAULT_PAGINATION_LIMIT,
+        MasterListRowRepository, NameRow, NameRowRepository, Pagination, StockLineRow,
+        StockLineRowRepository, StoreRow, StoreRowRepository, StringFilter,
+        DEFAULT_PAGINATION_LIMIT,
     };
 
     use super::{Item, ItemSort, ItemSortField};
@@ -662,6 +668,39 @@ mod tests {
             )
             .unwrap();
         assert_eq!(results.len(), 2);
+
+        // Test has_stock_on_hand filter
+
+        // Add stock for item 3 (which is invisible)
+        StockLineRowRepository::new(&storage_connection)
+            .upsert_one(&StockLineRow {
+                id: "stock_line_for_item_3".to_string(),
+                item_link_id: "item3".to_string(),
+                store_id: "name1_store".to_string(),
+                available_number_of_packs: 5.0,
+                pack_size: 1.0,
+                ..Default::default()
+            })
+            .unwrap();
+
+        // get visible rows + non visible rows with stock on hand
+        let results = ItemRepository::new(&storage_connection)
+            .query(
+                Pagination::new(),
+                Some(ItemFilter::new().is_visible(true).has_stock_on_hand(true)),
+                None,
+                Some("name1_store".to_string()),
+            )
+            .unwrap();
+
+        // item 1 & 2 == visible, item 3 == has stock
+        assert_eq!(
+            results
+                .into_iter()
+                .map(|r| r.item_row.id)
+                .collect::<Vec<String>>(),
+            vec!["item1", "item2", "item3"]
+        );
     }
 
     #[actix_rt::test]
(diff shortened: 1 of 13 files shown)


server/repository/src/db_diesel/item.rs
```diff
@@ -239,10 +247,36 @@ fn create_filtered_query(store_id: String, filter: Option<ItemFilter>) -> BoxedI
...
-            Some(false) => query.filter(item_dsl::id.ne_all(visible_item_ids)),
-            None => query,
+        let item_ids_with_stock_on_hand = item_link_dsl::item_link
+            .select(item_link_dsl::item_id)
+            .inner_join(stock_on_hand_dsl::stock_on_hand)
+            .filter(stock_on_hand_dsl::available_stock_on_hand.gt(0.0))
+            .group_by(item_link_dsl::item_id)
+            .into_boxed();
+
+        query = match (is_visible, has_stock_on_hand) {
```
mark-prins, line 257: bit of a nit pick - some weird cases aren't covered.
If I want `is_visible = Some(true) && has_stock_on_hand = Some(false)` then I will potentially get items which aren't visible and do have stock on hand.
Similarly the 'false' 'false' case is giving me items that aren't visible but do have stock

mark-prins (approved):
👍 That's great!
I have a couple of minor comments: that the item filter can potentially return slightly unwanted results, and that the internal order test didn't pass; I am seeing my 'invisible' item when creating an internal order

lache-melvin:
@mark-prins strange, I didn't have a failing internal order test? And CI tests passed? Anyway, I tried to simplify the filter a bit, hopefully more obvious what it does... could I get your thoughts on it before I merge?

server/repository/src/db_diesel/item.rs
```diff
@@ -239,10 +248,28 @@ fn create_filtered_query(store_id: String, filter: Option<ItemFilter>) -> BoxedI
             .filter(store_dsl::id.eq(store_id.clone()))
             .into_boxed();
 
-        query = match is_visible {
-            Some(true) => query.filter(item_dsl::id.eq_any(visible_item_ids)),
-            Some(false) => query.filter(item_dsl::id.ne_all(visible_item_ids)),
-            None => query,
+        let item_ids_with_stock_on_hand = item_link_dsl::item_link
```
mark-prins, line 251: very slick! that's a nice solution - makes it clearer for the consumer too.

mark-prins (approved):
Yep - like it! thanks 🙏

PR #4277, opened 2024-06-25
4264 demographic growth rates not saving

### LINKED ISSUE #4264: iSC/GAPS/Cold Chain: funky Demographic Growth number behaviour
## What went wrong? 😲

I am unsure of how to describe this behaviour, but it looks like when saving growth percentage, the numbers don't immediately reflect the change? For example: 

* Growth is set to be 10%, but the population numbers stay the same
* After navigating away and returning, the growth number reverts back to 0%
* After setting growth to be 10% again, changes took effect
* After navigating away and returning, the growth number again reverts back to 0% but the population numbers show accurately

https://github.com/msupply-foundation/open-msupply/assets/153777484/5e7f6bb5-d424-476f-ae29-4aaa2e38c859

## Expected behaviour 🤔

If growth % is entered, it should be saved and remain the same regardless if you navigate away from the page or not. Similarly, population numbers should increase by the right number - if it's 10% growth from a million people, the new number should be 1,100,000. 

## How to Reproduce 🔨

1. Go to Manage -> Demographics
2. Set your population to a certain number (mine was 1,000,000)
3. Play around with the growth percentage fields - set them to 10%, 0%, negative amount
4. Navigate out and back to the Demographics page to see the changes in-between
5. Note the inconsistencies 

## Your environment 🌱

- Open mSupply Version: V2.1.0-RC3
- Legacy mSupply Central Server Version: V7.15.05

- Platform: Desktop App on Windows

- Database type: SQLite3

lache-melvin:
Fixes #4264

# 👩🏻‍💻 What does this PR do?

Demographics projections (growth rates) weren't being invalidated after save. On the first time you created the demographics, if you hit the save button twice or more before refreshing, it wouldn't know to update - it would try to do another insert (resulting in an error)

This PR invalidates the projections query after upsert of demographics, so second save does an update correctly!

# 🧪 Testing

- [ ] Reset DB/ensure `demographic_projection` table is empty
- [ ] On central server, go to manage > demographics
- [ ] edit growth rates/add some indicators/edit population - click save multiple times
- [ ] TEST: every save says saved successfully, you do not see the `DemographicProjectitonAlreadyExistsForBaseYear` error
- [ ] Refresh the page/navigate away and come back - indicators and growth rates are all still up to date

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
+++ b/client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
@@ -4,6 +4,7 @@ import {
   Box,
   ColumnAlign,
   DataTable,
+  DateUtils,
   RecordPatch,
   TableProvider,
   createTableStore,
@@ -46,7 +47,7 @@ const IndicatorsDemographicsComponent = () => {
   const t = useTranslation();
 
   const { draft, setDraft } = useDemographicData.indicator.list(headerDraft);
-  const baseYear = headerDraft?.baseYear ?? 2024;
+  const baseYear = headerDraft?.baseYear ?? DateUtils.getCurrentYear();
 
   const { data: projection, isLoading: isLoadingProjection } =
     useDemographicData.projection.get(baseYear);
--- a/client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
+++ b/client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
@@ -46,14 +46,25 @@ const IndicatorsDemographicsComponent = () => {
   const t = useTranslation();
 
   const { draft, setDraft } = useDemographicData.indicator.list(headerDraft);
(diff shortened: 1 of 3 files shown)


client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
```diff
@@ -46,14 +46,25 @@ const IndicatorsDemographicsComponent = () => {
   const t = useTranslation();
 
   const { draft, setDraft } = useDemographicData.indicator.list(headerDraft);
+  const baseYear = headerDraft?.baseYear ?? 2024;
```
lache-melvin, line 49: Do you think we should be defaulting this to something like `new Date().getFullYear()` ?

mark-prins: yes, much better - thanks

mark-prins (approved):
👍  working nicely!
good find - that's much better thanks

PR #4282, opened 2024-06-26
3738 transfer comment and reference updates

### LINKED ISSUE #3738: Comments and reference field values don't sync in stock transfer
## What went wrong? 😲
 
Comments and reference field values don't sync in stock transfer

## Expected behaviour 🤔
ALong with quantities, comments & references should also sync?

## How to Reproduce 🔨
1. Setup at least 2 oms-stores
2. Create an outbound from one storeA to storeB with comments(eg: commentABC) and references(eg: refABC)
3. Just put status upto `picked` and sync
4. On storeB, you can already see that shipment under Inbounds with `picked` status
5. You cannot edit yet but can see the previous comments and references all
6. Now in storeA, make changes in comments (change it to sth like commentXYZ), quantities & references (change it tom sth like refXYZ) and confirm status to `shipped` now, and sync
7. In storeB, you can finally see that shipment as editable
8. See error: the old comments(commentABC) & references(refABC) are there instead of updated ones, only the quantities are updated-ones

## Your environment 🌱

- Open mSupply Version: v2.0.0-rc2
- Legacy mSupply Central Server Version: v7-14-00

- Platform: oms-central on server exe, remote-oms on server + browser (chrome)

- Database type: sqlite

lache-melvin:
Fixes #3738

# 👩🏻‍💻 What does this PR do?

Ended up fixing a few things!

The issue: comments/references weren't being transferred on update (only on insert). I.e. if you set these things before setting outbound to picked, they were transferred, but if you update them between picked and shipped, the changes do not map through. I also updated transport reference and tax percentage to transfer through on updates as well.

THEN I found a few bugs along the way, which honestly, I'm not sure how we didn't find them before??

- Service charge lines have no cost price per pack, nor pack size... so when running `convert_invoice_to_single_pack`, we would end up with a `NaN` and transfers stopped working completely?!
- Transferred service charge totals were showing as 0, becuase they were being calculated as `cost_price_per_pack * number_of_packs` ... both of which were 0 😁 

So added some logic around service lines 😅 

Would appreciate a double check on the logic/if I've transferred anything that we actually don't want to be? I don't think I have but 🙏 

# 🧪 Testing

- [ ] Add a comment/reference/transport reference/some service charge etc
- [ ] Set to picked
- [ ] Log in to receiving store
- [x] All the above mapped through correctly
- [ ] Go back to sending store, update all those fields to something else
- [ ] Set to shipped
- [ ] Go back to receiving store
- [x] All your changes are reflected

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/processors/transfer/invoice/test.rs
+++ b/server/service/src/processors/transfer/invoice/test.rs
@@ -876,6 +876,7 @@ impl InvoiceTransferTester {
                 &ctx,
                 inline_init(|r: &mut UpdateOutboundShipment| {
                     r.id.clone_from(&self.outbound_shipment.id);
+                    r.their_reference = Some("some updated reference".to_string());
                     r.status = Some(UpdateOutboundShipmentStatus::Shipped);
                 }),
             )
@@ -896,6 +897,8 @@ impl InvoiceTransferTester {
             inline_edit(&inbound_shipment, |mut r| {
                 r.status = InvoiceStatus::Shipped;
                 r.shipped_datetime = self.outbound_shipment.shipped_datetime;
+                r.their_reference =
+                    Some("From invoice number: 20 (some updated reference)".to_string());
                 r
             })
         );
--- a/server/service/src/processors/transfer/invoice/test.rs
+++ b/server/service/src/processors/transfer/invoice/test.rs
@@ -59,6 +59,10 @@ async fn invoice_transfers() {
         r.id = uuid();
     });
 
+    let service_item = inline_init(|r: &mut ItemRow| {
+        r.id = uuid();
+    });
+
     let site_id_settings = inline_init(|r: &mut KeyValueStoreRow| {
         r.id = KeyType::SettingsSyncSiteId;
         r.value_int = Some(site_id);
@@ -79,7 +83,7 @@ async fn invoice_transfers() {
         inline_init(|r: &mut MockData| {
             r.names = vec![inbound_store_name.clone(), outbound_store_name.clone()];
             r.stores = vec![inbound_store.clone(), outbound_store.clone()];
-            r.items = vec![item1.clone(), item2.clone()];
+            r.items = vec![item1.clone(), item2.clone(), service_item.clone()];
             r.key_value_store_rows = vec![site_id_settings];
         }),
     )
@@ -93,6 +97,7 @@ async fn invoice_transfers() {
         outbound_store_name,
         item1,
         item2,
+        service_item,
     );
 
     let number_of_instances = 6;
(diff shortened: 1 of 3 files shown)


server/service/src/processors/transfer/invoice/test.rs
```diff
@@ -59,6 +59,10 @@ async fn invoice_transfers() {
         r.id = uuid();
     });
 
+    let service_item = inline_init(|r: &mut ItemRow| {
```
lache-melvin, line 62: Added a service line to the transfer tests, hopefully will help catch anything funky going forward 😁

server/service/src/processors/transfer/invoice/test.rs
```diff
@@ -1269,24 +1317,41 @@ fn check_line(connection: &StorageConnection, inbound_id: &str, outbound_line: &
...
-    assert_eq!(inbound_line.r#type, InvoiceLineType::StockIn);
+
+    match outbound_line.r#type {
+        InvoiceLineType::Service => {
+            assert_eq!(inbound_line.r#type, InvoiceLineType::Service);
+            assert_eq!(
+                inbound_line.total_before_tax,
+                outbound_line.total_before_tax
+            );
+            assert_approx_eq::assert_approx_eq!(
```
lache-melvin, line 1328: Test was failing will `110.0 != 110.00000000000001` ... in DB the value looks correct... is assert_approx ok here or should I have some rounding in the transfer logic?

server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
```diff
@@ -93,10 +93,41 @@ impl InvoiceTransferProcessor for UpdateInboundInvoiceProcessor {
             invoice_line_repository.upsert_one(line)?;
         }
 
+        let outbound_invoice_row = &outbound_invoice.invoice_row;
+
+        let formatted_ref = match &outbound_invoice_row.their_reference {
+            Some(reference) => format!(
+                "From invoice number: {} ({})",
```
mark-prins, line 100: the returning nagging thought that we need to have translatable text in the backend, somehow

server/service/src/processors/transfer/invoice/test.rs
mark-prins: I think approx is fine - trying to approximate floating point numbers in binary 🤔 
If we change, I'd update `calculate_total_after_tax` and have that round the numbers slightly. But I'd prefer not to - just have dodgy floats in the database and round in the UI layer.

Though I would also do this:

```
total_after_tax: calculate_total_after_tax(total_before_tax, tax_percentage),
```

in `/server/service/src/processors/transfer/invoice/common.rs:65` to make the calculation consistent.

mark-prins (approved):
👍 lovely! thanks - that works nicely!

PR #4302, opened 2024-06-27
4284 export asset match import

### LINKED ISSUE #4284: iSC/Cold Chain: when importing a list of Cold chain equipment, the template does not match an exported list
## What went wrong? 😲

Me and @AnushaUp noticed, that when you export a list of Cold Chain Equipment under `Cold chain -> Equipment -> Export`, the file template is not the same we need for import. When importing that same file, we run into the following error: 

photo_2024-06-26 16 35 59

Seems like **Store** and **Catalogue** columns are removed while exporting the lists.

## Expected behaviour 🤔

It would be helpful to have the export and import in identical formats :)

## How to Reproduce 🔨

1. Go to Cold chain -> Equipment -> Export
2. Save the file
3. Go to Cold chain -> Equipment -> Import
4. Try importing the file you just saved without changing anything
5. See error

## Your environment 🌱

- Open mSupply Version: V2.1.0-RC3
- Legacy mSupply Central Server Version: V7.15.05

- Platform: Desktop app for Windows

- Database type: SQLite3

lache-melvin:
Fixes #4284

# 👩🏻‍💻 What does this PR do?

Both the `download an example` csv and the `export` csv file shapes now work map to the import structure (neither were before 😅) 

Though I suppose the export didn't necessarily need to match, as that data already exists, importing again results in duplicate errors. But at least its the same shape of the data everywhere.

# 🧪 Testing

- [ ] Coldchain -> equipment -> import
- [ ] click download template
- [ ] reupload that template
- [ ] it works (albeit dates won't parse)
- [ ] on list view, click export
- [ ] try to import that
- [ ] you can view the data in the table (though import will fail on duplicate asset numbers)

# 📃 Documentation

  1.
  2.


client/packages/coldchain/src/Equipment/ImportAsset/ImportReviewDataTable.tsx
```diff
@@ -52,27 +52,38 @@ export const ImportReviewDataTable: FC<ImportReviewDataTableProps> = ({
...
+    },
+    {
+      key: 'installationDate',
+      width: 100,
+      sortable: false,
+      label: 'label.installation-date',
+      Cell: TooltipTextCell,
+    },
+    {
+      key: 'replacementDate',
```
lache-melvin, line 72: replacement date was in the export, figured it may as well also be in the import 🤷‍♀️

client/packages/coldchain/src/Equipment/utils.ts
```diff
@@ -11,6 +11,7 @@ function baseAssetFields(t: TypedTFunction<LocaleKey>) {
   return [
     'id',
     t('label.asset-number'),
+    t('label.catalogue-item-code'),
```
lache-melvin, line 14: actual fixes are in this file - the order of these arrays matter, but they were in all sorts of orders so weren't mapping correctly 😁

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -155,11 +155,6 @@ function getImportHelpers<T, P>(
     const prop = t(localeKey) as keyof P;
     const value = row[prop] ?? '';
     if (value === undefined || (value as string).trim() === '') {
-      rowErrors.push(
```
lache-melvin, line 158: Thoughts here please!

Removes mandatory requirement of catalogue item code, and store code if on central... docs say these things are required, but when you create individually via the modal, you don't need to assign a store nor create from catalogue... so why do we mandate it here?

client/packages/coldchain/src/Equipment/utils.ts
```diff
@@ -78,31 +79,38 @@ export const importEquipmentToCsvWithErrors = (
...
       node.assetNumber,
       node.catalogueItemCode,
     ];
     if (isCentralServer) mapped.push(node.store?.code);
-    mapped.push(node.notes);
-    mapped.push(node.serialNumber);
-    mapped.push(node.installationDate);
-    mapped.push(node.lineNumber);
-    mapped.push(node.errorMessage);
+    mapped.push(
```
mark-prins, line 106: not exporting the properties columns in the 'withError' csv?

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
mark-prins: we did have the store code mandatory on central, though that has changed and we now have 'unassigned' equipment, so the store change is valid.

for the catalogue item - we'd require a type to be assigned, if the catalogue item isn't used.

( the allowance of non-catalogue items was another change part way through development, hence the functional mismatch here )

mark-prins (approved):
👍 looks good thanks! with a couple of caveats : the category item code is mandatory if you don't have the type specified and we're not handling that in the import are we?

I was expecting the properties to be shown when exporting the 'import with errors' and the standard 'export this list of stuff' button. but I'm fine with them not being in either of those

PR #4326, opened 2024-07-01
2118 breadcrumb refactor

### LINKED ISSUE #2118: Breadcrumb refactor
A breadcrumb can currently have some urlParts and a suffix. The urlParts are taken from the current browser location while the suffix can be set be a component to add another Breadcrumb or other component.  In some places this is currently used to add an additional breadcrumb + another "label" component, i.e. its not only one suffix.

Find way to make this a bit nicer, e.g.:
- have another `appendBreadcrumb` method
- OR accept multiple suffixes in a `setSuffixes` method

For more details please see the original discussion: https://github.com/openmsupply/open-msupply/pull/2085#discussion_r1295291575

### LINKED ISSUE #4063: Add customisable nested breadcrumbs

We want to be able to customise nested breadcrumbs values to give users more context of where they are when going through nested pages.

For new Programs route we go to:

Immunization programs page
-> Immunization Program detail page (which includes a list of Vaccine Courses)
    -> Vaccine course detail view

Breadcrumbs look like this:
Image 07-06-2024 at 10 06

Where 'id' would be a UUID and could be confusing for users.

Allow breadcrumbs to customise nested values.

### Moneyworks Jobcode 🧰

Could be under CCEIDEV as this is only being used currently in cold chain module.
Or maybe a more generic code?
I am unsure.

### LINKED ISSUE #4226: Breadcrumbs show the UUID rather than asset name
## What went wrong? 😲

Slowed down for your viewing pleasure... it does get updated, and then overwritten!

https://github.com/msupply-foundation/open-msupply/assets/55115239/707ecb0a-8adc-44e4-a7ae-0597228fc6df

## Expected behaviour 🤔

Asset/Immunizations name stays in the breadcrumb!

## How to Reproduce 🔨

1. Go to Coldchain > Equipment
2. Click on an asset to go to the detail view
3. See breadcrumb: UUID!

## Your environment 🌱

- Open mSupply Version: 2.1 RC1

- Platform: browser (chrome)

- Database type:

lache-melvin:
Fixes #2118 

Fixes #4063

Fixes #4226

# 👩🏻‍💻 What does this PR do?

Replaces breadcrumb `suffix`/`setSuffix` with `setBreadcrumbRenderers` , which allows a renderer to be provided for any index in the breadcrumb.

So now we can do this! Though there is now a PR to turn vaccine details into a modal 👀 

Screenshot 2024-07-01 at 3 30 12 PM

Also fixes the suffix sometimes clearing problem.

Generally curious for opinions on the API, could a different pattern be nicer?

# 🧪 Testing

- [ ] Go to programs > immunisation program > [detail] > [vaccine course detail]
- [x] Both the program and vaccine course names are in the breadcrumb
- [ ] Go to coldchain > equipment > [detail view] and refresh the page
- [x] Breadcrumb shows asset name, not ID
- [ ] Go to view stock > [detail view]
- [ ] Click on log tab, then back to detail tab
- [x] Breadcrumb still shows item name, not ID
- [x] General click around of pages, breadcrumb looks right 😅 

# 📃 Documentation

  1. Vaccine course screenshots
  2.


client/packages/system/src/ContactTrace/DetailView/DetailView.tsx
```diff
@@ -38,7 +38,7 @@ export const DetailView: FC<DetailViewProps> = ({
 }) => {
   const t = useTranslation('dispensary');
   const navigate = useNavigate();
-  const { setSuffix, urlParts } = useBreadcrumbs([AppRoute.ContactTrace]);
```
lache-melvin, line 41: Only need to pass in `topLevelRoutes` ... like for the Sync and Settings pages, not needed here

client/packages/system/src/ContactTrace/DetailView/DetailView.tsx
```diff
@@ -125,27 +125,29 @@ export const DetailView: FC<DetailViewProps> = ({
...
-              contactData.patient?.firstName,
-              contactData.patient?.lastName
-            )}
-          </Breadcrumb>
-          <span>{` / ${contactData.programName} - ${dateFormat.localisedDate(
-            contactData.documentData.datetime
-          )}`}</span>
-        </span>
-      );
+      setBreadcrumbRenderers({
```
lache-melvin, line 128: This pattern to render an additional breadcrumb part seems not great, but a little out of scope?

client/packages/system/src/Immunisation/VaccineCourseView/VaccineCourseView.tsx
```diff
@@ -104,6 +105,9 @@ export const VaccineCourseView: FC = () => {
     query: { data, isLoading },
     isDirty,
   } = useVaccineCourse(id);
+  const {
+    query: { data: programData },
+  } = useImmunisationProgram(t, data?.programId ?? '');
```
lache-melvin, line 110: wondering if it might be nicer to return the program name on the vaccine course node??

client/packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
```diff
@@ -11,49 +11,65 @@ export interface UrlPart {
 }
 
 type BreadcrumbState = {
-  setSuffix: (suffix?: string | React.ReactElement) => void;
-  suffix?: string | React.ReactElement;
+  setBreadcrumbRenderers: (
+    renderers?: Record<number, (part: UrlPart) => string | React.ReactElement>
```
jmbrunskill, line 15:
```suggestion
    renderers?: Record<part: string, (part: UrlPart) => string | React.ReactElement>
```
Could we do something like this?

So that you call it with...
`setBreadcrumbRenderers({ data?.id: () => data?.assetNumber ?? '' });`
or
`setBreadcrumbRenderers({ 'assets': () => t('label.assets')});

To me it's not intuative where index starts? Seems like it's an index from the end currently?
So we want to use 1 for the last element, 2 for the second to last?

I'd be tempted to just accept reactNode rather than renderer function too?

Depends if we want a single render which reacts to different values, but if we'd still need to specify all the indexes currently already, were you imagining passing in a render that would take an ID, so an API look up to determine the right label for that element?

Maybe something like this would be nicer? 
`setBreadcrumbRenderers({ data?.id: data?.assetNumber ?? '' });`
`setCustomBreadcrumbs({ 'assets': t('label.assets')});`

client/packages/system/src/Immunisation/VaccineCourseView/VaccineCourseView.tsx
jmbrunskill: Probably would be cleaner, I'm fine with this though...

client/packages/system/src/ContactTrace/DetailView/DetailView.tsx
jmbrunskill: I agree it might be worth rolling this in, doesn't go nicely with my url part lookup idea though...

client/packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
mark-prins: yeah, that's nice!
personally, I prefer thinking in terms of indices rather than static values - so I would keep with the current implementation. I also like the suggestion to use a react node rather than a renderer, though if you do that you lose the ability to respond to a part, actually fine in practice as we have no need for this; all the same its kinda neat to have a function which takes the url part and provides an appropriate response. that's more flexible.

mark-prins (approved):
🚢 that's great - three issues in one 😄 
I like the approach, and even though I'm worried it might be a little confusing to use for some later devs, it's clear enough I think.

client/packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
lache-melvin: Thanks for input! Yeah I like the just providing a value rather than a function, that should make it a little less unwieldy. 

Yeah I thought accepting the part and being able to do something with it would be helpful, but I think in any case where we want that, we still have access to the URL parts, can do the logic and then pass that into `setCustomBreadcrumbs` 👀 

I think I'll stick with index - potentially undefined object keys feels like asking for trouble in JS 🤣 Though I do like how easy it is to tell what you are replacing!

jmbrunskill:
I'm ok to merge this how it is but did have another idea which might  worth considering.

What about just `setBreadcrumbs(BreadCrumbRenderer[])`?

Would be clear what you're doing then from the name, overriding the whole breadcrumb display.
It would solve the question of how we can add an extra part, remove parts, etc and avoid confusion about indexes to breadcrumbs vs url parts etc?

If we could make a nice way to get the default behaviour I think it could be a nicer and more intuative design.

```
type BreadCrumbRenderer = (urlParts[], breadCrumbIndex) => (part: UrlPart) => string | React.ReactElement;
```

`setBreadcrumbs([defaultBreadCrumbRender,()=>'My Custom Breadcrumb', ()=> 'My Custom breadcumb with no url part');`

@lache-melvin @mark-prins  thoughts?
I assume `defaultBreadcrumbRender` could use the index and url parts to continue the existing behaviour.

lache-melvin:
> I'm ok to merge this how it is but did have another idea which might worth considering.
> 
> What about just `setBreadcrumbs(BreadCrumbRenderer[])`?
> 
> Would be clear what you're doing then from the name, overriding the whole breadcrumb display. It would solve the question of how we can add an extra part, remove parts, etc and avoid confusion about indexes to breadcrumbs vs url parts etc?
> 
> If we could make a nice way to get the default behaviour I think it could be a nicer and more intuative design.
> 
> ```
> type BreadCrumbRenderer = (urlParts[], breadCrumbIndex: number) => string | React.ReactElement;
> ```
> 
> `setBreadcrumbs([defaultBreadCrumbRender,()=>'My Custom Breadcrumb', ()=> 'My Custom breadcumb with no url part');`
> 
> @lache-melvin @mark-prins thoughts? I assume `defaultBreadcrumbRender` could use the index and url parts to continue the existing behaviour.

Thanks James! I like this interface! I think there's a bit more to consider here, as there was a bit of state we were managing based on the URL parts which I'm not sure how we would use/not use... I think I'll merge this to get these issues closed for 2.1 but would like to come back to it in another refactor 🔥

PR #4328, opened 2024-07-01
3947 stock line changes from stocktake not captured in activity log

### LINKED ISSUE #3947: Stock line changes from stocktake not captured in activity log
## What went wrong? 😲

If you edit stock line info (e.g. expiry date, cost price...) from the `View Stock` page, editing the stock line directly, then each of those changes are captured in the log tab:

Screenshot 2024-05-20 at 4 53 42 PM

However, there is also the ability to edit this information during a stocktake. These are not recorded... 

I updated the batch name for this stock line via stocktake, no log:
Screenshot 2024-05-20 at 5 03 11 PM

Note that when the stock level/number of packs is changed, we do have a trail of that from the ledger tab. This is just for changes to the additional stock line info. 

## Expected behaviour 🤔

Changes to stock line information are logged, regardless of where the changes are made.

## How to Reproduce 🔨

1. Create a new stocktake with at least one line
2. Click on that stocktake line to open the edit modal
3. Change some data for that stock line (e.g. change the batch name)
4. Click OK, then finalise the stocktake
5. In the `View stock` page, find the stock line you updated
6. The batch name should be updated
7. Go to the log tab for that stock line
8. BUG: there is no log for the change

## Your environment 🌱

- Open mSupply Version: 2.1.0

- Platform: 

- Database type:

roxy-dao:
Fixes #3947

# 👩🏻‍💻 What does this PR do?
Logs stocktake changes to stock line... also found a couple bugs along the way. I did the changes on finalised since the stocktake lines get pruned when they are finalised and any changes to stock line that hasn't been counted will be reverted....

The stock line doesn't get invalidated when updating stocktake...

# 🧪 Testing

- [ ] Create a stocktake
- [ ] Change the details of a stock line (must count stock as well)
- [ ] Finalise
- [ ] Go to Stock Log (Should see changes have been logged)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/icons/BarChart2.tsx
+++ b/client/packages/common/src/ui/icons/BarChart2.tsx
@@ -7,7 +7,7 @@ export const BarChart2Icon = (props: SvgIconProps): JSX.Element => {
       {...props}
       viewBox="0 0 24 24"
       stroke="currentColor"
-      stroke-width="2"
+      strokeWidth="2"
     >
       <line x1="18" y1="20" x2="18" y2="10"></line>
       <line x1="12" y1="20" x2="12" y2="4"></line>
--- a/server/service/src/stocktake/update.rs
+++ b/server/service/src/stocktake/update.rs
@@ -242,7 +242,7 @@ fn generate_update_inventory_adjustment_reason(
 
 /// Returns new stock line and matching invoice line
 fn generate_stock_in_out_or_update(
-    connection: &StorageConnection,
+    ctx: &ServiceContext,
     store_id: &str,
     inventory_addition_id: &str,
     inventory_reduction_id: &str,
@@ -302,8 +302,10 @@ fn generate_stock_in_out_or_update(
 
     let update_inventory_adjustment_reason = generate_update_inventory_adjustment_reason(
         invoice_line_id.clone(),
-        row.inventory_adjustment_reason_id,
+        row.inventory_adjustment_reason_id.clone(),
     );
+
+    log_stock_changes(ctx, stock_line_row.clone(), row.clone())?;
 
     let stock_in_or_out_line = if delta > 0.0 {
         StockChange::StockIn(InsertStockInLine {
@@ -339,8 +341,8 @@ fn generate_stock_in_out_or_update(
             batch: row.batch,
             pack_size: row.pack_size,
             expiry_date: row.expiry_date,
-            cost_price_per_pack: None,
-            sell_price_per_pack: None,
+            cost_price_per_pack: Some(cost_price_per_pack),
+            sell_price_per_pack: Some(sell_price_per_pack),
             total_before_tax: None,
             tax_percentage: None,
         })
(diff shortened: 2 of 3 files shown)


client/packages/common/src/ui/icons/BarChart2.tsx
```diff
@@ -7,7 +7,7 @@ export const BarChart2Icon = (props: SvgIconProps): JSX.Element => {
       {...props}
       viewBox="0 0 24 24"
       stroke="currentColor"
-      stroke-width="2"
+      strokeWidth="2"
```
roxy-dao, line 10: Side quest: Wrong casing 👀

server/service/src/stocktake/update.rs
```diff
@@ -339,16 +341,16 @@ fn generate_stock_in_out_or_update(
             batch: row.batch,
             pack_size: row.pack_size,
             expiry_date: row.expiry_date,
-            cost_price_per_pack: None,
-            sell_price_per_pack: None,
+            cost_price_per_pack: Some(cost_price_per_pack),
```
roxy-dao, line 344: Not sure why the cost and sell prices weren't being updated for stock out? If the stock line hasn't been zero'd then the cost and sell prices should still update?

server/service/src/stocktake/update.rs
```diff
@@ -361,6 +363,82 @@ fn generate_stock_in_out_or_update(
     })
 }
 
+fn log_stock_changes(
```
roxy-dao, line 366: This file is getting really big.... Would really like a refactor into generate.rs/validate.rs files....

lache-melvin: 10000000% agree yeehaw

server/service/src/stocktake/update.rs
```diff
@@ -302,9 +302,11 @@ fn generate_stock_in_out_or_update(
 
     let update_inventory_adjustment_reason = generate_update_inventory_adjustment_reason(
         invoice_line_id.clone(),
-        row.inventory_adjustment_reason_id,
+        row.inventory_adjustment_reason_id.clone(),
     );
 
+    log_stock_changes(ctx, stock_line_row.clone(), row.clone())?;
```
lache-melvin, line 308: This needs to move up, above the check for `delta == 0.0`

Currently only saves logs if there is an inventory adjustment, but should work if counted # packs is the same 😁

lache-melvin (changes requested):
🔥 Thank you! Just need to capture logs for stock lines where counted packs is same as snapshot!

server/service/src/stocktake/update.rs
roxy-dao: [bceb341](https://github.com/msupply-foundation/open-msupply/pull/4328/commits/bceb341509cb37783750bca27fe060443c719b46)

lache-melvin (approved):
Beaut TY

PR #4329, opened 2024-07-01
4281 Inbound shipment: tax percentage is being displayed as the tax value

### LINKED ISSUE #4281: Inbound shipment: tax percentage is being displayed as the tax value
## What went wrong? 😲

When tax is 10%, it shows as $10:

Screenshot 2024-06-26 at 1 53 10 PM

## Expected behaviour 🤔

should show $30, not $10!

## How to Reproduce 🔨

1. create an Inbound shipment
2. Add a line with a cost price per pack, so you end up with a stock total
3. add tax to total
4. See tax total is the percentage number, not the calculated amount of tax to pay!

## Your environment 🌱

- Open mSupply Version: 2.1 RC3

- Platform:

- Database type:

roxy-dao:
Fixes #4281

# 👩🏻‍💻 What does this PR do?
Use correct tax amount in Inbound Shipment
Screenshot 2024-07-01 at 7 51 09 PM

# 🧪 Testing

- [ ] Create an Inbound Shipment
- [ ] Add lines with some cost price 
- [ ] Add tax
- [ ] Check tax amount is correct

# 📃 Documentation

  1. Inbound/Outbound screenshots that show tax input
  2.


client/packages/common/src/intl/locales/en/distribution.json
```diff
@@ -35,7 +35,6 @@
   "heading.supply-to-requested": "Set supply to requested",
   "heading.supply-to-approved": "Set supply to approved",
   "heading.tax": "Tax",
-  "heading.edit-tax-rate": "Edit tax rate percentage",
```
lache-melvin, line 38: mad nit but I think if tax rate is a common component, the translation should be in common rather than replenishment 👀

lache-melvin (approved):
Splendid TY

client/packages/common/src/intl/locales/en/distribution.json
roxy-dao: I was thinking about that...

PR #4330, opened 2024-07-01
4280 tax displays incorrectly on totals below $1 in outbound shipments

### LINKED ISSUE #4280: Tax displays incorrectly on totals below $1 in outbound shipments
## What went wrong? 😲

Seems fairly edge case (maybe less so for service charges?) but still a bit random...

If sub-total is less that $1, the tax percentage display gets confused - seems to show the whatever the tax value as a percentage, rather than the tax percentage?

In this case, I set tax to 50% for both the service charge and the items, which you can see the totals are calculated correctly, but the percentages are then 40% and 10%:

Screenshot 2024-06-26 at 12 02 11 PM

When you open the tax edit modal, it correctly shows `50` before refreshing the page, then it shows incorrectly (e.g. 10).

## Expected behaviour 🤔

Tax percentage displays the tax percentage 👀 

## How to Reproduce 🔨

1. Go to Outbound Shipment
2. Adds some items and/or service charges where total adds to less than a dollar
3. Add tax percentage
4. See tax percentage displayed is not the tax percentage you entered

## Your environment 🌱

- Open mSupply Version: 2.1 RC

- Platform:

- Database type:

roxy-dao:
Fixes #4280

# 👩🏻‍💻 What does this PR do?
Use tax percentage from invoice to calculate stock tax since we don't do stock tax by line... and also refactor effective tax to return correct tax percentage for service charges.
Screenshot 2024-07-01 at 8 46 39 PM
Screenshot 2024-07-01 at 8 46 23 PM

# 🧪 Testing

- [ ] Add lines with sell price less than $1
- [ ] Add tax
- [ ] Should calculate tax correctly

- [ ] Have service line in master list for store
- [ ] Add multiple service charges with different tax percentages
- [ ] Tax percentage and total should calculate correctly

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/utils/pricing/PricingUtils.ts
+++ b/client/packages/common/src/utils/pricing/PricingUtils.ts
@@ -5,8 +5,9 @@ export const PricingUtils = {
     return Math.max(total - subtotal, 0);
   },
   effectiveTax: (subtotal: number, total: number) => {
+    const taxAmount = PricingUtils.taxAmount(subtotal, total);
     return (
-      Math.round(((total - subtotal) / subtotal) * 100 * ROUNDING_PRECISION) /
+      Math.round((taxAmount / (subtotal || 1)) * 100 * ROUNDING_PRECISION) /
       ROUNDING_PRECISION
     );
   },
--- a/client/packages/common/src/utils/pricing/PricingUtils.ts
+++ b/client/packages/common/src/utils/pricing/PricingUtils.ts
@@ -5,11 +5,9 @@ export const PricingUtils = {
     return Math.max(total - subtotal, 0);
   },
   effectiveTax: (subtotal: number, total: number) => {
-    const taxAmount = PricingUtils.taxAmount(subtotal, total);
-    const x =
-      Math.round(
-        (taxAmount / Math.max(subtotal, 1)) * 100 * ROUNDING_PRECISION
-      ) / ROUNDING_PRECISION;
-    return x;
+    return (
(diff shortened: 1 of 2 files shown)


client/packages/common/src/utils/pricing/PricingUtils.ts
```diff
@@ -5,11 +5,9 @@ export const PricingUtils = {
...
-    const taxAmount = PricingUtils.taxAmount(subtotal, total);
-    const x =
-      Math.round(
-        (taxAmount / Math.max(subtotal, 1)) * 100 * ROUNDING_PRECISION
-      ) / ROUNDING_PRECISION;
-    return x;
+    return (
+      Math.round(((total - subtotal) / subtotal) * 100 * ROUNDING_PRECISION) /
+      ROUNDING_PRECISION
+    );
```
lache-melvin, line 11: I think this must be doing something spicy? 

When no service charges:
Screenshot 2024-07-02 at 11 46 09 AM

Once you add service charges, above looks fine, but tax edit modal does this:
Screenshot 2024-07-02 at 11 48 02 AM

lache-melvin (changes requested):
Prices below $1 look good, but getting NaNs now 😁

client/packages/common/src/utils/pricing/PricingUtils.ts
roxy-dao: Using Math.max for subtotal is returning some wacky numbers even though it returns 1 for undefined 👀

lache-melvin (approved):
Yay stunning thank you 🚢

PR #4359, opened 2024-07-03
2663 Allow to enter decimal in issue quantity for Requisitions

### LINKED ISSUE #2663: Allow to enter decimal in issue quantity for Requisitions
## What went wrong? 😲
 
In OG mSupply, we can issue stocks in decimal. Maybe we'll soon need the same feature in oms as well, since ordering (internal order) is allowed in both units and packs. In case oms receives a requisition (in units), then it'll be difficult to complete that request in a single shipment since currently, there's neither any feature to supply in units nor decimal values allowed in issue quantity. 
## Expected behaviour 🤔
Maybe simply allow decimal entry in issue quantity like the OG mSupply? 
(or you wanna add units/packs toggle like the one in IO window?)

## How to Reproduce 🔨
1. First of all, turn the `pack to one` pref OFF for all the stores you're playing/testing with (for ease)
2. To replicate this specific issue, have a setup with both desktop and oms stores
3. From desktop, issue some units (by placing decimal values in CI) to oms-store so that oms-store now has decimal valued pack sizes, sth like this below:
image

4. Also make internal order to oms-store in such a way decimal value have to be entered from oms-side (see above recording for reference)
5. After sync, in oms-store, try to fulfill that request
6. See error: you simply can't fulfill that request in a single transaction! [A reference to what I mean in this step is below recording: here, allocating simply doesn't work! At some point, I saw the system crash as well in this step, but not crashing every single time]

https://github.com/msupply-foundation/open-msupply/assets/59193490/5ca24dea-c968-49e6-a268-308983e30a77

## Additional Information [need to investigate]
Also, throghout this, I've come across some random crashings while trying to access Internal Order (in the same oms-site where decimal pack Qty were introduced after above steps) and one also while processing the requisition (in received reqistions>Supply requested>Create outbound shipment>crash!) 

**Note**: This additional one is not replicable every single time with same steps, so will need to investigate further

## Your environment 🌱

 - Version: v1.5.4; v2.0.0-rc5 exe with mSupply v7-14-01
 - Platform: server exe + browser(chrome)

roxy-dao:
Fixes #2663

# 👩🏻‍💻 What does this PR do?
Allows decimal input if the line is being requested in packs instead of units. Currently set at 2 dp.

# 🧪 Testing

- [ ] Add a line that has a default pack size other than 1
- [ ] Try order in decimal 

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
@@ -265,6 +265,7 @@ export const RequestLineEditForm = ({
                       requestedQuantity / currentItem.defaultPackSize,
                       2
                     )}
+                    decimalLimit={2}
                     width={100}
                     onChange={quantity => {
                       update({


lache-melvin (approved):
🔥

PR #4371, opened 2024-07-04
3444 Move create_missing_master_list_and_program into a migration 

### LINKED ISSUE #3444: Move create_missing_master_list_and_program into a migration
## The suggestion

On server startup we run this code everytime
```
    // CREATE MISSING MASTER LIST AND PROGRAM
    service_provider
        .general_service
        .create_missing_master_list_and_program(&service_provider)
        .unwrap();
```
This upserts 3 rows with `MISSING_PROGRAM` for `master_list`, `context` and `program`

If we move this to a migration you'll save 3 sql queries every startup, and reduce the logic in the server crate, which makes the server code easier to follow.

Seems like a low risk change, but we need to check if this approach was taken for a particular reason?

Maybe 1 hour?

roxy-dao:
Fixes #3444

# 👩🏻‍💻 What does this PR do?
Move missing_program and system_user creation to migration

# 🧪 Testing

- [ ] Initialise V2.1.
- [ ] Change to this branch and upgrade

2.
- [ ] Run app
- [ ] Should work

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_02_00/create_missing_master_list_and_program.rs
```diff
@@ -0,0 +1,23 @@
+use crate::{migrations::sql, StorageConnection};
+use util::constants::MISSING_PROGRAM;
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    sql!(
+        connection,
+        r#"
+            INSERT INTO master_list (id, name, code, description, is_active)
+            SELECT '{MISSING_PROGRAM}', '{MISSING_PROGRAM}', '{MISSING_PROGRAM}', '{MISSING_PROGRAM}', false
+            WHERE NOT EXISTS (SELECT 1 FROM master_list WHERE id = '{MISSING_PROGRAM}');
```
andreievg, line 10: that really cool

andreievg (approved):
Thanks looks great

Issue #4069, opened 2024-06-07
KeepAwake not available on initial load


lache-melvin:
## What went wrong? 😲

Screen goes to sleep during intialisation, requiring `Retry`.

#4048 raised an issue where the Initialisation screen couldn't be typed in on android, because the KeepAwake plugin throwing a "not available" error. 

Workaround hotfix was introduced in #4050, catching/swallowing the error and moving on. However, we shouldn't have this problem in the first place! 

(See #4048 & #4050 for more details)

It's working fine in 2.0, but in 2.1 the error arises. This needs investigation - I did a little, and couldn't find any of our code changes in that vicinity recently, but needed to get a working build going so didn't spend too much time.

Now the screen goes to sleep during initialisation again, which stops the process & shows an error (you have to press Retry and it continues)  - not a great regression 😭 

## Expected behaviour 🤔

KeepAwake works without errors on initial load of the initialisation screen.

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Run debug version of APK
2. Start initialising
3. Screen will go to sleep
4. You have to click `Retry` for it to continue

OR for original issue 

1. Back out changes or use previous commit than #4050
1. Run debug version of APK
1. On app start, select run in server mode
1. You can't type in the input fields, and `Initialise` button says `Retry`

## Your environment 🌱

- Open mSupply Version: 2.1
- Legacy mSupply Central Server Version:

- Platform: Android

- Database type:

lache-melvin:
![image](https://github.com/msupply-foundation/open-msupply/assets/55115239/14e4f008-a34b-4cae-a794-51725135dcfc)

This is the error that @regotaina is getting sometimes, when the tablet goes to sleep. I think it's related? Hopefully this looks familiar to anyone who was around before the KeepAwake was introduced? 🙏

mark-prins:
That doesn't look familiar sorry - though it seems logical that this is due to the app sleeping and closing all connections

andreievg:
Triage, potential candidate for hotfix, would recommend high priority for 2.2, I think it would still be ok as initialisation should still happen in the background so even though it's likely, it's not critical, with work around

kahn-msupply:
@regotaina made me aware of this issue - one thing we can do is change MeMDM settings for timeout of the display - maybe just for the initialisation stage. Tested with @regotaina's tablet and the screen stays on.

mark-prins:
[...]

PR #4377, opened 2024-07-05
Mount capacitor js to web view onLoadResource

### LINKED ISSUE #4373: Can't use barcode scanner for Outbound Shipments or update item details (Android)
## What went wrong? 😲

While using V2.1.RC6 .apk,  @louisaw123 and I noticed the barcode scanning buttons are not available at:

- Outbound shipments: 'scan' button to add item lines
- Stock > View Stock: 'scan' icon to add batch and expiry date are not working

It is available on V2.1, which prob leads us to think it disappeared somewhere in between these versions.

- Open mSupply Version: V2.1.RC6
- Platform: Android

andreievg:
Fixes #4373

# 👩🏻‍💻 What does this PR do?

I found that native capacitor plugins were not always available, sometimes they were there and sometimes not. On tablet with dev options on and usb debugging plugged in I could inspect app while it was running with chrome using 'chrome://inspect'. Managed to see a pattern, when adding [console.log here](https://github.com/msupply-foundation/open-msupply/blob/f8dacaf0164d5e8c3442ff54067870ae40ecf19d/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java#L66), and it seem when the log was showing after bugsnag was mounted, web app was not able to recognise native functionality, but when the log was showing before bugsnag mounted log, native functionality worked as expected. This lead me to thing that capacitor JS injection was happening late, which was confirmed by the fix.

No 100% happy about it all, can discuss future improvement on this front a bit later, but I think this change is safe (relative to what is currently happening

We had to move to using evaluateJavascript because existing capacitor 'injection', that happens here:

https://github.com/ionic-team/capacitor/blob/f4e7f19f186e334404b2cd0decc3205e57bf4469/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java#L372-L381

Relies on below method which fails on self signed certs:

https://github.com/ionic-team/capacitor/blob/f4e7f19f186e334404b2cd0decc3205e57bf4469/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java#L348

Some extra material: https://github.com/ionic-team/capacitor/discussions/6166

# 🧪 Testing

On tablet open stock view for one batch, next to barcode you should see a scan icon. This icon is sometimes missing (if you restart and start app multiple times you should see it, for me it happens quite oftern)

### DIFF
--- a/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
+++ b/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
@@ -29,7 +29,7 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
     // and plugin definitions etc is not injected
 
     // Using onLoadResource vs onPageStarted because sometimes onPageStarted is too late
-    // but with onLoadResource we need to make sure that injection happens only one in a web view
+    // but with onLoadResource we need to make sure that injection happens only once in a web view
     @Override
     public void onLoadResource(WebView webView, 
             String url) {
@@ -79,7 +79,7 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
 
             // .post to run on UI thread
             webView.post(() -> {
-                // To only run once on a page we check for existence ot nativeBridge
+                // To only run once on a page we check for existence of nativeBridge
                 // which is instantiated by above scripts
                 webView.evaluateJavascript("typeof nativeBridge", new ValueCallback<String>() {
                     @Override
--- a/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
+++ b/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
@@ -1,6 +1,7 @@
 package org.openmsupply.client;
 
 import android.graphics.Bitmap;
+import android.webkit.ValueCallback;
 import android.webkit.WebView;
 
 import com.getcapacitor.Bridge;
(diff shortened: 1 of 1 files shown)


andreievg:
I think fix would solve this issue: https://github.com/msupply-foundation/open-msupply/issues/4069

client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
```diff
@@ -71,9 +77,20 @@ public void onPageStarted(WebView webView,
                     cordovaPluginsJS +
                     "\n\n";
 
-            // From JSInjector.getScriptString()
             // .post to run on UI thread
-            webView.post(() -> webView.evaluateJavascript(fullScript, null));
+            webView.post(() -> {
+                // To only run once on a page we check for existence ot nativeBridge
```
jmbrunskill, line 82:
```suggestion
                // To only run once on a page we check for existence of nativeBridge
```

client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
```diff
@@ -26,10 +27,14 @@ public ExtendedWebViewClient(Bridge bridge) {
     // but since it manually uses net.URL to fetch the content of request, this
     // fails for self signed certificates
     // and plugin definitions etc is not injected
+
+    // Using onLoadResource vs onPageStarted because sometimes onPageStarted is too late
+    // but with onLoadResource we need to make sure that injection happens only one in a web view
```
jmbrunskill, line 32:
```suggestion
    // but with onLoadResource we need to make sure that injection happens only once in a web view
```

jmbrunskill (approved):
Looks legit to me, haven't got any experience in this area though.
Nice find!

andreievg:
Thanks @jmbrunskill for review !

PR #4387, opened 2024-07-08
4161 Hardcode filtering out of patients from names query

### LINKED ISSUE #4161: Hardcode filtering out of patients from names query

## The suggestion

Context: #4146

Issue introduced by #4023

Couple hours?

roxy-dao:
Fixes #4161

# 👩🏻‍💻 What does this PR do?
Removes hardcoded patient filtering from the name service layer and have moved this to the repository layer, and remove is_patient filter and Patient from NameNodeType. 

Prescription insert/update now use a `check_patient_exists` for patient validation instead of the `check_other_party` function.

# 🧪 Testing

- [ ] Make sure suppliers/customers are still loaded correctly
- [ ] Make sure that prescriptions can still be created
- [ ] (dev) Call name API with graphql. Patients shouldn't show at all even without filtering.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/name.rs
+++ b/server/repository/src/db_diesel/name.rs
@@ -40,7 +40,6 @@ pub struct NameFilter {
     pub is_customer: Option<bool>,
     pub is_supplier: Option<bool>,
     pub is_donor: Option<bool>,
-    pub is_patient: Option<bool>,
     pub is_store: Option<bool>,
     pub store_code: Option<StringFilter>,
     pub is_visible: Option<bool>,
@@ -197,7 +196,6 @@ impl<'a> NameRepository<'a> {
                 address2,
                 country,
                 email,
-                is_patient,
                 code_or_name,
             } = f;
 
@@ -229,12 +227,6 @@ impl<'a> NameRepository<'a> {
 
             query = match is_donor {
                 Some(bool) => query.filter(name_dsl::is_donor.eq(bool)),
-                None => query,
-            };
-
-            query = match is_patient {
-                Some(true) => query.filter(name_dsl::type_.eq(NameType::Patient)),
-                Some(false) => query.filter(name_dsl::type_.ne(NameType::Patient)),
                 None => query,
             };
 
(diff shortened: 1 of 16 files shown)


server/repository/src/db_diesel/name.rs
```diff
@@ -176,6 +175,7 @@ impl<'a> NameRepository<'a> {
                     .left_join(store_dsl::store),
             )
             .inner_join(name_oms_fields_alias)
+            .filter(name_dsl::type_.ne(NameType::Patient))
```
roxy-dao, line 178: Filtering out all patients in repo layer

andreievg:
@roxy-dao this is where i got to with the refactor i was doing: https://github.com/msupply-foundation/open-msupply/tree/names-and-patients-separate-fitler-draft (the idea was to remove extra filtering in graphql and service layer, move them to repo layer and remove filter types that doin't apply to relevant repos like type Patient form NameRowType)

jmbrunskill (approved):
Looks good to me, tried prescriptions and all seems to work well.

Love PRs that reduce the amount of code!

roxy-dao:
> Looks good to me, tried prescriptions and all seems to work well.
> 
> Love PRs that reduce the amount of code!

How do you feel about trying changing it up a little and getting some of Andrei's refactoring in~?

jmbrunskill:
> > Looks good to me, tried prescriptions and all seems to work well.
> > Love PRs that reduce the amount of code!
> 
> How do you feel about trying changing it up a little and getting some of Andrei's refactoring in~?

I think you've achieved the goal with what you've done, but can see some advantages with Andrei's. Sorry I'd missed his comment when I first reviewed as he hadn't assigned himself, So I thought no one was looking at it. If I understand correctly Andrei's basically just cleans up the filter options so you can't send a filter type that doesn't make sense (like Name Type: `Patient`) to the `name` service. 

Personally I'm happy either way :)

andreievg (approved):
Looks great, happy for this to go through without the type changes for filters, but I wouldn't close the issue yet, and would also add to the issue:
relocate: GenderInput from Name to Patient in graphql

PR #4390, opened 2024-07-08
Android is just endlessly loading...

### LINKED ISSUE #4383: Unable to run android
## What went wrong? 😲

I tried to test #4069 now that #4377 has been implemented, and android doesn't run for me if I build and deploy the apk to a tablet. I get an infinite loader when running the 2.1 release branch.

Looking at the console I'm getting

image

when loading. Reverting the changes in #4377 and the app loads.
However, if I re-initialise then I'm still having issues

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #4383 #4069

# 👩🏻‍💻 What does this PR do?
I'm not 100% convinced of this one - it is running correctly when creating debug builds and deploying the apk, whereas the version currently in _develop_ is not running for me. So I think it's worth putting this change in. However I am not seeing _entirely_ reliable results, it's running 99% of the time correctly, and doesn't cause issues with initialisation - but sometimes the barcode scanner won't show as available, and it should always show.

The same code (develop branch) in a release build however, that is running ok. which means this isn't a high priority, I still think that it's worth implementing because
* the develop code isn't running correctly in debug
* it is (very slightly) more complex code, and any simplification is good
* there's an additional fix in here for the non-resolving promises, which causes the 'unable to initialise' issue

There's a few minor changes:
* return to use `onPageStarted` : the `onLoadResource` is called for every resource file requiring 
* add a check for embedded images in this method
* change the initialisation hook callbacks to run in an async method and use `await`

This is what I'm seeing with the current `onLoadResource` implementation:
Screenshot 2024-07-09 at 9 56 39 AM

which is the problem as described with the original implementation.

The 'unable to initialise' problem is this:

Screenshot 2024-07-08 at 3 58 13 PM

the state is `PRE_INITIALISATION', however the button says 'Retry' and the fields are all read-only. The problem here is that the callback on the promises isn't called, so the state isn't updated correctly. Changing to an async method has fixed for me.

One extra issue with `onPageStarted` is that it is called for embedded images, like this:

Screenshot 2024-07-08 at 3 28 56 PM

so I've added a check; no need to inject the plugin code in this case.

# 🧪 Testing

- [ ] Build the debug apk and deploy
- [ ] set the display timeout to the minimum (15s on samsung)
- [ ] clear storage and initialise
- [ ] the tablet shouldn't sleep while syncing
- [ ] log in, view a stock line: should have the barcode scanner available
- [ ] re-deploy and test while already initialised

# 📃 Documentation

  1.
  2.


andreievg:
> which is the problem as described with the original implementation.

Yep, it's troubling that it's popping up for you for onLoadResource, i can't seem to replicate.

Btw the issue with barcode scanning only shows after you initialise. If you initialise then close the app then open and try to navigate to page with scan button, in this branch i get missing button:

Screenshot 2024-07-09 at 1 29 42 PM

In the onLoadResource branch it seems to always show the scan button.

Maybe we need to do both onPageStart and onLoadResource and keep the check for nativeBridge undefined ?

Also i think the js string should be saved rather then being generated, then onLoadResource should be quite low impact, even if it happens on every resource load (it would only be injected if nativeBridge does not exist)

mark-prins:
I was beginning to doubt - I'd added more logging and couldn't repro.. but then rebuilt and here it is again:

image

this is injecting with `onLoadResource` only, the `window.Capacitor` object exists and has all the plugins.

will retest after merging #4374, which I've belatedly realised isn't in yet 🙄

mark-prins:
Oki - have tested with the capacitor upgrades and cannot make this version fail. 
Have tried on samsung and lenovo tablets, using Android studio debugging, a deployed debug apk, and a release apk.

With the `onLoadResource` version I was still able to repro the error, so have gone with the `onPageStart` version. It's also a simpler approach.

Can you retry with this version?

andreievg:
@mark-prins With this latest version, i am getting problem with both onLoadResource and onPageStart, pretty consistently. I would go to initialise page, not enter anything, and then quit app and then open app, and pretty much 100% of the time it would have 'Retry'. For some reason I was thinking it's due to JS string inject calculation, and after doing below change, i am constantly getting proper initialise and it's seems to be injecting timely in both onLoadResoure and onPageStart. I much prefer onPageStart. Happy to approve with onPageStart but also with below (since it seems to be consistently injecting on time, well, can't be sure 100% but 10 out of 10 for me)

https://github.com/msupply-foundation/open-msupply/compare/4383-update-from-dev...4383-update-from-dev(Pre-calculate-js-inject)

Please note above is branched from another branch from this branch (4383-update-from-dev) as I wasn't sure if there was update from dev done. So I don't think you can directly merge above diff into this branch

mark-prins:
oh 🙄 I realise that I committed my changes yesterday to a different branch, not this PR one..
Have moved them across now, and moved the js script into the init fn.

In your example branch, the script wasn't being run - `jsInject` isn't actually assigned, so the executed code is an empty string. maybe that's why it didn't block the initialisation page? 

the version in this branch is working for me now, and is looking pretty clean. shall we just push ahead with it? I can make an apk for the testers to try out

andreievg (approved):
Thanks @mark-prins, looks great

PR #4392, opened 2024-07-09
4040 refactor changelog triggers to use rust code

### LINKED ISSUE #4040: Refactor changelog triggers to use rust code
## The suggestion

Currently we use database triggers to record when changes are made to records that we need to sync.
This has some advantages (particularly that manual database interventions will be automatically captured), however it also has disadvantages such as the need to re-create triggers when some kinds of database changes are made,  making it harder to see where and when triggers are used, and work arounds such as db manipulation to skip changelog for db migrations.

A new pattern was established with the asset sync work, that would allow the rust code to manage the changelog entries using the Upsert Trait.

## Example use case
Currently we have an Upsert Trait like this used for sync.
We should remove the `upsert_sync` function from this trait and refactor the associated `_row.rs` file to insert the changelog record.

```rust
pub trait Upsert: DebugTrait {
    // TODO:(Long term) DELETE THIS TRAIT METHOD AND REMOVE TRIGGERS?
    fn upsert_sync(&self, con: &StorageConnection) -> Result<(), RepositoryError>;

    fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError> {
        self.upsert_sync(con)?;
        // When not using triggers to create changelog records, this is where you may want to implement changelog logic
        // This function should return the id of the changelog record created...
        Ok(None)
    }

    // Test only
    fn assert_upserted(&self, con: &StorageConnection);
    // Test only, can be used to drill down to concrete type (see test below)
    // also casting to any must be implemented by concrete type to be able to downcast to it
    // This is needed for integration test (where test record is generated for inventory adjustment, but id is not know until site is created)
    fn as_mut_any(&mut self) -> Option<&mut dyn Any> {
        None
    }
}
```

Example code to inserting changelog in rust code.
```
   pub fn upsert_one(&self, asset_row: &AssetRow) -> Result<i64, RepositoryError> {
        self._upsert_one(asset_row)?;
        self.insert_changelog(
            asset_row.id.to_owned(),
            RowActionType::Upsert,
            Some(asset_row.clone()),
        )
    }

    fn insert_changelog(
        &self,
        asset_id: String,
        action: RowActionType,
        row: Option<AssetRow>,
    ) -> Result<i64, RepositoryError> {
        let row = ChangeLogInsertRow {
            table_name: ChangelogTableName::Asset,
            record_id: asset_id,
            row_action: action,
            store_id: row.map(|r| r.store_id).unwrap_or(None),
            ..Default::default()
        };
        ChangelogRepository::new(&self.connection).insert(&row)
    }
```

Certain types of migrations such as Name & Store merge have required re-creating all triggers in the system which creates a high risk of something being missed. Rust's code checking will ensure 
Allow more flexibility and logic to be managed around when and how changelog entries are created (e.g. related entries might need to be referenced to find the associated name, or store)
Makes code most consistent and maintainable.

I think it should ultimately reduce risk, but the risks of the change would be removing changelog triggers but not creating the associated `insert_changelog` functions. We also have a risk that manual changes to the database (most likely at central) wouldn't automatically be synced as we'd need a changelog record to trigger sync

Quite a bit as there's a lot of tables with change log that need to be considered and some testing would be required to make sure things are working correctly. Maybe 2-3 days?

jmbrunskill:
Fixes #4040

# 👩🏻‍💻 What does this PR do?

Removes database changelog triggers, and replaces them with rust code to insert change logs.

Adds `asset_internal_location` change logs as that appears to have been missed.

`Delete` trait is now treated the same as the `Upsert` trait with source site_id added on sync.

Does **not** remove `is_sync_update` column from changelog but I think we can in a later PR, using source_site_id for everything.

I have removed the `toggle_is_sync_update` functions and associated diesel tables, but not the columns themselves...

I've commented out a few delete functions where they're not used in the code base, and no trigger existed for delete.
Figure we should probably either a) remove the function or b) add changelog? Or a we happy to have those functions there without change log?

Should I create a PR removing `is_sync_update` too?
What should I do with the `is_sync_update` column on some tables?

Please double check I haven't got any changelog table names wrong, feels like it would be easy to do that with copy and pasting...
Also check if I'm missing any `name_link_id` or `store_id` references.

# 🧪 Testing

- [ ] This needs lots of sync testing I guess?
- [ ] Check that no records are sent with the wrong changelog type.

# 📃 Documentation

  1.
  2.


server/repository/src/db_diesel/barcode_row.rs
```diff
@@ -125,62 +112,3 @@ impl Upsert for BarcodeRow {
         )
     }
 }
-
-#[cfg(test)]
```
jmbrunskill, line 129: I've removed these tests currently. Do you think we need to re-implement a similar pattern for every `changelog` table?

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -89,6 +89,7 @@ pub enum ChangelogTableName {
     AssetCatalogueItem,
     AssetCatalogueItemProperty,
     AssetCatalogueProperty,
+    AssetInternalLocation,
```
jmbrunskill, line 92: Found this wasn't being sync'd, but should be.

server/repository/src/db_diesel/changelog/test.rs
```diff
@@ -39,6 +39,7 @@ async fn test_changelog() {
             r.table_name = ChangelogTableName::Location;
             r.record_id = mock_location_1().id;
             r.row_action = RowActionType::Upsert;
+            r.store_id = Some(mock_location_1().store_id.clone());
```
jmbrunskill, line 42: `store_id` wasn't be set in location upsert trigger. I've added it now, feels like the right thing to do, but let me know if not...

server/repository/src/db_diesel/changelog/test.rs
```diff
@@ -461,7 +468,7 @@ async fn test_changelog_name_and_store_id_in_trigger() {
         |connection, r| {
             InvoiceLineRowRepository::new(connection)
                 .delete(&r.id)
-                .unwrap()
+                .unwrap();
```
jmbrunskill, line 471: These `;` are needed as otherwise would be returning `change_log_id` which isn't handled.

server/repository/src/db_diesel/clinician_link_row.rs
```diff
@@ -106,8 +106,9 @@ impl<'a> ClinicianLinkRowRepository<'a> {
 }
 
 impl Upsert for ClinicianLinkRow {
-    fn upsert_sync(&self, con: &StorageConnection) -> Result<(), RepositoryError> {
-        ClinicianLinkRowRepository::new(con).upsert_one(self)
+    fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError> {
+        ClinicianLinkRowRepository::new(con).upsert_one(self)?;
+        Ok(None) // Table not in Changelog
```
jmbrunskill, line 111: Added this everywhere I found a `Upsert` without a changelog.
Hopefully easier to find that way? With OMS Central we'll want changelog for all these tables I think.

server/repository/src/db_diesel/clinician_store_join_row.rs
```diff
@@ -41,28 +42,30 @@ impl<'a> ClinicianStoreJoinRowRepository<'a> {
...
+        &self,
+        row: &ClinicianStoreJoinRow,
+        action: RowActionType,
+    ) -> Result<i64, RepositoryError> {
+        let row = ChangeLogInsertRow {
+            table_name: ChangelogTableName::ClinicianStoreJoin,
+            record_id: row.id.clone(),
+            row_action: action,
+            store_id: Some(row.store_id.clone()),
+            name_link_id: None,
```
jmbrunskill, line 66: I don't think there should be name here right?

server/repository/src/db_diesel/document.rs
```diff
@@ -233,27 +234,27 @@ impl<'a> DocumentRepository<'a> {
...
-
-        Ok(())
-    }
+    fn insert_changelog(
+        &self,
+        row: &Document,
+        action: RowActionType,
+    ) -> Result<i64, RepositoryError> {
+        let row = ChangeLogInsertRow {
+            table_name: ChangelogTableName::Document,
```
jmbrunskill, line 250: This table_name is duplicated, so probably mean something's wrong, maybe just need one of these?

server/repository/src/db_diesel/document_registry_row.rs
```diff
@@ -56,14 +57,26 @@ impl<'a> DocumentRegistryRowRepository<'a> {
...
             .do_update()
             .set(row)
             .execute(self.connection.lock().connection())?;
-        Ok(())
+        self.insert_changelog(&row.id, RowActionType::Upsert)
+    }
+
+    fn insert_changelog(&self, uid: &str, action: RowActionType) -> Result<i64, RepositoryError> {
+        let row = ChangeLogInsertRow {
+            table_name: ChangelogTableName::Document,
```
jmbrunskill, line 72: This table_name is duplicated, so probably mean something's wrong, maybe just need one of these?

server/repository/src/db_diesel/name_row.rs
```diff
@@ -294,20 +294,18 @@ impl<'a> NameRowRepository<'a> {
...
 
-        self.insert_changelog_oms_fields(name_id.to_owned(), RowActionType::Upsert, None)
+        self.insert_changelog_oms_fields(name_id.to_owned(), RowActionType::Upsert)
     }
 
     fn insert_changelog(
         &self,
         record_id: String,
         row_action: RowActionType,
-        store_id: Option<String>,
```
jmbrunskill, line 304: `store_id` was never set anywhere for Name.
Should it be? Was this a bug?

server/service/src/sync/sync_status/test.rs
```diff
@@ -110,7 +110,7 @@ async fn sync_status() {
                 .map(|i| {
                     inline_init(|r: &mut LocationRow| {
                         r.id = i.to_string();
-                        r.store_id = mock_store_a().id;
+                        r.store_id = mock_store_b().id;
```
jmbrunskill, line 113: Store needs to be on the right `sync_site` or it won't be in the outgoing records list...
(Result of adding store_id to Location sync change logs)

server/repository/src/migrations/v2_02_00/remove_changelog_triggers.rs
```diff
@@ -0,0 +1,118 @@
+use crate::migrations::*;
+
+#[cfg(not(feature = "postgres"))]
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    sql!(
+        connection,
+        r#"
+        DROP TRIGGER IF EXISTS location__insert_trigger;
```
andreievg, line 8: looks like double underscore here.

server/repository/src/db_diesel/invoice_line_row.rs
```diff
@@ -97,14 +101,36 @@ impl<'a> InvoiceLineRowRepository<'a> {
...
+        self.insert_changelog(row, RowActionType::Upsert)
+    }
+
+    fn insert_changelog(
+        &self,
+        row: &InvoiceLineRow,
+        action: RowActionType,
+    ) -> Result<i64, RepositoryError> {
+        let invoice = InvoiceRowRepository::new(self.connection).find_one_by_id(&row.invoice_id)?;
+        let invoice = match invoice {
```
andreievg, line 120: My opinionated stylistic preference, not asking for change, happy with match

`invoice.ok_or(RepositoryError::NotFound)?`

server/repository/src/db_diesel/invoice_line_row.rs
```diff
@@ -161,10 +187,18 @@ impl<'a> InvoiceLineRowRepository<'a> {
         Ok(())
     }
 
-    pub fn delete(&self, invoice_line_id: &str) -> Result<(), RepositoryError> {
+    pub fn delete(&self, invoice_line_id: &str) -> Result<Option<i64>, RepositoryError> {
+        let old_row = self.find_one_by_id(invoice_line_id)?;
+        let change_log_id = match old_row {
+            Some(old_row) => self.insert_changelog(&old_row, RowActionType::Delete)?,
```
andreievg, line 193: Do you think this i really needed ? Why not just insert any way ?

andreievg (commented):
Thanks James, I am still reviewing, but so far look very good, i've updated integration tests and actually tried changing to source_site_id being passed on as parameters rather then saved in Integrationoperaiton, see what you think

https://github.com/msupply-foundation/open-msupply/compare/4040-refactor-changelog-triggers-to-use-rust-code...4040-refactor-changelog-triggers-to-use-rust-code(Source-site-id-passed-as-parameter)

One of the integration tests is failing, I am sure it's not because of changes in this PR, i need to look into it a bit closer (it's service charges $ totals)

server/repository/src/migrations/v2_02_00/remove_changelog_triggers.rs
jmbrunskill: image
I think it's correct right?

server/repository/src/db_diesel/invoice_line_row.rs
jmbrunskill: If we didn't look up the old record, we don't get the `store_id` and `name_link_id` assigned? Is that ok for a Delete? Maybe it is? I wasn't sure so went this way :)

jmbrunskill:
> Thanks James, I am still reviewing, but so far look very good, i've updated integration tests and actually tried changing to source_site_id being passed on as parameters rather then saved in Integrationoperaiton, see what you think
> 
> [4040-refactor-changelog-triggers-to-use-rust-code...4040-refactor-changelog-triggers-to-use-rust-code(Source-site-id-passed-as-parameter)](https://github.com/msupply-foundation/open-msupply/compare/4040-refactor-changelog-triggers-to-use-rust-code...4040-refactor-changelog-triggers-to-use-rust-code(Source-site-id-passed-as-parameter))

Thanks for your reviews!
I think passing the `source_site_id` as a parameter is nicer! I remember not being super happy with the pattern I had of assigning it later using add_source_site_id. I had a quick look at your branch and seems logical but not familiar enough with the code paths to know if there's some edge case. If you're happy I'm happy to merge it, hopefully we can get a bit of time testing sync out.

> 
> One of the integration tests is failing, I am sure it's not because of changes in this PR, i need to look into it a bit closer (it's service charges $ totals)

I haven't tried them, let know if you haven't got to it and I can investigate, (need to get integration testing setup again)

server/repository/src/db_diesel/invoice_line_row.rs
jmbrunskill: That's is nice! Thanks for the reminder. I tend to revert to the pattern I know, but very happy with that approach too.

jmbrunskill:
> Thanks for your reviews! I think passing the `source_site_id` as a parameter is nicer! I remember not being super happy with the pattern I had of assigning it later using add_source_site_id. I had a quick look at your branch and seems logical but not familiar enough with the code paths to know if there's some edge case. If you're happy I'm happy to merge it, hopefully we can get a bit of time testing sync out.

Feel free to just merge that branch when you're ready and I can give it another quick test here.

server/repository/src/lib.rs
```diff
@@ -44,21 +44,13 @@ pub fn run_db_migrations(connection: &StorageConnection) -> Result<(), String> {
...
-    // TODO:(Long term) DELETE THIS TRAIT METHOD AND REMOVE TRIGGERS?
-    fn upsert_sync(&self, con: &StorageConnection) -> Result<(), RepositoryError>;
-
-    fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError> {
-        self.upsert_sync(con)?;
-        // When not using triggers to create changelog records, this is where you may want to implement changelog logic
-        // This function should return the id of the changelog record created...
-        Ok(None)
-    }
+    fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError>;
```
andreievg, line 53: I think a small comment about Option<i64> wouel be great in the trait description

server/repository/src/migrations/v2_02_00/remove_changelog_triggers.rs
andreievg: Thanks !

I ran both sqlite and postgres, and looked at schema, 0 triggers.

I did see some artifact trigger methods though, maybe worth deleting, but not critical.

server/repository/src/db_diesel/clinician_link_row.rs
andreievg: Thanks !

andreievg (approved):
Thanks James, i had another quick looks, and I think this is good to go !

Cheers for all of those changes ! 

I fixed integration tests and merged into this branch

server/repository/src/migrations/v2_02_00/remove_changelog_triggers.rs
andreievg: Screenshot 2024-07-22 at 12 35 07 PM

jmbrunskill: Thanks Andrei had been meaning to come back to that, but forgot thanks for the reminder.
I think it's probably worth cleaning up, as could cause confusion in future...

PR #4402, opened 2024-07-11
3964 Tax Edit modal value is not updated if you edit tax percentage from service line edit modal

### LINKED ISSUE #3964: Tax Edit modal value is not updated if you edit tax percentage from service line edit modal
## What went wrong? 😲

Tax Edit modal value does not get updated when we edit the tax percentage via the service line edit modal - only updates on refresh:

https://github.com/msupply-foundation/open-msupply/assets/55115239/6903b945-78e4-42ba-8663-1e2af8d8be66

Same for both inbound and outbound shipments.

## Expected behaviour 🤔

Tax Edit modal reflects latest tax percentage immediately.

## How to Reproduce 🔨

1. Go to an Outbound shipment
2. Add a service charge
3. Set a tax percentage
4. Then open tax edit modal
5. See error: tax percentage does not reflect what you entered in the service charge edit modal

## Your environment 🌱

- Open mSupply Version: 2.0.0
- Legacy mSupply Central Server Version: 7.14

- Platform: Browser

- Database type: PG

## Additional context

[TaxEdit](https://github.com/msupply-foundation/open-msupply/blob/main/client/packages/invoices/src/OutboundShipment/DetailView/modals/TaxEdit.tsx#L17) modal uses state, which is not updated/invalidated when we save the service charges. Easiest way is probably to only mount the tax edit modal (with the useState) when we click the open/edit button, so that the state is set with a fresh value. 

Might also be a nice chance to consolidate Inbound and Outbound TaxEdit modals to be shared... they are identical, and I can imagine us wanting to include service charges in returns at some point too? 👀

roxy-dao:
Fixes #3964

# 👩🏻‍💻 What does this PR do?
Val wasn't being updated when tax was updated, so have used useEffect to prompt change.

https://github.com/msupply-foundation/open-msupply/assets/61820074/29fb9e66-0036-4dfb-b28a-25375c345b06

# 🧪 Testing

- [ ] Have service item in master list
- [ ] Go to an Outbound shipment
- [ ] Add a service charge
- [ ] Set a tax percentage
- [ ] Then open tax edit modal

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/TaxEdit.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/TaxEdit.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   useTranslation,
   InputModal,
@@ -19,6 +19,10 @@ export const TaxEdit = ({ disabled = false, tax, onChange }: TaxEditProps) => {
   const t = useTranslation('replenishment');
   const [val, setVal] = useState<number | undefined>(tax);
 
+  useEffect(() => {
+    setVal(tax);
+  }, [tax]);
+
   return (
     <>
       <IconButton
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
🚢  works nicely thanks!

PR #4404, opened 2024-07-14
4401 reports display available report in report menu

### LINKED ISSUE #4401: Reports: Display available report in report menu
Screenshot 2024-07-11 at 2 20 29 PM

We can use existing useReports hook to get reports list.

Current we use `Dispensary` report type for Program reports, we will be moving this to general report page.

Would recommend adding ContextType = Report ("type of form" in mSupply configuration), and using subContext for category name for the very first iteration. I think we need to make this a bit more type safe because we will be introducing translations and icon etc. 

In the long run Report Category as another field, with "Other" as mapping (we will do in follow up issue).

For first iteration no icons or translations (just display sub context)

roxy-dao:
Fixes #4401

# 👩🏻‍💻 What does this PR do?
Displays available reports in the widgets associated with those reports (using context to determine this, probably want this to be type safe). The reports don't lead anywhere. This just displays text of available reports.
Screenshot 2024-07-15 at 8 02 25 AM

# 🧪 Testing

- [ ] Add some reports in mSupply with subcontext = "StockAndItem" OR "Expiring"
- [ ] Sync
- [ ] Those reports should display in the list view

# 📃 Documentation

  1.
  2.


client/packages/reports/src/ListView/ListView.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { AppBarButtons } from './AppBarButton';
+import { SidePanel } from './SidePanel';
+import { StockAndItemWidget } from './StockAndItemWidget';
+import { useReport } from '@openmsupply-client/system';
+import { ExpiringWidget } from './ExpiringWidget';
+
+export const ListView = () => {
+  const { data } = useReport.document.list({});
+  const stockAndItemReports = data?.nodes?.filter(
+    report => report?.subContext === 'StockAndItems'
```
roxy-dao, line 12: Probably want an enum of sub-contexts so we can reliability search for the correct sub-context...needs change in mSupply? 🤔

client/packages/common/src/ui/components/domain/Dashboard/Widget.tsx
```diff
@@ -29,9 +32,21 @@ export const Widget: React.FC<PropsWithChildren<WidgetProps>> = ({
       boxShadow: theme => theme.shadows[2],
     }}
   >
-    <Typography sx={{ fontSize: '14px', fontWeight: 'bold' }}>
-      {title}
-    </Typography>
+    <Grid container alignItems="center">
+      {Icon && (
```
roxy-dao, line 36: Can change later... not a fan of the background colour on icon but easy to change (:

client/packages/reports/src/ListView/StockAndItemWidget.tsx
```diff
@@ -0,0 +1,56 @@
...
+              }}
+              key={report?.id}
+            >
+              <Typography
+                variant="h6"
+                sx={{
+                  paddingBottom: 2,
+                }}
+              >
+                {report?.name}
```
roxy-dao, line 42: =/ the report name would have to be descriptive enough for the user to understand what it does =/

andreievg: Yep, it would need to be descriptive, we also might have name and "display name" (for translation), and maybe some extra info to describe the report on the report records

client/packages/reports/src/ListView/ListView.tsx
andreievg: I think we would need another column on report, we can get to it when we transfer program report from dispensary menu to here. Since sub context in those reports seems to be linked to the program (although I can't find anywhere in code where sub context for reports is used ¯\_(ツ)_/¯)

andreievg (changes requested):
Thanks Roxy. 

Really want to approve this one, because it looks good enough to go in. But I am a bit unsure about the abstraction of using Dashboard/Widget. i.e. you added the icon to Widget and using that in StockAndItemWidget and ExpiringWidget. I think it should be a new 'widget' type, that accepts Icon and report list. I feel that styles of the <Card> will change, although there is no example of this in the main section of designs, but there would be more categories that fit on the screen horizontally:

Screenshot 2024-07-15 at 4 42 37 PM
Screenshot 2024-07-15 at 4 42 33 PM

The change request is to revert the change to Dashboard/Widget, and to add new widget in reports section that accepts Icon and report list and use that in ListView

client/packages/reports/src/ListView/ExpiringWidget.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface ExpiringReportProps {
+  reports: ReportRowFragment[] | undefined;
+}
+
+export const ExpiringWidget: React.FC<ExpiringReportProps> = ({ reports }) => {
+  const t = useTranslation('reports');
+
+  return (
+    <>
```
mark-prins, line 20: :nit no need for the `<>` surely?

client/packages/reports/src/ListView/StockAndItemWidget.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface StockReportProps {
+  reports: ReportRowFragment[] | undefined;
+}
+
+export const StockAndItemWidget: React.FC<StockReportProps> = ({ reports }) => {
+  const t = useTranslation('reports');
+
+  return (
+    <>
```
mark-prins, line 20: :nit this one too..

andreievg (approved):
Thanks @roxy-dao, looks great

PR #4427, opened 2024-07-16
Save the replacement date of an imported equipment item

### LINKED ISSUE #4399: Asset Import doesn't save Replacement Date
## What went wrong? 😲

Noticed that the import seems to validate the replacement due date ok

telegram-cloud-photo-size-1-4934212593657163008-y

but when I look at the asset there is no actual date populated

telegram-cloud-photo-size-1-4934212593657163009-y

## Expected behaviour 🤔

Date should save correctly

## How to Reproduce 🔨

1. Go to Cold Chain -> Equipment
2. Import an asset with with replacement date
3. Check if replacement due shows up

## Your environment 🌱

- Open mSupply Version: 2.1

mark-prins:
Fixes #4399

# 👩🏻‍💻 What does this PR do?

There are two conversions of the import object made prior to sending to the API.
The first of these was stripping out some of the properties - I've refactored to take all props and only override those that need to be.

# 🧪 Testing

- [ ] Try to import an asset, with a replacement date specified. Verify that the date is saved.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
+++ b/client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
@@ -58,42 +58,30 @@ export type ImportRow = {
 export type LineNumber = {
   lineNumber: number;
 };
+
 export const toInsertEquipmentInput = (
   row: ImportRow,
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Much better thanks Mark!

Issue #4412, opened 2024-07-15
Create R&R forms, prepopulated with Open mSupply data


lache-melvin:
Add a create button to the list view.

This should open a modal, where you can select the program, and the period.

You can only create an R&R form for a given period once the form for the previous period has been submitted. So the earliest period without a form should be automatically selected.

If no R&R forms for that master list exist yet, select the earliest period.

R&R form should only be able to be created if the previous one is submitted. (TODO: confirm this - and what do we do for the first one?)

From the given Master List, for each item, create R&R row:

- Item name (from item)
- strength (from item)
- unit (from item)
- VEN (from item - will require translation from mSupply)
- initial balance (at the start of the period, from item ledger)
- quantity received  (within the period, from item ledger)
- quantity consumed (within the period, from item ledger)
- adjusted quantity consumed - (quantity consumed x 30)/(30 - stock out duration) _(30, or days in month?)_
- losses/adjustments - blank, to be entered
- stock out duration (days) - (could be calculated... # days where stock = 0)
- final balance -  (initial balance) + (quantity received) - (quantity consumed) ± (losses / adjustments)
- AMC
- Maximum quantity - AMC * 2
- expiration date  - blank, to be entered
- requested quantity - Maximum quantity - final balance
- Approved quantity - blank, to be entered
- comment - blank, to be entered

PR #4439, opened 2024-07-22
4412 create rnr form (part 1)

lache-melvin:
Part 1 of #4412

# 👩🏻‍💻 What does this PR do?

Getting in this here because its beasty as is... sorry 😭  Leaving comments on TODOs :)

Adds modal to create R&R form:
Screenshot 2024-07-22 at 12 22 00 PM

Clicking save creates the form. 

# 🧪 Testing

- [ ] Go to Programs > R&R forms
- [ ] You are redirected to the detail route (but will be blank, no detail view yet)
- [ ] Newly created form appears in the list view

# 📃 Documentation

  1.
  2.


client/packages/programs/src/RnRForms/ListView/AppBarButtons.tsx
```diff
@@ -20,7 +20,6 @@ export const AppBarButtons = ({
     <AppBarButtonsPortal>
       <Grid container gap={1}>
         <ButtonWithIcon
-          startIcon={<PlusCircleIcon />}
```
lache-melvin, line 23: side quest: on small screens, having both `startIcon` and `Icon` meant the plus was showing twice?

client/packages/programs/src/RnRForms/ListView/RnRFormCreateModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+          }
+        />
+        <InputWithLabelRow
+          label={t('label.schedule')}
+          Input={
+            <Autocomplete
+              width={width}
+              disabled={!draft.program}
+              optionKey="name"
+              // TODO: autoselect!
```
lache-melvin, line 106: Was thinking it would be nice, if you only have one program/schedule, that it is selected automatically

server/service/src/rnr_form/insert.rs
```diff
@@ -0,0 +1,160 @@
...
+        .find_many_by_program_id(&input.program_id)?
+        .iter()
+        .map(|s| s.period_schedule_id.clone())
+        .collect::<Vec<String>>();
+
+    if !period_schedule_ids.contains(&period.period_schedule_id) {
+        return Err(InsertRnRFormError::PeriodNotInProgramSchedule);
+    }
+
+    if check_rnr_form_exists_for_period(connection, &input.period_id, &input.program_id)?.is_some()
```
lache-melvin, line 112: Realising I should probably also check that this is next period in the sequence here right? Prevent skipping of periods...

And maybe also that its not a period in the future?

server/service/src/rnr_form/schedules_with_periods.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub fn get_schedules_with_periods_by_program(
+    ctx: &ServiceContext,
+    program_id: &str,
+) -> Result<Vec<PeriodSchedule>, RepositoryError> {
+    let settings_repo = ProgramRequisitionSettingsRowRepository::new(&ctx.connection);
+    let period_repo = PeriodRepository::new(&ctx.connection);
+
+    let settings = settings_repo.find_many_by_program_id(program_id)?;
```
lache-melvin, line 23: Maybe this just should have been another join in the period table?

client/packages/programs/src/api/hooks/useProgramList.ts
```diff
@@ -0,0 +1,39 @@
...
+  const { api, storeId } = useProgramsGraphQL();
+
+
+  const queryKey = [PROGRAM, LIST];
+  const queryFn = async (): Promise<{
+    nodes: ProgramFragment[];
+    totalCount: number;
+  }> => {
+    const filter = {
+      isImmunisation: false,
```
lache-melvin, line 19: Currently this filter still returns `missing_program` - where's the best place to filter that out do you think?

client/packages/programs/src/RnRForms/ListView/RnRFormCreateModal.tsx
jmbrunskill: That does sound nice!

server/service/src/rnr_form/insert.rs
jmbrunskill: I guess if we do allow RnR forms to be skipped, we'd want to do some business logic, such as mark the old ones as skipped or something?
Keen to hear Mark's thoughts on that one. As far as the ledger & stock usage goes how bad is skipping?
I'm thinking potentially you could come to store that hasn't submitted and RnR form for a year, I assume in that scenario you'd probably want to just skip the old ones and start fresh somehow?

jmbrunskill (commented):
I'm getting a permission error on this query when navigating to R&R forms...

image


```
{
    "data": null,
    "errors": [
        {
            "message": "Forbidden",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "schedulesWithPeriodsByProgram"
            ],
            "extensions": {
                "details": "Store id not specified in request, Required permissions: And([HasStoreAccess, HasPermission(RnrFormQuery)]), Store: None"
            }
        }
    ]
}
```


Commented out store access to continue. 

```
    map.insert(
        Resource::QueryRnRForms,
        PermissionDSL::And(vec![
            // PermissionDSL::HasStoreAccess,
            PermissionDSL::HasPermission(PermissionType::RnrFormQuery),
        ]),
    );
```

Are you getting this? Seems like it would always happen as you're passing `None` to `validate_auth`

client/packages/programs/src/RnRForms/ListView/RnRFormCreateModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+              const result = await create();
+              if (result)
+                navigate(
+                  RouteBuilder.create(AppRoute.Programs)
+                    .addPart(AppRoute.RnRForms)
+                    .addPart(result.id)
+                    .build()
+                );
+            } catch (e) {
+              console.error(e);
```
jmbrunskill, line 56: Will errors be shown using the default toast? (haven't checked)

client/packages/programs/src/api/hooks/useDeleteSelectedImmunisationPrograms.ts
```diff
@@ -40,7 +40,7 @@ export const useDeleteSelectedImmunisationPrograms = () => {
       selectedRows.map(
         async id => await mutateAsync({ immunisationProgramId: id })
       )
-    ).then(() => queryClient.invalidateQueries([PROGRAM]));
```
jmbrunskill, line 43: I think I had deliberately shared keys to allow for the possiblity that an immunisation program could be a normal program at the same time. Makes sense to me that we might want to procure vaccines, track their usage, etc under the same program. I think changing it is fine for now, and probably for future to realistically, shouldn't be deleted often...

client/packages/programs/src/api/hooks/useProgramList.ts
jmbrunskill: I think create_filtered_query should filter it out, just like `is_deleted`, in fact i'd be tempted to add `is_deleted` to that table and then mark `missing_program` as deleted.
It's only used during sync translation from what I can tell.

client/packages/programs/src/api/hooks/useProgramList.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+
+    const query = await api.programs({
+      storeId,
+      first: 1000, // TODO: remove arbitrary limit
```
jmbrunskill, line 24: Should be enough even for a huge country I think!

server/graphql/programs/src/queries/r_and_r_form.rs
```diff
@@ -44,3 +45,27 @@ pub fn r_and_r_forms(
...
+
+pub fn get_schedules_with_periods_by_program(
+    ctx: &Context<'_>,
+    program_id: String,
+) -> Result<PeriodSchedulesResponse> {
+    let user = validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::QueryRnRForms,
+            store_id: None,
```
jmbrunskill, line 57: Need a store_id here?

jmbrunskill (commented):
Is the intention to allow only 1 R&R form per sync site, or is it per store?
image

In this case, Q3 is blanked out because I've entered it under a different store?

I'm assuming it should be per store instead of per site?

server/service/src/rnr_form/schedules_with_periods.rs
jmbrunskill: In most cases, I think it's fine to run multiple queries in a single request.

The more important goal is to make the code readable. In other projects I'd probably have leant towards a join here, but if we don't need data from the joined table, and with the extra complexity of diesel I think it's probably a good call to do it here.

There's a slight overhead to each query but should be minimal unless someone decides to run their database at the otherside of the world from their app server.

1vs1 queries isn't a big concern, n+1 problem is a bit more significant.
https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it
I think this is an N+1 thing as we need to run a query for every `period_schedule_id`

Assuming `N` is small ( e.g. there's not too many period schedules) this approach is fine.

server/service/src/rnr_form/schedules_with_periods.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let mut period_schedule_ids = settings
+        .iter()
+        .map(|s| s.period_schedule_id.clone())
+        .collect::<Vec<String>>();
+
+    // There can be duplicates in the program settings due to name tags
```
jmbrunskill, line 30: Good to have the comment here.
Could also consider deduping in the SQL layer. Would probably require a dedicated repository method but I'm ok with those...

jmbrunskill (changes requested):
It's looking good @lache-melvin I think we just need to fix the logic to support multiple stores on the same site. Checked with @mark-prins  and that is the intention.
If there are multiple stores on the site then we need one to be able to create an R&R form for the same `program`/`period`. Not sure how likely it is to happen in practice but we probably want to make sure we don't break everything if it happens...

Also need to sort out that permission issue, but that might sort it self out when you add stores in?

server/service/src/rnr_form/insert.rs
lache-melvin: Yeah talking with Mark and Craig, we won't allow skipping. Sounds like if it has been a really long time, they'd possibly start a new period schedule (i.e. 2024 schedule with monthly periods) ... but if not, just going through and generating r&rs with the pre-populated data wouldn't take too long, then can make adjustments in the current one

server/service/src/rnr_form/schedules_with_periods.rs
lache-melvin: Its a good question! Probably only one active one? Will ask 🙏

lache-melvin:
> I'm getting a permission error on this query when navigating to R&R forms...
>
> Are you getting this? Seems like it would always happen as you're passing `None` to `validate_auth`

Oof haha thank you! Turns out I still had debug no access control turned on 😅

server/repository/src/migrations/v2_02_00/rnr_form.rs
```diff
@@ -41,7 +41,10 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
         r#"
             ALTER TYPE permission_type ADD VALUE IF NOT EXISTS 'RNR_FORM_QUERY';
             ALTER TYPE permission_type ADD VALUE IF NOT EXISTS 'RNR_FORM_MUTATE';
-            ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'rn_r_form';
+
+            ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'rnr_form';
```
mark-prins, line 45: actually - this breaks postgres for me. the correct name is 
```
            ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'rn_r_form';
            ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'rn_r_form_line';
```

server/repository/src/db_diesel/period/period.rs
```diff
@@ -93,16 +93,35 @@ fn to_domain((period_row, period_schedule_row, rnr_form_row): PeriodJoin) -> Per
     }
 }
 
+// rnr_form_dsl::period_id.eq(period_dsl::id)
+type PeriodIdEqualToId = Eq<rnr_form_dsl::period_id, period_dsl::id>;
+// rnr_form_dsl::store_id.eq(store_id)
+type StoreIdEqualToStr = Eq<rnr_form_dsl::store_id, String>;
+// rnr_form_dsl::program_id.eq(store_id)
```
jmbrunskill, line 100:
```suggestion
// rnr_form_dsl::program_id.eq(program_id)
```

server/service/src/rnr_form/insert.rs
```diff
@@ -88,6 +92,7 @@ fn validate(
         CheckOtherPartyType::Supplier,
     )?;
 
+    // TODO... for store! How?
```
jmbrunskill, line 95: Programs are masterlists, so they are global.
We could check if it's available it the current store but I don't think it's necessary.

jmbrunskill (approved):
Looks great, thanks for the updates!

server/service/src/rnr_form/insert.rs
lache-melvin: While testing the different stores, I realised I could see/search the programs, but that the underlying master_list has store-based visibility. I think I'll add a join on the query....

lache-melvin: Can I check this is all good with you @jmbrunskill ? [0969e6f](https://github.com/msupply-foundation/open-msupply/pull/4439/commits/0969e6f047ce73a3ab1587ce7302690ea158235f) 🙏

jmbrunskill: Looks good to me! Thanks!

PR #4446, opened 2024-07-23
4412 generate blank rows

lache-melvin:
Part of #4412

# 👩🏻‍💻 What does this PR do?

Generates blank rnr form rows, so that we can split work i.e. create UI to view/edit.

Adds translation of VEN category and item strength.

Note that some clean up/additional logic from #4439 will be merged into here soon, and just ignore the generation/data population logic, that's coming in a separate PR :)

# 🧪 Testing

- [ ] Create a new R&R form for a program
- [ ] In DB browser, ensure that rnr_form_lines have been added for each item in the program master list

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_02_00/item_ven.rs
```diff
@@ -0,0 +1,40 @@
...
+    const VEN_CATEGORY_ENUM_TYPE: &str = if cfg!(feature = "postgres") {
+        "ven_category"
+    } else {
+        "TEXT"
+    };
+
+    sql!(
+        connection,
+        r#"
+            ALTER TABLE item ADD COLUMN strength TEXT;
```
mark-prins, line 26: actually not certain that this is much used in OG - it was added fairly recently and afaik mostly the name is stuffed with this detail and not split out into the strength field 🤷

mark-prins (approved):
👍  LGTM! thanks

server/repository/src/migrations/v2_02_00/item_ven.rs
lache-melvin: Yeah, had some fun trying to find the field in the first place 😅 But figured if they wanted to populate it, it might as well come through? Unless you think this is causing more confusion than help, maybe we just don't support the strength column in the rnr rows?

mark-prins: yeah, I thought the same.. might as well bring it through. They had strength on the R&R form, so it makes sense to store it somewhere

PR #4452, opened 2024-07-25
4451 Print button for report detail view 

### LINKED ISSUE #4451: Print button for report detail view
Add print button to report detail view

### Moneyworks Jobcode 🧰
OMS:REPT

roxy-dao:
Fixes #4451

# 👩🏻‍💻 What does this PR do?
Adds print button to report detail view
Screenshot 2024-07-26 at 8 50 49 AM

# 🧪 Testing

** Will update testing instructions in epic after MVP done**

- [ ] Have reports set up (will need context as either `StockAndItems` or `Expiring` with context = Report
- [ ] Alternatively can just try print one of the program reports (needs program module store pref)
- [ ] Click on a report in omSupply
- [ ] Click print button
- [ ] Should show up with preview to print
- [ ] Alternatively, filter report
- [ ] Click on print
- [ ] Should show up with preview to print

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -20,7 +20,7 @@ import { JsonData } from '@openmsupply-client/programs';
 export const DetailView = () => {
   const { id } = useParams();
   const { setCustomBreadcrumbs } = useBreadcrumbs();
-  const { data: report } = useReport(id ?? '');
+  const { data: report, isLoading: isReportLoading } = useReport(id ?? '');
   const { mutateAsync, isLoading } = useGenerateReport();
   const [fileId, setFileId] = useState<string | undefined>();
   const { print, isPrinting } = usePrintReport();
@@ -29,17 +29,16 @@ export const DetailView = () => {
     updateQuery,
     urlQuery: { reportArgs: reportArgsJson },
   } = useUrlQuery({ skipParse: ['reportArgs'] });
+  const reportArgs =
+    (reportArgsJson && JSON.parse(reportArgsJson.toString())) || undefined;
 
   // When reportWithArgs is undefined, args modal is closed
   const [reportWithArgs, setReportWithArgs] = useState<
     ReportRowFragment | undefined
   >();
 
-  // Report should be loaded if id is available
-  const isReportDescriptionLoading = !report?.id;
-
   useEffect(() => {
-    if (isReportDescriptionLoading) return;
+    if (!report?.id) return;
 
     setCustomBreadcrumbs({ 0: report.name ?? '' });
 
@@ -49,9 +48,6 @@ export const DetailView = () => {
       return;
     }
 
-    let reportArgs =
-      (reportArgsJson && JSON.parse(reportArgsJson.toString())) || undefined;
-
     if (!!reportArgs) {
       generateReport(report, reportArgs, false);
       return;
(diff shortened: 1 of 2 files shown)


client/packages/reports/src/DetailView/DetailView.tsx
```diff
@@ -82,6 +85,21 @@ export const DetailView = () => {
     setReportWithArgs(report);
   }, []);
 
+  const printReport = useCallback(() => {
+    if (report === undefined) {
+      return;
+    }
+
+    let reportArgs =
```
CarlosNZ, line 93:
```suggestion
    const reportArgs =
```

CarlosNZ: You're not in Rustland any more ;)

CarlosNZ: Also -- the `reportArgs` value is defined in multiple places -- you should be able to just define it once (up the top immediately below the `useUrlQuery` I would say)

CarlosNZ (commented):
Looks good, just a couple of small nits.

One other thing, it's not directly related to this PR, but I'm wondering about this:
```
  if (isReportDescriptionLoading) {
    return <NothingHere />;
  }
```

Looks a bit weird to see the "Nothing Here" page, which then turns into a Loading spinner, and then the report. Shouldn't it just use the loading spinner for both?

CarlosNZ (approved):
LGTM, thanks @roxy-dao

client/packages/reports/src/DetailView/DetailView.tsx
```diff
@@ -78,25 +77,35 @@ export const DetailView = () => {
...
   return (
     <>
-      {isLoading && <BasicSpinner />}
+      {(isGeneratingReport || isReportLoading) && <BasicSpinner />}
       {fileId ? (
         <>
           <iframe src={url} width="100%" />
           <AppBarButtons
-            isDisabled={!report.argumentSchema}
+            isDisabled={!report?.argumentSchema}
```
andreievg, line 105: isFilterDisabled

client/packages/reports/src/DetailView/DetailView.tsx
```diff
@@ -46,17 +48,14 @@ export const DetailView = () => {
...
-
     if (!!reportArgs) {
       generateReport(report, reportArgs, false);
       return;
     }
 
     // No urlQuery parameters exist, open modal
     openReportArgumentsModal();
-  }, [isReportDescriptionLoading]);
+  }, [report]);
```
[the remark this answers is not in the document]
andreievg: Actually i am kind of annoyed with the constantly needing to check for report (with either approach), i think i would prefer something like this: https://github.com/msupply-foundation/open-msupply/compare/4451-Print-button-for-report-detail-view...4451-Print-button-for-report-detail-view(Suggestion-with-wrapper-component)

andreievg (commented):
In general looks very good, there is a suggestion for what i think is a more straight forward managment of loading state

client/packages/reports/src/DetailView/DetailView.tsx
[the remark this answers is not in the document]
andreievg: Also i realised in my original suggestions there are a few callbacks with `[]` as the dependency, while they are actually dependent on report (so filter sometimes won't show up when button is pressed), but should be fixed in this PR

PR #1235, opened 2023-02-24
1201 html printing

clemens-msupply:
Part of #1201 

This fixes a problem I had where the iframe lost some css styling.

To test: test that browser printing is still working.

I didn't manage to get a page counts working in HTML/css. There are many articles about how to do this but they all don't work! Maybe there was support for it in the passed but today's browser don't implement it as described. So at the moment we don't have headers and footers support for HTML printing :-(

PR #4454, opened 2024-07-26
#3098 Excel from html

### LINKED ISSUE #3098: Minimal prototype to generate Excel from HTML (no custom attributes)
TODO:
- [ ] Identify HTML parsing lib
- [ ] Identify Excel rust crate (e.g. the one used in HSH)
- [ ] Use this lib to walk the HTML abstract syntax tree and extract the first HTML table
- [ ] Transform this table to an Excel file (while doing so please keep other features from the Epic, e.g. styling or positioning, in mind so that they can easily be implemented at a later stage)

CarlosNZ:
Fixes #3098

# 👩🏻‍💻 What does this PR do?

Minimal implementation of Excel output. Just **bolds** the column headers, no other styling or header info yet.

To test, use the branch from [this PR](https://github.com/msupply-foundation/open-msupply-reports/pull/74) in the reports repo, and go into `standard-reports/stock-status`, and run `./excel.sh`.

Adds another formatting option "excel" to the report output endpoint. Added function `print_html_report_to_excel` which generates excel file and returns file ID -- same as the PDF output.

I think we need to clarify what the structure of the HTML input can be guaranteed to be. The HTML being generated here is adding an extra incomplete (the tags don't actually close) "paging" table, so I had to do a bit extra and target the _actual_ table we want by its surrounding `div` (using its class name "container"), which is probably more hard-coding than we would like. We might need to fix the HTML output so it only has one table in, but we should have a chat about why it's doing that and the best way to fix it.


andreievg:
> We might need to fix the HTML output so it only has one table in, but we should have a chat about why it's doing that and the best way to fix it.

I think it's required for headers/footers (although I am not sure how it works): 
https://github.com/msupply-foundation/open-msupply/pull/1235
https://stackoverflow.com/questions/1360869/how-to-use-html-to-print-header-and-footer-on-every-printed-page-of-a-document

mark-prins:
> We might need to fix the HTML output so it only has one table in, but we should have a chat about why it's doing that and the best way to fix it.

It was so that the table headers and footers repeated on every page - when the table is longer than a single printed page. Any changes would have to retain that functionality.

server/service/src/report/report_service.rs
```diff
@@ -167,6 +172,70 @@ fn generate_html_report_to_html(
...
+    let _ = book
+        .set_sheet_name(0, sheet_name)
+        .map_err(|err| ReportError::DocGenerationError(format!("{}", err)))?;
+
+    let fragment = Html::parse_fragment(&format_html_document(document));
+    let container_selector = Selector::parse(r#"div[class="container"]"#).unwrap();
+    let table_head_selector = Selector::parse("thead").unwrap();
+    let table_body_selector = Selector::parse("tbody").unwrap();
+    let row_selector = Selector::parse("tr").unwrap();
+    let cell_selector = Selector::parse("td").unwrap();
```
andreievg, line 193: I wanted to add some suggestions and check out the functionality in more detail. I realise you have to rebuild are re-run server every time any changes happen here, at the start I wanted to show you how to add quick tests and check functionality via tests rather then by re-running the whole thing. I think it's good to have this also for future devs to quickly see how things work and try changes themselves. This is kind of what I mean (at least for selectors):

https://github.com/msupply-foundation/open-msupply/compare/3098-excel-from-html...3098-excel-from-html(Test-example)

andreievg: You'll also notice in that branch, that i do `let cell` rather then `get_cell_mut` twice and also `let sheet` rather then repeated `get_sheet_by_name_mut`

andreievg (commented):
Looks pretty good, i have some preliminary suggestions. There are some other changes that I would like to suggest. Would prefer us to do a peer programming session though, when you in the office tomorrow

CarlosNZ:
> Looks pretty good, i have some preliminary suggestions. There are some other changes that I would like to suggest. Would prefer us to do a peer programming session though, when you in the office tomorrow

Sounds good, see you soon.

andreievg (approved):
Thanks, good to go, together with the changes we've done in peer programming session

PR #4456, opened 2024-07-26
4414 editing/saving R&R progress

### LINKED ISSUE #4414: Be able to save your progress in the R&R form
Add a `confirm` checkbox column to the R&R detail view. This boolean should be saved on the R&R row.

Selecting this checkbox fires a mutation to save the row. This helps the user to track their progress.

Also implement save button - this should save the state of the whole form, regardless of whether the rows are confirmed

lache-melvin:
Fixes #4414

# 👩🏻‍💻 What does this PR do?

Allows the updating of values for the R&R form rows. Updates from just `quantity` to `snapshot` and `entered` quantities, as discussed, to help us track what the system knows vs. what the user entered.

This only adds backend side of updates, waiting to sync up with what @mark-prins has done on the display/UI side :)

# 🧪 Testing

- [ ] Run tests
- [ ] Use graphql playground to update a line

# 📃 Documentation

  1.
  2.


server/service/src/rnr_form/tests/insert.rs
```diff
@@ -252,7 +262,7 @@ mod query {
         RnRFormRowRepository::new(&context.connection)
             .upsert_one(&RnRFormRow {
                 status: RnRFormStatus::Finalised,
-                ..mock_rnr_form_a()
+                ..mock_rnr_form_b()
```
lache-melvin, line 265: Edits to this file were a bit of a side quest changing some inputs after adding more mock data

server/graphql/programs/src/mutations/rnr_form/update.rs
```diff
@@ -0,0 +1,136 @@
...
+    let service_context = service_provider.context(store_id.clone(), user.user_id)?;
+    match service_provider.rnr_form_service.update_rnr_form(
+        &service_context,
+        &store_id,
+        UpdateRnRFormInput::to_domain(input),
+    ) {
+        Ok(RnRForm {
+            rnr_form_row,
+            name_row,
+            store_row: _,
```
mark-prins, line 62: would be good to pass down and validate that the R&R form belongs to the supplied store

server/service/src/rnr_form/update.rs
```diff
@@ -0,0 +1,199 @@
...
+
+            for line in rnr_form_lines {
+                rnr_form_line_repo.upsert_one(&line)?;
+            }
+
+            activity_log_entry(
+                ctx,
+                ActivityLogType::RnrFormUpdated,
+                Some(input.id.clone()),
+                None,
```
mark-prins, line 73: thought for later: adding in some detail, so the user can see what the change was.
this is a can o worms, which is why I think later!

mark-prins (commented):
🚢  looks great! thanks. 
have been testing with the graphQL and it's working nicely

wondering if the calculated columns should be updatable via the API - or if they should be calculated in the API. currently I think f/e as we have plugins available there, so we can customise for clients. that's a spurious reason though!

mark-prins (approved):
Oops. meant to approve, not just comment!

PR #4465, opened 2024-07-29
4464 Report filter doesn't filter if you type too fast

### LINKED ISSUE #4464: Report filter doesn't filter if you type too fast
## What went wrong? 😲
The report filter doesn't correctly apply the text that was inputted if the user types/presses ok too fast

## How to Reproduce 🔨

1. Set up new reports in mSupply (download from reports repo)
2. Sync
3. Click on one of the reports
4. Type an item name really quickly and press Ok
5. See bug

## Your environment 🌱

- Open mSupply Version: 2.2

- Platform:

- Database type:

roxy-dao:
Fixes #4464

# 👩🏻‍💻 What does this PR do?
Debounce ok button for args...  Still a bit finicky depending on how fast you type then click.... but I just type/click super fast... so maybe just me... 

# 🧪 Testing

- [ ] Set up new reports (run upsert command in folder, or download and upload to OG with StockAndItems or Expiring for context)
- [ ] Sync
- [ ] Go to report page
- [ ] Click on the report
- [ ] Enter in something for item name or code
- [ ] Press Ok

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -111,7 +111,7 @@ const UIComponent = (props: ControlProps) => {
       ? t('error.json-bad-format-with-examples', {
           examples: examples.join('", "'),
         })
-      : zErrors ?? errors ?? customErrors;
+      : (zErrors ?? errors ?? customErrors);
 
   if (!props.visible) {
     return null;
--- a/client/packages/system/src/Report/components/ReportArgumentsModal.tsx
+++ b/client/packages/system/src/Report/components/ReportArgumentsModal.tsx
@@ -2,7 +2,7 @@ import React, { FC, useState } from 'react';
 
 import { JsonData, JsonForm } from '@openmsupply-client/programs';
 import { ReportRowFragment } from '../api';
-import { useDialog } from '@common/hooks';
+import { useDebounceCallback, useDialog } from '@common/hooks';
 import { DialogButton } from '@common/components';
 import { useAuthContext } from '@openmsupply-client/common';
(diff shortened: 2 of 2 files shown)


CarlosNZ (changes requested):
This doesn't seem to work -- I can still type quickly in the input and click okay and the args aren't captured. 

I thought we were just going to disable the button while debouncing?

andreievg (dismissed):
Looks great, thanks !

roxy-dao:
> This doesn't seem to work -- I can still type quickly in the input and click okay and the args aren't captured.
> 
> I thought we were just going to disable the button while debouncing?

Can't disable the okay button, since the store preferences is being populated in data now... 👀

CarlosNZ (commented):
Hey, just noticed that there is a (fortunately debounced) infinite re-render going on here too. (The `setData` method in `JsonForm` is being called every 750ms when the form is active even when nothing changes).

I don't think there's any rush to get this in though -- the original problem isn't a deal breaker, so we can keep experimenting.

CarlosNZ:
> Hey, just noticed that there is a (fortunately debounced) infinite re-render going on here too. (The `setData` method in `JsonForm` is being called every 750ms when the form is active even when nothing changes).

Actually, this problem is already in the base branch. I might have a quick go at fixing this today.

CarlosNZ:
> Actually, this problem is already in the base branch. I might have a quick go at fixing this today.

Okay, did a hotfix PR for the infinite re-render, as that needs to go into `develop` ASAP, whereas this is less urgent: https://github.com/msupply-foundation/open-msupply/pull/4483

PR #4467, opened 2024-07-29
Include source site id sync buffer rows after v6 push to OMS Central

### LINKED ISSUE #4466: Remote data not integrating after v6 push
## What went wrong? 😲

`sync_on_central` push method is not setting source site ID on sync buffer rows, so rows aren't being integrated.

The data is only integrated when the central server syncs and integrates.

## Expected behaviour 🤔

Remote data should be integrated on central server after v6 push is complete.

## How to Reproduce 🔨

1. Have OMS central and a remote site running
2. Create some records in remote site that sync to OMS Central (asset/R&R form...)
3. Sync the remote site
4. View sync_buffer on central - new sync buffer rows have been created, but with no source site id and no integration datetime
5. That data is not available in the relevant table on central server
6. Sync the central server
7. Sync buffer rows now have integration datetime/data exists in the right tables

## Your environment 🌱

- Open mSupply Version: 2.2 (develop)

- Platform:

- Database type:

jmbrunskill:
Fixes #4466

# 👩🏻‍💻 What does this PR do?

In a recent refactor we changed how we set the `source_site_id` on sync buffer rows. #4040
However, one function was hard coded to set this to none, resulting in incorrect setting of this source_site_id

Anything else we might have missed?

# 🧪 Testing

- [ ] Create a Cold Chain Equipment (or some other remote data on an OMS Site that syncs to OMS Central)
- [ ] Sync to OMS Central
- [ ] Check the records are integrated straight away (before would need to integrate)
- [ ] Re-initialise your remote server, check that newly create asset is downloaded correctly on re-initialisation

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/sync/api/common_records.rs
+++ b/server/service/src/sync/api/common_records.rs
@@ -108,8 +108,11 @@ impl CommonSyncRecord {
 
     pub(crate) fn to_buffer_rows(
         rows: Vec<CommonSyncRecord>,
+        source_site_id: Option<i32>,
     ) -> Result<Vec<SyncBufferRow>, ParsingSyncRecordError> {
-        rows.into_iter().map(|r| r.to_buffer_row(None)).collect()
+        rows.into_iter()
+            .map(|r| r.to_buffer_row(source_site_id))
+            .collect()
     }
 }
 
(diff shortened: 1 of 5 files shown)


lache-melvin (approved):
Works for me, thanks James! 🚀

andreievg (approved):
Looks great thanks, here is the test that fails in develop but passes with this change: https://github.com/msupply-foundation/open-msupply/pull/4469

PR #4469, opened 2024-07-30
4466-integration-test-for-remote-omsupply-data

andreievg:
Helper for validating #4466 in develop this test fails, but in the following PR it passes: https://github.com/msupply-foundation/open-msupply/pull/4463

# 👩🏻‍💻 What does this PR do?

Current omSuppy <-> omSupply central integration tests are limited to central data. Added remote data tester (basically a copy of existing [remote test for syncing with mSupply](https://github.com/msupply-foundation/open-msupply/blob/af61d89f9aa1541ed382e0d9d251e4bbd1f6605a/server/service/src/sync/test/integration/remote/mod.rs#L39-L107). Added one test case for asset row

We talked with @jmbrunskill briefly, when do we add sync integration test ? For now have the following:

* At least one integration test should be added for a records type (i.e. for omSupply central data -> pack_variant one for omSupply remote data -> asset, transfers -> invoice/requisition transfer tests)
* When sync errors arise, replicate with integration test for that records type (if possible)

It would be great if we had sync integration tests for all records types, not sure if we can invest the time into doing this though

# 🧪 Testing

Please see automated test description in integration/README.md

# 📃 Documentation

### DIFF
--- a/server/service/src/sync/test/integration/omsupply_central/mod.rs
+++ b/server/service/src/sync/test/integration/omsupply_central/mod.rs
@@ -1,10 +1,13 @@
+mod asset;
 mod pack_variant;
 mod test;
 
 use std::time::Duration;
 
+use repository::ChangelogRepository;
 use reqwest::Client;
 use url::Url;
+use util::assert_variant;
 
 use crate::sync::{
     test::{
@@ -14,11 +17,16 @@ use crate::sync::{
             init_test_context, SyncIntegrationContext,
         },
     },
+    translation_and_integration::integrate,
     CentralServerConfig,
 };
 
 use super::{GraphqlRequest, SyncRecordTester};
 
+/// For each test step:
+/// Upsert data to omSupply central via graphql
+/// Synchronise remote site
+/// Check integrated records exit
 async fn test_omsupply_central_records(identifier: &str, tester: &dyn SyncRecordTester) {
     // util::init_logger(util::LogLevel::Info);
     // Without re-initialisation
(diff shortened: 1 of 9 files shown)


server/service/src/sync/test/integration/omsupply_central/mod.rs
```diff
@@ -14,11 +17,16 @@ use crate::sync::{
             init_test_context, SyncIntegrationContext,
         },
     },
+    translation_and_integration::integrate,
     CentralServerConfig,
 };
 
 use super::{GraphqlRequest, SyncRecordTester};
 
+/// For each test step:
```
jmbrunskill, line 26: Thanks for adding all these explaining comments!

server/service/src/sync/test/integration/transfer/mod.rs
```diff
@@ -31,7 +31,6 @@ struct SyncIntegrationTransferContext {
     item1: ItemRow,
     item2: ItemRow,
     service_item: ItemRow,
-    currency: CurrencyRow,
```
jmbrunskill, line 34: image
I think we can remove currency, but need to remove it in more places...

jmbrunskill (approved):
Nice! Less code than I expected...

Just need to sort out currently properly.

server/service/src/sync/test/integration/transfer/mod.rs
andreievg: oops, i did `cargo check --features integration_test`, but I think it also needed a 'test' flag ¯\_(ツ)_/¯, thanks ! fc25690d81513ff42e148d3fe466b5dd2daceb2f

PR #4473, opened 2024-07-30
4411 detail view edit

### LINKED ISSUE #4411: Create R&R form detail view
Create a detail view for an R&R form. 

image

image

Items (from master list) should be grouped by category

Columns are (* = editable):

- Item name
- strength
- unit
- VEN
- initial balance
- quantity received *
- quantity consumed *
- adjusted quantity consumed 
- losses/adjustments *
- stock out duration (days) *
- final balance
- AMC
- Maximum quantity
- expiration date *
- requested quantity

Are these for requesting or responding side?
- Approved quantity
- comment

There should be a footer with save and submit buttons

lache-melvin:
Fixes #4411 

# 👩🏻‍💻 What does this PR do?

Adds editable detail view:

Screenshot 2024-07-31 at 11 32 18 AM

There are a few pieces that aren't quite ideal, but hoping ok to fix in the cleanup period?

# 🧪 Testing

- [ ] Create an R&R form
- [ ] You can edit the values
- [ ] The readonly calculated values adjust appropriately (adjusted consumption, final balance, AMC, requested quantity...)
- [ ] Confirming the lines greys them out
- [ ] Refresh before confirming clears the changes
- [ ] Finalise the form, you can no longer edit

# 📃 Documentation

  1.
  2.


client/packages/programs/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -1,60 +1,329 @@
...
+    const stockAvailableDays = periodLength - stockOutDuration;
+    const adjustedQuantityConsumed = stockAvailableDays
+      ? quantityConsumed * (periodLength / stockAvailableDays)
+      : quantityConsumed;
+
+    // This calculation might be a plugin in future!
+    const averageMonthlyConsumption =
+      // Average of the last 3 months (including the current month)
+      // TODO: what if don't want to consider previous months?
+      (2 * previousAverageMonthlyConsumption + adjustedQuantityConsumed) / 3;
```
lache-melvin, line 134: i.e. on first R&R form, probably DO want AMC to only consider this period...

- make AMC an editable field?
- return list of previous monthly consumption values, so if it's empty we only consider this period? ... Would we store this on the line or query it each time?

client/packages/programs/src/RnRForms/DetailView/Footer.tsx
```diff
@@ -6,17 +6,38 @@ import {
...
   const { navigateUpOne } = useBreadcrumbs();
+  const { error, info, success } = useNotification();
   const {
     query: { data },
     finalise: { finalise, isFinalising },
   } = useRnRForm({ rnrFormId });
 
+  const onFinalise = async () => {
+    if (linesUnconfirmed) {
+      info(t('messages.all-lines-must-be-confirmed'))();
```
lache-melvin, line 30: This isn't quite ideal, requiring them to click through and confirm each line...

Probably should have a dirty state instead, if any lines are dirty, prevent finalising (and warn on refresh/navigate away) - this felt a little more complex given the edit state is managed on a per line basis, so was hoping to do this in another PR...

server/service/src/rnr_form/generate_rnr_form_lines.rs
```diff
@@ -171,48 +174,66 @@ pub fn get_amc(
...
-        RnRFormLineRowRepository::new(connection).find_many_by_rnr_form_ids(prev_form_ids)?;
+    for form in prev_forms {
+        let period_length_in_days = get_period_length(&form.period_row);
+        let period_months = period_length_in_days as f64 / NUMBER_OF_DAYS_IN_A_MONTH;
 
-    for row in rows.into_iter() {
-        let amc_values = form_lines_by_item_id
-            .entry(row.item_id.clone())
-            .or_insert(vec![]);
-        amc_values.push(row.average_monthly_consumption);
```
lache-melvin, line 212: Realised that returning previous `average_monthly_consumption` values here meant we were averaging the averages, rather than averaging the consumption over the last 3 periods...

server/service/src/rnr_form/update.rs
```diff
@@ -209,19 +236,20 @@ fn generate(line_data: Vec<(UpdateRnRFormLine, RnRFormLineRow)>) -> Vec<RnRFormL
                     average_monthly_consumption,
                     stock_out_duration,
                     adjusted_quantity_consumed,
+                    initial_balance, // TODO; snapshot and entered?
```
lache-melvin, line 239: Still a bit unsure about what we want to do if they provide a different initial_balance... we probably want to check whether it matches what we think the initial balance should have been, and create another adjustment as of start of period?

Whether to do this with another snapshot/entered check, or just recalculate on finalise and adjust accordingly?

client/packages/programs/src/RnRForms/DetailView/RnRFormLine.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+    const adjustedQuantityConsumed = stockAvailableDays
+      ? quantityConsumed * (periodLength / stockAvailableDays)
+      : quantityConsumed;
+
+    // This calculation might be a plugin in future!
+    const averageMonthlyConsumption =
+      // Average of the last 3 months (including the current month)
+      // TODO: what if don't want to consider previous months?
+      (2 * previousAverageMonthlyConsumption + adjustedQuantityConsumed) / 3;
```
lache-melvin, line 61: i.e. on first R&R form, probably DO want AMC to only consider this period...

- make AMC an editable field?
- return list of previous monthly consumption values, so if it's empty we only consider this period? ... Would we store this on the line or query it each time?

client/packages/programs/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -6,54 +6,78 @@ import {
...
+  data,
+  saveLine,
+  periodLength,
+  disabled,
+}: ContentAreaProps) => {
   const t = useTranslation('replenishment');
 
-  return data.length === 0 ? (
+  // TODO: move to backend, should join on item and sort by name!
+  const lines = data.sort((a, b) => (a.item.name > b.item.name ? 1 : -1));
```
lache-melvin, line 27: Please note this will be fixed in #4475!

client/packages/programs/src/RnRForms/DetailView/RnRFormLine.tsx
mark-prins: I like the idea of storing a list of previous average consumption values. then averaging across all available values. I think that's a more accurate result than the (2*previous AMC + current month)/3

unsure how to store the list though. we could have a hard coded count of 3, and just add two 'previous AMC' columns to the table. or a JSON - could we deserialize that to a Vec?

other clients may well want a different number of previous AMCs to be used

client/packages/programs/src/RnRForms/DetailView/RnRFormLine.tsx
```diff
@@ -0,0 +1,268 @@
...
+      : draft.item.venCategory;
+
+  const textColor =
+    disabled || draft.confirmed
+      ? theme.palette.text.disabled
+      : theme.palette.text.primary;
+
+  const readOnlyColumn = {
+    backgroundColor: theme.palette.background.drawer,
+    padding: '5px',
```
mark-prins, line 90: :nit given that this is used in a `style` prop, it can be just numeric i.e. `padding: 5`

server/service/src/rnr_form/update.rs
mark-prins: if the process is working then we'll only be taking initial_balance from the previous R&R, and it's only enterable on the first R&R created for the period & program combination. that's creating validation complexity which I don't think we need right now

mark-prins (approved):
🚢  I think that's fabulous!
There's a few things to tweak, as you say. The layout looks nice, and the edit cells are great; overall I think it's great to use and we should get it in 😉 

I played with fixing the first two columns and have those changes in [this branch](https://github.com/msupply-foundation/open-msupply/compare/4411-detail-view-edit...4411-detail-view-edit(suggestion))

it's not perfect but possibly usable: here's how it looks..

https://github.com/user-attachments/assets/e8d79071-c92d-4e68-a2c0-052bd5f88fa4

A couple of other minor things:
* If you get an error, you get two errors. providing an `onError` to the mutation resolves that one (see below)
* It would be nice to round the decimal points of the calculated values (see below)

Screenshot 2024-07-31 at 3 10 23 PM

Screenshot 2024-07-31 at 2 40 11 PM

PR #4474, opened 2024-07-31
Improve sync messaging

mark-prins:
Fixes #4006

# 👩🏻‍💻 What does this PR do?
Clarifies the messaging on the sync page.
This now shows some explanatory text, the datetime of the last sync start and completion, the duration of the last sync, and the completion datetime of the last successful sync.

Here's how that looks:

Screenshot 2024-07-31 at 11 54 38 AM

Screenshot 2024-07-31 at 11 54 45 AM

Screenshot 2024-07-31 at 11 55 10 AM

## 💌 Any notes for the reviewer?

no particular notes.
# 🧪 Testing

- [ ] test the migrations with postgres and sqlite
- [ ] check that the duration column is populated
- [ ] check that the display looks ok

# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [x] **These areas should be updated or checked**: 
  1. Need to update sync screenshots

PR #4477, opened 2024-07-31
4417 create internal order on rr form submit

### LINKED ISSUE #4417: Create Internal Order on R&R form submit
Once the user is happy with the R&R form, they should click the `Submit` button, this should create an internal order to supplying store. 

After submit, R&R form should no longer be editable.

Acceptance Criteria
- [ ] Internal Order is created to supplying store with requested quantity
- [ ] Should be some way to identify that the internal order came from a R&R form
- [ ] Internal Order is automatically marked as `sent`

Nice to haves
- [ ] Option to keep internal order as editable, incase they want to change supplier or split some of the lines to a different store (is possible in some installations)
- [ ] Populate R&R Form style fields in open-mSupply requisition where they match...

jmbrunskill:
Fixes #4417

# 👩🏻‍💻 What does this PR do?

Creates an Internal Order (aka Requisition) when an R&R Form is initialised.
It goes to same supplier.

Tests are fixed in Mark's PR here: https://github.com/msupply-foundation/open-msupply/pull/4474
Will merge when he's merged to minimise conflicts

I wanted to re-use service layer logic for requistions, or program requisitions but neither quite fitted the logic, so creating at the repo layer. Interested in thoughts on that? Is it ok? Better to combine something?

We should probably create some kind of more formal link between and R&R form and Requisition.
If we store it on the requisition we'd need to update mSupply to capture the new data, maybe a back link from the R&R form to the requistion?

# 🧪 Testing

- [ ] Create and R&R form
- [ ] Finalise it
- [ ] Check that internal order is created

# 📃 Documentation

  1.
  2.


server/repository/src/db_diesel/item_row.rs
```diff
@@ -153,7 +153,7 @@ impl<'a> ItemRowRepository<'a> {
         Ok(result)
     }
 
-    pub fn find_many_by_id(&self, ids: &[String]) -> Result<Vec<ItemRow>, RepositoryError> {
+    pub fn find_many_by_id(&self, ids: &Vec<String>) -> Result<Vec<ItemRow>, RepositoryError> {
```
jmbrunskill, line 156: Nothing else was using this surprisingly!
Seems much easier to create a vec rather than an array when we don't know the size...

server/service/src/rnr_form/finalise.rs
```diff
@@ -90,3 +102,89 @@ impl From<RepositoryError> for FinaliseRnRFormError {
...
+        colour: None,
+        comment: Some("Automatically created from R&R Form".to_string()),
+        their_reference: None,
+        max_months_of_stock: 0.0,
+        min_months_of_stock: 0.0,
+        approval_status: None,
+        linked_requisition_id: None,
+        program_id: Some(rnr_form.program_id),
+        period_id: Some(rnr_form.period_id),
+        order_type: None, // Should we capture this in the RnR form?
```
jmbrunskill, line 136: We don't have order type in R&R Forms but is required for program requisitions?
Do we want to add to R&R? If we can default when there's only one, it wouldn't be too bad...

server/service/src/rnr_form/finalise.rs
```diff
@@ -90,3 +102,89 @@ impl From<RepositoryError> for FinaliseRnRFormError {
...
+        let requisition_line = RequisitionLineRow {
+            id: uuid(),
+            requisition_id: requisition_row.id.clone(),
+            item_link_id: rnr_form_line.item_id.clone(),
+            item_name: item_map
+                .get(&rnr_form_line.item_id)
+                .map(|item| item.name.clone())
+                .unwrap_or_default(),
+            requested_quantity: rnr_form_line.requested_quantity,
+            suggested_quantity: rnr_form_line.requested_quantity, // TODO This is user enterable, should we also re-calc this?
```
jmbrunskill, line 175: Not quite sure what the suggested_quantity should be, I think it probably should be the calculated value (but user can over write it) Do I re-calc here?

server/service/src/rnr_form/finalise.rs
```diff
@@ -90,3 +102,89 @@ impl From<RepositoryError> for FinaliseRnRFormError {
...
+            supply_quantity: 0.0,
+            available_stock_on_hand: rnr_form_line.final_balance,
+            average_monthly_consumption: rnr_form_line.average_monthly_consumption,
+            snapshot_datetime: rnr_form.finalised_datetime,
+            approved_quantity: 0.0,
+            approval_comment: None,
+            comment: None,
+            // TODO add Cust_pre_stock_balance, Cust_stock_received, Cust_stock_ord, Cust_stock_adj (in mSupply but no OMS Yet)
```
jmbrunskill, line 183: Should we bring these fields though open-mSupply? Maybe we can do it just at the translation layer in sync?

server/service/src/sync/test/test_data/rnr_form_line.rs
```diff
@@ -34,10 +34,13 @@ fn rnr_form_line_1() -> RnRFormLineRow {
...
-        quantity_received: 0.0,
-        quantity_consumed: 0.0,
+        snapshot_quantity_received: 0.0,
+        entered_quantity_received: None,
+        snapshot_quantity_consumed: 0.0,
+        entered_quantity_consumed: None,
         adjusted_quantity_consumed: 0.0,
-        adjustments: 0.0,
+        snapshot_adjustments: 0.0,
+        entered_adjustments: None,
```
jmbrunskill, line 43: This test isn't fully fixed, just fixed the compilation part waiting for Mark's PR to get merged first.

server/service/src/rnr_form/finalise.rs
mark-prins: the order type came from IC requirements, to have 'normal' and 'emergency' orders. In this context, having a 'normal' order makes the most sense.

in theory.. we can link an RNR form to `period` and then to `program_requisition_settings` and then to `program_requisition_order_type` to get the `name` out. however, we'd only want to take non-emergency and we aren't pulling that field through.

might have to add to the R&R form - but yes, defaulting the value if there's only one

server/service/src/rnr_form/finalise.rs
```diff
@@ -90,3 +100,89 @@ impl From<RepositoryError> for FinaliseRnRFormError {
...
+        let requisition_line = RequisitionLineRow {
+            id: uuid(),
+            requisition_id: requisition_row.id.clone(),
+            item_link_id: rnr_form_line.item_id.clone(),
+            item_name: item_map
+                .get(&rnr_form_line.item_id)
+                .map(|item| item.name.clone())
+                .unwrap_or_default(),
+            requested_quantity: rnr_form_line.requested_quantity,
+            suggested_quantity: rnr_form_line.requested_quantity, // TODO This is user enterable, should we also re-calc this?
```
mark-prins, line 173: the 'maximum quantity' on the R&R form is essentially the suggested quantity and I'd use that.

server/service/src/rnr_form/finalise.rs
mark-prins: I believe that we'll need them at some point. will need to discuss before we implement that though - I'm not sure why they haven't been done yet.

server/service/src/rnr_form/finalise.rs
```diff
@@ -90,3 +102,89 @@ impl From<RepositoryError> for FinaliseRnRFormError {
...
+        name_link_id: rnr_form.name_link_id,
+        store_id: rnr_form.store_id,
+        r#type: RequisitionType::Request,
+        status: RequisitionStatus::Sent,
+        created_datetime: Utc::now().naive_utc(),
+        sent_datetime: None,
+        finalised_datetime: None,
+        expected_delivery_date: None,
+        colour: None,
+        comment: Some("Automatically created from R&R Form".to_string()),
```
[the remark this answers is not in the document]
mark-prins: agree - would be nice to have a link between R&R and requisition. I would add a requisition ID field to the R&R, as you say, that's least impact

mark-prins (approved):
👍 Working nicely 😃 
that looks great thanks James!
Agree about adding a requisition / R&R link; and there's a couple of other comments re fields to populate. nothing show stoppery though.

server/service/src/rnr_form/finalise.rs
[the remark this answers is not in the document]
jmbrunskill: Looks like we already have one, I'll populate it now...

server/service/src/rnr_form/finalise.rs
jmbrunskill: Just to confirm, are you suggesting we try to look it up here, or we add it to the RNR Form process?

mark-prins: the more I think about it, the more i'm leaning to doing a lookup here. though we could equally leave it blank, as I don't see that it is appropriate for the resulting internal order anyway.

Just leave it as `None`.

PR #2913, opened 2024-02-08
2720 cold chain missing data create temporary breaches

jmbrunskill:
Fixes #2720

# 👩🏻‍💻 What does this PR do? 
Creates temporary breach records if a temperature log references a breach that doesn't already exist.
This shouldn't be necessary with other code changes in CCA, but did solve a problem for me with breaches created before the app was integrated with Open-mSupply...

# 🧪 How has/should this change been tested? 

## 💌 Any notes for the reviewer?

Should we include this code? Or just rely on the CCA app update?

I had to add Excusion to the frontend as a breach type, panic seemed to extreme a response to this!

## 📃 Documentation

_No user facing changes_

PR #2589, opened 2023-12-05
Fix temperature breach filter

mark-prins:
Fixes #2537

# 👩🏻‍💻 What does this PR do? 
 
Could not repro the stated issue. A related issue was raised in conversation - that the breaches aren't showing.

That's the part which I've addressed here - it turns out that the breach filter was a little over complicated and didn't function as required. The filter was looking for `temperature_breach` records.. the query was

```
SELECT "temperature_log"."id", "temperature_log"."temperature", "temperature_log"."sensor_id", "temperature_log"."location_id", "temperature_log"."store_id", "temperature_log"."datetime", "temperature_log"."temperature_breach_id" 
FROM "temperature_log" 
WHERE "temperature_log"."temperature_breach_id" IN 
(SELECT "temperature_breach"."id" FROM "temperature_breach" WHERE "temperature_breach"."id" = 'fff4bac7-2256-43cf-acb9-78acfc8cc7e0') 
ORDER BY "temperature_log"."datetime" DESC
```

which will not work for new breaches, the breach has to be added before the breach can be added!

I've changed the filter to look for an `id` only, which simplifies things and allows breaches to be added.

# 🧪 How has/should this change been tested? 

Configure a CCA with a sensor and sync with omSupply. get the sensor into a breach and check if the breach shows in oms.

## 📃 Documentation

_No user facing changes_

PR #4479, opened 2024-07-31
Add min/max for excursions

### LINKED ISSUE #3033: Cold Chain: Some breaches don't have a Max/Min temperature attached to it
## What went wrong? 😲

We noticed some of the temperature breaches don't have a max/min temperature attached to it on the 'Monitoring' screen:

Screenshot 2024-02-21 at 10 04 23 AM

Maria and Adam noticed there was data on the cold chain app, but for some reason it was not brought across oms.

## Expected behaviour 🤔

All breaches should have a min/max temperature associated with.

## How to Reproduce 🔨

1. Go to 'Cold chain'
2. Click on 'Monitoring' > breaches
3. See column 'Max/min' empty for some breaches

## Your environment 🌱

- Version: 1.6.02
- mSupply version: v7.12.01
- Platform:

jmbrunskill:
Fixes #3033

# 👩🏻‍💻 What does this PR do?

Provides a min or max temperature for an excusion breaches.
If the excursion contains record above and below the limits, the last min or max will be taken.
I don't think these are really used but does solve at least one potential reason for missing the min/max value.

This doesn't directly address the concerns in the original issue as this wasn't for excursion events.
However, as per Roxy's investigation the temperature_log row are missing from the database in these cases.

I haven't been able to recreate that with my setup, and there have been a number of improvements to both CCA and OMS to reduce the chance of this happening since.

Most importantly:
https://github.com/msupply-foundation/open-msupply/pull/2913
But also:
https://github.com/msupply-foundation/msupply-cold-chain/issues/280
https://github.com/msupply-foundation/open-msupply/pull/2589

Based on this I'm going to declare this issue closed, and if we have production issues with it happy to investigate further at that time.

Having no min/max is possible by design, for example with some Q tags we just get a breach notification without any temperature logs.

# 🧪 Testing

- [ ] Create a excursion event with temperature logs
- [ ] Check the min/max comes through correctly

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/cold_chain/query_temperature_breach.rs
+++ b/server/service/src/cold_chain/query_temperature_breach.rs
@@ -2,8 +2,8 @@ use repository::temperature_breach::{
     TemperatureBreach, TemperatureBreachFilter, TemperatureBreachRepository, TemperatureBreachSort,
 };
 use repository::{
-    EqualFilter, PaginationOption, RepositoryError, StorageConnection,
-    TemperatureBreachRowRepository, TemperatureBreachType, TemperatureLogFilter,
+    EqualFilter, PaginationOption, RepositoryError, StorageConnection, TemperatureBreachRow,
+    TemperatureBreachRowRepository, TemperatureBreachType, TemperatureLog, TemperatureLogFilter,
     TemperatureLogRepository,
 };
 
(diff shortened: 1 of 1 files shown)


server/service/src/cold_chain/query_temperature_breach.rs
```diff
@@ -46,6 +46,34 @@ pub fn get_temperature_breach(
...
+        if log.temperature_log_row.temperature < breach.threshold_minimum
+            && log.temperature_log_row.temperature < min_or_max
+        {
+            min_or_max = log.temperature_log_row.temperature;
+            min_or_max_option = Some(min_or_max);
+        } else if log.temperature_log_row.temperature > breach.threshold_maximum
+            && log.temperature_log_row.temperature > min_or_max
+        {
+            min_or_max = log.temperature_log_row.temperature;
+            min_or_max_option = Some(min_or_max);
```
lache-melvin, line 70: Forgive my lack of knowledge in this area... is an excursion breach only in one direction?

I.e. could this list of logs include some logs below min and some above max? In which case you'd need to consider which was further out of range to be the min/max??

lache-melvin (approved):
Ah sorry meant to approve, just with the question! LGTM!!

server/service/src/cold_chain/query_temperature_breach.rs
jmbrunskill: I think the intention is that there should be two excursion created if we somehow went from below min the above max without anything in between.

Right now I don't think excursion are really used, so I think this solution of the last min or max is good enough for 99% of situations.

PR #4483, opened 2024-07-31
Reports - fix infinite re-render

### LINKED ISSUE #4482: Reports -- infinite re-render of report arguments modal
## What went wrong? 😲

The `setData` method is being called over and over.

CarlosNZ:
Fixes #4482

# 👩🏻‍💻 What does this PR do?

Hotfix for infinite re-render. Problem was the "data" object being passed to the JsonForm component was a different object from that being set by "updateData", so was being considered "changed" even when the content hadn't actually changed.

(If you actually want to test/demo, put a `console.log()` in the `updateData` prop of JsonForm component -- in the base branch it is looping indefinitely, now it isn't)


client/packages/system/src/Report/components/ReportArgumentsModal.tsx
```diff
@@ -62,14 +62,14 @@ export const ReportArgumentsModal: FC<ReportArgumentsModalProps> = ({
           variant="ok"
           disabled={!!error}
           onClick={async () => {
-            onArgumentsSelected(report, data);
+            onArgumentsSelected(report, jsonData);
```
CarlosNZ, line 65: `jsonData` is derived from `data` (adds 3 fieds), so should only get sent to the output function, not the local form state

roxy-dao: The three fields don't load for me anymore.

CarlosNZ: Whereabouts? I'm seeing them in both the URL and the GraphQL query. 🤔

CarlosNZ: Ooh, I've never seen them on the modal at all, I guess I don't have a ui-schema that includes them. 🤷‍♂️

Screenshot 2024-08-01 at 6 00 47 PM

If they're meant to be there, I'll just need to rejig this a little bit.

roxy-dao: If you re-upserts any of the new reports in reports repo (still in PR) they should be there

CarlosNZ: Okay @roxy-dao, have changed it to initialise the `data` state with these values already included. 03b1104bb

andreievg (approved):
Looks good to me. Thanks

roxy-dao (approved):
Cool~ thanks Carl

PR #4487, opened 2024-08-01
filter report contexts for r&r forms

### LINKED ISSUE #4415: Print the current R&R form
Add a Print button to the App Bar Buttons to print the current R&R form.

This should include all of the rows, as well as the following header data:

- Health facility
- District
- Reporting period (? unsure of translation)
- Requisition period

Create a new report context for R&R forms, and add a new report!

lache-melvin:
Fixes #4415 

# 👩🏻‍💻 What does this PR do?

R&R report PR [here](https://github.com/msupply-foundation/open-msupply-reports/pull/76) - use the upsert script there to insert the report and test this PR.

This PR adds filtering of Requisition subcontexts, so only R&R report shows on R&R page, and it doesn't show on other requisition pages.

# 🧪 Testing

- [ ] Upsert the R&R report
- [ ] Go to an R&R form
- [ ] Click Print - report is generated and print window opens immediately
- [ ] Only Internal Order and Requisition pages, there is no option to print an R&R report

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
@@ -59,6 +59,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
         <ReportSelector
           context={ReportContext.Requisition}
           onPrint={printReport}
+          // Filters out reports that have a subContext (i.e. `R&R`)
           queryParams={{ filterBy: { subContext: { equalAnyOrNull: [] } } }}
         >
           <LoadingButton
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
@@ -59,6 +59,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
         <ReportSelector
           context={ReportContext.Requisition}
           onPrint={printReport}
+          queryParams={{ filterBy: { subContext: { equalAnyOrNull: [] } } }}
         >
           <LoadingButton
             variant="outlined"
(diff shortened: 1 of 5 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
```diff
@@ -59,6 +59,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
         <ReportSelector
           context={ReportContext.Requisition}
           onPrint={printReport}
+          queryParams={{ filterBy: { subContext: { equalAnyOrNull: [] } } }}
```
lache-melvin, line 62: Would have preferred `notEqualTo: "R&R"`, but this only returns reports where subcontext is not null as well... and its my understanding that Requisition reports don't use any subcontexts? Any risk in adding this filter now??

mark-prins: this has the advantage of also excluding the next sub-context-specific report as well as the 'R&R' one. Using the notEqualTo would require keeping up to date if we did this again 🤷 

shame though - I thought that `notEqualTo` would work in this case. It needs to be a notEqualToOrNull I guess

mark-prins: this line might need a comment explaining why it was necessary. If I was coming in cold, I'd probably say we could remove that filter!

client/packages/system/src/Report/api/hooks/useReportList.ts
```diff
@@ -39,15 +39,12 @@ export const useReportList = ({
...
-            context: context ? { equalTo: context } : null,
-            subContext: subContext ? { equalTo: subContext } : null,
-          }
-        : null;
     const query = await reportApi.reports({
-      filter: { ...filterBy, ...filter },
+      filter: {
+        ...filterBy,
+        ...(context ? { context: { equalTo: context } } : null),
+        ...(subContext ? { subContext: { equalTo: subContext } } : null),
```
mark-prins, line 46: this is a much nicer way to construct the params!
I'm not sure the scenario of having a subContext without a context is valid - but there's no reason to change that existing logic, nor to have a test for it in here

mark-prins (approved):
🎉  works like a charm!

Issue #4405, opened 2024-07-14
Report and Requisition (R&R) forms


mark-prins:
### Overview

[Spec document](https://docs.google.com/document/d/1wZ4R1Gq2GbqAKfCJ0KDXpMDnqEgBNAOLNnnXzEkcQ4I/edit)

Add the ability for health facilities to report on a customisable set of usage metrics as well as the consumption and stock levels of a pre-defined list of items.

The period of reporting should be configurable. This reporting system replaces an existing, excel based, form and should have a similar look and feel to using excel. Some fields / values are calculated, as in the excel version where some cells are showing the results of formulae. 

When complete, forms are submitted and will be sent to the supplying store and can no longer be edited.

### Tasks
- [x] #4408
  - [x] #4409
  - [x] #4410 
- [x] #4412
- [x] #4411
  - [x] #4414
  - [x] #4415
  - [x] #4416
  - [x] #4417 
- [x] #4457
- [x] #4510
- [x] #4513

Future:
- [ ] #4418 (implement with backdated transactions?)
- [ ] #4421 (these can just happen as a dashboard for now)
- [ ] #4475 (pending decision)
- [ ] The number of months used in the calculation is configurable (what calculation?)
- [ ] R&R forms are filterable

#### list view

![Screenshot 2024-07-15 at 12 16 18 PM](https://github.com/user-attachments/assets/a5e60fff-1b3f-4750-9b6e-fdbd657373b2)

#### detail view

![Screenshot 2024-07-15 at 12 22 49 PM](https://github.com/user-attachments/assets/6c2a9ba7-9f08-445f-80f8-8f0837bf0131)

#### editing

![Screenshot 2024-07-15 at 12 21 05 PM](https://github.com/user-attachments/assets/b2f01d99-2b87-4d44-9bc3-a985b25201bd)

### Tests

SETUP
1. You'll need to enable the `Open mSupply: Uses program module` store pref in mSupply
2. You'll need at least one master list configured to be visible in your store, which is a program master list
3. You'll need at least one period schedule with some periods configured - and the period schedule to the master list under the `Program settings` tab
4. In Open mSupply UI, go to Replenishment > R&R Forms
5.  Ensure there has been some activity for the items in your master list during the period you are creating the R&R form for
  a. Have transactions on the days bordering the start/end of the period to confirm values are captured correctly!
[...]

PR #4489, opened 2024-08-02
Fix opening balance for first R&R Form when stock is reserved.

### LINKED ISSUE #4472: R&R initial balance returning negative values!
## What went wrong? 😲

Screenshot 2024-07-31 at 10 40 47 AM

Its calculating based on AVAILABLE stock on hand now (from the the stock_on_hand view) but it actually needs to use TOTAL stock on hand...

## Expected behaviour 🤔

Balances should always be >= 0.

## How to Reproduce 🔨

1. Have some reserved stock for an item in program master list
2. Create R&R form for that program
3. Backwards calcs of the SOH at start of period are negative!

## Your environment 🌱

- Open mSupply Version: 2.2 develop

- Platform:

- Database type:

jmbrunskill:
Fixes #4472

# 👩🏻‍💻 What does this PR do?

Adds total_stock_on_hand to stock_on_hand view
Changes RnR Form logic to use total_stock_on_hand

Also fixed an RnR Form test.
There's another unrelated test failing though.

# 🧪 Testing

- [ ] Setup for testing R&R forms (see: https://github.com/msupply-foundation/open-msupply/issues/4405)
- [ ] Create a stockline that will be part of your RnR Form with less available stock, than total stock. (This can be done by creating an outbound shipment, with a stock stock assigned, but leave the outbound shipment status as `Allocated` (Don't confirm `Picked`)
- [ ] Check the opening `stock on hand` is calculated correctly when the R&R form is created for that item.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
+++ b/server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
@@ -5,7 +5,13 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
     sql!(
         connection,
         r#"
+        DROP VIEW IF EXISTS stock_on_hand;
         DROP VIEW IF EXISTS store_items;
+        "#,
+    )?;
+    sql!(
+        connection,
+        r#"
         CREATE VIEW store_items AS
         SELECT i.id as item_id, sl.store_id, sl.pack_size, sl.available_number_of_packs, sl.total_number_of_packs
         FROM
@@ -18,8 +24,6 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
     sql!(
         connection,
         r#"
-        DROP VIEW stock_on_hand;
-
         CREATE VIEW stock_on_hand AS
         SELECT
           'n/a' AS id,
--- a/server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
+++ b/server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
@@ -1,0 +1,60 @@
+use crate::{migrations::sql, StorageConnection};
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    // re-create stock_on_hand and store_items
+    sql!(
+        connection,
+        r#"
+        DROP VIEW IF EXISTS store_items;
+        CREATE VIEW store_items AS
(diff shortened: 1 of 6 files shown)


server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
```diff
@@ -0,0 +1,64 @@
+use crate::{migrations::sql, StorageConnection};
+
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    // re-create stock_on_hand and store_items
+    sql!(
+        connection,
+        r#"
+        DROP VIEW IF EXISTS stock_on_hand;
```
jmbrunskill, line 8: Postgres required both views to be re-created as they're dependant...

server/service/src/rnr_form/tests/finalise.rs
```diff
@@ -71,11 +71,6 @@ mod finalise {
             .context(mock_store_a().id, "".to_string())
             .unwrap();
 
-        let last_requisition_number = RequisitionRowRepository::new(&context.connection)
```
jmbrunskill, line 74: This was just a test warning, I'd removed the need for this already but hadn't removed this query...

server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
```diff
@@ -0,0 +1,64 @@
...
+          LEFT OUTER JOIN (
+            SELECT
+              item_id,
+              store_id,
+              SUM(pack_size * available_number_of_packs) AS available_stock_on_hand,
+              SUM(pack_size * total_number_of_packs) AS total_stock_on_hand
+            FROM
+              store_items
+            WHERE
+              store_items.available_number_of_packs > 0 OR store_items.total_number_of_packs > 0
```
lache-melvin, line 54: I wondered whether this would affect anything, if there was anywhere that we were expecting all stock_on_hand lines to have available stock, but a quick code search seems like it's fine... 🤷‍♀️

lache-melvin (approved):
LGTM! Thanks James!

server/repository/src/migrations/v2_02_00/stock_on_hand_add_total_stock.rs
jmbrunskill: Good call to check! I also think it's ok, but maybe worth mentioning in the chat just so people are aware...

PR #4490, opened 2024-08-02
Allow settings to be reset

### LINKED ISSUE #4485: Display settings not updated when cleared
## What went wrong? 😲

If the custom display settings (theme and/or logo) are set and then cleared - only the browser which clears the settings will see the cleared version.

## Expected behaviour 🤔
Changing or clearing should update for all users.

## How to Reproduce 🔨

using browser (a)
1. login, navigate to settings
2. set the theme and/or logo

using browser (b)
1. login, you'll see the display changes

using browser (a)
1. clear the custom display settings
2. now logout and login and refresh the page. you'll keep the cleared settings

using browser (b)
1. logout and login: the custom settings are shown
2. refresh page: still custom settings
3. go to settings > display settings and clear: the custom settings are cleared

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #4485

# 👩🏻‍💻 What does this PR do?

When the theme or logo settings are disabled, the value is set to an empty string. The GQL endpoint was interpreting empty as the same condition as a matching hash i.e. no change required on the client.

This change is to distinguish those two cases. Now the GQL query will only return `null` when the hash matches.
The original bug is this:

* set a custom theme or logo in one browser
* browse to the site in another browser
* clear the setting in the first browser

in the second browser - it requests display settings. It get a response of 

```
{
  "data": {
    "displaySettings": {
      "customTheme": null,
      "customLogo": null
    }
  }
}
```

which the client app interprets as 'no change required' and therefore it retains the customised logo and/or theme.

This change allows the second browser to get an empty string response, which it then uses to update its display settings. 

# 🧪 Testing

- [ ] set settings in one browser, login in another, clear in first browser, return to second. you will likely need to reload page to get the changes to show (not ideal but better than it was!)

# 📃 Documentation

  1.
  2.


andreievg (approved):
Looks good to me, thanks

server/service/src/rnr_form/tests/finalise.rs
```diff
@@ -71,10 +71,10 @@ mod finalise {
             .context(mock_store_a().id, "".to_string())
             .unwrap();
 
-        let last_requisition_number = RequisitionRowRepository::new(&context.connection)
-            .find_max_requisition_number(repository::RequisitionType::Request, &mock_store_a().id)
-            .unwrap()
-            .unwrap_or_default();
+        // let last_requisition_number = RequisitionRowRepository::new(&context.connection)
```
mark-prins, line 74: there were complaints. so I've commented out 🤷

server/repository/src/mock/rnr_form.rs
```diff
@@ -58,6 +59,7 @@ pub fn mock_rnr_form_b_line_a() -> RnRFormLineRow {
         snapshot_adjustments: -1.0,
         final_balance: 7.0,
         average_monthly_consumption: 7.0,
+        adjusted_quantity_consumed: 7.0,
```
mark-prins, line 62: popped these in to fix the tests

server/service/src/rnr_form/tests/insert.rs
```diff
@@ -301,7 +301,7 @@ mod insert {
             mock_rnr_form_b_line_a().final_balance
         );
         // AMC considers previous form
-        assert_eq!(form_lines[0].average_monthly_consumption, 4.0); // 5 (A) + 7 (B) + 0 (this period) / 3
+        assert_eq!(form_lines[0].average_monthly_consumption, 4.026696329254727); // (5 * 31 / 30) (A) + (7 * 31  30) (B) + 0 (this period) / 3
```
mark-prins, line 304: offroad: update to fix the tests

server/service/src/sync/test/test_data/rnr_form_line.rs
```diff
@@ -23,7 +23,8 @@ const RNR_FORM_LINE_1: (&str, &str) = (
         "expiry_date": null,
         "requested_quantity": 0.0,
         "comment": null,
-        "confirmed": false
+        "confirmed": false, 
+        "previous_average_monthly_consumption": 0.0
```
mark-prins, line 27: offroad: popped these in to fix the tests

andreievg (changes requested):
oops, sorry my approval was just for the display settings, will have a look at other changes in a bit

andreievg (approved):
Thanks @mark-prins, looks good as far as i can tell

Issue #385, opened 2022-07-26
Date of birth / age control


clemens-msupply:
Widget that allows to set the date of birth either through a calendar or by setting the age.

PR #4491, opened 2024-08-02
#3641 Show age in months/days for <1

### LINKED ISSUE #3641: Patient ages: display as months and days if under 1 year old

Issue raised by PLM Sierra Leone. Seems small but actually makes a big difference to users. If we don't do it and display the age as 0 then they have to work out the actual age in months and days.

If a patient's age is less than 1 year then display as "xx months, yy days". They paid to have this done in mobile (now live) and will need this in Open mSupply before they roll out to more sites (planned for Q4 2024 or Q1 2025 approximately).

### Additional context 💌

Gary said we'd do this for free because it's small and they're good people (strategic decision will little real cost to us but lots of goodwill earned).

CarlosNZ:
Fixes #3641

# 👩🏻‍💻 What does this PR do?

Update the "DateOfBirth" component to display the Age in months/days when the person's age is less than 1. To avoid additional complexity, when in this state, we just *display* the age rather than making it editable:

Screenshot 2024-08-02 at 3 59 42 PM

Do you think we should make this a configurable option for the DOB component, or is it fine to work like this all the time?

# 🧪 Testing

Go to a Patient, and change their date of birth to a value less than one year ago.


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -162,6 +162,11 @@
   "label.additional-info": "Additional Info",
   "label.address": "Address",
   "label.age": "Age",
+  "label.age-months_one": "{{months}} month",
+  "label.age-months_other": "{{months}} months",
+  "label.age-days_one": "{{days}} day",
+  "label.age-days_other": "{{days}} days",
+  "label.age-months-days": "$t(label.age-months, { \"count\": {{months}}, \"months\": {{months}} }), $t(label.age-days, { \"count\": {{days}}, \"days\": {{days}} })",
```
CarlosNZ, line 169: Nested language strings! https://www.i18next.com/translation-function/nesting

I think this is the first time we've done this. The syntax is pretty gnarly, but it seems this is the only way to do a string with more than one pluralisable value within it.

andreievg: Can we just concatenate when used ?

roxy-dao:
I think there was a reason why it was editable... since some people might not know their DOB and only have an age (guessimate(?)) There wasn't much information in https://github.com/msupply-foundation/open-msupply/issues/385#issue-1317595755...

CarlosNZ:
> I think there was a reason why it was editable...

It's still going to be editable for everyone except under one-year olds. And their DOB should be known :)

roxy-dao:
> > I think there was a reason why it was editable...
> 
> It's still going to be editable for everyone except under one-year olds. And their DOB should be known :)

Oh sorry didn't see the condition XD

roxy-dao (approved):
Nice~ Thanks Carl

I think it looks fine like this (:

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -101,21 +100,24 @@ const UIComponent = (props: ControlProps) => {
...
+              <NumericTextInput
+                value={age}
+                sx={{ width: 65 }}
+                onChange={onChangeAge}
+                disabled={!props.enabled}
+              />
+            </Box>
+          ) : (
+            <Typography fontSize="85%" whiteSpace="nowrap">
+              {t('label.age-months-days', { months, days })}
```
andreievg, line 118: I think it's worth hiding months or days if they are 0. Which also means using translations string individually rather then age-months-days

andreievg (changes requested):
I think good to have a comment and also change to use individual translation

client/packages/common/src/intl/locales/en/common.json
CarlosNZ: In practice that would probably be fine in this case, but we can't be sure that a comma-separated concatenation is correct for all languages. 

For example, in Chinese:

> 3个月15天

CarlosNZ: I guess we could stack them vertically and lose the comma?

andreievg: Or even not have a comma, just space ?

andreievg:
@CarlosNZ, with the comment i forgot to add the comment that I did:

```typescript
  ageInMonthsAndDays: (date: Date | string) => {
    const months = differenceInMonths(Date.now(), date);
    // from: 01-03-2024 to 02-08-2024 = 5 months, 1 day
    // from: 03-03-2024 to 02-08-2024 = 4 months, 30 days
    const days = differenceInDays(Date.now(), addMonths(date, months));
    return { months, days };
  },
```

Apart from that i still think we should just concatenate, separated by space, I can see there could be a difference in gramma and punctuation for some languages, but it's much simpler to concatenate I think (and also allows you to not show the months or days when they are 0

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
CarlosNZ: > I think it's worth hiding months or days if they are 0.

Yeah, makes sense to hide months if it's 0, though I'm less sure about 0 days. If you read "3 months" you kind of assume it's been rounded from within 2.5 - 3.5 months -- you need the 0 days to make it clear that we're being more precise.

client/packages/common/src/intl/locales/en/common.json
mark-prins: or put the separator in the translated string
```
"label.age-months-and": "{{months}} months, ",
"label.age-days_one": "{{days}} day",
```

etc.

then `${t('label.age-months-and', { count: months })}${t('label.age-days, {count: days})}`
or something like that. it's a little easier to read than the nested version

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -101,21 +100,25 @@ const UIComponent = (props: ControlProps) => {
...
+              <NumericTextInput
+                value={age}
+                sx={{ width: 65 }}
+                onChange={onChangeAge}
+                disabled={!props.enabled}
+              />
+            </Box>
+          ) : (
+            <Typography fontSize="85%" whiteSpace="nowrap">
+              {months > 0 && t('label.age-months', { count: months })}{' '}
```
mark-prins, line 118: I would recommend putting the separator (comma and space) into the 'month' translated string then you don't need the `{' '}` and it is truly conditional and more flexible.

CarlosNZ: Yup, which means we'll definitely need to keep the "days" part, even if it's 0 (re: Andrei's earlier comment)

CarlosNZ: c482d1438

andreievg (approved):

PR #4495, opened 2024-08-04
4161 relocate gender input

### LINKED ISSUE #4161: Hardcode filtering out of patients from names query

### LINKED ISSUE #4162: GAPS: soft delete immunization programs

roxy-dao:
Fixes #4161

# 👩🏻‍💻 What does this PR do?
Move GenderInput/Type from name type to patient type and also formatted gender in frontend to remove the `_` (: 

# 🧪 Testing

- [ ] Changing patient gender should still work
- [ ] Filtering by patient gender should work too

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -1032,7 +1032,7 @@ export type ContactTraceFilterInput = {
   datetime?: InputMaybe<DatetimeFilterInput>;
   documentName?: InputMaybe<StringFilterInput>;
   firstName?: InputMaybe<StringFilterInput>;
-  gender?: InputMaybe<EqualFilterGenderInput>;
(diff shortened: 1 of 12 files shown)


mark-prins (approved):
👍 changes look good to me thanks!

andreievg (approved):
Looks great, thanks !

andreievg:
@roxy-dao I think one more things remains for the main issues, and that is to remove the `patient` from nametype input (all the way through, to from repo to graphql layer)

PR #4498, opened 2024-08-04
4497 Report args modal backdrop blackening

### LINKED ISSUE #4497: Report args modal backdrop blackening
## What went wrong? 😲
Screenshot 2024-08-05 at 9 28 16 AM

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #4497

# 👩🏻‍💻 What does this PR do?
Move report args out of react fragment
Screenshot 2024-08-05 at 9 30 58 AM

# 🧪 Testing

- [ ] Set up new reports from reports repo
- [ ] Click on a report to generate from the Report page
- [ ] The modal shouldn't be bugging anymore

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -128,6 +128,7 @@ const DetailViewInner = ({
         isPrinting={isPrinting}
       />
       <ReportArgumentsModal
+        key={report.id}
         report={reportWithArgs}
         onReset={() => setReportWithArgs(undefined)}
         onArgumentsSelected={generateReport}
--- a/client/packages/reports/src/components/ReportWidget.tsx
+++ b/client/packages/reports/src/components/ReportWidget.tsx
@@ -1,20 +1,15 @@
-import React, { PropsWithChildren, useCallback } from 'react';
+import React, { PropsWithChildren } from 'react';
 import { ChevronDownIcon, SvgIconProps } from '@common/icons';
-import {
-  ReportRowFragment,
-  ReportArgumentsModal,
-} from '@openmsupply-client/system';
+import { ReportRowFragment } from '@openmsupply-client/system';
 import {
   BasicSpinner,
   Link,
   RouteBuilder,
-  useNavigate,
   Card,
   Grid,
   Typography,
 } from '@openmsupply-client/common';
 import { AppRoute } from '@openmsupply-client/config';
-import { JsonData } from '@openmsupply-client/programs';
 
 interface ReportWidgetProps {
   height?: number | string;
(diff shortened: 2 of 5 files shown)


client/packages/reports/src/components/ReportWidget.tsx
```diff
@@ -142,13 +142,13 @@ export const ReportWidget: React.FC<PropsWithChildren<ReportWidgetProps>> = ({
...
                       </Grid>
                     </Link>
-                    <ReportArgumentsModal
-                      report={reportWithArgs}
-                      onReset={() => setReportWithArgs(undefined)}
-                      onArgumentsSelected={reportArgs}
-                    />
                   </React.Fragment>
                 ))}
+                <ReportArgumentsModal
```
mark-prins, line 147: doesn't fix it for me 😢 
Looking at the source - there are three modals showing:


image

mark-prins (commented):
Still getting the issue, am looking into it

mark-prins:
Have fixed in this [suggestion branch](https://github.com/msupply-foundation/open-msupply/compare/4497-Report-args-modal-backdrop-blackening...4497-Report-args-modal-backdrop-blackening(suggestion)).

The issue is that the modal is rendered for every `ReportWidget` component and all of these are using the same state object to determine visibility : that is, the state in the `ListView` component. when this is set with an object that has args, all of the report modals are shown. 

My suggestion is to move the modal to the ListView, so that there is only a single instance of it.

mark-prins (changes requested):
Have added a suggested change

client/packages/reports/src/DetailView/DetailView.tsx
```diff
@@ -128,6 +128,7 @@ const DetailViewInner = ({
         isPrinting={isPrinting}
       />
       <ReportArgumentsModal
+        key={report.id}
```
mark-prins, line 131: don't see why the `key` is necessary, or what value that adds 🤷

mark-prins (approved):
🙏  thanks - that does the trick

client/packages/reports/src/DetailView/DetailView.tsx
CarlosNZ: Sorry, @mark-prins, this was from my PR (#4501 ) which I've merged into here already (I intended to wait, had an absent-minded moment).

Regarding the key:
> add a "key" to the ReportArgsModal so that it'll force a remount (and therefore data reset) when the report type (id) changes

This was because I removed a "clearData" function elsewhere and this was a way to ensure the form gets reset when the report type changes.

mark-prins: oh, nice! slick way to get the component to remount

PR #4501, opened 2024-08-05
#4496 Persist Report args

### LINKED ISSUE #4496: Reports args vanish
## What went wrong? 😲

From https://github.com/msupply-foundation/open-msupply/pull/4483... Reports args aren't being retained and if you click okay twice then you get error below

Screenshot 2024-08-05 at 8 43 02 AM
Screenshot 2024-08-05 at 8 40 07 AM

## How to Reproduce 🔨

1. Upsert reports from reports report with StockAndItems or Expiring
2. Generate report
3. Go to `filter`
4. Click `ok`
5. Go back to `filter`
6. See args are gone

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #4496 

Have based on @mark-prins 's suggested branch for #4498 as that looks like it was dealing in the same area. So probably want to get that sorted before this one.

# 👩🏻‍💻 What does this PR do?

- makes sure any existing URL report args are included in modal data on load
- *don't* clear modal data on close
- add a "key" to the ReportArgsModal so that it'll force a remount (and therefore data reset) when the report type (id) changes

# 🧪 Testing

Follow the steps in the issue (#4496) and check that the problem has been fixed.

Also, if possible, try different reports and make sure args data is cleared when changing reports.

### DIFF
--- a/client/packages/system/src/Report/components/ReportArgumentsModal.tsx
+++ b/client/packages/system/src/Report/components/ReportArgumentsModal.tsx
@@ -2,7 +2,7 @@ import React, { FC, useState } from 'react';
 
 import { JsonData, JsonForm } from '@openmsupply-client/programs';
 import { ReportRowFragment } from '../api';
-import { useDialog } from '@common/hooks';
+import { useDialog, useUrlQuery } from '@common/hooks';
 import { DialogButton } from '@common/components';
 import { useAuthContext } from '@openmsupply-client/common';
 
@@ -19,6 +19,7 @@ export const ReportArgumentsModal: FC<ReportArgumentsModalProps> = ({
   onArgumentsSelected,
 }) => {
   const { store } = useAuthContext();
+  const { urlQuery } = useUrlQuery();
 
   const {
     monthlyConsumptionLookBackPeriod,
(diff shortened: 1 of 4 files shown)


client/packages/system/src/Report/components/ReportArgumentsModal.tsx
```diff
@@ -32,12 +33,12 @@ export const ReportArgumentsModal: FC<ReportArgumentsModalProps> = ({
     monthsOverstock,
     monthsUnderstock,
     monthsItemsExpire,
+    ...JSON.parse((urlQuery?.['reportArgs'] ?? '{}') as string),
   });
   const [error, setError] = useState<string | false>(false);
 
   // clean up when modal is closed
   const cleanUp = () => {
-    setData({});
```
CarlosNZ, line 40: Hopefully this doesn't cause any other problems. I've added the `key` prop to the Modal which should force a remount when the report type changes, which *should* work in lieu of clearing the data here.

client/packages/system/src/Report/components/ReportArgumentsModal.tsx
```diff
@@ -32,12 +33,12 @@ export const ReportArgumentsModal: FC<ReportArgumentsModalProps> = ({
     monthsOverstock,
     monthsUnderstock,
     monthsItemsExpire,
+    ...JSON.parse((urlQuery?.['reportArgs'] ?? '{}') as string),
```
roxy-dao, line 36: Nice~

roxy-dao (approved):
LGTM~ Thanks Carl~

PR #4504, opened 2024-08-05
4161 move patient out of name type node

### LINKED ISSUE #4161: Hardcode filtering out of patients from names query

roxy-dao:
Fixes #4161

# 👩🏻‍💻 What does this PR do?
Last bit of 4161, move patient out of name type

# 🧪 Testing

- [ ] Make sure customers and suppliers are still loading correctly

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/rnr_form/tests/finalise.rs
+++ b/server/service/src/rnr_form/tests/finalise.rs
@@ -5,8 +5,8 @@ mod finalise {
     use repository::test_db::setup_all;
     use repository::{
         EqualFilter, RequisitionFilter, RequisitionLineFilter, RequisitionLineRepository,
-        RequisitionRepository, RequisitionRowRepository, RequisitionStatus,
-        RnRFormLineRowRepository, RnRFormRowRepository, RnRFormStatus,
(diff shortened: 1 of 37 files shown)


server/service/src/rnr_form/tests/finalise.rs
```diff
@@ -71,11 +71,6 @@ mod finalise {
             .context(mock_store_a().id, "".to_string())
             .unwrap();
 
-        let last_requisition_number = RequisitionRowRepository::new(&context.connection)
```
roxy-dao, line 74: This variable isn't being used anywhere in the test 🤷🏻

server/service/src/sync/test/test_data/rnr_form_line.rs
```diff
@@ -48,6 +48,7 @@ fn rnr_form_line_1() -> RnRFormLineRow {
         requested_quantity: 0.0,
         comment: None,
         confirmed: false,
+        previous_average_monthly_consumption: 0.0,
```
mark-prins, line 51: I think I've fixed this somewhere too 😆

mark-prins (approved):
🎉  thanks Roxy - looks great!

PR #4511, opened 2024-08-06
Set R&R line stock out duration to 0 if no transactions

### LINKED ISSUE #4509: R&R Stock out days should be set to 0 if there was no consumption
## What went wrong? 😲

if there is no consumption data we are setting stock out duration to be the length of the period... 
image

## Expected behaviour 🤔

Given the current use case (not having consumption data) it would be better to default to 0

## How to Reproduce 🔨

1. Have no consumption for period
2. Create R&R
3. See full stock out days

## Your environment 🌱

- Open mSupply Version: 2.2 develop

- Platform:

- Database type:

lache-melvin:
Fixes #4509 

# 👩🏻‍💻 What does this PR do?

If there is no transaction data for the period, and the closing balance is 0, we were setting stock out duration to be the length of the period... 

However for the current use case, R&R forms are used without consumption data tracked in OMS, so we should default this duration to 0 and let the user input the duration.

Feels like a future improvement, but wondered about checking whether there were _any_ transactions in the period, for any item, rather than by item... can imagine that if you _were_ tracking transactions, and you had some items with no transactions, then you _would_ want the stock out duration to be set 🤷‍♀️ 

# 🧪 Testing

- [ ] Create the first R&R form for a program, with an item that had a closing balance of 0.0 and no transactions in the period
- [ ] See stock out duration - it should be 0

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/rnr_form/generate_rnr_form_lines.rs
+++ b/server/service/src/rnr_form/generate_rnr_form_lines.rs
@@ -313,9 +313,14 @@ pub fn get_stock_out_duration(
         .historic_stock
         .into_iter()
         .filter(|point| point.quantity == 0.0)
-        .count();
+        .count() as i32;
 
-    Ok(days_out_of_stock as i32)
+    if days_out_of_stock == days_in_period as i32 {
+        // If there was no consumption data, we'll set stock out duration to 0 and let the user input this
+        Ok(0)
+    } else {
+        Ok(days_out_of_stock)
+    }
 }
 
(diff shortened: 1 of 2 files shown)


mark-prins (approved):
🚢  agree - it would be better to check or to have a setting which tells us that we are operating in some sot of 'stock tracking in oms' mode. this fix though does the business and we can adjust in the next release if needs be 🙏

PR #4512, opened 2024-08-06
Fix initial R&R AMC calculations for frontend

### LINKED ISSUE #4488: R&R: Frontend AMC calculation wrong for first forms
## What went wrong? 😲

Frontend AMC calculation is for 3 periods:

(previous monthly consumption * 2 + this period monthly consumption) / 3

HOWEVER frontend can't distiguish between `previousMonthlyConsumption` being 0 because there are no previous R&R forms, or because there has been 0 consumption for the last two periods..

Instead of a `previousMonthlyConsumption` value, frontend should get a list of previous monthly consumption values, which it can then average. This also allows a frontend plugin to consider more than 3 periods for AMC if it so chooses.

## Expected behaviour 🤔

If this is my first R&R form, and I consume 10 packs, my AMC should be 10, not 3.333...

## How to Reproduce 🔨

1. Go to R&R forms
2. Create your first form
3. Add some consumption for the period
4. Calculated AMC will be a third of that value...

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

lache-melvin:
Fixes #4488

# 👩🏻‍💻 What does this PR do?

Frontend AMC calculation was always expecting there to have been at least 2 previous forms, and so the average was wrong on initial forms... i.e. consumption of 10 resulted in AMC of 3.33

This change now stores previous monthly consumption values as a stringified list on the R&R form row, which the frontend parses and uses to calculate AMC.

Per current requirements, we are only storing 2 previous values, so AMC is across 3 periods (current + two previous). However the other calculations have been setup to support a changing number of lookback periods if that is needed in future.

# 🧪 Testing

- [ ] Create the first R&R form for a program
- [ ] Add some consumption, see AMC value change
- [ ] AMC should match consumption
- [ ] FInalise and add another form
- [ ] Add some consumption data
- [ ] AMC should be average of monthly consumption for this and last period

# 📃 Documentation

  1.
  2.


client/packages/programs/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -45,7 +45,7 @@ export const ContentArea = ({
           },
           '.second-column': {
             position: '-webkit-sticky',
-            left: 89,
+            left: 88,
```
lache-melvin, line 48: Side quest - the border between item name and strength was being hidden 😁

mark-prins (approved):
👍 Looks great! thanks

Issue #4224, opened 2024-06-20
Drop and re-create view during migrations


andreievg:
## The suggestion

Quite often when schema changes we have to re-create database views. This is quite a bit of work and can be prone to errors.

@CarlosNZ in Conforma implemented a drop and re-create view in migrations. We can do somethings similar (drop all views before migrations and then re-create them after migration finishes

## Example use case

An example of dropping and re-creating views

### Why should we invest time in this?

Would save time an d have all views in one place with their existing 'state' in code

### Are there any risks associated with this change?

Low risk, most views should be tested already

View tests would need to be done in repo code vs in the migrations, as they won't be available in normal migrations

### How much effort is required?

Up to a day

### Agreed Solution

Drop views at the start of each migration with one script and re-create them at the end of migrations with another script.

mark-prins:
Just curious - how does that handle migrating views? Do you have a way of specifying that a particular view isn't recreated? Is the migration to mutate the 'recreate' function? It's useful to know the full scope of a migration, which includes having the change to the view within it. Perhaps just a comment, to note that the view is changed in this version of the migration?

CarlosNZ:
> Just curious - how does that handle migrating views?

You only need to worry about the current version of the View -- the history of what it used to be is not particularly relevant or helpful (and if it is, just leave a comment)

Do you have a way of specifying that a particular view isn't recreated? 

```sql
DROP VIEW IF EXISTS xyz_view;
-- Deprecated, no longer required
```

> Drop views at the start of each migration with one script and re-create them at the end of migrations with another script.

You can actually go a step further and just have the one script at the end. As the migrations progress, if you're using CASCADE as you drop columns and tables, any connected views automatically get dropped (which can be a problem if you've forgotten about a view linked to a table being modified). So the finishing script just needs to drop all views and re-create the "current" version of them, with explanatory comments.

Might look a little something like this:

```sql
-- View to display combination of various data sources
DROP VIEW IF EXISTS abc;
CREATE VIEW AS abc (
  SELECT a, b, c from some_table...
);

-- This view no longer required
DROP VIEW IF EXISTS xyz;

-- View to do something else
DROP VIEW IF EXISTS my_view;
CREATE VIEW AS my_view (
  SELECT....
)

-- etc...
```

It's quite reassuring to have all the views (and functions if you have them) collected together in one place and know they'll they'll definitely be correct after every migration. And you only have to look in once place to see what they're doing instead of following a complex chain of migration logic.

[...]

PR #4517, opened 2024-08-07
#4224 Drop and recreate Views

### LINKED ISSUE #4424: Sync R&R forms to OMS central server

R&R forms need to be synced, in order to be backed up.

These should be synced to the OMS central server.

### Moneyworks Jobcode 🧰

[DFAT:TDEV]

CarlosNZ:
Fixes #4424

# 👩🏻‍💻 What does this PR do?

Adds functionality to drop all views before migration runs, then re-create them afterwards.

This means that, from now on, any changes to views can be done directly on the view rather than having to add additional migration to change them.

I've handled the error, in that the server will panic and stop if there is an error in the SQL mutations. However, I'd like to be able to bring through the actual error message from the database and print that out -- is that possible @andreievg ?

What did you want to do for test functions, @andreievg ?

# 🧪 Testing

- [ ] Check all views are correct after starting server
- [ ] Manually delete all the views. Re-start the server and they should all be re-instated.
- [ ] Check for postgres as well as SQLite
- [ ] Set the database version to something below 2.2.0 and the drop/rebuild shoudn't run (won't print anything to console)


andreievg:
>  However, I'd like to be able to bring through the actual error message from the database and print that out -- is that possible 

Yep, just have to store the error in the enum variant, as per below suggestion

> What did you want to do for test functions, @andreievg ?

I was hoping that test functions already exist on all of the view, i pretty confident they exist on most (and the rest should be used in reports, which we would add tests for anyways).

I actually couldn't see the migration run when migrating from scratch (i.e. when wiping database and doing `cargo run`). Code suggest they should run during startup if we are already on certain database version, but not as we migrate to that version or past that version from earlier version. I had a play around and this should work

https://github.com/msupply-foundation/open-msupply/compare/4224-drop-and-recreate-views...4224-drop-and-recreate-views(Drop-and-re-create-view-when-database-is-migrating)

There is an issue with postgres views and postgres migration (actually first of all there is a fix in develop for PgType on NameTypeRow, that might need to be pulled in). I'll make a separate issue about it shortly. Postgres tests actually fail

andreievg:
By the way for view that are different between postgres and sqlite, it's probably best to show the difference -> conditionally change just part of the view that is different (this can be done in another issue)

andreievg:
@CarlosNZ i did another check and I think the invoice_stats needs to be swapped (sqlite and postgres):
https://github.com/msupply-foundation/open-msupply/compare/4224-drop-and-recreate-views...4224-drop-and-recreate-views(Swap-postgres-and-sqlite-invoice-line-view)

andreievg:
@roxy-dao i merged my suggested change, then added the latest changes to the requisitions_in_period view (from one of the RC), re-run tests for postgres and sqlite and also added something about adding to internal docs in this issue: https://github.com/msupply-foundation/open-msupply/issues/4579

It should be ready to merge i think

roxy-dao (approved):
LGTM ~ Thanks Andrei and Carl (:

roxy-dao:
Merged into wrong branch, but have merged develop into 2.2 since no new PRs have gone into develop >.>

Issue #3231, opened 2024-03-07
Handle RC migrations


mark-prins:
## Is your feature request related to a problem? Please describe 👀
Currently if you install an RC version and initialise - if you then install a later RC which includes database migrations, you'll have to reinitialise the database.

This can take a considerable time (for example, if using cold chain) and so is a frustration for testers.
The release schedule has us creating the RC builds at the end of the two development sprints, however for the upcoming release, the first feature is already done. We have the ability to create an RC build early which allows the QA team to spread the testing workload better. I don't want the difficulty of migration being a blocker in our ability to test effectively.

## Describe the solution you'd like 🎁
Could the RC version be stored, and migrations grouped in batches according to RC version, within the major.minor version migration?

andreievg:
The main use case for disallowing RC upgrades is due to migration change, here are two use case, let's say we working on version  1.7.0

 1 - RC version is made/used, but we decided that the field name is not correct, if we disallow RC upgrade this is quite easy (just change the name in the original RC version), if we do allow RC upgraded then we have to migrate the name changes in the RC build
2 - RC version i made/used, but our existing data migration was incorrect, similar to the above we have to re migrate the data (potentially, with original data already 'lost')

I was wondering if we can ask testers to re-initialise, or keep base sqlite datafile from previous version ? Maybe even build tools for this ? Or reference data files (considering things like cold chain data will get outdated, maybe it's actually better to have a reference data for it, with date updates)

mark-prins:
> I was wondering if we can ask testers to re-initialise

the issue says "This can take a considerable time (for example, if using cold chain) and so is a frustration for testers."
it's not a deal breaker, no. but I find re initialising painful and best avoided.

> or keep base sqlite datafile from previous version ?

tricky on android, and tricky if using postgres 😉 
It would mean any changes made to the current (migrated) database would be lost / if they were sync'd to mSupply then things get tangled.

Perhaps we need something in the installer which warns / alerts the user if updating an RC version? Or we disallow this and force an uninstall and reinstall, which then encourages a re-initialisation? if you are using postgres it doesn't help - but it might solve the issue for testers? (while creating another issue, of having to re-init every time)

andreievg:
Hmm, maybe we can set RC migration as 'breaking' ? And only those are considered un-upgradable ? 
The 'breaking' RC version will be manually decided and set by devs ? (when say we update schema between RC versions) ?

[...]

PR #4519, opened 2024-08-07
Ability to migrate with fragments

andreievg:
Fixes https://github.com/msupply-foundation/open-msupply/issues/3231

# 👩🏻‍💻 What does this PR do?

Although https://github.com/msupply-foundation/open-msupply/pull/4361 seems to deal with the issue of migrating RC versions and allows for dev help with migrations, I struggled to understand fully how it should be applied/used.

There were a lot of references to migration steps, which was perviously considered as a way to deal with multiple parallel migrations. It also deals with the two use cases that we have:

* Being able to migrate during RC
* Run migrations that were added in parallel (during feature, by differed devs)

Removed RC restrictions on migrations and added 'migration fragments' which are explained in the doc and example in this PR. But essentially a way to record and run migration 'steps' on top of migration by version. 

There was a question that migrations should be decoupled from version. I would agree with that if we have a mechanism of testing migrations, especially data migrations, by migration fragments. Although I think migration by versions does give a structure and reduces the chance of some old migrations being updated (expecting it to be auto applied ?)

Please see readme.

Should we allow fragment migrations of previous versions ? (like if we did patch in main, should it also apply id dev)

# 🧪 Testing

See example in readme

# 📃 Documentation

  1.
  2.


mark-prins:
Took me a long time to understand the intention. Can I check that I have understood correctly?
* The current migration process remains: the `migration` fn is run if the migration version is greater than the database version. After migration the version in the database is updated, so the migration won't run again.
* There is a new fn added to the trait, which is run once only if the migration version is greater or equal to the database version (this is tracked by a record in a new table)

In which case: why not use the new table for _all_ migrations? Too many records in the table?

Side note: the 'migration fragments' are also one-time; so naming the 'migration' as 'one time' is not clear. I'm thinking in terms of 'major' and 'minor' migration aka semver; where you run the 'minor' when the major version is matching, and the 'major' migrations are a version bump. (though yes, it really is an RC migration, since major/minor/patch version changes all result in migrations. perhaps 'minor' is confusing things 🤔 )

Chris-Petty:
weird

server/repository/src/migrations/README.md
```diff
@@ -17,22 +17,58 @@ During startup server will run [these steps sequentially](mod.rs):
...
+
+1 - We start with adding a [new migration and a new table](https://github.com/msupply-foundation/open-msupply/compare/c95609f818f171bf106e8124c7ee87815d5f996e...4ede6643460459cb9a1bff4bbeea924bdb6c2e54).
+
+2 - [A field is added to that table](https://github.com/msupply-foundation/open-msupply/compare/4ede6643460459cb9a1bff4bbeea924bdb6c2e54...c53a65e19f1894baae7c2568aa8800fb69941a8d) (make sure it has reasonable defaults if it the column is not optional !). You can see that when devs are working in parallel on separate migrations, it should be possible to add consolidate migrations with these method, quite easily (since each fragment should be applied independently).
+
+3 - Modifying a field would require quite a lot of SQLite code, but we [can drop table and re-create it](https://github.com/msupply-foundation/open-msupply/compare/c53a65e19f1894baae7c2568aa8800fb69941a8d...71fbfb75a003f78cb29b678c6d1ade8810b98748).
+
+*Gotcha*
+
+Above quite works for simple cases, but it introduces too many use cases to work through, you should be at least aware of the surface area:
```
lache-melvin, line 61: What is this sentence trying to say? 🙈 

```suggestion
The above works for simple cases, but it often introduces too many use cases to work through. You should be at least aware of the surface area:
```
?

server/repository/src/migrations/README.md
```diff
@@ -17,22 +17,58 @@ During startup server will run [these steps sequentially](mod.rs):
...
+
+1 - We start with adding a [new migration and a new table](https://github.com/msupply-foundation/open-msupply/compare/c95609f818f171bf106e8124c7ee87815d5f996e...4ede6643460459cb9a1bff4bbeea924bdb6c2e54).
+
+2 - [A field is added to that table](https://github.com/msupply-foundation/open-msupply/compare/4ede6643460459cb9a1bff4bbeea924bdb6c2e54...c53a65e19f1894baae7c2568aa8800fb69941a8d) (make sure it has reasonable defaults if it the column is not optional !). You can see that when devs are working in parallel on separate migrations, it should be possible to add consolidate migrations with these method, quite easily (since each fragment should be applied independently).
+
+3 - Modifying a field would require quite a lot of SQLite code, but we [can drop table and re-create it](https://github.com/msupply-foundation/open-msupply/compare/c53a65e19f1894baae7c2568aa8800fb69941a8d...71fbfb75a003f78cb29b678c6d1ade8810b98748).
```
lache-melvin, line 57: This compare is confusing... there's two copies of the `delete_after_rc_add_new_table` file, and the `delete_after_rc_drop_new_table` fragment is called twice??

However, want to chat about this recommendation - I think renaming the existing migration file, then creating a new one with the same name but a different identifier, feels pretty error prone 👀 Could we not just create another fragment called `update_data_to_double`, which is run after `add_new_table`, and in it we drop the table and recreate with the updated columns?

server/repository/src/migrations/README.md
```diff
@@ -17,22 +17,58 @@ During startup server will run [these steps sequentially](mod.rs):
...
+
+Above quite works for simple cases, but it introduces too many use cases to work through, you should be at least aware of the surface area:
+
+If there is a field name change or field deletion or type change in a new branch, to which you switched and migrated. But then you switch back to develop for example, now we are in undefined state since rust logic will reference old fields and would expect old data types etc.
+
+If we added a reference to new table in another table, in parallel while between 2-3 were being worked on then migration will fail.
+
+Looking at 3 we may think it could be simplified by changing the identifier and [adding one drop clause](https://github.com/msupply-foundation/open-msupply/compare/c53a65e19f1894baae7c2568aa8800fb69941a8d...70827e95d03d359e80d967ac8e7ef29fe9ee72b3). This could work for in you in the branch you are working in, but then if you switch back to base branch, the previous migration fragments will try to be executed and will cause migration error
+
+Please be very mindful and vigilant when working with migration fragments, especially when doing major schema changes. And be aware of logic error that won't be caught at compile time.
```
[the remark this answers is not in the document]
mark-prins: +1 for the 'examples in code' comment; that was my key takeaway after reading the readme - would be great to have a code example. I think that'd be a more concise way to understand things

lache-melvin:
Hi just updating here for transparency - discussed that we'll back out the example/recommendation of deleting some migrations after the RC phase. It may mean a few more mutations of the schema, but will be simpler to reason about. Once a migration fragment has been created, it should not be edited.

roxy-dao (approved):
Thanks! Just a carry over for the docs (:

andreievg:
I created carry over issue to deal with docs.

Also added a commit to make sure RC3 is migrate-able: https://github.com/msupply-foundation/open-msupply/pull/4519/commits/6615973558a48e48b0e4412a455becdd1057a64f, re-ran all of the tests

Which looks at this diff
`git diff tags/v2.2.0-rc1 v2.2.0 server/repository/src/migrations/*`

```diff
diff --git a/server/repository/src/migrations/v2_02_00/rnr_form.rs b/server/repository/src/migrations/v2_02_00/rnr_form.rs
index f1ae19ac8f..c9d331a34a 100644
--- a/server/repository/src/migrations/v2_02_00/rnr_form.rs
+++ b/server/repository/src/migrations/v2_02_00/rnr_form.rs
@@ -39,6 +39,24 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
         "#
     )?;
 
+    #[cfg(feature = "postgres")]
+    sql!(
+        connection,
+        r#"
+          CREATE TYPE rn_r_form_low_stock AS ENUM (
+            'OK',
+            'BELOW_HALF',
+            'BELOW_QUARTER'
+          );
+        "#
+    )?;
+
+    const RNR_LOW_STOCK_ENUM_TYPE: &str = if cfg!(feature = "postgres") {
+        "rn_r_form_low_stock"
+    } else {
+        "TEXT"
+    };
+
     sql!(
         connection,
         r#"
@@ -46,6 +64,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
                 id TEXT NOT NULL PRIMARY KEY,
                 rnr_form_id TEXT NOT NULL REFERENCES rnr_form(id),
                 item_id TEXT NOT NULL REFERENCES item(id),
+                requisition_line_id TEXT REFERENCES requisition_line(id),
                 average_monthly_consumption {DOUBLE} NOT NULL,
                 previous_monthly_consumption_values TEXT NOT NULL,
 
@@ -63,6 +82,7 @@ pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
                 expiry_date {DATE},
                 calculated_requested_quantity {DOUBLE} NOT NULL,
                 entered_requested_quantity {DOUBLE},
+                low_stock {RNR_LOW_STOCK_ENUM_TYPE} NOT NULL,
 
                 comment TEXT,
                 confirmed BOOLEAN NOT NULL DEFAULT FALSE

```

PR #4523, opened 2024-08-07
Add useDebounce option to text input

### LINKED ISSUE #4464: Report filter doesn't filter if you type too fast

andreievg:
Partially Fixes #4464

# 👩🏻‍💻 What does this PR do?

Should fix issue for existing report, there is underlying issue in the debouncing of inputs and performance, please see linked issue. 

I've updated some reports to use this prop: https://github.com/msupply-foundation/open-msupply-reports/pull/79

This shouldnt' close the issue, and issue can be changed to look at the debouncing overall, as we've been able to replicate it in the program module

# 🧪 Testing

Upsert one of the reports with ./upsert.sh script. Try to quickly item code or name and press OK. Filter should be respected in the report

# 📃 Documentation

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -45,12 +45,7 @@ type Options = z.infer<typeof Options>;
 // Validates the option and parses the pattern into RegEx
 const useOptions = (
   options?: Record<string, unknown>
-): {
-  errors?: string;
-  options?: Options;
-  pattern?: RegExp;
-  useDebounce?: boolean;
-} => {
+): { errors?: string; options?: Options; pattern?: RegExp } => {
   const [regexError, setRegexErrors] = useState<string | undefined>();
   const { errors: zErrors, options: schemaOptions } = useZodOptionsValidation(
     Options,
@@ -102,7 +97,6 @@ const UIComponent = (props: ControlProps) => {
     errors: zErrors,
     options: schemaOptions,
     pattern,
-    useDebounce = true,
   } = useOptions(props.uischema.options);
   const customErrors = usePatternValidation(path, pattern, data);
   const error = !!errors || !!zErrors || !!customErrors;
@@ -133,6 +127,7 @@ const UIComponent = (props: ControlProps) => {
   const rows = schemaOptions?.rows;
 
   const width = schemaOptions?.width ?? '100%';
+  const useDebounce = schemaOptions?.useDebounce ?? true;
 
   return (
     <DetailInputWithLabelRow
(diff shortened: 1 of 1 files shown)


client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -41,7 +45,12 @@ type Options = z.infer<typeof Options>;
 // Validates the option and parses the pattern into RegEx
 const useOptions = (
   options?: Record<string, unknown>
-): { errors?: string; options?: Options; pattern?: RegExp } => {
+): {
+  errors?: string;
+  options?: Options;
+  pattern?: RegExp;
+  useDebounce?: boolean;
```
CarlosNZ, line 52: I don't think `useDebounce` is a separate prop, is it? It should be on the `options` object.

client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -130,7 +142,10 @@ const UIComponent = (props: ControlProps) => {
         value: text ?? '',
         sx: { width },
         style: { flexBasis: '100%' },
-        onChange: e => onChange(e.target.value ?? ''),
+        onChange: e =>
+          useDebounce
+            ? onDebounceChange(e.target.value ?? '')
+            : onChange(e.target.value ?? ''),
```
CarlosNZ, line 148: As mentioned above, this doesn't actually work for me unless I use `schemaOptions?.useDebounce` here

CarlosNZ: You correctly added the `useDebounce` field to the Zod "Options" type above, but then didn't use it.

CarlosNZ (changes requested):
Hey, looks good. Bit confused about the `useDebounce` prop -- looks like it should be in the `options` object, and it only worked for me when I used the "options" one.

Am I missing something?

client/packages/programs/src/JsonForms/common/components/Text.tsx
andreievg: thanks @CarlosNZ, nice catch, i rushed this out and I thought I tested it, but not properly: 10596a49888d272bd9c06853ea4e16521cfee6f1

CarlosNZ (approved):

CarlosNZ:
Nice, thanks for the fix @andreievg -- working great now :)

PR #4528, opened 2024-08-07
3315 Allow changing other party for manual returns

### LINKED ISSUE #3315: Allow changing other party for manual returns

Currently, once you set the other party for the return, you can't edit it. However, I should be able to change the other party for manual returns.

If you generate a return from its related shipment, the current behaviour should remain. 

However, if there is no related shipment (i.e. the return was generated manually) I should be able to change who I am returning to/receiving the return from!

Therefore, this issue should probably be resolved after #3299

roxy-dao:
Fixes #3315

# 👩🏻‍💻 What does this PR do?
Allow users to change other party in Inbound and Outbound returns

# 🧪 Testing

Outbound Return
- [ ] Create an Outbound Return in Store A to Store B
- [ ] Change status to Picked
- [ ] Go to Store B and confirm that Inbound Return has been created
- [ ] Go back to Store A's Outbound Return
- [ ] Change supplier to Store C
- [ ] Go to Store C and check that Inbound Return has been generated correctly
- [ ] Go to Store B and check that Inbound Return is no longer there

Inbound Return
- [ ] Create an Inbound Return in Store A
- [ ] Try change customer name
- [ ] Should change correctly

# 📃 Documentation

  1.
  2.


client/packages/invoices/src/Returns/api/operations.graphql
```diff
@@ -88,9 +100,6 @@ fragment InboundReturn on InvoiceNode {
 
   otherPartyId
   otherPartyName
-  otherPartyStore {
```
roxy-dao, line 91: Noticed this wasn't being used anywhere

server/graphql/types/src/types/invoice_query.rs
```diff
@@ -95,19 +94,6 @@ impl InvoiceNode {
         self.invoice.other_party_id()
     }
 
-    pub async fn other_party_store(&self, ctx: &Context<'_>) -> Result<Option<StoreNode>> {
```
roxy-dao, line 98: We have other_party which has store so have removed. Also doesn't look like it's being used anywhere

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
```diff
@@ -0,0 +1,381 @@
...
+
+#[derive(SimpleObject)]
+#[graphql(name = "UpdateOutboundReturnNameError")]
+pub struct UpdateNameError {
+    pub error: UpdateReturnNameErrorInterface,
+}
+
+#[derive(Union)]
+#[graphql(name = "UpdateOutboundReturnNameResponse")]
+pub enum UpdateNameResponse {
```
lache-melvin, line 32: NIT, but I think `updateOtherParty` would have been clearer naming to me. I know it sets the name_link_id at the repository layer, but everywhere else we call it other party... I thought we were naming the returns like instead of referring to them by their invoice numbers for a sec LOL

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
```diff
@@ -0,0 +1,381 @@
...
+    UpdateOutboundReturnName as ServiceInput, UpdateOutboundReturnNameError as ServiceError,
+};
+
+use crate::mutations::outbound_shipment::error::InvoiceIsNotEditable;
+
+#[derive(InputObject)]
+#[graphql(name = "UpdateOutboundReturnNameInput")]
+pub struct UpdateNameInput {
+    pub id: String,
+    other_party_id: Option<String>,
```
lache-melvin, line 21: Nit - I think if the whole purpose of this mutation is to change the other_party_id, this shouldn't need to be an option? Would save a few checks...

lache-melvin (approved):
Looks great thanks Roxy! A couple of nits but nothing blocking... though I would love to know the reasoning behind not using the existing outbound return update??

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
```diff
@@ -0,0 +1,381 @@
...
+}
+
+#[derive(Union)]
+#[graphql(name = "UpdateOutboundReturnNameResponse")]
+pub enum UpdateNameResponse {
+    Error(UpdateNameError),
+    Response(InvoiceNode),
+}
+
+pub fn update_name(
```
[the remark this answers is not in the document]
roxy-dao: Outbound has transfers which can fudge up, so name needs to update separately

client/packages/invoices/src/Returns/InboundDetailView/Toolbar.tsx
```diff
@@ -48,10 +55,18 @@ export const Toolbar: FC = () => {
       >
         <Grid item display="flex" flex={1}>
           <Box display="flex" flex={1} flexDirection="column" gap={1}>
-            {otherPartyName && (
+            {otherParty && (
               <InputWithLabelRow
-                label={t('label.customer-name')}
-                Input={<BasicTextInput value={otherPartyName} disabled />}
+                label={t('label.supplier-name')}
```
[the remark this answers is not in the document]
roxy-dao: Opps, they all started blending in at some point XD good catch!

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
roxy-dao: Will still need to do all the checks tho? What checks are you referring to? 🤔  only check that would save is the optional match 👀

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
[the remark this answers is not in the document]
roxy-dao: Outbound has transfers which can mess up when changing name (same for tax), so name needs to update separately (same as tax)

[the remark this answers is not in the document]
lache-melvin: Ah that makes sense, thanks 😁

server/graphql/invoice/src/mutations/outbound_return/update_name.rs
lache-melvin: Oh yeah that's all I meant, the option matches through the validate/generate - felt weird to be running a method called update_name with possibly no name to update to?

PR #4530, opened 2024-08-07
4506 Reports page is blank if the reports aren't configured

### LINKED ISSUE #4506: Reports page is blank if the reports aren't configured
## What went wrong? 😲

The empty sections render while the page is loading, and then once it realises I don't have reports configured, it goes completely blank?

https://github.com/user-attachments/assets/a9c0a054-8b40-48fd-a17b-6cd21b2a7f93

## Expected behaviour 🤔

Would be nice to see the sections still, or the nothing here component, or something...

## How to Reproduce 🔨

1. Ensure you don't have any reports with the `Report` context in your DB
2. Go to reports page
3. Sections load for half a second, then blank

## Your environment 🌱

- Open mSupply Version: 2.2

- Platform:

- Database type:

roxy-dao:
Fixes #4506

# 👩🏻‍💻 What does this PR do?
Add Nothing here with a message to contact support about setting up reports if there are no reports.

# 🧪 Testing

- [ ] Don't have any reports set up (not even program reports)
- [ ] Go to report page
- [ ] See message

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/intl/locales/en/reports.json
+++ b/client/packages/common/src/intl/locales/en/reports.json
@@ -9,5 +9,5 @@
     "label.stocktake-frequency": "Stocktake frequency",
     "label.threshold-for-overstock": "Threshold for overstock",
     "label.threshold-for-understock": "Threshold for understock",
-    "message.contact-support": "Please contact support at support@msupply.foundation about configuring reports."
+    "message.contact-support": "Please contact support@msupply.foundation about configuring reports"
 }
--- a/client/packages/common/src/intl/locales/en/reports.json
+++ b/client/packages/common/src/intl/locales/en/reports.json
@@ -8,5 +8,6 @@
     "label.monthly-consumption-look-back-period": "Monthly consumption look back period",
     "label.stocktake-frequency": "Stocktake frequency",
     "label.threshold-for-overstock": "Threshold for overstock",
-    "label.threshold-for-understock": "Threshold for understock"
+    "label.threshold-for-understock": "Threshold for understock",
+    "message.contact-support": "Please contact support at support@msupply.foundation about configuring reports."
 }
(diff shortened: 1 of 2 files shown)


client/packages/common/src/intl/locales/en/reports.json
```diff
@@ -8,5 +8,6 @@
     "label.monthly-consumption-look-back-period": "Monthly consumption look back period",
     "label.stocktake-frequency": "Stocktake frequency",
     "label.threshold-for-overstock": "Threshold for overstock",
-    "label.threshold-for-understock": "Threshold for understock"
+    "label.threshold-for-understock": "Threshold for understock",
+    "message.contact-support": "Please contact support at support@msupply.foundation about configuring reports."
```
CarlosNZ, line 12: The "support at support" is a bit weird. Just looking through the other examples, seems we just have the email.

An existing example: "The mSupply Central Server needs to be upgraded to a newer version before you can connect - please contact support@msupply.foundation"

Also -- don't need the trailing "." -- seems we don't put a period for single sentence messages.

client/packages/reports/src/ListView/ListView.tsx
```diff
@@ -63,6 +64,15 @@ export const ListView = () => {
     },
     [navigate]
   );
+
+  if (
+    !store?.preferences?.omProgramModule ||
+    !stockAndItemReports?.length ||
+    !expiringReports?.length
```
CarlosNZ, line 71: This fails if *any* of the conditions are `true`, so I'm getting the "Nothing here" even when I have reports enabled.

I think they should probably be `&&`, right?

And maybe doesn't need `!store?.preferences?.omProgramModule`? Would you ever hide the reports even though they are available?

CarlosNZ (changes requested):
Cool, see comments.

client/packages/reports/src/ListView/ListView.tsx
roxy-dao: I guess not since the script for program pre-configs the reports too...

CarlosNZ (approved):

CarlosNZ:
Nice, thanks for the fixes @roxy-dao

PR #4549, opened 2024-08-09
Add test connection binary

### LINKED ISSUE #4544: Integrate config check tool
As an [R&D day project](https://trello.com/c/kcmXCydt/10-sync-check-tool) I created a standalone binary to test the sync config. Brian has now made it usable and integrated into the OMS codebase.

This now performs the following basic checks, to validate 
* **config** file is parseable
* **ping**, fetching details of sync URL from config & database
* **login** to sync site
* **sync v5** testing for a valid connection to the site
* **sync v6** tests connection to the oms central server

It would be good to bring this into the OMS codebase, and then include the binary in the build and add to the setup artefacts, in the same way the cli is.

mark-prins:
Fixes #4544

# 👩🏻‍💻 What does this PR do?
Tidies up the config testing tool and integrates with omSupply

You can run the test connection app using

```
cargo run --bin test_connection
```

or build and run the binary. pass in `--features postgres` to run the postgres version.
Parameters are available too:
* `--gui` will show the GUI version
* `--username [user]` to specify a username to login to oms with
* `--password [pass]` to specify the password for login test
* `--server_uuid [uuid]` to specify the hardware id

There's a command line output:

Screenshot 2024-08-09 at 2 41 28 PM

and a GUI output if you prefer:

Screenshot 2024-08-09 at 2 48 15 PM

I've only tested the postgres version fully, so there may well be extra bugs. I thought it might be useful to get in place and road test with a few servers / setups.

# 🧪 Testing

- [ ] with valid / invalid config
- [ ] in/correct database config
- [ ] when oms server is / isn't running
- [ ] when mSupply is / isn't running
- [ ] when central server is / isn't running

# 📃 Documentation

  1. This is a helpful support tool - for the public documentation a quick overview
  2. detail the params possible


build/windows/omsupply_server.suf
```diff
@@ -523,7 +669,7 @@ end
 
 -- advance to the next screen
 Screen.Jump(sNextScreenName);
-</Script>
+					</Script>
```
mark-prins, line 672: don't know why setup factory has reformatted the file. I think it's a new version of sf on the server, so possibly that's why 🤷

server/cli/src/test_connection.rs
```diff
@@ -0,0 +1,499 @@
...
+        let username = test_data
+            .args
+            .username
+            .clone()
+            .unwrap_or("admin".to_string());
+        let password = test_data
+            .args
+            .password
+            .clone()
+            .unwrap_or("pass".to_string());
```
lache-melvin, line 220: These defaults really only help us in development right? This is going to be a publicly available tool... would probably just make username and password required arguments?

server/cli/src/test_connection.rs
```diff
@@ -0,0 +1,499 @@
...
+
+    let args = Args::parse();
+    let gui = args.gui;
+
+    let (gui_tx, gui_rc) = mpsc::channel(10);
+    let test_task = tokio::spawn(perform_tests(gui_tx, args));
+
+    if gui {
+        let options = eframe::NativeOptions {
+            viewport: egui::ViewportBuilder::default().with_inner_size([320.0, 240.0]),
```
lache-melvin, line 47: Not a blocker, but if we're making it this small, is there a way to add a scrollbar? 😁

lache-melvin (approved):
Working great for me thanks Mark! Like you say, let's get it in and used 🙏 

Next step, but a re-run button in the GUI would be nice hehe

server/cli/src/test_connection.rs
```diff
@@ -0,0 +1,499 @@
...
+    sync::{
+        api::{SyncApiSettings, SyncApiV5},
+        api_v6::SyncApiV6,
+        settings::SyncSettings,
+    },
+};
+use tokio::sync::mpsc;
+
+#[derive(Parser, Debug)]
+struct Args {
```
andreievg, line 24: I was wondering about the gui, is this tester meant to be run by users or admins ? If Admins don't really need gui if users then don't need command line arguments (enter in gui)

server/cli/src/test_connection.rs
```diff
@@ -0,0 +1,499 @@
...
+
+        let server_folder = config
+            .server
+            .base_dir
+            .clone()
+            .ok_or(anyhow!("No server base dir in config"))?;
+
+        let hardware_id = test_data.args.server_uuid.clone().unwrap_or(
+            AppDataService::new(server_folder.as_str())
+                .get_hardware_id()
```
andreievg, line 266: the hardware_id would only be populated after the server runs, is the use case for this app before or after first server run ? Was also wondering why `--server_uuid` would be a parameter ? How would Admin and User get hardware_id ?

server/cli/src/test_connection.rs
```diff
@@ -0,0 +1,499 @@
...
+            Ok(msg) => {
+                info!("{} test passed: {}", gui_state.tests[i].0, msg);
+            }
+            Err(msg) => {
+                error!("{} test failed: {}", gui_state.tests[i].0, msg);
+            }
+        }
+
+        if gui {
+            gui_state.tests[i].1 = match result {
```
andreievg, line 110: not a fan of indexing an array by another array, i think we need to use type safety of rust as much as possible (this can really be one array I think, with test implementing a method returning string in tuple)

andreievg (commented):
Looks pretty good thanks!

There is a lot of copy paste from omSupply main code to gui here, i think this will make it hard to maintain in the future, I think we should make another issue to consolidate this logic to use omSupply functions (to make sure we are indeed comparing apples to apples and make it easier to maintain in the future), also the other comments I've added can be included (or i guess just a link to this PR saying deal with extra comments), happy to merge this though

server/cli/src/test_connection.rs
mark-prins: good point - I had initially thought of this as a gui tool for users.. thinking that admins would already have things set up correctly. Now that it's done, I like the command line version better, but I'm reluctant to dump the gui now that it exists. 

I've changed so that if you run with command line args, then you get the command line version (admin usage) and if you just start the binary then you get the gui, with an admin & username input.

mark-prins:
thanks for reviews!
Have removed the username and password defaults and the hardware uuid arg.

The app now detects command line usage and acts accordingly. The GUI has inputs for username and password, and has a button to run, and re-run, along with a larger window format and vertical scrollbars.

PR #4556, opened 2024-08-11
Reports UI

### LINKED ISSUE #4494: Stock status report UI review
## What went wrong? 😲

A couple of UI improvements to the expiring items report please

image

* The breadcrumb doesn't need to have an ID in it. I'd recommend having `Reports / Expiring items` where the reports links to the  reports list view (currently, clicking 'expiring items' takes me to 'Reports' which is misleading)
* Add some padding to the status chips - the text and dot are hard against the edges
* remove the border on the iframe
* The font size of the report is smaller than our usual which looks a little out of place. I think it's probably ok, but worth checking how it looks slightly larger

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #4494

# 👩🏻‍💻 What does this PR do?

Removed topLevel from urlParts if there are more than 1 element in urlParts... if we want Report/{Report Name} probably best to move reports out of top level? But it's easy enough to get to reports from Nav, so probably fine? 
Screenshot 2024-08-12 at 8 23 32 AM

Also added shortcut for reports (g+r) and removed border from iframe

# 🧪 Testing

- [ ] Have reports set up
- [ ] Press g+r to navigate to reports with shortcut
- [ ] Click on a report
- [ ] Breadcrumb should show as above (with only report name)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
+++ b/client/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
@@ -41,11 +41,6 @@ export const Breadcrumbs = ({
       ? t('breadcrumb.item', { id: part.value })
       : t(part.key);
 
-  // Removes the first element if it is a top level path to get correct breadcrumbs
-  if (topLevelPaths.includes(urlParts[0]?.value ?? '') && urlParts.length > 1) {
-    urlParts.shift();
-  }
-
   const crumbs = urlParts.map((part, index) => {
     const customCrumb = customBreadcrumbs[index];
 
--- a/client/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
+++ b/client/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
@@ -41,6 +41,11 @@ export const Breadcrumbs = ({
       ? t('breadcrumb.item', { id: part.value })
       : t(part.key);
 
+  // Removes the first element if it is a top level path to get correct breadcrumbs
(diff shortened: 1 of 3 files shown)


client/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,6 +41,11 @@ export const Breadcrumbs = ({
       ? t('breadcrumb.item', { id: part.value })
       : t(part.key);
 
+  // Removes the first element if it is a top level path to get correct breadcrumbs
```
mark-prins, line 44: I rather like having `Reports / ` in the breadcrumb.
If you remove this change and instead do this in the `reports/DetailView/DetailView.tsx`
```
    setCustomBreadcrumbs({ 1: report.name ?? '' });
```

i.e. change the `0` to a `1` then that give this result:

image

mark-prins: though sometimes the `useEffect` timing is whack and you get 

image

so I'd change to 

```
  const { setCustomBreadcrumbs } = useBreadcrumbs(['reports']);
```

in the same file

mark-prins (changes requested):
❤️  Looks great! thanks Roxy

Can I request a minor change? with two small tweaks to the breadcrumb call, we can have the Reports breadcrumb show. can you see if the suggestion works for you?

mark-prins (approved):
🚢  thanks for the change 🙏

PR #4570, opened 2024-08-13
4534 map requisition approved quantities

### LINKED ISSUE #4534: R&R forms additional columns
We ran a demo of the R&R and one of the issues notes is that there are columns on the sample spreadsheet which weren't implemented. These are

* Alarm
* Approved quantity

I think that the approved quantity is best handled in a different way - as the approval process is outside of the facility level R&R process, it happens with the internal order.

In which case, we just need to add an `alarm` column.
Looking at the example excel, this has the formula

`=IF((L50<N50/4),"!!", IF((L50<N50/2),"!",""))`

i.e. 
* final balance < (maximum quantity / 4) => `!!`
* final balance < (maximum quantity / 2) => `!`

though I wonder if we can do something more elegant, perhaps the warning icon in orange & red

image

If we do though, this would be for on-page display only, when printing that would need to be the exclamation mark in order to show correctly in excel

### Moneyworks Jobcode 🧰

`DFAT:TDEV`

lache-melvin:
Fixes #4534

# 👩🏻‍💻 What does this PR do?

Shows requisition approved quantities in the R&R form:
Screenshot 2024-08-13 at 3 26 55 PM

This PR:
- Adds a new transfer processor, once response requisition has been approved, map the approved quantities back to the request requisition
- Links request requisition lines to the R&R form lines, so we can expose these approved quantities

# 🧪 Testing

- [ ] You'll need the authorisation module enabled
  - [ ] Configure your user as an authoriser for your master list
  - [ ] Add use authorisation module preference in supplying store, and include requisitions in suppliers remote authorisation processes in requesting store
- [ ] Create and finalise R&R form
  - [ ] At this time, approved quantities should show as 0
- [ ] Sync (probably twice)
- [ ] Open the remote authorisation web app
- [ ] New requisition should be there, change some approved quantities and approve
- [ ] Back in omSupply, sync again
- [ ] View your R&R form - approved quantities should be there

# 📃 Documentation

  1. Maybe mention use of authorisation module?
  2.


client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -48,10 +48,7 @@ export const ContentArea = ({
 }: ContentAreaProps) => {
   const t = useTranslation('replenishment');
 
-  // TODO: move to backend, should join on item and sort by name!
-  const lines = data.sort((a, b) => (a.item.name > b.item.name ? 1 : -1));
```
lache-melvin, line 52: after implementing RnRFormLineRepository, could remove this 😁

server/service/src/processors/transfer/requisition/update_request_requisition_approved_quantities.rs
```diff
@@ -0,0 +1,98 @@
...
+                &request_requisition.requisition_row.id,
+                &line.requisition_line_row.item_link_id,
+                line.requisition_line_row.approved_quantity,
+            )?;
+        }
+
+        // Execute
+        let updated_request_requisition = RequisitionRow {
+            // 6.
+            approval_status: response_requisition.requisition_row.approval_status.clone(),
```
lache-melvin, line 85: I'm setting the approval status on the request requisition as well, so we have something to validate against for step 4... is there any issue with that? Is it a misuse of the field?

server/service/src/rnr_form/finalise.rs
```diff
@@ -36,15 +36,28 @@ pub fn finalise_rnr_form(
...
-            let rnr_form_repo = RnRFormRowRepository::new(connection);
+            requisition_repo.upsert_one(&requisition_row)?;
+            rnr_form_repo.upsert_one(&finalised_rnr_form)?;
 
-            rnr_form_repo.upsert_one(&finalised_form)?;
+            for (rnr_form_line_id, requisition_line) in rnr_form_line_and_requisition_lines {
+                requisition_line_repo.upsert_one(&requisition_line)?;
+
+                rnr_form_line_repo
+                    .update_requisition_line_id(&rnr_form_line_id, &requisition_line.id)?;
```
lache-melvin, line 59: Sorry, did a bit of rejigging in here so all the inserts/updates happened at the top level - this is the actual change, adding requisition_line_id to the rnr_form_lines

server/repository/src/db_diesel/rnr_form_line_row.rs
```diff
@@ -38,10 +41,13 @@ table! {
 }
 
 joinable!(rnr_form_line -> rnr_form (rnr_form_id));
-joinable!(rnr_form_line -> item (item_id));
+joinable!(rnr_form_line -> item_link (item_id));
```
lache-melvin, line 44: How bad behaviour is it to have the field as `item_id` on rnr_form_line instead of `item_link_id`? 🙈

roxy-dao (changes requested):
Thanks Lache! I think we should change item_id to item_link_id~

Have a few nits... but nothing blocking (:

server/service/src/processors/transfer/requisition/update_request_requisition_approved_quantities.rs
```diff
@@ -0,0 +1,98 @@
...
+            ..request_requisition.requisition_row.clone()
+        };
+
+        RequisitionRowRepository::new(connection).upsert_one(&updated_request_requisition)?;
+
+        let result = format!(
+            "Internal order ({}) updated to approved, per requisition ({})",
+            updated_request_requisition.id, response_requisition.requisition_row.id
+        );
```
lache-melvin, line 94: Should there be some kind of system activity log here?

server/repository/src/db_diesel/requisition_line/requisition_line_row.rs
```diff
@@ -69,6 +69,23 @@ impl<'a> RequisitionLineRowRepository<'a> {
...
+    ) -> Result<(), RepositoryError> {
+        diesel::update(requisition_line_dsl::requisition_line)
+            .filter(
+                requisition_line_dsl::requisition_id
+                    .eq(requisition_id)
+                    .and(requisition_line_dsl::item_link_id.eq(item_id)),
+            )
+            .set(requisition_line_dsl::approved_quantity.eq(approved_quantity))
+            .execute(self.connection.lock().connection())?;
+        Ok(())
```
lache-melvin, line 86: Should this insert a changelog? Initial thought was yes (it's a change, we want it to sync) but a lot of these field updater methods don't seem to call insert_changelog? Was that because the old changelog triggers used to capture these changes, and we missed these field updater methods in the update to manage changelogs in Rust? Or do they not add changelogs on purpose?

server/repository/src/db_diesel/rnr_form_line_row.rs
```diff
@@ -108,6 +114,18 @@ impl<'a> RnRFormLineRowRepository<'a> {
...
+    pub fn update_requisition_line_id(
+        &self,
+        rnr_form_line_id: &str,
+        linked_requisition_line_id: &str,
+    ) -> Result<(), RepositoryError> {
+        diesel::update(rnr_form_line)
+            .filter(id.eq(rnr_form_line_id))
+            .set(requisition_line_id.eq(linked_requisition_line_id))
+            .execute(self.connection.lock().connection())?;
+        Ok(())
```
lache-melvin, line 126: Same changelog question here?

lache-melvin:
Thanks Roxy - have added your changes, and per Telegram, haven't updated item_link yet... pending decision on if we migrate R&R forms to be requisitions or now!

Have added a couple of questions about changelogs though?

server/service/src/processors/transfer/requisition/update_request_requisition_approved_quantities.rs
roxy-dao: Could log that status has changed (:

server/repository/src/db_diesel/requisition_line/requisition_line_row.rs
roxy-dao: Some of the field updaters are coupled together with the normal upserts, so the changelogs are created for the upsert, and when sync runs it catches the field updates too... maybe we should add it in as a fail safe? 🤔 Should definitely be adding changelog if the fields are being updated independently tho...

lache-melvin: Woah is that an explicit design decision, or just how it's happened? Feels like a pretty big risk that the field updater could be used somewhere else and the change not captured? Like the repository layer can't know whether its being used independently or after a regular upsert 👀

roxy-dao: Was missed during changelog changes.... can create a new issue for it (:

lache-melvin:
Changelog/broken test data situation has been fixed, issue to look into the other field updaters here: #4595

Thanks Roxy!

roxy-dao (approved):
Thanks for the changes Lache~

The 2nd await doesn't fix the race condition... I think probably fine to merge and create an issue to fix overall race condition? 🤔

lache-melvin:
Race condition bug: https://github.com/msupply-foundation/open-msupply/issues/4608 🙏

PR #4573, opened 2024-08-13
4555 fix "Unexpected Null" error on Internal Orders page

### LINKED ISSUE #4555: Diesel deserialisation error on Internal Orders page
## What went wrong? 😲

I get this error when I open the Internal Order list view:
Screenshot 2024-08-09 at 4 28 01 PM

It's from the programSettings query - querying the requisitions_by_period view, which hasn't been updated to handle nullable orderType values (Internal Orders generated from R&R forms don't have an order type)

## Expected behaviour 🤔

No error

Not sure whether the better to update the view, or set R&R form requisitions to Normal ordering type?

But in the `requisition` table `order_type` is nullable, so the view should probably expect the same type?!

## How to Reproduce 🔨

1. Have programs configured
2. Create and finalise an R&R form
3. View the Internal Orders pages
4. See the error

## Your environment 🌱

- Open mSupply Version: 2.2 RC1

- Platform:

- Database type: Postgres

lache-melvin:
Fixes #4555

# 👩🏻‍💻 What does this PR do?

Filters requisitions_in_period view to only include requisitions where order_type is not null, in line with the Rust types we have defined

# 🧪 Testing

- [ ] Have R&R forms configured - create and finalise at least one
- [ ] Also have usual programs configured for program based requisitions (i.e. also with the name tags)
- [ ] Open Internal Order page
- [ ] No error

# 📃 Documentation

  1.
  2.


andreievg (approved):
I added migration to migration fragment, and ran both postgres and sqlite tests

PR #4577, opened 2024-08-13
Fix tests

roxy-dao:
Fixes #3423423231

# 👩🏻‍💻 What does this PR do?
Really should have actions tests running again... anyway, need to import mock data to test against valid stock lines since we now set stock_line_id to None if stock line doesn't exist

# 🧪 Testing

- [ ] Run postgres + sqlite tests

# 📃 Documentation

  1.
  2.


andreievg (approved):
Looks good, thanks @roxy-dao

Action to test would be great but we are running out of space on github actions runners (vms), really need to setup self hosted runners, then can do both postgres and sqlite action tests

PR #4583, opened 2024-08-13
Prevent R&R line rerenders

### LINKED ISSUE #4568: R&R Forms perfomance
## What went wrong? 😲

When you have a number of rows in an R&R form the performance degrades

## Expected behaviour 🤔
Editing a single line should be pretty snappy

## How to Reproduce 🔨

1. Create an R&R form using a master list with 50+ rows
2. Edit a row on a tablet, or
3. add a `console.info` to the RnRFormLine component

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #4568

# 👩🏻‍💻 What does this PR do?
The `RnRFormLine` component is rendered 3x for every row, on every update.
This is sluggish on android and increasingly so the more lines you have.

I've hobbled the dirty state, as this contributes a render cycle - now the blue dot shows, but there's no global state, so there's no info snack telling you to update. I figure that if you are editing, then the row isn't confirmed and so the finalise button will still prompt.

I had also moved the `useConfirmOnLeaving` hook into the footer as this adds a render cycle when it updates. This is now on every row, and while I'm worried that this adds overhead, it seems to run ok.

Have added the loading indicator to the row. 

With the changes, when you edit a row the row itself renders - previously all rows were re-rendered. 
When you confirm, there are 2 renders for every row, so there's an improvement, even if not ideal.

Here's my old lenovo coping with 50 rows:

https://github.com/user-attachments/assets/91fbc589-6744-457b-ab9e-71991cc3ebf6

# 🧪 Testing

- [ ] try editing a medium/large form and get the vibes
- [ ] does it prompt if you navigate away after editing?
- [ ] does it warn you when finalising, if you have un-committed rows?

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
+++ b/client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import {
   Box,
   GlobalStyles,
@@ -7,6 +7,7 @@ import {
   NothingHere,
   Table,
   useTranslation,
+  ViewportList,
 } from '@openmsupply-client/common';
 import { RnRFormLineFragment } from '../api/operations.generated';
 import { RnRFormLine } from './RnRFormLine';
@@ -14,6 +15,7 @@ import { RnRFormLine } from './RnRFormLine';
 interface ContentAreaProps {
   data: RnRFormLineFragment[];
   saveLine: (line: RnRFormLineFragment) => Promise<void>;
+  markDirty: (id: string) => void;
   periodLength: number;
   disabled: boolean;
 }
@@ -41,10 +43,12 @@ const HeaderCell = ({ label, tooltip }: HeaderCellProps) => {
 export const ContentArea = ({
   data,
   saveLine,
+  markDirty,
   periodLength,
   disabled,
 }: ContentAreaProps) => {
   const t = useTranslation('replenishment');
+  const ref = useRef<HTMLDivElement>(null);
 
   // TODO: move to backend, should join on item and sort by name!
   const lines = data.sort((a, b) => (a.item.name > b.item.name ? 1 : -1));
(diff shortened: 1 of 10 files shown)


andreievg:
Just had a quick look at the code, I think it's worth to see what happens if you try to navigate away when multiple lines are dirty, it's looking to me like it would show multiple confirmation modals

lache-melvin:
> does it warn you when finalising, if you have un-committed rows?

It warns that some lines are unconfirmed, but then allows you to continue. I think if we're backing out the dirty state changes, we should at least make that warning text a bit more dramatic - that if you have any unsaved changes, they will be LOST

Also, if i have dirty rows, and then finalise, I still get the confirm on leaving warnings - these should be cleared somehow I think! Or not used...

>  it's looking to me like it would show multiple confirmation modals

Yup, this is happening for me

Can we pause this PR until the tech chat tomorrow? Aware there's already been a bit of effort sunk into this, but need it to be usable for this release... I'm kind of curious to try the zustand slice idea?

lache-melvin:
It's definitely a performance improvement though, thanks for looking into it Mark!

server/omsupply-chc-baoknana.sqlite.zip
lache-melvin, line 1: maybe remove 😁

client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -47,36 +48,55 @@ export const ContentArea = ({
...
             borderRight: '1px solid blue',
+            left: 72,
+            minWidth: '300px',
+            position: '-webkit-sticky',
           },
         }}
       />
       <Table
         sx={{
+          height: '500px',
```
lache-melvin, line 98: turns out adding `borderCollapse: 'separate'` here solves the borders not staying sticky on side scroll!

client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
```diff
@@ -47,36 +48,55 @@ export const ContentArea = ({
...
             width: 80,
           },
           '.second-column': {
-            position: '-webkit-sticky',
-            left: 88,
-            minWidth: '300px',
             borderRight: '1px solid blue',
+            left: 72,
+            minWidth: '300px',
+            position: '-webkit-sticky',
```
lache-melvin, line 92: I think we might need the column widths...

https://github.com/user-attachments/assets/0e042b4c-c2a9-48de-b981-38f0173b6799

client/packages/requisitions/src/RnRForms/DetailView/ContentArea.tsx
mark-prins: ❤️  legend!!

lache-melvin (approved):
Looks great Mark! Working well for me, just the column widths that I'm concerned about.

And obviously the changes clearing when you scroll, but all good with that in a separate PR 🙏

Issue #4589, opened 2024-08-14
Clearer reason why report could not be generated


lache-melvin:
## What went wrong? 😲

I didn't have the right permissions... would have expected that to give me a permission denied error? But just got this, which wasn't very helpful to try to troubleshoot:
![Screenshot 2024-08-14 at 4 49 59 PM](https://github.com/user-attachments/assets/a2153ae5-416a-4bd9-ad08-22df99a09086)

## Expected behaviour 🤔

Looks like we could pull out the forbidden response? 
![Screenshot 2024-08-14 at 4 48 46 PM](https://github.com/user-attachments/assets/bc71f7b7-7a7b-4e6a-a8a1-77185d8a9c10)

Would love to see the typical not authorised popup, or at least translate the structured error... not sure if there are other structured/expected errors as well?

`Unable to generate report` should mean there's nothing the user can do, unexpected error

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Have wrong permissions (i.e. can't view stock)
2. Generate stock details report
3. See `Unable to generate report`

## Your environment 🌱

- Open mSupply Version: v2.2.0 RC3
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

roxy-dao:
Closed in #4594

AnushaUp:
# 🧪 Testing
Tested in `V2.2.02-RC1`


- [ ] Remove view item permissions in OG
- [ ] Load stock status report 
- [ ] Sync OMS 
- [ ] Try to view stock status report
- [ ] Should show Nothing here page with no permission message

AnushaUp:
@roxy-dao @lache-melvin These item view permission doesn't make any difference in viewing the report
![Screenshot 2024-09-13 at 11 37 37 am](https://github.com/user-attachments/assets/9069a1b4-b9bb-4b66-9315-f4a595f508dd)

lache-melvin:
Hey @AnushaUp, the `view items` permission actually doesn't do anything in Open mSupply currently! (It's just based on store access) - I think the stock status and item usage reports are only affected by the report permission. The other two (stock detail and expiring items) should be affected by the view stock permission. So I think this is expected behaviour 🙏

PR #4594, opened 2024-08-14
4587 basic reports error shows eternal loading indicator

### LINKED ISSUE #4587: Basic Reports error shows eternal loading indicator
## What went wrong? 😲

If something goes wrong generating the report, you get the error snack, and then the loading indicator for the rest of time:

image

## Expected behaviour 🤔

Nothing here component? Something went wrong? Some way to indicator that we tried and failed, but we aren't still trying 🤷‍♀️ 

## How to Reproduce 🔨

1. Bang something up with your report config (or in my case, don't have the right permissions on for the report query hehe)
2. Generate the report
3. Its a disaster, you get an error
4. Circles of doom

## Your environment 🌱

- Open mSupply Version: V2.2.0 RC3
- Legacy mSupply Central Server Version: 7.16

- Platform: Android, also verified in browser

- Database type: sqlite

roxy-dao:
Fixes #4587 & #4589

# 👩🏻‍💻 What does this PR do?
Cover print structure error and change the spinner to a nothing here with the error message.
Screenshot 2024-08-15 at 7 29 46 AM

# 🧪 Testing

- [ ] Remove view item permissions in OG
- [ ] Load stock status report 
- [ ] Sync OMS 
- [ ] Try to view stock status report
- [ ] Should show Nothing here page with no permission message

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -2,9 +2,11 @@ import React, { useCallback, useEffect, useState } from 'react';
 import {
   BasicSpinner,
   FileUtils,
+  NothingHere,
   PrintFormat,
   useBreadcrumbs,
   useParams,
+  useTranslation,
   useUrlQuery,
 } from '@openmsupply-client/common';
 import {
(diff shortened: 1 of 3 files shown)


client/packages/reports/src/DetailView/DetailView.tsx
```diff
@@ -29,21 +34,24 @@ export const DetailView = () => {
     (reportArgsJson && JSON.parse(reportArgsJson.toString())) || undefined;
 
   return !report?.id ? (
-    <BasicSpinner />
```
andreievg, line 32: I think this one is ok to be a spinner (basically won't ever happen, unless user tries to edit url ?)

andreievg (approved):
Looks great thanks, I tested by building front end and accessing via bundle served by back end.

I think it's completely fine as is, but I would have moved error state to useGenerateRporet (so it returns muateAsync and error)

client/packages/reports/src/DetailView/DetailView.tsx
roxy-dao: That's what I did 😉

PR #4606, opened 2024-08-16
Round numbers when displayed in R&R forms

### LINKED ISSUE #4602: R&R AMC calculation and rounding
Request in from client:

- If there is no data for the last two months, AMC must be the same as adjusted consumption.  if the adjs. consumption is 200 the AMC is 200 too.
- If we have data previous for 3 months recently, some time the result can have decimal points, ex  677.42 or 677.66
 if it is possible the Quantity can be rounded up automatically?   from  677,.42 automatically will show  677 tabs. 

The rounding will have to apply to AMC, Maximum & Requested values. 
Will need to happen in the report too.

If it doesn't adversely affect performance, add a tooltip for the more accurate value.
For the AMC calculation - stick with concurrent periods; i.e. don't look for a previous form while skipping intervening, uncompleted periods. Look back until you reach an uncompleted form, up to a maximum of the 3 months.

mark-prins:
Fixes #4602 

# 👩🏻‍💻 What does this PR do?
Just rounds the display values. 
The AMC part is already handled - the confusion may have come about because the period length was 3mo in one of the examples.

# 🧪 Testing

- [ ] Check that the display has round numbers

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RnRForms/DetailView/RnRFormLine.tsx
+++ b/client/packages/requisitions/src/RnRForms/DetailView/RnRFormLine.tsx
@@ -10,6 +10,7 @@ import {
   LowStockStatus,
   NumericTextInput,
   NumUtils,
+  Tooltip,
   useBufferState,
   useNotification,
   useTheme,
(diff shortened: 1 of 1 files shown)


lache-melvin (approved):
LGTM!

> The AMC part is already handled - the confusion may have come about because the period length was 3mo in one of the examples.

I am a little worried about this - for monthly periods, adjusted consumption is for the whole period, but AMC is based on 30 days for a month... so 31 day period looks like this:

Screenshot 2024-08-19 at 9 55 22 AM

This is what we have elsewhere for our AMC calculations... do we need to do something different here?

PR #4617, opened 2024-08-19
Adjust translation namespace calls

### LINKED ISSUE #4616: Refreshing page results in components going everywhere??
## What went wrong? 😲

The UI has lost all understanding of space and time...

Screenshot 2024-08-19 at 12 17 58 PM

I've tracked the cause down to the addition of the useTranslation hook in the ReportArgumentsModal? But why??

https://github.com/msupply-foundation/open-msupply/blob/aa482282bbf125029b4aa19edec8117ae005189d/client/packages/system/src/Report/components/ReportArgumentsModal.tsx#L23

## Expected behaviour 🤔

Rendering as per usual

## How to Reproduce 🔨

1. Go to Internal Orders/Requisition detail view, or stocktake detail (these seem to be the worst affected?)
2. Refresh the page

## Your environment 🌱

- Open mSupply Version: v2.2.0 RC4

- Platform:

- Database type:

mark-prins:
Fixes #4616

# 👩🏻‍💻 What does this PR do?
I tracked this as far as the call to `useTranslation('reports')` which then causes the reports namespace to be loaded - but due to the timing / placement in the DOM of the ReportArgumentsModal this is loaded after the app bar portal is rendered:

Screenshot 2024-08-19 at 5 18 00 PM

which means that you get a ref for the portal, and then, without the div being unloaded, the ref is `null` 🤷 

a simple fix is to load the reports namespace earlier, so I've lifted that and used it in the appbar components.

A similar problem with internal orders, where the `distribution` namespace was being used, for a component which is in the `replenishment` section; a new namespace is loaded and the portal refs disappear.

So, no conclusive understanding of what's gone wrong - this at least will fix the problem.

# 🧪 Testing

Check the following pages, navigate to them and refresh: should look ok
- [ ] R&R forms detail view
- [ ] Requisition detail view
- [ ] Internal order detail view
- [ ] Stocktake detail view

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
+++ b/client/packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
@@ -25,7 +25,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   onAddItem,
 }) => {
   const { OpenButton } = useDetailPanel();
-  const t = useTranslation();
+  const t = useTranslation('reports'); // note: using 'reports' due to issue #4616
   const { print, isPrinting } = usePrintReport();
   const { data } = useStocktake.document.get();
   const isDisabled = !data || isStocktakeDisabled(data);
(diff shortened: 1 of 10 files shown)


roxy-dao (approved):
Thanks Mark! I wonder what is really going on though 🤔

PR #4619, opened 2024-08-19
4559 Total/Available SoH number looks wrong after stocktake for existing lines whose packsize is greater than 1

### LINKED ISSUE #4559: Total/Available  SoH number looks wrong after stocktake for existing lines whose packsize is greater than 1
## What went wrong? 😲
If you do stocktake of existing lines (with pack size greater than 1), after that stocktake, the SoH number is messed up.

## Expected behaviour 🤔
SoH should be correct as per the stocktake taken

## How to Reproduce 🔨
1. Have a item stock whose pack-size is greater than 1. (You can repack existing ones to do so). In my case, I've 1 pack `apple` item, whose pack-size=2; thus SoH=2
2. Create new stocktake
3. Select that item
4. Existing batch of packsize-2 appears with `snapshot packs`=1
5. Change `counted packs` to 5. As per this, now new SoH should be 2*5=10
6. Finalise stocktake and View stocks
7. See 1st error: `apple` item's pack-size is 1 (despite that I changed existing lines whose pack-size was 2 and fixed)
8. See 2nd error: SoH shows 9: wrong value here

## Your environment 🌱

- Open mSupply Version: v2.2.00-RC2
- Legacy mSupply Central Server Version: v7-16-04

- Platform: oms-server, browser(chrome)

- Database type: sqlite

roxy-dao:
Fixes #4559

# 👩🏻‍💻 What does this PR do?
OG docs only talk about converting to 1 when receiving stock, so have removed this for Inventory Adjustments. 

# 🧪 Testing

- [ ] Add stock with pack size greater than 1
- [ ] Do stocktake with that stock and add more stock
- [ ] Finalise stocktake
- [ ] Go back to stock view and make sure numbers are correct

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/invoice_line/stock_in_line/insert/generate.rs
+++ b/server/service/src/invoice_line/stock_in_line/insert/generate.rs
@@ -30,12 +30,11 @@ pub fn generate(
 ) -> Result<GenerateResult, RepositoryError> {
     let store_preferences = get_store_preferences(connection, &existing_invoice_row.store_id)?;
 
-    let new_line = generate_line(input.clone(), item_row, existing_invoice_row.clone());
+    let mut new_line = generate_line(input.clone(), item_row, existing_invoice_row.clone());
 
-    let mut new_line = match store_preferences.pack_to_one {
-        true => convert_invoice_line_to_single_pack(new_line),
-        false => new_line,
-    };
+    if StockInType::InventoryAddition != input.r#type && store_preferences.pack_to_one {
+        new_line = convert_invoice_line_to_single_pack(new_line);
+    }
 
     let barcode_option = generate_barcode(&input, connection)?;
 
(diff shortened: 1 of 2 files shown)


server/service/src/invoice_line/stock_in_line/insert/generate.rs
```diff
@@ -59,12 +58,7 @@ pub fn generate(
         // If a new stock line has been created, update the stock_line_id on the invoice line
         new_line.stock_line_id = Some(batch.id.clone());
 
-        let new_batch = match store_preferences.pack_to_one {
```
roxy-dao, line 62: Already being repacked to 1 above, don't need to repack 1 to 1 again.

mark-prins (approved):
🚢  works for me!

one thing to add though: line 5 of `/insert/generate.rs` the `convert_stock_line_to_single_pack` is no longer required

PR #4631, opened 2024-08-20
4588 Extend pagination on stock lines for reports

### LINKED ISSUE #4588: Basic Reports only capture first 100 lines
## What went wrong? 😲

Only the first hundred lines are being returned in any of the reports! I can't see anywhere to enter pagination preferences...

## Expected behaviour 🤔

All lines are included in the report (or a way to select the range)

## How to Reproduce 🔨

1. Have more than 100 stock lines
2. Create stock status detail report
3. Only the first 100 are included in the report!

(Same issue for all reports) 

## Your environment 🌱

- Open mSupply Version: v2.2.0 RC3

- Platform: browser

- Database type:

andreievg:
Fixes #4588 

Requires for fix in report repo: https://github.com/msupply-foundation/open-msupply-reports/pull/87

# 👩🏻‍💻 What does this PR do?

Extend pagination for stock line

Ideally we would have no pagination for reports or have a way to grab individual pages, added to reports epic to think about in the future.

# 🧪 Testing

Run  stock usage, stock detail, expiring items and stock status reports for data file where there are more then 100 items and stock lines, should see all rows.

Can add items with this (for sqlite):

```sql
WITH RECURSIVE
  cnt(x) AS (VALUES(1) UNION ALL SELECT x+1 FROM cnt WHERE x<1000)
insert into item(id, code, name, type, legacy_record) select lower(hex(randomblob(16))), 'gen ' || lower(hex(randomblob(16))) , 'gen ' || lower(hex(randomblob(16))) , 'STOCK', ''  FROM cnt;

insert into item_link(id, item_id) select id, id from item where name like 'gen %';

insert into master_list_line(id, item_link_id, master_list_id)  select lower(hex(randomblob(16))), item.id, master_list.id from item join master_list where item.name like 'gen %';
```

# 📃 Documentation

### DIFF
--- a/server/service/src/stock_line/query.rs
+++ b/server/service/src/stock_line/query.rs
@@ -6,7 +6,7 @@ use repository::{
     EqualFilter, PaginationOption, StockLine, StockLineFilter, StockLineRepository, StockLineSort,
 };
 
-pub const MAX_LIMIT: u32 = 1000;
+pub const MAX_LIMIT: u32 = 5000;
 pub const MIN_LIMIT: u32 = 1;
 
 pub fn get_stock_line(ctx: &ServiceContext, id: String) -> Result<StockLine, SingleRecordError> {


roxy-dao (approved):
Thanks Andrei~

PR #4632, opened 2024-08-20
Add fallback to loading indicator for lazy loads

### LINKED ISSUE #4434: Whole page reloads when navigating to sub menu
## Expected behaviour 🤔

Make sure only the page area that needs the loading state, shows the loading state (rather then the whole page).

## Suggested Solution

Add another suspense, currently lazy loads: https://github.com/msupply-foundation/open-msupply/blob/fa22d29835933b131d6a00ed938651a9cc6d5c94/client/packages/host/src/routers/DistributionRouter.tsx#L6 are fallen back to top level suspense https://github.com/msupply-foundation/open-msupply/blob/fa22d29835933b131d6a00ed938651a9cc6d5c94/client/packages/host/src/Host.tsx#L78, causing the whole app to reload.

I think we need a more strategic and thorough think about how we handle supporting non locally served front end, to avoid issues like this (where we tried to optimise the bundle and the front end speed for non locally server front end but ended up making the whole app reload, including graphql queries like isCentralServer, me, etc..), maybe have a way to test this and having a requirements about what we want to support, accepted speed etc (I think it would have been simpler and arguably better user experience if there is just one delay, the first time you load the app, rather then delays throughout the app)

## How to Reproduce 🔨

Add throttling (in browser dev tools,  network), try to go to sub menu (inbound shipments etc..), see whole page needing to refresh

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #4434

# 👩🏻‍💻 What does this PR do?

Adds fall back to lazy loading routers

Adds loading indicating skeleton which matches more closely the second loading of list views which is likely to occur with slower internet.

You can test this by navigating to any of the changed routes with throttle enabled in dev tools

# 🧪 Testing

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/layout/skeletons/DetailFormSkeleton.tsx
+++ b/client/packages/common/src/ui/layout/skeletons/DetailFormSkeleton.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { Box, Skeleton, Stack } from '@mui/material';
 import { AppBarContentPortal, AppFooterPortal } from '../../components/portals';
 import { ButtonSkeleton } from './ButtonSkeleton';
+import { BasicSpinner } from '@common/components';
 
 const AppBarContent = () => (
   <Box display="flex" flexDirection="column" gap={1} width="100%">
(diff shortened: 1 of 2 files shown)


lache-melvin (approved):
Looks great to me! Thanks Ferg 🚀 

Awesome to see a few less queries in the dev tools as a result of these changes, saving us some unnecessary API calls and re-renders! 🎉

PR #4636, opened 2024-08-20
4592 Add store id to item visibility stock on hand filter

### LINKED ISSUE #4592: Store's item visibility is not really/strictly respected in case of multiple stores in an oms-site
## What went wrong? 😲
Store's item visibility is not really/strictly respected in case of multiple stores in an oms-site

## Expected behaviour 🤔
Strictly follow item-visibility as per the master-list set for the store

## How to Reproduce 🔨
1. Have multiple stores for your oms-site. Let's say StoreA with MasterListA only and StoreB with MasterListB only
2. Setup user permissions to login accordingly
image
image

3. Login to StoreA
4. See Catalogue>Items
5. See error: you'll see items from MasterListB as well
image

6. Login to StoreB
7. See Catalogue>Items
8. See error: you'll see items from MasterListA as well

## Your environment 🌱

- Open mSupply Version: v2.2.00-RC3
- Legacy mSupply Central Server Version: v7-16-04

- Platform: oms-server + browser(chrome)

- Database type: sqlite

andreievg:
Fixes #4592 

# 👩🏻‍💻 What does this PR do?

Looks like we needed extra filter by store id for item stock on hand visibility check

I added extra test that fails without this change

# 🧪 Testing

Have two stores on the same site, add extra item and new master list to one of the stores, check that it is not shown for second store. Now add some stock for that item in the store, item should remain invisible in second store

# 📃 Documentation

### DIFF
--- a/server/repository/src/db_diesel/item.rs
+++ b/server/repository/src/db_diesel/item.rs
@@ -251,7 +251,11 @@ fn create_filtered_query(store_id: String, filter: Option<ItemFilter>) -> BoxedI
         let item_ids_with_stock_on_hand = item_link_dsl::item_link
             .select(item_link_dsl::item_id)
             .inner_join(stock_on_hand_dsl::stock_on_hand)
-            .filter(stock_on_hand_dsl::available_stock_on_hand.gt(0.0))
+            .filter(
+                stock_on_hand_dsl::available_stock_on_hand
+                    .gt(0.0)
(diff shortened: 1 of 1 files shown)


lache-melvin (approved):
LGTM 🙏 

> I added extra test that fails without this change

Why not keep the test in the code? 👀

andreievg:
> Why not keep the test in the code? 👀

Yep it's there now, was just saying that without the fix the test fail, as in a TDDed it : 0)

PR #4639, opened 2024-08-20
4250 Status set as cancelled but pop up for save encountered as visited appears

### LINKED ISSUE #4250: Status set as cancelled but pop up for save encountered as visited appears
## What went wrong? 😲

Unsure if this is bug or expected behaviour.

Screenshot 2024-06-24 at 2 54 49 PM

When Encounters status is saved as "Cancelled" and you click save, a pop up appears and asks if you would like to save encounter as visited. Feels like it would be easy to make the mistake of selecting "Save as visited" which changes the status from cancelled to visited.

## Expected behaviour 🤔

If the status is selected as "Cancelled" wonder if it should just prompt as "Save encounter as cancelled" and then you select either "save" or "cancelled" and not have the third option of "save as visited".

## How to Reproduce 🔨

## Your environment 🌱

open msupply, programs hiv module

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #4250

# 👩🏻‍💻 What does this PR do?
Check jsonData status as well since status doesn't get changed until the `Save` button is pressed. Thing the modal prompt was implemented to remind users to tick off pass encounters as visited, so have kept it as is and not displaying any modal for cancelled appointments.

# 🧪 Testing

- [ ] Have patient programs set up
- [ ] Enrol a patient into a program
- [ ] Add an encounter for today or in the past 
- [ ] Change status to `Cancelled`
- [ ] Click `Save`
- [ ] Shouldn't see prompt and encounter is cancelled

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Encounter/DetailView/DetailView.tsx
+++ b/client/packages/system/src/Encounter/DetailView/DetailView.tsx
@@ -213,9 +213,13 @@ export const DetailView: FC = () => {
       data as unknown as EncounterSchema,
       updateEncounter
     );
+  const dataStatus = data
+    ? (data as Record<string, JsonData>)['status']
+    : undefined;
   const suggestSaveWithStatusVisited = encounter
     ? new Date(encounter.startDatetime).getTime() < Date.now() &&
-      encounter.status === EncounterNodeStatus.Pending
+      encounter.status === EncounterNodeStatus.Pending &&
+      dataStatus === EncounterNodeStatus.Pending
     : false;
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -213,9 +213,13 @@ export const DetailView: FC = () => {
       data as unknown as EncounterSchema,
       updateEncounter
     );
+  const dataStatus = data
+    ? (data as Record<string, JsonData>)['status']
```
andreievg, line 217: Not a fan of using this 'as', i would imagine, it's common in program module work, would be great to relocate it further up the chain at some point.

andreievg (approved):
Looks good and makes sense !

PR #4644, opened 2024-08-21
4425 add better error messaging for unsynced passwords

### LINKED ISSUE #4425: Better login error messaging
## What went wrong? 😲

I've had this one a few times when trying to login:

Screenshot 2024-07-16 at 10 30 48 AM

Looking at the explanatory comment in the code:

```
  VerifyPasswordError::InvalidCredentialsBackend(_) => {
      LoginError::InternalError("Failed to read credentials".to_string())
...

    /// Invalid account data on the backend
    InvalidCredentialsBackend(bcrypt::BcryptError),
```

does not explain the problem.

## Expected behaviour 🤔
I would prefer a message which told me something in English.
A couple of thoughts:
1. The system shouldn't get into this state. If the problem is a duplicate user (see below) then how did it create a second user with the same login? If the issue is a blank password (also below) how is this possible?
2. The message should tell me how to get out of this mess. I want to login and it isn't helping me resolve the problem. Perhaps `The user account has invalid credentials stored in the database. Please try logging in as another user, or contact support.` This should show instead of `The server experienced a problem` and not require the tooltip

## How to Reproduce 🔨

Looking at my database, I have this:

image

and from memory, having a duplicate user record has been a problem before.

These are the steps I took:

1. Have a site configured in mSupply
4. Sync this to oms and then login to the site
5. Delete the oms database
6. Recreate the database and sync, try to login

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #4425 

Simple fix to filter for user accounts with valid hash when logging in to remote sites.

I also thought might be nice to have a support prompting message.

This fix is not exhaustive to other potential internal errors. I'm not sure what these might be? 

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

# 📃 Documentation

  1.
  2.


server/service/src/user_account.rs
```diff
@@ -166,11 +166,23 @@ impl<'a> UserAccountService<'a> {
...
         let repo = UserRepository::new(self.connection);
-        repo.query_one(
+        let result = repo.query_one(
             UserFilter::new()
                 .id(EqualFilter::equal_to(user_id))
+                .hashed_password(EqualFilter::not_equal_to(""))
                 .site_id(EqualFilter::equal_to_i32(site_id)),
-        )
+        );
+        println!(
```
roxy-dao, line 175: Remove~?

server/service/src/user_account.rs
```diff
@@ -181,17 +193,15 @@ impl<'a> UserAccountService<'a> {
...
-            .find_one_by_user_name(username)
+            .find_one_active_by_username(username)
             .map_err(VerifyPasswordError::DatabaseError)?
         {
             Some(user) => user,
             None => return Err(VerifyPasswordError::UsernameDoesNotExist),
         };
         // verify password
-        let valid = verify(password, &user.hashed_password).map_err(|err| {
-            error!("verify_password: {}", err);
```
roxy-dao, line 192: was there a reason you removed the error loggin~?

client/packages/common/src/intl/locales/en/app.json
```diff
@@ -55,6 +55,7 @@
   "error.integration-timeout-reached": "The central server took too long integrating pushed records. Please retry later.",
   "error.invalid-url": "Invalid url",
   "error.login": "Invalid username or password",
+  "error.login-support": "Please contact support at support@msupply.foundation if you are unable to login to your account",
```
mark-prins, line 58: our other 'contact support' messaging has this format: no need to say 'support at support@'
```suggestion
  "error.login-support": "Please contact support@msupply.foundation if you are unable to login to your account",
```

lache-melvin (changes requested):
Hey Ferg, thanks for this! Was just reading the issue, I think @mark-prins was asking for more clarity for the user around what the error actually was... the reproducible error seems to be when OMS can't hit OG to get the hashed_password - in that case my inputted credentials aren't invalid, its a connection issue, and we should communicate that!

I had a chat with Mark, think the way forward is:

- On login, if there _is_ a hashed password in the DB, do usual validate credentials and return the invalid credentials error if wrong password
- If, when doing the validate credentials, the hashed_password == "", check to see if the `fetch_user_from_central` call in the `do_login` method got a `ConnectionError`. If so, return a new error to the user saying that mSupply Central couldn't be reached. 
  - This should be the only valid reason that hashed_password is still "" by the time we get to the validating credentials step - otherwise fetch_user_from_central should have populated the password
  - It's something the user can do something about, validating the connection with the central server
- If we _did_ reach the central server (no error on the fetch_user_from_central), but somehow hashed_password is still "", something went wrong that neither the system nor the user can recover from. In that case we should show the InternalError.
  - Can add your `contact support` hint to the internal error rather than the invalid credentials here!

I can't replicate the duplicate user entries part that Mark mentioned - maybe we keep an eye on this as a team, if we can get reproduction steps then we can determine if it's a bug or a dev set up issue 😁

server/service/src/user_account.rs
fergie-nz: No reason! Adding back logging

server/service/src/login.rs
```diff
@@ -627,6 +639,33 @@ mod test {
...
+                },
+                0,
+            )
+            .await;
+
+            assert_matches!(
+                result,
+                Err(LoginError::MSupplyCentralNotReached)
+            );
+        }
```
fergie-nz, line 668: I wanted also here to add a check for internal error when the password was missing and mSupply central was reached, but I couldn't get the set up working

server/graphql/general/src/queries/login.rs
```diff
@@ -47,6 +47,15 @@ impl MissingCredentials {
     }
 }
 
+pub struct CentralSyncRequired;
+#[Object]
+impl CentralSyncRequired {
+    pub async fn description(&self) -> &str {
+        "Missing hash sync"
```
lache-melvin, line 54: Yeah think "couldn't reach mSupply central server" is clearer what went wrong

server/service/src/login.rs
lache-melvin: Had a go here! https://github.com/msupply-foundation/open-msupply/compare/4425-add-better-error-messaging-for-unsynced-passwords...4425-add-better-error-messaging-for-unsynced-passwords(test-internal-error)

lache-melvin (changes requested):
Looking great Ferg! Mostly just wording changes, and removing the active user query 🙏

server/graphql/general/src/mutations/update_user.rs
```diff
@@ -82,7 +82,9 @@ pub async fn update_user(ctx: &Context<'_>) -> Result<UpdateResponse> {
                 | LoginError::LoginFailure(LoginFailure::NoSiteAccess)
                 | LoginError::InternalError(_)
                 | LoginError::DatabaseError(_)
-                | LoginError::FailedToGenerateToken(_) => {
+                | LoginError::FailedToGenerateToken(_) 
+                | LoginError::MSupplyCentralNotReached
```
fergie-nz, line 86: Mapping mSupply not reached + hash password to internal_error for update user

client/packages/common/src/intl/locales/en/app.json
```diff
@@ -57,12 +57,14 @@
...
   "error.invalid-url": "Invalid url",
   "error.login": "Invalid username or password",
+  "error.login-support": "Please contact support@msupply.foundation if you are unable to login to your account",
   "error.account-blocked": "Account is blocked until the lockout period has expired",
   "error.no-site-access": "Your account does not have access to any stores on this site",
   "error.server-failed-to-start": "Error! The server failed to start",
   "error.server-not-found": "We are having a problem finding an Open mSupply server",
   "error.site-auth-timeout": "Site authentication timed out, please try again later",
   "error.site-has-no-store": "Site has no store",
+  "error.missing-central-sync": "Please sync with mSupply central to validate your user credentials",
```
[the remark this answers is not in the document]
fergie-nz: Nice sounds good

server/service/src/login.rs
fergie-nz: That is epic! Thanks Laché!

I fear it won't test the exact use case of receiving an OK user_info from mSupply central rather than an internal error. But I'm guessing that would add complexity for simulating the response from mSupply.

Do you reckon worthwhile trying to add that use case?

lache-melvin: That _shouldn't_ really be possible...

You can see here, if fetch from central succeeds, we call update_user, which hashes the inputted password. If we somehow don't get a hashed password, something has gone rather awry! So I think fine to rely on internal error 😁 

https://github.com/msupply-foundation/open-msupply/blob/a2744c6d316dd1f2df938ba5df66a2bed93e69bb/server/service/src/login.rs#L118-L123

server/service/src/login.rs
```diff
@@ -152,6 +157,13 @@ impl LoginService {
                         LoginError::InternalError("Failed to read credentials".to_string())
                     }
                     VerifyPasswordError::DatabaseError(e) => LoginError::DatabaseError(e),
+                    VerifyPasswordError::EmptyHashedPassword => {
+                        if connection_failure {
+                            LoginError::MSupplyCentralNotReached
+                        } else {
+                            LoginError::InternalError("Corrupted credentials".to_string())
```
lache-melvin, line 164: Changed this error message, as it's currently mapping all the way through to the frontend, and this felt a little too raw:

Screenshot 2024-09-04 at 4 11 28 PM

Arguably we shouldn't display the backend internal error messages to the user, but probably more helpful than not at the moment 😅

lache-melvin (approved):
Sweet!! Hope you don't mind, I pushed a couple of small changes, just clarifying a comment in the test I suggested, and changing the error message 😁 

Thanks for changes Ferg 🙏

PR #4645, opened 2024-08-21
#3641 (2) Fix default patient schema DOB component

### LINKED ISSUE #3641: Patient ages: display as months and days if under 1 year old

CarlosNZ:
Fixes #3641 (again 🙄)

# 👩🏻‍💻 What does this PR do?

Updated the default (fallback) Patient UI schema to use the proper "Date of birth" component rather than generic date picker.

# 🧪 Testing

If you have an existing Programs schema installed (eg PNG), you won't see the default one (hence why we missed this the first time). In order to test/confirm, go to `PatientView.tsx` and change line 112 temporarily to:
```ts
  const patientRegistry = DEFAULT_SCHEMA;
```

### DIFF
--- a/client/packages/system/src/Patient/PatientView/DefaultPatientUISchema.json
+++ b/client/packages/system/src/Patient/PatientView/DefaultPatientUISchema.json
@@ -45,12 +45,9 @@
           "label": "Last Name"
         },
         {
-          "type": "Control",
+          "type": "DateOfBirth",
           "label": "Date of Birth",
-          "scope": "#/properties/dateOfBirth",
-          "options": {
-            "disableFuture": true
-          }
+          "scope": "#"
         },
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
LGTM~

PR #4648, opened 2024-08-21
2389 Carry customer reference field through from requisition to outbound shipment

### LINKED ISSUE #2389: Carry customer reference field through from requisition to outbound shipment
It would be nice for the customer reference field value to flow through from the customer requisition through to the outbound shipment - as @louisaw123 said, it "just seems like something that will be easily missed by the user cos they typed it in during the requisition then when the create shipment opens u would think its already there and not look in that spot again"

Could it flow through?

roxy-dao:
Fixes #2389

# 👩🏻‍💻 What does this PR do?
Carry over response requisition reference to outbound

# 🧪 Testing

- [ ] Have a requisition
- [ ] Type some reference or keep as since pre-populates with `From internal order X`
- [ ] Supply some stock
- [ ] Click `Create Shipment`
- [ ] See reference has been carried through

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/requisition/response_requisition/create_requisition_shipment/generate.rs
+++ b/server/service/src/requisition/response_requisition/create_requisition_shipment/generate.rs
@@ -38,13 +38,13 @@ pub fn generate(
         status: InvoiceStatus::New,
         created_datetime: Utc::now().naive_utc(),
         requisition_id: Some(requisition_row.id),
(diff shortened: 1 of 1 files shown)


CarlosNZ (approved):
LGTM! Such a small change :)

PR #4649, opened 2024-08-21
1633 disable actions if invoice/requisition is uneditable

### LINKED ISSUE #1633: Disable actions if invoice/requisition is uneditable
## What went wrong? 😲
Screenshot 2023-04-28 at 12 18 40 PM

## Expected behaviour 🤔
Note: This is behaviour is blocked in backend. 
Disable actions if invoice/requisition is uneditable (see stocktake for example), maybe also disable the checkboxes too? 

## How to Reproduce 🔨
- Make invoice/requisition uneditable by changing status
- See that you can still select `Delete lines` in actions

roxy-dao:
Fixes #1633

# 👩🏻‍💻 What does this PR do?
Disable actions that aren't available. Thought of disabling whole action bar if user can't interact with it, but we now have returns, so have decided to just disable the action for unification. Also fixed dup return in prescriptions

# 🧪 Testing

- [ ] Have finalised Outbound/Inbound/Returns
- [ ] Click action
- [ ] See actions that aren't available with status disabled.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/Footer/StatusChangeButton.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Footer/StatusChangeButton.tsx
@@ -139,20 +139,16 @@ export const StatusChangeButton = () => {
     lines?.totalCount === 0 ||
     lines?.nodes?.every(l => l.type === InvoiceLineNodeType.UnallocatedStock);
 
-  if (!selectedOption) return null;
-  if (isDisabled) return null;
-
   const noLinesNotification = useDisabledNotificationToast(
     t('messages.no-lines')
   );
-
-  if (!selectedOption) return null;
-  if (isDisabled) return null;
 
   const onStatusClick = () => {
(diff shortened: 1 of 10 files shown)


client/packages/invoices/src/Prescriptions/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -139,20 +139,16 @@ export const StatusChangeButton = () => {
...
     lines?.nodes?.every(l => l.type === InvoiceLineNodeType.UnallocatedStock);
 
-  if (!selectedOption) return null;
-  if (isDisabled) return null;
-
   const noLinesNotification = useDisabledNotificationToast(
     t('messages.no-lines')
   );
 
-  if (!selectedOption) return null;
```
roxy-dao, line 149: Side quest... dup return

mark-prins (approved):
🎉  very nice thnx Roxy!

PR #4659, opened 2024-08-22
#4233 Fix filter layout ordering

### LINKED ISSUE #4233: Layout bug with multi-item filters
## What went wrong? 😲

I *think* this is only a problem with multi-element filters (like Date range).

If you add a single filter after a "Range" filter (like Date), the order gets messed up when you change focus.

For example, the "Code or Name" filter jumps to the beginning after clicking in one of the Date filters:

## Expected behaviour 🤔

Filter order should persist during interaction.

## How to Reproduce 🔨

1. On "View Stock" list page, add an "Expiry" filter
2. Add a "Code or Name" filter -- it should appear after the two date inputs
3. Click in one of the Date inputs. Then click elsewhere on the page
4. the "Code or Name" filter jumps to the start of the filter line. 😢

## Your environment 🌱

- v2.1.0-rc1
- Desktop
- Firefox browser

CarlosNZ:
Fixes #4233

# 👩🏻‍💻 What does this PR do?

Prevents the ordering of active filters being changed when they are updated

The problem was in the `useEffect` that runs when the filters are changed -- they were being recreated using the overall filter definitions list and following that ordering (it's the order they show up in the Filter Menu), and if the user had instantiated the filters in a different order, this would be overwritten.

Solution is to check the existing "activeFilters" and only add new filters (if they're in the URL) if not already active.

# 🧪 Testing

- Follow the steps in the [original issue](#4233) to see the the problem. However, note that the problem occurs whenever you load the filters in a different order to how they are listed in the menu, not just with the "range" filters as suggested in the issue.
- Switch to this branch and ensure the problem goes away.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Filters/FilterMenu.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/FilterMenu.tsx
@@ -90,12 +90,7 @@ export const FilterMenu: FC<FilterDefinitions> = ({ filters }) => {
   // i.e. choosing a filter option in one filter changes the options in another
   useEffect(() => {
     setActiveFilters(
-      flattenFilterDefinitions(filters).filter(
-        fil =>
-          Object.keys(urlQuery).includes(fil.urlParameter) ||
-          fil.isDefault ||
-          activeFilters.some(f => f.name === fil.name)
-      )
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
Thanks Carl~

Issue #1392, opened 2023-03-20
Android: Server error after upgrade


mark-prins:
## What went wrong? 😲
 
"after upgrading a second tablet from v1.1.5 to v1.1.6 I again get an "oops something went wrong error" after the splash screen, and can't go further."

## Expected behaviour 🤔
You can go further

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Install v1.1.5 via ME MDM
2. Configure as client or server + initialise
3. Install v1.1.6 via ME MDM
4. Start omsupply

## Your environment 🌱
 - Version: any
 - Platform: tablet

mark-prins:
DB migrations are failing in this scenario

![image](https://user-images.githubusercontent.com/9192912/226517049-92862074-e608-4d95-a69d-30f59716a7bc.png)

mark-prins:
created a build with more logging (1.1.7), cleared the tablet, installed 1.1.5 via ME and then 1.1.7 and could not reproduce the issue

mark-prins:
closing with unable to reproduce. please reopen if resurfacing

PR #4666, opened 2024-08-22
1404 Add some logging to migrations

### LINKED ISSUE #1404: Add some logging to migrations
Coming from #1392 - the issue there is in the diesel migrations.
However the `migrations/mod.rs` file has this:

```
// TODO logging
pub fn migrate(
```

I'd like to have some logging.

I'm unsure of the intention of the TODO comment. At the least I'd like the error thrown by the diesel migration to be logged ( I was unable to see it in the log I checked ) and if the migration drops out due to version mismatch etc, then that should be logged, at an error level. 

Please check that the current _panic_ will get logged - if so, then that's sufficient.

roxy-dao:
Fixes #1404

# 👩🏻‍💻 What does this PR do?
Logs migration errors

# 🧪 Testing

- [ ] Change json version to lower than database version
- [ ] Have logging set to 'Error'
- [ ] Run
- [ ] Check log, error should have been logged

# 📃 Documentation

  1.
  2.


CarlosNZ (commented):
Seems good, although we're already seeing the error when it panics and stops.

Before:
Screenshot 2024-08-23 at 12 24 38 PM

After:
Screenshot 2024-08-23 at 12 23 15 PM

Do we need both?

CarlosNZ:
@andreievg had a suggestion to capture the panic error and log it as an Error.

roxy-dao:
> Seems good, although we're already seeing the error when it panics and stops.
> 
> Before: Screenshot 2024-08-23 at 12 24 38 PM
> 
> After: Screenshot 2024-08-23 at 12 23 15 PM
> 
> Do we need both?

The log file doesn't have the log... only in console...

andreievg:
We would just need to enable [this](https://github.com/msupply-foundation/open-msupply/blob/3615e4b961f3cf65221baca099266f58a9a7ca2d/server/android/src/android.rs#L69) in `server/src/main.rs` (not in `server/src/lib.rs`, as lib.rs is used by windows service which already does something else with [panic](https://github.com/msupply-foundation/open-msupply/blob/3615e4b961f3cf65221baca099266f58a9a7ca2d/server/windows/src/windows.rs#L70-L71))

Btw i am making conclusion that https://github.com/msupply-foundation/open-msupply/issues/1404 is only for mac or linux build/dev work. And related issue https://github.com/msupply-foundation/open-msupply/issues/1392, has already been fixed with panic logging in android 

I've committed my change straight to this branch since Roxy is away, please re-review and merge if happy thanks, thanks.

CarlosNZ (approved):
LGTM, will merge :)

PR #4669, opened 2024-08-22
4116 catalogue items list is querying twice for each render

### LINKED ISSUE #4116: Catalogue Items List is querying twice for each render
## What went wrong? 😲

Catalogue Items list queries twice on each render

Only occurs if user first clicks into detail view, and then clicks back to list view.

Maybe a global issue?

https://github.com/msupply-foundation/open-msupply/assets/90807420/e63ed903-9830-4394-a383-3540c3a8c73b

roxy-dao:
Fixes #4116

# 👩🏻‍💻 What does this PR do?
Fix itemWithStats being called twice 

# 🧪 Testing

- [ ] Go to item page and check network to see if api still being called twice

# 📃 Documentation

  1.
  2.


client/packages/system/src/Item/ListView/ListView.tsx
```diff
@@ -108,7 +111,12 @@ const ItemListComponent: FC = () => {
         isError={isError}
         isLoading={isLoading}
         onRowClick={row => {
-          navigate(`/catalogue/items/${row.id}`);
+          navigate(
```
roxy-dao, line 114: Sideqwest~

CarlosNZ (changes requested):
Oops, this broke the items filter. 😢

Screenshot 2024-08-23 at 12 52 39 PM

I'll try and fix it this afternoon.

andreievg:
I see that Laché assigned @CarlosNZ to fix, which leaves someone else to assign as reviewer, thank you

CarlosNZ:
@andreievg, have had a look at this.  Was planning to pause on it until I can inspect what's happening on Production server -- I suspect it may not be needed at all.

CarlosNZ:
@lache-melvin @andreievg 

So, it seems that this issue occurs on most of the List and Detail view components that are using the original API pattern. In the places where we've refactored it (Stock, Reports), it's behaving as it should (i.e. single GraphQL query on load). 

So I don't think we need to fix this specific case. The problem will naturally go away as we gradually refactor the API hooks to the new pattern.

Okay to close this?

lache-melvin:
Thanks for investigating Carl! Yeah that sounds good to me, another motivator to keep refactoring to the new pattern! 😁

CarlosNZ:
Closing as per [above comment](https://github.com/msupply-foundation/open-msupply/pull/4669#issuecomment-2319718938)

PR #4688, opened 2024-08-26
4676 add rust embed to lower dependency

### LINKED ISSUE #4675: Create vaccinations table
During an immunisation program encounter, each vaccination event needs to be captured.

This should be a new table `vaccination`:

id: String
encounter_id: String
user_id: String (user making the entry)
created_datetime: NaiveDateTime (time of entry)
vaccination_date: NaiveDate (user entered date of vaccination)
vaccine_course_schedule_id: String
vaccine_item_id: Option<String>
stock_line_id: Option<String> (item and stock line are optional 
clinician_link_id: Option<String> (healthcare professional administering the dose)
vaccine_given: bool
reason: Option<String> (customisable? REFUSAL/OUT OF STOCK/NO REASON)
comment: Option<String>

TODO: do we think date and clinician can be from the encounter level or needed for each individual vaccination?

Needs to sync via v6 as remote data

stock lines/items etc. references shouldn't break when synced to different sites! So no ref constraints

### LINKED ISSUE #4676: Relocate embed to lower dependency
Add rust embed for locales in service layer so that it is accessible.

Part of epic #4682

fergie-nz:
Fixes #4676

# 👩🏻‍💻 What does this PR do?

Adds rust embed to service layer so that it is visible throughout the app.

I wasn't sure if I should also be moving asset rust embed to deeper in the app also.

# 🧪 Testing

Developer Testing only, will be part of reports translation feature.

# 📃 Documentation

  1.
  2.


server/service/src/service_provider.rs
```diff
@@ -63,6 +63,12 @@ use repository::{
     PaginationOption, RepositoryError, StorageConnection, StorageConnectionManager, Store,
     StoreFilter, StoreSort,
 };
+use rust_embed::RustEmbed;
+
+#[derive(RustEmbed)]
+// Relative to server/Cargo.toml
+#[folder = "../../client/packages/host/dist"]
```
andreievg, line 70: I think you can point directly to the locale directory, and can also do cfg flag to point to translations src in for dev `#[cfg(debug_assertions)]` and `#[not(cfg(debug_assertions))]`

server/service/src/service_provider.rs
```diff
@@ -150,12 +156,15 @@ pub struct ServiceContext {
...
+
+
 impl ServiceProvider {
     // TODO we should really use `new` with processors_trigger, we constructs ServiceProvider manually in tests though
     // and it would be a bit of refactor, ideally setup_all and setup_all_with_data will return an instance of ServiceProvider
     // {make an issue}
     pub fn new(connection_manager: StorageConnectionManager, app_data_folder: &str) -> Self {
-        ServiceProvider::new_with_triggers(
+
+        return ServiceProvider::new_with_triggers(
```
andreievg, line 167: Is this 'return' added by some prettier/clipy config ? I actually prefer no return statement now (unless early return), it removes extra bit of verboseness.

andreievg (approved):
Looks great, thanks !

server/service/src/service_provider.rs
```diff
@@ -1,68 +1,43 @@
...
+#[cfg(debug_assertions)]
+#[derive(RustEmbed)]
+#[folder = "../../client/packages/common/src/intl/locales"] // Path for development
+struct DevLocalisations;
+
+
+// Define a struct for production
+#[cfg(not(debug_assertions))]
+#[derive(RustEmbed)]
+#[folder = "../../client/packages/host/dist/locales"] // Path for production
```
jmbrunskill, line 39: Pretty sure we can just have a single implementation here.
At build time the client packages should be available, and would simplify this...

jmbrunskill: The JSON for translations doesn't get modified in the webpack build process as far as I know.

server/service/src/service_provider.rs
```diff
@@ -171,6 +149,20 @@ impl ServiceProvider {
         sync_trigger: SyncTrigger,
         site_is_initialised_trigger: SiteIsInitialisedTrigger,
     ) -> Self {
+
+        // println!("new with triggers");
```
jmbrunskill, line 153: Comments to remove?

Issue #4682, opened 2024-08-23
Add front end translations to omSupply reports


fergie-nz:
Add report translations to mSupply

- [x] https://github.com/msupply-foundation/open-msupply/issues/4676
- [x] https://github.com/msupply-foundation/open-msupply/issues/4677
- [x] https://github.com/msupply-foundation/open-msupply/issues/4678
- [x] https://github.com/msupply-foundation/open-msupply/issues/4679
- [x] https://github.com/msupply-foundation/open-msupply/issues/4680
- [x] https://github.com/msupply-foundation/open-msupply/issues/4681
- [x] https://github.com/msupply-foundation/open-msupply/issues/4720
- [x] #4739

Later we can implement similar translations for Health Supply Hub emails

roxy-dao:
All tasks seem complete. Can this be closed? @fergie-nz

DhanyaHerath:
Same question as above but thought I'd just close it :) @fergie-nz

marthakeezy:
Report translations were tested in V2.3.0 Release Testing, so moving this issue to Done ✅

PR #4698, opened 2024-08-27
4676 add translation files to service provider

### LINKED ISSUE #4676: Relocate embed to lower dependency

fergie-nz:
Fixes #4676 

# 👩🏻‍💻 What does this PR do?

Adds localisations to service layer in back end.

Still some work to do and this is as yet incomplete:

- follow up work will continue as part of epic #4682 

# 🧪 Testing

# 📃 Documentation

  1.
  2.


server/service/src/localisations.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    // Load translations from embedded files
+    pub fn load_translations(&mut self) -> Result<(), std::io::Error> {
+        // Languages - need to extract these from the files themselves?
+        let mut languages = Vec::new();
+
+        // read all dirs within locales dir
+        let folder_paths: Vec<String> = fs::read_dir("../client/packages/host/dist/locales")
```
andreievg, line 38: This is not available in production, I think you would use `EmbeddedLocalisations::iter()`

server/service/src/localisations.rs
```diff
@@ -0,0 +1,87 @@
...
+            .unwrap()
+            .filter_map(|e| e.ok())
+            .map(|e: fs::DirEntry| e.path().as_os_str().to_string_lossy().to_string().split('/').last().unwrap().to_string())
```
andreievg, line 41: using `::iter()` would return `Cow` str, so should be easier to work with !

server/service/src/localisations.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        for lang in languages {
+            if let Some(content) = EmbeddedLocalisations::get(&format!("locales/{}/common.json", &lang)) {
+                let json_data = content.data;
+                let translations: HashMap<String, String> = serde_json::from_slice(&json_data).unwrap();
+                self.translations.insert(lang.to_string(), translations);
+
+            }
+        }
+        // later need to think about how to concatonate all translation json files per language
```
andreievg, line 56: I think we talked about this, we still need to access each string by 'name space' so shouldn't concatenate.

server/service/src/localisations.rs
```diff
@@ -0,0 +1,87 @@
...
+
+        for path in folder_paths {
+            languages.push(path)
+        }
+
+        for lang in languages {
+            if let Some(content) = EmbeddedLocalisations::get(&format!("locales/{}/common.json", &lang)) {
+                let json_data = content.data;
+                let translations: HashMap<String, String> = serde_json::from_slice(&json_data).unwrap();
+                self.translations.insert(lang.to_string(), translations);
```
andreievg, line 52: this will replace previous names space

server/service/src/localisations.rs
```diff
@@ -0,0 +1,87 @@
...
+
+            }
+        }
+        // later need to think about how to concatonate all translation json files per language
+
+        Ok(())
+    }
+
+    // Get a translation for a given key and language
+    pub fn get_translation(&self, key: &str, language: &str) -> String {
```
andreievg, line 62: I think there would be 4 parameters:

language, namespace, key, fallback

andreievg (changes requested):
Thanks in general look good, but a few requests.

andreievg:
I was just playing around a little bit, and had something like this (for reference, I thought it would help with readability), untested:

```rust
type Translations = HashMap<
    /* Language */ String,
    HashMap<
        /* Namespace */ String,
        HashMap</* TranslationKey */ String, /* TranslationString */ String>,
    >,
>;

const DEFAULT_LANGUAGE: &str = "en";

fn example(language: &str, namespace: &str, key: &str, placeholder: &str) -> String {
    let translation: Translations = HashMap::new();

    translation
        .get(language)
        .map(|l| l.get(namespace).map(|n| n.get(key)))
        .flatten()
        .flatten()
        .map(String::to_string)
        .unwrap_or_else(|| {
            if DEFAULT_LANGUAGE == language {
                placeholder.to_string()
            } else {
                example(language, namespace, key, placeholder)
            }
        })
        .clone()
}
```

server/service/src/localisations.rs
fergie-nz: Nice, I see now. I've updated to have a more nested hashmap so we can access via language, namespace, and key

server/service/src/service_provider.rs
```diff
@@ -180,9 +139,9 @@ impl ServiceProvider {
         sync_trigger: SyncTrigger,
         site_is_initialised_trigger: SiteIsInitialisedTrigger,
     ) -> Self {
-        if let Some(content) = Localisations::get("locales/en/common.json") {
-            println!("{:?}", content.data);
-        }
+        let mut localisations = Localisations::new();
```
andreievg, line 142: load_translations could be with ::new() -> inline when boxing

andreievg (approved):
Looks great, thanks for changes !

Issue #4652, opened 2024-08-21
Mark programs as Immunisation programs


lache-melvin:
Currently immunisation programs are a separate kind of program, only stored in OMS Central. We need to change this to creating a program in legacy mSupply, then marking that program as an immunisation in Open mSupply. This will allow us to use existing patient program functionality.

STEPS:
- [x] mSupply change - move is_immunisation flag to mSupply
- [x] Update translation to map in this new flag
- [x] bump sync versions
- [x] UI changes - remove Immunisation program create button - list should already contain all immunisation programs
- [x] Remove delete immunisation programs... they would be deleted in mSupply - ensure this is a soft delete though? if there is patient data they need to be soft deleted
- [ ] Remove immunisation program context?

lache-melvin:
> * [ ]  mSupply change - move is_immunisation flag to mSupply


https://github.com/msupply-foundation/msupply/pull/15168

marthakeezy:
Tested by Tainá in #4642 - moving to Done ✅

PR #4708, opened 2024-08-29
4652 remove immunisation program oms mutations

lache-melvin:
Start of #4652

# 👩🏻‍💻 What does this PR do?

THIS IS THE AGGRESSIVE DELETION PR

We are moving to managing immunisation programs through mSupply, like other programs (needed to loop in with patient program functionality)

So no need to create/delete programs through OMS

Kept the update mutation because I figured we'll probably use something close to that for the adding age ranges etc. - probably faster than having to reimplement from scratch

Not much to review, just check she all still works. Just wanted all this noise separate from actual code changes 😁 

# 🧪 Testing

- [ ] Immunisation programs view on OMS central has no buttons to do anything with 😁 

# 📃 Documentation

  1.
  2.


server/service/src/stock_line/tests/query.rs
```diff
@@ -22,14 +22,14 @@ mod query {
             service.get_stock_lines(
                 &context,
                 Some(PaginationOption {
-                    limit: Some(2000),
+                    limit: Some(6000),
```
lache-melvin, line 25: This test was failing in develop, looks like max limit got increased?

Chris-Petty:
Looks nuked to me!

Tests pass locally. Some fluff the compiler is complaining about you might want to trim out ?

image

Chris-Petty (approved):

lache-melvin:
did some more deleting - could you double check happy with deleting the deleted datetime 😅

Chris-Petty (approved):
Hmm LGTM. Was deleted date time just programs that you make independently of the OG central server and master list (and `master_list.is_active`)?

lache-melvin:
> Hmm LGTM. Was deleted date time just programs that you make independently of the OG central server and master list (and `master_list.is_active`)?

Yep was only added for the independent immunisation programs 👍 Had a quick look and don't think we check whether master list is active in current OMS program queries... might need to at some point but I think a separate issue 🙏

PR #4716, opened 2024-08-30
4652 translate is immunisation program

### LINKED ISSUE #4652: Mark programs as Immunisation programs

lache-melvin:
Fixes #4652

# 👩🏻‍💻 What does this PR do?

Translates is_immunisation from mSupply 

You will need this mSupply branch: https://github.com/msupply-foundation/msupply/pull/15168

# 🧪 Testing

- [ ] Per epic, set up at least one program as a master list
- [ ] Sync
- [ ] It shows in OMS central server in programs > immunisations 

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/sync/translations/program_requisition_settings.rs
+++ b/server/service/src/sync/translations/program_requisition_settings.rs
@@ -26,6 +26,7 @@ pub struct LegacyListMasterRow {
     is_program: bool,
     #[serde(rename = "programSettings")]
     program_settings: Option<LegacyProgramSettings>,
+    is_immunisation: bool,
 }
 
 #[derive(Deserialize, Clone)]
@@ -180,7 +181,7 @@ fn generate_requisition_program(
         master_list_id: Some(master_list.id.clone()),
         name: master_list.description.clone(),
         context_id: context_row.id.clone(),
-        is_immunisation: false,
+        is_immunisation: master_list.is_immunisation.clone(),
         deleted_datetime: None,
     };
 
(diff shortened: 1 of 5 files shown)


server/service/src/sync/translations/program_requisition_settings.rs
```diff
@@ -180,7 +181,7 @@ fn generate_requisition_program(
         master_list_id: Some(master_list.id.clone()),
         name: master_list.description.clone(),
         context_id: context_row.id.clone(),
-        is_immunisation: false,
+        is_immunisation: master_list.is_immunisation.clone(),
```
Chris-Petty, line 184: nit: bool doesn't need cloning

Chris-Petty (changes requested):
Not sure why but here test error:

```
failures:

---- sync::test::pull_and_push::test_sync_pull_and_push stdout ----
thread 'sync::test::pull_and_push::test_sync_pull_and_push' panicked at repository/src/db_diesel/master_list_name_join.rs:92:9:
assertion `left == right` failed
  left: Ok(None)
 right: Ok(Some(MasterListNameJoinRow { id: "A7A06D78361041448B836857ED4330C4", master_list_id: "87027C44835B48E6989376F42A58F7E3", name_link_id: "name_store_a" }))
```

Fine on develop though!

lache-melvin:
> Not sure why but here test error:

Ah great catch thanks Chris, was only running the program tests 🙈 Forgot to update the other master list translation test data! 🙏

Chris-Petty (approved):
beaut

PR #4717, opened 2024-08-30
4653 vaccine course dose edit modal

### LINKED ISSUE #4653: Add dose schedule to Vaccine Course edit modal
Replace the `Number of doses` field:

Screenshot 2024-08-22 at 10 28 16 AM

With the dose schedule table:

Screenshot 2024-08-22 at 10 28 01 AM

Number of doses can be calculated based on number of rows.

Per existing VaccineCourseSchedule - just add dose number and label for now, additional fields in separate issue

lache-melvin:
Fixes #4653

# 👩🏻‍💻 What does this PR do?

Adds dose table to vaccine course edit modal:

Screenshot 2024-08-30 at 5 40 43 PM

Age/interval columns don't save - thats a separate issue

Put the button in the table header to try save some vertical real estate, but plz be honest if you hate the layout lol

# 🧪 Testing

- [ ] For a configured immunisation program
- [ ] Add a vaccine course
- [ ] You can add and delete doses from the course
- [ ] When you save, the number of doses in list view is correct

# 📃 Documentation

  1.
  2.


adamdewey:
This is great @lache-melvin :))

If it's not super tricky I'd put the dose button above the dose column, and replace the current column header?

But don't worry if it's a faff, I think it will be fine as it is too!

lache-melvin:
Screenshot 2024-09-02 at 8 55 59 AM

After quick discussion have put outside the table but still on the right, as is the typical location in other places (shipments/stocktakes etc) ... thoughts @adamdewey

adamdewey:
@lache-melvin - I think I would put it on the left, but also on the right is fine for version 1! 

Perhaps @mariyamsupply has some low effort suggestions?

adamdewey:
More thought: I regret making alternative suggestions, only the central user will see this and not very frequently - I think fine where it is, thanks @lache-melvin ! 😅

client/packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -227,7 +227,12 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
       >
         {title ? <ModalTitle title={title} /> : null}
         <form
-          style={{ display: 'flex', flexDirection: 'column', flex: '1 1 auto' }}
+          style={{
+            display: 'flex',
+            flexDirection: 'column',
+            flex: '1 1 auto',
+            overflow: 'auto',
```
CarlosNZ, line 234: What difference does this make?

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -217,3 +216,125 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
...
+};
+
+const NumericDoseTableCell = ({
+  value,
+  onChange,
+}: {
+  value: number;
+  onChange: (num: number) => void;
+}) => {
+  const [buffer, setBuffer] = useBufferState<number | undefined>(value);
```
CarlosNZ, line 326: I thought we were trying to move away from using this hook these days? Is it necessary?

Also, this component seems pretty generic -- can we just use the existing `NumberInputCell`? (There doesn't seem be anything specific to this particular use case here)

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -217,3 +216,125 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
...
+              <NumericDoseTableCell
+                value={dose.minAgeMonths}
+                onChange={num => updateDose(dose.id, { minAgeMonths: num })}
+              />
+              <NumericDoseTableCell
+                value={dose.minIntervalDays}
+                onChange={num => updateDose(dose.id, { minIntervalDays: num })}
+              />
+              <td style={{ display: 'flex', justifyContent: 'center' }}>
+                <IconButton
```
CarlosNZ, line 301: Is there anywhere else we've used the Delete icon like this? (I can't find one). If so, I guess just copy that, but if not, I feel like it could be a tiny bit smaller? Not totally sure though.

CarlosNZ (commented):
Looking good :) 

Couple of comments, see what you think.

client/packages/common/src/hooks/useDialog/useDialog.tsx
lache-melvin: Means modal contents will scroll, but the action buttons at the bottom are always visible, rather than having to scroll down to the buttons

lache-melvin: Thanks, sorry forgot to add a comment about this 🙏

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
lache-melvin: Ah yeah you're right, had some other logic in here but didn't end up needing it... Didn't use the `NumberInputCell` because wasn't using our usual DataTable component, but realised I probably should have! Updated [6d02161](https://github.com/msupply-foundation/open-msupply/pull/4717/commits/6d021613ce812a31627bd0e33bde52e0640cfa4b) 😁

CarlosNZ: Hmm, looks-wise I think it was better how you had it. I don't think the DataTable UI suits a small space like this. Not sure what to suggest. Quick chat at tomorrow stand-up?

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -210,10 +215,117 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
...
+          onClick={addDose}
+        />
+      </Box>
+      <TableProvider createStore={createTableStore}>
+        <DataTable
+          id={'doses-list'}
+          columns={columns}
+          data={doses}
+          noDataMessage={t('message.add-a-dose')}
+          dense
```
lache-melvin, line 326: That was easy @CarlosNZ 😁 

Screenshot 2024-09-03 at 9 45 06 AM

CarlosNZ: Love it 👍

CarlosNZ (approved):
LGTM, thanks for the changes :)

PR #4723, opened 2024-09-02
4654 add min age and interval columns

### LINKED ISSUE #4654: Add minimum_interval and minimum_age to vaccine course schedules
We need to capture the min number of days between doses, and when a patient should be able to receive the dose:

Screenshot 2024-08-22 at 10 42 02 AM

- new `Int` field on vaccine_course_schedule: `min_interval_days`
  - Min interval input should be visible/enterable for all doses except the last one 
  - Can default to 0 for the last dose
- Period input is now age (minimum age to receive the dose)
  - Measured in months, just a number input for now (may move to days/months/years selector at some point)

lache-melvin:
Fixes #4654

# 👩🏻‍💻 What does this PR do?

Adds min age and min interval fields in backend, so saving the vaccine course edit modal now saves those fields!

Also used this PR to do the rename VaccineCourseSchedule -> VaccineCourseDose... sorry about the PR bloat, tried to do all the renaming in the first commit so that might be easier way to review 🙏 

# 🧪 Testing

- [ ] Ensure latest migrations have run
- [ ] Create/edit doses on a vaccine course, edit the age and interval
- [ ] Save
- [ ] Open the vaccine course again - your changes should have persisted

# 📃 Documentation

  1.
  2.


server/graphql/types/src/types/vaccine_course_dose.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[Object]
+impl VaccineCourseDoseNode {
+    pub async fn id(&self) -> &str {
+        &self.row().id
+    }
+
+    pub async fn dose_number(&self) -> &i32 {
+        &self.row().dose_number
+    }
```
CarlosNZ, line 18: Did you want to remove the dose number in this PR or make a follow-up issue?

lache-melvin: Will do here, thanks Carl 🙏

CarlosNZ (approved):
Looks good (though I skimmed through the ones that just looked like renaming, which is most of it). 

One tiny question about `f64`, very trivial :)

CarlosNZ:
> Looks good (though I skimmed through the ones that just looked like renaming, which is most of it).
> 
> One tiny question about `f64`, very trivial :)

I'll have a final quick look once you've made the final tweaks to the dose number field :)

lache-melvin:
Changes added, thanks Carl 🙏

server/graphql/vaccine_course/src/mutations/update_vaccine_course.rs
```diff
@@ -46,10 +46,12 @@ pub fn update_vaccine_course(
 }
 
 #[derive(InputObject, Clone)]
-pub struct UpsertVaccineCourseScheduleInput {
+pub struct UpsertVaccineCourseDoseInput {
     pub id: String,
     pub label: String,
     pub dose_number: i32,
+    pub min_age: f64,
```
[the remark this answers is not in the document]
lache-melvin: Hm yeah probably... I've only seen f64 in the codebase for floating point values so was just keeping consistent, I think you're right here through 🤷‍♀️

CarlosNZ (changes requested):
Sweet, nearly there. I think we need to put an explicit sort on the dose order though. It seems to be relying on integrity checks for input and insert, but doesn't actually sort by age when querying. If I manually change the `min_age` field in the database they're returned/displayed out of order:

Screenshot 2024-09-03 at 3 29 08 PM

lache-melvin:
Sweet done - thanks for such thorough reviewing Carl! 🙏

CarlosNZ (approved):

Issue #4658, opened 2024-08-21
Initialising a new Open mSupply site with a lot of central server data takes a long time


DhanyaHerath:
## What went wrong? 😲
From Mark G: 

> Why does an Open mSupply store which:
> 
> * Handles items from one master list containing a dozen items
> * Receives stock from 1 supplier
> * Issues to 6 non-store customers
> Needs to be initialised with several hundred thousand records?  This takes literally hours at the end of a the average  cellular phone line, and is impractically long in the field.  The current work-around is to configuring devices where the internet is better than average, and then take them to the field.  We won't get away with this forever.  The time will come when we need to initiate a site in the field.
> 
> Why must every database have all system data (every item, master list, non-patient name, item-store-join, etc.)?  Why can't these details be synced as visibility to them is turned on?
> 

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

andreievg:
Triage, setting priority as must have to priorities refinement

andreievg:
Needs investigation, what is the breakdown of sync outs for small new sites.
The goal is to try and reduce number of records needed to be sent so let's see if we can identify any low hanging fruit.
Also discussed trying out connection timeout in reqwest crate and maybe do retries of connection.

andreievg:
During dev test day, me and @lache-melvin had issues initialising tablets, it was very slow and sometimes connection would be stuck for a long time, only to show up with `retry`. I did some investigation and even put caddy between mSupply and the outside world to setup timeouts, this didn't really have any effect. I added logging and saw that when connection is being 'timed' out, there wasn't even any caddy logs present (so was outside control of central servers).

There is a way to set connection timeout for reqwest, using client builder, while I was trying to set timeouts and replies I stumbled accros gzip header, in reqwest docs looks like it's not automatically turned on, and need gzip features for it to be enabled: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.gzip

[In this branch](https://github.com/msupply-foundation/open-msupply/compare/4662-omSupply-backups...Checking-sync-connection) I experimented with gzip, connection timeout and retries, and batch size increase. I did a few benchmarks connecting to the same central servers on debug build apk and it looks like it's 100-150% faster. I don't think the timeout and connection is helping, because today I never got the 'retry' msg.

**Conclusions**

[...]

PR #4724, opened 2024-09-02
4658 reduce central record sync volume kdd

Chris-Petty:
KDDs #4658


jmbrunskill:
I wonder if it's worth getting some general metrics, eg. how many bytes are we talking about transfering?
E.g. 
I'm wondering if things like increasing (or customising the batch size) could be beneficial?
Based on my example db, GZIP compression means we have a file size around 1/3rd the original size.
I'm not sure the sync v5 format, but thinking we should be talking about approximately 1MB compressed for 43,000  `list_master_line`  
I'm wondering if the number of round trips is more of a problem than the response size in bytes?

According to this website, if the whole `initial sync` was around 5MB, and we have a 256K connection it should be able to download in around 30s.
image

The initialization file `example1` in the OMS repo is currently 21mb but compressing it brings it down to 3MB...

image


Maybe it's worth considering some kind of master central datafile that gets downloaded on initialisation, then changelogs can take over from there? Just thinking of some alternatives...

decisions/2024-08-30_central_sync_queue_pull_reduction.md
```diff
@@ -0,0 +1,164 @@
...
+- If 5000 records are requested and we filter out some (or all) of them, do we scan ahead to grab some more records till the batch size is fulfilled (and continuing until done)? Probably negligible but none the less a slower.
+- Added overhead to getting records
+- If they actually need all of the master list lines, it's more overhead doing filtering for no reason. Can probably optimise this case though.
+- Doing work in LCS when we're trying to focus on OMS
+- Eventually we want OMS only deployments where OMS controls central data, so a solution in LCS land would require us eventually fixing it again in OMCS land.
+- Only fixes for OMS
+
+### Option 3 - OMCS change_log filtering per log processed
+
+The filtering could be more or less the same as in Option 2, but we do it in OMCS. Go full hog: We change to OMCS changelog pull and drop pulling LCS changelog.
```
jmbrunskill, line 127: In general I like this idea, especially as a bunch of R&D Work has been done on this possibility.
Then we're developing in the OMS code base rather than more code in legacy.

decisions/2024-08-30_central_sync_queue_pull_reduction.md
```diff
@@ -0,0 +1,164 @@
...
+- Future proof solution for the ages
+
+_Cons:_
+
+- Maybe not all records on sync v6 could come through? I don't see why that wouldn't be possible
+- Only fixes for OMS
+
+### Option 4 - OMCS change_log filter based on change_log.parent_id
+
+When master list line records are added to the changelog, we include a `parent_id` (could crudely use store_id for non-store records...)
```
jmbrunskill, line 142: This sounds a little bit confusing to me, does every record have a parent id? We need to look up master lists for every record before we can save it?

I think I'd rather do the filtering when we're querying change log...

jmbrunskill: I had a similar idea though. I was thinking we could have a single map with `record_id -> Vec<store_id>`
That would make the filtering quick at filtering time, just check if the requesting store needs this record or not.
It would have the same problem though, where we're need to calculate visibility for every central record when we create or update it.

decisions/2024-08-30_central_sync_queue_pull_reduction.md
```diff
@@ -0,0 +1,164 @@
...
+
+- Still potentially queuing lots of records to many sites. Some users will have >1000 sites and a general master list used across them with >5000 items. These instances will still suffer millions of queued records that central_change_log set out to fix to begin with!
+- Eventually we want OMS only deployments where OMS controls central data, so a solution in LCS land would require us eventually fixing it again in OMCS land.
+- Doing work in LCS when we're trying to focus on OMS
+
+### Option 2 - LCS central_change_log filtering per log processed
+
+We implement a system that allows sync v5 change log requests to have some filtering on the LCS side.
+
+Suppose a site requests 5000 change_log records from cursor 0:
```
jmbrunskill, line 104: What if they aren't querying from cursor 0?

E.g. I've initialised my site, and initially I don't need `Master List A`
Then `Master List A` is added to my site (Update on `master_list_name_join` I think?)

I start my cursor at 1000, and I see the new `master_list_name_join` record but not associated `master_list` would be sent right?

decisions/2024-08-30_central_sync_queue_pull_reduction.md
jmbrunskill: As per my question above I'm starting to think that we might need to have some kind of state management for what's visible at each site when records are saved, and related records might need to be sent (e.g. create a new cursor_id if visiblity changes) This has the challenge that it would potentially create 1000s of out going records to other sites that don't need the update!

Chris-Petty: > This sounds a little bit confusing to me, does every record have a parent id? We need to look up master lists for every record before we can save it?
>

For relevant tables it'd be recorded in the trigger where we already have the FK to put int the `changelog.parent_id`.
 
> I think I'd rather do the filtering when we're querying change log...

I agree but looking for ideas on how to do that effectively. Not sure how to do it without an ID to check. In a nutshell need to check `changelog.master_list_line.master_list_id IN site.stores.master_list_store_join.master_list_id // only for the changelogs for table == master_list_line`

decisions/2024-08-30_central_sync_queue_pull_reduction.md
Chris-Petty: ded :( yes it would be borken.

Chris-Petty: This actually affects all the filtering? For instance, it is an occasional activity to move a store from central to an active remote site (or even from another remote site). Currently past records for that store may be earlier in the changelog than the new site's cursor right? I suppose in this case we could query all changelogs by a store_id and bump the sequence number on them to the end of changelog.

jmbrunskill: Yes I think so, if we're filtering out record we'll run into this kind of issue...

Chris-Petty:
> I wonder if it's worth getting some general metrics, eg. how many bytes are we talking about transfering in these slow cases? E.g. I'm wondering if things like increasing (or customising the batch size) could be beneficial? Based on my example db, GZIP compression means we have a file size around 1/3rd the original size. I'm not sure the sync v5 format, but thinking we should be talking about approximately 1MB compressed for 43,000 `list_master_line` I'm wondering if the number of round trips is more of a problem than the response size in bytes?
> 
> According to this website, if the whole `initial sync` was around 5MB, and we have a 256K connection it should be able to download in around 30s. image
> 
> If we have 93861, and we get them in 500 record batches (this I think we do by default?) this is nearly 200 round trips to the server. If each one takes 1 second you're already at 3minutes vs 30seconds.
> 

Yup round trips certainly something we want to do better with bigger batch sizes. Plan is to do a refactor task implementing dynamic batch size approach such as in legacy mobile and I've already bumped LCS max batch size to 5000. There is certainly a trade off though, in some scenarios we have a reliable 256k connection would be god tier. It's often much lower than that and unstable at best, so even getting a batch of 100s of records can be undoable. This is why mSupply desktop in many scenarios has a default batch size of 20, and in certain countries (such as the one that raised this issue) it's lower still. Annoyingly a trade off with efficient gzip compression ratios though. The scenarios where connections are good batch of 20 is ridiculously low.

For the purpose of this KDD I was trying to keep it focused entirely on what we can do to reduce the count of records as there is something of a consensus to dynamically increase batch size already.

> The initialization file `example1` in the OMS repo is currently 21mb but compressing it brings it down to 3MB... image
> 
> Maybe it's worth considering some kind of master central datafile that gets downloaded on initialisation, then changelogs can take over from there? Or just see how much benefit a big batch size could give us?

Yea an initialisation file has legs. Similarly Legacy Desktop approach is to generate the datafile centrally and get it to the remote site either via USB or dropbox as doing the transfer that way is much easier. For some big sites where their data file is in the GBs and their connection is mediocre this is pretty much the only viable option. That's including millions of records of store data though, where as we're potentially just talking central data that goes into the changelog being included in the file.

For OMS there was a sharded central server idea where each site would have it's own DB on the central server - initialising would just be getting a copy of that db. Fell off as too complex for time pressure though, and in favour of known message queue system. (Also all store data again, complexities with where patient data goes, also pg vs sqlite).

decisions/2024-08-30_central_sync_queue_pull_reduction.md
andreievg: Such a great point @jmbrunskill.

So if we are filtering by a join (actually similar for name), we have an issue like this. 

I've had a good think of alternative way to do this, and I am struggling to figure out how to do it with logic that I would be ok with, apart from syncing all of that master data everywhere. Even updating master list and lines in changelog to the end when master list name join changes, doesn't sit well with me.

Although this issues is about trying to reduce number of records that are required for initialisation, i fee like it also crosses the question, can we actually sync all master data efficiently if we do some optimisation ?

andreievg:
@Chris-Petty @jmbrunskill 

How about we just filter out `reports`, i didn't realise how big of a difference that makes:

On my data file, i have 1223 central_change_log records. I bumped max for central sync in LCS and requested all central data in postman, my response was 28.5mb and 5.3mb compressed, i was very surprised that it would be 28.5mb.

After a quick look, i realised there are huge report blobs, and most reports are not even omSupply related. So filtered our reports on LCS, and my new response was 682kb and 76kb compressed.

Some quick math:

1 uuid: 16 bytes 

Say we have 100k master list and master list lines, so that would be 100k unique ids (not compressible) and the rest would be repeats, so lets say we double the final compressed size: 100k * 16bytes * 2 = 3.2mb <- !!! 

I think we should be able to sync up to 10mb during initialisation, even in places where internet is really really bad

andreievg:
In summary I think we can add the challenges listed by James and add another option just to filter out non omSup reports for now, conclude that it's hard to filter out some things without queue and suggest that dynamic batch size would be the next things to try before going for filtering (due to complexity especially with edge cases)

Chris-Petty:
Yea I'm happy to park for and address the known fixables thanks! Might become worth revisiting in future as central change log grows (5000000 facility names 🤪)

decisions/2024-08-30_central_sync_queue_pull_reduction.md
Chris-Petty: Yea I'm on board with parking and just focusing other optimisations due to this, it's a good catch @jmbrunskill!

Chris-Petty:
For my datafile with 90k records and lots of custom reports, compressed size of all records is 13MB, down from 128MB. Filtering out all non OMS reports is 7MB down from 98MB. 5000 is 9.5MB to 0.4MB. (n.b. compression level 6. Tried 9 but difference was marginal)

jmbrunskill:
> For my datafile with 90k records and lots of custom reports, compressed size of all records is 13MB, down from 128MB. Filtering out all non OMS reports is 7MB down from 98MB. 5000 is 9.5MB to 0.4MB. (n.b. compression level 6. Tried 9 but difference was marginal)

Maybe we should consider using `sync files` for custom reports in future?
Especially when they start having WASM in them, the size might grow...

andreievg (approved):
Thanks @Chris-Petty

jmbrunskill (approved):
Thanks Chris!

PR #4725, opened 2024-09-02
Update server-tests.yaml

Chris-Petty:
Crudely seeing if cargo clean will clear out some space. That said, I'd assume it's a fresh ubuntu runner anyway and would be starting from nothing. So this is a stab in the dark...

Other things to look at:

https://github.com/actions/runner-images/issues/2840
https://github.com/actions/runner-images/issues/2875 (leads to https://github.com/marketplace/actions/free-disk-space-ubuntu)
https://stackoverflow.com/questions/75536771/github-runner-out-of-disk-space-after-building-docker-image

### DIFF
--- a/.github/workflows/server-tests.yaml
+++ b/.github/workflows/server-tests.yaml
@@ -29,6 +29,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clean
+          args: --manifest-path server/Cargo.toml
       - uses: actions-rs/cargo@v1
         with:
           command: test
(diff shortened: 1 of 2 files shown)


Chris-Petty:
image
yep nothing to clean as suspected. Will try the [free-disk-space-ubuntu](https://github.com/marketplace/actions/free-disk-space-ubuntu) thing, basically all threads I can find suggest doing what this action does anyway (nuke a bunch of tooling directories for other tools e.g. .NET, Android...

Chris-Petty:
Ooo that did the trick. Has added a few minutes to the action running time though. Better than running for several minutes and aborting though!

PR #4733, opened 2024-09-03
4675 vaccinations table

### LINKED ISSUE #4675: Create vaccinations table

lache-melvin:
Fixes #4675

# 👩🏻‍💻 What does this PR do?

Just adds the table/repositories, no insert/data access yet

# 🧪 Testing

- [ ] Compiles/passes

# 📃 Documentation


server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
```diff
@@ -0,0 +1,36 @@
...
+        sql!(
+            connection,
+            r#"
+                CREATE TABLE vaccination (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    program_id TEXT NOT NULL,
+                    encounter_id TEXT NOT NULL,
+                    created_datetime TIMESTAMP NOT NULL,
+                    user_id TEXT NOT NULL REFERENCES user_account(id),
+                    vaccine_course_dose_id TEXT NOT NULL REFERENCES vaccine_course_dose(id),
```
jmbrunskill, line 22: Any plans to support vaccinations that aren't part of a course? E.g. Influenza?
Just wondering if `Not Null` is too strong constraint here...

server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
```diff
@@ -0,0 +1,36 @@
...
+                    store_id TEXT NOT NULL,
+                    clinician_link_id TEXT,
+                    invoice_id TEXT,
+                    vaccination_date {DATE} NOT NULL,
```
jmbrunskill, line 26: The issue mentions having a  `vaccine_item_id: Option` is that still required?
I'm thinking we probably do want to record exactly which item was administered?
Might need to be `vaccine_item_link_id`

Pls ignore me if I'm off base here...
Was just interested to see what the table structure is looking like :)

lache-melvin: Ah great catch! Was wondering whether just the linked invoice would be enough - though would need to go via that to the invoice_line to see what the item was. Probably easiest to duplicate here right?

And realising it's maybe better to just have item_link_id? I think it's possible that vaccine items change over time, some might be deleted if no longer recommended for the vaccine course. But would want whichever item was administered at the time of this vaccine to be persisted?

server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
lache-melvin: Great question... do you have preferences/upcoming requirements here @mark-prins?

server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
adamdewey: Yes, the vaccine item should be 'stamped' on the patient's record along with the batch

server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
mark-prins: No requirement at this stage to be able to administer items that aren't part of the course.
Doing so wouldn't fit into the concept of a vaccination card; which is to define a course of vaccines which are then ticked off / administered according to the schedule. So I'm fine with keeping the course_dose_id a hard link

server/repository/src/migrations/v2_03_00/add_vaccinations_table.rs
lache-melvin: Have gone with adding a stock line id - we can easily get the item from the stock line, and by not storing both IDs we don't have to have logic to handle e.g. stock_line_id is Some() but item_link_id is None

It's possible this will evolve a little whenever we are able to implement bundled items, but I think that's fine!

jmbrunskill: > Ah great catch! Was wondering whether just the linked invoice would be enough - though would need to go via that to the invoice_line to see what the item was. Probably easiest to duplicate here right?

Yeah I would probably duplicate... I think `item_link_id` would be correct. Tempting to put `item_id` for that trace back purpose Adam mentioned, but the item_link_id will still be the id for the old item so can still be associated if needed, we want the system to remain consistent in how it operates...

jmbrunskill: > Have gone with adding a stock line id - we can easily get the item from the stock line, and by not storing both IDs we don't have to have logic to handle e.g. stock_line_id is Some() but item_link_id is None
> 
> It's possible this will evolve a little whenever we are able to implement bundled items, but I think that's fine!

Sweet all good!

CarlosNZ (approved):
Looks good, thanks.

Issue #4679, opened 2024-08-23
Change existing reports to use translations


fergie-nz:
## Is your feature request related to a problem? Please describe 👀

Use preferred language and if not use english and if not add fallback in Tera to prevent failed renders.

Only add translations to core reports - don't need to worry about client specific reports at this stage.

Also remove reports.json (this won't be added to weblate).

In stead all of these will be added to common.json with the label of "report.x" rather than "label.x" to differentiate if required.

Part of epic https://github.com/msupply-foundation/open-msupply/issues/4682

fergie-nz:
We also have the future option of adding standard reports into the omSupply repo.

This has not been included in this issue however as needs more discussion

marthakeezy:
Anusha is smashing it in #4899, moving to Done ✅

PR #4743, opened 2024-09-05
4739 refactor translation function

### LINKED ISSUE #4739: Refactor translation function to handle errors and fall backs

Also would be helpful to see:

Documentation of functionality in tera
Examples of use in tera + testing of translation rendering in example report

fergie-nz:
Fixes #4739 

# 👩🏻‍💻 What does this PR do?

Refactors translate function to handle fallbacks
Makes translation function less verbose in use cases, and makes function more legible.
Adds documentation for translation function usage in readme
Adds actual html rendering tests in report service

# 🧪 Testing

Developer testing only at this stage.

More thorough testing helpful when standard reports have been translated as part of #4679

# 📃 Documentation

Docs updated for translation function usage


server/report_builder/README.md
```diff
@@ -183,3 +183,39 @@ For example, instead of `icon1` an alternative name can be used by using a refer
...
+```
+
+Where the letters are short hand for the following:
+- t for translate
+  The name of the function
+- k for key 
+  This is the locale key as is used in front end translations. 
+- l for language
+  This is an optional override for language. the default lang will use the language of the front end user.
+  To override the language of a particular key, you can substitute it in like:
```
fergie-nz, line 204: I thought an improvement would be adding the LocalisationService provider to the translation function somehow. So it would be an optional field in tera, and get_translation would default to the user language.

Then we could just have 
```
{{t(k="label.name", n="reports.json", f="Name")}}
``` 
throughout the html reports for an even less verbose option

However I couldn't quite figure out how!

server/service/src/localisations.rs
```diff
@@ -60,18 +72,44 @@ impl Localisations {
...
+        let fallback = args.get("f").and_then(serde_json::Value::as_str);
+
+        // make cascading array of fallback options:
+        for (language, namespace, key, fallback) in [
+            // first look for key in nomionaed namespace
+            (language, namespace, key, None), 
+            // then look for key in common.json
+            (language, "common.json", key, None), 
+            // then look for fallback
+            ("", "", "", fallback ),
```
fergie-nz, line 88: This structure just allows for more visible cascading fallback logic

andreievg (changes requested):
Thanks @fergie-nz, looks great, i haven't quite looked at every thing yet, but I would suggest this:

* Use this error
* First check specified namespace and only after that check common namespace
* Fallback at the very end (although I see why you would want to try fallback when we fail matching language, but would assume that we would use english language as first fallback and last fallback that is specified in the parameters would be english, because all translated reports would be english)

https://github.com/msupply-foundation/open-msupply/compare/4739-refactor-translation-function...4739-refactor-translation-function(Translation-example)

Another change, you can do in this PR or add to issue, namespace should not have `.json` in it, once again the consumer focused (make it as easy to use by consumer as possible)

andreievg:
There is no requirement to override language currently, technically speaking language can be overwritten by user changing their language settings. You can pass through default language of the user to the translation function, as per the macro examples in tera https://keats.github.io/tera/docs/#functions, and this should be passed on through the graphql query. 
Let me know, can have a peer programming session (tomorrow), and I can show you what I mean

andreievg (approved):
@fergie-nz this looks great thank you. 

I do have a couple more suggestions, but I think they need to be done in another PR:

* Pass current_language directly to get_translations (it can figure out the state of the `Option`, vs doing it in generate_report) -> more consumer focused
* Remove contex.insert("lang") is it still needed ?
* Map terra error so we can preserver the error source (and differentiate between different error types of `TranslationError`)  `.map_err(|e| TeraError::call_function("t", e))?;`
* We shouldn't need to read the code to much to figure out how the function is used, although it's reasonably simple to see what `get_translations` expects, I think we can improve is slightly, and also make it even easier on consumer following the example of tera `tera.register_function("url_for", make_url_for(urls));`. I really want to have a quick peer session with you on that front, to show you what I mean.
* Remove `let localisations = Localisations::new();` (just, translations_service: Box::new(Localisations::new()),  )
* Formatting, prettier (can sort this out in our peer session), also keen to talk about self PR review, I think you might have noticed quite a bit of formatting diff in [this PR](https://github.com/msupply-foundation/open-msupply/pull/4688/files), when you submitted it (formatting wise)

andreievg:
@fergie-nz thank you! happy for this to be merged

PR #4752, opened 2024-09-08
4679 add translations to standard reports

### LINKED ISSUE #4679: Change existing reports to use translations

fergie-nz:
Fixes #4679 

# 👩🏻‍💻 What does this PR do?

Adds translations for standard reports.

Adds keys into common.json (and removes the report.json namespace which won't be used in weblate)

# 🧪 Testing

You can test this branch by building reports from omsupply-reports-repo with branch:
#4679 (https://github.com/msupply-foundation/open-msupply-reports/pull/88)

And then generating them with different selected user languages

# 📃 Documentation

Documentation added in the reports readme. 

Front end user facing changes added:
https://github.com/msupply-foundation/msupply_docs/pull/81


client/packages/common/src/intl/locales/en/reports.json
```diff
@@ -1,18 +0,0 @@
-{
-    "description.notification-preferences": "By default, the system will tell you if you have more than 3 or 6 months of stock on the homepage dashboard.",
```
andreievg, line 2: Since this file already existed, have you thought about it being used already ? 
And apart from relocating the 'common', what else would need to be done ?

Typically when refactoring/renaming we should look at all of the instances where the renamed 'thing' is used. In rust type safety would be guaranteed, but in typescript, unfortunately, it's not entirely so.

fergie-nz: Good point thanks @andreievg 

I'm just thinking now a possible missed use case is duplicate keys where reports.json was previously used.

I will check now

fergie-nz: Possibly complicated by mechanisms from this PR:

https://github.com/msupply-foundation/open-msupply/pull/4617

andreievg (changes requested):
@fergie-nz without removal of reports namespace completely (including useTranslation hooks), i am getting: 

Screenshot 2024-09-10 at 1 08 51 PM

andreievg (approved):
Thanks @fergie-nz, appreciate your hard work on all of the translation stuff !

PR #4760, opened 2024-09-09
4711 select vaccination batch

### LINKED ISSUE #4711: Select batch to administer
When administering vaccination, clicking the batch (Lot) button:

Screenshot 2024-08-30 at 11 56 42 AM

opens a new modal, where I can select the batch to issue from:
Screenshot 2024-08-30 at 11 56 27 AM

lache-melvin:
Fixes #4711

# 👩🏻‍💻 What does this PR do?

Adds batch selector to dose administration modal:

Screenshot 2024-09-09 at 6 14 32 PM

I'm pretty sure this will all work correctly, as doses are administered, that there would be decimal remaining packs/units. Need to check once the prescriptions are being created though

# 🧪 Testing

- [ ] From the encounter -> vaccination tab
- [ ] Open the administer dose modal
- [ ] Click `given`
- [ ] Select an item
- [ ] Batch selection should now become enabled - click it to open the modal
- [ ] Select a batch
- [ ] Save the vaccination
- [ ] Check the stock line is saved in the DB

# 📃 Documentation

  1.
  2.


client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -88,7 +103,17 @@ export const VaccinationModal = ({
       width={550}
       slideAnimation={false}
     >
-      {modalContent}
+      <>
+        {batchModalOpen && (
```
CarlosNZ, line 107: I don't love the Modal over a Modal thing, but maybe can't be helped? Is it possible to show the batch list just as a Drop-down in the main modal, or is there too much info that needs to be squeezed in?

I think it's possible for a Drop-down item to be its own mini-components, so we could have the batch info formatted nicely in a little "card" within the drop-down selector?

CarlosNZ (commented):
Looking good. See what you think about the Drop-down suggestion. Happy to leave that to a later PR, but be good to get your thoughts regarding modal vs dropdown.

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
lache-melvin: As discussed in Telegram - modal in modal was what was originally designed, but does digress from existing patterns... have gone back to product, I think would be easier to make it a dropdown if we can cull a column or two!

Will refactor in another PR if/when decision is made :)

Are you happy with these changes otherwise @CarlosNZ

CarlosNZ: Sounds good, I'm happy to have a play with Drop-down options if/when we decide to do that.

CarlosNZ (approved):
LGTM :)

PR #4777, opened 2024-09-12
4665 vax card api

### LINKED ISSUE #4665: Show the status of each vaccination
Join vaccine_course_schedules rows for the vaccine card on vaccinations to show vaccination info where the vaccine has been recorded as given/not given. 

For the next dose of each vaccine course: if current date is beyond suggested date, show `overdue` status, otherwise show `pending` status.

lache-melvin:
Fixes #4665

# 👩🏻‍💻 What does this PR do?

Adds backend query for vaccine card, and wires up to the table in the frontend

# 🧪 Testing

- [ ] View patient vaccine card
- [ ] If you created some vaccinations, they should populate in here

# 📃 Documentation

  1.
  2.


server/graphql/types/src/types/mod.rs
```diff
@@ -101,9 +101,6 @@ pub use self::vaccine_course_item::*;
 pub mod vaccine_course_dose;
 pub use self::vaccine_course_dose::*;
 
-pub mod vaccination;
-pub use self::vaccination::*;
```
lache-melvin, line 105: realised the rest of the program related types were in the programs folder

server/repository/src/db_diesel/vaccination_card.rs
```diff
@@ -0,0 +1,86 @@
...
+    async fn test_vaccination_card() {
+        let (_, connection, _, _) =
+            test_db::setup_all("test_vaccination_card", MockDataInserts::all()).await;
+
+        let result = VaccinationCardRepository::new(&connection)
+            .query_by_enrolment_id(mock_immunisation_program_enrolment_a().id)
+            .expect("Failed to query vaccination card by enrolment id");
+
+        // 3 doses in the 1 course of the program
+        assert_eq!(result.len(), 3);
```
lache-melvin, line 84: Wanted some kind of test on the view... but maybe this is too brittle?

server/service/src/vaccination/insert/validate.rs
```diff
@@ -66,14 +68,6 @@ pub fn validate(
 
             let stock_line = check_stock_line_exists(connection, store_id, stock_line_id)?;
 
-            // This shouldn't be possible (mSupply ensures doses is at least 1 for vaccine items)
```
lache-melvin, line 69: Just moved this down, to return the known errors first

server/service/src/vaccination/insert/validate.rs
```diff
@@ -42,6 +42,8 @@ pub fn validate(
         return Err(InsertVaccinationError::VaccinationAlreadyExistsForDose);
     }
 
+    // TODO: check is the next dose! (can't give a dose if the previous one hasn't been given)
```
lache-melvin, line 45: will do in a separate issue!

server/graphql/types/src/types/program/vaccination_card.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+    pub async fn min_age_months(&self) -> &f64 {
+        &self.item.row.min_age
+    }
+    pub async fn min_interval_days(&self) -> &i32 {
+        &self.item.row.min_interval_days
+    }
+    pub async fn given(&self) -> &Option<bool> {
+        &self.item.row.given
+    }
```
CarlosNZ, line 74: Now I think about it, do we actually need the "given" field at all? It's implied by the `vaccination_date`, isn't it? I'd be inclined to remove it -- even if we want to return that field to front-end, we should just derive it rather than store it separately.

client/packages/system/src/Patient/api/hooks/usePatientVaccineCard.ts
```diff
@@ -1,4 +1,6 @@
+import { useQuery } from '@openmsupply-client/common';
 import { usePatientGraphQL } from '../usePatientGraphQL';
+import { VACCINATION_CARD } from './keys';
 
 export const usePatientVaccineCard = (
   patientId: string,
```
CarlosNZ, line 6: Ah, I see we only need the `programEnrolmentId` -- so we can get rid `patientId` from the input parameters.

(I can do this in my PR if you like?)

server/repository/src/db_diesel/vaccination_card.rs
CarlosNZ: I'd be inclined to set it up so that the mock patient has some vaccinations done and some not, and check that this is correctly reflected in the view. But not a huge priority -- the view itself looks fine by inspection.

server/repository/src/migrations/views/mod.rs
```diff
@@ -384,6 +385,27 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
       CREATE VIEW requisitions_in_period AS
           SELECT 'n/a' as id, program_id, period_id, store_id, order_type, type, count(*) as count FROM requisition WHERE order_type IS NOT NULL
             GROUP BY 1,2,3,4,5,6;   
+
+  CREATE VIEW vaccination_card AS
+    SELECT 
+      vcd.id || pe.id AS id,
```
CarlosNZ, line 391:
```suggestion
      vcd.id || "_" ||  pe.id AS id,
```
(For purely cosmetic reasons ;) )

server/repository/src/migrations/views/mod.rs
```diff
@@ -384,6 +385,27 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
...
+      vcd.min_interval_days, 
+      vcd.min_age, 
+      vc.id as vaccine_course_id, 
+      v.id as vaccination_id, 
+      v.vaccination_date, 
+      v.given, 
+      v.stock_line_id, 
+      pe.id as program_enrolment_id
+    FROM vaccine_course_dose vcd 
+    INNER JOIN vaccine_course vc
```
CarlosNZ, line 403: Specifying `INNER` not strictly necessary (you didn't for the `JOIN program_enrolment` below)

CarlosNZ (changes requested):
Nice,  thanks for getting this set up.

See comments --  I realise getting rid of the "given" field might be a bit of a headache to re-factor. Perhaps we should keep it, but make it a [generated column ](https://www.sqlite.org/gencol.html) ([Postgres version](https://www.postgresql.org/docs/current/ddl-generated-columns.html))-- then we can continue to use it but don't need to explicitly update it and ensure it's consistent.

server/repository/src/db_diesel/vaccination_card.rs
lache-melvin: This is actually done in the service layer, maybe that is sufficient!

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -10,66 +10,71 @@ import {
   DataTable,
   useColumns,
   NothingHere,
-  LocaleKey,
+  useIntlUtils,
```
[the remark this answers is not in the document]
lache-melvin: Oof, sorry... will merge this ASAP

server/graphql/types/src/types/program/vaccination_card.rs
lache-melvin: No its slightly different: vaccination_date is set on all vaccination "events", where you can record whether it was administered or refused (given true/false). Given would be undefined/None before the event 👍

client/packages/system/src/Patient/api/hooks/usePatientVaccineCard.ts
lache-melvin: Ah yep missed that one - eh its a quick change, I'll chuck it in with the other requested changes now

server/graphql/types/src/types/program/vaccination_card.rs
CarlosNZ: Okay, and that's different to the encounter date that it's linked to?

lache-melvin: Yep 🫠 It probably will be mostly the same, but the clinician might enter historic vaccinations during an encounter

CarlosNZ: Okay, but would they really ever enter historical non-vaccinations (i.e. a date but not "given")?

lache-melvin: 🤷‍♀️  Maybe catching up on a backlog, if they we're entering at the time? Probably pretty unlikely though 👍 But regardless, the given true/false on a `vaccination` serves a purpose for non-historical vaccinations, and I don't think the given field should mean different things based on the date

CarlosNZ (approved):
Looking good thanks. You merge this, then I'll take care of getting my PRs consistent with it. 

Cheers.

PR #4778, opened 2024-09-12
4663(2) Patient vax card (formatting)

### LINKED ISSUE #4663: Vaccination card (read only) UI
List all doses for all vaccine courses in the immunisation program. Order by age range, then by suggested date:

Suggested date should be the date patient reaches the age range milestone, or the date once min interval since last dose has past (whichever is later). Suggested date should only be provided for the next dose of each vaccine course. This should be a value calculated on backend and just displayed in UI.

(ignore the progress bar for now)
Screenshot 2024-08-22 at 5 32 06 PM

(Joining on vaccinations to show the given doses can come later)

CarlosNZ:
Fixes #4663 (a bit more)

# 👩🏻‍💻 What does this PR do?

Mostly just basic formatting/styling here:

- Age column text
- conditional row styling (if done)
- Localised date fields
- "Status" badge/chip thingy

Screenshot 2024-09-12 at 12 36 14 PM

Minor thing to tweak -- the alternating row colours is still being applied (pale grey every second row). Would be good to disable this.

The Status cell is based off the designs. I notice there is already this "badge" type cell used in Cold chain:

Screenshot 2024-09-12 at 12 36 31 PM

Would be good to consolidate these into a single general component for UI consistency. Will make an issue to do so once this is approved.

See inline comments for more detail.


client/packages/common/src/styles/theme.ts
```diff
@@ -193,6 +194,7 @@ export const themeOptions = {
       menu: '#f2f2f5',
       toolbar: '#fafafc',
       white: '#fff',
+      success: 'rgb(237, 247, 237)',
```
CarlosNZ, line 197: This the pale green used in the Alert backgrounds. I couldn't find anywhere that it's defined, it's just a Material defaul -- the general `success.main` colour defined is just the darker icon colour. So I've just copied it from the CSS here for re-use later. If anyone knows how to access the default Material colour definitions directly, that'd be cool.

client/packages/common/src/ui/layout/tables/components/Cells/StatusCell/StatusCell.tsx
```diff
@@ -0,0 +1,61 @@
...
+        />
+      )}
+      <CircleIcon
+        sx={{
+          color,
+          transform: 'scale(0.4)',
+        }}
+      />
+      <Typography sx={{ paddingRight: 1 }}>{label}</Typography>
+    </Box>
```
CarlosNZ, line 59: This is a combination of the Cold chain "Status" cell (the "chip") and the "DotCell" (the dot icon). You have the option to specify a background colour explicitly, but by default it will use a faded-out version of the main colour. I did this so we could just use existing palette colours and not have to define a bunch of new background colours. But perhaps made the logic a little over-complicated: when bgColor is defined, we can apply it to the main container, but if not, we have to apply it to a separate, absolutely positioned block, otherwise the entire component gets the low-opacity effect.

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -12,30 +11,63 @@ import {
...
+  useEffect(() => {
+    if (!rows) return;
+
+    setRowStyles(
+      rows.filter(row => !!row.dateGiven).map(row => row.id),
+      {
+        backgroundColor: `${theme.palette.background.success} !important`,
+      }
+    );
+  }, [rows]);
```
CarlosNZ, line 36: I guess I could set Row Styles for the remaining rows to explicitly set their background to white, which would solve the alternating colour problem described above? (Unless there's a better way?)

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -12,30 +11,63 @@ import {
...
+
+  useEffect(() => {
+    if (data) {
+      setCustomBreadcrumbs({
+        1: data?.vaccineCardItems?.patient?.name ?? '',
+        2: t('label.vaccination-card'),
+        3: data?.vaccineCardItems?.programName,
+      });
+    }
+  }, [data]);
```
CarlosNZ, line 62: Moved, not changed.

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -12,30 +11,63 @@ import {
...
   const { setCustomBreadcrumbs } = useBreadcrumbs();
   const { localisedDate } = useFormatDateTime();
+  const theme = useTheme();
 
   const {
     query: { data, isLoading },
   } = usePatientVaccineCard(patientId, programEnrolmentId);
 
-  const tableData = buildTableData(data);
+  const tableData = useMemo(() => buildTableData(data), [data]);
```
CarlosNZ, line 50: Need to memo-ise this to avoid infinite re-render. But we maybe able to remove this after the back-end is wired up.

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -45,32 +77,35 @@ export const VaccinationCardComponent: FC = () => {
...
+          statusMap={{
+            'label.status-given': {
+              color: theme.palette.success.light,
+              label: t('label.status-given'),
+            },
+            'label.status-not-given': {
+              color: theme.palette.error.main,
+              label: t('label.status-not-given'),
+            },
+          }}
```
CarlosNZ, line 93: Presumably the status values will be back-end enums?

CarlosNZ:
> Would be good to consolidate these into a single general component for UI consistency. Will make an issue to do so once this is approved.

Made the issue: https://github.com/msupply-foundation/open-msupply/issues/4779

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
lache-melvin: Yeah great point - agree it probably should be! 

Especially once we implement the age ranges - then we should implement a `Late` status, which comes into effect once the suggested date has passed, and the patient is beyond the end of the age range 👍

client/packages/common/src/ui/layout/tables/components/Cells/StatusCell/StatusCell.tsx
```diff
@@ -0,0 +1,61 @@
...
+        borderRadius: 4,
+        textAlign: 'center',
+        backgroundColor: bgColor,
+      }}
+      paddingY={0.1}
+      paddingX={0.5}
+      display="flex"
+      alignItems="center"
+      width="fit-content"
+      minWidth={100}
```
lache-melvin, line 35: Nit, but I'm not sure the min width is needed, I think it would look nicer just fitting the contents?

Screenshot 2024-09-12 at 3 49 20 PM

client/packages/common/src/ui/layout/tables/components/Cells/StatusCell/StatusCell.tsx
lache-melvin: It might be simpler to always render/user the separate Box, with styles something like

```ts
          backgroundColor: bgColor ?? color,
          opacity: bgColor ? 1 : 0.2,
```

rather than setting the background colour in two different places?

lache-melvin (approved):
Looks good to me Carl! I think could simplify that bgColor logic a little, but happy for you to merge when ready 🙏

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
lache-melvin: Can also be in the next PR, but would be good to refactor the table to a separate component from the detail view, so we can reuse in the encounter, without the breadcrumb bits etc 🚀

client/packages/common/src/ui/layout/tables/components/Cells/StatusCell/StatusCell.tsx
CarlosNZ: Yup, good idea: 7ed8975781

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
CarlosNZ: Done, but I had to add an additional flag to the `setRowStyles` method to not reset the row styles, because when you do two sequential `setRowStyles()`, the styles get reset in between and only the last one sticks.

7ed8975781

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
CarlosNZ: For now I've just used the `given` value (true/false) as the status map key. Can be easily changed to enum values at some point.

7ed8975781

CarlosNZ:
@lache-melvin Suggestions implemented, going to merge down now and we can have another look at the full thing if we like.

PR #4783, opened 2024-09-12
4774 backlog prescription entry backdated stock levels service

jmbrunskill:
Fixes #4474

# 👩🏻‍💻 What does this PR do?

New function `get_historical_stock_lines` and test case

- I'm using the stock movement view, does that miss stock that's just allocated rather than actually picked/sent etc?
- Does inventory adjustments or stock takes mean we need to account for something differently?
- Are there any test cases/scenarios I've missed?
- Is it safe to just adjust available_packs as I'm doing? Do we need to do something with total packs too?
- Is the overall approach/math valid? What am I missing?

I think there is a potential for a temporary ledger issue with this approach.
We need to take the minimum stock level as the available amount (the latest stock level might not be the minimum)
Will update this in another PR...

# 🧪 Testing

- [ ]  Test cases are in the rust code. Please suggest any other scenarios to try...

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/stock_line/tests/historical_stock.rs
+++ b/server/service/src/stock_line/tests/historical_stock.rs
@@ -186,12 +186,12 @@ mod query {
             .unwrap();
 
         // Check there's no stock to start with, if there is some mocks might have slipped through?
-        // let result = service_provider
-        //     .stock_line_service
-        //     .get_historical_stock_lines(&ctx, store_id.clone(), item_id.clone(), date_now().into())
-        //     .unwrap();
+        let result = service_provider
+            .stock_line_service
+            .get_historical_stock_lines(&ctx, store_id.clone(), item_id.clone(), date_now().into())
+            .unwrap();
 
-        // assert!(result.rows.is_empty());
+        assert!(result.rows.is_empty());
 
         // Here's our test scenario
         // 3 stock lines, A, B, &C
@@ -277,12 +277,12 @@ mod query {
         adjust_test_stock(&ctx, stock_movements);
 
         // // Check we can see 2 stock lines now (stock line A is fully consumed)
-        // let result = service_provider
-        //     .stock_line_service
-        //     .get_historical_stock_lines(&ctx, store_id.clone(), item_id.clone(), date_now().into())
-        //     .unwrap();
+        let result = service_provider
+            .stock_line_service
+            .get_historical_stock_lines(&ctx, store_id.clone(), item_id.clone(), date_now().into())
+            .unwrap();
 
-        // assert_eq!(result.rows.len(), 2);
+        assert_eq!(result.rows.len(), 2);
 
         // +++ 2020-01-01
         let result = service_provider
--- a/server/service/src/stock_line/tests/historical_stock.rs
+++ b/server/service/src/stock_line/tests/historical_stock.rs
@@ -22,6 +22,20 @@ mod query {
     static STOCK_LINE_B: &str = "stock_line_b";
     static STOCK_LINE_C: &str = "stock_line_c";
 
+    fn get_midnight(year: i32, month: u32, day: u32) -> NaiveDateTime {
+        NaiveDate::from_ymd_opt(year, month, day)
+            .unwrap()
+            .and_hms_opt(0, 0, 0)
+            .unwrap()
+    }
+
+    fn get_midday(year: i32, month: u32, day: u32) -> NaiveDateTime {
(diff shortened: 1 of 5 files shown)


server/service/src/stock_line/tests/historical_stock.rs
```diff
@@ -0,0 +1,479 @@
...
+            mock_name_customer_a().id // Customer
+        };
+
+        let old_stock_line = StockLineRowRepository::new(&ctx.connection)
+            .find_one_by_id(&stock_line_id)
+            .unwrap()
+            .unwrap_or_default();
+        let new_number_of_packs = old_stock_line.total_number_of_packs + number_of_packs;
+
+        let invoice_number = next_invoice_number();
```
andreievg, line 76: Btw for tests I wouldn't bother too much with invoice nubmers or other selecting correct supplier/customer (unless we using service), I think that woudl reduce the helper methods for tests, also would try to comment inline (just the SOH), so the data definition for test would be something like this:

```rust
        let stock_movements = vec![
            (100, StockIn, "A", "2021-01-01"),  // A, SOH = 100,
            (50, StockOut, "A", "2021-01-02"),  // A, SOH = 50
            (50, StockOut, "A", "2021-01-03"),  // A, SOH = 0
            (100, StockIn, "A", "2021-01-04"),  // A, SOH = 100
            (100, StockOut, "A", "2021-01-05"), // A, SOH = 0
        ];
```

Not a change request, just a thought for balancing out test helpers code vs clarity of data definition

andreievg:
@jmbrunskill 

I just realised something the stock_movement view looks at adjustments affection 'total number of pack': 

https://github.com/msupply-foundation/open-msupply/blob/aabe9159425a49fe1ee8061ad846406b8f7388c7/server/repository/src/migrations/views/mod.rs#L131-L140

Typically when we 'issue' stock we would consider available number of packs vs total number of packs. I think the use case would be, when there is some stock reserved for historic invoice that is not `picked`, it would not be considered in historic available stock. Heads up for a use case to verify, we probably need two fields on stock_movement, 'available_stock_movement_datetime' and 'total_stock_movement_datetime'.

andreievg (approved):
Looks great thanks, please consider the comment made, and extract anything relative to a new issue.

One other comment, is about 'get_historic_stock_lines' returning a stock line vs get historic stock line stock level. (or something like that), I wonder if we should just return the latter, since to be accurate the former would also need to have the correct stock line info (expiry data, batch etc.. all of those could have change from current stock line)

jmbrunskill:
> One other comment, is about 'get_historic_stock_lines' returning a stock line vs get historic stock line stock level. (or something like that), I wonder if we should just return the latter, since to be accurate the former would also need to have the correct stock line info (expiry data, batch etc.. all of those could have change from current stock line)

Good point, I think we're probably ok though in practice as things like batch number & expiry shouldn't change for a stock line unless it was miss entered right? If a new batch comes in with different expiry it would go in as a new line right?

jmbrunskill:
> Typically when we 'issue' stock we would consider available number of packs vs total number of packs. I think the use case would be, when there is some stock reserved for historic invoice that is not `picked`, it would not be considered in historic available stock. Heads up for a use case to verify, we probably need two fields on stock_movement, 'available_stock_movement_datetime' and 'total_stock_movement_datetime'.

I'm not sure if I understand the logic correctly.

Here's what I understand.

`stockLine` has `total_number_of_packs` and `available_number_of_packs`

When stock is allocated to an invoice, the `available_number_of_packs` for the stock line is reduced.
After it is picked the `total_number_of_packs` get reduced.

If I understand correctly `stock_movements` includes all the picked movements, if something is still just allocated, it's stock will be accounted for in the current StockLines, which is the base for my calculation?
Since we're taking the minimum (or will be soon) I don't think it needs to be accounted for again in movements?

Does that makes sense or am I missing something?

server/service/src/stock_line/tests/historical_stock.rs
jmbrunskill: This is a good idea, would help with keeping comments in sync with code too...

andreievg:
> if I understand correctly stock_movements includes all the picked movements, if something is still just allocated, it's stock will be accounted for in the current StockLines, which is the base for my calculation?
Since we're taking the minimum (or will be soon) I don't think it needs to be accounted for again in movements?

> Does that makes sense or am I missing something?

What you've described makes sense, I did quick calculation and all seems well, we wouldn't be doing available number of packs ledger check (as that's not a use case for us).

PR #4787, opened 2024-09-13
Fix reports hook

mark-prins:
Fixes #4783

# 👩🏻‍💻 What does this PR do?
Move the hook call to be within a hook

Unsure why no one else has complained about this?
It happens for me on every report generation.

# 🧪 Testing

- [ ] Try to view the item usage report

# 📃 Documentation

  1.
  2.


andreievg (approved):
Thanks @mark-prins

> Unsure why no one else has complained about this? 

This is only recent on develop branch. 

Sorry I approved without testing

PR #4788, opened 2024-09-13
Backdating Prescriptions - Save Date

### LINKED ISSUE #3672: Backlog prescription entry
This request has been raised by PLM in Sierra Leone. In some hospitals and other facilities, prescriptions are entered as a backlog, on a day different to when they were actually dispensed, either because staff are not available or because the number of people coming to the dispensary is too large to keep up. This can mess up reports and they get more messed up the bigger the time difference between when the prescription was dispensed and when it ws entered into the system. 

To be able to choose the confirmation date to be some date in the past, maybe with limits. But making sure that it doesn't cause ledger issues by using stock that wasn't in stock on the confirmation date - so some sort of checking required on confirmation date selection and on line added to prescription (or on OK button).

Only the above.

### Additional context 💌

PLM will pay for this work if it's a reasonable cost.

### Agreed Solution 💌

- Only allow stock to be issued that has arrived before the 'CI'  (outbound :-)  )  confirmed date.  

### Moneyworks Jobcode 🧰

 OMS:PLMB

jmbrunskill:
Fixes #3672 (Part 1)

# 👩🏻‍💻 What does this PR do?

Adds a new date selection to prescriptions, and saves the selected date in the allocatedDatetime (currently unused for prescriptions.

UI is very basic right now.
image

Should we have some kind of "Back Date Prescription" button or something?
Should we style this differently?
Do we need some kind of store pref to enable this???

Validations are coming in the next PR (Keen to see https://github.com/msupply-foundation/open-msupply/pull/4783 approved first)

# 🧪 Testing

- [ ] Test you can change the date on prescriptions

# 📃 Documentation

  1.
  2.


server/graphql/invoice/src/mutations/prescription/update.rs
```diff
@@ -20,6 +21,7 @@ pub struct UpdateInput {
     pub status: Option<UpdatePrescriptionStatusInput>,
     pub patient_id: Option<String>,
     pub clinician_id: Option<String>,
+    pub prescription_date: Option<NaiveDate>,
```
andreievg, line 24: This should be, Option<DateTime<Utc>>, similar to filter inputs (otherwise we are sending timezoned datetime, without timezone, that will be consider as UTC in back end):

https://github.com/msupply-foundation/open-msupply/blob/aabe9159425a49fe1ee8061ad846406b8f7388c7/server/graphql/core/src/generic_filters.rs#L71-L75

Also I don't think we need `naiveDateTime` in front end utils.

I think this would need some internal wiki description at some point, i've made a task for self on trello, as a placeholder, will turn into issue or wiki TODo

andreievg (changes requested):
Thanks, I haven't had a thorough look noticed graphql input and requested changes, will look closer tomorrow

server/graphql/invoice/src/mutations/prescription/update.rs
jmbrunskill: This was going to be date, with no time (assumed midnight) but timezone are an issue good point.
I was copying another pattern that did you naive date, I think because there wasn't a time component at all.

Definitely can remove `naiveDateTime` in front end utils, I put that in earlier then didn't end up using it and forgot to remove thanks for picking this up!

jmbrunskill: [1ca5619](https://github.com/msupply-foundation/open-msupply/pull/4788/commits/1ca56196bc1ddf1d5aaf0b726445053ec4c03c4a)

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -85,7 +92,11 @@ fn should_update_batches_total_number_of_packs(
     }
 }
 
-fn set_new_status_datetime(invoice: &mut InvoiceRow, status: &Option<UpdatePrescriptionStatus>) {
+fn set_new_status_datetime(
+    invoice: &mut InvoiceRow,
+    status: &Option<UpdatePrescriptionStatus>,
+    prescription_datetime: Option<NaiveDateTime>,
```
andreievg, line 98: This would mean prescription datetime would need to be present in the API call when status is changed ? How would this work ?

andreievg (approved):
Looks great thanks !

server/service/src/invoice/prescription/update/generate.rs
andreievg: It's been clarified that prescription_date will default = existing_invoice.allocated_datetime. Thus prescription picked and verified datetime will always = allocated_datetime, which I think works nicely, we just have to document that at some point

PR #4802, opened 2024-09-16
#4797 Add Vaccine card to Encounter

### LINKED ISSUE #4797: Add vaccination card to encounter

Vaccination card (read only mode) exists from the patient view. I want to access the edit view from an encounter.

Add the vaccination card table to the encounter view, opening the modal on row click

CarlosNZ:
Fixes #4797

# 👩🏻‍💻 What does this PR do?

- Vaccine card appears under Encounter as well as Patient
- Clicking on Vax card rows opens the "Vaccination" modal
  - Conditions: Under Patient, only rows that have vaccination records can be opened, and none can be edited (have added `disabled` to modal in this case)
- Adding vaccination (via modal) causes the Vax card to refresh to reflect the new Vaccination status
- Added "programEnrolmentId" to the Encounter query, so it can be passed to Vax card (it can only be accessed from URL when in Patient)
- Refactored the "VaccineCardTable" out of the DetailView component for easier sharing between Encounter and Patient
- DetailView (Patient) shows the whole "VaccinationTab" component, which includes the vaccination modal
- Hide "Encounter" tab when there are no "Encounter" items (this is a bit hacky, we can probably improve this)

See inline comments for further detail

# 🧪 Testing

In "Patient" vax card page:
- [ ] Can click on rows that have a vaccination, which opens modal
- [ ] Modal should have all inputs disabled
- [ ] Can't click on rows that have no vaccination

In "Encounter" vax card page
- [ ] Can click on all rows
- [ ] Can administer additional vaccines via modal
- [ ] Administering vaccine causes table to refresh to show new Vax status


client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -257,6 +257,24 @@ export const DetailView: FC = () => {
...
+        <VaccinationsTab
+          encounterId={encounter.id}
+          programEnrolmentId={encounter.programEnrolment.id}
+          clinician={encounter.clinician ?? undefined}
+        />
+      ),
+      value: t('label.vaccinations'),
+    });
+    if (encounter.document.documentRegistry?.uiSchema.elements.length > 0)
+      tabs.push({ Component: JsonForm, value: t('label.encounter') });
```
CarlosNZ, line 275: So this is the only way I could figure out how to tell if there are "boxes" to appear on the Encounter tab. I'm assuming that, initially, Vaccine program encounters will have no UI components defined?

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { FC, useEffect } from 'react';
```
CarlosNZ, line 1: Mostly no changes in here, just copied from previous DetailView component. I'll put a comment on the bits that have changed.

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <DataTable
+        id={'Vaccine Course List'}
+        columns={columns}
+        data={data?.items ?? []}
+        isLoading={isLoading}
+        onRowClick={row =>
+          (encounterId ?? row.vaccinationId)
+            ? openModal(row.vaccinationId, row.vaccineCourseDoseId)
+            : undefined
+        }
```
CarlosNZ, line 126: One thing I don't like about this -- even though you can't open the Patient Vax card non-vaccination rows, you still get the "clickable" icon when hovering over the rows. The only way to prevent this is to not have return a function here for those rows, but the problem is we need access to the *row* data in order to determine the state, which we can only do via a function. 🤷‍♂️

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
```diff
@@ -0,0 +1,142 @@
...
+    <InlineSpinner />
+  ) : (
+    <>
+      <DataTable
+        id={'Vaccine Course List'}
+        columns={columns}
+        data={data?.items ?? []}
+        isLoading={isLoading}
+        onRowClick={row =>
+          (encounterId ?? row.vaccinationId)
```
CarlosNZ, line 123: `encounterId` is only passed to the "Encounter" version of the card, and I'm using it as the condition to determine row-clicking behaviour.

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
```diff
@@ -1,63 +1,24 @@
...
   const t = useTranslation('dispensary');
   const { programEnrolmentId = '' } = useParams();
   const { setCustomBreadcrumbs } = useBreadcrumbs();
-  const { localisedDate } = useFormatDateTime();
   const { getLocalisedFullName } = useIntlUtils();
-  const theme = useTheme();
 
   const {
     query: { data, isLoading },
   } = usePatientVaccineCard(programEnrolmentId);
```
CarlosNZ, line 20: I don't like calling this hook in here as well as in the child "VaccineCardTable" component, but couldn't see any other straightforward way to access the patient data for the breadcrumbs. Presumably the caching will prevent duplicate network request?

client/packages/system/src/Patient/api/hooks/keys.ts
```diff
@@ -1 +1 @@
-export const VACCINATION_CARD = 'vaccination-card';
+export const VACCINATION = 'vaccination';
```
CarlosNZ, line 1: So this key needs be the same as the one in `client/packages/system/src/Vaccination/api/keys.ts` in order to cause the table refresh after vaccination.

Should we refer back to that one so that the Vax Card hook is referencing the same constant? Not sure what pattern we've established when two hooks that share a key are distantly separated in the file structure.

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -111,7 +111,7 @@ export function useVaccination({
     isDirty: Object.keys(patch).length > 0,
     updateDraft: (update: Partial<VaccinationDraft>) =>
       setPatch({ ...patch, ...update }),
-    create: insert,
+    create: encounterId ? insert : () => {},
```
CarlosNZ, line 114: Bit ugly, but need to make sure the mutation can't be called when there is no `encounterId` (and can't conditionally call the `useInsert` hook, so passing a dummy value to it.)

client/packages/system/src/Patient/api/hooks/keys.ts
lache-melvin: I think better would be to keep this as `VACCINATION_CARD`, and add `VACCINATION_CARD` to the invalidate queries call after the vaccination insert mutation?

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
lache-melvin: Oof yeah, that's frustrating... maybe we can do something in the setRowStyles, for rows that don't have a vaccinationId?

```css
        '& td': {
          cursor: 'default',
        },
        ```

File: client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx (hunk 4 above)
Comment (lache-melvin) [reply client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx]: Maybe having something like `const canOpenCreateModal = !!encounterId`, and using that here instead, would be a bit more descriptive for someone coming back to this later... or a comment?

File: client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx [hunk 8]
```diff
@@ -1,63 +1,24 @@
...
-const useStyleRowsByStatus = (rows?: VaccinationCardItemFragment[]) => {
-  const { setRowStyles } = useRowStyle();
-  const theme = useTheme();
-
-  useEffect(() => {
-    if (!rows) return;
-
-    const doneRows = rows.filter(row => row.given).map(row => row.id);
-    const notDoneRows = rows.filter(row => !row.given).map(row => row.id);
+import { VaccinationsTab } from 'packages/system/src/Encounter/DetailView/VaccinationsTab';
```
lache-melvin, line 10: Needs the @openmsupply-client/system import

client/packages/system/src/Encounter/DetailView/VaccinationsTab.tsx
```diff
@@ -2,36 +2,51 @@ import React from 'react';
 import { VaccinationModal } from '../../Vaccination';
 import { useEditModal } from '@common/hooks';
 import { Clinician } from '../../Clinician';
+import { VaccineCardTable } from '../../Patient/VaccinationCards/VaccineCardTable';
 
 export const VaccinationsTab = ({
```
lache-melvin, line 7: I think you mentioned this already - we should probably move this, and put it with VaccineCardTable component. The "Tab" name doesn't make much sense, it's just making rendering the card with the modal (maybe we call this component VaccineCard?)

If it was doing anything specific here to be a tab in the patient view, this separate component would make sense, but I don't think it's adding any value on that front anymore :)

client/packages/system/src/Vaccination/api/useVaccination.ts
lache-melvin: What if we let useInsert accept an optional encounterId, and do an early return from the mutationFn there? Same deal but feels a little more type safe - if the id is there, we know it's actually populated

client/packages/system/src/Patient/VaccinationCards/DetailView/DetailView.tsx
lache-melvin: Ah yeah, oh well. Shooould get the caching? But not too worried, should only need to run once here aynway. And I suppose the same as running a separate query just for this data 🤷‍♀️

lache-melvin (changes requested):
Looks great Carl - just few minor things, let me know your thoughts 🙏

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+    {
+      key: 'dateGiven',
+      label: 'label.date-given',
+      accessor: ({ rowData }) => localisedDate(rowData.vaccinationDate ?? ''),
+    },
+  ]);
+
+  const isPreviousDoseGiven = (row: VaccinationCardItemFragment) => {
+    if (!data?.items) return false;
+    const doseIndex = data?.items.findIndex(dose => dose.id === row.id);
```
lache-melvin, line 115: Ohh, actually, this first needs to filter by vaccineCourseId - this only works for the first row in the card 😁 

The vaccine card rows have vaccine_course_id on them, just needs to be exposed to frontend!

CarlosNZ: Ha, good catch! b41fe97cef

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
CarlosNZ: Yeah, good call. The logic is a bit gnarly but it works 👍

dbe2c1214c

client/packages/system/src/Patient/VaccinationCards/VaccineCardTable.tsx
CarlosNZ: I just called it `isEncounter`, as the logic to determine whether the modal can open is a bit more complicated.

client/packages/system/src/Vaccination/api/useVaccination.ts
CarlosNZ: 5e61809fd8

client/packages/system/src/Encounter/DetailView/VaccinationsTab.tsx
CarlosNZ: Yeah, I think they all could do with a bit of re-organisation. Will quickly look at it with you at tomorrow morn's stand-up.

lache-melvin (approved):
Looks and works great! Thanks for changes Carl 🔥

PR #4806, opened 2024-09-17
Plugins AMC example

andreievg:
This PR serves a few purposes:
* Keep team updated on plugin prototype work
* Creates some context for technical discussions
* Historic reference for progression of plugin technical design

** Plugin Provider **

Uses a static/const global RwLock (multiple simultaneous readers, ones writer), makes it more convenient to use plugin implementation without passing provider everywhere. Sometimes using global variables can create more problems then they solve, we need to be aware of that (I think in this case it should be fine).

I was trying to keep the impact of overwriting, quite happy with it. 

This is a limited functionality prototype, I have a pretty big list of questions and improvements, that would be summarised in an issue and prioritised, so if you are looking at this PR I would suggest waiting for that list before going into a deep dive of thoughtful improvements/suggestions. (still in exploratory stage)

Next up is another example, of overwriting requested quantity in requisitions with API call (macro eye use case)

** How it works **

There is global plugin provider which stores dynamically dispatched trait implementations, those can be overwritten at runtime, 'plugin' endpoint allows you to upload one plugin for AMC calculation.

AMC calculation plugin implements a trait of filter/store_id input and output of HashMap of itemIds and AMC data, this plugin is called if it exists in plugin provider, or default implementation, when item stats are calculated.

Plugin is bound using extism wasm plugin framework, at the time of creating plugin instance, service provider reference is captured in a host function that is passed on to the plugin (a host function to do sql query)

** How to test/run **

`cargo run` -> check AMC for items (catalogue -> item)
See commands in `amc.js` (in the comment), should be able to build and upload plugin -> check AMC after upload of plugin

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -443,11 +443,20 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli 0.28.1",
+]
+
+[[package]]
+name = "addr2line"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


andreievg:
Closing this prototype draft as we've started implementation into dev

PR #4809, opened 2024-09-18
4714 edit vaccination

### LINKED ISSUE #4714: Edit vaccinations
Implement edit of vaccinations

If batch or item changes, we will need to create a Inventory Addition to reverse the Prescription, and create a new Prescription.

If change from given to not given, need to create the Inventory Addition to reverse the Prescription.

If change from not given to given, create the prescription like in create vaccination

lache-melvin:
Fixes #4714

# 👩🏻‍💻 What does this PR do?

This is the backend/service layer piece, will do a separate PR to wire up with the card.

So many lines, I do apologise 😭 Though about half is just the tests!

Keen for thoughts on the system inventory adjustment reason...
- Reasons are required for inventory adjustments, if any reasons are defined in the system. Reasons are defined as options in mSupply, this inserts a custom one for system use, so we know it is always there
- This adds complexity - need to filter it out from options for when users are inputting an inventory adjustment, and translate it in the ledger (other reasons would already be defined in the country's main language)
- Other option would be to not put a reason - add logic to not require one if it's the system creating the adjustment. However, I think seeing only "patient name" and "Inventory addition" in the ledger is confusing - reversed prescription seems like helpful context!

# 🧪 Testing

- [ ] Run the tests 😁 

# 📃 Documentation

  1.
  2.


client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -19,6 +19,8 @@ export enum ColumnKey {
   'Number' = 'number',
 }
 
+const REVERSED_PRESCRIPTION = 'reverse_prescription';
```
lache-melvin, line 22: Not stoked to define this in two places... but haven't seen us define types for constants in graphql layer anywhere else?

server/service/src/stocktake_line/validate.rs
```diff
@@ -66,6 +66,7 @@ pub fn check_reason_is_valid(
                 InventoryAdjustmentReasonFilter::new()
                     .r#type(InventoryAdjustmentType::Positive.equal_to())
                     .is_active(true)
+                    .include_system_reasons(true)
```
lache-melvin, line 69: Include the system reasons as valid inventory adjustment reasons here, otherwise defaulting to filtering them out everywhere else

server/service/src/vaccination/generate.rs
```diff
@@ -0,0 +1,71 @@
...
+};
+
+#[derive(Debug)]
+pub struct CreatePrescription {
+    pub insert_prescription_input: InsertPrescription,
+    pub insert_stock_out_line_input: InsertStockOutLine,
+    pub update_prescription_input: UpdatePrescription,
+}
+
+pub fn generate_create_prescription(
```
lache-melvin, line 16: Nothing actually new in here, just pulled out the create prescription from the insert method, to share with update

server/service/src/vaccination/update/generate.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub new_stock_line: Option<StockLine>,
+}
+
+pub struct GenerateResult {
+    pub vaccination: VaccinationRow,
+    pub create_inventory_adjustment: Option<InsertInventoryAdjustment>,
+    pub create_prescription: Option<CreatePrescription>,
+}
+
+pub fn generate(
```
lache-melvin, line 28: There's quite a bit of complexity in terms of changing to/from given, or changing the stock line etc... not sure if there is a clearer way to denote the forks/where logic is shared? Let me know any areas that seem particularly unclear 😢

server/repository/src/migrations/v2_03_00/add_reverse_prescription_inventory_addition_reason.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+               INSERT INTO inventory_adjustment_reason (id, type, is_active, reason) 
+                VALUES ('{REVERSE_PRESCRIPTION_REASON_ID}', 'POSITIVE', TRUE, '{REVERSE_PRESCRIPTION_REASON_ID}');
+            "#
+        )?;
+
+        Ok(())
```
CarlosNZ, line 20: Just wondering, if this is a "system" reason that is always available, perhaps it would be better to hard-code it rather than put it in the database?

Putting it in the database makes it susceptible to user tampering. I know we can prevent it being edited, but if it wasn't here, then we wouldn't need the additional filtering/handling you've added, and we could just inject into the list of reasons (in the service layer?) in the few places it's needed?

CarlosNZ (commented):
Looks good, I'll test it more thoroughly when I try the front-end.

I have just have one reservation about the "reverse_prescription" reason. Not sure if what I'm suggesting makes things more or less complicated -- see what you think.

client/packages/invoices/src/utils.ts
```diff
@@ -121,11 +121,7 @@ export const getNextInboundStatus = (
 export const getNextCustomerReturnStatus = (
   currentStatus: InvoiceNodeStatus
 ): InvoiceNodeStatus | null => {
-  const currentStatusIdx = customerReturnStatuses.findIndex(
-    status => currentStatus === status
-  );
-  const nextStatus = customerReturnStatuses[currentStatusIdx + 1];
-  return nextStatus ?? null;
+  return getNextInboundStatus(currentStatus);
```
lache-melvin, line 124: Side quest - the next statuses couldn't be found on the customer return view and so no status change button was rendering! I think a missed thing from the inbound return -> customer return change

lache-melvin:
Ok @CarlosNZ, have replaced the inventory adjustment with a customer return. Had to add a couple of changes, so that we could update the existing stock line rather than creating new stock lines every time a vaccination was edited. Let me know what you think! Have also updated the UI PR, so you can test it there :)

server/graphql/invoice/src/mutations/customer_return/insert.rs
```diff
@@ -131,6 +131,7 @@ impl InsertInput {
                 .into_iter()
                 .map(|line| line.to_domain())
                 .collect(),
+            is_patient_return: false,
```
lache-melvin, line 134: Needed to distinguish the usual returns, from a customer store, from the new patient returns..

server/graphql/invoice/src/mutations/customer_return/insert.rs
```diff
@@ -157,6 +158,7 @@ impl CustomerReturnLineInput {
             expiry_date,
             batch,
             pack_size,
+            stock_line_id: None,
```
lache-melvin, line 161: We'll continue to create new stock lines from existing customer return functionality, but provision is now there to increase stock of existing stock lines (if you were to return from an outbound shipment)

server/service/src/invoice/customer_return/update/generate.rs
```diff
@@ -148,51 +149,29 @@ pub fn generate_lines_and_stock_lines(
...
-                supplier_link_id: Some(supplier_id.to_string()),
-                barcode_id: None,
+        if line.number_of_packs > 0.0 {
+            let existing_stock_line = if let Some(id) = &return_line.stock_line_id {
+                StockLineRowRepository::new(connection).find_one_by_id(id)?
+            } else {
+                None
             };
+
+            let stock_line = generate_batch(
```
lache-melvin, line 159: Updated this method to use the shared `generate_batch` from `stock_in_line` abstraction, as this already handles the creating new/adding to existing stock lines

server/service/src/invoice/customer_return/update/generate.rs
```diff
@@ -148,51 +149,29 @@ pub fn generate_lines_and_stock_lines(
...
+                connection,
+                line,
+                StockLineInput {
+                    stock_line_id: return_line.stock_line_id,
+                    store_id: store_id.to_string(),
+                    on_hold: existing_stock_line.map_or(false, |stock_line| stock_line.on_hold),
+                    barcode_id: None,
+                    supplier_link_id: supplier_id.to_string(),
+                    // Update existing stock levels if the stock line already exists
+                    overwrite_stock_levels: false,
```
lache-melvin, line 169: I find the `overwrite` naming confusing... but felt there was enough going on in this PR not to go trying to make any more improvements right now 😅

CarlosNZ: In what way is it confusing?

CarlosNZ (approved):
Whoa, quite a big change. Looks good thought, think it's worth it.

👍

server/service/src/invoice/customer_return/update/generate.rs
lache-melvin: Haha not sure, it just sits a bit better for me the other way... `adjust_existing_stock_levels` or something. It feels weird to call it overwriting when you are creating a new stock line 🤷‍♀️ 

The only really valid `overwrite` case is where we have an invoice that introduces stock, and then we edit those invoice lines (e.g. introducing a stock line with 6 packs instead of 5)

PR #4814, opened 2024-09-18
Plugin Suggested Quantity. via API call example

andreievg:
Following up from [this PR:](https://github.com/msupply-foundation/open-msupply/pull/4806), another prototype of adding suggested quantity plugin, that uses API call to get quantities:

* Upload plugin endpoint now accepts json config and url parameter for plugin. name
* Json config from plugin includes common configurations (which hosts are allowed for http requests) and plugin specific configs (in this case url of the server)
* Expose suggested quantity calculation as a plugin (refactor slightly, to only be called once for all item ids being processed)
* Add suggested quantity plugin that queries API for suggested quantities for item (using mock json server)


andreievg:
Closing this prototype draft as we've started implementation into dev

PR #4821, opened 2024-09-19
fix translations -  Move string to common

### LINKED ISSUE #4811: Add from master list string missing
## What went wrong? 😲

On my develop branch i seem to get the following, when adding internal order from master list (same on demo server)

Screenshot 2024-09-18 at 12 47 41 PM

Hmm how did it sneak through testing ?
Also how did we make this code change without noticing/testing and how to we avoid this in the future ?

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #4811

# 👩🏻‍💻 What does this PR do?

Fix translation namespace missing from common

I've found it snuck in [here](https://github.com/msupply-foundation/open-msupply/commit/7474f03d7c068c7b55cf1d40e57e02aeb09f8477)

No reason that I can see that other strings would be affected - but does highlight our not quite type safe locale keys as this is hard to see and trace.

Maybe an RnD project to make sure we don't encounter gaps like this in future?

# 🧪 Testing

You can check this by going to internal orders and adding from master list
# 📃 Documentation

no documentation changes required

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -548,6 +548,7 @@
   "messages.upload-invite": "Drag & drop to upload",
   "messages.upload-or": "or",
   "messages.yes": "Yes",
+  "messages.confirm-add-from-master-list": "Do you want to add all of the items from this master list?",
   "multiple": "[multiple]",
   "placeholder.filter-by-status": "Filter by status",
   "placeholder.filter-items": "Filter items",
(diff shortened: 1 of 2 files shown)


andreievg (approved):
Thank you @fergie-nz 

> Maybe an RnD project to make sure we don't encounter gaps like this in future?

Yeah would be great to make translations type safe, although there might be a move towards combining all translations to one common namespace, but we haven't quite thought about this in detail yet

PR #4824, opened 2024-09-19
#4173 Prevent dose skipping

### LINKED ISSUE #4173: Test DBs are inconsistent between production and test migrations
## What went wrong? 😲

In migrating to 2.01 we get this error (depending on if the test or non-test version of the migrations are run
```
   "UNKNOWN" ("\"\\\"table store_new has 7 columns but 8 values were supplied\\\"\""), version: Version { major: 2, minor: 1, patch: 0, pre_release: None } }
```
## Expected behaviour 🤔

Developer should reasonably expect the same state of migrations between test and prodcution builds.

## How to Reproduce 🔨

1. Delete template marker files in `server/repository/test_output`
2. Run `cargo test --package repository --lib -- migrations::v2_01_00 --show-output`
3. See tests fail

## Your environment 🌱

- Open mSupply Version: develop

- Platform:

- Database type:

### LINKED ISSUE #4713: Prevent dose skipping
In the encounter view of the vaccine card, lines should be disabled while:

- Previous dose in that course has not been administered
- ~~Minimum interval has been passed~~

i.e. only enable row click if the previous dose status is `Given`

Minimum interval informs suggested date (maybe we put a warning in the modal if you open it before the suggested date?) but we shouldn't _prevent_ this dose from being administered, we'll leave that to the clinician's discretion

CarlosNZ:
Fixes #4713

# 👩🏻‍💻 What does this PR do?

Just adds logic to check that the previous dose has been when attempting to insert a vaccination.

I refactored the `get_related_vaccinations` method so that it takes a vaccination dose input rather than a vaccination, as the vaccination doesn't yet exist when inserting. The output of the method (previous dose, next dose) remains unchanged, we just get there a different way.

Update tests confirmed still working. Not sure if I should add another case for the Insert?

# 🧪 Testing

- [ ] Run tests. Try turning off the restriction in the front-end and confirm that an error is received?

### DIFF
--- a/server/graphql/programs/src/mutations/vaccination/insert.rs
+++ b/server/graphql/programs/src/mutations/vaccination/insert.rs
@@ -101,6 +101,7 @@ fn map_error(error: ServiceError) -> Result<InsertVaccinationResponse> {
         | ServiceError::ReasonNotProvided
         | ServiceError::StockLineNotProvided
         | ServiceError::StockLineDoesNotExist
+        | ServiceError::VaccineIsNotNextDose
         | ServiceError::ItemDoesNotBelongToVaccineCourse => BadUserInput(formatted_error),
 
         ServiceError::CreatedRecordNotFound
--- a/server/repository/src/mock/vaccination.rs
+++ b/server/repository/src/mock/vaccination.rs
@@ -15,6 +15,7 @@ pub fn mock_vaccination_a() -> VaccinationRow {
         program_enrolment_id: mock_immunisation_program_enrolment_a().id,
         vaccine_course_dose_id: mock_vaccine_course_a_dose_a().id,
         encounter_id: mock_immunisation_encounter_a().id,
+        given: true,
         created_datetime: NaiveDate::from_ymd_opt(2024, 2, 1)
             .unwrap()
             .and_hms_opt(0, 0, 0)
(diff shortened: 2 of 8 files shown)


lache-melvin (changes requested):
Looks great Carl! As mentioned in stand up, noticed the missed check for non-existent next doses... created a [suggestion branch](https://github.com/msupply-foundation/open-msupply/compare/4173-prevent-dose-skipping-back-end...4173-prevent-dose-skipping-back-end(next-dose-check-fixes)), and then tested it and realised I was always getting the not next dose error!

So added a couple of tests, apologies this bloated the suggestion branch a bit. Found that it was program_id instead of program_enrolment_id getting passed in. Have added [comments against the commit](https://github.com/msupply-foundation/open-msupply/commit/c41e082852fd03aaef3edca333a3aa3507b3d372#diff-5be53bd732da8677e526cf53541d3b44ab4647c4d9114371c7b31e17c8016396R175)

Feel free to merge/apply changes as you like, otherwise looking good 🙏

CarlosNZ:
@lache-melvin -- great thanks for the fixes.

lache-melvin (approved):
LGTM - pushed a fix to another test I broke 😅 now looks good to go!

CarlosNZ:
> LGTM - pushed a fix to another test I broke 😅 now looks good to go!

Sweet, I see this is actually branched of your one, so I'll hold off on merging until I've had another look at that (after lunch)

PR #4827, opened 2024-09-20
4714 file rearrange

lache-melvin:
Last PR of #4714

# 👩🏻‍💻 What does this PR do?

Does the discussed rearrange of component files to the one `Vaccination` folder

Also found that the view was wrong, there was a where clause in the wrong place (doses that I had given to person A but not person B, didn't show in the card at all for person B)! Just chucked the fix in here, rather than creating a new issue 🙈 

# 🧪 Testing

- [ ] All the frontend still works, API calls fire correctly
- [ ] Administer some doses for a patient
- [ ] check the vax card of another patient
- [ ] those doses are still in the card as empty lines

# 📃 Documentation

  1.
  2.


CarlosNZ:
> Also found that the view was wrong, there was a where clause in the wrong place (doses that I had given to person A but not person B, didn't show in the card at all for person B)!

Ha, I'd just noticed this (one of the doses was missing from the card when there were multiples courses) and was about to point it out. But you've fixed it already! 🔥

CarlosNZ (approved):
Nice, much tidier :)

PR #4820, opened 2024-09-19
4790 add standard report

fergie-nz:
Fixes #4790

# 👩🏻‍💻 What does this PR do?

Adds expiring-items standard report with manifest.json to openmSupply.

Next PR will add cli and report reading to facilitate building and upserting report.

## 💌 Any notes for the reviewer?

# 🧪 Testing

Nothing to test

# 📃 Documentation

Documentation part of future epic

Issue #4800, opened 2024-09-16
Report-versioning epic


fergie-nz:
## Is your feature request related to a problem? Please describe 👀

Epic for reports versioning

- [x] #4790
- [x] #4799 
- [x] #4863
- [x] #4867
- [x] #4843 
- [x] #4792 
- [x] #4822
- [ ] #4791 
- [ ] #4794 
- [ ] #4795 
- [ ] #4798
- [ ] Update docs for standard reports in the repo

## Describe the solution you'd like 🎁

# Versioning, Testing and Easy of Deployment 

## Problem Space

First of all, report are configurable and sync out to sites, and sites can be on different versions of omSupply (updating all sites is not an easy task). Consider this in the context of the following

* Report might have new views or graphql added
* We might decide to change shape of graphql (not likely)
* We can decide to change shape of our sql schema

For deployment

* It's not that easy to update reports, have to do a lot of clicking and easy to miss report updates, this needs to be improved for support

## Solution

Standard reports will be moved to omSupply repository and cli can generate one json file containing all reports in a directory, this file is bundled into omSupply application and can also be uploaded to mSupply central to update all reports at once

When omSupply starts (or via cli command), these reports are upserted by the following rules:

```
found_report = find report with this code

if not found_report {
    upsert report
    return
}

if found_report.version >= this report.version {
    return
}

upsert report
```

**code** is a new field that identifies the same report for different versions

**version** follows omSupply major and minor version, for omSupply to select correct report version for current omSupply instance, and patch identifies any updates to current report version

When omSupply generates report it will select report based on version with the follow rules

```
After narrowing down on report by context and sub context, and then by code

report_versions = all reports for code

if report_versions have custom reports {
    report_versions = only custom reports
}

// Report selection logic will be such to satisfy the following use cases (path version is ignored)
report_version = [2.3.0, 2.3.5, 2.8.2, 2.8.3, 3.0.1, 3.5.1]
if omSupply.version = 2.3 selected report = 2.3.5
if omSupply.version = 2.4 selected report = 2.3.5
if omSupply.version = 2.8 selected report = 2.8.3
if omSupply.version = 3.2 selected report = 3.0.1
if omSupply.version = 4.5 selected report = 3.5.1
```

[...]

PR #4828, opened 2024-09-20
4790 move one standard report to oms repo

fergie-nz:
Fixes #4820 

# 👩🏻‍💻 What does this PR do?

Adds cli command to upsert all standard reports (and all of their versions) to database from OMS repo.
Report upsert params are accessed from manifest file within OMS repo for each report.

A little bit untidy with string handling and unwraps in manipulating the paths within OMS repo. 
Wasn't sure if that needed to be super type safe or was OK to leave as is.

# 🧪 Testing

You can test this by running the command:

```
./target/debug/remote_server_cli upsert-standard-reports
```

from oms server

# 📃 Documentation

**Part of an epic**:  Documentation added at completion of #4800


server/cli/src/cli.rs
```diff
@@ -435,6 +508,10 @@ async fn main() -> anyhow::Result<()> {
                 FormSchemaRowRepository::new(&con).upsert_one(form_schema_json)?;
             }
 
+            let is_custom = false;
+            let version = "1.0".to_string();
+            let code = None;
```
fergie-nz, line 513: Later will add these as external params from OMS reports repo.

server/repository/src/db_diesel/report_row.rs
```diff
@@ -157,3 +158,37 @@ impl Upsert for ReportRow {
         )
     }
 }
+
+// impl<'de> Deserialize<'de> for ContextType {
```
jmbrunskill, line 162: We can remove this now right? Or is it needed in future for something?

server/cli/src/cli.rs
```diff
@@ -396,6 +398,77 @@ async fn main() -> anyhow::Result<()> {
             info!("Refresh data result: {:#?}", result);
         }
         Action::SignPlugin { path, key, cert } => sign_plugin(&path, &key, &cert)?,
+        Action::UpsertStandardReports => {
+            let connection_manager = get_storage_connection_manager(&settings.database);
```
jmbrunskill, line 402: This cli.rs file is getting huge! Would be nice to break it up a bit. Even if we just do it for this new report for now. 
Perhaps we do a re-arrange PR separately to avoid a huge PR review :)

server/cli/src/cli.rs
```diff
@@ -396,6 +398,77 @@ async fn main() -> anyhow::Result<()> {
...
+            let con = connection_manager.connection()?;
+            let base_reports_dir = "./reports";
+
+            let report_names = fs::read_dir(base_reports_dir)?
+                .map(|res| res.map(|e| e.path().into_os_string().into_string().unwrap()))
```
jmbrunskill, line 407: Maybe better to just keep this as a path at this stage?

server/cli/src/cli.rs
```diff
@@ -396,6 +398,77 @@ async fn main() -> anyhow::Result<()> {
...
+                .collect::<Result<Vec<String>, std::io::Error>>()?;
+
+            for name_dir in report_names {
+                let report_versions = fs::read_dir(&name_dir)?
+                    .map(|res| res.map(|e| e.path().into_os_string().into_string().unwrap()))
+                    .collect::<Result<Vec<String>, std::io::Error>>()?;
+
+                let (_, name) = name_dir.rsplit_once('/').unwrap();
```
jmbrunskill, line 415: Should be able to use this method rather than rsplit if we keep this a path, rather than converting to string? 
https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.file_name

server/cli/src/cli.rs
```diff
@@ -396,6 +398,77 @@ async fn main() -> anyhow::Result<()> {
...
+
+                for version_dir in report_versions {
+                    let (_, version) = version_dir.rsplit_once('/').unwrap();
+
+                    let arguments_path = format!("{version_dir}/arguments.json");
+                    let arguments_ui_path = format!("{version_dir}/arguments_ui.json");
+
+                    let manifest_file = fs::File::open(format!("{version_dir}/manifest.json"))
+                        .expect("file should open read only");
```
jmbrunskill, line 424:
```suggestion
                        .expect("Can't open manifest for report {} {}", reportname, version);
```
Might be helpfull to make this error a bit more explicit?

server/cli/src/cli.rs
```diff
@@ -396,6 +398,77 @@ async fn main() -> anyhow::Result<()> {
...
+
+                    let manifest: Manifest = serde_json::from_reader(manifest_file)
+                        .expect("manifest json not formatted");
```
jmbrunskill, line 427:
```suggestion
                        .expect("manifest json parsing error, check it's in the right format report{}:version{}", report, version);
```

jmbrunskill (commented):
LOoking good, I'll try this out properly on Monday. Added a few comments in the mean time...

fergie-nz:
Close as duplicate changes to #4842

Issue #4801, opened 2024-09-16
Historic vaccinations should not create a prescription


lache-melvin:
## Is your feature request related to a problem? Please describe 👀

If a vaccination is being entered for a past date, do not capture item/stock line details, just that the patient has received this dose. Therefore no prescription should be created.

## Describe the solution you'd like 🎁

Design:

<img width="713" alt="Screenshot 2024-09-16 at 2 38 45 PM" src="https://github.com/user-attachments/assets/ea4c5286-df51-4719-b92b-0e41161c0eb4">

### Describe alternatives you've considered 💭

While doing this, would probably be good to handle the case where no items are defined for the vaccine course, and also allow no prescription/no item required in that case!

And hiding the item select for when a different facility is chosen!

jmbrunskill:
@lache-melvin FYI I'm currently working on backdating prescriptions if that helps this situation?
https://github.com/msupply-foundation/open-msupply/issues/3672

lache-melvin:
> FYI I'm currently working on backdating prescriptions if that helps this situation?

Yeah it may at some point - for now though it seems sufficient to be capturing vaccinations just for informational purposes 🙏

marthakeezy:
Tested by Tainá in #4642 - moving to Done ✅

PR #4833, opened 2024-09-20
4796 add facility to vaccination

### LINKED ISSUE #4796: Add facility to vaccination

I want to know where the vaccination I am recording has occurred. For now, default to this facility, or if desired, switch a free text field. In future, users will be able to select other facilities as well:

Add facility field to vaccination modal: 

Screenshot 2024-09-16 at 11 48 28 AM

lache-melvin:
Fixes #4796

# 👩🏻‍💻 What does this PR do?

Adds inputs to the vaccination modal for this facility/free text other facility.

Screenshot 2024-09-20 at 6 00 32 PM
Screenshot 2024-09-20 at 6 00 41 PM

Haven't done the part of not requiring item/batch when its for another facility. Will do that with #4801, as need to make some changes to the "batch is required" logic...

# 🧪 Testing

- [ ] Add a vaccination for your facility
- [ ] Add a vaccination for a different facility - you can input custom name
- [ ] Edit a vaccination, you can change to/from "other"

# 📃 Documentation

  1.
  2.


server/graphql/programs/src/mutations/vaccination/insert.rs
```diff
@@ -94,6 +100,7 @@ fn map_error(error: ServiceError) -> Result<InsertVaccinationResponse> {
     let graphql_error = match error {
         ServiceError::VaccinationAlreadyExists
         | ServiceError::ClinicianDoesNotExist
+        | ServiceError::FacilityNameDoesNotExist
```
CarlosNZ, line 103: Small nit, but could we call this "FacilityDoesNotExist"? It makes it seem like we're checking the name field, but we're not, we're checking that the ID exists as a Facility name (right?) (I was initially confused as I wondered why we'd check the name for a free text field). (Plus it's slightly more consistent with the "ClinicianDoesNotExist" right above it)

CarlosNZ, line 103: Small nit, but could we call this "FacilityDoesNotExist"? It makes it seem like we're checking the name field, but we're not, we're checking that the ID exists as a Facility name (right?) (I was initially confused as I wondered why we'd check the name for a free text field). (Plus it's slightly more consistent with the "ClinicianDoesNotExist" right above it)

CarlosNZ (approved):
Nice. :)

Couple of very small nit picks, update as you see fit and merge away :)

client/packages/system/src/Vaccination/api/useVaccinationsGraphQL.ts
```diff
@@ -8,8 +8,8 @@ import { getSdk } from './operations.generated';
 export const useVaccinationsGraphQL = () => {
   const { client } = useGql();
   const queryClient = useQueryClient();
-  const { storeId } = useAuthContext();
+  const { storeId, store } = useAuthContext();
   const api = getSdk(client);
 
-  return { api, queryClient, storeId };
+  return { api, queryClient, storeId, store };
```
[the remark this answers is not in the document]
lache-melvin: I tried... storeId is required in the graphql call, but the whole `store` node is possibly undefined. Was a bit icky having the is defined check everywhere, so just pulled both off 🤷‍♀️

Issue #4334, opened 2024-07-01
Can create patient without first & last name in patient-program setup


nishaDangol-Sussol:
## What went wrong? 😲
Can create patient without first & last name in patient-program setup
**Note:** For normal dispensary (without patient-program setup), there is no problem; without mandatory fields, it doesn't allow for name creation further, which is correct 

Tag severity level depending on programs-feature usage
## Expected behaviour 🤔
shouldn't allow to proceed with patient creation without mandatory fields: `first name` & `last name`

## How to Reproduce 🔨

Steps to reproduce the behaviour:
1. Do the minimum to setup patient-programs in your store
2. Create new patient
3. Enter nothing and proceed
4. See error: it lets you proceed without mandatory fields: `first name` & `last name`
5. Enter nothing and save
6. See error: nameless patient without any NID nor NUIC is generated

## Your environment 🌱

- Open mSupply Version: v2.1.0-rc4
- Legacy mSupply Central Server Version: v7-15-09

- Platform: browser(chrome)

- Database type: sqlite

andreievg:
Triage, must have because it would be one of the first things to look at fixing when we have future programs budget

CarlosNZ:
This is just a JSON Forms config. I've updated the Patient type in the [ProgramSchemas repo](https://github.com/msupply-foundation/programschemas) to make `firstName` and `lastName` required.

Don't think there's anything else to do in the actual codebase is there?

roxy-dao:
I think patient programs first name and last name wasn't made mandatory cause they can have anon... but I guess they can just put anon 🤷🏻  closing issue since fix in programschema repo

PR #4841, opened 2024-09-23
#4656 Maximum age for vaccination dose (1)

### LINKED ISSUE #4656: Add maximum age to a dose
UPDATED: SEE COMMENT BELOW!

When editing a vaccine course, I should be able to assign one or more age ranges to each dose in the vaccine course schedule. ~~The age range options should be the age ranges defined for the program.~~

CarlosNZ:
Fixes #4656

# 👩🏻‍💻 What does this PR do?

Initial basic implementation of #4656. Main thing missing is the input for "Years"/"Months" -- still working on that, might want to have a chat about how far to go with that. I think to do it really properly will actually be quite a bit of work -- I'll show you my ideas tomorrow morn @lache-melvin ? (It might better if I do #4334 and come back to this)

The display order on the Vax card is already by min-age, though there's nothing specific about the order *within* the group that share the same age.r

# 🧪 Testing

- [ ] Open the "Immunizations" config in "Programs". Should see the "Max age" column
- [ ] Can edit and save the "Max age" value


client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -228,7 +228,7 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
         />
       }
       height={height}
-      width={750}
+      sx={{ maxWidth: 'unset' }}
```
CarlosNZ, line 231: Lets it expand to the width it requires -- should we set a Max?

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -259,6 +259,7 @@ const VaccineCourseDoseTable = ({
           // temp - will be overwritten by the backend to assign unique dose number (even if previous doses were deleted)
```
lache-melvin, line 259: Side note but I think we should remove this comment, we don't have a dose number at all anymore!

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -259,6 +259,7 @@ const VaccineCourseDoseTable = ({
...
           label: `${courseName} ${doses.length + 1}`,
           minAgeMonths: (previousDose?.minAgeMonths ?? 0) + 1,
+          maxAgeMonths: previousDose?.minAgeMonths ?? 1,
```
lache-melvin, line 262: Shouldn't this have a +2, or be prev.maxAge +1 or something? Otherwise maxAge is less than min age when you add a row?

server/service/src/sync/test/test_data/vaccine_course_dose.rs
```diff
@@ -12,6 +12,7 @@ const VACCINE_COURSE_DOSE1: (&str, &str) = (
         "vaccine_course_id": "test_vaccine_course",
         "label": "test dose label",
         "min_age": 12.0,
+        "min_age": 13.0,
```
lache-melvin, line 15:
```suggestion
        "max_age": 13.0,
```
Causing a failing test...

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
lache-melvin: I think its okay, the modal only expands to a margin before the edge of the screen, so will just side scroll if needed on a smaller tablet 🙏

lache-melvin (changes requested):
Looking good! Just a failing test, and the add a dose default value to fix...

Were you meaning to link a different issue in your description? #4334 looks closed, and I'm not sure how it relates... we can discuss in stand up though 👍

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
lache-melvin: kidding, I think keeping a set width is good, otherwise it jumps around when you add the first dose:

Screenshot 2024-09-24 at 8 54 10 AM
Screenshot 2024-09-24 at 8 54 15 AM

CarlosNZ: Cool, done. It can still collapse smaller when it needs to, too.

CarlosNZ:
> Were you meaning to link a different issue in your description? #4334 looks closed, and I'm not sure how it relates... we can discuss in stand up though 👍

Er yeah, I was just referring to one of the other smaller issues in our list. 🙄

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
CarlosNZ: Yeah, I thought maybe we were going to be getting new defaults from back-end, but never mind, have updated now.

server/service/src/sync/test/test_data/vaccine_course_dose.rs
CarlosNZ: Done, thanks

lache-melvin (approved):
Looks great, thanks Carl!

PR #4842, opened 2024-09-23
4822 add cli commands to oms standard reports

### LINKED ISSUE #4822: Add cli command to upsert standard report within OMS repo

Part of epic #4800 

Add cli command to replace bash scripting in upserting standard reports within OMS repo

- [ ] Access upsert params from the manifest.json file.
- [ ] Upserts all reports and all of their versions to OMS central database

fergie-nz:
Fixes #4822

# 👩🏻‍💻 What does this PR do?

Adds CLI command to upsert one or all standard reports.
This will upsert all versions of new reports
Also adds multiple versions of this report

This is part of epic #4800 

# 🧪 Testing

You can test this by building the server binary, then running 
```
./target/debug/remote_server_cli upsert-standard-reports
```

Or by adding the optional parameter name for a specific report:

```
./target/debug/remote_server_cli upsert-standard-reports --name expiring-items
```

# 📃 Documentation

**Part of an epic**: #4800 documentation will be completed for the feature as a whole


server/cli/Cargo.toml
```diff
@@ -22,6 +22,8 @@ util = { path = "../util" }
 service = { path = "../service" }
 server = { path = "../server" }
 graphql = { path = "../graphql" }
+report_builder = {path = "../report_builder"}
```
fergie-nz, line 25: I've added report builder as a dependency rather than running the report builder binary.
I thought this made more sense? Though I know dependencies have caused slower build times so unsure if this is the right approach

server/cli/src/cli.rs
```diff
@@ -139,6 +142,12 @@ enum Action {
     /// User can specify max number of backup to keep, see example configuration file
     Backup,
     Restore(RestoreArguments),
+    // command to upsert standard reports. Will later move this into rust code
+    UpsertStandardReports {
+        /// Optional report name. If none supplied, all standard reports are uploaded
+        #[clap(short, long)]
+        name: Option<String>,
```
fergie-nz, line 149: Add optional name in case we want to be specific in what report we upload

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                .map(|res| res.map(|e| e.path().into_os_string().into_string().unwrap()))
+                .collect::<Result<Vec<String>, std::io::Error>>()?;
+
+            let report_names = match &name {
+                Some(name) => report_names
+                    .into_iter()
+                    .filter(|report| report == name)
+                    .collect(),
+                None => report_names,
+            };
```
fergie-nz, line 423: filter for specific name if passed as a param

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                    let args = BuildArgs {
+                        dir: format!("{version_dir}/src"),
+                        output: format!("{version_dir}/generated/{name}.json").into(),
+                        template: "template.html".to_string(),
+                        header: None,
+                        footer: None,
+                        query_gql: Some("query.graphql".to_string()),
+                        query_default: None,
+                        query_sql: None,
+                    };
```
fergie-nz, line 444: later this will need to be more specific to include headers and footers, and also to be flexible to different types of standard report queries.
Some of these might just need renaming to a standard graphql name, others will need to search and pass a vec of sql queries. (You can see this in item-usage report which has a large number of sql queries)

fergie-nz: Have added an issue for this #4843

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+
+                    let _ = build(args);
+
+                    let arguments_path = format!("{version_dir}/argument_schemas/arguments.json");
+                    let arguments_ui_path =
+                        format!("{version_dir}/argument_schemas/arguments_ui.json");
+
+                    let manifest_file = fs::File::open(format!("{version_dir}/manifest.json"))
+                        .expect("file should open read only");
```
fergie-nz, line 453: Feels a little strange having these as hard coded strings. I don't really like it! But I guess is also the same as hard coded cli command paths?

I would like to add better error handling and/or good testing for these which would help

server/cli/src/cli.rs
```diff
@@ -435,6 +540,10 @@ async fn main() -> anyhow::Result<()> {
                 FormSchemaRowRepository::new(&con).upsert_one(form_schema_json)?;
             }
 
+            let is_custom = false;
+            let version = "1.0".to_string();
+            let code = None;
```
fergie-nz, line 545: This is for the upsert custom report (old report upsert method from oms reports repo). Maybe later we would want to have these as passable params?

server/cli/src/cli.rs
jmbrunskill: I think the idea was to specify the header, footer, queries, etc all in the manifest file. Then you should only need to read the one file here? Or at least get the paths from the manifiest to add to pass to the buildier?

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                    let id = format!("{name}_{version}");
+                    let report_path = format!("{version_dir}/generated/{name}.json");
+                    let context = manifest.context;
+                    let report_name = manifest.name;
+                    let is_custom = manifest.is_custom;
+
+                    let sub_context = manifest.sub_context;
+                    let code = manifest.code;
+
+                    let filter = ReportFilter::new().id(EqualFilter::equal_to(&id));
```
jmbrunskill, line 467: I think Andrei's idea was to use the `code` to identify the matching report.
I think this is probably fine too, as long as there's no duplicate names...

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                        .and_then(|r| r.argument_schema.as_ref().map(|r| r.id.clone()));
+
+                    let form_schema_json = schema_from_row(FormSchemaRow {
+                        id: argument_schema_id.unwrap_or(format!("for_report_{}", id)),
+                        r#type: "reportArgument".to_string(),
+                        json_schema: fs::read_to_string(arguments_path)?,
+                        ui_schema: fs::read_to_string(arguments_ui_path)?,
+                    })?;
+
+                    FormSchemaRowRepository::new(&con).upsert_one(&form_schema_json)?;
```
jmbrunskill, line 481: Not every report has a schema....

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                        output: format!("{version_dir}/generated/{name}.json").into(),
+                        template: "template.html".to_string(),
+                        header: None,
+                        footer: None,
+                        query_gql: Some("query.graphql".to_string()),
+                        query_default: None,
+                        query_sql: None,
+                    };
+
+                    let _ = build(args);
```
jmbrunskill, line 446: Would be good to return the error if there is one...

server/cli/src/cli.rs
```diff
@@ -396,6 +405,102 @@ async fn main() -> anyhow::Result<()> {
...
+                        ui_schema: fs::read_to_string(arguments_ui_path)?,
+                    })?;
+
+                    FormSchemaRowRepository::new(&con).upsert_one(&form_schema_json)?;
+
+                    ReportRowRepository::new(&con).upsert_one(&ReportRow {
+                        id,
+                        name: report_name,
+                        r#type: repository::ReportType::OmSupply,
+                        template: fs::read_to_string(report_path)?,
```
jmbrunskill, line 487: Would be nice if we can just return the string from teh report builder rather than creating the .json file?
Can re-look at that when we do the embedding stuff I think.

jmbrunskill (commented):
Nice, looking good!

Will give it a test tomorrow, and better review.
Can you change the version to `2_3_0` please?
Still showing as `2_3_1` in this branch?

server/cli/Cargo.toml
jmbrunskill: I think that's a good idea for now.
Might make sense to move more of this logic around I think, but need some more time to think about it...

server/cli/src/cli.rs
fergie-nz: Yes thank you James, I had just picked up on this also after adding this PR

client/packages/invoices/src/Returns/SupplierDetailView/SidePanel/RelatedDocumentsSection.tsx
```diff
@@ -21,7 +21,7 @@ export const RelatedDocumentsSectionComponent = () => {
   const { originalShipment } = data ?? {};
 
   const getLabel = (createdDatetime: string, username?: string) => {
-    const label = t('messages.outbound-shipment-created-on', {
+    const label = t('messages.inbound-shipment-created-on', {
```
jmbrunskill, line 24: Where's this change from ? Seems unrelated to this?

jmbrunskill (approved):
Approved

PR #4844, opened 2024-09-23
4691 Wrapped AppBarButtons have no margin

### LINKED ISSUE #4691: Wrapped AppBarButtons have no margin
## What went wrong? 😲

Sometimes the buttons wrap, if there are a lot of them, on a just-big-enough-that-the-button-are-expanded sized screen (especially when you open the side panels).

When they do this, they lose their top and bottom margin:
Screenshot 2024-08-27 at 12 41 41 PM

Looks pretty ugly with the drop shadows lol:
Screenshot 2024-08-27 at 12 45 19 PM

## Expected behaviour 🤔

When they're all in one line, they have some margin - this should still exist if they wrap across two lines:
Screenshot 2024-08-27 at 12 40 21 PM

## How to Reproduce 🔨

1. Go to a page w lots of buttons, like Internal Order
2. Make screen as narrow as you can without the button text collapsing to only show icon
3. And/or open the nav drawer/more info panel to reduce the width of the main page
4. See buttons wrapping, with no margin above/below

## Your environment 🌱

- Open mSupply Version: 2.2

- Platform:

- Database type:

roxy-dao:
Fixes #4691

# 👩🏻‍💻 What does this PR do?
Was trying to find where the padding/margin was... but looks like it was a min-height set on some element... so have added padding to account for wrapping in App Bar Button

# 🧪 Testing

- [ ] Go to internal order
- [ ] Inspector
- [ ] Change sizing of screen

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/portals/AppBarButtons/AppBarButtons.tsx
+++ b/client/packages/common/src/ui/components/portals/AppBarButtons/AppBarButtons.tsx
@@ -27,7 +27,12 @@ export const AppBarButtonsPortal: FC<BoxProps> = props => {
 
   return (
     <Portal container={appBarButtonsRef.current}>
-      <Box {...props} />
+      <Box
+        sx={{
+          padding: 1.5,
+        }}
+        {...props}
+      />
     </Portal>
   );
 };


lache-melvin (approved):
LGTM TY

PR #4845, opened 2024-09-23
4801 prevent historic vaccination transactions

### LINKED ISSUE #4801: Historic vaccinations should not create a prescription

lache-melvin:
Fixes #4801

# 👩🏻‍💻 What does this PR do?

Adds messages about when stock transactions can and can't be created, and allows the user to choose whether to update them:

Screenshot 2024-09-24 at 10 24 52 AM
Screenshot 2024-09-24 at 10 25 04 AM

Screenshot 2024-09-24 at 10 18 53 AM
Screenshot 2024-09-24 at 10 19 06 AM

Screenshot 2024-09-24 at 10 16 31 AM

# 🧪 Testing

There's a lot of variants, plz engage your inner QA and try break it:

- [ ] Create/edit vaccination in all the variations/combinations you can
  - [ ] This facility/other
  - [ ] Today/historic
  - [ ] Given/not given
  - [ ] When there is an existing prescription/not
- [ ] Ensure the toggles show the right text at the right time, and that the invoices are created/not created appropriately

# 📃 Documentation

  1.
  2.


client/packages/system/src/Vaccination/Components/SelectItemAndBatch.tsx
```diff
@@ -0,0 +1,127 @@
...
+  SxProps,
+  Typography,
+  useTranslation,
+} from '@openmsupply-client/common';
+import React, { PropsWithChildren, useMemo } from 'react';
+import { VaccinationDraft } from '../api';
+import { VaccinationCourseDoseFragment } from '../api/operations.generated';
+import { OTHER_FACILITY } from './FacilitySearchInput';
+
+export const SelectItemAndBatch = ({
```
lache-melvin, line 15: The modal component was getting huge, pulls this out into separate component

client/packages/system/src/Vaccination/Components/SelectItemAndBatch.tsx
```diff
@@ -0,0 +1,127 @@
...
+      })) ?? []
+    );
+  }, [dose.id]);
+
+  if (!draft.given) {
+    return null;
+  }
+
+  if (draft.facilityId === OTHER_FACILITY) {
+    return <InfoText>{t('messages.no-transaction-other-facility')}</InfoText>;
```
lache-melvin, line 44: This text and the switch below are the new pieces here

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -87,47 +88,19 @@ export const VaccinationModal = ({
...
-                >
-                  {t('button.view-prescription')}
-                  <ArrowRightIcon />
-                </Link>
-              )}
-            </Box>
-          ),
-        }
-      : undefined;
+  const InfoBox = <GivenInfoBox vaccination={vaccination} />;
```
lache-melvin, line 91: Most of the changes in this file are just rearranging, moving out this infobox component and the batch select

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -267,41 +247,35 @@ const VaccinationForm = ({
...
+
+      <SelectItemAndBatch
+        dose={dose}
+        draft={draft}
+        openBatchModal={openBatchModal}
+        updateDraft={updateDraft}
+        hasExistingSelectedBatch={!!vaccination?.stockLine?.id}
+      />
+
+      {(shouldRevertInvoice ||
```
lache-melvin, line 259: This is the actual change, adding the switch

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -43,34 +45,13 @@ export function useVaccination({
...
-  });
-  const { data: vaccination, isLoading: vaccinationLoading } = useQuery({
-    queryKey: [VACCINATION, vaccinationId],
-    queryFn: async () => {
-      if (!vaccinationId) {
-        return null;
-      }
-      const result = await api.vaccination({ vaccinationId, storeId });
+  const { data: dose, isLoading: doseLoading } =
+    useDoseQuery(vaccineCourseDoseId);
```
lache-melvin, line 51: heck sorry more rearranging, moves the guts of these queries outside the top level hook, was getting unwieldy

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -103,27 +85,22 @@ export function useVaccination({
...
     // Populate with existing vaccination data
-    comment,
+    comment: comment ?? '',
     stockLine,
     given,
     notGivenReason,
     itemId: stockLine?.itemId,
+
+    recordHistoricalTransaction: false,
+    editExistingTransactions: false,
```
lache-melvin, line 95: Only juicy bit is these

server/graphql/programs/src/mutations/vaccination/insert.rs
```diff
@@ -106,7 +106,6 @@ fn map_error(error: ServiceError) -> Result<InsertVaccinationResponse> {
         | ServiceError::ProgramEnrolmentDoesNotMatchVaccineCourse
         | ServiceError::VaccinationAlreadyExistsForDose
         | ServiceError::ReasonNotProvided
-        | ServiceError::StockLineNotProvided
```
lache-melvin, line 109: Removes this error - there are so many cases now where we allow a stock line to not be provided/don't want to generate an invoice. Easier to let the caller pass in stock line if they care, otherwise don't worry about it, than validate every case.

The stockline part is not the core function (we want to record there was a vaccination, the stock transactions are helpful additions) so I think best not to block the vaccination being recorded

CarlosNZ (commented):
Looks good, some pretty complicated logic there, but seems to be doing what it's supposed to. Would be good to have all the "rules" written down in one place somewhere (even if just in comments). Can do that later though if we want to get this in.

Just a couple of cases I wonder about:

1. When giving a vaccination today, at the current facility, it seems like recording the stock movement would be the norm. And yet, it feels a little too easy to just skip over it and push OK. I wonder if we want to either:
- require an "Are you sure?" confirmation when doing the above without entering a stock item
- have another toggle switch like the other ones you've added that says "Don't record stock transaction" that you have to explicitly set to "yes" before it lets you submit with no transaction
2. Similarly, when you select a "Vaccine item", maybe you should be required to enter a batch before you can submit. It seems a bit easy to just select an item and submit and assume that it's recorded the stock movement, but it hasn't.

Those can be dealt with in other issues if we want to deal with them though.

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -345,9 +319,40 @@ const VaccinationForm = ({
...
+                alignItems: 'center',
+                display: 'flex',
+              }}
+              to={RouteBuilder.create(AppRoute.Dispensary)
+                .addPart(AppRoute.Prescription)
+                .addPart(vaccination.invoice.invoiceNumber.toString())
+                .build()}
+            >
+              {t('button.view-prescription')}
+              <ArrowRightIcon />
```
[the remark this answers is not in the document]
mark-prins: bearing in mind the distaste that users have for our current links which open in a new tab. it is something that I've had a number of minor complaints about. 

worth checking in android too, I don't recall how we're handling new tab links on android.

lache-melvin:
Hey @CarlosNZ thanks for feedback! have consolidated the switch to one place, hopefully a bit easier to make sense of?

Great point about the being too easy to not add the item/batch, might leave to do with the batch select refactor 🙏

adamdewey:
> Similarly, when you select a "Vaccine item", maybe you should be required to enter a batch before you can submit. It seems a bit easy to just select an item and submit and assume that it's recorded the stock movement, but it hasn't.

Agree with this - I think it shouldn't be possible to select a vaccine item but not the batch

lache-melvin:
> Agree with this - I think it shouldn't be possible to select a vaccine item but not the batch

Yeah agree - we just also didn't want to block the workflow e.g. you select the item, but the system says there are no available batches, even though you might have one in hand (Chris mentioned it is common to eek 21 doses out of a 20 dose vial!)

You should be able to enter the vaccination... but I think a confirmation from the user that not stock has been linked would be great 🚀

adamdewey:
> **Yeah agree - we just also didn't want to block the workflow e.g. you select the item, but the system says there are no available batches, even though you might have one in hand (Chris mentioned it is common to eek 21 doses out of a 20 dose vial!)

Ah, I think there is some confusion here between the dose and the stock vaccine item?

You should still be able to record that a dose from the vaccine course was given regardless of stock, but if you're selecting a vaccine item that's indicating you're involving stock and then you need to also capture the batch. 

There's a point to be made that the user doesn't need to capture batches for all stock items so the batch may not exist, but the reality for vaccines is that they will all 100% have batch numbers (in fact we should almost make it mandatory to capture batches for vaccine items in stock management)

adamdewey:
Ah ok, I think I see what you're saying about the bonus dose - that the batch won't be available to select if there is no stock available. 

Hmm, maybe we need to address that edge case specifically?

CarlosNZ (approved):
Looks good, thanks. Tests make it clear what all the conditions are, thanks :)  

Some follow-up regarding some of this it would seem, based on discussion. Should I make a new issue for it?

Also, couple of other things I noticed -- wonder what the restrictions should be:

Screenshot 2024-09-25 at 4 46 10 PM

- It's a bit weird to have a "Date Given" date for doses that are "Not Given". Should we re-word this somehow?
- Should we restrict the dates for given doses so that they must be sequential? It's a bit weird to insist that we give doses in order, but you can just put the dates out of order anyway?

lache-melvin:
Haha those are great points! Yeah would be good to make follow up issues for them. Have noted the above comments in the batch select refactor issue, will see what feels like it still needs fixing after that 🙏

PR #4847, opened 2024-09-24
4351 refactor stocktake in service layer

### LINKED ISSUE #4351: Refactor Stocktake in service layer
## The suggestion
The stocktake insert/update files are getting really big and would be nice if some of the functions can be moved out into their own validate/generate files like in the rest of the repo.

Nicer to read! And way easier to navigate through~

Shouldn't be any risks? 

Might take a few hours to move things out and make sure nothing has broken~

roxy-dao:
Fixes #4351

# 👩🏻‍💻 What does this PR do?
Refactor stocktake/stocktake lines into validate/generate/mod files under each action.

Some validation functions need to be reviewed as well since we have dups but didn't wanna crowd this PR with that

P.S would like to hold merging this PR until a 2.3.0 branch is created (: 

# 🧪 Testing

- [ ] Make sure stocktake is still working as it should

# 📃 Documentation

  1.
  2.


server/service/src/stocktake/delete/mod.rs
```diff
@@ -29,34 +29,6 @@ pub struct DeleteStocktake {
...
-    }
-
-    if existing.is_locked {
-        return Err(DeleteStocktakeError::StocktakeIsLocked);
-    }
-
-    if !check_stocktake_not_finalised(&existing.status) {
-        return Err(DeleteStocktakeError::CannotEditFinalised);
-    }
-    // Note that lines are not deleted when an invoice is deleted, due to issues with batch deletes.
```
roxy-dao, line 52: Note from very, very long ago. Deletes seem to be working as it should...

fergie-nz:
I'm sure you've seen but needs base updating for 2.3 release 🚀

roxy-dao:
> I'm sure you've seen but needs base updating for 2.3 release 🚀

Refactor fortnight issues go into 2.4 unless it's a small change (:

fergie-nz:
I am wondering when reading this about test coverage for refactors, and when we should be checking test coverage. 

It looks to me like a safe moving around of the same functions to different files, but I also wonder if we are wanting comprehensive test coverage in these cases anyway? 

I also find it kind of hard to tell how comprehensive our test coverage is, and when facing that uncertainty, is it the role of the reviewer to fully understand all use cases and ensure that we do have good testing?

roxy-dao:
> I am wondering when reading this about test coverage for refactors, and when we should be checking test coverage.
> 
> It looks to me like a safe moving around of the same functions to different files, but I also wonder if we are wanting comprehensive test coverage in these cases anyway?
> 
> I also find it kind of hard to tell how comprehensive our test coverage is, and when facing that uncertainty, is it the role of the reviewer to fully understand all use cases and ensure that we do have good testing?

There are already tests covering the stocktake logic, and the logic itself hasn't changed so those tests should still pass.

The reviewer should ensure that the refactor doesn't break existing behaviours since refactors shouldn't really be changing any logic... If the refactor touches an area that doesn't already have existing tests, then tests should definitely be added to ensure everything is still working as it should! (:

fergie-nz (approved):

PR #4848, opened 2024-09-24
4656(2) Age Month/Year input

### LINKED ISSUE #4656: Add maximum age to a dose

CarlosNZ:
Fixes #4656 

Last part of this issue, adding the Year/Month age inputs

# 👩🏻‍💻 What does this PR do?

Created a new general-purpose "Multi-Unit" input cell, where any number of arbitrary units can be defined (works for H/M/S if we need it!), and implement it for the Vaccine course age ranges.

Screenshot 2024-09-24 at 2 25 51 PM

New component is essentially very similar to the existing "NumberInputCell" -- this just allows an *array* of inputs to be defined, each related to the base unit by a ratio.

Added some test cases to test the functions used to convert to and from multiple units to a single value.

I'd like to have more functionality for this input, as previously discussed, but this is a start :) 

Do we think the abbreviations ("Y"/"M") are acceptable? I don't love them, but they seem like a necessary evil in order to conserve screen width, and also to avoid the pluralisation headache we'd have to deal with if we showed the full word. Plus, intention to improve later with the moving label/placeholder text, as discussed.

Main improvements I'd like to make:
- Labelling/placeholder text improvements (to be more like the Filters)
- Auto-conversion of input: user can enter "24" in months and it would auto-format to "2Y 0M". I could do this easily enough already, but it's a bit annoying without adding some debouncing, or waiting till `onBlur`, which is getting a bit complex for now.

# 🧪 Testing

- [ ] See inputs on Vaccine Course modal, and check they behave as expected
- [ ] Run tests with `yarn test multi`


client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
```diff
@@ -0,0 +1,167 @@
...
+          // unit, e.g. "Months" would be limited to 11.9999
+          max={setMax(units, index, max)}
+          decimalLimit={decimalLimit}
+          step={step}
+          multiplier={multiplier}
+          allowNegative={allowNegative}
+          defaultValue={defaultValue}
+          value={cellValues[index]}
+          width={width}
+          endAdornment={unit.label}
```
CarlosNZ, line 123: Yay for end adornment -- worked a treat for now.

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { RecordWithId } from '@common/types';
+import { useBufferState, useDebounceCallback } from '@common/hooks';
+
+interface Unit {
+  key: string;
+  ratio: number;
+  label: string;
+}
+
+export const MultipleNumberInputCell = <T extends RecordWithId>({
```
CarlosNZ, line 61: Most of this is just copied straight from "NumberInputCell". Only changes are in relation to multi-unit functionality.

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const ratio = units[index]?.ratio ?? 1;
+    return sum + ratio * val;
+  }, 0);
+
+export const setMax = (
+  units: Unit[],
+  index: number,
+  max: number | undefined
+) => {
+  if (index === 0) return max;
```
CarlosNZ, line 157: I made the component's `max` property applicable to whatever the highest unit is (i.e. years). But now that I think about it, it should probably be in terms of the *base* unit, right? Makes the logic a tad more complex, but probably worth it? (It would also be more consistent with the `min` prop)

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -229,7 +230,7 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
       }
       height={height}
       sx={{
-        width: 875,
+        width: 950,
```
CarlosNZ, line 233: Pretty much the minimum value that doesn't require side-scrolling. Fortunately the modal still shrinks as the viewport narrows.

lache-melvin: I think this is ok, its only accessible on central server 👍

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
```diff
@@ -0,0 +1,167 @@
...
+            newValues[index] = newValue;
+            const newTotal = computeTotal(newValues, units);
+            setBuffer(newTotal);
+            updater({ ...rowData, [column.key]: Number(newTotal) });
+          }}
+          min={min}
+          // Limit smaller units to less than the value of the next highest
+          // unit, e.g. "Months" would be limited to 11.9999
+          max={setMax(units, index, max)}
+          decimalLimit={decimalLimit}
```
lache-melvin, line 116: Do you think we could require all the earlier units to be integers, and only the last accept decimals? This seems really unintuitive to me:


https://github.com/user-attachments/assets/96f14f08-0b5d-4a08-a97a-a23c08a23720

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
```diff
@@ -0,0 +1,167 @@
...
+) => {
+  if (index === 0) return max;
+  const thisRatio = (units[index] as Unit).ratio;
+  const higherRatio = (units[index - 1] as Unit).ratio;
+  // If not the last value, should be limited to the integer below the next
+  // highest ratio. But for the last value, we need to be able to accept
+  // fractional input right up to the ratio, so we allow this to an arbitrary
+  // decimal precision (can't see needing more than 4 d.p for this kind of
+  // thing!)
+  const limitReduction = index === units.length - 1 ? 0.0001 : 1;
```
lache-melvin, line 165: Given how narrow the cells are, I think might need to go with 1 or 2 dp (maybe needs to be configurable for different use cases, but that can be done in future as needed) ... I thought this was jumping to `999`


https://github.com/user-attachments/assets/b433fc68-37c4-41a8-816a-bbc236bc838d

lache-melvin (changes requested):
Awesome thanks Carl, great to have the test coverage on these elements! Just a couple of requests, let me know what you think 🙏

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
lache-melvin: Hm, max years actually makes more sense to me than max months, though I see your point... I think document as a comment, and see if there is any feedback if/when this component gets reused 🙏

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
CarlosNZ: Yes, for sure, there's no reason to ever need decimals for anything other than the smallest unit.

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
CarlosNZ: Yes, for sure, there's no reason to ever need decimals for anything other than the smallest unit.

CarlosNZ: Yes, for sure, there's no reason to ever need decimals for anything other than the smallest unit.

client/packages/common/src/ui/layout/tables/components/Cells/MultiNumberInputCell/MultiNumberInputCell.tsx
CarlosNZ: I actually thought there was a default limit of 2dp, but it doesn't seem to be the case here. I've added that to this component now.

CarlosNZ: Having said that, the limit here is just for internal precision, so the num of dp here just needs to be more than anything we'd realistically expect to be used, hence 4 seemed reasonable.

CarlosNZ:
Thanks, @lache-melvin , have made the following changes:
- Comment about the `max` prop
- Only the smallest unit can ever have non-integer values
- Added default of 2dp for this component

6b0d9b8613

lache-melvin (changes requested):
Hm, the dp change doesn't seem to fix for set to max... e.g. still get this if I type 20 in the months field:

Screenshot 2024-09-25 at 11 05 28 AM

Also, one more request... would it be easy enough to ensure the `To` value is greater than the `From`? If its a rabbithole I wouldn't worry for now, seeing as I don't this any of the logic really depends on it, but would be nice 😁

CarlosNZ:
> Also, one more request... would it be easy enough to ensure the `To` value is greater than the `From`? If its a rabbithole I wouldn't worry for now, seeing as I don't this any of the logic really depends on it, but would be nice 😁

Little bit of a rabbit hole, but I think it's working okay now. I ended up changing the approach to handle the maximum and minimum to just constrain the numeric total after re-combination. The logic to correctly set min and max values for each individual cell is actually quite ridiculous to do 100% properly, so this seemed more straightforward. Downside is you can get a little bit of "lag" as the input is reset to its constrained values, but probably okay for now.

But we have interdependent min/max for the age ranges now! c016125450

lache-melvin (approved):
Awesome thanks Carl! Works great 🚀 

Only thing I'm wondering is whether, instead of the min age having a max, the max age increases as you increase the min age... just as you're entering left to right, I kind of want to be able to set the correct min age without having to first set a higher max age 👀 

But let's get this in as is, and see what feedback we get 🙏

CarlosNZ:
> Only thing I'm wondering is whether, instead of the min age having a max, the max age increases as you increase the min age... just as you're entering left to right, I kind of want to be able to set the correct min age without having to first set a higher max age 👀

That's true. I actually thought this would be easy to implement, but I've quickly tried it and it's not quite working right. There's some strange interplay going on between the rowData value and the buffer value that is not correct and it'll take a little while to disentangle. I'll make a new issue to address this and some of the other little quirks I've noticed.

CarlosNZ:
Issue to make further improvements to this: https://github.com/msupply-foundation/open-msupply/issues/4874

PR #4851, opened 2024-09-24
4799 add embedding standard reports in oms build

### LINKED ISSUE #4799: Report-versioning: Make standard reports appear by default in open-mSupply

Need to include standard reports in OMS installations.

Maybe via bundling? 
Need some discussion and planning for how we bundle and select compatible reports + manage avoiding excess sync traffic where we update local omsupply installations.

Also need standard reports to be compiled as one large json inclusive all variables - and also upsertable via CLI with one file argument.
This makes the solution consumer focussed to support so that they can easily update 

We create a new Central Server Build for open-mSupply.
This version will register itself in windows as open-mSupply-Central (or similar)

Only the oms Central server install will come with standard reports
These will be synced out to remote sites (assuming they are on a new enough version to check for report updates)

For now ALL report version will go to ALL sites. Each site will select the latest version that's compatible.

Filtering in Sync could be done in future issues?

Bundled reports will be bundled using `rust embed` similar to frontend code.
The bundled reports will be inspected at startup, and inserted into the reports table if new.

1. Standard Reports are bundled into every OMS version
This has the challenge that because OMS Central has the bundled reports it will still try to sync everywhere anyway.
Not clear how to avoid this problem if we do want to be able to update bundled reports via sync...

2. Standard Reports are locked to OMS version
This greatly simplifies the versioning question at the cost of flexibility.
A new installation would be required to update standard reports.
Versioning is likely to be needed for custom reports anyway, so would probably just as much work...

fergie-nz:
Fixes #4799

# 👩🏻‍💻 What does this PR do?

Updates CLI for embedded standard reports.
This removes the UpsertStandardReports command from 4822, and adds instead:
BuildStandardReports (which builds all standard reports from src files within the OMS repo into one large json also within the OMS repo).
UpsertReportsJson upserts standard reports from the embedded json in the OMS repo, or, if an alternative json_path is provided, upserts reports based on this json.
This allows a cli to upsert standard reports to the db from embedded files, or for support to upsert specific reports.

# 🧪 Testing

You can test this by building the back end, then running 

```
./target/debug/remote_server_cli build-standard-reports
``` 

first to build standard reports, and then

```
./target/debug/remote_server_cli upsert-reports-json
```

# 📃 Documentation

- [ ] **Part of an epic**: documentation to be added with #4800


server/cli/src/cli.rs
```diff
@@ -639,3 +696,24 @@ pub struct TestReportArguments {
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub struct ReportData {
+    pub id: String,
+    pub name: String,
+    pub r#type: repository::ReportType,
+    pub template: ReportDefinition,
+    pub context: ContextType,
+    pub sub_context: Option<String>,
+    pub argument_schema_id: Option<String>,
```
fergie-nz, line 708: There is a bunch of nested structs here.
I wasn't sure the best way to do this, and doing this way adds a bit of bloat in structs.

I also didn't like how it wasn't actually type safe because fields are optional. ie the 'arguments' field in manifest.json is optional as well as the fields within arguments.
This means it is quite easy to miss spell a field resulting in it not being found and added to the large json file (and is quite easy to miss!). I wonder if there is a way to automate this? Make a test which checks the json fields? Or perhaps structuring it in a different way? having all fields but explicitly calling None / null in the fields where they are empty?

Later I will add more tests to hopefully make

server/report_builder/src/build.rs
```diff
@@ -260,3 +260,10 @@ pub fn build(args: BuildArgs) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+pub fn build_report_definition(args: BuildArgs) -> anyhow::Result<ReportDefinition> {
+    let project_dir = Path::new(&args.dir);
+    let files = find_project_files(project_dir)?;
+    let definition = make_report(&args, files)?;
+    Ok(definition)
+}
```
fergie-nz, line 269: New function basically duplicate of 'build', but returns the struct rather than saving as json in dir

server/cli/src/cli.rs
```diff
@@ -502,29 +507,81 @@ async fn main() -> anyhow::Result<()> {
...
+        Action::UpsertReportsJson { json_path } => {
+            let connection_manager = get_storage_connection_manager(&settings.database);
+            let con = connection_manager.connection()?;
+            let base_reports_dir = "./reports";
+            let generated_dir = format!("{base_reports_dir}/generated");
+
+            let json_file = fs::File::open(
+                json_path.unwrap_or(format!("{generated_dir}/standard_reports.json")),
+            )
+            .expect("json not found");
```
jmbrunskill, line 561:
```suggestion
            .expect("{generated_dir}/standard_reports.json not found");
```
Would be nice to be more explicit with error case...

server/report_builder/src/build.rs
jmbrunskill: Would be nice to get the one the writes the JSON to call this one, then save the file.
Just means we don't have two code paths that can get out of sync.

jmbrunskill (approved):
Nice work, Ferg!

Super fast to run too.

This PR isn't quite doing the embedding stuff, and does't quite meet the requirements for issue #4799.

Good to merge this PR, but lets get on to embedding tomorrow.
Should be similar to this code you did for localisations
https://github.com/msupply-foundation/open-msupply/blob/78b2a5efa5254e3acced81b62718fd77399c0c32/server/service/src/localisations.rs#L10

Just need to embed the standard-reports json, then upsert on startup for now I think.

PR #4854, opened 2024-09-24
3672 show historical stock in prescription

### LINKED ISSUE #3672: Backlog prescription entry

jmbrunskill:
Fixes #3672

# 👩🏻‍💻 What does this PR do?

Shows the historical stock available when prescribing/dispensing

This basically just swaps out the stocklines query for the historical stock lines if a date is set for the prescription.

We only adjust the available stock level, so in Stock Packs still shows what's available now.

image

Ran this past @mark-prins  we think this is good, as the user sees what's in the store now (what they're expecting to see) and how much they can issue (in the past). Hopefully intuitive enough but we need to update the docs.

# 🧪 Testing

- [ ] Introduce stock (today?)
- [ ] Create a backdated prescription
- [ ] Try to issue stock
- [ ] Make sure you can't issue the newly introduced stock from today
- [ ] Try other variations, e.g. different dates, different stock lines, etc.
- [ ] Make sure you can edit the lines correctly, e.g. if you allocate stock on a line then edit it, should still be able to allocate that stock.

# 📃 Documentation

- [ ] Dispensing docs need to be updated to include backdated prescriptions


jmbrunskill:
Raised issue https://github.com/msupply-foundation/open-msupply/issues/4872 for not showing the error when changing date. New PR coming for this...

andreievg (changes requested):
@jmbrunskill I think we need to use total for backwards ledger calculations, i was getting negative values otherwise (interesting enough negative values were issuable 😨): https://github.com/msupply-foundation/open-msupply/compare/3672-show-historical-stock-in-prescription...3672-show-historical-stock-in-prescription(Use-total-for-back-ledge-calculation) -> tests pass and my manual test passes too.

Above is replicated, by having another prescription or invoice that  reduces available levels. And then in another prescription back data to before stock existed, and try to add that item/stock line

I am not getting an error toast when trying to back date prescription with existing line (beyond it's stock levels).

jmbrunskill:
Your changes seem to work well for me. Not sure exactly if I'm testing the same thing but my tests seem ok too after your change.

andreievg (approved):
Thanks @jmbrunskill

PR #4868, opened 2024-09-24
4657 vaccine course dose soft delete

### LINKED ISSUE #4657: Persist/soft delete vaccine course doses
Currently, every time a vaccine course is edited, each vaccine_course_dose is deleted and recreated.

However, we need to implement soft deletes for vaccine_course_dose, which means the delete all and recreate will no longer suit. Instead, we should receive the current list of vaccine_course_dose from the frontend, and compare that with all the vaccine_course_dose that we have for that vaccine_course.

Any that aren't in the current list should be soft deleted - all others can be upserted. 

It should be noted (in the code) that we are soft-deleting so as not to break references for patients that have already received that dose.

- [ ] Implement soft delete of vaccine_course_dose (for when a user explicitly deletes a dose)
- [ ] Upsert remaining vaccine_course_doses provided by the frontend

lache-melvin:
Fixes #4657 

# 👩🏻‍💻 What does this PR do?

Soft deletes vaccine course doses, so you can edit a vaccine course now 🎉 

# 🧪 Testing

- [ ] Run unit tests
- [ ] Create some vaccinations for a patient for a specific dose/course
- [ ] Now edit that course
- [ ] Even delete the dose
- [ ] You can save the vaccine course fine
- [ ] Vaccination records are still there :)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/vaccine_course/vaccine_course_dose.rs
+++ b/server/repository/src/db_diesel/vaccine_course/vaccine_course_dose.rs
@@ -113,6 +113,10 @@ fn create_filtered_query(filter: Option<VaccineCourseDoseFilter>) -> BoxedVaccin
             vaccine_course_dose_dsl::vaccine_course_id
         );
     }
+
+    // Filter out deleted rows
+    query = query.filter(vaccine_course_dose_dsl::deleted_datetime.is_null());
+
     query
 }
 
--- a/server/repository/src/db_diesel/vaccine_course/vaccine_course_dose.rs
+++ b/server/repository/src/db_diesel/vaccine_course/vaccine_course_dose.rs
@@ -28,6 +28,7 @@ pub struct VaccineCourseDose {
 pub struct VaccineCourseDoseFilter {
     pub id: Option<EqualFilter<String>>,
     pub vaccine_course_id: Option<EqualFilter<String>>,
+    pub include_deleted: Option<bool>,
 }
 
 impl VaccineCourseDoseFilter {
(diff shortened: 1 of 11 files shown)


CarlosNZ:
First thing I noticed: the deleted course doses don't appear in the Vaccination Modal any more, even though they still appear in the card list:
Screenshot 2024-09-25 at 2 08 18 PM

Also, wondering -- obviously we want to display the previously existing items in the Vax card list once the vaccination record exists. But for doses they haven't had yet, should we still display them? Are patients supposed to be able receive doses that have been removed from the course?

CarlosNZ (changes requested):
Looks good as far as the delete logic goes. Just need to make sure the vaccination card can still display the details for the deleted doses.

lache-melvin:
Ah great point! Yeah I think existing ones should continue to be visible, but shouldn't be able to create new vaccinations. There might be an edge case where they delete a dose, let's say "dose 2", and create a new dose also called "dose 2"... the cards would have 2 lines in them both called "dose 2", one given and one blank... not much we can do in that case though, and I think pretty low probability

lache-melvin:
Thanks for catching Carl!

- Only show deleted dose on card if there is an existing vaccination: [69b36a7](https://github.com/msupply-foundation/open-msupply/pull/4868/commits/69b36a79bcb8ba6d711f261c155e265fedbd690a)
- Display modal correctly: [3fb9567](https://github.com/msupply-foundation/open-msupply/pull/4868/commits/3fb95673c9ea28c831f325fa37ffd5e6b3c34f01)

CarlosNZ (approved):
Great, thanks for the fixes :)

PR #4876, opened 2024-09-25
Map and display update prescription stock not available error

### LINKED ISSUE #4872: No error displayed when changing dispensing date, and no stock available.
## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

andreievg:
Fixes #4872

Display an error toast when trying to back date prescription to a date where stock is not available

We've discussed this as a very edge case, the main use case for back dating prescriptions is when you do have stock but forgot to enter it (this is also indicated by Gary saying we don't need to do back dated inbound shipments). Because of this we though just a toast for now will do

I am not very happy with the implementation but didn't have time to dig into the abstraction (which I think would need to be removed, or understood enough to know how it can be modified, and not affect all the places it's used)

# 🧪 Testing

Create prescription
Add some stock
Try to back date prescription to be before the stock was introduced, see error

# 📃 Documentation

  1.
  2.


jmbrunskill (approved):
Works like a charm

PR #4877, opened 2024-09-25
4864 back date UI

### LINKED ISSUE #4864: Backdating prescription UI
A tweak to the UI please.
The date field is floating in the middle of the toolbar area, not vertically aligned with the other inputs:

image

Can we adjust so that the inputs align, and bring the date input closer to the others

image

In addition to the UI move above, I'd also like the following changes:

I'd also like to default the value to today's date - leaving the input empty invites the user to enter something, which would be the wrong intention. Defaulting to today gives the user a clue that the date is the date of the prescription and therefore will be today's date by default.

Additional thoughts
* maybe move it to the 'more info' panel?
* if so, shift clinician up to make the toolbar smaller vertically
* rename? The `Date` label isn't telling me what the intention is. Perhaps if we move to the info panel we could add some help text?

jmbrunskill:
Fixes #4864

# 👩🏻‍💻 What does this PR do?

Aligns datepicker with rest of the inputs in dispensing.

image

Defaults `prescription date` to created datetime

Sets picked_date to prescription date if we change the date time for a prescription (by setting it at verified time)

Disables date input after verified

Do we still want the prescription date in the more tab?

# 🧪 Testing

- [ ] Start dispensing, check date picker shows up correctly..

# 📃 Documentation

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
@@ -39,7 +39,13 @@ export const Toolbar: FC = () => {
         alignItems="flex-end"
       >
         <Grid item display="flex" flex={1}>
-          <Box display="flex" flex={1} flexDirection="column" gap={1}>
+          <Box
+            display="flex"
+            flex={1}
+            flexDirection="column"
+            gap={1}
+            maxWidth={'fit-content'}
+          >
             {patient && (
               <InputWithLabelRow
                 label={t('label.patient')}
@@ -69,7 +75,12 @@ export const Toolbar: FC = () => {
               }
             />
           </Box>
-          <Box display="flex" flexDirection="column" flex={1}>
+          <Box
+            display="flex"
+            flexDirection="column"
+            flex={1}
+            marginLeft={'8px'}
+          >
             <InputWithLabelRow
               label={t('label.date')}
               Input={
--- a/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
@@ -85,6 +85,7 @@ export const Toolbar: FC = () => {
               label={t('label.date')}
               Input={
                 <DateTimePickerInput
+                  disabled={isDisabled}
                   value={DateUtils.getDateOrNull(prescriptionDate)}
                   format="P"
                   onChange={async prescriptionDate => {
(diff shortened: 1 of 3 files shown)


lache-melvin:
This is to go into 2.3 right?

jmbrunskill:
> This is to go into 2.3 right?

Updated...

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
```diff
@@ -69,11 +75,17 @@ export const Toolbar: FC = () => {
               }
             />
           </Box>
-          <Box display="flex" flex={1}>
+          <Box
+            display="flex"
+            flexDirection="column"
+            flex={1}
+            marginLeft={'8px'}
+          >
```
mark-prins, line 83:
```suggestion
          <Box display="flex" flexDirection="column" flex={1} marginLeft={3}>
```

two things - we need to remove hard coded pixel values for spacing and fonts in order to support responsive UI and I think the date field is too close to the patient field, given the gap between `Date:` and the input: it's easy to confuse the label & input association

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
```diff
@@ -39,7 +39,13 @@ export const Toolbar: FC = () => {
         alignItems="flex-end"
       >
         <Grid item display="flex" flex={1}>
-          <Box display="flex" flex={1} flexDirection="column" gap={1}>
+          <Box
+            display="flex"
+            flex={1}
+            flexDirection="column"
+            gap={1}
+            maxWidth={'fit-content'}
```
mark-prins, line 47: nice!

mark-prins (approved):
looks great! thanks James 🙏 
a small tweak request, as commented.
and a query : for existing prescriptions, I understood from Andrei's comment on the issue that the created date would be used to populate the date input. I'm not seeing it in this PR though. Is it something that might happen?

jmbrunskill:
> looks great! thanks James 🙏 a small tweak request, as commented. and a query : for existing prescriptions, I understood from Andrei's comment on the issue that the created date would be used to populate the date input. I'm not seeing it in this PR though. Is it something that might happen?

I'm just assigning this at the point where the prescription is created.
So there is an edge case where any prescriptions created before the update don't get a prescription date assigned.
I think this is ok, so will merge...

@mark-prins if you're not happy, can re-open...

mark-prins:
ah right. yes - that's what I was testing for : editing existing prescriptions, which have an empty date field.  I was expecting that the date would be populated with the picked date of the prescription, from my reading of the comment on the issue : "And the reason to also use picked date, is because otherwise todays date would be shown on historic prescriptions ?"

PR #4878, opened 2024-09-25
4852 user friendly errors for trying to give/ un give doses in the wrong order

### LINKED ISSUE #4852: User friendly errors for trying to give/"un-give" doses in the wrong order
## What went wrong? 😲

The user can access the edit modal after a vaccination has been given. Let's say we have given 3 doses. Then I go to dose 1 and try to change to not given. 

The system does not allow this (they should change dose 3 to not given - and adjust other vaccination dates if needed - because they user is actually trying to say that the patient has only received 2 doses).

However, seeing as this is a valid error the user can cause, this should be a structured error, with a translatable error message, rather than this:

Screenshot 2024-09-24 at 3 27 34 PM

Also sometimes getting the double snack, would be great to fix here too!

## Expected behaviour 🤔

Add structured error for NotMostRecentGivenDose and translate on frontend.

## How to Reproduce 🔨

1. Follow steps above

## Your environment 🌱

- Open mSupply Version: v2.3.0 develop

- Platform:

- Database type:

roxy-dao:
Fixes #4852

# 👩🏻‍💻 What does this PR do?
Turn NotMostRecentGivenDose into a structured error for tranlsation.

https://github.com/user-attachments/assets/f53a2c99-2c27-462b-aecb-d629a6ed7144

# 🧪 Testing

- [ ] Set up vaccinations
- [ ] Enrol patient into program and create encounter
- [ ] Give 3 doses of some vaccine
- [ ] Change first dose to 'Vaccine not given`
- [ ] Click ok. See toast

# 📃 Documentation

  1.
  2.


client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -78,11 +78,20 @@ export const VaccinationModal = ({
...
+      }
+
+      if (result?.__typename === 'UpdateVaccinationError') {
+        if (result.error.__typename === 'NotMostRecentGivenDose') {
+          const errorSnack = error(t('error.not-most-recent-given-dose'));
+          errorSnack();
+        }
+      }
     } catch (e) {
-      error(t('error.failed-to-save-vaccination'))();
```
roxy-dao, line 85: Removes the extra toast

lache-melvin (approved):
Mean thank you 🔥

Issue #4779, opened 2024-09-12
Standardise the "Chip"/Badge component


CarlosNZ:
## The suggestion

For vaccination card, we created a Badge component for vaccination status, as per [the UI designs](https://github.com/msupply-foundation/open-msupply/issues/4663):

![360298251-3a602ca5-9cf7-477d-8d69-675f5dcc79e8](https://github.com/user-attachments/assets/0dcb76bb-209d-4616-a286-576b429a6f05)

However, there is already a slightly different type of Badge used in Cold chain: 
![Screenshot 2024-09-12 at 12 36 31 PM](https://github.com/user-attachments/assets/a85d436f-e408-4026-9a60-7e9e2999ce99)

The one in cold chain uses a component that is within the "cold-chain" package, so presumably not intended for use outside that.

I think we should agree on a standard Badge component then re-factor to use the same one everywhere for UI consistency. The one I made for Vax card is intentionally generic (it takes a `statusMap` prop to specify the colors/labels based on the input), so would be easy to re-use.

### Why should we invest time in this?

UI consistency, reduce code duplication.

### Are there any risks associated with this change?

Not that I can think of.

### How much effort is required?

Probably just a few hours.

### Agreed Solution

TBD, but based on the [Vaccine card designs](https://github.com/msupply-foundation/open-msupply/issues/4663), something like that is probably where we're going.

mariyamsupply:
Agree, suggest we use created for vaccination status **Badge component**

PR #4881, opened 2024-09-25
#4805 Vaccination card statuses

### LINKED ISSUE #4805: Create vaccine card status values in backend

Currently we are calculating this on frontend, but should be calculated on backend and included in card data to the frontend, as an enum that we can match on and render the correct chip:

Statuses: Given, not given, pending, late.

Screenshot 2024-09-18 at 8 50 44 AM

Screenshot 2024-09-18 at 8 51 41 AM

- Given: vaccination.given = true
- Not Given: vaccination.given = false
- Pending: No vaccination yet, but the patient is younger than max age for dose
- Late: No vaccination yet, and the patient is older than max age for dose
- Otherwise, no status

CarlosNZ:
Fixes #4805

(In Draft cos there's one Rust compilation error in the GraphQL layer that I can't quite resolve)

# 👩🏻‍💻 What does this PR do?

- Returns "status" from server as an enum value "Given", "NotGiven", "Pending", "Late" or `null`, based on patient age/vax status etc.
- Front-end maps this enum value to appropriate display chip instead of just the boolean `given` field

See inline comments

# 🧪 Testing

- [ ] Vax card table displays the correct "Status" chip based on the rules in the [original issue](#4805)
- [ ] Colours for each status are appropriate and correct

### DIFF
--- a/client/packages/common/src/styles/theme.ts
+++ b/client/packages/common/src/styles/theme.ts
@@ -95,6 +95,12 @@ declare module '@mui/material/styles/createPalette' {
       hot: PaletteColor;
       lines: string[];
     };
+    vaccinationStatus: {
+      given: string;
+      notGiven: string;
+      pending: string;
+      late: string;
+    };
     drawerDivider: string;
     gray: PaletteColor & { pale: string };
     outline: Palette['primary'];
@@ -196,7 +202,6 @@ export const themeOptions = {
       white: '#fff',
       success: 'rgb(237, 247, 237)',
     },
-
     form: {
       field: '#555770',
       label: '#28293d',
@@ -214,6 +219,12 @@ export const themeOptions = {
       notFunctioning: '#de0001',
       notInUse: '#b0b0b0',
       text: '#fff',
+    },
+    vaccinationStatus: {
+      given: 'success.light',
+      notGiven: 'error.main',
+      pending: 'cceStatus.notInUse',
+      late: 'cceStatus.functioningButNeedsAttention',
     },
   },
   zIndex: {
--- a/client/packages/common/src/styles/theme.ts
+++ b/client/packages/common/src/styles/theme.ts
@@ -223,8 +223,8 @@ export const themeOptions = {
     vaccinationStatus: {
       given: 'success.light',
       notGiven: 'error.main',
-      pending: 'cceStatus.notInUse',
-      late: 'cceStatus.functioningButNeedsAttention',
+      pending: 'info.light',
+      late: 'error.main',
     },
   },
   zIndex: {
--- a/server/graphql/types/src/types/program/vaccination_card.rs
+++ b/server/graphql/types/src/types/program/vaccination_card.rs
@@ -7,9 +7,21 @@ use graphql_core::{
     loader::{NameByIdLoader, NameByIdLoaderInput, StockLineByIdLoader},
     ContextExt,
 };
-use service::vaccination::get_vaccination_card::{VaccinationCard, VaccinationCardItem};
+use serde::Serialize;
+use service::vaccination::get_vaccination_card::{
+    VaccinationCard, VaccinationCardItem, VaccinationCardItemStatus,
+};
 
 use crate::types::StockLineNode;
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum VaccinationCardItemNodeStatus {
+    Given,
+    NotGiven,
+    Pending,
+    Late,
+}
 
 pub struct VaccinationCardNode {
     pub vaccination_card: VaccinationCard,
@@ -87,6 +99,13 @@ impl VaccinationCardItemNode {
     pub async fn suggested_date(&self) -> &Option<NaiveDate> {
         &self.item.suggested_date
     }
+    pub async fn status(&self) -> &Option<VaccinationCardItemNodeStatus> {
+        &Some(VaccinationCardItemNodeStatus::Given)
+        // &self
+        //     .item
+        //     .status
+        //     .map(|status| VaccinationCardItemNodeStatus::from_domain(&status))
+    }
 
     pub async fn stock_line(&self, ctx: &Context<'_>) -> Result<Option<StockLineNode>> {
         let loader = ctx.get_loader::<DataLoader<StockLineByIdLoader>>();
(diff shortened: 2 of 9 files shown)


client/packages/common/src/styles/theme.ts
```diff
@@ -215,6 +220,12 @@ export const themeOptions = {
       notInUse: '#b0b0b0',
       text: '#fff',
     },
+    vaccinationStatus: {
+      given: 'success.light',
+      notGiven: 'error.main',
+      pending: 'cceStatus.notInUse',
+      late: 'cceStatus.functioningButNeedsAttention',
+    },
```
CarlosNZ, line 228: Thought it was easier to define these explicitly, since the CCE stuff is. Fortunately, we can refer to existing colours within this!

server/graphql/types/src/types/program/vaccination_card.rs
```diff
@@ -87,6 +99,13 @@ impl VaccinationCardItemNode {
     pub async fn suggested_date(&self) -> &Option<NaiveDate> {
         &self.item.suggested_date
     }
+    pub async fn status(&self) -> &Option<VaccinationCardItemNodeStatus> {
+        &Some(VaccinationCardItemNodeStatus::Given)
+        // &self
+        //     .item
+        //     .status
+        //     .map(|status| VaccinationCardItemNodeStatus::from_domain(&status))
+    }
```
CarlosNZ, line 108: @lache-melvin this is the line I can't quite get working. Commented out it what I'm trying to do. 

Compiler suggests:
```
cannot move out of `self.item.status` which is behind a shared reference
help: consider calling `.as_ref()` or `.as_mut()` to borrow the type's contents
```
But when I use `as_ref()`, the `#Object` macro above complains 🤷‍♂️

server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -18,8 +18,19 @@ pub struct VaccinationCard {
...
 
+#[derive(Clone, Debug)]
+pub enum VaccinationCardItemStatus {
+    Given,
+    NotGiven,
+    Pending,
+    Late,
+}
+
+const MONTHS_PER_YEAR: f64 = 365.25 / 12.0;
```
CarlosNZ, line 32: Since we're re-using it. Do we have a global place to store constants we might want to refer to elsewhere?

server/graphql/types/src/types/program/vaccination_card.rs
lache-melvin: Ah yeah, they're a bit pedantic 😅 This should work:

```rust
// remove & from return type, because the reference is created within this fn (the reference created by .as_ref() wouldn't live beyond the end of this fn call, so need to return the option for the caller to own)
pub async fn status(&self) -> Option<VaccinationCardItemNodeStatus> { 
  // so remove & here too
  self.item
    .status
    // take a reference of item status (could also clone, but more expensive)
    .as_ref()
    // status here is now &VaccinationCardItemStatus :)
    .map(|status| VaccinationCardItemNodeStatus::from_domain(status))
}
```

server/service/src/vaccination/get_vaccination_card.rs
lache-melvin: yes! `server/util/src/constants.rs` 🔥

server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -123,6 +137,36 @@ pub fn get_suggested_date(
     }
 }
 
+pub fn get_vaccination_status(
+    row: &VaccinationCardRow,
+    patient_dob: Option<NaiveDate>,
+) -> Option<VaccinationCardItemStatus> {
+    if row.given == Some(true) {
+        Some(VaccinationCardItemStatus::Given);
```
lache-melvin, line 145: all these early return `Some()` values need a `return` before them 😁

server/service/src/vaccination/get_vaccination_card.rs
lache-melvin: This is `DAYS_PER_MONTH` though right?

server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -123,6 +137,36 @@ pub fn get_suggested_date(
...
+
+    let patient_age_in_months =
+        ((Local::now().date_naive() - patient_dob.unwrap()).num_days() as f64) / MONTHS_PER_YEAR;
+
+    if row.given == None {
+        if patient_age_in_months < row.max_age {
+            Some(VaccinationCardItemStatus::Pending);
+        };
+        if patient_age_in_months > row.max_age {
+            Some(VaccinationCardItemStatus::Pending);
```
lache-melvin, line 163: Should be `Late`

server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -123,6 +137,36 @@ pub fn get_suggested_date(
...
+    if row.given == Some(false) {
+        Some(VaccinationCardItemStatus::NotGiven);
+    }
+
+    if patient_dob == None {
+        return None;
+    }
+
+    let patient_age_in_months =
+        ((Local::now().date_naive() - patient_dob.unwrap()).num_days() as f64) / MONTHS_PER_YEAR;
```
lache-melvin, line 156: Two things here! 

All our backend date calculations should use Utc, so dates are consistent when we sync.

For the `paitient_dob` - while it should be safe given you have done the `None` check above, we pretty strictly don't use `.unwrap` in production code, as it could panic if it fails. If you need to unwrap something, it's recommended to use `.unwrap_or(..)` to provide a fallback. There's a number of other ways around it e.g. `match` statement, `.map()`, `if let Some()`.. example below!

server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -123,6 +137,36 @@ pub fn get_suggested_date(
...
+        if patient_age_in_months < row.max_age {
+            Some(VaccinationCardItemStatus::Pending);
+        };
+        if patient_age_in_months > row.max_age {
+            Some(VaccinationCardItemStatus::Pending);
+        };
+    }
+
+    None
+}
```
lache-melvin, line 168: Here's an alternative implementation... take what you like!

```rust
pub fn get_vaccination_status(
    row: &VaccinationCardRow,
    patient_dob: Option<NaiveDate>,
) -> Option<VaccinationCardItemStatus> {
    match row.given {
        // slightly more concise version of the `if` blocks you have
        Some(true) => return Some(VaccinationCardItemStatus::Given),
        Some(false) => return Some(VaccinationCardItemStatus::NotGiven), 
        None => {}
    }
    
    // If patient_dob is Some, take that value and use it inside the block
    if let Some(patient_dob) = patient_dob {
        let patient_age_in_months =
            ((Utc::now().date_naive() - patient_dob).num_days() as f64) / MONTHS_PER_YEAR;

        if patient_age_in_months < row.min_age {
            Some(VaccinationCardItemStatus::Pending)
       // shouldn't need the other `if`
        } else {
            Some(VaccinationCardItemStatus::Late)
        }
    // If patient_dob is None, return None
    } else {
        None
    }
}
```

lache-melvin (commented):
Looks great! Have left some suggestions for the backend part 🙏

CarlosNZ (commented):
Thanks for the suggestions, @lache-melvin .

Have updated the front-end too to only show the next status label as we discussed.

client/packages/common/src/styles/theme.ts
lache-melvin: The screenshots in the issue are from the designs...

lache-melvin: I think all the colours we need are already in the theme (success, error, secondary)? So can probably just use those directly in the vaccine card component, rather than another layer of indirection here...

Otherwise, I think something like `chip` in the theme would be better, and pull colours from there, so can be reused. But yes, happy not to refactor CCE stuff right now 🙏

CarlosNZ: > Otherwise, I think something like `chip` in the theme would be better, and pull colours from there

Yup, will definitely do this as well as get some definitive badge colour patterns established when we do https://github.com/msupply-foundation/open-msupply/issues/4779

server/service/src/vaccination/get_vaccination_card.rs
lache-melvin: Oo great to know, thank you! Yeah when working with months, I think we're pretty safe 😁

lache-melvin (commented):
Works a treat thanks Carl! If you could just change the badge colours to match the designs (pending blue, late in red) then good to go!

CarlosNZ:
> Works a treat thanks Carl! If you could just change the badge colours to match the designs (pending blue, late in red) then good to go!

Okay, @lache-melvin, updated as discussed with Maria. Can revisit the colours more generally in the re-factor issue :)

add12fb3ff

lache-melvin (approved):
Epic, thanks for changes Carl 🙏

PR #4887, opened 2024-09-26
4883 create return when not this facility

### LINKED ISSUE #4883: Vaccine card: customer return not created when changing this facility -> other
## What went wrong? 😲

Can't reverse the prescription when changing to another facility 😢 

## Expected behaviour 🤔

If you check the `reverse prescription` toggle... it reverse!

## How to Reproduce 🔨

1. Go to a vaccination that was `given` at this facility (that has a prescription created)
2. Change facility to `Other`
3. Toggle the `Revert existing stock transaction` ON
4. Save
5. BUG: Reopen - still shows prescription at the top of the vaccination, no customer return was created

## Context

I think this is from the change to allow `given` vaccinations with no associated stockline... stockline might need to be NullableUpdate now? Currently, if given == true and stockline == null, we keep the existing stockline!

## Your environment 🌱

- Open mSupply Version: 2.3 RC1

- Platform:

- Database type:

lache-melvin:
Fixes #4883

# 👩🏻‍💻 What does this PR do?

Ensures invoices are updated when facility is changed from this facility to other.

Also fixes the `record historic transaction` toggle showing up when it shouldn't

# 🧪 Testing

- [ ] Go to a vaccination that was given at this facility (that has a prescription created)
- [ ] Change facility to Other
- [ ] Toggle the Revert existing stock transaction ON
- [ ] Save
- [ ] prescription no longer shows at the top of the vaccination, customer return was created

ALSO
- [ ] Create vaccination at Other facility
- [ ] Enter for date in the past
- [ ] Set to given
- [ ] Ensure no toggle shows up to `record historic stock transaction`

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Vaccination/api/useVaccination.ts
+++ b/client/packages/system/src/Vaccination/api/useVaccination.ts
@@ -209,6 +209,8 @@ const useUpdate = (vaccinationId: string | undefined) => {
       throw new Error(t('error.failed-to-save-vaccination'));
     }
 
+    const isOtherFacility = input.facilityId === OTHER_FACILITY;
+
     const apiResult = await api.updateVaccination({
       storeId,
       input: {
@@ -218,17 +220,15 @@ const useUpdate = (vaccinationId: string | undefined) => {
         clinicianId: setNullableInput('id', input.clinician),
         comment: input.comment,
         notGivenReason: !input.given ? input.notGivenReason : null,
-        stockLineId: input.given ? input.stockLine?.id : null,
+        stockLineId: {
+          value: input.given && !isOtherFacility ? input.stockLine?.id : null,
+        },
 
         facilityNameId: {
-          value:
-            input.facilityId === OTHER_FACILITY ? undefined : input?.facilityId,
+          value: isOtherFacility ? undefined : input?.facilityId,
         },
         facilityFreeText: {
-          value:
-            input.facilityId === OTHER_FACILITY
-              ? input.facilityFreeText
-              : undefined,
+          value: isOtherFacility ? input.facilityFreeText : undefined,
         },
 
         updateTransactions: input.createTransactions,
--- a/client/packages/system/src/Vaccination/api/useVaccination.ts
+++ b/client/packages/system/src/Vaccination/api/useVaccination.ts
@@ -156,6 +156,8 @@ const useInsert = ({
   const mutationFn = async (input: VaccinationDraft) => {
     if (!encounterId) return;
 
+    const isOtherFacility = input.facilityId === OTHER_FACILITY;
+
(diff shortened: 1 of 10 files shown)


roxy-dao (changes requested):
Thanks Lache~

The return does get created for me but the prescription link is still showing on top for me. 
Screenshot 2024-09-27 at 9 25 04 AM

Also, shouldn't a prescription be created at a store if the user accidentally enters some facility and changes it to an mSupply facility?

lache-melvin:
Yeah expected behaviour, I'll double check it all now 🫠:

- Give at this facility: creates a prescription
- Change to `other` (or to not given): creates return, removes link to prescription
- Change back to given at this facility: creates new prescription, adds link to it
- Change the item/batch: creates return for old batch and new prescription

server/service/src/vaccination/update/generate.rs
```diff
@@ -62,44 +64,53 @@ pub fn generate(
...
-
-            let create_return = InsertCustomerReturn {
-                id: uuid(),
-                other_party_id: patient_id.clone(),
-                is_patient_return: true,
-                outbound_shipment_id: None,
-                customer_return_lines: vec![CustomerReturnLineInput {
+    let create_customer_return = if (stock_line_has_changed || should_revert) && update_transactions
+    {
+        existing_prescription_line
```
lache-melvin, line 72: added in prescription line - realised there were a couple MORE cases:

- If you give (and create prescription) -> change batch and _don't_ update transactions -> un-give and _do_ revert transactions, it was creating a return of the new batch, rather than what was in the prescription!
- Also needed to handle give (create prescription) -> change to Other facility (which clears vaccination `stock_line`) but don't revert prescription -> reopen modal and click revert... so needed to check if the prescrizzle existed rather than if the vaccination stock line was changing!

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -218,17 +220,15 @@ const useUpdate = (vaccinationId: string | undefined) => {
         clinicianId: setNullableInput('id', input.clinician),
```
roxy-dao, line 220: Should Clinician id also be nulled if isOtherFacility?

server/service/src/vaccination/update/generate.rs
roxy-dao: can use and_then instead of map.flatten (:

client/packages/system/src/Vaccination/api/useVaccination.ts
lache-melvin: Oo great catch, thank you!

roxy-dao (approved):
Thanks for the changes Lache (:

PR #4888, opened 2024-09-26
4704 create locally hosted GitHub actions runner

### LINKED ISSUE #4704: Create Locally hosted Github Actions runner
## The suggestion

Setup a github action runner on the mac mini we have already purchased for the office.

## Example use case

In the last 2 OMS Retrospectives, we've raise the issue that we don't always have confidence Postgres & Other tests have been run on every branch and merge.

Using Github Actions Runners in the cloud could work but we have issues with space & ability to setup & install software such as postgres.

See: https://github.com/msupply-foundation/open-msupply/issues/2900

- [ ] Time spent changing features, recompiling postgres and running tests slows down developer workflows
- [ ] Existing CI processes are broken causing people to bypass output potentially important checks.

Nothing significant that I can see?

I'd guess this is a couple of days work, plus some related issues such as adding the postgres, integration test, and so forth

### Agreed Solution

Github runner is setup and ready to run more specific workflows from CI.

### LINKED ISSUE #4706: Setup postgres tests in CI
## The suggestion
Following on from #4704 we can then setup a Postgres server to run tests against in github actions

This reduces the risk of postgres related issues not being pickup during development or code review.
Could save time for devs running postgres tests for every PR if they know they can wait for CI to complete the test.

Not really

1-2 days work

### Agreed Solution

Might be possible to have two checked out versions of the app, or different compile cache directories? So we don't need to recompile the full app for both SQLite and Postgres every time we commit?
Would be more work to setup but could be much faster?

jmbrunskill:
Fixes #4704 
Fixes #4706

# 👩🏻‍💻 What does this PR do?

Fixes 

# 🧪 Testing

# 📃 Documentation

  1.
  2.

### DIFF
--- a/.github/workflows/build-mac-demo.yaml
+++ b/.github/workflows/build-mac-demo.yaml
@@ -2,17 +2,17 @@ on:
   workflow_dispatch:
     inputs:
       architecture:
-        description: 'Architecture'
+        description: "Architecture"
         required: true
-        default: 'intel'
+        default: "intel"
         type: choice
         options:
           - intel
           - arm
       includeDemoData:
-        description: 'Include demo data'
+        description: "Include demo data"
         required: true
-        default: 'false'
+        default: "false"
         type: choice
         options:
           - false
@@ -23,7 +23,7 @@ name: Build mac demo
 jobs:
   build_and_test:
     name: Build mac demo
-    runs-on: macos-latest
+    runs-on: self-hosted
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


jmbrunskill:
Some times getting this

`libunwind: stepWithCompactEncoding - invalid compact unwind encoding`

Will investigate, sound like upgrading xcode might help...

lache-melvin (approved):
Looks great to me, thanks James! Yeah I think let's get it in and iterate from there 🔥 

> I think maybe we could add a manual workflow to trigger the clean up?

Yeah sounds like the cargo clean step would be likely needed at some point, but happy to add that in when the issue arises?

> Do we want to leave other workflows on github e.g. bundle size, client tests, I'm currently thinking yes but open to suggestions (under new issue I think)

Yep these run fairly quick, I think leaving them as is makes sense 🙏

Issue #4891, opened 2024-09-26
Update stocktake price column headers


mark-prins:
## What went wrong? 😲

In French, the headers have 'cost price' and 'sell price' and don't indicate whether this is a pack or unit price.
English is better, though we should improve 'cost' too:

<img width="551" alt="image" src="https://github.com/user-attachments/assets/3f2ae7ed-06a7-46d1-b756-1479848152db">

![telegram-cloud-photo-size-1-5165869931168902409-y](https://github.com/user-attachments/assets/6de97d13-728e-4158-9ee1-d999ffcecbaa)

## Expected behaviour 🤔
* Change fr to 'pack sell price' and 'pack cost price' or equivalent
* Change en to 'pack cost price'

**Update**
Have pushed the updated translation. For this issue though:

* add a new translation key for `label.pack-cost-price` for the `Pack cost price` and update the `costPricePerPack` to use this
* rename `label.sell` to be `label.pack-sell-price`
* update the inbound shipment detail view `useExpansionColumns` function to use `label.cost` - there isn't much space in the expansion area, and cost/pack relationship is clear in this case:

<img width="305" alt="image" src="https://github.com/user-attachments/assets/d7777a9d-567f-4c77-8b8f-41e62263fecf">

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

lache-melvin:
Note that if the translation string key is `label.cost` or something generic, we would want to check the same label isn't used somewhere else that we _do_ mean unit cost!

mark-prins:
It is in fact using `label.cost` but this is the only instance of it.
```
  costPricePerPack: {
    label: 'label.cost',
    key: 'costPricePerPack',
    width: 50,
    align: ColumnAlign.Right,
    Cell: CurrencyCell,
    format: ColumnFormat.Currency,
  },
```

I'll update the translation directly - would like the label to be more specific as well though. have updated the description.

roxy-dao:
Closed in #5019

roxy-dao:
The cost + sell price for above is for the pack size

roxy-dao:
Can do... though to me the stock page implies 'price per pack' maybe not obvious enough? @mark-prins what do you think?

mark-prins:
[...]

Issue #4673, opened 2024-08-22
Document translation for release patches


andreievg:
## Is your feature request related to a problem? Please describe 👀

Since translations will now go to the develop branch (from translation server), I can see a use case where translations are added in translations server during testing/release phase, or as part of patch, would be great to document a process of how these translations should be pulled through to release safely (especially when translation keys have changed in develop). 

As part of this would also be great to document what is the process of safely changing translation keys (how this works with translation server)

### Additional context 💌

https://wiki.sussol.net/doku.php/msupply:support:important_stuff:translation_server_hosting?s[]=translation

andreievg:
The issue I was thinking relates to when changes need to be done in weblate but are needed in patch.

Dhanya suggested similar issues with new columns etc (only going to master, table changes)

andreievg:
Closing, to be done as part of: https://github.com/msupply-foundation/open-msupply/issues/4060

Issue #4060, opened 2024-06-06
Organise translation namespaces


mark-prins:
## The suggestion

The current organisation of translation namespaces might not be entirely fit for purpose and it would be good to review how it is working and if we can improve.

The original intention was to divide translation strings into functional areas, so that the app would only need to download strings for the area that the user is working in, reducing the overall download size. 
I would still strongly recommend splitting the strings into namespaces, and I also prefer having the translation keys prefixed in a way to divide them e.g. `button.save` `settings.label-printing`

However.. given that the JSON files are cached, perhaps the huge initial download isn't critical? In which case we could split across tasks e.g.

`buttons.json`
`errors.json`

even though this would require editing multiple translation files while working on a feature.

Here are the current namespaces:

| namespace | description |
| ------------| ----------- |
| app | Application-wide strings such as menu items and keyboard shortcuts |
| common | Strings which are used in more that one area and strings which are called from the `common` package by shared components |
| dashboard | Dashboard-specific strings  |
| distribution | Anything used solely in  the distribution area: Outbound shipments, Requisitions &c. |
| inventory | Inventory only strings  |
| replenishment | Replenishment only strings  |
| catalogue | Catalogue only strings  |
| dispensary | Dispensary only strings  |
| programs | Programs specific: the HIV program work for PNG  |
| coldchain | All things coldchain, including the assets module now  |
| plugin | A separate space for plugin specific strings  |
| manage | The new central server-only administration aspects; for centrally managed things e.g. `name`s for GAPS |

### Why should we invest time in this?

There's possibly a small overhead in developer time with the current split. I have heard complaints, though nothing tangible enough to flesh out this issue. There may also be a small improvement in app responsiveness if we get the split perfect, and conversely, a decrease in performance if we prioritise developer's needs.

A concern would be having a single file, as this would increase the incidence of development conflicts. Splitting by functional areas is a sensible way to remove this risk.

### Are there any risks associated with this change?
* There is a small risk we decrease performance
* a risk we confuse developers with the change
* a possible increase in development conflicts

### How much effort is required?
Probably not much 🤷 
The development change is minor, perhaps half a day to move strings around.
Deciding what to do, that will take days.

jmbrunskill:
One problem I have with the breaking down of things into `buttons.add-name` and `label.add-name` is that I feel like we get a lot of duplication. Maybe we should have a difference between a button text and label text, I currently don't feel clear on it.
[...]

PR #4902, opened 2024-09-29
4089 ledger table change name 'quantity' to 'units'

### LINKED ISSUE #4089: Ledger table: change name 'Quantity' to 'Units'
## What went wrong? 😲

On the ledger tab the column 'quantity' means number of Units.

Without any specification it can be confusing as to whether means 'Packs' or 'Units'.

telegram-cloud-photo-size-1-5132363654342225164-y

## Your environment 🌱

Desktop app v2.1.0-IA-test-2

roxy-dao:
Fixes #4089

# 👩🏻‍💻 What does this PR do?
Changed quantity in ledger to unit quantity... 
Screenshot 2024-09-30 at 8 58 12 AM

The quantity in Returns is being used in the stepper, so I thought it was fine to leave it but idk? 

# 🧪 Testing

- [ ] Go to ledger
- [ ] See `Unit Qty` instead of quantity

# 📃 Documentation

  1. Screenshots for ledger
  2.


client/packages/common/src/intl/locales/en/distribution.json
```diff
@@ -66,7 +66,6 @@
   "label.outbound-shipment": "Outbound shipment",
   "label.placeholder": "Placeholder",
   "label.delete": "Delete",
-  "label.quantity": "Quantity",
```
roxy-dao, line 69: Removed cause it's in common

client/packages/system/src/Stock/Components/Ledger/useLedgerColumns.ts
```diff
@@ -13,7 +13,7 @@ export enum ColumnKey {
   'DateTime' = 'datetime',
   'Time' = 'time',
   'Name' = 'name',
-  'Quantity' = 'quantity',
+  'UnitQuantity' = 'unitQuantity',
```
lache-melvin, line 16: I think the value still needs to be `quantity` - ledger is showing blank values now! 😁 

Screenshot 2024-09-30 at 11 01 26 AM

lache-melvin (changes requested):
Thanks Roxy, just need to fix the display 😅 

Assuming all the extra translations you just pulled through from develop? I think its good to include them in the release, though seeing them here is reminding me we need that way to keep the RC branch up to date with translations/what to do if we change translation strings between versions...

roxy-dao:
> Thanks Roxy, just need to fix the display 😅
> 
> Assuming all the extra translations you just pulled through from develop? I think its good to include them in the release, though seeing them here is reminding me we need that way to keep the RC branch up to date with translations/what to do if we change translation strings between versions...

The translations used to go to main... but now it's going straight to develop... maybe can change it back to main to get latest translations? 👀

lache-melvin:
> The translations used to go to main... but now it's going straight to develop... maybe can change it back to main to get latest translations? 👀

I think we're sticking with develop, but there's some issues to hash out, #4060, #4673 :)

lache-melvin (approved):
Mean as TY 🚀

mark-prins:
Two reasons for using `develop` rather than `main`
1. Branch protection on `main` meant that weblate was unable to push to it without creating a PR ( there's a way around this though, which we're doing for develop, so this isn't an issue now )
2. It allows our translators to translate new functionality before release

I wouldn't want to change back to `main` due to the second point - that situation is more common than the need to pull translations into an RC / hotfix branch; which I think is what's needed here?

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -211,7 +211,7 @@
   "label.contact-tracing": "Contact Tracing",
   "label.contact-tracing-type": "Type of contact",
   "label.context": "Context",
-  "label.cost": "Cost",
+  "label.cost": "Pack cost price",
```
jmbrunskill, line 214: Shouldn't we use a more descriptive `label.cost_per_pack` to make translation easier and have less mistakes?
I remember Adam raised something about french translations not having the correct details in them?

mark-prins: yes! is requested in #4891

PR #4905, opened 2024-09-29
4832 Improve vaccination batch selection

### LINKED ISSUE #4832: Autoselect single batch
If there is only a single batch available, can we select it by default please

If there is only one batch, no need to have the batch selection modal or the onclick. simply select that batch.

lache-melvin:
Fixes #4832

# 👩🏻‍💻 What does this PR do?

Addresses the bug where selected batch didn't render if it didn't have a batch label, and adds the auto-select and all-in-one-modal requests.

- Changes batch select from a separate modal to a table in the same modal
Screenshot 2024-09-30 at 12 51 55 PM

- Autoselects the batch if there is only one
Screenshot 2024-09-30 at 12 52 06 PM

- Asks for confirmation if no batch selected
Screenshot 2024-09-30 at 12 44 49 PM

# 🧪 Testing

- [ ] Open vaccination modal, give vaccination, select an item
- [ ] Batch options are now within the same modal
- [ ] If you select an item with only one batch available, it is selected automatically
- [ ] If you select an item with multiple batches, no line is selected
- [ ] Save the modal without selecting a batch, you should get a confirmation popup

# 📃 Documentation

  1.
  2.


client/packages/programs/src/ImmunisationProgramListView/ListView.tsx
```diff
@@ -57,7 +57,6 @@ const ImmunisationProgramListComponent: FC = () => {
             ? UNDEFINED_STRING_VALUE
             : rowData.vaccineCourses?.map(n => n.name).join(', '),
       },
-      'selection',
```
lache-melvin, line 60: sorry small side quest, noticed this column while taking docs screenshots, shouldn't be there anymore, this is done in OG

CarlosNZ (approved):
Looks great, I only have one suggestion, if it's not too tricky to do:

> Autoselects the batch if there is only one

Could it also auto-select if there are multiple batches but only one has any stock?

Screenshot 2024-09-30 at 5 11 04 PM

Not sure how much extra this would be, so leave it for now if it's a bit much.

lache-melvin:
> Could it also auto-select if there are multiple batches but only one has any stock?
> 
> Screenshot 2024-09-30 at 5 11 04 PM


Ohh I think this is actually a bug! I just tested and when the stock line is actually reduced to 0, it doesn't show up there. But I think there is a rounding issue in the vax card stock transactions:

Screenshot 2024-09-30 at 5 20 52 PM

 The batch select is rounding the decimals, but the stock line you're seeing is actually _almost_ 0, rather than 0 🥲 Now that I think about it, doses should always be an integer, shouldn't need the rounding!

Will log a separate issue 🙏

PR #4906, opened 2024-09-30
#4834 Group vaccines by age

### LINKED ISSUE #4834: Group vaccine card rows by age

Currently the age shows on every row in the card, rather that just once (so as to group all the rows by age) - it should look like this:

image

The merging cells might be harder given our current styling, but this would be a good start:

image

CarlosNZ:
Fixes #4834

# 👩🏻‍💻 What does this PR do?

Quite a bit of CSS hackery here to get this all to work! The current "box-shadow" property, which is used to add the row dividers doesn't work for over-riding per cell. So I've turned off the box-shadow and replaced it with a "border", matching as best I can by eye (could probably be tweaked further, but I think it's pretty close).

Then set up a bunch of rules per-cell, and by status to make Vax table match the designs as much as possible.

Screenshot 2024-09-30 at 1 02 35 PM

At some stage, we probably want to provide a more general mechanism to do some of this table styling (ie. merged cells)

### DIFF
--- a/client/packages/common/src/ui/layout/tables/context/TableContext.tsx
+++ b/client/packages/common/src/ui/layout/tables/context/TableContext.tsx
@@ -142,7 +142,10 @@ export const createTableStore = () =>
 
         // Set new styles for the ids passed.
         ids.forEach(id => {
-          rowState[id] = getRowState(state, id, { style });
+          const currentRowStyle = rowState?.[id]?.style ?? {};
+          rowState[id] = getRowState(state, id, {
+            style: { ...currentRowStyle, ...style } as AppSxProp,
+          });
         });
 
         return { ...state, rowState: { ...rowState } };
--- a/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
+++ b/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
@@ -149,6 +149,7 @@ export const VaccinationCardComponent: FC<VaccinationCardProps> = ({
         label: 'label.age',
         sortable: false,
         accessor: ({ rowData }) => {
+          // Only show age label for first of each "block", when repeated
           const index =
             data?.items.findIndex(item => item.id === rowData.id) ?? 0;
           const sameAsPrev =
--- a/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
+++ b/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
@@ -53,26 +53,32 @@ const useStyleRowsByStatus = (
   const { setRowStyles } = useRowStyle();
   const theme = useTheme();
 
+  // This replaces the default "box-shadow", and is not an exact replacement,
+  // but pretty close. Can be refined in future.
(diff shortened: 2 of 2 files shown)


client/packages/common/src/ui/layout/tables/context/TableContext.tsx
```diff
@@ -142,7 +142,10 @@ export const createTableStore = () =>
 
         // Set new styles for the ids passed.
         ids.forEach(id => {
-          rowState[id] = getRowState(state, id, { style });
+          const currentRowStyle = rowState?.[id]?.style ?? {};
+          rowState[id] = getRowState(state, id, {
+            style: { ...currentRowStyle, ...style } as AppSxProp,
+          });
```
CarlosNZ, line 148: This allows sequential "setRowStyles" calls to update the row style rather than replace it.

mark-prins: :nit in which case the 'set' is misleading and may well cause issues later. Can you instead add an 'updateRowStyle' method? Also with a change like this - are all current usages happy about an update rather than a set?

lache-melvin (approved):
Looks amazing Carl, thank you!

Yeah the setRowStyles setup is getting pretty convoluted... I wondered whether it would have any kind of performance impact, but I think we should have a fairly limited number of lines in this table, and it's virtualised, so should be okay.

Happy merging like this, but when time allows, would it be simpler just to pass an sx prop into the `DataTable`??

Did notice this on smaller width screens, maybe should add a width to the age column?

https://github.com/user-attachments/assets/9a41221a-6019-4d3f-996e-b64209ec8cc4

CarlosNZ:
> Happy merging like this, but when time allows, would it be simpler just to pass an sx prop into the `DataTable`??

How were you imagining the conditional styling would work on a per-row basis?

I was thinking of adding a new method to useRowStyles that would do all the conditional styling in a single request, haven't given it too much thought though.

Thanks for the suggestions, 1ba2993253 @mark-prins @lache-melvin

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -73,55 +73,42 @@ const useStyleRowsByStatus = (
...
-      false
-    );
-    setRowStyles(
-      lastOfEachAgeRange,
-      {
-        '& td:first-child': {
-          borderBottom: BORDER_STYLE,
-          fontWeight: 'bold',
-        },
+      '& td:nth-child(2)': {
```
lache-melvin, line 98: Oo one more nit... I know we aren't server side rendering, so should be fine, but the could we change this to clear the console warnings?
Screenshot 2024-10-01 at 9 28 16 AM

CarlosNZ: That's an annoying warning, but fortunately easy to replace :) a89072883a

lache-melvin (approved):
Epic -thanks for changes Carl!

PR #4911, opened 2024-09-30
4900 improve stock numbers backdated

### LINKED ISSUE #4900: Backdate prescriptions - stock numbers are strange
## What went wrong? 😲

When you change the date on the prescription, then the stock numbers (available & in store) show values from the start of the date selected. This is problematic when the date is today, as any transactions made today aren't included.

This was compounded by an existing weirdness, which is that the 'in store' total is reduced by the quantity issued on the current prescription, but the 'available' is not. Thus you can have less stock in store than is available:

image

## Expected behaviour 🤔
Two changes would make things clearer:

* Change the time to 23:59 for the selected date
* Adjust the displayed value for the stock in store to be stock + quantity allocated to this prescription

The second item is how the available stock column behaves; we need to have both stock columns behave consistently

## How to Reproduce 🔨

1. Add some stock for item a
2. Create a prescription
3. Add item a
4. see that the stock which you've just added is not available: the full total shows in the 'in store' total but is not in the 'available' total

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #4900

# 👩🏻‍💻 What does this PR do?

Removes default of `allocated_date` to `created_date`
Shows today's date in the Date Picker (even though no date is specified in backend)
Defaults the time of a backdating to midnight local time (from the frontend)
Adjusts the Total Stock in store to show as the total stock in store plus what allocated in this line.
This mi required because prescription lines automatically go to `picked` but this means they are removed from the stock line, it's not then clear how the stock line can be adjusted.

Doing a separate PR for adding a `backdated_datetime` into the backend and some logic cleanup

# 🧪 Testing

- [ ] Test prescriptions with backdated prescription time and without.
- [ ] Check the total available, always shows the current level of stock in the store plus the amount on the line
- [ ] Check the available total is correct (should always be less than the total stock in store) 

# 📃 Documentation

- [X] **These areas should be updated or checked**: Prescriptions

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/useDraftPrescriptionLines.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/useDraftPrescriptionLines.tsx
@@ -78,6 +78,7 @@ export const useDraftPrescriptionLines = (
               invoiceLine,
               invoiceId,
               stockLine: batch,
+              adjustTotalNumberOfPacks: true,
             });
           } else {
             return createDraftStockOutLineFromStockLine({
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -145,6 +145,12 @@ export const DateUtils = {
   },
   startOfDay,
   startOfToday,
+  endOfDayOrNull: (date?: Date | string | number | null): Date | null => {
+    if (!date) return null;
+
+    const d = dateInputHandler(date);
+    return isValid(d) ? endOfDay(d) : null;
+  },
   endOfDay,
   startOfYear,
   previousMonday,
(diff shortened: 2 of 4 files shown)


client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/useDraftPrescriptionLines.tsx
```diff
@@ -78,6 +78,7 @@ export const useDraftPrescriptionLines = (
               invoiceLine,
               invoiceId,
               stockLine: batch,
+              adjustTotalNumberOfPacks: true,
```
jmbrunskill, line 81: I went with this as it wasn't easy to map through the status of the invoice, could be added to outbound shipments too if we wanted?

andreievg (approved):
Looks great thanks. Yeah this is evident here too: https://github.com/msupply-foundation/open-msupply/issues/4936, will ask triage team about it

PR #4914, opened 2024-09-30
4900 backdated datetime

### LINKED ISSUE #4900: Backdate prescriptions - stock numbers are strange

jmbrunskill:
Fixes #4900

# 👩🏻‍💻 What does this PR do?

This PR adds `backdated_datetime` to invoices.
The plan is to add this optional timestamp to mSupply as well.

Adding the extra field makes the purpose clearer (vs using allocated_datetime as the `backdated_datetime`).
This also makes it easier to update the other datetimes when the backdated date is changed...

Is this change too big to go into 2.3.0? Is it safe, or do we need to re-test all the prescription stuff?

There is an edge case that if someone does manually change their prescription to as today, it will be treated as backdated, even though it's kind of going into the future (aka at midnight) I think that's ok?

# 🧪 Testing

- [ ] Create a normal prescription.
- [ ] Check that the appropriate date times are set e.g, `picked_datetime` and `created_datetime` are set correctly
- [ ] Create a backdated prescription
- [ ] Check that all the datetimes are set to midnight of the date backdated  too
- [ ] Try changing the backdated date at various stages make sure the dates are updated appropriately...

# 📃 Documentation


client/packages/invoices/src/Prescriptions/api/operations.graphql
```diff
@@ -2,7 +2,7 @@ fragment PrescriptionRow on InvoiceNode {
   __typename
   comment
   createdDatetime
-  prescriptionDate: allocatedDatetime
+  prescriptionDate: backdatedDatetime
```
jmbrunskill, line 5: Kept this as prescriptionDate to minimise changes in frontend, I think it's ok how it is but maybe it's better to make it clearly backdatedDatetime in frontend?

server/service/src/invoice/prescription/insert/generate.rs
```diff
@@ -40,7 +40,7 @@ pub fn generate(
         comment: None,
         their_reference: None,
         transport_reference: None,
-        allocated_datetime: Some(current_datetime),
+        allocated_datetime: None,
```
jmbrunskill, line 43: Removed this default, so allocated_datetime won't be set now, unless backdating...

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -27,28 +27,29 @@ pub(crate) fn generate(
...
         should_update_batches_total_number_of_packs(&existing_invoice, &input_status);
     let mut update_invoice = existing_invoice.clone();
 
-    // prescription_date is None if the invoice is not backdated, this means we get actual picked/verified date
-    // If we have a backdated invoice, we use the backdated date for picked/verified date
-    let prescription_date = match prescription_date {
-        Some(prescription_date) => Some(prescription_date),
+    let backdated_datetime = match backdated_datetime {
+        Some(backdated_datetime) => {
+            backdate_status_datetimes(&mut update_invoice, backdated_datetime);
```
jmbrunskill, line 40: Now always updating datetimes if we change the backdated time.

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -95,10 +96,30 @@ fn should_update_batches_total_number_of_packs(
...
 }
 
+fn backdate_status_datetimes(invoice: &mut InvoiceRow, backdated_datetime: NaiveDateTime) {
+    match &invoice.status {
+        InvoiceStatus::New => {} // Created datetime is never backdated
+        InvoiceStatus::Allocated => {
+            invoice.allocated_datetime = Some(backdated_datetime);
+        }
+        InvoiceStatus::Picked => {
+            invoice.allocated_datetime = Some(backdated_datetime);
```
jmbrunskill, line 106: Is this safe, should we always have an allocated date? Maybe we should check if it's not null before setting it?

server/service/src/invoice_line/stock_out_line/mod.rs
```diff
@@ -35,9 +35,9 @@ impl StockOutType {
 }
 
 pub(crate) fn invoice_backdated_date(invoice: &InvoiceRow) -> Option<NaiveDateTime> {
-    if let Some(allocated_date) = invoice.allocated_datetime.clone() {
-        if allocated_date < invoice.created_datetime {
-            return Some(allocated_date);
+    if let Some(backdated_datetime) = invoice.backdated_datetime.clone() {
+        if backdated_datetime < invoice.created_datetime {
```
jmbrunskill, line 39: I think this logic is still good, if backdated time is after created_datetime, it's sort of considered not a backdated invoice, although the field is populated in the db which could cause some confusion in reporting (if that was to happen) e.g. # of backdated invoiced could be incorrect if people set it to today a lot...

client/packages/invoices/src/Prescriptions/api/operations.graphql
andreievg: Yeah I think backdateDatetime would be better for front end

server/service/src/sync/translations/invoice.rs
```diff
@@ -181,6 +181,11 @@ pub struct LegacyTransactRow {
     #[serde(rename = "om_original_shipment_id")]
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub original_shipment_id: Option<String>,
+
+    #[serde(default)]
+    #[serde(rename = "backdated_datetime")]
```
andreievg, line 186: I think shis would be om_backdated_datetime in mSupply

server/service/src/sync/translations/invoice.rs
```diff
@@ -298,6 +303,14 @@ impl SyncTranslation for InvoiceTranslation {
             }
         };
 
+        // If it was created after allocation, then it must be a backdated invoice
+        let backdated_datetime = match mapping.created_datetime
```
andreievg, line 307: Is this for backwards compatibility with RC build ? I don't think we need to support it

server/service/src/invoice_line/stock_out_line/mod.rs
andreievg: > There is an edge case that if someone does manually change their prescription to as today, it will be treated as backdated,

With this change that other edge case is dealt with ?

server/service/src/invoice/prescription/update/generate.rs
andreievg: I find it easer to think about if we first do normal update of statuses, and then adjust them by backdate (for those dates that are set)

https://github.com/msupply-foundation/open-msupply/compare/4900-backdated_datetime...4900-backdated_datetime(PR-review-suggestion)

andreievg (changes requested):
Thanks, have some small suggestions.

Also do you think we should be updating prescription datetime if you click in and out of prescription datetime, but it hasn't changed ? (like when by default it's current date, and we click in and out of it?)

server/service/src/sync/translations/invoice.rs
jmbrunskill: Was thinking for mSupply backdated, sounds like better to use picked_datetime from legacy...

server/service/src/sync/translations/invoice.rs
jmbrunskill: [2b93947](https://github.com/msupply-foundation/open-msupply/pull/4914/commits/2b9394723aac585574ed65df47b1477c7e62b5c7)

server/service/src/sync/translations/invoice.rs
jmbrunskill: [9124e6c](https://github.com/msupply-foundation/open-msupply/pull/4914/commits/9124e6c461955beee432d202b8ced76f2c01aa6f)

lache-melvin:
@jmbrunskill was going to review, but I think develop got merged in here instead of v2.3.0 😁

jmbrunskill:
@lache-melvin dang, that was me, was trying to resolve conflicts...
One tick...

server/service/src/invoice_line/stock_out_line/mod.rs
jmbrunskill: @andreievg  I think we agreed to leave this for now? I'll raise an issue for follow up though...

andreievg (approved):
Thanks @jmbrunskill, look great !

jmbrunskill:
Also added https://github.com/msupply-foundation/open-msupply/issues/4980 for follow up...

PR #4919, opened 2024-09-30
Show Average unit price in prescriptions

### LINKED ISSUE #4764: Prescription: Price is only calculated from the first line if multiple packsizes are selected.
## What went wrong? 😲

telegram-cloud-photo-size-1-5114005898967232095-y

telegram-cloud-photo-size-1-5114005898967232096-y

## Expected behaviour 🤔

Total Price should calculated correctly
If there's more than one `Unit Price` it should show `[multiple]` like other columns do

## How to Reproduce 🔨

1. Make sure you have an item in stock with different packsizes available and prices
2. Create a prescription for a patient and select stock from two different packsizes
3. Check the prices are calculated correctly.
4. Test with default prices lists and discount price lists in case this affects things

## Your environment 🌱

- Open mSupply Version: 2.2

jmbrunskill:
Fixes #4764

# 👩🏻‍💻 What does this PR do?

Previously the unit price was shown as the sum of all the pack price per line.

Now it shows the average...

image

We maybe should just show [Multiple] if there are different prices?
Should the title be updated to show average?

# 🧪 Testing

- [ ] Make sure you have an item in stock with different packsizes available and prices
- [ ] Create a prescription for a patient and select stock from two different packsizes
- [ ] Check the prices are calculated correctly.
- [ ] Test with default prices lists and discount price lists in case this affects things

# 📃 Documentation

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
+++ b/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
@@ -268,11 +268,17 @@ export const usePrescriptionColumn = ({
       Cell: CurrencyCell,
       accessor: ({ rowData }) => {
         if ('lines' in rowData) {
-          return Object.values(rowData.lines).reduce(
-            (sum, batch) =>
-              sum + (batch.sellPricePerPack ?? 0) / batch.packSize,
-            0
-          );
+          // Multiple lines, so we need to calculate the average price per unit
+
+          let totalSellPrice = 0;
+          let totalUnits = 0;
+
+          for (const line of rowData.lines) {
+            totalSellPrice += line.sellPricePerPack * line.numberOfPacks;
+            totalUnits += line.numberOfPacks * line.packSize;
+          }
+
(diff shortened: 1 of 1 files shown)


roxy-dao (changes requested):
Ooh thanks James! Much simpler than it was in my mind 😆 

Can we have this round to 2 dp with ellipsis to indicate it's longer than 2 dp and tooltip to show full price? 

Think it's fine not to show multiple for the different prices since one is average unit price and the other is total price for line...

P.S. You might want to bring this change to Outbound Shipments as well since the unit total there is wrong!

roxy-dao (approved):
Or maybe it's okay with rounded to 2dp?

roxy-dao (changes requested):
Sorry...
Screenshot 2024-10-01 at 1 40 08 PM

roxy-dao (approved):
Did changes in Outbound PR thanks James~

Issue #4928, opened 2024-09-30
Update Vaccination age range UI


CarlosNZ:
As suggested by Maria's designs, the UI for inputting Vaccination age ranges could do with some improvement. This is the preferred option:

![photo_2024-10-01 11 30 40](https://github.com/user-attachments/assets/5316f70c-14ec-4f9b-8aa5-6ee57c3d1473)

You add periods outside the table, in *either* months or years, and then **Add** them:

![photo_2024-10-01 11 34 11](https://github.com/user-attachments/assets/277cdfb5-2976-4032-93f6-a4fd510a4eed)

Another suggestion was if we kept the existing UI, to put an explanatory message above:
![photo_2024-10-01 11 30 44](https://github.com/user-attachments/assets/a5202972-5cc6-4fa0-9073-00cb7c3eb67f)

I *don't* think this is a good idea -- we shouldn't be relying on helpers to make up for unclear UI, especially if there's a clearer solution available.

The only downside of the preferred option is not being able to enter a range in both months and years, e.g. 1 year, 6 months. This would require some discussion, but it seems to me it's fine to think in "months" if you're dealing with partial years (i.e. it's common to say "18 months" for a child's age and then deal in whole years past a certain point, like 2 years old).

Needs further discussion / agreement...

andreievg:
Triage, just a suggestion to check the scope, seeing as it's one time configuration and sounds like the requirement has already been met

andreievg:
Also wasn wondering about Mariya's design time, in terms of priority, is this something that you see as quite urgent @mark-prins ?

mark-prins:
not urgent - we can get by with the current ui. I think it needs changing though - the wall of input boxes is hard to parse visually!
and I don't think the loss of functionality we get when dropping year and month entry is a big deal, not enough to worry me. I'd go with the single input for number and a drop down for month/year.

andreievg:
Another issue: https://github.com/msupply-foundation/open-msupply/issues/6102

PR #4931, opened 2024-09-30
#4884 Vax config age range improvement

### LINKED ISSUE #4884: Vaccine course - Issue filling Age Range for doses
## What went wrong? 😲

When creating a new Vaccine Course, if you try to edit the age range starting with 'From age' the field won't accept any number smaller than the 'To age'. This causes a problem when the default value for 'To age' is pre-filled with 0Y 2M.

Considering the design, the user is lead to start filling the form by 'From age', but with this restriction, for now the user needs to edit 'To age' first.

For example, if the interval is supposed to be "From age" = 5Y ", To age"= 10Y, the only way to make it work is inputting 10Y first. See video below:

https://github.com/user-attachments/assets/32bea7b8-08b4-48a7-bdca-b9437d2b0194

If not done in this order, the system will not accept the number 5Y, as the default value for 'To age' is 0Y 2M
Screenshot 2024-09-26 at 4 19 49 PM

## Expected behaviour 🤔

If 'To age' < ' From age', error message appears.

## How to Reproduce 🔨

1. Go to 'Programs' >Immunisations
2. Click on 'Add Vaccine course'
3. Scroll down to '+ Dose'
4. Try typing 'From age' = 3Y 
5. See error

## Your environment 🌱

- Open mSupply Version: V2.3.RC1
- Legacy mSupply Central Server Version: V7.17

- Platform: Desktop app

CarlosNZ:
Fixes #4884 (sort of)

# 👩🏻‍💻 What does this PR do?

All it does currently is:
- remove the min/max constraints
- set the new min default to the previous dose maximum

I had started to add some additional constraints (max must be more than min, doses can't overlap), but have stopped that for now for two reasons:

- we're probably going to update this UI anyway, as per #4928, which would make this unnecessary for now
- I encountered a limitation of the `useBufferState` hook, where it doesn't update the local state value if the external data has been *reverted* to it's existing value after a user input change. So we'd need to either change this, or find a way to pass the *full* data set into the cell so we can compare rows within the cell.

# 🧪 Testing

- [ ] Can change min age value without being constrained by the current max
- [ ] Ranges can overlap (obviously this is not what we want, but solution to that will be done elsewhere)

### DIFF
--- a/client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
+++ b/client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
@@ -276,10 +276,9 @@ const VaccineCourseDoseTable = ({
   };
 
   const updateDose: ColumnDataSetter<VaccineCourseDoseFragment> = newData => {
-    const validData = constrainInput(newData, doses);
     updatePatch({
       vaccineCourseDoses: doses.map(dose =>
-        dose.id === newData.id ? { ...dose, ...validData } : dose
+        dose.id === newData.id ? { ...dose, ...newData } : dose
       ),
     });
   };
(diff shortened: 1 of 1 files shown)


client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -261,8 +261,8 @@ const VaccineCourseDoseTable = ({
           __typename: 'VaccineCourseDoseNode',
           id: FnUtils.generateUUID(),
           label: `${courseName} ${doses.length + 1}`,
-          minAgeMonths: (previousDose?.minAgeMonths ?? 0) + 1,
-          maxAgeMonths: (previousDose?.minAgeMonths ?? 0) + 2,
+          minAgeMonths: previousDose?.maxAgeMonths ?? 0,
+          maxAgeMonths: (previousDose?.maxAgeMonths ?? 0) + 1,
```
lache-melvin, line 265: I wonder whether this would be more a helpful default... so if the first range was 0-6mnths, next would be 6-12, rather than 6-7?
```ts
  const addDose = () => {
    const previousDose = doses[doses.length - 1];

    const previousMin = previousDose?.minAgeMonths ?? 0;
    const previousMax = previousDose?.maxAgeMonths ?? 0;
    const previousRange = previousMax - previousMin;

    updatePatch({
      vaccineCourseDoses: [
        ...doses,
        {
          __typename: 'VaccineCourseDoseNode',
          id: FnUtils.generateUUID(),
          label: `${courseName} ${doses.length + 1}`,
          minAgeMonths: previousMax,
          maxAgeMonths: previousMax + (previousRange || 1),
          minIntervalDays: previousDose?.minIntervalDays ?? 30,
        },
      ],
    });
  };
```

lache-melvin (commented):
Thanks for fixing this Carl - let me know what you think of the defaults.. otherwise happy to merge 🙏

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
CarlosNZ: Yeah, sure. Hard to know for sure without knowing real-world cases, but this seems reasonable: 8cdd1b7520

lache-melvin (approved):
ship it 🚀

Issue #4918, opened 2024-09-30
Total doesn't display the correct value


marthakeezy:
## What went wrong? 😲

Numbers don't show properly in different cases, depending on what numbers get entered

<img width="781" alt="Screenshot 2024-09-30 at 5 32 36 PM" src="https://github.com/user-attachments/assets/8daa9c2a-9be8-43c5-a74b-9c0ae5f41066">

<img width="925" alt="Screenshot 2024-09-30 at 5 38 12 PM" src="https://github.com/user-attachments/assets/e950333a-2b93-429b-948d-ac6d0a7b31bb">

## Expected behaviour 🤔

They should show normally without the three dots at the end :)

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to Dispensary -> Prescriptions
2. Create a new Prescription or edit an existing one
3. Add an item -> under Pack Qty, try and add decimal numbers
4. Save the changes -> see inconsistent number display

## Your environment 🌱

- Open mSupply Version: V2.3.0-RC3
- Legacy mSupply Central Server Version: V7.17.10

- Platform: Tablet (apk)

- Database type: SQLite3

marthakeezy:
<img width="1092" alt="Screenshot 2024-10-01 at 10 23 41 AM" src="https://github.com/user-attachments/assets/2dddaaf1-cd2d-4a69-a553-b6ec536b9904">

The numbers also don't show fully once added to the prescription & picked/verified

roxy-dao:
Closing as the eclipses is there to indicate that there are trailing numbers. If you hover over the numbers the tooltip should show 0.647777777777 or whatever the number is

lache-melvin:
For triage clarity - this issue is not about actually repeating decimals, but for numbers like Maria has, total should be 4.4. But our `hasMoreThanTwoDp` calculation (to determine whether to show the `...`) ends up with 4.40000000000006, hence the ellipsis. 

I think this should be resolved...

marthakeezy:
Thanks Roxy & Laché - just a note it's impossible to hover over on a tablet, and holding my finger on the cell doesn't work either, so there actually isn't a way to see the rest of decimals on the tablet anyway

Looking at the first screenshot, the total value should be 4.4 with nothing trailing behind

roxy-dao:
Also check if outbound totals and placeholder values are correct while doing this issue as well.

marthakeezy:
[...]

PR #4939, opened 2024-10-01
4591 group by item hides batch even if only one line present

### LINKED ISSUE #4591: Group by item hides batch even if only one line present
## Describe the bug
A. 
image
image

We show most other columns with [multiple] if there are are several lines, or if there is just one line we typically show the value for the one line - except for batch, which appears to show nothing?

B.
Pack sell-prices when grouped by item not matching correctly. see below

### Environment

Local on `develop`, though testers have reported on their systems too.

## To Reproduce

### A. Group by item hides batch even if only one line present
1. Go to open an inbound shipment 
2. Add an invoice line that has a batch
3. Toggle group by item on and off, notice that batch is hidden when **on**

### B. `Pack sell price` value is not updated correctly when grouped by item
1. In an Inbound, add multiple lines of same item with different pack-sell prices, sth like this:
image

2. Note the pack-sell prices there
3. Group by item
4. See error: the total sums up to 0; should either be total of all OR just [multiple]
image
 
## Expected behaviour

A. Should show batch if there is just one line and the UI is grouping by item.
B. should either be total of all OR just [multiple]

## Agreed Solution

TBD

## Additional context

Thinking that if the columns aren't that useful showing [multiple] in `group by item`, should we remove the columns? Could show some aggregate columns instead? We already have unit quantity, perhaps swap the costs columns for total cost, total sell price? If wanting broad nitty gritty then don't group by item.

roxy-dao:
Fixes #4591 and #4918

# 👩🏻‍💻 What does this PR do?
### Inbound Shipment
- Fixes batch not showing up even with only 1 line
- Fixes pack size not showing up with [multiple] for multiple lines 
- Changed sell price per pack to `cost price per unit` to match Outbound Shipment. Makes more sense for Inbound to show cost price than sell price.
- Refactor has2Dp checks + add comments (: 
Screenshot 2024-10-01 at 4 18 28 PM

# 🧪 Testing

- [ ] Create inbound
- [ ] Have multiple lines for one line. Check changed lines mentioned above are correct
- [ ] Have one line: Should see all column info

# 📃 Documentation

  1. Screenshots + descriptions
  2.


client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.ts
```diff
@@ -233,35 +239,27 @@ export const useInboundShipmentColumns = () => {
         },
       ],
       {
-        label: 'label.sell',
-        key: 'sellPricePerPack',
+        label: 'label.cost-per-unit',
+        key: 'costPricePerUnit',
```
roxy-dao, line 243: Changed this to cost per unit to match with OS sell price per unit... The pack sizes are can be all different, so not sure how useful pack per unit price really is?

client/packages/common/src/utils/numbers/NumUtils.ts
```diff
@@ -29,6 +29,13 @@ export const NumUtils = {
     const multiplier = 10 ** dp;
     return Math.round(value * multiplier) / multiplier;
   },
+  /**
+   * Checks if precision is greater than 2. Some numbers have floating precision
+   * errors e.g. 2.05 * 100 = 204.99999999999997 so using 1000 for a more reliable result...
+   */
+  hasMoreThanTwoDp: (value: number): boolean => {
+    return ((value * 1000) / 10) % 1 !== 0;
+  },
```
fergie-nz, line 38: Nice! Javascript is weird.

I wonder is this new check 100% safe? or just safer? I guess just safer is fine if it is safe enough?

fergie-nz (approved):
Nice Roxy looks great to me :)

client/packages/common/src/utils/numbers/NumUtils.ts
roxy-dao: Safer XD

PR #4954, opened 2024-10-02
Add standard reports docs

fergie-nz:
partially Fixes [4800](https://github.com/msupply-foundation/open-msupply/issues/4800) docs requirement.

Note: docs will probably need further updating when we finalise front end + doc syncing methods from OMS central server.

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

No code changes

# 📃 Documentation

Is docs 🎸

### DIFF
--- a/server/report_builder/README.md
+++ b/server/report_builder/README.md
@@ -288,7 +288,13 @@ Optional fields in the manifest json are marked as '// optional'
 
 This manifest.json file takes the place of cli param inserts seen in bash scripts such as upsert.sh in the open-mSupply-reports repo.
 
-Standard reports in the reports dir can be built into the generated json by running the following cli command:
+The command to build reports uses the build cli, so you will first need to run
+
+```bash
+cargo build
+```
+
+Standard reports in the reports dir can then be built into the generated json by running the following cli command:
 
 ```bash
 ./target/debug/remote_server_cli build-standard-reports
@@ -304,22 +310,23 @@ from the open-msupply/server dir.
 
 ## Standard reports versioning
 
-Standard reports include versions for updates of reports. open mSupply central will automatically sync the most recent standard report compatible with the remote site build of open mSupply.
+Standard reports include versions for updates of reports. Open mSupply central will automatically sync all standard reports.
 
-First open mSupply will check for custom reports of the same code which will override any standard report.
-open mSupply central will then sync to the remote site the most recent minor version which matches the major version of the remote site.
-For example:
+2.4 release will include filtering checks to show only the most up to date report compatible with the remote site build of open mSupply.
+
+<!-- For future release - confirm this will occur -->
+<!-- For example:
 
 for report_versions = [2.3.0, 2.3.5, 2.8.2, 2.8.3, 3.0.1, 3.5.1]
 if remote omSupply.version = 2.3 selected report = 2.3.5
 if remote omSupply.version = 2.4 selected report = 2.3.5
 if remote omSupply.version = 2.8 selected report = 2.8.3
 if remote omSupply.version = 3.2 selected report = 3.0.1
-if remote omSupply.version = 4.5 selected report = 3.5.1
+if remote omSupply.version = 4.5 selected report = 3.5.1 -->
 
 ### Adding a new standard report version
 
-Add a new version dir within the name of the report you are adding a new version to.
+Add a new version dir within the directory for the report you are adding a new version to.
 
 > Note that the dir names are for developer convenience only. The name and version of each report read by omSupply is from the manifest.json file.
 
--- a/server/report_builder/README.md
+++ b/server/report_builder/README.md
@@ -197,10 +197,11 @@ This can be implemented in the report by adding the following translation functi
 By default,
 
 Where the letters are short hand for the following:
+
 - t for translate
   The name of the function
-- k for key 
-  This is the locale key as is used in front end translations. 
+- k for key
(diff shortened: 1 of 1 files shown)


server/report_builder/README.md
```diff
@@ -197,10 +197,11 @@ This can be implemented in the report by adding the following translation functi
...
 Where the letters are short hand for the following:
+
 - t for translate
   The name of the function
-- k for key 
-  This is the locale key as is used in front end translations. 
+- k for key
+  This is the locale key as is used in front end translations.
 - n for namespace
   The file namespace where the translation key is. The .json exention is automattically added ie catalogue (which refers to the catalogue.json namespace).
```
jmbrunskill, line 206:
```suggestion
  The file namespace where the translation key is. The .json extention is automatically added ie catalogue (which refers to the catalogue.json namespace).
```
But maybe we can remove this line now that we just have `common.json`

server/report_builder/README.md
```diff
@@ -221,3 +222,103 @@ Next it will fallback to the english translation of the nominated key in the com
...
+
+```bash
+./target/debug/remote_server_cli build-standard-reports
+```
+
+from the open-msupply/server dir.
+
+## Standard reports versioning
+
+Standard reports include versions for updates of reports. open mSupply central will automatically sync the most recent standard report compatible with the remote site build of open mSupply.
```
jmbrunskill, line 298:
```suggestion
Standard reports include versions for updates of reports. open mSupply central will automatically sync all standard report compatible with the remote site build of open mSupply.
```
I think the simplest idea is to sync all reports, then the local version will know which one to use.
It does mean a bit more sync traffic but that is the current plan.

Note: This isn't currently true, so not sure if we should put it in the docs. Maybe it should say from 2.4.0, open-msupply will sync?

server/report_builder/README.md
```diff
@@ -221,3 +222,103 @@ Next it will fallback to the english translation of the nominated key in the com
...
+
+First open mSupply will check for custom reports of the same code which will override any standard report.
```
jmbrunskill, line 300: Again I don't think it does yet right?

server/report_builder/README.md
```diff
@@ -221,3 +222,103 @@ Next it will fallback to the english translation of the nominated key in the com
...
+for report_versions = [2.3.0, 2.3.5, 2.8.2, 2.8.3, 3.0.1, 3.5.1]
+if remote omSupply.version = 2.3 selected report = 2.3.5
+if remote omSupply.version = 2.4 selected report = 2.3.5
+if remote omSupply.version = 2.8 selected report = 2.8.3
+if remote omSupply.version = 3.2 selected report = 3.0.1
+if remote omSupply.version = 4.5 selected report = 3.5.1
+
+### Adding a new standard report version
+
+Add a new version dir within the name of the report you are adding a new version to.
```
jmbrunskill, line 313: With the version of open-mSupply that the report is compatible with...

server/report_builder/README.md
```diff
@@ -221,3 +222,103 @@ Next it will fallback to the english translation of the nominated key in the com
...
+
+> Note that the dir names are for developer convenience only. The name and version of each report read by omSupply is from the manifest.json file.
```
jmbrunskill, line 315: Nice! Good to have these kinds of developer notes, for non-obvious things!

server/report_builder/README.md
```diff
@@ -221,3 +222,103 @@ Next it will fallback to the english translation of the nominated key in the com
...
+  }
+}
+```
+
+This manifest.json file takes the place of cli param inserts seen in bash scripts such as upsert.sh in the open-mSupply-reports repo.
+
+Standard reports in the reports dir can be built into the generated json by running
+
+```bash
+./target/debug/remote_server_cli build-standard-reports
```
jmbrunskill, line 291: Are they just built or also upserted? Isn't there two commands you added?
Should we also mention how to build the cli?

jmbrunskill (commented):
Nice @fergie-nz looking good, but can you review the suggestions and comments I've made?

Maybe an in person discussion would be helpful?

Not sure what we do about the future things in here that aren't in 2.3.0 but I guess this is going into develop so maybe we can assume it will be done before develop is released?
@andreievg thoughts on this?

fergie-nz:
Yes good point thank you James.

I wrote these when we were still planning to add sync + select right version but now won't come out with this release.

In this case is it better to write reports just for the software as it stands, and add sync + selection later?

jmbrunskill:
> In this case is it better to write reports just for the software as it stands, and add sync + selection later?
Yes, I think so unless we're about to do the sync stuff? @andreievg ?

server/report_builder/README.md
jmbrunskill:
```suggestion
Add a new version dir within the directory for the report you are adding a new version to.
The version relates to the earliest mSupply version, the report is compatible with.
```

jmbrunskill (approved):
Keen to get this one in @fergie-nz, 

There's a couple of other comments I'd like addressed but merge when you're ready.
I think probably ok to have the docs explaining what will happen in the future which is what you've done, but please do a review after the next round of reports stuff...

I'd also like to see a video demoing the process to the team.

server/report_builder/README.md
fergie-nz: Have updated to to specify that we aren't using namespace anymore, but functionality is retained (in case we use customer specific namespaces in future)

PR #4961, opened 2024-10-02
4831 dose age labels

### LINKED ISSUE #4831: Add age label to dose
Now that we don't have periods, the label which groups vaccine doses is not configurable by the user.
We'd like the ability to change the words shown in column one here:

image

e.g. `Naissance` rather than `0 months`

When editing a dose, add a text input where you can enter a label.
So the dose edit would have
* label (for the dose)
* start age (in months)
* end age (in months)
* age label
* minimum interval

I'm wondering about the feasibility of using an autocomplete with chips, like this:

image

where the input is limited to a single entry being selected - or a free text.
Behaving in the same way that the tag input behaves in OG. 
This would allow for existing labels to have a consistent spelling, and for the entry of new labels.

We currently generate the label based on the age. Would prefer to have them customisable.

lache-melvin:
Fixes #4831

# 👩🏻‍💻 What does this PR do?

Adds custom label field to vaccine course dose table: 
Screenshot 2024-10-02 at 5 40 29 PM

To support custom labels in the vaccine card:
Screenshot 2024-10-02 at 3 21 38 PM

Sorry, made a number of side quests - I had this branch going while writing docs and was making small fixes as I went. I should have made a separate branch 🙈 

# 🧪 Testing

- [ ] Configure some custom labels
- [ ] They show correctly in the card
- [ ] Doses > 1 year show as years in the card (rather than e.g. 144 months)

# 📃 Documentation

  1.
  2.


client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -294,7 +295,7 @@ const VaccineCourseDoseTable = ({
       },
       {
         key: 'label',
-        Cell: props => <TextInputCell fullWidth {...props} />,
+        Cell: LabelCell,
```
lache-melvin, line 298: Side quest, it was doing the lose focus when you type thing

client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -272,7 +272,7 @@ export const DetailView: FC = () => {
       value: t('label.vaccinations'),
     });
     if (encounter.document.documentRegistry?.uiSchema.elements.length > 0)
-      tabs.push({ Component: JsonForm, value: t('label.encounter') });
+      tabs.push({ Component: JsonForm, value: t('label.details') });
```
lache-melvin, line 275: Side quest, naming request from a while ago that we forgot

client/packages/system/src/Patient/PatientView/PatientView.tsx
```diff
@@ -262,7 +262,7 @@ export enum PatientTabValue {
   Programs = 'programs',
   Encounters = 'encounters',
   ContactTracing = 'contact-tracing',
-  VaccinationCards = 'vaccination-cards',
+  Vaccinations = 'vaccinations',
```
lache-melvin, line 265: Naming request again

client/packages/system/src/Vaccination/Components/SelectItemAndBatch.tsx
```diff
@@ -32,6 +32,18 @@ export const SelectItemAndBatch = ({
...
+
+  const selectBatch =
+    !isHistorical || draft.createTransactions || hasExistingSelectedBatch;
+
+  // Auto-select if there is only one item (and not already selected)
+  useEffect(() => {
+    if (vaccineItemOptions.length === 1 && !draft.itemId) {
+      updateDraft({ itemId: vaccineItemOptions[0]!.value });
+    }
+  }, [vaccineItemOptions]);
```
lache-melvin, line 45: Added the autoselect to item as well, seemed weird to have one without the other? Can back out if you're judging me rn 😁

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -40,11 +40,11 @@ const isPreviousDoseGiven = (
   return itemsForCourse[doseIndex - 1]?.given;
 };
 
-const isRowClickable = (
-  isEncounter: boolean,
+const includeRow = (
```
lache-melvin, line 43: honestly I don't know what to name this, but needed to be more generic, because in Patient view, the "next" doses should NOT be clickable, but we do want to show their status

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -124,23 +124,40 @@ export const VaccinationCardComponent: FC<VaccinationCardProps> = ({
 
   useStyleRowsByStatus(data?.items, isEncounter);
 
+  const getAgeLabel = (row: VaccinationCardItemFragment) => {
```
lache-melvin, line 127: this is the juicy bit frontend wise

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -174,7 +174,8 @@ const useInsert = ({
         vaccinationDate: Formatter.naiveDate(input.date ?? new Date()),
         comment: input.comment,
         notGivenReason: input.notGivenReason,
-        stockLineId: input.stockLine?.id,
+        stockLineId:
+          input.given && !isOtherFacility ? input.stockLine?.id : undefined,
```
lache-melvin, line 178: Small fix to ensure if you select a batch, then change to not given, it doesn't save the batch

roxy-dao (commented):
Thanks Lache! I just have one question >.>

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
lache-melvin: Yep - discussion with design team was that the grouping by age is most important, and from there it makes more sense to all be under one label than awkwardly have some lines of each kind of label. Added to the docs that the expectation is that if you are using custom labels, the same label should be used for all doses of that age, across all courses 🙏  (Ideally we'd have the labels as some kind of reusable chip select thing, but that was too much to implement at this stage of the release!)

roxy-dao (approved):
Cool! Awesome new thang. Thanks Lache (: <3

lache-melvin:
@roxy-dao thanks for review! I pushed a fix for the stale line styling, want to check it before i merge? 🙏

roxy-dao:
> @roxy-dao thanks for review! I pushed a fix for the stale line styling, want to check it before i merge? 🙏

Looks good thanks!

PR #4976, opened 2024-10-03
delete doses on course delete

### LINKED ISSUE #4924: Deleting a vaccine course doesn't remove the course doses from the vaccine card
## What went wrong? 😲

Vaccine course doses aren't being marked deleted when the vaccine course is deleted. The view only checks if the dose is deleted before rendering, so we end up with this:

Screenshot 2024-10-01 at 10 01 08 AM

## Expected behaviour 🤔

If a course has been deleted, and the patient hasn't already received the doses from that course, those doses should not show in the card.

Could check both the course and the doses in the view, but I think more correct to update the vaccine course delete logic to also mark the doses as deleted

## How to Reproduce 🔨

1. Create vaccine course
2. add some doses
3. delete the course
4. view the vaccine card
5. the doses are still there!

## Your environment 🌱

- Open mSupply Version: v2.3.0 rc3

- Platform:

- Database type:

lache-melvin:
Fixes #4924

# 👩🏻‍💻 What does this PR do?

Soft deletes vaccine course doses when vaccine course is soft deleted

# 🧪 Testing

- [ ] Create a vaccine course with doses
- [ ] See doses in vax card
- [ ] Delete the vaccine course
- [ ] Check vax card - doses should not be there
- [ ] (Note - doses where vaccination has already be recorded will continue to show in the card, just can't administer new ones!)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/vaccine_course/delete.rs
+++ b/server/service/src/vaccine_course/delete.rs
@@ -28,13 +28,13 @@ pub fn delete_vaccine_course(
         .transaction_sync(|connection| {
             let doses = validate(connection, &id)?;
 
-            // Make the vaccine course doses as deleted
+            // Mark the vaccine course doses as deleted
             let dose_row_repo = VaccineCourseDoseRowRepository::new(connection);
             for dose in doses {
                 dose_row_repo.mark_deleted(&dose.vaccine_course_dose_row.id)?;
             }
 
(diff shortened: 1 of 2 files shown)


roxy-dao (approved):
LGTM~

PR #4988, opened 2024-10-03
4963 demographic permissions

### LINKED ISSUE #4963: Question regarding demographic permissions
## What went wrong? 😲
If I have edit central data ticked, and now `View Assets`... I can add indicator but it doesn't show up on my list and the permission toast doesn't give much information about what permission the user doesn't have. 

A bit confused about whether the demographics needs the view assets permission? 

Also getting a `Cannot read properties of undefined (reading '__typename')` error

https://github.com/user-attachments/assets/b75809e2-d1f8-4fdd-8fa4-27c7abca5e1a

## How to Reproduce 🔨

1. Tick edit central data permission and untick view asset permission
2. Go to demographics
3. Try add new indicator

## Your environment 🌱

- Open mSupply Version: 2.3

- Platform:

- Database type:

roxy-dao:
Fixes #4963

# 👩🏻‍💻 What does this PR do?
- Use the QueryDemographics permission for all demographic query endpoints (this requires no permission to view)
- Fixes the `__typename` error
- Passes in store_id to check user has correct permission for store

QueryDemographics no longer requires permission, so should never get the permission denied pop up in vaccination... but will need to refactor this bit of code if permission does change in the future.

# 🧪 Testing

- [ ] Be an OMS central 
- [ ] Go to demographics
- [ ] Should be able to view demographics without permission

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -5465,6 +5465,7 @@ export type QueriesDemographicIndicatorsArgs = {
   filter?: InputMaybe<DemographicIndicatorFilterInput>;
   page?: InputMaybe<PaginationInput>;
   sort?: InputMaybe<Array<DemographicIndicatorSortInput>>;
+  storeId: Scalars['String']['input'];
 };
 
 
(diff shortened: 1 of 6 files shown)


lache-melvin (approved):
Awesome! 🙏  What a win with the QueryDemographic perm, wonder why that didn't get applied earlier 😅

PR #5011, opened 2024-10-06
4916 batches with <1 doses remaining are showing as 0

### LINKED ISSUE #4916: Vaccination card transactions aren't rounding correctly
## What went wrong? 😲

For vaccines where doses per unit is more than one, after creating a few vaccination transactions, we end up with this:

Screenshot 2024-09-30 at 5 20 52 PM

Also, the batch select (in the vax card) is rounding the decimals, so if you end up with and almost-zero stock line, it shows there as 0. Doses should always be an integer, shouldn't need the rounding!

Image

## Expected behaviour 🤔

Decimal reductions in stock from vaccinations add up to whole units. None of this 0.00000000000001 business.

## How to Reproduce 🔨

1. Have a vaccine item with many doses per unit. Bonus points for something like 7 per unit 😱 
2. Give some vaccinations
3. View stock levels

## Your environment 🌱

- Open mSupply Version: v2.3.0 RC3

- Platform:

- Database type:

lache-melvin:
Fixes #4916

# 👩🏻‍💻 What does this PR do?

Per issue comment - not fixing anything up in the transactions (i.e. still can end up with 0.00000000001) but instead of rounding that down to 0, show `<1` and disable the line:

Screenshot 2024-10-05 at 4 59 54 PM

Could also just hide the line, but I think better to show user that there is an unusable line sitting there, prompting them to clean up with stocktake/adjustment?

# 🧪 Testing

- [ ] Have less than a dose worth of a vaccination left (set something like 3 doses per vial (0.3333 units each vaccination, so end up with 0.000000001 units after 3 vaccinations)
- [ ] Try to give another dose of that vaccine item
- [ ] That batch should be disabled, and show <1

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Vaccination/Components/SelectBatch.tsx
+++ b/client/packages/system/src/Vaccination/Components/SelectBatch.tsx
@@ -5,10 +5,10 @@ import {
   ColumnAlign,
   createTableStore,
   DataTable,
-  NumUtils,
   TableProvider,
   useColumns,
   useRowStyle,
+  useTableStore,
   useTranslation,
 } from '@openmsupply-client/common';
 import React, { useEffect } from 'react';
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
LGTM~

PR #5020, opened 2024-10-07
Fix broken report id urls

### LINKED ISSUE #4990: App crashes if you refresh or change language on reports page
## What went wrong? 😲

https://github.com/user-attachments/assets/d81cdd32-9427-4766-b3a7-9efece0de986

Recording is on browser, APK also crashes with `File not found`
 
Screenshot 2024-10-04 at 11 38 25 AM

Probably low likelihood for user to change language, but we don't love app crashes 😁 

## Expected behaviour 🤔

App does not crash - maybe just redirects you back to base reports page??

At the VERY LEAST we should fail to the error boundary, not this not found page!

## How to Reproduce 🔨

1. Go to Reports
2. Generate a report
3. Change the language
4. App crashes

## Your environment 🌱

- Open mSupply Version: v2.3.0-RC5

- Platform:

- Database type:

fergie-nz:
Fixes #4990 

# 👩🏻‍💻 What does this PR do?

Changes report ids to use "_" for report versions rather than "."
"." causing issues in url for some reason.
I couldn't quite figure out why? Would love to hear insights from the wiser folk around :) 

Note that this occurs anytime you refresh with a full stop in the url. Changing language was just one instance of causing a page refresh while with a full stop. This fix just removes full stops from the ids of the reports which were being used to call the reports in the url.

Note - this fix won't get rid of old reports in your database, so you will need to remove old standard reports from the database with "." in the url.

Note big dif in standard reports json is just the regeneration of the same reports with new ids.

# 🧪 Testing

Try changing the language (or refreshing) with a new report

# 📃 Documentation


fergie-nz:
I believe something related to this:

https://stackoverflow.com/questions/13298542/apicontroller-returns-404-when-id-contains-period

Where it is checking for a file. 

Maybe a more elegant solution is available?

I am thinking for us that given the ids are arbitrary, this fix will be fine.

server/cli/src/cli.rs
```diff
@@ -440,7 +440,9 @@ async fn main() -> anyhow::Result<()> {
                     let code = manifest.code;
 
                     let version = manifest.version;
-                    let id = format!("{code}_{version}");
+                    let id_version = str::replace(&version, ".", "_");
```
jmbrunskill, line 443: I'm wondering if we are better to just use a UUID here? Consistent with all our other processes?
I quite like this approach of having a readable ID, but wondering if it creates other problems...

jmbrunskill (changes requested):
Can you add a migration to fix the existing ids please?
Otherwise it will still break for RC builds.

I'm also thinking we should probably bit the bullet and remove id from the standard_reports.json, and just generate UUIDS for each report, or possibly pre-generate them in the report manifest so they're the same everywhere but still a UUID?

server/cli/src/cli.rs
lache-melvin: What other problems do you imagine James? I quite liked the idea of the version being included, if ever needing to troubleshoot (It helped me realise why I had duplicates from v2.2 😁)

lache-melvin:
> "." causing issues in url for some reason. I couldn't quite figure out why?

I think due to everything after the `.` being considered the file extension! And the server couldn't find a file of that type?

server/cli/src/cli.rs
jmbrunskill: I think this is the blog post I'm looking for?
https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/

I'm on the fence with this one, as like you I kind of line having it how it is. 

I just know that trying to produce reliable ids like this can have unintended consequences (such as we've just seen), and the assumptions behind them might not always hold in the future?

I'm fine with it how it is personally, just need the migration to avoid the issues for tests in the RC build.

andreievg: I am ok with underscore for this fix (would also include is_custom in 2.4)

I think with pre generated ids we might have issue with uniqueness (especially when there are custom reports in separate repos)

server/repository/src/migrations/v2_03_00/remove_stops_from_report_ids.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                SELECT
+                    REPLACE(id, '.', '_')
+                FROM 
+                    report;
```
jmbrunskill, line 17:
```suggestion
                UPDATE report SET id =  REPLACE(id, '.', '_');
                UPDATE form_schema id REPLACE(id, '.', '_'); 
```
Doing just a select does nothing sadly, need to an update!
I think we need to do form_schema as well?

jmbrunskill (changes requested):
Migration isn't quite right yet, pretty sure you'll be ok with my suggestion though?

server/repository/src/migrations/v2_03_00/remove_stops_from_report_ids.rs
fergie-nz: Ah nice thank you James. I must have tested this wrong to not pick that up!

fergie-nz:
> Migration isn't quite right yet, pretty sure you'll be ok with my suggestion though?

Yes looks great thank you

jmbrunskill (approved):
Thanks Fergus

PR #5024, opened 2024-10-07
Fix intermittent test failure for transfers

### LINKED ISSUE #4608: Intermittent test failure: approved quantity transfer
## What went wrong? 😲

There's some kind of race condition/dependency in the tests - is it an actual bug?

## Expected behaviour 🤔

Tests consistently passss

## How to Reproduce 🔨

1. Run backend tests
2. Maybe run many time?
3. Requisition transfer test will fail, approved quantity didn't get transferred?

## Your environment 🌱

- Open mSupply Version: V2.2.0 RC3

- Platform:

- Database type:

jmbrunskill:
Fixes #4608 

# 👩🏻‍💻 What does this PR do?

Turned out to be a simple fix, but was very hard to debug and track down.

Basically the problem turned out to be that in the test case, we were first updating the status of the request requisition to approved, then updating the approved quantities second. This created a short period of time, where if the processor ran, it could get the updated status for the request requisition and process the lines even though these hadn't been updated yet.

I'm pretty confident that we always have lines inserted before updating invoices, although I'm a bit worried about sync.
It will update the invoice first I think, but does do it in a transaction now, and calls the processor only after sync is finished so *I think* we're ok, 

*Keen for input/confirmation that we don't have the same potential order of operations error in production code!*

I've added some thread ids and debug logs for future debugging help. I think it's worth keeping it in, if reviewers are happy...

Example log of what was going wrong, filtered debug output...
The `0:` prefix is from the particular tokio thread that had this issue.
The Processor lines don't have a number prefix

```
0: Updating line 01926556-7cd0-7960-a58b-37ce747a91bb to approved (42.0) 1/2
Processing rq: "0_request_requisition" Sent
About to update request lines rq: "0_request_requisition" Sent response lines: 2
Updated lines for: "0_request_requisition" to approved quantity: 0.0
Updated lines for: "0_request_requisition" to approved quantity: 0.0
0: Updating line 01926556-7cd0-7960-a58b-37dd1a7b4e3a to approved (42.0) 2/2
0: await_events_processed
Processing requisition transfers
Processing await process queue
0: check_request_requisition_approved
0: Checking line 0_request_requisition_line_1
thread 'tokio-runtime-worker' panicked at service/src/processors/transfer/requisition/test.rs:584:5:
assertion `left == right` failed
  left: 42.0
 right: 0.0
```

Updating the lines first and then the status seems to have solved it.

With SQLite we could just wrap it in a transaction but I think this approach is closer to real operation in OMS?

# 🧪 Testing

- [ ] `cargo test --package service --lib -- processors::transfer::requisition::test::requisition_transfer --exact --show-output`

# 📃 Documentation

### DIFF
--- a/server/service/src/processors/transfer/requisition/test.rs
+++ b/server/service/src/processors/transfer/requisition/test.rs
@@ -90,28 +90,39 @@ async fn requisition_transfer() {
     let test_handle = exec_concurrent(
         test_input,
         number_of_instances,
-        |_, test_input| async move {
+        |thread_num, test_input| async move {
             let (service_provider, request_store, response_store, item1, item2) = test_input;
 
             let ctx = service_provider.basic_context().unwrap();
+            let mut tester = RequisitionTransferTester::new(
+                &request_store,
+                &response_store,
+                &item1,
+                &item2,
+                thread_num,
+            );
 
-            let mut tester =
-                RequisitionTransferTester::new(&request_store, &response_store, &item1, &item2);
-
+            log::debug!("{}: insert", thread_num);
             tester.insert_request_requisition(&ctx.connection);
             // manually trigger because inserting the requisition doesn't trigger the processor
             ctx.processors_trigger
                 .requisition_transfer
                 .try_send(())
                 .unwrap();
+            log::debug!("{}: await_events_processed", thread_num);
             ctx.processors_trigger.await_events_processed().await;
+            log::debug!("{}: check_response_requisition_not_created", thread_num);
             tester.check_response_requisition_not_created(&ctx.connection);
+            log::debug!("{}: update_request_requisition_to_sent", thread_num);
             tester.update_request_requisition_to_sent(&service_provider);
             ctx.processors_trigger.await_events_processed().await;
-            tester.check_response_requisition_created(&ctx.connection);
+            log::debug!("{}: check_response_requisition_created", thread_num);
+            tester.check_response_requisition_created(&ctx.connection, thread_num);
             ctx.processors_trigger.await_events_processed().await;
+            log::debug!("{}: check_request_requisition_was_linked", thread_num);
             tester.check_request_requisition_was_linked(&ctx.connection);
-            tester.update_response_requisition_to_approved(&service_provider);
+            log::debug!("{}: update_response_requisition_to_approved", thread_num);
+            tester.update_response_requisition_to_approved(&service_provider, thread_num);
             // Response requisition approval is usually done by mSupply
             // Processor would be triggered after sync
             // We've approved manually for testing, so need to manually trigger the processor as well
@@ -119,12 +130,14 @@ async fn requisition_transfer() {
                 .requisition_transfer
                 .try_send(())
                 .unwrap();
+            log::debug!("{}: await_events_processed", thread_num);
             ctx.processors_trigger.await_events_processed().await;
-            // Sometimes processor isn't finished the first time, so we await twice?
-            ctx.processors_trigger.await_events_processed().await;
-            tester.check_request_requisition_approved(&ctx.connection);
+            log::debug!("{}: check_request_requisition_approved", thread_num);
+            tester.check_request_requisition_approved(&ctx.connection, thread_num);
+            log::debug!("{}: update_response_requisition_to_finalised", thread_num);
             tester.update_response_requisition_to_finalised(&service_provider);
             ctx.processors_trigger.await_events_processed().await;
+            log::debug!("{}: check_request_requisition_status_updated", thread_num);
             tester.check_request_requisition_status_updated(&ctx.connection);
         },
     );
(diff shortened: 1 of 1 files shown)


server/service/src/processors/transfer/requisition/test.rs
```diff
@@ -90,41 +90,54 @@ async fn requisition_transfer() {
...
             // Response requisition approval is usually done by mSupply
             // Processor would be triggered after sync
             // We've approved manually for testing, so need to manually trigger the processor as well
             ctx.processors_trigger
                 .requisition_transfer
                 .try_send(())
                 .unwrap();
+            log::debug!("{}: await_events_processed", thread_num);
             ctx.processors_trigger.await_events_processed().await;
-            // Sometimes processor isn't finished the first time, so we await twice?
```
jmbrunskill, line 123: Took this out, as it seems like it was an attempt to fix something, but might be masking an issue?

server/service/src/processors/transfer/requisition/test.rs
```diff
@@ -221,9 +234,10 @@ impl RequisitionTransferTester {
         response_store: &StoreRow,
         item1: &ItemRow,
         item2: &ItemRow,
+        thread_number: u32,
     ) -> RequisitionTransferTester {
         let request_requisition = inline_init(|r: &mut RequisitionRow| {
-            r.id = uuid();
+            r.id = format!("{}_request_requisition", thread_number);
```
jmbrunskill, line 240: This helps with the debug output and looking at the db, but could go back to UUID here?

server/service/src/processors/transfer/requisition/test.rs
```diff
@@ -408,34 +429,48 @@ impl RequisitionTransferTester {
...
+                line_count
+            );
             line.approved_quantity = 42.0;
             RequisitionLineRowRepository::new(&ctx.connection)
                 .upsert_one(&line)
                 .unwrap();
+            i += 1;
         }
 
+        let approved_response_requisition = RequisitionRow {
```
jmbrunskill, line 457: This is the actual fix, setting the status after we're already updated the lines...

server/service/src/processors/transfer/requisition/test.rs
```diff
@@ -90,41 +90,54 @@ async fn requisition_transfer() {
...
-                RequisitionTransferTester::new(&request_store, &response_store, &item1, &item2);
-
+            log::debug!("{}: insert", thread_num);
             tester.insert_request_requisition(&ctx.connection);
             // manually trigger because inserting the requisition doesn't trigger the processor
             ctx.processors_trigger
                 .requisition_transfer
                 .try_send(())
                 .unwrap();
+            log::debug!("{}: await_events_processed", thread_num);
```
roxy-dao, line 112: Will all the debug logs stay~? 

Also backend tests still seem to be failing....

server/service/src/processors/transfer/requisition/test.rs
```diff
@@ -90,41 +90,54 @@ async fn requisition_transfer() {
     let test_handle = exec_concurrent(
         test_input,
         number_of_instances,
-        |_, test_input| async move {
+        |thread_num, test_input| async move {
```
fergie-nz, line 93: Hey James, I'm struggling to figure out how this closure is actually finding the thread count. Would you be willing to explain how it is capturing it?

server/service/src/processors/transfer/requisition/test.rs
jmbrunskill: The same test is failing for you?

server/service/src/processors/transfer/requisition/test.rs
jmbrunskill: @fergie-nz have a look at the exec_concurrent code, it's generated in there...
It's just a for loop based on the number of concurrent instances

server/service/src/processors/transfer/requisition/test.rs
jmbrunskill: Or do you mean this ? https://github.com/msupply-foundation/open-msupply/issues/5029

jmbrunskill: > Will all the debug logs stay~?

I'm suggesting Yes, why not?

roxy-dao: Just curious

lache-melvin (approved):
Looks great to me! Thanks for the effort investigating this James 🙏 

Hopefully the added logging makes for faster troubleshooting next time 🚀

lache-melvin:
> I'm pretty confident that we always have lines inserted before updating invoices, although I'm a bit worried about sync. It will update the invoice first I think, but does do it in a transaction now, and calls the processor only after sync is finished so _I think_ we're ok,
> 
> _Keen for input/confirmation that we don't have the same potential order of operations error in production code!_

Approved quantities are only updated through the remote authorisation module (OG), and processors are only triggered once sync is complete, so I think we are okay? It's a good prompt if we might have any other race conditions like this anywhere though... would have thought the transactions would protect us, like you say?

jmbrunskill:
> would have thought the transactions would protect us, like you say?

That's not actually given! In SQLite, Yes as there's only ever 1 writer at a time, but in postgres it depends on isolation level and locks that are used. I think by default if the processor ran while a transaction was underway to update the lines, it would see the old lines, as it would see the last committed state.
I think we're save though by the fact that the updated status also wouldn't be there so the processor wouldn't try to process the record. 

https://www.postgresql.org/docs/current/transaction-iso.html
I'm sure there's a nice transaction levels learning aid somewhere, would be worth a quick look one day...

PR #5040, opened 2024-10-08
Use invoice status to adjustStockLines

### LINKED ISSUE #4936: Available can be bigger then total in outbound line edit
## What went wrong? 😲

Available can show higher then total in picked outbound shipment line edit

## Expected behaviour 🤔

Total should never be lower then available.

## How to Reproduce 🔨

Create new outbound, issue some stock, change status to picked, see that total is lower then available in outbound line edit.

## Agreed solution

Adjust total + number of packs issued for picked outbound lines.

Related:
https://github.com/msupply-foundation/open-msupply/pull/4911
https://github.com/msupply-foundation/open-msupply/issues/3252

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #4936 

# 👩🏻‍💻 What does this PR do?

Changes the code for creating draftStockOutLines to decide when to adjust stock levels based on 

Logic should be

Available Stock show is the total of the available stock in store (e.g total stock - allocated to other invoices) plus the amount  you have allocated or picked on this invoice line.
Total stock (in store) is also adjusted to include what you have allocated to this line too, once the invoice is in the picked state.

This is because by updating that line you can adjust the total stock and or available stock levels, so we need to show the adjusted stock levels to avoid confusion about how much you can or can't allocate to the invoice.

Any status things I'm missing?
Is there some kind of magic logic from before that I've missed?

# 🧪 Testing

- [ ] Try editing lines, saving, re-opening, etc, check that the available stock is always less than total stock (in store)
- [ ] Update status e.g allocated, repeat earlier tests
- [ ] Update status to Picked, repeat earlier tests
- [ ] Do the same but with prescriptions (these are affected by this change too)

# 📃 Documentation
Is there some documentation on this that needs to be updated?

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useDraftOutboundLines.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useDraftOutboundLines.tsx
@@ -73,11 +73,12 @@ export const useDraftOutboundLines = (
           const invoiceLine = lines?.find(
             ({ stockLine }) => stockLine?.id === batch.id
           );
-
           if (invoiceLine) {
             return createDraftStockOutLine({
               invoiceLine,
+              stockLine: batch,
               invoiceId,
+              invoiceStatus: status,
             });
           } else {
             return createDraftStockOutLineFromStockLine({
@@ -98,7 +99,11 @@ export const useDraftOutboundLines = (
           const placeHolderItem = lines?.find(l => l.item.id === item.id)?.item;
           if (!!placeHolderItem) placeholder.item = placeHolderItem;
           rows.push(
-            createDraftStockOutLine({ invoiceId, invoiceLine: placeholder })
+            createDraftStockOutLine({
+              invoiceId,
+              invoiceLine: placeholder,
+              invoiceStatus: status,
+            })
           );
         } else {
           rows.push(createStockOutPlaceholderRow(invoiceId, item.id));
--- a/client/packages/invoices/src/StockOut/hooks/usePackSizeController.test.tsx
+++ b/client/packages/invoices/src/StockOut/hooks/usePackSizeController.test.tsx
@@ -2,6 +2,7 @@ import { usePackSizeController } from './usePackSizeController';
 import { act } from '@testing-library/react';
 import {
   InvoiceLineNodeType,
+  InvoiceNodeStatus,
   renderHookWithProvider,
 } from '@openmsupply-client/common';
 import {
@@ -37,6 +38,7 @@ const testLine = ({
 }: TestLineParams): DraftStockOutLine =>
   createDraftStockOutLine({
     invoiceId: '',
+    invoiceStatus: InvoiceNodeStatus.New,
     invoiceLine: {
       id,
       expiryDate,
--- a/client/packages/invoices/src/StockOut/utils.tsx
+++ b/client/packages/invoices/src/StockOut/utils.tsx
@@ -101,9 +101,7 @@ export const createDraftStockOutLine = ({
   // Other statuses such as Shipped shouldn't show the stock line as available, so we don't need to adjust the available number of packs
   // If the invoice is New, no adjustments are needed, as the stockLines shouldn't be updated yet
 
-  const adjustAvailableNumberOfPacks =
-    invoiceStatus === InvoiceNodeStatus.Picked ||
-    invoiceStatus === InvoiceNodeStatus.Allocated;
+  const adjustAvailableNumberOfPacks = invoiceStatus !== InvoiceNodeStatus.New;
   const adjustTotalNumberOfPacks = invoiceStatus === InvoiceNodeStatus.Picked;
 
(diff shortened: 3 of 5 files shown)


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useDraftOutboundLines.tsx
```diff
@@ -73,11 +73,12 @@ export const useDraftOutboundLines = (
           const invoiceLine = lines?.find(
             ({ stockLine }) => stockLine?.id === batch.id
           );
-
           if (invoiceLine) {
             return createDraftStockOutLine({
               invoiceLine,
+              stockLine: batch,
```
jmbrunskill, line 79: Wasn't 100% if this should be passed in or not, I think it wasn't included before to try to trigger the adjustment logic but there might have been another reason I'm not aware of too?

client/packages/invoices/src/StockOut/hooks/usePackSizeController.test.tsx
```diff
@@ -37,6 +38,7 @@ const testLine = ({
 }: TestLineParams): DraftStockOutLine =>
   createDraftStockOutLine({
     invoiceId: '',
+    invoiceStatus: InvoiceNodeStatus.New,
```
jmbrunskill, line 41: Didn't look at these tests but they do seem to pass with `New` so I left it as New for now

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -93,28 +93,43 @@ export const createDraftStockOutLineFromStockLine = ({
...
-});
+  invoiceStatus,
+}: DraftStockOutLineSeeds): DraftStockOutLine => {
+  // When creating a draft stock out line from an invoice line we may need to adjust the available and total number of packs
+  // This is because, once an invoice line is created and it is in the allocated state, the available number of packs is reduced by the number of packs in the invoice line
+  // After it is in picked status, the total number of packs is also reduced by the number of packs in the invoice line
+  // Other statuses such as Shipped shouldn't show the stock line as available, so we don't need to adjust the available number of packs
+  // If the invoice is New, no adjustments are needed, as the stockLines shouldn't be updated yet
+
+  const adjustAvailableNumberOfPacks =
```
andreievg, line 104: I am pretty sure we adjustAvailableNumberOfPacks for all invoice statuses

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -93,28 +93,43 @@ export const createDraftStockOutLineFromStockLine = ({
 export const createDraftStockOutLine = ({
   invoiceLine,
   stockLine,
```
andreievg, line 95: Wouldn't invoiceLine always have stockLine ?

andreievg (changes requested):
Thanks James, only request is for the availableNumberOfPacks to be adjusted in all invoice statuses

client/packages/invoices/src/StockOut/utils.tsx
jmbrunskill: apart from new right? Then the lines don't actually exist. Although I think we don't call this code when the invoice is new... We call the draftFromStockLIne one instead...

client/packages/invoices/src/StockOut/utils.tsx
jmbrunskill: Yeah I'm a little confused by the reasoning for this being passed as well, as the `invoice.stockLine` which should already exist...
Maybe better to remove it to remove any confusion...

client/packages/invoices/src/StockOut/utils.tsx
jmbrunskill: What do we want to do with Shipped for example I'm ok with always adjusting I'm just not sure if adjusting really makes sense...

client/packages/invoices/src/StockOut/utils.tsx
jmbrunskill: Was scared to remove it and find somewhere it was needed but missed in testing.
That being said it feels like it's unnecessary, maybe some state thing when only one of the queries is refreshed but even then it's the invoice one that should matter right?

andreievg: Yeah could be true, I am ok to keep, especially since we will be refactoring this area

client/packages/invoices/src/StockOut/utils.tsx
andreievg: > apart from new right? Then the lines don't actually exist. Although I think we don't call this code when the invoice is new... We call the draftFromStockLIne one instead...

You can add and edit lines in new status.

>  I'm just not sure if adjusting really makes sense...

We've been following mSupply logic for this (even thought statuses are different, adjustment for available always happens)

andreievg (approved):
Thanks @jmbrunskill

PR #5054, opened 2024-10-09
5053-Store-hardware-id-in-memory

### LINKED ISSUE #5053: Hardware id to be fetched from memory

We had an issue on a server that caused service to quit and then fail to start, this is the log file.

https://drive.google.com/drive/u/1/folders/15L6VLXbC6SCWUbL-qKy4ZcZGAhrhia4d

In the log file it look like panic around app data (when fetching hardware id), which happens on startup and in synchroniser (a bit puzzled why so many logs for the synchroniser, that doesn't cause the app to quit, if it's protected from panic since it's on it's own thread that makes sense, but why does the service quit in the end), on startup it makes sense as there is a sync operation in main thread.

After app data file is deleted, service manage to start (new app data file is created as expected)

Fetch hardware id from memory, look it up on startup

andreievg:
Fixes #5053

# 👩🏻‍💻 What does this PR do?

Uses global constant RwLock to get and set hardware id

Would need to create a carry over refactor issue to:
* Remove app data service and set and get hardware id from the const
* Remove hardware id from settings and use set hardware id in android ?
* For non android, first time get_hardware_id is called get system hardware id ? 
* Above would allow integration test to work without too much extra (with this change I think integration tests will be broken)

# 🧪 Testing

Sync for existing sites does not break and existing hardware id is used

# 📃 Documentation

  1.
  2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.2.04",
+  "version": "2.2.05",
   "private": true,
   "scripts": {
     "start": "cd ./server && cargo run & cd ./client && yarn start-local",
--- a/server/server/src/lib.rs
+++ b/server/server/src/lib.rs
@@ -136,7 +136,7 @@ pub async fn start_server(
     }
 
     // SET HARDWARE UUID
-    info!("Setting hardware uuid..");
+    info!("Getting hardware uuid..");
     #[cfg(not(target_os = "android"))]
     let machine_uid = machine_uid::get().expect("Failed to query OS for hardware id");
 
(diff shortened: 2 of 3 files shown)


lache-melvin (approved):
LGTM, thanks for pumping this out Andrei! 🚀

lache-melvin (changes requested):
Screenshot 2024-10-10 at 8 37 04 AM

Whoops - approved before the tests finished running 👀 Assuming this is related?

lache-melvin (approved):
I kid, the same test is failing in main, will make another issue to fix! Thanks Andrei 😅

jmbrunskill:
@andreievg I'm ok with this change, but I suspect the hardware_id isn't coming back correctly for this machine.
The line of code that's failing is failing to parse the yaml file, not failing to open it.

I'm suspicious that we're getting a weird value from the machine_id::get() ?

I wonder if it's worth logged the machine_uid too just so we can see what it's getting? Or actually I guess we can just look at that yaml file to see what it has in it?

lache-melvin:
https://github.com/msupply-foundation/open-msupply/issues/5067

andreievg:
> @andreievg I'm ok with this change, but I suspect the hardware_id isn't coming back correctly for this machine.
The line of code that's failing is failing to parse the yaml file, not failing to open it.

@jmbrunskill that's true,

PR #5068, opened 2024-10-10
Fix rc migration

jmbrunskill:
Fixes #

# 👩🏻‍💻 What does this PR do?

Migration tried to update ID, but foreign keys weren't update yet.
Since this is just for RC, I think it's fine if we have a few duplicated records so I've removed that line from the migration

Sorry I didn't test my suggestion properly :(

# 🧪 Testing

- [ ] Run RC7?

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
mean as, thanks james!

PR #5071, opened 2024-10-10
Remove unique constraint on breach description

### LINKED ISSUE #4755: Importing a fridge tag fails if an import has happened in another store
## What went wrong? 😲

Given:
* you are running postgres
* there are mutliple stores configured for your site
* a fridge tag has been imported in another store

Then:
any fridge tag import will fail

The issue is that this call to check matching sensor config (berlinger.rs, line 204)

```
let result = get_matching_sensor_breach_config(
        connection,
        &sensor_row.store_id,
        temperature_breach_config,
        &breach_row_type,
    )?;

    if !result.is_empty() {
        return Ok(());
    };
```

is checking for a config based on store_id, duration_milliseconds, minimum_temperature, maximum_temperature and type.
However, the constraint on the table is for a unique description and this does not include the store_id, so the insert following the code above will fail.

## Expected behaviour 🤔
Able to import the fridge tag data for a new store.

## How to Reproduce 🔨

1. Go to Cold chain / monitoring
2. Import a fridge tag
3. Switch to another store and repeat

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #4755

# 👩🏻‍💻 What does this PR do?

It kind of fixes the issue above...

It's a dis-satisfying end though.

Removes the restriction on `temperature_breach_config`:`description`
This means if the same fridge tag is imported on different stores, the `config` will be duplicated for each store.
The fridge tag data is then imported in the other store.

The way the endpoint works is that the sensor `last_connection_datetime` is checked and you get what looks like an error if you try to import the same data into a different store saying 

image

The API needs to improve to clearly state that no `NEW` datapoints were found in the import.
Probably for a real use case when you're plugging in the device this won't be a problem as there should be new datapoints?

It feels like breach config should probably either be independent to the store it's loaded against, or perhaps linked to the `sensor_id`. However to change this we'll need to change mSupply as it's also using `store_id` it will also cause challenges for sync if there's no `store_id` so put it in the too hard bucket for now...

Maybe Adrian's suggestion of forcing the user to change store for a sensor is a good one? Although we might want to have Cold Chain devices that are deliberately imported into different stores though, e.g. if they're transported with product in future.

So I think that space needs some more design thinking rather than a hack now...

# 🧪 Testing

It would be best to test this with a real device, to avoid the issues with 

- [ ] Import data from a sensor into one store
- [ ] wait for new datapoint to be generated
- [ ] import data into another store
- [ ] Check new datapoints show up in the second store
- [ ] If you get an error than the one saying nothing to import then it's probably an issue

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/migrations/v2_04_00/remove_unique_description_on_tmp_breach.rs
+++ b/server/repository/src/migrations/v2_04_00/remove_unique_description_on_tmp_breach.rs
@@ -1,0 +1,44 @@
+use crate::migrations::*;
+
+pub(crate) struct Migrate;
+
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "remove_unique_description_on_tmp_breach"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        if cfg!(feature = "postgres") {
+            sql!(
+                connection,
+                r#"
+                    ALTER TABLE temperature_breach_config DROP CONSTRAINT temperature_breach_config_description_key;
+                "#
+            )?;
+        } else {
+            sql!(
+                connection,
+                r#"
+                CREATE TABLE tmp_temperature_breach_config (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    duration_milliseconds INTEGER NOT NULL,
+                    type TEXT NOT NULL,
+                    description TEXT NOT NULL,
+                    is_active BOOLEAN,
+                    store_id TEXT NOT NULL REFERENCES store(id),
+                    minimum_temperature {DOUBLE} NOT NULL,
+                    maximum_temperature {DOUBLE} NOT NULL
+                );
+                INSERT INTO tmp_temperature_breach_config SELECT * FROM temperature_breach_config;
+
+                PRAGMA foreign_keys = OFF;
+                DROP TABLE temperature_breach_config;
+                ALTER TABLE tmp_temperature_breach_config RENAME TO temperature_breach_config;
+                PRAGMA foreign_keys = ON;
+                "#
+            )?;
+        }
+
+        Ok(())
+    }
+}
--- a/server/repository/src/migrations/mod.rs
+++ b/server/repository/src/migrations/mod.rs
@@ -24,6 +24,7 @@ mod v2_02_00;
 mod v2_02_01;
 mod v2_02_02;
 mod v2_03_00;
+mod v2_04_00;
 mod version;
 mod views;
 
@@ -119,6 +120,7 @@ pub fn migrate(
         Box::new(v2_02_01::V2_02_01),
         Box::new(v2_02_02::V2_02_02),
         Box::new(v2_03_00::V2_03_00),
+        Box::new(v2_04_00::V2_04_00),
     ];
 
     // Historic diesel migrations
--- a/server/repository/src/migrations/v2_04_00/mod.rs
+++ b/server/repository/src/migrations/v2_04_00/mod.rs
@@ -1,0 +1,43 @@
+use super::{version::Version, Migration, MigrationFragment};
+
+use crate::StorageConnection;
+mod remove_unique_description_on_tmp_breach;
+
+pub(crate) struct V2_04_00;
+
+impl Migration for V2_04_00 {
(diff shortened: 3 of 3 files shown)


server/repository/src/migrations/v2_04_00/remove_unique_description_on_tmp_breach.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn identifier(&self) -> &'static str {
+        "remove_unique_description_on_tmp_breach"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        if cfg!(feature = "postgres") {
+            sql!(
+                connection,
+                r#"
+                    ALTER TABLE temperature_breach_config DROP CONSTRAINT temperature_breach_config_description_key;
```
jmbrunskill, line 15: I'm pretty sure this is ok but but I'm not 100% sure about an old version of posgres for example.
Officially you're supposed to look up the constraint name first then run a command line this.
Tempted to also modify the original table migration so any new installs won't ever get this broken version?

jmbrunskill: Added a warn message just incase there's an issue hopefully with postgres people will be able to login and fix this if there was a problem...

server/repository/src/migrations/v2_04_00/remove_unique_description_on_tmp_breach.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        if cfg!(feature = "postgres") {
+            let result = sql!(
+                connection,
+                r#"
+                    ALTER TABLE temperature_breach_config DROP CONSTRAINT temperature_breach_config_description_key;
+                "#
+            );
+            if (result.is_err()) {
```
lache-melvin, line 18: These parentheses are giving a ``#[warn(unused_parens)]` warning 😁

lache-melvin:
Hey James - as discussed at stand up, I've committed my tests for this problems, can be run against develop and this branch to see:

- Should be able to import a new sensor into a different store <- this is the one that is currently failing in develop
- Per Adrian's comment in the issue, should NOT be able to import the same sensor into a different store: however, in develop, currently you can!
  - This is due to the code sometimes using sensor_row.store_id, rather than the store_id passed in. 
  - As discussed, maybe that's helpful! Can upload sensor data to the right store, from any store!
  - However, is that breaking anything with sync (if store B syncs a record with `store_a` as the store_id, does it ever actually make it back to store A?

Need decision around 1) whether we care about allowing same tag to be imported into multiple stores, then 2) which store_id to use.

This has already taken quite an investment of time, think this should go back through some kind of triage before much more time spent 🤷‍♀️

lache-melvin:
Hey @jmbrunskill - think it would be good to be able to close this... wonder if we should merge this as is, and create another issue for the import tag from another store question?

jmbrunskill:
> Hey @jmbrunskill - think it would be good to be able to close this... wonder if we should merge this as is, and create another issue for the import tag from another store question?

I think that's a good plan,I might have to update the migration to go into 2.5 though...

jmbrunskill:
@lache-melvin updated this to be in 2.5, and commented out the broken test. Want to review again when you get a chance?

lache-melvin (approved):
Mean thanks James! If you could create quick carry over issue and link in that commented out test that would be great - though imagine it won't be a priority for a while 😢 

Issue should probably include that if you have created/imported the fridge tag in one store, and then switch to different store, and import it again, you get a success message, but no sensor/logs appear (they were imported into first store)

jmbrunskill:
> Mean thanks James! If you could create quick carry over issue and link in that commented out test that would be great - though imagine it won't be a priority for a while 😢
> 
> Issue should probably include that if you have created/imported the fridge tag in one store, and then switch to different store, and import it again, you get a success message, but no sensor/logs appear (they were imported into first store)

I'm hesitant to add to our issue count, and I don't think we've got a clear plan on what we actually want here.
@mark-prins do you want a new issue for this, or do we just wait until we have a real use case planned?

PR #5076, opened 2024-10-10
5056 report descriptions

### LINKED ISSUE #5056: Report descriptions
There is currently no way to see what the data in the individual reports are for and what the columns mean.

Add `How to read { report name } report? >` link that expands in the `App Bar` (see screenshots below and login error for implementation). This should show the column descriptions on expansion see [expiring items README](https://github.com/msupply-foundation/open-msupply-reports/blob/main/standard-reports/expiring-items/README.md), [item usage README](https://github.com/msupply-foundation/open-msupply-reports/blob/main/standard-reports/item-usage/README.md) and [stock status README](https://github.com/msupply-foundation/open-msupply-reports/blob/main/standard-reports/stock-status/README.md) for descriptions.

photo_2024-10-10 08 25 18
photo_2024-10-10 07 44 33

### Moneyworks Jobcode 🧰
OMS:REPT

roxy-dao:
Fixes #5056

# 👩🏻‍💻 What does this PR do?
Adds report description for columns in the report detail view's toolbar. An info box will appear or disappear when the user clicks, `How to read { report name } report?` as per @mark-prins's request (:

https://github.com/user-attachments/assets/0473b53f-3a20-4746-af77-27d23a2fb2dd

# 🧪 Testing

- [ ] Go to any report
- [ ] Click on `How to read { report name of the report you clicked } report?`
- [ ] See info box
- [ ] Click link again
- [ ] See box disappear

# 📃 Documentation

  1. Report screenshots + a little blub about what this is.
  2.


client/packages/reports/src/DetailView/Toolbar.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <Grid sx={{ paddingLeft: 3 }}>
+            <Typography sx={{ textWrap: 'wrap' }} variant="body2">
+              {t(`messages.how-to-read-${reportFormat}` as LocaleKey)
+                .split('\n')
+                .map((line, index) => (
+                  <React.Fragment key={index}>
+                    {line}
+                    <br />
+                  </React.Fragment>
+                ))}
```
roxy-dao, line 71: Had to do this since \n wasn't inputting the text on new line... 👀 tho... it does input in a new line in console 👀

lache-melvin: did a quick google, looks like you can use CSS instead... add `whiteSpace: `pre-line` to the Typography component... maybe a little nicer?

roxy-dao: Oh the more you learn about CSS!

roxy-dao:
> Looks great! Before approving, I just want to understand if the following has been considered/might be coming soon:
> 
> * In future, we add more reports, but forget/don't think we need to add a "how to read" for one of them... should the "how to read" only show if defined? I think we'd get the `messages.how-to-read` thing rendering otherwise...
> * Same goes with if I have any custom reports uploaded! They shouldn't have a how-to-read...
> * If I upload a custom version of the expiring items report, the how to read instructions could be wrong...
> 
> All this to say, wondering if the instructions should be associated with the report at a different level than just a translation key? Possibly returned from backend?

- We can exclude how-tos from all custom reports
- We currently don't have a way to set 'how-tos' in backend right now, and could definitely be considered in the future. Current translations for reports work like this: Locale passed in from frontend when the report is generated, then stringified and generated into template.

lache-melvin (approved):
LGTM! 🙏 Still think report `code` would be more reliable `name` for the translation key, but happy to merge for now, especially if description likely to move to backend at some point 👀

PR #5097, opened 2024-10-13
5090 bug fix eternal loading 

### LINKED ISSUE #5090: Quitting report filter givings eternal loading on top of existing loaded report
## What went wrong? 😲
Could possibly be from #4991 change. 

https://github.com/user-attachments/assets/04040454-634b-4ca8-8b4e-ed4bda0abf41

## How to Reproduce 🔨

1. Go to a report
2. Click on the filter button
3. cancel
4. See loading

## Your environment 🌱

- Open mSupply Version: 2.3-RC8

- Platform:

- Database type:

fergie-nz:
Fixes #5090

# 👩🏻‍💻 What does this PR do?

Adds check so set to loading isn't called if report arguments do not change

# 🧪 Testing

Attempt to replicate issue in #5090 

# 📃 Documentation

### DIFF
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -54,10 +54,13 @@ const DetailViewInner = ({
   t: TypedTFunction<LocaleKey>;
 }) => {
   const { setCustomBreadcrumbs } = useBreadcrumbs(['reports']);
-  const [errorMessage, setErrorMessage] = useState('');
+  const [state, setState] = useState<
+    | { s: 'loading' }
+    | { s: 'error'; errorMessage: string }
+    | { s: 'loaded'; fileId: string }
+  >({ s: 'loading' });
   const { mutateAsync } = useGenerateReport();
-  const [reportLoading, setReportLoading] = useState<boolean>(true);
-  const [fileId, setFileId] = useState<string | undefined>();
+
   const { print, isPrinting } = usePrintReport();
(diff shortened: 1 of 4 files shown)


andreievg (changes requested):
@fergie-nz we actually already have isLoading state and that's fileId === undefined, but that also has to be combined with errorMessage state to show the right thing, I suggest we add a state object rather then keeping state with 3 different state variable: https://github.com/msupply-foundation/open-msupply/compare/5090-bug-fix-eternal-loading-...5090-bug-fix-eternal-loading-(Example-of-state-handling)

andreievg:
There is also the excel export, which I think would take awhile, do you mind creating an issue for that ?

fergie-nz:
> @fergie-nz we actually already have isLoading state and that's fileId === undefined, but that also has to be combined with errorMessage state to show the right thing, I suggest we add a state object rather then keeping state with 3 different state variable: [5090-bug-fix-eternal-loading-...5090-bug-fix-eternal-loading-(Example-of-state-handling)](https://github.com/msupply-foundation/open-msupply/compare/5090-bug-fix-eternal-loading-...5090-bug-fix-eternal-loading-(Example-of-state-handling))

That is awesome, thanks Andrei!

For some reason I had in my head that the iframe would take some time to render the report after making the request with the fileId, but I can see that is not the case on testing

andreievg (approved):
Looks great thanks Ferg !

PR #5108, opened 2024-10-14
Add Demographics to separate out usage of demographics for Vaccine Courses from the demographic projections

### LINKED ISSUE #4147: GAPS: Demographics must be linked to year.
## What went wrong? 😲

Because we have decided to records demographic against a `baseYear` to avoid confusion about when population growth is applied too, we now only have the option of creating demographic records against a particular year.

image

## Expected behaviour 🤔

When selecting a demographic in an immunisation program, it should be independant of the year.

## How to Reproduce 🔨

1. Login to Central OMS (with module:vaccine enabled)
2. Click on Demographics
3. Create a new demographic indicator
image

4. Click on Program -> Immunisation
5. Create a Program, and Vaccine Course
6. Try to select a demographic (see the year is included)

## Your environment 🌱

## Agreed solution

- [ ] Create two tables, one for demographic (independent of year) and one for demographic projection (which does include the baseYear)
- [ ] Update the Vaccine Course component to select the demographic (not the projection) record.
- [ ] Automatically create both the demographic and projection on save in the Demographics Edit page.

jmbrunskill:
Fixes #4147

# 👩🏻‍💻 What does this PR do?

This will break sync if we have someone using Vaccine Cards before this change goes in.

e.g. If you upgrade central server to 2.4 and you have 2.3.0 sites, they won't get any new demographic_indicators which might be needed for vaccine courses.
Also vaccine courses will not have a demographic_indicator_iid which the 2.3.0 site won't know how to deal with.

Similarly if you upgrade the remote sites first they won't know what to do with `demographic_indicators` records from central so they'll break too.

# 🧪 Testing

- [ ] Create an Immunisation Program, Demographic Projections, and Vaccine Courses
- [ ] Test vaccine card (see vaccine card setup)
- [ ] Test that everything syncs correctly to remote sites.

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_04_00/add_demographic_table.rs
```diff
@@ -0,0 +1,46 @@
...
+            r#"
+                CREATE TABLE demographic (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    name TEXT NOT NULL
+                );
+            "#
+        )?;
+
+        // Insert any existing data into the new table
+        // We'll assume we don't have duplicate names (yet!)
```
lache-melvin, line 26: In the very low likelihood that somewhere, someone plays with demographics before upgrading to 2.4, and might end up with duplicates here... there's no major impact right? There would be duplicate named records, but we could update the indicators to point to one, then delete the other(s)? Or am I missing other implications 👀

lache-melvin (changes requested):
Looks great, thanks James. 

Getting this error during sync after upgrade... I think given its a bit late to have this in 2.3, we should handle this off chance that indicator changelog records exist? Also makes it easier for QA to upgrade 😅 Idk, what are your thoughts? Maybe it's the same conversation as the vaccine courses..

```
  "\"DIESEL_DESERIALIZATION_ERROR\" (\"DeserializeFieldError { field_name: Some(\\\"table_name\\\"), error: \\\"Unrecognized enum variant: 'demographic_indicator'\\\" }\")",
```

Otherwise just the one NIT clean up of the validate/generate pattern for me!

Per your comment and our convo yesterday, keen to get this in before vaccine card is used, thinking 2.3.1 could be good - have branch [here](https://github.com/msupply-foundation/open-msupply/tree/v2.3.1) if you want to rebase?

server/repository/src/migrations/v2_04_00/add_demographic_table.rs
jmbrunskill: Yeah I don't think it's too bad, problem if it did happen is we don't have soft delete, or sync for deletes so we can't easily clean up remote sites. I think it's pretty unlikely particularly if we get it in `2.3.1`

server/service/src/demographic/insert_demographic_indicator.rs
```diff
@@ -90,10 +116,12 @@ pub fn generate(
         year_4_projection,
         year_5_projection,
     }: InsertDemographicIndicator,
+    demographic_id: String,
 ) -> DemographicIndicatorRow {
     DemographicIndicatorRow {
         id,
-        name: name.unwrap_or_default(),
+        demographic_id,
+        name: name.unwrap_or_default(), // This should really happen due to validation, but possibly should be a required field?
```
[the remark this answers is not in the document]
jmbrunskill: Good idea!

jmbrunskill:
Re-raised for v2.3.1 (rebase was causing issues for me) 
https://github.com/msupply-foundation/open-msupply/pull/5126

jmbrunskill:
@lache-melvin Should I close this one now that i'm merging to 2.3.1?

lache-melvin:
> @lache-melvin Should I close this one now that i'm merging to 2.3.1?

yes please!

PR #5110, opened 2024-10-14
5055 fix integration tests

### LINKED ISSUE #5055: Run integration tests for 2.3
We might need to adjust as per this PR: https://github.com/msupply-foundation/open-msupply/pull/5054 

And this telegram msg:
```
Hi, we would need to do integration tests at some point for 2.3, with the latest PR for hardware_id integration tests will fail unless we set hardware id manually (this is integration test with omSupply central, since it need to run as a server which will set hardware id, but integration test won't have it, so when new site is created with test endpoin in mSupply hardware id will fail to match)
```

lache-melvin:
Fixes #5055

# 👩🏻‍💻 What does this PR do?

2.3 integration test fixes

# 🧪 Testing

- [ ] Run the integration tests

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -47,7 +47,6 @@ log-panics = { version = "2.1.0", features = ["with-backtrace"] }
 
 # TODO: Remove. async_graphql is using 1.0 but SET_COOKIE was being imported from wrong version.
 http2 = { package = "http", version = "1.0.0" }
-log-panics = { version = "2.1.0", features = ["with-backtrace"] }
 
 [workspace.lints.clippy]
 wrong_self_convention = "allow"
--- a/server/service/src/sync/test/integration/remote/program_requisition.rs
+++ b/server/service/src/sync/test/integration/remote/program_requisition.rs
@@ -21,7 +21,7 @@ impl SyncRecordTester for ProgramRequisitionTester {
         // STEP 1 - insert
         let period_schedule1 = PeriodScheduleRow {
             id: uuid(),
-            name: "Weekly".to_string(),
+            name: uuid(),
         };
         let period_schedule_json1 = json!({
             "ID": period_schedule1.id,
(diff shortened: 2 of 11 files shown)


server/service/src/sync/test/integration/remote/program_requisition.rs
```diff
@@ -21,7 +21,7 @@ impl SyncRecordTester for ProgramRequisitionTester {
         // STEP 1 - insert
         let period_schedule1 = PeriodScheduleRow {
             id: uuid(),
-            name: "Weekly".to_string(),
+            name: uuid(),
```
lache-melvin, line 24: This one I was a little surprised hadn't failed before now? It does a find_one_by_name in the translation, which once you've run the integration tests a few times, always finds the first one, rather than this newly created one

server/Cargo.toml
```diff
@@ -47,7 +47,6 @@ log-panics = { version = "2.1.0", features = ["with-backtrace"] }
 
 # TODO: Remove. async_graphql is using 1.0 but SET_COOKIE was being imported from wrong version.
 http2 = { package = "http", version = "1.0.0" }
-log-panics = { version = "2.1.0", features = ["with-backtrace"] }
```
roxy-dao, line 50: Why was this removed? 👀

lache-melvin: Oh... it's above on line 46... maybe my base branch is out of date, one mo...

roxy-dao: I did see a commit in 2.3 that removed it 👀  was committed straight to 2.3?

lache-melvin: yeah... there's only one in 2.3 though, I'm confused..

lache-melvin: looks right now

roxy-dao: tests pass tho!

roxy-dao (approved):
Thanks Lache

PR #5116, opened 2024-10-14
5022 fix ledger issues in backdated prescriptions

### LINKED ISSUE #5022: Create Ledger Issue of Available Stock with backdated Prescription
## What went wrong? 😲

I can't quite work out if the UX problem above is related to a bug that breaks my ledger. I think the backend is getting different updates to what is ultimately on the frontend, and "available" stock is getting updated incorrectly and irrecoverably through UI. On a **back dated** prescription (if it's current dated it seems fine. What if it's just an __older__ prescription? Should that be treated as essentially the same as a back dated prescription? Hmmm) when you modify quantities on a line that already had a quantity, it appears to adjust the stock line's available quantity down irrecoverably. Even setting issued to 0, it will not return to the baseline available. Doing it repeatedly will trend available further downward. The second line in the video I have churned lot - the available quantity should 458, not 88! Likewise the top line should have returned to 292, but is at 269. Changing the date does not fix this. Nor does deleting the line or even the whole prescription, so this is certainly the stock line irrecoverably adjusted:

https://github.com/user-attachments/assets/2bf42f58-2216-49cf-915a-e2c9b2300004

Screenshot_20241007_144316_Open mSupply

image

image

image

If I could guess, I'd guess these bugs are from 2 footguns in react:
1. not using prev state when updating state
2. useEffect race conditions making lots of rerenders and state jumping around.

Also, if the F/E is handling updating of available then I think that is wrong - the backend should be responsible at least for the final numbers. 

## Expected behaviour 🤔

There should be no ledger problems created by updating the quantities issued on any prescription, current or backdated.

## How to Reproduce 🔨

1. Have a back dated prescription and a stock line with some good history data. Ideally with days between adjustments and an adjustment today adding some amount, say 50. (In my datafile I have multiple stock lines however IDK if having multiple stock lines for the item is a prerequisite, consider it if you can't reproduce)
2. In the back dated prescription add that line and set the quantity to available and OK
3. Edit the same line and reduce it and press OK
4. Again but increase it some amount and press OK
5. Repeat 3 and 4 chaotically until the available amount starts decreasing erratically
6. Check ledger back on OG to confirm damage.

## Your environment 🌱

- Open mSupply Version: 2.3.0-RC5 apk on Samsung Tab A8 SM-X200
- Legacy mSupply Central Server Version: `master` on 7.17.15.
- OMS Central: `develop` on 2.3.0-RCX

- Database type: sqlite

Chris-Petty:
**IDK if this branch base is right or should it be v2.3.1?**

Fixes #5022

# 👩🏻‍💻 What does this PR do?

- Tweaks the logic of validation to not munge the "batch pair"
- Tests cover ledger problems
- Also made a struct field adjustment type no longer optional - there was an error always thrown in if it was missing. This is in a separate commit and can be reverted if unhappy.

# 🧪 Testing

- [ ] Have a stock line with some amount of stock introduced in the past, and some more introduced today
- [ ] Make a prescription 
- [ ] Backdate it between today and the date of stock introduced in the past
- [ ] Add a line of the stock above and press OK (Available should be the same if you open the line again - there should be no ledger issues created)
- [ ] Update the line with a new value (Available should be the same if you open the line again - there should be no ledger issues created)
- [ ] In outbound shipments experiment with adding/removing lines, adding placeholders and allocating them etc. All works as expected.
- [ ] No situation where you're inappropriately rejected from a change because there isn't enough stock left when there clearly is (e.g. blocked from issuing 30 when there is 30 available)

### DIFF
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -1,5 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import {
+  BasicSpinner,
   LocaleKey,
   noOtherVariants,
   NothingHere,
@@ -23,7 +24,7 @@ import { JsonData } from '@openmsupply-client/programs';
 
 export const DetailView = () => {
   const { id } = useParams();
-  const { data: report } = useReport(id ?? '');
+  const { data: report, isLoading } = useReport(id ?? '');
   const t = useTranslation();
 
   const {
@@ -32,6 +33,9 @@ export const DetailView = () => {
 
   const reportArgs =
     (reportArgsJson && JSON.parse(reportArgsJson.toString())) || undefined;
+  if (isLoading) {
+    return <BasicSpinner messageKey="loading" />;
+  }
 
   return !report?.id ? (
     <NothingHere body={t('error.report-does-not-exist')} />
@@ -52,6 +56,7 @@ const DetailViewInner = ({
   const { setCustomBreadcrumbs } = useBreadcrumbs(['reports']);
   const [errorMessage, setErrorMessage] = useState('');
   const { mutateAsync } = useGenerateReport();
+  const [reportLoading, setReportLoading] = useState<boolean>(true);
   const [fileId, setFileId] = useState<string | undefined>();
   const { print, isPrinting } = usePrintReport();
   const { updateQuery } = useUrlQuery();
@@ -180,14 +185,24 @@ const DetailViewInner = ({
       <ReportArgumentsModal
         key={report.id}
         report={reportWithArgs}
-        onReset={() => setReportWithArgs(undefined)}
+        onReset={() => {
+          setReportWithArgs(undefined);
+          setReportLoading(true);
+        }}
         onArgumentsSelected={generateReport}
       />
-
-      {!fileId ? (
-        <NothingHere body={errorMessage} />
+      {reportLoading && (
+        <BasicSpinner messageKey="messages.loading-report"></BasicSpinner>
+      )}
+      {fileId ? (
+        <iframe
+          src={url}
+          width="100%"
+          style={{ borderWidth: 0 }}
+          onLoad={() => setReportLoading(false)}
+        />
       ) : (
-        <iframe src={url} width="100%" style={{ borderWidth: 0 }} />
+        !reportLoading && <NothingHere body={errorMessage} />
       )}
     </>
   );
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -128,7 +128,6 @@ const DetailViewInner = ({
 
   const openReportArgumentsModal = useCallback(() => {
     setReportWithArgs(report);
-    setReportLoading(false);
   }, []);
 
   const printReport = useCallback(() => {
@@ -188,9 +187,9 @@ const DetailViewInner = ({
         report={reportWithArgs}
         onReset={() => {
           setReportWithArgs(undefined);
+          setReportLoading(true);
         }}
         onArgumentsSelected={generateReport}
-        onSetLoading={() => {setReportLoading(true)}}
       />
       {reportLoading && (
         <BasicSpinner messageKey="messages.loading-report"></BasicSpinner>
--- a/client/packages/reports/src/DetailView/DetailView.tsx
+++ b/client/packages/reports/src/DetailView/DetailView.tsx
@@ -128,6 +128,7 @@ const DetailViewInner = ({
 
   const openReportArgumentsModal = useCallback(() => {
     setReportWithArgs(report);
+    setReportLoading(false);
   }, []);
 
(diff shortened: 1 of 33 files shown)


client/packages/reports/src/DetailView/DetailView.tsx
Chris-Petty, line 1: This file and others in `/client/` are from commits that I've acquired? Is this from the rebasing mess?

server/service/src/invoice_line/stock_out_line/update/mod.rs
```diff
@@ -18,7 +18,7 @@ use super::StockOutType;
 #[derive(Clone, Debug, PartialEq, Default)]
 pub struct UpdateStockOutLine {
     pub id: String,
-    pub r#type: Option<StockOutType>,
+    pub r#type: StockOutType,
```
Chris-Petty, line 21: I got annoyed at one point that I neglected this field and immediately got failing tests due to the `NoInvoiceType` error being thrown... appears it's not optional so I just made it not optional. I'm not 100% sure the GQL layer provides this 100% of the time, though.

If the argument is we can't trust the frontend to send it, I think that's moot as surely then we can't trust it to send the CORRECT one either.

client/packages/reports/src/DetailView/DetailView.tsx
roxy-dao: I've reverted the change (:

It's not supposed to be in 2.3 yet. There's an open PR for it

server/service/src/stock_line/historical_stock.rs
```diff
@@ -63,7 +63,7 @@ pub fn get_historical_stock_lines_available_quantity(
         min_available_and_pack_size
             .entry(stock_line_id)
             .and_modify(|m| {
-                m.total = m.total - quantity;
+                m.total -= quantity;
```
jmbrunskill, line 66: Is this a Clippy change? I prefer the old way :(

server/service/src/invoice_line/stock_out_line/update/mod.rs
jmbrunskill: Slightly worried this was done for a reason in the frontend, feels like a risky change to make given there's not much testing time left...

server/service/src/invoice_line/stock_out_line/update/mod.rs
```diff
@@ -625,37 +624,86 @@ mod test {
...
             verified_datetime: None,
             status: InvoiceStatus::Picked,
-            backdated_datetime: Some(datetime.clone()), // Backdated to 2 days ago
+            backdated_datetime: Some(datetime), // Backdated to 2 days ago
             ..Default::default()
         };
 
         prescription_invoice.upsert(&context.connection).unwrap();
 
         // Add a stock out line to the prescription (using all available stock)
```
jmbrunskill, line 637: Is this comment wrong now, are we only allocating 5 out of 10 available?

jmbrunskill (commented):
Looks generally good to me, thanks for finding and fixing this!
Haven't tried it out but I'm assuming Roxy is on that!

jmbrunskill:
> Haven't tried it out but I'm assuming Roxy is on that!

Ok turns out she's not! I'll do some smoke testing before approving!

server/service/src/invoice_line/stock_out_line/update/mod.rs
Chris-Petty: I agree 😁

server/service/src/stock_line/historical_stock.rs
roxy-dao: I prefer this change o;

Chris-Petty: I think this one was purely subjective! I'm not attached to though, just passing through.

I've actually turned clippy off since about halfway through doing this fix - it was just waaaay too slow on our codebase. Cargo check takes a few seconds, but clippy was like 30s+. It was so painful!

server/service/src/stock_line/historical_stock.rs
```diff
@@ -63,7 +63,7 @@ pub fn get_historical_stock_lines_available_quantity(
         min_available_and_pack_size
             .entry(stock_line_id)
             .and_modify(|m| {
-                m.total -= quantity;
+                m.total = m.total - quantity;
```
roxy-dao: 😢

Chris-Petty: Lol, you want the short one back???

jmbrunskill: IMHO
 `m.total -= quantity;` looks too much like ` m.total = quantity;`
I had to look twice which made me worry for readability but obviously both are fine!

Chris-Petty: I agree both are fine so I'm leaving it as is 😁

roxy-dao:
How's smoke testing going~?

jmbrunskill (approved):
Looks good to me, haven't broken it :)

Nice work!

PR #5126, opened 2024-10-15
4147 gaps demographics must be linked to year 2.3.1

### LINKED ISSUE #4147: GAPS: Demographics must be linked to year.

jmbrunskill:
Fixes #4147

# 👩🏻‍💻 What does this PR do?

Adds demographics table, which is created from records created for demographic indicators, but doesn't include year, or projection information (only required for GAPS)

Changes vaccine card to use this new table.

Removes demographic_indicators from sync (replaced with demographics)

This will break sync if we have someone using Vaccine Cards before this change goes in.

e.g. If you upgrade central server to 2.4 and you have 2.3.0 sites, they won't get any new demographic_indicators which might be needed for vaccine courses.
Also vaccine courses will not have a demographic_indicator_iid which the 2.3.0 site won't know how to deal with.

Similarly if you upgrade the remote sites first they won't know what to do with `demographic_indicators` records from central so they'll break too.

# 🧪 Testing

- [ ] Create an Immunisation Program, Demographic Projections, and Vaccine Courses
- [ ] Test vaccine card (see vaccine card setup)
- [ ] Test that everything syncs correctly to remote sites.

# 📃 Documentation

  1.
  2.


lache-melvin:
Just getting this test failure now, need to update package.json, sorry! 😅 

``Err` value: MigrationAboveAppVersion(Version { major: 2, minor: 3, patch: 1, pre_release: None }, Version { major: 2, minor: 3, patch: 0, pre_release: None })`

jmbrunskill:
> Just getting this test failure now, need to update package.json, sorry! 😅
> 
> `Err value: MigrationAboveAppVersion(Version { major: 2, minor: 3, patch: 1, pre_release: None }, Version { major: 2, minor: 3, patch: 0, pre_release: None })`

Dang, thought I'd run tests but clearly not!

server/repository/src/migrations/v2_03_01/add_demographic_table.rs
```diff
@@ -0,0 +1,67 @@
...
+            sql!(
+                connection,
+                r#"
+                ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'demographic';
+            "#
+            )?;
+        }
+
+        // Essentially in `add_demographic_table.rs` we are moving the demographic_indicator table to a new table called demographic (while maintaining demographic_indicators on central (not synced)
+        // So we need to update any changelog records that reference the demographic_indicator table to reference the new demographic table
```
lache-melvin, line 54: Thanks for the comments! 🔥

lache-melvin (approved):
Working great 🔥Thanks for all this effort James!

Didn't test the sync between 2.3.0/2.3.1 remote/central, but I think that's okay, hoping we can just avoid that with deployment management 😁

Issue #5149, opened 2024-10-16
Handle unbackdating a prescription properly


jmbrunskill:
## The suggestion

Currently if we back date a prescription, then set it today's date, it's not clear what times should be applied.
At the moment, the current time will be set as teh backdated time, but is likely to be after the created_date so it doesn't really make sense...

```rust
            // if the backdated_datetime is in the future, we assume we actually mean now
            // This is a fix for where backdated_date is set to the end of the day even though the invoice is created in the morning
            // TODO: backdated datetime Should be a nullable update, and should be unset if it's in the future
            let patch = match patch.backdated_datetime {
                Some(backdated_datetime) => {
                    match backdated_datetime > chrono::Utc::now().naive_utc() {
                        true => UpdatePrescription {
                            backdated_datetime: Some(chrono::Utc::now().naive_utc()),
                            ..patch
                        },
                        false => patch,
                    }
                }
                None => patch,
            };
```

### Why should we invest time in this?

Will hopefully result in fewer bugs, and cleaner databases.

### Are there any risks associated with this change?

Backdating is highly problematic in general so could break something that was working.

### How much effort is required?

A couple of hours

### Agreed Solution

Deliberately unset the backdated time in the frontend end and handle it in the backend as a nullable update.

jmbrunskill:
Duplicate of #4979

marthakeezy:
Passed the test ✅

Issue #4979, opened 2024-10-03
When backdating an invoice to `today` the dates aren't correct


jmbrunskill:
If backdated time is after created_datetime, it's sort of considered not a backdated invoice, although the field is populated in the db which could cause some confusion in reporting (if that was to happen) e.g. # of backdated invoiced could be incorrect if people set it to today a lot...

_Originally posted by @jmbrunskill in https://github.com/msupply-foundation/open-msupply/pull/4914#discussion_r1780383673_

jmbrunskill:
Possible Solution: Just unset the backdated_datetime if it's set to something > `created_datetime`

andreievg:
Triage, restrict date to be only settable before prescription date, probably need to refine this together with general issue on how to set backdate date

lache-melvin:
Closed by #5168

marthakeezy:
Passed the test! Moving to Done ✅

marthakeezy:
@jmbrunskill the issue came back with `V2.3.2-RC1`

<img width="1252" alt="Screenshot 2024-11-08 at 12 18 17 PM" src="https://github.com/user-attachments/assets/308c73ec-4930-4bfc-ba60-e2862271a455">

PR #5150, opened 2024-10-16
Bug fix for backdated prescriptions

### LINKED ISSUE #5146: Cannot backdate prescriptions - "stock not available" when it should be
## What went wrong? 😲

Video attached -> [here](https://drive.google.com/drive/folders/1vVZi9r26J4W4NIYqGNyWa30mAjvphiHt). It's long and hefty but captures at least 3 errors?

- Stock of 878 should be available for today and yesterday. And it is! But requires a bit of a workaround, as on the Ledger you will be able to see 877 issued in a prescription. However the system returns errors when trying to issue that quantity.
- When there is an error, it should show up once and not have duplicates occurring at the same time
- Available quantity showed as double (timestamp 2:28)

What concerns me is if you set a smaller value (5 instead of 878), it works fine both ways. It’s only if you set a higher number issues occur both ways. 877 or 828 causes issues too. Only one stock line for the item I was testing (e.g. only one batch). 

Database files attached ->>> [here](https://mega.nz/fm/NEw0lBCT)

## Expected behaviour 🤔

Able to backdate prescriptions without random errors occuring.

## How to Reproduce 🔨

1. Select an item that had some stock movement in the past, create an Outbound Shipment or a Prescription for this item for today.
2. Create a second prescription with that same item -> try to backdate it -> see if you run into same errors.
3. Might be quite hard to recreate so attaching my database above and in the video you can see an item I was testing. 

## Your environment 🌱

- Open mSupply Version: V2.3.0-RC9

- Platform: Tablet (apk)

- Database type: SQLite3

jmbrunskill:
Fixes #5146 

# 👩🏻‍💻 What does this PR do?

When we changed to setting the backdated invoice time to the end of the day, we broke the historical stock levels calculation.

This is because we're querying for all stock movements from `Now()` back until the backdated time.
However, the code was assuming the current invoice would be included in calculations to see how much stock was available at that time. If the current invoice is in the future as far as the database is concerned it will not be seen as a stock movement.

This changes the behaviour so that if the backdated time is in the future, we set it to the current time.
This slightly breaks the behaviour in that stock introduced after the backdated time is not available to be used in a prescription, but I think this is a rare scenario and arguably wrong to allow an invoice to issue stock introduced after it's picked date anyway.

Created an issue to clean this up. https://github.com/msupply-foundation/open-msupply/issues/5149
Although we already have this, which is kind of a duplicate... https://github.com/msupply-foundation/open-msupply/issues/4979

What else might have gone wrong...

# 🧪 Testing

- [ ] Backdate a prescription to say, yesterday
- [ ] Allocate all available stock to it (including todays)
- [ ] Change the prescription date to today
- [ ] Check there's no error
- [ ] Check you can't change the prescription date to before the stock was introduced
- [ ] Check other things you can think of :)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/invoice/prescription/update/mod.rs
+++ b/server/service/src/invoice/prescription/update/mod.rs
@@ -60,6 +60,22 @@ pub fn update_prescription(
     let invoice = ctx
         .connection
         .transaction_sync(|connection| {
+            // if the backdated_datetime is in the future, we assume we actually mean now
+            // This is a fix for where backdated_date is set to the end of the day even though the invoice is created in the morning
+            // TODO: backdated datetime Should be a nullable update, and should be unset if it's in the future
+            let patch = match patch.backdated_datetime {
+                Some(backdated_datetime) => {
+                    match backdated_datetime > chrono::Utc::now().naive_utc() {
+                        true => UpdatePrescription {
+                            backdated_datetime: Some(chrono::Utc::now().naive_utc()),
+                            ..patch
+                        },
+                        false => patch,
+                    }
+                }
+                None => patch,
+            };
+
             let (invoice, status_changed) = validate(connection, &ctx.store_id, &patch)?;
             let GenerateResult {
                 batches_to_update,
--- a/server/service/src/invoice/prescription/update/mod.rs
+++ b/server/service/src/invoice/prescription/update/mod.rs
@@ -62,7 +62,8 @@ pub fn update_prescription(
         .transaction_sync(|connection| {
             // if the backdated_datetime is in the future, we assume we actually mean now
             // This is a fix for where backdated_date is set to the end of the day even though the invoice is created in the morning
-            // TODO: backdated datetime Should be a nullable update, and should be unset if it's in the future
+            // TODO: backdated datetime Should possibly be a nullable update, and should be unset if it's in the future?
+            // https://github.com/msupply-foundation/open-msupply/issues/4979
             let patch = match patch.backdated_datetime {
                 Some(backdated_datetime) => {
                     match backdated_datetime > chrono::Utc::now().naive_utc() {
--- a/server/service/src/stock_line/historical_stock.rs
+++ b/server/service/src/stock_line/historical_stock.rs
@@ -10,6 +10,7 @@ use crate::{service_provider::ServiceContext, ListError, ListResult};
 
 use super::query::get_stock_lines;
 
+#[derive(Debug)]
 struct MinAvailableAndPackSize {
     pack_size: f64,
     min: f64,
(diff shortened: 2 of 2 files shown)


server/service/src/invoice/prescription/update/mod.rs
```diff
@@ -60,6 +60,23 @@ pub fn update_prescription(
     let invoice = ctx
         .connection
         .transaction_sync(|connection| {
+            // if the backdated_datetime is in the future, we assume we actually mean now
+            // This is a fix for where backdated_date is set to the end of the day even though the invoice is created in the morning
+            // TODO: backdated datetime Should possibly be a nullable update, and should be unset if it's in the future?
+            // https://github.com/msupply-foundation/open-msupply/issues/4979
+            let patch = match patch.backdated_datetime {
```
jmbrunskill, line 67: Not entirely happy with this, the backdated_datetime needs to go into the Validate method, and also apply to the generate, so have put the logic here for now, hoping to clean up in 4979.

Possibly a frontend fix would be nicer and simpler?

server/service/src/stock_line/historical_stock.rs
```diff
@@ -38,28 +39,46 @@ pub fn get_historical_stock_lines_available_quantity(
...
-                    total,
-                },
-            )
-        })
-        .collect();
+    let mut min_available_and_pack_size: HashMap<String, MinAvailableAndPackSize> =
+        stock_lines_plus_reserved
+            .iter()
+            .map(|(stock_line, reserved_available_number_of_packs)| {
+                // Any stock already allocated to this invoice, will have reduced the available stock in stock lines.
```
jmbrunskill, line 46: Added comments here as it wasn't obvious the intention was to undo the currently allocated invoice lines to correct the total.

server/service/src/stock_line/historical_stock.rs
```diff
@@ -38,28 +39,46 @@ pub fn get_historical_stock_lines_available_quantity(
...
+                    stock_line.id.clone(),
+                    MinAvailableAndPackSize {
+                        pack_size: stock_line.pack_size,
+                        min: available,
+                        total,
+                    },
+                )
+            })
+            .collect();
+    log::debug!(
```
jmbrunskill, line 68: THinking to leave these in for future debugging purposes, very helpful!

lache-melvin (changes requested):
Can't if it's related to these changes but:

- Create backdated prescription with stock introduced on previous day (issue all stock)
- Create inventory adjustment, to introduce more stock (today)
- Try to change date of prescription to different "past" date

This should still work, I'm not trying to issue any more stock (the stuff that was issued today)?

Bit worried about the number of hard-to-pin-down bugs we've found in this area...

jmbrunskill:
Closing in favour of #5153

PR #5156, opened 2024-10-16
fix stocktake tests again

roxy-dao:
Fixes #

Third time a charm?


Chris-Petty (approved):
Whoa you got a diff that's so cool

PR #5161, opened 2024-10-17
Don't allow changing backdated date with lines assigned.

### LINKED ISSUE #5146: Cannot backdate prescriptions - "stock not available" when it should be

### LINKED ISSUE #5153: Prevent changing prescription date, once lines are allocated
## What went wrong? 😲

If we change the date of a prescription, it's hard to correctly account for stock movements as the stocklines have already been modified be the invoice, so if we allow changing of prescription date we need to somehow undo the stockline changes, but also correctly account for the stock movements which might or might not be in the date range for the new invoice date.

## Expected behaviour 🤔

If a user tries to change the invoice date for a prescription after it's had lines added, they will get a prompt

Changing the prescription date will remove all lines in this prescription, do you want to continue?

If they select yes, all lines will be deleted and the date changed.

## How to Reproduce 🔨

1. Create a prescription
2. Add a line
3. Try to change the date
4. See prompt
5. Check Yes, and No both work as expected.

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #5153
Fixes #5146

# 👩🏻‍💻 What does this PR do?

Stops users from changing prescription date when lines are assigned.

If they do they'll get a prompt asking if they want to delete lines and continue.

image

# 🧪 Testing

- Create a prescription
- Add a line
- Try to change the date
- See prompt
- Check Yes, and No both work as expected.

# 📃 Documentation

  1.Backdate Prescription


server/service/src/invoice/prescription/update/validate.rs
```diff
@@ -39,9 +37,16 @@ pub fn validate(
...
-        // Check that any lines already assigned won't create a ledger discrepancy
-        check_stock_available_at_date(connection, &invoice.id, backdated_datetime)?;
+    if patch.backdated_datetime.is_some() {
+        // Check if we have any lines allocated to this invoice, if so we can't backdate
+        let lines = InvoiceLineRepository::new(connection).query_by_filter(
+            InvoiceLineFilter::new().invoice_id(EqualFilter::equal_to(&patch.id)),
+        )?;
+        if lines.len() > 0 {
+            return Err(CantBackDate(
+                "Can't Invoice has allocated lines".to_string(),
```
jmbrunskill, line 47:
```suggestion
                "Can't backdate as invoice has allocated lines".to_string(),
```

server/service/src/invoice/prescription/update/validate.rs
```diff
@@ -39,9 +37,16 @@ pub fn validate(
...
 
-    if let Some(backdated_datetime) = &patch.backdated_datetime {
-        // Check that any lines already assigned won't create a ledger discrepancy
-        check_stock_available_at_date(connection, &invoice.id, backdated_datetime)?;
+    if patch.backdated_datetime.is_some() {
+        // Check if we have any lines allocated to this invoice, if so we can't backdate
+        let lines = InvoiceLineRepository::new(connection).query_by_filter(
+            InvoiceLineFilter::new().invoice_id(EqualFilter::equal_to(&patch.id)),
+        )?;
+        if lines.len() > 0 {
```
andreievg, line 45: there is count() method for InvoiceLineRepository

jmbrunskill: Nice, good point will change...

andreievg (changes requested):
If i click into the date and click out of it i get the delete lines msg, I suspect it should only happen when date is changed ?

jmbrunskill:
> If i click into the date and click out of it i get the delete lines msg, I suspect it should only happen when date is changed ?

On it!

jmbrunskill:
Thanks @lache-melvin for the assist!

lache-melvin:
Hold up, for some reason "onCancel" is getting called on confirm...

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
```diff
@@ -76,22 +127,16 @@ export const Toolbar: FC = () => {
             />
           </Box>
           <Box display="flex" flexDirection="column" flex={1} marginLeft={3}>
+            <Typography></Typography>
```
andreievg, line 130: Is there a need for empty typography here ?

lache-melvin:
> Hold up, for some reason "onCancel" is getting called on confirm...

Oh man this is our default ConfirmationModal behaviour 😭

andreievg (approved):
Looks great thanks.

Would have been nice if DateTimePickerInput would only call updates when value changed from default etc..

andreievg:
Maybe next time we go for the bare minimum change ? (an error message from backend and page/state refresh)

andreievg (approved):
Looks good, @lache-melvin  mentioned she tried another couple of datepickers and they didn't explode

lache-melvin:
> they didn't explode

a low bar 😁

PR #5163, opened 2024-10-17
4768 full screen rnr

### LINKED ISSUE #4768: Full screen R&R entry
It's been requested to have the edit table able to be viewed / used in a full screen mode to make the most use of the limited screen size available on tablets.

Here are some mock ups:

image

image

image

Additionally, could we have `escape` also close the full screen mode; this is an established pattern.
If doing so, a hint on entering full screen would be good to include, such as this version from chrome:

Screenshot 2024-09-10 at 1 01 29 PM

We could use an info snackbar to show the 'press esc' message in order to reuse existing functionality.

lache-melvin:
Fixes #4768

# 👩🏻‍💻 What does this PR do?

Adds full screen R&R mode:
Screenshot 2024-10-17 at 4 59 31 PM

Created this as reusably as I could - should be able to add the `FullScreenButton` to any `AppBarButtons` to allow that feature on any page 👀 

Just sets the portals (toolbar/footer/appdrawer) to `display: none`, so anything their doing isn't actually unmounted, just invisible.

# 🧪 Testing

- [ ] Full screen yourself in and out
- [ ] Test the `esc` key binding
- [ ] Try her on android!
- [ ] Generally just try to break it please hehe

# 📃 Documentation

  1. Using full screen mode
  2.


client/packages/host/src/components/AppBar/AppBar.tsx
```diff
@@ -26,30 +30,56 @@ const StyledContainer = styled(Box)(({ theme }) => ({
...
+              <Breadcrumbs />
+              <AppBarButtons />
+            </Toolbar>
+          )}
+          <AppBarContent />
+          {!isDashboard && <AppBarTabs />}
+        </StyledContainer>
+      </Box>
+
+      {/* Show the exit button in the top right of the page if full screen mode enabled */}
```
lache-melvin, line 62: Is this the best place for this button?

jmbrunskill (approved):
Looks good to me.
As for where the button goes to exit full screen, the top right makes sense but that does obscure the last column in R&R forms.
I'm ok with that for now, no obviously better place. Arguably maybe we should make it with a transparent background? So it's easier to read what ever is under it?

I think it's probably ok, as people can exit full screen to check what it is if they forget.

Tried doing 
`background-color: rgba(255, 255, 255, 0.8);`

The result is not terrible?
image

jmbrunskill:
FYI - tested on tablet too which seems to work fine

Issue #4702, opened 2024-08-28
Standard report sorting


andreievg:
## Is your feature request related to a problem? Please describe 👀

Add report arguments for column sorting in standard reports, `expiring-items`, `stock-status`, `stock-details`, `item-usage`. Will require this PR as pre-requisite https://github.com/msupply-foundation/open-msupply/issues/4701

roxy-dao:
Do in JS

fergie-nz:
I notice there is an unused variable `PrintReportSort` which takes a param from generate report queries and adds this to database queries rather than sorting with wasm functions.

I just wanted to clarify do we want to use wasm functions for sorting? Or add in this variable?

fergie-nz:
> I notice there is an unused variable `PrintReportSort` which takes a param from generate report queries and adds this to database queries rather than sorting with wasm functions.
> 
> I just wanted to clarify do we want to use wasm functions for sorting? Or add in this variable?

You can see this in reports/src/print.rs line 182 where a None is passed through rather than the optional sort

andreievg:
As per our convo yesterday `PrintReportSort` is used by graphql queries, it's kind of hard coded for print to respect the sorting done with front end for stocktakes, invoices and requisitions. 

It would be a bit harder to try and pass one the input of json form sort as PrintReportSort (to be later used by convert_data method).

My conclusion is we keep PrintReportSort and add new sort through filter arguments for now

PR #5165, opened 2024-10-17
5058 simplify report logic with wasm functions

### LINKED ISSUE #5058: Identify what reports has complex conditions and use wasm to simplify logic

Part of #4668

fergie-nz:
Fixes #5058 

# 👩🏻‍💻 What does this PR do?

Adds custom wasm functions to expiring-items and item-usage reports to remove logic from tera

I've left the other two standard reports (stock-detail and stock-status)

Adds tests + testing pathway into server dir for .test.js files

# 🧪 Testing

You can test this by building and generating standard reports and comparing to previous standard reports

# 📃 Documentation


server/reports/expiring-items/2_3_0/convert_data_js/convert_data.test.ts
```diff
@@ -0,0 +1,111 @@
...
+//     spyCalculateDaysUntilExpired.mockImplementation((x) => x);
+//     let res = processStockLines(inputData.stockLines.nodes);
+//     expect(spyCalculateDaysUntilExpired).toHaveBeenCalledTimes(2);
+//     expect(spyProcessStockLines).toHaveBeenCalledTimes(1);
+//     // expect(mockRoundDaysToInteger).toHaveBeenCalledWith(2.3);
+//     // expect(mockRoundDaysToInteger).toHaveBeenCalledWith(3.1);
+//     // expect(mockRoundDaysToInteger).toHaveBeenCalledWith(5.5);
+//     spyCalculateDaysUntilExpired.mockRestore();
+//   });
+// });
```
fergie-nz, line 111: I really wanted to get this mocked function testing working to ensure functions were being called correctly, but I couldn't figure this out.

Left here as comments in case reviewer is more familiar with this part of Jest!

server/reports/utils.js
```diff
@@ -0,0 +1,30 @@
...
+      }
+    }
+  });
+  return newNode;
+};
+
+module.exports = {
+  cleanUpObject,
+  cleanUpNodes,
+};
```
fergie-nz, line 30: Have added this to remove empty strings from json objects which will fail to parse. However, tera will throw errors if it tries to render non existent values.

I'm not sure if it is likely that we will return non existent values from queries, but might be safe to use 'ignore missing' in tera to handle this

server/reports/expiring-items/2_3_0/src/template.html
```diff
@@ -16,24 +16,18 @@
...
-                {% if diff <= arguments.monthsItemsExpire * 30 %}
-                  <span class="soon">{{ diff }}</span>
-                {% else %}
-                  <span class="all-good">{{ diff }}</span>
-                {% endif %}
-              {% else %}
-               <span class="expired">Expired</span>
-              {% endif %}
+          {% if stockLine.daysUntilExpired %}
+            {% if stockLine.daysUntilExpired <= 0 %}
```
andreievg, line 24: I would add another field 'expiryStatus' or something like that and calculate it in the convert_data

fergie-nz:
Have now added bundling to this PR so that we can run tests on the JS files.

I've kept this the same as the docs (here)[https://github.com/extism/js-pdk] so that it is easy to understand.

However this does decrease the ease for future development as there is more boilerplate to add. I am assuming we will need to add bundled wasm building for all reports if we want to add sorting to all reports?

I'm also slightly worried about the size of overhead for each version of each report. Would this be a concern?

server/server/src/lib.rs
```diff
@@ -139,6 +140,10 @@ pub async fn start_server(
             .unwrap();
     }
 
+    // SET LOG CALLBACK FOR WASM FUNCTIONS
+    info!("Setting wasm function log callback..");
+    let _ = set_log_callback(|log| println!("log {:?}", log), "info");
+
```
fergie-nz, line 146: Have added call for logging here which fixes https://github.com/msupply-foundation/open-msupply/issues/5182

I added this here because it was helpful for trouble shooting in this PR, and was a small change.

However when I unwrap, it is still throwing an error from being called multiple times. This was unexpected for me? I wonder if it needs to be elsewhere

server/reports/expiring-items/2_3_0/src/template.html
fergie-nz: I am wanting to wait to refactor further until after https://github.com/msupply-foundation/open-msupply/issues/4702

Reason being is that expiryStatus is a function of monthsItemsExpire which is an argument and currently not accessible in wasm function. 

And this PR is already massive!

Thinking of adding further refactor for another issue post adding sorting (and with that passing arguments into wasm).

What do you reckon?

package.json
```diff
@@ -7,5 +7,6 @@
     "start": "cd ./server && cargo run & cd ./client && yarn start-local",
     "build": "cd ./client && yarn build && cd ../server && cargo build --release",
     "test": "cd ./server && ulimit -n 1000 && cargo test && cd ../client && yarn test"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
andreievg, line 11: Is this required ?

server/server/Cargo.toml
```diff
@@ -48,6 +48,7 @@ regex = { workspace = true }
 actix-multipart = { workspace = true }
 futures-util = { workspace = true }
 serde_json = { workspace = true }
+extism = "1.5.0"
```
andreievg, line 51: can you use workspace pattern please like the other creates

server/server/src/lib.rs
andreievg: I think this is the correct place, use info! rather then println

server/reports/.gitignore
```diff
@@ -1,3 +1,6 @@
 # ignore all built wasm functions
 target/
-*.wasm
\ No newline at end of file
+*.wasm
+
+# Dependency directories
+**/node_modules/
```
andreievg, line 6: do we need `dist` directory ? I think we should ignore it too

server/reports/item-usage/2_3_0/convert_data_js/src/convert_data.js
```diff
@@ -0,0 +1,9 @@
+function convert_data() {
+  const res = JSON.parse(Host.inputString());
+  res.items.nodes = processItemLines(res);
```
andreievg, line 3: where is processItemLines come from here ?

andreievg (changes requested):
Looking better and better, thanks Ferg.

I have some change requests. I also have some comments and questions about test, would love to smash this out with a pair programming session after afternoon tea : )

server/server/Cargo.toml
andreievg: general rule is when more then one crate uses them, btw I was also hoping you would do the same for server crate

andreievg (approved):
Thanks, looks great, just one little suggestion left, but otherwise ready to go.

There is a bit of carry over that we talked about, would be great to capture it in follow up issues, if you haven't already. 

One more things to add to that, is to check performance on tablet, i see that standard_reports.json is around 5 mb, we might need to cache the was function and probably move the convert_data out of the json (actually we should reshape report table to be less json at some point (since we are moving it to omSupply central anyways)

fergie-nz:
FYI @andreievg the esbuild.js file requires package.json dependencies to be installed to build the bundled reports, so other devs won't be able to build reports without installing

I was thinking of adding to the script `yarn &&` to make it `yarn && node esbuild.js...`

This is also the source of the yarn packageManager showing up in the main client package.json.

I'm not familiar with this, but believe it is a method of reconciliation of dependency versions.

fergie-nz:
> Thanks, looks great, just one little suggestion left, but otherwise ready to go.
> 
> There is a bit of carry over that we talked about, would be great to capture it in follow up issues, if you haven't already.
> 
> One more things to add to that, is to check performance on tablet, i see that standard_reports.json is around 5 mb, we might need to cache the was function and probably move the convert_data out of the json (actually we should reshape report table to be less json at some point (since we are moving it to omSupply central anyways)

Will make issues for all of these

server/server/Cargo.toml
fergie-nz: Ah nice I missed the service crate extism import, thank you

Issue #5118, opened 2024-10-14
Add service + endpoint to create a Response Requisition


roxy-dao:
Add service and endpoint to create a `Response Requisition`.

See `request_requisition/insert.rs` for examples

### Moneyworks Jobcode 🧰
OMS:CIMDV

PR #5166, opened 2024-10-17
5118 add service + endpoint to create a response requisition

roxy-dao:
Part of #5118

# 👩🏻‍💻 What does this PR do?
Service and endpoint to create response requisitions (normal).

## PS ADDING LINES COMING IN NEXT PR

# 🧪 Testing

- [ ] Use GraphQL to call new endpoint

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/graphql/requisition/src/mutations/response_requisition/insert.rs
+++ b/server/graphql/requisition/src/mutations/response_requisition/insert.rs
@@ -111,8 +111,6 @@ pub fn map_error(error: ServiceError) -> Result<InsertErrorInterface> {
         // Standard Graphql Errors
         ServiceError::RequisitionAlreadyExists => BadUserInput(formatted_error),
         ServiceError::OtherPartyDoesNotExist => BadUserInput(formatted_error),
-
-        ServiceError::OtherPartyIsNotAStore => BadUserInput(formatted_error),
         ServiceError::NewlyCreatedRequisitionDoesNotExist => InternalError(formatted_error),
         ServiceError::DatabaseError(_) => InternalError(formatted_error),
     };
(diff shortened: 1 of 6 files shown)


server/graphql/requisition/src/mutations/response_requisition/insert.rs
```diff
@@ -0,0 +1,341 @@
...
+#[graphql(name = "InsertResponseRequisitionInput")]
+pub struct InsertInput {
+    pub id: String,
+    pub other_party_id: String,
+    pub max_months_of_stock: f64,
+    pub min_months_of_stock: f64,
+}
+
+#[derive(Interface)]
+#[graphql(name = "InsertResponseRequisitionErrorInterface")]
```
fergie-nz, line 27: I wonder do we rename graphql fields throughout the code? I find the renaming slightly confusing to trace types, but maybe that is just me?

roxy-dao: Yes, we use this for the Object name throughout the codebase.

Chris-Petty: (Not requesting any change!)

>Yes, we use this for the Object name throughout the codebase.

You mean the frontend code base? I think Ferg's point is that when trying to trace backend implementation it's not always intuitive to search for `InsertErrorInterface` which is generically named within this code context, but in a code search things become harder to find (and when working you have to be constantly aware of directory structure of the file you're editing to apply context to the variable name - bit of a cognitive load when trying to jump around different files.

image

Trade off though - once you are contextualised the short names are cozy and it favours copy pasting all this boiler plate.

That said not everywhere has the different names:

https://github.com/msupply-foundation/open-msupply/blob/7a31855c50d62f04c22ac8014bc10bf3abdef6e5/server/graphql/requisition/src/mutations/request_requisition/insert_program.rs#L17-L19

Chris-Petty (approved):
Looks good to me!

server/graphql/requisition/src/mutations/response_requisition/insert.rs
roxy-dao: Ah that's true. I usually go through by folder name then action to find them 🤔

PR #5168, opened 2024-10-17
4979 when backdating an invoice to today the dates arent correct

### LINKED ISSUE #4979: When backdating an invoice to `today` the dates aren't correct

jmbrunskill:
Fixes #4979 

# 👩🏻‍💻 What does this PR do?

In the generate method, if created_datetime < backdated_datetime, we set backdated date to null and picked date to now (if already set)

Usually we don't have this kind of logic in the generate function but it's the cleanest place to put it for now...

Not handling other dates and I'm pretty confident in prescriptions we can't change the date unless in picked or new status.

# 🧪 Testing

- [ ] Back date a prescription
- [ ] Set it back to today
- [ ] Check dates aren't in the future

# 📃 Documentation

  1.
  2.


lache-melvin (commented):
It works well for me, but agree we usually haven't used `generate` for this kind of change - more consistent would be to put in the nullable update... and clearer to consumer, they can clear the date, rather than it magically being unset.

Though there's also an argument for making the frontend "dumb" and making these decisions in our backend, seeing as we have to run all the validation here anyway 🤷‍♀️ 

@andreievg curious if you had many thoughts around this - how much we responsibility we put on the frontend vs backend?

andreievg:
This looks good, but I the update of invoice status in two places doesn't quite sit well with me, I would do something like this: https://github.com/msupply-foundation/open-msupply/compare/4979-when-backdating-an-invoice-to-today-the-dates-arent-correct...4979-when-backdating-an-invoice-to-today-the-dates-arent-correct(PR-suggestion)

jmbrunskill:
@andreievg Pushed a change, which I think cleans up the logic a bit. I found your suggestion a bit hard to read. Let me know if you're happy with this...

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -91,12 +93,28 @@ fn should_update_batches_total_number_of_packs(
...
+
+// Handle a change to backdated_time
+fn handle_new_backdated_datetime(invoice: &mut InvoiceRow, backdated_datetime: NaiveDateTime){
+    
+    // We need to check if we are actually needing to back_date here, if the backdated_datetime is in the future, we unset the backdated_datetime as it isn't possible to future date.
+    if invoice.created_datetime < backdated_datetime {
+        // Not actually backdated!
+        let now = Utc::now().naive_utc();
+        invoice.backdated_datetime = None;
+        replace_status_datetimes(invoice, now);
```
andreievg, line 111: I think in the example I was giving this would only happen if backdated_datetime was actually set already on the invoice, although I am happy with the way you have it (would just mean picked and allocated datetime would reset anytime backdated_datetime is set > created_datetime).

andreievg (changes requested):
I think we only want to set this only in handle_new_backdate_datetime

https://github.com/msupply-foundation/open-msupply/blob/7721ca5112694fa32146ad176051fb8c618f89bc/server/service/src/invoice/prescription/update/generate.rs#L49

Otherwise could change backdated_datetime > created_datetime, it would still be recorded on invoice, and next time status changes the datetime would be set to backdated_datetime rather then now ?

Also I think it's good to destructure to input `input_backdated_datetime`, for a bit more clarity (at least for consistency in this case), might have been easier to spot the backdated_datetime being set.

> @andreievg Pushed a change, which I think cleans up the logic a bit. I found your suggestion a bit hard to read. Let me know if you're happy with this...

Happy with your way, would be great to get some feedback though about readability of my proposal though (want to know what to avoid in the future)

server/service/src/invoice/prescription/update/generate.rs
jmbrunskill: Ah true, that is kind of nice...

jmbrunskill:
> I think we only want to set this only in handle_new_backdate_datetime
[077c563](https://github.com/msupply-foundation/open-
msupply/pull/5168/commits/077c563f58de8fe2c0720260f746f2b69b82b2d0)

Great catch thank you!

andreievg (approved):
Looks great, thanks !

PR #5195, opened 2024-10-23
4702 add report sorting via wasm functions

### LINKED ISSUE #4702: Standard report sorting

fergie-nz:
Fixes #4702 

# 👩🏻‍💻 What does this PR do?

Adds sorting to standard reports

This adds params sort and dir to argument schemas of standard reports which are used for sorting.

Also adds wasm functions for all standard reporting to conduct the sorting

Also adds tests + abstracts some report functions which are called by multiple reports (such as the sorting function)

# 🧪 Testing

- [ ] Go to reports
- [ ] Try rendering reports with various sorting options

# 📃 Documentation

- [x] **Part of an epic**: documentation will be added alongside remaining 2.4 release report changes


server/service/src/report/report_service.rs
```diff
@@ -455,15 +455,26 @@ fn wasm_sql(
     })
 }
 
+#[derive(Serialize)]
+struct WasmData {
```
andreievg, line 459: I am keen to try and standardies this, rather then manually inserting in context:https://github.com/msupply-foundation/open-msupply/compare/4702-add-report-sorting-via-wasm-functions...4702-add-report-sorting-via-wasm-functions(Use-from-serialise-rather-then-inserting-manually)

This would also mean you don't have to change existing templates (otherwise all reports would break, even existing ones ?)

server/reports/utils.js
```diff
@@ -24,7 +24,35 @@ const cleanUpObject = (node) => {
   return newNode;
 };
 
+const getNestedValue = (node, key) => {
+  key = key + "";
+  return key.split(".").reduce((value, part) => value && value[part], node);
+};
+
+const sortNodes = (nodes, sort, dir) => {
```
andreievg, line 32: I would actually prefer to use sortBy method from lodash, also I am not a big fan of drilling through the "." separator string to access nested values. mainly because it's slightly inneficient, adds complexity that can be avoided, hard to validate all use cases, and harder to transition to typescript.

I think ideally it would look like this:

Input to convert_data:

`{ data: { /* nested, ungrouped etc * / } , arguments: { .. } }`

First there is conversion method, let's say 'convert', which will take in above structure, and return the same structure but with another key, 'convertedData' or something like that, which is data ready for presentation with template.html, so acess to the data and sorting would be much simpler:

 `{% for stockLine in data.data.stockLines.nodes %}` -> `{% for row in data.convertedData %}`

`{{stockLine.item.code}}` -> `{{row.itemCode}}`

This way we can also sort on the whole object simply with

`const sorted = sortBy(convertedData, [sort])`
`if (dir === 'DESC') sorted.reverse();`

In terms of helper methods I would go for this abstraction:

``` 
function convert_data() {
  const res = JSON.parse(Host.inputString());
  const unsortedConvertedData = convert(res); // from convert.js 
  const convertedData = sort(unsortedConvertedData, DEFAULT_SORT_KEY, DEFAULT_SORT_DIRECTION) // from default_sort.js
  Host.outputString(JSON.stringify( { ..res, convertedData }));
}
```

Not requesting any of the above changes in this PR, but I do think it need to be done at some stage

andreievg (changes requested):
Thanks Ferg, looks great, there is one change request, and that's as per this comment: https://github.com/msupply-foundation/open-msupply/pull/5195#discussion_r1814027639

And there is also a comment to think about in carry over, plus this thing to think through at some point
* standard_reports.json is getting too big, shouldn't be committed at this stage (some way to generate it only when upserting report for dev/test and when building for production, figure out an alternative to extism, for bundle size)

server/service/src/report/report_service.rs
fergie-nz: This makes perfect sense and looks neat having

server/reports/utils.js
fergie-nz: That is awesome, agree the nested . structure looks fragile.

server/service/src/report/report_service.rs
andreievg: Additional keys should always be backwards compatible for JSON/Object data

andreievg (approved):
Thanks for your work and the changes @fergie-nz !

Issue #4368, opened 2024-07-04
GAPS: Capture cold storage temperature range per item variant


adamdewey:
### Additional context 💌

| User story  | B.2.8 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to specify the cold storage temperature requirements for each type of vaccine variant and also other products
So that… | the cold storage requirements can be broken down by temperature range

**Solution B.2.8**

The system will allow central users to specify the cold storage temperature requirements for vaccine variants in the Master Data product catalogue.

3 options of cold storage temperature range will be available:

- +5 °C
- -20 °C
- -70 °C

Additionally the system will allow central users to specify a cold storage temperature range for non-vaccine item variants (such as diluents).

### Moneyworks Jobcode 🧰

`CCEIDEV`

mark-prins:
Refinement: we'll leave up to the implementing team to refine further and report back to the refinement team if there are decisions to confirm

marthakeezy:
# 🧪 Testing

Tested in V2.4.0-RC2 and as a part of feature release testing :)

- [ ] Login to your central OMS
- [ ] Edit item variants
- [ ] Check you can select a cold storage location type (I'm assuming you have them too!) 
- [ ] Make sure changing/updating works correctly
- [ ] Everything displays right

marthakeezy:
Passed!

Issue #3923, opened 2024-05-16
GAPS: Specify item variants that should be bundled with other item variants


mark-prins:
Part of epic #3662 

## Acceptance criteria
* [ ] Able to record items which should be bundled with a particular item variant

## Describe the solution you'd like 🎁
On central server, add another tab to the item detail page.
This lists all items which should be bundled with the current item.
Have the options to 
* add an item
* remove an item
* specify the quantity when adding a bundled item or editing an existing bundled item

### Additional context 💌

| User story  | B.2.9 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to specify that each type of vaccine should be 'bundled' with diluents and / or syringes in certain ratios
So that… | any additional cold storage requirements can be taken into account if the diluents need to be stored cold too (and any additional dry storage requirements for syringes can be known too) 

**Solution B.2.9**

The system will allow central users to specify ‘Bundling’ rates for vaccines in the Master Data product catalogue. 
An unlimited number of other products from the Master Data product catalogue may be ‘bundled’ with vaccines. 
For each ‘bundled’ product the central user will specify the ratio in which they are bundled. 
The default ratio will be 1:1

Note: we need to be clear that bundling is used for storage estimate needs and will not automatically update your shipments with bundled items. Can we add an info note on the page when editing bundled items please

### Moneyworks Jobcode 🧰
`CCEIDEV`

adamdewey:
So here the quantity should be a ratio, which could be < or > than 1:1

Example 1

1 vial of `Vaccine B` should be bundled with 10 units of `Syringe 1` - **Ratio: 1:10**
(assuming base units of each are vial and unit respectively)

Example 2

1 vial of `Vaccine B` should be bundled with 0.5 bottles of `Diluent 1` - **Ratio = 2:1** 
(assuming base units of each are vial and bottle respectively)

The idea being that these ratios could be used for: 

- GAPS calculations
- Distribution (possibly)
- Procurement module (in the future)

mark-prins:
Yes, I was thinking to have a decimal value, so you would enter `10` for the first example and `0.5` in the second.

Do you think entering as a ratio would be easier to understand though? Me, I'm mentally converting the ratio to a fraction / decimal.. so I find it easier to enter it as numbers, like you have in the initial part of the example:

**Vaccine B**
Bundled with
|    Item    |  Quantity  |   Unit   |
|----------|---------:|--------|
| Syringe 1 |      10      | Unit     |
[...]

Issue #3917, opened 2024-05-16
GAPS: Capture vaccine item variant volume requirements


mark-prins:
Part of epic #3662 

## Acceptance criteria
* [ ] Able to specify the type of vaccine storage which is used in a particular facility
* [ ] Able to specify the storage requirement of a vaccine variant, for `Primary` `Secondary` and `Tertiary`

## Describe the solution you'd like 🎁
Add custom field support for items
* add a column to the item table for `custom_data`
* include this column in sync
* extend the mSupply item import to include custom_data
* display custom field data in the item detail page
Add a custom field to the footrunner script used to create names for the vaccine storage type

### Additional context 💌

| User story  | B.2.7 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to specify the cold storage volume requirements per dose for **each presentation** of each type of vaccine
So that… | this volume can be taken into account when calculating cold storage space requirements

**Solution B.2.7**

The system will allow central users to specify individual vaccine volume requirements (in litres) on the Master Data product catalogue. 
This data is required to calculate the total cold storage capacity required per person per immunization program. 

| User story  | B.2.7.1 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | to specify the cold storage volume requirements per dose for **each presentation** of each type of vaccine per Primary, Secondary and Tertiary packaging
So that… | the volume calculation is accurate when different forms of packaging are used to store the vaccines

**Solution B.2.7.1**

The system will allow central users to specify up to 3 levels of vaccine volume requirements (in litres) on the Master Data product catalogue to reflect the different forms of packaging used at different levels of the supply chain. 

| User story  | B.2.7.2 |
| :------------- | :------------- |
As a… | National EPI manager
I want… | the appropriate form of packaging (Primary, Secondary, Tertiary) used by each facility to be captured
So that… | the volume calculation is accurate for each facility

**Solution B.2.7.2**

The system will allow facilities to specify the level of packaging they are using to store their vaccines. 
The facility users will be able to select one of the levels for their facility as defined by the central users in the Master Data product catalogue. 

**Extra context**

Example of WHO IGA item granularity: 

![image](https://github.com/user-attachments/assets/9b908ffc-f4f5-472e-bdf1-5a13a21b5461)

### Moneyworks Jobcode 🧰
`CCEIDEV`

adamdewey:
There was an oversight in the requirements, User Story B.2.7.1 should read: 

| User story  | B.2.7.1 |
[...]

PR #5196, opened 2024-10-23
4789 item variant UI (PT 1)

### LINKED ISSUE #4789: GAPS: Implement item variants: Item UI changes
Part of epic #4362

There are several aspects of the GAPS work which require item variant capabilities
For reference this is the full [item variant PRD](https://docs.google.com/document/d/1fRZslRohRGaZnUo13a2eKZjfO0DGWwXtIFsbEah7A4Y/edit#heading=h.hyld495fuxe5)

We won't need all aspects of that; but we will need the ability to create and edit and sync item variants.

The variants will need to have these properties for the GAPS work:

| Name | Type   | Reference |
|:------ |:------ |:---- |
| Volume (x3) | float | #3917 |
| Bundle items | reference to n item variants | #3923 |
| Temperature range | enum | #4368 |

See #5141

UI - on OMS central:

image

### Handling stock

Done in separate issue: #5145

lache-melvin:
Fixes #4789

# 👩🏻‍💻 What does this PR do?

This is part one, it was getting big enough! Adds base UI for item variants, with a mocked API sending some dummy data.

Screenshot 2024-10-24 at 8 43 21 AM

Still need to do the save mutation/manage edit state, and the manufacturer search component.

Things that will be done in separate issues:
- cold storage type dropdowns #4368
- The packaging variants edit modal #3917
- Bundling #3923

# 🧪 Testing

- [ ] On OMS central
- [ ] Items -> item -> there is a variants tab
- [ ] Add variant button works
- [ ] It looks kind of like the design
- [ ] When running not as central - no variants tab

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/locales/pt/common.json
```diff
@@ -1032,5 +1032,72 @@
   "title.stock-on-hand": "Estoque Disponível",
   "tools": "Ferramentas",
   "warning.caps-lock": "Advertência: Caps Lock está ativada",
-  "warning.nothing-to-supply": "Nada para fornecer!"
+  "warning.nothing-to-supply": "Nada para fornecer!",
```
lache-melvin, line 1035: Ah sorry, ignore these, I think I pulled a more recent version of develop into my branch than the base of this PR 👀

adamdewey:
Nice!

@lache-melvin  Instead of `Cold storage type` can we use something a little less restrictive such as `Storage temperature` ?

adamdewey:
Actually, I think just `Temperature` would be fine and more concise!

When we have the rest of the data fields in there can we also get @mariyamsupply to cast her eye over how best to refine the layout? 🙏

lache-melvin:
Oh yeah, I missed doses per unit 😅 FYI am going off this design: https://github.com/msupply-foundation/open-msupply/issues/4789#issuecomment-2427772421

adamdewey:
Thanks @lache-melvin - apologies, I also added a comment after your commit that maybe `Temperature` would be better as it's more concise?

For the layout, I see - hmm, it sort of looks a bit confusing to me with the rows of the 2 tables not lining up but being next to each other without very clear delineation? 

I think likely that eventually there will be a lot of information in each item variant - for example when it starts being used for stock management, having barcodes associated with them etc. 

Am conscious we have limited time / budget in this release - wondering if there are some simple suggestions that @mariyamsupply could give? Otherwise I fear we'll condemn it to the same depressing fate as the Item page 😿

lache-melvin:
@adamdewey Updated the screenshot to latest changes, but don't want to spend too long on this (it's central server only UI)

This is using Mariya's designs as best I can for now with the component library we have... hopefully in future we can iterate on designs before we get into dev time! 😅

adamdewey:
Ah it's central UI only?

Ok, will be interested to see how item variants are displayed to the standard user!

Totally agree it would be good to review designs before getting to dev stage, if only we had more time 😅

Thanks Laché!

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
```diff
@@ -0,0 +1,214 @@
...
+} from '@openmsupply-client/common';
+import { ItemPackagingVariantsTable } from './ItemPackagingVariantsTable';
+import { ItemVariantFragment, PackagingVariantFragment } from '../../../api';
+
+const dummyPackVariants: PackagingVariantFragment[] = [
+  {
+    __typename: 'PackagingVariantNode',
+    id: '1',
+    packagingLevel: 1,
+    name: 'Primary',
```
jmbrunskill, line 28: I was thinking we might have a mapping something like `label.packaging_level.1` = `Primary` so it was translatable, and potentially could still allow the user to still set a name, but I think this is good. We might need some way to default to this when we add new variants I guess?

server/graphql/types/src/types/item.rs
```diff
@@ -133,6 +135,19 @@ impl ItemNode {
         Ok(result)
     }
 
+    // TODO: maybe do this in a loader!
```
jmbrunskill, line 138: Loader does sound useful. I think we need to think about how we can split out our types though, as in some cases you won't want to load all the extra stuff in on item (if you just need name and code for example). I'm  not sure we have a good pattern established for that?

server/graphql/types/src/types/item_variant.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+#[Object]
+impl ItemVariantNode {
+    pub async fn id(&self) -> &String {
+        &self.item_variant.id
+    }
+
+    pub async fn item_id(&self) -> &String {
+        &self.item_variant.item_link_id // TODO join to item for item_id
```
jmbrunskill, line 20: Just do an item loader? Not sure what we'll actually use this for?

I think I remember @andreievg saying that repository or service layer should map the item_link_id to item_id.
Is there a wiki about this?

Not for this PR but I realise I'm not clear on how we are doing it :)

server/graphql/types/src/types/item_variant.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub async fn name(&self) -> &String {
+        &self.item_variant.name
+    }
+
+    pub async fn doses_per_unit(&self) -> &Option<f64> {
+        &self.item_variant.doses_per_unit
+    }
+
+    pub async fn manufacturer_id(&self) -> &Option<String> {
+        &self.item_variant.manufacturer_link_id // TODO join to name for manufacturer_id
```
jmbrunskill, line 32: Name loader?

jmbrunskill (approved):
Looks great to me, lets ship it!

client/packages/system/src/Item/DetailView/DetailView.tsx
```diff
@@ -44,11 +47,16 @@ export const ItemDetailView: FC = () => {
...
-  Environment.FEATURE_PACK_VARIANTS &&
+  isCentralServer &&
     tabs.push({
-      Component: <PackVariantsTab itemId={data.id} />,
-      value: t('label.pack-variants'),
-    });
+      Component: <ItemVariantsTab itemVariants={data.variants} />,
+      value: t('label.variants'),
+    }),
+    Environment.FEATURE_PACK_VARIANTS &&
```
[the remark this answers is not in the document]
lache-melvin: Ah yeah definitely!

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
lache-melvin: Yep, my thinking when implementing the `add` button properly was to have these translatable - hadn't thought about them being able to change the name though - I suppose as long as we display the packaging level then we could!

server/graphql/types/src/types/item.rs
lache-melvin: What do you mean? If we use the loader then these wouldn't get loaded if only selecting a subset of fields..

Unless you're saying we should add another query to get the variants directly. Which probably, though I think the current cases we have, we'd only be querying the variants in the context of an item 🤷‍♀️

server/graphql/types/src/types/item_variant.rs
jmbrunskill: I think my point was if we're generally loading the item_variant from a loader on item, looking up or even providing item_id on the node feels a bit redundant. I guess, we just wait to see what we actually need for the front end. Maybe we can even just get rid of this for now until we need it.

Definitely keen for that Loader write up. `link_id` one would be good too!

server/graphql/types/src/types/item.rs
jmbrunskill: I take it back this is already being done.
I see in Item API we have 

```
fragment ItemRow on ItemNode {
  __typename
  id
  code
  name
  unitName
}

fragment ItemWithPackSize on ItemNode {
  ...ItemRow
  defaultPackSize
}
fragment ItemStockOnHand on ItemNode {
  ...ItemWithPackSize
  availableStockOnHand(storeId: $storeId)
}
fragment ItemRowWithStats on ItemNode {
  ...ItemStockOnHand
  stats(storeId: $storeId) {
    __typename
    averageMonthlyConsumption
    availableStockOnHand
    availableMonthsOfStockOnHand
    totalConsumption
  }
}
```

So I think we already have what I was thinking of.
Just didn't want the loader running unless we actually need it in the context...

PR #5226, opened 2024-10-24
5139 delete response requisition lines

### LINKED ISSUE #5139: Delete response requisition lines
Implement deletion for requisition lines now that users can create manual requisitions. 

1. Service
2. API
3. UI hook up

fergie-nz:
Fixes #5139 

# 👩🏻‍💻 What does this PR do?

Adds delete response requisition lines & line back end service and graphql layers.

Adds front end api for batch delete requisition lines. 

Have added delete singular response requisition lines but doesn't expose this in API (similar to request requisition lines).
We have the singular delete, but all deletes are handled through the batch delete.

Adds tests, but graphql delete lines are a little limited.

# 🧪 Testing

- [ ] Go to distribution -> requisitions
- [ ] select some lines
- [ ] delete via drop down
- [ ] (You might need to create lines in the DB as no create lines method currently exists. Or you can reassign lines from another (request) requisition for data to delete.

# 📃 Documentation


client/packages/requisitions/src/ResponseRequisition/api/api.ts
```diff
@@ -191,22 +191,37 @@ export const getResponseQueries = (sdk: Sdk, storeId: string) => ({
...
+    if (result.batchResponseRequisition.deleteResponseRequisitionLines) {
+      const failedLines =
+        result.batchResponseRequisition.deleteResponseRequisitionLines.filter(
+          line =>
+            line.response.__typename === 'DeleteResponseRequisitionLineError'
+        );
+      if (failedLines.length === 0) {
+        return result.batchResponseRequisition.deleteResponseRequisitionLines;
+      }
+    }
```
fergie-nz, line 207: This uses the batch delete api even for singular.

Not currently used, but will be used for edit single line in future

server/graphql/requisition_line/src/lib.rs
```diff
@@ -53,4 +53,13 @@ impl RequisitionLineMutations {
...
     }
+
+    async fn delete_response_requisition_line(
+        &self,
+        ctx: &Context<'_>,
+        store_id: String,
+        input: response_requisition_line::DeleteInput,
+    ) -> Result<response_requisition_line::DeleteResponse> {
+        response_requisition_line::delete(ctx, &store_id, input)
+    }
```
fergie-nz, line 64: I've made this as per delete_request_requisition_line, but neither are currently exposed

server/service/src/requisition/request_requisition/delete.rs
```diff
@@ -94,13 +94,13 @@ fn validate(
...
-        return Err(OutError::CannotEditRequisition);
-    }
-
     if requisition_row.r#type != RequisitionType::Request {
         return Err(OutError::NotARequestRequisition);
     }
+
+    if requisition_row.status != RequisitionStatus::Draft {
+        return Err(OutError::CannotEditRequisition);
+    }
```
fergie-nz, line 103: I've rearranged error handling for mutating requisitions throughout. 

To me this makes more sense returning the wrong type of requisition error before returning a status error.

server/service/src/requisition/request_requisition/update/test.rs
```diff
@@ -4,7 +4,7 @@ mod test_update {
     use repository::{
         mock::{
             mock_draft_request_requisition_for_update_test,
-            mock_full_draft_response_requisition_for_update_test, mock_name_store_c,
+            mock_full_new_response_requisition_for_update_test, mock_name_store_c,
```
fergie-nz, line 7: We don't have draft status for response requisitions.

Have renamed

fergie-nz: This is also what drew my attention to the ordering of errors. When I changed the status to new from draft, tests would return errors of wrong status when the requisition was also the wrong type.

This is why I changed the order of error returning rather than restructure the mocks and tests.

I am thinking best order of error returning importance would be doesn't exist > wrong store > wrong type (request vs response) > wrong status.

This touches a few other places where I haven't changed the order, but I think it would be helpful to do so

server/graphql/requisition_line/src/lib.rs
roxy-dao: It's being used by the batch mutation. Ok not to be exposed

server/service/src/requisition_line/response_requisition_line/delete.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+
+pub enum DeleteResponseRequisitionLineError {
+    RequisitionLineDoesNotExist,
+    NotThisStoreRequisition,
+    NotAResponseRequisition,
+    RequisitionDoesNotExist,
+    CannotEditRequisition,
```
roxy-dao, line 24: Can you add a `CannotDeleteTransferredRequisition` er or some better naming... This needs to check that the requisition is not created from an Internal Order

server/service/src/requisition_line/response_requisition_line/delete.rs
```diff
@@ -0,0 +1,184 @@
...
+        return Err(OutError::NotThisStoreRequisition);
+    }
+
+    if requisition_row.r#type != RequisitionType::Response {
+        return Err(OutError::NotAResponseRequisition);
+    }
+
+    if check_approval_status(&requisition_row) {
+        return Err(OutError::CannotEditRequisition);
+    }
```
roxy-dao, line 69: This error is supposed to check for status since you cannot delete a requisition that was created from an Internal Order so approval status doesn't really matter.

```suggestion
    if requisition_row.status == RequisitionStatus::Finalised
        return Err(OutError::CannotEditRequisition);
    }
```

roxy-dao (changes requested):
Thanks Ferg (: Excellent job on the deletes~

I just have a few comments below that need to be addressed before the PR can be approved (:

roxy-dao:
Oh! And the postgres tests are failing, can you fix this (:

server/graphql/requisition_line/src/lib.rs
fergie-nz: of course!

client/packages/requisitions/src/ResponseRequisition/api/hooks/line/useDeleteResponseLines.ts
```diff
@@ -28,21 +27,25 @@ export const useDeleteResponseLines = () => {
...
+    mutateAsync(selectedRows).catch(err => {
+      throw err;
+    });
   };
 
-  return { onDelete };
+  const confirmAndDelete = useDeleteConfirmation({
+    selectedRows,
+    deleteAction: onDelete,
+    canDelete: !isDisabled,
```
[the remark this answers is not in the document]
fergie-nz: I'm wondering about making a new utils hook to just check if it has a linked requestRequisition

For the toolbar drop down, I disabled by checking if there was a theirReference. Here I am reluctant to make a unique call for the full requisition node just for this check.

fergie-nz:
> Thanks for the changes Ferg! I just have a few more comments and have put them in this [example branch](https://github.com/msupply-foundation/open-msupply/compare/5139-delete-response-requisition...5139-example) (:

Awesome thank you Roxy! I have added those.

roxy-dao (approved):
Awesome! Thanks for all the changes Ferg /o/

PR #5231, opened 2024-10-24
4367(2) Cold storage display

### LINKED ISSUE #4367: GAPS: Link stock data to specific storage temperature ranges
 
Part of Epic https://github.com/msupply-foundation/open-msupply/issues/4362

You can already assign multiple locations to a cold chain asset:

Image

So just need to add temperature_range to a location. This would need to be stored in OMS central, on the `location` table (much like the `name_oms_fields` extension of the `name` table)

Show the temperature range in brackets after the location name, when selecting the location for a stock line (i.e. in inbound shipment)

This can then be used in dashboard queries

### Additional context 💌

| User story  | A.6.2 |
As a… | Health Care Worker (HCW) or warehouse worker
I want… | Stock inventory data to be associated with the correct cold storage location
So that… | I can see exactly what stock each CCE is storing at any given time 

**Solution A.6.2**

The system will allow for users to make the association between stock storage locations and CCE by providing a mechanism to create new stock storage locations that represent the cold storage within CCE.

Multiple locations will be possible for each CCE and the user will be able to specify which type of cold storage they are associated with - e.g. +5 °C and -20 °C for hybrid refrigerator / freezers.

Note: the `exactly what stock each CCE is storing` would be nice but not required at this stage 🙏 

### Moneyworks Jobcode 🧰

`CCEIDEV`

CarlosNZ:
Fixes #4367

# 👩🏻‍💻 What does this PR do?

- Add GraphQL loaded for `coldStorageType` and make available in Location node
- Fetch and display storage temperature range in Cold Chain Equipment locations
- Display Cold storage info in "Locations" table

Cold Chain -> Equipment -> Details page
Screenshot 2024-10-25 at 12 27 13 PM
Inventory -> Locations -> List view
Screenshot 2024-10-25 at 12 26 58 PM

Haven't added editing functionality to the "Edit Locations" modal yet as wanted to get this out now. Is this in the scope of this issue? (Would just be a "Cold storage type" drop-down)

Haven't done sync translations yet, will do next.

Also, please not that I've changed the migration script to define the field as `DOUBLE` rather than `REAL` as it was getting a runtime error. So you'll have to drop the `cold_storage_type` table and re-run the migration if it's already been previously created.

# 🧪 Testing

- [ ] Temperature range shows on the "Locations" chips in Equipment Detail page
- [ ] Temperature range shows on "Locations" dropdown in Equipment Details page
- [ ] Cold storage type info shows on "Locations" inventory table


client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
```diff
@@ -98,7 +99,7 @@ export const EquipmentDetailView = () => {
 
   const locations =
     locationData?.nodes.map(location => ({
-      label: location.code,
+      label: formatLocationLabel(location as LocationNode),
```
CarlosNZ, line 102: The exact type returned from the hook is not named, and differs slightly from this `location` object. However, it's essentially the same (for this purpose anyway), hence as `as`. If there's a better way that doesn't involve manually defining a type....?

client/packages/coldchain/src/Sensor/api/operations.graphql
```diff
@@ -1,4 +1,4 @@
-fragment LocationRow on LocationNode {
+fragment SensorLocationRow on LocationNode {
```
CarlosNZ, line 1: This is just renamed, as the codegen complained about having two different types called `LocationRow` (it was fine when they were exactly the same, but now one of them is fetching additional info)

server/repository/src/migrations/v2_04_00/add_cold_storage_type_table.rs
```diff
@@ -14,8 +14,8 @@ impl MigrationFragment for Migrate {
                 CREATE TABLE cold_storage_type (
                     id TEXT NOT NULL PRIMARY KEY,
                     name TEXT NOT NULL,
-                    min_temperature REAL,
-                    max_temperature REAL
+                    min_temperature {DOUBLE},
+                    max_temperature {DOUBLE}
```
CarlosNZ, line 18: Gave a runtime error when type was `REAL`, so changed to 'double precision' in keeping with existing values of similar type.

adamdewey:
Can we rename `Cold storage type` to `Storage type` in Locations to make it less restrictive?

client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
lache-melvin: I'd just change the `formatLocationLabel` function below to accept `LocationRowFragment` (the type of `location` here)!

lache-melvin (changes requested):
Looks awesome, thanks Carl! Just to finish up:

- In location list: change column from `Cold Storage Type` to `Storage type`
- Per product chat, change the location picker to show the `name` instead of `min - max`
- `LocationSearchInput` should also include the storage type (for inbound shipment/view stock detail etc)


Yep would need to add dropdown to add/edit location, but agree can be next PR 🙏

CarlosNZ:
> Looks awesome, thanks Carl! Just to finish up:
> 
>     * In location list: change column from `Cold Storage Type` to `Storage type`
> 
>     * Per product chat, change the location picker to show the `name` instead of `min - max`
> 
>     * `LocationSearchInput` should also include the storage type (for inbound shipment/view stock detail etc)
> 
> 
> Yep would need to add dropdown to add/edit location, but agree can be next PR 🙏

Changes done: 9a03a8e60c

lache-melvin (approved):
Looks awesome, thanks for changes Carl 🔥

PR #5227, opened 2024-10-24
4789 item variant UI pt2

lache-melvin:
Fixes #4789

# 👩🏻‍💻 What does this PR do?

Moves add/edit of item variants to modal:
Screenshot 2024-10-25 at 10 30 27 AM

## 💌 Any notes for the reviewer?

# 🧪 Testing

- [ ] No backend wired up, but check add/edit open the modal as expected, and send the right payload to the backend

# 📃 Documentation

- [x] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

PR #5232, opened 2024-10-25
5208 gaps item variant service layer

### LINKED ISSUE #5208: GAPS: item_variant server & graphql framework

Able to query item_variant, and packaging_variant from graphql (doesn't have to be final API)
Insert/update won't be in this PR...

jmbrunskill:
Fixes #5208

# 👩🏻‍💻 What does this PR do?

Service layer with upsert and delete for, `item_variant` and `packaging_variant`
Will do a new PR for graphql based on https://github.com/msupply-foundation/open-msupply/pull/5227

Haven't done much if any validation.

Will need to add this to next PR. Added some todos in the test cases for validation. Interested in feedback on what validation is needed?

# 🧪 Testing

- [ ]  No QA needed, it will be tested as part of the UI stuff in the next PR
- [ ] Check if code test cases make sense, and what needs to be added...

# 📃 Documentation


server/service/src/item/item_variant/test/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    #[actix_rt::test]
+    async fn validate_item_variant() {
+        // TODO validation tests
+
+        // Test that the item variant name is set?
+
+        // Test that we can't create a record with an item_id that doesn't exist
+
+        // Test that we can't change the item_id on an existing record???
```
lache-melvin, line 148: Hm, this one is a good point! I guess fine to determine if we are doing update or insert in the validation and apply some different checks accordingly! Same with the unique name check

server/service/src/item/packaging_variant/upsert.rs
```diff
@@ -0,0 +1,81 @@
...
+        volume_per_unit,
+        deleted_datetime: None,
+    }
+}
+
+fn validate(
+    connection: &StorageConnection,
+    input: &UpsertPackagingVariant,
+) -> Result<(), UpsertPackagingVariantError> {
+    let item_variant = check_item_variant_exists(connection, &input.item_variant_id)?;
```
lache-melvin, line 75: Wondering if we can actually get away without this step - given graphql layer, you can only insert packaging variant through item variant 👀

lache-melvin (approved):
Awesome thanks James! Cool to see you removing the _dsl stuff too 🔥

server/service/src/item/packaging_variant/upsert.rs
jmbrunskill: Yeah I guess it depends if we want to keep this option available in the service layer?

Doesn't hurt I think...

PR #5236, opened 2024-10-25
5228 add item variant

### LINKED ISSUE #5228: Add item_variant_id to stock_line/invoice_line

Need to capture item_variant_id on all incoming stock for GAPS (optional). Should be added to both stock_line and invoice_line.

Will need to go be added in mSupply, the translated and synced.

lache-melvin:
Fixes #5228

# 👩🏻‍💻 What does this PR do?

Adds item variant ID to stock line and invoice line

# 🧪 Testing

- [ ] Pull latest `master` of mSupply
- [ ] Check sync versions work correctly
- [ ] Everything's compiling as it should
- [ ] Actual saving of these fields will be next PR!

# 📃 Documentation

  1.
  2.


server/service/src/invoice_line/stock_in_line/update/generate.rs
```diff
@@ -140,6 +141,7 @@ fn generate_line(
         currency_id,
         currency_rate,
     )?;
+    update_line.item_variant_id = item_variant_id.or(update_line.item_variant_id);
```
jmbrunskill, line 144: This pattern would mean we can't update it to null once it's been set right?
Not a big deal for this PR but feels like something we'd want to be able to do right?

jmbrunskill (approved):
Looks good to me.

Does stocktakeLine also need this field? So we can set item variant via a stocktake, or do you think we can work around it?

server/service/src/invoice_line/stock_in_line/update/generate.rs
lache-melvin: That's a great point, it probably should be nullable update - can fix with the UI PR!

lache-melvin:
> Does stocktakeLine also need this field? So we can set item variant via a stocktake, or do you think we can work around it?

Yeah I wondered about that - I thought it might be okay for now - usually you're doing a stocktake on existing stock, which should already have the variant set, and if you add new batches, can always amend the variant from the stock detail page? Agree it would be better to have it in, but was thinking it's probably okay until such a time as we need item variants to be more integrated in the app? 👀 Can check with Mark though...

mark-prins:
I agree : better to have it, but a it's a low priority for the current usage. It would add complexity to the stocktake which wouldn't be helpful for most use cases. would the variant selection show if there aren't any defined though? if that's the case then it's low impact to put it in now. but down the priority list as we don't actually need it for the current feature

Issue #3987, opened 2024-05-22
Returns modal horizontal scroll


mark-prins:
## What went wrong? 😲
I took this screen cap while testing something the other day. The distribution return (Inbound return) looks ok when the item name is short, but has a horizontal scroll and doesn't look great if the item name is a little longer:

<img width="1067" alt="Screenshot 2024-05-20 at 4 29 55 PM" src="https://github.com/msupply-foundation/open-msupply/assets/9192912/d71cd822-8729-404d-991a-09e07b03e2cd">

The replenishment return (Outbound) also has an issue with horizontal scolling:

![image](https://github.com/msupply-foundation/open-msupply/assets/9192912/43917834-cc3c-4a6a-b790-aa49322ecf67)

I tried to click next without realising that I had to enter a quantity, because the input is hidden off to the right somewhere.
The input field can be on the left, though the other numeric fields need to be nearby too, as they influence the amount to return.
The modal could be wider on desktop, but probably not on mobile. The unit column is wider than strictly necessary, and perhaps the item name column needs a limit on it?

## Expected behaviour 🤔
No horizontal scrolling in the Inbound return modal table
The input field should be visible in the Outbound return modal

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

andreievg:
Refinement, quick note, bigger modal, inputs on left side

mariyamsupply:
For these examples I see that the item Name and Code repeat the ones at the top (in the dropdown).
Can we just drop them from the table? So there is should be enough space for the other columns.

mariyamsupply:
Alternatively, I think we can replace all 'complex' modals with the LineViewEdit page and keep things in accordions

<img width="1024" alt="Image" src="https://github.com/user-attachments/assets/d768472f-5a69-4958-8970-a7300aee7cab" />

<img width="1024" alt="Image" src="https://github.com/user-attachments/assets/3f18cfaf-7ad1-43b5-9e2c-0e99a1aa55cf" />

andreievg:
Closing, [added to this issue](https://github.com/msupply-foundation/open-msupply/issues/8608), please re-open if scheduled

PR #5255, opened 2024-10-30
5145 item variant on all incoming stock

### LINKED ISSUE #5145: Capture item variant on all incoming stock

- Add extra column with a dropdown input (before batch) anywhere stock comes in/changes:
  - Inbound Shipment
  - Customer Return
  - Stocktake
  - View stock detail
  - Inventory adjustment
  - Repack
- Add item_variant_id to stock_line, and probably invoice_line (mSupply change!) & sync

Screenshot 2024-10-16 at 2 12 36 PM

### Additional context 💌

At this point, we're just capturing the variant on the stock line to calculate current storage capacity used. i.e. we don't need to change anything with regards to distribution based on # packs per variant at this stage

### Moneyworks Jobcode 🧰

CCEIDEV

lache-melvin:
Fixes #5145

# 👩🏻‍💻 What does this PR do?

Capture item variant on incoming stock:

Screenshot 2024-10-30 at 1 11 19 PM

Also changes the `updates` to accept nullable input (to be able to unset the variant)

Put the variant column fairly early in the input tables (after batch) per the designs, but do you think it should be more to the right/deprioritised column at this point?

Also, should I add a column to the view stock list view?

# 🧪 Testing

- [ ] Have some item variants configured for an item

For each of the below, try add and remove item variants:

- [ ] Create & edit customer return
- [ ] Create inbound shipment
- [ ] Create new stock line from View Stock
- [ ] Edit stock line from view stock

# 📃 Documentation

  1.
  2.


jmbrunskill:
Looks like it might have been a pre-existing problem but seems bad that quantity to return is off the screen :(
image

client/packages/system/src/Item/api/operations.graphql
```diff
@@ -266,6 +266,29 @@ query itemById($storeId: String!, $itemId: String!) {
   }
 }
 
+fragment ItemVariantOption on ItemVariantNode {
```
jmbrunskill, line 269: Nice, feels like a potential pattern to re-use?

server/service/src/invoice_line/stock_in_line/insert/validate.rs
```diff
@@ -31,6 +31,12 @@ pub fn validate(
         return Err(LocationDoesNotExist);
     }
 
+    if let Some(item_variant_id) = &input.item_variant_id {
+        if !check_item_variant_exists(connection, item_variant_id)? {
```
jmbrunskill, line 35: I guess to be thorough we should check that it not only exists but that it's for the right `item_id`?
Not overly worried though...

server/service/src/invoice_line/stock_in_line/update/validate.rs
```diff
@@ -51,6 +52,17 @@ pub fn validate(
     if !check_location_exists(connection, store_id, &input.location)? {
         return Err(LocationDoesNotExist);
     }
+    // TODO
```
jmbrunskill, line 55: Is this TODO correct?

server/service/src/invoice_line/stock_in_line/update/validate.rs
```diff
@@ -51,6 +52,17 @@ pub fn validate(
...
+    match &input.item_variant_id {
+        Some(NullableUpdate {
+            value: Some(item_variant_id),
+        }) => {
+            if !check_item_variant_exists(connection, item_variant_id)? {
+                return Err(ItemVariantDoesNotExist);
+            }
+        }
+        _ => {}
+    }
```
jmbrunskill, line 65:
```suggestion
    match &input.item_variant_id {
        Some(NullableUpdate {
            value: Some(item_variant_id),
        }) => {
            if !check_item_variant_exists(connection, item_variant_id)? {
                return Err(ItemVariantDoesNotExist);
            }
        }
        _ => {} //  We don't need to check item_variant if it's not being update, or if it's being updated to None
    }
```

I feel like this is sufficiently complicated to justify a comment? Took me a bit to parse it, possibly dad brain slowness though haha.

lache-melvin:
> Looks like it might have been a pre-existing problem but seems bad that quantity to return is off the screen :( 

Yup :/ #3987

> Also related, and I guess highly debated, but it feels wrong that the customer return under distribution, is actually adding to your stock, vs other transactions under this heading which are reducing it.
> 
> I was confused as I was looking under replenishment...

And yep, this has gone around many a time on Telegram, grouping by "Incoming stock"/"Outgoing stock" vs "Customer related"/"Supplier related" 🤷‍♀️

jmbrunskill (approved):
Nice! Working well, merge away when ready!

One thought I had was that it might be nice to hide the item variant inputs if there's no variants configured. Basically so as to not change the UI  and save screen realestate for people who aren't using them, which might be a bulk of our users? A global pref might be nice?

Worthing discussing or adding to backlog @mark-prins ?

Personally I like the idea of `item_variant` showing up in the `ViewStock` list view, but also happy for that to come along in future if asked for :)

client/packages/system/src/Item/api/operations.graphql
lache-melvin: I wondered about it! In cases where we are doing a query to populate the options, this is nicer than having to map in the component - but I think many queries are reused from another list... arguably that's not great though!

server/service/src/invoice_line/stock_in_line/update/validate.rs
lache-melvin: Jaja it was at one point... 🙈

lache-melvin:
> One thought I had was that it might be nice to hide the item variant inputs if there's no variants configured. Basically so as to not change the UI and save screen realestate for people who aren't using them, which might be a bulk of our users? A global pref might be nice?
> 
> Worthing discussing or adding to backlog @mark-prins ?

Yeah we had this for pack variants, I think it's nice to hide for majority of users, especially while they don't do much! It is more work though... let's see what Mark thinks 🙏 

> Personally I like the idea of `item_variant` showing up in the `ViewStock` list view, but also happy for that to come along in future if asked for :)

Was going to do if it was a 2 min job, but I think needs a new item_variant_by_id loader etc., think we can do later as needed 😁

PR #5275, opened 2024-10-31
5259 gaps sync translation for location type cold storage type

### LINKED ISSUE #5259: GAPS: Sync translation for location_type -> Cold Storage Type

When translating LocationType from mSupply Legacy, we should create cold_storage types (if the `min` or `max` is defined for the location type)

jmbrunskill:
Fixes #5259

# 👩🏻‍💻 What does this PR do?

Translates `location_type` from mSupply into `cold_storage_type` in OMS

Named the translator after the Legacy record, I think that's probably what we'd want but wasn't sure.

# 🧪 Testing

- [ ] Create/Edit Location Types
- [ ] Check that they're translated into `cold_storage_types`
- [ ] Check that item variants translate correctly when these are added at the same time as new cold storage types

# 📃 Documentation

  1.
  2.


server/service/src/sync/translations/item_variant.rs
```diff
@@ -21,7 +24,11 @@ impl SyncTranslation for ItemVariantTranslation {
     }
 
     fn pull_dependencies(&self) -> Vec<&str> {
-        vec![ItemTranslation.table_name()]
+        vec![
+            ItemTranslation.table_name(),
+            NameTranslation.table_name(),
```
jmbrunskill, line 29: This one should be in my other PR...

server/service/src/sync/translations/location_type.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub temperature_max: f64,
+}
+// Needs to be added to all_translators()
+#[deny(dead_code)]
+pub(crate) fn boxed() -> Box<dyn SyncTranslation> {
+    Box::new(LocationTypeTranslation)
+}
+
+/// Translates between the legacy LocationTypeRow and the new ColdStorageTypeRow
+pub(super) struct LocationTypeTranslation;
```
lache-melvin, line 24: Hm, comparing to trans_line/invoice_line: https://github.com/msupply-foundation/open-msupply/blob/5259-gaps-sync-translation-for-location_type-cold-storage-type/server/service/src/sync/translations/invoice_line.rs#L93

I think the translator, and the file names should be the OMS name (cold_storage_type), only references to legacy row should be location_type?

lache-melvin (approved):
Looks great to me James, just requesting the naming change to be consistent with other tables we've renamed from OG, merge when ready 🙏

server/service/src/sync/translations/location_type.rs
jmbrunskill: Updated!

lache-melvin (approved):
changes looks good 😊

PR #5276, opened 2024-10-31
#5247 Naive date fix

### LINKED ISSUE #5247: Date filters (general) + patient birth date on creation displays the incorrect date (date before)
## What went wrong? 😲
Reported by @alainsussol & @kahn-msupply 

When they filter by date (in stock and patient) the date selection jumps to the date before and the same thing happens when entering a patient's birth date on creation.

Possible timezone problem.

https://github.com/user-attachments/assets/5155487a-3949-4a9e-a2a2-41c429baa798

https://github.com/user-attachments/assets/2db382bd-ca6c-46b4-91a8-3324e1933c72

photo_2024-10-30 09 06 15

## How to Reproduce 🔨

1. Go to a date filter (stock expiry date or patient DOB)
2. Click on a date, see that it changes to the date before

## Your environment 🌱

- Open mSupply Version: 2.3.1

- Platform: Android & Desktop

- Database type:

CarlosNZ:
Fixes #5247

# 👩🏻‍💻 What does this PR do?

Note: in order to see the problem (and the solution), you'll probably need to set your computer's timezone to something that is still in "yesterday".

Okay, so the problem is that the values in question are "naive" dates (i.e. timezone is irrelevant), but the date filters and Patient summary were using the `getDateOrNull` method, which assumes a time stamp of "00:00:00" if not is specified, so dates get converted to the day before if local timezone is still in "yesterday" relative to UTC.

There was a method to return a "naive" version of the date, which I've used, but made a couple of small changes to it:

- changed the name from `getLocalTime` to `getNaiveDate` (it's kind of the opposite of "get local time", so I found the current name confusing/misleading)
- moved the method out of the `useFormatDateTime` hook into the main `DateUtils` object. This is so it can be used in functions that are not inside hooks. (There's actually several functions in that hook that don't need to be there -- they only need to be inside a hook if they use another hook, e.g. `useTranslation`, but that's a problem for another day)

See inline comments in PR

# 🧪 Testing

- On original branch, change your computer timezone to something that reveals the problem as shown in the issue
- Switch to this branch
- Problem should be fixed:
  - Stock filters
  - Patient Summary
- Also, the Date pickers were auto-switching the displayed value to the day before after entry due to the same problem -- this should be fixed now too (See Stock expiry date modal and Patient DOB entry)

### DIFF
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -104,6 +104,35 @@ export const DateUtils = {
         : new Date(date);
     return isValid(maybeDate) ? maybeDate : null;
   },
+  /**
+   * While getDateOrNull is naive to the timezone, the timezone will still
+   * change. When converting from the assumed naive zone of GMT to the local
+   * timezone, the dateTime will be wrong if the timezone is behind GMT.
+   * For example: for a user in -10 timezone, a date of 24-02-2024 will become
+   * 2024-02-23T13:00:00.000Z when rendered for mui datepicker.
+   * This function acts in the same way as getDateOrNull, but will create a
+   * datetime of start of day local time rather than start of day GMT by
+   * subtracting the local timezone offset.
+   * You can use this function anytime you need a datetime for mui date picker
+   * to be created from a date only string. This includes date of birth, date of
+   * death or any other date which is time and timezone agnostic.
+   */
+  getNaiveDate: (
+    date?: Date | string | null,
+    format?: string,
+    options?: ParseOptions,
+    timeZone?: string
+  ): Date | null => {
+    // tz passed as props options for testing purposes
+    const tz = timeZone ?? new Intl.DateTimeFormat().resolvedOptions().timeZone;
+    const UTCDateWithoutTime = DateUtils.getDateOrNull(date, format, options);
+    const offset = UTCDateWithoutTime
+      ? getTimezoneOffset(tz, UTCDateWithoutTime)
+      : 0;
+    return UTCDateWithoutTime
+      ? addMilliseconds(UTCDateWithoutTime, -offset)
+      : null;
+  },
   minDate: (...dates: (Date | null)[]) => {
     const maybeDate = fromUnixTime(
       Math.min(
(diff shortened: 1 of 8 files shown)


client/packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -104,6 +104,35 @@ export const DateUtils = {
...
+   * For example: for a user in -10 timezone, a date of 24-02-2024 will become
+   * 2024-02-23T13:00:00.000Z when rendered for mui datepicker.
+   * This function acts in the same way as getDateOrNull, but will create a
+   * datetime of start of day local time rather than start of day GMT by
+   * subtracting the local timezone offset.
+   * You can use this function anytime you need a datetime for mui date picker
+   * to be created from a date only string. This includes date of birth, date of
+   * death or any other date which is time and timezone agnostic.
+   */
+  getNaiveDate: (
```
CarlosNZ, line 120: Moved and renamed, no change to functionality.

There's probably a simpler way to do this, but this is basically fine for now.

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
```diff
@@ -72,7 +72,7 @@ export const DateFilter: FC<{ filterDefinition: DateFilterDefinition }> = ({
 
 const getDateFromUrl = (query: string, range: RangeOption | undefined) => {
   const value = typeof query !== 'object' || !range ? query : query[range];
-  return DateUtils.getDateOrNull(value);
+  return DateUtils.getNaiveDate(value);
```
CarlosNZ, line 75: Date filter now uses Naive date, which means it will match the string in the URL as well as the displayed "value". 

I don't think this would cause any problems anywhere would it?

client/packages/programs/src/JsonForms/common/components/Date.tsx
```diff
@@ -52,7 +53,7 @@ const UIComponent = (props: ControlProps) => {
         <BaseDatePickerInput
           sx={{ width: '100%' }}
           // undefined is displayed as "now" and null as unset
-          value={formatDateTime.getLocalDate(data)}
+          value={DateUtils.getNaiveDate(data)}
```
CarlosNZ, line 56: Rename only

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -86,7 +86,7 @@ const UIComponent = (props: ControlProps) => {
         <Box display="flex" alignItems="center" gap={FORM_GAP} width="100%">
           <BaseDatePickerInput
             // undefined is displayed as "now" and null as unset
-            value={formatDateTime.getLocalDate(dob)}
+            value={DateUtils.getNaiveDate(dob)}
```
CarlosNZ, line 89: Rename only

lache-melvin (approved):
Looks great to me Carl, thanks for this 🚀

PR #5286, opened 2024-11-01
Improve confirm on leaving modal

### LINKED ISSUE #5285: Improve confirm on leaving alert
Following on from #5274 - this is an opportunity to update the modal style.
When prompting we are using the browser `confirm` method, giving:

image

It would be nicer to use the confirmation modal and prompt using a react element rather than the browser default.

mark-prins:
Fixes #5285

# 👩🏻‍💻 What does this PR do?
Uses the confirmation modal rather than the `confirm` method, giving:

Patients, when changing tabs:
image

R&R form, when navigating away from page:
image

Have used the confirmation modal.

It looks a little nicer

I think this should be targeted for develop, there's no urgency to get it into the patch release. Have just targeted the patch branch to make the PR cleaner to read, ahead of it being merged back into develop.

# 🧪 Testing

- [ ] Test the R&R form, making a change and navigating away or refreshing the page
- [ ] OK and Cancel should work as expected

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
+++ b/client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
@@ -1,7 +1,7 @@
 import { useCallback, useContext, useEffect } from 'react';
-import { BlockerFunction, useBeforeUnload, useBlocker } from 'react-router-dom';
+import { useBeforeUnload, useBlocker } from 'react-router-dom';
 import { useTranslation } from '@common/intl';
-import { ConfirmationModalContext } from '@openmsupply-client/common'; // '@common/components';
+import { ConfirmationModalContext } from '@openmsupply-client/common';
 
 /** useConfirmOnLeaving is a hook that will prompt the user if they try to navigate away from,
  *  or refresh the page, when there are unsaved changes.
@@ -23,18 +23,10 @@ export const useConfirmOnLeaving = (isUnsaved?: boolean) => {
     setOpen(true);
   }, [setMessage, setTitle, setOpen]);
 
-  const shouldBlock = useCallback<BlockerFunction>(
-    ({ currentLocation, nextLocation }) => {
-      if (!!isUnsaved && currentLocation.pathname !== nextLocation.pathname) {
-        showConfirmation();
-        return true;
-      }
-      return false;
-    },
-    [isUnsaved, showConfirmation]
+  const blocker = useBlocker(
+    ({ currentLocation, nextLocation }) =>
+      !!isUnsaved && currentLocation.pathname !== nextLocation.pathname
   );
-
-  const blocker = useBlocker(shouldBlock);
 
   // handle page refresh events
   useBeforeUnload(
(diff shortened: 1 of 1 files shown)


client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
```diff
@@ -1,32 +1,42 @@
...
-        return false;
-      },
-      [isUnsaved]
-    )
+  const { setOpen, setMessage, setOnConfirm, setTitle } = useContext(
+    ConfirmationModalContext
   );
 
+  const showConfirmation = useCallback(() => {
+    setMessage(t('heading.are-you-sure'));
```
mark-prins, line 21: manually doing this rather than using the `useConfirmationModal` hook as I had issues with that call, and having a captured enclosure around the blocker.

client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
```diff
@@ -1,32 +1,42 @@
...
+    setTitle(t('messages.confirm-cancel-generic'));
+    setOpen(true);
+  }, [setMessage, setTitle, setOpen]);
+
+  const shouldBlock = useCallback<BlockerFunction>(
```
mark-prins, line 26: no particular reason to split this out, it was simply the result of testing some changes. happy to revert

client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
```diff
@@ -38,5 +48,19 @@ export const useConfirmOnLeaving = (isUnsaved?: boolean) => {
...
 
+  // Reset the blocker if the dirty state changes
+  useEffect(() => {
+    if (blocker.state === 'blocked' && !isUnsaved) {
+      blocker.reset();
+    }
+  }, [blocker, isUnsaved]);
+
+  // update the onConfirm function when the blocker changes
+  useEffect(() => {
```
mark-prins, line 59: without this I had trouble with an enclosure

CarlosNZ (commented):
Looks good, but I'm not sure about the two `useEffects`. I removed them and I couldn't find any difference in functionality. Can you show me a case that they're required?

I'm not fully following the logic sequence, but it seems like we're using `useBlocker` in a different way to the [official example](https://reactrouter.com/en/main/hooks/use-blocker) due to a) it being inside a hook rather than a component, and b) defining the `onOk` method inside a context called from this hook rather than with an event handler inside the component. 

Anyway, not sure what I'm suggesting. But I was able to make a much simpler version of this overall hook (with less `useCalbacks` too) without any apparent loss of functionality. So maybe I'm missing some use cases?

mark-prins:
Agree that using effects isn't ideal!
If you remove them how are you calling `blocker.proceed()`?

The example requires rendering a react element, which the user then interacts with. I was trying to keep with the existing declarative hook implementation so that I didn't have to update the callers.

I've simplified the code though and taken out the reset effect.

CarlosNZ:
> If you remove them how are you calling `blocker.proceed()`?

I don't know, but it still works 🤷‍♂️

Even with the latest changes, I can remove the useEffect and it still behaves correctly (at least when trying it with Patients).

I put a console log inside the `useEffect` and monitored the `blocker.state` -- it never changes from "unblocked" (so `blocker.proceed()` is not being called), so I suspect the useEffect is getting a stale version of `blocker` and a new instance is being created on every render.

This is when testing with the "Patients" details. Have you seen a place where we're sure that `blocker.proceed()` is definitely being called?

mark-prins:
> Even with the latest changes, I can remove the useEffect and it still behaves correctly (at least when trying it with Patients).

patients is a different process - you'll need to test with the R&R form

CarlosNZ (approved):
> patients is a different process - you'll need to test with the R&R form too

Okay, looks good. Thanks @mark-prins 👍

PR #5287, opened 2024-11-01
#4367(3) Edit cold storage on Location modal

### LINKED ISSUE #4367: GAPS: Link stock data to specific storage temperature ranges

CarlosNZ:
Fixes #4367 (hopefully the last thing on this issue?)

# 👩🏻‍💻 What does this PR do?

- Add "ColdStorageType" selector to "Edit Location" modal
- Add "cold_storage_type_id" field to GraphQL update mutation
- Add to front-end mutations
- Modify "ColdStorageType" component to allow "null" input (since you should be able to remove this once you've added it)

The UI is a problem here:

Screenshot 2024-11-01 at 1 14 23 PM

Screenshot 2024-11-01 at 1 14 40 PM

Note how the "Temperature" label doesn't show up until something is selected. This is something I've dealt with before, and it was a bit of headache to get it working for the Filter drop-downs. Getting the Drop-down here to display a label the same as the basic text inputs will take quite a bit of CSS black magic. However, I don't think it's worth the time as this modal is inconsistent with everywhere else as far as I can tell -- the normal layout is to use the "InputWithLabelRow" component, which looks like this:

Screenshot 2024-11-01 at 1 17 38 PM

So it's probably better to change this modal to be consistent with all the others, right? I'll make a new issue for that once we're clear on that.

# 🧪 Testing

- [ ] Go to Inventory -> Locations
- [ ] Click a location that has no "Storage type"
- [ ] Can add a storage type in the Edit Location modal that pops up
- [ ] Can change a storage type on a location that already has one
- [ ] Can add a storage type when creating a new location ("New Location" button at the top)
- [ ] Can remove storage type by clearing the input field and saving


client/packages/system/src/Item/Components/ColdStorageTypeInput/ColdStorageTypeInput.tsx
```diff
@@ -28,18 +30,18 @@ export const ColdStorageTypeInput = ({
   return (
     <Autocomplete
       disabled={disabled}
-      clearable={false}
       value={buffer && { ...buffer, label: buffer.name }}
       loading={isLoading}
       onChange={(_, name) => {
         setBuffer(name);
-        name && onChange(name);
+        onChange(name);
```
CarlosNZ, line 37: Allow `onChange` to accept `null` input

client/packages/system/src/Item/Components/ColdStorageTypeInput/ColdStorageTypeInput.tsx
```diff
@@ -28,18 +30,18 @@ export const ColdStorageTypeInput = ({
...
       }}
       options={data?.coldStorageTypes.nodes ?? []}
       getOptionLabel={option => option.name}
       width={`${width}px`}
       popperMinWidth={width}
       isOptionEqualToValue={(option, value) => option?.id === value?.id}
+      inputProps={{ label }}
```
CarlosNZ, line 44: Will probably remove this when we update the modal UI as I outlined in PR description

client/packages/system/src/Location/ListView/LocationEditModal/LocationEditModal.tsx
```diff
@@ -139,6 +141,15 @@ export const LocationEditModal: FC<LocationEditModalProps> = ({
             label={t('label.code')}
             InputLabelProps={{ shrink: true }}
           />
+          <ColdStorageTypeInput
+            value={draft.coldStorageType ?? null}
+            label={t('label.temperature')}
+            onChange={coldStorageType =>
+              onUpdate({
+                coldStorageType: coldStorageType as ColdStorageTypeNode,
```
CarlosNZ, line 149: I always seem to have trouble with this. The incoming type is `ColdStorageTypeFragment`, which is basically the same as `ColdStorageTypeNode` except is has a subset of fields. It's fine in this case, as only the `id` actually gets passed through to the mutation, but would like to do it without needing `as` if possible?

adamdewey:
> Note how the "Temperature" label doesn't show up until something is selected. This is something I've dealt with before, and it was a bit of headache to get it working for the Filter drop-downs

Alternative could be to have 'Room temperature' as the default value for new Locations?

jmbrunskill:
>So it's probably better to change this modal to be consistent with all the others, right? I'll make a new issue for that once we're clear on that.

I agree! Create a new issue

jmbrunskill:
> > Note how the "Temperature" label doesn't show up until something is selected. This is something I've dealt with before, and it was a bit of headache to get it working for the Filter drop-downs
> 
> Alternative could be to have 'Room temperature' as the default value for new Locations?

We've gone with the solution to make these configurable, so we can't easily create a Default Temperature one sadly. If we'd gone with a standardised Enum that would probably have worked....

client/packages/system/src/Location/ListView/LocationEditModal/LocationEditModal.tsx
jmbrunskill: I think the trick is to use the same fragment in both graphql queries, so that types match.
Here's my suggestion branch: https://github.com/msupply-foundation/open-msupply/compare/4367(3)-edit-storage-on-location-modal...4367-suggestion-remove-as

However, I don't like importing the fragment from a different graphql file, seems to make the `../../../../` issue when doing `yarn generate` even worse.
Would be better to move the `location_types` query from `item` to `location` I think?

I also had to bring through the min and max temperature and do other updates to avoid other issues where those were it being used.

So possibly if you just did added min_max to the `ColdStorageTypeFragment` it might compile? I think the types should then be equivalent? Haven't tried though.

The idea of using a more narrow fragment feels good in that you only bring through the data you actually need, but does cause other headaches, so probably only worth trying to be frugal if we've got a lot of loaders or something that genuinely slow things down...

I'm sure there's other ways we could deal with this too.

jmbrunskill (approved):
Looks good Carl. I agree would be nice to fix that `as` issue, possibly my suggestion branch, but would be good to take it a bit further to move the fragment perhaps?

client/packages/system/src/Location/ListView/LocationEditModal/LocationEditModal.tsx
CarlosNZ: > So possibly if you just did added min_max to the `ColdStorageTypeFragment` it might compile? I think the types should then be equivalent? Haven't tried though.

Yeah that did the trick, thanks :) 

6f69e19b68

CarlosNZ:
> > So it's probably better to change this modal to be consistent with all the others, right? I'll make a new issue for that once we're clear on that.
> 
> I agree! Create a new issue

Done https://github.com/msupply-foundation/open-msupply/issues/5290

Issue #3984, opened 2024-05-22
QueryErrorHandler doesn't report repeated errors


lache-melvin:
## What went wrong? 😲

Error snack does not display again after first error:

https://github.com/msupply-foundation/open-msupply/assets/55115239/a0df8303-bbe8-4465-bc5c-b42939c6b187

This appears to be because the `QueryErrorHandler` only calls the `error` snack if its `errorMessage` reference changes. (So e.g. first error was `ReturnIsNotEditable`, second error is `NetworkError`, that would display fine.) 

The `errorMessage` state is reset when we change pages, but otherwise, if we encounter the same error message twice in a row, we will only get the snack the first time!

I think this is particularly bad because the snack disappears - maybe I didn't finish reading it in time, I want to click OK again to finish reading the error message. OR I think I fixed my entered data (but actually I hadn't fixed it, BE returning same error) so I click OK again, and now not getting any error message, but the modal won't close. 

## Expected behaviour 🤔

If the same problem happens twice, I am told about it twice....

## How to Reproduce 🔨

Steps to reproduce the behaviour:

I think stopping the server is the easiest/fastest - but have seen with many places - with either standard errors, or where we throw errors from the api.ts files...

1. Create something e.g. an inbound return with at least one line
2. Stop your server
3. Try to edit the line
5. First time, get `NetworkError`
6. Click OK again
7. BUG: no error snack

1. Create an inbound return with at least one line
2. Open the edit modal for that line
3. Now, in another window (incognito?) set that inbound return to verified
4. Back in the first window, now try to click OK in your edit modal
5. First time, get `ReturnIsNotEditable`
6. Click it again
7. BUG: no error snack

## Your environment 🌱

- Open mSupply Version: 2.0.0
- Legacy mSupply Central Server Version: 7.14

- Platform: browser

- Database type: pg

mark-prins:
That was deliberate - we were getting duplicate errors displayed on a single page and it looks a litle incompetent, or at the least hamfisted!

If fixing, we need to be able to distinguish between multiple calls to the same action / api resulting in the same error and the user initiating something that results in the same error. possibly useQuery is initiating an action multiple times and reporting the same error multiple times?

lache-melvin:
> If fixing, we need to be able to distinguish between multiple calls to the same action / api resulting in the same error and the user initiating something that results in the same error. 

Hm, or a way to reset error state once the toast disappears?

[...]

Issue #5241, opened 2024-10-28
GAPS link tables not handled properly....


jmbrunskill:
## What went wrong? 😲

Currently we aren't joining to`name_link` and `item_link` tables properly in GAPS work.
This affects `item_variant` which is planning to be used outside of GAPS in the future...

This feels low priority for GAPS given we are planning to use it on new installs, but is not ideal, and technical debt we need to revisit.

## Expected behaviour 🤔

TBD: I assume if we merge an item the item variants should also be merged? Although that might create duplication, e.g. if you have two item_variants with pack size 1 and level 1, should they both still exist in future? Potentially we need `item_variant_link` but I'd really rather avoid going down that road!

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

jmbrunskill:
Refinement: It's fine to have duplicates of `item_variants` use can fix by deleteing

jmbrunskill:
TODO: https://github.com/msupply-foundation/open-msupply/pull/5319#discussion_r1830230131

jmbrunskill:
This is closed by https://github.com/msupply-foundation/open-msupply/pull/11668

PR #5288, opened 2024-11-01
5250 gaps improve validation on packagingitem variants

### LINKED ISSUE #5250: GAPS: Improve validation on Packaging/Item Variants

Appropriate validations are added to API requests
Tests for validation rules are added...

jmbrunskill:
Fixes #5250 

# 👩🏻‍💻 What does this PR do?

Adds validations for item_variants and pack_variants.

The only one the user should see if our UI is right (Duplicate name) is a structured error.

I thought about not allowing an empty string for item_variant name, but I feel like that's probably fine?

I have a non-traditional solution for the duplicate error message issue https://github.com/msupply-foundation/open-msupply/issues/3984 
Interested on feedback, it's definitely a bit of duplication in the code...

What other validations should we have? Is there anything else we need to restrict?

# 🧪 Testing

- [ ] Create and edit item_variants (OMS Central Server)
- [ ] Try doing thing you think you should be allowed too like duplicate names

# 📃 Documentation

  1.
  2.


client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
```diff
@@ -52,9 +52,26 @@ export const ItemVariantModal = ({
...
+            save(draft)
+              .then(() => {
+                success(t('messages.item-variant-saved'))();
+                onClose();
+              })
+              .catch(e => {
+                // We create the same error message as we get from the default handler, but prevent duplicates
+                // This avoids the same error message being displayed multiple times, and the only appears once bug...
+                // https://github.com/msupply-foundation/open-msupply/issues/3984
+                if (
```
jmbrunskill, line 64: This is a bit of duplication, but kind of a nice hack for the issue.

Also makes me think maybe we want to prevent duplicated on the main API catch, but then allow the errors to occur multiple times. This would mean you won't see too many errors right rather than just waiting an arbitrary length of time?

server/service/src/item/item_variant/upsert.rs
```diff
@@ -124,5 +128,38 @@ fn validate(
         return Err(UpsertItemVariantError::ItemDoesNotExist);
     }
 
+    let old_item_variant = ItemVariantRowRepository::new(connection).find_one_by_id(&input.id)?;
+
+    if let Some(old_item_variant) = old_item_variant {
+        if old_item_variant.item_link_id != input.item_id {
```
lache-melvin, line 134: I think for item merge to work correctly, we need to query item by item_link here, and then compare the item ids?

server/service/src/item/packaging_variant/test/mod.rs
```diff
@@ -131,19 +134,136 @@ mod query {
...
+                ..Default::default()
+            },
+        );
+
+        assert_eq!(
+            result.unwrap_err(),
+            UpsertPackagingVariantError::ItemVariantDoesNotExist
+        );
 
         // Test that we can't change the item_variant_id on an existing record???
```
lache-melvin, line 205: the 3 question marks really add a sense of confidence 😁

server/service/src/item/packaging_variant/upsert.rs
```diff
@@ -76,5 +78,35 @@ fn validate(
...
+        PackagingVariantRowRepository::new(connection).find_one_by_id(&input.id)?;
+    if let Some(old_packaging_variant) = old_packaging_variant {
+        if old_packaging_variant.item_variant_id != input.item_variant_id {
+            return Err(UpsertPackagingVariantError::CantChangeItemVariant);
+        }
+    }
+
+    if input.packaging_level < 0 {
+        return Err(UpsertPackagingVariantError::LessThanZero(
+            "packaging_level".to_string(),
```
lache-melvin, line 91: This is a nice solution as opposed to 3 new error variants! 🚀

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
lache-melvin: Love the `preventDuplicate` find!

Still get the double error the first time, but that's part of the bigger refactor issue so I think that's fine. Would love to make some time to try the `preventDuplicate` in the common API catch during cooldown period, rather than in each call - then we could just ignore the handled errors here 🔥 

Screenshot 2024-11-04 at 2 22 12 PM

server/service/src/item/packaging_variant/upsert.rs
```diff
@@ -76,5 +78,35 @@ fn validate(
...
+        ));
+    }
+
+    if let Some(pack_size) = input.pack_size {
+        if pack_size < 0.0 {
```
lache-melvin, line 96: I think this one should be `<=`  - can't have a pack size of 0? Actually, maybe all of these checks should error if the value is 0? Not sure volume of 0, or packaging level of 0 make much sense either...

lache-melvin (changes requested):
Looks great James! Just requesting the item_link fix, and the less than 0 pack size check 🙏

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
lache-melvin: Oh huh - not even after a page refresh?

jmbrunskill: Hmm you're right I thought it was working on Friday but maybe I hadn't refreshed...

server/service/src/item/item_variant/upsert.rs
jmbrunskill: [7aa62da](https://github.com/msupply-foundation/open-msupply/pull/5288/commits/7aa62da25be439d59d3afdf2a62eaaae2dad2139)

Done here, but since we're not properly using item_link yet, it's a bit hard to test. We also don't seem to have any item_merge mock items?

I think we can revisit the tests in #5241 ?

server/service/src/item/packaging_variant/upsert.rs
jmbrunskill: Good point! Done!

lache-melvin (changes requested):
Looks great James! Just one more thing... I think either frontend should prevent user inputting 0 in the packsize/volume, or we should translate this error:

Screenshot 2024-11-05 at 8 40 39 AM

jmbrunskill:
I've pushed a hopefully good enough fix for the 0 pack size

[8bf7715](https://github.com/msupply-foundation/open-msupply/pull/5288/commits/8bf771501ddb6ccb98a48c2fbda18a1a21d7e89e)

I don't really want to translate the error as I think the UI shouldn't allow it. However, if we want to support decimal pack size we have to pick some arbitrary size. However I see in `PackSizeEntryCell` we do have a minimum of 1.

Also in this context it would nice to be able to clear the input as it's not actually required.
That's not supported by the NumericInput though as far as I can tell and would be quite a bit of work to add...

So think this is good enough for now? But open to suggestions!

lache-melvin:
Hm, bit of a hack/different idea, but what do you think about this James? https://github.com/msupply-foundation/open-msupply/compare/5250-gaps-improve-validation-on-packagingitem-variants...5250-gaps-improve-validation-on-packagingitem-variants(non-zero-input)

Was getting this, not able to enter a `0.xx` decimal without it defaulting to min...
https://github.com/user-attachments/assets/da2e1428-5cc9-4092-a292-f8f1ee1904d0

Agree would be good to clear the value - though looks like we also need to implement nullable update on the backend for that, so definitely sounding like more work!

jmbrunskill:
> Hm, bit of a hack/different idea, but what do you think about this James? [5250-gaps-improve-validation-on-packagingitem-variants...5250-gaps-improve-validation-on-packagingitem-variants(non-zero-input)](https://github.com/msupply-foundation/open-msupply/compare/5250-gaps-improve-validation-on-packagingitem-variants...5250-gaps-improve-validation-on-packagingitem-variants(non-zero-input))
> 
> Was getting this, not able to enter a `0.xx` decimal without it defaulting to min (I tried to type 0.2 here) ...
> 
>  Screen.Recording.2024-11-05.at.12.24.34.PM.mov 
> Agree would be good to clear the value - though looks like we also need to implement nullable update on the backend for that, so definitely sounding like more work!

Nice, I think this is a good solution, Pretty sure people will be able to figure out why their 0 value isn't valid :)
Merged to this branch.

lache-melvin (approved):
Sweeet, ship it!

PR #5292, opened 2024-11-01
3923 bundled items (Backend)

### LINKED ISSUE #3923: GAPS: Specify item variants that should be bundled with other item variants

jmbrunskill:
Fixes #3923

# 👩🏻‍💻 What does this PR do?

Repo (Thanks Lache), Service, and graphql for bundled items.

Still to come I think better to create new issues?:

- [ ] Bundled Items UI
- [ ] Bundled Items service layer validations

I'm a little bit concerned about calling this bundled items when it's really just bundled item variants.
Should we change it? Or if we think we'll support bundling items to items or variants to items in the future should we update the db design now so we have optional fields? It's a pain to change these afterwards...

# 🧪 Testing

- [ ] Just run the tests at this stage I think
- [ ] Check the new graphql endpoints work?

# 📃 Documentation

  1.
  2.


server/graphql/Cargo.toml
```diff
@@ -21,6 +21,7 @@ graphql_cold_chain = { path = "cold_chain" }
 graphql_asset_catalogue = { path = "asset_catalogue" }
 graphql_inventory_adjustment = { path = "inventory_adjustment" }
 graphql_item_variant = { path = "item_variant" }
+graphql_item_bundle = { path = "item_bundle" }
```
jmbrunskill, line 24: I went with `item_bundle` to keep it visually next to `item_variant` but I haven't been consistent with that.
Should this just be part of `item_variant` or should it be renamed??

server/repository/src/migrations/v2_04_00/add_bundled_item_table.rs
```diff
@@ -0,0 +1,26 @@
...
+        "add_bundled_item_table"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                CREATE TABLE bundled_item (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    principal_item_variant_id TEXT NOT NULL REFERENCES item_variant(id),
```
jmbrunskill, line 16: Should we add item_id NOT null and make this nullable?

server/repository/src/mock/item_variant.rs
```diff
@@ -0,0 +1,58 @@
+use crate::{item_variant::item_variant_row::ItemVariantRow, mock::item::*};
+
+pub fn mock_item_a_variant_1() -> ItemVariantRow {
```
jmbrunskill, line 3: Decided we'd reached the threshold to deserve mocks.
Haven't refactored any other test cases though. I think that's ok right?

server/service/src/item/mod.rs
```diff
@@ -75,6 +82,31 @@ pub trait ItemServiceTrait: Sync + Send {
     ) -> Result<String, DeletePackagingVariantError> {
         delete_packaging_variant(ctx, input)
     }
+
+    fn get_bundled_items(
```
jmbrunskill, line 86: There's no query for this at the graphql layer, I figured we might want to do a loader of item_variant like we did for item_variant off item?

I figure we can leave that till we do the UI part...

server/service/src/item/bundled_item/test/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(bundled_item.count, 0);
+    }
+
+    #[actix_rt::test]
+    async fn validate_bundled_item() {
+        // TODO validation tests
+
+        // Check that the two item variants are not the same
+
+        // Check that the two item variants are not from the same item (that would be bad)
```
jmbrunskill, line 116: Anything else to check here?

server/repository/src/migrations/v2_04_00/add_bundled_item_table.rs
lache-melvin: I've messaged Mark to double check from a product side, but I think my preferred design, if/when we end up with variants & bundling more integrated with the rest of OMS, would be to create a `default` variant for all items. That way we can handle items with and without many variants set up, without a bunch of forked logic and nullable fields to check... what do you reckon?

server/repository/src/mock/item_variant.rs
lache-melvin: Sounds fine to me, if we are back i n those test cases can always refactor later 🙏

server/service/src/item/bundled_item/test/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+            .unwrap();
+
+        // Update the ratio
+        let _bundled_item = service
+            .upsert_bundled_item(
+                &context,
+                UpsertBundledItem {
+                    id: test_bundled_item_record_id.to_string(),
+                    principal_item_variant_id: mock_item_a().id,
+                    bundled_item_variant_id: mock_item_a().id,
```
lache-melvin, line 62: These should be variant ids?

server/service/src/item/bundled_item/test/mod.rs
lache-melvin: I have two more:

- Can't bundle the same 2 variants multiple times (otherwise could configure same bundle with different ratios, which one should we pick?)
- Prevent nested bundling - check the `principal` variant isn't the `bundled` variant in another bundle

lache-melvin (approved):
Looks good to me James - let's chuck those validation cases in the next issue 🙏

lache-melvin (changes requested):
Whoops, realised I didn't run the tests - failing on `PrincipalItemVariantDoesNotExist` for me!

lache-melvin:
> Whoops, realised I didn't run the tests - failing on `PrincipalItemVariantDoesNotExist` for me!

Looks like one fix is to add `item_variants` to the `all_mock_data` object, other is due to item_id being used instead of variant id 🙏

server/repository/src/migrations/v2_04_00/add_bundled_item_table.rs
lache-melvin: > I'm a little bit concerned about calling this bundled items when it's really just bundled item variants.
Should we change it? Or if we think we'll support bundling items to items or variants to items in the future should we update the db design now so we have optional fields? It's a pain to change these afterwards...

Mark seemed happy with the default variant solution, if you are? And @andreievg?

server/service/src/item/bundled_item/test/mod.rs
jmbrunskill: Yes

server/repository/src/migrations/v2_04_00/add_bundled_item_table.rs
jmbrunskill: I think we can get away with not deciding for now? I'm thinking the default variant solution is probably a good one, but we don't need it yet right?  Maybe we should create an issue for the default Item variants?

lache-melvin (approved):
LGTM 🙏  Sorry forgot to approve after pushing fix 😁

PR #5301, opened 2024-11-04
5282 Allow initialisation from export

### LINKED ISSUE #5282: Initalising from export might be incorrect
@roxy-dao  I'm facing an issue when i try to login into the app. The issue is this : "Your account does not have access to any stores on this site". I'm using the Admin credentials from the Users.txt file.

_Originally posted by @myspace20 in https://github.com/msupply-foundation/open-msupply/issues/5219#issuecomment-2448820091_
        

Needs investigation as to why...

jmbrunskill:
Fixes #5282

# 👩🏻‍💻 What does this PR do?

Initialising from reference data was failing due to a missing `om_date_of_death`.
I've fixed a couple of other errors too such as missing fields on `activity_log` and `stock_take_line`

There's still a few other sync errors but don't seem to critical...

Are we ok with defaults for these fields?
At some point we might need to update the reference data file, but this seemed like a nice quick fix...

# 🧪 Testing

- [ ]  APP__DATABASE__DATABASE_NAME=reference1 cargo run --bin remote_server_cli -- initialise-from-export -n reference1
- [ ]  APP__DATABASE__DATABASE_NAME=reference1 cargo run 
- [ ] Check you can login!

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/sync/translations/stocktake_line.rs
+++ b/server/service/src/sync/translations/stocktake_line.rs
@@ -34,7 +34,6 @@ pub struct LegacyStocktakeLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub item_line_ID: Option<String>,
     pub item_ID: String,
-    #[serde(default)]
     pub item_name: String,
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub Batch: Option<String>,
--- a/server/service/src/sync/translations/stocktake_line.rs
+++ b/server/service/src/sync/translations/stocktake_line.rs
@@ -34,6 +34,7 @@ pub struct LegacyStocktakeLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub item_line_ID: Option<String>,
     pub item_ID: String,
+    #[serde(default)]
     pub item_name: String,
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub Batch: Option<String>,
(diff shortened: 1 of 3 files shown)


server/service/src/sync/translations/stocktake_line.rs
```diff
@@ -34,6 +34,7 @@ pub struct LegacyStocktakeLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub item_line_ID: Option<String>,
     pub item_ID: String,
+    #[serde(default)]
     pub item_name: String,
```
lache-melvin, line 38: I get it for the options, but why would this need a default? Maybe we need to fix the reference data if we have stocktake lines that don't have items/item_names?

lache-melvin (changes requested):
Sorry coming back to this... I'm wondering if actually all we need to do is update the reference data? I suspect we wouldn't need the `default`s if we do an `export-initialisation` on up-to-date data?

jmbrunskill:
> Sorry coming back to this... I'm wondering if actually all we need to do is update the reference data? I suspect we wouldn't need the `default`s if we do an `export-initialisation` on up-to-date data?

Yes, I agree that would probably work too. Do you know if the original datafile exists somewhere to re-generate from? Thought this was the quicker solution :)

server/service/src/sync/translations/stocktake_line.rs
jmbrunskill: Yeah, it's not ideal :) Especially for item_name maybe I take that one out but merge the other two?

jmbrunskill: Remove this... I figure the rest is ok to have, and get's people up and running for now, until we can re-generate the data...

lache-melvin (approved):
Ah ok yeah fair enough - yep this works for me @jmbrunskill! I was just thinking to generate from the [Demo primary datafile](https://drive.google.com/drive/folders/10EFbD6n--2_zlifHhmitgxRJl94kzMzY) but this is working, let's get it in to unblock 🙏

PR #5309, opened 2024-11-04
#4917 Fix patches with nullable values

### LINKED ISSUE #4917: Location disappears when i enter cost and press save
## What went wrong? 😲
on oms tablet version 2.2.01 - I had item and locations in but didn't have costs..

i added costs and pressed save and my location then disappears (note it only disappears after I press save but it is still there before I press save when I am editing)

Screenshot 2024-10-01 at 8 41 54 AM

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #4917

# 👩🏻‍💻 What does this PR do?

Problem was caused by the `setNullableInput` function when passing a partial update object to the mutation function -- any properties that were not defined were getting set to `{value: null}` by the nullable input function.

Have changed `setNullableInput` to require an explicit `null` value in order to set `null` in backend.

I'm pretty sure this doesn't break anything else -- have had a quick check and seems okay elsewhere.

See inline comments for other minor notes.

We're actually not being particularly consistent with how we handle `null` updates. Some use the `setNullableInput` in the front-end, whereas other places send a full update object (with `null` values) and are handled by the GraphQL layer on the server (where `null` => `None`). It would be good to consolidate this at some point and have a standard pattern.

Personally, I don't like the use of `setNullableInput` -- formatting null values as `{value: null}` is an implementation detail of the GraphQL abstraction, so the front-end shouldn't even need to know about it. Ideally we could:
- send partial updates
- `null` values are treated as `null`/`None`
- `undefined` values are not updated

As I understand it, there's some complexity in the GraphQL layer that makes this not completely straightforward, so would need a bit of careful re-factoring -- something for another day...

# 🧪 Testing

- [ ] Go to Inventory -> View Stock and click on an item
- [ ] If the item doesn't have a location, set one, save, then exit back to the list
- [ ] Go back to the same item, change some other values (e.g. costs)
- [ ] Save, and ensure the "Location" is kept as-is
- [ ] Maybe check the behaviour in some other places where user can set `null` values and ensure no regressions


client/packages/common/src/utils/functions/ApiUtils.tsx
```diff
@@ -4,9 +4,9 @@ export const setNullableInput = <NODE, KEY extends keyof NODE>(
   key: KEY,
   entity?: NODE | null | undefined
 ) => {
-  if (entity == null) return { value: undefined };
+  if (entity === null) return { value: null };
```
CarlosNZ, line 7: Classic error caused by the less strict equality operator (`==`). This returns true for `undefined` as well, so properties that were not defined were replaced with `null`s.

client/packages/common/src/utils/functions/ApiUtils.tsx
```diff
@@ -4,9 +4,9 @@ export const setNullableInput = <NODE, KEY extends keyof NODE>(
...
 
-  if (!entity) return undefined;
+  if (entity === undefined) return undefined;
```
CarlosNZ, line 9: Not completely required for this use case, but better to be explicit here. `!entity` would also match `0` or `""` .

client/packages/system/src/Stock/api/hooks/useStockLine.ts
```diff
@@ -81,8 +81,9 @@ export function useStockLine(id?: string) {
     return result;
   };
   const update = async () => {
-    updateMutation(patch);
     setIsDirty(false);
+    await updateMutation(patch);
+    setPatch({});
```
CarlosNZ, line 86: Unrelated, but something I noticed when testing. Best to clear the patch after update, so you start fresh with subsequent updates. (This was why the original "bug" didn't happen sometimes -- if you hadn't done something to clear the patch object, the original location would get updated, so no bug -- perhaps that's why you saw it "intermittently", @lache-melvin ?)

jmbrunskill:
It's a slightly different issue but Item Variant is also disappearing if you change price.

Can you investigate in this PR.

jmbrunskill (changes requested):
Nullable Input changes All looks good.
Also had a poke around, only thing I found was that item_variant issue which doesn't appear to be nullable input related but is on the same screen.

Totally agree about your comments about consistency between full and partial updates, and potential to simiplify nullable inputs.

CarlosNZ:
> It's a slightly different issue but Item Variant is also disappearing if you change price.
> 
> Can you investigate in this PR.

Yeah, good catch. Needed to add additional check for the entity object being defined, but with an `undefined` value on it: b65180023b

I think the front-end is doing too much work here for implementation details that should be behind an abstraction layer on the server. Would be good to refactor this at some point. You think it's worth making a "refactor" issue?

jmbrunskill:
> I think the front-end is doing too much work here for implementation details that should be behind an abstraction layer on the server. Would be good to refactor this at some point. You think it's worth making a "refactor" issue?

Yup absolutely!

jmbrunskill (approved):
Looks good, thanks!

PR #5317, opened 2024-11-05
#5164 Fix Prescriptions DatePicker/Confirmation conflict

### LINKED ISSUE #5164: Clicking on a datetime picker in Prescriptions shows the warning straight away, warning appears again after clicking anywhere on the screen
## What went wrong? 😲

**Issue 1** - the warning the lines will be removed appears straight away, without giving users a chance to even see the datetime picker. Should the flow be `user clicking on the datetime picker -> seeing the window -> clicking on the past date -> clicking Ok to confirm -> then only seeing the warning`? 

Or if the warning should appear before any actions are taken by users, then it shouldn't overlrap the datetime picker -> they should appear one by one. 

**Issue 2** - after cancelling out of the warning and clicking anywhere on the screen, the warning appears again

The video of the behaviour can be found ->>> [here](https://drive.google.com/drive/folders/1chwrPRBjIUMlbce6h2jpYFIsFNpJBbDo)

## Expected behaviour 🤔

**Issue 1** - user clicking on the datetime picker -> seeing the window -> clicking on the past date -> clicking Ok to confirm -> then only seeing the warning

**Issue 2** - the warning should not appear twice if you cancel out of it earlier

## How to Reproduce 🔨

1. Go to `Dispensary -> Prescriptions`
2. Create a new prescription and add an item to it
3. Click on the datetime picker to see the first issue
4. Click Ok or Cancel and then click anywhere else on the screen to see the second issue

## Your environment 🌱

- Open mSupply Version: V2.3.0-RC10

- Platform: Tablet (apk)

- Database type: SQLite3

CarlosNZ:
Fixes #5164

# 👩🏻‍💻 What does this PR do?

This seems too simple, but it appears to fix the problem! 🤷‍♂️

I just changed the `onChange` prop the DatePicker to `onAccept`, so the `handleDateChange` function doesn't get called when the picker opens (which it does on Mobile)

Haven't tested on Tablet specifically. However, can replicate the problem by changing to a mobile layout with browser DevTools, and this fix works for that.

# 🧪 Testing

- [ ] If using Desktop browser, open DevTools and switch to a mobile layout (the DatePicker will be different)
- [ ] Open a Prescription and click on date
- [ ] Should *not* get any item deletion confirmation modal
- [ ] Change the date and submit
- [ ] Should get the confirmation modal
- [ ] Also test in normal browser to confirm existing behaviour is unchanged

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
@@ -125,7 +125,11 @@ export const Toolbar: FC = () => {
                   defaultValue={new Date()}
                   value={DateUtils.getDateOrNull(prescriptionDate)}
                   format="P"
-                  onChange={handleDateChange}
+                  // Using onAccept rather than onChange -- on mobile, onChange
+                  // is triggered when first opening the picker, which causes UI
+                  // conflict with the confirmation modal
+                  onAccept={handleDateChange}
+                  onChange={() => {}}
                   maxDate={new Date()}
                 />
               }


client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
```diff
@@ -125,7 +125,11 @@ export const Toolbar: FC = () => {
                   defaultValue={new Date()}
                   value={DateUtils.getDateOrNull(prescriptionDate)}
                   format="P"
-                  onChange={handleDateChange}
+                  // Using onAccept rather than onChange -- on mobile, onChange
+                  // is triggered when first opening the picker, which causes UI
+                  // conflict with the confirmation modal
+                  onAccept={handleDateChange}
+                  onChange={() => {}}
```
CarlosNZ, line 132: `onChange` is a required prop, so passing a dummy function seemed simpler than refactoring DatePicker to handle non-compulsory `onChange`.

lache-melvin (approved):
Working great! As mentioned, found this issue (#5325) but not related to these changes. Thanks Carl 🚀

PR #5319, opened 2024-11-05
Bundled Item Validations

### LINKED ISSUE #5300: Item Bundling validation

Adds the item bundling validation logic from #5292

Only valid item bundles can be created

### Moneyworks Jobcode 🧰

CCEIDEV

jmbrunskill:
Fixes #5300

# 👩🏻‍💻 What does this PR do?

Adds validations for bundled items upsert

Anything else missing?
Should some of these be strucutured errors?
If so should we change that once we have UI?

# 🧪 Testing

- [ ] Try to create an bundled item using existing item variants
- [ ] Check that you can't bundle an item that already has bundled items (can't be nested)
- [ ] Check that you can't create the same bundled items twice
- [ ] Check that you can't do anything else that you feel like shouldn't be allowed...

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/item/bundled_item/upsert.rs
+++ b/server/service/src/item/bundled_item/upsert.rs
@@ -1,6 +1,9 @@
 use repository::{
-    item_variant::bundled_item_row::{BundledItemRow, BundledItemRowRepository},
-    RepositoryError, StorageConnection,
+    item_variant::{
+        bundled_item::{BundledItemFilter, BundledItemRepository},
+        bundled_item_row::{BundledItemRow, BundledItemRowRepository},
+    },
+    EqualFilter, RepositoryError, StorageConnection,
 };
 
 use crate::{check_item_variant_exists, service_provider::ServiceContext};
@@ -10,6 +13,9 @@ pub enum UpsertBundledItemError {
     CreatedRecordNotFound,
     PrincipalItemDoesNotExist,
     BundledItemDoesNotExist,
+    DuplicateBundledItem,
+    CanNotNestBundledItems,
+    CanNotBundleItemWithItself,
     DatabaseError(RepositoryError),
 }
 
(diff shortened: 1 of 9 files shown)


lache-melvin:
Looks like a test is failing here James?

server/service/src/item/bundled_item/upsert.rs
```diff
@@ -68,12 +74,57 @@ fn validate(
...
+            None => return Err(UpsertBundledItemError::BundledItemDoesNotExist),
+        };
+
+    if input.principal_item_variant_id == input.bundled_item_variant_id {
+        return Err(UpsertBundledItemError::CanNotBundleItemWithItself);
+    }
+
+    // Check that item_ids are not the same
+    // Technically this has a problem if the item is merged but should be very rare...
+    if principal_item_variant.item_link_id == bundled_item_variant.item_link_id {
```
lache-melvin, line 95: I ended up joining to item in #5324, if we merge those two, and use variant_repo rather than variant_row_repo in the `check_item_variant` call... can fix here 😅 

Aware/agree the risk is probably low here for now, but I'm a bit worried about the proliferation of item merge TODOs throughout the codebase, would love the get the right patterns in for us to start duplicating...

lache-melvin (commented):
Looking good James, just need to fix the failing test, and maybe the item link thing? Let me know your thoughts there...

Frontend PR should prevent these errors, so happy with standard errors for now 🙏

server/service/src/item/bundled_item/upsert.rs
jmbrunskill: > TODOs throughout the codebase, would love the get the right patterns in for us to start duplicating...

I have asked for this to be documented! @andreievg @roxy-dao any progress?

lache-melvin (approved):
🚀

server/service/src/item/bundled_item/upsert.rs
roxy-dao: Nope... sorry ;-;

PR #5323, opened 2024-11-05
#4343 Activity log for Demographics

### LINKED ISSUE #4343: Add activity logs for demographics mutations

Demographics backend was implemented without adding in `ActivityLog` entries.

We've got a few of these lying around now:

https://github.com/msupply-foundation/open-msupply/blob/8b222383c736f64fc5a4b9873bb82dfb30639c49/server/service/src/demographic/insert_demographic_indicator.rs#L47

Changes to demographics indicators and projections should be logged

CarlosNZ:
Fixes #4343

# 👩🏻‍💻 What does this PR do?

Just adds activity log entries for the Demographics mutations, with accompany Activity log enum values:
- `DEMOGRAPHIC_INDICATOR_CREATED`
- `DEMOGRAPHIC_INDICATOR_UPDATED`
- `DEMOGRAPHIC_PROJECTION_CREATED`
- `DEMOGRAPHIC_PROJECTION_UPDATED`

Just followed the existing pattern here, not sure if there's anything else to consider?

Notice there's no `store_id` inserted -- I see this comes from the Service Context directly, so presumably we don't need to change anything here?

# 🧪 Testing

- [ ] On Central Server, go to Manage -> Demographics
- [ ] Insert Demographic indicator
- [ ] Update Demographic indicator
- [ ] Edit the "Growth on previous year" values at the top (that's the Demographic Projection)
- [ ] Check that correct activity_log entries are added for each of the above operations (will need to manually inspect database)


client/packages/system/src/IndicatorsDemographics/DetailView/IndicatorsDemographics.tsx
```diff
@@ -225,22 +225,19 @@ const IndicatorsDemographicsComponent = () => {
...
           isLoading={isLoadingProjection}
           setData={handleGrowthChange}
-        ></GrowthRow>
+        />
         <DataTable
           data={Object.values(draft)}
           columns={columns}
           id={'indicators-demographics-table'}
-        ></DataTable>
+        />
```
CarlosNZ, line 240: Unrelated, but spotted when testing out this area -- self-closing tags are much more succinct and clear when no inner content :)

project-words.txt
```diff
@@ -6,6 +6,7 @@ coldchain
 conforma
 datasource
 datetime
+Insertable
```
jmbrunskill, line 9: Glad I'm not the only one adding these :)

jmbrunskill (approved):
Thanks Carl, looks good.

I think we also need to add these for Upsert/Delete `item_variant`, `packaging_variant`, and `bundled_item`.

Do you think we raise a new issue or add to this PR?

CarlosNZ:
> I think we also need to add these for Upsert/Delete `item_variant`, `packaging_variant`, and `bundled_item`.
> 
> Do you think we raise a new issue or add to this PR?

Sure, I may as well do some more copy-pasting in here. New Enum values for each of those?

CarlosNZ:
> > I think we also need to add these for Upsert/Delete `item_variant`, `packaging_variant`, and `bundled_item`.
> > Do you think we raise a new issue or add to this PR?
> 
> Sure, I may as well do some more copy-pasting in here. New Enum values for each of those?

Actually change of plan -- I realised the activity logs for those ones will be a bit tricker, as they're "Upserts" which we haven't made activity_log entries for that I can see, and the logic will be a bit trickier as currently the service layer doesn't know whether the upsert was a "create" or "update".

Will make new issue.

CarlosNZ:
> Will make new issue.

https://github.com/msupply-foundation/open-msupply/issues/5341

PR #5324, opened 2024-11-05
5302 bundled item UI

### LINKED ISSUE #5302: Bundled Item UI

See designs here: https://github.com/msupply-foundation/open-msupply/issues/3923#issuecomment-2418134318

UI to add/edit and remove bundled variants from other item variants.

It should also prevent the user from creating nested bundles

### Moneyworks Jobcode 🧰

CCEIDEV

lache-melvin:
Fixes #5302

# 👩🏻‍💻 What does this PR do?

Adds item bundling UI:

Screenshot 2024-11-06 at 11 43 49 AM

As principal variant:
Screenshot 2024-11-06 at 11 44 16 AM

As bundled variant:
Screenshot 2024-11-06 at 11 44 39 AM

Sorry she's a bit of a big one, did end up needing the join on the item row, so that we could get all variants for that item 😅 

# 🧪 Testing

- [ ] Create item variants for at least 2 items
- [ ] Bundle a variant of another item with your item variant
- [ ] Edit the bundle (e.g. change the ratio)
- [ ] Delete the bundle
- [ ] Check you can't bundle the same item with itself
- [ ] Check you can't create two bundles with the same two variants
- [ ] Check you can't create nested bundled items (i.e. if item A is bundled with item B, item B can't then be bundled with item C)

# 📃 Documentation

  1.
  2.


client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemPackagingVariantsTable.tsx
```diff
@@ -49,6 +49,7 @@ export const ItemPackagingVariantsTable = ({
       Cell: update ? NonZeroInputCell : TooltipTextCell,
       label: 'label.volume-per-unit',
       setter: updatePackaging,
+      width: 150,
```
lache-melvin, line 52: tiny side quest to fix label wrapping

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+          labelWidth="200"
+          Input={
+            <Box width="100%">
+              <StockItemSearchInput
+                autoFocus={!draft.itemId}
+                openOnFocus={!draft.itemId}
+                onChange={item => updateDraft({ itemId: item?.id })}
+                currentItemId={draft.itemId}
+                extraFilter={item => item.id !== principalVariant.itemId}
+              />
```
CarlosNZ, line 103: How hard would it be to filter the item list to only show ones that have Variants defined?

I guess you'd need to update the possible filters for Items, so might be a bit much for now. I think it would be useful though -- it seems a little annoying to be able to select an item and then discover it has not variants that can be bundled. Maybe a new issue?

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+          <>
+            <InputWithLabelRow
+              label={t('label.variant')}
+              labelWidth="200"
+              Input={
+                <Box width="100%">
+                  <ItemVariantSearchInput
+                    onChange={variantId =>
+                      updateDraft({
+                        variantId: variantId ?? '',
```
CarlosNZ, line 117: Shouldn't the fallback be `undefined` rather than `""`? (I guess `""` still works, but just seems a little more "correct" 🤷‍♂️)

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1169,6 +1176,7 @@
   "messages.confirm-changing-from-new": "Note: changing from 'New' will remove any lines with zero quantity.",
   "messages.confirm-delete-assets_one": "This will permanently remove this asset.",
   "messages.confirm-delete-assets_other": "This will permanently remove the {{count}} selected assets.",
+  "messages.confirm-delete-bundled-item": "This will permanently delete this bundled item",
```
CarlosNZ, line 1179: This kind of implies that it will delete the item rather than "unbundling" it. Perhaps "This will permanently remove this item from the bundle", or "This will un-bundle this item" ?

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemVariants.tsx
```diff
@@ -0,0 +1,157 @@
...
+      label: 'label.item-variant',
+      accessor: ({ rowData }) =>
+        `${rowData.principalItemVariant?.itemName} - ${rowData.principalItemVariant?.name}`,
+    },
+    {
+      key: 'ratio',
+      Cell: TooltipTextCell,
+      label: 'label.ratio',
+      description: 'description.bundled-item-ratio',
+      accessor: ({ rowData }) => 1 / rowData.ratio,
```
CarlosNZ, line 142: Probably want to put a decimal limit on this display. When the original ratio is `3`, this shows up as `0.3333333333333333`

client/packages/system/src/Item/api/hooks/bundledItem/useUpsertBundledItem.ts
```diff
@@ -0,0 +1,74 @@
...
+      throw new Error(t('error.failed-to-save-bundled-item'));
+    },
+    onSuccess: () => {
+      queryClient.invalidateQueries(keys.detail(principalVariant.itemId));
+    },
+  });
+
+  return {
+    draft,
+    isComplete: getIsComplete(draft),
```
CarlosNZ, line 64: I was a bit unsure what we meant by "complete" -- perhaps `canSubmit` would be more specific? Then again, I see this name has been used in a few places now, so probably not worth changing them all. See what you think.

client/packages/system/src/Item/api/operations.graphql
```diff
@@ -77,11 +77,31 @@ fragment PackagingVariant on PackagingVariantNode {
...
+  name
+  itemId
+  itemName
+}
+
+fragment BundledItem on BundledItemNode {
+  __typename
+  id
+  ratio
+  principalItemVariant {
```
CarlosNZ, line 91: Are we being inconsistent with naming? Is it the "base" item or the "principal" item? Can we keep it the same throughout -- I think I have a slight preference for "base", but not a strong one.

CarlosNZ (changes requested):
Awesome, looking very pretty :)

Just a couple of nitpicky things, re naming and wording, if you want to take a look. Cheers

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemModal.tsx
lache-melvin: Hm yeah nice idea, but yep would be quite a bit of extra work... I also don't know whether it would be confusing to only see a subset of the items? Would need to explain somewhere that only items with variants are available in the list 🤷‍♀️ 

Thinking it's okay for now, given this is only on Central Server, and only used for GAPS, where we're expecting a known data set. Future idea is to actually have "default" item variant, for any items that don't have variants defined, which would solve this issue 🙏

client/packages/common/src/intl/locales/en/common.json
lache-melvin: [1282793](https://github.com/msupply-foundation/open-msupply/pull/5324/commits/128279354807e9f9e7b6d3039d68463d8ecad2aa)

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemVariants.tsx
lache-melvin: [6e643f5](https://github.com/msupply-foundation/open-msupply/pull/5324/commits/6e643f5665039a911381c6191b63c0137c63f2cd)

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/BundledItemModal.tsx
lache-melvin: [36dc9bf](https://github.com/msupply-foundation/open-msupply/pull/5324/commits/36dc9bffaf203e1b0ff151951666d4f35121983f)

lache-melvin:
Thanks for review Carl! Let me know what you think of this change... thought it might be helpful to link to the principal item? [2f3f1b6](https://github.com/msupply-foundation/open-msupply/pull/5324/commits/2f3f1b6f48fb2720d667242f4582c97d829ed4a5)

CarlosNZ (approved):
Sweet, working nicely, looking good 👍🏼 

Cheers

PR #5327, opened 2024-11-06
5310 add graphql loader for values

### LINKED ISSUE #5310: Add graphql loader for indicator values for program_indicator queries

Builds on #5215
Adds loader to add values to program_indicator nodes

Adds a simple loader
Loader doesn't include more complex hash including period_id, supplier_id, customer_id. 
This means with multiple queries for indicators where those values differ might cause mixing from the loader, but we can iterate on this later.

Could query indicator values from service layer also

### Moneyworks Jobcode 🧰

OMS:CIMDV

fergie-nz:
Fixes #5310 

Adds indicator value loader for program indicator queries

Doesn't link up to repo layer yet as value table not yet implemented, so has just dummy service layer functions which return no values.

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

Cannot be tested yet - Wait until plumbed in to repo layer

# 📃 Documentation


server/service/src/programs/program_indicator/query.rs
```diff
@@ -153,32 +155,35 @@ impl ProgramIndicator {
...
-                None => ValueType::String,
-            },
+            // r#type: match column.value_type {
+            //     // TODO remove optional value type if we initialise default values on requisition creation?
+            //     Some(value_type) => match value_type {
+            //         IndicatorValueType::String => ValueType::String,
+            //         IndicatorValueType::Number => ValueType::Number,
+            //     },
+            //     None => ValueType::String,
+            // },
```
fergie-nz, line 182: I've left this commented as I'm not sure if we need it - and if we do need it, should it be here? Or should it be in the same node as the value?

I also wasn't sure if the value node should be more flattened rather than having a value node on the column.

You can see in IndicatorColumnNode, the value field returns another node which has within it the value table value and value.id.
 I had a feeling that leaving it nested as it is wasn't the best, but was reluctant to refactor too much in case flattening it also isn't correct.

I also wasn't quite sure how to take the loaded node result (which includes two values we want: id, and value), and then adding that two a flattened column value node

server/service/src/programs/program_indicator/query.rs
```diff
@@ -153,32 +155,35 @@ impl ProgramIndicator {
...
+            id: line.id,
         }
     }
 
     // TODO add to_domain utility function
 }
 
 impl IndicatorColumn {
-    pub fn from_domain(column: IndicatorColumnRow) -> IndicatorColumn {
+    pub fn from_domain(column: IndicatorColumnRow, line_id: String) -> IndicatorColumn {
```
fergie-nz, line 172: passing line_id down so that the loader in column can call with both line_id and column_id

server/graphql/core/src/loader/program_indicator_value.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    async fn load(
+        &self,
+        loader_inputs: &[IndicatorValueLoaderInput],
+    ) -> Result<HashMap<IndicatorValueLoaderInput, Self::Value>, Self::Error> {
+        let service_context = self.service_provider.basic_context()?;
+
+        let (period_id, supplier_store_id, customer_name_id) =
+            // TODO replace with logic to not assume only one combination queried at any time.
+            if let Some(loader_input) = loader_inputs.first() {
```
fergie-nz, line 45: This assumes that only one requisition is queried at anytime, otherwise values might get mixed on loader bundling

fergie-nz:
Ready for review! 

I had a play with some ways to make the loader better, but have settled on this most simple case which doesn't add any complexities and is easy to understand.

This is calling the loader twice - once for id, and once for value_id. 

This could cause performance drops. I'm not sure if that is enough to be of concern.

This could be improved by calling the loader once then caching those values to flatten into the indicatorColumnNode. I had a little play, but was struggling to get it working so have left as is for now.

Chris-Petty:
ya got conflicts @fergie-nz

Chris-Petty (approved):
Looks good to me!

roxy-dao:
@Chris-Petty sorry Chris.. gonna need you to re-review. There have been some changes in the base branch that have been brought forth to here (: and also did some refactoring

server/graphql/types/src/types/program/program_indicator.rs
```diff
@@ -180,6 +185,37 @@ impl IndicatorColumnNode {
     pub async fn column_number(&self) -> i32 {
         self.column.column_number
     }
+
+    pub async fn value(
```
roxy-dao, line 189: can grab Id from here, don't need a separate value_id that calls the same loader just to get id

Chris-Petty (approved):
Sweet looks good!

PR #5329, opened 2024-11-06
Improve handing of backdated date

### LINKED ISSUE #5325: Can't backdate prescriptions
## What went wrong? 😲

Can't backdate a prescription, with or without lines:

https://github.com/user-attachments/assets/b2701345-c61b-4045-a5bf-f6ad75f135dc

Maybbbe related, as shown at end of video, if you cancel the date change, it still shows the selected date until you refresh the page/

## Expected behaviour 🤔

I can backdate the prescription!

## How to Reproduce 🔨

1. Go to a prescription
2. Try to change the date

## Your environment 🌱

- Open mSupply Version: v2.3.1 (working in v2.3.0)

- Platform: Chrome 

- Database type:

jmbrunskill:
Fixes #5325

# 👩🏻‍💻 What does this PR do?

We're now comparing if a `backdated date` is in the future, rather than if just greater than created date.

We're displaying the following dates in the frontend (in order)

`backdated_datetime`
`picked_datetime`
`created_datetime`

We could handle the situation where we're backdating to a date after created_date differently, but simply setting the other fields such as `picked_date`, as it's kind of not backdated. 
This ways simpler though, which I like...

# 🧪 Testing

- [ ] Create a prescription in the past without setting a prescription date/backdating (more than 2 days is needed to test properly)
- [ ] This could either by an old record or, by waiting till a few days to continue testing, or editing the database
- [ ] Try to set the Backdated date, to something in the past, but after created date
- [ ] Try other combinations of date times to select
- [ ] Check that the date displays properly after updating/changing these dates.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
@@ -19,22 +19,14 @@ import { ClinicianSearchInput } from '../../../../system/src/Clinician';
 import { usePrescriptionRows } from '../api/hooks/line/usePrescriptionRows';
 
 export const Toolbar: FC = () => {
-  const {
-    id,
-    patient,
-    clinician,
-    prescriptionDate,
-    createdDatetime,
-    pickedDatetime,
-    update,
-  } = usePrescription.document.fields([
-    'id',
-    'patient',
-    'clinician',
-    'prescriptionDate',
-    'createdDatetime',
-    'pickedDatetime',
-  ]);
+  const { id, patient, clinician, prescriptionDate, createdDatetime, update } =
+    usePrescription.document.fields([
+      'id',
+      'patient',
+      'clinician',
+      'prescriptionDate',
+      'createdDatetime',
+    ]);
   const onDelete = usePrescription.line.deleteSelected();
   const onDeleteAll = usePrescription.line.deleteAll();
   const { items } = usePrescriptionRows();
@@ -81,7 +73,6 @@ export const Toolbar: FC = () => {
 
   const defaultPrescriptionDate =
     DateUtils.getDateOrNull(prescriptionDate) ??
-    DateUtils.getDateOrNull(pickedDatetime) ??
     DateUtils.getDateOrNull(createdDatetime) ??
     new Date();
 
--- a/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
@@ -19,13 +19,22 @@ import { ClinicianSearchInput } from '../../../../system/src/Clinician';
 import { usePrescriptionRows } from '../api/hooks/line/usePrescriptionRows';
 
 export const Toolbar: FC = () => {
-  const { id, patient, clinician, prescriptionDate, update } =
-    usePrescription.document.fields([
-      'id',
-      'patient',
-      'clinician',
-      'prescriptionDate',
-    ]);
+  const {
+    id,
+    patient,
+    clinician,
+    prescriptionDate,
+    createdDatetime,
+    pickedDatetime,
+    update,
+  } = usePrescription.document.fields([
+    'id',
+    'patient',
+    'clinician',
+    'prescriptionDate',
+    'createdDatetime',
+    'pickedDatetime',
+  ]);
   const onDelete = usePrescription.line.deleteSelected();
   const onDeleteAll = usePrescription.line.deleteAll();
   const { items } = usePrescriptionRows();
@@ -69,6 +78,12 @@ export const Toolbar: FC = () => {
       },
     });
   };
+
+  const defaultPrescriptionDate =
+    DateUtils.getDateOrNull(prescriptionDate) ??
+    DateUtils.getDateOrNull(pickedDatetime) ??
+    DateUtils.getDateOrNull(createdDatetime) ??
+    new Date();
 
   return (
     <AppBarContentPortal sx={{ display: 'flex', flex: 1, marginBottom: 1 }}>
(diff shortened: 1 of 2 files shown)


client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
```diff
@@ -70,6 +79,12 @@ export const Toolbar: FC = () => {
     });
   };
 
+  const defaultPrescriptionDate =
+    DateUtils.getDateOrNull(prescriptionDate) ??
+    DateUtils.getDateOrNull(pickedDatetime) ??
+    DateUtils.getDateOrNull(createdDatetime) ??
```
jmbrunskill, line 85: If it's not been at least picked do we want this to just show today? Even if we created it a few days ago?

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -183,3 +186,76 @@ fn lines_to_trim(
...
+
+#[cfg(test)]
+mod test {
+    use chrono::Utc;
+    use repository::InvoiceRow;
+    use repository::InvoiceStatus;
+    use repository::InvoiceType;
+
+    #[actix_rt::test]
+    async fn handle_new_backdated_datetime_test() {
```
jmbrunskill, line 198: Shame on me for not creating this test earlier :(

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
lache-melvin: I don't think I see the case for picked date? Would think this should show created date, unless the prescription is backdated, in which case show that date?

Possibly out of scope, but I almost wonder about showing created date and prescription date separately? At the very least, I don't find it particularly intuitive that the date in the list view and the date in the detail view don't always match (list view always shows created date...)

lache-melvin (changes requested):
Just would like to remove picked date from the default date, otherwise working great, thanks for jumping on this James 🚀

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
jmbrunskill: > I don't think I see the case for picked date? Would think this should show created date, unless the prescription is backdated, in which case show that date?
> 
> Possibly out of scope, but I almost wonder about showing created date and prescription date separately? At the very least, I don't find it particularly intuitive that the date in the list view and the date in the detail view don't always match (list view always shows created date...)

What about the case where it was created 2 days ago, then 1 day later you picked your items and verified it. You want the prescription date to show as yesterday when you re-open it right?
That was my logic anyway...

This whole question shows up because we've started showing the date on the prescription, previously you never saw it apart from the list view.

Speaking of which, it's a great point, that only created date is in the ListView. If we show something similar to this for prescription date, it's going to mess with sorting though :(

@mark-prins - and thought on that? Should we create new issue?

jmbrunskill: I think the simple solution is to just add `PickedDate` as another column in the List View...

mark-prins: I don't see the need for picked date in the mix. The prescription date is functioning as an adjustable created date - any actions taken against the prescription are a separate concern.
My take: remove picked date from that list of fallbacks, and show prescription date in the list view.
I'd also show 'created date' in the right hand panel - that's an extra bit of information which is interesting to know, when it differs from prescription date

jmbrunskill: My argument is that picked date is the date when the stock movement occurs so should be the one we show as the prescription date? But I agree it adds complexity, will try to see if I can get created_date/backdated_date as prescription date showing in the list view without breaking sorting...

lache-melvin (approved):
Happy to merge this as is James! If the list view column is an easy change that would be a nice touch, but I reckon that's also fine in another issue/wouldn't spend too much time.

It's an interesting point about the picked date - thinking of the prescription date as "the date the stock movement happened" as opposed to "the date I created this prescription, or the date I'm saying I created it". My gut feel is that end users would be more likely to be thinking in terms of the latter, but something to keep an ear out for 👀 

To be fair, I would imagine this is fairly unlikely to come up, can't see why anyone would create blank prescriptions and leave them hanging for days... though I suppose you never know! 😁

jmbrunskill:
> It's an interesting point about the picked date - thinking of the prescription date as "the date the stock movement happened" as opposed to "the date I created this prescription, or the date I'm saying I created it". My gut feel is that end users would be more likely to be thinking in terms of the latter, but something to keep an ear out for 👀

I disagree, the whole point of backdating is to record when the transfer of stock happened, so it feels like the transaction date what I'd want to see, however in normal operation the created date and picked date should be the same though, so agree it's not important...

mark-prins:
oh, right. I was thinking of prescription date as the date of the script / the actual prescription. and not a recording of when the transfer of stock happens. that's recorded by the picked date 🤔 or in OG, the confirm date - which is what this 'prescription date' is changing. perhaps it's a naming / labelling issue?

client/packages/invoices/src/Prescriptions/DetailView/Toolbar.tsx
GaryWilletts: @mark-prins @jmbrunskill The original request was to be able to backdate a precription when it's being created and it seems to be most logical to me that it's the date that affects the stock that is entered and displayed in that field. Doesn't really make sense to me to display anything else in there. Being able to backdate a prescription after it's been picked is fine as long as the rules about picking the stock available on that date are followed. So I guess setting it to the current date for a new prescription then only ever the picked date is what makes sense to me. I hope that answers the question you're asking! Probably pointless anyway since it's already merged and I think it agrees with the eventual conclusion 😄

jmbrunskill: Thanks @GaryWilletts you validated my thinking at least!

PR #5370, opened 2024-11-11
5100 sync items to legacy

### LINKED ISSUE #5100: Import GAPS data
"something extra I think we need to do for GAPS to be a success is to create a standard PQS vaccine item variant catalogue that can easily be loaded into the system at the central level and that has all the necessary data there ready for use in the calculations - e.g. volume per dose, temperature range etc.

I think it would be a tall order to expect the central level to gather all the information they need so would be good to smooth the path for them

There is a tab in the WHO IGA spreadsheet with all the information already there, hopefully there's a way we can it bundle along with the Open mSupply installer?"

Agree that this would be useful -  though there's no funding for the change. Raising an issue so that we can discuss with the triage team.

lache-melvin:
Fixes #5100

# 👩🏻‍💻 What does this PR do?

For adding all the GAPS data, it's much simpler if we can do that all from one script from OMS central. So adds the `to_sync_record` implementation for items, so new items can be created and synced to legacy from OMS central. 

Will leave with you @jmbrunskill, whether you think best to add data via DB script or if we want more code changes for this issues to try and add through the service layer + get the changelogs etc. for free?

# 🧪 Testing

- [ ] Create an item (in DB)
- [ ] Add changelog for that item
- [ ] Sync
- [ ] See item in Legacy mSupply

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/sync/translations/item.rs
+++ b/server/service/src/sync/translations/item.rs
@@ -1,6 +1,6 @@
 use repository::{
-    ChangelogRow, ItemRow, ItemRowDelete, ItemRowRepository, ItemType, StorageConnection,
-    SyncBufferRow, VENCategory,
+    ChangelogRow, ChangelogTableName, ItemRow, ItemRowDelete, ItemRowRepository, ItemType,
+    StorageConnection, SyncBufferRow, VENCategory,
 };
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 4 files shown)


jmbrunskill (approved):
Epic! Thanks!

server/service/src/sync/translations/item.rs
```diff
@@ -101,6 +120,60 @@ impl SyncTranslation for ItemTranslation {
...
+    fn change_log_type(&self) -> Option<ChangelogTableName> {
+        Some(ChangelogTableName::Item)
+    }
+
+    fn try_translate_to_upsert_sync_record(
+        &self,
+        connection: &StorageConnection,
+        changelog: &ChangelogRow,
+    ) -> Result<PushTranslateResult, anyhow::Error> {
+        let Some(item) = ItemRowRepository::new(connection).find_one_by_id(&changelog.record_id)?
```
jmbrunskill, line 133: Tempted to say we should put an `is_central_server` check in here, but I do know there's an idea of draft items coming that Craig's mentioned, so maybe should just mark anything as draft that's from remote?

PR #5373, opened 2024-11-11
5362 hide item variants if none configured

### LINKED ISSUE #5362: Hide Item Variant column if none are configured

GAPS feature includes `item variants`. As part of this, we needed to track what variant each batch we have on hand is, so we can calculate how much storage is used.

Therefore, there is a way to set item variant on all incoming stock:

Screenshot 2024-11-08 at 5 23 35 PM

However, most OMS deployments aren't using GAPS, and it's probably quite confusing to see this extra column, where the dropdown would always be empty.

Hide the column if no item variants are configured.

When rendering this column, first query and see if the `item_variants` table is empty. If so, just don't show the column.

Or maybe based on whether or not GAPS store properties are on?

Also not sure if it should be hidden per item (if the selected item doesn't have variants) or if _no_ items have variants?

lache-melvin:
Fixes #5362

# 👩🏻‍💻 What does this PR do?

Hides the item variant column on inbound modals/the stock detail page, if no item variants have been configured.

Make sure it feels okay to be hiding/showing base on there being _any_ variants set up, rather than for the item in question?

# 🧪 Testing

- [ ] Have no item variants configured (for any items)
  - [ ] There is no item variant column in:
    - [ ] Inbound shipment
    - [ ] Customer return
    - [ ] Stock detail view
- [ ] Configure an item variant
  - [ ] There is now an item variant column in:
    - [ ] Inbound shipment
    - [ ] Customer return
    - [ ] Stock detail view

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -26,7 +26,7 @@ import {
   ItemVariantInputCell,
   LocationRowFragment,
   PackSizeEntryCell,
-  useItemVariantsConfigured,
+  useIsItemVariantsEnabled,
 } from '@openmsupply-client/system';
 
 interface TableProps {
@@ -90,7 +90,7 @@ export const QuantityTableComponent: FC<TableProps> = ({
   isDisabled = false,
 }) => {
   const theme = useTheme();
-  const showItemVariantsColumn = useItemVariantsConfigured();
+  const showItemVariantsColumn = useIsItemVariantsEnabled();
 
   const columns = useColumns<DraftInboundLine>(
     [
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -137,7 +137,6 @@ export const QuantityTableComponent: FC<TableProps> = ({
     columnDefinitions,
   ]);
 
-  console.log('render');
   return (
     <DataTable
       id="inbound-line-quantity"
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -26,6 +26,7 @@ import {
   ItemVariantInputCell,
   LocationRowFragment,
   PackSizeEntryCell,
+  useItemVariantsConfigured,
 } from '@openmsupply-client/system';
 
 interface TableProps {
@@ -89,21 +90,28 @@ export const QuantityTableComponent: FC<TableProps> = ({
   isDisabled = false,
 }) => {
   const theme = useTheme();
+  const showItemVariantsColumn = useItemVariantsConfigured();
 
   const columns = useColumns<DraftInboundLine>(
     [
       getBatchColumn(updateDraftLine, theme),
       getExpiryColumn(updateDraftLine, theme),
-
-      {
-        key: 'itemVariantId',
-        label: 'label.item-variant',
-        width: 170,
-        Cell: props => (
-          <ItemVariantInputCell {...props} itemId={props.rowData.item.id} />
-        ),
-        setter: updateDraftLine,
-      },
+      ...((showItemVariantsColumn
+        ? [
+            {
+              key: 'itemVariantId',
+              label: 'label.item-variant',
+              width: 170,
+              Cell: props => (
+                <ItemVariantInputCell
+                  {...props}
+                  itemId={props.rowData.item.id}
+                />
+              ),
+              setter: updateDraftLine,
+            },
+          ]
+        : []) as ColumnDescription<DraftInboundLine>[]),
       [
         'numberOfPacks',
         {
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -90,53 +90,54 @@ export const QuantityTableComponent: FC<TableProps> = ({
   isDisabled = false,
 }) => {
   const theme = useTheme();
-  const showItemVariantsColumn = useIsItemVariantsEnabled();
+  const itemVariantsEnabled = useIsItemVariantsEnabled();
 
-  const columns = useColumns<DraftInboundLine>(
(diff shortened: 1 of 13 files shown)


client/packages/system/src/Item/api/hooks/useItemVariant/useItemVariantsConfigured.ts
```diff
@@ -0,0 +1,21 @@
...
+    queryKey: [ITEM_VARIANTS],
+    queryFn: async () => {
+      const result = await api.itemVariantsConfigured({
+        storeId,
+      });
+
+      return result.itemVariantsConfigured;
+    },
+    // Only call on page load
+    refetchOnMount: false,
```
lache-melvin, line 17: This means that if you set up an item variant, then immediately navigate to inbound shipment, the column won't be there... just need to refresh the page - I think this is okay? They are only configured on central server, and the no variants -> a variant exists change only happens once!

CarlosNZ:
> Make sure it feels okay to be hiding/showing base on there being _any_ variants set up, rather than for the item in question?

To me this seems fine, but I guess it depends on whether or not variants would be expected to be set up for most items -- I don't know the use case. But if only one or two items had variants amongst a large catalogue, I could see it being a bit annoying.

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -89,21 +90,28 @@ export const QuantityTableComponent: FC<TableProps> = ({
...
+              Cell: props => (
+                <ItemVariantInputCell
+                  {...props}
+                  itemId={props.rowData.item.id}
+                />
+              ),
+              setter: updateDraftLine,
+            },
+          ]
+        : []) as ColumnDescription<DraftInboundLine>[]),
```
CarlosNZ, line 114: I slightly prefer the `if (condition) then columns.push(...)` pattern for conditional columns, but that's just minor personal preference. Slightly easier to ready IMHO, but not worth changing.

client/packages/system/src/Item/api/hooks/useItemVariant/useItemVariantsConfigured.ts
```diff
@@ -0,0 +1,21 @@
+import { useQuery } from 'react-query';
+import { useItemGraphQL } from '../useItemApi';
+import { ITEM_VARIANTS } from '../../keys';
+
+export const useItemVariantsConfigured = () => {
```
CarlosNZ, line 5: Naming-wise, `useIsItemVariantsConfigured` is more consistent with other similar hooks -- or `useAreItemVariantsConfigured` if we want to be grammatically correct 🤷‍♂️

client/packages/requisitions/src/ResponseRequisition/api/operations.generated.ts
```diff
@@ -2,9 +2,9 @@ import * as Types from '@openmsupply-client/common';
 
 import { GraphQLClient, RequestOptions } from 'graphql-request';
 import gql from 'graphql-tag';
-import { ItemRowFragmentDoc } from 'packages/system/src/Item/api/operations.generated';
-import { NameRowFragmentDoc } from 'packages/system/src/Name/api/operations.generated';
-import { ReasonOptionRowFragmentDoc } from 'packages/system/src/ReasonOption/api/operations.generated';
+import { ItemRowFragmentDoc } from '../../../../system/src/Item/api/operations.generated';
+import { ReasonOptionRowFragmentDoc } from '../../../../system/src/ReasonOption/api/operations.generated';
+import { NameRowFragmentDoc } from '../../../../system/src/Name/api/operations.generated';
```
CarlosNZ, line 7: Is this a problem? Any idea why it's changed?

server/graphql/item_variant/src/lib.rs
```diff
@@ -1,8 +1,43 @@
...
+
+        let item_variants = service_provider
+            .item_service
+            .get_item_variants(&service_context, None, None, None)
+            .map_err(StandardGraphqlError::from_list_error)?;
+
+        Ok(item_variants.count > 0)
+    }
+}
+
```
CarlosNZ, line 40: Is this simpler than adding an `itemVariant` filter to the `items` node and just getting totalCount? That would seem cleaner that adding a whole top-level node just for this, but if this is less code, then makes sense.

client/packages/system/src/Item/api/hooks/useItemVariant/useItemVariantsConfigured.ts
CarlosNZ: Makes sense, but does that mean we don't really need a queryKey for this?

CarlosNZ (approved):
Nice, working well. Couple of tiny things, but basically nitpicks or questions, up to you what you want to address.

client/packages/system/src/Item/api/hooks/useItemVariant/useItemVariantsConfigured.ts
lache-melvin: Haha, probably not - that was an earlier thought, thanks

client/packages/requisitions/src/ResponseRequisition/api/operations.generated.ts
lache-melvin: I'm not sure, wondered if someone had tried to manually fix imports at some stage? I think it's usually been the `../..` way? Was going to keep an eye on it...

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -90,53 +90,54 @@ export const QuantityTableComponent: FC<TableProps> = ({
...
+    ]
   );
 
+  const columns = useColumns<DraftInboundLine>(columnDefinitions, {}, [
+    updateDraftLine,
+    lines,
+    columnDefinitions,
+  ]);
+
+  console.log('render');
```
CarlosNZ, line 140: `console...` 😉

client/packages/system/src/Item/api/hooks/useItemVariant/useIsItemVariantsEnabled.ts
```diff
@@ -1,11 +1,10 @@
 import { useQuery } from 'react-query';
 import { useItemGraphQL } from '../useItemApi';
-import { ITEM_VARIANTS } from '../../keys';
 
-export const useItemVariantsConfigured = () => {
+// Item variant inputs (i.e. in Inbound Shipment) should not be available if no item variants are configured
```
CarlosNZ, line 4:
```suggestion
// Item variant inputs (e.g. in Inbound Shipment) should not be available if no item variants are configured
```
pedantry 😉

CarlosNZ (approved):
Great, good catch on that bug.

There's a console.log left hanging around and one tiny comment nit, otherwise good to go :)

Issue #5266, opened 2024-10-30
On save, create prescription (picked status)


lache-melvin:
Hook in the with the `Ok` click of the JSON form modal, to create the prescription. Create it in `picked` status, so the stock movement is created, but more items could be added to the prescription if needed.

Ensure this only happens one (in future saves, prescription ID already exists, shouldn't create again)

CarlosNZ:
Couple of possible approaches I think of.

1. As a plugin:   
  From: https://github.com/msupply-foundation/open-msupply/blob/develop/client/packages/plugins/README.md 
    > Adding a button to a toolbar of a detail view which performs an external action like looking up shipping details from an external API. The plugin is provided with details of the object being viewed (e.g. draft shipment object) and can use that data when performing actions.
  
    This seems like a good case for such a button, which would supercede the default "OK/Save" button on any form where it's defined. We'd still need to specify the configuration in the form schemas somehow (i.e. define that a plugin should be used, and which one). It's be good to do this without changing the structure of either the form schema or the UI schema too much, but possibly an additional field `actions` on the UI schema would be okay.
2. As an extended functionality of the form component itself. Since the functionality will be associated with a particular form element (in this case, creating a prescription when an ItemSelector element is present), it kind of makes sense to define the behaviour with the form component logic. And this way, we wouldn't need to change the structure of the schemas at all -- this would just be configured by additional `options` on the form element UI schema definition. For this case we'd need to create a new component anyway ("Item Selector"), as the item selection logic is too specific for the general-purpose "Select" component. So we'd be adding a new concept to each form component -- an "action" that can be hooked up to the main form submission button. Then on submit, the JSON Forms module would check through the components used for any exposed actions, and if they are configured, execute them.

Having written this out, I'm leaning towards option 2, as it seems to make sense that certain actions would be fairly tightly coupled to specific form elements. Plus, the "scaffolding" is already in place -- the library of JSON Forms components plus the core form functionality, so I don't think exposing additional "action" functionality to existing components would be a big job.

CarlosNZ:
Regarding option 1 and option 2 -- I actually think that ultimately, most of the JSON Forms components should *be* plugins, as they're very case-specific.

The ones in the "common" folder are fairly generic, and can be kept in the core codebase, but I think it makes sense for the more specific ones to become plugins:
[...]

Issue #5264, opened 2024-10-30
Create JSON Form prescription renderer 


lache-melvin:
Create a new JSON forms renderer.

Input will be item_category_id. Item dropdown should be filtered by this category.

On item select, show available batches (much like the vaccination modal).

User should then be able to enter desired # packs to dispense. 

Will confirm - I think only need to be able to dispense from one batch here?

lache-melvin:
Closed by #5581

marthakeezy:
No specific testing steps -> tested as a part of #5676, moving to Done ✅

PR #5394, opened 2024-11-13
#5266 (Demo) Proof of concept for form submission actions

CarlosNZ:
Just a demo, not intended for merging.

# 👩🏻‍💻 What does this PR do?

This demonstrates how a JSON Forms component can be enabled to cause additional actions to occur when the form is saved. I've just added a very basic `console.log()` action on the Text component as an example. Obviously this will be removed and we'll use the mechanism for the Item Selector (Prescription renderer) component (#5264) when we get to it.

The basic mechanism is:
- an additional reference object on `useJsonForms` hook to put all the actions that will be run when the form is saved.
- a method "updateSubmitActions" that is passed down to individual components (via `config` prop) which allows them to "register" actions for saving. In most cases, this would be called with each component's `onChange` function, which would update the action with a closure over the current `data`, but this is up to the individual component to implement on a case-by-case basis
- in the `saveData` method (in this same hook, called whenever form is "saved"), we check the submit actions library, and iterate over the registered actions and execute them one by one
- conditional values can be included in the action by defining additional parameters on the component's `options` (configured in "uischema json") -- this is not implemented here, but it's just another option which is straightforward

To demonstrate here, go to a JSON form (e.g Patient details), change some "Text" input values. Then "Save" and check the console -- you should get message for each changed input with info about the data and the path.

The challenge here was to allow individual components to affect functionality in the top-level form manager (useJsonForms). A method passed down via the "config" object seemed the best way to achieve this without breaking the standard JsonForms implementation, but open to suggestions/refinements.


client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -121,6 +123,7 @@ export const useJsonForms = <R,>(
   }, [loadedData]);
   // current modified data
   const [data, setData] = useState<JsonData | undefined>();
+  const submitActions = useRef<Record<string, () => void>>({});
```
CarlosNZ, line 126: Using `useRef` rather than `useState` to reduce re-renders (the UI doesn't need updating in response to changes of this object)

client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -209,5 +223,7 @@ export const useJsonForms = <R,>(
     isDirty: isDirty ?? false,
     error,
     validationError,
+    submitActions,
+    updateSubmitActions,
```
CarlosNZ, line 227: Not actually doing anything with these return values currently, but could be handy for interacting with from outside. Not sure if will be useful, will remove if not necessary after #5264

CarlosNZ:
Tagging @andreievg if you have any comments regarding this new mechanism?

client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -129,6 +132,12 @@ export const useJsonForms = <R,>(
 
   useConfirmOnLeaving(isDirty);
 
+  const updateSubmitActions = useCallback((key: string, action: () => void) => {
```
lache-melvin, line 135: Wondering if `registerSubmitAction` would be clearer?

client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -140,6 +149,10 @@ export const useJsonForms = <R,>(
     try {
       const sanitizedData = stripEmptyAdditions(initialData, data);
       const result = await save?.(sanitizedData);
+      const actions = Object.values(submitActions.current);
+      for (const action of actions) {
+        await action();
+      }
```
lache-melvin, line 155: Seeing a possibility where we might want to be able to choose whether to run an action before or after the JSON form save, but I think this is fine for now. Probably wouldn't want to create the prescription if the JSON form submit gave an error.

Should be ok to create the prescription ID ahead of time to save on the JSON form (to look up prescription in future), then use that to create the prescription.

lache-melvin (approved):
Looks great, thank you for sharing PoC!

client/packages/programs/src/JsonForms/useJsonForms.tsx
CarlosNZ: Yup, was thinking something along those lines

client/packages/programs/src/JsonForms/useJsonForms.tsx
CarlosNZ: Yep, currently they get registered with just a key and an action. But I was envisaging a number of other properties that (when we need it), we could also register:
- pre/post save
- priority (a number, where higher numbers are executed first, similar to the "testers" priority in JSON forms)

These could be passed through via options, so could be configurable in the JSON UI schema. Will have to wait and see whether this becomes necessary, but fairly easy to add if we need it! :)

CarlosNZ: > Probably wouldn't want to create the prescription if the JSON form submit gave an error.

Similarly, we might need to save additional data to the JSON form on execution of the action, e.g. the ID of the prescription. I'll play around with this when doing the Prescription component.

andreievg (commented):
Just some minor comments, looking forward to seeing in real life

client/packages/programs/src/JsonForms/useJsonForms.tsx
CarlosNZ: image

That's one of the intended usages for useRef (along with direct DOM manipulation):

https://react.dev/reference/react/useRef

CarlosNZ:
Closing now -- this mechanism is now implemented in https://github.com/msupply-foundation/open-msupply/pull/5551, with a few improvements:
- put into its own hook containing all the logic
- can specify actions as either "pre" or "post" submit
- added some additional "state" into the hook for holding intermediate data (i.e. don't want to save it to JSON forms document, but need it for actions or component re-mounts)

PR #5405, opened 2024-11-14
change buffered state to draft

### LINKED ISSUE #4952: Message in automatic Customer Return doesn't update when status has been changed to Delivered
## What went wrong? 😲
Updating status to delivered in Customer Returns doesn't update the message in the toolbar. This doesn't block anything, but can potentially cause confusion if message contradicts what you can do 🤷🏻 

photo_2024-10-02 14 09 55

## Expected behaviour 🤔
Message should update.

## How to Reproduce 🔨

1. Create a Supplier Return to OMS supplier A
2. Set status to Shipped
3. Go to OMS supplier A store -> Customer Return
4. Click on newly generated return
5. Change status to Delivered
6. See that message still says you can't edit the details until status is changed to Delivered
-- Correct message below--
7. Refresh
8. See actual message

## Your environment 🌱

- Open mSupply Version: 2.3

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #4952 

# 👩🏻‍💻 What does this PR do?

Updated useCustomerReturn hook so any change in state would create a draft object for the CustomerReturnFragment. This is then patched in so the correct message banner will display when the status is changed from shipped to delivered

# 🧪 Testing

- [ ]  Create and ship a supplier return 
- [ ] Go to the supplier store
- [ ] Update the customer return status from Shipped to Delivered
- [ ] The message banner should update 

# 📃 Documentation

### DIFF
--- a/client/packages/invoices/src/Returns/CustomerDetailView/SidePanel/AdditionalInfoSection.tsx
+++ b/client/packages/invoices/src/Returns/CustomerDetailView/SidePanel/AdditionalInfoSection.tsx
@@ -18,13 +18,13 @@ export const AdditionalInfoSectionComponent = () => {
   const { debouncedMutateAsync } = useReturns.document.updateCustomerReturn();
   const isDisabled = useReturns.utils.customerIsDisabled();
 
-  const { bufferedState, setBufferedState } =
+  const { draft, setDraft } =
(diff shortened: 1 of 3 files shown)


lache-melvin (approved):
Epic!! Congrats on first PR 🎉

PR #5426, opened 2024-11-15
5383 Android - Load Plugins after they've be registered properly

### LINKED ISSUE #5383: Issues with Open mSupply v2.3.02 on Android 12
## What went wrong? 😲

When opening Open mSupply v2.3.02 on Android 12, we have a 'Loading' loop. When pressing the back button once, The orange man Open mSupply logo appears and is stuck on the screen. Pressing the back button one more time takes us to the server connection page, but shows no IP address in the bottom right-hand corner, and can't seem to choose client mode.

The apk was reinstalled and data was clearer, with the same issues occurring.

After upgrading the tablets to Android 13, the issue is now resolved.

image
image
photo_2024-11-12_16-03-32

## Expected behaviour 🤔

Hopefully to have Open mSupply working on Android 10+, or specific requirements updated so the support team can check if an upgrade is possible.

## How to Reproduce 🔨

1. New Samsung tablet with Android 12 installed
2. Install Open mSupply
3. See error
4. Upgrade to Android OS 13
5. See error is resolved

## Your environment 🌱

- Open mSupply Version: 2.3.02
- Legacy mSupply Central Server Version: N/A
- Platform: Android
- Database type: N/A

jmbrunskill:
Fixes #5383 

This behaviour with Android 11/12 isn't entirely consistent, this appears to be an improvement over the old method at least.

# 👩🏻‍💻 What does this PR do?

After much re-discovery of how things work and various attempts to fix similar issues.
E.g. https://github.com/msupply-foundation/open-msupply/pull/4390

I've changed the logic for the `ExtendedWebViewClient` to generate the javascript to inject on App Start and at page_load (as a backup method).

It seemed like in my emulators that the code that registers the plugins, and the web view constractor run concurrently so ordering of operations can't be guarenteed.

Here's what the `logcat` output looks like now.
```
2024-11-15 17:15:53.901  7282-7282  Capacitor               org.openmsupply.client               D  Starting BridgeActivity
2024-11-15 17:15:53.918  7282-7282  Capacitor               org.openmsupply.client               D  Registering plugin instance: CapacitorCookies
2024-11-15 17:15:53.919  7282-7282  Capacitor               org.openmsupply.client               D  Registering plugin instance: WebView
2024-11-15 17:15:53.920  7282-7282  Capacitor               org.openmsupply.client               D  Registering plugin instance: CapacitorHttp
2024-11-15 17:15:53.920  7282-7282  Capacitor               org.openmsupply.client               D  Registering plugin instance: NativeApi
```

Before it seemed to switch halfway through to the `generatePluginScript` code, and even putting in a thread sleep didn't work, I guess it's on the same thread, not familiar enough with how this works in Java?

# 🧪 Testing

- [ ] Build APK and try it on an android 10, 11, and/or 12 tablet, check you can initialize and then start the app again after

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/android/README.MD
+++ b/client/packages/android/README.MD
@@ -17,17 +17,17 @@ This requires a couple of steps.
 
 ![omSupply Android NDK](./doc/omSupply_android_ndk.png)
 
-Currently, we support two Android architectures;
+Currently, we support only 1 Android architecture (64bit);
 
 | rust target             | android ABI |
-| ----------------------- | ----------- |
 | aarch64-linux-android   | arm64-v8a   |
-| armv7-linux-androideabi | armeabi-v7a |
+| ----------------------- | ----------- |
+
 
 To add the required build targets to Rust run:
 
 ```bash
-rustup target add aarch64-linux-android && rustup target add armv7-linux-androideabi
+rustup target add aarch64-linux-android
 ```
 
 Next we have to tell Rust where to find the Android NDK binaries required to link the server library.
--- a/client/packages/android/README.MD
+++ b/client/packages/android/README.MD
@@ -12,7 +12,7 @@ The UI is not bundled within Capacitor, it is served by the server the client co
 Before building the Android app, the Rust server library that allows us to run the mSupply server on the mobile device needs to be built.
 This requires a couple of steps.
 
-- Install Android Studio
+- Install Android Studio (Ladybug at time of writing 15/11/24)
 - Install NDK (26.1.10909125 at the time of writing on the 8/4/24)
 
 ![omSupply Android NDK](./doc/omSupply_android_ndk.png)
--- a/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
+++ b/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
@@ -21,8 +21,13 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
 
     public ExtendedWebViewClient(Bridge bridge) {
         super(bridge);
-
         this.bridge = bridge;
+    }
+
+    public void loadJsInject() {
+        if(this.jsInject == null) {
+            this.jsInject = this.generatePluginScript();
+        }
     }
 
     // Have to manually inject Capacitor JS, this typically happens in
@@ -34,23 +39,13 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
         Logger.debug("onPageStarted" + url   );
         if (url.startsWith("data:text")) return;
 
-        if (this.jsInject == null){
-            this.jsInject = this.generatePluginScript();
-        }
+        // Just incase the js hasn't been generated yet, generate it here.
+        this.loadJsInject()
 
         if(this.jsInject != null) {
             Logger.debug("injecting JS");
             // .post to run on UI thread
             webView.post(() -> webView.evaluateJavascript(this.jsInject, null));
-        }
-    }
-
-
-    private void sleep(int delay) {
-        try {
-            Thread.sleep(delay);
-        } catch (InterruptedException e) {
-            throw new RuntimeException(e);
         }
     }
 
@@ -66,7 +61,7 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
             PluginHandle plugin = bridge.getPlugin(pluginName);
             if (plugin == null) {
                 Logger.error("Couldn't find plugin : " + pluginName);
-                continue;
+                return null;
             }
             pluginList.add(plugin);
         }
--- a/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
+++ b/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
@@ -71,7 +71,7 @@ public class ExtendedWebViewClient extends BridgeWebViewClient {
             String globalJS = JSExport.getGlobalJS(bridge.getContext(), bridge.getConfig().isLoggingEnabled(),
                     bridge.isDevMode());
             String bridgeJS = JSExport.getBridgeJS(bridge.getContext());
-            String pluginJS = JSExport.getPluginJS(pluginList);    
+            String pluginJS = JSExport.getPluginJS(pluginList);
             String cordovaJS = JSExport.getCordovaJS(bridge.getContext());
             String cordovaPluginsJS = JSExport.getCordovaPluginJS(bridge.getContext());
             String cordovaPluginsFileJS = JSExport.getCordovaPluginsFileJS(bridge.getContext());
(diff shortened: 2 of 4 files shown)


jmbrunskill:
NOTE: This is currently going to main, probably need to create a v2.3.3 branch?

client/packages/android/README.MD
```diff
@@ -12,22 +12,22 @@ The UI is not bundled within Capacitor, it is served by the server the client co
...
 This requires a couple of steps.
 
-- Install Android Studio
+- Install Android Studio (Ladybug at time of writing 15/11/24)
 - Install NDK (26.1.10909125 at the time of writing on the 8/4/24)
 
 ![omSupply Android NDK](./doc/omSupply_android_ndk.png)
 
-Currently, we support two Android architectures;
+Currently, we support only 1 Android architecture (64bit);
```
jmbrunskill, line 20: This change was actually made earlier, but we didn't update the docs...

client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
```diff
@@ -29,35 +31,52 @@ public ExtendedWebViewClient(Bridge bridge) {
...
-        );
+
+        // This function needs to run after plugins are registered, so can't be part of the constructor as order doesn't appear to be consistent.
+        List<String> pluginNames =  Arrays.asList("NativeApi","Keyboard", "WebView","BarcodeScanner","Preferences", "KeepAwake", "App", "Printer");
+        List<PluginHandle> pluginList = new ArrayList<>();
+        for (String pluginName : pluginNames) {
+            PluginHandle plugin = bridge.getPlugin(pluginName);
+            if (plugin == null) {
+                Logger.error("Couldn't find plugin : " + pluginName);
+                continue;
```
jmbrunskill, line 69: Not sure about this, Maybe better to keep the old behaviour and crash the app, however would be good if we could somehow present a better error state if this happens again? The `crash` state just shows as loading... Maybe could inject a substitute JS that displays an error?

client/packages/android/app/src/main/java/org/openmsupply/client/RemoteServer.java
```diff
@@ -12,6 +14,7 @@ public RemoteServer() {
     }
 
     public void start(int port, String filesDir, String cacheDir, String androidId) {
+        Logger.info("Starting OMS Rust Server");
```
jmbrunskill, line 17: At one stage It seemed like OMS wasn't starting, would be nice to get some kind of feedback in `logcat` about OMS server status. Maybe an R&D or backlog issue?

andreievg:
> I've changed the logic for the ExtendedWebViewClient to generate the javascript to inject, only when it's first needed.

We've had issues where the javascript inject generation would take longer then initial page load and the JS wouldn't recognise that the plugs are there.

I actually was playing around with something on the weekend, and did this approach (worth trying):

https://github.com/msupply-foundation/open-msupply/compare/4702-add-report-sorting-via-wasm-functions...4702-add-report-sorting-via-wasm-functions(Load-js-injecton-start)

client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
andreievg: Plugin missing would also cause an error down in the catch statement, was wondering if it's important to know what plugin hasn't been loaded ? 

With above we would get a logic/silent error which are harder to trace right ? (like Android wouldn't know that it's android ? or some other functionality would fail further up the track)

jmbrunskill: I agree a silent error isn't helpful! 
If everything is working right we shouldn't have any plugins missing.
This useful for me to find that sometimes plugins were missing, but not all of them.
I think we should probably either try to display a more useful error like `Plugin XYZ` missing rather than infinite loading?

andreievg:
https://github.com/msupply-foundation/open-msupply/blob/64a166bcefa1ca99770025afd44781758afcb43e/client/packages/android/app/src/main/java/org/openmsupply/client/MainActivity.java#L15-L16

NativeApi.class is registered before onCreate method for BridgeActivity

Which would add a NativePlugin before initialPlugs are added:

https://github.com/ionic-team/capacitor/blob/8b01b2bbb3c11d484952d904c7398939fb3da395/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java#L54-L57

Notice how the plugin is loaded

https://github.com/ionic-team/capacitor/blob/8b01b2bbb3c11d484952d904c7398939fb3da395/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L224-L225

https://github.com/ionic-team/capacitor/blob/8b01b2bbb3c11d484952d904c7398939fb3da395/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L637-L640

https://github.com/ionic-team/capacitor/blob/8b01b2bbb3c11d484952d904c7398939fb3da395/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L678-L683

You can see that it's put into into plugin array after PluginHandler is created which also initialises the plugin which would call 

https://github.com/ionic-team/capacitor/blob/8b01b2bbb3c11d484952d904c7398939fb3da395/android/capacitor/src/main/java/com/getcapacitor/PluginHandle.java#L104

And this would also try to generate JS bundle, before all plugins (even this plugin) is available in bridge (due to the above)

https://github.com/msupply-foundation/open-msupply/blob/64a166bcefa1ca99770025afd44781758afcb43e/client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java#L23

jmbrunskill:
Added this message if the javascript isn't available to inject. Which hopefully helps the next person who encounters a similar problem....
image

jmbrunskill:
Pushed changes based on Andrei's branch but still generating in onPageStarted if it's not there, just incase. Pretty sure @andreievg 's got it right though. Thanks for the Links and work @andreievg !

jmbrunskill:
https://drive.google.com/file/d/1czk09-I5O_Q9VbRoCIVewpCJeHvlS8bm/view?usp=sharing - APK

client/packages/android/app/src/main/java/org/openmsupply/client/ExtendedWebViewClient.java
```diff
@@ -31,26 +39,35 @@ public ExtendedWebViewClient(Bridge bridge) {
...
+        // Just incase the js hasn't been generated yet, generate it here.
+        this.loadJsInject();
+
         if(this.jsInject != null) {
+            Logger.debug("injecting JS");
             // .post to run on UI thread
             webView.post(() -> webView.evaluateJavascript(this.jsInject, null));
+        } else {
+            Logger.error("JS not generated, not injecting");
+            webView.post(() -> webView.evaluateJavascript("alert('Error unable to load javascript to inject. Please contact mSupply Support for assistance.')", null));
```
andreievg, line 51: great !

andreievg (approved):
Looks great ! Thanks

PR #5428, opened 2024-11-15
5331 translate item categories

### LINKED ISSUE #5331: Translate item categories
Item categories exist in OG mSupply, we need to translate these into our new categories schema.

Should be translated to known id (mSupply Categorisation or similar) as the category group, then the categories translated into the category table.

lache-melvin:
Fixes #5331

# 👩🏻‍💻 What does this PR do?

Not tested yet, just sharing what I have :)

- Removes changelog creation from category tables (as syncing from OG now)
- Translates 3 levels of the `category` table, from `item_category`, `item_category_level1` and  `item_category_level2`
- Translates item_category_join from item.category_ID

# 🧪 Testing

- [ ] Create item categories (in OG: Item > Item categories > Show Category 1)
- [ ] Assign a category to an item
- [ ] Sync to OMS
- [ ] No translation errors
- [ ] Should be category entries in the category table
- [ ] Should be a join table entry for the item
- [ ] Change the category for an item, sync again
- [ ] old join should have deleted_datetime, new join entry should exist

# 📃 Documentation

  1.
  2.


server/repository/src/db_diesel/category_row.rs
```diff
@@ -73,17 +54,20 @@ impl<'a> CategoryRowRepository<'a> {
...
         Ok(())
     }
 }
 
 impl Upsert for CategoryRow {
     fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError> {
-        let change_log_id = CategoryRowRepository::new(con).upsert_one(self)?;
-        Ok(Some(change_log_id))
+        CategoryRowRepository::new(con).upsert_one(self)?;
+        // Not in changelog
```
jmbrunskill, line 69: I'm keen to @andreievg's thoughts on this.
I think it should be safer to keep it out of changelog for now, but when we do do the central OMS things we do want it in change log. Maybe we'll do something like just translate the legacy records on OMS Central, but keep the changelog entries?

server/service/src/sync/translations/item.rs
```diff
@@ -173,6 +221,10 @@ impl SyncTranslation for ItemTranslation {
             strength,
             type_of: to_legacy_item_type(r#type),
             VEN_category: to_legacy_ven_category(ven_category),
+            // Item push is only used for GAPS, which doesn't use item categories
+            // Probably better to move management of categories to OMS Central than
+            // build out the syncing back and forth of categories to OG!
+            category_ID: None,
```
jmbrunskill, line 227: agreed, good comment!

server/service/src/sync/translations/category.rs
```diff
@@ -0,0 +1,135 @@
...
+            _ => false,
+        }
+    }
+
+    fn try_translate_from_upsert_sync_record(
+        &self,
+        _: &StorageConnection,
+        sync_record: &SyncBufferRow,
+    ) -> Result<PullTranslateResult, anyhow::Error> {
+        // Try to parse as a child category (item_category or item_category_level2)
```
jmbrunskill, line 64: I feel like this probably, can match both?

If you put `#[serde(default)]` on parent_id, you should be able to just use one legacy type?
Am I right the only difference is the existence of the parent_id?

jmbrunskill (approved):
Looks good, I have a suggestion for the legacy types.
I'm assuming you're testing but if you want me to test too I can :)

server/service/src/sync/translations/category.rs
lache-melvin: Oo nice spot, thank you!

lache-melvin:
Pushed a few changes if you're up for a quick re-review @jmbrunskill!

jmbrunskill (changes requested):
Looks like it's only translating catergories not the levels for me?

server/repository/src/db_diesel/item_category_row.rs
```diff
@@ -33,55 +31,45 @@ impl<'a> ItemCategoryRowRepository<'a> {
         ItemCategoryRowRepository { connection }
     }
 
-    pub fn upsert_one(&self, item_category_row: &ItemCategoryRow) -> Result<i64, RepositoryError> {
-        diesel::insert_into(item_category::table)
-            .values(item_category_row)
-            .on_conflict(item_category::id)
+    pub fn upsert_one(
+        &self,
+        item_category_join_row: &ItemCategoryRow,
```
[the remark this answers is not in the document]
lache-melvin: Heh probably 😁

lache-melvin:
> Looks like it's only translating catergories not the levels for me?

Found the culprit, filtering sync buffer rows by the translator `table_name`: https://github.com/msupply-foundation/open-msupply/blob/5331-translate-item-categories/server/service/src/sync/sync_buffer.rs#L44

What do you think of this solution? [af76dc4](https://github.com/msupply-foundation/open-msupply/pull/5428/commits/af76dc4086dee1d80facda95ac612573c48f7771)

jmbrunskill (approved):
Looks good to me,

I'm not 100% happy with the `additional_legacy_table_names()` solution,
I tried this but not sure if it's actually nicer or not as it would be possible to have a translator that doesn't have any table_name defined, but I think that's unlikely as we'll usually copy and paste and create tests?

https://github.com/msupply-foundation/open-msupply/compare/5331-translate-item-categories...5331-translate_time_categories-suggestion

What do you think of this?
Feel free to merge either version...

andreievg:
The least amount of changes to the existing traits would be best. So leaning towards @jmbrunskill -> however would need to be careful with pull dependencies for other translators.

I guess there is not reference constraints so should be safe in terms of deps between category and level.

I would add a refactor to remove table_name and use table_names everywhere (including dependencies).

Alternative is to create 3 translators that re-use the same functionality.

server/repository/src/db_diesel/category_row.rs
andreievg: I think since we would have source_id we should be able to have those in changelog (on remote on OMS central) regardless of where they come from, and then only sync to remote from OMS central if source_id = current site. I hope i understood the question correctly

PR #5434, opened 2024-11-17
5372 unit grouping symbol in stocktake difference

### LINKED ISSUE #5372: No unit grouping symbol in the Difference column of stocktakes
## What went wrong? 😲

All columns in the stocktake window (and all other windows I saw) use a digit grouping symbol based, I believe, on the Regional format you have. The Difference column doesn't seem to use any digit grouping symbol. I think that it should to ensure consistency in how we display numbers throughout omSupply.  

## Expected behaviour 🤔
The Stocktake Difference column should use a digit grouping symbol based on the user regional settings. 

## How to Reproduce 🔨

1. Go to Inventory > Stocktakes
2. Add a new stocktacke ("New Stocktake" button on the top right of your screen) 
3. Add a new item ("New Item" button on the top menu)  or open an already listed item
4. In the # of counted pack field, add a number equal or greater than a thousand than the Snapshot Packs number . Click OK
6. See error. The Number of counted packs has a digit grouping symbol but the number of the Difference Column don't 

## Your environment 🌱

- Open mSupply Version: 2.4.0test-1
- Legacy mSupply Central Server Version: v19-0RC28

- Platform: openmSupply desktop or Browser version

- Database type: SQLlite

aimee-mcneil-melville:
Fixes #5372 

# 👩🏻‍💻 What does this PR do?

- Add unit grouping to the 'difference' column on stocktake detail view via the NumberCell type
- Included that if the counted column is null (`-`) that difference would also appear as `-`, whereas previously the difference would appear as '0'

This creates consistency with other fields which are also formatted this way

Screenshot 2024-11-18 at 9 46 03 AM

# 🧪 Testing

- [ ] Go to Inventory > Stocktakes
- [ ] Add a new stocktacke ("New Stocktake" button on the top right of your screen)
- [ ] Add a new item ("New Item" button on the top menu) or open an already listed item - one of the stocklines will need at least 1,000 units available
- [ ] In the # of counted pack field, add a number for each of:

-  equal or greater than a thousand than the Snapshot Packs number 

-  equal or greater than a thousand *less* than the Snapshot Packs number 

-  equal to the Snapshot Packs number

- blank - not counted

- [ ] Click OK. See that the unit grouping and `-` values in the difference column are consistent with the counted column

# 📃 Documentation


node_modules/.yarn-integrity
lache-melvin, line 1: We should remove this file before merging... I think it might have been generated by running a yarn command from the root rather than the `client` directory?

client/packages/inventory/src/Stocktake/DetailView/ContentArea/columns.tsx
```diff
@@ -206,28 +206,32 @@ export const useStocktakeColumns = ({
...
-          return (total < 0 ? Math.abs(total) : -total).toString();
-        } else {
-          return (
-            (rowData.countedNumberOfPacks ?? rowData.snapshotNumberOfPacks) -
-            rowData.snapshotNumberOfPacks
-          );
-        }
-      },
-    },
+                const total =
```
lache-melvin, line 217: Looks like more formatting noise... we should check out your config after lunch!

lache-melvin (changes requested):
Looks great Aimee! I'm wondering if we add the same NumberCell in the stocktake modal as well, I'm seeing the same issue for the snapshot packs:

Screenshot 2024-11-18 at 1 34 35 PM

Did you get to ask someone about what it's doing? Otherwise I am free whenever to chat!

aimee-mcneil-melville:
Screenshot 2024-11-18 at 4 11 18 PM

Great, I have fixed those and also noticed the Pack Size in list view didn't have the number type, so have added that too.

Screenshot is pre this change, but also the column header formatting looks off?

node_modules/.yarn-integrity
aimee-mcneil-melville: Thanks, I did do this - now I know why it happened!

lache-melvin, line 1: Huh, looks like this file is still here, but empty now? Maybe try deleting the `node_modules` folder (the one you want is within the `client` folder 😁

lache-melvin:
> Screenshot 2024-11-18 at 4 11 18 PM
> 
> Screenshot is pre this change, but also the column header formatting looks off?

Do you mean that `Pack Size` is wrapped, but `Snapshot Packs` in all one line?

Yeah it's a little funky, but I think fine to leave for now. The snapshot packs column doesn't have a width set, so on your wider screen it is expanding to fill the space. You'll notice if you make the screen narrower, it will wrap. The pack size column _does_ have a width, but it's a bit hard to find...

The `'packSize'` string on line 143 here:

https://github.com/msupply-foundation/open-msupply/blob/23b44e0e97eee9083253cdf7507f7882e90c0629/client/packages/inventory/src/Stocktake/DetailView/ContentArea/columns.tsx#L142-L152

maps to this "base" pack size column definition:

https://github.com/msupply-foundation/open-msupply/blob/23b44e0e97eee9083253cdf7507f7882e90c0629/client/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts#L186-L191

125px is a bit too narrow, so that header always wraps 🤷‍♀️  We added the fixed widths because we were having issues on small screens of the headers getting cut off, but clearly haven't got them all! Seeing as pack size is usually a fairly small number, we'd rather the header wraps, and we get a bit more width for other columns 😁

lache-melvin (approved):
Looks great to me Aimee! Just remove that node_modules folder and merge when ready 🙏

PR #5436, opened 2024-11-17
new util function to display age

### LINKED ISSUE #5218: Age is showing as 0 for patients younger than 1 year old
## What went wrong? 😲

When clicking on a specific patient, if they are < 1 year old, the age display is inconsistent (e.g. one area shows actual age, another area shows 0)

Screenshot 2024-10-24 at 12 47 07 PM

Screenshot 2024-10-24 at 12 50 49 PM

## Expected behaviour 🤔

Would be nice to display months and days at the top and not just 0

## How to Reproduce 🔨

2. Create a new patient and set the date of birth to a date within the last year
3. Finalise the creation process (e.g. add code etc)
4. At the end, see the top showing 0 and the middle showing the actual age, just like on the screenshots

## Your environment 🌱

- Open mSupply Version: `V2.3.1-RC1`
- Legacy mSupply Central Server Version: `V7.18.06`

- Platform: `Browser (Google)`

- Database type: `SQLite3`

noel-yeldos:
new util function to display age below one year in months and days. TODO finishing updating tests

Fixes #5218 

# 👩🏻‍💻 What does this PR do?
Fix the issue in the patient toolbar where age for patients under 1 year old was displaying as 0. Age now displays in months and days.
 

- Created a new util function file in the same folder as patientSummary.tsx and updated the function being called on line 44 of patientSummary

To stop the age displaying as 0 and be consistent with date picker age

Screenshot 2024-11-18 at 11 01 30 AM

# 🧪 Testing

- [ ] Open patients tab
- [ ] Create or edit patient
- [ ] Add DOB
- [ ] Check DOB for more than 1 year old, less than 1 year old, less than 1 month old
- [ ] Check that dates display in days and months

# 📃 Documentation

  1.
  2.


CarlosNZ:
Thanks @noel-yeldos! We actually already have a method that does this in `DateUtils` called `ageInMonthsAndDays`. 

So you should be able to just use that in the Patient Summary toolbar. I think it should work the same way, let me know if there's any differences you think worth taking into account.

Tests are great though, so please add them to `DateUtils.test.ts`! Thanks :)

noel-yeldos:
@CarlosNZ the ```getDisplayAge``` util function uses ```ageInMonthsAndDays``` but also has logic to display the age with the appropriate labels when displaying ages less than 1 year. At the moment, I've left the function file and test file in the ```PatientView``` folder. Thanks!

lache-melvin:
@CarlosNZ let us know if you think this is generic enough for DateUtils? Only used in patient as yet but maybe nice to expose as reusable?

CarlosNZ (changes requested):
Looking good. I think the `getDisplayAge` function can go in the `useFormatDateTime` hook, which is in the `DateUtils.ts` file. It's a collection of date/time formatting methods, which this fits into nicely. The main difference between this hook and the core `DateUtils` methods is that it uses the localisation hook, so it must be a hook itself.

The advantage of this hook is that you won't need to pass the "translate" method (`t()`) into the function, as it's already available in the hook scope.

Also, can you replace the formatting in the DOB component (`DateOfBirth.tsx`) with this new function to keep them consistent? Thanks.

client/packages/system/src/Patient/PatientView/getDisplayAge.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (age >= 1) {
+    return String(age);
+  } else if (months >= 1) {
+    const ageInMonthsAndDays =
+      t('label.age-months-and', { count: months }) +
+      t('label.age-days', { count: days });
+    return ageInMonthsAndDays;
+  } else {
+    return t('label.age-days', { count: days });
+  }
```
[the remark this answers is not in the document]
noel-yeldos: Done!

noel-yeldos:
@CarlosNZ I've made the requested changes. Ready for review again. Thanks!

cc - Thanks @lache-melvin for helping me figure out testing!

client/packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -250,6 +251,16 @@ export const useFormatDateTime = () => {
...
+  const getDisplayAge = (dob: Date): string => {
+    const patientAge = DateUtils.age(dob);
+    const { months, days } = DateUtils.ageInMonthsAndDays(dob ?? '');
+
+    if (patientAge >= 1) {
+      return String(patientAge);
+    } else
+      return `${months > 0 ? t('label.age-months-and', { count: months }) : ''}${t('label.age-days', { count: days })}`;
+  };
+
```
CarlosNZ, line 263: Nice, might be good to have a comment above the function with an explanation/example, something like:
```
  // Returns a formatted age for the input date (of birth) relative to the
  // current date. Will format as whole number of years unless the age is less
  // than one year old, in which case it will format as months/days:
  // e.g:
  // - DOB is 2 years in the past => "2"
  // - DOB is 9 months and 2 days ago => "9 months, 2 days"
```

CarlosNZ (approved):
Sweet. Just one small improvement (and the comment for the function):

Can you just update the "DateOfBirth" component to use this method? Line 117-120 will become:
```tsx
            <Typography fontSize="85%" whiteSpace="nowrap">
              {formatDateTime.getDisplayAge(dob)}
            </Typography>
```
(And then you can remove the `DateUtils.ageInMonthsAndDays` on line 77)

It'll give you a type error. I'd update the `getDisplayAge` method to accept null/undefined inputs and return empty string:
```tsx
  const getDisplayAge = (dob: Date | null | undefined): string => {
    if (!dob) return '';
// etc...
```
(And maybe add one more test case for this case, but not too important)

noel-yeldos:
@CarlosNZ thanks for that! Ready for review again

client/packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -251,14 +251,23 @@ export const useFormatDateTime = () => {
...
     } else
       return `${months > 0 ? t('label.age-months-and', { count: months }) : ''}${t('label.age-days', { count: days })}`;
+
+    // Returns a formatted age for the input date (of birth) relative to the
+    // current date. Will format as whole number of years unless the age is less
+    // than one year old, in which case it will format as months/days:
+    // e.g:
+    // - DOB is 2 years in the past => "2"
+    // - DOB is 9 months and 2 days ago => "9 months, 2 days"
+    // - DOB is 5 days ago => "5 days"
```
CarlosNZ, line 270: Oh, function definition comments should go above the function:
```ts
// Returns a formatted age for the input date (of birth) relative to the
// current date. Will format as whole number of years unless the age is less
// than one year old, in which case it will format as months/days:
// e.g:
// - DOB is 2 years in the past => "2"
// - DOB is 9 months and 2 days ago => "9 months, 2 days"
// - DOB is 5 days ago => "5 days"
const getDisplayAge = (dob: Date | null | undefined): string => {
  if (!dob) return '';
```

client/packages/common/src/intl/utils/DateUtils.test.ts
```diff
@@ -36,4 +36,10 @@ describe('getDisplayAge', () => {
     const result = getDisplayAge(dob);
     expect(result).toBe('10 days');
   });
+
+  it('returns an empty string if dob is not defined', () => {
+    const dob = null;
+    const result = getDisplayAge(dob);
+    expect(result).toBe('');
+  });
```
CarlosNZ, line 44: In simple cases like this I think it's fine to not bother with the intermediate variables, particularly since the description tells you what it's doing. Just this would be okay:
```ts
expect(getDisplayAge(null)).toBe('')
```

It's fine though, no need to change it in this case, just a general comment.

CarlosNZ (approved):
Great, thanks @noel-yeldos . If you could just move the comment above the function, then merge when you're ready (no need for me to look again).

Thanks for your updates :)

PR #5441, opened 2024-11-17
4097 browser back button behaviour

### LINKED ISSUE #4907: Browser back button behaviour
## What went wrong? 😲

With the current workflow
* I'm on page (x)
* I navigate to the Internal orders list
* Click on 'new order'
* the newly created internal order loads
* click the back button in the browser
* you're on page (x)

## Expected behaviour 🤔
You should be back on the internal order list
Same problem with 
* Inbound shipment
* Outbound shipment
* Prescription
* Supplier return

## How to Reproduce 🔨

[ see above ]

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #4907 

# 👩🏻‍💻 What does this PR do?

Corrects navigation behavior using the browser back button. From the detail view of a page, the browser back button will now navigate to the list view of the same page. Changes made for the following pages:

- Internal Orders
- Requisitions
- Inbound Shipment
- Outbound Shipment
- Prescriptions
- Supplier Return

When using the browser back button from a detail view, navigation would go to the list view of a previously viewed/unrelated page, instead of the most recent list used to access the detail view.

Please check the navigation and URL for Supplier Returns as this was a new write.

# 🧪 Testing

Test each of the above pages, steps are for Internal Orders

- [ ] Start from any page _other than_ the one being tested
- [ ] Navigate to the list view of the page being tested
- [ ]  Click on 'new order' or equivalent, complete necessary selections, click OK
- [ ] The new order should be created and in detail view
- [ ] Click the back button in the browser
- [ ] The page should be in the list view for the tested page

# 📃 Documentation


client/packages/invoices/src/Returns/SupplierListView/AppBarButtons.tsx
```diff
@@ -61,6 +64,14 @@ export const AppBarButtonsComponent: FC<{
...
               supplierId: name?.id,
               supplierReturnLines: [],
+            }).then(invoiceNumber => {
+              navigate(
+                RouteBuilder.create(AppRoute.Replenishment)
+                  .addPart(AppRoute.SupplierReturn)
+                  .addPart(String(invoiceNumber))
+                  .build(),
+                { replace: false }
+              );
```
lache-melvin, line 74: Did you have a look into why this one didn't already have a navigate call here?

It's actually still using an old pattern, [calling the `navigate` from the API hook](https://github.com/msupply-foundation/open-msupply/blob/4097-browser-back-button-behaviour/client/packages/invoices/src/Returns/api/hooks/document/useInsertSupplierReturn.ts#L20)!

I think this is as good a time as any to update that! We can keep your navigate call here, and remove the one in the API hook. Currently the back function is getting a bit confused with both calls 😅

aimee-mcneil-melville: Ah, yes this makes sense now that I know where to look!

Customer Return has the navigation through the API as well. I'd noticed that it didn't have navigation in AppBarButtons.tsx, however it did behave as expected. 
Would this be one to update now while I'm in it?

lache-melvin: Yes please! Great spotting 😁

client/packages/invoices/src/InboundShipment/ListView/AppBarButtons.tsx
```diff
@@ -78,7 +78,7 @@ export const AppBarButtons: FC<{
                 .addPart(AppRoute.InboundShipment)
                 .addPart(String(invoiceNumber))
                 .build(),
-              { replace: true }
+              { replace: false }
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Can I confirm the expected behaviour for the requisition detail ones. I'm thinking of this in terms of breadcrumbs back, rather than back being 'whichever view it was on before'

- Requisition Detail -> add item/click item row (input page opens) -> back button will go back to Requisition Detail. Back once more is to Requisition List

- Requisition Detail -> add item/click item row (input page opens) -> close button takes you back to Requisition Detail.
-> if now clicking the back button I would expect it to take me to the Requisition List (not back to the input page)
Similar thing for Next/Previous, these should go back to the the Detail view, and then back to List view?

node_modules/.yarn-integrity
lache-melvin, line 1: haha this is back!

lache-melvin (approved):
Looks great Aimee, thank you! All good to merge once that .yarn-integrity file is removed again 😅

PR #5450, opened 2024-11-18
5100 import gaps data script

### LINKED ISSUE #5100: Import GAPS data

jmbrunskill:
Fixes #5100

# 👩🏻‍💻 What does this PR do?

Creates...

1. Master List for GAPS to provide item visibilty (will only exist in OMS Central)
2. `items`
3. `item_link` - Not synced should it be?
4. `item_variant` (for item variant and diluent variant)
5. 3 `packaging variant`s for each `item_variant`
6. If diluent is not bundled
7. Create diluent
8. Create diluent variant
9. Create item_bundle
10. Create diluent packaging variants (x3)

Have I missed something?

Sorry for the messy code and writing in python, seemed like the fastest way to get something working.

There's probably a bit of data clean up or filtering to add.
For example if there a 0 or no `volume_per_unit` we're just loading that in, but maybe should exclude those items till we have the actual data for them???

Any other data cleanups we need to consider?

# 🧪 Testing

- [ ] Run the gaps_Inserts.sql script in your OMS Central Postgres Server
- [ ] Check that the item_variant, bundles, etc line up with the data here: https://docs.google.com/spreadsheets/d/1H_5_GMNdSN6_eLlRqZ7z4_3gx9spShwvb1NtLBrMEjQ
- [ ] It won't be possible to check everything but spot checking random things is a good idea...

# 📃 Documentation

  1.
  12.


server/gaps/asset_catalogue/migration.sql
```diff
@@ -0,0 +1,179 @@
+INSERT INTO asset_class (id, name) VALUES ('fad280b6-8384-41af-84cf-c7b6b4526ef0', 'Cold chain equipment');
```
jmbrunskill, line 1: I added this too, as I believe we'll need to update this asset list soon...

server/gaps/gaps_item_variants.csv
```diff
@@ -0,0 +1,315 @@
+Item Universal Code,PQSVaccineID,VaccineType,VaccineTypeName,Manufacturer,CommercialName,PharmaceuticalForm,Presentation,DosesCount,ShelfLife,VaccinePreservative,AdministrationRoute,VaccineStorageTemperature,PrequalificationStatus,VaccineVialMonitor,MultidoseVialPolicy,VaccinePrimaryVolume,VaccineSecondaryVolume,VaccineTertiaryVolume,DiluentBundled,DiluentPrimaryVolume,DiluentSecondaryVolume,DiluentTertiaryVolume,FreezeSensitive,DiluentStorageTemperature,AlAdjuvant,Nbre doses dans calendrier,Prix du flacon ($/flacon),UNICEF PS
```
jmbrunskill, line 1: Exported from https://docs.google.com/spreadsheets/d/1H_5_GMNdSN6_eLlRqZ7z4_3gx9spShwvb1NtLBrMEjQ

server/gaps/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+        return '(SELECT id FROM cold_storage_type WHERE name = \'-20\' or min_temperature = -20 and max_temperature = -20 LIMIT 1)'
+    elif storage_temp == '-70°C':
+        return '(SELECT id FROM cold_storage_type WHERE name = \'-70\' or min_temperature = -70 and max_temperature = -70 LIMIT 1)'
+    else:
+        return 'null'
+
+
+def create_master_list():
+    print("INSERT INTO master_list (id, name, code, description, is_active) VALUES ('43491ce9-bc89-4ee5-988d-9dbb2856e398', 'gaps_items', 'gaps_items', 'All Items automatically create for Cold Chain Gap Analysis', true) ON CONFLICT DO NOTHING;")
+    print("INSERT INTO master_list_name_join (id, master_list_id, name_link_id) SELECT uuid_in(md5(random()::text || random()::text)::cstring), '43491ce9-bc89-4ee5-988d-9dbb2856e398', id FROM name_link WHERE id NOT IN (select name_link_id from master_list_name_join WHERE master_list_id = '43491ce9-bc89-4ee5-988d-9dbb2856e398');")
```
jmbrunskill, line 19: Should make the master list available to all names in the current sync site.
Make sure you sync the site first before running this script.
Seems unlikely that you're not have synced but still...

server/gaps/gaps_insert_functions.py
```diff
@@ -0,0 +1,167 @@
...
+    # nextval('changelog_cursor_seq'::regclass)
+    return f"INSERT INTO changelog (record_id, table_name, row_action) VALUES ('{record_id}', '{table_name}', 'UPSERT');"
+
+
+def item_link_stmt(item_id):
+    return f"INSERT INTO item_link (id, item_id) VALUES ('{item_id}', '{item_id}') ON CONFLICT DO NOTHING;"
+
+def upsert_item_stmt(item_id, row):
+    item_name = row['VaccineType']
+    item_code = row['VaccineTypeName']
```
lache-melvin, line 16: Typename might need to be included in item name, getting this with covid vaccines and rotavirus (codes are all different but this looks not great...) 
Screenshot 2024-11-19 at 11 00 46 AM

server/gaps/gaps_insert_functions.py
```diff
@@ -0,0 +1,167 @@
...
+
+    #Insert the changelog
+    print(changelog_stmt(item_id, 'item'))
+
+    return insert_statement
+
+def upsert_item_variant_stmt(item_variant_id, item_id, row):
+    item_variant_name = row['VaccineTypeName'] + " " + row['CommercialName']
+    if row['VaccineTypeName'] == row['CommercialName']:
+        item_variant_name = row['VaccineTypeName']
```
lache-melvin, line 61: NIT, we can improve later... but maybe a bit much duplication 😁 

Screenshot 2024-11-19 at 11 11 14 AM

lache-melvin (changes requested):
Script doesn't work on SQLite... are we safe to assume central server would always be running PG (probably right?)

I think I would benefit from a little more content in the readme 😁 

Perhaps we can put the .sql files in a `generated` folder? I don't think I mind committing them, nice not to need to run the generation scripts first, but good to be clear about what to do if you need to change any of the data.

> item_link - Not synced should it be?

No all good, links get created when items are inserted - once new items sync to other sites they'll create their own links 👍

lache-melvin:
Oh forgot to mention in review - business logic prevents duplicate item variant names, but this script is creating duplicates. I guess they do have duplicate names in the catalogue, but we need to make a decision either to include something unique in the variant names here, or to show more info in the item variant select dropdowns (currently would be no way to tell which variant you are bundling etc?)

jmbrunskill:
@adamdewey do you want to try to run the SQL inserts here on you Open mSupply Central Server? https://github.com/msupply-foundation/open-msupply/blob/13c5d3d445a4e90b609506d186bd843db432e39b/server/gaps/items/generated/gaps_item_inserts.sql

Have a look to see if you see any issues with the translation of the PQS Vaccine data.

jmbrunskill:
Just found a bug with this, running multiple times adds the items multiple times to the masterlist which creates some un-intended effects, we should delete the master list association on each run or check if the item exists on the master list before inserting...

jmbrunskill:
> Just found a bug with this, running multiple times adds the items multiple times to the masterlist which creates some un-intended effects, we should delete the master list association on each run or check if the item exists on the master list before inserting...

Fixed...

lache-melvin (changes requested):
Looking good thanks James! Main request is to add the diluents/bundled items to the master list as well, so we can see those items separately 🙏 

Wonder if it's worth adding a carry over issue to sync the master list... would be nice just to enable GAPS items in another store too - though maybe that's more of a dev desire, instead of controlling what's seen by each store 😁

server/repository/src/migrations/v2_04_00/mod.rs
```diff
@@ -42,6 +43,7 @@ impl Migration for V2_04_00 {
             Box::new(indicator_value_create_table::Migrate),
             Box::new(add_bundled_item_table::Migrate),
             Box::new(add_demographic_indicator_types_to_activity_log::Migrate),
+            Box::new(item_changelog::Migrate),
```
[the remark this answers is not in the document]
lache-melvin: See you haven't resolved this comment, did you have a recommended plan here @jmbrunskill? Just leaving for now?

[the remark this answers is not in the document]
jmbrunskill: Yup planning to just leave it for now.

jmbrunskill:
> Wonder if it's worth adding a carry over issue to sync the master list... would be nice just to enable GAPS items in another store too - though maybe that's more of a dev desire, instead of controlling what's seen by each store 😁

Yeah, I'm not 100% sure what is the 'correct' behaviour. I think we probably don't really want to use that master list for anything other than visibility, so while I was originally thinking that, now I'm thinking. Maybe the Central Server should have a special show hidden items button similar to mSupply?

adamdewey:
@jmbrunskill  Ah, good point - I was imagining there would be a Master List with all items on that are not visible to any stores, and then they would also add the items that they **are** using to a different Master List which is visible to stores?

jmbrunskill:
> @jmbrunskill Ah, good point - I was imagining there would be a Master List with all items on that are not visible to any stores, and then they would also add the items that they **are** using to a different Master List which is visible to stores?

Ok, that is essentially what we have. We have a masterlist with all the gaps items, but only on the OMS Central Server. If you want to use any of the items for other stores, you'll need to add them manually to a master list.

adamdewey:
I think that should be fine - as long as they're there somewhere it's pretty straightforward to add them to a new Master List

jmbrunskill:
@lache-melvin diluents are added to the master list now.
Doing this I found another small bug, as diluents were using the Vaccine Type Name not mSupply Item Name for lookups, that's fixed now too.

jmbrunskill:
FYI this does a decent job of clean up, although might delete non-gaps items so please don't run unless you're sure!
```
DELETE from bundled_item;
DELETE from packaging_variant;
DELETE FROM item_variant WHERE id not in (SELECT coalesce(item_variant_id,'noid') from stock_line)
AND id not in (SELECT coalesce(item_variant_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_variant_id,'noid') from stocktake_line);
DELETE FROM master_list_line WHERE id not in (SELECT item_link_id FROM stock_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from stocktake_line);
DELETE FROM item_link WHERE id not in (SELECT item_link_id FROM stock_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from stocktake_line);
DELETE FROM item WHERE id not in (select item_id from item_link);
```

lache-melvin (approved):
Looks great thanks for all this James!!

PR #5456, opened 2024-11-18
5388 add error for reasons in manual requisition changes

### LINKED ISSUE #5388: Error for reasons
Add backend error for Reasons if supply quantity is different from requested quantity and add checks in frontend to alert the user of this error

fergie-nz:
Fixes #5388

# 👩🏻‍💻 What does this PR do?

Adds error when suggested quantity differs from requested quantity in response requisition if no reason provided.
Adds error handling and info toast to front end.

This also only occurs on program requisitions; non program requisitions won't have the option to add reasons.
(Side question - should non program requisitions allow for requested != suggested? I have in this case, but am realising now that I'm not certain.)

No error handling on creation (because lines are created with no info and won't have requested quantities).

# 🧪 Testing

You can test this by:

- [ ] Adding a line to a program response requisition
- [ ] Change the requested quantity to be different from suggested without a reason
- [ ] See the error
- [ ] Add a reason for the change
- [ ] See no error when reason has been added and requested != suggested

# 📃 Documentation


client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/hooks.tsx
```diff
@@ -42,7 +46,26 @@ export const useDraftRequisitionLine = (item?: ItemRowFragment | null) => {
...
+        result.updateResponseRequisitionLine.__typename ===
+        'UpdateResponseRequisitionLineError'
+      ) {
+        switch (result.updateResponseRequisitionLine.error.__typename) {
+          case 'ReasonNotProvided':
+            error(t('error.requires-reason'))();
+            break;
+          default:
+            error(result.updateResponseRequisitionLine.error.description)();
+            break;
```
fergie-nz, line 62: Handling error toast inside hook.

I see there are a number of ways that we are handling errors - in the api, in the hooks, or in components. 

Would love to add some consistency to our patterns!

roxy-dao (changes requested):
The inputs are saved onBlur, which means once you set a requestedQuantity != suggestedQuantity, the requestedQuantity won't change -> error propped, requestedQuantity doesn't get saved -> infinite loop of trying to change requestedQuantity but being unable to because of 'reason'

client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
```diff
@@ -69,6 +73,11 @@ export const ResponseLineEdit = ({
       ? available / (draft?.averageMonthlyConsumption ?? 1)
       : 0;
 
+  const customRequestedDisabled =
+    !!hasLinkedRequisition ||
+    (!!isProgram && reasonsData?.nodes.length === 0) ||
+    reasonsLoading;
```
roxy-dao, line 79: Why have you disabled requested with these extra conditions? Users should be able to input requested quantity if the Requisition hasn't been generated from an Internal Order

roxy-dao (changes requested):
Hiiii. 

The change request was to remove the reason error check for lines from backend and implemented in frontend. I see that it is still there. The requested quantity cannot be updated at all because of the error. 

The expected behaviour is: If the user has entered a requested quantity that is different from the requested quantity then the reasons field should be highlighted to indicate to the user that this is now a required field.

For backend you can add the reasons check when the user tries to update the Requisition and the lines on the detail view can be highlighted. This will stop the user from being able to finalise the requisition without putting in reasons for the lines that need it.

Lemme know if you want to sit down and walk through this together (:

client/packages/requisitions/src/ResponseRequisition/DetailView/DetailView.tsx
```diff
@@ -67,52 +68,55 @@ export const DetailView: FC = () => {
...
+                      requisitionId: data.id,
+                      itemId: newItem.id,
+                    });
+                    navigate(
+                      RouteBuilder.create(AppRoute.Distribution)
+                        .addPart(AppRoute.CustomerRequisition)
+                        .addPart(String(data.requisitionNumber))
+                        .addPart(String(newItem.id))
+                        .build(),
+                      { replace: true }
```
roxy-dao, line 106: This should be removed. We don't wanna replace the history

client/packages/requisitions/src/ResponseRequisition/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -76,11 +87,41 @@ const useStatusChangeButton = (requisition: ResponseFragment) => {
...
+
   const onConfirmStatusChange = async () => {
     if (!selectedOption) return null;
+    let result;
     try {
-      await update({ status: selectedOption.value });
-      success(t('messages.saved'))();
+      result = await save({ id, status: selectedOption.value });
+      const errorMessage = mapStructuredErrors(result);
+      console.log('errorMessage', errorMessage);
```
roxy-dao, line 118: Remove console.log?

fergie-nz:
> The reasons being far right make it hard to see the error highlight since the user has to scroll all the way to the right to see it... but that can be discussed later >.>

yes agreed! I found that annoying too. I still wonder if error should prompt when the user first enters also.

I will make a carryover issue for this.

roxy-dao:
> > The reasons being far right make it hard to see the error highlight since the user has to scroll all the way to the right to see it... but that can be discussed later >.>
> 
> yes agreed! I found that annoying too. I still wonder if error should prompt when the user first enters also.
> 
> I will make a carryover issue for this.

Yup, I think highlighting the row on the page will also make it clear to the user that this needs to be inputted. The asterisk probably isn't obvious enough

roxy-dao (approved):
Cool thanks for changes Ferg

fergie-nz:
> > > The reasons being far right make it hard to see the error highlight since the user has to scroll all the way to the right to see it... but that can be discussed later >.>
> > 
> > 
> > yes agreed! I found that annoying too. I still wonder if error should prompt when the user first enters also.
> > I will make a carryover issue for this.
> 
> Yup, I think highlighting the row on the page will also make it clear to the user that this needs to be inputted. The asterisk probably isn't obvious enough

carryover issue: https://github.com/msupply-foundation/open-msupply/issues/5544

PR #5475, opened 2024-11-19
fixed show/hide all

### LINKED ISSUE #5396: Show hide all columns doesn't
## What went wrong? 😲

I experienced some confusion - and expect that others may have the same problem..

When showing and hiding columns, there is a very useful option to show/hide all columns. That's what I wanted to do, so I clicked it, and was disappointed with the result:

https://github.com/user-attachments/assets/12ba4034-8ee5-49fb-b101-545d4f06c329

## Expected behaviour 🤔
The column was the show/hide checkbox column and I was showing and hiding that - and not all the columns which were available.

If there was a 'select all' checkbox at the top of the column visibility popup then I wouldn't be so confused. Renaming the rogue column would also help, but might introduce other confusion!

## How to Reproduce 🔨

1. Go to the detail view of an outbound shipment, say (any table with column selection will do)
2. Click the show/hide column button

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

myspace20:
Fixes #5396

# 👩🏻‍💻 What does this PR do?
This adds the show or hide all logic to the data table custom column picker, reported in the detailed view of any outbound shipment. Previously, the hide/show all checkbox existed but the logic was unavailable.

# 🧪 Testing

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/layout/tables/hooks/useColumnDisplayState.ts
+++ b/client/packages/common/src/ui/layout/tables/hooks/useColumnDisplayState.ts
@@ -25,6 +25,10 @@ export const useColumnDisplayState = <T extends RecordWithId>(
     ])
   );
 
+  const [allIsSelected, setAllIsSelected] = useState(false)
+
+
+
   const toggleColumn = (colKey: string) => {
     const newState = {
       ...columnDisplayState,
@@ -36,6 +40,16 @@ export const useColumnDisplayState = <T extends RecordWithId>(
           ? false
           : !columnDisplayState[colKey],
     };
+    if(colKey === 'selection'){
+      if(!allIsSelected){
+        Object.keys(newState).forEach((key)=>newState[key] = true)
+        setAllIsSelected(true)
+      }else{
+        Object.keys(newState).forEach((key)=>newState[key] = false)
+        setAllIsSelected(false)
+      }
+      
+    }
     setColumnDisplayState(newState);
     setHiddenColsStorage({
       ...hiddenColsStorage,


myspace20:
@roxy-dao I need a review on this.

CarlosNZ (changes requested):
Hi, thanks for this, it's working as you intended.

However, I think the problem here is that it doesn't make sense to be able to turn *off* all columns.

Also, there is some additional confusion caused by the incorrect labelling of this originally -- it said "Select / unselect all columns", but what it actually did was toggle the column that should be labelled "Select / unselect all *rows*", which is what this column is used for in the actual tables.

I think the correct behaviour should be as follows:

- The column label should be "Select / unselect all rows"
- In the column selection drop-down, this column (probably) should not be visible (i.e it's always on and we don't refer to it at all)
- If (and only if) there is any columns hidden, then in the selection drop-down, there should be a button/link to "Restore all columns" which resets all columns to visible. Not sure about the UI design here, but it needs to be fairly unobtrusive and not clutter the UI (perhaps we need a mockup from @mariyamsupply before we go any further?)
- There should be no option to hide all columns, it's purely a "restore" functionality.

Maybe wait till we get some design consensus on this before progressing further?

mark-prins:
Thanks @CarlosNZ I think you've summarised the requirements nicely. 
I was originally thinking to have a show/hide all columns option, but you're right, that hiding all columns makes little sense. It only helps if you want to show only a couple of columns. without the hide all option you have to manually hide each column which is pretty tedious. 

Agree with the first two changes - and I think there's a valid case for having a show/hide all option. happy to go with that or the 'restore' option. That could just be a standard link style, but yeah, will leave that for @mariyamsupply

mariyamsupply:
Now that I understand that 'select/unselect' is the name for the checkbox, not the action, I think it's actually shouldn't be on the list.
the checkbox is a function for actions, not the data. I would vote for: 
- we have 'select/hide all columns', 
- we don't have the checkbox column on the list
- if the last column (or all columns) are unselected we hide the column with check-boxes as well

mariyamsupply:
Potential design for the rearranging columns icon (after we remove Actions dropdown)
arrange columns

roxy-dao:
@myspace20  Hi are you still interested in working on this issue?

myspace20:
I am a little busy right now

mark-prins:
I'm taking over! superceded by #5907

PR #5476, opened 2024-11-19
5437 manual requisitions inconsistent behaviour in the reasons dropdown

### LINKED ISSUE #5437: Manual Requisitions: confusing interaction with the Reasons dropdown field
## What went wrong? 😲

Was testing the setup for the new feature, and noticed a few things I wanted to raise: 

1. Clicking between Reasons for Manual Requisitions is sometimes inconsistent and reverts back to the original Reason (e.g. user needs to double-click to select another one)
2. Reason only becomes interactable with when we change the quantities -> however, if we revert back to 0, the field does not clear and instead solidifies whatever reason we chose. 
3. Even though the field is faded out, the user actually is still able to click on it (misleading?) and that leads to the old numbers appearing in the fields (e.g. going from 0 to 3 as shown on the video)
4. If 0 are entered for Requested Quantity and Quantity to Supply, should the user be able to click on Next and proceed? Should there be some sort of warning to notify they haven't entered any values? Not sure if this is expected behaviour!

Video uploaded ->>> here

## Expected behaviour 🤔

1. Single click is required
2. Field clears if the user reverts back to 0
3. If the field UX/UI is faded out, it should not be interactable - to me the order of action is, it should be intractable with again only once the user enters the values greater than 0 in the Quantities fields
4. Unsure what the expected behaviour is -> should the user be allowed to continue if no changes were made to the requisition quantities? Should there be a prompt?

## How to Reproduce 🔨

2. Create a new Manual Requisition
3. Play around with the quantities fields and the reason dropdown as described above (make sure you have a few reasons set up in the OG mSupply!)
4. See inconsistent behaviours

## Your environment 🌱

- Database type: SQLite3

roxy-dao:
Fixes #5437

# 👩🏻‍💻 What does this PR do?
- Clears reason if suggested = requested
- Calculate remaining in frontend using (supply - already issued) to reduce number being fetched from API
- Uses draft for same line. Helps if the user enters inputs really fast, so the API will update all the input before invalidating... can cause numbers to not update straight away on slow networks if the user is swapping between items within 2 seconds... but better than it was before which was not saving... 👀  

# 🧪 Testing

- [ ] Have some lines in a requisition
- [ ] Throttle network
- [ ] Type inputs really fast 
- [ ] Inputs should stay what you changed them to

# 📃 Documentation

  1.
  2.


client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
```diff
@@ -229,8 +229,15 @@ export const ResponseLineEdit = ({
                 <NumericTextInput
                   width={INPUT_WIDTH}
                   value={draft?.requestedQuantity}
-                  onChange={value => update({ requestedQuantity: value })}
+                  onChange={value => {
+                    if (draft?.suggestedQuantity === value) {
+                      update({ requestedQuantity: value, reason: null });
```
roxy-dao, line 234: Clearing reason here since you can't edit suggested

client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
```diff
@@ -229,8 +229,15 @@ export const ResponseLineEdit = ({
...
+                    } else {
+                      update({ requestedQuantity: value });
+                    }
+                  }}
                   disabled={!!hasLinkedRequisition}
+                  onBlur={save}
```
roxy-dao, line 240: Missed save on blur so this value wasn't being saved if it was the only one edited..

client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
```diff
@@ -326,7 +332,9 @@ export const ResponseLineEdit = ({
             Input={
               <NumericTextInput
                 width={INPUT_WIDTH}
-                value={draft?.remainingQuantityToSupply}
+                value={
+                  (draft?.supplyQuantity ?? 0) - (draft?.alreadyIssued ?? 0)
```
lache-melvin, line 336: Screenshot 2024-11-27 at 9 02 54 AM

If you issue more than supply quan in the outbound shipment then this shows as negative... I guess that's fine?? Maybe should be 0 idk

server/service/src/requisition_line/response_requisition_line/update.rs
```diff
@@ -144,7 +144,7 @@ fn generate(
         u.addition_in_units = updated_addition_in_units.unwrap_or(u.addition_in_units);
         u.expiring_units = updated_expiring_units.unwrap_or(u.expiring_units);
         u.days_out_of_stock = updated_days_out_of_stock.unwrap_or(u.days_out_of_stock);
-        u.option_id = updated_option_id.or(u.option_id);
+        u.option_id = updated_option_id;
```
lache-melvin, line 147: Assuming you changed this so its clearable? I think this way okay for now, seeing as we know we always send the full draft in our update mutations, but I think this should be NullableUpdate, so that if we send `UpdateInput { expiring_units: 10 }` it doesn't clear the reason 😁

lache-melvin (approved):
LGTM! Thanks Roxy 🙏

server/service/src/requisition_line/response_requisition_line/update.rs
roxy-dao: The draft keeps the reason, so it doesn't clear unless the draft is updated to null

lache-melvin: Yep - just mean if we were to use this API from somewhere else!

roxy-dao: Oh yes definitely!

PR #5480, opened 2024-11-19
4702 add report sorting via wasm functions

fergie-nz:
# 👩🏻‍💻 What does this PR do?

Duplicates #5195 PR but into release branch


roxy-dao (approved):
Thanks Ferg!

PR #5484, opened 2024-11-20
5367 add category id to item filter

### LINKED ISSUE #5367: Add item category to item filter
Allow for filtering of items by item category, so that custom prescription picker can filter by selected item category id.

### Moneyworks Jobcode 🧰
DJIFPCDV

lache-melvin:
Fixes #5367

# 👩🏻‍💻 What does this PR do?

Adds category ID to item filter, to use from custom JSON form component

Note: we currently only join at, and therefore support search by, lowest level category - this is all we need for RH module... but where should I capture this? Don't want someone confused trying to filter by higher level category and it not working?

# 🧪 Testing

You can test with this query - compare results with item search in 4D:
```gql
query itemByCategoryId($storeId: String!, $categoryId: String!) {
  items(
    storeId: $storeId
    filter: { categoryId: $categoryId }
  ) {
    ... on ItemConnector {
      nodes {
        id
        name
      }
    }
  }
}

```

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/item.rs
+++ b/server/repository/src/db_diesel/item.rs
@@ -1,4 +1,5 @@
 use super::{
+    item_category_row::item_category_join,
     item_link_row::item_link::dsl as item_link_dsl,
     item_row::{item, item::dsl as item_dsl},
     master_list_line_row::master_list_line::dsl as master_list_line_dsl,
@@ -45,6 +46,7 @@ pub struct ItemFilter {
     pub name: Option<StringFilter>,
     pub code: Option<StringFilter>,
     pub r#type: Option<EqualFilter<ItemType>>,
+    pub category_id: Option<String>,
     /// If true it only returns ItemAndMasterList that have a name join row (void if is_visible_or_on_hand is true!)
     pub is_visible: Option<bool>,
     /// If true it returns ItemAndMasterList that have a name join row, or items with stock on hand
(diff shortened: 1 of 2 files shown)


server/repository/src/db_diesel/item.rs
```diff
@@ -227,6 +235,15 @@ fn create_filtered_query(store_id: String, filter: Option<ItemFilter>) -> BoxedI
             query = query.filter(item_dsl::is_vaccine.eq(is_vaccine));
         }
 
+        if let Some(category_id) = category_id {
+            let item_ids_for_category_id = item_category_join::table
+                .select(item_category_join::item_id)
+                .filter(item_category_join::category_id.eq(category_id.clone()))
+                .into_boxed();
+
+            query = query.filter(item_dsl::id.eq_any(item_ids_for_category_id));
```
jmbrunskill, line 244: Was wondering what should happen with merged items?

I think that it's better to just leave it how it is though, as you don't want to find by the category of the non-merged item right?

Maybe add a comment to that effect?

jmbrunskill (approved):
Looks legit! Haven't tested, but I think we'll find any issues pretty quickly while developing the rest of the feature or in testing.

server/repository/src/db_diesel/item.rs
lache-melvin: Yeah was wondering about this too, had a chat with Roxy... Seems like it's doing the right thing - if item was merged, we should find by the category of the "chosen" item - will add comment 🙏

PR #5518, opened 2024-11-22
5409 hide temperature ranges if unspecified

### LINKED ISSUE #5409: iSC/Cold chain/GAPS: do not show temperature ranges if they haven't been specified
## What went wrong? 😲

At the moment, it seems like when you add Location Type in OG mSupply, the temperature input is compulsory - e.g. it is pre-populated with 0, and there cannot be a blank value. However, not every Location Type needs to have temperature associated with it, and the user can leave it unchanged. 

Screenshot 2024-11-14 at 6 35 35 PM

In omSupply though, it comes out as a temperature range 0-0, which can be confusing for the user and does not look nice. 

Screenshot 2024-11-14 at 6 35 42 PM

## Expected behaviour 🤔

As a solution, we can either: 

- Hide the temperature range if both values entered are 0
- Hide the temperature range if it hasn't been tempered with in OG mSupply
- Or allow empty values in the cells, so if there is a value that is 0, the user can enter it themselves

## How to Reproduce 🔨

1. In OG mSupply, create a Location Type with both 0 as values
2. Sync omSupply 
3. Go to Inventory -> Locations -> click on an existing location or create a new one
4. Once done, click on it and set the test temperature range
5. See the range appear even if it is 0-0

## Your environment 🌱

- Platform: Google Chrome (browser for Mac) and Galaxy Tab A8 (Android)

- Database type: SQLite3

zachariah-at-msupply:
Fixes #5409

# 👩🏻‍💻 What does this PR do?

Changes the rust query to filter out all cold storage types with a min tempereature = 0 and a max temperature = 0, so that they are not displayed on the front end.
Changes the label of...
- the 'storage type' column in inventory>locations,

image

- the 'temperature' selector dropdown in the inventory>locations>add location menu

image

- the 'temperature' in catalogue>detail view>item variants

image

- the 'temperature in catalogue>detail view> item variant>add variant menu

image

...to 'cold storage type' to avoid confusion.

This required a new 'label.cold-storage-type', which requires translation.

Have a nice day :)

# 🧪 Testing

 

- [ ] In OG mSupply, create a Location Type with both 0 as values.
- [ ] Sync omSupply.
- [ ] Go to Inventory -> Locations -> click on an existing location or create a new one.
- [ ] Once done, click on it and set the test temperature range.
- [ ] See that ranges that are 0degC to 0degC don't appear.
- [ ] Navigate to the locations mentioned above and see that the labels make sense and fit well.

# 📃 Documentation

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -976,6 +976,7 @@
   "label.stocktake-date": "Stocktake Date",
   "label.stocktake-frequency": "Stocktake frequency",
   "label.storage-type": "Storage type",
+  "label.cold-storage-type": "Cold storage type",
   "label.store": "Store",
   "label.strength": "Strength",
   "label.sub-catalogue": "Sub catalogue",
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
@@ -121,7 +121,7 @@ const ItemVariantForm = ({
         />
 
         <InputWithLabelRow
-          label={t('label.temperature')}
+          label={t('label.cold-storage-type')}
           labelWidth="200"
           Input={
             <Box width="100%">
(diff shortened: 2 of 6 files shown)


jmbrunskill (approved):
Looks good to me, Also checked that a temperature range from 0 to 25 isn't filtered out, which is good!

image

PR #5521, opened 2024-11-22
Remove doses_per_unit from item_variant

### LINKED ISSUE #5512: Remove dose from the item variant

Part of https://github.com/msupply-foundation/open-msupply/issues/5432

Current vaccine variants define a number of doses that can be different from the item.
However this field isn't used when dispensing or vaccine cards, so it creates confusion for users as to the purpose.

Remove the field from the UI and database

Could remove just from the UI, but that could create other confusion in future.

jmbrunskill:
Fixes #5512

# 👩🏻‍💻 What does this PR do?

Remove doses_per_unit from the item_variant table an associate parts of the application

I haven't created a migration to remove the field from the db, so any RC databases will still have that field.
I think that's a good trade off vs adding an then removing the field in separate migrations.

# 🧪 Testing

- [ ] Create, and edit item_variants
- [ ] Make sure they all work as expected.

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
Looks good to me James! Yep, agree with the decision to remove from the migration and just leave that column in RCs 🙏

PR #5523, opened 2024-11-22
5162 response requisition indicator UI

### LINKED ISSUE #5162: Indicators UI
#5134 must be implemented first. 

- Designs for the detail view tab coming -

Image

HIV indicators open up in a new page for editing
Screenshot 2024-10-24 at 11 30 07 AM

- [ ] customer manual program requisition
  - [x] Add Indicator tab
  - [ ] Indicator tab only if other party is store and the program has indicators
  - [ ] (Program) Indicator table component
  - [ ] Indicator tab shows indicator table for each `programIndicator`
  - [ ] Clicking a row navigates to URL including `:indicatorLineId`, programIndicator editing page
  - [ ] Program indicator page replicates the `ResponseEditPage`/`ResponseLineEdit`
  - [ ] `IndicatorLineEdit` needs to be dynamic according to the `IndicatorColumns` defined
- [ ] internal order
  - [ ] Add Indicator tab
  - [ ] Indicator tab only if other party is store
  - [ ] (Program) Indicator table component
  - [ ] Indicator tab shows indicator table for each `programIndicator`
  - [ ] Clicking a row navigates to URL including `:indicatorLineId`, programIndicator editing page
  - [ ] Program indicator page replicates the `ResponseEditPage`/`ResponseLineEdit`
  - [ ] `IndicatorLineEdit` needs to be dynamic according to the `IndicatorColumns` defined

OMS:CIMDV

Chris-Petty:
Fixes #5162

# 👩🏻‍💻 What does this PR do?

Add indicator UI to manual customer program requisitions.

image
image

For now we just have buttons to access each program indicator's lines. In the future it'll be tables for each as per designs.

Problems I've noticed that need fixing:

HIV indicators never seems to select first indicator line:

image

Probably a number input problem, but default value is typically `0`. It is hard to actually remove this leading 0 without highlighting the digit and replacing it. This would feel even more awful on tablets:

image

# 🧪 Testing

- [ ] Central OG and Central OMS
- [ ] Stores configured to use programs with program indicators, periods etc. Check the "Is HIV" checkbox in program config
- [ ] Another program without "Is HIV"
- [ ] In OMS make a manual program customer requisition to another store and use the program that "Is HIV"
- [ ] Indicator tab shows
- [ ] It has Regimen and HIV buttons
- [ ] Pressing them takes you to corresponding set of questions
- [ ] Make another requisition using a program that is NOT "HIV"
- [ ] It only has the Regimen button

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -378,6 +380,7 @@
   "error.v6-server-not-configured": "V6 central server not configured",
   "error.v6-server-not-configured-hint": "Check the central server URL",
   "error.vaccine-course-update-failed": "Vaccine course failed to save",
+  "error.value-type-not-correct": "Value type not correct",
```
roxy-dao, line 383: User probably won't understand what this means... since we're only checking if Number is the correct type, maybe we can change the error message to be friendlier

client/packages/requisitions/src/RequisitionService.tsx
```diff
@@ -29,6 +30,13 @@ const customerRequisitionLineRoute = RouteBuilder.create(
   .addPart(':itemId')
   .build();
 
+const indicatorLineRoute = RouteBuilder.create(AppRoute.CustomerRequisition)
+  .addPart(':requisitionNumber')
+  .addPart('indicator')
+  .addPart(':programIndicatorCode')
+  .addPart(':programIndicatorLineId')
```
roxy-dao, line 37: Carryover to make the `indicator` and `code` values in breadcrumbs non-clickable...

client/packages/requisitions/src/ResponseRequisition/DetailView/DetailView.tsx
```diff
@@ -45,6 +57,27 @@ export const DetailView: FC = () => {
...
   }, []);
 
+  const onProgramIndicatorClick = useCallback(
+    (
+      programIndicator?: ProgramIndicatorFragment,
+      indicatorLine?: IndicatorLineRowNode,
+      response?: ResponseFragment
+    ) => {
+      // TODO: Snack?
+      if (!response || !indicatorLine) return;
```
roxy-dao, line 67: Wanna add a snack to let the users know why they can't click the button if they happen to click is and there's nothing there? or remove this and disable the buttons down below

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
```diff
@@ -0,0 +1,116 @@
...
+  next,
+  hasPrevious,
+  previous,
+  currentLine,
+}: IndicatorLineEditProps) => {
+  const columns = currentLine?.columns.sort(
+    (a, b) => a.columnNumber - b.columnNumber
+  );
+  const t = useTranslation();
+  const { mutateAsync } = useResponse.document.updateIndicatorValue();
```
roxy-dao, line 43: Can you create a carryover for turning saving the indicator values in a draft before upserting. The values will then be saved on onBlur instead of onChange.

Saving it directly could cause data discrepancies if there are multiple indicators and the user enters the data quickly/navigates away quickly after entering data

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/hooks.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }, [indicatorValue]);
+
+  const update = (patch: Partial<IndicatorValueFragment>) => {
+    if (draft) {
+      setDraft({ ...draft, ...patch });
+    }
+  };
+
+  return { draft, isLoading, save: () => draft && save(draft), update };
+};
```
roxy-dao, line 69: Not using this anywhere... probably can add to carryover

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const IndicatorsTab = ({
+  onClick,
+  isLoading,
+  response,
+  indicators,
+}: IndicatorTabProps) => {
+  const t = useTranslation();
+  if (isLoading) {
+    return <NothingHere body="There are no indicators for this requisition" />;
```
roxy-dao, line 31: make this translatable please XD

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+    indicator => indicator?.code === 'HIV'
+  );
+  const regimenIndicators = indicators?.filter(
+    indicator => indicator?.code === 'REGIMEN'
+  );
+
+  return (
+    <Box display="flex" flexDirection="column" padding={2} gap={2}>
+      <ButtonWithIcon
+        // disabled={disableAddButton}
```
roxy-dao, line 43: See above comment about the snack

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+        label={t('button.regimen')}
+        Icon={<PlusCircleIcon />}
+        onClick={() =>
+          onClick(
+            regimenIndicators?.[0],
+            regimenIndicators?.[0]?.lineAndColumns[0]?.line,
+            response
```
roxy-dao, line 50: Comment above about why this is getting the data at 0

roxy-dao (commented):
I probably shouldn't review this since we did most of it together! But....

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/hooks.tsx
Chris-Petty: whoops yolo removed only read half comment e75041cf5519

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const t = useTranslation();
+  if (isLoading) {
+    return <NothingHere body="There are no indicators for this requisition" />;
+  }
+  const hivIndicators = indicators?.filter(
+    indicator => indicator?.code === 'HIV'
+  );
+  const regimenIndicators = indicators?.filter(
+    indicator => indicator?.code === 'REGIMEN'
+  );
```
lache-melvin, line 38: Seems like it might be easiest just to hide/show these buttons based on if any indicators of that kind are defined? Though I suppose then what do you show if neither are available... snack seems nice 😁

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
lache-melvin: Don't think this can wait for carry over - at a minimum, we need to debounce the API call and have some frontend state... I tried with throttled network and everything I typed would get overwritten 😭

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorEditPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const sortedLines = lines.sort((a, b) => a.lineNumber - b.lineNumber);
+
+  const { hasNext, next, hasPrevious, previous } = usePreviousNextIndicatorLine(
+    sortedLines,
+    currentLine
+  );
+
+  useEffect(() => {
+    setCustomBreadcrumbs({
+      2: 'Indicators',
```
lache-melvin, line 49: Should be translated

lache-melvin (changes requested):
Found a couple issues:

- HIV indicators showing for all program requisitions, not just HIV ones
- Regimen indicators should be just for that program/period schedule, but they're showing up on all program requisitions
- I'm getting 3 entries for each HIV indicator, instead of 1?

Screenshot 2024-11-25 at 12 42 44 PM

Naturally, all these bad boys that I probably should be able to see, I also can't save (getting spammed with `RecordNotFound` errors when I try to input a value)

Possibly all of this is because the programIndicators query looks like it wants a period ID to filter by, and it's not getting one?
Screenshot 2024-11-25 at 12 43 14 PM

Side note - should I expect the default values from regimen data to map through or not at this stage?

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
roxy-dao: Wanna filter this by program ID too... to get the correct indicators >.>

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorEditPage.tsx
Chris-Petty: 288b2e731632

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const IndicatorsTab = ({
+  onClick,
+  isLoading,
+  response,
+  indicators,
+}: IndicatorTabProps) => {
+  const t = useTranslation();
+  if (isLoading) {
+    return <NothingHere body={t('error.no-indicators')} />;
```
Chris-Petty, line 31: @roxy-dao shouldn't this be a loading state page rather than a no indicators page?

roxy-dao: Yeah XD

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
Chris-Petty: I think it's better to hide the buttons as the niche of users trained on this feature should know the deal and if the button isn't there it saves them the effort of clicking to find out it's not there. We already hide the tab entirely if there are no indicators for the given requisition/program/period. HIV indicators for example, being present should really be self evident based on it being an HIV program to begin with.

Ideally to match the configurable schema these buttons (or tables in the future) would be dynamically generated rather than hardcoded, in which case you'd effectively/hide show based on existence too.

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
```diff
@@ -0,0 +1,113 @@
...
+  previous: IndicatorLineRowFragment | null;
+  currentLine?: IndicatorLineWithColumnsFragment | null;
+}
+
+const INPUT_WIDTH = 185;
+const LABEL_WIDTH = '150px';
+
+const InputWithLabel = ({ data }: { data: IndicatorColumnNode }) => {
+  if (!data?.value) {
+    return;
```
lache-melvin, line 34: Screenshot 2024-11-27 at 12 56 38 PM

Not sure if you were just returning here so the type is definitely available but... if no value on indicator yet then I get nothing!

lache-melvin: Probably should just pass through a default empty string if not defined...

roxy-dao: If the values don't exist for the indicator then the buttons or (tab if both indicators) shouldn't show!

lache-melvin: Alas we have determined this was actually because I had indicator lines with no values on an existing program requisition - works fine on newly created ones! Need to find a way to handle existing requisitions 🙏

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -0,0 +1,76 @@
...
+          onClick={() =>
+            onClick(
+              regimenIndicators[0],
+              regimenIndicators[0]?.lineAndColumns[0]?.line,
+              response
+            )
+          }
+        />
+      )}
+      {hivIndicators.length > 0 && (
```
lache-melvin, line 60: Screenshot 2024-11-27 at 1 16 32 PM

Still seeing HIV indicators on my non-HIV program?

lache-melvin (changes requested):
Looking good - just 2 things:

- HIV indicators available on non-hiv program requisitions
- Old requisitions show regimen data columns but no fields - should not show at all

Thanks for all the mahi Chris 🚀

lache-melvin (changes requested):
Well they're not there... but it would appear the buttons are still showing, so I get a blank page now 😭 
Screenshot 2024-11-27 at 3 35 13 PM
Screenshot 2024-11-27 at 1 16 32 PM

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
Chris-Petty: Hmm I think if you look at an old program requisition, the button would still be there but if you click it everything gets filtered out? It's getting a bit gross cause I already do the same filter in `IndicatorEditPage` and `IndicatorLineEdit`, and I'd have to do it again in `IndicatorTab`. In hindsight I really should have persisted with doing it in the backend, as this will pop up as a problem request requisitions too.

As discussed I've also been ignoring that you can turn HIV off on program and the indicators will still be there 😭. I'm not 100% sure if OG does it, but ideally OG would set the HIV program_indicator to be inactive:

https://github.com/msupply-foundation/open-msupply/blob/97f0eeecfcc247d733fe8099eaaa103863b391b8/server/repository/src/db_diesel/program_indicator_row.rs#L22

I don't believe we're doing any filtering on that yet but.

Chris-Petty:
> Well they're not there... but it would appear the buttons are still showing, so I get a blank page now 😭 Screenshot 2024-11-27 at 3 35 13 PM Screenshot 2024-11-27 at 1 16 32 PM

Dang yea, 'cause I did the filtering in the child routes but not IndicatorTab 🫠 . Really should do it in the backend so it's not so easily missed!

lache-melvin:
> Dang yea, 'cause I did the filtering in the child routes/components but not IndicatorTab 🫠 . Really should do it in the backend so it's not so easily missed!

Yeah, would like to see this as a clean up PR if poss - would be good to get the feature out to QA though?

lache-melvin:
whoops just saw that hiv indicators could have the same issue, pushed quick fix for that too!

client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorsTab.tsx
```diff
@@ -34,9 +34,14 @@ export const IndicatorsTab = ({
   if (!indicators || indicators.length === 0) {
     return <NothingHere body={t('error.no-indicators')} />;
   }
+
   const regimenIndicators = indicators.filter(
-    indicator => indicator.code === 'REGIMEN'
+    indicator =>
+      indicator.code === 'REGIMEN' &&
+      // Should only include regimen indicators if they have at least one column with a value
+      indicator.lineAndColumns.some(line => line.columns.some(c => c.value))
```
Chris-Petty, line 42: coming from rust more recently I was like ugh I though js had an `Array.any` 😭. Been using `filter` and `find` hahahaaaaaaaa

lache-melvin:
Actually just tried setting a new HIV program back to non-HIV, and the indicators are gone so... must just be my munted dev data, sorry!

PR #5533, opened 2024-11-22
4791 add report versioning

### LINKED ISSUE #4791: Reports-versioning: Select the right version of a report to include

When omSupply requests available reports, the available reports will follow these rules

```
found_report = find report with this code

if not found_report {
    upsert report
    return
}

if found_report.version >= this report.version {
    return
}

upsert report
```

**code** is a new field that identifies the same report for different versions

**version** follows omSupply major and minor version, for omSupply to select correct report version for current omSupply instance, and patch identifies any updates to current report version

When omSupply generates report it will select report based on version with the follow rules

```
After narrowing down on report by context and sub context, and then by code

report_versions = all reports for code

if report_versions have custom reports {
    report_versions = only custom reports
}

// Report selection logic will be such to satisfy the following use cases (path version is ignored)
report_version = [2.3.0, 2.3.5, 2.8.2, 2.8.3, 3.0.1, 3.5.1]
if omSupply.version = 2.3 selected report = 2.3.5
if omSupply.version = 2.4 selected report = 2.3.5
if omSupply.version = 2.8 selected report = 2.8.3
if omSupply.version = 3.2 selected report = 3.0.1
if omSupply.version = 4.5 selected report = 3.5.1
```

This will allow us to update existing report through open-mSupply Central without upgrading remote sites, for current and previous versions of omSupply. When omSupply is deployed are remote sites, it's not that easy to update them to later versions, but patching existing versions can be done through report configurations. Also custom reports would still be used instead of standard reports, where needed.

fergie-nz:
Fixes #4791

# 👩🏻‍💻 What does this PR do?

Two parts:

- Adds only upsert if the report to be upserted is a later version than the existing report of the same code (or if is_custom && is later version than an existing custom report)
- Adds filter for only the latest version of any report which is compatible with the app version

There is a unique case which could casue issues where if a db has a report version later than the app version, you wouldn't be able to upload older report versions. But I don't think we need to worry about that.

You can test this report with this database which includes a number of report versions and is_custom combinations:

[central_test.sqlite.zip](https://github.com/user-attachments/files/17876383/central_test.sqlite.zip)

# 🧪 Testing

- [ ] Go to reports
- [ ] See only latest versions displayed

# 📃 Documentation


server/service/src/standard_reports.rs
```diff
@@ -41,33 +44,66 @@ impl StandardReports {
...
+                    template: serde_json::to_string_pretty(&report.template)?,
+                    context: report.context,
+                    sub_context: report.sub_context,
+                    argument_schema_id: report.argument_schema_id,
+                    comment: report.comment,
+                    is_custom: report.is_custom,
+                    version: report.version,
+                    code: report.code,
+                })?;
+            };
```
andreievg, line 106: Hmm not very happy with logic here, do you mind doing a self review of the changes and existing logic in this file and seeing how it can/should be simplified ?

client/packages/system/src/Report/api/hooks/useLatestReportList.ts
```diff
@@ -0,0 +1,114 @@
...
+    const v1 = v1Parts[i] || 0;
+    const v2 = v2Parts[i] || 0;
+
+    if (v1 > v2) return 1;
+    if (v1 < v2) return -1;
+  }
+  return 0;
+};
+
+export const useLatestReportList = ({
```
andreievg, line 33: Front end should be report version agnostic, this should be back end logic

server/service/src/standard_reports.rs
fergie-nz: Hi Andrei,

I was wondering about just upserting all reports regardless of if they were older versions than the one existing on the server. 
Is this what you're thinking? Or are you thinking refactoring but keeping the current logic?

server/repository/src/db_diesel/report_row.rs
```diff
@@ -110,19 +130,6 @@ impl<'a> ReportRowRepository<'a> {
...
-        version: &str,
-    ) -> Result<Option<ReportRow>, RepositoryError> {
-        let result = report_dsl::report
-            .filter(report_dsl::code.eq(code))
-            .filter(report_dsl::version.eq(version))
-            .first(self.connection.lock().connection())
-            .optional()?;
-        Ok(result)
-    }
-
```
fergie-nz, line 125: Removing as not anymore

server/repository/src/mock/reports.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+pub fn mock_report_j() -> ReportRow {
+    inline_init(|r: &mut ReportRow| {
+        r.id = "mock_report_j".to_string();
+        r.version = "3.0.1".to_string();
+        r.is_custom = false;
+        r.code = "report_with_custom_option".to_string();
+    })
+}
```
fergie-nz, line 93: Added extra cases to test for prioritising is_custom reports

server/service/src/report/report_service.rs
```diff
@@ -325,16 +324,88 @@ fn get_report(ctx: &ServiceContext, id: &str) -> Result<Report, RepositoryError>
...
-    let pagination = get_default_pagination(pagination, MAX_LIMIT, MIN_LIMIT)?;
     let filter = filter
         .unwrap_or_default()
         .r#type(ReportType::OmSupply.equal_to());
-    Ok(repo.query(pagination, Some(filter.clone()), sort)?)
+
+    let reports_to_show_meta_data = report_filter_method(
+        repo.query_meta_data(Some(filter.clone()), None)?,
+        app_version,
+    );
```
fergie-nz, line 340: Added querying all for metadata first with a sub selection of code, version, id, and is_custom. I thought this would help if our report list gets long and we aren't wanting to query all reports and filter out later.

server/service/src/report/report_service.rs
```diff
@@ -325,16 +324,88 @@ fn get_report(ctx: &ServiceContext, id: &str) -> Result<Report, RepositoryError>
...
+            .into_iter()
+            .map(|r| r.id)
+            .collect(),
+    ));
+
+    let reports = repo.query(Some(filter), sort)?;
+    Ok(reports)
+}
+
+fn report_filter_method(
```
fergie-nz, line 353: I have a feeling this logic could be more concise. However, after taking some time to think about it, I figure it is not worth spending time on now given we are only passing around report_meta_data

server/repository/src/db_diesel/report.rs
```diff
@@ -1,15 +1,14 @@
...
+    form_schema_row::{self, form_schema::dsl as form_schema_dsl},
+    report_row::report::{self, dsl as report_dsl},
+    ContextType, ReportMetaDataRow, ReportRow, ReportType, StorageConnection,
 };
 
 use crate::{
     diesel_macros::{apply_equal_filter, apply_sort_no_case},
     schema_from_row, FormSchema, FormSchemaRow,
 };
-use crate::{EqualFilter, Pagination, Sort, StringFilter};
```
fergie-nz, line 12: I have removed pagination as we won't use it for these queries.

server/service/src/report/report_service.rs
```diff
@@ -325,16 +324,88 @@ fn get_report(ctx: &ServiceContext, id: &str) -> Result<Report, RepositoryError>
...
+    let app_version: Version = Version::from_package_json();
+
     let repo = ReportRepository::new(&ctx.connection);
-    let pagination = get_default_pagination(pagination, MAX_LIMIT, MIN_LIMIT)?;
     let filter = filter
         .unwrap_or_default()
         .r#type(ReportType::OmSupply.equal_to());
-    Ok(repo.query(pagination, Some(filter.clone()), sort)?)
+
+    let reports_to_show_meta_data = report_filter_method(
```
andreievg, line 337: I think it's better to return report_ids to show (that feels a bit more clear to me), maybe it's personal preference, how does it sound to you ?

server/service/src/report/report_service.rs
andreievg: > Added querying all for metadata first with a sub selection of code, version, id, and is_custom. I thought this would help if our report list gets long and we aren't wanting to query all reports and filter out later.

That's awesome, thank you

server/repository/src/db_diesel/report.rs
```diff
@@ -118,16 +128,34 @@ impl<'a> ReportRepository<'a> {
...
+            match sort.key {
+                ReportSortField::Id => {
+                    apply_sort_no_case!(query, sort, report_dsl::id);
+                }
+                ReportSortField::Name => {
+                    apply_sort_no_case!(query, sort, report_dsl::name);
+                }
+            }
+        }
+        Ok(query
```
andreievg, line 155: That's so good !

server/service/src/report/report_service.rs
andreievg: This is great, really easy to read and performant, thank you. There is a lot to like about this, abstraction for readability, the `reports_of_code and `custom_reports_of_code` reduces the cognitive load (one step at a time)

The only improvement I can think of, the query_meta_data could return a domain object, with a little bit of mapping after `.load`, to get the version number (from_str) and store it into that domain object (reduce mapping in compare_major_minor and in find_latest_report)

andreievg (approved):
Looking very good, a joy to review. At some point we need to document this a bit more, wondering if it should be done inline in code, or as a separate `.md` file

server/service/src/report/report_service.rs
fergie-nz: Yes I like that too - will amend

fergie-nz:
> Looking very good, a joy to review. At some point we need to document this a bit more, wondering if it should be done inline in code, or as a separate `.md` file

Cheers!

I am thinking in report .md file might be good? It documents a lot of other report editing how-tos, which would be accompanied by report version changes.

server/service/src/report/report_service.rs
fergie-nz: 1ce0ac42e7de686b25dbad610cf2a192f3a90f3a

server/service/src/report/report_service.rs
```diff
@@ -329,12 +330,56 @@ fn query_reports(
...
+    let app_version: String = Version::from_package_json().to_string();
+
     let repo = ReportRepository::new(&ctx.connection);
     let pagination = get_default_pagination(pagination, MAX_LIMIT, MIN_LIMIT)?;
     let filter = filter
         .unwrap_or_default()
         .r#type(ReportType::OmSupply.equal_to());
-    Ok(repo.query(pagination, Some(filter.clone()), sort)?)
+
+    let all_reports = repo.query(pagination, Some(filter.clone()), sort)?;
```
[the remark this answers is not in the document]
fergie-nz: 516f705324ee4739ccbe8a02dd7d7fe9fb32c7c7

fergie-nz:
Have added better explanation in .md file, and added from_domain mapping to remove need for version mapping in meta data.

server/repository/src/db_diesel/report_row.rs
```diff
@@ -93,6 +93,26 @@ impl Default for ReportRow {
     }
 }
 
+#[derive(Clone, Insertable, Queryable, Debug, PartialEq, Eq, AsChangeset, Selectable)]
```
roxy-dao, line 96: You can default here so you don't need to impl default (:

server/service/src/report/report_service.rs
```diff
@@ -339,21 +339,13 @@ fn query_reports(
         app_version,
     );
 
-    let filter = ReportFilter::new().id(EqualFilter::equal_any(
-        reports_to_show_meta_data
-            .into_iter()
-            .map(|r| r.id)
-            .collect(),
-    ));
+    let filter = ReportFilter::new().id(EqualFilter::equal_any(reports_to_show_meta_data));
```
andreievg, line 342: this variable name doesn't make that much sense now

server/service/src/report/report_service.rs
```diff
@@ -1151,27 +1143,57 @@ mod report_filter_test {
         let mut app_version = Version::from_str("2.3.00");
         let mut result = report_filter_method(reports.clone(), app_version);
         assert_eq!(result.len(), 1);
-        assert_eq!(result.first().unwrap().version, "2.3.5");
+        let mut report = reports
+            .clone()
+            .into_iter()
+            .filter(|r| r.id == result.clone().into_iter().next().unwrap())
```
andreievg, line 1145: result.first/pop().unwrap()

server/repository/src/db_diesel/report.rs
```diff
@@ -21,6 +22,25 @@ pub struct Report {
     pub argument_schema: Option<FormSchema>,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct ReportMetaData {
```
andreievg, line 26: our domain object are composites atm, we don't really copy them (a bit less mapping and less to change when new fields are added)

andreievg: So should just have Version as another key (of type Version)

server/service/src/report/report_service.rs
```diff
@@ -345,12 +345,10 @@ fn query_reports(
...
 
-fn report_filter_method(reports: Vec<ReportMetaDataRow>, app_version: Version) -> Vec<String> {
-    let reports_with_compatible_versions: Vec<ReportMetaDataRow> = reports
+fn report_filter_method(reports: Vec<ReportMetaData>, app_version: Version) -> Vec<String> {
+    let reports_with_compatible_versions: Vec<ReportMetaData> = reports
         .into_iter()
-        .filter(|r| {
-            compare_major_minor(Version::from_str(&r.version), &app_version) != Ordering::Greater
-        })
+        .filter(|r| compare_major_minor(r.version.clone(), &app_version) != Ordering::Greater)
```
andreievg, line 351: why clone ?

andreievg (approved):
looks great thanks

PR #5539, opened 2024-11-24
Add check for if extra fields in requisition showing

### LINKED ISSUE #5538: Shouldn't check for requested != suggested if store preferences for requisition extra fields is off
## What went wrong? 😲

Show extra fields on requisitions not set causes error when Suggested quantity differs from requested quantity.

In back end, we should only throw an error on differs IF show extra fields store preference is set.

fergie-nz:
Fixes #5538

# 👩🏻‍💻 What does this PR do?

Adds not checking for requested == suggested when extra fields aren't set up in store preferences.

# 🧪 Testing

- [ ] Set show extra response requisition fields to false
- [ ] create a response requisition
- [ ] set requested != suggested on a line
- [ ] confirm finalised
- [ ] See that status change works
- [ ] Set show extra response requisition fields to true
- [ ] Create a new response requisition
- [ ] Set requested != suggested
- [ ] Confirm finalised should fail

# 📃 Documentation

### DIFF
--- a/server/service/src/requisition/response_requisition/update.rs
+++ b/server/service/src/requisition/response_requisition/update.rs
@@ -5,6 +5,7 @@ use crate::{
         query::get_requisition,
     },
     service_provider::ServiceContext,
+    store_preference::get_store_preferences,
 };
 use chrono::Utc;
 use repository::{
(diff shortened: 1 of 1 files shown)


roxy-dao (approved):
Whoo thanks Ferg

server/service/src/requisition/response_requisition/update.rs
```diff
@@ -103,7 +104,9 @@ pub fn validate(
         RequisitionLineFilter::new().requisition_id(EqualFilter::equal_to(&requisition_row.id)),
     )?;
 
-    if requisition_row.program_id.is_some() {
+    if requisition_row.program_id.is_some()
+        && get_store_preferences(connection, &store_id)?.extra_fields_in_requisition
+    {
```
roxy-dao, line 109: Opps, might want to check that RequisitionLineVariance reasons are provided too

roxy-dao (changes requested):
Opps too trigger happy

PR #5547, opened 2024-11-25
5182 add logging to reports

### LINKED ISSUE #5182: Add logging function to report wasm functions
Logging in wasm functions would be very helpful for trouble shooting.

This should come out of the box with wasi (https://github.com/extism/js-pdk), but isn't working.

We could either get this logging working, or alternatively insert a logging function in as per how we are doing with sql functionality.

Get logging working fro reports

fergie-nz:
Fixes #5182 

# 👩🏻‍💻 What does this PR do?

Adds more verbose logging for ease of tracing reports.
Adds println to expose javascript logs from within wasm functions for more easy debugging.
console.log, console.error, and console.info are all exposed as println! in the server terminal.

# 🧪 Testing

- [ ] Add a console.log to the wasm function
- [ ] Watch it show up when running the wasm function

# 📃 Documentation

### DIFF
--- a/server/cli/src/cli.rs
+++ b/server/cli/src/cli.rs
@@ -503,7 +503,7 @@ async fn main() -> anyhow::Result<()> {
                     let form_schema_json = match (arguments_path, arguments_ui_path) {
                         (Some(_), None) | (None, Some(_)) => {
                             return Err(anyhow!(format!(
-                                "When arguments path are specified both paths must be present in report {:?} {:?}", name, version_dir
+                                "When arguments_path is specified arguments_ui_path must be specified too in report {:?} {:?}", name, version_dir
                             )))
                         }
                         (Some(arguments_path), Some(arguments_ui_path)) => {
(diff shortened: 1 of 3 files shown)


server/cli/src/cli.rs
```diff
@@ -496,9 +502,9 @@ async fn main() -> anyhow::Result<()> {
 
                     let form_schema_json = match (arguments_path, arguments_ui_path) {
                         (Some(_), None) | (None, Some(_)) => {
-                            return Err(anyhow!(
-                                "When arguments path are specified both paths must be present"
-                            ))
+                            return Err(anyhow!(format!(
+                                "When arguments path are specified both paths must be present in report {:?} {:?}", name, version_dir
```
jmbrunskill, line 506:
```suggestion
                                "When arguments_path is specified arguments_ui_path must be specified too in report {:?} {:?}", name, version_dir
```

jmbrunskill (approved):
Looks good to me. Haven't tried building the reports.
I assume you discussed with @andreievg about changing to `println`

server/cli/src/cli.rs
jmbrunskill: Thought this would be more helpful than the old message...

fergie-nz:
> Looks good to me. Haven't tried building the reports. I assume you discussed with @andreievg about changing to `println`

We have yet to discuss. However the inbuilt logging with info wasn't working for me when I tried, so opted for the println

PR #5549, opened 2024-11-25
Update AssetNumberAlreadyExists error message

### LINKED ISSUE #5467: iSC/Cold chain/GAPS: Inconsistent error messages when importing duplicated assets
## What went wrong? 😲

Scenario: I have 14 assets in my store, and I exported a list of them. I then imported that same list into omSupply to test what error messages I get. 

Screenshot 2024-11-19 at 2 41 56 PM

Result: inconsistent error messaging. 

1. Even though the list of assets I am importing is identical to what I already have in my store, I only get the 'duplicated' warning for some?

Screenshot 2024-11-19 at 2 39 04 PM

2. Even when importing the same asset with identical information, one is getting an error and another one doesn't

3. Should the system recognise the user is trying to upload identical assets and only upload one of them instead of both?

## Expected behaviour 🤔

1. The errors should be consistent
2. The errors should be consistent
3. It would make sense for omSupply to notify users their data has duplicated and offer to only upload one, if two assets are identical

## How to Reproduce 🔨

1. Go to cold chain -> equipment
2. Have a few assets set up (e.g. maybe 5 or 10 for testing - create both catalogue and non-catalogue types)
3. Export that list into a CSV file
4. Click Import -> upload the file you just downloaded
5. See error display inconsistent - even though you have _all_ the assets already existing in your store, you will only get the message for some

## Your environment 🌱

- Database type: SQLite3

noel-yeldos:
Fixes #5467 

# 👩🏻‍💻 What does this PR do?
Display appropriate error message when user attempts to import existing asset 

From this:
image

To this:
image

# 🧪 Testing

- [ ] Go to cold chain -> equipment
- [ ] Have a few assets set up (e.g. maybe 5 or 10 for testing - create both catalogue and non-catalogue types)
- [ ] Export that list into a CSV file
- [ ] Click Import -> upload the file you just downloaded
- [ ] If there are are non-catalogue items, remove these from the import and try again
- [ ] If there are multiple duplicates of the same asset in the import file itself, remove so that there is only 1 item with that asset number (otherwise you get error "Asset Number already exists in import file"
- [ ] After cleaning the import file to the above specifications, import again to see the appropriate error message

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/asset/insert.rs
+++ b/server/service/src/asset/insert.rs
@@ -96,7 +96,7 @@ pub fn validate(
     }
 
     if let Some(asset_number) = &input.asset_number {
-        if check_asset_number_exists(asset_number, connection)?.len() == 1 {
+        if check_asset_number_exists(asset_number, connection)?.len() >= 1 {
             return Err(InsertAssetError::AssetNumberAlreadyExists);
         }
     }


jmbrunskill (approved):
Thanks @noel-yeldos looks good to me

Issue #4642, opened 2024-08-21
Feature: EIR / Vaccination Card


lache-melvin:
# Vaccination Card

[Spec document](https://docs.google.com/document/d/1CejT3U8jOkdb-lnGfw5eVacuRtRImoLfHNCSonNQNVU/edit)

#### Create and Maintain Immunisation Programs
- [x] #4652 
- [x] #4653 
- [x]  #4654
- [x]  #4655 
- [x]  #4656 
- [x]  #4657

#### Vaccination Card

- [x]  #4660
- [x] #4661
- [ ]  #4663
  - [x] #4665
  - [x] #4737
  - [ ] Delivered dose modal links to related encounter
  - [ ] Add a progress bar summarising the immunisation schedule
- [ ] Capture adherence against each immunisation schedule
- [ ] Capture patient withdrawal - program withdrawn -> adherence shows this?
- [x] #4675
  - [x] #4793
  - [ ] Highlight the period of the patient's age
  - [ ] Highlight doses the patient is ready to receive in this encounter (in the period(s) that a patient is currently within, if min interval has passed)
  - [ ] Hide past periods, unless a vaccination was missed in that period
  - [x] #4710
    - [x] #4711
    - [x] Batch not required if encounter date is in the past
    - [x] Or if not given, capture reason (refused/out of stock)
  - [x] #4713

#### Encounters changes
- [x] #4712
- [x] #4709
- [x] #4714
- [ ] Delete vaccinations
- [ ]  Delete vaccinations when encounter is deleted
- [ ] When finalising, create next scheduled encounter
- [x] #4796 
- [x] Default encounter date to today
- [ ] When pulling a patient from central, also pull encounter data
  - [ ] Sync encounter data between sites (if a patient is at multiple)
- [x] Link to the created prescription from an encounter
- [x] Capture/show facility on encounter
- [ ] Can create an encounter from another facility

#### Patient Management improvements (if time)
- [ ] New patient creation in one single form
- [ ] Add Enrolment in program to creation form
- [ ] When there is only one program, it should be default selected during enrolment
- [ ] Improve the patient search flow when creating patients
- [ ] Remove patients from a facility (transfer to another facility/they move overseas)
- [ ] Delete a patient (if no transactions)

 
### Moneyworks Jobcode 🧰

[DJIGVCDV]

# Testing

### Setup
- [ ] Configure some vaccine items in mSupply
- [ ] Configure an immunisation program in mSupply
[...]

PR #5551, opened 2024-11-25
#5264 JSON Forms prescription renderer

### LINKED ISSUE #5264: Create JSON Form prescription renderer

CarlosNZ:
Fixes #5264

# 👩🏻‍💻 What does this PR do?

Implements a component to create enter prescription data and save a prescription invoice when the form is saved.

Have used the mechanism demonstrated in #5394 to implement the actions, but have refactored into its own hook (`useFormActions`) which also includes the ability to hold some "intermediate state" -- i.e. data that we don't want to actually save to the JSON form document, but need to keep around (in this case for when the prescription is created). 

Each instance of the component can only select a single item (with multiple batches). However, there is only one prescription created for the whole form. So you can add different items at different form sections (e.g. Contraceptive method, General Exam) and you'll only generate a single prescription on save.

The only piece of data actually saved back to the JSON form is the Invoice ID of the new prescription. When this component loads, it looks up the prescription and displays accordingly:
- if the prescription exists, we just show a link to the existing prescription -- any further editing must be done there for now:
Screenshot 2024-11-25 at 4 15 19 PM
- if the prescription doesn't exist, display the Item selector and Stockline inputs:
Screenshot 2024-11-25 at 4 09 36 PM

Still to do (probably in a subsequent PR):
- [ ] Validation for the quantity inputs (currently doesn't stop you entering more than available quantity, might need additional checks too)
- [ ] Only show items that belong to a specified item category (this would be an option -- some implementations might want access to the full list)
- [ ] General UI improvements based on design feedback

The end result here is not that complicated, but it took a while to get there as the solution to handle actions in a way that allows for the intended use cases (as well as being flexible enough to handle future Action implementations for other components) was a bit of a convoluted journey! 😆

See inline comments for more detailed explanation of specific elements.

# 🧪 Testing

First you need to set up the JSON schemas from the [program schemas repo](https://github.com/openmsupply/programschemas). These schemas are just minimal implementations to demonstrate and test the functionality here. Full Djibouti schemas are still in progress:
- The branch is called "djibouti_family_planning"
- Build the JSON schema for "djibouti" -> "Reproductive Health Program" and "Reproductive Health Encounter"
- Upload the generated schemas and their equivalent UI schemas to mSupply and sync to OMS
- The details for doing that are the same as in the Testing section of [this issue](https://github.com/msupply-foundation/open-msupply/issues/4642), but come and chat to me if you need a bit more guidance getting it set up

Then enrol one or more patients in the Reproductive Health program and start an Encounter. You should see an Encounter UI like this:
Screenshot 2024-11-25 at 4 33 00 PM

- [ ] Open "Contraceptive Method", select options for the "Method" selectors
- [ ] Start creating a prescription using the Item/Stock inputs
- [ ] Close form, the "Save" button should be enabled (but don't save just yet)
- [ ] Open the "Additional Prescription Item" modal, and add another item
- [ ] Now "Save" the form. You should see a Success notification for the save and for a new prescription created
- [ ] Open one of the above modals again. It should now display "Click to view details" instead of prescription inputs
- [ ] Click the link, and check the prescription opens in a new tab and that the contents of the prescription match what you entered earlier

Additional things to try:
- [ ] Delete a prescription and check that you can now create a new one for that encounter
- [ ] Ensure that the form becomes saveable (i.e. the Save button is clickable) after just entering prescription info (on the Contraceptive tab, the first three selectors are compulsory, so you won't be able to save until those are entered, so either do that in a separate operation, or use the "Additional Prescription" tab)
- [ ] Check that the Batches list changes when different items are selected
- [ ] Check that the state persists when closing the modals and re-opening them (the selected item and the amounts to prescribe)


client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -138,9 +143,13 @@ export const useJsonForms = <R,>(
 
     // Run mutation...
     try {
+      await formActions.run({ preSubmit: true });
+
       const sanitizedData = stripEmptyAdditions(initialData, data);
       const result = await save?.(sanitizedData);
 
+      await formActions.run({ preSubmit: false });
+
```
CarlosNZ, line 152: As suggested earlier, we now have the option to specify actions as "pre-submit" or "post-submit".

In future we might want to do some more error handling, and abort the save if any (or all?) of the pre-submit actions fail. But we need more real world use cases before that becomes worthwhile.

client/packages/programs/src/JsonForms/useFormActions.tsx
```diff
@@ -0,0 +1,86 @@
+/**
+ * Some JSONForm components can have actions associated with them, which are
+ * executed when the whole form is saved/submitted. This hook provides a
+ * mechanism to "register" such actions. They can be defined to be run either
+ * *before* or *after* the main "Save" action (in useJsonForms)
+ *
+ * This hook is only called from `useJsonForms` and the "formActions" object it
+ * returns is passed to all form components on the main `config` prop. This
+ * ensures all form components are sharing a comment Actions registry.
```
CarlosNZ, line 9: Another thought was to create this as a Zustand store and get components that need it to access it directly via a hook, rather than passing down through JsonForms component.

lache-melvin:
Could you rebase to 2.4.1 @CarlosNZ? 🙏

CarlosNZ:
> Could you rebase to 2.4.1 @CarlosNZ? 🙏

I've done so locally, but I'll need to make a new PR with just those changes, otherwise we end up with additional `develop` changes in here. I'm working on the correct branch for #5560, so probably best if you just review this one and I'll make any changes in the other branch. Unless you have better suggestion?

lache-melvin (approved):
Looking epic Carl! Thanks for all the mahi 😁 

Once base updating and tests pass, happy for this to go in!

Agree with all of your next steps, and yep sounds good to put them in follow up PRs! 🙏 

Only thing I found was with the 2nd prescription - I could click save, but it wouldn't create the prescription. I think this was actually because the reproductive health fields were required and not filled in... so I think a JSON form configuration thing, rather than a code issue! Particularly given that this use case is for only one prescription line to be created in the encounter, let's not worry too much about it for now 😁

CarlosNZ:
> Only thing I found was with the 2nd prescription - I could click save, but it wouldn't create the prescription. I think this was actually because the reproductive health fields were required and not filled in... so I think a JSON form configuration thing, rather than a code issue! Particularly given that this use case is for only one prescription line to be created in the encounter, let's not worry too much about it for now 😁

This was a very subtle JSON Forms bug. I've fixed it in https://github.com/msupply-foundation/open-msupply/pull/5570 -- the chain of events leading to this behaviour is very convoluted so I'll explain it to you IRL if you're interested.

Issue #5514, opened 2024-11-21
Difficult to know which item variant to select in inbounds hipment


jmbrunskill:
Part of https://github.com/msupply-foundation/open-msupply/issues/5432

## What went wrong? 😲
<img width="999" alt="image" src="https://github.com/user-attachments/assets/78d68152-2f3b-40fa-aa66-e49c5b7d294f">

## Expected behaviour 🤔

The item variant selector should make it possible to see other distinghuising information such as manufacturer

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create item variants with similar names
2. Create an inbound shipment
3. See it's hard to choose the right item variant

## Your environment 🌱

- Open mSupply Version: 2.4.1-RC2
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

adamdewey:
Thanks @jmbrunskill - I guess that workaround works for now!

But if we want to support the industry properly at some future stage then we should probably handle the `Name` and `Manufacturer` fields robustly?

mark-prins:
Would it be possible to enable `autoWidth` on the [select component](https://mui.com/material-ui/react-select/#auto-width)
It's hard to read when all squashed up like that.

jmbrunskill:
> Would it be possible to enable `autoWidth` on the [select component](https://mui.com/material-ui/react-select/#auto-width) It's hard to read when all squashed up like that.

I had a quick look but we're using AutoComplete which doesn't have the same option. Can leave your comment here for when we look at this... I think we'll need to customise the Popper Component. Or I guess we can probably switch to a Select component? I think the intention is to not have too many variants for an item so that's probably a good option?

mark-prins:
ah. in that case, even easier. Just set the `popperMinWidth` on the `Autocomplete`, as per something like this:

<img width="524" alt="image" src="https://github.com/user-attachments/assets/d399c4ac-2ec7-48a8-9b8d-24da56274f0b">

😄

mark-prins:
@jmbrunskill we now have manufacturer displayed

<img width="1041" height="532" alt="Image" src="https://github.com/user-attachments/assets/61fbb397-e898-4e6a-9d73-fade6b8e2c27" />

can we close this issue?

PR #5552, opened 2024-11-25
Stocktake item variants

### LINKED ISSUE #5515: Allow setting item_variant during stocktake
Part of https://github.com/msupply-foundation/open-msupply/issues/5432

During stock take, there's no way to set an item_variant on a stock_line
image

The same item_variant selector we have added to inbound shipment

jmbrunskill:
Fixes #5515

# 👩🏻‍💻 What does this PR do?

image

Allows the user to set item variants during stocktakes

The very edge of `Reason` is being cut off, not clear to me why.
Given timeframe I've left this for now, can be cleaned up later??

@mark-prins are you ok with `item_variant` appearing nex to batch?

Inbound looks like this, ordering isn't consistent so wasn't obvious where it should go

image

# 🧪 Testing

- [ ] Setup item_variants on your central OM server
- [ ] Try stocktaking items and assign item variants to them
- [ ] Check that the right variant shows up in View Stock after Finalising
- [ ] Check that different combinations of updates work, e.g. adding additional stock, equal stock but with the new variant, new stock lines and batches with and without variants.
- [ ] Check that if they already have an item variant set on the stock line that it populates when creating a stocktake.

# 📃 Documentation

  1.
  2.


server/service/src/sync/translations/stock_line.rs
```diff
@@ -44,6 +44,7 @@ pub struct LegacyStockLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string", rename = "barcodeID")]
     pub barcode_id: Option<String>,
     #[serde(rename = "om_item_variant_id")]
+    #[serde(default)]
```
jmbrunskill, line 47: I think I added this here by accident, but I think it doesn't hurt to allow this? Otherwise potentially could be an compatibility issue with older mSupply?

server/service/src/sync/translations/stocktake_line.rs
```diff
@@ -48,6 +48,11 @@ pub struct LegacyStocktakeLineRow {
     #[serde(rename = "optionID")]
     #[serde(deserialize_with = "empty_str_as_option_string")]
     pub inventory_adjustment_reason_id: Option<String>,
+
+    #[serde(rename = "om_item_variant_id")]
+    #[serde(deserialize_with = "empty_str_as_option_string")]
+    #[serde(default)]
+    pub item_variant_id: Option<String>,
```
jmbrunskill, line 55: mSupply iss created to handle this field
https://github.com/msupply-foundation/msupply/issues/15618

mark-prins:
For the reason input being truncated, it's because the column width is being used to set the input width.. but the column has some padding in it. So really the width of the input should be less than the column!

line 132 in `StocktakeLineEditTables.tsx` could be this:

```
          width={Number(column.width) - 16}
```

to resolve in a clumsy way.. it will show you what I mean anyway.

and for the other question: yep, item variant after batch is fine. the input is very narrow though, and with the text truncated it is hard to see which variant is actually selected. that's a minor thing, not sure how to fix that.

server/service/src/stocktake/update/generate.rs
```diff
@@ -114,6 +114,7 @@ fn generate_stock_in_out_or_update(
             cost_price_per_pack,
             sell_price_per_pack,
             expiry_date,
+            item_variant_id: stocktake_line.line.item_variant_id.clone(),
```
lache-melvin, line 117: Should this not follow the same pattern as the other fields?

```rust
let item_variant_id = row.item_variant_id.or(stock_line_row.item_variant_id)
```

Though it appears to be retaining the existing stock line data fine, because the initial stocktake line is created based on the stock line data in the frontend - but I think backend should continue to support partial updates

client/packages/system/src/Item/Components/ItemVariantSearchInput/ItemVariantSearchInput.tsx
```diff
@@ -30,7 +30,7 @@ export const ItemVariantSearchInput = ({
     <Autocomplete
       disabled={disabled}
       width={`${width}px`}
-      popperMinWidth={Number(width)}
+      popperMinWidth={200}
```
lache-melvin, line 33: Was there a reason you changed this? Now getting this in bundled items:

Screenshot 2024-11-26 at 11 26 45 AM

But I think it was working ok? Popper expands to whatever width is needed to show the variant names?

lache-melvin (approved):
LGTM James - just the little fixup with keeping the existing variant would be nice 🙏

client/packages/system/src/Item/Components/ItemVariantSearchInput/ItemVariantSearchInput.tsx
jmbrunskill: Dang, this was based on Mark's suggestion here: https://github.com/msupply-foundation/open-msupply/issues/5514#issuecomment-2496256771

jmbrunskill: Forgot to Comment with the reference, I'll do a min of the two maybe?

server/service/src/stocktake/update/generate.rs
jmbrunskill: I think it we do that we wouldn't be able to clear the `item_variant_id` with a stock take.
Maybe that's a good trade off? 

I'm personally not sure how partial updates should/could work with stock take?
If we leave `stocktake_line` columns null for things we don't want to change, how do we know we've captured all the data?


@andreievg thoughts?

I've updated for now, to just add a comment explaining...

client/packages/system/src/Item/Components/ItemVariantSearchInput/ItemVariantSearchInput.tsx
jmbrunskill: image

server/service/src/stocktake/update/generate.rs
jmbrunskill: Added this, which I think alleviates the concern that item_variant it might not be populated in stocktake_line. https://github.com/msupply-foundation/open-msupply/pull/5552/commits/6bf227193ace8efa5baea5dd23e0cd1a2a01e7bc

I think I'll go ahead and merge, but please feel free to create another issue if we think further cleanup is required...

Issue #5447, opened 2024-11-18
iSC/Cold Chain/GAPS: Equipment Installation Date follows a different format to the rest of the dates in omSupply


marthakeezy:
## What went wrong? 😲

Cold chain -> Equipment: the installation dates follow a different format to the rest of the dates in omSupply - could we please standardise the appearance?

How it looks now: **YYYY-MM-DD**

<img width="723" alt="2" src="https://github.com/user-attachments/assets/6d6cddc1-39bb-4d2e-b05f-ec91719a5ac9">

How it should look: **DD-MM-YYYY**

<img width="440" alt="1" src="https://github.com/user-attachments/assets/26f13bc2-1095-4a34-ac23-354a3397290e">

## Expected behaviour 🤔

Should be `DD-MM-YYYY`

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to Cold chain -> Equipment
2. Create a new asset or modify an existing one
3. Under Installation Date, select a Date (note - at that moment, the date appears normally)
4. Navigate back to the overall view of all assets - see column Installation Date now shows the date as YYYY-MM-DD
5. Inconsistency!

## Your environment 🌱

- Open mSupply Version: V2.4.0-RC1
- Legacy mSupply Central Server Version: V7.18.11

- Platform: Google Chrome (browser on Mac) + Samsung Galaxy Tab A8 (Android)

- Database type: SQLite3

marthakeezy:
<img width="1427" alt="Screenshot 2024-11-20 at 11 04 21 AM" src="https://github.com/user-attachments/assets/c8fc0422-72e4-4765-be23-3f5f11bc84d0">

Same spotted when importing cold chain assets into omSupply

noel-yeldos:
@marthakeezy I have a PR to fix first issue! Could you please give me steps to recreate the import related issue?

marthakeezy:
# 🧪 Testing



`V2.4.0-RC3`

- [ ] Go to Cold chain -> Equipment
- [ ] Create a new asset or modify an existing one
- [ ] Under Installation Date, select a Date
- [ ] Navigate back to the list view of all assets - column Installation Date now shows the date as DD/MM/YYYY

marthakeezy:
Hey Noel, do you think you could please take a look at this? 

In V2.4.1-RC2, Warranty Start & End dates follow a different format

[...]

PR #5553, opened 2024-11-25
5481 asset import missing properties

### LINKED ISSUE #5481: iSC/Cold chain/GAPS: some asset attributes are not configured/missing from export/import functionality
## What went wrong? 😲

I was playing around with importing/exporting cold chain assets, and noticed some attributes were not included in the import template or when exporting a CSV file -> therefore the user would have to go through each asset manually to update the missing details. 

**The missing ones are:** 

- Store
- Category
- Type
- Warranty Start
- Warranty End
- Current Status
- Reason
- Needs Replacement (Y/N)
- Donor

**Enhancements:** 

- Can lifespan column be placed after Refrigerant Type(s) just as it appears in omSupply? Currently, it's the last column in the template/export file which does not make sense to me
- Temperature Monitoring Device appears twice in the file

Screenshot 2024-11-20 at 10 56 31 AM

- Catalogue Item code appears in export and is editable in import template, however does not appear in omSupply (front-end wise?). Should this info be visible to the users? Possibly needs @adamdewey's input

## Expected behaviour 🤔

If the field is editable/fillable by the user, it should be in the import template & also exported, otherwise the user will have to manually update what's missing

## How to Reproduce 🔨

1. Go to cold chain -> assets
2. Create a new asset and fill out all the details (e.g. cold chain location, measurements, donor etc)
3. Export the list of assets - see some columns are missing, no way for the user to export some information
4. When uploading assets, download the template provided - notice the same columns are missing from the template
5. Inconsistency!

## Your environment 🌱

- Database type: SQLite3

lache-melvin:
Fixes #5481 

# 👩🏻‍💻 What does this PR do?

Adds columns to import/export of CCE Assets.

Doesn't include the category/type/manufacturer (as this is derived from the `Catalogue item code`

Doesn't set Donor - names would be harder to navigate at this time

# 🧪 Testing

- [ ] Have some CCE assets
- [ ] Export assets
  - [ ] warranty dates, current status and whether it needs replacing should come through
  - [ ] Properties should not be duplicated
- [ ] Click import asset
  - [ ] Download the import CSV
  - [ ] Fill it out with some new assets (at least one functioning, one with a different status, one that needs replacing)
  - [ ] Import the CSV
  - [ ] Assets import correctly (dates and statuses mapped! - note that if you don't set a status, it will be created as functioning)
  - [ ] Import again to create errors (already exists)
  - [ ] Export the errors CSV
  - [ ] Columns all look correct here
- [ ] Compare all the CSV files, columns should now all be in the same order
- [ ] Check all works on both central server and remote site (central should include store column, remote should not)

# 📃 Documentation

  1.
  2.


client/packages/coldchain/src/Equipment/ImportAsset/ImportReviewDataTable.tsx
```diff
@@ -29,24 +31,12 @@ export const ImportReviewDataTable: FC<ImportReviewDataTableProps> = ({
...
-      sortable: false,
-      label: 'label.asset-number',
-    },
-    {
-      key: 'catalogueItemCode',
-      width: 50,
-      sortable: false,
-      label: 'label.catalogue-item-code',
-    },
-  ];
```
lache-melvin, line 45: Sorry, little bit of jumping around - didn't get why everywhere was a different order

client/packages/coldchain/src/Equipment/ImportAsset/ImportPanel.tsx
```diff
@@ -4,6 +4,7 @@ import { styled, TabPanel, Box } from '@openmsupply-client/common';
 
 const StyledTabPanel = styled(TabPanel)({
   height: '100%',
+  width: '100%',
```
lache-melvin, line 7: Review table wasn't scrolling/overflowing correctly with added columns

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -210,13 +211,16 @@ export const EquipmentUploadTab: FC<ImportPanel & EquipmentUploadTabProps> = ({
...
         catalogueItemCode: '',
         store: undefined,
         notes: '',
         serialNumber: '',
-        installationDate: 'DD/MM/YYYY',
-        replacementDate: 'DD/MM/YYYY',
+        installationDate: t('label.date-format'),
+        replacementDate: t('label.date-format'),
+        warrantyStart: t('label.date-format'),
+        warrantyEnd: t('label.date-format'),
```
lache-melvin, line 222: realised these probably don't translate well!

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1152,7 +1153,7 @@
   "messages.cannot-view-vaccine-card": "You cannot view this patient's Vaccination Card from this store, because they were enrolled in the {{ programName }} by another store",
   "messages.cant-delete-generic": "You cannot delete one or more of the selected items",
   "messages.cant-delete-requisitions": "Can only delete requisitions with a status of 'Draft'",
-  "messages.cant-download-android": "File cannot be download on this device",
+  "messages.cant-download-android": "File cannot be downloaded on this device",
```
lache-melvin, line 1156: Side quest 😁

jmbrunskill (approved):
Happy to merge when you're ready...

I've found one bug of sorts though, spreadsheet program seems to love mangling date formats, so I think we should add a bit more validation around the dates.

Otherwise it's very easy to do this (i did it by accident)
image


image

image

I think we used to have it so that we had YYYY-MM-DD in the import. I think this is a much less ambiguous format for imports, I know it's not so common but I wonder if we should consider going back to it?

At least checking that we have 4 digits for the year would be good...

client/packages/coldchain/src/Equipment/ImportAsset/EquipmentImportModal.tsx
```diff
@@ -70,12 +75,30 @@ export const toInsertEquipmentInput = (
...
   return {
     ...rest,
+    needsReplacement: !!row.needsReplacement,
     catalogueItemId,
     store: store ? { ...store, __typename: 'StoreNode', storeName: '' } : null,
     parsedProperties,
   };
 };
 
+export const toStatusTypeInput = (status: StatusType): AssetLogStatusInput => {
```
jmbrunskill, line 85: Can we add these to the docs somewhere, so people know what options they have for that column?

noel-yeldos:
@jmbrunskill I tried re-importing the file and even in YYYY-MM-DD format it's still adding the leading zeroes. Is this what you what you were referring to?

Screenshot 2024-11-26 at 10 28 31 AM

FYI I had made a change to update the formatting in the import table to DD-MM-YYYY per issue #5447

jmbrunskill:
> @jmbrunskill I tried re-importing the file and even in YYYY-MM-DD format it's still adding the leading zeroes. Is this what you what you were referring to?
> 
> Screenshot 2024-11-26 at 10 28 31 AM
> 
> FYI I had made a change to update the formatting in the import table to DD-MM-YYYY per issue #5447

Yes, clearly that doesn't check properly either :)

It's all good I think would just be good to add a check because we can't just re-import to fix and could be a systematic error for 1000s of assets...

noel-yeldos (approved):

PR #5563, opened 2024-11-26
Bugfixes for asset locations

### LINKED ISSUE #5415: iSC/Cold chain/GAPS: buggy Cold Storage Location selection in Equipment
## What went wrong? 😲

Quite hard to describe, but I suspect (!) what the issue might be is when one asset is assigned to a location, the location selection disappears from the dropdown for other equipment entries? 

Also, when a location is removed for an asset, you cannot add it back in. For context, I have 4 different locations, none of them are on hold, each location has been used before for testing with equipment etc. 

Captured the bugs [in this video](https://drive.google.com/drive/folders/1FomYcW13X3xZeklPu8tCfgaNI_szHnrj)

## Expected behaviour 🤔

When we edit equipment, when one location is removed, it should still be in a selection even after removal -> when the location is selected, the number should go down to three, but then back to four if it's unassigned. Also more than one equipment can be assigned to a single location, so the option should not disappear for other entries. 

## How to Reproduce 🔨

1. Unclear! But what I've done while playing around: 
2. Go to Cold chain -> Equipment and assign a Cold storage location to an equipment you are testing
3. Go to another equipment and see if that same location is available to be assigned
4. When editing equipment, assign location -> save -> unassign it -> click on the dropdown again -> see that same location is not available to be selected again
5. Buggy behaviour!

## Your environment 🌱

- Database type: SQLite3

jmbrunskill:
Fixes #5415

# 👩🏻‍💻 What does this PR do?

Fixes issue with locations not being available to select when deleting and re-adding locations from assets.

2 problems
First, that after saving an asset, we didn't invalidate the location query and there for didn't see the locations as unassigned.
Second, if deleting a location, if before saving you wanted to add that back in you couldn't.

Pre-exising locations show up at the end of the list.
It might be nice if they were at the top, or we sorted them or something.
Doesn't feel important enough to spend time on to me...

# 🧪 Testing

See video in the issue to see how to re-create the issues.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
@@ -106,6 +106,8 @@ export const EquipmentDetailView = () => {
       value: location.id,
     })) || [];
 
+  // Any locations that are already assigned to the asset won't be returned by the query above
+  // So we add them in manually here...
   if (data && data?.locations.nodes.length) {
     const assignedLocations = data.locations.nodes.map(location => ({
       label: formatLocationLabel(location),
--- a/client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
@@ -100,11 +100,19 @@ export const EquipmentDetailView = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [data, setDraft]);
 
-  const locations =
+  let locations =
     locationData?.nodes.map(location => ({
       label: formatLocationLabel(location),
       value: location.id,
     })) || [];
+
+  if (data && data?.locations.nodes.length) {
+    const assignedLocations = data.locations.nodes.map(location => ({
+      label: formatLocationLabel(location),
+      value: location.id,
+    }));
+    locations.push(...assignedLocations);
+  }
 
   if (isLoading || isLoadingLocations) return <DetailFormSkeleton />;
 
--- a/client/packages/coldchain/src/Equipment/api/hooks/document/useAssetUpdate.ts
+++ b/client/packages/coldchain/src/Equipment/api/hooks/document/useAssetUpdate.ts
@@ -1,13 +1,18 @@
 import { useMutation, useQueryClient } from '@openmsupply-client/common';
 import { useAssetApi } from '../utils/useAssetApi';
 import { DraftAsset } from '../../../types';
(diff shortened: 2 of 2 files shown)


client/packages/coldchain/src/Equipment/api/hooks/document/useAssetUpdate.ts
```diff
@@ -1,13 +1,18 @@
...
 
 export const useAssetUpdate = () => {
   const queryClient = useQueryClient();
   const api = useAssetApi();
+  const locationApi = useLocationApi();
 
   return useMutation(async (asset: Partial<DraftAsset>) => api.update(asset), {
-    onSuccess: id => queryClient.invalidateQueries(api.keys.detail(id)),
+    onSuccess: id => {
+      queryClient.invalidateQueries(locationApi.keys.list());
```
jmbrunskill, line 13: This is a bit of an overkill to remove all lists, but don't want the additional complexity of trying to match two types of query keys here?

lache-melvin: Yeah I think this is fine 😃

client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
```diff
@@ -100,12 +100,20 @@ export const EquipmentDetailView = () => {
...
 
-  const locations =
+  let locations =
     locationData?.nodes.map(location => ({
       label: formatLocationLabel(location),
       value: location.id,
     })) || [];
 
+  if (data && data?.locations.nodes.length) {
+    const assignedLocations = data.locations.nodes.map(location => ({
```
lache-melvin, line 110: A comment might be helpful here - assigned ones are filtered out by API, then added here, then filtered by the dropdown - good to explain why we add here?

lache-melvin (approved):
LGTM! Just a comment might be nice 🙏

client/packages/coldchain/src/Equipment/DetailView/DetailView.tsx
```diff
@@ -100,12 +100,20 @@ export const EquipmentDetailView = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [data, setDraft]);
 
-  const locations =
+  let locations =
```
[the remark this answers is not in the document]
jmbrunskill: Fair enough, It doesn't help that `let` is immutable by default in rust 😓

PR #5565, opened 2024-11-26
add temperature range to item variants detail

### LINKED ISSUE #5410: iSC/Cold Chain/GAPS: Temperature selections should have ranges visible
## What went wrong? 😲

When editing Inventory -> Locations and setting a temperature, the selection should have ranges visible - at the moment the user cannot see the range they are choosing until they click on it (it only appears afterwards). The names won't always make sense, so it would be helpful for the user to see the ranges straight away

Screenshot 2024-11-14 at 8 09 53 PM

## Expected behaviour 🤔

Ranges should be visible :)

## How to Reproduce 🔨

1. Go to Inventory -> Locations
2. Choose an existing Location or create a new one
3. Click on the temperature field dropdown and see only names appear and not the ranges

## Your environment 🌱

- Platform: Google Chrome (browser for Mac) and Samsung Galaxy Tab8 (Android)

- Database type: SQLite3

aimee-mcneil-melville:
Fixes #5410

# 👩🏻‍💻 What does this PR do?

Allows the user to view the temperature range along with the cold storage type in an item variant in detail view

Screenshot 2024-11-26 at 3 27 14 PM

Is it worth setting a centralised definition for the temperature range display value? (Option renderer?)
It's used in three places, so would need to find and update them all separately if they need changing

# 🧪 Testing

- [ ] Go to Catalogue -> Items -> Select an item -> Variants tab -> Add Variant
- [ ] Create a new variant with a cold storage type selection
- [ ] Click OK
- [ ] See temperature range visible in the detail view

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
@@ -76,6 +76,10 @@ const ItemVariant = ({
   const t = useTranslation();
   const confirmAndDelete = useDeleteItemVariant({ itemId });
 
+  const coldStorageValue = variant.coldStorageType
+    ? `${variant.coldStorageType.name} (${variant.coldStorageType.minTemperature}°C to ${variant.coldStorageType.maxTemperature}°C)`
+    : null;
+
   return (
     <Box maxWidth="1000px" margin="25px auto" paddingBottom={6}>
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Looks good to me!

Can you raise a refactor issue to clean up the duplication?
Just go issue new then select refactor, and fill out the template.
Feels like it's probably a good thing to do, but also don't want to delay anything going into 2.4.

PR #5566, opened 2024-11-26
Basic spinner if current item isn't populated yet

### LINKED ISSUE #5562: Adding an item to an existing manual requisition results in a blank page
## What went wrong? 😲

When adding an item to a manual requisition (General, not Programs), the screen first appears blank and then when navigated away and back to the item, the detailed view appears

Video of the bug ->>> [here](https://drive.google.com/drive/folders/1XMzU62c-tqC-aSUZC52PdeD1UKPRXTAZ)

## Expected behaviour 🤔

The detailed view should be shown straight away

## How to Reproduce 🔨

2. Create a new manual requisition (important - navigate away from it, don't add the item straight away)
3. When back -> and prompted with choice between Program & General, choose General
5. Add an item -> see blank screen appear like in the video
6. Navigate away and back to the item you just added
7. Detailed view should now appear

## Your environment 🌱

- Open mSupply Version: `V2.4.0-RC3`

- Database type: `SQLite3`

roxy-dao:
Fixes #5562

# 👩🏻‍💻 What does this PR do?
Show spinner if current item isn't populated and NothingHere if the itemId isn't in the URL

https://github.com/user-attachments/assets/c2d338ed-39df-4a02-955f-3bbd5c5a59e9

# 🧪 Testing

- [ ] Throttle network
- [ ] Have an existing manual Requisition
- [ ] Go to that Requisition and add an item
- [ ] Edit page should load instead of showing nothing here

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
@@ -60,7 +60,7 @@ export const ResponseLineEditPage = () => {
   }, [currentItem]);
 
   if (isLoading || !currentItem) return <BasicSpinner />;
-  if (!data || !itemId) return <NothingHere />;
+  if (!data) return <NothingHere />;
 
   return (
     <>
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
@@ -37,19 +37,18 @@ const ResponseLineEditFormLayout = ({
 };
 
 export const ResponseLineEditPage = () => {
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Thanks Roxy seems good to me.
I was getting slightly different behaviour to the video where it was showing Nothing there, then updating after the query completed? 

This looks like it fixes it anyway...

I'm not clear why you're have no `item_id` probably isn't a realistic scenario?

An invalid `item_id` seems to return an infinite loading spinner but I think that's also Ok as it shouldn't really happen in general operation right?

roxy-dao:
> Thanks Roxy seems good to me. I was getting slightly different behaviour to the video where it was showing Nothing there, then updating after the query completed?
> 
> This looks like it fixes it anyway...
> 
> I'm not clear why you're have no `item_id` probably isn't a realistic scenario?
> 
> An invalid `item_id` seems to return an infinite loading spinner but I think that's also Ok as it shouldn't really happen in general operation right?

You would hope it doesn't happen in general operation XD I'll remove the item_id check...

Weird that shows the nothing here page on first load?

PR #5569, opened 2024-11-26
Fix item_variant selection on customer return

### LINKED ISSUE #5513: iSC/Cold chain/GAPS: unable to interact with Item Variant dropdown in Customer Returns on Android
## What went wrong? 😲

Unable to interact with Item Variant dropdown in Customer Returns on Android. When clicking on the field, the keyboard retracts and kicks you out of the selection - there is not enough time to click on any field presented :(

Video of the bug attached ->>> [here](https://drive.google.com/drive/folders/1RjCxXGs1WPowPF3X5u8ZTkcO9iQjRMER)

## Expected behaviour 🤔

Should be able to interact with the field and make a selection

## How to Reproduce 🔨

1. Go to Distribution -> Customer Returns
2. Create a new return or edit an existing one
3. Select an item (regardless if its variants have been configured or not - the behaviour is the same)
4. Attempt to interact with the Item Variant dropdown -> see Android behaviour

## Your environment 🌱

- Open mSupply Version: `V2.4.0-RC2`

- Platform: `Galaxy Tab A8 (Android 14)`

- Database type: `SQLite3`

jmbrunskill:
Fixes #5513

# 👩🏻‍💻 What does this PR do?

Move's batch to be beginning of the modal, this seems to help and is more consistent with other screens, although there's still some issues with tabing when changing item_variants

This is a deeply unsatisfying fix, someone better at React should probably help fix the underlying issue.
It seems like the reason this didn't happen on other screens is just luck that there's a default text component first in the modal so the keyboard height doesn't go haywire. Changing the order is good to do anyway, but we could run into this again I think unless we fix the underlying issue.

If you change the item_variant and hit tab, you don't tab to the next field in the table correctly.
I assume the component is being re-created for some reason?
I think it's worth getting this in 2.4 though even if it just changes the order to be consistent

# 🧪 Testing

- [ ] Have item variants setup
- [ ] On android, create a customer return
- [ ] Add an item and try to set the item_variant on that item
- [ ] If you can, success!

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnQuantitiesTable.tsx
+++ b/client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnQuantitiesTable.tsx
@@ -14,7 +14,7 @@ import {
   ItemVariantInputCell,
   useIsItemVariantsEnabled,
 } from '@openmsupply-client/system';
-import React from 'react';
+import React, { useMemo } from 'react';
 import { GenerateCustomerReturnLineFragment } from '../../api';
 import { PackSizeEntryCell } from '@openmsupply-client/system';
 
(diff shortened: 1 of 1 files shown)


lache-melvin (approved):
Haha, that is frustrating! Was still seeing the same issue when I built android, but pushed that useMemo fix from stocktake, seems to have solved the problem if you are happy with it!

Agree - let's get this fix in for now. I don't have time to investigate underlying issue further today but I would be keen to at some point... could you make a refactor issue for it? 🙏

PR #5570, opened 2024-11-26
#5560 JSON Forms Prescription -- category filtering

### LINKED ISSUE #5560: JSON Forms prescriptions -- handle categories
Changes required:
- Add option to "Prescription" component to specify the `itemCategoryPath`, i.e. where in the JSON data to find the category name
- Add filter functionality to "Items" endpoint to filter by category name
- Update "Prescription" component to retrieve `itemCategoryPath` and pass through to Item Selector ("StockItemSearchInput")
- Update "StockItemSearchInput" to take additional "category" prop to pass to above filter

CarlosNZ:
Fixes #5560

@lache-melvin I've made a new branch `5264-json-forms-prescriptions-2.4.1` which is correctly branched off 2.4.1 and has the fixes for #5551 in it. So now basing this PR of that so it's just the new changes.

Adds support for filtering the Prescription items list by category (name). The category is defined elsewhere in the JSON schema, so this component requires an `itemCategoryPath` option to tell it where to look. If not defined, the item list will show all.

# 👩🏻‍💻 What does this PR do?

Have updated the UI Schema in the Program schemas repo to reflect these changes, so you'll need to re-import that (I think just the UI schema has changed). (Branch still `djibouti_family_planning`)

You'll also need to add categories and assign some items to them. The category names are:

Screenshot 2024-11-26 at 11 40 29 AM

You can try excluding one of the categories, and see that it shows "No items" when it is selected.

You can see the Prescription component in "Contraceptive Method" requires the "Prescribed Method" to be selected before it becomes visible, and the item list reflects what is selected in that. The one in "Additional Prescription Item" has no category restriction (`itemCategoryPath` not set) and so shows the full item list.

# 🧪 Testing

- [ ] Re-import/update the UI schema for "Reproductive Health Encounter" as outlined in #5551 
- [ ] Create prescriptions as before (#5551), but note the category restriction on Contraceptive Method -- the Prescription component won't show until a "Prescribed Method" is selected
- [ ] Switching the "Prescribed Method" selection resets the Prescription component (no item selected or lines displayed)
- [ ] Closing the modal and re-opening should persist all the selections
- [ ] Prescription creation works as before, and no other regressions from #5551

### DIFF
--- a/client/packages/common/src/hooks/useQueryParams/types.ts
+++ b/client/packages/common/src/hooks/useQueryParams/types.ts
@@ -13,6 +13,10 @@ export type FilterRule = {
 
 export type FilterBy = Record<string, FilterRule | null>;
 export type FilterByWithBoolean = Record<string, FilterRule | null | boolean>;
+export type FilterByWithStringAndBool = Record<
+  string,
+  FilterRule | null | boolean | string
+>;
 
 export interface FilterController {
   filterBy: FilterByWithBoolean | null;
--- a/client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
+++ b/client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
@@ -28,7 +28,22 @@ import { DraftStockOutLine } from 'packages/invoices/src/types';
 
 export const prescriptionTester = rankWith(10, uiTypeIs('Prescription'));
 
-const Options = z.object({ prescriptionIdPath: z.string() }).strict();
+const Options = z
+  .object({
+    /**
+     * There should only be one prescription for the whole encounter, so where
+     * it's stored in the schema won't be this component's path. Specify the
+     * path in the JSON schema where this value should kept (must be defined in
+     * JSON schema)
+     */
+    prescriptionIdPath: z.string(),
+    /**
+     * Path on the data object to look for an item category name. If not
+     * specified, will display all items, not a particular category
+     */
+    itemCategoryPath: z.string().optional(),
+  })
+  .strict();
 type Options = z.infer<typeof Options>;
 
 const UIComponent = (props: ControlProps) => {
@@ -55,6 +70,13 @@ const UIComponent = (props: ControlProps) => {
   const { mutateAsync: updateLines } = usePrescription.line.save();
 
   const { success } = useNotification();
+
+  const itemCategoryPath = options?.itemCategoryPath;
+  const categoryName: string | undefined = extractProperty(
+    core?.data,
+    itemCategoryPath ?? '',
+    undefined
+  );
 
   // Ensures that when this component is re-mounted (e.g. in a Modal), it will
   // populate the draft line data with previously acquired state
@@ -139,6 +161,7 @@ const UIComponent = (props: ControlProps) => {
           <StockItemSearchInput
             onChange={selected => handleItemSelect(selected)}
             currentItemId={selectedItem?.id}
+            itemCategoryName={categoryName}
           />
           {selectedItem && (
             <StockLineTable
--- a/client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
+++ b/client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
@@ -71,7 +71,8 @@ const UIComponent = (props: ControlProps) => {
 
   const { success } = useNotification();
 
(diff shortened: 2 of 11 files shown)


client/packages/common/src/hooks/useQueryParams/types.ts
```diff
@@ -13,6 +13,10 @@ export type FilterRule = {
 
 export type FilterBy = Record<string, FilterRule | null>;
 export type FilterByWithBoolean = Record<string, FilterRule | null | boolean>;
+export type FilterByWithStringAndBool = Record<
+  string,
+  FilterRule | null | boolean | string
+>;
```
CarlosNZ, line 19: This seems a bit silly, but I'd rather not pull on this thread any further at the moment.

client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
```diff
@@ -56,6 +71,14 @@ const UIComponent = (props: ControlProps) => {
 
   const { success } = useNotification();
 
+  const itemCategoryPath = uischema.options?.['itemCategoryPath'];
```
CarlosNZ, line 74: We'd normally do `const itemCategoryPath = options.itemCategoryPath;`

But I noticed that the `options` object returned by the Zod checker is `undefined` on first render. This was making the logic to detect a change of category far too complicated, until I realised I could pull the value directly  from `uischema.options`. It's the same value, and Zod is still doing run-time type-checking.

client/packages/system/src/Item/Components/StockItemSearchInput/StockItemSearchInput.tsx
```diff
@@ -52,14 +57,14 @@ export const StockItemSearchInput: FC<StockItemSearchInputProps> = ({
...
-        extraFilter ? items.filter(extraFilter) ?? [] : items ?? [],
+        extraFilter ? (items.filter(extraFilter) ?? []) : (items ?? []),
         'name'
       ).sort((a, b) => a.label.localeCompare(b.label)),
     [items]
   );
 
   const cachedSearchedItems = useMemo(() => {
-    const newItems = [...items, ...(data?.nodes ?? [])];
+    const newItems = data?.nodes ?? [];
```
CarlosNZ, line 67: This caused problems whenever we switched the Category, as it didn't remove the old items, just added the new ones. I *think* this change won't cause any regressions elsewhere -- can you think of any?

CarlosNZ:
Have made a follow-up issue for the leftovers: https://github.com/msupply-foundation/open-msupply/issues/5571

client/packages/system/src/Item/Components/StockItemSearchInput/StockItemSearchInput.tsx
lache-melvin: I think this will break the paginated item search, was just looking at this with Aimee so will double check after lunch 🙏

client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
lache-melvin: nice find!

client/packages/system/src/Item/Components/StockItemSearchInput/StockItemSearchInput.tsx
lache-melvin: It does break pagination (would only affect if you have more than 500 items, but you can test by changing this line:

https://github.com/msupply-foundation/open-msupply/blob/8122021324a96625413720242c02209e07ca1411/client/packages/system/src/Item/Components/StockItemSearchInput/StockItemSearchInput.tsx#L35

to be:

```ts
  const { pagination, onPageChange } = usePagination(20);
```

You'll be able to scroll down, but then not scroll back up again!

However, I just tried reverting the change, and it seems to be invalidating the list fine? Could you try removing on your end and see if you still see the issue or if something else resolved it?

lache-melvin (changes requested):
Working great Carl 🚀 

Need to revert that cached items bit so item scroll still works, then should be good to go.

Also found this - bit niche... if the first dropdown you interact with is the category one, the item picker doesn't load the first time?:

https://github.com/user-attachments/assets/4cb99011-b626-4997-abae-2cb4d47e809b

lache-melvin (approved):
Epic thank you!

PR #5581, opened 2024-11-26
#5264 JSON Forms prescription renderer

### LINKED ISSUE #5264: Create JSON Form prescription renderer

CarlosNZ:
Fixes #5264

(Note: replaces PR #5551, which was based off the wrong branch)

# 👩🏻‍💻 What does this PR do?

Implements a component to enter prescription data and save a prescription invoice when the form is saved.

Have used the mechanism demonstrated in #5394 to implement the actions, but have refactored into its own hook (`useFormActions`) which also includes the ability to hold some "intermediate state" -- i.e. data that we don't want to actually save to the JSON form document, but need to keep around (in this case for when the prescription is created). 

Each instance of the component can only select a single item (with multiple batches). However, there is only one prescription created for the whole form. So you can add different items at different form sections (e.g. Contraceptive method, General Exam) and you'll only generate a single prescription on save.

The only piece of data actually saved back to the JSON form is the Invoice ID of the new prescription. When this component loads, it looks up the prescription and displays accordingly:
- if the prescription exists, we just show a link to the existing prescription -- any further editing must be done there for now:
Screenshot 2024-11-25 at 4 15 19 PM
- if the prescription doesn't exist, display the Item selector and Stockline inputs:
Screenshot 2024-11-25 at 4 09 36 PM

Still to do (probably in a subsequent PR):
- [ ] Validation for the quantity inputs (currently doesn't stop you entering more than available quantity, might need additional checks too)
- [ ] Only show items that belong to a specified item category (this would be an option -- some implementations might want access to the full list)
- [ ] I want to display the actual item and quantity alongside the link to the prescription, will need to save the invoice line IDs at the time of form saving, which is a bit of a headache right now (for reasons I can explain IRL if you want). I'd like to solve this, but can wait till after initial implementation.
- [ ] General UI improvements based on design feedback

The end result here is not that complicated, but it took a while to get there as the solution to handle actions in a way that allows for the intended use cases (as well as being flexible enough to handle future Action implementations for other components) was a bit of a convoluted journey! 😆

See inline comments for more detailed explanation of specific elements.

# 🧪 Testing

First you need to set up the JSON schemas from the [program schemas repo](https://github.com/openmsupply/programschemas). These schemas are just minimal implementations to demonstrate and test the functionality here. Full Djibouti schemas are still in progress:
- The branch is called "djibouti_family_planning"
- Build the JSON schema for "djibouti" -> "Reproductive Health Program" and "Reproductive Health Encounter"
- Upload the generated schemas and their equivalent UI schemas to mSupply and sync to OMS
- The details for doing that are the same as in the Testing section of [this issue](https://github.com/msupply-foundation/open-msupply/issues/4642), but come and chat to me if you need a bit more guidance getting it set up

Then enrol one or more patients in the Reproductive Health program and start an Encounter. You should see an Encounter UI like this:
Screenshot 2024-11-25 at 4 33 00 PM

- [ ] Open "Contraceptive Method", select options for the "Method" selectors
- [ ] Start creating a prescription using the Item/Stock inputs
- [ ] Close form, the "Save" button should be enabled (but don't save just yet)
- [ ] Open the "Additional Prescription Item" modal, and add another item
- [ ] Now "Save" the form. You should see a Success notification for the save and for a new prescription created
- [ ] Open one of the above modals again. It should now display "Click to view details" instead of prescription inputs
- [ ] Click the link, and check the prescription opens in a new tab and that the contents of the prescription match what you entered earlier

Additional things to try:
- [ ] Delete a prescription and check that you can now create a new one for that encounter
- [ ] Ensure that the form becomes saveable (i.e. the Save button is clickable) after just entering prescription info (on the Contraceptive tab, the first three selectors are compulsory, so you won't be able to save until those are entered, so either do that in a separate operation, or use the "Additional Prescription" tab)
- [ ] Check that the Batches list changes when different items are selected
- [ ] Check that the state persists when closing the modals and re-opening them (the selected item and the amounts to prescribe)


client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
```diff
@@ -0,0 +1,173 @@
...
+  ItemStockOnHandFragment,
+  StockItemSearchInput,
+} from '@openmsupply-client/system';
+import { AppRoute } from '@openmsupply-client/config';
+import { DefaultFormRowSx, useZodOptionsValidation } from '../../common';
+import { useJsonForms, withJsonFormsControlProps } from '@jsonforms/react';
+import { usePrescription } from '@openmsupply-client/invoices/src/Prescriptions';
+import { useDraftPrescriptionLines } from '@openmsupply-client/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks';
+import { StockLineTable } from './StockLineTable';
+import { DraftStockOutLine } from 'packages/invoices/src/types';
```
lache-melvin, line 27: one more packages import 😁

client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import {
+  DataTable,
+  NumberInputCell,
+  TableProvider,
+  createTableStore,
+  useColumns,
+  useTranslation,
+} from '@openmsupply-client/common';
+import { DraftStockOutLine } from 'packages/invoices/src/types';
```
lache-melvin, line 10: and here :)

lache-melvin (approved):
sweet thanks Carl! Just the couple last imports to fix, merge when ready 🙏

Issue #5458, opened 2024-11-18
Suggested quantity is not calculated for Manual requisitions in omSupply


alainsussol:
## What went wrong? 😲

The Suggested quantity seems to be always 0. 

https://drive.google.com/file/d/1sRNEtT-pSOdXn2BTx8cdqmrSrtOgUWLZ/view?usp=sharing

@roxy-dao is saying that Suggested quantity is not calculated yet for Manual Requisitions and that an issue should be opened

It doesn`t seem to be calculated either when you are doing a manual Requisition on mSupply OG.  When creating an internal order on mSupply OG in the IC context , the calculation seems to be the following

Suggested Quantity= [(Max MOS + Months lead time) * AMC] - Stock available
https://drive.google.com/file/d/1gbknIX3q3EyDFxE6m1h5402vvBPMgCRm/view?usp=sharing
 

## Expected behaviour 🤔
I am not sure. I suppose it depends on the requirements to know if it needs to be calculated or not considering that some of this data needs to be available and some of what is used is entered manually like the AMC

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create a new program requisition
2. Tap on a line and values  to initial SOH, Stock arriving , outgoing, Losses, Additions and AMC
3. Check Suggested quantity
4. See error. The Suggested quantity is still 0

## Your environment 🌱

- Open mSupply Version: 2.4.0RC1
- Legacy mSupply Central Server Version: v19-0RC27

- Platform: Android 14

- Database type: SQLite

mark-prins:
While EXF18 is for internal orders, EXF08 & EXF19 are for customer requisitions. In that there is the comment about calculated quantities:

<img width="640" alt="image" src="https://github.com/user-attachments/assets/987c0e5e-6eaa-4d35-b0e3-166ec4471700">

And then under the heading **EXF01 Bis, EXF06 Bis, EXF08 Bis -  Manual Customer requisition**
there is the line "All stock data columns must be editable except for the ones that are the result of a calculation (available stock, AMC, suggested quantities)"

So while mSupply doesn't calculate - there is something of an implicit expectation that there will be a suggested quantity value which is useful when comparing to the ordered quantity. Currently it would only compare to `0` for each item.

alainsussol:
@roxy-dao @andreievg 
[...]

PR #5585, opened 2024-11-26
Don't calculated suggested quantity

roxy-dao:
Part of #5458

# 👩🏻‍💻 What does this PR do?
Suggested, available and AMC was being calculated incorrectly for response requisitions as they are supposed to be the customer's values not the supplier's, so have removed them for now until discussion in 5458 is concluded.

# 🧪 Testing

- [ ] Create program requisition that has an item that would have a suggested quantity
- [ ] See that it isn't being calculated

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/requisition/response_requisition/insert_program.rs
+++ b/server/service/src/requisition/response_requisition/insert_program.rs
@@ -1,6 +1,5 @@
 use crate::{
     activity_log::activity_log_entry,
-    item_stats::{get_item_stats, ItemStatsFilter},
     number::next_number,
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
Looks good. I wonder if helpful to add // TODOs in this case if explicitly a temporary change?

roxy-dao:
> Looks good. I wonder if helpful to add // TODOs in this case if explicitly a temporary change?

Suggested will probably be done in plugin for CDI.

PR #5586, opened 2024-11-27
Add example of v6 central data sync

andreievg:
Add to sync docs example commits for v6 sync. I realised that there was already somethings like this for asset log reason, but this one is strictly for V6 sync central data sync, would be great to also differentiate other commit examples by type. 

I found this commits based examples of adding different layers very useful, would be great to have one common area for all the different examples like this (yes a bit to maintain but I think saves time on training in the future)


jmbrunskill (approved):
Thanks Andrei

PR #5593, opened 2024-11-27
5409 fix location filter

### LINKED ISSUE #5409: iSC/Cold chain/GAPS: do not show temperature ranges if they haven't been specified

jmbrunskill:
Fixes #5409

# 👩🏻‍💻 What does this PR do?

Correctly filters out only location_types with min and max temperature both 0.0.
E.g. you can now have a temperature range that goes from 0 to 15 for example.

image

Should have pick this up in code review.
Was actually worried about this and I thought I'd tested it... Here: https://github.com/msupply-foundation/open-msupply/pull/5518#pullrequestreview-2453186364

Not sure what happened!

# 🧪 Testing

- [ ] Test you can have a location_type with range 0 -> something and assign to a location in OMS

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/cold_storage_type.rs
+++ b/server/repository/src/db_diesel/cold_storage_type.rs
@@ -5,7 +5,7 @@ use super::{
     ColdStorageTypeRow, DBType, StorageConnection,
 };
 
-use diesel::prelude::*;
+use diesel::{dsl::not, prelude::*};
 
 use crate::{
     diesel_macros::{apply_equal_filter, apply_sort, apply_sort_no_case},
@@ -96,11 +96,9 @@ impl<'a> ColdStorageTypeRepository<'a> {
         let mut query = cold_storage_type_dsl::cold_storage_type.into_boxed();
         // Any cold storage types that don't have temperature set (OdegC to 0degC default value) are invalid => filter out
 
-        query = query.filter(
-            cold_storage_type_dsl::min_temperature
-                .ne(0.0)
-                .and(max_temperature.ne(0.0)),
-        );
+        query = query.filter(not(cold_storage_type_dsl::min_temperature
+            .eq(0.0)
+            .and(max_temperature.eq(0.0))));
 
         if let Some(f) = filter {
             let ColdStorageTypeFilter { id, name } = f;


server/repository/src/db_diesel/cold_storage_type.rs
```diff
@@ -96,11 +96,9 @@ impl<'a> ColdStorageTypeRepository<'a> {
...
         // Any cold storage types that don't have temperature set (OdegC to 0degC default value) are invalid => filter out
 
-        query = query.filter(
-            cold_storage_type_dsl::min_temperature
-                .ne(0.0)
-                .and(max_temperature.ne(0.0)),
-        );
+        query = query.filter(not(cold_storage_type_dsl::min_temperature
+            .eq(0.0)
+            .and(max_temperature.eq(0.0))));
```
roxy-dao, line 101: can't this just be 

```suggestion
        query = query.filter(cold_storage_type_dsl::min_temperature
            .eq(0.0)
            .and(max_temperature.ne(0.0)));
```


?

roxy-dao (approved):
Works great thanks James!

server/repository/src/db_diesel/cold_storage_type.rs
jmbrunskill: I don't think so? Did you try it? I'm happy with the solution I have...

roxy-dao: Yeah I tried both works XD. So I'm happy with either 🤷🏻

PR #5613, opened 2024-11-27
Sync return reasons

### LINKED ISSUE #5230: Not syncing return reasons
## What went wrong? 😲

Was just in the the invoice translation code, and saw this:

https://github.com/msupply-foundation/open-msupply/blob/a6f8ff3890aaf50515c795c7e2a957b849850ecf/server/service/src/sync/translations/invoice_line.rs#L268

I would assume this means return reasons are not being synced/saved!

OG only has the `optionID` field, so might have been pending some kind of decision on how we map the different reason id fields in OMS to OG?

## Expected behaviour 🤔

Return reasons sync

## How to Reproduce 🔨

1. Create return
2. Sync to another site
3. No reason?

## Your environment 🌱

- Open mSupply Version: v2.3.0

- Platform:

- Database type:

roxy-dao:
Fixes #5230

# 👩🏻‍💻 What does this PR do?
Sync return reasons

# 🧪 Testing

- [ ] Have reasons for IA and return 
- [ ] Make a Supplier Return with reason 
- [ ] Make a stocktake with reason
- [ ] Sync
- [ ] Re-initialise
- [ ] Make sure the data is still there

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/sync/translations/invoice_line.rs
+++ b/server/service/src/sync/translations/invoice_line.rs
@@ -79,7 +79,7 @@ pub struct LegacyTransLineRow {
     pub total_after_tax: Option<f64>,
     #[serde(rename = "optionID")]
     #[serde(deserialize_with = "empty_str_as_option_string")]
-    pub inventory_adjustment_reason_id: Option<String>,
+    pub option_id: Option<String>,
     #[serde(rename = "foreign_currency_price")]
     pub foreign_currency_price_before_tax: Option<f64>,
     #[serde(rename = "om_item_variant_id")]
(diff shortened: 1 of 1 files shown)


server/service/src/sync/translations/invoice_line.rs
```diff
@@ -269,8 +268,16 @@ impl SyncTranslation for InvoiceLineTranslation {
...
+            inventory_adjustment_reason_id: match invoice.r#type {
+                InvoiceType::InventoryAddition | InvoiceType::InventoryReduction => {
+                    option_id.clone()
+                }
+                _ => None,
+            },
+            return_reason_id: match invoice.r#type {
+                InvoiceType::CustomerReturn | InvoiceType::SupplierReturn => option_id,
+                _ => None,
+            },
```
roxy-dao, line 280: Couldn't do an overall match and have to check separately here since it kept giving me FK constraint even though the FK was correct....

server/service/src/sync/translations/invoice_line.rs
```diff
@@ -322,14 +329,32 @@ impl SyncTranslation for InvoiceLineTranslation {
...
-                    return_reason_id: _, // TODO
+                    return_reason_id,
                     foreign_currency_price_before_tax,
                     item_variant_id,
                 },
             item_row,
             ..
         } = invoice_line;
 
+        let invoice = match InvoiceRowRepository::new(connection).find_one_by_id(&invoice_id)? {
```
lache-melvin, line 340: You can just destructure invoice_row off of invoice_line above rather than extra query here 😁

roxy-dao: Oh you're so right :rofl

lache-melvin (approved):
LGTM! 🙏

PR #5637, opened 2024-11-28
5619 indicators should show indicator line name 

### LINKED ISSUE #5619: Indicators should show indicator line name
## What went wrong? 😲

Indicator configuration and OMS UI are not consistent. In short, it's intuitive that the "Name" of an indicator line is the survey question being asked, where as the "Code" is a short hand code for the indicator, or a variation of the question (as seemingly the case for HIV indicators config). Here is how it is configured in OG and displayed:

image

image

But in OMS, we decided to only show the code. Perhaps this was in consideration of standard HIV configuration in OG having duplicate names. Unfortunately this can be confusing for regular indicators configured by the user in the "Regimen" section:

image

For HIV it may not be clear what the actual question is that the code is pertaining to other than hoping the users memorise it from elsewhere

image

## Expected behaviour 🤔

OMS should show both the code and the name of an indicator as per the OG spec.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version: 2.4.0-rc4 dev
- Legacy mSupply Central Server Version: 7.18 dev

- Platform: m3

- Database type: pg

roxy-dao:
Fixes #5619

# 👩🏻‍💻 What does this PR do?
Show description + code for HIV indicators and code for regimen.

Screenshot 2024-11-29 at 9 11 54 AM
Screenshot 2024-11-29 at 9 12 04 AM

# 🧪 Testing

- [ ] Have indicators set up for programs (both HIV and manual)
- [ ] Create program requisition
- [ ] Go to indicators tab
- [ ] Click on either of the buttons
- [ ] See that they are more descriptive

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/ListIndicators.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/ListIndicators.tsx
@@ -3,6 +3,7 @@ import {
   ListOptions,
   RouteBuilder,
   useNavigate,
+  useParams,
 } from '@openmsupply-client/common';
 import { IndicatorLineRowFragment } from '../../api';
 
@@ -17,6 +18,7 @@ export const ListIndicatorLines = ({
   lines,
   route,
 }: ListIndicatorLineProps) => {
+  const { programIndicatorCode } = useParams();
   const navigate = useNavigate();
   const value = lines?.find(({ id }) => id === currentIndicatorLineId) ?? null;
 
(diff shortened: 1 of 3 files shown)


client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/ListIndicators.tsx
```diff
@@ -29,9 +31,9 @@ export const ListIndicatorLines = ({
         });
       }}
       options={
-        lines?.map(({ id, code }) => ({
+        lines?.map(({ id, name, code }) => ({
           id,
-          value: code,
+          value: programIndicatorCode === 'HIV' ? `${name}: ${code}` : code,
```
Chris-Petty, line 36: There are no rules for Regimen indicator lines - my intuition is to put the actual question in the `name`, not the `code` so I can't see my questions still:

image

I firmly believe that we should show name and code regardless of the program_indicator. If we ever actually make this feature configurable properly like it was designed to be this will break (What happens when they want an extra program_indicator for TB programs?).

Chris-Petty (changes requested):

Chris-Petty (approved):
image

Look at me go!

PR #5651, opened 2024-11-29
Allow editing of asset number

### LINKED ISSUE #5392: Allow user to enter an asset number when creating an asset via 2d Matrix Barcode Scan

When importing a cold chain equipment asset via a 2d Data Matrix, the user doesn't get an opportunity to enter an asset number (serial number is used instead)

User should get a prompt to enter an asset number when confirming they want to create the asset.

Could allow asset number editing? Might be a better solution?

Should also consider consolidating the saving logic between the new asset modal and the scan button.
Potentially logic such as adding the first functional status on creation should be moved to the backend???

CarlosNZ:
Fixes #5392

# 👩🏻‍💻 What does this PR do?

This is the simplest version of this -- it just makes the Asset number editable after the asset is created. I'll make a follow-up issue to get user input on scanning, as that's a bit more involved and could do with some further planning to make it work nicely.

I notice we're not enforcing unique-ness for Asset number. Should it be?

# 🧪 Testing

- [ ] Go to an Asset detail page
- [ ] Can edit the Asset Number
- [ ] Can save, and Asset number is updated in the Breadcrumb title at the top of the page

### DIFF
--- a/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/Tabs/Summary.tsx
@@ -180,6 +180,13 @@ export const Summary = ({ draft, onChange, locations }: SummaryProps) => {
               onChange={e => onChange({ serialNumber: e.target.value })}
             />
           </Row>
+          <Row label={t('label.asset-number')}>
+            <BasicTextInput
+              value={draft.assetNumber ?? ''}
+              fullWidth
+              onChange={e => onChange({ assetNumber: e.target.value })}
+            />
+          </Row>
           <Row label={t('label.installation-date')}>
             <DateTimePickerInput
               value={DateUtils.getDateOrNull(draft.installationDate)}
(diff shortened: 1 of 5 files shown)


CarlosNZ:
Follow-up: https://github.com/msupply-foundation/open-msupply/issues/5652

adamdewey:
> I notice we're not enforcing unique-ness for Asset number. Should it be?

It would be nice, but not sure how we would do it if assets can be added offline?

jmbrunskill:
We  an asset number duplication check in the `insert` method.
Now that we're updateing it from upsert it should be there too
```
    if let Some(asset_number) = &input.asset_number {
        if check_asset_number_exists(asset_number, connection)?.len() >= 1 {
            return Err(InsertAssetError::AssetNumberAlreadyExists);
        }
    }
```

Can you add this to the `validate` method in the update please?

I agree @adamdewey we can't guarentee it doesn't get duplicated with sync but would be good to keep it unique for a site at least.

jmbrunskill (approved):
Approved, but can you add that duplication check, or put a follow up issue for that to be picked up refactor week?

jmbrunskill:
> Can you add this to the `validate` method in the update please?

On second thoughts I think we'll need to make it check that it's not the same id that the current asset is using.
I think a follow up issue is better...

PR #5654, opened 2024-11-30
Fix issue with extism js bundle

### LINKED ISSUE #5312: Unable to run Item usage report
## What went wrong? 😲

Running the item usage report results in this:

Screenshot 2024-11-05 at 12 14 12 PM

image

## Expected behaviour 🤔
Report should display correctly.

## How to Reproduce 🔨

Using this version of item usage will cause a problem:
[item-usage.json](https://github.com/user-attachments/files/17625487/item-usage.json)

Taking the generated version from the reports repo is fine though - so there's an issue with the report JSON.
The questions is - can we display an error message, that the report has a problem? that would be more useful than the above error.

Issue raised by @marthakeezy : I have a copy of the sqlite datafile if you are unable to repro using the above JSON

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

andreievg:
Fixes #5312

# 👩🏻‍💻 What does this PR do?

This took a bit of time to investigate, first there was an issue with parsing large json inside extism js function, this was fixed by https://github.com/extism/js-pdk/issues/107, and extism-js reinstall helped, as per: https://github.com/extism/js-pdk?tab=readme-ov-file#install-script (@fergie-nz @roxy-dao I think you might have been the only two with js-pdk installed, can you please re-install)

After that I was getting error with linking, updating extism package helped (I'll try to replicate and make an issue)

https://github.com/extism/js-pdk/issues/108

# 🧪 Testing

Try to run item usage report with @mariyamsupply data file

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/reports/expiring-items/2_3_0/convert_data_js/package.json
+++ b/server/reports/expiring-items/2_3_0/convert_data_js/package.json
@@ -11,7 +11,6 @@
   "types": "./convert_data.d.ts",
   "description": "",
   "devDependencies": {
-    "@extism/js-pdk": "^1.1.1",
     "esbuild": "^0.24.0"
   }
 }
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -443,20 +443,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
-dependencies = [
- "gimli 0.28.1",
-]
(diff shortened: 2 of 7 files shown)


server/reports/expiring-items/2_3_0/convert_data_js/package.json
```diff
@@ -11,7 +11,6 @@
   "types": "./convert_data.d.ts",
   "description": "",
   "devDependencies": {
-    "@extism/js-pdk": "^1.1.1",
```
andreievg, line 14: These types were not actually used, wanted to avoid confusion with extism-js version so removed

roxy-dao (approved):
Working for me~

Thanks Andrei! 🙏🏻

PR #5668, opened 2024-12-02
5591 fix integration tests

### LINKED ISSUE #5591: Fix integration tests
## What went wrong? 😲

image
## Expected behaviour 🤔

Integration tests should at least compile!

## How to Reproduce 🔨

1. Run `cargo test --features=integration_test`

## Your environment 🌱

- Open mSupply Version:2.4.0

- Platform:

- Database type:

roxy-dao:
Fixes #5591

# 🧪 Testing

- [ ] Run integration tests

# 📃 Documentation

  1.
  2.


andreievg (approved):
Looks great, thanks

server/service/src/sync/central_data_synchroniser.rs
```diff
@@ -60,7 +62,9 @@ impl CentralDataSynchroniser {
 
             logger.progress(
                 SyncStepProgress::PullCentral,
-                max_cursor - last_cursor_in_batch,
+                // During integration tests got attempt to 'substract with overflow'
+                // There is a chance that max_cursor is lower the last cursor in batch
```
jmbrunskill, line 66: Interested to understand how this is possible...

jmbrunskill: I think this if fine though, better to be safe rather than crash...

server/service/src/sync/translations/item.rs
```diff
@@ -69,7 +69,9 @@ fn to_legacy_ven_category(ven_category: VENCategory) -> String {
 }
 
 pub(crate) fn ordered_simple_json(text: &str) -> Result<String, serde_json::Error> {
-    let json: serde_json::Value = serde_json::from_str(text)?;
+    let mut json: serde_json::Value = serde_json::from_str(text)?;
+    // Saving dose_picture 'picture' type as incoming 'empty string' causes issues in integration tests
+    json.as_object_mut().unwrap().remove("dose_picture");
```
jmbrunskill, line 74: Is unwrap() safe here? Should we consider what happens if it's not a json or `null` which I have seen sync'd once, perhaps not for item but still... will it block sync permanently if this happened??

server/service/src/sync/test/integration/transfer/requisition.rs
```diff
@@ -105,6 +106,7 @@ async fn integration_sync_requisition_transfers_initialisation() {
             &response_site.store,
             &item1,
             &item2,
+            6,
```
jmbrunskill, line 109: Any particular reason for `6` here? Would 0/1 be better or make it optional? Sorry I added this in but didn't update the integration test...

server/service/src/sync/translations/item.rs
jmbrunskill, line 74:
```suggestion
       if let Some(json) = json.as_object_mut() {
        json.remove("dose_picture");
    }
```
May not matter but this feels safer, if we're very confident then would like a comment as to why it's safe.

https://github.com/msupply-foundation/open-msupply/wiki/Code-Review-Comments---Rust#dont-use-unwrap-unless-in-tests

jmbrunskill (approved):
Happy to merge, but have a couple of questions. Main concern is the unwrap() in ordered_simple_json for `item_translation`.

server/service/src/sync/central_data_synchroniser.rs
andreievg: I didn't want to dwell on this for too long (and look through 4d code), but I would suspect max_cursor would be queried before the batch is compiled, and while batch is compiled more central records are added

server/service/src/sync/translations/item.rs
andreievg: technically not possible but i 100% agree we shouldn't be unwrapping, very nice catch thanks @jmbrunskill

andreievg: Thought to check in vs code and committed there: dea05985d5544bbe99a79242a10a50349a936dc1

Thanks so much for the catch

server/service/src/sync/test/integration/transfer/requisition.rs
roxy-dao: No particular reason XD. Optional would probably be better

PR #5684, opened 2024-12-03
5602 create feedback form table

### LINKED ISSUE #5602: Create a feedback_form table
New table with fields:

|Field | Type | 
|---------|---|
| id | String
reply_email | String
body | String
created_datetime | NaiveDatetime
user_id | String
store_id | String
site_id | String

This PR should include create a new migration for this table, and implement the repository layer (i.e. feedback_form and feedback_form_row files)

You'll also need to create a new `Changelog` type for the `feedback_form`. This will require an extra piece in the migration, just for postgres. See example here: 

https://github.com/msupply-foundation/open-msupply/blob/c3b6c04ea08aee398738272a14d8994633f1d9af/server/repository/src/migrations/v2_04_00/add_bundled_item_table.rs#L24-L32

noel-yeldos:
Fixes #5602 

# 👩🏻‍💻 What does this PR do?
Create repository layer for ~feedbackform~ contact_form table and add migration for table

Thanks to team Korimako for jamming to get this one over the line!

# 🧪 Testing

- [ ] Cargo run
- [ ] Check that ~feedbackform~ contact_form table is in db file

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_05_00/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[cfg(test)]
+#[actix_rt::test]
+async fn migration_2_05_00() {
+    use v2_04_00::V2_04_00;
+
+    use crate::migrations::*;
+    use crate::test_db::*;
+
+    let previous_version = V2_04_00.version();
```
lache-melvin, line 31: Just need to update this, the previous version is actually `V2_04_01`, so that is what we should test upgrading from 🙏

lache-melvin (changes requested):
Looking good! Just the test to fix up 🚀

server/repository/src/migrations/v2_05_00/mod.rs
noel-yeldos: Done!

lache-melvin (changes requested):
We can come back to this next week, but the next steps here are:

- [ ] Change `feedback_form` to `contact_form`
- [ ] Add new `contact_type` field to `FeedbackFormRow` (enum `Feedback` | `Support`)
- [ ] Add new column in the `contact_form` migration
- [ ] Add migration for the postgres enum (like the `indicator_value_type` enum defined [here](https://github.com/msupply-foundation/open-msupply/blob/develop/server/repository/src/migrations/v2_04_00/indicator_line_column_create_tables.rs))

server/repository/src/migrations/v2_05_00/add_contact_form_table.rs
```diff
@@ -0,0 +1,57 @@
...
+            r#"
+                CREATE TABLE contact_form (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    reply_email TEXT NOT NULL,
+                    body TEXT NOT NULL,
+                    created_datetime {DATETIME} NOT NULL,
+                    user_id TEXT NOT NULL REFERENCES user_account(id),
+                    store_id TEXT NOT NULL REFERENCES store(id),
+                    site_id TEXT NOT NULL,
+                    contact_type {contact_type}
```
lache-melvin, line 40:
```suggestion
                    contact_type {contact_type} NOT NULL
```

lache-melvin (changes requested):
Looks great Noel, thank you! 

One requested change to make the contact form column `NOT NULL` (we should only not include the `NOT NULL` piece if the field will be an `Option<SomeType>`)

lache-melvin (approved):
Mean thanks Noel, merge when ready 🔥

PR #5706, opened 2024-12-04
Reduce apk size

jmbrunskill:
Fixes #

# 👩🏻‍💻 What does this PR do?

Migrates the changes for building 64 bit only android from 2.4
Adds the `opt-level = "s"` flag to reduce the compiled build size...

Here's an APK to try out
https://drive.google.com/file/d/14oAuLnzkPlEvUDEP8LGoIW5-goWcWNgD/view?usp=sharing

# 🧪 Testing

- [ ] Check that the apk installs, syncs and generally doesn't have any problems.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/android/app/src/main/jniLibs/armeabi-v7a/.gitignore
+++ b/client/packages/android/app/src/main/jniLibs/armeabi-v7a/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore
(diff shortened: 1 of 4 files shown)


andreievg (approved):
Looks good to me, I think it's worth adding a remove command for the .so in the build scrip ¯\_(ツ)_/¯

PR #5714, opened 2024-12-08
Add comments for FE filter

### LINKED ISSUE #5624: Move program_indicator filtering from FE to BE and filter inactive records on generation

As a rough solution to a half cooked feature I put some filtering of program indicators in the frontend rather than backend. This is bad because the same filtering would be required for internal orders when we do that, and it's very easy to forget all the places you need to do it in the FE (or at least FE should be refactored to do it a different way than I have 😉).

The more I looked at it, the more confused I was about the correct way to do it in the backend, hence the frontend kludge.

There are several things to consider when filtering program_indicators and the related tables:

- [ ] When creating a program requisition we should 
  - [x] exclude inactive program indicators [(we do this here currently, but might be problematic)](https://github.com/msupply-foundation/open-msupply/blob/b0f700726a298317e4466a8fca9bd46187680dae/server/repository/src/db_diesel/program_indicator.rs#L56)
  - [x] exclude inactive indicator_lines (we don't do this, and need to take care to not filter inactive for _past_ requisitions querying for display in FE)
- [ ] When creating API response for program indicators display on a requisition
  - [ ] Ignore any `is_active` fields - we need the representation of indicators from the time that period is relevant to, so what was generated at the time the requisition was made.
  - [ ] We should include only program_indicators and relating tables if the relevant period **period** has any indicator values
  - [ ] We should include only the indicator_lines that have at least 1 related indicator_value

Both the generation and the gql API use the program_indicators service, which means care needs to be taken when implementing filtering to make sure that it effects each appropriately. Perhaps they should not use the same service, and the repo layer should allow overriding the default filtering out of program_indicators. GLHF.

It's logically a bit messy. I might be cooked. Above is a guide, but could be a better approach!

fergie-nz:
Fixes #5624

Also related to [docs changes](https://github.com/msupply-foundation/msupply_docs/pull/130)

# 👩🏻‍💻 What does this PR do?

Adds comment to explain presence of front end filtering for when values don't exist

# 🧪 Testing

No functional changes. No testing required

# 📃 Documentation

Docs updated in linked PR.

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorEditPage.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/IndicatorEdit/IndicatorEditPage.tsx
@@ -39,7 +39,8 @@ export const IndicatorEditPage = () => {
   const currentLineAndColumns = linesAndColumns.find(
     l => l.line.id == programIndicatorLineId
   );
-  // Lines may be added to a program_indicator after the requisition was made. We need to hide them, as they're not valid or necessary for past requisitions.
+  // Should only include indicators if they have at least one column with a value
+  // Filtering for !value done on FE because values are queried via loader
   const populatedLines = linesAndColumns.filter(l =>
     l.columns.find(c => c.value)
   );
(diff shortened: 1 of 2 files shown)


roxy-dao (approved):
Thanks~

PR #5700, opened 2024-12-04
5643 refactor prescription hooks

CarlosNZ:
Fixes #5643

# 👩🏻‍💻 What does this PR do?

Should be no noticeable changes in user experience.

Refactor hooks using new pattern. New hooks are:

- `usePrescription` -- manages state and provides database methods for a single prescription
- `usePrescriptionList` -- provides query/data for the prescription list
- `usePrescriptionDelete` -- provides "delete" database method. Didn't incorporate into the previous hooks as deletion of prescriptions is available from both the List and Details pages
- `usePrescriptionLines` -- provides database methods for create/updates/delete of individual prescription lines. I didn't fully manage the draft state in this hook, as the logic in the existing `useDraftPrescriptionLines` is pretty complicated and interacts with a whole bunch of other things, so I left it alone and just used the hook to provide the database mutations.

## 💌 Any notes for the reviewer?
See inline comments.

# 🧪 Testing

Basically just do Prescription stuff (create, delete, edit, add/remove lines, etc) and see if anything breaks or behaves unexpectedly.

Should also test the "Prescription" component in Encounters if possible as that also uses these hooks.

PR #5715, opened 2024-12-08
5643(2) Re-use usePatchState hook

### LINKED ISSUE #5643: Update prescriptions to use new Hook Pattern

CarlosNZ:
Fixes #5643 -- part 2

# 👩🏻‍💻 What does this PR do?

Putting this in a separate PR just to avoid making [the earlier one](#5700) even larger!

Created the `usePatchState` hook as part of that one since we're re-using this exact logic in several places. This PR just replaces the other implementations with the hook version to reduce duplication. Should be no functionality changes.

# 🧪 Testing

- [ ] Confirm no regressions in:
  - [ ] Vaccine course modal
  - [ ] Stock line edit


client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
```diff
@@ -147,6 +145,7 @@ export const VaccineCourseEditModal: FC<VaccineCourseEditModalProps> = ({
       error(t('error.failed-to-save-vaccine-course'))();
       console.error(e);
     }
+    resetDraft();
```
jmbrunskill, line 148: Do we want to reset if there's an issue saving? Sounds like potential data loss? Shouldn't this be next to the `success()` message?

jmbrunskill (approved):
Looks good to me, just that one potential issue with the resetState().

I'll let you fix it or convince yourself it's not needed...

client/packages/programs/src/VaccineCourseEditModal/VaccineCourseEditModal.tsx
CarlosNZ: Yeah, fair point. I've moved it up into the "Success" condition: ac9f263f8e

(It actually previously had the `setIsDirty(false)` at the top of this method, so would have had a similar problem previously -- good spotting)

PR #5716, opened 2024-12-09
5599 adjust settings page permissions

### LINKED ISSUE #5599: Adjust permissions for the Settings page
Currently the settings section ~~(now `Help`)~~ is restricted to only users with server admin permission. However, we now wish to open up the settings to users so that they can edit display settings.

- [ ] Make the ~~`Help`~~ `Settings` link in the AppDrawer visible to all
- [ ] Make these accordion sections hidden, unless you have the ServerAdmin permission 
  - [ ] Synchronisation
  - [ ] Support
  - [ ] Devices
  - [ ] Configuration
- [ ] The `Display Settings` section should always be visible, but hide these elements for non-ServerAdmins 
  - [ ] Custom theme
  - [ ] Custom logo

Image

zachariah-at-msupply:
Fixes #5599

# 👩🏻‍💻 What does this PR do?

- Makes the Settings icon in the app drawer available to every user.
- Hides all of the Settings sections except Display settings from every non-admin.
- Hides all the Display settings except the Language settings from every non-admin.

image

Is it secure: could a malicious party force the custom theme or logo sections, or other settings sections, to be visible somehow?

# 🧪 Testing

[ ] log in as an admin and check that all settings sections are visible

[ ] log in as a user with minimum permissions and check that the setting icon is available, but ONLY display settings are visible, and within those settings only language is visible.

# 📃 Documentation

[x] **Part of an epic**: documentation will be completed for the feature as a whole

### DIFF
--- a/client/packages/host/src/Admin/DisplaySettings.tsx
+++ b/client/packages/host/src/Admin/DisplaySettings.tsx
@@ -7,6 +7,8 @@ import {
   useTranslation,
   useNotification,
   LocalStorage,
+  useAuthContext,
+  UserPermission,
 } from '@openmsupply-client/common';
 import { themeOptions } from '@common/styles';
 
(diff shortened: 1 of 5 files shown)


lache-melvin (approved):
Looks great, thanks Zachariah 🚀 

Yep this is sufficiently secure, the content is not rendered at all if you don't have the permission (as opposed to hidden by CSS styles which could be removed 😁)

PR #5718, opened 2024-12-09
5666 generate indicators for internal orders

### LINKED ISSUE #5666: CDI 2.5: Generate indicators for program Internal Orders
Generate indicators for the program based on the period in `insert_program_request_requisition` function. 

### Moneyworks Jobcode 🧰
OMS:CIDV

fergie-nz:
Fixes #5666 & #5926

Adds generate indicator values for request requisition (similar to how they are generated for response requisition).

# 👩🏻‍💻 What does this PR do?

Adds generation function
Adds tests

# 🧪 Testing

- [ ] Add a new program internal order
- [ ] Check database to see indicator values are populated for this new internal order with default values
Currently there is no front end visibility for indicator values in internal orders.
- [ ] Run tests to ensure passing

# 📃 Documentation


roxy-dao (changes requested):
Thanks Ferg! Just a small refactor request pleaseee

roxy-dao:
Can you also do a check to only generate the indicators if the store doesn't have the NPSPClient tag but has the ESPC tag? And add a TODO to refactor this to not use name tags? The indicators shouldn't be generated for NPSPClient tag stores

server/service/src/requisition/request_requisition/generate.rs
```diff
@@ -92,3 +95,36 @@ pub fn generate_requisition_lines(
 
     Ok(result)
 }
+
+pub(crate) fn generate_program_indicator_values(
+    store_id: &str,
+    period_id: &str,
+    program_indicators: Vec<ProgramIndicator>,
+) -> Vec<IndicatorValueRow> {
+    let mut indicator_values = vec![];
```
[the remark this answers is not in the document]
fergie-nz: I had it abstracted initially, but for request requisitions it is slightly different because there is no 'other_party_id'.
I thought it might be confusing to use that function but pass in store_id twice, once as other_party_id, so moved it back to two separate functions.

 Is it still worth abstracting into one?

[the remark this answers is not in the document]
roxy-dao: You can add a check for type... so you know what id to use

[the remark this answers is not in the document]
roxy-dao: I think you missed this comment @fergie-nz

server/service/src/requisition/request_requisition/generate.rs
```diff
@@ -92,3 +95,36 @@ pub fn generate_requisition_lines(
...
+        for line in program_indicator.lines {
+            for column in line.columns {
+                let value = match column.value_type {
+                    Some(_) => column.default_value.clone(),
+                    None => line.line.default_value.clone(),
+                };
+
+                let indicator_value = IndicatorValueRow {
+                    id: uuid(),
+                    customer_name_link_id: store_id.to_string(),
```
roxy-dao, line 116: You're using the store_id instead of the name_id here

server/service/src/requisition/request_requisition/generate.rs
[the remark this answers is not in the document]
fergie-nz: thank you! I did

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -2,20 +2,26 @@ use crate::{
...
+    MasterListLineRepository, NameTagFilter, NameTagRepository, NumberRowType, Pagination,
+    ProgramIndicatorFilter, ProgramRequisitionOrderTypeRow, ProgramRow, RepositoryError,
+    Requisition, RequisitionLineRowRepository, RequisitionRowRepository, StoreFilter,
+    StoreRepository,
 };
 
-use super::generate_requisition_lines;
+use super::{generate_program_indicator_values, generate_requisition_lines};
+
+const NSPC_CLIENT_TAG: &'static str = "NPSPClient";
```
roxy-dao, line 23:
```suggestion
const NPSC_CLIENT_TAG: &'static str = "NPSPClient";
```

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +194,46 @@ fn generate(
...
+        NameTagFilter::new().store_id(EqualFilter::equal_to(&ctx.store_id)),
+    ))?;
+
+    let mut indicator_values: Vec<repository::IndicatorValueRow> = vec![];
+    let store_tag_names: Vec<String> = store_tags.into_iter().map(|t| t.name).collect();
+
+    if store_tag_names.contains(&ESPC_TAG.to_owned())
+        && !store_tag_names.contains(&NSPC_CLIENT_TAG.to_owned())
+    {
+        let indicators = match supplier_store {
```
roxy-dao, line 221: Why is this matching with supplier_store?

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +194,46 @@ fn generate(
...
+            Some(_) => {
+                generate_program_indicator_values(&ctx.store_id, &period_id, program_indicators)
+            }
+            None => vec![],
+        };
+        for indicator in indicators {
+            indicator_values.push(indicator)
+        }
```
roxy-dao, line 229: The generate program also generates a vec of indicators...

can you not indicator_values = ?

server/service/src/requisition/common.rs
```diff
@@ -102,3 +105,41 @@ impl From<RepositoryError> for OrderTypeNotFoundError {
         Self::DatabaseError(error)
     }
 }
+
+pub struct IndicatorGenerationInput<'a> {
```
roxy-dao, line 109: Why did you put lifelines on? 🤔

fergie-nz: I had thought you needed a lifetime to ensure referenced components outlive the struct.

Would this be preferable with cloning?

```
pub struct IndicatorGenerationInput {
    pub store_id: String,
    pub period_id: String,
    pub program_indicators: Vec<ProgramIndicator>,
    pub other_party_id: String,
}
```

roxy-dao: I think cloning is fine for something this small (:

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +195,42 @@ fn generate(
...
+        None,
+        Some(ProgramIndicatorFilter::new().program_id(EqualFilter::equal_to(&program.id))),
+    )?;
+
+    let store = StoreRepository::new(connection)
+        .query_one(StoreFilter::new().name_id(EqualFilter::equal_to(&ctx.store_id)))?;
+
+    let should_generate_indicators: bool =
+        if let Some(supplying_store_id) = store.and_then(|s| s.name_row.supplying_store_id) {
+            supplying_store_id == other_party_id
```
roxy-dao, line 213: This is not checking the correct ids. the supplying_store_id is a store id but you are checking it against a name id.

roxy-dao (changes requested):
Thanks for the changes so far Ferg. The conditional check is checking against an incorrect value and your tests are failing D:

server/service/src/requisition/request_requisition/insert_program.rs
fergie-nz: Hey @roxy-dao 

@Chris-Petty mentioned that some internal orders via programs will have an other party id which is not their typical supplying_store_id.

I wonder had you thought of that case? I am wondering if that check would still be valid.

roxy-dao: Had a convo with @Chris-Petty. Agree, that was something I've brought up before, but for now we will go with the supplying_store to generate indicators and a carry over issue should be created to allow stores to input indicators based on their facility level. #5844

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +199,82 @@ fn generate(
...
+    let stores_with_store_as_supplier_ids: Vec<String> = NameRepository::new(input.connection)
+        .query(
+            &input.store_id,
+            Pagination::all(),
+            Some(NameFilter::new().supplying_store_id(EqualFilter::equal_to(&input.store_id))),
+            None,
+        )?
+        .into_iter()
+        .map(|s| s.store_row.and_then(|s| Some(s.id)))
+        .filter_map(|id| id)
```
roxy-dao, line 248:
```suggestion
        .filter_map(|s| s.store_row.map(|s| s.id))
```

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +199,82 @@ fn generate(
...
+        .collect::<Vec<String>>();
+
+    if stores_with_store_as_supplier_ids.len() > 0 {
+        return Ok(generate_aggregate_indicator_values(
```
roxy-dao, line 252:
```suggestion
       Ok(generate_aggregate_indicator_values(
```

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +199,82 @@ fn generate(
...
+                store_id: input.store_id.clone(),
+                period_id: input.period_id,
+                program_indicators: input.program_indicators,
+                other_party_id: input.store_id.clone(),
+                customer_store_ids: stores_with_store_as_supplier_ids,
+                connection: input.connection,
+            },
+        )?);
+    } else {
+        return Ok(generate_program_indicator_values(
```
roxy-dao, line 263:
```suggestion
       Ok(generate_program_indicator_values(
```

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +199,82 @@ fn generate(
...
+            IndicatorGenerationInput {
+                store_id: input.store_id.clone(),
+                period_id: input.period_id,
+                program_indicators: input.program_indicators,
+                other_party_id: input.store_id,
```
roxy-dao, line 268: You're passing in the store_id but the indicator_value is saving the name_link_id. This will give a foreign constraint error..

roxy-dao (changes requested):

server/service/src/requisition/common.rs
```diff
@@ -102,3 +105,41 @@ impl From<RepositoryError> for OrderTypeNotFoundError {
         Self::DatabaseError(error)
     }
 }
+
+pub struct IndicatorGenerationInput {
+    pub store_id: String,
+    pub period_id: String,
+    pub program_indicators: Vec<ProgramIndicator>,
+    pub other_party_id: String,
```
roxy-dao, line 113: There was a request to name this to `customer_name_id`~

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +199,82 @@ fn generate(
...
+        indicator_values,
+    })
+}
+
+pub struct InternalOrderIndicatorGenerationInput<'a> {
+    pub connection: &'a StorageConnection,
+    pub store_id: String,
+    pub period_id: String,
+    pub program_indicators: Vec<ProgramIndicator>,
+    pub other_party_id: String,
```
roxy-dao, line 233: `customer_name_id` please

roxy-dao (commented):
P.S indicators should also only be generated if the supplier has a store and you should be checking if the value exists for aggregating values!

roxy-dao:
To sum up: The aggregate and generate values can be all in one function. 
- The values are aggregated if existing values are there and the other party is the supplying store.
- The name_link_id should be the customer name link id not the supplier or customer's store id

Rename other_party_id to `customer_id` and also add to the tests (:

Thanks Ferg!

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +197,144 @@ fn generate(
         .map(|line| line.item_id)
         .collect();
 
-    let requisition_line_rows =
+    let requisition_lines =
         generate_requisition_lines(ctx, &ctx.store_id, &requisition, program_item_ids)?;
 
-    Ok((requisition, requisition_line_rows))
+    let program_indicators = program_indicators(
```
roxy-dao, line 203: Filter by elmis code coming soon in carry over

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +197,144 @@ fn generate(
...
+            .store_id(EqualFilter::equal_any(customer_store_ids.clone())),
+    )?;
+
+    let mut indicator_values = vec![];
+
+    for program_indicator in program_indicators {
+        for line in program_indicator.lines {
+            for column in line.columns.clone() {
+                let value = match column.value_type {
+                    Some(IndicatorValueType::String) => column.default_value.clone(),
```
roxy-dao, line 281: Not sure what we want to do with strings.. 🤷🏻  so just defaulting

andreievg: Sounds good

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +197,144 @@ fn generate(
...
+                    value,
+                };
+
+                indicator_values.push(indicator_value);
+            }
+        }
+    }
+
+    Ok(indicator_values)
 }
```
andreievg, line 338: Looks good, can be simplified (with some helpers too): https://github.com/msupply-foundation/open-msupply/compare/5666-generate-indicators-for-internal-orders...5666-generate-indicators-for-internal-orders(Simplify-aggregated-or-default-value)

While doing above you'll notice I used it in validation check, i also tried to find, where value_type is used, and notice we not pre-calculating value type for front end, and then notice that we only take column value type into consideration, i think we need to consider both ?

In fact I think we should have stored valueType on the value earlier ¯\_(ツ)_/¯

andreievg (changes requested):
Looking good, one change requests and some suggestions, the change request is to make sure to use customer_name_id rather then store id to filter out customer and also would be great to add a small test

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +197,144 @@ fn generate(
...
+                            sum.to_string()
+                        } else {
+                            column.default_value.clone()
+                        }
+                    }
+                    None => match line.line.value_type {
+                        Some(IndicatorValueType::String) => line.line.default_value.clone(),
+                        Some(IndicatorValueType::Number) => {
+                            if store_pref
+                                .use_consumption_and_stock_from_customers_for_internal_orders
```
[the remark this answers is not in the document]
roxy-dao: Ah I added the check after (: checks both customers & store pref

server/service/src/requisition/request_requisition/insert_program.rs
roxy-dao: Ooh nice thanks! So much neater (:

server/service/src/requisition/request_requisition/insert_program.rs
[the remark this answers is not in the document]
roxy-dao: [3be0072](https://github.com/msupply-foundation/open-msupply/pull/5718/commits/3be00720e96aaa801a68ad7f42566cdb4e29250c)

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -177,10 +197,144 @@ fn generate(
...
+            Pagination::all(),
+            Some(
+                NameFilter::new()
+                    .supplying_store_id(EqualFilter::equal_to(store_id))
+                    .is_customer(true),
+            ),
+            None,
+        )?
+        .into_iter()
+        .map(|name| name.name_row.id)
```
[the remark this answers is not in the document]
roxy-dao: [9a0c400](https://github.com/msupply-foundation/open-msupply/pull/5718/commits/9a0c4009db7c5ffea649c111de4b5d578e638bcd) with description

andreievg (approved):
Looks great thanks !

One carry over is to make sure that front end knows correct type for values where type comes from line rather then column (unless i am missing something I think it's currently only taking the value type from column ?) ideally this would be returned correctly by back end, and I think it's possible since ColumnNode already take some info from line

roxy-dao (approved):

PR #5720, opened 2024-12-09
5650 dispensing redesigned item add & edit page view

### LINKED ISSUE #5650: Dispensing: Redesigned item add/edit page view
When dispensing in OMS, adding an item used a modal to create/edit the item data.

We want to move all this functionality to be done in a new page instead of modal to give more space and allow for extendable functionality while keeping the screen usable on a tablet.

image

image

> Note: we don't need to do all the other input updates, or other nice to have designs for the components.

The page has sections for different areas needed in the prescription, such as stock lines, but also provisions for other functionality,  `Directions`, `Repeats` etc.

For this issue, we should simply move all the existing add item functionality to the new design

**Acceptance Criteria**
- [x] Add item opens a new page
- [x] An Item selector in `New Item` page
- [x] After selecting an item, you can immediately (e.g without having to save) enter other details such as selecting your stocklines etc.
- [x] Able to issue stock from stock lines (using something similar to existing UI)
- [x] Able to manually assign stock to lines
- [x] Able to enter the `issue` field to automatically select from valid stocklines
- [x] Selecting an existing line in the list view opens the item page with that item selected
- [x] Lefthand bar selector lets you change between items in the prescription
image

jmbrunskill:
Fixes #5650

# 👩🏻‍💻 What does this PR do?

Changes dispensing to use a whole page to display line editing functionality
Uses the new approach for adding a new line too...

Raising this as a draft to get some feedback and input.

Feels a bit buggy in a number of ways currently, thinking we maybe do some testing and raise issues for little bugs to follow up on?

For example I ran into this bug, which appears to be the same on develop branch? https://github.com/msupply-foundation/open-msupply/issues/5753

- Saving on `OK and Next` / `Next` feels a bit weird somehow, and it a bit unwieldy in the code see useEffect solution to navigate
- I'm Saving on Previous too but the also seems a bit wrong? Maybe we should go back to auto saving idea?
- Navigating from the 
- Warning say things like Press Ok Again to confirm but we don't actually have an `Ok` button now. Is that ok?
image

# 🧪 Testing

- [ ] Create prescriptions, check everything works as it used too
- [ ] Check new item edit pages display ok on Tablets

# 📃 Documentation

  1.
  2.


client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PageLayout.tsx
```diff
@@ -0,0 +1,20 @@
...
+  Right: React.ReactElement;
+}
+
+export const PageLayout = ({ Left, Right }: PageLayoutProps) => {
+  return (
+    <Grid container spacing={2} direction="row" padding={2} paddingBottom={2}>
+      <Grid item xs={3}>
+        {Left}
+      </Grid>
+      <Grid item xs={9}>
```
jmbrunskill, line 15: This is a different ratio to Manual reqs (seems like they aren't using all `12` available slots for some reason?)

jmbrunskill: Should this be some kind of common component?

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+    let itemSet = new Set();
+    let items: ItemRowFragment[] = [];
+    lines.forEach(line => {
+      if (!itemSet.has(line.item.id)) {
+        items.push(line.item);
+        itemSet.add(line.item.id);
+      }
+    });
+    return items;
+  }, [data]);
```
jmbrunskill, line 44:
```suggestion
  }, [lines]);
```

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/useNextItem.tsx
```diff
@@ -1,6 +1,8 @@
 import { ItemRowFragment, toItemRow } from '@openmsupply-client/system';
 import { usePrescription } from '../../../api';
 
+// TODO: DELETE THIS FILE
```
jmbrunskill, line 4: This isn't needed in the new design but didn't want to break it just yet as I still have the modal...
Might be other stuff to remove too, worth checking....

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/usePreviousNextItem.ts
```diff
@@ -0,0 +1,40 @@
+interface itemId {
+  id: string;
+}
+
+interface prevNextState {
+  hasPrevious: boolean;
+  previous: null | string;
+  hasNext: boolean;
+  next: null | string;
```
jmbrunskill, line 9: in Manual reqs we have this as a structure with an id in it, I'm assuming we just want an ID here, but if that is the case we probably should name it `nextId` or something.
This could be a generic/common component if we only care about the id...

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/Footer.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const Footer: FC<FooterProps> = ({
+  hasNext,
+  next,
+  hasPrevious,
+  previous,
+  invoiceNumber,
+  loading,
+  isDirty,
```
jmbrunskill, line 31: Don't like having all this in the footer component, but don't know a way around it unless we do that auto saving thing...

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditPage.tsx
```diff
@@ -0,0 +1,89 @@
...
+    query: { data, loading: isLoading },
+  } = usePrescription();
+
+  const lines =
+    data?.lines.nodes.sort((a, b) => a.item.name.localeCompare(b.item.name)) ??
+    [];
+
+  let currentItem = lines.find(line => line.item.id === itemId)?.item;
+
+  let items = useMemo(() => {
```
jmbrunskill, line 33: Not sure if useMemo is strictly needed here, but have had problems with things change on re-render before so was thinking this might helpful...

CarlosNZ:
Thanks @jmbrunskill , I'll jump into reviewing this and building off it first thing tomorrow :)

CarlosNZ:
First thought is that if we're going to go down this path then we need to commit to the redesign of the actual line input UI. Having to side-scroll to see what's happening currently makes it even less useful than the modal version:

Screenshot 2024-12-12 at 10 12 13 AM

Do we know exactly what extent we're re-doing this, and what the agreed UI is? (For a start, just reducing the number of columns would help)

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEdit.tsx
```diff
@@ -36,24 +32,22 @@ import { PrescriptionLineEditForm } from './PrescriptionLineEditForm';
...
 
 interface PrescriptionLineEditModalProps {
-  isOpen: boolean;
-  onClose: () => void;
   draft: Draft | null;
   mode: ModalMode | null;
+  items: ItemRowFragment[];
 }
 
 export const PrescriptionLineEdit: React.FC<PrescriptionLineEditModalProps> = ({
```
CarlosNZ, line 43: Since this is no longer a modal, we probably want to do some renaming to remove all references to "Modal" (including the "mode" prop values)

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/hooks/usePreviousNextItem.ts
CarlosNZ: Agree, should be a common component, doesn't just need to be for items either. Would just need to change the input "items" to be an array of strings rather than a specific type of object. Or, an additional prop to specify the field to compare against to avoid having to map the input array to just the IDs.

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PrescriptionLineEditPage.tsx
```diff
@@ -0,0 +1,89 @@
...
+          Left={
+            <>
+              <ListItems
+                currentItemId={itemId}
+                items={items}
+                route={RouteBuilder.create(AppRoute.Dispensary)
+                  .addPart(AppRoute.Prescription)
+                  .addPart(String(invoiceNumber))}
+                enteredLineIds={enteredLineIds}
+                showNew={true}
```
CarlosNZ, line 72: We should only `showNew` when the Prescription is editable (i.e not "Verfied")

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/Footer.tsx
```diff
@@ -0,0 +1,105 @@
...
+  handleSave,
+}) => {
+  const navigate = useNavigate();
+  const { itemId } = useParams();
+
+  // When saving, we need to navigate to the next or previous prescription line, but we'll get a not saved warning if we don't wait for the save to complete
+  // Use effect gets around this issue (but is ugly)
+  let [navigateTo, setNavigateTo] = useState<string | null>(null);
```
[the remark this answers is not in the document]
jmbrunskill: NIce, that's essentially what I did have calling `onSave` then realised I needed other `handleSave` logic. Should have thought to try this!

client/packages/invoices/src/Prescriptions/DetailView/PrescriptionLineEdit/PageLayout.tsx
jmbrunskill: https://github.com/msupply-foundation/open-msupply/issues/5766

PR #5725, opened 2024-12-09
5537 paginated autocomplete

### LINKED ISSUE #5537: AutocompleteWithPagination not working
## What went wrong? 😲
The AutocompleteWithPagination component isn't working as it should. Currently this is being used by `StockItemSearchInput` and in #5524.

This component will only load the pagination set in backend currently, so if the user has more items than the set pagination then it won't show up in the search components.

I heard it's working in HSH, so could check HSH to see why it's not working

## Expected behaviour 🤔

Search works across all records

Scrollable list, once further records/pages are loaded, previous pages are still scrollable

## How to Reproduce 🔨

1. 

## Your environment 🌱

- Open mSupply Version: 2.4.0

- Platform:

- Database type:

lache-melvin:
Fixes #5537

# 👩🏻‍💻 What does this PR do?

Autocomplete pagination was implemented 3 different ways, between item select input, catalogue item input, and store select input.

This PRs aligns item & catalogue item with store select - I think this is the cleanest option (the other two were watching the pagination object with a useEffect to trigger the next page query, which was both hard to follow and required more renders)

Bug was that it "wasn't working" but I actually couldn't replicate that... if you can, please let me know? Hopefully changes fix it regardless 😅 

If you have takes/improvements on this pattern, please share!

# 🧪 Testing

- [ ] Have datafile with lots of items (OR for dev testing, you can change the `RECORDS_PER_PAGE` to something smaller, like 20)
- [ ] Open the item picker dropdown somewhere, e.g. add item to outbound shipment
  - [ ] As you scroll to bottom, next "page" of data appears
  - [ ] Your position in the list is maintained when next page appears
  - [ ] You can search for items not in the currently displayed set
- [ ] Same checks for catalogue item options in the `Create Asset` modal (Coldchain > Equipment)

# 📃 Documentation

  1.
  2.


CarlosNZ:
Hey @lache-melvin, this doesn't seem to be working for me as described -- currently I get no scroll bar at all, so can't fetch more items:

Screenshot 2024-12-11 at 10 13 29 AM

Also, less significant but worth noting -- the number of items shown is always 2 less than the `ROWS_PER_PAGE` value. I have it set to "6" in the above, but shows 4 🤷‍♂️

lache-melvin:
@CarlosNZ - the -2 rows per page thing is strange... I wonder why??

I think the "can't scroll to next page" is a separate thing, where if you don't have enough items in the list to create a scroll bar, then you don't have a "hit-the-bottom-of-the-scroll" event to trigger then next page query. Kind of think this isn't worth fixing - paginated autocomplete should only be needed if there are enough records to be scrolling, and I wouldn't anticipate us querying less than 20 records per page (probably even 100?)

If you need a datafile with more items I can share one?

CarlosNZ (changes requested):
Looking good, thanks for consolidating these, I hate having multiple implementations of the same thing!

Turns out the "2 items short" problem was just the due to the `extraFilter` passed from the parent, which filters out items already in the list. Which makes sense, but ideally this would be done server-side so you get a "page" of results of the correct size, but I guess it's fine in this case (with the default page size of 100, a handful less in the results list won't be noticeable).

I have a couple of suggestions, made a suggestion branch for each:

1. Would be good to show the "Loading" indicator while the next page is being fetched. Easy to do, just use the `isFetchingNextPage` value from `useInfiniteQuery`:  
https://github.com/msupply-foundation/open-msupply/commit/4b2860c446509ead44104e3943c75cd169143373
2. This is not really part of this PR (it's bothered me for a while), but it's slightly annoying how you lose the current options list when re-querying. If you're just trying to narrow the already-retrieved options, you should be able to see the filtering of the results instantly without having to wait for the new query to finish, as the items are already there. Here's one way to do it, there's probably others:  
https://github.com/msupply-foundation/open-msupply/commit/4964da59c83c78d9925e262d8b0618a9e40bec35

This is the difference in user experience (slow network artificially simulated)
How it works now:
Slow version
With my suggestion -- you get the "loading" indicator to tell you it's fetching more, but the previous results are still instantly available:
Fast version

I've just implemented both those suggestions for the StockItemSearchInput. It's be good if they were baked into the hook so that that functionality was automatically available everywhere, but that's a bit more work. If you don't want to re-factor that now and you think it's worth doing at some point, I can write up an issue for it.

lache-melvin:
Thanks Carl! Sorry for slow turnaround, really like your improvements 🚀 

Have added them in, plus refactored some of the options and pagination management into the common component - let me know what you think!

CarlosNZ (approved):
Very nice, thanks for the updates. :)

PR #5726, opened 2024-12-09
5691 Encounter delete fix

### LINKED ISSUE #5691: Programs Module: deleting an encounter results in duplicated confirmation messages and a warning
## What went wrong? 😲

Noticed the following behaviour when deleting an Encounter: 

https://github.com/user-attachments/assets/7db1f31c-6e4b-47f1-a765-bead4b630596

I have also once saw (and Carl was able to replicate), when deleting an Encounter, you run into an infinite loop of confirmation messages and a warning the information will be lost if navigating away. Looks something like this: 

Screenshot 2024-12-04 at 12 55 42 PM

## Expected behaviour 🤔

Should be able to delete an Encounter successfully, with a single confirmation message and no warning windows showing

## How to Reproduce 🔨

1. Have programs set up in your store
2. Create an encounter for a patient
3. Attempt to delete an encounter - see duplicated confirmation messages.
4. I am unclear on how to replicate the infinite messages & the warning one - saw it happen only once on my laptop and on Carl's computer. It definitely happens, but unable to pinpoint what causes it. 

## Your environment 🌱

- Open mSupply Version: V2.4.1-RC1

- Database type: SQLite3

CarlosNZ:
Fixes #5691

# 👩🏻‍💻 What does this PR do?

Resets the `deleteRequest` value to avoid the infinite loop over the `useEffect` function.

I think the `useEffect` chaining using the `deleteRequest` intermediate value is not great React usage, but changing that is a bit of a bigger job so won't do that now. The problem is we can't easily update the data using JSON Forms' `setData` method and then immediately save that changed data in a single render. Something to figure out another time...

# 🧪 Testing

- [ ] Try to delete an encounter
- [ ] Make sure it works and you are navigated back to the Encounter list without any unnecessary loops
- [ ] Try changing something in the Encounter first before deleting and also confirm no probs

### DIFF
--- a/client/packages/system/src/Encounter/DetailView/DetailView.tsx
+++ b/client/packages/system/src/Encounter/DetailView/DetailView.tsx
@@ -192,12 +192,14 @@ export const DetailView: FC = () => {
     updateEncounter({ status: EncounterNodeStatus.Deleted });
     setDeleteRequest(true);
   };
+
   useEffect(() => {
     if (!deleteRequest) return;
     if (
       (data as Record<string, JsonData>)['status'] ===
       EncounterNodeStatus.Deleted
     ) {
+      setDeleteRequest(false);
       (async () => {
         const result = await saveData(true);
         if (!result) return;


jmbrunskill (approved):
Looks good to me!

PR #5727, opened 2024-12-09
Fix queries on item usage report

### LINKED ISSUE #5579: Item usage report showing different data on mSupply vs on OMS
## What went wrong? 😲
Numbers are different on both reports. Would expect numbers to be the same. Am on latest version of mSupply and OMS.

As an example, see item code: PMC834

Report comparisons:
mSupply:
Screenshot 2024-11-27 at 10 58 06 AM

OMS:
Screenshot 2024-11-27 at 10 58 24 AM

This is the stocklines on OMS and Legacy:
Screenshot 2024-11-27 at 11 08 07 AM

Screenshot 2024-11-27 at 11 08 50 AM

In stock has 1 extra item, stock on order is 0, expiring months is 0. 

## Expected behaviour 🤔
All the values of mSupply report and OMS report should match

## How to Reproduce 🔨

Generate a report on mSupply and on OMS and compare

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #5579

# 👩🏻‍💻 What does this PR do?

Fixes queries to show data item usage report.

You can test this by removing the existing item-usage report in your database, rebuilding standard reports, and then upserting the updated report. Note starting the server will automatically upsert built report.

I write this here because documentation of these commands isn't comprehensive now.

Also added totalStockOnHand to item stats query. This is what is being used on stockLine list view and OG reports for 'in stock' column. Previously was using availableStockOnHand which caused discrepancy.

For reviewers - the fastest way to get the new report is like so:

- [ ] Do not stop your server running instance.
- [ ] Manually delete the item-usage report from your database
- [ ] Open a new terminal and run:

```
cargo build &&  ./target/debug/remote_server_cli build-standard-reports && ./target/debug/remote_server_cli upsert-reports-json
```

Which will rebuild and upsert reports to the database name set up in your .yaml file.

# 🧪 Testing

- [ ] Run commands above to delete, rebuild, and upsert reports
- [ ] Go to reports
- [ ] Open item usage report
- [ ] See expiring in 6/12 months matches expected from OG
- [ ] See stock on hand also populated

### DIFF
--- a/server/reports/item-usage/2_3_0/src/expiringInSixMonths.postgres.sql
+++ b/server/reports/item-usage/2_3_0/src/expiringInSixMonths.postgres.sql
@@ -13,5 +13,5 @@ FROM stock_line s
 INNER JOIN item_link i ON i.id = s.item_link_id
 INNER JOIN this_month ON true
 INNER JOIN six_months ON true
-WHERE s.expiry_date >= this_month AND s.expiry_date < six_months AND s.store_id = $storeId
+WHERE s.expiry_date < six_months AND s.store_id = $storeId
 GROUP BY s.id, i.item_id
--- a/server/service/src/item_stats.rs
+++ b/server/service/src/item_stats.rs
@@ -22,6 +22,7 @@ pub struct ItemStats {
     pub total_consumption: f64,
     pub average_monthly_consumption: f64,
     pub available_stock_on_hand: f64,
+    pub total_stock_on_hand: f64,
     pub item_id: String,
     pub item_name: String,
 }
@@ -140,6 +141,7 @@ impl ItemStats {
                     .get(&stock_on_hand.item_id)
                     .copied()
                     .unwrap_or_default(),
+                total_stock_on_hand: stock_on_hand.total_stock_on_hand,
             })
             .collect()
     }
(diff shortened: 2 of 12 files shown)


server/reports/item-usage/2_3_0/src/expiringInSixMonths.postgres.sql
```diff
@@ -13,5 +13,5 @@ FROM stock_line s
 INNER JOIN item_link i ON i.id = s.item_link_id
 INNER JOIN this_month ON true
 INNER JOIN six_months ON true
-WHERE s.expiry_date >= this_month AND s.expiry_date < six_months AND s.store_id = $storeId
+WHERE s.expiry_date < six_months AND s.store_id = $storeId
```
fergie-nz, line 16: The line changes here don't make sense with what was changed.

This change removes condition that stock isn't expired.
Now expiring in 6/12 months columns will also show stock which has already expired. 
This matches behaviour of OG reports

server/service/src/item_stats.rs
```diff
@@ -151,8 +153,9 @@ impl ItemStats {
             available_stock_on_hand: row.available_stock_on_hand,
             item_id: requisition_line.item_row.id.clone(),
             item_name: requisition_line.item_row.name.clone(),
-            // TODO: Implement total consumption
+            // TODO: Implement total consumption & total_stock_on_hand
             total_consumption: 0.0,
+            total_stock_on_hand: 0.0,
```
fergie-nz, line 158: This is a bit gross. Currently unused in the context where it is defaulted to 0.

Maybe worth separating into a different struct?

andreievg: I am tempted to make another query just for the report (and not add total_stock_on_hand, but I think since total_consumption is also set to zero here, I think it should be ok, thanks

andreievg (approved):
Looks good to me thanks

Issue #5138, opened 2024-10-15
OMS central server configurations for Manual Requisitions


roxy-dao:
Manual requisitions will use OMS central server to configure the columns for manual requisitions. Designs to come.

These configurations will need to sync to all sites.

1. Show list of available requisition columns
2. Have a toggle to enable/disable the visibility of the column
3. Have an optional override for the column label

### Moneyworks Jobcode 🧰

OMS:CIMDV

andreievg:
I played around a little bit with the translation structure, considering that global preference is stored in a table with key = preference type and value = json. In the use case of the requisition line column visibility some columns might be added via plugin, so they won't be strictly typed, but other should be I think, structure below should allow for it.

```

#[derive(Serialize, Deserialize, Debug)]
enum TypedColumnName {
    PreviousStockOnHand,
    PositiveAdjustment,
}
#[derive(Serialize, Deserialize, Debug)]
struct TypedColumn {
    column_name: TypedColumnName,
    is_visible: bool,
}
#[derive(Serialize, Deserialize, Debug)]
struct CustomColumn {
    column_name: String,
    is_visible: bool,
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(untagged)]
enum RequisitionColumnVisibility {
    TypedColumn(TypedColumn),
    CustomColumn(CustomColumn),
}

fn main() {
    let value = r#"[{
        "column_name": "PreviousStockOnHand", "is_visible": true },
         {"column_name": "PositiveAdjustment", "is_visible": false },
         {"column_name": "SomeCustomColumn", "is_visible": true }
    ]"#;

    println!(
        "{:#?}",
        serde_json::from_str::<Vec<RequisitionColumnVisibility>>(value)
    );

    println!(
        "{}",
        serde_json::to_string_pretty(&vec![RequisitionColumnVisibility::TypedColumn(
            TypedColumn {
                column_name: TypedColumnName::PositiveAdjustment,
                is_visible: true
            }
        )])
        .unwrap()
    );
}
```

Above can go all the way through to graphql and typescript layer i think.

I would also imagine that we access each preference type separately, and have a seperate method for each serialisation and mapping, although I played around with some generic preference types, not sure how they would work across all of the layers:

```

table! {
    glob_pref (id) {
        id -> Text,
        value -> Text,
    }
}

trait GlobalPref: Default + DeserializeOwned {
    fn key() -> &'static str;

[...]

PR #5729, opened 2024-12-10
Preferences KDD

### LINKED ISSUE #5683: KDD of preferences structure
Global preferences have come up as some to implement now, but going into the weeds reveals a lot of false assumptions and people have expressed thoughts on how we might it approach it better including how front end APIs could ergonomically provide preferences. Conceptually, data structures here likely are applicable to user permissions as well, especially in a role based system. Feels right to make a KDD addressing the full context and laying out different approaches.

### Moneyworks Jobcode 🧰

OMS:DDSP

Chris-Petty:
Fixes #5683

### DIFF
--- a/decisions/2024-12-03_preferences.md
+++ b/decisions/2024-12-03_preferences.md
@@ -114,6 +114,34 @@ _Cons:_
 
 - Some preferences might need more than one FK, e.g. user pref for a specific store or machine.
 
+### Option 3 - Similar to `properties` and `name_properties` structure
+
+With this option we'd define a list of `prefs` in the database that could be assigned, and what possible values they have e.g. float, int, boolean, controlled list.
+
+Each preference has a `unique_key` assigned, this key is used to assign the value to when serialising/deserialising to json.
+
+Each pref would record if it's relevent to apply at a global, store, or other level.
+
+In sync we'd sync a `preference_data` record for each store and one global (for now) could add user, or machine etc as needed in future. This would have a json payload similar to Option 1.
+
+There's nothing to stop us still deserialise them in a rust struct as needed
+
+This option would allow us to develop a preference editor UI similar to what we have with editing store/name properties.
+
+If you need some more complex type in your sync record, then it probably is worth creating a dedicated ui and management for this struct.
+
+_Pros:_
+
+- Re-uses an existing pattern
+- Clear path for creating a UI without lots of over head for adding a new pref
+- Can still deserialise to specific structs if needed
+- Could be used by plugins to add additional prefs as needed
+
+_Cons:_
+- Could be larger payload as prefs aren't broken down into different areas
+- Need to insert prefs into database rather than just adding to code? Maybe there's a code first way to achieve the same thing? Enums?
+
+
 ## Decision
 
 TBD
--- a/decisions/2024-12-03_preferences.md
+++ b/decisions/2024-12-03_preferences.md
@@ -1,4 +1,4 @@
-# Categorisation in Open mSupply
+# Preferences in Open mSupply
 
 - _Date_: 2024-12-03
 - _Deciders_: @Chris-Petty, @CarlosNZ, @jmbrunskill, @andreievg
@@ -103,7 +103,7 @@ _Cons:_
 - Potentially more upfront effort right now
 - Create need for design convention on how to structure preferences. Should you make an individual pref for every little thing, or group them together into many keys in the same pref's JSON? How many is too many? If ever concurrent editing of same record with race on saving, bigger JSON means more data lost without merge strategy. We can nest - how much nesting is TOO FAR?
   - structs defining pref shape should at least provide some tacit guardrails; if your struct is crazy maybe the pref should be broken down!
-   - Requires dedicated code for configuration UI (as we don't know how pref might be structured)
+  - Requires dedicated code for configuration UI (as we don't know how pref might be structured)
 
 ### Option 2 - Add more preference tables like the current OMS `store_preference`
 
@@ -139,9 +139,9 @@ _Pros:_
 - Could be used by plugins to add additional prefs as needed
 
 _Cons:_
+
 - Could be larger payload as prefs aren't broken down into different areas
 - Need to insert prefs into database rather than just adding to code? Maybe there's a code first way to achieve the same thing? Enums?
-
 
 ## Decision
 
--- a/decisions/2024-12-03_preferences.md
+++ b/decisions/2024-12-03_preferences.md
@@ -1,0 +1,119 @@
+# Categorisation in Open mSupply
+
+- _Date_: 2024-12-03
+- _Deciders_: @Chris-Petty, @CarlosNZ, @jmbrunskill, @andreievg
+- _Status_: DRAFT
+- _Outcome_: TBD
+
+## Context
+
+In mSupply we have numerous approaches to prefs (gory details saved for `Additional Info` below). We need a preferences system that allows:
+
+- Global preferences, where something is broadly set (we need to decided whether this needs to / can be overridden by say store prefs, or vice versa)
+- Store preferences, where something pertains to the operations of a particular store
+- User preferences. An example of this is remembering custom column order or widths in table views.
+- Machine preferences. Any preference that should be associated with a particular device. Prime example is choosing a default label printer for a certain machine, so that regardless of store logged in the labels print at the printer at the workstation the user is at.
+
+### Distinction from creating new tables
+
+It is really important to get rules down for when we should simply make a new table. Perhaps most importantly, if the entity should have relations then we want good indexing that dedicated tables provide. Beyond that, when a configuration is concrete enough, representing something correlating to the real world and likely doesn't need hierarchy or overrides then it may be appropriate to dedicate a table as it is more maintainable to have the data structure is apparent in the db schema with named, typed fields and columns (one could argue well defined structs for a preference also achieves this). Obvious candidates for distinct tables are invoices, requisitions or their lines. Master lists are a clear central configuration that benefits from RDB tables for scaling and relations. A good example of something in OG that is in preferences, but could be better in a table (particularly due to a lack of strict typing) is configured printers:
+
+- The parameters are pretty concrete, so fields are very stable.
+- We don't need any overrides.
+- We (currently!) want all printers to be available regardless of what store is logged in.
+- They don't need centralised control - i.e. configuration should be done on site, not on the central server.
+- We should want them to sync, to include in backups and easy redeployment when reinitialising a site.
+
+Note, there are other pref records for default printers associated with certain machines. These have a relation to some printer ID saved in the data pref that stores the configured printers (I presumed the IP of the printer, however I checked and each printer apparently has a SERIAL number derived from the `number` table).
+
+### Use cases
+
+- Machine preferences
+  - Default printer
+- Store preferences
+  - To show dispensing directions UI (can simplify UI if unneeded for particular store)
+  - Store specific label printing settings
+- Global prefs
+  - Errrr nothing applicable yet?
+  - One case is for a default set of prefs that are configurable, and store prefs give finer grained overrides.
+
+### Requirements
+
+#### Must
+
+- Sync. In many cases all central controlled, but potentially allow for local control (e.g. user preference controlled on remote site)
+- Can be related to: Global, Store (and extended to) Machine, User (and Site?)
+- Clean API with typing from backend through to frontend
+
+#### Should
+
+- Extendable for plugins
+
+#### Could
+
+- Defined hierarchy rules for overrides. At least a base definition and optional overriding preferences.
+- Partial definitions if hierarchy. (Means Option fields everywhere...)
+
+### Additional info
+
+#### OG
+
+OG has 2 sources of preferences which are similar but in storage field type, `pref_blob` and `pref`. The former is an arbitrary blob of data, the latter a 4D object field (serialised as JSON in sync). Neither have great rules for consistency. The blobs might be arrays of any value type, objects (possibly base64 encoded and stored as text, or primitive values). The `pref` objects at least are consistently object, but beyond that can have all the same wild adventures as the blobs. For instance, the boolean `pref` "Show direction entry in prescriptions" is the 28th bit in a 32bit integer used as a bit array. All the other bits are miscellaneous, mainly invoice related prefs.
+
+Both follow the same field structure which likely has some wisdom attached from our context over 20 years:
+
+- item: a unique string identifying the pref
+- store_id: the pref applies to given store
+- user_id: the pref applies to given user
+- network_id: the pref applies to given machine identified with MAC address (this might not be ideal for devices with multiple network adapters)
+
+Any, none or all of the above may be applied. If none are applied, then it's either a global pref OR a datafile specific pref. **Many settings in the preferences window are datafile specific, not global. There is no design pattern indicating which are which**
+
+The tables have a troubled relationship with sync. On initialisation, there is a lot of cruft such as all the central server datafile preferences being synced e.g. the default printer for the central server machine, various site version metadata records... Unlike most tables, there is no trigger code handling sync. It is handled by **explicit** calls to queue the pref record for sites whenever the pref record are modified and it is necessary to sync them, which is easy to get wrong or neglect to do. They do not go into the central change log.
+
+A big downside with these tables is the lack of strict typing around the shape of their pref data. It is often very hard to know what prefs exist in the system, why they exist and what shape their data is by looking at the code. It is often easier to analyse datafiles, dump prefs to JSON and text search to try find a relevant pref record if you're not sure what pref you're looking for. There is no enum or catalogue, and little to no dev documentation.
+
+#### OMS
+
+We currently just have a basic pref setup for store prefs. There is a `store_preference` table that has a column for each preference. This provides typing. There are some column limits of 1600 for postgres and ~32000 for sqlite, we're unlikely to hit those though.
+
+This doesn't cover things such as machine or user preferences by it self. To follow this pattern for other requirements, you'd have to make the tables `user_preference` and `machine_preference` and so on. If a future requirement, cross overs of those prefs might not really work without getting crazy e.g. a user pref for a particular store.
+
+## Options
+
+### Option 1 - Take the lessons from OG `pref` table by doing the same
+
+The OG pref table is battle tested for over a decade:
+
+- We create a `preference` (or `pref`?) table with `id`, `item`, `store_id`, `user_id`, `machine_id` and `data` fields.
+- The lack of typing that ailed OG can readily be addressed with with structs in OMS. These structs may carry through to the frontend gql types as appropriate.
+- Sync needs deeper thought, but at a glance: store prefs are easily identified and synced by the `store_id`. `user_id` could be added to `change_log` to handle that (if needed to sync... or we sync users and their prefs everywhere 😉). Perhaps it's OK to not sync and lose machine_id records.
+- We store the pref data in an JSON text field. We deserialize this with serde into structs for each pref as expected.
+  - This does mean that sync does not validate the `data` field, that validation occurs when the record is accessed. Though, we could quickly deserialize at sync buffer integration time just to check!
+- We can drop the `store_preference` table, but keep the repo the same for now, just querying from `preference` instead.
+
+_Pros:_
+
+- Consistency with OG can be easier to reason about between systems (provided it the OG pattern isn't hard to reason about to begin with...)
+- Just 1 table and logic (i.e. 1 repo etc.)
+- Generally flexible and battle tested
+
+_Cons:_
+
+- Potentially more upfront effort right now
+- Create need for design convention on how to structure preferences. Should you make an individual pref for every little thing, or group them together into many keys in the same pref's JSON? How many is too many? If ever concurrent editing of same record with race on saving, bigger JSON means more data lost without merge strategy. We can nest - how much nesting is TOO FAR?
+  - structs defining pref shape should at least provide some tacit guardrails; if your struct is crazy maybe the pref should be broken down!
+
+### Option 2 - Add more preference tables like the current OMS `store_preference`
+
+_Pros:_
+
+- Lower effort for now
+
+_Cons:_
+
+- Some preferences might need more than one FK, e.g. user pref for a specific store or machine.
+
+## Decision
+
+TBD
--- a/decisions/2024-12-03_preferences.md
+++ b/decisions/2024-12-03_preferences.md
@@ -91,6 +91,8 @@ The OG pref table is battle tested for over a decade:
 - We store the pref data in an JSON text field. We deserialize this into a struct in the rust code to which provides sensible defaults for anything missing.
   - This does mean that sync does not validate the `data` field, that validation occurs when the record is accessed. Though, we could quickly deserialize at sync buffer integration time just to check!
 - We can drop the `store_preference` table, but keep the repo the same for now, just querying from `preference` instead.
+- The list of preferences is defined by a single struct in the backend. The backend is responsible for merging Global, Store, and potentially if required `User Prefs/Machine Prefs`.
+The single struct can have additional complexity, e.g. struct within structs as required. This may require additional mapping to graphql though.
 
 _Pros:_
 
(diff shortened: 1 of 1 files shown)


Chris-Petty:
Anything to add, go for it!

Anything I should elaborate on?

I decided for first draft just to get the contextual stuff down and a crack at requirements. Put in the OG prefs option. I haven't gone much deep into implementation details, figured to just agree on requirements and preferred schema/data structure first.

jmbrunskill:
A thought to add, can we export a group of gloabl and/or default store prefs to apply to a new site/demo?
E.g. if we have a new feature that's relying on a set of prefs, it would be good if we can provide that to support & QA to import for testing or setup.

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Consistency with OG can be easier to reason about between systems (provided it the OG pattern isn't hard to reason about to begin with...)
+- Just 1 table and logic (i.e. 1 repo etc.)
+- Generally flexible and battle tested
+
+_Cons:_
+
+- Potentially more upfront effort right now
+- Create need for design convention on how to structure preferences. Should you make an individual pref for every little thing, or group them together into many keys in the same pref's JSON? How many is too many? If ever concurrent editing of same record with race on saving, bigger JSON means more data lost without merge strategy. We can nest - how much nesting is TOO FAR?
+  - structs defining pref shape should at least provide some tacit guardrails; if your struct is crazy maybe the pref should be broken down!
```
jmbrunskill, line 105:
```suggestion
  - structs defining pref shape should at least provide some tacit guardrails; if your struct is crazy maybe the pref should be broken down!
   - Requires dedicated code for configuration UI (as we don't know how pref might be structured)
```

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Option 2 - Add more preference tables like the current OMS `store_preference`
+
+_Pros:_
+
+- Lower effort for now
+
+_Cons:_
+
+- Some preferences might need more than one FK, e.g. user pref for a specific store or machine.
```
jmbrunskill, line 115:
```suggestion
- Some preferences might need more than one FK, e.g. user pref for a specific store or machine.

### Option 3 - Similar to `properties` and `name_properties` structure

With this option we'd define a list of `prefs` in the database that could be assigned, and what possible values they have e.g. float, int, boolean, controlled list.

Each preference has a `unique_key` assigned, this key is used to assign the value to when serialising/deserialising to json.

Each pref would record if it's relevent to apply at a global, store, or other level.

In sync we'd sync a `preference_data` record for each store and one global (for now) could add user, or machine etc as needed in future. This would have a json payload similar to Option 1.

There's nothing to stop us still deserialise them in a rust struct as needed

This option would allow us to develop a preference editor UI similar to what we have with editing store/name properties.

If you need some more complex type in your sync record, then it probably is worth creating a dedicated ui and management for this struct.

_Pros:_

- Re-uses an existing pattern
- Clear path for creating a UI without lots of over head for adding a new pref
- Can still deserialise to specific structs if needed

_Cons:_
- Could be larger payload as prefs aren't broken down into different areas
- Need to insert prefs into database rather than just adding to code? Maybe there's a code first way to achieve the same thing? Enums?

```

image

jmbrunskill (approved):
Thanks @Chris-Petty I'm suggesting a third option, which is kind of like Option 1 but a bit more limited/focussed. Mostly thinking about how we can use this to create a pref editor UI, think the approach I suggest would work.

More complex stuff wouldn't be done with preferences, or we'd need to extend teh capabilities of the pref options/types.

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Re-uses an existing pattern
+- Clear path for creating a UI without lots of over head for adding a new pref
+- Can still deserialise to specific structs if needed
+- Could be used by plugins to add additional prefs as needed
+
+_Cons:_
+- Could be larger payload as prefs aren't broken down into different areas
+- Need to insert prefs into database rather than just adding to code? Maybe there's a code first way to achieve the same thing? Enums?
+
```
Chris-Petty, line 143: Great suggestion @jmbrunskill !

Would work quite well for plugins as they'd just be centrally controlled records that sync everywhere. 

I'd be slightly less concerned about the larger payload. We have already the "custom_data" field and that hasn't blown up too badly despite the years. In some ways it's quicker to sync 1 big record over 1000 pref records, far few queries. Tricky if someone puts something chunky in it though, such as a base64 encoded file. e.g. a plugin for adding an album of photos of a store, but rather than using document sync they just encode it in the plugin config 😆.

Speaking of plug config, I had been thinking that if plugins had config then the plugin should include the UI for the configuration that the FE loads up. That might have some flexibility advantages, but the way you suggest would probably cover most cases and be easier than cooking a UI for every plugin.

I'll add to the other options details on what config options they have.

Chris-Petty:
@jmbrunskill what your suggestion quite reminds me of vscode extension configurations which got me sniffing around how vscode does it. Here is rust-analyser's [config definition](https://github.com/rust-lang/rust-analyzer/blob/248bd511aee2c1c1cb2d5314649521d6d93b854a/editors/code/package.json#L343) for example.

Here is the docs for extension configuration https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema. One thing to note is that it is a superset of JSON schema, which we have worked with on presenting with JSON forms previously.

All this kinda shows to me that your approach could easily handle things such as drop down options and some degree of grouping. Surely if the approach is flexible enough for vscode extensions and generating the settings UI from the JSON schema surely it'll be good enough for our configurations/preferences 🙂.

I am wondering if I'm diverging from option 3 though - or at least I'm guessing on how we manage the configuration data. Some amount of base `preference_config` for preferences of standard functionality, and separate records for each plugin. Haven't thought out exactly how it's queried for. 

n.b. I'm worried about calling the config record just `preference` cause it'll 100% create lots of confusion with those versed in OG or outside of dev (because from a UI perspective we talk of preference being the thing you turn on/off)

> Each preference has a unique_key assigned, this key is used to assign the value to when serialising/deserialising to json.

Just confirming how devs might use this - logic using a preference should be expected to use the unique key right? some scenarios to convey what I mean (and tbh wouldn't be any different to the first 2 options! "unique_key" is just the item):

- In backend, code can query the preference_data table based on a static `&str` (constant?)
- For frontend usage: 
  - gql layer, as the backend, can include relevant preferences in the gql schema, the backend would include them via as above 
  - We could have an gql endpoint dedicated to fetching a pref data, to be used arbitrarily by frontend logic (I sorta like this less than the above option)
  - Must eventually have an endpoint for mutating a pref data

Perhaps this is relevant and plausible for all the options, which indicates I may be in the weeds 😉.

mark-prins:
Looking good @Chris-Petty - nice and comprehensive!
I was also wondering about the vscode approach, I like the hierarchical nature of it. that's achievable with any of your options though, with a differing level of effort. 

Agree that having a pref editor is nice - and I like the approach of having a single table with fields for the association of a pref to a user/store etc.

which is to say that I like option 1 and then preferred Jame's third way when that popped up. happy with any of these approaches tho.

I'm relaxed about migrating from an existing OG implementation, though ideally we'd accomodate that somehow

CarlosNZ:
I basically like Option 3, but I'd like to offer what I think are important "requirements" and try and work towards a solution that achieves all if possible -- I think Option 3 is heading in the right direction, but would be good to see if we're all on the same page about the importance of the following:

**Requirements** (see if you agree):
- Invisible to front-end: app should just receive a single structure of all prefs and not care about the source (user pref/store pref, etc)
- Applied heirarchically
  - I.e. can be defined for store, user, or globally, with configurable rules about what the precedence is. e.g. some prefs might want the "store" setting to overrule any "user" setting, but others might need to be the other way round
  - Should aslo handle "partial" definitions, i.e. a user doesn't have to have a setting for each pref, it would just fall back to store, etc, until ultimately hitting a global, hard-coded default
- Minimise workload of keeping configurations UI up-to-date/in sync with definitions (Hello JSON Forms!)
- Capacity to handle plugin preferences
- Easily extensible to different preference structures without needing to create "special" tables for complex cases in the future
- Very broad interpretation of "preferences", i.e. anything "configurable" should be able to captured under this same umbrella structure, so it's all one main concept
- Types easily generated for front-end for any type of preference structure

To this end, I think it's useful to distinguish between 3 different "concepts" that each require their own implementation (although obviously inter-related), would be good if we could agree on decent terminology for each of the following:
1. Preference type definitions -- the name/code and data structure/type for each possible preference (could be as simple as just "boolean" up to complex nested objects)
2. Preference "structure": i.e. what preferences (from 1) can be applied at what level (global/store/user), and the priority/fallback order. I'm thinking this is globally configurable, but we may want to impose hard restrictions for some (e.g. some prefs might have to be only global to make sense)
3. The actual preferences -- i.e. the values set for each user/store/system

I have written up a more detailed design/schema that can hopefully accomodate all the above, but I won't post that here yet as it still needs a bit more thought and it'd be good to see if it's even required based on further discussion, but the key points are:
- Loosely akin to Option 3
- Rust types as primary source of truth for each pref definition (not database schema -- database just gets JSON)
- Auto-generated JSON schemas based on types
- JSON Forms UI for configuration based on those schemas
- Things like "grouping" and other visual relationships between prefs can be defined in UI Json Schema, and don't need to be considered in the underlying schema
- Plugins can provide their own defintions for both the preference "type definitions" and "structure" (1 & 2 above -- probably need better naming to distinguish those), and the same core "engine" would generate schemas and UI.

Chris-Petty:
> I'm relaxed about migrating from an existing OG implementation, though ideally we'd accomodate that somehow

I'm not sure what exactly you mean by accommodate the OG implementation? I guess I can see some teething pains:

Old OMS sites will still rely on the old OG central server for existing prefs. So we'll have to configure anything twice, not sure if there is any mechanism that can address that without feeling very... spaghetti. e.g. the OMS central server could map the new structure to old structures and sync that to OG central...

Old OMS sites will also have to smoothly migrate the old pref data structures to the new, that shouldn't be too hard in any of the solutions IMO.

---

One note for option 3 is there is possibly some overhead, or at least it's moved the overhead. In Option 1 many prefs will just be essentially:

```sql
SELECT data FROM pref WHERE item = 'some pref' AND store_id = 'some id';
```
->
```json
{ "value": true }
```

Complex data structures in option 3 might just need a bit more parsing work. That's probably negligible though :).

Chris-Petty:
Notes from meeting:


- Thinking about delineation of when we generally use JSON in a database
- Note on carls requirements: _Everything_ configurable should be in preferences.
  - Could be too eager and everything is deep in JSON? Maybe perhaps discretion needed is enough
  - Rust types would strictly define everything that is in it though
  - I think we all like having JSON schema create the UI for many cases
- All solutions are some form of JSON, and changing that structure has same problems for different versions compat. across sync system (if changed type of existing pref is synced out from central, may break older remotes. Vice versa if older remote has permission to edit a pref and sync to central)
- Main difference in solutions is how to configure/structure that JSON and provide good typing and validation.
- Consensus that the "pref" table structure similar to OG rather than having a properties table. Preferring shallow preference structures and mapping them to structs for backend/fe consumption
- API approach:
  - A big struct that has the prefs for a given store/user. Backend does any resolution/merging of prefs that we may support and frontend gets that big object and applies it given contexts
  - Github puts permissions/prefs as fields on gql objects. E.g. querying all comments for an issue, an issue has a field `isEditable`. `isEditable` is populated based on the current user's org role, team role, being author of the comment etc. by the graphql backend.
    - Any of the solutions we've got so far should allow for doing this as we see fit. An example for us is say default printer for a particular machine could be `isDefault` field on the printer object, populated by backend gql logic querying preferences. 
  - Conclude we'll do the big boy struct endpoint as a baseline and the github style as we see fit. Any specific backend solution doesn't really affect frontend as such, gql as a contract driven by the principles above 

---

- Preferences table
  - Opt store id, user id, machine id
  - JSON for data
  - Backend merges records if needed
  - Base rust struct with defaults set
  - No type def in DB schema
  - Basic fields (e.g. booleans) can have UI obviously done
  - More complex things we can figure out, few options either building out on gql shape or on JSON forms (potentially generate from struct via some crate derive)

Spitball of some backend stuff - WIP (relates to https://github.com/msupply-foundation/open-msupply/issues/5138#issuecomment-2461246641):

Chris-Petty:
I think loosely speaking given the discussions had, we're really leaning towards doing option 1 as far as the DB schema is concerned

- 1 pref record per store pref per store. 
- Global pref is just a pref identified by `key` that hasn't got any store/user etc. id set
- Fall back to global pref is defaults defined in rust land
- as in last comment will hash out the deets of exactly how we'll do in rust land, few ways to shave the yak but probably don't need a kdd for that part?  I'd clarify and edit into the above comment, or add to consequences section of KDD. Really that could be detailed in an issue at this stage. (I desperately want this KDD closed haha)

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+This doesn't cover things such as machine or user preferences by it self. To follow this pattern for other requirements, you'd have to make the tables `user_preference` and `machine_preference` and so on. If a future requirement, cross overs of those prefs might not really work without getting crazy e.g. a user pref for a particular store.
+
+## Options
+
+### Option 1 - Take the lessons from OG `pref` table by doing the same
+
+The OG pref table is battle tested for over a decade:
+
+- We create a `preference` (or `pref`?) table with `id`, `item`, `store_id`, `user_id`, `machine_id` and `data` fields.
+- The lack of typing that ailed OG can readily be addressed with with structs in OMS. These structs may carry through to the frontend gql types as appropriate.
```
jmbrunskill, line 89:
```suggestion
- The lack of typing that ailed OG can readily be addressed with a big struct in OMS. Thus structs may carry through to the frontend gql types as appropriate.
```

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+- Sync needs deeper thought, but at a glance: store prefs are easily identified and synced by the `store_id`. `user_id` could be added to `change_log` to handle that (if needed to sync... or we sync users and their prefs everywhere 😉). Perhaps it's OK to not sync and lose machine_id records.
+- We store the pref data in an JSON text field. We deserialize this with serde into structs for each pref as expected.
```
jmbrunskill, line 91:
```suggestion
- We store the pref data in an JSON text field. We deserialize this into a struct in the rust code to which provides sensible defaults for anything missing.
```

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+  - This does mean that sync does not validate the `data` field, that validation occurs when the record is accessed. Though, we could quickly deserialize at sync buffer integration time just to check!
+- We can drop the `store_preference` table, but keep the repo the same for now, just querying from `preference` instead.
+
```
jmbrunskill, line 94:
```suggestion
- The list of preferences is defined by a single struct in the backend. The backend is responsible for merging Global, Store, and potentially if required `User Prefs/Machine Prefs`.
The single struct can have additional complexity, e.g. struct within structs as required. This may require additional mapping to graphql though.

```

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+_Pros:_
+
+- Consistency with OG can be easier to reason about between systems (provided it the OG pattern isn't hard to reason about to begin with...)
+- Just 1 table and logic (i.e. 1 repo etc.)
+- Generally flexible and battle tested
```
jmbrunskill, line 99:
```suggestion
- Generally flexible and battle tested
-  Preferences rust struct gives single source of the all preferences defined in the system
```

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+
+_Cons:_
+
+- Potentially more upfront effort right now
+- Create need for design convention on how to structure preferences. Should you make an individual pref for every little thing, or group them together into many keys in the same pref's JSON? How many is too many? If ever concurrent editing of same record with race on saving, bigger JSON means more data lost without merge strategy. We can nest - how much nesting is TOO FAR?
```
jmbrunskill, line 104: I'm not entirely sure if I still agree with this comment, given it's one struct we don't have different `prefs` there's just one set of prefs for the whole system (in my understanding?)

decisions/2024-12-03_preferences.md
```diff
@@ -0,0 +1,148 @@
...
+- Could be used by plugins to add additional prefs as needed
+
+_Cons:_
+
+- Could be larger payload as prefs aren't broken down into different areas
+- Need to insert prefs into database rather than just adding to code? Maybe there's a code first way to achieve the same thing? Enums?
+
+## Decision
+
+TBD
```
jmbrunskill, line 148:
```suggestion
Option 1
```

The list of available prefs will be defined by a struct in the Rust backend and available in graphql.
The UI will be manually created for each new pref as added to the struct.
This approach allows us to start small, with just Booleans for Global and Store Prefs initially, but can be expanded to more complex scenarios as required.

### Consequences

-  New prefs will required additional configuration UI to be manually created. But it should be a simple copy and paste type job
- There's still some potential risk for incompatible changes to come via sync to an old version for example. We'll need to be vigilant in code reviews to avoid this issue, and potentially favour creating new prefs if say a boolean pref needs to become a number for example.

jmbrunskill (changes requested):
I've put a bunch of suggestions in to bring the KDD up to speed based on our discussions and @Chris-Petty notes. I think once those or similar changes are in we're good to merge.

Thanks for co-ordinating Chris!

jmbrunskill:
> * Rust land applies resolution/merging of prefs into coherent struct (i.e. the pref struct shouldn't have optional fields right?)

Not sure if this is hard and fast rule? But I agree in principle I think.

jmbrunskill:
> * as in last comment, we will hash out the deets of exactly how we'll do in rust land, few ways to shave the yak but probably don't need a kdd for that part?  I'd clarify and edit into the above comment, or add to consequences section of KDD. Really that could be detailed in an issue at this stage. (I desperately want this KDD closed haha

Agreed, don't need a KDD for this part.

CarlosNZ:
> * Rust land applies resolution/merging of prefs into coherent struct (i.e. the pref struct shouldn't have optional fields right?)

You mean because they'll always be populated with a default if not defined in source JSON?

I think there is a possibilty of truly optional fields in more complex pref structures, but we can deal with that as it arises, probably?

jmbrunskill:
```rust
use serde::Serialize;
use serde::Deserialize;


#[derive(Serialize, Deserialize, Debug)]
struct ComplexStruct {
    pub something_here: String
}

#[derive(Serialize, Deserialize, Debug)]
struct Prefs {
    pub use_payments_in_prescriptions: Option<bool>,
    pub complex: Option<ComplexStruct>
}

impl Default for Prefs {
    fn default() -> Self { 
        Prefs {
            use_payments_in_prescriptions: Some(false),
            complex: None // Default to not having this complex thing?
        }
    }
}

fn get_merged_prefs(store_id: String) -> Prefs {

    // Would be a query for global pref record 
    let global_prefs = Prefs {
        use_payments_in_prescriptions: Some(false),
        complex: None
    };
    
    // Would be a query for store pref record
    let store_prefs = Prefs {
        use_payments_in_prescriptions: Some(false),
        complex: Some(ComplexStruct{ something_here: "I am here!".to_string() })
    };
    
    let mut merged = Prefs::default();
    
    
    // Possibly can use https://github.com/idubrov/json-patch or something from https://stackoverflow.com/questions/47070876/how-can-i-merge-two-json-objects-with-rust
    // Not sure how we're ensure we can control order of preference application though if so?
    
    // Merge global first
    if let Some(use_payments) = global_prefs.use_payments_in_prescriptions {
        merged.use_payments_in_prescriptions = Some(use_payments)
    };

    if let Some(complex) = global_prefs.complex {
        merged.complex = Some(complex)
    };
    
    // Then overwrite with Store pref if defined
    if let Some(use_payments) = store_prefs.use_payments_in_prescriptions {
        merged.use_payments_in_prescriptions = Some(use_payments)
    };
    if let Some(complex) = store_prefs.complex {
        merged.complex = Some(complex)
    };
    
    merged
}


fn main() {

    let merged_prefs = get_merged_prefs("ABC".to_string());
    println!("{:?}",merged_prefs);

    
}
```

FWIW, I'm thinking something like this for the backend.

Just to check we're on the same page :)

I found that we might need to make the backend struct fully optionals to allow for merging based on if something is set or not. I think we can translate to the frontend (graphql type) making sure there are no optionals at that point though.

Or we could have two structs one with optionals, and one without, which we'd map the optionals from global or store pref, to the non-optional one, but feels like that might be double handling...

decisions/2024-12-03_preferences.md
Chris-Petty: I'm not sure what you mean by having different prefs, but having a set (of different?) prefs? If it helps clarify there was a bit of implementation discussion that came up - basically I thought we were heading towards a bigger json pref that was all of a store's prefs, but was worried that we'd be pinned for potential sync situations, leading to clarifying another thing...

In OG store prefs and global prefs are both controlled on the central server, we're not worried about merging pref records through sync atm. That's likely to continue to be the case in OMS, though it'd be nice to keep the door open for some store prefs to be controlled on the remote site. At any rate, if prefs are being edited on both remote and central for a given store, individual records for prefs makes conflict resolution easier. Unless we use CRDTs 😁.

Anyway, so assuming that there are lots of individual prefs, to initialise the BIG STRUCT we'd do something like query all the prefs for a given store_ID then have fns on the struct for getting a particular pref from that selection or default. 

Does that make sense, or does that raise more questions? In fairness if we're all on the same page then this _con_ can be removed because the convention is taking shape already!

jmbrunskill: Does my rust code below help clarify the `BIG STRUCT` approach?
I was thinking for the first implementation at least, we'd have a single database row per store, and a single global row. The internal JSON structure would be the same, although we wouldn't need to serialise the `None` fields.

If we want to allow remote sites to edit their own store_prefs, then I think we go with the last update wins initially.
Could alway extend to allow some kind of CRDT for merging, or split store_prefs into `remote_store_prefs` and `central_store_prefs` as a simpler version of broken up prefs.
 
I'm personally, trying to avoid having lots of different prefs in the database to need more complex querying, mental overhead to decide is this a new pref row, is it part of another pref, etc. To me the simplicity of the big struct is that everything is there in one place. It's easy to follow through the flow from db records to what gets passed to in the UI.

I might be on a different page to others though so wanted to share that rust example to explain.
I agreed there might be limitations and challenges that come up with this approach, but I think it's a good way to start and only with good reason e.g requirements & use case do we extend it further.

Chris-Petty:
> You mean because they'll always be populated with a default if not defined in source JSON?

Yup that was my thinking! And not having to deal with Options everywhere or `boolean | undefined` in the frontend would be nicer I reckon.

Some OG context, there we have spaghetti of `pref_get_boolean(key: String, store_id: String, etc.)` functions that if the record doesn't exist then creates the pref record with a value of `false`. Similar funcs for strings, just `""`, numbers `0`... prefs are just written in phrasing that the scalar type default is the wanted default behaviour. That's why a lot of things are named in the negative (e.g. `inactive`) if the desired default is for the pref to be on.

Chris-Petty:
Very crude pseudo code based on a whiteboard doodle, certainly does not run in a playground 😁:

```rust
struct Prefs {
    prefs: Hashmap<PrefRow> // some hashmap structure of all the prefs /key/store_id ? Trying to avoid every pref fn from querying db.
}

// could just about put #[Object] here as it looks very similar to async_gql object implementations at a glance.
impl Prefs {
    pub fn new() -> Prefs { // prob need db connection. Perhaps store_id etc. too
        todo(); // queries for all the prefs for all the getters to use. Merging logic in here? Or the getters :think:
    }

    pub fn use_vaccine_module(&self, store_id: String) -> bool {
        todo!();  // Read from the prefs. or_default(false);
    } 

    // ..Every pref
    // Would have helper fns to make it fairly trivial and concise.
    // Possibly define setters/delete here too?
}

fn main() {
    let prefs = Prefs::new();
    let use_vaccine_module = prefs.use_vaccine_module();
}
```

There was another doodle but it was very similar to James's [above](https://github.com/msupply-foundation/open-msupply/pull/5729#issuecomment-2652295163), which is more fleshed out anyway.

jmbrunskill (approved):
Looks great to me! Thanks @Chris-Petty

Chris-Petty:
s e n d  i t

PR #5730, opened 2024-12-10
5025 hidden items showing on item usage and stock details reports

### LINKED ISSUE #5025: Hidden items are appearing in item usage and stock details reports
## What went wrong? 😲

Hidden items are appearing in item usage and stock details reports. 
Screenshot 2024-10-07 at 10 09 02 PM

## Expected behaviour 🤔
If item is hidden it is probably not meant to be ordered etc. Would expect it to not be showing.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #5025

Removes hidden items from report rendering.

There might be some confusion about what visible means for items in oms. We don't sync the hidden boolean from OG. in OMS visible means being part of a masterlist which is visible on the store.

I've updated explanations in comments.

Also graphql report queries updated to only show items both active and with stock on hand.
I think this should be sufficient because items cannot be hidden on OG if there is stock available.

Also added default sorting ascending and by item name on all reports

# 👩🏻‍💻 What does this PR do?

For reviewers - the fastest way to get the new report is like so:

- [ ] Do not stop your server running instance.
- [ ] Manually delete the item-usage report from your database
- [ ] Open a new terminal and run:

```
cargo build &&  ./target/debug/remote_server_cli build-standard-reports && ./target/debug/remote_server_cli upsert-reports-json
```

Which will rebuild and upsert reports to the database name set up in your .yaml file.

# 🧪 Testing

- [ ] Run commands above to delete, rebuild, and upsert reports
- [ ] Go to reports
- [ ] Render reports and don't see empty items

# 📃 Documentation

### DIFF
--- a/server/reports/stock-detail/2_3_0/src/query.graphql
+++ b/server/reports/stock-detail/2_3_0/src/query.graphql
@@ -2,7 +2,11 @@ query StockDetail($storeId: String, $itemCodeOrName: String) {
   stockLines(
     storeId: $storeId
     page: { first: 5000 }
-    filter: { itemCodeOrName: { like: $itemCodeOrName }, hasPacksInStore: true }
+    filter: {
+      itemCodeOrName: { like: $itemCodeOrName }
+      hasPacksInStore: true
+      isActive: true
+    }
     sort: { key: itemName, desc: false }
   ) {
     ... on StockLineConnector {
--- a/server/reports/stock-detail/2_3_0/src/query.graphql
+++ b/server/reports/stock-detail/2_3_0/src/query.graphql
@@ -7,7 +7,6 @@ query StockDetail($storeId: String, $itemCodeOrName: String) {
       hasPacksInStore: true
       isActive: true
     }
-    sort: { key: itemName, desc: false }
   ) {
     ... on StockLineConnector {
       nodes {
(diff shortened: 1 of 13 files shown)


server/reports/stock-detail/2_3_0/src/query.graphql
```diff
@@ -2,8 +2,11 @@ query StockDetail($storeId: String, $itemCodeOrName: String) {
   stockLines(
     storeId: $storeId
     page: { first: 5000 }
-    filter: { itemCodeOrName: { like: $itemCodeOrName }, hasPacksInStore: true }
-    sort: { key: itemName, desc: false }
+    filter: {
+      itemCodeOrName: { like: $itemCodeOrName }
+      hasPacksInStore: true
+      isActive: true
```
andreievg, line 8: I got confused about this, i thought it won't work (graphql won't allow it), but it seems to work somehow ? (isActive is not part of filter input)

andreievg (commented):
Looks great, one comment about isActive on stockLine, and do we need to commit standard_reports.json ?

server/reports/stock-detail/2_3_0/src/query.graphql
roxy-dao: Have added the filter on stock line

andreievg:
I updated from 2.4.1 (to get the other report fix in) and then generated standard reports, thanks @fergie-nz thanks @roxy-dao

andreievg (approved):
Looks great thanks

PR #5734, opened 2024-12-10
5733 remove numeric input mode for number inputs

### LINKED ISSUE #5733: Can't enter negative numbers on Samsung tablet
## What went wrong? 😲

Can't enter negative numbers in R&R fields on tablet, even though the field accepts negative (i.e. on browser)

This is because Samsung numeric keyboard doesn't have a `-` key!!

## Expected behaviour 🤔

Can enter negative numbers

## How to Reproduce 🔨

- On samsung tablet
- try to enter negative number in R&R form `Adjustments` columns
- Can't!

## Your environment 🌱

- Open mSupply Version: v.2.3.4

- Platform:

- Database type:

lache-melvin:
Fixes #5733

# 👩🏻‍💻 What does this PR do?

Couldn't enter negative values in R&R forms on samsung tablet!

This PR switches back to alphanumeric keyboard, so users have access to the `-` key as needed.

R&R form and demographic growth rates are the only two places where we allow entering negative values... it feels a shame to remove the simplicity of the numeric keyboard everywhere for these two cases?

Considered something like `inputMode = allowNegative ? "text" : "numeric"` - but this was quite ick as you clicked through negative and non-negative fields in the R&R form. Probably best just to be consistent, but curious for thoughts! 😢 

Other interim solution for demo has been to install Gboard app, and set this as default keyboard. This keyboard has `-` key and allows us to keep the numeric keyboard. But I don't think this is a feasible option long term...

May we keep an eye out for the inevitable update where the `-` comes back!

# 🧪 Testing

- [ ] On Samsung tablet
- [ ] Have R&R form
- [ ] Edit a row 
  - [ ] Non-negative fields, like quantity received, you still can't make negative
  - [ ] Adjustments field, you can type a negative number

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -195,6 +195,7 @@ export const NumericTextInput = React.forwardRef<
       noFormatting = false,
       fullWidth,
       endAdornment,
+      inputMode,
       ...props
     },
     ref
@@ -265,8 +266,6 @@ export const NumericTextInput = React.forwardRef<
 
     return (
       <BasicTextInput
-        // NOTE: not setting input mode as numeric here, because on Samsung tablets,
-        // the numeric keyboard doesn't allow entering negative numbers!
         ref={ref}
         sx={{
           '& .MuiInput-input': {
@@ -275,6 +274,7 @@ export const NumericTextInput = React.forwardRef<
           },
           ...sx,
         }}
+        inputMode={inputMode ?? 'numeric'}
         InputProps={{
           endAdornment: endAdornment ? (
             <InputAdornment position="end" sx={{ paddingBottom: '2px' }}>
--- a/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
@@ -265,6 +265,8 @@ export const NumericTextInput = React.forwardRef<
 
     return (
       <BasicTextInput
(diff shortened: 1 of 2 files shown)


client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
```diff
@@ -273,7 +275,6 @@ export const NumericTextInput = React.forwardRef<
           },
           ...sx,
         }}
-        inputMode="numeric"
```
jmbrunskill, line 276: Could this be based on the `allowNegative` bool?
Feels like it might be an ok trade off? Or maybe that's just a weird ui thing?

Personally I'm ok with always seeing the whole keyboard, but would require switching to number mode to enter numbers which is a bit of a slow down...

jmbrunskill (commented):
I feel like we need some use feedback on this one, I imagine having numeric inputs at least for things like stock take where we need to enter lots of numbers is probably important???

client/packages/common/src/ui/components/inputs/TextInput/NumericTextInput.tsx
jmbrunskill: Sorry that's what you said you didn't like :( Need to read the PR properly first !

jmbrunskill: Maybe yet another prop to control this so we can limit it just to R&R forms for now?

lache-melvin:
@jmbrunskill have updated to your suggestion, alphanumeric keyboard on the R&R form but keep the number pad elsewhere 🙏

jmbrunskill (approved):
Looks great, thanks Lache!

Issue #5743, opened 2024-12-10
Remove site_id column from contact_form


GeronimoJohn:
EDIT: see updated solution below

Currently we've set up the site_id for Contact_Form to a String or a text. Depending on the context that it's in. 

The actual type that we get for site_id is `i32`. To be specific we get an optional `i32`. We need to update files that are relating to contact form and is using site_id in order to match the type that is given. 

Service layer `contact_form`
- current type is string which is to be converted to be a optional `i32`
- we need to support the case of this value being optional within this file

Repository layer `contact_form`
- current type is text which is to be converted to be `integer`

lache-melvin:
Hi - I think easiest option actually is to remove site_id from contact form row. We already have store_id, which we will need to join to later to get extra info for email anyway. store_row has site_id already 🚀

PR #5744, opened 2024-12-10
5603 service layer contact

### LINKED ISSUE #5603: Add service layer to create feedback form
Implement service layer logic to save a feedback form.

Add a new `contact_form` service, with an `insert_contact_form` function.

Should accept an input with the following properties:

Key | Type
-- | --
id | String
type | ContactFormType
reply_email | String
body | String

`ContactFormType` enum: `Feedback` | `Support`

Validation should include:
- Feedback form with that ID doesn't already exist
- Body is not empty
- Email is valid format

Other fields on feedback_form_row: 
- `created_datetime` should be set to `Utc::now()`
- `site_id` should come from the `site_info_service`

aimee-mcneil-melville:
Fixes #5603 

# 👩🏻‍💻 What does this PR do?

Create a service layer for a contact form. This form is 
Includes:
- [ ] Insert function
- [ ] Validate file
- [ ] Generate file
- [ ] Unit tests

Validation includes email regex, empty email, empty body, and if id already exists

Tests include several email regex variations, empty email, empty body, if id already exists, and Success

This will link to the graphql layer and the repository layer for the back end piece of this form

Site_id types to be adjusted in #5743 

# 🧪 Testing

- [ ] Automated tests pass

# 📃 Documentation

  1.
  2.


server/service/src/contact_form/insert/validate.rs
```diff
@@ -0,0 +1,37 @@
...
+) -> Result<(), InsertContactFormError> {
+    if check_contact_form_record_exists(&input.id, connection)? {
+        return Err(InsertContactFormError::ContactIdAlreadyExists);
+    }
+
+    if input.reply_email.is_empty() {
+        return Err(InsertContactFormError::EmailDoesNotExist);
+    }
+
+    let email_regex = Regex::new(r"^[\w\.=-]+@[\w\.-]+\.[\w]{2,}$").unwrap();
```
jmbrunskill, line 18: This regex will fail on an address like this `james+test@gmail.com` which is a valid email.
If we're going to have a email regex in open-mSupply I think it should have tests associated with in, and probably live in the `utils` package.

Validating email addresses is difficult, and I think we're generally better to be super conservative and then just let the email fail to send if it's invalid? 

In notify currently we're using this in the frontend, I think a frontend check is fine here for this purpose...

https://github.com/msupply-foundation/notify/blob/e97525854006a6e247cf34b9aef5222c37891cf3/frontend/packages/common/src/utils/regex/RegexUtils.tsx#L49

If you do want a backend check, I'd potentialy just do this. `^[^@]+@[^@]+$`
That means any characters except for an `@` sign followed by an `@` sign and then any other non-`@` characters.
That basically just ensures we have only one @ sign, so it would accept something like `james @msupply.foundation` which isn't valid but remember this record will be synced so we'll still have the record even if the email is invalid...

lache-melvin: @jmbrunskill - confirming, when you say "conservative", you mean conservative by erring on the side of accepting the email, rather than conservative preventing potentially invalid emails? 😁 

I was wondering if it was better to try constrain more... as if record syncs but email never sends due to invalid address, how long until that is discovered (particularly if it was a support request that was sent)? I suppose we could recommend configuring notify alerts against mail send failures?

jmbrunskill: > @jmbrunskill - confirming, when you say "conservative", you mean conservative by erring on the side of accepting the email, rather than conservative preventing potentially invalid emails? 😁

Basically yes, even if the email is syntactically valid. it can be miss entered or have typos and might not send anyway, so that situation needs to be accounted for. If I understand correctly, we'll send these to support or some known email address anyway, so it would only be the email to the user that would go missing.

I'm just aware how many times email address haven't been considered valid that should have been, some websites don't accept `msupply.foundation` as a domain for example.

It's just my 2cents though, feel free to try for something more strict, but I'm worried it might create more problems than it solves :)

Also might be worth considering if you can use the user's email from the `user_account` table, pre-populate for them mabye? Not sure how often this is filled in though...

lache-melvin: Yes have a future todo to pre-populate with user account email if defined 🎉 

Yeah that's true, it should make it to support regardless, which is the main thing 🙏 Thanks for your input James, makes sense, we'll update to be less restrictive :)

aimee-mcneil-melville: Great feedback, thank you James. I have updated the regex to ```[^@]+@[^@]+\.[^@]+``` very similar to your suggestion but requiring the .  to get the basic structure in

Would this be a refactor issue to bring the validation into the utils?

jmbrunskill: > Would this be a refactor issue to bring the validation into the utils?

It wouldn't hurt to create refactor issue for it...

server/service/src/contact_form/insert/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+use validate::validate;
+
+use crate::service_provider::ServiceContext;
+
+#[derive(PartialEq, Debug)]
+pub enum InsertContactFormError {
+    ContactIdAlreadyExists,
+    EmailIsInvalid,
+    EmailDoesNotExist,
+    MessageDoesNotExist,
```
lache-melvin, line 18: Would rename these:

 - ContactFormAlreadyExists
 - `DoesNotExist` -> `NotProvided`

"does not exist" is convention for a DB record that we couldn't find, rather than an input not provided the way we expect 🙏

lache-melvin (changes requested):
Looks great Aimee! 🚀  Just the error names would be good to clean up before we merge this 🙏 

And yes please on the refactor issue for email validation utils!

server/service/src/contact_form/insert/generate.rs
```diff
@@ -0,0 +1,41 @@
+use chrono::Utc;
+use repository::contact_form_row::{ContactFormRow, ContactType};
+
+use super::InsertContactForm;
+
+//inputs from graphql layer
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Great makes sense thank you!

server/service/src/contact_form/insert/mod.rs
aimee-mcneil-melville: Sounds good!

lache-melvin (approved):
ship it! 🚀

lache-melvin:
@GeronimoJohn sorry will be some types to update when you merge again 😅

PR #5754, opened 2024-12-11
Handle optional gs1 warranty dates

jmbrunskill:
Fixes potential issue of missing waranty dates in GS1 codes
Especially since this standard isn't verified yet, we need to allow this to be empty when scanning a GS1 Matrix Barcode for assets

# 👩🏻‍💻 What does this PR do?

Makes waranty dates optional

Haven't create a test GS1 for this but I think the code change is safe.

# 🧪 Testing

- [ ] This would require creating a new GS1 barcode without a `91` field.

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
LGTM, thanks James :)

PR #5755, opened 2024-12-11
Vagabond/#5578

### LINKED ISSUE #5578: Rename 'useJsonForms' hook to avoid confusion
## The suggestion

There are currently two different `useJsonForms` hooks in use -- one is part of the core JSONForms package and provides full context of the JSON Forms state. The other is one we've created which we use to manage the forms and provide an overall "wrapper" for the entire structure.

This can be mildly confusing -- most individual components use the first one, and the implementations of JSON forms use the latter. It would be fairly trivial to rename our custom hook to something distinct, for example "useJsonFormsHandler" or "useJsonFormsWrapper".

It's not a huge point of confusion, but it's very simple to fix, so probably worth it IMHO.

- Less confusion when trying to understand JSON forms code
- Easier to search for the correct thing

Doubt it, the compiler will yell at us if we miss any references to the renamed hook

Less than 1 hour.

### Agreed Solution

... ? (My vote is for `useJsonFormsHandler`)

vagabond-0:
Fixes #5578

# 👩🏻‍💻 What does this PR do?

Rename every the function useJsonForm to useJsonFormsHandler

To avoid confusion

No change in ui

Fixes #

# 👩🏻‍💻 What does this PR do?

Nil

NIl

NIl

### DIFF
--- a/client/packages/programs/src/JsonForms/useJsonForms.tsx
+++ b/client/packages/programs/src/JsonForms/useJsonForms.tsx
@@ -106,7 +106,7 @@ export type JsonFormData<R> = {
 };
 
 /**
- * This hook add provides functionality to save form data and keep track if form data has been
+  * This hook add provides functionality to save form data and keep track if form data has been
  * modified.
  *
  * What data is shown and how it is saved can be customized through the `jsonFormData` form
--- a/client/packages/programs/src/JsonForms/useJsonForms.tsx
+++ b/client/packages/programs/src/JsonForms/useJsonForms.tsx
@@ -112,7 +112,7 @@ export type JsonFormData<R> = {
  * What data is shown and how it is saved can be customized through the `jsonFormData` form
  * parameter.
  */
-export const useJsonForms = <R,>(
+export const useJsonFormsHandler = <R,>(
   config: JsonFormsConfig,
   jsonFormData: JsonFormData<R>
 ) => {
--- a/client/packages/system/src/Patient/PatientView/PatientView.tsx
+++ b/client/packages/system/src/Patient/PatientView/PatientView.tsx
@@ -135,8 +135,8 @@ const PatientDetailView = ({
         isCreating: true,
       };
     } else if (!!currentPatient && !currentPatient.document) {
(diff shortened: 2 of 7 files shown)


client/packages/programs/src/JsonForms/useJsonForms.tsx
```diff
@@ -106,13 +106,13 @@ export type JsonFormData<R> = {
 };
 
 /**
- * This hook add provides functionality to save form data and keep track if form data has been
+  * This hook add provides functionality to save form data and keep track if form data has been
```
CarlosNZ, line 109: Could you remove this whitespace change, please?

client/packages/system/src/Patient/PatientView/PatientView.tsx
```diff
@@ -199,7 +199,7 @@ const PatientDetailView = ({
       };
 
   const { JsonForm, saveData, isSaving, isDirty, validationError } =
-    useJsonForms(
+  useJsonFormsHandler (
```
CarlosNZ, line 202: Revert whitespace change (If you have Prettier set up, this should happen automatically)

client/packages/system/src/Patient/ProgramEnrolment/Components/DetailModal.tsx
```diff
@@ -61,7 +61,7 @@ export const ProgramDetailModal: FC = () => {
     handleSave
   );
   const { JsonForm, isLoading, isSaving, saveData, isDirty, validationError } =
-    useJsonForms(
+  useJsonFormsHandler(
```
CarlosNZ, line 64: Same here.

CarlosNZ (changes requested):
Thanks for contributing @vagabond-0

Aside from those minor whitespace changes, could you also rename the *file* to `useJsonFormsHandler.tsx` as well please?

Thanks

vagabond-0:
Removed white spaces and changed the file name to useJsonFormsHandler.tsx

CarlosNZ (approved):
LGTM!

Thanks for contributing.

PR #5759, opened 2024-12-11
Update inbound shipment generate.rs to use default item packsize

### LINKED ISSUE #5561: Manual Inbound Shipment - default pack size not displayed when adding lines using "Add from a Master List" button
## What went wrong? 😲

possibly similar to https://github.com/msupply-foundation/open-msupply/pull/5558

While the system is picking up the default pack size when adding a line using the "Add an item" button, it does not do the same when adding the same item but using the "Add from a master list" button: 

**Add an item button**

Default pack size (20) is automatically displayed

Screenshot 2024-11-26 at 12 03 33 PM

**Add from a master list**

Pack size is showing 1 while it should 20

Screenshot 2024-11-26 at 12 06 15 PM

## Expected behaviour 🤔

See above
## How to Reproduce 🔨

See above

## Your environment 🌱

- Open mSupply Version: 2.3.02
- Platform: browser (chrome)

noel-yeldos:
Fixes #5561

# 👩🏻‍💻 What does this PR do?
Fix issue where items added to inbound shipment from masterlist had a default pack size of 1 instead of item default pack size

Screenshot 2024-12-12 at 10 55 41 AM
Screenshot 2024-12-12 at 10 56 07 AM

# 🧪 Testing

- [ ] Go to Replenishments -> Inbound Shipments
- [ ] Click on existing shipment or create a new one
- [ ] Click add from masterlist button
- [ ] Select a list and add items
- [ ] Check to see that pack size is using item default rather than settting to 1 (Note some packs have a default size of 1)
- [ ] Click on item to see that pack size displayed matches that from list view

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/invoice/inbound_shipment/add_from_master_list.rs
+++ b/server/service/src/invoice/inbound_shipment/add_from_master_list.rs
@@ -38,7 +38,7 @@ pub fn add_from_master_list(
         .connection
         .transaction_sync(|connection| {
             let invoice_row = validate(connection, &ctx.store_id, &input)?;
-            let new_invoice_line_rows = generate(ctx, invoice_row, &input)?;
+            let new_invoice_line_rows = generate(connection, ctx, invoice_row, &input)?;
 
             let invoice_line_row_repository = InvoiceLineRowRepository::new(connection);
 
@@ -83,6 +83,7 @@ fn validate(
 }
 
 fn generate(
+    connection: &StorageConnection,
     ctx: &ServiceContext,
     invoice_row: InvoiceRow,
     input: &ServiceInput,
@@ -107,7 +108,7 @@ fn generate(
         .map(|master_list_line| master_list_line.item_id)
         .collect();
 
-    generate_empty_invoice_lines(ctx, &invoice_row, items_ids_not_in_invoice)
+    generate_empty_invoice_lines(connection, ctx, &invoice_row, items_ids_not_in_invoice)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


server/service/src/invoice/inbound_shipment/generate.rs
```diff
@@ -24,7 +24,7 @@ pub fn generate_empty_invoice_lines(
                     location_id: None,
                     batch: None,
                     expiry_date: None,
-                    pack_size: 1.0,
+                    pack_size: item.default_pack_size,
```
lache-melvin, line 27: Per the store pref `pack_to_one` thing - I think we actually need to factor in the same thing here!

You can use the `get_store_preferences` function, like here:
https://github.com/msupply-foundation/open-msupply/blob/20f5074f9cd298f89ad180cf04ea110d629da2b5/server/service/src/invoice_line/stock_in_line/insert/generate.rs#L30

If pack_to_one is true, should be 1.0, otherwise can use the default pack size 🙏

lache-melvin (changes requested):

lache-melvin (approved):
Epic thanks for changes! 🔥

server/service/src/invoice/inbound_shipment/add_from_master_list.rs
```diff
@@ -107,7 +108,7 @@ fn generate(
         .map(|master_list_line| master_list_line.item_id)
         .collect();
 
-    generate_empty_invoice_lines(ctx, &invoice_row, items_ids_not_in_invoice)
+    generate_empty_invoice_lines(connection, ctx, &invoice_row, items_ids_not_in_invoice)
```
jmbrunskill, line 111: @noel-yeldos FYI `connection` is available from the `ServiceContext` object `ctx`, to access the connection you can just do `ctx.connection`, so adding the extra parameter here isn't necessary...

noel-yeldos: @jmbrunskill ah good to know! I'll fix as sidequest on current PR :)

PR #5760, opened 2024-12-11
5604 graphql layer contact

### LINKED ISSUE #5604: Add insert_feedback_form mutation, and use from frontend
Add graphql layer for creating the feedback form.

Input should include:

Key | Type
-- | --
id | String
reply_email | String
body | String

Other fields to pass to domain/service input:
- `store_id` should be passed in as separate graphQL parameter (as per our convention)
- `user_id` can be pulled from `ctx`

Steps:
- [ ] Add `FeedbackFormMutations`, which includes a `insert_feedback_form` method
- [ ] Calls the feedback form service > insert_feedback_form
- [ ] Add new `operations.graphql` with the mutation & generate types
- [ ] Use this mutation in `useFeedbackForm`

GeronimoJohn:
Fixes #5604 

# 👩🏻‍💻 What does this PR do?

Created the ContactForms graphql layer and added one insert mutation: `insert_contact_form`.  This mutation was connected to the ContactForms service layer

The newly created mutation was integrated into `useFeedbackForm` and used within the existing FeedbackForm on the help page. 

In order to establish the connection with our api, we also created a useFeedbackFormGraphQL hook

Site_id to be removed in #5743 

`totalStocksOnHand` needed to be generated and added in the graphql schema within Client. This was part of Team Ruru's changes and have been communicated that we are doing this fix within this PR.

# 🧪 Testing

Relied on the service layer for tests

# 📃 Documentation

  1.
  2.


server/graphql/contact_form/src/mutations/insert.rs
```diff
@@ -0,0 +1,105 @@
...
+use graphql_core::{standard_graphql_error::validate_auth, ContextExt};
+use repository::contact_form_row::ContactFormRow;
+use service::{
+    auth::{Resource, ResourceAccessRequest},
+    contact_form::{InsertContactForm as ServiceInput, InsertContactFormError as ServiceError},
+};
+
+#[derive(InputObject)]
+#[graphql(name = "InsertContactFormInput")]
+pub struct InsertContactFormInput {
```
lache-melvin, line 12: NIT: don't need the `#[graphql(..)]` renames now that the struct name is the same 😁

lache-melvin (approved):
LGTM! Epic to have this flow in and saving 🔥

server/graphql/contact_form/src/mutations/insert.rs
GeronimoJohn: Applied!

PR #5767, opened 2024-12-12
remove contact forms side id

### LINKED ISSUE #5743: Remove site_id column from contact_form

GeronimoJohn:
Fixes #5743 

# 👩🏻‍💻 What does this PR do?

This drops the `site_id` column from the `contact_form` table. After further discussions and digging into things we've realised that we don't need `site_id`. 

We are to drop the use of this field in the different functions within our different layers. Ensuring that the overall flow and functionality from the `Client` all the way to `Server` still works.

Dependent on the PRs #5744 and #5760. With 5744 the priority to go in first

# 🧪 Testing

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
🔥

.github/workflows/server-cargo-clean.yaml
```diff
@@ -0,0 +1,40 @@
+on: 
+  pull_request: 
+    branches: 
+      - main
+      - develop
+    paths: 
+      - 'server/**'
```
lache-melvin, line 7: Awesome! Yeah let's remove this piece now that CI is passing, once we merge into develop (our default branch) the workflow_dispatch should be available :)

lache-melvin (approved):
Ship it! ...after removing PR trigger 🙏

PR #5788, opened 2024-12-13
Move selection to be first column in list view tables

### LINKED ISSUE #5655: Move table select checkboxes to the first column
Tables throughout the app have the `Selection` column as the last column:

Screenshot 2024-12-02 at 9 59 02 AM

This is hard to access/use on smaller screens.

Go through the frontend and move the `Selection` column from the last to the first column on all tables.

Context:
- Throughout the frontend code, tables are configured by defining an array of column definitions, and passing this into a `useColumns` hook. You'll want to review all implementations of this hook.
- The column should use `GenericColumnKey.Selection`, but some places have the hard coded string `'selection'` - please use this opportunity to update all the `GenericColumnKey.Selection`!

noel-yeldos:
Fixes #5655

# 👩🏻‍💻 What does this PR do?
Move selection to be first column and also rename ```'selection'``` to ```genericColumnKey.Selection```

Changes needed to make it easier for user to select rows in list view

Screenshot 2024-12-13 at 12 59 39 PM
Screenshot 2024-12-13 at 1 05 40 PM
Screenshot 2024-12-13 at 1 05 57 PM
Screenshot 2024-12-13 at 1 06 15 PM
Screenshot 2024-12-13 at 1 06 31 PM
Screenshot 2024-12-13 at 1 06 54 PM
Screenshot 2024-12-13 at 1 07 10 PM
No screenshot for programs
Screenshot 2024-12-13 at 1 09 11 PM
Screenshot 2024-12-13 at 1 09 29 PM
Screenshot 2024-12-13 at 1 09 48 PM
Screenshot 2024-12-13 at 1 10 09 PM
Screenshot 2024-12-13 at 1 10 23 PM

Note sure why padding is being added to the left of the column in Inventory -> Locations and Stocktakes

# 🧪 Testing

- [ ] Check the areas in the screenshots 
- [ ] See that selection column is now the first column in the tables
- [ ] Test that selection is working correctly

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/ListView/ListView.tsx
+++ b/client/packages/coldchain/src/Equipment/ListView/ListView.tsx
@@ -16,6 +16,7 @@ import {
   DotCell,
   RouteBuilder,
   ColumnFormat,
+  GenericColumnKey,
 } from '@openmsupply-client/common';
 import { useAssets } from '../api';
 import { Toolbar } from './Toolbar';
@@ -52,7 +53,7 @@ const AssetListComponent: FC = () => {
 
   const columnsToCreate: ColumnDescription<AssetRowFragment>[] = [];
   if (isCentralServer)
-    columnsToCreate.push({
+    columnsToCreate.push(GenericColumnKey.Selection, {
       key: 'store',
       label: 'label.store',
       accessor: ({ rowData }) => rowData.store?.code,
@@ -121,8 +122,7 @@ const AssetListComponent: FC = () => {
       key: 'notes',
       label: 'label.notes',
       sortable: false,
-    },
-    'selection'
+    }
   );
 
   const columns = useColumns(
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/columns.ts
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/columns.ts
@@ -250,7 +250,7 @@ export const useRequestColumns = () => {
       accessor: ({ rowData }) => rowData.linkedRequisitionLine?.approvalComment,
     });
   }
-  columnDefinitions.push(GenericColumnKey.Selection);
+  columnDefinitions.unshift(GenericColumnKey.Selection);
 
   const columns = useColumns<RequestLineFragment>(
     columnDefinitions,
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/columns.ts
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/columns.ts
@@ -50,6 +50,7 @@ export const useRequestColumns = () => {
   const { getError } = useRequestRequisitionLineErrorContext();
 
   const columnDefinitions: ColumnDescription<RequestLineFragment>[] = [
+    GenericColumnKey.Selection,
     getCommentPopoverColumn(),
     [
       'itemCode',
@@ -250,7 +251,6 @@ export const useRequestColumns = () => {
       accessor: ({ rowData }) => rowData.linkedRequisitionLine?.approvalComment,
     });
   }
-  columnDefinitions.unshift(GenericColumnKey.Selection);
 
   const columns = useColumns<RequestLineFragment>(
     columnDefinitions,
--- a/client/packages/system/src/Asset/ListView/ListView.tsx
+++ b/client/packages/system/src/Asset/ListView/ListView.tsx
@@ -11,6 +11,7 @@ import {
   useToggle,
   useIsCentralServerApi,
   ColumnDescription,
+  GenericColumnKey,
 } from '@openmsupply-client/common';
 import { AssetCatalogueItemFragment, useAssetData } from '../api';
 import { Toolbar } from './Toolbar';
@@ -79,7 +80,7 @@ const AssetListComponent: FC = () => {
     },
   ];
 
-  if (isCentralServer) columnDescriptions.push('selection');
+  if (isCentralServer) columnDescriptions.unshift(GenericColumnKey.Selection);
   const columns = useColumns<AssetCatalogueItemFragment>(
     columnDescriptions,
     {
--- a/client/packages/system/src/Asset/ListView/ListView.tsx
+++ b/client/packages/system/src/Asset/ListView/ListView.tsx
@@ -41,6 +41,7 @@ const AssetListComponent: FC = () => {
   const importModalController = useToggle();
 
   const columnDescriptions: ColumnDescription<AssetCatalogueItemFragment>[] = [
+    ...(isCentralServer ? [GenericColumnKey.Selection] : []),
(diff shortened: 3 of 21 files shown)


client/packages/system/src/Location/ListView/ListView.tsx
```diff
@@ -38,6 +39,7 @@ const LocationListComponent: FC = () => {
   const t = useTranslation();
   const columns = useColumns<LocationRowFragment>(
     [
+      GenericColumnKey.Selection, // adds padding to the left not sure why
```
GeronimoJohn, line 42: Just a bit unsure which padding in the left we're referring to. Would it be the green or the blue one?
Screenshot 2024-12-16 at 8 41 29 AM

GeronimoJohn (approved):
Looks good to me! Awesome stuff

GeronimoJohn (commented):
Sorry for the build ups. Forgot one bit, we should probably remove the comment once we understood what's going on under the hood 😄

client/packages/coldchain/src/Equipment/ListView/ListView.tsx
```diff
@@ -52,7 +53,7 @@ const AssetListComponent: FC = () => {
 
   const columnsToCreate: ColumnDescription<AssetRowFragment>[] = [];
   if (isCentralServer)
-    columnsToCreate.push({
+    columnsToCreate.push(GenericColumnKey.Selection, {
```
roxy-dao, line 56: This can just be 'selection', if you check `ColumnDefinitionSetBuilder.ts` you can see a list of predefined columns (:

client/packages/system/src/Location/ListView/ListView.tsx
noel-yeldos: @GeronimoJohn thanks for checking! It doesn't add padding to all columns. It's only visible in my second screenshot (Stocktakes) and my last screenshot (Inventory Locations).

lache-melvin (changes requested):
Hi @noel-yeldos - would you be able to apply this change for the `DetailView` components that use tables as well? E.g. Outbound Shipment: 

Screenshot 2024-12-16 at 3 42 23 PM

client/packages/system/src/Location/ListView/ListView.tsx
lache-melvin: It's not padding... but when the screen gets wider, it needs somewhere to put the extra width.

If we have some columns with set widths, and some with none, then the extra width only goes on the columns with no set width, because they can grow as much as needed. (This is usually the case)

Checkbox does have width, so usually stays fixed at 60:

https://github.com/msupply-foundation/open-msupply/blob/5b2aecdc71c6c48902b3524b6e9b2f0cda04a75c/client/packages/common/src/ui/layout/tables/columns/CheckboxSelectionColumn.tsx#L34-L40

 However, if _all_ the columns have a defined width (a lot of them are defined in `ColumnDefinitionSetBuilder` if you're curious 😅) then it shares the extra space between all the columns. 


I guess we didn't notice this when select was on the end, because of the `align: ColumnAlign.Right`. Would be an improvement to make it `align: ColumnAlign.Left,` now!
Screenshot 2024-12-16 at 4 34 00 PM

Potentially, could also remove the `width` property from some columns of those two tables that are playing up, so only the columns we actually want to grow will grow 🤷‍♀️

GeronimoJohn: Right sorry misread that. Comment makes sense!

I think I found where the issue is happening. The component `getCheckboxSelectionColumn`. Sets the check box's alignment to be on the right. 

Screenshot 2024-12-16 at 4 43 43 PM

Changing it to left seems to fix the issue. 

Screenshot 2024-12-16 at 4 46 28 PM

The extra spacing are either being caused by Material UI or just how we've set up `DataTable`. Regardless of how the width is set, `ColumnAlign.Right` just uses that given space and yea puts it on right.

IMO we should probably move the alignment to left now. 😃

noel-yeldos: Thank you both! Updated here 17be43b6324e0a82409a1a5e4244ac890a4e219e :)

noel-yeldos:
> Hi @noel-yeldos - would you be able to apply this change for the `DetailView` components that use tables as well? E.g. Outbound Shipment:
> 
> Screenshot 2024-12-16 at 3 42 23 PM

Thanks for catching this! Updated here: 3c7ba087fdec091cdbf0ec862416c89ee7934267 :)

client/packages/system/src/Location/ListView/ListView.tsx
GeronimoJohn: Hahaha didn't see Laché's comment

GeronimoJohn (approved):
LGTM! 🔥

client/packages/requisitions/src/RequestRequisition/DetailView/columns.ts
```diff
@@ -250,7 +250,7 @@ export const useRequestColumns = () => {
       accessor: ({ rowData }) => rowData.linkedRequisitionLine?.approvalComment,
     });
   }
-  columnDefinitions.push(GenericColumnKey.Selection);
+  columnDefinitions.unshift(GenericColumnKey.Selection);
```
lache-melvin, line 253: Thanks for changes Noel! One last little request from me - I do think just putting `GenericColumnKey.Selection` as the first array item, as you have in the other places, would be clearer than the unshifts 🙏

noel-yeldos: Fixed here 3b8bb42be443cadfdef8514732cf9c3c9861af50 :)

client/packages/system/src/Asset/ListView/ListView.tsx
```diff
@@ -79,7 +80,7 @@ const AssetListComponent: FC = () => {
     },
   ];
 
-  if (isCentralServer) columnDescriptions.push('selection');
+  if (isCentralServer) columnDescriptions.unshift(GenericColumnKey.Selection);
```
lache-melvin, line 83: hehe one more 😅

lache-melvin (approved):
Epic thanks Noel! One more unshift to replace, merge once ready 🔥

client/packages/system/src/Asset/ListView/ListView.tsx
noel-yeldos: Ah the elusive one! Fixed here 808a5cb26c5f6e0847d99d6c64b27ea7bee6d4bc :)

alainsussol:
Will be tested while testing the whole GEDSI Feature

PR #5796, opened 2024-12-13
actions footer component

### LINKED ISSUE #5738: Replace the `Actions` dropdown with an actions menu at the bottom of the screen
Screenshot 2024-12-11 at 10 12 14 AM

- [ ] Remove the Actions dropdown from the Toolbar
- [ ] Define an `ActionsFooter`, with buttons for the same options that were in the dropdown
- [ ] When any of the rows are selected, replace the normal footer (the one with the statuses etc) with the ActionsFooter

This PR is to prove out the implementation. Just implement this in `Outbound Shipment` detail view. We can apply the pattern across other areas in a separate refactor PR.

GeronimoJohn:
Fixes #5738 

# 👩🏻‍💻 What does this PR do?

We've created a new ActionsFooter component. Where it moves all the action that users can take when a row or rows are selected. This effectively removes the action dropdown menu in outbound shipment details view page. 

This looks out for if there are any rows are selected or `selectedIds`. If there are, it shows actions footer otherwise it renders the default footer.

# 🧪 Testing

- [ ] Navigate to Outbound Shipments
- [ ] Selected a store
- [ ] Tick the checkbox for a row or rows
- [ ] Actions Footer should render at the bottom of the page
- [ ] Actions Dropdown menu should not be in the page

# 📃 Documentation

  1.
  2.


lache-melvin:
Just copying over Mark's feedback, which he left on the issue 😅 

This is awesome! 😀 
very nice to use, much more intuitive

Couple of minor things:
* can you check the height, the footer is taller than the current footer and it feels like it's using more space than necessary
* can you remove the shouty case on the buttons please

image

client/packages/common/src/ui/components/footer/ActionsFooter/ActionsFooter.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <Typography
+        fontSize="16px"
+        sx={{
+          fontWeight: 'bold',
+        }}
+      >
+        {selectedRowCount} {t('label.selected')}
+      </Typography>
+      {actions.map(({ label, icon, onClick, disabled }) => (
+        <Button
```
lache-melvin, line 47: Can we use our `FlatButton` component here? Should resolve the all caps situation 😅

lache-melvin (dismissed):
Just the little things from Mark's feedback, if we could use `FlatButton` for the buttons, and adjust the height so the two footers are consistent? 🙏 

Screenshot 2024-12-16 at 2 00 50 PM
Screenshot 2024-12-16 at 2 01 04 PM

client/packages/common/src/ui/components/footer/ActionsFooter.tsx
```diff
@@ -3,18 +3,18 @@ import {
   Stack,
   useTranslation,
   Typography,
-  Button,
+  FlatButton,
 } from '@openmsupply-client/common';
 
-interface ActionFooter {
+export interface ActionButtonFooter {
```
lache-melvin, line 9: I think `FooterAction`... or even just `Action` if it's not already taken as a name, might be better - this sounds like you are defining the props for the footer, rather than one of the actions in the footer?

lache-melvin (approved):
One more naming NIT - happy for you to merge when ready 🔥

mark-prins:
I'm too late - but also noticed this:

image

it would have been better if the 'process return of selected lines' button wasn't shown, or was disabled if the status of the shipment isn't 'shipped'.

PR #5811, opened 2024-12-15
5737 validate inputs contact form frontend

### LINKED ISSUE #5737: Validate feedback form inputs in the frontend
We should give the user fast feedback on why their contact form is invalid, rather than sending to the backend and suddenly getting an error.

Disable the `send` button while:
- email is empty
- message is empty
- email is not a valid email

After a slight delay from typing in the email, if not valid, show an error message, something like:

Screenshot 2024-12-11 at 10 08 02 AM

(See `error` and `helperText` props on `BasicTextInput`)

noel-yeldos:
Fixes #5737 

# 👩🏻‍💻 What does this PR do?
Validate frontent contact form inputs

Screenshot 2024-12-16 at 12 42 59 PM
Screenshot 2024-12-16 at 12 43 26 PM
Screenshot 2024-12-16 at 12 43 40 PM

# 🧪 Testing

- [ ] Navigate to ```Help``` page from app drawer
- [ ] When no inputs, check that send button is disabled
- [ ] When invalid email entered, check input field has red border and error helper text displaying, check send button still disabled
- [ ] Enter valid email but no message, check send button still disabled
- [ ] Enter valid email and valid message, check send button is enabled

# 📃 Documentation

  1.
  2.


client/packages/host/src/api/hooks/help/useFeedbackForm.ts
```diff
@@ -27,37 +31,59 @@ export function useFeedbackForm() {
     setDraft(defaultDraft);
   };
 
+  const checkEmailIsValid = (email: string) => {
+    const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+    return emailRegex.test(email);
+  };
```
lache-melvin, line 37: Actually, could we add this function to [RegexUtils](https://github.com/msupply-foundation/open-msupply/blob/5737-validate-inputs-contact-form-frontend/client/packages/common/src/utils/regex/RegexUtils.tsx), and add some unit tests for it? 🙏 

Will be nice to have this as reusable when we implement more features involving emails

client/packages/host/src/Help/FeedbackForm.tsx
```diff
@@ -36,8 +53,10 @@ export const FeedbackForm = () => {
         Input={
           <BasicTextInput
             value={draft.replyEmail}
-            onChange={e => updateDraft({ replyEmail: e.target.value })}
+            onChange={handleChange}
             fullWidth
+            helperText={emailError}
+            error={!isValidEmail && draft.replyEmail !== ''}
```
lache-melvin, line 59:
```suggestion
            error={!!emailError && draft.replyEmail !== ''}
```

and can then get rid of `isValidEmail` const.

Shouldn't need the extra isValidEmail check, you've already done that in the hook!

--- Actually, wondering if we should take off the `&& draft.replyEmail !== ''` part... I don't think it makes sense for the red border to sometimes go away, but the helper text to stay. Either not-empty check for both fields, or neither 🙏

lache-melvin (changes requested):
Looking good! Just a couple requests - making the email check reusable (and tested 🔥) and a bit of logic clean up 🙏

client/packages/host/src/api/hooks/help/useFeedbackForm.ts
noel-yeldos: Fixed in 039df99a1cd4844a2749c65e9760047125d7eea5 :)

client/packages/host/src/Help/FeedbackForm.tsx
noel-yeldos: Fixed in f0151a536631c474c6d0ff633a26b9b26073d9db 😃

client/packages/common/src/utils/regex/RegexUtils.test.ts
```diff
@@ -112,3 +112,27 @@ describe('formatTemplateString', () => {
...
+    expect(RegexUtils.checkEmailIsValid('test@email.')).toBeFalsy;
+  });
+  it('rejects empty string before @ symbol', () => {
+    expect(RegexUtils.checkEmailIsValid('@email.com')).toBeFalsy;
+  });
+  it('rejects blank email', () => {
+    expect(RegexUtils.checkEmailIsValid('')).toBeFalsy;
+  });
+  it('accepts correctly formatted email', () => {
+    expect(RegexUtils.checkEmailIsValid('test@email.com')).toBeTruthy;
```
lache-melvin, line 136: Awesome test cases! Just need to call the `toBeFalsy()`/`toBeTruthy()` functions, so that the assertions actually run 😁 

Also might be good to add one for different/funkier emails (e.g. test.user_1@msupply.foundation) - the long domain extension should be okay, but does fail on some email regexes so is a good one to test!

Maybe also things like multiple @@, mulitple ..... in the domain name, etc? Bit of a rabbithole - AI might suggest some good options to validate?

lache-melvin:
Looks great! Just fixing up those tests and then good to go 🔥

client/packages/common/src/utils/regex/RegexUtils.test.ts
noel-yeldos: Thank you! 😄  Added here b2e6feb422a5fcc33a677d463755e616eba84a59

lache-melvin (approved):
🔥

PR #5816, opened 2024-12-16
5690 create email queue table

### LINKED ISSUE #5690: Create email_queue table
```sql
CREATE TABLE email_queue (
    id TEXT PRIMARY KEY,
    to_address TEXT NOT NULL,
    subject TEXT NOT NULL,
    html_body TEXT NOT NULL,
    text_body TEXT NOT NULL,
    status {email_queue_status} NOT NULL,
    created_at TIMESTAMP NOT NULL,
    updated_at TIMESTAMP NOT NULL,
    sent_at TIMESTAMP,
    retries INTEGER NOT NULL DEFAULT 0,
    error TEXT
)
```

Where email_queue_status is an enum (`Queued | Sent | Errored | Failed`) (stored as text in SQLite, enum for PG)

Repository implementation should look the same as in [Notify](https://github.com/msupply-foundation/notify/blob/main/backend/repository/src/db_diesel/email_queue_row.rs)

noel-yeldos:
Fixes #5690 

# 👩🏻‍💻 What does this PR do?
Create email queue table and add migration

Changes needed as part of GEDSI project to add contact form which sends requests to support

Sidequest to fix https://github.com/msupply-foundation/open-msupply/pull/5759#discussion_r1886107764 here 6efc405aeefef5c7f5d58e9fdb7d65ccf6169923

# 🧪 Testing

- [ ] cargo run
- [ ] check db table is created

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/repository/src/db_diesel/email_queue_row.rs
+++ b/server/repository/src/db_diesel/email_queue_row.rs
@@ -25,8 +25,8 @@ table! {
 pub enum EmailQueueStatus {
     Queued,
     Sent,
-    Errored,
-    Failed,
+    Errored, // Errored will be re-tried
+    Failed,  // Failed will NOT be re-tried
 }
 
 impl Default for EmailQueueStatus {
--- a/server/repository/src/db_diesel/email_queue_row.rs
+++ b/server/repository/src/db_diesel/email_queue_row.rs
@@ -38,7 +38,7 @@ impl Default for EmailQueueStatus {
 #[derive(
     Clone, Queryable, Insertable, Identifiable, Debug, PartialEq, Eq, AsChangeset, Default,
 )]
-#[table_name = "email_queue"]
+#[diesel(table_name = email_queue)]
 pub struct EmailQueueRow {
     pub id: String,
     pub to_address: String,
--- a/server/repository/src/db_diesel/email_queue_row.rs
+++ b/server/repository/src/db_diesel/email_queue_row.rs
@@ -1,0 +1,111 @@
+use super::email_queue_row::email_queue::dsl::*;
+use crate::{RepositoryError, StorageConnection, Upsert};
+use chrono::NaiveDateTime;
+use diesel::prelude::*;
+use diesel_derive_enum::DbEnum;
+
+table! {
+    email_queue (id) {
+        id -> Text,
+        to_address -> Text,
+        subject -> Text,
+        html_body -> Text,
+        text_body -> Text,
+        status -> crate::db_diesel::email_queue_row::EmailQueueStatusMapping,
+        created_at -> Timestamp,
+        updated_at -> Timestamp,
+        sent_at -> Nullable<Timestamp>,
+        retries -> Integer,
+        error -> Nullable<Text>,
+    }
+}
+
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Hash)]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+pub enum EmailQueueStatus {
+    Queued,
+    Sent,
+    Errored,
+    Failed,
+}
+
+impl Default for EmailQueueStatus {
+    fn default() -> Self {
+        EmailQueueStatus::Queued
+    }
+}
+
+#[derive(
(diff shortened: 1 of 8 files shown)


server/repository/src/db_diesel/email_queue_row.rs
```diff
@@ -0,0 +1,111 @@
...
+        error -> Nullable<Text>,
+    }
+}
+
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Hash)]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+pub enum EmailQueueStatus {
+    Queued,
+    Sent,
+    Errored,
```
jmbrunskill, line 28:
```suggestion
    Errored, // Errored will be re-tried
    Failed,  // Failed will not be re-tried
```

I had this clarification in the notify version of this table, might be useful to copy here?

server/service/src/invoice/inbound_shipment/add_from_master_list.rs
```diff
@@ -38,7 +38,7 @@ pub fn add_from_master_list(
         .connection
         .transaction_sync(|connection| {
             let invoice_row = validate(connection, &ctx.store_id, &input)?;
-            let new_invoice_line_rows = generate(connection, ctx, invoice_row, &input)?;
+            let new_invoice_line_rows = generate(ctx, invoice_row, &input)?;
```
jmbrunskill, line 41: It's a good idea to add a github comment for a side quest like this, luckily I know the context here, but normally it's good to point it out otherwise the reviewer would be confused why this code has changed.
In general we try to keep side quests to a minimum, unless it's closely related to the main change.

server/repository/src/migrations/v2_05_00/add_email_queue_table.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        if cfg!(feature = "postgres") {
+            sql!(
+                connection,
+                r#"
+                CREATE TYPE email_queue_status_enum AS ENUM (
+                    'QUEUED',
+                    'SENT',
+                    'ERRORED',
+                    'FAILED',
```
jmbrunskill, line 19:
```suggestion
                    'FAILED'
```
I don't think SQL likes a trailing `comma` :)

jmbrunskill (changes requested):
Looks good, but I think you need to fix the postgres tests.

Run with `cargo test --features=postgres`

It's failing on the `CREATE TYPE` right not, but after that passed you'll probably get a error with refresh dates.

You'll probalby need to add
`created_datetime`, `updated_datetime`, `sent_at` etc to 
`get_exclude_timestamp_fields()`

One thing you might want to add is a `retry_at` field similar to the one in notify here:

https://github.com/msupply-foundation/notify/blob/e97525854006a6e247cf34b9aef5222c37891cf3/backend/repository/src/db_diesel/notification_event_row.rs#L20

This will allow us to wait longer between each email send retry message...

Here's how we/ethan did it in Notify...
https://github.com/msupply-foundation/notify/commit/1716d7a413962b438d0f965e05642da33abc7a2d

server/service/src/invoice/inbound_shipment/add_from_master_list.rs
noel-yeldos: Ah yes good to know, I've added a note in the original comment

server/repository/src/db_diesel/email_queue_row.rs
noel-yeldos: Add here 56690bedfddd39b76e00ca771067e77ae9d102a6

server/repository/src/migrations/v2_05_00/add_email_queue_table.rs
noel-yeldos: Added here 56690bedfddd39b76e00ca771067e77ae9d102a6 :)

noel-yeldos:
Thanks @jmbrunskill! I have made a comment here: https://github.com/msupply-foundation/open-msupply/issues/5821#issuecomment-2547056619 to address adding of ```retry_``` column and exponential back-off.

jmbrunskill (approved):
Looks great, let's get it in!

PR #5822, opened 2024-12-16
5650(2) Dispensing Line Edit UI additional changes

### LINKED ISSUE #5650: Dispensing: Redesigned item add/edit page view

CarlosNZ:
Fixes #5650 (Part 2)

# 👩🏻‍💻 What does this PR do?

Changes to the way this works, based on UI designs and IRL discussions.

Main functionality is to manage draft line state in the top level component(LineEditView) and save all changes at once when clicking Save.

Also, Prev/Next nav bar is separate to main Footer buttons

Screenshot 2024-12-17 at 11 25 42 AM

Still a work in progress. Needs UI updates for styling etc., and to implement inner components of the line edit view.

Also, not yet handling errors properly. I've left some of the existing code for this in, but some of it is commented out for now. I think we'll need to address this in its own issue after we figure out exactly what the UI behaviour should be.

Also, one of the fields in the Line Edit form is still editable on Verified prescriptions, as the UI component didn't have a simple "disabled" flag, but we will be re-doing this form anyway, so ignore for now.

# 🧪 Testing

- [ ] From prescription "Detail" view (status: "Picked" or "New"), click an item to go to "LineEditView"
- [ ] Clicking in left-hand bar takes you to the correct item in the list
- [ ] Can select "New" item to add a new item
- [ ] "Save" button should remain disabled until some changes are made (dirty state)
- [ ] No "you haven't saved" alert should appear when switching between items, but *should* appear when leaving the page or refreshing
- [ ] Prev/Next navigation should behave as expected
- [ ] After "Saving" a "New Item", you should be directed to that item's newly created entry in the Item list
- [ ] For "Verified" prescriptions:
  - [ ] You should still be able to access this LineEditView from the Detail page (this is new)
  - [ ] "New Item" should not appear
  - [ ] Can't edit anything (bar the exception mentioned above)


client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -1,31 +1,54 @@
...
   const { setCustomBreadcrumbs } = useBreadcrumbs();
+  const navigate = useNavigate();
 
   const {
     query: { data, loading: isLoading },
   } = usePrescription();
 
+  const {
+    save: { saveLines, isSavingLines },
+    // delete: { deleteLines },
```
CarlosNZ, line 36: To-DO: add Delete button

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -1,31 +1,54 @@
...
+  } = usePrescriptionLines(data?.id);
+
+  const newItemId = useRef<string>();
+
+  const { isDirty, setIsDirty } = useDirtyCheck();
+
   const lines =
-    data?.lines.nodes.sort((a, b) => a.item.name.localeCompare(b.item.name)) ??
-    [];
+    data?.lines.nodes.sort((a, b) => a.id.localeCompare(b.id)) ?? [];
```
CarlosNZ, line 44: We need the order to be in the order that items were added (so the new item gets added to the bottom of the list). Fortunately, with v7 UUIDs, sorting by ID achieves this, as they have a built-in timestamp. I know relying on the ID for timestamp is not 100% reliable, but there's currently no other indicator of the actual added order of items to an invoice (is there?)

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -1,31 +1,54 @@
...
+
+  const invoiceId = data?.id ?? '';
+  const status = data?.status;
+
+  const [allDraftLines, setAllDraftLines] = useState<
+    Record<string, DraftStockOutLine[]>
+  >({});
```
CarlosNZ, line 51: So this object keeps the draft changes for all the items (indexed by Item ID), and we just pass the correct one down to the individual line edit forms (along with an "update" method)

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -51,9 +74,78 @@ export const PrescriptionLineEditView = () => {
...
+        ? {
+            id: invoiceId,
+            status: InvoiceNodeStatus.Picked,
+          }
+        : undefined;
+
+    await saveLines({
+      draftPrescriptionLines: flattenedLines,
+      patch,
+    });
```
CarlosNZ, line 126: This stuff directly copied from original child component.

client/packages/invoices/src/Prescriptions/LineEditView/NavBar.tsx
```diff
@@ -0,0 +1,67 @@
+/**
+ * TO-DO: Make this a generic re-useable component (in "common"), and update
+ * styling to match new designs
+ */
+
+import {
```
CarlosNZ, line 6: This component just bare-bones for now. The UI designs have new colour scheme, etc, but don't want to start down that path just yet as that requires a more global approach.

client/packages/invoices/src/Prescriptions/LineEditView/NavBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+        label={t('button.previous')}
+        Icon={<ArrowLeftIcon />}
+        disabled={!hasPrevious}
+        onClick={() => {
+          setItem(items[currentIndex - 1] ?? '');
+        }}
+      />
+      <Typography>
+        {/* Only display when not "NEW" item */}
+        {currentCount <= totalCount && `${currentCount}/${totalCount}`}
```
CarlosNZ, line 55: I figured it didn't really make sense to include the "New" item in the total count since it isn't actually part of the prescription yet. But when we're on the "New" page, the current count would be higher than the total, so I've chosen to just hide this completely in that case.

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEdit.tsx
```diff
@@ -138,27 +110,27 @@ export const PrescriptionLineEdit: React.FC<PrescriptionLineEditModalProps> = ({
...
+  //     // setIsDirty(false);
+  //     if (!!placeholder) {
+  //       const infoSnack = info(t('message.placeholder-line'));
+  //       infoSnack();
+  //     }
+  //     return onSaved();
+  //   } catch (e) {
+  //     // console.error(e);
+  //   }
+  // };
```
CarlosNZ, line 133: This is some of the logic needed for UI error indications, so just leaving here commented out for now so we don't forget about it or lose it.

client/packages/programs/src/JsonForms/components/Prescription/Prescription.tsx
```diff
@@ -72,8 +72,16 @@ const UIComponent = (props: ControlProps) => {
     useState<ItemStockOnHandFragment | null>(
       formActions.getState(`${path}_item`) ?? null
     );
-  const { draftStockOutLines, setDraftStockOutLines } =
-    useDraftPrescriptionLines(selectedItem);
+
+  const [draftPrescriptionLines, setDraftPrescriptionLines] = useState<
+    DraftStockOutLine[]
+  >([]);
```
CarlosNZ, line 78: Because of the changes to state management, this component now needs to hold its own `draftLines` state, and it gets manipulated by the `useDraftPrescriptionLines` hook -- functionality should be unchanged.

client/packages/invoices/src/Prescriptions/LineEditView/Footer.tsx
```diff
@@ -66,35 +37,17 @@ export const Footer: FC<FooterProps> = ({
...
-                  handleSave(() =>
-                    setNavigateTo(
-                      RouteBuilder.create(AppRoute.Dispensary)
-                        .addPart(AppRoute.Prescription)
-                        .addPart(String(invoiceNumber))
-                        .addPart(String(previous))
-                        .build()
-                    )
-                  );
+                  // TO-DO
```
jmbrunskill, line 43: Maybe the same as close for now? Feels like this button should work?

jmbrunskill: I guess it needs to bypass the ConfirmOnLeaving thing too? is it just a `resetDraft()` call?

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
jmbrunskill: I think this is ok, although I did kind of like the ordering by name... For a new item I agree it makes sense what you've done.

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -51,9 +74,78 @@ export const PrescriptionLineEditView = () => {
...
+  useConfirmOnLeaving(
+    isDirty,
+    // Need a custom checking method here, as we don't want to warn user when
+    // switching to a different item within this page
+    (current, next) => {
+      if (!isDirty) return false;
+
+      const currentPathParts = current.pathname.split('/');
+      const nextPathParts = next.pathname.split('/');
+      // Compare URLS, but don't include the last part, which is the ItemID
```
jmbrunskill, line 86: Good comment

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -51,9 +74,78 @@ export const PrescriptionLineEditView = () => {
...
+
+  const onSave = async () => {
+    if (!isDirty) return;
+
+    const flattenedLines = Object.values(allDraftLines).flat();
+
+    // needed since placeholders aren't being created for prescriptions yet, but
+    // still adding to array
+    const isOnHold = flattenedLines.some(
+      ({ stockLine, location }) => stockLine?.onHold || location?.onHold
```
jmbrunskill, line 110: onHold will just not be queried in future I think, when we do the new stockline selector it won't be shown...

jmbrunskill (commented):
Nice, work a lot to think about, and work through here!

Feel like it's a bit hard to review all the code, probably didn't give myself enough time sorry not quite ready to sign off yet.

A few things concern me though.

1. Lots of commented out code (especially the error handling) can you create a follow up issue to come back to it?
2. It still feels a bit buggy...

- I was able to get a Line Already Exists error (not 100% of repo yet)
- Cancel button doesn't work
- Adding an item requires a save button to be clicked but you don't get a confirmation if you change page (you changes are just lost?)
- If entering a new item and type in a number of items to issue, that state seems to be persisted to all the items? Probably needs to re-calculate from stocklines on page change?
- Sometimes if editing a stock line quantity directly isDirty doesn't seem to be set? So the save button appears to be disabled.

I'll try to take a closer look tomorrow morning and see if I can figure out some repro steps.
Definitely good learning that the this kind of change takes a tonne of time, and is hard to get back to the previous level of reliable functionality...

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
jmbrunskill: Nice solution!

CarlosNZ:
> Cancel button doesn't work

Oops, just forgot to come back to that one. Done now: e0ee1a90fa

CarlosNZ:
> * Adding an item requires a save button to be clicked but you don't get a confirmation if you change page (you changes are just lost?)

Yeah, I've added a Save with confirmation to that action now: 12f272c2e3

Not totally convinced it's the best approach, but works for now. Also, I realise it's possible to get into an `isDirty` state with a new item even though it's not yet valid (I'll show you how IRL), so will need to address this at some point too (but at least it doesn't actually save invalid data)

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
CarlosNZ: The default ordering for the Detail view is in order added, so makes sense to be the same in here.

jmbrunskill (approved):
Thanks for the Update Carl,

As discussed lets merge this in to the feature branch, then we can asset where we're at tomorrow to see if we have any blockers for merging into develop.

PR #5401, opened 2024-11-14
5390 correctly handle gs1 2d data matrix format

jmbrunskill:
Fixes #5390

# 👩🏻‍💻 What does this PR do?

I tried to make minimal changes for this, but it should allow us to scan GS1 formatted barcodes and create assets from them. 

## 💌 Any notes for the reviewer?

Maybe/probably should remove the parsing of human readable string? Figured I'd leave it there for now as it might come in handy, and I didn't want to have to re-create data for all my tests cases and aiming for minimal changes here :)

Thoughts on this?

# 🧪 Testing

- [ ] Using either Android (with camera), or Desktop (with 2d Barcode Scanner)
- [ ] Open the cold chain equipment section, and click the `Scan` button (this doesn't show on web)
- [ ] Scan one of the barcodes below
- [ ] You should be prompted to create a new asset
- [ ] If you accept, it should show a new asset with the serial number, asset type, and Warenty date populated
- [ ] The Installation date should default to today
- [ ] Scanning the same barcode again, should navigate you to the right asset (once it has been created)

> NOTE, scanning these 2d Matrixes in github `dark mode` or with a black background doesn't work well...
> NOTE: in general scanning 2d Matrixes seems much less reliable than QR Codes, need to have camera straight on for example

Scan a GS1 formatted Data Matrix here (Note: or create your own for testing)

PQS Asset Type: E003/002
Serial Number: S12345678
https://bwipjs-api.metafloor.com/?bcid=gs1datamatrix&text=(01)09506000134352(11)241007(21)S12345678(241)E003/002(3121)820000(3131)670000(3111)630000(90)001(91)241007-310101
image

PQS Asset Type: E004/055
Serial Number: S1
https://bwipjs-api.metafloor.com/?bcid=gs1datamatrix&text=(01)09506000134352(11)241007(21)S1(241)E004/055(3121)820000(3131)670000(3111)630000(90)001(91)241007-310101

image

PQS Asset Type: E004/055
Serial Number S2
https://bwipjs-api.metafloor.com/?bcid=gs1datamatrix&text=(01)09506000134352(11)241007(21)S1(241)E004/055(3121)820000(3131)670000(3111)630000(90)001(91)241007-310101

image

# 📃 Documentation

- [X] **Part of an epic**: documentation will be completed for the feature as a whole
- [X] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

PR #5839, opened 2024-12-17
add permission check to scan button

### LINKED ISSUE #5223: App crashes when trying to scan without permission
## What went wrong? 😲
Found while testing [THIS](https://github.com/msupply-foundation/open-msupply/issues/1912)

If there's no permission and we try to scan QR from the equipment tab in Android, the app crashes.
Screenshot 2024-10-24 at 4 37 27 pm

https://github.com/user-attachments/assets/02444d74-f731-40cc-af7b-4cfe7ad283e9

## How to Reproduce 🔨

1. Go to user permission and remove permission from the Permissions3> Assets
2. Now sync with the tablet and try to add equipment from scanner button
3. See error

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:  `V2.3.01-RC1`

- Platform: Android (V2.3.01-RC1)

- Database type:

aimee-mcneil-melville:
Fixes #5223 

Has a sidequest PR #5855 

# 👩🏻‍💻 What does this PR do?

Adds permission checks to the scan button in ```Cold Chain -> Equipment``` and notifies the user if they are not permitted

Screenshot 2024-12-17 at 5 26 38 PM

The original issue referenced the app crashing in v2.3.1 if the user did not have permission to scan, I could not replicate this.

I have added checks to notify the user if they are trying to access a scan function they are not permitted for, which also prevents the scanning going any further.

There are three pathways:
- if the user has no 'view' permission for assets, they are notified of this immediately upon clicking the scan button
- if the user has 'view' permissions but not 'add/edit', they can use the scan function to find assets by gs1 or Id. If the asset already exists, it will show. If it does not exist, the user will notified they do not have permission to create a new asset
- if the user has 'add/edit' permissions, they will be able to scan an asset, view it if it already exists, and create a new one if it doesn't

The user will benefit from feedback on why they cannot access the scan function/part of the scan function. This also prevents any further code being actioned if the permissions are not valid

~Is there any use case for a user with permission to add/edit but not view? There is no workflow created for this. As it is, this permission combo will block any access to scanning~
Checked with James, this should be OK!

Noticed permissions don't always persist after refresh (see comment) will raise this separately as it may be across more areas.

Not directly linked to this PR but I'm going to create a new branch for adding permission checks to prevent updating the status of an asset, while I'm in the code area and setup for the android testing

# 🧪 Testing

Can also complete testing for #5855 at the same time

#### Initial setup
- [ ] Setup on Android emulator as a site
- [ ] Setup with label printer
- [ ] Test label (GS1 Data Matrix) - use one of the examples in #5401 

#### With all permissions on
- [ ] Turn user permissions on for ```View Asset``` and ```Add/Edit Asset``` in 4D  -> sync
- [ ] Go to Cold Chain -> Equipment
- [ ] Add a new asset and print a label for it, or print a label for an existing Asset
- [ ] Scan the asset - it will show the detail view for the asset!

#### With view only permissions on
- [ ] Have user permission on for ```View Asset``` and off for ```Add/Edit Asset``` in 4D -> sync -> go to equipment page
- [ ] Scan the label you printed - you should see the detail view
- [ ] Scan a test code in the link above - you should see a notification "You do not have permission to add an asset" 

#### With no view or add permissions on
- [ ] Turn user permissions off for ```View Asset``` and ```Add/Edit Asset``` in 4D -> sync -> go to equipment page
- [ ] Click the scan button -> see a notification "You do not have permission to view assets" and scan doesn't start

# 📃 Documentation

  1.
  2.


aimee-mcneil-melville:
I found whilst doing the side branch that these permission checks only work until the page is refreshed - unsure if that is replicatable specifically for scanning on tablet, however is definitely an issue on a computer for 'New Asset' and the change status button

aimee-mcneil-melville:
> I found whilst doing the side branch that these permission checks only work until the page is refreshed - unsure if that is replicatable specifically for scanning on tablet, however is definitely an issue on a computer for 'New Asset' and the change status button

This is intermittent. I could not replicate this today - however Noel tested and could replicate this bug. It appears that the 'permissions' graphql does not get recalled on refresh so may be inconsistent caching.

I will raise it as a separate issue as this may affect other areas of the code and will need further investigation

client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
```diff
@@ -102,26 +107,29 @@ export const AddFromScannerButtonComponent = () => {
...
-      stopScan();
-    } else {
-      try {
-        await startScanning(handleScanResult);
-      } catch (e) {
-        error(t('error.unable-to-start-scanning', { error: e }))();
+    if (userHasPermission(permission)) {
+      if (!isConnected) {
+        show(e);
+        return;
```
lache-melvin, line 120: NIT, but I think it would be a nice clean up - our typical convention when we need to check something before continuing is to do an early return, like you can see is already here with the `isConnected` check.

It's a little easier to read than having to track all way to the bottom of the if block to see what would happen without the permission.

So could be:

```ts
if (!userHasPermission(permission)) {
  info(t('error.no-asset-view-permission'))();
  return;
}
... rest of fn
```

lache-melvin (approved):
Code changes look great Aimee! I don't have a tablet here to test, I wonder if you could ask someone in the office to take it for a quick spin? And merge when ready 🔥

client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
aimee-mcneil-melville: Awesome, have done this now!

jmbrunskill:
Looks great @aimee-mcneil-melville I've found a bug though, but it's not directly related to your change.
Created a PR here https://github.com/msupply-foundation/open-msupply/pull/5875, although I'm not sure if it's the best solution...

jmbrunskill (approved):
Ship it!

lache-melvin:
Thanks for gorgeous testing steps too 🔥

PR #5840, opened 2024-12-17
5805 support option on contact form

### LINKED ISSUE #5805: Add `support` request option to contact form
Screenshot 2024-12-16 at 8 55 20 AM

- [x] Update backend graphql and service layers to include a new field in their input structs: contact_type (enum type)
- [x] Add radio buttons for contact form type in frontend
- [x] Include the selected contact form type in the draft state, and send to backend graphql request
- [x] Ensure send button is disabled until one of the contact types is selected
- [ ] Update any frontend code references that still say "feedbackform" rather than "contact form

zachariah-at-msupply:
Fixes #5805

# 👩🏻‍💻 What does this PR do?

Adds the ability to input a reason for contacting us on the contact form.

This helps us separate feedback from support requests (assuming it's always used properly 😏 ).

Screenshot 2024-12-17 at 16 56 22

Hello, how's the day?

Also, should these be other issues, or included in this branch, or not issues at all?
1. The 'Reason for contacting us' label is squished over two lines.
2. The original issue asked for radio buttons (which would mean the squished label doesn't exist).
3. Radio buttons are not currently in our component library in the code, so we would need to make a radio button component.
4. The contact us form will make the whole app crash if it gets an invalid input because the form validation is done on the back end, and we haven't linked those back end errors to fornt-end behaviour.

Cheers!

The changes affected the client's host package, and the server's service and graphql layers.
I also threw in a quick little bit of code to grey out the send button if the email field is blank - I'm not sure this is necessary, especially if we address point 4 above, but it seemed like a convenient time to include that part of the front-end validation, as it only took a few extra characters to do.

# 🧪 Testing

- [ ] go to the help page
- [ ] check that the send button is disabled 
- [ ] check that the send button is only enabled after you enter both a reason for contacting us and something in the email field
- [ ] check the text in the dropdown is suitable

# 📃 Documentation

- [ x ] **Part of an epic**: documentation will be completed for the feature as a whole
  1.
  2.


client/packages/host/src/api/hooks/help/useFeedbackForm.ts
```diff
@@ -38,13 +38,22 @@ export function useFeedbackForm() {
...
-  const mutationFn = async ({replyEmail, body}: ContactFormInput) => { 
+  const mutationFn = async ({contactFormType, replyEmail, body}: ContactFormInput) => { 
+    let contactType;
+    if (contactFormType == ContactFormNodeType.Feedback) {
+      contactType = ContactFormNodeType.Feedback;
+    }
+    if (contactFormType == ContactFormNodeType.Support) {
+      contactType = ContactFormNodeType.Support;
+    }
+    if (contactType == undefined) {throw Error("contactType is undefined")} // this is ok as the front end will disable the send button if this is the case.
```
lache-melvin, line 49: I think this `if` block is a bit redundant, and might be easy to miss if another option got added - nicer just to use the enum values throughout. Assuming this was done as a result of the `onChange` of the Select component handling a string rather than the enum value?

I think this is one of the cases where the `as` keyword is ok... put a suggestion and comment here, let me know if you have questions:

https://github.com/msupply-foundation/open-msupply/commit/20a3472c951500e3e36f2e0237cbfae111c6f179#r150471171

Also looks like there are a number of white space differences - we might need to take a look at your prettier configuration, I think it is not formatting your files on save 🙏

lache-melvin (changes requested):
Thanks Zachariah, looking good! 🔥 Just requesting the mapping from handling contact type as string, to just using the enum values 🙏 

Your other questions:

> The 'Reason for contacting us' label is squished over two lines.

Wondering whether `Request type` would be an ok label? Probably not a huge deal as like you say, should move away from it anyway


> Radio buttons are not currently in our component library in the code, so we would need to make a radio button component.

Yes could you please create a new `enhancement` issue for this? Thanks

> The contact us form will make the whole app crash if it gets an invalid input because the form validation is done on the back end, and we haven't linked those back end errors to fornt-end behaviour.

Frontend validation is going in with Noel's PR #5811 - but once we get everything merged we should do a thorough test - possibly still missing an error catch at this stage 🙏 

> I also threw in a quick little bit of code to grey out the send button if the email field is blank

Hehe might cause a small conflict with the validation PR but helpful to see in this PR, I think no worries, we'll resolve at merge time 🚀

client/packages/host/src/api/hooks/help/useFeedbackForm.ts
zachariah-at-msupply: Cool! I had a look at my formatter settings and I think they're correct now... we'll see in the future I suppose.
I'll deal with that if block now.

zachariah-at-msupply (commented):
THanks for your help Laché! I merged the suggested changes in to this branch. I'll request another review now 👍

lache-melvin (approved):
🔥 Thanks Zachariah! Looks like there are some conflicts with develop now to resolve before merging, I think they're with John & Noel's changes so if any of the conflicts are finnicky, you should be able to jam with them to get context 🚀

zachariah-at-msupply:
I merged in all the changes, and I've requested a re-review to make sure everything's ship-shape ;)

lache-melvin (approved):
Ship it!

PR #5862, opened 2024-12-19
5801 dispensing diagnosis selection

### LINKED ISSUE #5801: Dispensing: Diagnosis Selection
Display the select list for diagnosis
The options are taken from the diagnosis table, with just the ICD_Description field shown to the user
Only diagnosis options where the VALID_TILL is greater than today, or not set are shown
An option of ‘None’ is available for selection
Saved as [transact]diagnosis_ID in mSupply

image

Acceptance Criteria
- [x] Add `diagnosis_id` to invoice table
- [x] Sync selected `diagnosis` with mSupply
- [x] Diagnosis shows in side panel under Patient Details
- [x] Use can set the diagnosis to an option that's still valid from the list
- [x] Diagnosis can be set to none/nothing if desired
- [x] Only Description is shown when selecting a diganosis

> No need to do new create prescription modal in this issue...

jmbrunskill:
Fixes #5801

# 👩🏻‍💻 What does this PR do?

Adds the ability to select a diagnosis on a prescription.

Add some diagnoses in mSupply

image

Open a prescription in OMS, and go to the `More` section

image

See the selection box under Patient Details

image

Global pref for diagnosis in prescription isn't used yet. It's just always available in OMS if people want to use it.

# 🧪 Testing

- [ ] Make sure prescriptions work ok with no diagnoses setup
- [ ] Setup diagnoses on mSupply
- [ ] Make sure you can select and clear diagnosis on a prescription
- [ ] Check that diagnosis selection is disabled after prescription is verified
- [ ] Check that no invalid diagnoses show up, except in old prescriptions (which should still show)

# 📃 Documentation

- [X] **These areas should be updated or checked**: Prescriptions


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -443,6 +443,9 @@
   "heading.details": "Details",
   "heading.devices": "Devices",
   "heading.dispensary-details": "Dispensary Details",
+  "heading.pricing": "Pricing",
```
jmbrunskill, line 446: Pricing is in the new designs, and section was internally labeled as pricing, so I added the translation here.

client/packages/invoices/src/Prescriptions/api/hooks/useDiagnosisOptions.ts
```diff
@@ -0,0 +1,32 @@
+import { DiagnosisNode, useQuery } from '@openmsupply-client/common';
+import { usePrescriptionGraphQL } from '../usePrescriptionGraphQL';
+import { DIAGNOSIS, LIST } from './keys';
+
+export const useDiagnosisOptions = () => {
```
jmbrunskill, line 5: Went with a specific `Options` hook, less generic but fits the use case. Interested if others like this idea...

client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
```diff
@@ -196,10 +197,12 @@ const useUpdate = (id: string) => {
   const { prescriptionApi, storeId, queryClient } = usePrescriptionGraphQL();
 
   const mutationFn = async (patch: RecordPatch<PrescriptionRowFragment>) => {
+    console.log('patch', patch);
     const input: UpdatePrescriptionInput = {
       ...patch,
       id,
       status: mapStatus(patch),
+      diagnosisId: setNullableInput('diagnosisId', patch),
```
jmbrunskill, line 205: This is in two places, I wonder if we need to generalise this some how. Maybe even the draft patch state thing can do something special with Nullable Updates?

server/repository/src/migrations/v2_05_00/diagnosis_add_to_invoice.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+            ALTER TABLE invoice ADD COLUMN diagnosis_id TEXT REFERENCES diagnosis(id);
+        "#
+        )?;
+
+        // Retranslate all invoices on the next sync, to make sure we capture the diagnosis_id if populated.
```
jmbrunskill, line 18: This might be overkill and could be slow on big sites. Wasn't sure if I should try to optimise by querying for Prescriptions, or do we jsut not care?
@andreievg  thoughts?

server/service/src/auth.rs
```diff
@@ -580,6 +581,10 @@ fn all_permissions() -> HashMap<Resource, PermissionDSL> {
         Resource::QueryVaccineCourse,
         PermissionDSL::NoPermissionRequired,
     );
+    map.insert(
+        Resource::NoPermissionRequired,
```
jmbrunskill, line 585: I think this is nice and clear and can easily change to something more specific if needed, would allow more easy re-use for similar situations?

server/service/src/invoice/inventory_adjustment/add_new_stock_line/generate.rs
```diff
@@ -78,6 +78,7 @@ pub fn generate(
         requisition_id: None,
         clinician_link_id: None,
         backdated_datetime: None,
+        diagnosis_id: None,
```
jmbrunskill, line 81: There's a tonne of these, manually added to each one, as I think in general we probably want to think about each use case when adding something to invoice, however `..Default:default()` seems rather tempting at this stage?
@andreievg

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PatientDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { diagnosis } = data ?? {};
+
+  const {
+    query: { data: diagnosisOptions },
+  } = useDiagnosisOptions();
+
+  return (
+    <DetailPanelSection title={t('heading.patient-details')}>
+      <Grid container gap={0.5}>
+        <>
```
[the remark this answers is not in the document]
CarlosNZ: Actually, I don't even thing we need the `<Grid>`. It just adds another level of spurious `div` to the HTML. (which has been the cause of numerous layout bugs that I've fixed before )

(Looks like you just copied it from one of the other Panel Rows -- I guess we can't fix them all right now, but we can be the change we want to see!)

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PatientDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+              }}
+              onChange={(_e, selected) => {
+                if (selected) {
+                  update({ diagnosisId: selected.value });
+                } else {
+                  // Updated value needs to be null for nullable input to work correctly
+                  update({ diagnosisId: null });
+                }
+              }}
+              disabled={isDisabled}
```
CarlosNZ, line 50: You get a bit of an annoying flicker when selecting/changing the option. This is due to the value not actually being updated until the round-trip to/from the server is complete.

We could use the `useBufferState` hook, but that uses the ill-advised `useEffect` to update state, so we should really only use it when the external value can possibly be different to the locally selected value. So in this case, I'd just use this pattern, where keep a local value, but fall back to the external value when local state is not yet defined (i.e when first loading).

Here's a suggestion branch demonstrating: https://github.com/msupply-foundation/open-msupply/commit/e6e1143a152b6958626866b05a0cb1ad0c7d4be3

If you throttle your network speed significantly and compare the difference, it's quite noticeable.

CarlosNZ (changes requested):
Looks good, thanks. Only a few minor comments/suggestions.

server/repository/src/migrations/v2_05_00/diagnosis_add_to_invoice.rs
andreievg: Great point, I think it's fine not to re-integrate (would only be a problem when migrating mSup to omSup before this change, which i think is super unlikely)

jmbrunskill: Don't we have some migrated sites already? At least from Mobile?

client/packages/invoices/src/Prescriptions/api/hooks/useDiagnosisOptions.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { prescriptionApi } = usePrescriptionGraphQL();
+
+  const queryKey = [DIAGNOSIS, LIST, 'all'];
```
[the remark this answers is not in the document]
jmbrunskill: Makes sense, although I think this is fine too and feels more consistent to me...
Now that Mark's moved sync to a modal we might need to thinking about this invalidation on sync case...

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PatientDetails.tsx
jmbrunskill: [d2c6249](https://github.com/msupply-foundation/open-msupply/pull/5862/commits/d2c62498eb4969bd8164bdb4504c52a8e6200e23)

server/service/src/diagnosis.rs
```diff
@@ -0,0 +1,15 @@
...
+pub fn get_all_active_diagnoses(
+    connection_manager: &StorageConnectionManager,
+) -> Result<Vec<Diagnosis>, RepositoryError> {
+    let connection = connection_manager.connection()?;
+    let repository = DiagnosisRepository::new(&connection);
+
+    let rows = repository.query_by_filter(DiagnosisFilter::new().is_active(true))?;
+
+    Ok(rows)
+}
```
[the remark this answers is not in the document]
jmbrunskill: [9f005a7](https://github.com/msupply-foundation/open-msupply/pull/5862/commits/9f005a7cc5900cc72654af794ee43e324551a2ac)

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PatientDetails.tsx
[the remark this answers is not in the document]
jmbrunskill: [cab8dc1](https://github.com/msupply-foundation/open-msupply/pull/5862/commits/cab8dc1df894709cec4cf31d0cbd6fca44d4fa71)

client/packages/invoices/src/Prescriptions/api/operations.graphql
```diff
@@ -391,3 +398,15 @@ fragment historicalStockLine on StockLineNode {
...
 }
+
+fragment diagnosis on DiagnosisNode {
+  id
+  code
+  description
+}
+
+query activeDiagnoses {
+  allActiveDiagnoses {
```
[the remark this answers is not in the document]
jmbrunskill: [8fafe30](https://github.com/msupply-foundation/open-msupply/pull/5862/commits/8fafe3064a9c032e68c59f7d1817260231e9ee05)

CarlosNZ (approved):
Nice, thanks for the changes 👍

PR #5864, opened 2024-12-19
modify the tooltips of table column headers

### LINKED ISSUE #5235: Tooltip repetition in column headers
## What went wrong? 😲

When hovering over SOH, AMC and MOS, I find the tooltip repeats the column title which is already clear enough. It should elaborate on what the acronym stands for, without repeating what is already on the screen. It looks even worse when translating those strings, where you can't turn everything into an acronym so a lot of information ends up being repeated twice.

https://github.com/user-attachments/assets/b4a332e5-7158-4385-a510-bed160874c75

## Expected behaviour 🤔

No repetition, e.g. let's remove the column title from the tooltip

## How to Reproduce 🔨

1. Go to Catalogue -> Items
2. Hover over SOH, AMC or MOS -> see repetition in the tooltip

## Your environment 🌱

- Open mSupply Version: V2.3.1-RC1
- Legacy mSupply Central Server Version: V7.18.06

- Platform: Browser (Google)

- Database type: SQLite3

zachariah-at-msupply:
Fixes #5235

# 👩🏻‍💻 What does this PR do?

I modified the tooltips for header cells, so that now, if a description of the column has been shown, the label is not shown - if the description has already been shown, showing the label will not add anything. 
That prevents things being repetitive

image

I changed the tooltips so that now it says 'sort by [column label in lower case]' - before it did not transform it to lower case. Is this ok?

Is it worth creating an issue for this - at the moment, you get tooltips for every single entry in a table, which add no information, and stop you interacting with what's covered by them. Here are some examples:

This one stops you clicking on two lines:

Screenshot 2024-12-19 at 16 40 32

These tooltips don't add information, and might lead to confusion due to their position:

Screenshot 2024-12-19 at 16 40 02
Screenshot 2024-12-19 at 16 39 51

This one gives you no new information but stops you from sorting the table by name. This tooltip comes up if the cursor is even slightly below the header cell, meaning it might be frustrating for users trying to sort a column quickly:

Screenshot 2024-12-19 at 16 40 18

# 🧪 Testing

- [ ] Go to catalogue>items and hover on the columns' headers
- [ ] Go to other tables around the app and hover on the columns' headers
- [ ] See if the tooltips avoid repeating themselves

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
@@ -72,10 +72,10 @@ export const HeaderCell = <T extends RecordWithId>({
         {sortable ? (
           <div>
             {t('label.click-to-sort')}
-            {` ${columnLabel}`}
+            {` ${columnLabel.toLocaleLowerCase()}`}
           </div>
         ) : (
-          columnLabel
+          !description && columnLabel
         )}
       </>
     );


lache-melvin (approved):
Epic, LGTM! 🔥 

> Is it worth creating an issue for this - at the moment, you get tooltips for every single entry in a table, which add no information, and stop you interacting with what's covered by them

This is there so that if the cell content is bigger than the cell width, it would show `Start of item nam...` and then the full name with the tooltip on hover. I agree it feels a bit overkill to have them everywhere, but wouldn't raise issue for now. The tooltips don't work well on touch screens, so are getting a rethink in the new designs anyway 😁

mark-prins:
> I changed the tooltips so that now it says 'sort by [column label in lower case]' - before it did not transform it to lower case. Is this ok?

No, I disagree with the change. It makes some sense when reading the line as a sentence however it makes more sense to me to have the column name as a distinct object: e.g. `Serial Number` is a thing and you can sort by it so the phrase `Sort by Serial Number` is valid and makes more sense than `Sort by serial number` as the 'serial number' here is an abstract concept.

It also causes problems with abbreviations; I don't think this is valid:

Screenshot 2024-12-30 at 9 07 27 AM

will make an issue to revert.

zachariah-at-msupply:
Yep that makes sense! I'll revert the change once that issue has been made, could you post a link to it here?

PR #5867, opened 2024-12-19
remove currentItem loading check

### LINKED ISSUE #5857: Infinite Loading spinner when adding some items to a Requisition
## What went wrong? 😲

Encountering this behaviour: 

[VIDEO HERE](https://drive.google.com/drive/folders/1KaY2iaMJvMa766FVSTmPM0zNbbmSlyBv)

Upon refresh the item seems to be added, also I know we had an issue similar to this raised somewhere, just cannot locate it!

Spotted by @noel-yeldos, confirmed recreated by myself and @GeronimoJohn 

## Expected behaviour 🤔

Shouldn't be loading spinner

## How to Reproduce 🔨

1. Create a requisition
2. Try adding an item - loading spinner does not happen for every item, but everyone was able to recreate with Ibuprofen
3. See loading spinner
4. Refresh the page - the item is now added to the requisition

## Your environment 🌱

- Open mSupply Version: V2.4.3

- Database type: SQLite3

GeronimoJohn:
Fixes #5857 

# 👩🏻‍💻 What does this PR do?

Adding an item to a requisition can sometimes cause the browser to enter an infinite loading state. This is due to race conditions in the code.

The loading check relies on `isLoading || !currentItem` to determine if data has been fetched and if `currentItem` is populated. However, `currentItem` is derived from the fetched data, and there’s a moment when `isLoading` becomes false (data fetched), but `currentItem` is still empty. This timing mismatch triggers the issue.

Screenshot 2024-12-20 at 10 02 09 AM

As seen as well with the length of the array return from data. Lines grows in size once data is retrieved.

Screenshot 2024-12-20 at 10 19 25 AM

Removing `currentItem` from the loading check fixes the infinite loading state issue.

# 🧪 Testing

- [ ] Add a few items to a requisition
- [ ] Page shouldn't be stuck in a infinite loading state at any point

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
@@ -38,7 +38,7 @@ export const ResponseLineEditPage = () => {
     });
   }, [currentItem]);
 
-  if (isLoading || !currentItem) return <BasicSpinner />;
+  if (isLoading) return <BasicSpinner />;
   if (!data) return <NothingHere />;
 
   return (
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditPage.tsx
@@ -7,7 +7,7 @@ import {
   useBreadcrumbs,
   useParams,
 } from '@openmsupply-client/common';
-import { useResponse } from '../../api';
+import { ResponseFragment, useResponse } from '../../api';
 import { ListItems } from '@openmsupply-client/system';
 import { ResponseLineEdit } from './ResponseLineEdit';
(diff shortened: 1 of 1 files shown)


andreievg:
Would be happy to approve with this: https://github.com/msupply-foundation/open-msupply/compare/5857-requisition-infinite-loading-spinner...5857-requisition-infinite-loading-spinner(User-wrapper-comonent-to-wait-for-state)

and this:

https://github.com/msupply-foundation/open-msupply/compare/5857-requisition-infinite-loading-spinner(User-wrapper-comonent-to-wait-for-state)...5857-requisition-infinite-loading-spinner(User-wrapper-comonent-to-wait-for-state)(Remove-default-empty-array)

As per our covo, it's a bit hard to figure out all of the race conditions and best to just wait for all of the state to be available before getting currentItem

GeronimoJohn:
> The explanation doesn't match the behaviour I'm seeing, which is as you've described, but I don't have the problem:
> 
> image

That's interesting. I think this is the point where it happens for me, but I might have missed another factor. Though as per Andrei's comment, the recent push should cover the race conditions.

andreievg (approved):
Looks great, thanks @GeronimoJohn

PR #5875, opened 2024-12-19
Check for 01 ids before trying to create an asset...

### LINKED ISSUE #5233: Make Inbound & Outbound Shipments plural in the navigation menu
## What went wrong? 😲

Let's make Inbound and Outbound Shipments plural in the navigation menu! Everything else is plural and it's far more likely users have many many entries in that area of omSupply than just one.

Screenshot 2024-10-25 at 4 23 55 PM

## Expected behaviour 🤔

They should be plural!

## How to Reproduce 🔨

1. Log into your omSupply
2. Look at the navigation menu

## Your environment 🌱

- Open mSupply Version: `V2.3.1-RC1`
- Legacy mSupply Central Server Version: `V7.18.6`

- Platform: `Browser (Google), Desktop (Windows), Tablet`

- Database type: `SQLite3`

jmbrunskill:
Fixes #5233

# 👩🏻‍💻 What does this PR do?

I found that scanning our own printed asset labels wasn't working for me due to IDs that started with `01` which is a kind of Valid GS1. The id was interpreted badly as GS1 and didn't navigate as expected.

I've just created a check to see if an ID looks like an ID, but maybe we should look at changing our label QR code structure to have more information in it which would make it more identifiable. Would be nice to know which store it's for for example, so it could warn if you're scanning a barcode from a different store for example.

# 🧪 Testing

- [ ] Create an ASSET with an id that starts with `01`
- [ ] Print asset label
- [ ] Try to scan asset label
- [ ] Check it navigates to the right place
- [ ] Try to see if we can create a valid GS1 that looks like our ID?

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
+++ b/client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
@@ -17,6 +17,7 @@ import {
   AssetLogStatusInput,
   UserPermission,
   useAuthContext,
+  RegexUtils,
 } from '@openmsupply-client/common';
 import { AppRoute } from '@openmsupply-client/config';
 import { useAssets } from '../api';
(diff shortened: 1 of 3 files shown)


jmbrunskill:
Seems like this was more isolated to the ID `0193e0d1-a3da-7b21-aa83-604efbc7c17e`
It just happens to have a `21` Serial Number field in potentially the right place or the GS1 detection code is very forgiving?
Since it's not happening for ever `01*` id I think we close this until we do something more structured for the Asset Labels

PR #5883, opened 2024-12-20
5841 gaps item import script is incorrect

### LINKED ISSUE #5841: GAPS item import script is incorrect
## What went wrong? 😲

1.
Some items were incorrectly conflated due to duplicate mSupply item_codes created.
We need to separate these out so they're not duplicated.

2.
Packaging Volumes are off by a factor of 1000 (needs a conversion)

3.
Cold storage types are mapped based on names or strict mapping, should be based on a range e.g. +5 should be between 0 and 10 degrees -20 between -30 and -10, etc.

4.
Variant names are too long and can't be viewed correctly in the UI

## Expected behaviour 🤔

Improved setup of gaps items

## How to Reproduce 🔨

1. Run `create_gaps_item_sql_script`
2. Run the output in the generated file (on a fresh db)
3. Check the items are setup correctly

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #5841

# 👩🏻‍💻 What does this PR do?

Uses a new mSupply Item Code (created using the python script in this PR) but now is in the spreadsheet as a VLOOKUP.
Volumes are now divided by 1000 to make them in litres
Cold Storage Type look up is improved.

If you're brave and want to mostly start again.. you could consider running this.
However might not be correct, definitely don't run it on production..

```
DELETE from bundled_item;
DELETE from packaging_variant;
DELETE FROM item_variant WHERE id not in (SELECT coalesce(item_variant_id,'noid') from stock_line)
AND id not in (SELECT coalesce(item_variant_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_variant_id,'noid') from stocktake_line);
DELETE FROM master_list_line WHERE id not in (SELECT item_link_id FROM stock_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from stocktake_line);
DELETE FROM item_link WHERE id not in (SELECT item_link_id FROM stock_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from invoice_line)
AND id not in (SELECT coalesce(item_link_id,'noid') from stocktake_line)
AND id not in (SELECT coalesce(item_link_id, 'noid') FROM vaccine_course_item);
DELETE FROM item WHERE id not in (select item_id from item_link);
```

# 🧪 Testing

- [ ] start with afresh GAPS database.
- [ ] Run gaps_item_inserts.sql on your Central Server Postgres database
- [ ] Check that items that previously had been conflated are now fixed
- [ ] Check the looks for cold storage works correctly with new "default" ones setup in the issue
- [ ] Check volumes are correct

# 📃 Documentation

  1.
  2.


server/gaps/items/gaps_code_creator.py
```diff
@@ -0,0 +1,108 @@
+import csv
```
CarlosNZ, line 1: As discussed, a comment up the top here to explain the purpose/status of this script would be good

CarlosNZ (commented):
Holding off on further review pending @jmbrunskill looking at updating the spreadsheet to include item ID, as discussed IRL

jmbrunskill:
@CarlosNZ I've decided that there's too many ids to manage in that spreadsheet.
However, I've changed the spreadsheet to use the item_code as the unique lookup, which means we could change item names without loosing our link to the item_id in the `id_lookup.json` if we want to change the item_codes again then they'll get new ids, let's cross that bridge if we need too.

CarlosNZ (approved):
Cool, I've tested it as best I can for now. I think I need a bit more depth of knowledge to really look for edge cases, but I think should be good for now. Thanks

PR #5891, opened 2024-12-23
5665 extra window to show item info

### LINKED ISSUE #5665: CIV 2.5: Extra window to show item info
Screenshot 2024-12-03 at 8 21 13 AM

Only shown if `useExtraFieldsForRequisitions` + `useConsumptionAndStockFromCustomersForInternalOrders` is enabled

- Show customer store information for each item: 
	- AMC
	- Stock on hand
	- Adjustment
	- Period
	- Outgoing stock 

Get data from requisitions of the same period (?)

Data is aggregated from lower district facilities. 

Needs backend implementation to get all values and also frontend implementation

### Moneyworks Jobcode 🧰

OMS:CIDV

roxy-dao:
Fixes #5665

# 👩🏻‍💻 What does this PR do?
- Migration to add elmis code to program to allow filtering out different programs by elmis code
- Logic for getting item information from requisition
- Loader to get item information to load on Requisition Line node
- Remove unused file which has already been refactored into the line edit page for Internal Orders
- Display the item information window in Internal Order's line edit if the user has the correct preferences on

Get customer item information based on program. These values are based on any requisitions they have for the same program (or program with the same elmis code). The majority of the values used for customers is based on their latest requisition ( the one with the last update based on either: finalised datetime, sent datetime, or created datetime )

For customers with Requisitions: 
- Customer's stock in unit is based on their available stock on hand (to be changed to total stock on hand) + their daily usage (following OG)
- Adjustments: incoming + Inventory Adjustments - outgoing 

For customers without requisitions (or 0 values for AMC in said requisitions): 
- AMC: When a store has no AMC for an item line for a program and a period, instead of using “0”, pull the previous month’s AMC (m-1), if there is no AMC at m-1, pull the AMC at m-2, etc. until data > 0 is found or until m-6. If they still have no AMC at m-6 then it is the average AMC of all customers with requisitions. 

For current logged in store: 
- AMC: Area AMC which is the sum of all customer AMCs
- Stock in units: the stock on hand at the end of the period ( if none then it is the balance closest to the end of the period )
- Adjustment: (based on start and end date of the period) additions in units (all delivered Inbound Shipments with item) + Inventory Adjustments ( all Inventory Additions and Inventory Reductions for period) - losses in units ( all picked Outbound Shipments in period )

The table header isn't sticky since I had to fix the height of the table and allow for overflow >.>. Also not sure why the values differ from OG. Have double checked the DB entries and seems to be correct on my end?

# 🧪 Testing

- [ ] Note: The item info window will only show for stores who have customers that use them as a supplying store
- [ ] Make sure you have past requisition orders from customers and maybe some item consumption/replenishment history
- [ ] Create a program Internal Order for a store that has the preferences `Extra fields in requisition` and `use consumption data from customer's` 
- [ ] Click on an item in the Internal Order to be redirected to the line edit page
- [ ] You should see a table with customer item info below the inputs

# 📃 Documentation

  1.
  2.


client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/ItemInformation.tsx
```diff
@@ -0,0 +1,98 @@
...
+      dense
+    />
+  );
+};
+
+export const ItemInformationView = ({
+  itemInformation,
+  storeNameId,
+}: ItemInformationProps) => (
+  <Box maxHeight={200} width="100%" borderRadius={3} overflow="auto">
```
roxy-dao, line 90: Had to fix height cause of the limited real estate... ;-; headers aren't stickyyyy

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
```diff
@@ -77,7 +85,7 @@ export const RequestLineEdit = ({
             label={t('label.stock-on-hand')}
             sx={{ marginBottom: 1 }}
           />
-          {isProgram && useConsumptionData && (
+          {isProgram && extraFields && (
```
roxy-dao, line 88: These were using the wrong preferences... Values should be calculated in backend based on consumption preference

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
roxy-dao, line 1: Removed this since it was forgotten in another PR. No longer used, has all been refactored into the page edit

server/service/src/requisition/request_requisition/requisition_item_info.rs
```diff
@@ -0,0 +1,349 @@
...
+            > latest_requisition.requisition_row.created_datetime
+        {
+            *latest_requisition = requisition.clone();
+        }
+    }
+
+    Ok(latest_requisitions.into_values().collect())
+}
+
+fn get_store_information(
```
roxy-dao, line 257: All logic here should follow OG... >>

mark-prins (approved):
🎉  pheeewww! that's a lot of logic 🙏 Nice work getting thru all of that.

For the store calculations - the mSupply version is looking at any prior customer requisition rather than only those in the current period. This gives differences in calcs. The AMC is (I think) different bc of the days out of stock calculation  which should be improved with the back end plugin.

Screenshot 2025-01-03 at 10 40 31 AM


Here's a [suggestion](https://github.com/msupply-foundation/open-msupply/compare/5665-extra-window-to-show-item-info...5665-extra-window-to-show-item-info(layout-suggestion)) for updating the layout to make the table fit and add v-scroll.

Can I suggest this in lines 66+ of `ItemInformation.tsx`?

```
  const itemInformationSorted = draft?.itemInformation
    ?.sort((a, b) => a.name.name.localeCompare(b.name.name))
    .sort((a, b) => b.amcInUnits - a.amcInUnits)
    .sort((a, b) => b.stockInUnits - a.stockInUnits);
```

I was trying to find my store and having the lines sorted randomly was annoying 😉

mark-prins:
oh! another suggestion: in the client-side sort routine either
(a) sort by store name only, or
(b) show non-zero amc/unit qty values at the top

I would use the store name sort myself but it isn't what OG does. that has non-zero at the top and sorted within that

PR #5892, opened 2024-12-23
5826-Plugin-base

### LINKED ISSUE #5826: Plugins via BoaJS
Epic: https://github.com/msupply-foundation/open-msupply/issues/2951

Add plugin provider and example of AMC and suggested quantity override as per draft PRs but with BoaJS

andreievg:
Fixes #5826

# 👩🏻‍💻 What does this PR do?

Add base plugin provider with BoaJS as the engine for JS plugins.

Only AMC pluging is implemented, although suggested quantity is mentioned in the issue, this will be done in follow up PRs.

Changed the plugin provider slightly, to try to reduce number of places to change when adding new plugin. 

Also tried to balance out funky/complex rust types with readability and dryness.

Another big difference to drafts is error handling, wanted to have pretty strong error handling, which also meant that where item stats is used, repository error was not sufficient so added PluginOrRepositoryError which lead to quite a few files to be touched. 

There are quite a few TODO, they would be captured in the epic (carry over).

I would love to do internal docs, but the shape might still change a little bit, once it's finalised will get into docs

# 🧪 Testing

`cargo run --bin remote_server_cli -- initialise-from-export -n reference1`

`cargo run --bin remote_server_cli -- refresh-dates`

`cargo run` 

See AMC in item catalogue

Then run curl command in amc.js

See AMC in item catalogue (should be changed)

Play around with curl command in amc.js, reload again and see the change

# 📃 Documentation

  1.
  2.


server/server/src/upload_plugin.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub fn config_upload_plugin(cfg: &mut web::ServiceConfig) {
+    cfg.service(plugin);
+}
+
+// Should be in plugin_provider
+#[derive(Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct UrlParams {
+    r#type: PluginType,
```
jmbrunskill, line 29: I assume you're refine this a bit in a later PR?
Or at least add some docs about what these types are?
I think `variant_type` could have a better name but struggling to thing of one right now, maybe `plugin-technology`? Could also just leave that off until we actually have a different type?

server/server/src/upload_plugin.rs
```diff
@@ -0,0 +1,73 @@
...
+    url_params: web::Query<UrlParams>,
+    settings: Data<Settings>,
+) -> HttpResponse {
+    let file_service = StaticFileService::new(&settings.server.base_dir).unwrap();
+
+    let static_file = file_service
+        .move_temp_file(file, &StaticFileCategory::Temporary, None)
+        .unwrap();
+
+    let bundle_base64 = BASE64_STANDARD.encode(fs::read(static_file.to_path_buf()).unwrap());
```
jmbrunskill, line 57: Just a random thought, but do we actually need to BASE64 this? Can't it just be a string?
Not sure what the advantages/disadvantages would be though...

server/service/src/backend_plugin/boajs/call_plugin.rs
```diff
@@ -0,0 +1,82 @@
...
+        .cloned()
+        .unwrap();
+
+    let key = js_string!(name);
+    let plugin = plugins
+        .get(key, context)
+        .unwrap()
+        .as_callable()
+        .cloned()
+        .unwrap();
```
jmbrunskill, line 73: Is this unwrap safe? Should probably handle this being missing right?

server/service/src/backend_plugin/boajs/methods/log.rs
```diff
@@ -0,0 +1,16 @@
+use boa_engine::*;
+use log::info;
+
+// Log method will log all arguments
+pub(crate) fn bind_method(context: &mut Context) -> Result<(), JsError> {
+    context.register_global_callable(
+        JsString::from("log"),
+        0,
+        NativeFunction::from_copy_closure(move |_, args, _| {
+            info!("from js {:#?}", args);
```
jmbrunskill, line 10: Would be nice to handle strings here as a special case to make the log a bit cleaner.

```
ebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36" 0.039822
2025-01-06 17:16:32.156279000 [INFO ] <service::backend_plugin::boajs::methods::log:10>:from js [
    String(
        "Running SQL query",
    ),
]
```

server/service/src/backend_plugin/examples/amc/amc.js
```diff
@@ -0,0 +1,45 @@
...
+
+// TODO type sharing and type sharing
+
+// TODO Should come from settings
+const DAY_LOOKBACK = 800;
+const DAYS_IN_MONTH = 30;
+
+let plugins = {
+  average_monthly_consumption: ({ store_id, filter }) => {
+    const item_ids = filter.item_id.equal_any;
```
jmbrunskill, line 12: What if the filter is more complex?
I think our interfaces should be much more closely controlled for plugins?
Passing a filter object seems a bit to easy to screw up? Although I guess it's up to the plugin developer to know what to expect...

It's a bit of a higher level question about how we'll defined out plugin interfaces, would be keen for a discussion...

jmbrunskill (commented):
Looking good Andrei.

Keen to talk about a couple of things in person.

1. Defining what we pass to plugins, how do we decide what to expose and how do we document it? `Filter` feels dodgy to expose to me, but maybe it's ok as it is in the graphql...
2. Do we need to provide some kinds of backwards compatibility, if interfaces change for example?

server/server/src/upload_plugin.rs
jmbrunskill: Discussed in person, happy to keep base64

jmbrunskill (changes requested):
Happy to merge once we have 

- [ ] https://github.com/msupply-foundation/open-msupply/pull/5929 (or similar)
- [ ] Fix the unwraps()

server/service/src/backend_plugin/boajs/methods/log.rs
andreievg: > Would be nice to handle strings here as a special case to make the log a bit cleaner.

Logging is only for devs, yes it would be great to handle a bit better here, but it's could be any array of args, and extracting string and formatting them differently seems like quite a bit of code. I feel like the display is adequate for the target consumer. 

Please let me know if you feeling strong on this one

andreievg:
> Do we need to provide some kinds of backwards compatibility, if interfaces change for example?

Versioning would work similar to the reports, we will try to be backwards compatible but there is a chance that interface would change, which would require a new version, will explain in full plugin documentation (in the epic versioning is mentioned a few time, i think that's should capture an action for this comment)

server/service/src/backend_plugin/boajs/call_plugin.rs
andreievg: Thanks, i missed this: 7563e8d975642cbad1cc7878b70b05a8563cc22c

jmbrunskill (approved):
Thanks for the update @andreievg 
I haven't tested again, but looks good. Not sure if you're ready to merge but happy for you to merge if you're ready...

andreievg:
Thanks for reviews @jmbrunskill !
It's ready to merge yeah, there is test error but it's fixed in downstream PRS

PR #5896, opened 2024-12-23
inbound shipment cost price is NaN

### LINKED ISSUE #5768: Inbound shipment placeholder lines have `$NaN` cost price when grouped by item
## What went wrong? 😲

Screenshot 2024-12-12 at 2 54 21 PM

`0 / 0 == NaN`:
https://github.com/msupply-foundation/open-msupply/blob/20f5074f9cd298f89ad180cf04ea110d629da2b5/client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.ts#L214-L221

The `getSellPrice` has an `isPlaceholderRow` check but I think needs to be updated to get the cost price??

I feel like we've fixed this, or something very similar, before. Would be good to add unit tests around the returned cost price value so we don't have another regression 🙏 

## Expected behaviour 🤔

Shows $0.00

## How to Reproduce 🔨

1. New inbound shipment
2. Add items from master list
3. group by item
4. View cost price column

## Your environment 🌱

- Open mSupply Version: v2.4.1

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #5768

# 👩🏻‍💻 What does this PR do?

Creates an ArrayUtil function to handle the calculation of unit cost price (average from a 'grouped by' array). Added tests to check the new function.

Also updated SellPrice to CostPrice for the ungrouped result.

The existing function was returning NaN where ```costPricePerPack``` and ```numberOfPacks``` were both 0. Typescript does not check if NaN can be a result of a function that contains only number types, and allows ```0/0 = NaN```. 

There are already several NaN checks in place (eg in the Currency cell definition). This calculated cost price for the 'grouped by' option is not picked up as NaN from the cell type, so implemented the ArrayUtil to handle it.
This matches the pattern of other columns that use ArrayUtils for calculations

It might be worth having another ArrayUtils calculation for 'Sell Price' 
I can replicate this error for grouped sell price in ```Outbound Shipment``` -> ```Detail View``` 

Some other areas also using sell price column calculations, but I was unable to replicate NaN without the grouped toggle being available:
- ```Outbound Shipment`` -> ```Detail View``` -> ```Outbound Line Edit```
-  ```Supplier Returns``` -> ```Detail View``` 
- ```Prescriptions``` -> ```Detail View``` 

# 🧪 Testing

- [ ] Create new inbound shipment
- [ ] Add items from master list
- [ ] Toggle group by item
- [ ] View cost price as $0.00 or where there is data, a valid currency format

# 📃 Documentation

  1.
  2.


mark-prins (changes requested):
If I can pop in a quick request? the outbound shipment suffers the same fate, as you've noted:

Screenshot 2024-12-30 at 12 27 56 PM

Would be nice to fix that too.

I would also say that there's no need to add a function to ArrayUtils if it is only used in one place.. though yes, it makes the test creation easier 😄 

To that end.. I'd make the ArrayUtils function more generic and implement in both inbound and outbound. ticks both boxes

I've made the change in this [suggestion branch](https://github.com/msupply-foundation/open-msupply/compare/5768-inbound-shipment-NaN-cost-price...5768-inbound-shipment-NaN-cost-price(suggestion)(suggestion))

mark-prins:
An alternative would be to fix the CurrencyCell.. as that input is demanding a number and not handling gracefully the bad data. I'd prefer to have `-` shown rather than `$0.00` as the error state but hey, can live with it 🙏

aimee-mcneil-melville:
Great thank you Mark! 
Your suggestion branch looks awesome, and definitely makes sense calculating as an average regardless of if cost or sell price is passed in - I'll pull this in!

In regards to handling the data in Currency Cell, I did look at this but didn't come up with any answers. To me it looked like the NaN is handled prior to the calculation of the average, thus not returning the error or ' ' in this case.

https://github.com/msupply-foundation/open-msupply/blob/d36246d37ea8f48331f3d5a7e3fd68fa3dbbd157/client/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx#L73-L82

If you have any wisdom on this that'd be great!

mark-prins:
> In regards to handling the data in Currency Cell,

The `formatter` function is only used by the `BasicCell`
You could change the `CurrencyCell` to handle string input values:

```
export const CurrencyCell = <T extends RecordWithId>({
  column,
  rowData,
  currencyCode,
}: CellProps<T> & { currencyCode?: Currencies }): React.ReactElement<
  CellProps<T>
> => {
  const { c } = useCurrency(currencyCode);
  const value = column.accessor({ rowData });
  if (typeof value === 'string') {
    return <div>{value}</div>;
  }

  const price = Number(value ?? 0);
...
```

and update `ArrayUtils.getAveragePrice` to do this:

```
    return Number.isNaN(averagePrice) ? UNDEFINED_STRING_VALUE : averagePrice;
```

which gives this:

image

but I wouldn't! 😄 
you'd also have to update the un-grouped display and it's just not worth the dev or QA time!

aimee-mcneil-melville:
Great thank you for going into that for me! 
Is this one ready to merge now?

aimee-mcneil-melville:
I can't merge until it's been ticked as approved 🙏

mark-prins (approved):
🚢

PR #5898, opened 2024-12-24
5827-Backend-plugin-table

### LINKED ISSUE #5827: Backend plugin table

A table that holds plugin bundle, with unique plugin code that is central data syncs to remotes

andreievg:
Fixes #5827

# 👩🏻‍💻 What does this PR do?

Add backend plugin table and translations for central sync.

Moved some types to repository, there are questions about dev efficency and needing to add edit/update things in many places, there is no real reason to have plugins types and plugin variant type in different service (apart from organisation), ideally though it will be in common type (as a root crate)

Btw strum allow you to check all of the enum variants (to make sure enum is added to postgres)

As per the comment left below, ended up changing from single 'type' to 'types' on backend_plugin table. A single plugin should be able to server multiple interfaces

# 🧪 Testing

I ran both postgres and sqlite test, pass, further test will be done when plugins are added and when they sync to remote sites

# 📃 Documentation

  1.
  2.


andreievg:
In follow up PR schema is changed to have `types` as an array, will update this PR

jmbrunskill (approved):
Looks good to me, I haven't really tested as we need to get the actual upload endpoints working to test it properly. Nothing actually adds a plugin to this table currently right?

Issue #5554, opened 2024-11-25
Item variant usability


mark-prins:
## What went wrong? 😲

I added some item variants, and there's a few aspects of the UI that could be easier to use.

## Expected behaviour 🤔

1. The 'no variants' message could have a link to add an item on it
2. Pressing enter to select a manufacturer on the 'add/edit variant'  modal saved my changes and closed the modal. I was expecting enter to select the one manufacturer that was returned by my search 
3. It would be helpful to have the first input field on the form auto focussed ( I use keyboard as much as possible )
4. if there are no cold storage types then I get a 'loading...' message forever. This would be better to say that there are no types configured

![Screenshot 2024-11-25 at 7 03 42 PM](https://github.com/user-attachments/assets/a12e4919-fddd-49d9-bab4-13a85ffc71a9)

![Screenshot 2024-11-25 at 7 01 47 PM](https://github.com/user-attachments/assets/c7f5fba5-28eb-4675-9c88-25f66954fbe0)

## How to Reproduce 🔨

Steps to reproduce the behaviour:

6. Click on '....'
7. Scroll down to '....'
8. See error

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

DhanyaHerath:
good first issue, but make PR for each one

zachariah-at-msupply:
Hi, I couldn't replicate point number 2 - for me it works as you said you wanted it to work Mark. I'm not sure why. Do you have any ideas about how to replicate it? I'm using the latest version of develop to test this - if you try to use that, has the issue resolved itself? Let me know; I'll keep working on this.

mark-prins:
the Enter key issue is still there and slightly annoying but can probably leave it out. it sort of work, just caught me out when using the keyboard..

https://github.com/user-attachments/assets/99d97957-ef03-47b0-a3d4-300ca0ae2f5c

if you enter a name, then open the auto complete for the manufacturer, start typing and press enter then the modal closes. If you use the down arrow and highlight an entry that works.. although if you click enter again then the modal is closed even though the cursor is in the autocomplete.

my take: fixing will probably break the 'Enter clicks ok and closes the modal' behaviour and that would be worse than the current quirk. best to leave it.

zachariah-at-msupply:
[...]

PR #5904, opened 2024-12-29
add link to create new variant if none

zachariah-at-msupply:
Fixes Part 1 of #5554

# 👩🏻‍💻 What does this PR do?

Adds a create new variant link on the catalogue>item>variants tab, if there are no variants present.

Screenshot 2024-12-30 at 09 02 38

The capitalisation seems a bit dodgy. It seems all the 'create a new one' links will be similar. Using a function called toLocaleLowerCase() in the NothingHere.tsx file can solve this - should I create an issue to do that?

# 🧪 Testing

- [ ] log in to your _open_ mSupply central server
- [ ] go to cataogue>item page and then click on an item that has no variants to see its detail view
- [ ] go to the variants tab
- [ ] see if the link that says 'create a new one' functions the same as the 'add variant' button on the top right of the variants tab page.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
@@ -48,7 +48,12 @@ export const ItemVariantsTab = ({
       </AppBarButtonsPortal>
       <Box flex={1} marginX={2}>
         {itemVariants.length === 0 ? (
-          <NothingHere body={t('messages.no-item-variants')} />
+          <>
+            <NothingHere
+              body={t('messages.no-item-variants')}
+              onCreate={() => onOpen()}
+            />
+          </>
         ) : (
           itemVariants.map(v => (
             <ItemVariant
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
@@ -48,12 +48,10 @@ export const ItemVariantsTab = ({
       </AppBarButtonsPortal>
       <Box flex={1} marginX={2}>
         {itemVariants.length === 0 ? (
-          <>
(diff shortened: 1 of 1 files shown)


client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
```diff
@@ -48,7 +48,12 @@ export const ItemVariantsTab = ({
       </AppBarButtonsPortal>
       <Box flex={1} marginX={2}>
         {itemVariants.length === 0 ? (
-          <NothingHere body={t('messages.no-item-variants')} />
+          <>
+            <NothingHere
+              body={t('messages.no-item-variants')}
+              onCreate={() => onOpen()}
+            />
+          </>
```
mark-prins, line 56: Couple of minor things to change:

* the syntax `() => ` here will create a new function, which is an overhead that isn't required. The `onOpen` doesn't need parameters so can be passed directly
* The `<>...</>` isn't required. that's an empty element which you'll only need if there are multiple children and you need to provide only one

```suggestion
          <NothingHere
            body={t('messages.no-item-variants')}
            onCreate={onOpen}
          />
```

mark-prins (changes requested):
👍 Looks good thanks!
I have a requested change - it's a minor point but worth getting into the habit of which is why I mention it.

Re the question about capitalisation - the place that you would change it is in the translation, and not calling a lowercase function. but don't 😄 I think the caps is ok - if anything is needed, it would be a separator or some punctuation.. but it's fine as it. at least it's consistent!

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
zachariah-at-msupply: Thanks Mark! I'll bear that in mind, and put those changes into a commit.

mark-prins (approved):
🚢  Ship it! (thanks!)

PR #5915, opened 2025-01-05
5821 email sending

### LINKED ISSUE #5821: Implement sending of emails
Using [Notify](https://github.com/msupply-foundation/notify/blob/main/backend/service/src/email/mod.rs) implementation as a reference, we need:

- [ ] Email service to process and try send all unsent emails
- [ ] Scheduled task to trigger this email process
- [ ] Mailhog set up for testing
- [ ] Add email configuration to app config (required only for OMS central)

lache-melvin:
Fixes #5821

# 👩🏻‍💻 What does this PR do?

Adds the email processing/sending (same implementation from Notify)

Adds use of Mailhog for testing, and the `email-tests` feature flag.

Should I add mail test_connection to the test_connection tool? I don't think we'll know whether it is central or remote (to "know" whether mail should be configured) but we could just info `mail not configured` at all times, imagine its usually support using this tool and they could use it to double check mail config on central?

Probably can be a next PR?

# 🧪 Testing

- [ ] Run `cargo test --features=email-tests --package service --lib -- email::email_test --nocapture`
- [ ] Is there a good way to write a test for actual sending of email? 
  - [ ] I just manually added an `email_queue` row, started my server, and saw that cron job had run because I got that email in MailHog
  - [ ] Probably easier to test once queueing of emails is in

# 📃 Documentation

  1.
  2.


client/packages/coldchain/src/Equipment/ListView/ListView.tsx
```diff
@@ -51,9 +51,11 @@ const AssetListComponent: FC = () => {
     AppRoute.Equipment
   );
 
-  const columnsToCreate: ColumnDescription<AssetRowFragment>[] = [];
+  const columnsToCreate: ColumnDescription<AssetRowFragment>[] = [
+    GenericColumnKey.Selection,
+  ];
```
lache-melvin, line 56: Sorry side quest, I saw this had gone missing

jmbrunskill:
> Should I add mail test_connection to the test_connection tool? I don't think we'll know whether it is central or remote (to "know" whether mail should be configured) but we could just info `mail not configured` at all times, imagine its usually support using this tool and they could use it to double check mail config on central?

I think it's a good idea? Can definitely be new issue/PR though. Maybe check with Kahn about how useful it would be...

jmbrunskill:
> * Is there a good way to write a test for actual sending of email?
Here's an example of what we're doing in HSH.

**Testing template rendering**
https://github.com/msupply-foundation/health-supply-hub/blob/94ee1099eb07ba23f76b41fe4228044df1552857/backend/service/src/quote/email/quote_confirmation.rs#L178

**Testing email sending**
https://github.com/msupply-foundation/health-supply-hub/blob/94ee1099eb07ba23f76b41fe4228044df1552857/backend/service/src/quote/email/quote_confirmation.rs#L211

I think it's fine to wait till you have an actual email to send to test...

server/server/src/scheduled_tasks.rs
```diff
@@ -0,0 +1,32 @@
+use service::service_provider::ServiceProvider;
+use std::{sync::Arc, time::Duration};
+use tokio::task::JoinHandle;
+
+static TASK_INTERVAL: Duration = Duration::from_secs(10);
```
jmbrunskill, line 5: I think we should maybe bump this up to 60s or something for now? Especially since we're spamming the logs on every run...

server/server/src/scheduled_tasks.rs
```diff
@@ -0,0 +1,32 @@
...
+        let send_emails = service_provider
+            .email_service
+            .send_queued_emails(&service_context);
+        match send_emails {
+            Ok(num) => {
+                if num > 0 {
+                    log::info!("Sent {} queued emails", num);
+                }
+            }
+            Err(error) => log::error!("Error sending queued emails: {:?}", error),
```
jmbrunskill, line 29: When not configured, we get almost duplicated errors here...
```
2025-01-06 14:29:05.157412000 [ERROR] <service::email:118>:Email settings not configured
2025-01-06 14:29:05.157883000 [ERROR] <server::scheduled_tasks:29>:Error sending queued emails: NotConfigured
```
Maybe remove the other one and keep this?

server/server/src/scheduled_tasks.rs
jmbrunskill: Although it is really nice to have it respond quickly when testing emails, so I'm ok with leaving this!

jmbrunskill (changes requested):
Looks good Lache.

2 things to follow up on, can be in new issues if you want to merge this?

- [ ] Adding `retry_at` logic https://github.com/msupply-foundation/open-msupply/issues/5821#issuecomment-2547056619
- [ ] Limiting the errors to just central server
```
        log::debug!("Processing Scheduled Tasks");
        if CentralServerConfig::is_central_server() {
            // Email sending is only supported on the central server
            let send_emails = service_provider
                .email_service
                .send_queued_emails(&service_context);
            match send_emails {
                Ok(num) => {
                    if num > 0 {
                        log::info!("Sent {} queued emails", num);
                    }
                }
                Err(error) => log::error!("Error sending queued emails: {:?}", error),
            };
        }
```
This would do it, but I'm not sure if it's better to do some kind of check for pending emails first (before even checking NotConfigured state)?
Then we could return the number of queued emails even if the configuration is missing? 

That's a nice to have though.

server/server/src/scheduled_tasks.rs
lache-melvin: Done [34f4aeb](https://github.com/msupply-foundation/open-msupply/pull/5915/commits/34f4aebee2c542578429f74a9b920480328a5c1b)

server/server/src/scheduled_tasks.rs
lache-melvin: Added it as a config option? What do you reckon? [532aa0e](https://github.com/msupply-foundation/open-msupply/pull/5915/commits/532aa0e50ef60811d30793367cad21251ac75b57)

lache-melvin:
Thanks @jmbrunskill, like these improvements 🙏

server/server/src/scheduled_tasks.rs
```diff
@@ -2,16 +2,19 @@ use service::service_provider::ServiceProvider;
 use std::{sync::Arc, time::Duration};
 use tokio::task::JoinHandle;
 
-static TASK_INTERVAL: Duration = Duration::from_secs(10);
+pub fn spawn_scheduled_task_runner(
+    service_provider: Arc<ServiceProvider>,
+    interval_secs: u64,
+) -> JoinHandle<()> {
+    let period: Duration = Duration::from_secs(interval_secs);
```
jmbrunskill, line 9: Why switch between period and interval?
It looks fine but wondering if we should stick to one name?

server/server/src/scheduled_tasks.rs
jmbrunskill: Nice, that's handy, and good for support to be able to tweak it too.

jmbrunskill (approved):
Thanks @lache-melvin looks great.

server/server/src/scheduled_tasks.rs
lache-melvin: Oh haha was just using the naming from the Duration package inputs, but I reckon I could just do it all in one go 😅

PR #5917, opened 2025-01-06
5673 civ 2.5 indicators for ds internal orders

### LINKED ISSUE #5673: CIV 2.5: Indicators for DS Internal Orders
- Use supplying store to identify higher level facilities
- In backend when an Internal Order is created for a higher level facility, sum all the indicator values (don't worry about strings for now)
- Extra panel to show customer indicator details
Screenshot 2024-12-03 at 8 20 24 AM
- Data can be surmised from Requisitions of the same period

### Moneyworks Jobcode 🧰
OMS:CIDV

roxy-dao:
Fixes #5673 

# 👩🏻‍💻 What does this PR do?
- Logic to get indicator values for the store's customers
- The indicator values are summed if it is a number (if they have made more than 1 requisition for the period)
- Displays the indicator information table in the indicator edit page under the input (must have use consumption data from customers for internal orders preference turned on)
- All columns for the line are shown in the indicator information table 

Screenshot 2025-01-07 at 7 56 33 AM
Screenshot 2025-01-07 at 7 56 37 AM

# 🧪 Testing

- [ ] Must have use consumption data from customers for Internal Orders preference turned on
- [ ] Create program Internal Orders (with indicators) for customers that have a supplying store
- [ ] Enter some indication data
- [ ] Send IO
- [ ] Go to supplying store
- [ ] Create a program Internal Order (with indicators) for a store that has customers using it as a supplying store. (Must be same period and program as the above IOs)
- [ ] Go to indicators
- [ ] See that there is now a table under the indicators input with information collected from customers

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/strings/IntlStrings.ts
```diff
@@ -38,3 +38,8 @@ export const useTranslation = (ns?: Namespace): TypedTFunction<LocaleKey> => {
     [t]
   );
 };
+
+export const translationExistsInLocale = (key: string): boolean => {
```
roxy-dao, line 42: Label in columns is now a string, so we want to check if the translation exists in the translation file first so determine if we should use the string or translation

client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/CustomerIndicatorInfo.tsx
```diff
@@ -0,0 +1,105 @@
...
+      return 'label.value';
+    default:
+      return columnName;
+  }
+};
+
+const CustomerIndicatorInfo = ({
+  customerInfos,
+}: CustomerIndicatorInfoProps) => {
+  const columnNames = customerInfos?.[0]?.indicatorInformation?.map(
```
roxy-dao, line 37: Checking first element to get all the column names (should be the same for all elements after the first)

server/graphql/requisition/src/program_indicator.rs
```diff
@@ -19,7 +19,7 @@ pub fn program_indicators(
     let user = validate_auth(
         ctx,
         &ResourceAccessRequest {
-            resource: Resource::QueryProgram,
+            resource: Resource::QueryRequisition,
```
roxy-dao, line 22: QueryPrograms is the wrong permission.

server/graphql/core/src/loader/requisition_indicator_information.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        let mut result = HashMap::new();
+
+        let indicator_info = self
+            .service_provider
+            .requisition_service
+            .get_indicator_information(&service_context, &store_id, &period_id, &program_id)?;
+
+        for loader_input in loader_inputs {
+            result.insert(loader_input.clone(), indicator_info.clone());
```
andreievg, line 58: I wasn't going to touch this PR and just approve as is, but I wasn't confident about this one, loaders should ideally return something that maps to the node requesting the loader, in this case I think the same result will be sent to all of the nodes requesting the info. 

I thought that customer indicator info should really come by line, so moved the loader there, which also means we don't need to be bothered with program_indicator id or columns when calculating IndicatorInfo, this suggestion is quite large, but I think it would be worth it long term:https://github.com/msupply-foundation/open-msupply/compare/5673-CIV-2.5-Indicators-for-DS-Internal-Orders...5673-CIV-2.5-Indicators-for-DS-Internal-Orders(Refactor-aggregate-indicator-info)

Sorry I haven't fully tested it, if it doesn't work happy to revert it and do sometime in the refactor period

andreievg (changes requested):
Thanks looks good, the main request is with customer_name_link_id check again name_id in indicator_information.rs, although I have quite a big suggestion to refactor the loader (this suggestion is not necessary but we can make a not of it for future work on this feature)

server/service/src/requisition/request_requisition/indicator_information.rs
```diff
@@ -0,0 +1,167 @@
...
+            match entry {
+                Some(entry) => {
+                    let indicator_info = entry
+                        .indicator_information
+                        .iter_mut()
+                        .find(|info| info.column_id == value.indicator_column_id);
+
+                    match indicator_info {
+                        Some(info) => {
+                            if let Ok(num) = value.value.parse::<f64>() {
```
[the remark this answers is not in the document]
roxy-dao: It is summing up the numbers if there is more than one value for period/customer ;3 and only gets 1 value if it is a string. In OG, they only show one value and date range (haven't checked the code yet)

client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/CustomerIndicatorInfo.tsx
```diff
@@ -0,0 +1,105 @@
...
+interface CustomerIndicatorInfoProps {
+  customerInfos?: CustomerIndicatorInfoFragment[] | null;
+}
+
+enum ColumnName {
+  Comment = 'Comment',
+  Value = 'Value',
+}
+
+const columnNameToLocal = (columnName: string) => {
```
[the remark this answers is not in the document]
roxy-dao: [70283ce](https://github.com/msupply-foundation/open-msupply/pull/5917/commits/70283ce52f75183e1205fedf405e12d7f95ba290)

server/graphql/core/src/loader/requisition_indicator_information.rs
roxy-dao: Looks good to me ~ thanks Andrei

andreievg (approved):
Looks good to me and thanks for all of the changes : )

PR #5922, opened 2025-01-06
5920 prescriptions filter out on hold stock

### LINKED ISSUE #5920: Prescriptions: Filter out on hold stock

jmbrunskill:
Fixes #5920

# 👩🏻‍💻 What does this PR do?

The new Prescriptions UI doesn't show the On Hold Column, instead we just won't show any on hold batches at all.

Did this as a front end change, I think it should be fine to do here rather than making backend changes and filters on the historical stock. Also makes it possible to show on hold easily again. There isn't usually a lot of stock on hold so a backend filter shouldn't be a big deal IMHO.

# 🧪 Testing

- [ ] Try creating a prescription with an item that has some stock on hold
- [ ] You shouldn't see the old hold stock.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/hooks/useDraftPrescriptionLines.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/hooks/useDraftPrescriptionLines.tsx
@@ -58,7 +58,10 @@ export const useDraftPrescriptionLines = (
     const invoiceLineStockLines = (lines ?? []).flatMap(l =>
       l.stockLine ? [l.stockLine] : []
     );
-    const stockLines = uniqBy([...data.nodes, ...invoiceLineStockLines], 'id');
+    const stockLines = uniqBy(
+      [...data.nodes, ...invoiceLineStockLines],
+      'id'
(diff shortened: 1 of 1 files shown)


client/packages/invoices/src/Prescriptions/LineEditView/hooks/useDraftPrescriptionLines.tsx
```diff
@@ -58,7 +58,10 @@ export const useDraftPrescriptionLines = (
     const invoiceLineStockLines = (lines ?? []).flatMap(l =>
       l.stockLine ? [l.stockLine] : []
     );
-    const stockLines = uniqBy([...data.nodes, ...invoiceLineStockLines], 'id');
+    const stockLines = uniqBy(
+      [...data.nodes, ...invoiceLineStockLines],
+      'id'
+    ).filter(stockLine => stockLine.onHold === false); // Filter out on hold stock lines
```
CarlosNZ, line 64:
```suggestion
   ).filter(stockLine => !stockLine.onHold ); // Filter out on hold stock lines
```

CarlosNZ (approved):
Works for me, thanks 👍

PR #5928, opened 2025-01-07
5431 when the supplier and the customer have different master list for the same program, you need to display the customer one when creating a manual requisition

### LINKED ISSUE #5431: When the supplier and the customer have different master list for the same program, you need to display the customer one when creating a manual requisition
## What went wrong? 😲

In Ivory Coast context, the same program have more than one master list. For the same program, there is one master list for the District, one for the Health centers (CS) and one for the Hospitals. They share a lot of items but there are some differences between each master list as you can see below there are items present in the Health center master list (in red) that are not in the District master list.
 
https://drive.google.com/file/d/1pPJYRrta-IGA5h4yxrCUqFbErBrP2GwC/view?usp=sharing

So Normally, you do a request requisition (internal order) using your master list ie the Health Center (CS) will use its master list to create an internal order to send to the District. So here you can see that all the requisitions have the master list version of the program that belongs to the Health Center that did the internal order.
 
https://drive.google.com/file/d/1Dnw6wLr-GDYCodFxc9qCZ-Hv2Aij1z1B/view?usp=sharing

But when you do a manual requisition for a customer you need to use the program master list of the customer rather than your own and that is why in mSupply OG, you choose the customer first and then the program. See the video below:

https://drive.google.com/file/d/1gb2oxkwDlRLptk3tFjrCaW8MyK-4BAQ8/view?usp=sharing

Currently in omSupply when you do a manual requisition you have to choose the program master list first and my only choice is the master list of the district but I should see the master list of the Health Center (CS)
https://drive.google.com/file/d/1kitLmNbEAzwzT7EHpVxg6CPJNXxEyhET/view?usp=sharing

The second problem, this causes is that by using the Supplier master list instead of the customer master list is that the list of customers is populated by other stores/facilities that uses that master list to order from a higher or equal level facility (See below)
https://drive.google.com/file/d/1n34Y86cIiJzR_LDbqcWnWeLzj6neSSom/view?usp=sharing

## Expected behaviour 🤔
You should be able to choose the master list that belongs to the customer instead of the master list that belongs to the supplier and for this you need to choose the customer before you choose the program.

## How to Reproduce 🔨

1. Go to Requisitions
2. Click on Create new requisition
3. Choose program 
4. See error. The supplier master list is showing instead of the customer one

## Your environment 🌱

- Open mSupply Version: 2.4.0
- Legacy mSupply Central Server Version: v19-0dev27

- Platform: Android 14

- Database type: SqLite 3

roxy-dao:
Fixes #5431

# 👩🏻‍💻 What does this PR do?
Allows users to select customers who aren't using the same master list based on elmis codee

# 🧪 Testing

- [ ] Create some master lists with matching elmis codes. 
- [ ] Set tags according to get program master lists to show
- [ ] Give one store one master list and another store the other master list
- [ ] Sync to OMS
- [ ] Try to create an IO/Requisition
- [ ] Should be able to select the master list for the store even if the store isn't using the master list (because of matching elmis code)

Scenario 2
- [ ] Don't set any elmis codes
- [ ] Set master lists and program tags accordingly for stores
- [ ] Create IO/Requisition
- [ ] Should be able to create program requisition for those master lists

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
Wooop wooop thanks for fix 😁

PR #5935, opened 2025-01-07
5018 discovery intermittently crashes

### LINKED ISSUE #5018: Discovery intermittently crashes
## What went wrong? 😲

An issue with [astro-dnssd](https://github.com/AstroHQ/astro-dnssd) crate causing OMS server to crash when ~~running two instances from the same machine (ie central server OMS + local OMS).~~

Edit: happening with only one instance now!

## How to Reproduce 🔨

Reproduction or error is intermittent, and might not always panic. 

1. Run OMS
3. See error, ie:

```
2024-10-07 12:26:26.495206000 [INFO] <astro_dnssd::os::apple::register:44>:Got reply
thread '<unnamed>' panicked at /Users/fergusroache/.cargo/registry/src/index.crates.io-6f17d22bba15001f/astro-dnssd-0.3.4/src/os/apple/register.rs:77:40:
called `Result::unwrap()` on an `Err` value: SendError { .. }
thread '<unnamed>' panicked at library/core/src/panicking.rs:221:5:
panic in a function that cannot unwind
```

Probably occuring as a result of unwrapping the error in astro-dnssd crate:

```
            Ok((name, regtype, domain)) => {
                if error_code == kDNSServiceErr_NoError {
                    let reply = DNSServiceRegisterReply {
                        regtype,
                        name,
                        domain,
                    };
                    tx.send(Ok(reply)).unwrap();
                    info!("Reply info sent");
                } else {
                    error!("Error in reply: {}", error_code);
                    tx.send(Err(RegistrationError::ServiceError(error_code)))
                        .unwrap();
                }
            }
```

## Your environment 🌱

- Open mSupply Version: Develop (but occuring on main / 2.3.0 also)

roxy-dao:
Fixes #5018

# 👩🏻‍💻 What does this PR do?
astro-dnssd crate had an update. There's no changelog, but what I can gather, they've removed the unwrap and have proper error handling now

# 🧪 Testing

- [ ] Run app for a while to see if it will crash with astro-dnssd error

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -230,7 +230,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.79",
+ "syn 2.0.95",
 ]
 
 [[package]]
@@ -269,7 +269,7 @@ dependencies = [
  "parse-size",
  "proc-macro2",
  "quote",
- "syn 2.0.79",
+ "syn 2.0.95",
 ]
 
 [[package]]
@@ -310,7 +310,7 @@ dependencies = [
  "futures-core",
  "futures-util",
  "mio",
- "socket2 0.5.7",
+ "socket2 0.5.8",
  "tokio",
  "tracing",
 ]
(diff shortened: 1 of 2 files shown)


lache-melvin (changes requested):
Screenshot 2025-01-08 at 2 15 57 PM

Don't think we can put this in for now - we have to be running rust 1.79 for discovery not to crash the app, until we put it in separate process... and these updates need at least 1.81 😭

PR #5940, opened 2025-01-07
5938 Move letter behind not android target flag

### LINKED ISSUE #5938: Unable to build android due to openssl
## What went wrong? 😲

Due to android build we use rustls rather then openssl, however `lettre` was added that uses openssl. 

## Suggested solution

Put functionality behind `letter` under a target flag

andreievg:
Fixes #5938

# 👩🏻‍💻 What does this PR do?

Move functionality hat uses lettre behind not android target flag, and lettre import also behind target flag.

I didn't change anything in the email functionality just moved it

I also change some error to just be string, and notice that some errors not used ?

Are we ok with unimplemented for android @lache-melvin, or should I just log it without panic ? Didn't lok at where it's used but was wondering if testers/devs are using android as central server if this would have an issue ? If so what should i return from send_queued_emails and test_connection ?

# 🧪 Testing

Would be great to also test that openssl would be ok on windows, not sure if it's alway available for dynamic linking


lache-melvin:
> Are we ok with unimplemented for android @lache-melvin, or should I just log it without panic ? Didn't lok at where it's used but was wondering if testers/devs are using android as central server if this would have an issue ? If so what should i return from send_queued_emails and test_connection ?

I'm ok with unimplemented, currently not supporting emails on android - will just need to let QA know they can't use this feature will tablet acting as central server:

https://github.com/msupply-foundation/open-msupply/blob/5938-Put-email-functionality-behind-not-android-target-flag/server/android/src/android.rs#L67-L68

lache-melvin (approved):
LGTM thanks andrei

lache-melvin:
> Would be great to also test that openssl would be ok on windows, not sure if it's alway available for dynamic linking

Are you happy for me to raise this as something we'd like QA to test or we should test asap?

andreievg:
@lache-melvin can you have another look please ? 
I had some doubts about openssl being available on windows and tries something else, which is way easier, using rust-tls as features flags for lettre. So reverted it all

server/service/Cargo.toml
```diff
@@ -61,7 +61,7 @@ rust-embed = { version = "8.4.0", features = ["include-exclude"] }
 extism = { workspace = true }
 base64 = "0.22.1"
 regex = { workspace = true }
-lettre = "0.11.11"
+lettre = { version = "0.11.11", default-features = false, features = ["rustls-tls", "smtp-transport", "builder"] }
```
roxy-dao, line 64: ❤️

roxy-dao (approved):
Yay LGTM

PR #5952, opened 2025-01-08
#5846 Fix Stock issue quantities for JSON Forms prescription

### LINKED ISSUE #5846: Family Planning/Reproductive Health module: we need to make it clear the dispensing if in units
## What went wrong? 😲

@Rajajajaja @richardmoizeau - need more context from you dudes, but from my understanding the client only dispenses in units, and not packs. We need to make it more clear how the dispensing works, and if it actually is dispensed in units

Currently this is what we are seeing: 

Screenshot 2024-12-18 at 3 33 42 PM

## Expected behaviour 🤔

The title of the column should be clear if it's units or packs
Also we need to clarify, what is being dispensed - if it is packs, is it possible to change it to units?

## How to Reproduce 🔨

1. Make sure the programs module is enabled for your store
2. Make sure Reproductive Health/Family Planning is enabled
3. Create an encounter for an existing/new patient
4. In the encounter, click on contraception method and issue a prescription from within the encounter
5. Note the editing window is unclear - are we issuing packs or units?

## Your environment 🌱

- Open mSupply Version: V2.4.2

- Database type: SQLite3

CarlosNZ:
Fixes #5846

# 👩🏻‍💻 What does this PR do?

The actual labelling to "units" had already been done -- this just updates the functionality to properly handle units for the input and display.

See inline comments

# 🧪 Testing

- Create prescription data for Family Planning encounter (as per previous PR, eg. https://github.com/msupply-foundation/open-msupply/pull/5581)
- Input some values to issue, ensuring you can't enter more than the available quantity
- Close the modal, and "Save"
- Should see "Prescription created" notification
- Go to the newly created prescription and ensure all values are consistent with what you entered previously (in terms of units/packs)

### DIFF
--- a/client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
+++ b/client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import {
+  ColumnAlign,
   DataTable,
+  ExpiryDateCell,
   NumUtils,
   TableProvider,
   createTableStore,
@@ -8,7 +10,7 @@ import {
   useTranslation,
 } from '@openmsupply-client/common';
 import { DraftStockOutLine } from '@openmsupply-client/invoices/src/types';
-import { PackQuantityCell } from '@openmsupply-client/invoices/src/StockOut';
+import { UnitQuantityCell } from '@openmsupply-client/invoices/src/StockOut';
 
 interface StockLineTableProps {
   stocklines: DraftStockOutLine[];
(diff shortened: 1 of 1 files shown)


client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
```diff
@@ -34,32 +36,51 @@ export const StockLineTable = ({
...
   const columns = useColumns<DraftStockOutLine>(
     [
+      {
+        width: 60,
+        key: 'expiry',
+        label: 'label.expiry',
+        accessor: ({ rowData }) => rowData.expiryDate,
+        Cell: ExpiryDateCell,
+        align: ColumnAlign.Right,
+      },
```
CarlosNZ, line 46: Using correct `ExpiryDateCell` component now, which displays expired lines in red.

CarlosNZ: Also, now that the date is right-aligned, it looks better as the first column

client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
```diff
@@ -34,32 +36,51 @@ export const StockLineTable = ({
...
         label: 'label.units-issued',
-        accessor: ({ rowData }) => rowData.numberOfPacks,
-        setter: handleUpdateDraft,
-        Cell: PackQuantityCell,
+        accessor: ({ rowData }) =>
+          NumUtils.round(
+            (rowData.numberOfPacks ?? 0) * (rowData.packSize ?? 1),
+            3
+          ),
+        setter: row => {
```
CarlosNZ, line 73: I could avoid using the `as` by keeping the key as `numberOfPacks`, but this would mean temporarily storing the unit count in the `numberOfPacks` field, then updating it by multiplying. But I think it's bad practise to put a value into an incorrectly named field, even if only temporarily. This is a bit hacky, but I think it's less misleading than the alternative.

jmbrunskill: I think its probably better make that a comment in the actual code rather than here, I agree though :)

jmbrunskill (approved):
Change look good to me, havent tested but about to go to QA so I think it's ok :)

client/packages/programs/src/JsonForms/components/Prescription/StockLineTable.tsx
CarlosNZ: Sure: 2f70cf0991

PR #5954, opened 2025-01-08
Move the ServerInfo component to the settings page

### LINKED ISSUE #5953: QR code and server info show when sync modal is displayed
## What went wrong? 😲

On any page with the app bar, open the sync modal. The QR code is shown top right:

Screenshot 2025-01-08 at 4 06 42 PM

## Expected behaviour 🤔
Display on the settings page only

## How to Reproduce 🔨

1. Go to any page except dashboard
2. Click on Sync

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #5953

# 👩🏻‍💻 What does this PR do?
Moves the QR code to the settings page, and only displays on that page not in the AppBarPortal

Screenshot 2025-01-08 at 5 00 24 PM
Screenshot 2025-01-08 at 5 00 35 PM

# 🧪 Testing

- [ ] On any page other than dashboard, view the sync modal: should be no QR
- [ ] On the settings page there should be QR (which still expands when clicked)
- [ ] The server, site and version are shown on the settings page

# 📃 Documentation

  1. The 'scan your QR' info in the cold chain application docs needs updating
  2. The synchronisation page needs to be updated


lache-melvin (approved):
Looks great, thanks Mark!! Side question... how come we show the frontend URL as the server url? wouldn't we want the URL of... the server?

mark-prins:
🙏 

> how come we show the frontend URL as the server url?

Perhaps that's misleading naming (again!) the frontend URL is actually the server URL - but of course without the port and `/graphql` endpoint.
The use-case of the server info an QR here is actually so that the cold chain app can scan the QR and easily configure the API URLs - for that we don't want the graphql endpoint

lache-melvin:
But its showing localhost:3003 - isn't your server at localhost:8000? Or is CC app proxied through the frontend?

mark-prins:
It's a bit not-good when running the local dev server. Under normal usage the site is served by the backend which means:

* https://some-domain.org/ serves the front end
* https://some-domain.org/coldchain/v1/temperature-log (etc) is the cold chain API
* https://some-domain.org/graphql is the graphQL endpoint

in which case `window.location.origin` is correct

PR #5955, opened 2025-01-08
5933 update backend crates 

### LINKED ISSUE #5933: Update backend crates

roxy-dao:
Fixes #5933

# 👩🏻‍💻 What does this PR do?
Updating backend crates

# 🧪 Testing

- [ ] Run app 
- [ ] Make sure nothing's broken

# 📃 Documentation

  1.
  2.


server/server/src/logging.rs
```diff
@@ -16,13 +16,18 @@ pub fn logging_init(settings: Option<LoggingSettings>, level: Option<Level>) {
     let config = match settings.mode {
         LogMode::File => file_logger(&settings)
             .level(log_level.to_string())
+            .expect("Cannot determine log level")
```
jmbrunskill, line 19: I wonder if it's better to default rather than panic?
With a warning message perhaps?

jmbrunskill (changes requested):
I'm a little bit worried about the expects in logging. Can we just default to Info instead?

server/server/src/logging.rs
roxy-dao: I think it's fine since above sets level to Info by default, this little piece of code~

``` rust
  let settings = settings.unwrap_or(LoggingSettings::new(
        LogMode::Console,
        service::settings::Level::Info,
    ));

    let log_level = level.unwrap_or(settings.level.clone());
```

andreievg: I think it's ok, although took me a bit of time to figure out why it's ok (clicking through to the parse_level doesn't help that much tbh) <- for that reason probably worth propagating the error. Although we did have "Unable to initialise logger" expect below, so probably ok. Only concern here is that in windows service if there is a panic before logging is mounted, there is no way to capture it (but it seems to be not possible atm)

We can probably simplify a lot of the logging (or how we use that crate and translate from settings), but I am ok with how it is for now

jmbrunskill (approved):
If @andreievg's happy I'm happy.

I'd still prefer to default to `INFO` just incase, but it sounds like it's fine how it is:)

roxy-dao:
> If @andreievg's happy I'm happy.
> 
> I'd still prefer to default to `INFO` just incase, but it sounds like it's fine how it is:)

It defaults it at the top 🙈

PR #5959, opened 2025-01-08
#5957 Remove duplicate "New item" link

### LINKED ISSUE #5957: 2 New Item Buttons in Prescriptions
## What went wrong? 😲

image
## Expected behaviour 🤔

Only one button

## How to Reproduce 🔨

1. Open a prescription
2. Add a new item, or open an existing line on a prescription
3. See extra button

## Your environment 🌱

- Open mSupply Version:2.5.0-RC1

- Platform:

- Database type:

CarlosNZ:
Fixes #5957

# 👩🏻‍💻 What does this PR do?

I don't think this is a complete solution, more a quick fix to get rid of the duplicate for 2.5 release.

I've made it match the designs (for Prescriptions at least), but there's still an inconsistency in how the "New" link works between Prescriptions and Internal orders. For internal orders, it seems to work more like a "button" in that it just provides a temporary stock selector which is then immediately added to the order (and re-directed), whereas in Prescriptions it remains in local state until explicitly "Saved", so the user stays on the "New" page until the entire item is filled it.

So I think we need to make this behaviour consistent and agree on a standard pattern for this component, but that will require a bit of discussion. We also need to consider how to easily access the "New" link when it's at the bottom of a really long list, as mentioned by @lache-melvin [here](https://github.com/msupply-foundation/open-msupply/issues/5957#issuecomment-2578579614). (Making it sticky doesn't quite look right to me at the moment -- perhaps it could jump to the top of the list once the list gets beyond a certain length? Thoughts, @mariyamsupply ?)

Shall we make a new issue for it?

# 🧪 Testing

Make sure there is only one "New Item" link in the list, and confirm behaviour is as expected for both Prescriptions and Internal Orders (anywhere else @roxy-dao ?)

### DIFF
--- a/client/packages/system/src/Item/Components/ListItems/ListItems.tsx
+++ b/client/packages/system/src/Item/Components/ListItems/ListItems.tsx
@@ -7,8 +7,6 @@ import {
   useTranslation,
   useConfirmationModal,
   Box,
-  FlatButton,
-  PlusCircleIcon,
 } from '@openmsupply-client/common';
 import { ItemRowFragment } from '../../api';
 
@@ -65,19 +63,6 @@ export const ListItems = ({
             enteredLineIds={enteredLineIds}
           />
         </Box>
-        {showNew && (
-          <FlatButton
-            label={t('label.new-item')}
-            onClick={() => navigate(route.addPart('new').build())}
-            startIcon={<PlusCircleIcon />}
-            sx={{
-              position: 'sticky',
-              bottom: 0,
-              padding: '1em',
-              justifyContent: 'flex-start',
-            }}
-          />
-        )}
       </Box>
     </Tooltip>
   );


CarlosNZ:
Thanks @jmbrunskill, I'll wait till the end of the day to merge this in case any one else wants to comment.

mariyamsupply:
Agree, it should be consistent. Happy to have a discussion. 
My idea is that you don't need an instant access to the 'New item' button as you're already in the process of adding one.  There's 'New item' button at the top of the page on the prev page and it's at the bottom of the workflow when you're done with that one. Plus, it's also aligned with navigation to a previous item.
But I'm happy to get user feedback and consider different placement.

mariyamsupply:
Also, maybe I'm missing the context of navigating the items list, would be good to look at Internal orders implementation.

mark-prins:
FYI here is how things look with this PR in place:

Adding an item:
Screenshot 2025-01-09 at 4 30 00 PM

Editing an item:
Screenshot 2025-01-09 at 4 30 17 PM

lache-melvin:
Wondering if at least putting it at the top of the list might be better than the bottom? So when you are navigated to this page you see it?

CarlosNZ:
> Wondering if at least putting it at the top of the list might be better than the bottom? So when you are navigated to this page you see it?

I think @mariyamsupply 's argument was that you would normally get to the "New" page by clicking the "Add New Item" button from the Detail page. And once you've added a "New", you can easily get to another "New" just by pushing the "Next" arrow button from the just-added item. This is how it works in Prescriptions, anyway, which I think is fine. Perhaps it's not so good for the Internal Orders (I haven't really tried it extensively), hence the suggestion for a follow-up issue to standardize this new UI usage.

mariyamsupply:
Yes, my thoughts that you don't need 'New item' button when the Item dropdown is open. Once you chose and item and it's closed you can see the 'New item' button.
Same for Internal orders.
But I hear your point about seeing a page structure when you land on it. Maybe don't auto open the dropdown in this case?

mariyamsupply:
NHLMIS product has the same structure 

3b
Screenshot 2024-09-11 at 16 10 29 1

CarlosNZ:
Okay, I'm going to merge this now -- have created follow-up issue for further discussion to come up with a proper standardised pattern: #5973

PR #5962, opened 2025-01-08
Update picked_date when adding or updating lines in prescriptions

### LINKED ISSUE #4980: Prescription picked_date isn't quite correct
## What went wrong? 😲

When we first add a line to a prescription, the status for the invoice is set to `picked`.

This might not be correct, for example you could start a prescription, add a line, delete it then add line in again the next day.
Since picked date is used for stock movement, it could cause problems with ledger, especially if stock is only available after the date the new line is added.

## Expected behaviour 🤔

Suggested:

Update `picked_datetime` on every line added to a prescription?

## How to Reproduce 🔨

1. Create a prescription
2. Add a line
3. Delete the line
4. Wait a day? Nah any time is ok, but more obvious if left for a long time
5. Add a line to the prescription
6. Check picked datetime

## Your environment 🌱

- Open mSupply Version: 2.3.0

jmbrunskill:
Fixes #4980

# 👩🏻‍💻 What does this PR do?

Automatically updates the picked_datetime on prescriptions to the datetime of the latest new line update.
Note: This shouldn't affect any backdated invoices...

The update will happen on every line we insert in the batch mutation.
So that means if we have 1000 lines added, the invoice will be updated 1000 times.
This isn't how we use the API the frontend but worth considering if we end up with patterns like that.

Adding something to detect if there's been line updates, and then 
We could put in a cut off such as if the picked date is within the last 5 minutes we don't update?
I think we're ok for now though.

# 🧪 Testing

- [ ] Open an old (but not backdated) prescription in picked status
- [ ] Add a new line or update an existing line
- [ ] Check that picked datetime is updated
- [ ] Open or create a backdated prescription
- [ ] Add a new line or update an existing line
- [ ] Check that picked datetime is NOT updated

# 📃 Documentation

  1.
  2.


server/service/src/invoice/prescription/update/update_picked_date.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn update_picked_date(
+    connection: &StorageConnection,
+    invoice: &InvoiceRow,
+) -> Result<(), RepositoryError> {
+    // We only want to update the picked date if the invoice is in the picked status and is a prescription
+    if invoice.status != InvoiceStatus::Picked {
+        return Ok(());
+    }
+
+    if invoice.r#type != InvoiceType::Prescription {
```
jmbrunskill, line 16: Because of this condition, I think this function can live in the prescription update section, even though it's not called from here...

jmbrunskill:
Pls hold off on review, we need to update the picked_date for all outbound shipments...

server/service/src/invoice/update_picked_date.rs
```diff
@@ -0,0 +1,25 @@
...
+    connection: &StorageConnection,
+    invoice: &InvoiceRow,
+) -> Result<(), RepositoryError> {
+    // We only want to update the picked date if the invoice is in the picked status and is a prescription
+    if invoice.status != InvoiceStatus::Picked {
+        return Ok(());
+    }
+
+    // Use the invoice's backdated datetime if it's set, otherwise set the status to now
+    let status_datetime = invoice.backdated_datetime.unwrap_or(Utc::now().naive_utc());
```
andreievg, line 17: I feel like this logic should be in a helper method (just this line would be ok), and used both here and in prescription/update#generate

andreievg (changes requested):
> This isn't how we use the API the frontend

We do have auto allocate lines, which would cause mass mutations.

> We could put in a cut off such as if the picked date is within the last 5 minutes we don't update?

I kind of like this tbh, pretty simple to add, I think it's worth adding (requested changes are for this and for the helper method to account for backedate_datetime when figuring out new picked_datetime, as per comment).

server/service/src/invoice/update_picked_date.rs
jmbrunskill: https://github.com/msupply-foundation/open-msupply/pull/5962/commits/ac4bda7a7ddd57ec836bf5c5373f2f664b91bc95

andreievg (approved):
Looks great @jmbrunskill thanks, I just had one more thing (don't update unless picked datetime was changed)

https://github.com/msupply-foundation/open-msupply/compare/4980-prescription-picked_date-isnt-quite-correct...4980-prescription-picked_date-isnt-quite-correct(Dont-update-if-picked-datetime-not-changed)

PR #5966, opened 2025-01-09
4971 add GitHub action to sort json files

### LINKED ISSUE #4971: Add github action to order translation json files alphabetically

Order translation strings!

Building on this PR: https://github.com/msupply-foundation/open-msupply/pull/4968#discussion_r1785315532

Could also include custom ordering of plural translation strings to put them as: one, few some many, etc (a more intuitive ordering than alphabetical)

fergie-nz:
Fixes #4971

# 👩🏻‍💻 What does this PR do?

Adds github action to sort json files upon pull request

I am not sure about the caching of the jq and if that is an effective method. Maybe need to use an alternative sorting mechanism? I feel there must be something lightweight and effective.

# 🧪 Testing

I've only tested this locally on my own repository as it needs to have access to github secrets to work.

# 📃 Documentation


jmbrunskill (commented):
Looks legit to me!
Could run it on our mac mini actions runner if you wanted to avoid having the install steps and caching. JQ would just have to be installed on the mac mini.

We could probably also sort using a javascript function, and node it probably installed by default but I haven't checked...
This looks fine to me though, but haven't actually tested it

fergie-nz:
> Looks legit to me! Could run it on our mac mini actions runner if you wanted to avoid having the install steps and caching. JQ would just have to be installed on the mac mini.
> 
> We could probably also sort using a javascript function, and node it probably installed by default but I haven't checked... This looks fine to me though, but haven't actually tested it

Both those options sound great! I saw that using a custom runner and/or something with preinstalled package would be better, but wasn't sure how so tried this MVP.

I'll see if I can run on minimac with either jq or node

andreievg:
@fergie-nz is this PR ready for review ? Also was wondering, if it's better to do as a commit hook, similar to how we do pre commit compilation ?

fergie-nz:
> @fergie-nz is this PR ready for review ? Also was wondering, if it's better to do as a commit hook, similar to how we do pre commit compilation ?

Yes can do.

Though I already find the commit hook a little bit frustrating if it takes some time to check things over and have a mild preference to reduce load on that.

What do you think?

jmbrunskill:
> > @fergie-nz is this PR ready for review ? Also was wondering, if it's better to do as a commit hook, similar to how we do pre commit compilation ?
> 
> Yes can do.
> 
> Though I already find the commit hook a little bit frustrating if it takes some time to check things over and have a mild preference to reduce load on that.
> 
> What do you think?

I like it being a github action. We're supposed to be moving the frontend compile check to be an action too

fergie-nz:
To revisit during next cooldown period

.github/workflows/sort-locale-json.yaml
```diff
@@ -0,0 +1,50 @@
...
+permissions:
+  contents: write
+
+jobs:
+  sort-json:
+    name: Sort Json
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+      - uses: actions/checkout@v3
```
mark-prins, line 17: could use @v4 for this
might need to add
```
        with:
          ref: ${{ github.ref }}
```

.github/workflows/sort-locale-json.yaml
```diff
@@ -0,0 +1,50 @@
...
+          if [ "$(git status --porcelain)" ]; then
+            echo "Changes detected, committing..."
+            git config --global user.name "github-actions[bot]"
+            git config --global user.email "github-actions[bot]@users.noreply.github.com"
+            git add .
+            git commit -m "chore: sorted JSON files"
+            git push https://x-access-token:${{ secrets.PAT }}@github.com/${{ github.repository }} HEAD:${{ github.head_ref }}
+          else
+            echo 'No changes to commit.'
+          fi
```
mark-prins, line 50: an alternative to this block is simply:

```
      - uses: stefanzweifel/git-auto-commit-action@v5
        with:
          commit_message: 'Sorted JSON translation files'
          commit_options: '--no-verify'
```

when I was playing around I did a `git add $file` in the loop - you can try and see if that's needed or no

mark-prins (commented):
I think it's great! You should get it merged in 🙏

PR #5971, opened 2025-01-09
5368 Fix ledger with migrated cancelled invoices

### LINKED ISSUE #5368: Open mSupply Views don't account for negative stock correctly
## What went wrong? 😲

Discovered while investigating this issue...
https://github.com/msupply-foundation/open-msupply/issues/5359

If an invoice is cancelled it creates a duplicate invoice with negative values in it.

image

However, our stock movements view filters out these canceled invoices, but still includes the uncancelled version.

image

This means our ledger is incorrect, and results in miss-calculations when backdating stock

## Expected behaviour 🤔

We either need to include cancelled invoice movements in the stock movements OR
filter out both the original and the cancellation (treating it as though the stock movement never happened)

I think the first one sounds easier but might have unexpected effects on `item_stats`?

## How to Reproduce 🔨

1. Create an invoice in mSupply, cancel that invoice, then migrate the store to OMS
2. Check Ledger in open-mSupply (See the cancelation is missing from ledge)
3. This might also allow you to back date a prescription to before you first introduced stock as it's using the same data source

## Your environment 🌱

- Open mSupply Version:2.3.2

- Platform:

- Database type:

andreievg:
Fixes #5368

# 👩🏻‍💻 What does this PR do?

During translation if invoice line has negative number of packs convert to StockIn type and make sure negative numbers are turned to positive

I didn't want to have negative invoice lines quantities, so changed the type and abs_ed() the numbers

Technnically speaking we woudl also want to remove cancelled invoice from any AMC calculations, I feel like cancellation is rare enough to not consider it for now

# 🧪 Testing

Use data file provider or migrate a store to omsupply (a store that had cancelled invoice), check catalogue/item/ledger make sure ledger balance matches correctly and ledger looks good.

[simple sync site.4DD.zip](https://github.com/user-attachments/files/18356021/simple.sync.site.4DD.zip)

 `user 1 -> user 1` to configure in OG (register, hadrwareid for site etc..)(
 `test -> pass` for everything else

# 📃 Documentation

  1.
  2.


jmbrunskill (approved):
Looks good to me @andreievg

PR #5975, opened 2025-01-09
5620 Add index to sync buffer

### LINKED ISSUE #5620: Performance issue with sync integration with large data file
## What went wrong? 😲

Integration quite slow on large data files

## Expected behaviour 🤔

Integration is quick

## Suggested fix

I did some investigate, and noticed that query for sync buffer was taking awhile: 

https://github.com/msupply-foundation/open-msupply/blob/d9240a8f6ea18e44178f0e4de81c94cf467e4123/server/service/src/sync/sync_buffer.rs#L63-L70

Even though individual index exists it was taking some 500ms or so to query for one table, and since we do a query for each table it was adding up, below table fixed the performance issue

CREATE INDEX "index_sync_buffer_action_integration_datetime" ON "sync_buffer" ("action",integration_datetime)

## How to Reproduce 🔨

Use data files here: 

https://drive.google.com/drive/u/1/folders/1rbscWr9_tWs817Nmdd7gP9soE0E-xSU1

Add some configuration (new item or touch a master list) on central mSupply server, sync, see integration step taking awhile

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

andreievg:
Fixes #5620

# 👩🏻‍💻 What does this PR do?

Add index to sync buffer

Went with suggestion by james, seems fast

# 🧪 Testing

Test with data file from the issue (for both omsup and msup), try manual sync and see integration speed being slow, now try with 2.6, it should be fast

# 📃 Documentation

  1.
  2.


jmbrunskill (approved):
This is slightly different to my suggestion, in that I didn't have `integration_error`
Am a little bit worried that could inflate the index size? Not a major for now...

andreievg:
> This is slightly different to my suggestion, in that I didn't have integration_error
> Am a little bit worried that could inflate the index size?

You so right ! Thanks for review, i remove integration_error from the index https://github.com/msupply-foundation/open-msupply/pull/5975/commits/83e33623fb3407579b091d4640f5ec147d73d715

Issue #5577, opened 2024-11-26
Create a component for rendering the temperature range used in iSC/Cold Chain/GAPS


aimee-mcneil-melville:
## The suggestion

Create a single component for rendering of the Cold Chain type/temperature selection.

We currently have this defined in each of the three areas it is currently used in. 
- coldStorageTypeInput component.
- location list view.
- item variant tab (detail view).

### Why should we invest time in this?

Having these defined as a component will:
- ensure that any changes to the rendering are consistent across all areas.
- if any change is needed to be made, it will only need to be made in one place.
- save time finding the areas and possibly needing multiple issues/PR's if not all areas are updated at once

### Are there any risks associated with this change?

This change is low risk, involving only frontend code and will not create any user-facing changes.

### How much effort is required?

This should be a small change, I estimate 1 - 2 hours to implement

### Agreed Solution

Create an option renderer for the temperature change, which can then be imported into in the list view, detail view and the coldStorageType input

lache-melvin:
Closing as resolved by #5983

alainsussol:
Will be tested through #5814

PR #5983, opened 2025-01-09
create cold storage translation label

### LINKED ISSUE #5814: Cold storage range not translated
## What went wrong? 😲

Hardcoded `to` rather than translation for range:

https://github.com/msupply-foundation/open-msupply/blob/d07273e6e6e59a16eee84f23ae402341213888c1/client/packages/system/src/Location/ListView/ListView.tsx#L48

This is currently in 3 places, so could be fixed as part of #5577

Not sure if we just translate "to" or if ranges are presented differently in different languages, maybe we need to pass the two values into the translation string to be handled appropriately??

## Expected behaviour 🤔

If I switch language, the `to` translates appropriately

## How to Reproduce 🔨

1. Go to Inventory > Locations
2. Have at least one location with a temp range set
3. Change to a language other than english
4. See temp range is still in english

## Your environment 🌱

- Open mSupply Version: v2.4.2

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #5814 

# 👩🏻‍💻 What does this PR do?

Allows for translation of the cold storage type/temperature range

Changed from a hardcoded string to a translation string, which inputs the values for 'type'(fridge, freezer etc) as well as min and max temperatures

This allows for translation of the entire string, so that languages that have a different sentence structure can be translated appropriately - quick google translate suggests that this is the case for Arabic and Russian

Is #5577 still needed in this case? As the string format is being handled by translation, it is now easier to:
1: Change the content of the string in the translation file
2: Find where the translation is being used and update data passed in if needed

# 🧪 Testing

- [ ] Go to Inventory -> Locations
- [ ] Add a new location, select a Cold storage Type - see the temperature range here. Add the item
- [ ] See the temperature range in the Locations List
- [ ] This can also be seen in Catalogue -> Items -> Variants tab (may need to add a new variant)

No translations yet, it will all be English!

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
Looks great thank you!!

> Is https://github.com/msupply-foundation/open-msupply/issues/5577 still needed in this case?

And no! Great point, will close 🙏

PR #5988, opened 2025-01-09
5987 fix tests

### LINKED ISSUE #5987: Fix tests
## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #5987

# 👩🏻‍💻 What does this PR do?
Don't compare `None` elmis code

# 🧪 Testing

- [ ] Tests pass

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/service/src/requisition/program_settings/customer_program_settings/map.rs
+++ b/server/service/src/requisition/program_settings/customer_program_settings/map.rs
@@ -19,18 +19,8 @@ pub(super) fn map_customer_program_settings(
                 .iter()
                 .filter(|(customer, _)| {
                     customer.program.id == program_setting.program_row.id
-                        || customer
-                            .program
-                            .elmis_code
-                            .as_ref()
-                            .map_or(false, |elmis_code| {
-                                elmis_code
-                                    == program_setting
-                                        .program_row
-                                        .elmis_code
-                                        .as_ref()
-                                        .unwrap_or(&"".to_string())
-                            })
+                        || (customer.program.elmis_code == program_setting.program_row.elmis_code
+                            && customer.program.elmis_code.is_some())
                 })
                 .map(|(customer, requisitions_in_periods)| {
                     // Filter available periods for program settings for the customer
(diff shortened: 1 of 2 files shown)


server/service/src/requisition/program_settings/customer_program_settings/map.rs
```diff
@@ -19,7 +19,18 @@ pub(super) fn map_customer_program_settings(
...
+                            .elmis_code
+                            .as_ref()
+                            .map_or(false, |elmis_code| {
+                                elmis_code
+                                    == program_setting
+                                        .program_row
+                                        .elmis_code
+                                        .as_ref()
+                                        .unwrap_or(&"".to_string())
+                            })
```
jmbrunskill, line 33:
```suggestion
                        || (customer.program.elmis_code == program_setting.program_row.elmis_code && !customer.program.elmis_code.is_none())
```

jmbrunskill: Can't we just do this?

Was trying to get my head around the filter change as if wasn't entirely clear to me what it was doing.
This seems much clearer and simpler?
Maybe I'm missing something but tests are passing now...

If we need the more complicated filter that's fine but would be nice to have a comment explaining it as it's not immediately obvious, at least to me :)

jmbrunskill (approved):
Tests pass which is good.
I don't have enough context to fully understand what un-intended consequences there might be but feels safe enough.

I've made a suggestion to simplify the filter, hopefully that's correct and you agree it's simpler. Otherwise would like a comment and/or a function call to wrap the functionality with a useful name to explain the intention of the filter.

server/service/src/requisition/program_settings/customer_program_settings/map.rs
roxy-dao: That also looks fine to me! Will test in a bit

Sometimes stores might have different programs that are essentially the same (this is very CIV specific), CIV uses elmis code to match master lists

Issue #5043, opened 2024-10-08
Dynamic sync batch size


Chris-Petty:
## Is your feature request related to a problem? Please describe 👀

As per [KDD](https://github.com/msupply-foundation/open-msupply/blob/develop/decisions/2024-08-30_central_sync_queue_pull_reduction.md), we can speed up sync a bunch by just increasing the batch size. Further it may be nice to do this dynamically as that's proved quite nice in the past for OG mobile. Fast when connection is good, takes it easy when connection is bad. 

## Describe the solution you'd like 🎁

Approximately exponentially increase/decrease sync batch size based on a round trip size. 

### Additional context 💌

Mobile's implementation targetting 5s batch round trip and integration time.
https://github.com/msupply-foundation/mobile/blob/38602080d644736191062a02c024a1c7e804678c/src/sync/Synchroniser.js#L461-L480

andreievg:
Using this issue as place to record carry over from another sync api performance related issue:https://github.com/msupply-foundation/open-msupply/pull/5998#issuecomment-2603445169

* Smarter way to do timeout for connection retry (and do we capture failed connection attempts in `with_retries` method ?) 
* Connection type errors should be displayed to the user in readable format

PR #5998, opened 2025-01-10
3813 Retry api on connection error or timeout

### LINKED ISSUE #3813: Sync error on remote site - 'EOF while parsing'
## What went wrong? 😲

After doing any transaction on a remote site(android), I press 'sync now' and get the following error message:

Screenshot 2024-05-07 at 3 35 24 PM

the problem is resolved if I press the sync button for the 2nd time.

## Expected behaviour 🤔

No need to press sync twice to get it to work.

## How to Reproduce 🔨

1. Go to 'remote site (android)
2. Do any transactions (internal order, outbound shipment, etc..)
3. Go to sync and press the button 'sync now'

## Your environment 🌱

- Open mSupply Version: V2.0 RC3

- Platform: android

andreievg:
Fixes #3813

# 👩🏻‍💻 What does this PR do?

While I was doing a benchmark I notice error that were very similar:

Screenshot 2025-01-10 at 5 13 09 PM

[this branch](https://github.com/msupply-foundation/open-msupply/tree/Bench-sever-initial) and this diff [commit](https://github.com/msupply-foundation/open-msupply/compare/develop...Bench-sever-initial)

This was on a benchmark branch and pretty large data file (see quick [README.md](https://github.com/msupply-foundation/open-msupply/blob/a92d6e8afe52eda533065dcb2cf57895ad83b4d6/bench-server/README.md#L3) note)

Having a quick look at[ the code](https://github.com/msupply-foundation/open-msupply/blob/a92d6e8afe52eda533065dcb2cf57895ad83b4d6/server/service/src/sync/api_v6/mod.rs#L229) it seems we are not actually checking status (which is 408) when there is a connection timeout.

I remember doing a retry and connection timeout to try and overcome some difficulties we have with syncing as per [this comment](https://github.com/msupply-foundation/open-msupply/issues/4658#issuecomment-2309107827).

After I added this to the benching, this issue disappeared: https://github.com/msupply-foundation/open-msupply/compare/Bench-sever-initial...Bench-server

You can see below that there are some retires show in log:

Screenshot 2025-01-10 at 4 39 01 PM

Applied the same to our sync client API 

Is 2 second too little and actually will compromise existing low bandwidth clients (never enough time to connect ?)

# 🧪 Testing

Make sure sync still work for central and remote (and especially when internet to cloud server is slow).

Try bench branch with mentioned data file. Both with retries and without retires changes


jmbrunskill:
> Is 2 second too little and actually will compromise existing low bandwidth clients (never enough time to connect ?)

I think it probably is a bit too short. I'm not sure how we can defined a worse case though. Maybe 10s would be a more realistic worse case? Will delay how long it takes to re-try on this error though.
I wonder if we do an exponential back off where we try waiting a bit longer each time on a re-try?

andreievg:
@jmbrunskill I agree, updated with RetrySeconds Vec rather then number of retries and connection timeout.

server/util/src/api_helper.rs
```diff
@@ -0,0 +1,42 @@
...
+pub async fn with_retries<F>(connection_timeouts: RetrySeconds, f: F) -> Result<Response>
+where
+    F: Fn(Client) -> RequestBuilder,
+{
+    let mut index = 0;
+    loop {
+        let client = Client::builder()
+            .connect_timeout(Duration::from_secs(connection_timeouts.0[index]))
+            .build()
+            .unwrap();
```
jmbrunskill, line 24:
```suggestion
            .unwrap();// This method fails if a TLS backend cannot be initialized, or the resolver cannot load the system configuration.
```

Might nice to put a comment on the unwrap? I think it's safe enough though.

server/util/src/api_helper.rs
```diff
@@ -0,0 +1,42 @@
...
+        let result = f(client).send().await;
+
+        let (status, is_connect_error) = match result.as_ref() {
+            Ok(r) => (Some(r.status().clone()), false),
+            Err(e) => (e.status().clone(), e.is_connect()),
+        };
+
+        if (is_connect_error || status == Some(StatusCode::REQUEST_TIMEOUT))
+            && (index + 1) < connection_timeouts.0.len()
+        {
```
jmbrunskill, line 35:
```suggestion
        {
            log::info!(
                "Connection timeout, retrying in {} seconds",
                connection_timeouts.0[index]
            );
```

Not really needed but like the idea of logging things like this that might help support figure out an issue.
What do you think about logging this or similar?

For this you'd need to include `log` in the `Cargo.toml` too

jmbrunskill (approved):
Thanks Andrei, looks good, the retry vec seems good!

One issue I thought of is that some site connections might always fail after 2 seconds, so for them the connection would take 7seconds instead of say 5 before. I don't know how often that would be a problem but might be worth a follow up issue to look for a nice solution, like a configuration parameter for the retries?

One other issue is that I see is that we now get Deadline Exceeded as the error message in the sync error, would be nice if that was more explicitly a `Timeout` even better something human readable such as `The server didn't respond in time`?

image

I'm happy to merge how it is but wanted to raise those thoughts for follow up...

server/util/src/api_helper.rs
andreievg: Agree, thanks

server/util/src/api_helper.rs
andreievg: > What do you think about logging this or similar?

I think it's good to have logs, but I think this log will just be noise, I can't see how we can action on it? If there is actually a problem connection user should see this in UI (and pass along to support) ?

andreievg:
Thanks for review @jmbrunskill,

> One issue I thought of is that some site connections might always fail after 2 seconds, so for them the connection would take 7seconds instead of say 5 before. I don't know how often that would be a problem but might be worth a follow up issue to look for a nice solution, like a configuration parameter for the retries?

Agree added to this issue: https://github.com/msupply-foundation/open-msupply/issues/5043

> One other issue is that I see is that we now get Deadline has ellapsed as the error message in the sync error, would be nice if that was more explicitly a Timeout even better something human readable such as The server didn't respond in time?

Currently connection errors are not explained to the user, just says there is an error. I got 'connection refused', 'host error' on top of the error you mentioned (although I haven't been able to replicate your error). I feel like we can do better with error coms to user, but doesn't feel like big priority atm, still added to above issue

PR #6010, opened 2025-01-13
Update cost per unit when items not grouped

### LINKED ISSUE #4591: Group by item hides batch even if only one line present

noel-yeldos:
Fixes #4591 

# 👩🏻‍💻 What does this PR do?
Fix issue where cost per unit was not being calculated correctly when items were not grouped. 

Screenshot 2025-01-13 at 4 35 09 PM
Screenshot 2025-01-13 at 4 36 02 PM
Screenshot 2025-01-13 at 4 44 08 PM
Screenshot 2025-01-13 at 4 37 04 PM

# 🧪 Testing

- [ ] Open inbound shipment
- [ ] Add item 
- [ ] Add a batch and update  pack size (e.g. 3, pack quantity (e.g. 1), and pack cost price (e.g $6).
- [ ] Go back to detail view and toggle Group by OFF
- [ ] Cost per unit should display $2.00 (e.g total cost (6 * 1) divided by pack size (3) = $2 cost per unit)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.ts
+++ b/client/packages/invoices/src/InboundShipment/DetailView/ContentArea/columns.ts
@@ -38,7 +38,7 @@ export const useInboundShipmentColumns = () => {
     queryParams: { sortBy },
   } = useUrlQueryParams({ initialSort: { key: 'itemName', dir: 'asc' } });
   const getCostPrice = (row: InboundLineFragment) =>
-    isInboundPlaceholderRow(row) ? 0 : row.costPricePerPack;
+    isInboundPlaceholderRow(row) ? 0 : row.costPricePerPack / row.packSize;
   const { getColumnPropertyAsString, getColumnProperty } = useColumnUtils();
 
   const columns = useColumns<InboundLineFragment | InboundItem>(


lache-melvin (approved):
Love it thank you!

PR #6024, opened 2025-01-13
#5977 Ability to clear clinician name

### LINKED ISSUE #5977: Prescription: Unable to remove Clinician's name
## What went wrong? 😲

When creating prescriptions, the clinician's name can be changed but cannot be removed.
Here's video:

https://github.com/user-attachments/assets/49a4f173-b04a-4fb6-ae2a-93612ac9e98c

## How to Reproduce 🔨

2. Make a prescription and add Clinician
3. Try to remove it
4. See error (can't remove completely)

## Your environment 🌱

- Open mSupply Version:`V2.5.00-RC1`
- Legacy mSupply Central Server Version: `V8.0-dev32`
- 
- Platform:

- Database type:

CarlosNZ:
Fixes #5977

# 👩🏻‍💻 What does this PR do?

Make the "clinicianID" field nullable on both back and front-end, and correctly passes `null` value when input is cleared.

Have also added a `useState` value for "clinician" so that it can be updated and reflected in the UI immediately, otherwise you get a delay for the round-trip to the server before the value is updated in the UI.

Also, added the `isDisabled` condition to the input to match the other elements in the Toolbar

# 🧪 Testing

- [ ] Go to an unverified prescription -> detail page
- [ ] Can change the Clinician
- [ ] Can clear the Clinician
- [ ] Changes are reflected immediately in the UI (no network delay)
- [ ] Changes persist on page refresh
- [ ] On a verified prescription, clinician field is not editable


jmbrunskill (approved):
Looks good to me! Thanks Carl!

PR #6025, opened 2025-01-13
5079 remove raw graphql error from snacks

### LINKED ISSUE #5079: Remove raw GraphQL error detail from error snacks
## What went wrong? 😲

I am not exactly sure how this happened but was a nasty one. But what i do remember is, i selected the past expiry date.
Below is the error.
Screenshot 2024-10-10 at 21 56 11

## How to Reproduce 🔨

Not sure

## Your environment 🌱

- Open mSupply Version: 2.2.01
- Legacy mSupply Central Server Version: Not sure
- Platform: Desktop, Google Chrome
- Database type: Not sure

GeronimoJohn:
Fixes #5079 

# 👩🏻‍💻 What does this PR do?

Created `errorWithDetail` Snackbar within `useNotification`. It displays an error Snackbar with a general message and an info icon. Clicking the info icon opens a dialog containing a detailed error message.

Our users sometimes receive irrelevant information, such as raw GraphQL errors. While these errors are helpful for developers, we need to improve the user-facing error messages.

Screenshot 2025-01-14 at 11 29 01 AM

Screenshot 2025-01-14 at 11 29 10 AM

To manually test it, I've updated one of GraphQL queries to panic (`panic!()`) in order to see an error.

# 🧪 Testing

- [ ] Refer to reviewer notes
- [ ] Visit the page where we've temporarily updated a query or a mutation to fail
- [ ] Snackbar error should show up with message `Click info icon to see error`
- [ ] Clicking on the icon opens a dialog which returns a detailed explanation of the error

# 📃 Documentation

  1.
  2.


client/packages/common/src/hooks/useNotification/useNotification.tsx
```diff
@@ -24,18 +29,47 @@ export const useNotification = (): NotificationHook => {
...
+
   const error = (message: string, options?: OptionsObject) => () =>
     enqueueSnackbar(message, {
       variant: 'error',
       action,
       ...options,
     });
+
+  const errorWithDetail = (message: string, options?: OptionsObject) => () =>
+    enqueueSnackbar(`${t('label.error-info-icon')}:`, {
```
lache-melvin, line 59: Wondering if this might be good with a `error.something-wrong` first? 

 Then would read something like "Oops! Something's gone wrong. Click info icon to see error details: "

lache-melvin (commented):
Hm - I'm thinking might be best to put this one aside until @mark-prins is back, maybe we can build a test build for him and Mariya to have a play with?

I really like the concept, but I think we need to refine it a bit:
- Not sure about the toast persisting until you dismiss it - maybe the modal was a good idea after all? Or if there is some other way for it to stick around while the paper popover is open, then dismiss 🤷‍♀️ 
- We have a lot of cases like this:
Screenshot 2025-01-14 at 4 56 32 PM

It's a separate issue, the double reporting, but it feels more apparent this way. I wonder if we can do better filtering of the errors at the query error handler?

Maybe we only do the error with detail if the error exceeds a certain length?

mark-prins:
Good points @lache-melvin! thanks for pausing
I agree - really like the concept and nicely done! 
there's a few aspects to tweak..

* the toast persistence is a problem; you can navigate to other pages and you have an error showing which is now unrelated to the page. this is compounded by the fact that the toast is shown again if you come back to the problem page.. but the original is still there too!

image

* as noted: the usability of hiding the short errors behind an extra mouse click is a little annoying too
* :nit the `:` at the end of the message bothered me.. I would just say `Click info icon for details` and leave it at that
* there's a double colon too - it's both in the errorWithDetail method and the translation

..so my suggestion is this:

* auto hide the `errorWithDetail` toast
* increase the auto hide timeout for the error toast
* only show the `errorWithDetail` if the error message is longer than x characters


I went with 10s (twice the default) and that gives enough time to click the info icon, but tidies itself out of the way and reduces duplicates etc.

```
  const errorWithDetail = (message: string, options?: OptionsObject) => () =>
    enqueueSnackbar(t('error.something-wrong-info-icon'), {
      variant: 'error',
      action: key => actionWithDetail(key, message),
      autoHideDuration: 10000,
      ...options,
    });
```

mark-prins:
this is how it looks


https://github.com/user-attachments/assets/c63ddf62-702d-411d-944a-53ae2b0322a0

lache-melvin (changes requested):
Marking as requesting changes outlined by Mark above 🙏

mark-prins:
thanks @GeronimoJohn! is it possible to also add the `only show the errorWithDetail if the error message is longer than x characters` request?  less essential but would make it nicer 😃

GeronimoJohn:
Sorry missed that bit! Thanks for that @mark-prins. Can see the changes here: [c244516](https://github.com/msupply-foundation/open-msupply/pull/6025/commits/c2445165b595aa38896f238f34f345ea7c9ca868).

I've set it to only show if over a 100 for now. No reason for it. Just thought its a good start. Can change to what we think might be best

PR #6042, opened 2025-01-14
5184 use correct formula to get days in month

### LINKED ISSUE #5184: Are we translating requisition_line.average_monthly_consumption correctly?
### Month length is 30 rather than 365/12

https://github.com/msupply-foundation/open-msupply/blob/b654ba6800c8b5c2f1f68d2e255fe88443dade54/server/util/src/constants.rs#L7-L8
In OMS we're doing 30 for the length of a month.

https://github.com/msupply-foundation/mobile/blob/38602080d644736191062a02c024a1c7e804678c/src/database/utilities/constants.js#L16
In Mobile after much debate the consensus was to use `365/12` religiously.

OG _mostly_ uses [`365.25/12`](https://github.com/msupply-foundation/msupply/blob/develop/Project/Sources/Methods/day_month_convert_ratio.4dm). IDK why mobile opted for 365 😆.

### We store as average monthly consumption rather than (average )daily consumption

https://github.com/msupply-foundation/open-msupply/blob/ecba9246933555cfd800b15270eb6536cdde7707/server/service/src/sync/translations/requisition_line.rs#L111

On pull we translate we convert from daily to monthly and `ceil()` the value. My concern here is that at the translation layer we're applying a different month length constant and rounding saving it to the record - this could break the math of any of the other fields that were calculated based on the daily_consumption and a different length of month right? So possibly subtly breaking past data. This would bleed back to the central server if the records were pushed again.

Why do we take the `ceil()`? I'm afraid again of how this may force a facet into all calculations that clients may not be expecting (say in future plugin based suggestions).

Why didn't we just use `average_daily_consumption`? I'm guessing probably because OMS frontend found it was only showing AMC (for now 😉), so thought to just make the schema match?

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #5184

# 👩🏻‍💻 What does this PR do?
Use 365.25/12.0 to get correct days in month

# 🧪 Testing

- [ ] Run tests

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -183,7 +183,7 @@
   "description.rnr-amc": "The Average Monthly Consumption is calculated using the Adjusted consumption value, using this R&R form and the previous two, if available",
   "description.rnr-approved-quantity": "The quantity approved to be ordered. Populated after the R&R form is finalised and authorized.",
   "description.rnr-consumed": "The quantity of this item distributed during this period (via Outbound Shipments or Prescriptions)",
-  "description.rnr-consumed-adjusted": "Shows the consumption adjusted to account for stock out days. =(Consumed x 30)/(30 - Stock out duration)",
+  "description.rnr-consumed-adjusted": "Shows the consumption adjusted to account for stock out days. =(Consumed x (365.25/12))/((365.25/12) - Stock out duration)",
```
roxy-dao, line 186: I wasn't sure what this 30 represented, assuming it's number of days in month but please correct me if I'm wrong

andreievg: This should be:

Shows the consumption adjusted to account for stock out days. = (Consumed x (Days in period  / (Days in period - Stock out duration))

andreievg (approved):
Can you please revert `description.rnr-consumed-adjusted` to what it was and we would need to make a note to change that one to what I suggested. 

Thanks @roxy-dao

client/packages/common/src/intl/locales/en/common.json
roxy-dao: Should I just change it to this then? or do you want me to add TODO like the approved message?

server/util/src/constants.rs
```diff
@@ -5,7 +5,7 @@ pub const INVENTORY_ADJUSTMENT_NAME_CODE: &str = "invad";
 /// Code for the special repack name
 pub const REPACK_NAME_CODE: &str = "repack";
 /// Number of days in a month (used in AMC calculation)
-pub const NUMBER_OF_DAYS_IN_A_MONTH: f64 = 30.0;
+pub const NUMBER_OF_DAYS_IN_A_MONTH: f64 = 365.25 / 12.0;
```
CarlosNZ, line 8: This constant is already defined in this file as:

```rs
/// Use this value for accurate average month length (i.e. when considered over
/// a long time period)
pub const DAYS_PER_MONTH: f64 = 365.25 / 12.0;
```

Can we combine them now, since they're the same value now?

roxy-dao: [b70b580](https://github.com/msupply-foundation/open-msupply/pull/6042/commits/b70b5808cd79a9affb56f579ec35f0f9eb4f742d)

PR #6054, opened 2025-01-14
4015 Allow filtering "View Stock Page" by master list

### LINKED ISSUE #4015: Allow filtering "View Stock Page" by master list

Came out of a demo - I think this is a nice to have that might be a quick win. 
It would be good to be able to filter the stock page by a masterlist, as for some countries they organise stock (and who manages that stock) by masterlist. 

New filter on view stock page, that filters by masterlist name and add master list name to the stock page columns

roxy-dao:
Fixes #4015

# 👩🏻‍💻 What does this PR do?
Display master list names in stock view and allow the user to filter by master list

nn

Screenshot 2025-01-15 at 8 16 41 AM

# 🧪 Testing

- [ ] Go to stock view
- [ ] See master list name listed
- [ ] Click filter -> filter by master list name
- [ ] Start filtering

# 📃 Documentation

  1. Stock screenshots
  2.

### DIFF
--- a/client/packages/system/src/Stock/ListView/ListView.tsx
+++ b/client/packages/system/src/Stock/ListView/ListView.tsx
@@ -21,6 +21,7 @@ import { AppBarButtons } from './AppBarButtons';
 import { Toolbar } from './Toolbar';
 import { AppRoute } from '@openmsupply-client/config';
 import { useStockList } from '../api/hooks/useStockList';
+import { ChipTableCell } from '../../Patient';
 
 const StockListComponent: FC = () => {
   const {
@@ -39,11 +40,14 @@ const StockListComponent: FC = () => {
         key: 'expiryDate',
         condition: 'between',
       },
+      {
+        key: 'masterList.name',
+      },
     ],
   });
   const navigate = useNavigate();
   const queryParams = {
-    filterBy,
+    filterBy: filterBy ?? undefined,
     offset,
     sortBy,
     first,
@@ -71,7 +75,13 @@ const StockListComponent: FC = () => {
       Cell: TooltipTextCell,
       width: 350,
     },
-    // TODO:: Add a column for the master list name
+    {
+      key: 'masterList',
+      label: 'label.master-list',
+      Cell: ChipTableCell,
+      width: 150,
+      accessor: ({ rowData }) => rowData.masterList.map(m => m.name),
+    },
     { key: 'batch', label: 'label.batch', Cell: TooltipTextCell, width: 100 },
     {
       key: 'expiryDate',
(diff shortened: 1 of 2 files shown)


client/packages/system/src/Stock/api/hooks/useStockList.ts
```diff
@@ -13,7 +13,7 @@ export type ListParams = {
   first?: number;
   offset?: number;
   sortBy?: SortBy<StockLineRowFragment>;
-  filterBy?: FilterByWithBoolean | null;
+  filterBy?: StockLineFilterInput;
```
roxy-dao, line 16: Had to change this so can spread filter by master list...

client/packages/system/src/Stock/ListView/ListView.tsx
```diff
@@ -71,7 +75,13 @@ const StockListComponent: FC = () => {
       Cell: TooltipTextCell,
       width: 350,
     },
-    // TODO:: Add a column for the master list name
+    {
+      key: 'masterList',
+      label: 'label.master-list',
+      Cell: ChipTableCell,
```
lache-melvin, line 81: Screenshot 2025-01-20 at 4 06 26 PM

I think if we're going to show them like this, we should hide master list column by default. Item could be on a LOT of master lists....

lache-melvin (changes requested):
Filter working well, just concerned about display 🙏

client/packages/system/src/Stock/ListView/ListView.tsx
roxy-dao: Yeah... but Mark wanted to show master list column since we're filtering by it now... the same thing will happen to program names too in patient list if programs is configured... I think we should just show default number of chips (will need issue) and show ellipsis or a more > button.. will need design

mariyamsupply:
I tried to mockup a few versions, but they don't look ideal and feels like more work at this point. 
**I think it'd be better to hide that column for now.**

masterlists

client/packages/system/src/Stock/ListView/ListView.tsx
```diff
@@ -71,7 +74,14 @@ const StockListComponent: FC = () => {
       Cell: TooltipTextCell,
       width: 350,
     },
-    // TODO:: Add a column for the master list name
+    // TODO: Add back when design has been decided
```
lache-melvin, line 77: Aw I was hoping we could use the `columnDisplayState` but looks like we only support the user hiding columns, rather than having some columns not show by default 🥲

lache-melvin (approved):
LGTM... definitely odd to have a filter for a column you can't see, but we can leave that with design friends 😁 

Do we need to make a separate issue for the needs design piece if we are going to close this one?

roxy-dao:
> LGTM... definitely odd to have a filter for a column you can't see, but we can leave that with design friends 😁
> 
> Do we need to make a separate issue for the needs design piece if we are going to close this one?

Will make soon

PR #6057, opened 2025-01-14
5194 remove diesel dls alias

### LINKED ISSUE #5194: Refactor to remove redundant aliasing of diesel DSLs
## The suggestion

Throughout our codebase while copying swathes of boilerplate we've aliased all of the diesel DSLs for each table like so:
```rust
db_diesel::{
  master_list_name_join::master_list_name_join::dsl as master_list_name_join_dsl,
  master_list_row::master_list::dsl as master_list_dsl,
  name_link_row::name_link::dsl as name_link_dsl, name_row::name::dsl as name_dsl,
  name_store_join::name_store_join::dsl as name_store_join_dsl,
  program_row::program::dsl as program_dsl, store_row::store::dsl as store_dsl,
}  
```

Diesel's [getting started page](https://diesel.rs/guides/getting-started) has this to say:

>The use self::schema::posts::dsl::* line imports a bunch of aliases so that we can say posts instead of posts::table, and published instead of posts::published. It’s useful when we’re only dealing with a single table, but that’s not always what we want. It’s always important to keep imports to schema::table::dsl::* inside of the current function to prevent polluting the module namespace.

My TL;DR: Use DSL for convenience if just working with one table in a module.

So if we don't do that, it becomes a lot less verbose to do imports:

```rust
db_diesel::{
  master_list_name_join::master_list_name_join, master_list_row::master_list,
  name_link_row::name_link, name_row::name, name_store_join::name_store_join,
  program_row::program, store_row::store,
}
```

And if you check out the code in this [commit for example](https://github.com/msupply-foundation/open-msupply/pull/5183/commits/eea9ec19ebcbb32c523b4514e4dc820a764afbec), it's a lot less verbose having shorter variable names throughout our diesel code and I think more readable having `master_list::table` rather than `master_list_dsl::master_list`

### Ironically, we don't like using DSL as intended

If you do have a module where you are just working with one table @andreievg and I don't much like using a DSL as intended i.e. for `stock_on_hand` instead of `my_table::stock_on_hand`, because it's very easy to create name space conflicts with variable and type names and the errors when you do are really cryptic. Because they're all hidden behind the import `*` wildcard, you don't realise what things are in the modules namespace.

## Example use case

https://github.com/msupply-foundation/open-msupply/pull/5183/commits/eea9ec19ebcbb32c523b4514e4dc820a764afbec

Literally just more readable code. The longer that we have `DSL` aliasing the more it'll get copy pasted.

This is a true refactor - it is functionally identical as far as the compiler is concerned!

IMO a few hours of grinding it out going through the repository module and touching up every table's files. Within a file search and replace text speeds things up well and reduces errors. Some savvy person can probably think how to script it and do whole code base but probably more effort than grinding 😉.

GeronimoJohn:
Fixes #5194 

# 👩🏻‍💻 What does this PR do?

This removes redundant aliasing of Diesel DSLs. According to Diesel's documentation, the dsl alias is only useful when dealing with a single table. We should use DSL for convenience if we are working with just one table in a module, but for the rest, we don't need to alias them.

This is quite the hefty review. In order to make things easier, I've pushed the changes in batches. I was considering different PRs, but I didn't want our project to be filled with them.

See the link below for each commit. Hope it helps 😃 

1. [Initial Batch](https://github.com/msupply-foundation/open-msupply/pull/6057/commits/772aedd47d7c9215e9f3e8b59bdbdd0b96d2700b)
2. [2nd Batch](https://github.com/msupply-foundation/open-msupply/pull/6057/commits/5c84338bbc182dbcc3703a2977ba3bc2ca09869b)
3. [3rd Batch](https://github.com/msupply-foundation/open-msupply/pull/6057/commits/e54f4e796dc9192a3d38ad20bbe8d6be564d0a4f)
4. [4th Batch](https://github.com/msupply-foundation/open-msupply/pull/6057/commits/ecfe582ef4ba9b4d44ffae5d4bea7ad610c0975c)
5. [Last Batch](https://github.com/msupply-foundation/open-msupply/pull/6057/commits/775ce60ed300190bb69cb66782e21bbecf21107a)

# 🧪 Testing

# 📃 Documentation

  1.
  6.

### DIFF
--- a/server/repository/src/db_diesel/activity_log.rs
+++ b/server/repository/src/db_diesel/activity_log.rs
@@ -1,7 +1,4 @@
-use super::{
-    activity_log_row::{activity_log, activity_log::dsl as activity_log_dsl},
-    ActivityLogRow, DBType, StorageConnection,
-};
+use super::{activity_log_row::activity_log, ActivityLogRow, DBType, StorageConnection};
 use diesel::prelude::*;
 
 use crate::{
@@ -69,20 +66,20 @@ impl<'a> ActivityLogRepository<'a> {
         if let Some(sort) = sort {
             match sort.key {
                 ActivityLogSortField::Id => {
-                    apply_sort_no_case!(query, sort, activity_log_dsl::id)
+                    apply_sort_no_case!(query, sort, activity_log::id)
                 }
                 ActivityLogSortField::ActivityLogType => {
-                    apply_sort_no_case!(query, sort, activity_log_dsl::type_)
+                    apply_sort_no_case!(query, sort, activity_log::type_)
                 }
                 ActivityLogSortField::UserId => {
-                    apply_sort_no_case!(query, sort, activity_log_dsl::user_id)
+                    apply_sort_no_case!(query, sort, activity_log::user_id)
                 }
                 ActivityLogSortField::RecordId => {
-                    apply_sort_no_case!(query, sort, activity_log_dsl::record_id)
+                    apply_sort_no_case!(query, sort, activity_log::record_id)
                 }
             }
         } else {
-            query = query.order(activity_log_dsl::datetime.asc())
+            query = query.order(activity_log::datetime.asc())
         }
 
         let result = query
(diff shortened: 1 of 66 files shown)


roxy-dao (approved):
Nice! Massive effort ~ Just a nitpicky comment, but I would have alias some of it to have the same name as the _dsl without the _dsl XD If that makes sense...

Just did a quick search and it seems there are still 551 more results for _dsl.
Screenshot 2025-01-16 at 8 43 45 AM

Could you keep the issue opened and do this in another PR since this one is already massive~?

PR #6058, opened 2025-01-15
Fix error boundary

### LINKED ISSUE #6049: Unexpected Application Error!
## What went wrong? 😲

After loading huge client's datafile, we can experience unexpected application error in different modules while loading data if the session is left for longer period of time or maybe due to slowness.
While loading items:
Screenshot 2025-01-14 at 3 11 20 pm
for prescriptions: 
Screenshot 2025-01-14 at 9 26 35 am
Others:
Screenshot 2025-01-14 at 3 11 13 pm

## Expected behaviour 🤔
No error 

## How to Reproduce 🔨

1. Go to items, patients or prescription
2. Leave the page for few hours or over night
3. We can also see error just after oms cloud setup and opening the list for the first time
4. Click on the list (for me patients and items)
5. See error

## Your environment 🌱

- Open mSupply Version: `V2.5-rc3`
- Legacy mSupply Central Server Version: `V8.0dev39`

- Platform: Google Chrome

- Database type: Postgres

lache-melvin:
Fixes #6049

# 👩🏻‍💻 What does this PR do?

Brings back this guy when something goes wrong:
Screenshot 2025-01-15 at 1 56 19 PM

I left the old error boundary in Host, as high as we can have it, for things that go wrong in those common providers. But we needed to provide another one inside the router.

# 🧪 Testing

- [ ] Cause a crashing error somewhere in the app
  - [ ] As a dev, I found the easiest was to add `throw new Error` somewhere in a component 
- [ ] See error boundary rather than the HTML error 
  - [ ] If you're in dev mode, you'll need to close the error overlays first :)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/host/src/Host.tsx
+++ b/client/packages/host/src/Host.tsx
@@ -75,31 +75,34 @@ const router = createBrowserRouter(
     <Route
       path="*"
       element={
-        <Viewport>
-          <ErrorAlert />
-          <BackButtonHandler />
-          <Box display="flex" style={{ minHeight: '100%' }}>
-            <Routes>
-              <Route
-                path={RouteBuilder.create(AppRoute.Initialise).build()}
-                element={<Initialise />}
-              />
-              <Route
-                path={RouteBuilder.create(AppRoute.Login).build()}
-                element={<Login />}
-              />
-              <Route
-                path={RouteBuilder.create(AppRoute.Discovery).build()}
-                element={<Discovery />}
-              />
-              <Route
-                path={RouteBuilder.create(AppRoute.Android).build()}
-                element={<Android />}
-              />
-              <Route path="*" element={<Site />} />
-            </Routes>
-          </Box>
-        </Viewport>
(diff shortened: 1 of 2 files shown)


client/packages/host/src/Host.tsx
```diff
@@ -108,28 +111,28 @@ const router = createBrowserRouter(
...
-        <ErrorBoundary Fallback={GenericErrorFallback}>
-          <QueryClientProvider client={queryClient}>
-            <GqlProvider
-              url={Environment.GRAPHQL_URL}
-              skipRequest={skipRequest}
-            >
-              <AuthProvider>
-                <Init />
-                <AppThemeProvider>
+      <AppThemeProvider>
```
lache-melvin, line 114: moving up app theme, so both error boundaries both have the theme applied (for same coloured buttons)

roxy-dao (approved):
LGTM~

PR #6071, opened 2025-01-15
6070 Localise ledger date + fix adjustment calculation

### LINKED ISSUE #6070: Localise ledger date + fix adjustment calculation
## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #6070

# 👩🏻‍💻 What does this PR do?
Fix calculation for ledger adjustment + localise date

# 🧪 Testing

- [ ] Have `use consumption data from customers` pref on
- [ ] Create an IO at a store that is a supplying store for another customer
- [ ] Add an item line that has previous requisitions made for it in the same period (note should have other data such as IA + Inbound/Outbound Shipments for this item in the same period)
- [ ] Check the adjustment column for this item for the store (table at the bottom of the input). The calculation should be for period: Delivered Inbound - Verified Outbound + IA (any reductions/additions)

Second:
- [ ] Change timezone and make sure the date is correct with OG

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemLedger.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemLedger.tsx
@@ -37,7 +37,7 @@ const ItemLedgerTable = ({
   const pagination = { page, first, offset };
 
   const { data, isLoading } = useItemLedger(itemId);
-  const { localisedTime } = useFormatDateTime();
+  const { localisedTime, localisedDate } = useFormatDateTime();
 
   const columns = useColumns<ItemLedgerFragment>(
     [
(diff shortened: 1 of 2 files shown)


client/packages/system/src/Item/DetailView/Tabs/ItemLedger.tsx
```diff
@@ -53,6 +53,7 @@ const ItemLedgerTable = ({
         label: 'label.date',
         format: ColumnFormat.Date,
         sortable: false,
+        accessor: ({ rowData }) => localisedDate(rowData.datetime),
```
lache-melvin, line 56: this one isn't needed (already done by `columnFormat.date` and adding it is banging up the display, date column type can't be string like this 😅

lache-melvin (approved):
🚀

PR #6089, opened 2025-01-16
Get invoice line cost/sell price instead of stock line

### LINKED ISSUE #6040: Changes to the item's cost and sell prices get reflected in the item's historic ledger entries
## What went wrong? 😲

If the user makes changes to the item's cost and/or sell prices, the changes will be reflected under Catalogue -> Items -> Item Ledger

https://github.com/user-attachments/assets/e10087a6-5019-48f8-908b-64f31b6a9f93

## Expected behaviour 🤔

To me, this does not make sense, especially since `Total before tax` stays the same - so you can have both sell and cost price sitting at $0, but the total will stay at $30 or $40. Will it make more sense to keep the historic entries as they are, e.g. not change the cost and sell prices in the past?

## How to Reproduce 🔨

1. Choose an item you are testing with
2. Go to Inventory -> View Stock and add some stock to the item
3. Define cost & sell price straight away
4. Go to Catalogue -> Items -> the item's ledger and note the cost, sell and the total before tax prices (you can open it in a new tab for ease of testing)
5. Go back to the stockline and change the prices to $0
6. See under the item's ledger the prices are now $0 (e.g. historic data got changed), but the total before tax stayed the same
7. Does not make sense

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC2

- Database type: PostgreSQL

roxy-dao:
Fixes #6040

# 👩🏻‍💻 What does this PR do?
Use invoice line sell/cost price for historic data instead of stock line ones since those can change...

Should we just be using all details from invoice line? Like batch and such too? 🤔 

# 🧪 Testing

- [ ] Create a stocktake to add new batch for an item. Set sell /cost price as something like $2
- [ ] Go to stock -> change sell/cost price to something like $1
- [ ] Go to Item -> Ledger
- [ ] Check that sell/cost price hasn't changed for historic line

# 📃 Documentation

  1.
  2.


lache-melvin (approved):
> Should we just be using all details from invoice line? Like batch and such too? 🤔

I kind of think yes? I think ledger row should look the same as the related invoice if you were to click on it... But also happy just to get this change in for now

Issue #6090, opened 2025-01-16
Can dispense in fractions even though dispensing is in units of 1


roxy-dao:
## What went wrong? 😲
I can dispense in fractions even if the tablets are in units of 1.

## Expected behaviour 🤔
Shouldn't be able to dispense in fractions if the unit is 1!

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create a prescription
2. Add an item
3. Enter a fractional number
4. Save

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version: v2.5.0-RC3

- Platform:

- Database type:

CarlosNZ:
Closing -- leaving as allowing fractional input for now, after confirmation from. @andreievg

Issue #6078, opened 2025-01-15
Prescriptions add item keeps refocusing on first accordion


roxy-dao:
## What went wrong? 😲
The screen keeps going back to the first opened accordion

https://github.com/user-attachments/assets/0872c922-1d00-4acd-8d1c-7692e759ad68

## Expected behaviour 🤔
Shouldn't jump around the place

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Add new item to prescription
2. Go through inputs
3. See screen keeps scrolling back up

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC3
- Legacy mSupply Central Server Version:

- Platform: Android

- Database type:

roxy-dao:
Setting as high. Makes dispensing very annoying to use since it keeps focusing back on the first accordion and you have to keep scrolling back down with every new input entered

CarlosNZ:
Closed by #6127

AnushaUp:
Tested and Passed✅

Issue #6061, opened 2025-01-15
'Line Already Exists' error when changing the quantity dispensing for the first time in a prescription


marthakeezy:
## What went wrong? 😲

Seeing this 'Line Already Exists' error when updating the quantity for an item that's being issued in a prescription. But only for the first time! - any further changes are saved as per normal with no errors.

Video attached ->>> [here](https://drive.google.com/drive/folders/17u87zwn6kHffeH2rMdK6Q5XKGZDZ8wTc)

## Expected behaviour 🤔

No error

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to Dispensing -> Prescriptions
2. Create a new prescription -> add an item
3. Issue quantity -> Save
4. Change the quantity issued -> save again
5. See error

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC3
- Legacy mSupply Central Server Version: V7.18.00

- Platform: Google Chrome (browser for Mac)

- Database type: PostgreSQL

andreievg:
Hotfix

CarlosNZ:
Close by #6104

AnushaUp:
Moving to done as it can't be recreated in latest build✅

Issue #6059, opened 2025-01-15
Prescriptions: Unit Quantity calculation is wrong if the item's Pack Size is higher than 1


marthakeezy:
## What went wrong? 😲

The title is just my assumption - I am not actually sure where the calculation is wrong, but I think if the Pack Size is higher than 1, then the unit quantity for the item is wrong

<img width="862" alt="Screenshot 2025-01-15 at 2 13 57 PM" src="https://github.com/user-attachments/assets/81156b9d-b6e8-450d-a85d-58c670fa4843" />

## Expected behaviour 🤔

Unit Quantity should be calculated correctly, e.g. 10 for the example above

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Have an item that has Pack Size > 1 (either create a new one or modify an existing one)
2. Go to Dispensary -> Prescriptions and create a new prescription
3. Add the item to it
4. Add a few other batches so you are able to expand the prescription
5. See for the batch that has a higher Pack Size, the Unit Quantity is off

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC3
- Legacy mSupply Central Server Version: V7.18.00

- Platform: Google Chrome (browser for Mac)

- Database type: PostgreSQL

andreievg:
Putting as hotfix in 2.5, we need to be correct with simple calculations

CarlosNZ:
Closed by #6104

PR #6104, opened 2025-01-16
#6083 #6090 #6061 #6059 Prescription hotfixes x4

### LINKED ISSUE #6083: Can still edit details of a verified Prescriptions
## What went wrong? 😲

Verified a prescription but the input is still enabled, thus can change issue and directions. Also even if you don't edit this, and navigate away, the "unsaved changes" modal pops up

Screenshot_20250116_113221_Open mSupply

## Expected behaviour 🤔
Shouldn't be able to edit details of a finalised prescription.

## How to Reproduce 🔨

1. Create a prescription
2. Add an item
3. Verify prescription
4. Go back to the edit page
5. See that input is still enabled

## Your environment 🌱

- Open mSupply Version: v2.5.0-RC3
- Legacy mSupply Central Server Version: Android

- Platform:

- Database type:

CarlosNZ:
Fixes #6083 #6090 #6061 #6059

(There's one more dispensing hotfix (#6078), but that'll require a bit more thinking about, so will do separately.)

# 👩🏻‍💻 What does this PR do?

Four small fixes for the above issues.

I've done each in a different commit, so you can review them individually if you like:

#6083 e249b567c5
#6090 d6767fb83d
#6061 9b825cdb5a
#6059 3953898422

I've also mentioned the issue # in my review comments below.

See review comments for more details

# 🧪 Testing

These are all really tiny, so please just refer to the original issues and confirm behaviour is fixed when doing the "How to reproduce" steps.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
@@ -138,6 +138,7 @@ export const PrescriptionLineEditView = () => {
       );
     }
     setIsDirty(false);
+    setAllDraftLines({});
   };
 
   if (isLoading || !itemId) return <BasicSpinner />;
--- a/client/packages/invoices/src/Prescriptions/LineEditView/hooks/usePrescriptionLineEditRows.ts
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/hooks/usePrescriptionLineEditRows.ts
@@ -3,7 +3,11 @@ import { useTableStore, SortUtils } from '@openmsupply-client/common';
 import { isA } from '../../../utils';
 import { DraftStockOutLine } from '../../../types';
 
-export const usePrescriptionLineEditRows = (rows: DraftStockOutLine[]) => {
+export const usePrescriptionLineEditRows = (
+  rows: DraftStockOutLine[],
+  isDisabled: boolean
+) => {
+  console.log('rows', rows);
   const tableStore = useTableStore();
 
   const isOnHold = (row: DraftStockOutLine) =>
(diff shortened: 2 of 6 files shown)


client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -138,6 +138,7 @@ export const PrescriptionLineEditView = () => {
       );
     }
     setIsDirty(false);
+    setAllDraftLines({});
```
CarlosNZ, line 141: Fix for #6061. Problem was the item ended up in the `allDraftLines` array twice -- once indexed by `"new"` and once indexed by its ID. Best to just start with a fresh object after a save (same as when you first load the page).

client/packages/invoices/src/Prescriptions/LineEditView/hooks/usePrescriptionLineEditRows.ts
```diff
@@ -54,8 +57,9 @@ export const usePrescriptionLineEditRows = (rows: DraftStockOutLine[]) => {
   }, [allocatableRows, wrongPackSizeRows, onHoldRows, noStockRows]);
 
   const disabledRows = useMemo(() => {
+    if (isDisabled) return orderedRows;
```
CarlosNZ, line 60: Fix for #6083

If the prescription is verified, it is `disabled`, so we make sure *all* rows are disabled.

Unfortunately adds another prop that must be passed down through about 3 levels 🙄

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -406,8 +406,9 @@ export const UnitQuantityCell = (props: CellProps<DraftStockOutLine>) => (
       (props.rowData.stockLine?.packSize ?? 1)
     }
     id={getPackQuantityCellId(props.rowData.stockLine?.batch)}
-    decimalLimit={2}
     min={0}
+    // We're not allowing non-integer values (decimalLimit=0), as you can only
+    // issue whole units
```
CarlosNZ, line 411: Fix for #6090. Remove `decimalLimit` prop which forces an integer.

roxy-dao: idk if we want to fix this one because of the use case I mentioned in issue 🤔 maybe revert this change and just close the bug? 🤔

roxy-dao (approved):
Awesome~ Thanks Carl

Issue #5973, opened 2025-01-09
Standardise "LineEditView" behaviour and UI


CarlosNZ:
** Updated **
Mark's summary of the various discussions we've had on this. Putting in here at the top for clarity

* ‘Add item’ button should be top right for consistency with other screens
* The left hand menu component for ‘New item’ only shows when you are adding an item and this isn’t a button it’s a placeholder
* ‘Previous’ button is removed
* ‘Next item’ button is at the bottom right of the edit form
* ‘Next item’ text and behaviour changes to ‘Add item’ if the current item is the last one in the list
* If editing, records are auto-saved
* If adding, records are auto-saved when the last input loses focus or the ’Next item’ button is clicked
* Only create a new record once data has been entered (as per #6890) 
* There is an `OK` button in the footer which closes the edit and returns to the detail view
* The text ‘auto-saved at [time]’ is shown in the footer
* There is a header on the left (item) list stating ‘[action] [number] items’
* The condition for the green tick should be specified by the calling component as this differs between usages
* There is a delete icon button beside the item

Left hand header:
* Dispensing 5 products (dispensary)
* Ordering 5 products (internal orders)
* Supplying 5 products (requisitions)
* Returning 5 products (customer returns)

Related issues:
* Also Re-usable Page Line Edit Layout component #5766
* Adjust the loading indicator a bit higher while entering items #6400
* New internal order input should capture "Next" press rather than dismissing keyboard #6116

### Additional screenshots and notes from @mariyamsupply:

<img width="1024" alt="Image" src="https://github.com/user-attachments/assets/8fb32b8e-9771-4754-8705-e6939de927b4" />
<img width="1024" alt="Image" src="https://github.com/user-attachments/assets/fbe5b83d-8cd6-4a93-a5d0-9e6cb587800f" />

**Check styling please:**
- min font size - 14pt 
- reduce header height, make it white, remove grey line and shadow
- add shadow to the sticky footer
- default directions droppdown styling match inputs (shorter length, plus match the border)

---------------- original issue description -----------------

Follow-on issue from #5959 (see discussion on that PR).

With the new "LineEditView", with the associated "List" sidebar on the left, we need to establish a consistent UI pattern as it's currently implemented slightly differently for Prescriptions and Internal Orders.

![Screenshot 2025-01-09 at 5 04 33 PM](https://github.com/user-attachments/assets/97db39cf-3c48-4a43-a8a1-6f129746e7b7)

[...]

PR #6127, opened 2025-01-16
#6078 Prevent unnecessary scrolling

### LINKED ISSUE #6078: Prescriptions add item keeps refocusing on first accordion

CarlosNZ:
Fixes #6078

# 👩🏻‍💻 What does this PR do?

Turns out the `ref` definition in `<ListOptions>` was calling `scrollIntoView` on every re-render, which was pulling the whole page up.

I've changed it to define the ref in a parent component, and passed it down to the ListOptions, and also defined a `scrollIntoView` method which calls `scrollIntoView()` on that ref from the Nav Bars (Prev/Next buttons).

Also, this problem was affecting *all* `<ListOptions>` implementations, so have made the same fix (in addition to Prescriptions) for:
- Requisitions
- Internal Orders
- Indicators for both of the above

There's a lot of duplicate code, plus the different implementations of the Navigation bars are inconsistent, so hopefully can consolidate this a bit better with #5973 .

# 🧪 Testing

Go to a prescription "Line Edit View" with a decent number of items, and make your browser window small enough to require a scroll bar on the list of items, like this:
Screenshot 2025-01-17 at 12 40 52 PM

Then, scroll down so the selected item is no longer visible:
Screenshot 2025-01-17 at 12 41 01 PM

- [ ] Then, click the "Next" button and the next item should smoothly scroll back into view :)
- [ ] Try out different variations with the "Previous" button, etc.
- [ ] Do similar for the Requisition line edit views
- [ ] Please also test the "How to reproduce" from the original issue (#6078) to confirm the unwanted scrolling is no longer a problem,

### DIFF
--- a/client/packages/common/src/ui/components/navigation/ListOptions/ListOptions.tsx
+++ b/client/packages/common/src/ui/components/navigation/ListOptions/ListOptions.tsx
@@ -21,7 +21,7 @@ interface ListProps {
   options: ListOptionValues[];
   currentId?: string;
   enteredLineIds?: string[];
-  scrollRef?: React.MutableRefObject<HTMLLIElement | null>;
+  scrollRef: React.MutableRefObject<HTMLLIElement | null>;
 }
 
 export const ListOptions = ({
--- a/client/packages/common/src/ui/components/navigation/ListOptions/ListOptions.tsx
+++ b/client/packages/common/src/ui/components/navigation/ListOptions/ListOptions.tsx
@@ -21,6 +21,7 @@ interface ListProps {
   options: ListOptionValues[];
   currentId?: string;
   enteredLineIds?: string[];
+  scrollRef?: React.MutableRefObject<HTMLLIElement | null>;
 }
 
 export const ListOptions = ({
@@ -28,6 +29,7 @@ export const ListOptions = ({
   options,
   currentId,
   enteredLineIds,
+  scrollRef,
 }: ListProps) => {
   const { height } = useWindowDimensions();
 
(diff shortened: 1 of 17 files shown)


client/packages/common/src/ui/components/navigation/ListOptions/ListOptions.tsx
```diff
@@ -64,12 +66,7 @@ export const ListOptions = ({
           <ListItem
             sx={{ padding: '5px 0px', cursor: 'pointer' }}
             onClick={() => onClick(option.id)}
-            ref={
-              option.id === currentId
-                ? l =>
-                    l?.scrollIntoView({ behavior: 'smooth', block: 'center' })
-                : null
-            }
+            ref={option.id === currentId ? scrollRef : null}
```
CarlosNZ, line 69: The `scrollRef` gets applied to the element containing the current selection, which changes. I've had to add a small time delay to the `scrollIntoView` method to ensure that the ref has been applied to the *new* selection before scrolling to it. Without that, it scrolls to the previously select (so it's off by one) as it starts scrolling before the re-render has changed the ref assignment.

I don't love it, but it seems the most straightforward for this particular situation.

jmbrunskill:
@CarlosNZ I'm building an APK to give this a test on tablet...

CarlosNZ:
> I'm interested to know what the desired behaviour is, and why the scroll inter view is needed. Are we trying to focus the item selector when changing pages?

I think it was just so when you have a long list of items, we can make sure the currently selected item is *always* in view. Without it, you can keep clicking "Next" until the selected item is outside the visible area, which is not that nice.

roxy-dao:
> > I'm interested to know what the desired behaviour is, and why the scroll inter view is needed. Are we trying to focus the item selector when changing pages?
> 
> I think it was just so when you have a long list of items, we can make sure the currently selected item is _always_ in view. Without it, you can keep clicking "Next" until the selected item is outside the visible area, which is not that nice.

Yes! This was why (: but I can have a quick whirl if you want @jmbrunskill

roxy-dao (approved):
LGTM~ Thanks Carl (:

PR #6129, opened 2025-01-17
6075 'Invoice Number' column can be clearer

### LINKED ISSUE #6075: 'Invoice Number' column can be clearer
## What went wrong? 😲

I wonder if anyone else finds the invoice numbers confusing if they repeat? I know they are per 'type' but... wondering if they could be clearer?

what if invoice numbers were IA-1 (for inventory additions) or SR-2 (for supplier returns) etc - e.g. maybe we can have identifications before the numbers to indicate which invoice?

photo_2025-01-16 09 48 50

## Expected behaviour 🤔

as per above :)

## How to Reproduce 🔨

1. Choose an item you are testing with
2. Make some changes to it - e.g. add to an outbound shipment, inbound shipment, issue prescription with it - so you have ledger entries to look at
3. Go to Catalogue -> Items -> the item you are testing and click on Ledger
4. See different types of transactions show Invoice Numbers, and sometimes they repeat
5. Even though the numbers are per type of transaction, I still catch myself confused sometimes

## Your environment 🌱

- Database type: PostgreSQL

roxy-dao:
Fixes #6075

# 👩🏻‍💻 What does this PR do?
Concat type + invoice num in ledgers... should change the header? I think it's fine... 🤷🏻 

# 🧪 Testing

- [ ] Go to ledger either through item or stock
- [ ] See type now has invoice type + number

# 📃 Documentation

  1. ledger ss updates
  2.

### DIFF
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemLedger.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemLedger.tsx
@@ -45,7 +45,7 @@ const ItemLedgerTable = ({
         key: 'type',
         label: 'label.type',
         accessor: ({ rowData }) =>
-          t(getInvoiceLocalisationKey(rowData.invoiceType)),
+          `${t(getInvoiceLocalisationKey(rowData.invoiceType))} #${rowData.invoiceNumber}`,
         sortable: false,
       },
       {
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Seems fine to me, I think leave it as Type is ok. Not sure what the column name would be otherwise.
To be fair I think Andrei's suggestion of re-ordering the columns would have been fine too.

PR #6143, opened 2025-01-17
6088 dispensary directions popover shows batch but unclear this is batch

### LINKED ISSUE #6088: Dispensary directions popover shows batch but unclear this is batch
## What went wrong? 😲
Dispensary directions popover in the detail view shows batch number but it's unclear that this is the batch number. Also is it useful to show batch in the Directions? 

Also if you dispense more than 1 batch like the screenshot below, it only shows one batch number

Screenshot 2025-01-16 at 1 46 09 PM

Screenshot 2025-01-16 at 1 46 00 PM

## Expected behaviour 🤔
Should probably remove batch from Directions since you can't give specific directions for each batch and I don't see why you should since it's all the same medicine?

## How to Reproduce 🔨

1. Create a prescription
2. Add item with direction
3. Go back to detail view
4. Long press note icon
5. See batch number... Just the batch number without any description of what the number means

## Your environment 🌱

- Open mSupply Version: v2.5.0-RC3

- Platform:

- Database type:

jmbrunskill:
Fixes #6088

# 👩🏻‍💻 What does this PR do?

image

Dispensing `directions` pop over not longer shows batch numbers, and only one directions section is shown

UI looks a little bit untidy. could be Cleaned up at some point.

# 🧪 Testing

- [ ] Create a prescription
- [ ] Add lines with directions
- [ ] Add multiple batches to a single line.
- [ ] Got to the main view for the prescriptoin and hoverover the notes button
- [ ] You should see the directions shown, but no batch numbers.

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
+++ b/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
@@ -73,18 +73,22 @@ export const usePrescriptionColumn = ({
         accessor: ({ rowData }) => {
           if ('lines' in rowData) {
             const { lines } = rowData;
-            const noteSections = lines
-              .map(({ batch, note }) => ({
-                header: batch ?? '',
-                body: note ?? '',
-              }))
-              .filter(({ body }) => !!body);
+            if (!lines) return null;
+
+            // All the lines should have the same note, so we just take the first one
+            const lineWithNote = lines.find(({ note }) => !!note);
+            if (!lineWithNote) return null;
+
+            const noteSections = [
+              {
+                header: null,
+                body: lineWithNote.note ?? '',
+              },
+            ];
+
             return noteSections.length ? noteSections : null;
-          } else {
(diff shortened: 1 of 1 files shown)


client/packages/invoices/src/Prescriptions/DetailView/columns.ts
```diff
@@ -73,18 +73,22 @@ export const usePrescriptionColumn = ({
...
+            const noteSections = [
+              {
+                header: null,
+                body: lineWithNote.note ?? '',
+              },
+            ];
+
             return noteSections.length ? noteSections : null;
-          } else {
-            return rowData.batch && rowData.note
```
jmbrunskill, line 84: Don't think we need this for prescriptions maybe makes sense for outbound shipments? Not sure...

roxy-dao (approved):
LGTM~

PR #6154, opened 2025-01-19
#5977(2) Fix prescription update tests

CarlosNZ:
Follow-up to #5977 to fix tests broken by clinician_id field to NullableUpdate.

I think @jmbrunskill wanted to have a quick look at the `unwrap()` to try and improve it.

# 🧪 Testing

- Run `cargo test` -- should all pass

### DIFF
--- a/server/service/src/invoice/prescription/update/mod.rs
+++ b/server/service/src/invoice/prescription/update/mod.rs
@@ -147,6 +147,7 @@ mod test {
     use crate::{
         invoice::prescription::{UpdatePrescription, UpdatePrescriptionStatus},
         service_provider::ServiceProvider,
+        NullableUpdate,
     };
 
     use super::UpdatePrescriptionError;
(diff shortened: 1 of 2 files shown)


server/service/src/invoice/prescription/update/mod.rs
```diff
@@ -352,7 +357,7 @@ mod test {
                     diagnosis_id: _,
                 } = get_update();
                 u.name_link_id = patient_id.unwrap();
-                u.clinician_link_id = clinician_id;
+                u.clinician_link_id = clinician_id.unwrap().value;
```
jmbrunskill, line 360:
```suggestion
                u.clinician_link_id = clinician_id.map(|c| c.value).unwrap_or_default();
```
This works, but feels a bit clunky, I think what you have is totally fine.

Slightly better is this, but feels weird having two functions, and the second one is the one that runs on a `Some` value?
```suggestion
               u.clinician_link_id =  clinician_id.map_or_else(|| None, |c| c.value);
```

The problem with just use the map function is that it automatically wraps whatever you're mapping in Some()
So it doesn't let you just return a Some yourself. THis is why we were getting an extra `Some` in the type definition...

jmbrunskill (approved):
Thanks @CarlosNZ added some ideas for how to fix that unwrap() for learning purposes, don't need to change to using it.

server/service/src/invoice/prescription/update/mod.rs
CarlosNZ: Thanks for the tips :)

PR #6157, opened 2025-01-19
5579 standardise total stock on hand vs available stock on hand

### LINKED ISSUE #5579: Item usage report showing different data on mSupply vs on OMS

fergie-nz:
Fixes #5579 

# 👩🏻‍💻 What does this PR do?

Standardises SOH to total stock on hand (rather than available stock on hand)

Updates report sidepanel, argument modal, and 'how to read' to show default monthly lookback period when it is being used.

Set Stock Status report to show total stock on hand rather than available stock on hand
Set Item Usage report to show total stock on hand rather than available stock on hand
Show AMC which is used in MOS calculation in Item Usage report

There is a lot of noise in reviewing due to duplication of report source files.

For ease of reviewing, I recommend not viewing anything within the reports/ folder in github, and instead use the following terminal commands from the OMS repo:

`diff -bur --exclude=dist server/reports/item-usage/2_3_0/ server/reports/item-usage/2_5_0/`
`diff -bur --exclude=dist server/reports/stock-status/2_3_0/ server/reports/stock-status/2_5_0/`

# 🧪 Testing

- [ ] Navigate to reports
- [ ] Open any of the standard reports filter modal, and see that AMC look back period is visible as the default (3 months) when no store preference is set for AMC lookback period.
- [ ] Generate stock status report
- [ ] See that values are for total stock on hand, not available stock on hand (will exclude item counts which are allocated)
- [ ] Generate item usage report
- [ ] See new column AMC (lookback period)
- [ ] View the 'how to read report' text in item usage report and see mention of default AMC lookback period
- [ ] View the side panel ('More' button on report list view), and see default amc lookback period is used when no store preference is set for AMC lookback period.

# 📃 Documentation


server/reports/item-usage/2_5_0/convert_data_js/src/utils.js
```diff
@@ -0,0 +1,70 @@
...
+    item.expiringInTwelveMonths = calculateQuantity(
+      data.expiringInTwelveMonths,
+      item.id
+    );
+    item.stockOnOrder = calculateQuantity(data.stockOnOrder, item.id);
+    item.AMC12 = calculateQuantity(data.AMCTwelve, item.id);
+    item.AMC24 = calculateQuantity(data.AMCTwentyFour, item.id);
+    item.SOH = calculateStatValue(item?.stats?.stockOnHand);
+    item.MOS = calculateStatValue(item?.stats?.availableMonthsOfStockOnHand);
+    item.AMCDefault = calculateStatValue(item?.stats?.averateMonthlyConsumption)
```
lache-melvin, line 34: Can you explain why this is called AMCDefault? Is item stats AMC always using the default lookback period or something? I wonder if there is a clearer name...

lache-melvin: Actually, what is the point of this column? I don't see it in the OG report or mentioned in the issue?

lache-melvin (changes requested):
Needing some clarity on the AMC default column? 

SOH changes look good, I wonder if we should only have these?

server/reports/item-usage/2_5_0/convert_data_js/src/utils.js
fergie-nz: There was confusion about how 'Months Cover' column was calculated.
It is calculated using the consumption over the store preferences lookback period (or the default 3 months lookback period).

Confusion could occur because the months cover might differ from what users would expect it to be given the AMC12 and AMC24 provided. Adding it in shows the users the AMC which is used to calculated the months cover column.

@andreievg mentioned this to me in person, but I hadn't updated the issue. What do you reckon?

fergie-nz: > Can you explain why this is called AMCDefault? Is item stats AMC always using the default lookback period or something? I wonder if there is a clearer name...

Good point - I'm thinking just 'AMC'

lache-melvin: Ah ok that makes sense - it should probably show something other than 0 then? 😁 

For testing... this should match the item detail view AMC value right?

Screenshot 2025-01-21 at 4 09 45 PM

fergie-nz: > Ah ok that makes sense - it should probably show something other than 0 then? 😁

Yikes! Yes it should be.
Unless of course you haven't used any of those medicines at that store in the last (lookback period) months?
Default is 3 if not set in store preferences.
 
> For testing... this should match the item detail view AMC value right?

Correct - both pull directly from itemStats node

client/packages/system/src/Report/components/ReportArgumentsModal.tsx
```diff
@@ -22,13 +22,16 @@ export const ReportArgumentsModal: FC<ReportArgumentsModalProps> = ({
...
-  const {
+  let {
     monthlyConsumptionLookBackPeriod,
+    defaultMonthlyConsumptionLookBackPeriod,
     monthsOverstock,
     monthsUnderstock,
     monthsItemsExpire,
   } = store?.preferences ?? {};
 
+  monthlyConsumptionLookBackPeriod = monthlyConsumptionLookBackPeriod !== 0.0 ? monthlyConsumptionLookBackPeriod : defaultMonthlyConsumptionLookBackPeriod;
```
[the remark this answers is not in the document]
fergie-nz: I'm just making this change, though I was wondering why mutable is no good? Is it unsafe?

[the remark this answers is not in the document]
lache-melvin: Not unsafe in terms of code giving unexpected bugs, but does mean that those values could now change anywhere, and it's hard to tell where. `const` gives us more confidence as devs that the value we see here, where it is being defined, is that value we'll get when we use the variable. Just a best practice 🚀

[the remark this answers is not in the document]
fergie-nz: Nice! Thank you :)

server/reports/item-usage/2_5_0/convert_data_js/output.json
```diff
@@ -0,0 +1,49 @@
...
+        "name": "Item A",
+        "stats": {
+          "stockOnHand": 200,
+          "totalConsumption": 100,
+          "availableMonthsOfStockOnHand": 4.5
+        },
+        "stockOnOrder": 5,
+        "AMC12": 150.8,
+        "AMC24": 250.8,
+        "AMCDefault": 0,
```
lache-melvin, line 38: Test is failing, I think because this is still here...

lache-melvin (changes requested):
I'm still getting the 0s... lookback period 3 months:
Screenshot 2025-01-22 at 12 13 12 PM

Item stats do have values, e.g.:
Screenshot 2025-01-22 at 12 13 21 PM

Is it working for you? What am I doing wrong lol?

Side question... the 2_3_0 reports no longer exist - assuming we've gone with that decision from the discussion [here](https://github.com/msupply-foundation/open-msupply/discussions/6155)? Did they get deleted in the RC base branch?

client/packages/reports/src/ListView/SidePanel.tsx
```diff
@@ -70,7 +70,7 @@ export const SidePanelComponent = () => {
               label={t('label.monthly-consumption-look-back-period')}
               decimalLimit={2}
               endAdornment={t('label.months')}
-              value={preferences?.monthlyConsumptionLookBackPeriod}
+              value={preferences?.monthlyConsumptionLookBackPeriod !== 0.0 ? preferences?.monthlyConsumptionLookBackPeriod : preferences?.defaultMonthlyConsumptionLookBackPeriod}
```
andreievg, line 73: I think it's better to make monthlyConsumption not nullable return of preference with the default value, we shouldn't be handling defaults in front end, and as per before, let's make it easier on consumer

server/reports/item-usage/2_5_0/argument_schemas/arguments.json
```diff
@@ -0,0 +1,49 @@
+{
```
andreievg, line 1: If we not changing API, don't need to update these

fergie-nz: API has changed to include stockOnHand rather than availableStockOnHand in item stats. 

However I think some fixes made to the 2.3.0 report which were made after I copied them into 2.5.0 - so I will add these fixes in now.

lache-melvin (changes requested):
Thanks for changes, AMC coming through in the report correctly now...

Tests still failing though, and this not translated... need to regenerate again?

Screenshot 2025-01-23 at 9 29 32 AM

mark-prins:
Can you also change/check the requisition graphs as per #6208 - best if that is showing total stock too

roxy-dao:
Please update with develop and fix the conflicts so the bloat on your PR is gone 😢

server/graphql/types/src/types/store_preference.rs
```diff
@@ -82,6 +83,18 @@ impl StorePreferenceNode {
...
+                    .monthly_consumption_look_back_period
+                    == 0.0
+                {
+                    DEFAULT_AMC_LOOKBACK_MONTHS.into()
+                } else {
+                    store_preference.monthly_consumption_look_back_period
+                },
+                ..store_preference
+            },
+        }
```
fergie-nz, line 98: @andreievg FYI I added this here because store preferences are called from a couple of places which bypass the service layer, and I liked having this mapping all in one place.

However I think this is relatively unique for from_domain functions, so I wasn't sure if this was the right way to go.

server/reports/item-usage/2_5/manifest.json
```diff
@@ -1,6 +1,6 @@
 {
   "is_custom": false,
-  "version": "2.3.0",
+  "version": "2.5.0",
```
fergie-nz, line 3: Noting here that 2.5.0 version change is required because stockOnHand in itemStatsNode was only included in 2.5 release

lache-melvin:
Hi @andreievg are you ok to review this one? Feels like there might be context/convos I'm missing...

fergie-nz:
> Hi @andreievg are you ok to review this one? Feels like there might be context/convos I'm missing...

Hi @lache-melvin apologies yes we had a discussion about how to make reviewing with big difs from changing report versions easier for reviewers.

Update to docs inbound shortly.

Basically in future we are duplicating report files in a separate PR, then making a second PR against the first. This means the dif only shows changes.

I've done that retroactively here by duplicating the reports on a new PR, and rebased this one to that PR.

andreievg (changes requested):
Looks great, just one small change request please, and can you merge this change please: https://github.com/msupply-foundation/open-msupply/compare/5579-standardise-SOH...5579-standardise-SOH(Use-store-preference-defaults)
 
Basically the way we were using store preference wasn't quite as i planned. Should always have a default store preference, i think that cause confusion with amc lookback months display, above change is a bit broader then this PR but I wanted to fix it once and for all

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1604,6 +1604,8 @@
   "reason.stored": "Stored",
   "reason.unknown": "Unknown",
   "replenishment": "Replenishment",
+  "report.AMC-lookback": "AMC Lookback Period",
+  "report.amc": "AMC (lookback period)",
```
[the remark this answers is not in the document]
fergie-nz: Thank you I think merge conflict. I have consolidated to amc-lookback and removed the duplicate caps option also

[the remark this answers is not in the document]
lache-melvin: FWIW i thought first one was "this is the AMC lookback period" and the second was "this is the AMC value for the lookback period" ?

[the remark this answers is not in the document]
fergie-nz: > FWIW i thought first one was "this is the AMC lookback period" and the second was "this is the AMC value for the lookback period" ?

You're right! So maybe the AMC (lookback period) is more clear for that use case.

I don't have an opinion on this. Worth re adding?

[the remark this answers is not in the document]
lache-melvin: Yeah I'd say so, they're meaning quite different things. Maybe you can call the second one `report.amc-for-lookback-period` or something to make it clearer when reading the translations out of context?

server/service/src/requisition_line/response_line_stats/response_line_stats.rs
```diff
@@ -50,8 +50,7 @@ pub fn response_store_stats(
     )?;
 
     let stock_on_hand = stock_lines.iter().fold(0.0, |sum, stock_line| {
-        sum + stock_line.stock_line_row.available_number_of_packs
-            * stock_line.stock_line_row.pack_size
+        sum + stock_line.stock_line_row.total_number_of_packs * stock_line.stock_line_row.pack_size
```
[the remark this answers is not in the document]
fergie-nz: In this case the item_usage report access the response_line_stats node in its graphql query which is why I have added this here. 

The only other place I can see this node is being used is in the ResponseStoreStats, which makes sense to me. But worthwhile waiting until we consolidate total everywhere do you think?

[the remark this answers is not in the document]
andreievg: oki, sounds good, thanks

andreievg (approved):
Looks great thanks ! Thanks for all your work on this

PR #6166, opened 2025-01-20
6162 in stock and available show wrong numbers once the prescription is saved

### LINKED ISSUE #6162: In Stock and Available show wrong numbers once the prescription is saved
## What went wrong? 😲

Noticing this weird behaviour: In Stock (Units) and Available (Units) show wrong numbers once the prescription is saved. Upon refreshing/navigating away and back, the numbers reset.

Video ->>> [here](https://drive.google.com/drive/folders/1k3CUXTJkNS3mGrwc2aEnt_E142DKBk8g)

## Expected behaviour 🤔

Numbers should be accurate

## How to Reproduce 🔨

1. Go to Dispensary -> Prescriptions
2. Create a new prescription and add an item (note available/in stock numbers)
3. Issue quantity and click Save (note available/in stock numbers have changed)
4. Refresh or navigate away and back - the numbers are back to normal

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC4

- Platform: Galaxy Tab A8 (Android 14)

- Database type: SQLite3

jmbrunskill:
Fixes #6162

# 👩🏻‍💻 What does this PR do?

Invalidates the historical stock query after updating stock lines.
This causes the code to correctly re-calculate the available stock, including what's already allocated to this prescription

# 🧪 Testing

See issue for testing instructions

# 📃 Documentation

  1.
  2.


client/packages/invoices/src/Prescriptions/api/hooks/usePrescriptionLines.ts
```diff
@@ -13,6 +13,7 @@ import { usePrescriptionGraphQL } from '../usePrescriptionGraphQL';
 import { PrescriptionRowFragment } from '../operations.generated';
 import { PRESCRIPTION, PRESCRIPTION_LINE } from './keys';
 import { createInputObject, mapStatus } from './utils';
+import { HISTORIAL_STOCK_LINES } from 'packages/system/src/Item/api/keys';
```
CarlosNZ, line 16: I think we're supposed to do:
```ts
import { HISTORICAL_STOCK_LINES } from '@openmsupply-client/system/src/Item/api/keys';
```

(Similarly for `import { DraftStockOutLine }` above -- be good to fix this too)

CarlosNZ: Also -- spelling of "HISTORI**C**AL" ;)

client/packages/system/src/Item/api/hooks/useHistoricalStockLines/useHistoricalStockLines.tsx
```diff
@@ -12,7 +13,14 @@ export const useHistoricalStockLines = ({
   const sdk = getSdk(client);
   const { storeId } = useAuthContext();
 
-  const result = useQuery(`stocklinesAt${storeId}${itemId}${datetime}`, () =>
+  const key: string[] = [
```
CarlosNZ, line 16: The type definition can be implied here, so don't really. need `: string[]`. Although if you're just adding it for clarity, then that's cool

CarlosNZ (changes requested):
>  This causes the code to correctly re-calculate the available stock, including what's already allocated to this prescription

This seems to be going the wrong way. If I issue 50 units, then save, the "Available" amount has gone *up* by 50. Shouldn't it go *down*?

client/packages/system/src/Item/api/hooks/useHistoricalStockLines/useHistoricalStockLines.tsx
jmbrunskill: I think I added it when trying to resolve my confusion around the datetime, thought I had to convert it to string, but the realised it was just nullable
I think it's ok though?

jmbrunskill:
> > This causes the code to correctly re-calculate the available stock, including what's already allocated to this prescription
> 
> This seems to be going the wrong way. If I issue 50 units, then save, the "Available" amount has gone _up_ by 50. Shouldn't it go _down_?

The reason this was happening is because when we have stock assigned to the current prescription, we want to show available stock including the stock assigned to that line from this prescription. This aims to make it consistent between what was available before an after you click save, and makes is clearer to the user what they can assign to the invoice.

For example If we start with Batch: ABC with 100 packs available.

In a prescription, I can allocate 75 of those to the current prescription

When I hit save, the stock line for batch ABC would be updated to say only 25 are now available.

If we then refreshed the stocklines on save, it would look like we've allocated 75 out of 25 available, to compensate for that, after saving we add the 75 assigned to this invoice back on to the stock lines, increasing the available stock.

What was happing before is that we were adding on the 75 stock, but we hadn't refreshed the stocklines (giving us the correct updated available stock, so it was adding the 75 on to the existing 100 creating 175.

Hopefully that explains it?

client/packages/invoices/src/Prescriptions/api/hooks/usePrescriptionLines.ts
jmbrunskill: > Also -- spelling of "HISTORI**C**AL" ;)

Turns out Historcal is correct in Spanish (not that I knew that before...)

CarlosNZ:
> When I hit save, the stock line for batch ABC would be updated to say only 25 are now available.
> 
> If we then refreshed the stocklines on save, it would look like we've allocated 75 out of 25 available, to compensate for that, after saving we add the 75 assigned to this invoice back on to the stock lines, increasing the available stock.

Okay, so if I'm understanding correctly, what you're saying is the "available" number should stay the same so it doesn't look like you have less available than what you're issuing. This makes sense, and this is what I see *after* refreshing the page. But what's happening when I press "Save" is that the "available" number immediately goes *up*, which doesn't make sense, and I have to refresh the page to bring it back to a meaningful value.

See this recording:



Whatever the "correct" value is, it should at least be the same before and after a refresh, wouldn't you think?

jmbrunskill:
> Whatever the "correct" value is, it should at least be the same before and after a refresh, wouldn't you think?

Agreed, in that video my fix clearly isn't working. Maybe it's a firefox thing or my changes to the keys aren't correct. I'll test again...

jmbrunskill:
> > Whatever the "correct" value is, it should at least be the same before and after a refresh, wouldn't you think?
> 
> Agreed, in that video my fix clearly isn't working. Maybe it's a firefox thing or my changes to the keys aren't correct. I'll test again...

It's definitely working for me on my branch in both chrome and firefox.


Can you check again @CarlosNZ ? Are you sure you were on my branch when you took that recording?

CarlosNZ:
> Can you check again @CarlosNZ ? Are you sure you were on my branch when you took that recording?

Yup, definitely the correct branch. I just tried again and I can reproduce it, but only for some items -- some of them work as expected. I can have a look at it with you at/after stand-up if you want.

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -103,7 +103,8 @@ export const createDraftStockOutLine = ({
 
   const adjustTotalNumberOfPacks = invoiceStatus === InvoiceNodeStatus.Picked;
 
-  let adjustedStockLine = stockLine ? stockLine : invoiceLine?.stockLine;
+  // Note to future self, the stockLine spread here is important, if not spread you'll be modifying the tanStack Query Cache, with unintended effects!
+  let adjustedStockLine = stockLine ? { ...stockLine } : invoiceLine?.stockLine;
```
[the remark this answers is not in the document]
jmbrunskill: Thanks, that seems to work, modifying subsequent statements seemed to result in some werid behaviour.
Have push, please check again...

CarlosNZ (approved):
Thanks James, let's get it out :)

PR #6170, opened 2025-01-20
4707 automate android build

### LINKED ISSUE #4707: Automate Android Build
## The suggestion

Use the NZ Mac Mini to automatically create an android build when a release tag being create.
Alternatively we could do it in Jenkins, if that made it easier somehow?

Would save time for release team creating a android release for every release
Would ensure the full build process was defined in code, reducing the reliance on sharing knowledge & keys for production builds.

Automating a process may take more time than running it manually.

Feels like a 2-4 day job? Could go faster potentially but I think it's worth spending a few days on it if needed.

aimee-mcneil-melville:
Fixes #4707 

# 👩🏻‍💻 What does this PR do?

Automates the Android Build via Github Action which uses the mac mini as a self-hosted runner. The result is the APK being downloadable as an artifact from the completed action

Saves time for the release team - is fully automated in creating the APK, including keys needed. This is triggered on a tag being pushed that starts with 'v'.

Re the mentioned risk of this potentially taking more time: Runtime for the action is around 10 minutes (if not queued by another action)

There are a few warnings in relation to a depreciated command. I used the same set-up as other actions doing the same steps. Should be an easy update to do on all of them at the same time.

Is there a readme or other internal docs that might need updating?

# 🧪 Testing

- [ ] Add a tag on a commit, begin the tag with 'v' & push it to Github
- [ ] Go to Github -> Actions tab
- [ ] You should see the action in progress, then when it has a green tick it is complete
- [ ] Click into the action and scroll to the 'Artifacts' section
- [ ] Download the Artifact
- [ ] If you unzip the Artifact in your downloads folder, you will get a folder named Android_Build. This contains a file named open-msupply-[version number in package.json]-release.apk, plus a file of metadata

# 📃 Documentation


lache-melvin (approved):
LGTM! Heeeeaps of work Aimee, well done getting it over the line 🔥 

I think there are some things we can iterate on - those set_output warnings, and perhaps uploading just the APK file rather than the folder. But let's get this much in, that can be follow up 🙏

Issue #5990, opened 2025-01-09
Numeric Text input -- Error state pattern


CarlosNZ:
Part of changes described [here](https://github.com/msupply-foundation/open-msupply/issues/4213#issuecomment-2581450867)

Create a mechanism (probably a hook) to hold the error state of a form, and pass down the error state and a "setError" method into form components.

~~Number component should then use this mechanism instead of modifying the input for invalid values.~~ -- Initial version, don't remove existing Number component behaviour, come back to it once this pattern and its UI are established (#5992)

marthakeezy:
`2.20.00-develop-06070242`

Testing insurance form and vaccine course

marthakeezy:
Raised #11982 for insurance policies - everything seems to be all good when adding a vaccine course. Moving to Done ✅

PR #6172, opened 2025-01-20
5990(1) Allow custom props for table cells without creating new component.

CarlosNZ:
Basically a pre-requisite for #5990, but just wanted to make this is separate PR as it's essentially a separate refactor of existing functionality and didn't want to over-complicate #5990 proper.

# 👩🏻‍💻 What does this PR do?

It's been an ongoing hassle when making Table cells that require more props than are available from the table provider. If you try and define a new Component inline, you get problems with losing input focus on re-render. So you have to define a component outside the table column definitions. But even with that, I was having problems implementing what I need for #5990 without getting input focus issues on the Number Input.

So this PR just provides a very simple mechanism to pass props into whatever cell is defined in the `Cell` field of a table column definitions -- the column definition can take an additional `cellProps` property, which is then passed into the Cell component where it is rendered (inside `DataRow`)

I've implemented it just in one place to demonstrate that it works as expected -- in `StocktakeLineEditTables`. There will be a more substantial implementation in #5990 proper, 

Any potential problems? I don't think so, but haven't tried too hard to break it.

# 🧪 Testing

Check that the "Counted Number Of Packs" field in the Stocktake Line Edit modal behaves exactly the same as before.


client/packages/common/src/ui/layout/tables/columns/types.ts
```diff
@@ -90,6 +90,9 @@ export interface Column<T extends RecordWithId> {
   backgroundColor?: string;
 
   Cell: JSXElementConstructor<CellProps<T>>;
+  // For passing additional props to the above Cell
+  cellProps?: Record<string, unknown>;
```
lache-melvin, line 94: Hm ... I played around with something like this: 

` cellProps?: Omit<ComponentProps<Column<T>['Cell']>, keyof CellProps<T>>;`

It's not quite right but might be good to experiment with for a sec? Would be nice for cellProps to be typed based on the Cell

lache-melvin (approved):
Happy with this as it is, better than the separately defined Cells (is there a plan to clean these up?) but if we can type it, that would be my preference 🙏

client/packages/common/src/ui/layout/tables/columns/types.ts
CarlosNZ: Yeah, I had it *almost* working too. I'll have another look, won't spend too long on it.

mark-prins: thanks - that was my concern too. allowing untyped spreading and working around the type system isn't ideal

CarlosNZ: The difficulty is that the `Cell` prop itself isn't actually correctly defined: `JSXElementConstructor<CellProps<T>>` but it can can have props beyond `CellProps<T>`, so it's hard to define props type for props the `Cell` property isn't actually aware of (if that makes any sense). 

I tried passing in a second generic `U` to the Column definition, but this requires every column to have the same props for its "Cell" which is obviously not the case. 

I think we inherently lose type information by defining `Cell` like this. Any other suggestions?

CarlosNZ: For now I've just added a comment to use with caution. I think a more significant refactor is in order to properly type these 🤔

PR #6198, opened 2025-01-20
5818 Creating a new encounter does not take you to a newly created encounter, instead keeps you on a landing page

### LINKED ISSUE #5818: Creating a new encounter does not take you to a newly created encounter, instead keeps you on a landing page
## What went wrong? 😲

@Rajajajaja @richardmoizeau 

When creating a new encounter for a patient, the action of saving an encounter does not bring you to the newly created page, but instead keeps you on the landing page

Video of the behaviour ->>> [here](https://drive.google.com/drive/folders/1rAw_e6Jva7ad3tS-WgB4HAx30Oyo2SZS)

## Expected behaviour 🤔

I think it is intuitive and makes sense for the user to create an encounter and then be brought to the said encounter to continue editing the details. For example, things like start/end time, you cannot do it from the first screen when you are just adding an encounter, so naturally they would want to keep editing. And especially in context of family planning module, when the main use case would be to prescribe contraception, the midwives will need to go into the encounter anyway, so it doesn't make sense for them to see the overall page?

## How to Reproduce 🔨

1. Make sure you have Programs and Family Planning module enabled for your store
2. Go to your Dispensary -> Patients -> click on an existing one or create a new one
3. Add an encounter to the patient -> see the action of adding one does not bring you to the detailed encounter page, but keeps you on the landing one

## Your environment 🌱

- Open mSupply Version: `V2.4.1`
- Legacy mSupply Central Server Version: `V7.19.00`

- Database type: `SQLite3`

roxy-dao:
Fixes #5818

# 👩🏻‍💻 What does this PR do?
Redirecting to Encounter if the datetime is less than right now (with 5 minute buffer). Have also allowed users to select a time to save in the start datetime so they don't have to go into future encounters to configure this

# 🧪 Testing

- [ ] Create an encounter for tomorrow
- [ ] Don't get redirected
- [ ] Create an encounter for today or in the past
- [ ] Get redirected to Encounter edit page

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
+++ b/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
@@ -296,6 +296,7 @@ export const CreateEncounterModal: FC = () => {
                           highlightedDays: getHighlightedDays(),
                         } as any,
                       }}
+                      showTime
                     />
                   }
                 />
--- a/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
+++ b/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
@@ -173,9 +173,7 @@ export const CreateEncounterModal: FC = () => {
   ]);
 
   const setStartDatetime = (date: Date | null): void => {
-    const startDatetime = DateUtils.formatRFC3339(
-      DateUtils.addCurrentTime(date)
-    );
+    const startDatetime = date?.toISOString();
     setDraft({
       ...currentOrNewDraft(),
       startDatetime,
--- a/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
+++ b/client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
@@ -231,19 +231,23 @@ export const CreateEncounterModal: FC = () => {
                 return;
               }
               const startDatetime = new Date(draft?.startDatetime ?? 0);
-              if (DateUtils.addHours(startDatetime, 1).getTime() > Date.now()) {
+              const dateNow = Date.now();
+              if (
+                startDatetime.getTime() <=
+                DateUtils.addMinutes(dateNow, 5).getTime()
+              ) {
                 navigate(
                   RouteBuilder.create(AppRoute.Dispensary)
-                    .addPart(AppRoute.Patients)
-                    .addPart(patientId)
-                    .addQuery({ tab: PatientTabValue.Encounters })
+                    .addPart(AppRoute.Encounter)
+                    .addPart(id)
                     .build()
                 );
               } else {
                 navigate(
                   RouteBuilder.create(AppRoute.Dispensary)
-                    .addPart(AppRoute.Encounter)
-                    .addPart(id)
+                    .addPart(AppRoute.Patients)
+                    .addPart(patientId)
+                    .addQuery({ tab: PatientTabValue.Encounters })
                     .build()
                 );
               }


client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
```diff
@@ -231,19 +231,23 @@ export const CreateEncounterModal: FC = () => {
...
+              if (
+                startDatetime.getTime() <=
+                DateUtils.addMinutes(dateNow, 5).getTime()
+              ) {
                 navigate(
                   RouteBuilder.create(AppRoute.Dispensary)
-                    .addPart(AppRoute.Patients)
-                    .addPart(patientId)
-                    .addQuery({ tab: PatientTabValue.Encounters })
+                    .addPart(AppRoute.Encounter)
```
roxy-dao, line 241: Just swapping the conditions around XD

client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
```diff
@@ -292,6 +296,7 @@ export const CreateEncounterModal: FC = () => {
                           highlightedDays: getHighlightedDays(),
                         } as any,
                       }}
+                      showTime
```
lache-melvin, line 299: This doesn't work... onChange sets it back to now 😅  Would just remove? Haven't heard request to be able to set the encounter time (though not sure why not lol)

client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
```diff
@@ -231,19 +231,23 @@ export const CreateEncounterModal: FC = () => {
                 return;
               }
               const startDatetime = new Date(draft?.startDatetime ?? 0);
-              if (DateUtils.addHours(startDatetime, 1).getTime() > Date.now()) {
+              const dateNow = Date.now();
+              if (
+                startDatetime.getTime() <=
+                DateUtils.addMinutes(dateNow, 5).getTime()
```
lache-melvin, line 237: Given can't set time, we can prob just do `if (startDatetime.getTime() > Date.now())` here?

lache-melvin (changes requested):
Thankssss, just the time thing 🙏

client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
roxy-dao: We can set time now >.>

client/packages/system/src/Patient/Encounter/CreateEncounterModal.tsx
roxy-dao: O_O It worked for me... I set time... got redirected and Encounter startdatetime was set but will double check

roxy-dao: [d01ca5c](https://github.com/msupply-foundation/open-msupply/pull/6198/commits/d01ca5c764223410058a7b669d559f8c999dfd78)

lache-melvin (approved):
Working well for me - thanks Roxy 😁 

Only concern, maybe @mark-prins can help, is why that time picker got removed? Wondering if it was too slow having to enter the time? Can be a separate issue if needed - maybe we should have current date/time as the default for the encounter time when you open the modal 👀

roxy-dao:
> Working well for me - thanks Roxy 😁
> 
> Only concern, maybe @mark-prins can help, is why that time picker got removed? Wondering if it was too slow having to enter the time? Can be a separate issue if needed - maybe we should have current date/time as the default for the encounter time when you open the modal 👀

I feel like it was one of those that Clemens just removed just cause >.>

lache-melvin:
Actually looks like we already have some start-date setting, for when there is a suggested next encounter. Maybe not too hard to set? https://github.com/msupply-foundation/open-msupply/compare/develop...5818-Creating-a-new-encounter-does-not-take-you-to-a-newly-created-encounter%2C-instead-keeps-you-on-a-landing-page(autoset-start-date)

roxy-dao:
> Actually looks like we already have some start-date setting, for when there is a suggested next encounter. Maybe not too hard to set? [develop...5818-Creating-a-new-encounter-does-not-take-you-to-a-newly-created-encounter,-instead-keeps-you-on-a-landing-page(autoset-start-date)](https://github.com/msupply-foundation/open-msupply/compare/develop...5818-Creating-a-new-encounter-does-not-take-you-to-a-newly-created-encounter,-instead-keeps-you-on-a-landing-page(autoset-start-date))

Doesn't look like it is! 😄

PR #6209, opened 2025-01-21
Catch panic in discovery

### LINKED ISSUE #5018: Discovery intermittently crashes

jmbrunskill:
Fixes #5018 (hopefully?)

# 👩🏻‍💻 What does this PR do?

Catches the panic on discovery, and just logs an error.

We still probably want to get a proper fix, and possibly re-try adding discovery

Not 100% sure if this solved the issue as it's not happening for me here at home at the moment.
Would be good for someone to test it who's having the problem regularly...

Branched off V2.5.0 RC branch as that is where I was seeing the problem more regularly.

@andreievg  also suggested we could do add a configuration file to disable discovery.
Thought I'd try this first to see if it helps...

# 🧪 Testing

- [ ] start your remote server in the office a number of times, see if it randomly crashes sometimes or not...

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/server/src/discovery.rs
+++ b/server/server/src/discovery.rs
@@ -1,3 +1,5 @@
+use std::panic;
+
 use crate::certs::Protocol;
 use {astro_dnssd::DNSServiceBuilder, std::collections::HashMap};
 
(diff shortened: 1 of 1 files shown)


CarlosNZ (approved):
LGTM -- tested for several minutes after a period of heavy crashing, seems good 👍

andreievg:
> LGTM -- tested for several minutes after a period of heavy crashing, seems good

Did you get the panic log ?

andreievg:
I think before we merge I would like to see the panic log, since I remember putting an unwrap in the tokio task and it didn't cause a panic, which for some reason tells me that the error can escape panic (especially since it happens in the unsafe code ¯\_(ツ)_/¯), i'll just have this running for a little bit, fingers crossed will get the log

jmbrunskill:
> I think before we merge I would like to see the panic log, since I remember putting an unwrap in the tokio task and it didn't cause a panic, which for some reason tells me that the error can escape panic (especially since it happens in the unsafe code ¯_(ツ)_/¯), i'll just have this running for a little bit, fingers crossed will get the log

With my manual unwrap() I did get the panic log as well as not crashing, but agree would like to see it properly with the issue...

andreievg (approved):
I guess only time will tell, thanks @jmbrunskill !

jmbrunskill:
@andreievg merging this for now, as people are apparently getting a lot of panics but don't want to be stuck on this branch. If there's any issues or concerns with this approach happy to roll back as appropriate.

jmbrunskill:
@mark-prins reports this isn't working sadly :(

mark-prins:
I've had the server crash a couple of times this morning - and have now commented out the disco 🪩 😢

PR #6211, opened 2025-01-21
6144 add check for exceeding max orders

### LINKED ISSUE #6144: You can create more than the max number of  emergency internal orders by period defined for a program.
## What went wrong? 😲

Each program order type as a max order a customer can create for a period. I was able to create more than the max order specified for the program emergency orders type. It is working for normal order type.

Emergency order Setup of max 2
https://drive.google.com/file/d/1izMh1R9OIPqmP9tNuJCRyQ-0WLrmgVkX/view?usp=drive_link

Created 3 omSupply emergency orders
https://drive.google.com/file/d/1l_D_D6SCjwrC6XA4qG_Mr17tv9WXhw00/view?usp=sharing

## Expected behaviour 🤔
The max number order for a program emergency order type should be enforced. The number of max orders for normal order type is enforced for a internal order by not hiding the period for which the maximum number of orders has been reached
## How to Reproduce 🔨

Have a datafile with programs setup
1. In OG Go to 'Item > Item master list'
2. Click on a 'program' and setup the max order of emergency order type to 2
3. Sync with omSupply
4. on omSupply create more than 2 emergency orders for the same program and period
5. See error. You can.

## Your environment 🌱

- Legacy mSupply Central Server Version: v8.00 dev40

- Platform: Android

- Database type: SQLite3

fergie-nz:
Fixes #6144 

# 👩🏻‍💻 What does this PR do?

Adds back end check to ensure we don't create program requisitions when the maximum number for that period has been reached.

Also adds front end error handling when attempting to create more than the specified number of 

I think we should spend time adding back end tests to creating program requisitions. Right now there is quite complicated logic with no tests validating. 

Any fixes such as this one are time consuming to deal with. It is also currently unclear if there are other gaps in the logic without tests.

# 🧪 Testing

The way to test this using in app behaviour is to attempt to create duplicate requisitions which combined would exceed the maximum number together. 
This requires 2 browsers:

- [ ] Create a program active on a site with an order type of maximum orders of 1
- [ ] Log in to OMS on two separate browsers
- [ ] Open the UI to create a requisition of this order type for a given period
- [ ] Create 2 requisitions at the same time from the 2 different browsers
- [ ] See error prompt

# 📃 Documentation


roxy-dao (changes requested):
Thanks for the changes Ferg. I've given it a quick whirl. Don't wanna comment on code yet until this is fixed..

As you can see my max orders for this order type is 2~ but I have managed to make 4 orders for this order type both through having two instances opened + directly calling endpoint 😭 
Screenshot 2025-01-22 at 12 13 32 PM

Screenshot 2025-01-22 at 12 10 41 PM

fergie-nz:
> Thanks for the changes Ferg. I've given it a quick whirl. Don't wanna comment on code yet until this is fixed..
> 
> As you can see my max orders for this order type is 2~ but I have managed to make 4 orders for this order type both through having two instances opened + directly calling endpoint 😭 Screenshot 2025-01-22 at 12 13 32 PM
> 
> Screenshot 2025-01-22 at 12 10 41 PM

Oh no! Thanks Roxy,

I saw a gap in the logic where you would be able to create one more when existing = max, but not 4.

I'm getting this error response when I try to create more. Are you still able to recreate making many more than max?

server/service/src/requisition/common.rs
```diff
@@ -97,6 +98,42 @@ pub fn check_emergency_order_within_max_items_limit(
     ))
 }
 
+pub struct CheckExceededOrdersForPeriod<'a> {
+    pub connection: &'a StorageConnection,
```
roxy-dao, line 102: As mentioned in another PR, please move this out of the struct

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -129,6 +132,18 @@ fn validate(
         return Err(OutError::SupplierNotValid);
     }
 
+    if check_exceeded_max_orders_for_period(CheckExceededOrdersForPeriod {
+        connection,
+        program_id: &program_setting.program_requisition_settings.program_row.id,
+        period_id: &input.period_id,
+        program_order_type_id: &input.program_order_type_id,
+        max_orders_per_period: i64::from(order_type.order_type.max_order_per_period),
+        requisition_type: RequisitionType::Response,
```
roxy-dao, line 141: This is the wrong type (:

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -129,6 +132,18 @@ fn validate(
...
+    })? == true
```
roxy-dao, line 142: Doesn't need == true. The function is already returning a bool, so if true return error (:

roxy-dao (changes requested):
Thanks for the changes Ferg! Frontend changes look good to me, but this still doesn't work for request requisitions because you are passing in the wrong type~. 

Have made some suggestions [here](https://github.com/msupply-foundation/open-msupply/compare/6144-add-check-for-exceeding-max-orders...6144-example) with comments below (:

fergie-nz:
That is awesome, thanks @roxy-dao. Have merged your branch with one change to checking logic to include current is equal to max.

server/service/src/requisition/common.rs
```diff
@@ -124,7 +124,7 @@ pub fn check_exceeded_max_orders_for_period(
 
             let current_orders = RequisitionRepository::new(connection).count(Some(filter))?;
 
-            Ok(current_orders > input.max_orders_per_period)
+            Ok(current_orders >= input.max_orders_per_period)
```
roxy-dao, line 127: If current order is equal or greater than max then it will return true and will prompt error? Or am I missing something here?

fergie-nz: Current orders is how many orders already exist in the DB. So being able to create one more would exceed the maximum.

roxy-dao: Ah right~ makes sense

roxy-dao (approved):
Cool~ thanks for the changes Ferg!

PR #6229, opened 2025-01-21
6192 Remove mislabelled column

### LINKED ISSUE #6192: Wrong column header for target Stock in requisition report
## What went wrong? 😲
The requisition report has "Batch" has column header but the data seems to be the Target stock number for the internal order and correspond to nothing I can see in the requisition but I assume it is also for the Target stock 

Internal order
https://drive.google.com/file/d/1P3qsk8qg_0ff1QYOW77BYrchwbiCHTZA/view?usp=sharing

Requisition
https://drive.google.com/file/d/1R_scjTUdrhFNUI-C7Mw6qF5JoXA2nUlJ/view?usp=sharing

## Expected behaviour 🤔
I believe it should be named Target Stock for internal order and I am not sure what for the requisition
## How to Reproduce 🔨

1. Open an internal order or requisition
2. Click print
3. Open the report

## Your environment 🌱

- Legacy mSupply Central Server Version: v8.00dev43

- Platform: Open mSupply desktop on Windows 11 and Android

- Database type:  SQLite3

roxy-dao:
Fixes #6192

# 👩🏻‍💻 What does this PR do?
Mislabelled column in requisition report `batch` but calculation was for `Stock to order`

# 🧪 Testing

- [ ] Open IO report 
- [ ] Batch column should now say MOS (months of stock)

# 📃 Documentation

  1.
  2.


server/reports/requisition/2_5_0/src/template.html
```diff
@@ -61,7 +62,7 @@
         <td class="item_name" style="width: 25%">{{ line.item.name }}</td>
         <td class="quantity" style="width: 8%">{{ line.itemStats.availableStockOnHand }}</td>
         <td class="pack" style="width: 10%">{{ line.itemStats.averageMonthlyConsumption | round(precision=2) }}</td>
-        <td class="batch" style="width: 8%">
+        <td class="stock_to_order" style="width: 8%">
           {{ data.requisition.maxMonthsOfStock * line.itemStats.averageMonthlyConsumption }}
         </td>
```
lache-melvin, line 67: Is it supposed to be obvious what "stock to order" means? This looks like its calculating something more like max stock?

lache-melvin: also should the column be rounding? 

Screenshot 2025-01-22 at 1 43 13 PM

lache-melvin (approved):
If stock to order is definitely the label, then lgtm. doesn't really make sense to me though?

server/reports/requisition/2_5_0/src/template.html
roxy-dao: Actually it doesn't make sense because it's basically suggested but without taking away stock on hand.. I'm not sure what the original intention of this report was 🤔 it's quite an old report... maybe we should just remove the column?

lache-melvin: Probs.. whats OG report do?

@mark-prins ?

roxy-dao: OG doesn't have

lache-melvin: ah... i reckon take it out then - all the other important columns are there?

roxy-dao: yeah

roxy-dao: I've removed, and added MOS to match with OG report

server/reports/requisition/2_5_0/src/template.html
```diff
@@ -62,13 +65,11 @@
         <td class="item_name" style="width: 25%">{{ line.item.name }}</td>
         <td class="quantity" style="width: 8%">{{ line.itemStats.availableStockOnHand }}</td>
         <td class="pack" style="width: 10%">{{ line.itemStats.averageMonthlyConsumption | round(precision=2) }}</td>
-        <td class="stock_to_order" style="width: 8%">
-          {{ data.requisition.maxMonthsOfStock * line.itemStats.averageMonthlyConsumption }}
-        </td>
+        <td class="mos" style="width: 8%">{{ line.itemStats.monthsOfStockOnHand }}</td>
```
lache-melvin, line 68: Nice- can we just round so it is consistent w AMC column 🙏 

Screenshot 2025-01-22 at 4 28 04 PM

And maybe suggested quan should be right aligned hehe

lache-melvin (changes requested):

server/reports/requisition/2_5_0/src/requisition.css
```diff
@@ -25,7 +25,7 @@ body {
     text-align: right;
 }
 
-.line_number, .quantity, .pack, .mos, .expiry, .sell_price, .cost_price, .total_quantity, .total_extension {
+.columns {
```
lache-melvin, line 28: Nit... `.number-column` ?

lache-melvin (approved):
Thanks for changes 🙏

Issue #4213, opened 2024-06-20
Numeric text input should support multiple `0`s as incomplete value


lache-melvin:
## Is your feature request related to a problem? Please describe 👀

I want to change from 50,000 to 40,000, by back-spacing the 5, and typing 4. However, multiple `0`s are getting reformatted to just `0`:

https://github.com/msupply-foundation/open-msupply/assets/55115239/18954563-a383-4484-9310-5c3410078c4f

## Describe the solution you'd like 🎁

While the input is in focus, support many `0`s as a incomplete value

CarlosNZ:
Wow, that's a new one. I've never seen anyone edit numbers like that before!

But fair enough I guess, it should be allowed.

lache-melvin:
> Wow, that's a new one. I've never seen anyone edit numbers like that before!

Oh hahaha is that just me 🤣

CarlosNZ:
> > Wow, that's a new one. I've never seen anyone edit numbers like that before!
> 
> Oh hahaha is that just me 🤣

Personally, I would just highlight the 5 then type 4 (which works), but each to their own 😁

andreievg:
Triage, need to refine, leaving it out of milestone. We've had quite a lot of issues regarding number input, i think it's worth investigating why

mark-prins:
Two more aspects of the numeric input behaving strangely:
* with 2dp specified, enter [any number].[any number] and press `5`. keep pressing `5`: the number goes from x.y5 to x.y6 and then reverts to x.y5 and starts again
* with decimal places, enter a number with decimals. have the cursor at the end of the input, press the up key: this increments the integer component regardless of where the cursor is. The natural behaviour is for the closest number to the cursor to be incremented / decremented

andreievg:
If can be done easily go for it, otherwise inconclusive.

andreievg:
Also want to validate at some point how important and urgent this is, i think highest deployment is 1.7, which doesn't have the refactor of number input

jmbrunskill:
Related: If there;s a default value of 0, and you click to the left of it on Android, the 0 is kept at the cursor is at the begining of the input, this means you clikc and type your number say `2`, but the input shows `20`. Feels broken in R&R forms.

lache-melvin:
More things to review:

- Need a `clearable` NumericTextCell (currently it defaults to 0) (See #5288)
[...]

PR #6232, opened 2025-01-21
fix internal order value after backspace

### LINKED ISSUE #6185: Doesn't let you enter value when you click on backspace
## What went wrong? 😲

While creating Internal order in android oms (Happens in browser as well), after adding an item it will show`NAN` when you click on backspace and doesn't let you enter value in Requested quantity field.
Video: [HERE](https://drive.google.com/drive/folders/1agIyorH5S0fX08UCnQyq0qRtN3zGN6yv) 

## How to Reproduce 🔨

1. Go Replenishment -> Internal Orders
2. Click on `New Order`
3. Create new order and add an item
4. add requested quantity and click on backspace
5. Now you will see NAN and can't delete it or add value

## Your environment 🌱

- Open mSupply Version: `2.5.00-RC4`

- Platform: Android

- Database type:

aimee-mcneil-melville:
Fixes #6185 

# 👩🏻‍💻 What does this PR do?

When backspacing a value in Internal Order -> New Item -> Requested Quantity, the empty value populates as 'NaN'
Added a NaN check, if the value is not a number it will show '0'. The user will then be able to type in the field with the new number

# 🧪 Testing

- [ ] Go Replenishment -> Internal Orders
- [ ] Click on New Order
- [ ] Create new order and add an item
- [ ] add requested quantity and click on backspace
- [ ] See '0' and that you can type into the field with a new value

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
@@ -259,7 +259,11 @@ export const RequestLineEdit = ({
                   Input={
                     <NumericTextInput
                       width={INPUT_WIDTH}
-                      value={Math.ceil(draft?.requestedQuantity)}
+                      value={
+                        isNaN(Number(draft?.requestedQuantity))
+                          ? 0
+                          : Math.ceil(draft?.requestedQuantity)
+                      }
                       disabled={isPacks}
                       onChange={value => {
                         if (draft?.suggestedQuantity === value) {
@@ -320,7 +324,8 @@ export const RequestLineEdit = ({
                         decimalLimit={2}
                         width={100}
                         onChange={value => {
-                          const newValue = (value ?? 0) * (draft?.defaultPackSize ?? 0);
+                          const newValue =
+                            (value ?? 0) * (draft?.defaultPackSize ?? 0);
                           if (draft?.suggestedQuantity === newValue) {
                             update({
                               requestedQuantity: newValue,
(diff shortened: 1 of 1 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
```diff
@@ -259,7 +259,11 @@ export const RequestLineEdit = ({
                   Input={
                     <NumericTextInput
                       width={INPUT_WIDTH}
-                      value={Math.ceil(draft?.requestedQuantity)}
+                      value={
+                        isNaN(Number(draft?.requestedQuantity))
+                          ? 0
+                          : Math.ceil(draft?.requestedQuantity)
+                      }
```
lache-melvin, line 266: This is how we've done it elsewhere - preventing the storing of the `NaN` value on the draft at all, rather than then having to handle it anywhere we might want to use the draft.requestedQuantity:

https://github.com/msupply-foundation/open-msupply/blob/62ce998db4fa0e5788bca0b4747bceca0cf4f600/client/packages/requisitions/src/RnRForms/DetailView/RnRFormLine.tsx#L356-L358

Will hopefully get cleaned up as part of [NumericTextInput fixes](https://github.com/msupply-foundation/open-msupply/issues/4213#issuecomment-2581450867) coming soon, but I think best to update this to match how we are handling everywhere else 🙏

lache-melvin (changes requested):

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
aimee-mcneil-melville: Have updated the solution :)

lache-melvin (approved):
lovely thank you!

PR #6237, opened 2025-01-22
6085 use path in report building cli

### LINKED ISSUE #6085: Reports CLI for upserting reports json is not using pathing package. Doesn't work on Windows.
## What went wrong? 😲

Report upsert CLI uses hard coded '/' rather than a pathing package.
This causes issues when running on Windows (or is it mac? unsure.)
Adding a pathing package as per elsewhere in the codebase should fix this!

Pathing package needs to be added in the BuildStandardReports action and UpsertReportsJson action, and throughout the cli utility functions which use hard coded '/' for pathing.

## How to Reproduce 🔨

1. Failure is visible when attempting to use the UpsertReportsJson command on a windows machine
See [the report builder readme)[https://github.com/msupply-foundation/open-msupply/blob/develop/server/report_builder/README.md] for details on using the CLI.

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #6085 

# 👩🏻‍💻 What does this PR do?

Changes cli to use pathing package rather than hard coded strings

I've attached 

# 🧪 Testing

- [ ] Attempt to upsert attached json to server using CLI on mac environment
- [ ] Attempt to upsert attached json to server using CLI on windows environment

# 📃 Documentation


server/cli/src/cli.rs
```diff
@@ -388,38 +388,46 @@ async fn main() -> anyhow::Result<()> {
...
+                let report_versions: Vec<PathBuf> = fs::read_dir(&name_dir)?
                     .filter_map(|r| r.ok())
                     .map(|e| e.path())
                     .filter(|p| p.is_dir())
-                    .map(|p| p.into_os_string().into_string().unwrap())
+                    .map(|p| p)
                     .collect();
 
-                let (_, name) = name_dir.rsplit_once('/').unwrap();
+                let parent = name_dir
```
andreievg, line 411: what is the reason for this parent and the `name` ? Is it just for logs ?

server/cli/src/cli.rs
```diff
@@ -482,7 +490,7 @@ async fn main() -> anyhow::Result<()> {
                     let form_schema_json = match (arguments_path, arguments_ui_path) {
                         (Some(_), None) | (None, Some(_)) => {
                             return Err(anyhow!(format!(
-                                "When arguments_path is specified arguments_ui_path must be specified too in report {:?} {:?}", name, version_dir
+                                "When arguments_path is specified arguments_ui_path must also be specified in report {:?} {:?}", name, version_dir
```
andreievg, line 493: I don't think you need both anyhow! and format

server/cli/src/cli.rs
fergie-nz: And also the name of the name.json generated output file for that report. 
FYI this is not the name used on the front end as this is extracted from the manifest file.

Issue #5675, opened 2024-12-02
CIV 2.5: Link a manual Inbound Shipment to an Internal Order


roxy-dao:
## Step 1: 
Amend the `InsertInboundShipment` input to accept a `requisition_id` which should only be inputted if the store has the store preference: `Manually link Internal Order to Inbound Shipment`

## Step 2: 
- Note: This window will only show if the store has the preference `Manually link Internal Order to Inbound Shipment`
- Requisition endpoint to get all finalised Internal Orders
	- Filter by: 
		- Type: Request
		- Status: Sent
		- OtherPartyId: Selected from add customer (From `New Shipment` -> `Clicked supplier`)
- If the store has the above preference and the requisition endpoint returns some values then show below table for user to select which Internal Order they want to link to the Inbound Shipment.
![Screenshot 2024-12-03 at 8 13 11 AM](https://github.com/user-attachments/assets/e645c43a-e341-4c11-8749-1af5aabe1d1e)

## Step 3
- Add an `Add from Internal Order` option to the `Add item` button 
![Screenshot 2024-12-03 at 8 11 14 AM](https://github.com/user-attachments/assets/829e81b6-6c66-4aee-b290-7254b0dc725f)
- Select items from the Internal Order to add to the Inbound Shipment
![Screenshot 2024-12-03 at 8 11 45 AM](https://github.com/user-attachments/assets/b28a2075-0147-4f46-8a22-6ebc1004ecd7)
- When an item is added, the invoice line should be populated with information from the Internal Order line

### Moneyworks Jobcode 🧰
OMS:CIDV

PR #6247, opened 2025-01-22
5675.1 pass requisition id on insert inbound

roxy-dao:
Part 1 of #5675

# 👩🏻‍💻 What does this PR do?
Attempting to break down the issue for easier review. This PR allows endpoint to pass in requisition_id and should throw an error if you try to link requisition to inbound shipment without `manually_link_internal_order_to_inbound_shipment` store preference.

# 🧪 Testing

- [ ] Don't turn on `manually_link_internal_order_to_inbound_shipment` store pref
- [ ] Use graphiql to create a new inbound shipment and link to an internal order by providing requisition_id
- [ ] Should have error
- [ ] Now turn on pref and try again
- [ ] Should be okay to link 

# 📃 Documentation

  1.
  2.


fergie-nz (approved):
Nice Roxy looks awesome. Thank you for splitting up for easeful reviewing.

I was wondering when it is good to add structured errors for more descriptive handling? I guess this would typically never become visible on the front end in most customer use cases.

roxy-dao:
> Nice Roxy looks awesome. Thank you for splitting up for easeful reviewing.
> 
> I was wondering when it is good to add structured errors for more descriptive handling? I guess this would typically never become visible on the front end in most customer use cases.

Yeah, the errors I've added have very low case of happening since frontend should show all we need to not ever get those errors 👀 unless something goes very wrong, so they won't be structured errors

PR #6249, opened 2025-01-22
Don't convert dob to naivedate

### LINKED ISSUE #5869: Unable to select certain dates of births when editing a patient
## What went wrong? 😲

Noticed this behaviour with a date picker - unable to select certain dates of births for a patient. And when selecting a date, it seems to select the day before?

https://github.com/user-attachments/assets/d4baee0d-e830-47c3-8351-87b7c352c791

Unsure if related to #5438 - also have a suspicion it might be something to do with the patient being exactly one month old?

## Expected behaviour 🤔

Should be able to select any date as the date of birth

## How to Reproduce 🔨

2. Create a new patient or edit an existing one
3. Play around with their date of birth - try setting it to be 29 days, 30 days, 1 month and 1 day, etc. Save and refresh, go in and make changes - see if you can run into the issue above

## Your environment 🌱

- Open mSupply Version: V2.5.0-reports

- Database type: SQLite3

roxy-dao:
Fixes #5869

# 👩🏻‍💻 What does this PR do?
Don't convert selected value to naivedate

- [ ] Go to patient
- [ ] Try selecting a variety of dates as the birth date

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
+++ b/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
@@ -61,6 +61,7 @@ const UIComponent = (props: ControlProps) => {
 
   useEffect(() => {
     if (!data) return;
+    const dob = DateUtils.getNaiveDate(data.dateOfBirth);
     setDoB(dob);
     if (dob === null) {
       setAge(undefined);
--- a/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
+++ b/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
@@ -84,7 +84,7 @@ const UIComponent = (props: ControlProps) => {
         <Box display="flex" alignItems="center" gap={FORM_GAP} width="100%">
           <BaseDatePickerInput
             // undefined is displayed as "now" and null as unset
-            value={DateUtils.getNaiveDate(dob)}
+            value={dob}
             onChange={onChangeDoB}
             format="P"
             width={135}
--- a/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
+++ b/client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
@@ -61,13 +61,13 @@ const UIComponent = (props: ControlProps) => {
 
   useEffect(() => {
     if (!data) return;
-    const dob = DateUtils.getNaiveDate(data.dateOfBirth);
-    setDoB(dob);
-    if (dob === null) {
+    const naiveDoB = DateUtils.getNaiveDate(data.dateOfBirth);
+    setDoB(naiveDoB);
+    if (naiveDoB === null) {
       setAge(undefined);
       return;
     }
-    setAge(DateUtils.age(dob));
+    setAge(DateUtils.age(naiveDoB));
   }, [data]);
 
   if (!props.visible) {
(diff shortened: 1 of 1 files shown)


mark-prins (approved):
❤️  I'm a little nervous.. this is a minefield of unintended consequences.. but let's do it!

CarlosNZ:
Didn't we just change this to this fairly recently? Have we tested that this doesn't display the wrong date at certain times of the day?

roxy-dao:
> Didn't we just change this to this fairly recently? Have we tested that this doesn't display the wrong date at certain times of the day?

Yeah I tested it and converting to naive date does change the date to day before.. but also don't see why we are converting to naive date? DOB is saved as a date and no matter what time zone you are in... if you are born on 1st Jan then you are born on first Jan... or is this supposed to change to 31st December when you change timezones?

CarlosNZ:
> if you are born on 1st Jan then you are born on first Jan... or is this supposed to change to 31st December when you change timezones?

No, you are correct, but that's what "naive date" is supposed to be -- a date that doesn't care about what timezone it's in. Is that correct? Changing this to naive date was the fix for the DOB jumping back a day before, so I'm struggling to understanding how changing it back would help.

I'll give this a quick test after lunch just to convince myself either way.

roxy-dao:
> > if you are born on 1st Jan then you are born on first Jan... or is this supposed to change to 31st December when you change timezones?
> 
> No, you are correct, but that's what "naive date" is supposed to be -- a date that doesn't care about what timezone it's in. Is that correct? Changing this to naive date was the fix for the DOB jumping back a day before, so I'm struggling to understanding how changing it back would help.
> 
> I'll give this a quick test after lunch just to convince myself either way.

But it was jumping with the naive date 🤔

CarlosNZ:
> But it was jumping with the naive date 🤔

It might be because the double conversion. Probably only needs naive date for input/save, and then can just display that in local time.

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -61,7 +61,6 @@ const UIComponent = (props: ControlProps) => {
 
   useEffect(() => {
     if (!data) return;
-    const dob = DateUtils.getNaiveDate(data.dateOfBirth);
     setDoB(dob);
```
CarlosNZ, line 64: I think you still need this one. 

Firstly,  you're just setting `dob` to itself here now, if you remove the locally-defined `dob` 😆 -- that's why you're not seeing the "error", but you'll notice if you save and refresh, the date is gone altogether!

I would put the original back, but change the locally scoped `dob` to something else to avoid this confusion.

This seems to work for me:
```ts
  useEffect(() => {
    if (!data) return;
    const newDob = DateUtils.getNaiveDate(data.dateOfBirth);
    setDoB(newDob);
    if (newDob === null) {
      setAge(undefined);
      return;
    }
    setAge(DateUtils.age(newDob));
  }, [data]);
```

To convince yourself it's correct, do the following:
- set your system timezone to something that is still "yesterday" relative to us (I used "New York", but that'll only work for the next 4 hours)
- adjust the DOB and save, and refresh to check that is has saved and is correct
- change your system back to NZ timezone
- refresh the page, the DOB should show the same date, not have jumped forward a day.

roxy-dao: Yeah sorry, was supposed to revert. But then I did something else and have just come back from lunch

CarlosNZ (approved):
Don't love the name (`localDob` is pretty unclear if you weren't part of our previous convo), but I won't hold it up. 

Thanks 🙏

PR #6286, opened 2025-01-27
fix confirm on leaving with multiple blockers

### LINKED ISSUE #6321: No 'You have unsaved changes' warning when navigating away from Vaccination encounters
## What went wrong? 😲

I can navigate away from the encounter without a prompt even when I have unsaved changes!

This is due to two `useBlocker` calls - one from `DetailTabs` and one from `useJsonForms`.

useBlocker can only have one active instance at a time, so naturally neither are working.

## Expected behaviour 🤔

Vaccination encounters still have the navigation warnings if I have unsaved changes

Update the implementation so that we can register multiple blockers in as many places as we need, and this can inform one useBlocker call at the base of the app.

## How to Reproduce 🔨

1. Go to a vaccination encounter
2. Edit some JSON form data, e.g. visit type, encounter start time...
3. Save button should be enabled - you have changes that need to be saved - do NOT save the form
4. Try to navigate away
5. You are not blocked from navigating/warned that you have unsaved changes!

## Your environment 🌱

- Open mSupply Version: 2.5

- Platform:

- Database type:

lache-melvin:
Fixes #6321

# 👩🏻‍💻 What does this PR do?

useBlocker can only have one active instance at a time. In encounters, there were two blockers (one from JSON form, one from `DetailTabs`) - so neither were working. You could edit encounter, then navigate away without saving with no warning.

This updates the implementation so that we can register multiple blockers in as many places as we need, and this informs the one `useBlocker` call at the base of the app. 

Actually ended up fixing the detail tabs so that it doesn't use the confirm on leaving hook - see comment - but will have the same multiple blockers rendered problem with my next vax card PR. Just didn't want this bloat in that PR 🙏 

# 🧪 Testing

- [ ] Setup for vaccine card - see testing setup: #4642 
- [ ] Create/go to a vaccination encounter
- [ ] Make a change e.g. set visit type or start time
- [ ] Try to navigate away - you get the are you sure confirmation modal
- [ ] Also test it still works in the other places:
  - [ ] Refreshing page in inbound/outbound shipment edit modals
  - [ ] R&R form
  - [ ] Editing CCE
  - [ ] Editing prescription
    - [ ] Check you still _can_ change between items in prescription, just not to a different page

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
@@ -7,7 +7,6 @@ import {
   RouteBuilder,
   useBreadcrumbs,
   useConfirmOnLeaving,
-  useDirtyCheck,
   useNavigate,
   useParams,
 } from '@openmsupply-client/common';
@@ -46,8 +45,6 @@ export const PrescriptionLineEditView = () => {
     // the list before scrolling to it
     setTimeout(() => scrollRef.current?.scrollIntoView(), 100);
 
-  const { isDirty, setIsDirty } = useDirtyCheck();
-
   const lines =
     data?.lines.nodes.sort((a, b) => a.id.localeCompare(b.id)) ?? [];
 
@@ -82,8 +79,9 @@ export const PrescriptionLineEditView = () => {
     });
   }, [currentItem]);
 
-  useConfirmOnLeaving(
-    isDirty,
+  // tODO: does the state turn up right lol
+  const { isDirty, setIsDirty } = useConfirmOnLeaving(
+    'prescription-line-edit',
     // Need a custom checking method here, as we don't want to warn user when
     // switching to a different item within this page
     (current, next) => {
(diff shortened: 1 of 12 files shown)


client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -82,8 +79,9 @@ export const PrescriptionLineEditView = () => {
     });
   }, [currentItem]);
 
-  useConfirmOnLeaving(
-    isDirty,
+  // tODO: does the state turn up right lol
```
lache-melvin, line 82: chris has brought it to my attention that it indeed does not 😭

lache-melvin: Fixed in f0f0cce153, thanks Chris 🙏

Chris-Petty (approved):
Yisss

PR #6293, opened 2025-01-28
6261 mark as visited on change

### LINKED ISSUE #6261: When user tries to leave the encounter after interacting, prompt to mark as visited
Show a confirmation modal on leaving the encounter page, if changes have been made.

Add dirty state to encounter, handling both changes to the `Details` and `Vaccination` tabs. 

If data on either tab is changed, the user should be prompted to  mark the encounter as Visited as they leave.

lache-melvin:
Fixes #6261

# 👩🏻‍💻 What does this PR do?

Prompt user to mark as visited on leaving encounter after giving some vaccinations:
Screenshot 2025-01-28 at 2 37 08 PM

I put this in only if user has made some changes (per requirement: `When user tries to leave the encounter after interacting, prompt to mark as visited` ... so if user opens and closes a pending encounter, there is no prompt.) However, do you think it would be better if after encounter start date and still pending, prompt to mark visited? Maybe its a later clean up problem..

Note that "mark as visited" does not redirect user - next step is to automatically open "schedule next encounter" modal after confirming visited, so won't redirect here 🙏 

# 🧪 Testing

- [ ] Open/create vaccination encounter
- [ ] Give a vaccination
- [ ] Try to navigate to another page
- [ ] See prompt
- [ ] Select leave as pending - you are redirected to the next page
- [ ] Go back to encounter
- [ ] Edit some encounter data (e.g. visit type, start time)
- [ ] Click save in the footer (only save, don't save as visited)
- [ ] Try to navigate to another page
- [ ] See prompt
- [ ] Select leave as pending - you are redirected to the next page
- [ ] Go back to encounter
- [ ] Give another vaccination (or edit the previous one somehow)
- [ ] Try to navigate to another page
- [ ] See prompt
- [ ] Select `Mark as Visited` - you stay on encounter page, with "saved successfully" toast message and status updated

# 📃 Documentation

  1.
  2.


client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
```diff
@@ -1,29 +1,26 @@
...
 
 /**
  * useConfirmOnLeaving is a hook that will prompt the user if they try to
  * navigate away from, or refresh the page, when there are unsaved changes.
  */
-export const useConfirmOnLeaving = (
-  key: string,
-  customCheck?: (currentLocation: Location, nextLocation: Location) => boolean
-) => {
+export const useConfirmOnLeaving = (key: string, options?: BlockerOptions) => {
```
lache-melvin, line 11: Heh sorry more changes to this hook - updated from passing in just a `customCheck` to `options`, so we can also provide a custom confirmation modal to open instead of the generic `are you sure`

client/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -17,6 +17,7 @@ interface ConfirmationModalProps {
   info?: string | undefined;
   iconType?: 'alert' | 'info' | 'help';
   buttonLabel?: string | undefined;
+  cancelButtonLabel?: string | undefined;
```
lache-melvin, line 20: also had to add support for custom cancel button label here

client/packages/system/src/Encounter/DetailView/DetailView.tsx
```diff
@@ -269,6 +301,12 @@ export const DetailView: FC = () => {
 
   if (!isSuccess && !isError) return <DetailViewSkeleton />;
 
+  const suggestSaveWithStatusVisited = encounter
+    ? new Date(encounter.startDatetime).getTime() < Date.now() &&
+      encounter.status === EncounterNodeStatus.Pending &&
+      dataStatus === EncounterNodeStatus.Pending
+    : false;
```
lache-melvin, line 309: this is just moved down

client/packages/system/src/Encounter/DetailView/StatusChangeButton.tsx
```diff
@@ -21,9 +21,18 @@ export const StatusChangeButton = ({
   const onStatusClick = () => {
     if (!selectedOption.value || selectedOption.value === currentStatus) return;
     onSave(selectedOption.value);
-    setSelectedOption(selected => ({ ...selected, isDisabled: true }));
   };
 
+  useEffect(() => {
+    setSelectedOption(selected =>
```
lache-melvin, line 27: updated to reset the selected option state whenever the status changes, rather than on button click, because status can also be updated from other places

Chris-Petty (approved):
Interestingly if you edit a field such as the date at the top first, then add a vaccination, you get the dirty form state prompt instead of mark as visited prompt when navigating away

image

But if you press cancel instead of navigating it shows the mark it as visited prompt. Not sure which is more important nor that bothered by it!

Otherwise works all good!

Is it normal that you can do many of the vaccines all in the same encounter? I'd have thought that they'd have a different encounter for each vaccination in the schedule. I guess allows back entry of data more easily though. Heck you can even mark it as cancelled and continue editing everything haha.

lache-melvin:
> But if you press cancel instead of navigating it shows the mark it as visited prompt. Not sure which is more important nor that bothered by it!

I'm not able to replicate this? Hoping whatever is happening it's not too high likelihood for now, can fix up in a clean up issue 🙏 

> Is it normal that you can do many of the vaccines all in the same encounter? 

Yeah you might be giving like the first dose of 3 different courses in the one encounter 🥇

Chris-Petty:
> Yeah you might be giving like the first dose of 3 different courses in the one encounter 🥇

Oh but I was giving doses 1, 2, 3 and 4 in encounter 😆 Missed your first 3 doses? AH JUST TAKE THEM ALL NOW

PR #6303, opened 2025-01-28
update icon color to white on orange buttons

### LINKED ISSUE #6005: Sync icon is not updated to white colour, only visible when hovered over the button
## What went wrong? 😲

https://github.com/user-attachments/assets/672343f7-360c-4c4a-b76f-05939c2a4653

## Expected behaviour 🤔

When not hovered over, the icon should match the text and show up as white. Remain orange when hovered over

## How to Reproduce 🔨

1. Upgrade to V2.5.0-RC2
2. Click on Sync which is now a pop-up window and does not redirect to a new page
3. Note only text shows in white and not the icon
4. Hover over the button - now the icon is visible

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC2

- Database type: PostgreSQL

aimee-mcneil-melville:
Fixes #6005

# 👩🏻‍💻 What does this PR do?

Fixes buttons that are orange and not showing icons (without hover). Icons were displaying as orange on orange background, where the icons should be white

I had a look at if there was a quick refactor as the icons colors are now being controlled by two different components, and the icons being in parent components to these. This is probably a separate refactor issue though

Buttons that had an orange background, also had an orange icon which was not visible. Icon is now white

Screenshot 2025-01-28 at 6 29 46 PM

# 🧪 Testing

- [ ] See Sync modal, Sync button should have a white icon
- [ ] See Settings -> Devices -> 'Save' button should be orange with white icon (may need to fill IP address)

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/buttons/standard/ShrinkableBaseButton.tsx
+++ b/client/packages/common/src/ui/components/buttons/standard/ShrinkableBaseButton.tsx
@@ -26,12 +26,13 @@ export const StyledShrinkableBaseButton = styled(StyledBaseButton, {
           marginLeft: 8,
         }
       : {},
-
-    '&.MuiButton-root:not(:hover) .MuiSvgIcon-root': {
-      color:
-        color === 'primary' && !disabled
-          ? translateColor(theme, color)
-          : undefined,
+    '&.MuiButton-outlined': {
+      '&.MuiButton-root:not(:hover) .MuiSvgIcon-root': {
+        color:
+          color === 'primary' && !disabled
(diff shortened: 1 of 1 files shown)


lache-melvin:
> I had a look at if there was a quick refactor as the icons colors are now being controlled by two different components, and the icons being in parent components to these. This is probably a separate refactor issue though

This is a good point - given the plan to move away from MUI, I think what you have here is great for now, how we apply the styles will need to be redone in the near future anyway 🙏

Issue #6315, opened 2025-01-28
Tables/contents on modals need some styling love


CarlosNZ:
They're looking a bit sad right now:

![Image](https://github.com/user-attachments/assets/26e2cb89-7942-417b-af75-fc78254bba21)

The suggested designs were like this, but that was for when it was a normal modal, not the new full-size ones:

![Image](https://github.com/user-attachments/assets/8bee4dea-3b73-4cf0-8dde-79aeacb7653f)

Would be good to get an agreed "look" for these full-size table modals and implement globally.

Tagging @mariyamsupply :)

mariyamsupply:
That's new :)

I suggest we:
- Align the header and the 'close' icon with the table container
- Make 'close' icon blue
- outline the table with a light grey line (to show)

- improve alignment of table headers (prob out of scope for this)

CarlosNZ:
Modals with not much on them (e.g Encounter sections) are looking particularly barren:

![Image](https://github.com/user-attachments/assets/9d33a804-135e-44b7-aafa-e6a7983f25ef)

We may need to keep some of these simpler ones as "regular" modals?

andreievg:
This issue is just for the table formatting as per designs (no extra functionality, like buttons and search and checkbox)

CarlosNZ:
Also, I notice the cutoff condition for full-screen modals has been set to `'(max-height: 850px)'` -- I wonder if this should be something smaller -- we basically have to go full-screen on our Macs to not see the fullscreen modal. And shouldn't it be based on a width rather than a height? @lache-melvin I think you set this -- was there any particular reasoning behind it?

lache-melvin:
Wildly arbitrary, figured we could adjust. I think most tablets are under 800px high, so could go a little lower. I think it should be related to height though, the point for full screen modal is mostly to accomodate tablet keyboard. Maybe the check should actually just check if we're running on android, rather than a breakpoint?

mariyamsupply:
Checking if we're running on android sounds like a good idea.
Other than that for breakpoints:
- width
i'd say go full screen when the width is less than 1024px?
- height
Height wise -  it's sort of amount of content dependant, but maybe less than 650px? 
Would be good to play around. Need to install dev demo data.

mariyamsupply:
Also, thinking whatever we should put heading, close icon and buttons closer to the rest content?

current version: 
(is the content always centered within the height?)

[...]

Issue #6311, opened 2025-01-28
New full-screen modals aren't properly responsive to viewport changes


CarlosNZ:
## What went wrong? 😲

With the new full-window modals, if you resize your browser window while open, the contents doesn't resize. If you refresh/reload the page, and launch it again, it will be correctly laid out, but it won't change with the window changes.

E.g.

Launching modal like this:

![Image](https://github.com/user-attachments/assets/b12d19ce-f8a9-4e05-ac8b-309b306b8a9f)

When I narrow my window, it goes like this:

![Image](https://github.com/user-attachments/assets/7ce969d3-db16-42b4-8c7b-cdf05c468f23)

And then, after a reload:

![Image](https://github.com/user-attachments/assets/88d29de1-6542-42fc-a8c1-b77b32087fa9)

And the, after widening it again:

![Image](https://github.com/user-attachments/assets/abee2766-f260-408d-a320-2d9859e42258)

## Expected behaviour 🤔

The contents should resize with the viewport.

This would affect tablets too, if they rotate their screen while modal is open.

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Open a (full-size) modal
2. Resize your browser
3. Notice the contents doesn't resize until you reload

## Your environment 🌱

- Open mSupply Version: 2.6
- Platform: MacOS, Firefox & Chrome
- Database type: Postgres

lache-melvin:
I think this has always been the case with modals, same thing if you open a wide modal like inbound shipment edit line on big screen (so not fullscreen modal), then make screen narrower. We set modal size when it opens - not sure if that was done to resolve an issue or if resizing would be easy?

CarlosNZ:
> I think this has always been the case with modals, same thing if you open a wide modal like inbound shipment edit line on big screen (so not fullscreen modal), then make screen narrower. 

Yeah, I guess it was, but wasn't so apparent as I wasn't narrowing the screen to less than the modal content very often.

> We set modal size when it opens - not sure if that was done to resolve an issue or if resizing would be easy?

But now that the modal is the full width of the viewport, I would have thought the contents should be able to resize within it just like a normal page. 🤷‍♂️. I'd be keen to investigate why it isn't.

CarlosNZ:
Just quickly -- I notice that the `Dialog` component is rendering an additional  `form` element around the content, which doesn't seem to be the case on the MUI demo page.

[...]

PR #6316, opened 2025-01-28
#5756 Dispensing history modal

### LINKED ISSUE #5756: Dispensing: Display previously prescribed items
NewModal in the Prescription `History`

Opening the tab modal displays a list of previously prescribed items for that patient (Paginated if needed)

image

Acceptance Criteria

- [ ] Able to see the list of previously prescribed items for a patient
- [ ] Table should contain item name, total prescription quantity, directions, prescription date, and clinician
- [ ] All data should be for the current store only (don't need top look up remote data)
- [ ] Doesn't need to have new UI yet (e.g. focus on functionality first)

CarlosNZ:
Fixes #5756

# 👩🏻‍💻 What does this PR do?

This is a very bare-bones initial implementation -- I think it needs quite a few extras to be useful, particularly looks-wise, but I've created a new issue for these new tables in general: https://github.com/msupply-foundation/open-msupply/issues/6315

The modal just uses the existing "usePrescriptionList" query, but filtered for specific patient. It pulls the most recent 20 prescriptions, then organises them into lines, combining multiple batches of the same item within a single prescription. 

Is there anything else we want to add right now?

Please confirm the logic for summing the lines is correct. I'm fairly confident it it now, but would pay to double-check.

# 🧪 Testing

- [ ] Click on an existing prescription or start a new one
- [ ] Click the new "History" button
- [ ] History modal should load
  - [ ] All items should be for this patient
  - [ ] Items should be ordered by most recent at the top
  - [ ] Multiple batches of the same item within a prescription should be combined (check against previous prescriptions)

# 📃 Documentation

Would recommend getting this a bit further along (at least looking good and havinig the "Add to new prescription" option implemented before doing any documentation for this.


client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    rawData.forEach(prescription => {
+      const { clinician, pickedDatetime, lines } = prescription;
+      compiled.push(
+        ...combineCommonLines(lines.nodes as InvoiceLineNode[]).map(line => ({
+          id: line.id,
+          itemName: line.itemName,
+          unitQuantity: line.unitQuantity,
+          directions: line.note ?? '',
+          date: pickedDatetime ? new Date(pickedDatetime) : null,
```
jmbrunskill, line 70:
```suggestion
          date: pickedDatetime ? new Date(pickedDatetime) : new Date(createdDatetime),
```
Not critical but feels like the right thing to do?

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const item = uniqueItems?.[line.item.id];
+    if (item) {
+      item.unitQuantity += line.numberOfPacks * line.packSize;
+    } else {
+      uniqueItems[line.item.id] = {
+        ...line,
+        unitQuantity: line.numberOfPacks * line.packSize,
+      };
+    }
+  });
```
jmbrunskill, line 119:
```suggestion
  lines.forEach(line => {
    let item = uniqueItems?.[line.item.id];
    if (!item) {
          // new item found, default to 0
         item = {
                      ...line,
                      unitQuantity: 0,
                      };
          uniqueItems[line.item.id] = item;
      } 
    
      item.unitQuantity += line.numberOfPacks * line.packSize;    
    }
  });
```

That's probably not quite the right code (haven't tested it), but I prefer this kind of pattern, as the logic for the calculation is in only one place.

In this case I think it's unlikely to change and looks correct to me, so I'm totally fine with what you had, just thought I'd suggest it to see what your thoughts are.

client/packages/invoices/src/Prescriptions/DetailView/DetailView.tsx
```diff
@@ -77,7 +93,16 @@ export const PrescriptionDetailView: FC = () => {
             },
           })}
         >
-          <AppBarButtons onAddItem={onAddItem} />
+          <AppBarButtons onAddItem={onAddItem} onViewHistory={onViewHistory} />
+          {isHistoryOpen && (
```
jmbrunskill, line 97: I feel like I've had issues in the past with this pattern, but can't remember the details. I think it caused a problem in Health Supply Hub where modals weren't opening correctly for some reason.

We don't actually need to have the conditional here right? Does this reduce the render effort or something?

I'm ok to keep it but thought I'd just query it.

client/packages/programs/src/api/operations.generated.ts
jmbrunskill, line 1: Can you try re-generating and if you need to edit save without formatting? This makes for a big diff, and might cause more merge conflicts etc. Other option would be to always format our generated files potentially? But we probably need to all agree if that's our plan...

jmbrunskill (changes requested):
Thanks Carl, Looking good.

BTW: Not sure if that was deliberate, or not but it's not showing as fullscreen modal for me? I'm not worried either way but thought I'd mention as your other issue suggests it is full screen?

Just two things please.

1. Can we add a close button to the modal. (or allow it to close by clicking out?)
Haven't tried this on the tablet but feels like relying on escape doesn't work on tablet,and isn't obvious UX for some users.
The new sync modal has a potential pattern to follow.
image

2. Fix diff for `operations.generated.ts`

Otherwise looks great.

In the original issue there was a request to show `Invoice Number` not a requirement for now, but could be added if you want show it?

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }, [data]);
+
+  return (
+    <TableProvider
+      createStore={createTableStore}
+      queryParamsStore={createQueryParamsStore({
+        initialSortBy: { key: 'expiryDate' },
+      })}
+    >
+      <Modal title={t('heading.history')} width={1125} height={600}>
```
jmbrunskill, line 89: Assuming this width here might be why it's not going full screen?

client/packages/programs/src/api/operations.generated.ts
CarlosNZ: Sorry, think that was left over from Chris' stuff. Hadn't actually regenerated myself. Have reverted: 99dbc6985f

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
CarlosNZ: Yeah, sure. 85fdd182d7

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
CarlosNZ: Yup, I hear what you're saying. The main difficulty I had here was fighting with Typescript to assign a value to an object indexed with a non-literal propery name (hence the separate `item` intermediate variable).

 But I've taken the suggestion and put the unitQuantity calculation first, so ended up with this: 97c51e191b

client/packages/invoices/src/Prescriptions/DetailView/DetailView.tsx
CarlosNZ: No, you're right, it's unecessary -- another one of Chris's leftovers that I didn't really consider :) 7870fbbc4c

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
CarlosNZ: Yes and no. The `width` and `height` properties are actually converted into `minHeight` and `minWidth` in the Modal component, so they're just minimums really. Have reduced the width somewhat now. 301b8b483a

Have made a separate issue for the lack of responsiveness when re-sizing: https://github.com/msupply-foundation/open-msupply/issues/6311

client/packages/invoices/src/Prescriptions/DetailView/History/HistoryModal.tsx
jmbrunskill: Nice looks great

CarlosNZ:
Okay @jmbrunskill, sorted that now. Have added an "OK" button to close the modal: 71d667e900

I've also changed the heading on the modal to "Recently prescribed items", as we probably should make it clear that it's not a *complete* history.

jmbrunskill (approved):
Nice, looking good.

Would be good if we can improve the message if there's no previously prescribed items.
image

Another small issue I just discovered, is that we're showing data from the current prescription in the list, which I think shouldn't be included?
image

Maybe we merge this as it's all working then create a follow up issue for those two items?

Was also thinking it might be nice to show the number of days ago next to the date, which would help the user rather than mentally calculating it, but that might be a question for @mark-prins ?

PR #6337, opened 2025-01-29
6336 Make central record processor generic

### LINKED ISSUE #6336: Refactor central processor to allow new processors to be added

Contact form processor has most of the infrastructure needed to be generically used as a central record processor, but still missing a little bit of abstraction

Make CentralRecordProcessor generic so that it can be used for processing plugins (to bind them to PluginProvider

andreievg:
Fixes #6336 

# 👩🏻‍💻 What does this PR do?

Allow for central record processor to be generic. 
Expand central record processor trait to have extra methods:
change_log_table_names
changelogs_filter
cursor_type
should_run

Processor trigger accepts type of central record processor

Relocated processor files to be more structured (now that there is central processor, it lives in parent dir and has contact form as child)

# 🧪 Testing

- [ ] Contact form emails still work


server/service/src/processors/central_records/mod.rs
```diff
@@ -27,21 +26,39 @@ pub(crate) enum ProcessCentralRecordsError {
...
+    }
 
     let ctx = service_provider
         .basic_context()
         .map_err(Error::DatabaseError)?;
-
     let changelog_repo = ChangelogRepository::new(&ctx.connection);
-    let cursor_controller = CursorController::new(KeyType::ContactFormProcessorCursor);
+
+    let cursor_controller = CursorController::new(processor.cursor_type());
```
andreievg, line 58: processor now defines cursor_type and changelog_filter

server/service/src/processors/mod.rs
```diff
@@ -29,14 +30,14 @@ const CHANNEL_BUFFER_SIZE: usize = 30;
 pub struct ProcessorsTrigger {
     requisition_transfer: Sender<()>,
     invoice_transfer: Sender<()>,
-    contact_form_to_email: Sender<()>,
+    central_record: Sender<CentralRecordProcessorType>,
```
andreievg, line 33: When triggered we specify which central record type should be processed

server/service/src/processors/central_records/mod.rs
```diff
@@ -27,21 +26,39 @@ pub(crate) enum ProcessCentralRecordsError {
 
 const CHANGELOG_BATCH_SIZE: u32 = 20;
 
+#[derive(Clone)]
+pub enum CentralRecordProcessorType {
```
andreievg, line 30: Type is used because processor will be triggered in generic way, and need to specify which records to process

server/service/src/processors/central_records/mod.rs
```diff
@@ -74,8 +88,27 @@ pub(crate) fn process_central_records(
...
+
+    /// Default to empty array in case chanelogs_filter is manually implemented
+    fn change_log_table_names(&self) -> Vec<ChangelogTableName> {
+        Vec::new()
+    }
+
+    /// Extra check to see if processor should trigger, like if it's central for contact form email
+    fn should_run(&self) -> bool {
+        true
+    }
```
lache-melvin, line 109: If it's the central server processor, shouldn't it always only run on central server?

server/service/src/processors/central_records/mod.rs
```diff
@@ -27,21 +26,39 @@ pub(crate) enum ProcessCentralRecordsError {
...
 
+#[derive(Clone)]
+pub enum CentralRecordProcessorType {
+    ContactFormEmail,
+}
+
+impl CentralRecordProcessorType {
+    pub(super) fn get_processor(&self) -> Box<dyn CentralServerProcessor> {
+        match self {
+            CentralRecordProcessorType::ContactFormEmail => Box::new(QueueContactEmailProcessor),
```
lache-melvin, line 37: Should we keep the capability to have multiple processors for the same record type, like we have for invoices?

lache-melvin (changes requested):
Thanks for this Andrei, knew this needed a little cleaning up 😁 

This feels now like it's just a Processor - we could consolidate the requisition and invoice processors into this one (with should_run always true) - or it should actually just be central only?

server/service/src/processors/central_records/mod.rs
andreievg: I feel like the transfer processors are a bit different, because:

They prepare data in a particular shape before processors run (maybe can be done with some generic, but a bit complex for now ?)

I think we can use the cursor increase function of this generic processors also for transfers, you are right ! I'll have a little go at this now, won't spend too long on it though

andreievg: > I think we can use the cursor increase function of this generic processors also for transfers, you are right ! I'll have a little go at this now, won't spend too long on it though

Actually I am getting a bit worried with 'order of operation', so will keep existing transfer logic as is for now

andreievg:
@lache-melvin thanks for your review, here is the change for the renaming: c666f7d04e6b43c7d8bf2dd0a8d9dc65a3d7e48e

I think we can use this 'Processor' in requisition and invoice transfers, but thought to not change too much and was worried a little bit about ordering. The `try_process_record` can implement getting the required records from the changelog and then running a bunch of specific processors, this can potentially be made generic but I think it would be a big hard with generic types, like

```rust
 // fn try_process_record(
fn prepare_record_for processing(&self) -> T // Get all the pre requisite records ready to pass to ProcessRecordFucntion
fn get_process_record_functions(&sefl) ->Vec<Box<dyn ProcessRecordFunction<T>>>> { 
// ...
// The above should work but when it comes to Box Processor, need to actually specify "T" (so stuck on that)
```

server/service/src/processors/general_processor.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[derive(Clone)]
+pub enum ProcessorType {
+    ContactFormEmail,
+}
+
+impl ProcessorType {
+    pub(super) fn get_processor(&self) -> Box<dyn Processor> {
+        match self {
+            ProcessorType::ContactFormEmail => Box::new(QueueContactEmailProcessor),
```
lache-melvin, line 36: Only thought is whether this should support a vec of processors, like we have for invoices etc... there might end up being multiple things we want to process with the contact form? But guess we can add that when we get there 😁

lache-melvin (approved):
LGTM, thanks for changes 🙏

Issue #6248, opened 2025-01-22
When a new item is added to a prescription, all previous items quantity is reset to 0


marthakeezy:
## What went wrong? 😲

When a new item is added to a prescription, all previous items quantity is reset to 0 (only on the Issue ... field at the top. From what I have seen, it does not impact the ledger (!) and actually resets once you click back on the item.

Video of the behaviour ->>> [here](https://drive.google.com/drive/folders/1YAHfxK-88YAukryfthqpY2mBRWJQtL1m)

@jmbrunskill 

## Expected behaviour 🤔

Quantity issued shows the same regardless whether a new item has been added or not

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to Dispensary -> Prescriptions
2. Create a new prescription and add an item to it
3. Note the quantity you issued
4. Add another item -> revisit the old item -> see quantity updated number reset to 0
5. Click on another item -> click back on the old one -> see the quantity is back :)

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC7
- Legacy mSupply Central Server Version: V7.19.03

- Platform: Galaxy Tab A8 (Android 14)

- Database type: SQLite3

alainsussol:
Will be build tested in feature 2.6.0 Release testing #6000

aimee-mcneil-melville:
I'm still able to replicate this bug 😢 
When testing item directions on #6456, the allocated quantity either doesn't allocate at all, or un-allocates when adding a direction

andreievg:
@lache-melvin says i can close

PR #6345, opened 2025-01-29
6180 prescribed quantity

### LINKED ISSUE #6180: Dispensing: Prescribed Quantity
## Add Prescribed Quantity Prescription LineEditView

New field in the Line Edit View to record the prescribed quantity of the item (might be more than dispensed)

## Notes

Need to figure out Where is this stored in mSupply?
Not 100% sure where this field shows up for editing? (probably next to the all `Issue` field? Under quantity)

This column is editable manually
This column doesn’t trigger any calculation
No need to to Implement the store preference for ‘edit prescribed quantity’

## Acceptance Criteria 🧾

- [ ] New Prescribed Quantity field in the Prescription Line Edit View
- [ ] Field should sync to mSupply
- [ ] Field should be editable while the prescription is not verified
- [ ] No need to check if the field is valid, but must be a number
- [ ] Display field in table in DetailView

### Epic Link 🔗

#6000

### Moneyworks Jobcode 🧰

OMS:CIDV

GeronimoJohn:
Fixes #6180 

# 👩🏻‍💻 What does this PR do?

This update introduces a new field in the Line Edit View to record the prescribed quantity of an item. The field is editable and does not perform any calculations. The data entered in this field is saved to the database and synchronized with mSupply.

https://github.com/user-attachments/assets/ab922a19-f1d2-4f2e-96a0-3acfeb4d7ef7

Left a comment for an improvement that will be worked on a different PR

# 🧪 Testing

- [ ] Edit or create a new prescription
- [ ] A field called Prescribed Quantity is on the screen
- [ ] The field should either be pre-filled with a default value or zero if there's none
- [ ] The field can be updated to our desired prescribed quantity

# 📃 Documentation

  1.
  2.


GeronimoJohn:
A conversation about how the input behaves are still in discussion within it's - converting to a draft until we've decided the way to go

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -221,6 +224,8 @@ export const allocateQuantities =
       ...batch,
       numberOfPacks: 0,
       isUpdated: batch.numberOfPacks > 0,
+      prescribedQuantity:
+        batch.numberOfPacks > 0 && prescribedQuantity ? prescribedQuantity : 0,
```
GeronimoJohn, line 228: This is a temporary solution. The main goal is to only add prescribedQuantity to one line. But we've decided to tackle that in a different ticket for ease of review (this one is already big enough sorry 🙏🏽)

[Improvements Ticket](https://github.com/orgs/msupply-foundation/projects/21/views/7?pane=issue&itemId=95764653&issue=msupply-foundation%7Copen-msupply%7C6353)

GeronimoJohn: Potential solution for the above: https://github.com/msupply-foundation/open-msupply/commit/164f6cba1e4d754012ca980a9f5ff2402a1e2bdf

CarlosNZ:
Looking good John, I haven't dug into the code yet, will do that next. Just wanted to comment on the layout/behaviour first. I see you've had some discussion on the issue, so I don't want to contradict everything that was agreed there. My main concern was minimising the effort required for the user.

In the vast majority of cases the Prescribed quantity will match the Issue quantity, so the user should only have to type the value once. But the way it is here, is initially focuses to the "Issue" quantity, and the "Prescribed" value stays at 0. Wouldn't it be better to put the inital focs on "Prescribed" so that the "Issue" value is filled automatically?

Following on from that, I was just chatting to James, and we thought that the "Tab" key from "Prescribed" should take you straight to Directions, as most of the time that's what you'll want.

Probably best just to have a quick IRL chat before changing anything though, so pop over and see me when you have a chance :)

client/packages/invoices/src/StockOut/utils.tsx
GeronimoJohn: PR for the ticket mentioned in the thread: https://github.com/msupply-foundation/open-msupply/pull/6387

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -168,13 +183,31 @@ export const PrescriptionLineEditForm: React.FC<
...
+      numPacks,
+      Number(packSize),
+      quantityType === 'prescribed' ? inputUnitQuantity : prescribedQuantity
+    );
   };
 
-  const prescriptionLineWithNote = draftPrescriptionLines.find(l => !!l.note);
-  const note = prescriptionLineWithNote?.note ?? '';
+  const handlePrescribedQuantityChange = (inputPrescribedQuantity?: number) => {
+    if (!inputPrescribedQuantity) return;
```
CarlosNZ, line 194:
```suggestion
    if (inputPrescribedQuantity == null) return;
```

This one has caught me out a few times in the past -- if you just do`!inputPrescribedQuantity`, it'll return on *any* falsy value, including `0`. We need to accept "0", so check needs to be for explicitly non-values.

Just an annoying quirk of JS to always keep in mind ;)

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -239,31 +270,42 @@ export const PrescriptionLineEditForm: React.FC<
...
+                    {t('label.prescribed-quantity')}
+                  </InputLabel>
+                  <NumericTextInput
+                    autoFocus
+                    disabled={disabled}
+                    value={prescribedQuantity}
+                    onChange={handlePrescribedQuantityChange}
+                    min={0}
+                    decimalLimit={2}
+                  />
```
CarlosNZ, line 287: So one thing I noticed is that when you move from Prescribed input field to Issued input, the "Issued" value is briefly changed to the Prescribed until it is re-allocated, like this:



This is because the "change" method is actually being called "onBlur" as well (internally to the Numeric input). You can get around this just by passing a dummy "onBlur" function to override it:

```tsx
<NumericTextInput
    autoFocus
    disabled={disabled}
    value={prescribedQuantity}
    onChange={handlePrescribedQuantityChange}
    min={0}
    decimalLimit={2}
    onBlur={()=> {}}
  />
```

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -239,31 +270,42 @@ export const PrescriptionLineEditForm: React.FC<
...
+                </Box>
               </Grid>
               <Grid>
-                <InputLabel style={{ fontSize: 12 }}>
-                  {t('label.unit-plural', {
-                    count: issueUnitQuantity,
-                    unit: item?.unitName,
-                  })}
-                </InputLabel>
+                <Box display="flex" alignItems="center" gap={1}>
```
CarlosNZ, line 291: I don't think you need the `<Box>` -- all it ends up doing is rendering an additional `<div>` around a single child. 

This is something I've started noticing quite a bit, as it leads to annoying bugs such as this one: https://github.com/msupply-foundation/open-msupply/issues/6370

If you need to add styling, just put it on the `Grid` component, as at the end of the day it's just another `<div>` in the output:

```
 <Grid display="flex" alignItems="center" gap={1}>
```

client/packages/invoices/src/StockOut/utils.tsx
GeronimoJohn: 6387 is merged to this branch

client/packages/invoices/src/Prescriptions/api/hooks/utils.ts
```diff
@@ -32,16 +32,17 @@ export const createInputObject = (
...
   const stockLineId = stockLine?.id ?? '';
-  const output = { id, numberOfPacks, stockLineId, note };
+  const output = { id, numberOfPacks, stockLineId, note, prescribedQuantity };
 
   switch (type) {
     case 'delete':
       return { id };
     case 'update':
-      return output;
+      return { ...output };
```
CarlosNZ, line 45: I don't think you need the ...spread here, do you?

CarlosNZ (changes requested):
Looking good, John.

I've left a few comments, mostly pretty minor stuff.

There's one bug I noticed that I can't quickly figure out what's happening:
- Go into a prescription that already has lines (or create one), make sure there is a prescribed quantity and issue quantity for each.
- Add a new item, then save
- Now go back to the first item on the list and open the "Quantity" accordion -- the "Issue" quantity has been reset to 0! (If you refresh the page it'll come back).

Good luck tracking that one down :) j

client/packages/invoices/src/Prescriptions/api/hooks/utils.ts
GeronimoJohn: Woops, forgot to undo this one

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -80,21 +81,32 @@ export const PrescriptionLineEditForm: React.FC<
...
   const { format } = useFormatNumber();
   const { rows: items } = usePrescription();
 
+  const [issueUnitQuantity, setIssueUnitQuantity] = useState(0);
+  const [prescribedQuantity, setPrescribedQuantity] = useState(0);
+  const [allocationAlerts, setAllocationAlerts] = useState<StockOutAlert[]>([]);
+
+  const key = item?.id ?? 'new';
+  const prescriptionLineWithNote = draftPrescriptionLines.find(l => !!l.note);
+  const note = prescriptionLineWithNote?.note ?? '';
```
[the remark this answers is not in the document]
GeronimoJohn: This is more of a preference, but will revert back

CarlosNZ (approved):
Nice, thanks for the fixes.

And as discussed IRL, the other issue I found is actually [this one](https://github.com/msupply-foundation/open-msupply/issues/6248) and so nothing to do with your changes.

Happy to merge this now 👍

PR #6347, opened 2025-01-29
6295 add crawl function for generating reports

### LINKED ISSUE #6295: Update CLI to crawl through and find all dirs containing manifest.json

Current cli command to build reports is rigid to an arbitrary file structure. Would be more user friendly if we can crawl through all sub dirs of a nominated dir (which is passed through with the cli command), and generate all reports within that dir. 

Report dirs can be identified by them containing the manifest.json file.

Also rename manifest.json to report-manifest.json for specifificity.

This would make the command flexible to different file structures. For example, we may want to separate between 'forms' and 'reports' and have a number of reports in each dir. Alternatively we may want to pass a path to a specific report dir within that dir. This crawling would create flexibility for devs to generate a json with exactly the reports they are needing.

### Moneyworks Jobcode 🧰

OMS:REPT

fergie-nz:
Fixes #6295

# 👩🏻‍💻 What does this PR do?

- Adds crawl function to flexibly build reports in any folder structure.
- Also adds better error handling for anything cropping up when building reports
- Renames manifest.json to report-manifest.json

I've also updated error handling to make diagnosing errors with report development more easeful.
I was a little unsure about when to throw an error vs when to continue with other reports which might be working fine.

# 🧪 Testing

(Testing for devs - the report build cli is not accessible in testers environment)
- [ ] Move reports into a new file structure (ie separate reports into client sub folders, move them into folders within each other)
- [ ] See that report generation still works fine
- [ ] Generated json file contains all reports within the nominated dir

# 📃 Documentation


server/cli/src/report_utils.rs
```diff
@@ -0,0 +1,262 @@
...
+    let argument_schema_id =
+        existing_report.and_then(|r| r.argument_schema.as_ref().map(|r| r.id.clone()));
+
+    let form_schema_json = match (arguments_path, arguments_ui_path) {
+        (Some(_), None) | (None, Some(_)) => {
+            return Err(Error::FailedToGenerateArgumentSchema(path.to_path_buf()))
+        }
+        (Some(arguments_path), Some(arguments_ui_path)) => Some(
+            schema_from_row(FormSchemaRow {
+                id: argument_schema_id.unwrap_or(format!("for_report_{}", id)),
```
andreievg, line 155: I think format!("for_report_{}") is enough for all cases. Getting some id from local database doesn't make sense to me, reports are not database specific. Also I don't think that report making should rely on any database operations, please remove con. (It's not a new change in this PR, i understand, but would be great to get rid of it for now )

andreievg (changes requested):
Looks good thanks, although I haven't tested.

There is one change please, sorry it's not related to your new chagnes, but I would like it cleared up

andreievg (approved):
Looks great, thank !

PR #6351, opened 2025-01-30
6251 filter by next of kin name

### LINKED ISSUE #6251: Add a filter for the Next of kin's name (first and last) on the patient list

lache-melvin:
Fixes #6251

# 👩🏻‍💻 What does this PR do?

Adds next of kin filter to patient list view

I have patient with Basil Coriander as next of kin:
Screenshot 2025-01-30 at 2 43 24 PM

Search by name:
Screenshot 2025-01-30 at 2 43 41 PM 

This is currently just using `name` field, which is `[lastName], [firstName]`. Same filter applies when e.g. creating new prescription. I figure this is okay for first pass, and get feedback, might need to support searching in either order?

# 🧪 Testing

- [ ] Go to patient list view
- [ ] There is a `next of kin` option in the filters
- [ ] Search by the next of kin name of a patient
- [ ] That patient appears

# 📃 Documentation

  1.
  2.


Chris-Petty (approved):
ur so fas

image

Works sweet. I wonder UI wise if it might be useful showing a column of the next of kin name? Just cause I apparently have 4 different "Richard Moizeau"s and all these patients actually have a different next of kin record. Well... the column wouldn't really clarify in this case, but if you have many Richards with different surnames it might help!

lache-melvin:
>  I wonder UI wise if it might be useful showing a column of the next of kin name?

Was thinking this too, will make new issue, to do once next of kin node exposed on patient. Also not sure if we should show as 2 sep columns, or one `Moizeau, Richard` ?

PR #6356, opened 2025-01-30
6175 prescription price columns

### LINKED ISSUE #6175: Dispensing: Cost price in Prescription Detail View
## Add Cost Price to the Prescription Detail View

New Column in the Table in the Prescription Detail View to show the cost price of the item.

## Note

Sell price might get more complicated with insurance etc, but for now, we can just show the sell price as the pack quantity \* sell price.

## Acceptance Criteria 🧾

- [ ] add `Cost Price` column = `pack quantity * cost price`
- [ ] add `Sell Price` column = `pack quantity * sell price`

### Epic Link 🔗

#6000

### Moneyworks Jobcode 🧰

OMS:CIDV

zachariah-at-msupply:
Fixes #6175 

# 👩🏻‍💻 What does this PR do?

Adds cost and sell price columns to the prescription view.

image

I removed some of the other columns from the screenshot to highlight the new ones that are available. All the original columns are still available but the 'line total' column is now 'sell price'.

Sup, how's it going?

# 🧪 Testing

- [ ] Make sure you have multiple batches of the same medecine, with different sell and cost prices.
- [ ] Create a prescription and sdd both of those different batches of the same medecine to it.
- [ ] Go to the prescription's detail view, and check that the sell price column = sell price of one batch * quantity of packs of that batch being dispensed + sell price of the other batch * quantity of packs of that other batch being dispensed, and similar with the cost column.

# 📃 Documentation

- [x] **These areas should be updated or checked**: New sell and cost price columns in the Prescriptions detail view.


client/packages/invoices/src/Prescriptions/DetailView/columns.ts
```diff
@@ -262,64 +262,61 @@ export const usePrescriptionColumn = ({
...
           );
         } else {
-          return (rowData.sellPricePerPack ?? 0) / rowData.packSize;
+          return (rowData.costPricePerPack ?? 0) * rowData.numberOfPacks;
         }
       },
     },
+
     {
-      label: 'label.line-total',
```
roxy-dao, line 299: Are we removing the line total?

zachariah-at-msupply: The 'line total' column is now named 'sell price'. The issue asks for 'sell price', and it turns out the line total column is exactly the same content as sell price.

jmbrunskill: @zachariah-at-msupply just had a chat with Mark Prins, can we change this back to `unit-total` for now.
We can update the french version for CIV to be what they want it to be.

Also, can we add a new translation for `label.total-sell-price` as normally sell price is the unit price (not the total, so would be nice to distinguish.

I'm also a bit confused by the diff, I think maybe due to the ordering, would you mind putting the lineTotal first, then total cost price at the end of the list. Most users won't be using costPrice so I think it's good to have this one last.

jmbrunskill (approved):
Thanks @zachariah-at-msupply given @mark-prins is away, lets get this merged in.
If we need to remove that column later, it should be easy to do...

I tweaked the wording a little bit to just call it `Total Cost` which displays a bit nicer at least in French...
image

The other thing I put a suggestion in for the comment which seems wrong to me, if you agree can you add that suggestion too before merging.

PR #6375, opened 2025-01-31
6266 todays encounter report

### LINKED ISSUE #6266: Report for patients due for an encounter today
As a vaccination officer, I want to track children who are due for their vaccination today, so that I can anticipate how many children are expected at my health center and send reminders if needed for no-shows.

lache-melvin:
Fixes #6266

# 👩🏻‍💻 What does this PR do?

This is the `schedule for a vaccination today` report: https://docs.google.com/spreadsheets/d/13jcRALuMLXlmw13SwrIAO5SFhzIrp5vD94jE00D9wcU/edit?gid=0#gid=0

Screenshot 2025-01-31 at 3 10 00 PM

Screenshot 2025-01-31 at 2 25 55 PM

It should work for non-today encounters too, but that report wants some more columns too 😁 

"Today to tomorrow" is maybe not the most intuitive input.. would be nice if the "end" filter would use the end of the date, so it was inclusive, but I'm not sure how to configure this in the argument schema....

Hoping this is okay for now, will get feedback on Monday how to prioritise compared to other pieces...

# 🧪 Testing

- [ ] Have some patients scheduled for an encounter today (ensure still pending)
- [ ] At least some of these patients should have a next of kin, who has a phone number, to test those fields
- [ ] Also set a status on the patient's program enrolment
- [ ] Go to reports - select pending encounters
- [ ] Select your program
- [ ] Set date range for todays encounters (start of today to start of tomorrow)
- [ ] See report - your pending encounters should show

# 📃 Documentation

  1.
  2.


server/reports/generated/standard_reports.json
lache-melvin, line 1: If you haven't review reports yet, FYI that u just ignore this one, it's generated based on the other report files 🙏

server/graphql/types/src/types/program/patient.rs
```diff
@@ -217,6 +217,25 @@ impl PatientNode {
         })
     }
 
+    pub async fn next_of_kin(&self, ctx: &Context<'_>) -> Result<Option<PatientNode>> {
+        let loader = ctx.get_loader::<DataLoader<PatientLoader>>();
+
+        if let Some(next_of_kin_id) = &self.patient.next_of_kin_id {
+            let result = loader
```
Chris-Petty, line 224: probably copy pasta things but nit: 

```suggestion
    pub async fn next_of_kin(&self, ctx: &Context<'_>) -> Result<Option<PatientNode>> {
        if let Some(next_of_kin_id) = &self.patient.next_of_kin_id {
            let loader = ctx.get_loader::<DataLoader<PatientLoader>>();
            let result = loader
```

Don't need to initialise the loader for the other branch?

server/reports/encounters/2_6_0/src/template.html
```diff
@@ -0,0 +1,49 @@
+{% macro formatDate(datetime) %} {% if datetime %}{{ datetime |
+date(format="%d/%m/%Y") }}{% endif %} {% endmacro input %}
+
+<style>
+  {% include "encounters.css" %}
```
Chris-Petty, line 5: I haven't checked the other reports' css, but I assume there should be some pretty common css that can be shared amongst reports? That said perhaps there would be some nastiness with doing that with the report versioning.

Chris-Petty (changes requested):
image

When you are selecting a date, the OK button makes it look like it will select the currently highlighted date, but it just closes the date picker. You must instead click the date you want. I think it'd be clearer if it was like _Expiring Items_ report's "Clear" button:

image

Going past the scope of this PR I think that most report date pickers should default to today, at least the start date anyway. Maybe best as config because some reports would have other reasonable defaults (OG has a whole mechanism for saving your reporting settings!). Also the spacing between date pickers is a bit rough in the json form renderer:

image

client/packages/reports/src/ListView/ListView.tsx
```diff
@@ -45,8 +45,10 @@ export const ListView = () => {
   );
   const programReports = data?.nodes?.filter(
     report =>
-      report?.subContext === 'HIVCareProgram' &&
-      report?.context === ReportContext.Dispensary
+      report?.context === ReportContext.Dispensary &&
+      (report?.subContext === 'HIVCareProgram' ||
+        // TODO: Also check vaccine module enabled?
```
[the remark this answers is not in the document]
lache-melvin: Oh hehe - actually I think I will take this out. I thought the wording/query might be more closely tied to the vaccinations, but its actually just encounters, which could be useful for other kinds of programs too!

server/reports/encounters/2_6_0/src/template.html
lache-melvin: Yeah I think I remember it being a versioning thing... though something shared would be nice 🥲

server/graphql/types/src/types/program/patient.rs
lache-melvin: [3d8dde3](https://github.com/msupply-foundation/open-msupply/pull/6375/commits/3d8dde39240d8d59ed2addf7a92f4ac07ef589e4)

client/packages/reports/src/ListView/ListView.tsx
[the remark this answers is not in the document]
lache-melvin: [8a2b69d](https://github.com/msupply-foundation/open-msupply/pull/6375/commits/8a2b69da397cb7d63673835d36897b8b80f82e73)

server/graphql/types/src/types/program/patient.rs
```diff
@@ -218,22 +218,22 @@ impl PatientNode {
     }
 
     pub async fn next_of_kin(&self, ctx: &Context<'_>) -> Result<Option<PatientNode>> {
+        let Some(next_of_kin_id) = &self.patient.next_of_kin_id else {
+            return Ok(None);
+        };
```
Chris-Petty, line 223: Noice! The sorta funky `else` got me thinking if there was any kind of `let None = &self... {}`, so I asked chatgpt

---

Yes! You can make the code more concise using `?`, which propagates `None` early without requiring an explicit `return` statement:

```rust
let next_of_kin_id = self.patient.next_of_kin_id.as_ref()?;
```

### Explanation:
- `as_ref()` is used to avoid moving `next_of_kin_id` (assuming it's an `Option<T>`).
- The `?` operator will return `Ok(None)` early if `next_of_kin_id` is `None`.

This approach is more idiomatic and keeps the code tidy.

--- 

TIL

Chris-Petty (approved):
🚢

server/graphql/types/src/types/program/patient.rs
lache-melvin: I had a "woah cool how come we don't do that anywhere" moment... alas it would appear ChatGPT is lying 😱  `?` is only available on `Result`, rather than `Option` types. Possibly could map the types, but at that point I think not worth it 😢

Chris-Petty: Hahaha damn it looked so legit, seemed feasible to me

Chris-Petty: > Possibly could map the types, but at that point I think not worth it 😢

i.e. idiomatic rust, but madness in any other lang lol

Chris-Petty: ChatGPT losing the plot when called out
```rust
let next_of_kin_id = self.patient.next_of_kin_id.as_ref().ok_or(Ok(None))??;
```

Issue #5037, opened 2024-10-08
Restructure report docs for ease


fergie-nz:
## Is your feature request related to a problem? Please describe 👀

mention higher in docs that you will need to build the cli before report building commands work

lache-melvin:
In the docs or our readmes @fergie-nz? Can this just be a change in the docs?

fergie-nz:
Some notes to add:

Need to detail how report name is translated to display on front end. 
This is done by a utils function at `client/packages/reports/src/utils.ts`.
Translation done on report code rather than report name.

PR #6377, opened 2025-01-31
5863 Add overwrite flag to upsert json cli

### LINKED ISSUE #5863: Add overwrite flag to upsert json cli

[Reports versioning](https://github.com/msupply-foundation/open-msupply/pull/5533) introduced conditional upserting of reports on server start and with the `upsert-reports-json`. Ie, if a report already exists, it won't be upserted.

This makes developing reports difficult, as devs will likely want to rapidly upsert the same report many times while testing without bumping the version.

To change this, we want to introduce an -o (overwrite) flag on the `upsert-reports-json` command. This will force overwrite reports even if they exist already. 
Note this was the behaviour of the `upsert-reports-json` command prior to the report versioning PR.

Also we want to add a pathing package to the upsert-reports-json command when a reports.json file path is specified. Currently it will search compatibly with mac OS. We want this to be flexible to linux or other systems wherever OMS may be running.

Also need a separate CLI command for use on tester's environment which can access embedded standard reports (not looking through file structure).
This would need to overwrite existing reports in the database.

## Describe the solution you'd like 🎁|

- [ ] Add -o flag to `upsert-reports-json` command.
- [x] Add path package to optional reports json file location.
- [ ] Separate CLI to upsert bundled standard reports which is usable in tester's environment

EthanMcQ-TMF:
Fixes #5863 

# 👩🏻‍💻 What does this PR do?

- Add `--overwrite / -o` flag to `upsert-json` command. If used, overwrite any pre-existing reports.
- Add new `reload-embedded-reports` command. Reloads the embedded reports and overwrites them in the database

# 🧪 Testing

- Run `upsert-reports` without `-o` flag, observe 0 reports updated
- Run `upsert-reports` with `-o` flag, observe reports being updated
- Run `reload-embedded-reports`, observe reports being updated

# 📃 Documentation

- Additional command and flag are added to the usage/help text. Is there any extra docs needed considering this is a dev tool?


andreievg (changes requested):
This is great ! Thanks @EthanMcQ-TMF 

I just noticed that we don't error on UpsertReports and ReloadEmbeddedReports, I think we should. I know it's not really part of this PR scope, but can you please add.

> Is there any extra docs needed considering this is a dev tool? 

There is this issue, can you please add comment: https://github.com/msupply-foundation/open-msupply/issues/5037

EthanMcQ-TMF:
> I just noticed that we don't error on UpsertReports and ReloadEmbeddedReports, I think we should.

@andreievg I'm sorry, error in what scenario exactly? This makes it sound like you want this feature to just always throw an error...

server/cli/src/cli.rs
```diff
@@ -515,6 +521,12 @@ async fn main() -> anyhow::Result<()> {
 
             info!("Report upserted");
         }
+        Action::ReloadEmbeddedReports => {
+            let connection_manager = get_storage_connection_manager(&settings.database);
+            let con = connection_manager.connection()?;
+
+            let _ = StandardReports::load_reports(&con, true);
```
andreievg, line 528: `let _` basically discards the result.

Should be something like `StandardReports::load_reports(&con, true)?` or `StandardReports::load_reports(&con, true).expect("Failed to load reports")`

andreievg (approved):
Thank you @EthanMcQ-TMF, looks great now

PR #6378, opened 2025-01-31
Only run pre-commit lint if there are actually changes in the client dir

EthanMcQ-TMF:
This was just annoying me when working on `server` code. Why lint something that hasn't been changed?


jmbrunskill (approved):
Thanks!

BTW: The plan is to move this lint to be a github action so it doesn't hold up commits, haven't searched for the issue but would like to see this happen too!

PR #6383, opened 2025-02-02
Update utils.ts

### LINKED ISSUE #6382: Report name translation only replacing first space
## What went wrong? 😲

Report name translation only searches and replaces first space rather than all of them in a report name.

## How to Reproduce 🔨

Make a report with multiple spaces in its name. See how only first space is converted

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #6382 

# 👩🏻‍💻 What does this PR do?

Fix to how report names (code) is translated in the front end.

Currently only replaces first instance of ' ' in string with '-'. Should replace all

# 🧪 Testing

You can test this by adding multiple words to a report 'code' in your database, then view how it appears in the list view on the front end with both this branch and develop.
You will see this branch searches and replaces all spaces rather than just the first.

# 📃 Documentation

### DIFF
--- a/client/packages/reports/src/utils.ts
+++ b/client/packages/reports/src/utils.ts
@@ -4,5 +4,5 @@ export const translateReportName = (
   t: ReturnType<typeof useTranslation>,
   reportName: string
 ) => {
-  return t(`report.${reportName.replace(' ', '-').toLowerCase()}` as LocaleKey);
+  return t(`report.${reportName.replace(/ /g, '-').toLowerCase()}` as LocaleKey);
 };


jmbrunskill:
> Fixes #6382
> 
> # 👩🏻‍💻 What does this PR do?
> Fix to how report names (code) is translated in the front end.
> 
> Currently only replaces first instance of ' ' in string with '-'. Should replace all

Why not make the code with no spaces, or aligned directly with the translation key?
I thought that was part of the point of calling it a code, it doesn't need to be directly readable by humans?

fergie-nz:
> > Fixes #6382
> > # 👩🏻‍💻 What does this PR do?
> > Fix to how report names (code) is translated in the front end.
> > Currently only replaces first instance of ' ' in string with '-'. Should replace all
> 
> Why not make the code with no spaces, or aligned directly with the translation key? I thought that was part of the point of calling it a code, it doesn't need to be directly readable by humans?

Agreed! I've made a separate issue here to restructure how we translate report names:

https://github.com/msupply-foundation/open-msupply/issues/6385

I added this because it was causing issues in unexpected ways with how we are currently managing translations for front end reports.

jmbrunskill (approved):
Approving as it does what it says, and is an improvement, but would like to see #6385 done for a better solution.

PR #6393, opened 2025-02-03
Add Abbreviation graphql layer

### LINKED ISSUE #6334: Add abbreviations graphql

Need to get all abbreviations. They don't relate to any other entity, and there will rarely be more than say 1 or 2 hundred records with 3 fields. So probably don't need any pagination - with no paginating means we can fully apply the expansions on the front end, otherwise we'd need to hit the back end on every debounce keystroke.

- [ ] New graphql endpoint for the abbreviations table. This will be pretty similar to [`Diagnosis`](https://github.com/msupply-foundation/open-msupply/blob/f0f0cce153c7d2715c158ee26b0e41b7241f32ce/server/graphql/types/src/types/diagnosis.rs)

### Moneyworks Jobcode 🧰

OMS:DDSP

aimee-mcneil-melville:
Fixes #6334

# 👩🏻‍💻 What does this PR do?

Adds graphQl layer for the abbreviations query + a service file

Can query by filter (optional)

This will allow the abbreviations to be used in the prescription directions

# 🧪 Testing

- [ ] GraphQl endpoint testing - query abbreviations, you should have the option for expansion, id, and text
- [ ] Can query these fields to get all results
- [ ] Can query by the 'text' value of one abbreviation, and get the this abbreviation object

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/graphql/general/src/lib.rs
+++ b/server/graphql/general/src/lib.rs
@@ -6,13 +6,14 @@ pub mod types;
 pub use self::queries::sync_status::*;
 use self::queries::*;
 
+use abbreviation::abbreviations;
 use diagnosis::diagnoses_active;
 use graphql_core::pagination::PaginationInput;
 use service::sync::CentralServerConfig;
 
 use crate::store_preference::store_preferences;
 use graphql_types::types::{
-    CurrenciesResponse, CurrencyFilterInput, CurrencySortInput, DiagnosisNode,
+    AbbreviationNode, CurrenciesResponse, CurrencyFilterInput, CurrencySortInput, DiagnosisNode,
     MasterListFilterInput, StorePreferenceNode,
 };
 use mutations::{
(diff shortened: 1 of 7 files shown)


server/graphql/general/src/lib.rs
```diff
@@ -426,6 +428,14 @@ impl GeneralQueries {
     pub async fn diagnoses_active(&self, ctx: &Context<'_>) -> Result<Vec<DiagnosisNode>> {
         diagnoses_active(ctx)
     }
+
+    pub async fn abbreviations(
+        &self,
+        ctx: &Context<'_>,
+        filter: AbbreviationFilterInput,
```
lache-melvin, line 435: Per our other queries, I think we should make this `Option<AbbreviationFilterInput>` and pass that all the way through (`repo.query()` should accept optional filter, rather that `query_by_filter` where filter is required)

This way if we just want to query all abbreviations from frontend, don't need to pass filter

lache-melvin (changes requested):
Working great! Just the optional filter would be great 🙏

server/graphql/general/src/lib.rs
aimee-mcneil-melville: Have updated the filter to be optional!

PR #6395, opened 2025-02-03
6204 name insurance join

### LINKED ISSUE #6204: Sync and Translate nameInsuranceJoin table
## Sync and Translate nameInsuranceJoin table

Note: To enable insurance in mSupply, you need to use the Ivory Coast customisation.

After enabling this, a new insurance providers menu appears where these can be configured.

https://docs.msupply.org.nz/dispensing:patient_insurance#insurance_providers

You'll need at least one insurance provider to be able to add insurance to patients in mSupply.
Eventually you'll be able to setup these insurance provider links to patients in open-mSupply, however right now you'll need to setup them up in mSupply to test sync.
This might require moving an existing mSupply site with patients to become an open-mSupply site (re-initialied as open-mSupply) to test this feature.

## Tasks

- [ ] New Database table in oms `name_insurance_join`
- [ ] NameInsuranceJoinRow (find_by_id)
- [ ] Translations for `Upsert`

A Similar Reference PR: https://github.com/msupply-foundation/open-msupply/pull/5752

## Acceptance Criteria 🧾

- [ ] Can sync `nameInsuranceJoin` records to and from mSupply

### Epic Link 🔗

#6000

### Moneyworks Jobcode 🧰

OMS:CIDV

jmbrunskill:
Fixes #6204

# 👩🏻‍💻 What does this PR do?

Adds sync for `nameInsuranceJoin` to open-mSupply

# 🧪 Testing

For dev testing, we can do something like this...

- [ ] Create an insuranceProvider in 4d
- [ ] Choose a patient in your database and get their ID
Run an insert record something like this in the database
`INSERT INTO "main"."name_insurance_join"
("id", "name_link_id", "insurance_provider_id", "policy_number", "policy_type", "discount_percentage", "expiry_date", "is_active")
VALUES ('TEST_NAME_INSURANCE_JOIN', 'PATIENT_NAME_ID', 'INSURANCE_PROVIDER_ID', '2312', 'business', 10, '2027-01-01', 'true');`
Then insert the changelog
`INSERT INTO changelog (cursor, table_name, record_id, row_action) VALUES ((SELECT max(cursor)+1 FROM changelog), 'name_insurance_join', 'TEST_NAME_INSURANCE_JOIN' ,'UPSERT')`
- [ ] Sync
- [ ] Check the name has the new record assigned

# 📃 Documentation

  1.
  2.


server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
```diff
@@ -0,0 +1,56 @@
...
+                r#"
+                    ALTER TYPE changelog_table_name ADD VALUE IF NOT EXISTS 'name_insurance_join';
+                "#
+            )?;
+
+            // Postgres enum variant for policy_type
+            sql!(
+                connection,
+                r#"
+                    CREATE TYPE insurance_policy_type AS ENUM ('Personal', 'Business');
```
CarlosNZ, line 30: As discussed, probably needs to be ALL_CAPS like the other PG enums.

server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
```diff
@@ -0,0 +1,56 @@
...
+        sql!(
+            connection,
+            r#"
+                CREATE TABLE name_insurance_join (
+                    id TEXT NOT NULL PRIMARY KEY,
+                    name_link_id TEXT NOT NULL REFERENCES name_link(id),
+                    insurance_provider_id TEXT NOT NULL REFERENCES insurance_provider(id),
+                    policy_number_person TEXT,
+                    policy_number_family TEXT,
+                    policy_number TEXT NOT NULL,
```
CarlosNZ, line 44: As discussed, probably want to make "person" and "family" just boolean flags in OMS, and just translate them to OG?

server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
```diff
@@ -0,0 +1,56 @@
...
+                    policy_type {policy_type} NOT NULL,
+                    discount_percentage INTEGER NOT NULL,
```
CarlosNZ, line 46: Can policy_type be `null`?

CarlosNZ (changes requested):
Looks good as far as I can tell. Just that failing sync due the PG Enums to sort out, plus those suggestions re the OMS schema fields.

server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
jmbrunskill: Done!

server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
jmbrunskill: Doesn't look like it in OG, so I think it's ok to keep it as NOT NULL.

server/repository/src/migrations/v2_06_00/add_name_insurance_join.rs
jmbrunskill: I've checked in mSupply, it looks like if both fields are filled in, they're combined in to the `full` field.
E.g. 123-321

So I think we probably want to keep it how it is...

CarlosNZ (commented):
Couple of quick things before approve:
- The "discount percentage" in mSupply accepts non-integer values, so we should probably make our field float rather than integer?
- The syncing only seems to be going one way (from OMS-> to OG) -- if I add a policy to a user in mSupply it doesn't sync back. Is this what we're expecting at the moment?

jmbrunskill:
> * The "discount percentage" in mSupply accepts non-integer values, so we should probably make our field float rather than integer?

Yes, sorry that's a great point!

jmbrunskill:
> * The syncing only seems to be going one way (from OMS-> to OG) -- if I add a policy to a user in mSupply it doesn't sync back. Is this what we're expecting at the moment?

Yes, this is expected. It should come down when you initialise the store if created from that store though. I guess we can test by re-initialising our mSupply site after adding the nameInsuranceJoin Manually. If we synced all patients and all patient insurances to every store it would be too much data. It does make me wonder though if you can take over being the patient's main store, does that somehow interfere?

CarlosNZ (approved):
Cool, thanks @jmbrunskill LGTM!

PR #6397, opened 2025-02-03
Add cli to generate and install plugin bundle

### LINKED ISSUE #6396: Make it easier to generate and install plugin bundle
While testing load plugin to plugin provider work, needed to re-generate and upsert plugin bundle, following process in `amc.js`, which was very time consuming. For the sake of PR reviewer and future backend plugin devs, implement CLI to generate and isntall plugin bundle

andreievg:
Fixes #6396 

# 👩🏻‍💻 What does this PR do?

Add install-plugin-bundle cli and general-and-install-plugin-bundle cli, see `amc.js` on notes of how to use.

# 💌 Any notes for the reviewer?

There is now API accessor in the cli (api struct), which is similar to what was done in the `report_builder` but a bit easier on the consumer, we would use this abstraction when time comes to delete report builder

# 🧪 Testing

Make changes to` amc.js`
Run
`cargo run --bin remote_server_cli -- generate-and-install-plugin-bundle -i './service/src/backend_plugin/examples/amc' -u 'http://localhost:8000' --username 'test' --password 'pass'`, and see your AMC values update when looking at catalogue


server/cli/src/cli.rs
```diff
@@ -154,6 +154,10 @@ enum Action {
     },
     /// Will generate a plugin bundle
     GeneratePluginBundle(GeneratePluginBundle),
+    /// Will insert generated plugin bundle
+    InstallPluginBundle(InstallPluginBundle),
```
andreievg, line 158: Was thinking maybe we just need generate and install, kept this one however as a manual was to install via CLI

server/cli/src/graphql/auth.rs
```diff
@@ -0,0 +1,16 @@
+pub(super) const AUTH_QUERY: &str = r#"
+query AuthToken($username: String!, $password: String) {
+  root: authToken(password: $password, username: $username) {
```
andreievg, line 3: this root relates to the `GraphQlResponse` struct and `Root` struct, it seems cleaner to use struct with auto derived de/se and map graphql field to root, rather then manually accessing serde_json::Value type

server/cli/src/graphql/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+            serde_json::json! ({
+              "username": username,
+              "password": password,
+            }),
+            None,
+            Some("AuthToken"),
+        )
+        .await?;
+
+        let token = result["token"].as_str().unwrap().to_string();
```
andreievg, line 70: extracting and saving token, unwrap should be save as we've checked 'expected_typename' and query is const

server/cli/src/graphql/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub async fn upload_file(&self, path: PathBuf) -> Result<UploadedFile, Error> {
+        let url = self.url.join("upload").unwrap();
+
+        let auth_cooke_value = format!(r#"auth={{"token": "{}"}}"#, self.token);
+        let built_request = reqwest::Client::new()
+            .post(url.clone())
+            .header("Cookie", auth_cooke_value);
+
+        // Add file to request
+        let mut file_handle =
```
andreievg, line 100: Copied from https://github.com/msupply-foundation/open-msupply/blob/86be0d6b3cef63051dfad3655582944800232c13/server/service/src/sync/api_v6/upload_file.rs#L60-L61

server/cli/src/graphql/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+        send_and_parse(built_request, url).await
+    }
+}
+
+async fn _gql(
+    url: &Url,
+    query: &str,
+    variables: serde_json::Value,
+    token: Option<&str>,
+    expected_typename: Option<&str>,
```
andreievg, line 121: Will check typename at the 'root' if specified

server/service/src/backend_plugin/examples/amc/amc.js
```diff
@@ -36,6 +42,9 @@ let plugins = {
     const sql_result = sql(sql_statement);
     const response = {};
 
+    // Fill all item_ids with default
+    item_ids.forEach((itemId) => (response[itemId] = { average_monthly_consumption: 1 }));
```
andreievg, line 46: If there is not 'consumption' then it will be zero, changed to '1' to see effect of the plugin

server/cli/src/graphql/mod.rs
jmbrunskill: I agree this is ok

server/cli/src/plugins.rs
```diff
@@ -125,3 +163,70 @@ fn process_manifest(bundle: &mut PluginBundle, path: &PathBuf) -> Result<(), Err
...
+    }: GenerateAndInstallPluginBundle,
+) -> Result<(), Error> {
+    let out_file = PathBuf::from("report_temp.json");
+
+    generate_plugin_bundle(GeneratePluginBundle {
+        in_dir,
+        out_file: out_file.clone(),
+    })?;
+
+    let api = Api::new_with_token(url.clone(), username, password).await?;
```
jmbrunskill, line 212: Potentially can call the function `install_plugin_bundle` from with in this one, but I agree it's not too much code duplication, so I'm happy with this..

jmbrunskill (approved):
THanks @andreievg works well for me...

One thing to note, not related to this PR directly, but if your plugin has an error in it, or fails the whole request seems to be killed.

image

```
    if (Math.random() > 0.5) {
      throw new Error("Random error");
    }
```

I'm not sure if we've talked about expected error handling? Maybe wort thinking about, I think ideally this shouldn't stop me seeing my whole item list...
But won't be easy to handle it an show an error else where. Maybe it's fine how it is but I can imagine a small plugin bug breaking the whole system for a client....

server/cli/src/plugins.rs
andreievg: Agree, thanks: 532f988b3270fd2a061c470fa42ece2b8fee261b

andreievg:
> One thing to note, not related to this PR directly, but if your plugin has an error in it, or fails the whole request seems to be killed.

Thanks for flagging, made an issue: https://github.com/msupply-foundation/open-msupply/issues/6447

andreievg:
Thanks for review @jmbrunskill

PR #6399, opened 2025-02-03
5829 load plugins via processor

### LINKED ISSUE #5829: Load plugin to plugin provider

At this point plugin should exist in database.

Do this on:

* Startup
* After sync (when remote site receives a plugin)
* After it's been installed/loaded on central

(suggest processor function that looks at plugins table, for the last two)

andreievg:
Fixes #5829 

# 👩🏻‍💻 What does this PR do?

Adds a new processor to load plugins to plugin provider (so that after sync plugins are loaded without restart).
Plugins are also loaded at startup.
Added enum test, key_type_enum, to make sure key types in key value store match database type definitions

# 🧪 Testing

Setup remote site and central site on this version. Sync them. Then run

cargo run --bin remote_server_cli -- generate-and-install-plugin-bundle -i './service/src/backend_plugin/examples/amc' -u 'http://localhost:8000' --username 'test' --password 'pass' 

Above will upload plugin to central sever, after syncing remote site you should see AMC calculated per amc.js file. On servers restart you should still see AMC calculater as per amc.js file


server/repository/src/db_diesel/key_value_store.rs
```diff
@@ -16,8 +16,9 @@ table! {
     }
 }
 
-// Database:  https://github.com/openmsupply/open-msupply/blob/d6645711184c63593949c3e8b6dc96b5a5ded39f/server/repository/migrations/postgres/2022-02-11T15-00_create_key_value_store/up.sql#L2-L16
+// Snippet for adding new, including migration : https://github.com/msupply-foundation/open-msupply/wiki/Snippets "New Key Type for KeyValueStore"
```
andreievg, line 19: Added a snippet for this, for future additions

server/repository/src/db_diesel/key_value_store.rs
```diff
@@ -16,8 +16,9 @@ table! {
...
 #[derive(DbEnum, Debug, Clone, PartialEq, Eq, Default)]
+#[cfg_attr(test, derive(strum::EnumIter))]
```
andreievg, line 21: This test wasn't stricly necessary, but was quick to add and doesn't add too much extra code, sorry to the reviewer

server/server/src/lib.rs
```diff
@@ -147,11 +147,14 @@ pub async fn start_server(
     }
 
     // PLUGIN CONTEXT
-
```
andreievg, line 150: Oops, sorry for format change

server/service/src/plugin/mod.rs
```diff
@@ -56,10 +66,11 @@ pub trait PluginServiceTrait: Sync + Send {
 
                 for row in plugin_bundle.backend_plugins.clone() {
                     backend_repo.upsert_one(row.clone())?;
-                    // Bind plugin to provider (this would ideally happen via processor, going over changelog)
-                    PluginInstance::bind(row)
                 }
 
+                ctx.processors_trigger
```
andreievg, line 71: A trigger would work here, since rows have been upserted

server/service/src/processors/general_processor.rs
```diff
@@ -11,15 +11,15 @@ use crate::{
     service_provider::{ServiceContext, ServiceProvider},
 };
 
-use super::contact_form::QueueContactEmailProcessor;
+use super::{contact_form::QueueContactEmailProcessor, load_plugin::LoadPlugin};
 
 #[derive(Error, Debug)]
 pub(crate) enum ProcessorError {
-    #[error("{0:?} not found: {1:?}")]
+    #[error("{0} not found: {1}")]
```
andreievg, line 18: Chnaged error mapping, tehnically should just do `[source]` or from for existing error and they will be added to the error message without needing to print them out

server/service/src/processors/load_plugin/load_plugin.rs
```diff
@@ -0,0 +1,43 @@
...
+        changelog: &ChangelogRow,
+    ) -> Result<Option<String>, ProcessorError> {
+        let plugin = BackendPluginRowRepository::new(connection)
+            .find_one_by_id(&changelog.record_id)?
+            .ok_or(ProcessorError::RecordNotFound(
+                "Backend plugin".to_string(),
+                changelog.record_id.clone(),
+            ))?;
+
+        PluginInstance::bind(plugin);
```
andreievg, line 31: Pretty simple processor to just just call 'bind'

server/service/src/plugin/mod.rs
```diff
@@ -43,6 +44,15 @@ pub trait PluginServiceTrait: Sync + Send {
         uploaded_file.as_json_file(settings)
     }
 
+    fn reload_all_plugins(&self, ctx: &ServiceContext) -> Result<(), RepositoryError> {
+        let repo = BackendPluginRowRepository::new(&ctx.connection);
+        for row in repo.all()? {
```
jmbrunskill, line 49: Just realised, that we don't have soft delete for plugins, Is there an issue for this, I think we definitely want to be able to remove them...

jmbrunskill (approved):
Working great for me @andreievg 

Just want to follow up on the deleting situation, especially given we have potential for a plugin to be breaking a big part of the app.

server/service/src/plugin/mod.rs
andreievg: Agree, was on my list but not captured, thanks: https://github.com/msupply-foundation/open-msupply/issues/6448

PR #6408, opened 2025-02-03
#6333 item directions gql

### LINKED ISSUE #6333: Add item_directions to item gql node

Need API to provide data for prescription item directions #5648 

- [ ] Create `item_direction` gql loader (something similar to the 2 fields mentioned below)
- [ ] Add `item_directions` to `ItemNode`, it'll be similar to [`available_batches`](https://github.com/msupply-foundation/open-msupply/blob/f0f0cce153c7d2715c158ee26b0e41b7241f32ce/server/graphql/types/src/types/item.rs#L97-L113) or [`variants`](https://github.com/msupply-foundation/open-msupply/blob/f0f0cce153c7d2715c158ee26b0e41b7241f32ce/server/graphql/types/src/types/item.rs#L136-L144)
- [ ] Ultimately can query for something like 
```jsonc
{
  data:
    invoice:
     lines:
       item:
         item_directions: [ nodes ]
}
```

Making a separate endpoint for querying just for item directions. We don't need to do any modifying right now and otherwise only really need it when querying for an item. Perhaps I am unfamiliar with how this `async_graphql` crate works and we'll end up with it regardless.

### Moneyworks Jobcode 🧰

OMS:DDSP

Chris-Petty:
Fixes #6333 

# 👩🏻‍💻 What does this PR do?

Creates GQL on `item` to access `itemDirections`.

There are a bunch of commented out tests that need updating (they were basically copied from `barcode` gql type). I've noticed that item_variant doesn't have tests and I'm tempted to just delete the tests rather than study how to make them work.

# 🧪 Testing

In a gql client query something like:

```gql
query invoices_with_directions {
  invoices(storeId: "1E31B3575A284B7DB27EE755102446FE") {
    ... on InvoiceConnector {
      nodes {
        id
        lines {
          nodes {
            item {
              name
              itemDirections {
                id
                itemLinkId
                priority
                directions
              }
            }
          }
        }
      }
    }
  }
}
```

# 📃 Documentation


server/graphql/core/src/loader/item_direction.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Ok(map)
+    }
+}
+
+pub struct ItemDirectionByItemDirectionIdLoader {
+    pub service_provider: Data<ServiceProvider>,
+}
+
+impl Loader<String> for ItemDirectionByItemDirectionIdLoader {
```
Chris-Petty, line 41: This isn't actually used and is here for boiler plate cargo cult reasons

server/graphql/core/src/loader/item_direction.rs
```diff
@@ -0,0 +1,62 @@
...
+        let service_context = self.service_provider.basic_context()?;
+        let repo = ItemDirectionRepository::new(&service_context.connection);
+
+        let item_directions = repo.query_by_filter(
+            ItemDirectionFilter::new().id(EqualFilter::equal_any(item_direction_ids.to_vec())),
+        )?;
+
+        let mut map: HashMap<String, ItemDirection> = HashMap::new();
+        for direction in item_directions {
+            map.insert(direction.item_row.id.clone(), direction);
```
Chris-Petty, line 58: That said, this is probably wrong and should be 
```suggestion
            map.insert(direction.item_direction_row.id.clone(), direction);
```

server/graphql/general/src/queries/requisition_line_chart.rs
```diff
@@ -167,7 +167,6 @@ mod graphql {
         requisition_line::RequisitionLineServiceTrait,
         service_provider::{ServiceContext, ServiceProvider},
     };
-    use util::{inline_edit, inline_init};
```
Chris-Petty, line 170: Random cargo check fix

server/graphql/types/src/types/item.rs
```diff
@@ -143,6 +143,14 @@ impl ItemNode {
         Ok(ItemVariantNode::from_vec(result))
     }
 
+    pub async fn item_directions(&self, ctx: &Context<'_>) -> Result<Vec<ItemDirectionNode>> {
+        let loader = ctx.get_loader::<DataLoader<ItemDirectionsByItemIdLoader>>();
+        let result = loader.load_one(self.row().id.clone()).await?;
+        let result = result.unwrap_or_default();
```
Chris-Petty, line 149: Shadowed variable 'cause using `dbg!(&result);` before unwrap_or_default()

server/graphql/types/src/types/item_direction.rs
```diff
@@ -0,0 +1,142 @@
...
+            nodes: item_directions
+                .rows
+                .into_iter()
+                .map(ItemDirectionNode::from_domain)
+                .collect(),
+        }
+    }
+}
+
+// #[cfg(test)]
```
Chris-Petty, line 82: Tempted to delete these rather than fix? Not all gql types have tests it seems. E.g. stock_lines, item_variants

server/graphql/core/src/loader/item_direction.rs
lache-melvin: Does this translate to "my next PR will use it"?

Is long term that a prescription would store item_direction_id and directions would be loaded via this?

Otherwise burn it lol

server/graphql/core/src/loader/item_direction.rs
lache-melvin: Yeah map lookup would be by direction id so this will break 😁

server/graphql/types/src/types/item_direction.rs
```diff
@@ -0,0 +1,142 @@
...
+    nodes: Vec<ItemDirectionNode>,
+}
+
+#[Object]
+impl ItemDirectionNode {
+    pub async fn id(&self) -> &str {
+        &self.item_direction.item_direction_row.id
+    }
+
+    pub async fn item_link_id(&self) -> &str {
```
jmbrunskill, line 23: I think we've agreed that in the API we should only expose `item_id` not `item_link_id`.
I've been asking for these things to be documented but yeah...

server/graphql/types/src/types/item_direction.rs
lache-melvin: Can remove these - the barcode tests you copied feel to me like they're in the wrong place, they're testing the result of the `barcode` query rather than the node type - you don't have a standalone query for item directions. So if you were to add to a test, would add to the item one (but item seems to only be testing legacy fields 😭)

BUT zooming out I have THOUGHTS about our graphql tests, I don't think they add near enough value for the effort right now. They're painful to write, and either haven't caught the bug, or they were just brittle and failed because I added something and the objects no longer lined up. 

- I think these nodes should have unit tests, that just call these little functions, if anything juicier than return this field is happening (i.e. test `from_domain` mappings, in the item test I would check that directions got loaded right)
  - This should need much less instantiating of graphql test biz, and therefore run much faster
  - Should be small easy fast to write
- Probably also valuable would be some kind of `all fields` query, which we would have to keep up to date as we add new fields to each query. 
  - Primary purpose of this would be to check that the API doesn't change in unexpected ways, contract test for the frontend, catch where we might have enthusiastically renamed or changed types.
  - Honestly not even how valuable this is, seeing as we have the generated types, which will break frontend compilation if anything changes 
  - I think more valuable when we get to the point of the API needing to be stable for other consumers
- Mutations should have the same unit tests wherever input/output mappings occur. Otherwise all biz logic happens in service layer so not a whole lot to test.

lache-melvin (changes requested):
Looking good! Can remove those tests, and fix up id mappings, then good to gooooo

client/packages/common/src/types/schema.ts
```diff
@@ -3557,7 +3557,7 @@ export enum InvoiceNodeStatus {
    * Outbound Shipment: Becomes not editable
    * Inbound Shipment: For inter store stock transfers an inbound Shipment
    * becomes editable when this status is set as a result of corresponding
-   * outbound Shipment being chagned to shipped (this is similar to New status)
+   * outbound Shipment being changed to shipped (this is similar to New status)
```
lache-melvin, line 3560: I think need another yarn generate too, this file should show the item direction changes?

Chris-Petty: Will do! I wasn't sure if the yarn generate is done for the backend PR or the frontend trying to consume.

server/repository/src/db_diesel/item_direction.rs
```diff
@@ -40,25 +50,40 @@ impl<'a> ItemDirectionRepository<'a> {
...
-        }
+fn create_filtered_query(filter: Option<ItemDirectionFilter>) -> BoxedItemDirectionQuery {
+    let mut query = item_direction::table
+        .inner_join(item_link::table.inner_join(item::table))
+        .into_boxed();
 
-        query
+    if let Some(filter) = filter {
+        apply_equal_filter!(query, filter.id, item_direction::id);
+        apply_equal_filter!(query, filter.item_id, item_direction::item_link_id);
```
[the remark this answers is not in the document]
Chris-Petty: Nice catch!

server/graphql/core/src/loader/item_direction.rs
Chris-Petty: No I see no use case for it, that is the only feasible idea long term that might and it ain't one we wanna do.

server/graphql/types/src/types/item_direction.rs
Chris-Petty: https://github.com/msupply-foundation/open-msupply/pull/6408/commits/43151d5a6f4b78d7c3c9de63b079d5ed1616b589

server/graphql/core/src/loader/item_direction.rs
Chris-Petty: https://github.com/msupply-foundation/open-msupply/pull/6408/commits/f03fa967ad7a81e9e48b21dd09516ee9551638ff

server/graphql/core/src/loader/item_direction.rs
Chris-Petty: https://github.com/msupply-foundation/open-msupply/pull/6408/commits/f03fa967ad7a81e9e48b21dd09516ee9551638ff

client/packages/common/src/types/schema.ts
Chris-Petty: d6d85dbc4b94

server/graphql/types/src/types/item_direction.rs
Chris-Petty: > Can remove these - the barcode tests you copied feel to me like they're in the wrong place, they're testing the result of the `barcode` query rather than the node type - you don't have a standalone query for item directions. So if you were to add to a test, would add to the item one (but item seems to only be testing legacy fields 😭)
> 

GONE. Yea I looked at item tests and decided too much effort. I started removing the inline_inits... nope. Also I tried to follow how the gql test methods work and quickly figured that was probably a futile/not useful effort haha.

> BUT zooming out I have THOUGHTS about our graphql tests, I don't think they add near enough value for the effort right now. They're painful to write, and either haven't caught the bug, or they were just brittle and failed because I added something and the objects no longer lined up.
> 
>     * I think these nodes should have unit tests, that just call these little functions, if anything juicier than return this field is happening (i.e. test `from_domain` mappings, in the item test I would check that directions got loaded right)
>       * This should need much less instantiating of graphql test biz, and therefore run much faster
>       * Should be small easy fast to write
> 

YAY

>     * Probably also valuable would be some kind of `all fields` query, which we would have to keep up to date as we add new fields to each query.
>       * Primary purpose of this would be to check that the API doesn't change in unexpected ways, contract test for the frontend, catch where we might have enthusiastically renamed or changed types.
>       * Honestly not even how valuable this is, seeing as we have the generated types, which will break frontend compilation if anything changes
>       * I think more valuable when we get to the point of the API needing to be stable for other consumers
>     * Mutations should have the same unit tests wherever input/output mappings occur. Otherwise all biz logic happens in service layer so not a whole lot to test.

I think the goals for the tests are mainly:

- Avoid API breaking changes. As you've noted, field name changes are gql API breaking changes - doing so needs deprecating the old field (and leaving it) and adding the new field name separately (or going against gql "best" practice of no versioning). This is very important given we advertise the API as an integration layer, not to mention our reports! As it is I feel like the leaning here would be to just rename things in tests to fix them given the amount of times we've borken reports. If simplifying like this made it easier for people to reason to not do that then it'll be good!
- Easy testing/dev velocity by just running a test rather than compiling/running the server and using an external API request app. Only helps velocity if writing the tests isn't a nightmare!

I think maybe make a refactor issue, or a github discussion?

server/repository/src/db_diesel/item_direction.rs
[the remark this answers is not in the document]
Chris-Petty: 77693c5b3a25

server/graphql/types/src/types/item_direction.rs
lache-melvin: YES PLZ (maybe both, discussion so as not to crazy bloat the issue, but create an issue so it doesn't go to the discussion board to die 😁)

lache-melvin (approved):
STUNNING TY

PR #6420, opened 2025-02-04
Disable pluralisation in prescription line edit view

### LINKED ISSUE #5978: Prescription: 'S' appearing for item with no units
## What went wrong? 😲

As per the title
photo_2025-01-09 1 35 50 pm
Screenshot 2025-01-09 at 1 39 34 pm

## How to Reproduce 🔨

2. Make a prescription and add item with no units
3. See error(extra 'S' appearing in place of unit)
4. Extra 'S' for items with units as well

## Your environment 🌱

- Open mSupply Version:  `V2.5.00-RC1`
- Legacy mSupply Central Server Version: `V8.0-dev32`

- Platform: Browser

- Database type:

zachariah-at-msupply:
Fixes #5978 

# 👩🏻‍💻 What does this PR do?

Disables pluralisation in line edit view. (Did you read the title? What did you expect? 😜)

Which means that now there are no silly 's'ses like in this sentens.

Now it looks like this for items without units:
image

And like this for items with units:
image

Have a nice day :)

# 🧪 Testing

- [ ] Go to Dispensary > Prescriptions > New Prscription (it's in the top right hand corner).
- [ ] In the input box which appears, type a character to search for a patient, and select a patient.
- [ ] Click 'Add item'.
- [ ] In the input box which appears, click the dropdown arrow at the end of the box, and select an item that doesn't have units (e.g. '78373 Clotrimazole cream 60 ml').
- [ ] Issue 0 units and check that there's no s after the number issued.
- [ ] Issue 1 unit and check that again.
- [ ] Issue more than 1 unit and check that again.
- [ ] Now select a medicine that _does_ have units (e.g. 030062 Acetylsalicylic Acid 300mg tabs).
- [ ] And do the same checks as before.
Thanks!
 
# 📃 Documentation

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -335,7 +335,10 @@ export const PrescriptionLineEditForm: React.FC<
                   decimalLimit={2}
                 />
                 <InputLabel sx={{ fontSize: 12 }}>
-                  {t('label.unit-plural', {
+                  {t('label.unit-plural_one', {
+                    // unit-plural_one has been specified to deactivate pluralisation.
+                    // This is to handle the case where units have not been entered for the item.
+                    // see https://github.com/msupply-foundation/open-msupply/issues/5978
                     count: issueUnitQuantity,
                     unit: item?.unitName,
                   })}


GeronimoJohn (approved):
Thanks for this @zachariah-at-msupply. LGTM! 🚀 

Though personally I found the comment a bit overwhelming for a one line change

mark-prins:
I would prefer to retain the pluralisation - the 'Issue 10 tablet' doesn't read well.
Doing this would retain the plural and prevent the problem raised:

```
                {!!item?.unitName && (
                  <InputLabel sx={{ fontSize: 12 }}>
                    {t('label.unit-plural', {
                      count: issueUnitQuantity,
                      unit: item?.unitName,
                    })}
```

Ideally we'd be checking for how to generate plurals properly - for english a simple solution would be to add [this package](https://www.npmjs.com/package/pluralize). it doesn't help with other languages, but it's a start! 

I was keen to have configuration for both singular and plural forms of the unit - until that can be done we'll just have to compromise.

mark-prins:
Also - we shouldn't call the `_one` form of a translation key directly. that's used when the `count` provided is `1`
If you change to use that `_one` versoin, there's no need to provide the `count` prop

A better approach is to provide `count: 1` but an even better would be to use a different translation.

zachariah-at-msupply:
Cool, gotcha. I'll just check with my team leader if he wants me to spend time on this before implementing the code you've suggested. 

It seems like in the issue you made (#6445, https://github.com/msupply-foundation/open-msupply/issues/6445) you would prefer the pluralize package to be added rather than the code - would you prefer the code added now as a temporary measure, or to wait for the pluralize package to be added?

mark-prins:
thanks - I'd recommend doing the pluralize approach; that should be pretty quick to do and is a slightly more complete solution. No need to implement this change and then rework later, the site is usable as it is now

Chris-Petty:
Sorry @mark-prins, you go away ONE DAY and we go crazy! Refinement we decided to just nix it because this has come up multiple times and reckoning people find bad pluralisation worse than lack of pluralisation.

PR #6423, opened 2025-02-05
6410 next of kin name

### LINKED ISSUE #6355: Show next of kin column in patient list

Per Chris:

Image

Show column with next of kin name. 2, for first and last names, or just one? If 2 columns, should there be 2 filters?

### LINKED ISSUE #6410: Quick add patient from `Next of Kin` selector

per @richardmoizeau 

When adding/editing a patient, there is `Next of kin` field. However, it only lists existing patients. This is a bit of hassle if you are adding a new patient, and their next of kin is also not yet in the system. Have to navigate away, add that patient, and come back.

Proposing quick add button here:

Image

Could open a modal that just accepts the min fields required to add a patient

## Considerations

Probably has to be the whole patient UI schema, because we won't know which fields are required on custom patient. Maybe we can at least get away with putting this all in one modal, rather than the two separate steps?

Main concern is whether we would need the central lookup step too? And how we would piece that flow together. How bad is it to end up with an occasional duplicated patient (particularly if they are in the system as a next of kin)?

Sounds like this is needed for 2.6, so adding `needs refinement` and `needs design` for how can we get a quick but robust solution for this! We can iterate in future but don't want to cause data issues!

## Alternatives

- Don't make next of kin a patient at all! Instead use the OG `next_of_kin_relative` field, which is just a text field. Expose this on name to continue to allow ease of filtering
  - Means no linking for any patients

lache-melvin:
Fixes #6410
Fixes #6355

# 👩🏻‍💻 What does this PR do?

Support plain text entry of next of kin name. Using existing SearchWithUserSource component, thanks Clemens 🔥 

Base patient input now looks like this:
Screenshot 2025-02-05 at 1 38 31 PM

You can either search for existing patient, and matching options are available for you to select:
Screenshot 2025-02-05 at 1 38 47 PM
Screenshot 2025-02-05 at 1 38 51 PM

Or type in plain text:
Screenshot 2025-02-05 at 1 39 07 PM

This means we store linked patient name ID if selected, or just a name for faster entry.

Updates report for just `next of kin` field:
Screenshot 2025-02-05 at 1 42 33 PM

(Note, phone now looks at the patient's own phone field. Idea is that if a country specific implementation needs the next of kin phone, we can change the label in the custom patient schema, and populate from selected next of kin if there!)

Also added Next of Kin column to patient list bc it was easy and I wanted to check everything was coming through good :)
Screenshot 2025-02-05 at 1 41 30 PM

# 🧪 Testing

- [ ] Be using the default OMS patient schema
- [ ] Create/edit 2 patients
- [ ] First one - search for another existing patient to be the next of kin
  - [ ] Correct options appear, selecting on disables the input and shows the name
  - [ ] Save/refresh, it persists
- [ ] Second one - type in some name and don't select a matching patient
- [ ] Go back to patient list view - both patients show the correct next of kin name
- [ ] Enrol both patients in program and schedule an encounter for each
- [ ] Upsert latest reports: `./target/debug/remote_server_cli upsert-reports -o`
- [ ] Generate Pending Encounters report and ensure next of kin name is correct in report

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1677,9 +1677,6 @@
   "report.net-decrease": "Net Decrease",
   "report.net-increase": "Net Increase",
   "report.net-variance": "Net Variance",
-  "report.next-of-kin-first-name": "Next of kin's first name",
-  "report.next-of-kin-last-name": "Next of kin's last name",
-  "report.next-of-kin-phone": "Next of kin's phone",
```
lache-melvin, line 1682: no longer needed

client/packages/programs/src/JsonForms/components/Search/SearchWithUserSource.tsx
```diff
@@ -100,54 +95,44 @@ export const SearchWithUserSource = (
         renderers={renderers}
         enabled={!isPatientSelected}
       />
-      {(isPatientSelected || results.length > 0) && (
-        <DetailInputWithLabelRow
```
lache-melvin, line 104: this diff is just remove this component wrapper, seeing as there was no label...

client/packages/system/src/Patient/PatientView/Footer.tsx
```diff
@@ -104,7 +104,7 @@ export const Footer: FC<FooterProps> = ({
                 gap={2}
               >
                 <Typography sx={{ fontSize: 18, fontWeight: 700 }}>
-                  Document Edit History
+                  {t('label.document-edit-history')}
```
lache-melvin, line 107: i found more!

client/packages/system/src/Patient/PatientView/PatientView.tsx
```diff
@@ -164,7 +165,13 @@ const PatientDetailView = ({
           isDeceased: currentPatient.isDeceased ?? undefined,
           phone: currentPatient.phone ?? undefined,
           address1: currentPatient.address1 ?? undefined,
-          nextOfKinId: currentPatient.nextOfKinId ?? undefined,
+          nextOfKin:
+            currentPatient.nextOfKinId || currentPatient.nextOfKinName
+              ? {
+                  id: currentPatient.nextOfKinId ?? undefined,
+                  name: currentPatient.nextOfKinName ?? undefined,
```
lache-melvin, line 172: small bit of ick - works best for JSON schema to have nextOfKin as an object, but it needs to be a _different_ nextOfKin object than the `nextOfKin` field on `currentPatient`, which is a whole PatientNode from the backend, only if the nextOfKinId is set. 

Found the best was just to query the two distinct fields from API, and build the JSON form nextOfKin object here, then map back to separate fields on save 🥲

server/repository/src/migrations/v2_06_00/add_name_next_of_kin_id.rs
```diff
@@ -16,7 +16,7 @@ impl MigrationFragment for Migrate {
             "#
         )?;
 
-        // Reset translate all items on the next sync
+        // Reset translate all names on the next sync
```
lache-melvin, line 19: Not resetting integration_datetime in the new migration seeing as it is already in this one 😁

Chris-Petty (approved):
tested, works gud

PR #6426, opened 2025-02-05
6410 fix backend patient schema

### LINKED ISSUE #6410: Quick add patient from `Next of Kin` selector

lache-melvin:
Fixes #6410

# 👩🏻‍💻 What does this PR do?

Updates backend NextOfKin schema field to match the frontend one, and the one on base patient in programschemas repo now (in PR)

# 🧪 Testing

- [ ] Patient next of kin add/edit still works with default patient
- [ ] Upload these schemas to document registry: https://github.com/msupply-foundation/programschemas/pull/14
- [ ] Patient next of kin add/edit works with custom patient
- [ ] 

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/programs/src/schema_types.ts
+++ b/client/packages/programs/src/schema_types.ts
@@ -85,6 +85,8 @@ export interface PersonSchema {
 
   notes?: NoteSchema[];
 
+  phone?: string;
+
   nextOfKin?: NextOfKinSchema;
 }
 
--- a/server/service/src/programs/patient/patient_schema.rs
+++ b/server/service/src/programs/patient/patient_schema.rs
@@ -33,8 +33,7 @@ impl Default for SchemaPatient {
             marital_status: Default::default(),
             contacts: Default::default(),
             extension: Default::default(),
-            next_of_kin_id: Default::default(),
-            next_of_kin_name: Default::default(),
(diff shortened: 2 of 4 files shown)


Chris-Petty (approved):
🔥

PR #6432, opened 2025-02-06
6297 Restructure reports file system

### LINKED ISSUE #6297: Restructure report file system

A number of report file system restructures and re-namings:
- [x] Separation of reports into separate 'forms' (previously called picklists) and 'reports' dirs.
- [x] Rename 'reports' dir to be 'standard_reports' dir.
- [x] Rename 'manifest.json' files to 'report-manifest.json' in all reports (and update cli to read report-manifest.json).
- [x] Move reports dir out of the backend, and into the open-msupply root dir
- [x] Simplify version dir names to only major and minor (with patch changes, we will just overwrite).
Currently file names are 'x.x.x', but we can simplify to 'x.x'. When new patches are made, we overwrite the old patch (no need to keep old versions of patches).

### Moneyworks Jobcode 🧰

OMS:REPT

EthanMcQ-TMF:
Fixes #6297 

# 👩🏻‍💻 What does this PR do?

- Move `reports` dir to root of repo, rename to `standard_reports`
- Split `standard_forms` into a separate dir
- Standardise `manifest.json` files to `report-manifest.json`
- Remove patch numbers from report dir names
- Update CLI to point to new dirs

Splitting the forms at the same dir level as the reports was a style choice on my part, but it does result in the default `build-reports` command now skipping forms. I'm not entirely sure if that's something we want in this new dir layout, so I'm happy to adjust that if needed.

# 🧪 Testing

- [ ] Confirm `build-reports` command correctly builds from the newly renamed directories
- [ ] Confirm `upsert-reports` command correctly uploads from the newly renamed directories
- [ ] Confirm `reload-embedded-reports` command correctly loads from the newly renamed directories

# 📃 Documentation

- [ ] Updated the path in the report-builder README


roxy-dao (changes requested):
Thanks for the changes Ethan~ Just need to fix two things before this can be merged (: 

1. The standard forms aren't being built anymore 
2. The requisition report in `standard_report` folder is a form
3. .gitignore path for node_modules and dist need to be amended

server/cli/src/cli.rs
```diff
@@ -421,70 +421,82 @@ async fn main() -> anyhow::Result<()> {
...
-
-            fs::write(&output_path, serde_json::to_string_pretty(&reports_data)?).map_err(
-                |_| {
-                    anyhow::Error::msg(format!(
-                        "Failed to write to {:?}. Does output dir exist?",
-                        output_path
-                    ))
-                },
-            )?;
+            for base_dir in dir_list {
```
EthanMcQ-TMF, line 430: For the sake of running `standard_reports` and `standard_forms` back to back I wanted to loop this. Fun bonus is now you can build multiple report paths in a single command if you want.

```
cargo run --bin remote_server_cli build-reports --path ../some/example/path ../some/other/path
```

server/cli/src/cli.rs
```diff
@@ -421,70 +421,82 @@ async fn main() -> anyhow::Result<()> {
...
+                None => vec![standard_reports_dir, standard_forms_dir],
+            };
 
-            StandardReports::upsert_reports(reports_data, &con, overwrite)?;
+            for file_path in file_list {
+                let json_file = fs::File::open(file_path.clone())
+                .expect(&format!(
+                    "{} not found for report",
+                    file_path.display()
+                ));
```
roxy-dao, line 491:
```suggestion
                    .unwrap_or_else(|_| panic!("{} not found for report", file_path.display()));
```

roxy-dao (approved):
LGTM~ Thanks for all the changes (:

PR #6435, opened 2025-02-06
add nothing found feedback to new prescription modal

### LINKED ISSUE #5689: New prescription -- patient search gives no useful feedback when no results found
## What went wrong? 😲

When creating a new prescription, if your search term for the patient yields no results, there is no visual feedback for the user. You get a brief "Loading" placeholder, then nothing. As a user, it feels like something has gone wrong rather than no results.

## Expected behaviour 🤔

A message saying "No patients found" or equivalent

## How to Reproduce 🔨

1. Prescriptions -> New Prescription
2. Type some random characters into the Patient search
3. Observer no useful feedback

## Your environment 🌱

- Open mSupply Version: 2.4.01-RC1
- Legacy mSupply Central Server Version: n/a

- Platform: Browser (Firefox)

- Database type: Postgres

zachariah-at-msupply:
Fixes #5689 

# 👩🏻‍💻 What does this PR do?

Adds feedback to the patient search in the create prescription modal, so that you know when no patients are found.

This is what it looks like when you open the modal:
&nbsp;
image
&nbsp;
And this is what it looks like when you search for something that matches no patients:
&nbsp;
image
&nbsp;
It is not ideal that the feedback covers the create patient button, but apart from moving that button, I was unable to find a better solution without taking days over it, which seemed excessive. Perhaps one possibility might be to move the create patient button down a bit - what do you think about this, those who are in charge of UI?

Have a nice day :)

# 🧪 Testing

- [ ] Go to Dispensary > Prescriptions > New Prescription (in the top right hand corner of the page).
- [ ] Check that 'No patients found' is displaying beneath the search box.
- [ ] Enter a search term that displays some results.
- [ ] Check that the results appear as expected, without 'No patients found'.
- [ ] Enter a search term that is definitely not going to return any results.
- [ ] Check that 'No patients found' is displaying beneath the search box.

# 📃 Documentation

- [x] **These areas should be updated or checked**: Screenshots should be updated in the make a new prescription documentation.


jmbrunskill:
> Perhaps one possibility might be to move the create patient button down a bit - what do you think about this, those who are in charge of UI?

Can we try to put it at the bottom of the modal? I think that would solve it and look nicer?

roxy-dao:
> > Perhaps one possibility might be to move the create patient button down a bit - what do you think about this, those who are in charge of UI?
> 
> Can we try to put it at the bottom of the modal? I think that would solve it and look nicer?

Or move the create patient button to the top >.>

zachariah-at-msupply:
I'll just wait for a final decision and then go ahead and implement it :)

jmbrunskill:
> I'll just wait for the decision to be finalised and then go ahead and implement it, to avoid doing lots of commits :)

Put it at the bottom of a modal if you can figure out a nice way, I just had a quick look and couldn't figure out how, maybe see if you can figure it out in 30min?
If it's too difficult to put it at the bottom of the modal, Then lets just add some padding so there's just enough room for the message?

Moving the Create button to the top seems a bit risky as people might click it instead of searching.
We want to improve this more the the future but it's not the priority right now...

zachariah-at-msupply:
Here it is! Is this what you had in mind?
image

client/packages/system/src/Patient/Components/PatientSearchModal/PatientSearchModal.tsx
```diff
@@ -65,10 +65,10 @@ const PatientSearchComponent: FC<PatientSearchModalProps> = ({
           onChange={(_, name) => {
             if (name && !(name instanceof Array)) onChange(name);
           }}
-          noOptionsText=""
+          noOptionsText={t('messages.no-patients-found')}
         />
       </Box>
-      <Box p={2}>
+      <Box marginTop="100%" padding={2}>
```
GeronimoJohn, line 71: Like the idea! Though I would suggest using or adding `flexGrow` within the `sx` field for the `Box` above this one (`sx: {{ flexGrow: 1}}`) and remove marginTop. 

Margins are great tool but can usually lead to layout issues unlike flex box which has more of a predictable outcome. `flexGrow: 1` allows the element to take up as much space. Effectively reaching the desired layout

Screenshot 2025-02-10 at 9 08 42 AM

GeronimoJohn (commented):
Great idea! Though left one thing about it

client/packages/system/src/Patient/Components/PatientSearchModal/PatientSearchModal.tsx
zachariah-at-msupply: Awesome trick! I'll do that :+1:

GeronimoJohn (approved):
LGTM! Thanks for the changes @zachariah-at-msupply. If we're all happy with the location of the button then 🚀

PR #6450, opened 2025-02-09
restore the border

### LINKED ISSUE #6449: Multi line input style has changed
## What went wrong? 😲

The multi line input has lost the border it once had:

Image

## Expected behaviour 🤔
The input used to have a border like this:

Image

There needs to be some demarcation to show where the input is. Please investigate and see why this happened - if it was deliberate then isolate your solution so that the settings looks ok, and the original intention isn't overwritten

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #6449

# 👩🏻‍💻 What does this PR do?
Restores the border on the settings inputs

It's another MUI6 problem. 
* The slotProps needed to be on the `input` rather than the `htmlInput`
* Within the props, the `sx` provided by the caller was being overwritten 

The borders were missing and it's hard to see where the input is now!

### Side quest
Fix the background colour on the numeric inputs for prescriptions. These are showing as grey and should be white:

Screenshot 2025-02-09 at 4 52 59 PM

# 🧪 Testing

- [ ] Check the settings multi-line inputs; they should have a border
- Check other usages to see if they look ok:
  - [ ] help form email body input
  - [ ] create encounter form notes input
  - [ ] Customer requisition comment input
  - [ ] Internal order comment input

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
+++ b/client/packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
@@ -1,6 +1,7 @@
 import React, { FC } from 'react';
 import { StandardTextFieldProps } from '@mui/material/TextField';
 import { BasicTextInput } from '../TextInput';
+import merge from 'lodash/merge';
 
 export const TextArea: FC<StandardTextFieldProps> = ({
   value,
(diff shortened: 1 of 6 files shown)


roxy-dao (approved):
LGTM~

PR #6460, opened 2025-02-10
6417 insurance graphql and service layer query

### LINKED ISSUE #6417: Dispensing: Insurance GraphQL and Service Layer Query
## Insurances GraphQL and Service Layer Query

Add a GraphQL endpoint for `insurances` within the general queries directory

Add the Service layer ensuring the GraphQL endpoints able to retrieve data from the database

This depends on #6204

## Tasks

- [ ] Add GraphQL endpoint - add queries
- [ ] Add Service Layer - ensuring GraphQL's endpoint are able to perform operations needed

## Acceptance Criteria 🧾

- [ ] GraphQL endpoint for getting insurance related to a patient

### Epic Link 🔗

#6000

### Moneyworks Jobcode 🧰

OMS:CIDV

GeronimoJohn:
Fixes #6417 

# 👩🏻‍💻 What does this PR do?

Adds an API or GraphQL endpoint for retrieving a patient's insurance data. The query requires the patient's ID to return the correct information.

# 🧪 Testing

- [ ] Start the OMS server locally
- [ ] Open the GraphQL playground
- [ ] Write and execute a query for insurance data
- [ ] Add the appropriate parameters
- [ ] Retrieve the data!

# 📃 Documentation

  1.
  2.


server/graphql/general/src/queries/insurances.rs
GeronimoJohn, line 1: Debating wether to move the types to the types folder

server/graphql/general/src/queries/insurances.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+#[derive(PartialEq, Debug)]
+pub struct InsuranceNode {
+    insurance: NameInsuranceJoinRow,
+}
+
+#[Object]
+impl InsuranceNode {
```
GeronimoJohn, line 62: Realised that pattern follows `row()` rather than the node name. Will update after review

server/graphql/general/src/queries/insurances.rs
```diff
@@ -0,0 +1,189 @@
...
+pub struct InsuranceConnector {
+    nodes: Vec<InsuranceNode>,
+}
+
+impl InsuranceConnector {
+    pub fn from_domain(insurances: Vec<NameInsuranceJoinRow>) -> InsuranceConnector {
+        InsuranceConnector {
+            nodes: insurances
+                .into_iter()
+                .map(|insurance| InsuranceNode { insurance })
```
GeronimoJohn, line 118: Found that we impl a method for this. Will update after review

server/graphql/general/src/queries/insurances.rs
```diff
@@ -0,0 +1,199 @@
...
+    ctx: &Context<'_>,
+    store_id: String,
+    name_link_id: String,
+    filter: Option<InsuranceFilterInput>,
+    sort: Option<Vec<InsuranceSortInput>>,
+) -> Result<InsuranceResponse> {
+    let user = validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::QueryInsurances,
```
jmbrunskill, line 128:
```suggestion
            resource: Resource::QueryPatient,
```

I think we want to link this to if you're allowed to see patients, you're allowed to see their insurances?
What you have is probably ok, but I think this is more correct.

server/repository/src/db_diesel/name_insurance_join_row.rs
```diff
@@ -1,12 +1,29 @@
...
+    diesel_macros::{apply_equal_filter, apply_sort},
+    repository_error::RepositoryError,
+    Sort, Upsert,
+};
 use diesel::prelude::*;
 use diesel_derive_enum::DbEnum;
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Default, PartialEq, Debug)]
+pub struct NameInsuranceJoinFilter {
```
jmbrunskill, line 16: I don't see any need for this filter in the frontend, as far as I know we don't need to filter in that view other than by the patient name which we're passing in.
I'd be tempted to just remove the filter entirely....

server/graphql/general/src/queries/insurances.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    pub async fn expiry_date(&self) -> &NaiveDate {
+        &self.row().expiry_date
+    }
+
+    pub async fn is_active(&self) -> &bool {
+        &self.row().is_active
+    }
+
+    pub async fn entered_by_id(&self) -> &Option<String> {
```
jmbrunskill, line 103: I'm not sure this is being populated by mSupply? Maybe best to leave it out for now, we're not displaying it right?

jmbrunskill (dismissed):
Looks great!

Can you please
- [x] Change the permissions to use `QueryPatient` (editing/adding will require `MutatePatient`)
- [x] Change/remove usage of name_link_id in graphql
- [x] Remove `Resource::QueryInsurances`
- [x] Optional: Remove Filter from query, don't think we'll use it and just adds to compile time...

server/repository/src/db_diesel/name_insurance_join_row.rs
GeronimoJohn: I was thinking about too, doesn't seem like we need it. Will remove it!

server/graphql/general/src/queries/insurances.rs
```diff
@@ -0,0 +1,199 @@
...
+    insurance: NameInsuranceJoinRow,
+}
+
+#[Object]
+impl InsuranceNode {
+    pub async fn id(&self) -> &str {
+        &self.row().id
+    }
+
+    pub async fn name_link_id(&self) -> &str {
```
jmbrunskill, line 67: We're not supposed to expose `name_link_id` directly in the graphql.
The idea is that the backend should handle it and the frontend can just assume it has the right name.

In this case I think we can just leave this out of the response as we already querying based on a name, we don't need to have this in the response? Or if we do, then we could use the Name Row Loader to return the name information rather than just the name_link_id which would require another look up anyway...

GeronimoJohn: That makes sense!

server/repository/src/db_diesel/name_insurance_join_row.rs
```diff
@@ -1,12 +1,29 @@
...
+pub enum NameInsuranceJoinSortField {
+    ExpiryDate,
+    IsActive,
+}
+
+pub type NameInsuranceJoinSort = Sort<NameInsuranceJoinSortField>;
+
 #[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 #[DbValueStyle = "SCREAMING_SNAKE_CASE"]
```
jmbrunskill, line 29: I think this is probably a good approach for joining to name but not needing to return it...

https://github.com/msupply-foundation/open-msupply/compare/6417-patient-insurance-graphql-and-service-layer...6417-name-link-suggestion

jmbrunskill (approved):
Thanks John looks good!

PR #6476, opened 2025-02-10
6459 Can send blank internal orders 

### LINKED ISSUE #6459: Can send blank internal orders
## What went wrong? 😲

I sent an internal order with no lines D:

## Expected behaviour 🤔

You should not be able to send blank internal orders in a _regular setup_ - N.B. There may be a valid exception for program orders that have indicator data attached! I am not sure on this front. That said you can make a blank requisition with indicator data too atm, all the lines with 0 stock from the master list get pruned.

## How to Reproduce 🔨

1. Make internal order
2. Add a line
3. Mark as "sent"
4. The line gets pruned because it has 0 requested
5. So an empty requisition is sent

Note I do NOT have the pref set to allow 0 requested quantities on lines, besides my interpretation is that this pref should prevent the pruning of these lines.

Image

## Your environment 🌱

- Open mSupply Version: 2.5.1 Android

roxy-dao:
Fixes #6459

# 👩🏻‍💻 What does this PR do?
Don't allow users to finalise Internal Orders if:
1. There are no lines
2. All lines have requested quantity = 0

# 🧪 Testing

- [ ] Shouldn't be able to send IO
- [ ] Add lines but don't put anything for requested quantity
- [ ] Shouldn't be able to send

# 📃 Documentation

  1.
  3.

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1254,7 +1254,7 @@
   "messages.cant-download-android": "File cannot be downloaded on this device",
   "messages.cant-return-shipment": "Cannot process return of lines until the status is 'Shipped'",
   "messages.cant-return-shipment-replenishment": "Cannot return lines until the status is 'Delivered'",
-  "messages.cant-send-order": "Cannot send Internal Order because there are no lines",
+  "messages.cant-send-order": "Cannot send Internal Order because there are no lines or because all lines have a requested quantity of 0",
   "messages.catalogue-property": "This property is defined in the catalogue",
   "messages.cce-created": "CCE created successfully",
   "messages.change-server": "Change server",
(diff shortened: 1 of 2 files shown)


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1254,7 +1254,7 @@
   "messages.cant-download-android": "File cannot be downloaded on this device",
   "messages.cant-return-shipment": "Cannot process return of lines until the status is 'Shipped'",
   "messages.cant-return-shipment-replenishment": "Cannot return lines until the status is 'Delivered'",
-  "messages.cant-send-order": "Cannot send Internal Order because there are no lines",
+  "messages.cant-send-order": "Cannot send Internal Order because there are no lines or because all lines have a requested quantity of 0",
```
noel-yeldos, line 1257: NIT - Would this be better as two separate toasts? Only because that seems like a lot to read in one toast. If no lines then show no lines toast and if line but 0 requested quantity then show 0 requested quantity toast.

noel-yeldos (approved):
LGTM! Tested with no lines and with lines but no requested quantity.

client/packages/common/src/intl/locales/en/common.json
roxy-dao: It's being checked in the same check so it's fine.

PR #6480, opened 2025-02-11
6086 Unclear which store's SOH & AMC are shown when clicking on an item in Requisitions

### LINKED ISSUE #6086: Unclear which store's SOH & AMC are shown when clicking on an item in Requisitions
## What went wrong? 😲

When creating a requisition and clicking on an item you are supplying, it is unclear which store's details are shown on the screen - are those SOH and AMC the issuing store's or the receiving/requesting one's?

Screenshot 2025-01-16 at 12 18 00 PM

Also, as always, some column header inconsistencies 

## Expected behaviour 🤔

1. Can we please specify if it's 'Our SOH & AMC' or 'Their SOH & AMC'
2. If possible, can we add both - e.g. Our SOH & AMC and Their SOH & AMC to the same page, so the user can view both?
3. Why is it 'Stock on hand' but 'AMC'? Can it please be either SOH & AMC _or_ Stock on Hand & Average Monthly Consumption?
4. Between the list view and the specific view, there are column header inconsistencies too: 

Screenshot 2025-01-16 at 12 23 16 PM

5. Our SOH & Their SOH in the list view but 'Stock on hand' in the detailed view - maybe should just be SOH in the detailed view too (to the 3d point above)
6. Only 'Suggested' has Quantity next to it, and the rest just have the name - e.g. Requested, Supply etc. Can either all of them have 'Qty' next to them or none at all?
7. Continuing the point, when in detailed view, only 3 out of 5 lines have 'Quantity' added - can all of them also please be either with Quantity, or none at all? And make them align with the list view - whatever the decision is for point 6, should be the same for point 7

Screenshot 2025-01-16 at 12 25 23 PM

## How to Reproduce 🔨

2. Create a new Requisiton
3. Add some items
4. Play around with the list view and the detailed view
5. See the discrepancies 

## Your environment 🌱

- Database type: PostgreSQL

roxy-dao:
Fixes #6086

# 👩🏻‍💻 What does this PR do?
Add their/our stock on hand label instead of just stock on hand for Requisitions and Internal Orders + show our stock on hand for Requisitions

Screenshot 2025-02-11 at 2 56 11 PM

Screenshot 2025-02-11 at 2 56 05 PM

# 🧪 Testing

- [ ] Go to requisition line edit
- [ ] See label has clear labelling now 
- [ ] See that there is a column representing store's soh
- [ ] Go to Internal Orders
- [ ] See label has clear labelling now

# 📃 Documentation

  1. Screenshots for requisition and internal order line edit page
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
@@ -115,7 +115,6 @@ export const RequestLineEdit = ({
                     width={INPUT_WIDTH}
                     value={draft?.itemStats.availableStockOnHand}
                     disabled
-                    autoFocus
                   />
                 }
                 labelWidth={LABEL_WIDTH}
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
@@ -119,7 +119,7 @@ export const RequestLineEdit = ({
                   />
                 }
                 labelWidth={LABEL_WIDTH}
-                label={t('label.stock-on-hand')}
+                label={t('label.our-soh')}
                 sx={{ marginBottom: 1 }}
               />
               {isProgram && useConsumptionData && (
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEdit.tsx
@@ -114,6 +114,19 @@ export const ResponseLineEdit = ({
           <>
             <Box paddingLeft={4} paddingRight={7}>
               {/* Left column content */}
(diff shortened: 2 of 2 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
```diff
@@ -115,11 +115,10 @@ export const RequestLineEdit = ({
                     width={INPUT_WIDTH}
                     value={draft?.itemStats.availableStockOnHand}
                     disabled
-                    autoFocus
```
noel-yeldos, line 118: What does removing autoFocus do for a field that is disabled? I tried adding it back but didn't notice any difference. Was the field initially editable - is that why it had the autoFocus in the first place?

noel-yeldos (approved):
Looks good to me! Tested and labels appear as expected. Just had a general question about the autoFocus prop

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
roxy-dao: There shouldn't be an autofocus on the disabled field. We want it to autofocus on the first input so that the user doesn't have to click the first input to input something

PR #6508, opened 2025-02-13
Refactor utils to have overridable name as displaying report name

### LINKED ISSUE #6385: Update report name translations

A few things that would help report name translations to make them more useable.

Report name translations are currently done on the front end here: client/packages/reports/src/utils.ts
This replaces the report 'code' with a string which we translate in common.json.

This makes it slightly difficult to manage custom reports:
- We are required to add custom report translation strings into the main OMS repo for specific clients
- Doesn't allow for hard coding of report names in custom reports
- Adds restrictions when a custom report is used to replace a standard report (ie only one translation per locale) 
- Translations and sub translation functions are conducted in more places than necessary.

Also side note is that this method is not documented which adds to confusion.

- [ ] Remove front end string manipulation
- [ ] Translate based on 'name' rather than 'code'
- [ ] Display 'name' translated without any conversions or simply display name.
This would allow us to use report.standard-report as the name, or alternatively a hard coded string for standard reports

### Moneyworks Jobcode 🧰

OMS:CIRDV

fergie-nz:
Fixes #6385 

# 👩🏻‍💻 What does this PR do?

Alternative front end translation manipulation so that raw name is displayed when translation function goes to fallback.

I had wondered about changing translation to work on code rather than name, and have the name field as an optional override. This would allow consolidation of translation manipulation into less places.
However, this is so much more simple and is backwards compatible, so opted for this option.

Allows reports to show raw name field as name without adding client specific translation strings for custom reports.

# 🧪 Testing

- [ ] Try editing the name of a report to a custom name. Any standard report is suitable for this (must be the latest version of that report to display
- [ ] Regenerate reports with `cargo build && ./target/debug/remote_server_cli build-reports && ./target/debug/remote_server_cli upsert-reports -o`
- [ ] View raw report name on OMS

# 📃 Documentation

- [ ] **Part of an epic**: Readme to be updated after [readme update PR](https://github.com/msupply-foundation/open-msupply/issues/6385) is merged

### DIFF
--- a/client/packages/reports/src/utils.ts
+++ b/client/packages/reports/src/utils.ts
@@ -2,8 +2,8 @@ import { LocaleKey, useTranslation } from '@common/intl';
 
 export const translateReportName = (
   t: ReturnType<typeof useTranslation>,
-  reportName: String
+  reportName: string
 ) => {
   let key = `report.${reportName.replace(/ /g, '-').toLowerCase()}` as LocaleKey
-  return (key == t(key)) ? reportName.toString() : t(key);
+  return (key == t(key)) ? reportName : t(key);
 };
--- a/client/packages/reports/src/utils.ts
+++ b/client/packages/reports/src/utils.ts
@@ -2,7 +2,8 @@ import { LocaleKey, useTranslation } from '@common/intl';
 
 export const translateReportName = (
   t: ReturnType<typeof useTranslation>,
-  reportName: string
+  reportName: String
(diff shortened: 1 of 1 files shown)


noel-yeldos (approved):
LGTM! Tested and name changes as expected.

Screenshot 2025-02-18 at 10 16 19
Screenshot 2025-02-18 at 10 16 28

Issue #6343, opened 2025-01-29
Customised Report: Stockouts


roxy-dao:
Displays a list of items which have had a stockout event

Filter: Date range

| Column | Description |
| --- | --- |
| Item Code | - | 
| Item name | - |
| Number of days of stockout | Number of consecutive days for which the SOH was 0 | 
| Program name | - |

Moneyworks Jobcode 🧰
OMS:CIRDV

alainsussol:
# Testing
- [ ] These columns are present and are translated
   - [ ] Item Code **-PASS-**
   - [ ] Item name **-PASS-**
   - [ ] Number of days of stockout | Number of consecutive days for which the SOH was 0 **-PASS-**
   - [ ] Program name **-PASS-**
 - [ ] Test SOH when nbr of day of SOH are summed when separated by non stockout days **-PASS-**
 - [ ] Use filter to make sure that only stockout days in the period specified are included  **-PASS-**
         - [ ] Have the filter be in a period that start in the middle of stockout period to see if only the day of the period are included

alainsussol:
@noel-yeldos @mark-prins @roxy-dao 

Everything works as I understand it is intended however there is this case when using the filter where some items that are out of stock will show 0 days out of stock because the Stockout event ie having stock and then not having stock will be outside the period specified in the filter.

So lets says that you had stock for an item on the 2nd of January 2025 and then you issued it all on the 3rd of January. If you run the report for the period that starts on the 1st of January to today (1st of April), the report will show you the right amount of days ie 87 days
https://drive.google.com/file/d/1yeKqCNd1B2vPmYKSs75gq9rrWRXniUel/view?usp=sharing

If you run the report from the 10th of January it indicates 0 stockout days because the event is outside the period specified by the filter. Part of me finds this a bit problematic.
https://drive.google.com/file/d/1bMg0gAYTwkjyXfZQG_2-8GHeiDheIONW/view?usp=sharing

 What do you think ?

roxy-dao:
> If you run the report from the 10th of January it indicates 0 stockout days because the event is outside the period specified by the filter. Part of me finds this a bit problematic.

Are you running the report until an end date? Or just with a start date for the

alainsussol:
until an end date

PR #6514, opened 2025-02-14
6343 Add datetime filter to itemledger endpoint for customised stockouts report

noel-yeldos:
Translation and updating endpoint for #6343

# 👩🏻‍💻 What does this PR do?
Add translation for stockouts and also add datetime filter to itemledger endpoint

I'm not sure if I've added the datetime filter correctly because it isn't working at the moment. Would like some guidance there pls

# 🧪 Testing

See for report PR https://github.com/msupply-foundation/open-msupply-reports/pull/144

# 📃 Documentation

  1.
  2.


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1720,6 +1720,7 @@
   "report.stock-take-sheet": "Stock take sheet",
   "report.stock-take-variance-report": "Stock Take Variance Report",
   "report.stock-to-date": "Stock to date",
+  "report.stockouts": "Stockouts",
```
fergie-nz, line 1723: I think won't need this in common.json.

Can hard code the report name as it is a custom report

fergie-nz (changes requested):
Looks great @noel-yeldos,

I think you'll also need to add pagination to this change. Otherwise you will only query a small subset of lines when querying for the report.

client/packages/common/src/intl/locales/en/common.json
roxy-dao: Pagination for ledger defaults to unlimited so it's fine.

roxy-dao:
> Looks great @noel-yeldos,
> 
> I think you'll also need to add pagination to this change. Otherwise you will only query a small subset of lines when querying for the report.

Pagination for ledger defaults to unlimited so it's fine.

client/packages/common/src/intl/locales/en/common.json
noel-yeldos: Removed here dd19cf0c2cb1eff7bd601c4e09fc1049021fec18

Issue #6505, opened 2025-02-12
Add pref to enable canceling prescriptions


jmbrunskill:
Cancelling prescriptions is (I think) behind a customisation code in mSupply.
We probably want to create a global pref to allow customers to disable this feature if they don't want to use it.

jmbrunskill:
Waiting on #5729 & implementation

jmbrunskill:
Adding needs triage  to see if it should be prioritised for 2.6, 2.7 or Just dropped?

jmbrunskill:
@mark-prins do we want this pref, or should we just allow every customer to cancel untill requested to add this?

mark-prins:
there isn't a preference in OG; only the 'can cancel finalised invoices' user permission.
I would like to have that brought through and checked before cancellation - no need for a preference though thanks

Issue #6500, opened 2025-02-12
On Prescription cancellation, create return to balance ledger


CarlosNZ:
When "Cancelled" a Reverse Invoice is to be automatically generated in the back-end in order to balance the ledger, and marked as `is_cancellation:true`

This extra invoice will not show in the UI (only the single cancelled invoice will appear in the Prescription list)
However, in the ledger you'll see two records one for the cancellation and one for the original prescription.

![Image](https://github.com/user-attachments/assets/a0c403b5-98e5-4764-9ad4-06d09a5caaa8)

Note -- need to check [this](https://github.com/msupply-foundation/open-msupply/pull/6516/files#r1955522714)

jmbrunskill:
In mSupply this shows as a record Prescription with the same status, but negative quantities.
I think in open-mSupply we're going with a whole different type e.g. `CustomerReturn` 

I haven't looked into how we're mapping that with sync to mSupply.

In mSupply you see both `prescriptions` but I think we don't want to show both transactions in OMS, just the single transaction as cancelled.
<img width="879" alt="Image" src="https://github.com/user-attachments/assets/f9723d43-df72-4d70-bf3b-f7cd287399ab" />

Here's the sync record for the example above (notice the "is_cancellation": true,`) on the new record, but the original one is unchanged in mSupply Land.

```
{
    "Colour": 0,
    "Date_order_received": "0000-00-00",
    "Date_order_written": "0000-00-00",
    "ID": "EAB1E4D03C1D4C78AF8B55A8E92BA38D",
    "amount_outstanding": 0,
    "arrival_date_actual": "0000-00-00",
    "arrival_date_estimated": "0000-00-00",
    "authorisationStatus": "",
    "budget_period_ID": "",
    "category2_ID": "",
    "category_ID": "",
    "comment": "Cancellation of customer invoice 16",
    "confirm_date": "2025-02-11",
    "confirm_time": 73114,
    "contact_id": "",
    "currency_ID": "currency_a",
    "currency_rate": 1,
    "custom_data": null,
    "diagnosis_ID": "",
    "donor_default_id": "",
    "encounter_id": "",
    "entry_date": "2025-02-11",
    "entry_time": 73114,
    "export_batch": 0,
    "finalised_date": "2025-02-11",
    "finalised_time": 73114,
    "foreign_currency_total": 0,
    "goodsReceivedConfirmation": null,
    "goods_received_ID": "",
    "hold": false,
[...]

PR #6516, opened 2025-02-14
6499 Cancel verified prescriptions

### LINKED ISSUE #6499: Allow cancelling of verified prescriptions
In UI, can select "Cancelled" (not sure the design currently)

CarlosNZ:
Fixes #6499

Just takes care of the actual cancellation, we still need to generate another invoice to reverse the stock changes in #6500, and hide it behind a preference in #6505.

# 👩🏻‍💻 What does this PR do?

- Adds an option to "Cancel" the current prescription in the "More" side panel, but only if the prescription is already "Verified". Presents a confirmation dialog before cancelling:

Screenshot 2025-02-14 at 4 33 39 PM

- Show cancelled in the Status bar, but only if Cancelled. Will stop at Verified otherwise:

Screenshot 2025-02-14 at 4 33 57 PM

- Make Cancelled prescriptions greyed out in the list (like Verified), and prevent them from being edited:

Screenshot 2025-02-14 at 4 33 51 PM

- migration to add "Prescription Cancelled" status to activity log enum
- make "Cancelled" available as a status in GraphQL Upsert node

Still to-do: I'm just using the "Verified" date-time for cancelled at the moment -- not sure where to find the "Cancelled datetime" yet. And I don't think we fully figured out how we're going to handle it.

Also -- note the "activity log" migration is added to the invoice status migration (from yesterday), so you might to manually reset that migration fragment to get it to work.

# 🧪 Testing

Have some prescriptions in your list..

- [ ] Make sure no option to "Cancel" makes itself known until presription is "Verified"
- [ ] Once "Verified", there is a "Cancel" button in the "More" side panel
- [ ] "Verified" prescriptions should *Not* show a "Cancelled" step in the Status Progress (footer)
- [ ] Click "Cancel" -- should show confirmation:
  - [ ] Can close modal without cancelling
  - [ ] Click "OK" -- prescription gets updated with "Cancelled" status in status bar (footer)
- [ ] "Cancelled" prescriptions appear greyed out in Prescriptions list
- [ ] "Cancelled" prescriptions can not be edited anywhere


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1133,6 +1133,7 @@
   "label.variants": "Variants",
   "label.ven": "VEN",
   "label.verified": "Verified",
+  "label.cancel": "Cancel",
```
CarlosNZ, line 1136: We have a button with "Cancel" text, but I figured these could actually be different words in other languages.

client/packages/common/src/types/schema.ts
```diff
@@ -3290,6 +3291,56 @@ export type InsertVaccineCourseInput = {
...
+  isActive: Scalars['Boolean']['output'];
+  policyNumber: Scalars['String']['output'];
+  policyNumberFamily?: Maybe<Scalars['String']['output']>;
+  policyNumberPerson?: Maybe<Scalars['String']['output']>;
+  policyType: InsurancePolicyNodeType;
+};
+
+export enum InsurancePolicyNodeType {
+  Business = 'BUSINESS',
+  Personal = 'PERSONAL'
```
CarlosNZ, line 3315: Insurance stuff from earlier PRs, but we hadn't re-generated since then

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/SidePanel.tsx
```diff
@@ -55,16 +60,30 @@ export const SidePanelComponent = () => {
...
   return (
     <DetailPanelPortal
       Actions={
         <>
-          <DetailPanelAction
-            icon={<DeleteIcon />}
-            title={t('label.delete')}
-            onClick={onDelete}
-            disabled={!canDelete}
-          />
```
CarlosNZ, line 67: Since we show *either* "Delete" or "Cancel", makes sense for Delete to conditionally appear rather than be disabled sometimes.

server/repository/src/migrations/v2_06_00/add_cancelled_status_to_invoice.rs
```diff
@@ -12,9 +12,12 @@ impl MigrationFragment for Migrate {
                 connection,
                 r#"
                 ALTER TYPE invoice_status ADD VALUE IF NOT EXISTS 'CANCELLED';
+                ALTER TYPE activity_log_type
+                ADD VALUE IF NOT EXISTS
+                    'PRESCRIPTION_STATUS_CANCELLED' AFTER 'PRESCRIPTION_STATUS_VERIFIED';
```
CarlosNZ, line 17: Added to existing migration, so might need to manually reset the migration fragment.

server/service/src/dashboard/invoice_count.rs
```diff
@@ -124,7 +124,8 @@ fn invoices_count(
             invoice_filter = invoice_filter.verified_datetime(datetime_filter)
         }
         InvoiceStatus::Cancelled => {
-             invoice_filter = invoice_filter. verified_datetime(datetime_filter)
+            // TO-DO: I don't think this is correct
+            invoice_filter = invoice_filter.verified_datetime(datetime_filter)
```
CarlosNZ, line 128: Unrelated, but came across it -- surely we should be excluding "Cancelled" invoices from this count altogether?

Should we update this with #6500?

client/packages/common/src/intl/locales/en/common.json
jmbrunskill: this is a button too though right? Maybe label.mark_invoice_as_cancelled or something to give more context if it's needed for the translation?

server/service/src/dashboard/invoice_count.rs
jmbrunskill: Yes pls!

jmbrunskill: I think its called with a single status at a time so don't think it will matter, but definitely worth following up on....

jmbrunskill (approved):
Works good for me.
Would like to resolve that datetime thing, I think Zachariah's solution of adding the field is probably a good idea now, if you want that I can throw that in today?

jmbrunskill (changes requested):
Actually just noticed we have failing tests. Should fix those before merging

client/packages/common/src/intl/locales/en/common.json
CarlosNZ: Okay, I updated both the key and the text to make it more explicit:

Screenshot 2025-02-17 at 11 45 35 AM

19c86ea798

CarlosNZ:
> Actually just noticed we have failing tests. Should fix those before merging

Cool, that's all sorted now with that little bit of extra logic.

jmbrunskill (approved):
Thanks for the changes!

Issue #6512, opened 2025-02-13
Show Cancelled Datetime in a Cancelled Prescription


jmbrunskill:
```
  if (statusIdx >= 3) {
    statusLog[InvoiceNodeStatus.Cancelled] = invoice.cancelledDatetime;
  }
```
We might want something like this added, I think that will require changing the PrescriptionRowFragment as well as looking up the time (either by recording it on update, or by looking up the time for the cancellation record invoice)

_Originally posted by @jmbrunskill in https://github.com/msupply-foundation/open-msupply/pull/6509#discussion_r1955254838_

Issue #6504, opened 2025-02-12
Sync is_cancellation & translate for prescription records


jmbrunskill:
When we get a cancellation from mSupply, we should apply the Canceled status to the linked invoice (if not already) and map this to a field in OMS.

Should also map the cancelled_datetime to a new om_cancelled_datetime field in mSupply if we decide we do want that field.

marthakeezy:
# 🧪 Testing



`V2.6.1-RC4`

- [ ] Cancel Prescription in OG
- [ ] Sync
- [ ] Prescription should appear as "Cancelled" in Prescriptions list
- [ ] The "Cancelled" date won't be correct in the Invoice details (Status progress bar) as that part still needs to be done in a later PR

marthakeezy:
Hey @CarlosNZ, I don't think the cancellation still syncs to OMS

Are you able to check please?

Cancelled 2 prescriptions for Noel

<img width="868" alt="Image" src="https://github.com/user-attachments/assets/7c8efe6e-71aa-4e59-8359-2c11b0d71a0e" />

They still show as `Verified` in the list view

<img width="1276" alt="Image" src="https://github.com/user-attachments/assets/45482246-8d68-498a-82b0-8c542ba517ce" />

This is one of the Logs

<img width="1278" alt="Image" src="https://github.com/user-attachments/assets/94218613-1878-46e3-943a-f82fca45e7f6" />

And the statuses show nothing

<img width="392" alt="Image" src="https://github.com/user-attachments/assets/43d52b36-0120-4f43-8bba-f10124537d39" />

PR #6543, opened 2025-02-18
#6500 Auto-create cancellation prescription

### LINKED ISSUE #6500: On Prescription cancellation, create return to balance ledger

CarlosNZ:
Fixes #6500 & #6512

Note: ~~In Draft because haven't finished writing test case. But all functionality should be in place for review.~~ Test cases added now 👍

# 👩🏻‍💻 What does this PR do?

- Add "cancelled_datetime" and "is_cancellation" fields to "Invoice" table
- Ensure "cancelled_datetime" gets set when prescription is cancelled
- Auto-generate a "reverse" invoice when a prescription is "Cancelled" (with the same Invoice number as its source), and set its `is_cancellation` field to `true`
- Prevent "is_cancellation" invoices appearing in the Prescriptions list in client
- Prevent "is_cancellation" invoice appearing when fetching Invoice by number
- Use "cancelled_datetime" as the source for Cancellation date shown in UI (in Breadcrumb status footer)
- Adds "Status" filter to Prescriptions list view

# 🧪 Testing

Following on from test steps in #6516:

- [x] Cancel another prescription
- [x] The "Cancelled" date in the Status Breadcrumb footer shows the correct cancellation date (cancel a prescription that was verified more than a day ago to see the difference)
- [ ] After Cancellation, look up the stock line in Inventory -> View Stock, go to the Ledger tab and see the prescription and its reverse are both listed, with the latter undoing the stock movements (with negative values) of the former.
- [ ] Go to Prescriptions list, confirm that you can filter by "Status", and that it behaves correctly


server/repository/Cargo.toml
```diff
@@ -17,7 +17,7 @@ diesel = { version = "2.2.6", default-features = false, features = [
     "r2d2",
     "numeric",
     "chrono",
-    "32-column-tables",
+    "64-column-tables",
```
CarlosNZ, line 20: Hit the 32 column limit! 😔

server/service/src/invoice/prescription/update/generate.rs
```diff
@@ -151,6 +151,9 @@ fn set_new_status_datetime(invoice: &mut InvoiceRow, status: &Option<UpdatePresc
     let status_datetime = get_invoice_status_datetime(invoice);
 
     match (&invoice.status, new_status) {
+        (InvoiceStatus::Verified, UpdatePrescriptionStatus::Cancelled) => {
+            invoice.cancelled_datetime = Some(status_datetime)
+        }
```
CarlosNZ, line 156: This will *only* set the cancelled_datetime if "Cancelled" is being set *from* "Verified", but we don't actually have a check to prevent setting Cancelled from "Picked" (except in the front-end UI). This okay?

jmbrunskill (commented):
Looks good to me Carl, We'll follow up on the sync stuff tomorrow [#6504](https://github.com/msupply-foundation/open-msupply/issues/6504)

I'll let @GeronimoJohn give this a test run, I haven't actually tried it.

GeronimoJohn:
Gave it a test run. Followed the testing guides and all are passing except one. I'm assuming that one of them should be the original value - meaning positive right?

Screenshot 2025-02-19 at 10 58 11 AM

CarlosNZ:
> Great stuff @CarlosNZ. Gave it a test run. Followed the testing guides and all are passing except one. I'm assuming that one of them should be the original value - meaning positive right?
> 
Yeah, I'd think so. I thought that changing it from a "STOCK_OUT" to a "STOCK_IN" would reverse it, but maybe not. Might need to explicitly negative the quantities. Will look into it, thanks.

CarlosNZ:
> > Great stuff @CarlosNZ. Gave it a test run. Followed the testing guides and all are passing except one. I'm assuming that one of them should be the original value - meaning positive right?
> 
> Yeah, I'd think so. I thought that changing it from a "STOCK_OUT" to a "STOCK_IN" would reverse it, but maybe not. Might need to explicitly negative the quantities. Will look into it, thanks.

Duh, turns out I had Stock In and Stock Out the wrong way round in my function 🙄

Have fixed now, seems to be working properly for me: 2cad56172a

GeronimoJohn (approved):
Great stuff! LGTM 

Screenshot 2025-02-19 at 1 16 00 PM

CarlosNZ:
Note, have squeezed in adding an additional "Status" filter to the "Prescriptions" list view if you want to check it quickly @GeronimoJohn: 1c44be8b12

client/packages/common/src/ui/components/inputs/Filters/EnumFilter.tsx
```diff
@@ -35,14 +35,7 @@ export const EnumFilter: FC<{
...
       placeholder={name}
-      slotProps={{
-        input: {
-          sx: {
-            width: FILTER_WIDTH,
-          },
-        },
-      }}
-      sx={FilterLabelSx}
+      sx={{ ...FilterLabelSx, width: FILTER_WIDTH }}
```
CarlosNZ, line 38: Seems like another victim of MUI6 migration, good to catch it here.

client/packages/common/src/ui/components/inputs/Filters/FilterMenu.tsx
```diff
@@ -42,7 +42,7 @@ interface FilterDefinitions {
 
 // CONSTANTS
 const RESET_KEYWORD = 'RESET';
-export const FILTER_WIDTH = 220;
+export const FILTER_WIDTH = 180;
```
CarlosNZ, line 45: I couldn't find any filters that really needed this additional width, although we should keep a lookout as this applies everywhere 😬

client/packages/system/src/Stock/ListView/Toolbar.tsx
```diff
@@ -27,6 +27,7 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
               name: t('label.code-or-name'),
               urlParameter: 'itemCodeOrName',
               placeholder: t('placeholder.enter-an-item-code-or-name'),
+              isDefault: true,
```
CarlosNZ, line 30: Okay, this is totally off-topic, but it was annoying me and is consistent with other pages.

GeronimoJohn:
> Note, have squeezed in adding an additional "Status" filter to the "Prescriptions" list view if you want to check it quickly @GeronimoJohn: [1c44be8](https://github.com/msupply-foundation/open-msupply/commit/1c44be8b12d96cbc2992434b222823228e5316de)

Sorry forgot to mention that I saw the changes. LGTM!

PR #6554, opened 2025-02-18
6533 pagination for period

### LINKED ISSUE #6533: Pagination for period
## What went wrong? 😲
Some reports require filtering by period but this currently doesn't implement any kind of pagination. 

## Expected behaviour 🤔
Implement pagination for period and use AutocompleteWithPagination for the json component

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #6533

# 👩🏻‍💻 What does this PR do?
Defaults shown periods to 15 and only show past periods. 

# 🧪 Testing

- [ ] Have lots of periods over 30? 
- [ ] Set up CIV stocktake report 
- [ ] Go to report and click on stocktake report
- [ ] Only limited of periods shown until you keep scrolling

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
+++ b/client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
@@ -3,7 +3,6 @@ import { rankWith, ControlProps, uiTypeIs } from '@jsonforms/core';
 import { z } from 'zod';
 import {
   AutocompleteWithPagination,
-  DateUtils,
   DetailInputWithLabelRow,
   extractProperty,
   Formatter,
@@ -33,14 +32,13 @@ const UIComponent = (props: ControlProps) => {
     ? extractProperty(core?.data, 'programId')
     : null;
   const today = new Date();
-  const lastDayOfMonth = DateUtils.lastDayOfMonth(today);
   const { data, isFetching, fetchNextPage } = usePeriodList(
     RECORDS_PER_PAGE,
     programId,
     options?.findByProgram ? !!programId : true,
     {
-      endDate: {
-        beforeOrEqualTo: Formatter.naiveDate(lastDayOfMonth),
+      startDate: {
+        beforeOrEqualTo: Formatter.naiveDate(today),
       },
     }
   );
--- a/client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
+++ b/client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
@@ -2,9 +2,11 @@ import React, { useState } from 'react';
 import { rankWith, ControlProps, uiTypeIs } from '@jsonforms/core';
 import { z } from 'zod';
 import {
-  Autocomplete,
+  AutocompleteWithPagination,
+  DateUtils,
   DetailInputWithLabelRow,
   extractProperty,
+  Formatter,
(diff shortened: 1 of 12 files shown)


client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
```diff
@@ -29,11 +32,23 @@ const UIComponent = (props: ControlProps) => {
...
+  const today = new Date();
+  const lastDayOfMonth = DateUtils.lastDayOfMonth(today);
+  const { data, isFetching, fetchNextPage } = usePeriodList(
+    RECORDS_PER_PAGE,
     programId,
-    options?.findByProgram ? !!programId : true
+    options?.findByProgram ? !!programId : true,
+    {
+      endDate: {
+        beforeOrEqualTo: Formatter.naiveDate(lastDayOfMonth),
```
roxy-dao, line 43: Want to get all periods until the last day of current month in case there are transactions done in this time

client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
```diff
@@ -29,11 +32,23 @@ const UIComponent = (props: ControlProps) => {
   const programId = options?.findByProgram
     ? extractProperty(core?.data, 'programId')
     : null;
-  const { data, isLoading } = usePeriodList(
+  const today = new Date();
+  const lastDayOfMonth = DateUtils.lastDayOfMonth(today);
+  const { data, isFetching, fetchNextPage } = usePeriodList(
```
fergie-nz, line 37: nit but I like consistency of isLoading which we use everywhere else. Is there a reason to change to fetching?

roxy-dao: cause we are fetching next page so the loading was just changed to fetching to match, but guess can use isLoading

client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
fergie-nz: This is great but I think it misses some edge cases:

If a period is monthly but doesn't begin on the first day of the month, then it won't capture this period (in my case I had a monthly period schedule which began on the 10th, and wasn't visible). I'm not sure if this is a common case, but there it is.

I wonder if it would be better to filter instead by startDate and filter for startDate < now.

I recognise requires refactoring of backend filters, but makes for a more simple front end use.

fergie-nz (changes requested):
Looks awesome - some issue showing the current period when not mapped exactly to months

client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
fergie-nz: I'm easy, just a minor preference

client/packages/programs/src/JsonForms/components/PeriodSearch.tsx
roxy-dao: Yup good idea

fergie-nz (approved):
nice LGTM! 🎸

Issue #2951, opened 2024-02-14
[Ruru] Back end plugin investigation


mark-prins:

andreievg:
Setting priority as must have, because we are committed to get this in 2.2

andreievg:
I have quite a bit of carry over to capture somewhere (nothing is critical)

* Update documentation based on TODOs
* Plugin data to not have record type (STOCK_LINE), and to just have an identifier for plugin, also plugin data to have optional store id, without store id plugin data sync everywhere, this is where global configuration for plugins will be stored, configurations for plugins should be set on central server via another plugin (also central only plugins)
* Plugin data api to be in core and easily available for plugin consumption
* Versioning, like reports, see more in documentation PR
* Rename plugin provider and all types there to backend plugin provider
* Deal with decoding error of base64 (since plugins are loaded on startup, don't want to brick the server)
* Plugin validation and signing, reinstate
* Suggest going to V3 (change v2.6 release to be V3 since plugins are added ?)
* UI Interface for plugin upload ?
* We should be able to not need singleton context if shared context is exposed in common
* generate_and_install_pluigin_bundle and install plugin bundle to use the same config file for usename and password as for reports ?
* Front end plugins to reload after sync (clear cache ? but only if plugins arrived ?)
* variant_type on backend front end was suggested to be something like mechanism (better nameing)
* PluginInfoNode => PluginBundleInfoNode
* Backend plugin to have some tests 
* Plugin type to be loaded during bundling from the actual exports of plugin.tsx and plugin.ts some more detail in documentation PR

PR #6556, opened 2025-02-18
6555 interface to transform requisition line

### LINKED ISSUE #6555: Expose suggested quantity modification as backend plugin

Need to modify suggested quantity when inserting or updating requisition line, there are other requirements for plugin to populate other field, so having a way to 'transform' requisition line would be ideal

Add transform_requisition_line plugin interface

Overwriting suggested quantity calculation only. We would still need to change other fields so a more generic interface is prefered

andreievg:
Fixes #6555

# 👩🏻‍💻 What does this PR do?

* Adds a new transform requisition line interface

Ideally multiple plugins should implement this interface and still work and/or a plugin should be able to implement multiple instances of this interface, i added to list here: https://github.com/msupply-foundation/open-msupply/issues/2951

# 🧪 Testing

Using this [plugin example](https://github.com/msupply-foundation/open-msupply-plugins/pull/8), try creating internal order from master list (with data file where there is amc and suggested quantity, can initialise reference1 in this branch and refresh date, but would need to make `pub fn is_central_server() -> bool {` return `true`). Load plugin as per instructions in `plugin.ts`, re create internal order from same master list, see different suggested quantity. Make changes to plugin.ts, re load it. Then update internal order (say max mos), see suggested quantity updated as per your new change


client/packages/plugins/backendTypes/generated/ApprovalStatusType.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
```
andreievg, line 1: There are all auto generated, only change in backendTypes dir is the types.ts

client/packages/plugins/backendTypes/types.ts
```diff
@@ -1,9 +1,12 @@
 import { PluginTypes } from './generated/PluginTypes';
 
 export type BackendPlugins = {
-  average_monthly_consumption: (
+  average_monthly_consumption?: (
```
andreievg, line 4: has to be optional, plugins only implement interfaces they need

server/repository/Cargo.toml
```diff
@@ -38,6 +38,7 @@ rust-embed = { version = "8.4.0", features = ["debug-embed"] }
 anyhow.workspace = true
 pretty_assertions = { workspace = true }
 strum = { workspace = true }
+ts-rs = { workspace = true }
```
andreievg, line 41: Needed since some repo rows are now exposed in plugin

server/repository/src/db_diesel/requisition/requisition_row.rs
```diff
@@ -91,9 +94,13 @@ pub struct RequisitionRow {
     #[diesel(column_name = type_)]
     pub r#type: RequisitionType,
     pub status: RequisitionStatus,
+    #[ts(as = "String")]
     pub created_datetime: NaiveDateTime,
```
andreievg, line 98: NaiveDateTime does not implement TS (should check maybe there is a features flag that implements it

server/service/src/backend_plugin/types/transform_requisition_lines.rs
```diff
@@ -0,0 +1,27 @@
...
+#[ts(rename = "TransformRequisitionLineInput")]
+pub struct Input {
+    pub requisition: RequisitionRow,
+    pub lines: Vec<RequisitionLineRow>,
+}
+
+#[derive(TS, Clone, Deserialize, Serialize)]
+#[ts(rename = "TransformRequisitionLineOutput")]
+pub struct Output {
+    pub transformed_lines: Vec<RequisitionLineRow>,
```
andreievg, line 17: Using transformed_lines rather then just Vec<RequisitionLineRow> because we would also need to expose pluginData, so that it's returned by the plugin, we would need to update suggested quantity based on the area AMC in some cases

andreievg:
i messed up the merge, closing, pr here: https://github.com/msupply-foundation/open-msupply/pull/6642

PR #6565, opened 2025-02-19
#6542 dispensary label printing 

### LINKED ISSUE #6542: Label Printing backend

For label printing MVP we should have a standard label format that follows similarly to the OG medicine label defaults. We need a back end request that 

- [ ] Get a good standard label template for MVP (just `format!` for now)
- [ ] Add endpoint similar to QR code label printing endpoint
- [ ] Use current printer setup - write issue for hooking up to new printer table instead (API should take ID of printer to use as parameter

- We could literally use reports as the templating for printing labels
- We using tera could be helpful, though maybe it's overboard as we don't likely need the

### Additional context 💌

- It's been brought to my attention at the 11th hour that multiple labels per line would be a great feature that OG does not have; if a patient is prescribed 2 boxes of 100, they need 2 labels printed with "part 1/2" and "part 2/2".
- Likewise, number of repeats on the label would be fantastic.

### Moneyworks Jobcode 🧰

OMS:DDSP

Chris-Petty:
Fixes #6542 

# 👩🏻‍💻 What does this PR do?

Adds API for _very_ basic label printing. The API shape itself may change a lot in the future, though!

I lost an embarrassingly long time trying to figure out why on earth my bearer tokens weren't going through (as that's how we auth the gql API):

```
Denied(
    NotAuthenticated(
        "Missing auth token",
    ),
)
```

Turns out the printing API and some file sync use cookies instead! Can't think of why we need to auth schemes. Perhaps the auth cookie had some extra data in it beyond `token`, though JWTs themselves hold extra data so it's a bit of a nesting doll situation there.

Anyway if you want to test this you need an API client that can include a cookie, like `curl` or insomnia.

Your cookie must have this JSON content shape:
```json
{
  "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE3Mzk5MzkxNzAsImF1ZCI6IkFwaSIsImlhdCI6MTczOTkzNTU3MCwiaXNzIjoib20tc3VwcGx5LXJlbW90ZS1zZXJ2ZXIiLCJzdWIiOiIwNzYzRTJFMzA1M0Q0QzQ3OEUxRTZCNkIwM0ZFQzIwNyJ9.COkzCGBJFMFNpFMDO348XRJOwad3UeTshY-j-06kt-g"
}
```

Where the token value is a JWT from the `authToken` endpoint (note for gql we just send the header `"Authorization: Bearer <JWT>"`)

# 🧪 Testing

- [ ] Setup printer in OMS Settings > Devices. Office printer is on `192.168.1.224`, port `9100` is the standard for zebras web server.
- [ ] use an http client to send a label payload and print a label

Something like

```json
{
  "patientDetails": "Chris Petty - 92e941",
  "itemDetails": "Amoxicillin Dry Powder for Suspension 125mg/5ml Bot/100ml",
  "itemDirections": "Dilute 50:1 with protein shake. Include oats for fibre and good gut health. Chug at least twice a day.",
  "warning": "Avoid excessive sun exposure",
  "details": "Amazing Store - Mr Pharmacist - 2024/02/19",
}
```

# 📃 Documentation

  1.
  2.

### DIFF
--- a/server/server/src/print/label.rs
+++ b/server/server/src/print/label.rs
@@ -5,7 +5,7 @@ use actix_web::{
 use repository::RepositoryError;
 use service::{
     auth_data::AuthData,
-    print::label::{host_status, print_qr_code},
+    print::label::{host_status, print_prescription_label, print_qr_code, PrescriptionLabelData},
     service_provider::ServiceProvider,
     settings::LabelPrinterSettingNode,
 };
@@ -13,7 +13,7 @@ use service::{
 use crate::authentication::validate_cookie_auth;
 
 #[derive(serde::Deserialize)]
-pub struct LabelData {
+pub struct QrLabelData {
     code: String,
     message: Option<String>,
 }
@@ -22,7 +22,7 @@ pub async fn print_label_qr(
     request: HttpRequest,
     service_provider: Data<ServiceProvider>,
     auth_data: Data<AuthData>,
-    data: web::Json<LabelData>,
+    data: web::Json<QrLabelData>,
 ) -> HttpResponse {
     let auth_result = validate_cookie_auth(request.clone(), &auth_data);
     match auth_result {
@@ -43,6 +43,35 @@ pub async fn print_label_qr(
 
     match print_qr_code(settings, data.code.clone(), data.message.clone()) {
         Ok(_) => HttpResponse::Ok().body("QR label printed"),
+        Err(err) => HttpResponse::InternalServerError().body(err.to_string()),
+    }
+}
+
+pub async fn print_label_prescription(
+    request: HttpRequest,
+    service_provider: Data<ServiceProvider>,
+    auth_data: Data<AuthData>,
+    data: web::Json<PrescriptionLabelData>,
+) -> HttpResponse {
+    let auth_result = validate_cookie_auth(request.clone(), &auth_data);
+    match auth_result {
+        Ok(_) => (),
+        Err(error) => {
+            let formatted_error = format!("{:#?}", error);
+            return HttpResponse::Unauthorized().body(formatted_error);
+        }
+    }
+
+    let settings = match get_printer_settings(service_provider) {
+        Ok(settings) => settings,
+        Err(error) => {
+            return HttpResponse::InternalServerError()
+                .body(format!("Error getting printer settings: {}", error));
+        }
+    };
+
+    match print_prescription_label(settings, data.into_inner()) {
+        Ok(_) => HttpResponse::Ok().body("Label printed"),
         Err(err) => HttpResponse::InternalServerError().body(err.to_string()),
     }
 }
--- a/server/service/src/print/label.rs
+++ b/server/service/src/print/label.rs
@@ -77,27 +77,27 @@ pub fn print_prescription_label(
 ^CI28
 
 ^A0,25
-^FO20,10
-^FB550,1,0,C
-^FD{item_details}\&^FS
+^FO10,10
+^FB570,2,0,C
+^FD{item_details}^FS
 
-^FO20,50
-^GB550,2,2^FS
+^FO10,65
+^GB570,2,2^FS
 
 ^A0,25
-^FO20,60
-^FB550,6,0,L
+^FO10,75
+^FB570,6,0,L
 ^FD{item_directions}\&{warning}^FS
 
-^FO20,210
-^GB550,2,2^FS
+^FO10,210
+^GB570,2,2^FS
 
 ^A0,25
-^FO20,220
+^FO10,220
 ^FD{patient_details}^FS
 
 ^A0,25
-^FO20,250
+^FO10,250
 ^FD{details}^FS
 ^XZ
 "#
--- a/server/service/src/print/label.rs
+++ b/server/service/src/print/label.rs
@@ -47,9 +47,9 @@ pub fn print_qr_code(
 pub struct PrescriptionLabelData {
     item_name: String,
     item_directions: String,
-    patient_details: String,
-    warning: Option<String>,
-    details: String,
+    patient_details: String, // e.g patient name, possibly code etc.
+    warning: Option<String>, // Some items come with a defined warning (OG field "Message") that should be printed on all labels regardless of directions e.g. avoid sun exposure, avoid alcohol...
+    details: String, // General details to include e.g. store name, prescriber name, date/time...
 }
 
 pub fn print_prescription_label(
@@ -60,8 +60,8 @@ pub fn print_prescription_label(
         item_name,
         item_directions,
         patient_details,
-        details,
         warning,
+        details,
     } = label_data;
     let warning = warning.unwrap_or_default();
 
@@ -96,7 +96,7 @@ pub fn print_prescription_label(
         ^FD{details}^FS
         ^XZ"#
     );
-    let printer = Jetdirect::new(settings.address, settings.port); // This (and QR labels!) should be moved to using the new printer config table in some manner. Either FE sends the details or backend does query to get them (FE should probably already have them though)
+    let printer = Jetdirect::new(settings.address, settings.port);
     printer.send_string(payload, Mode::Print)
 }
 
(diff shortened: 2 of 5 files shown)


server/server/src/print/label.rs
```diff
@@ -47,6 +47,35 @@ pub async fn print_label_qr(
...
+        }
+    }
+
+    let settings = match get_printer_settings(service_provider) {
+        Ok(settings) => settings,
+        Err(error) => {
+            return HttpResponse::InternalServerError()
+                .body(format!("Error getting printer settings: {}", error));
+        }
+    };
```
Chris-Petty, line 71: basically duplicate of print_label_qr logic, but DRY it out after 3rd instance?

server/service/src/print/label.rs
```diff
@@ -32,6 +32,9 @@ pub fn print_qr_code(
     let payload = format!(
         r#"
         ^XA
+        ^FX CI command parameters:
+        ^FX - encoding (28 = UTF-8)
+        ^CI28
```
Chris-Petty, line 37: Should allow printing supporting more languages I think. Recommended [here](https://labelary.com/zpl.html).

server/service/src/print/label.rs
```diff
@@ -43,6 +46,66 @@ pub fn print_qr_code(
...
+^A0,25
+^FO20,220
+^FD{patient_details}^FS
+
+^A0,25
+^FO20,250
+^FD{details}^FS
+^XZ
+"#
+    );
```
Chris-Petty, line 104: This produces a label that looks something like this:



- If the item name is too long it will wrap onto itself. Will need to get more clever I think...
- The directions and warning (if present) will span 6 lines at most, before things start stacking up on the 6th line

format! is very limited and just a temp solution - it only works with string literals, so we can't change it at runtime at all. We could use tera templates, perhaps literally reports, to do the templating instead. Some benefits:

- Well reports would sync from central, so doing that would piggy back a lot
- Perhaps could be it's own table, but behave much the same way as reports...
- Tera Jinja2 style syntax allows potentially some better formatting. I was optimistic we'd get 99% of the way with just ZPL, but I suspect it's more like 80%. 80/20 rule coming in hot...

server/service/src/print/label.rs
```diff
@@ -43,6 +46,66 @@ pub fn print_qr_code(
...
+    let printer = Jetdirect::new(settings.address, settings.port);
+    printer.send_string(payload, Mode::Print)
```
Chris-Petty, line 106: Just running the same way as the qr codes, however you can literally just make a regular http POST with the ZPL text to print labels. At least for zebra printers anyway.

server/service/src/print/label.rs
```diff
@@ -43,6 +46,66 @@ pub fn print_qr_code(
...
 
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PrescriptionLabelData {
+    item_details: String, // usually the amount of units + the item name e.g. "10Tabs Paracetamol 500mg Tablets"
+    item_directions: String,
+    patient_details: String, // e.g patient name, possibly code etc.
+    warning: Option<String>, // Some items come with a defined warning (OG field "Message") that should be printed on all labels regardless of directions e.g. avoid sun exposure, avoid alcohol...
+    details: String, // General details to include e.g. store name, prescriber name, date/time...
+}
```
Chris-Petty, line 57: This is likely to shift. Requirements I got in the process of working on this:

- Must show the number of units in the item_details.
- Should show number of repeats.
- Could/Should support multiple labels for the same line! OG doesn't do this. Should be "part1", "partN"... sometimes you prescribe multiple physical boxes (e.g. 2x 200tabs of paracetamol for a 400tab prescription)
- Could include a "n/total" at the bottom. For 3 labels you'd have 1/3, 2/3, 3/3.
- Observation: it is currently one label per request. This seems to cause the printer to stop for >100ms per label, but I know it can just do a non stop stream, OG does 😉.

All of this makes me think that all of the label content should just be on the backend. The API could be

```rust
PrescriptionLabelsData {
  printer_id: String // the selected printer 
  label_ids: Vec<String> // duplicate IDs indicating to print multiple labels
}
```

The endpoint would handle querying for everything and building the ZPL payload on the backend. That's possibly redundant if the FE already had that data and could send it. Just occurred to me it could just be a `Vec<PrescriptionLabelData>` though. Many ways to shave the yak, though that'd save a few cheap db queries 🤷. Anyway,

Chris-Petty:
Another fun tidbit...



When I was printing today this started happening. Telling the printer to recalibrate itself didn't help. What did magically help... was printing a label from OG! Miraculously fixed it the printer's own config print out and my OMS label printing too. I suspect OG must start printing commands off customising the width/height etc. to avoid this happening.

server/server/src/print/label.rs
aimee-mcneil-melville: Yeah, for now looks straightforward having it separate! It also gives room for any significant differences between them as we expand on the prescription label printing

server/service/src/print/label.rs
aimee-mcneil-melville: Would it be a quick interim change to allow the item name space for two lines, and reduce the directions/warning to 5 lines total? This would allow most items to fit within the two lines without being cut off

Should we also consider raising an issue for adding front end checks, so the user is notified/limited on the directions length?

server/service/src/print/label.rs
aimee-mcneil-melville: Just thoughts - will itemDetails be a string of `units + itemName`?  - 2 each, Amoxicillin Dry Powder for Suspension 125mg/5ml Bot/100ml
With the extra length, it would be ideal if it had another line of space!

Great if it can all be handled in the backend, however I know there is something around having a preview of the label in the frontend - if that is implemented would that be then complicating the queries?

aimee-mcneil-melville: nit - testing JSON example has `itemName` where in code it is `itemDetails` - itemDetails makes more sense if it has the units in the same line and I see that was a change at the end!
Can you update that in the JSON example please 😄

aimee-mcneil-melville:
Works great! Just the change to allow a longer item name please 😄 

If OG has a pre print step to recalibrate the labels that would be fantastic to bring in here! A separate issue though!

I'll also add some specifics for the testing setup:
POST: http://localhost:8000/print/label-prescription
Cookie: `Key: auth`, `Value: {token ..etc}`
Body: JSON

aimee-mcneil-melville (changes requested):
Works great! Just the change to allow a longer item name please 😄 

If OG has a pre print step to recalibrate the labels that would be fantastic to bring in here! A separate issue though!

I'll also add some specifics for the testing setup:
POST: http://localhost:8000/print/label-prescription
Cookie: `Key: auth`, `Value: {token ..etc}`
Body: JSON

server/service/src/print/label.rs
Chris-Petty: Whoops stale JSON haha. Will add more space, will look a bit more ugly if the string is a 1 liner but good enough for now.

> Great if it can all be handled in the backend, however I know there is something around having a preview of the label in the frontend - if that is implemented would that be then complicating the queries?

I'd say the preview thing is very much a stretch goal - we'd need a library for rendering ZPL to image or svg and I'm yet to find a good candidate, writing one would just be way out of scope. We can just use [labelary](https://labelary.com/viewer.html) for quick previews when making new layouts. But yea, could complicate things for sure. I'm flip flopping so I might chat to Drei about it - we could do all the templating on the frontend if we wanted and just send the ZPL text to the backend.

Chris-Petty: For what it's worth, labelary does the rendering on the backend and sends the image in a response

image

I suspect that's to protect their proprietary renderer ;)

Chris-Petty:
I made things a smidge wider too just to get more use outta the label

server/service/src/print/label.rs
aimee-mcneil-melville: Cool, all things to work out for later but this does the trick for now!

aimee-mcneil-melville (approved):
Nice, looking great!

Chris-Petty:
rebased! Nice catch @aimee-mcneil-melville

PR #6569, opened 2025-02-19
#6500 (2) Replacement PR

CarlosNZ:
This is a replacement PR for #6543, since the original one got reverted due to the merge mess-up (fixed in #6566).

It should be identical to the original.


GeronimoJohn (approved):
Thanks Carl! Just a failing test but otherwise LGTM!

CarlosNZ:
> Thanks Carl! Just a failing test but otherwise LGTM!

Ha, was a single missing comma in the Russian translation file! Thanks for spotting -- all fixed now: 0732f8434a

PR #6586, opened 2025-02-20
6517 App crashes when I click on disabled Reason dropdown Supplier return edit modal

### LINKED ISSUE #6517: App crashes when I click on disabled ```Reason``` dropdown Supplier return edit modal
## What went wrong? 😲

Found when investigating #6066. App crashes with message below:

Image

Image

https://github.com/user-attachments/assets/c1fae51c-619b-4186-9072-0712ad1ba0ea
 

## Expected behaviour 🤔

App shouldn't crash when clicking on disabled reason dropdown.

## How to Reproduce 🔨

1. Go to Inbound shipment
2. Ensure inbound shipment status is delivered
3. Select items with quantity and return items
4. Enter amount to return and click next
5. Click reason dropdown
6. Watch app crash

- Open mSupply Version: 2.6.0
- Legacy mSupply Central Server Version: 8.00.04

- Platform: MacOS  Chrome Browser

- Database type:

roxy-dao:
Fixes #6517

# 👩🏻‍💻 What does this PR do?
Removes the custom renderer... it's the same as the AutoComplete input but with extra padding??

# 🧪 Testing

- [ ] Set up return reasons
- [ ] Create a supplier return
- [ ] Should be able to choose return reason

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
+++ b/client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
@@ -1,7 +1,6 @@
 import React, { FC } from 'react';
 import {
   Autocomplete,
-  BasicTextInput,
   Box,
   defaultOptionMapper,
   getDefaultOptionRenderer,
@@ -20,7 +19,6 @@ export const ReturnReasonSearchInput: FC<ReturnReasonSearchInputProps> = ({
   selectedReasonId,
   onChange,
   autoFocus = false,
-  isError,
   isDisabled,
 }) => {
   const { data, isLoading } = useReturnReason.document.listAllActive();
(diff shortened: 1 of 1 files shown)


client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
```diff
@@ -47,20 +45,6 @@ export const ReturnReasonSearchInput: FC<ReturnReasonSearchInputProps> = ({
...
           onChange(reason?.id ?? '');
         }}
-        renderInput={props => (
-          <BasicTextInput
-            {...props}
-            autoFocus={autoFocus}
-            slotProps={{
-              input: {
-                disableUnderline: false,
-                style: props.disabled ? { paddingLeft: 0 } : {},
```
roxy-dao, line 57: The extra padding.. though looks fine to me even if it isn't in?

client/packages/system/src/ReturnReason/Components/ReturnReasonSearchInput.tsx
```diff
@@ -20,7 +19,6 @@ export const ReturnReasonSearchInput: FC<ReturnReasonSearchInputProps> = ({
   selectedReasonId,
   onChange,
   autoFocus = false,
-  isError,
```
roxy-dao, line 23: isError isn't being used so have removed

lache-melvin (approved):
LGTM thank you! Looks like a missing mapped props, but good to use the defaults where we can! 😢

PR #6608, opened 2025-02-21
Translate labels

### LINKED ISSUE #6496: No translation for indicator field labels
## What went wrong? 😲

Image

Regimen data labels for `comment` and `value` are not being localised, showing the resource identifier instead.

## Expected behaviour 🤔

Localised for English and ideally other languages too!

## How to Reproduce 🔨

1. View indicators

## Your environment 🌱

- Open mSupply Version: 2.5.1 android

mark-prins:
Fixes #6496

# 👩🏻‍💻 What does this PR do?
The labels were being provided with a string that is supposed to be a translation key, but isn't.

I've changed the utility method to return a LocaleKey rather than a string, as that can be used as a string when required. It does remove the fallback, but we only have comment and value so I think it's ok.

image

# 🧪 Testing

- [ ] Create a program based internal order and check that the labels are correct in the indicators tab

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
@@ -17,7 +17,7 @@ import {
 } from '../../api';
 import { useDraftIndicatorValue } from './hooks';
 import { CustomerIndicatorInfoView } from './CustomerIndicatorInfo';
-import { indicatorColumnNameToLocaleKey } from '../../../utils';
+import { indicatorColumnNameToLocal } from '../../../utils';
 
 interface IndicatorLineEditProps {
   requisitionNumber: number;
@@ -95,7 +95,7 @@ const InputWithLabel = ({
     <InputWithLabelRow
       Input={inputComponent}
       labelWidth={LABEL_WIDTH}
-      label={t(indicatorColumnNameToLocaleKey(data.name))}
+      label={indicatorColumnNameToLocal(data.name, t)}
       sx={{ marginBottom: 1 }}
     />
   );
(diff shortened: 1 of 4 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
```diff
@@ -95,7 +95,7 @@ const InputWithLabel = ({
     <InputWithLabelRow
       Input={inputComponent}
       labelWidth={LABEL_WIDTH}
-      label={indicatorColumnNameToLocal(data.name)}
+      label={t(indicatorColumnNameToLocaleKey(data.name))}
```
andreievg, line 98: This looks like the one change that would fix the issue

andreievg: Actually `It does remove the fallback, but we only have comment and value so I think it's ok.`, we should pass `t` to `indicatorColumnNameToLocaleKey` and then use it there, so we can keep fallback

client/packages/requisitions/src/utils.ts
```diff
@@ -155,13 +155,13 @@ enum IndicatorColumnName {
...
+  columnName: string
+): LocaleKey => {
   switch (columnName) {
     case IndicatorColumnName.Comment:
       return 'label.comment';
-    case IndicatorColumnName.Value:
-      return 'label.value';
     default:
-      return columnName;
+      return 'label.value';
```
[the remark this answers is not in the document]
mark-prins: The HIV use their own labels rather than 'label'/'value':
image

client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
mark-prins: I've changed to pass around the TypedTFunction

andreievg (approved):
Thanks Mark

Issue #6445, opened 2025-02-08
Prescription ux


mark-prins:
## What went wrong? 😲

The prescription button order is non-standard and I've had problems using the detail page because of it.
On all our screens, the 'Add Item' button is the leftmost. it means that you don't have to read the buttons to add an item, you just click the one on the end.

However:

<img width="572" alt="Image" src="https://github.com/user-attachments/assets/24e66c40-6ed1-4682-8c2e-12ee4d602a3c" />

I'm not totally happy with the solution made in [this PR](https://github.com/msupply-foundation/open-msupply/pull/6420), which is to remove pluralisation of unit names. The original issue raised was that blank units get an 's' appended and the simple solution is to not show the 's' when the unit is blank.

## Expected behaviour 🤔
Add item should be on the left of all the buttons
The order is (left to right):
1. Add item
2. History
3. Print labels
4. More

I suggest adding the [pluralize](https://www.npmjs.com/package/pluralize) package and using this to generate the plural form, reinstate the translation key, and remove the comment added in the PR. 

## How to Reproduce 🔨

Steps to reproduce the behaviour:

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

lache-melvin:
Idk if should be same issue, but could we do something UI wise to distinguish disabled fields? If it's disabled with text, the text is greyer:

![Image](https://github.com/user-attachments/assets/39a641ff-09dd-4645-909c-d01cc0ad1e98)

but these new empty inputs look identical when disabled or enabled:

![Image](https://github.com/user-attachments/assets/9f4cdbea-a1c2-43b5-9a31-ffe2cc8bdc58)

zachariah-at-msupply:
I don't know if this will be fixed by implementing pluralise (to give it the correct spelling 😜) as you mentioned above, but for Russian and Arabic the line edit view of a prescription has wierd pluralisation:

## English

![Image](https://github.com/user-attachments/assets/20571a17-7337-44c4-a47c-a147b7713be4)

## Arabic

![Image](https://github.com/user-attachments/assets/3bdc1e8b-11b4-4424-866c-e43bd6ed8552)

## Russian

![Image](https://github.com/user-attachments/assets/cf5d2944-d830-4a33-962f-4780cb05c51f)

marthakeezy:
Hi @mark-prins 

Just checking - in the original issue, it states the order of the buttons should be (left to right)

[...]

PR #6609, opened 2025-02-21
6581 printing labels from print buttons

### LINKED ISSUE #6581: Printing labels from print buttons

Hooking up the glorious frontend to the basic printing API that now exists.

- [ ] Other dispensary work hijacked our print button to print receipt. We should add a new "Print Labels" button
- [ ] The "Print Labels" button should make the requests to /print/label-prescription endpoint to print labels for all prescription lines
- [ ] The action bar print labels button should make requests to print the selected lines' labels

Carry over for 2.7
- Selecting printer
- Setting default printer (needs prefs implemented)
- Changing the API to accept an array of print inputs so it's only one request

### Moneyworks Jobcode 🧰

OMS:DDSP

aimee-mcneil-melville:
Fixes #6581 

# 👩🏻‍💻 What does this PR do?

Prints dispensing labels for items on a prescription from Detail View

There are two buttons to do this:
- Print Labels (App Bar) prints labels for all items on the prescription
- Print Labels (Footer) appears when one or more lines are selected, and prints labels for the selected lines only

With both options, an array of labels is sent to the printer so it prints one request in a stream if there are multiples, rather than one print request for each label. This required some changes in the backend

Printed labels
labels

Error handling - I have kept consistent with errors that are used currently in each of the button locations
App Bar - no printer found
Screenshot 2025-02-21 at 4 45 41 PM

Footer - no printer found
Screenshot 2025-02-21 at 6 16 58 PM

For both - a printer exists but is not configured correctly
Screenshot 2025-02-21 at 6 02 54 PM

Overall page layout:
Screenshot 2025-02-21 at 6 15 15 PM

Sidequests: 
- Re-ordered the buttons in App Bar (#6445)
- Updated the import on useAbbreviations as it was incorrect. Found it as this import failed tests when used in the usePrinter hook

- There is no handling for the optional fields for the clinician name - suggestions? If it is not set, will print undefined. - Prescriptions can be saved without a clinician name
The alert modal could be set to navigate to the settings page on OK?
- The store name comes from the currently logged in store

# 🧪 Testing

If you have no printer configured - test this first! Once a printer is added, these error messages can't be replicated without deleting the printer from the database

Set up steps: 
- [ ] go to Dispensary -> Prescriptions -> Create a new prescription or edit an existing one
- [ ] Select a clinician (add one in OG if none configured)
- [ ] Add multiple items, include a quantity and add some directions via the abbreviations field or in the directions text area
- [ ] Save the prescription -> Close -> be in Detail View

No printer configured:
- [ ] Click 'Print Labels' in the App Bar - with no printer configured a popover will appear 
- [ ] Select a line -> Click 'Print Labels' in the Footer - with no printer configured a modal will appear -> Click OK to close

Printer incorrectly configured:
- [ ] Click 'Print Labels' in the App Bar - error snack appears with the error message
- [ ] Select a line -> Click 'Print Labels' in the Footer - error snack appears with the error message

Printer correctly configured:
Office printer is: IP - 192.168.1.224, port 9100
- [ ] Click 'Print Labels' in the App Bar - labels for all items print with the expected information
- [ ] Select a line -> Click 'Print Labels' in the Footer - labels for the selected items print with the expected information

# 📃 Documentation

  1.
  2.


client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx
aimee-mcneil-melville, line 1: Renaming the asset label messages to be generic 'label'. Also resulted in sorting of all the common.json files

client/packages/invoices/src/Prescriptions/DetailView/AppBarButton.tsx
```diff
@@ -29,19 +30,55 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
...
+  const handlePrintLabels = (e: React.MouseEvent<HTMLButtonElement>) => {
+    if (prescription) {
+      printPrescriptionLabels(prescription, prescription.lines.nodes, e);
+    }
+  };
+
   return (
     <AppBarButtonsPortal>
       <Grid container gap={1}>
+        <ButtonWithIcon
```
aimee-mcneil-melville, line 63: Moved the Add button to be the first button

Chris-Petty (changes requested):
Awesome! Code looks good just a couple cases we should handle to stop errors:

image

image

Ah no directions. Hmm perhaps just send "" from the front end if null, or maybe make optional on the backend?

Similarly for clinician? @mark-prins perhaps can give 2c on correct behaviour. Make alerts to user to tell them to set clinician first, set directions first?

Chris-Petty:
image

Another bug! So you can have multiple batches say you are issuing 100, and have 10 left of first expiring batch (FEFO procedure = First Expiring First Out) and 500 of another batch. From this you'd have 2 batches of same item periodically. They likely should just have the same directions printed twice. So that `null` should just have the same directions as the first.

Gah honestly perhaps sometimes you'd have say 5 boxes same batch and all may need a label. That's for another issue though :) (I'm told OG does not handle this well either)

Chris-Petty:
Next morning Chris be thinking like... what if all the JSON keys were Optional on the API side? Way harder to have a deserialise error then! 

Also, for another issue after this but



run out of space for the bottom details pretty quick. Should have been "undefined, undefined". Might make text smaller for that bottom section and see if we can get a 3rd line.

client/packages/invoices/src/Prescriptions/DetailView/Footer/Footer.tsx
```diff
@@ -88,6 +90,19 @@ export const FooterComponent: FC = () => {
...
+  const {
+    printLabels: printPrescriptionLabels,
+    isPrintingLabels,
+    printerExists,
+    setPrinterExists,
+  } = usePrintLabels();
+
+  const handlePrintLabels = () => {
+    if (data) {
+      printPrescriptionLabels(data, selectedRows);
```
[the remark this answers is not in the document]
aimee-mcneil-melville: I agree, done!

client/packages/invoices/src/Prescriptions/DetailView/hooks/usePrinter.tsx
```diff
@@ -0,0 +1,75 @@
...
+          const text = await response.text();
+          throw new Error(text);
+        }
+        success(t('messages.success-printing-label'))();
+      })
+      .catch(e => {
+        error(`${t('error.printing-label')}: ${e.message}`)();
+      })
+      .finally(() => setIsPrintingLabels(false));
+    setIsPrintingLabels(false);
```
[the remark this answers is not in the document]
aimee-mcneil-melville: It works the same without - have removed!

client/packages/common/src/types/schema.ts
```diff
@@ -4016,6 +4018,7 @@ export type LedgerConnector = {
 };
 
 export type LedgerFilterInput = {
+  datetime?: InputMaybe<DatetimeFilterInput>;
```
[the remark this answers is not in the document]
aimee-mcneil-melville: I've updated the front end to send "" if no directions for now. Though keen on some input if they should be optional or not (are directions a main function of these labels??), and the preferred method of handling if empty but required?

aimee-mcneil-melville:
> Another bug! So you can have multiple batches say you are issuing 100, and have 10 left of first expiring batch (FEFO procedure = First Expiring First Out) and 500 of another batch. From this you'd have 2 batches of same item periodically. They likely should just have the same directions printed twice. So that null should just have the same directions as the first.

Should directions be assigned per item, or per batch? First thought is by item however:
If there are tablets prescribed across two batches would the labels then be: box 1/2 100 tablets... and box 2/2 - 50 tablets... and therefore a slightly different label for each?

aimee-mcneil-melville:
Directions print on all batches of the same item for me..

Screenshot 2025-02-25 at 10 45 01 AM

Chris-Petty:
> > Another bug! So you can have multiple batches say you are issuing 100, and have 10 left of first expiring batch (FEFO procedure = First Expiring First Out) and 500 of another batch. From this you'd have 2 batches of same item periodically. They likely should just have the same directions printed twice. So that null should just have the same directions as the first.
> 
> Should directions be assigned per item, or per batch? First thought is by item however: If there are tablets prescribed across two batches would the labels then be: box 1/2 100 tablets... and box 2/2 - 50 tablets... and therefore a slightly different label for each?

Yes but we can do that another time

client/packages/invoices/src/Prescriptions/DetailView/hooks/usePrinter.tsx
```diff
@@ -61,8 +61,7 @@ export const usePrintLabels = () => {
       })
       .catch(e => {
         error(`${t('error.printing-label')}: ${e.message}`)();
-      })
-      .finally(() => setIsPrintingLabels(false));
+      });
     setIsPrintingLabels(false);
```
Chris-Petty, line 65: Heh my bad if not clear - I think we should keep the finally!

The problem is that setIsPrintingLabels(false) will run immediately here, and the promise above will run async in the background. So the state essentially is never `true`. `Finally()` will run after the last `then` or `catch`, so will correctly call `setIsPrintingLabels(false)` after the request/promise has completed

Chris-Petty:
> Directions print on all batches of the same item for me..
> 

@aimee-mcneil-melville /

1. add an item with 1 batch selected with directions and save
2. bonus bug: Add another item and 1 batch + directions
3. edit the first item and add 2 more batches and save
4. The recently added batches will have no directions D:
5. The recently added batches will be printed after the second item D:

client/packages/invoices/src/Prescriptions/DetailView/hooks/usePrinter.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const printPrescriptionLabel = (
+    prescription: PrescriptionRowFragment,
+    lines: PrescriptionLineFragment[]
+  ) => {
+    setIsPrintingLabels(true);
+    const labels = lines.map(line => ({
+      itemDetails: `${line.numberOfPacks * line.packSize} ${line.item.unitName}: ${line.itemName}`,
+      itemDirections: line.note,
+      patientDetails: `${prescription.patient?.name} - ${prescription.patient?.code}`,
+      details: `${store?.name} - ${new Date(prescription.createdDatetime).toLocaleDateString()} - ${prescription.clinician?.lastName}, ${prescription.clinician?.firstName}`,
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Response from Mark - current logged in store is what mSupply does. That is likely the same as dispensing store, as you can only see your own prescriptions

client/packages/invoices/src/Prescriptions/DetailView/hooks/usePrinter.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }));
+    fetch(Environment.PRINT_LABEL_PRESCRIPTION, {
```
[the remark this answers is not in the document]
lache-melvin: Yeah, we can use `useMutation` here, given this is a POST request. 

Example here: https://github.com/msupply-foundation/open-msupply/compare/6581-printing-labels-from-print-buttons...6581-printing-labels-from-print-buttons(with-use-mutation)

I think more would be gained if we set up something more robust for our non-graphql queries, to handle the credentials, application type, mapping 400/500s to errors etc. But nice to get the loading state for free at least?

FYI queries on click are a bit more hotly debated (there is some hackery with enabling/refetching that we have done a few times, but ideally, the click would render something new, which calls `useQuery` the usual way.)

aimee-mcneil-melville:
New version is in! 
The label generation has been separated into a  new utils file, with a test file. 

- [ ] prints ONE label per item - even if prescription is across multiple batches, it will generate and print one label only
- [ ] will find the first batch of the item with a direction note (currently the direction note is on the first batch, however had bugs where batches were changed after the first save)
- [ ] prints by default in the same order as the prescription lines

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.test.tsx
```diff
@@ -0,0 +1,438 @@
...
+  availableNumberOfPacks?: number;
+  numberOfPacks?: number;
+  prescribedQuantity?: number;
+  onHold?: boolean;
+  expiryDate?: string;
+  note?: string;
+  batch?: string;
+};
+
+const createTestLine = ({
```
Chris-Petty, line 26: good lord mock data is a pain haha

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+  itemDetails: string;
+  itemDirections: string;
+  patientDetails: string;
+  details: string;
+}
+
+export const groupItems = (
+  lines: PrescriptionLineFragment[]
+): ItemDetails[] => {
+  const linesByItem: Record<string, PrescriptionLineFragment[]> = {};
```
Chris-Petty, line 20: NOT A REQUEST: I don't want to bikeshed typescript. That said IDK if this is the semantically correct use of Record even if valid typing for the task haha I guess it's simpler than the `{ [key: string]: PrescriptionLineFragment[] }` we had yesterday though

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  // groups all batches of an item by id
+  lines.forEach(line => {
+    const { id } = line.item;
+    linesByItem[id] = [...(linesByItem[id] || []), line];
```
Chris-Petty, line 25: NOT A REQUEST CAUSE IT'LL NEVER MATTER HERE REALLY: I think Laché won't fight me but I'd argue for `array.push()` approach over this even though the performance in this context is likely negligible 😁. Case I refer to often: https://jpcamara.com/2023/03/07/making-tanstack-table.html

That said, in reactjs you very often want to create a new array (and reference) with `const new = [...old, newElement]` otherwise you get punked by components not rerendering due to the reference not changing. Common footgun in react! Hence the immutable craze.

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+      id: firstItem?.id ?? '',
+      unitName: firstItem?.item.unitName ?? '',
+      name: firstItem?.itemName ?? '',
+      sum: totalUnits,
+      itemDirections: itemWithNote?.note ?? '',
+    };
+  });
+  return items;
+};
+
```
Chris-Petty, line 50: beaut

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const finishedLabel = {
+      itemDetails: itemDetails,
+      itemDirections: result.itemDirections,
+      patientDetails: patientDetails,
+      details: `${store} - ${new Date(prescription.createdDatetime).toLocaleDateString()} ${clinicianDetails}`,
+    };
+    return finishedLabel;
+  });
+};
```
Chris-Petty, line 70: beaut^2 and @lache-melvin was right to suggest moving these out reads great

Chris-Petty (approved):
siiick

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
Chris-Petty: I think an pedant might demand using a map for this but idgaf

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.test.tsx
```diff
@@ -0,0 +1,438 @@
...
+
+    const store = 'Test Store';
+    const prescription = createTestPrescription();
+    const generate = generateLabel(item, prescription, store);
+
+    const labelOne = {
+      itemDetails: '500 tablet Ibuprofen',
+      itemDirections: 'first item note',
+      patientDetails: 'Patient A - code',
+      details: 'Test Store - 27/02/2025 - lastName, firstName',
```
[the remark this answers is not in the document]
Chris-Petty: I guess for you expect objects here you'll need to do something like `${new Date(prescription.createdDatetime).toLocaleDateString()}` here too to make sure it always matches locale

Chris-Petty (changes requested):

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.test.tsx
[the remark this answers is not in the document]
aimee-mcneil-melville: Oh thank you!

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
aimee-mcneil-melville: I've made some changes here! Thanks for the article, great learning 😄

client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
aimee-mcneil-melville: I've tried using a map here - it worked this way 🤷‍♀️  Also works with the [key: string] too!

Chris-Petty (approved):
🐟

PR #6618, opened 2025-02-23
6530 gender options are in english

### LINKED ISSUE #6530: French gender options show up in English in patient view
## What went wrong? 😲

Even though there is an option to specify patient's gender in French, when selected and saved, it still shows in English

Image

## Expected behaviour 🤔

Should show in French

## How to Reproduce 🔨

1. Have the patient form JSON files uploaded
2. Have vaccination card & immunization programs running for your datafile
3. Go to your patients and set the gender using French language
4. Note the gender at the top of the screen and in patient list view shows in English

## Your environment 🌱

- Open mSupply Version: V2.5.2
- Legacy mSupply Central Server Version: V7.18.03

- Platform: Galaxy Tab A8 (Android 14)

- Database type: SQLite3

lache-melvin:
Fixes #6530

# 👩🏻‍💻 What does this PR do?

Maps selected gender options to translation keys so they are translated to different languages.

# 🧪 Testing

- [ ] Switch to French
- [ ] Dispensary > Patients > Detail
- [ ] Select a gender and save
- [ ] Toolbar shows gender in french

# 📃 Documentation

  1.
  2.


client/packages/programs/src/hooks/usePatientStore.ts
```diff
@@ -7,9 +7,9 @@ export enum Gender {
   FEMALE = 'FEMALE',
   /** 703117000 Masculine gender*/
   MALE = 'MALE',
-  /** 407379008 transgendered transsexual, female-to-male*/
+  /** 407379008 Transgender, female-to-male*/
   TRANSGENDER_MALE = 'TRANSGENDER_MALE',
-  /** 407378000 Surgically transgendered transsexual, male-to-female*/
```
lache-melvin, line 12: About time this wasn't in our core codebase 💔

client/packages/system/src/Patient/PatientView/DefaultPatientUISchema.json
```diff
@@ -63,7 +63,7 @@
                   ["TRANSGENDER_MALE", "Transgender Male"],
                   ["TRANSGENDER_FEMALE", "Transgender Female"],
                   ["NON_BINARY", "Non-binary"],
-                  ["UNKOWN", "Unknown"]
```
lache-melvin, line 66: typo 😱

client/packages/system/src/Patient/PatientView/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { LocaleKey } from '@common/intl';
+import { GenderType } from '@common/types';
+
+export function getGenderTranslationKey(gender: GenderType): LocaleKey {
```
lache-melvin, line 4: I _think_ this is the right way to go - our API returns a set of gender types, which we can map to translation keys.

However, the options available in the patient form are from the patient schema, which is probably custom per client. So they could have an entirely different set of display values for these gender options. But very hard to extract this out of UI schema - I think better to push the custom translations, get them to update their translations to match their schema at that point?

server/cli/src/refresh_dates.rs
```diff
@@ -53,6 +53,7 @@ fn get_timestamp_fields() -> Vec<TableAndFieldName> {
         ("vaccination", "created_datetime"),
         ("invoice", "backdated_datetime"),
         ("contact_form", "created_datetime"),
+        ("invoice", "cancelled_datetime"),
```
lache-melvin, line 56: fixing tests

client/packages/system/src/Patient/PatientView/utils.ts
andreievg: Custom schemas should really have hard coded translations, only shareable schemas would need translations like this. Btw we do have [`t` now for translation of json forms](https://github.com/msupply-foundation/open-msupply/blob/0699638b856ea96dd33741aa549994c95450d9a9/standard_reports/stock-status/2_5/argument_schemas/arguments_ui.json#L39), this could also have been used in standard form (even for enums), I think I would prefer it done that way, unless we use all of the GenderType and translate them all. Also I think Hormone/Surgical shouldn't be translated (if no key exists)

lache-melvin: The problem I was meaning actually is the hard coded translations. They are available in the JSON form, but map to and enum type on base Patient which is queried by the toolbar here. 

Its really hard to pull the hard coded gender translations out of the UI schema (the gender picker may or may not exist, at any level of depth...) so hoping we are okay with OMS translations of gender enum in places where we arent rendering the form. 

I suppose other option would be to recommend always using translation strings even for custom gender options, once we get overrides in, so we can be sure they match?

andreievg: Ahh ok, that makes sense, I think what you have here is good cheers : )

andreievg (approved):
Thanks @lache-melvin !

PR #6622, opened 2025-02-23
6612 prescribed quantity gets overriden

### LINKED ISSUE #6612: Prescribed Quantity gets overridden/displays weird behavior
## What went wrong? 😲

See my demo video Around 4minutes in

https://drive.google.com/file/d/1cMHiuF0ql1LAun4ekssPYZa69uy-lx3O/view?usp=sharing

## How to Reproduce 🔨

1. Create a prescription
2. Add items
3. Add a prescribed quantity
4. (potenitally interact with other areas like directions)
5. Prescribed quantity seems to disappear!

## Your environment 🌱

- Open mSupply Version: 2.6.0-RC1

- Platform:

- Database type:

GeronimoJohn:
Fixes #6612 

# 👩🏻‍💻 What does this PR do?

Prescribed Quantity gets overriden on initial changes of its input. This is happening due to the logic that's within a `useEffect`.  It has a dependency that gets updated when we initially change prescribed quantity. Which re-triggers the prescribed quantity logic that's only meant to run on initial load.

Moved the initial setting of prescribed quantity logic to its own `useEffect` without any dependencies. Triggering the logic only on initial load

https://github.com/user-attachments/assets/9788dca0-5642-4e02-a5da-b794825bc653

# 🧪 Testing

- [ ] Navigate to a prescription
- [ ] Add a new item
- [ ] Edit the prescribed quantity field
- [ ] It should retain the initial value that the user inputs
- [ ] It should auto allocate quantity to lines
- [ ] Click save
- [ ] Item gets created!

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -210,19 +210,21 @@ export const PrescriptionLineEditForm: React.FC<
   const note = prescriptionLineWithNote?.note ?? '';
 
   useEffect(() => {
-    const selectedItem = items.find(
-      prescriptionItem => prescriptionItem.id === item?.id
-    );
+    // sets prescribed quantity on initial load
     if (preferences?.editPrescribedQuantityOnPrescription) {
-      const newPrescribedQuantity: number =
+      const selectedItem = items.find(
+        prescriptionItem => prescriptionItem.id === item?.id
+      );
+      const newPrescribedQuantity =
         selectedItem?.lines?.find(
           ({ prescribedQuantity }) =>
             prescribedQuantity != null && prescribedQuantity > 0
         )?.prescribedQuantity ?? 0;
-
       setPrescribedQuantity(newPrescribedQuantity);
     }
+  }, []);
 
+  useEffect(() => {
     const newIssueQuantity = Math.round(
       allocatedUnits / Math.abs(Number(packSizeController.selected?.value || 1))
     );
@@ -328,6 +330,13 @@ export const PrescriptionLineEditForm: React.FC<
                     min={0}
                     decimalLimit={2}
                     onBlur={() => {}}
+                    slotProps={{
+                      htmlInput: {
+                        sx: {
+                          backgroundColor: 'background.white',
+                        },
+                      },
+                    }}
                   />
                 </Grid>
               )}
@@ -345,18 +354,13 @@ export const PrescriptionLineEditForm: React.FC<
                   slotProps={{
                     htmlInput: {
                       sx: {
-                        backgroundColor: disabled
-                          ? undefined
-                          : 'background.white',
+                        backgroundColor: 'background.white',
                       },
                     },
                   }}
                 />
                 <InputLabel sx={{ fontSize: 12 }}>
-                  {t('label.unit-plural_one', {
-                    // unit-plural_one has been specified to deactivate pluralisation.
-                    // This is to handle the case where units have not been entered for the item.
-                    // see https://github.com/msupply-foundation/open-msupply/issues/5978
+                  {t('label.unit-plural', {
                     count: issueUnitQuantity,
                     unit: item?.unitName,
                   })}
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -211,7 +211,10 @@ export const PrescriptionLineEditForm: React.FC<
 
   useEffect(() => {
     // sets prescribed quantity on initial load
-    if (preferences?.editPrescribedQuantityOnPrescription) {
+    if (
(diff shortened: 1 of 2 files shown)


client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -328,6 +330,13 @@ export const PrescriptionLineEditForm: React.FC<
                     min={0}
                     decimalLimit={2}
                     onBlur={() => {}}
+                    slotProps={{
+                      htmlInput: {
+                        sx: {
+                          backgroundColor: 'background.white',
+                        },
+                      },
+                    }}
```
GeronimoJohn, line 339: Side Quest: Fixed the grey background of the input

client/packages/invoices/src/Prescriptions/api/hooks/utils.tsx
```diff
@@ -514,6 +514,13 @@ export const UnitQuantityCell = (props: CellProps<DraftPrescriptionLine>) => (
     id={getPackQuantityCellId(props.rowData.stockLine?.batch)}
     min={0}
     decimalLimit={2}
+    slotProps={{
+      htmlInput: {
+        sx: {
+          backgroundColor: 'background.white',
+        },
+      },
+    }}
```
GeronimoJohn, line 523: Side Quest: Fixed the grey background of the input

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
jmbrunskill: Thank you!

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -345,18 +354,13 @@ export const PrescriptionLineEditForm: React.FC<
                   slotProps={{
                     htmlInput: {
                       sx: {
-                        backgroundColor: disabled
-                          ? undefined
-                          : 'background.white',
+                        backgroundColor: 'background.white',
                       },
```
GeronimoJohn, line 358: removed disabled logic, disabled prop already manages it's own color

CarlosNZ (commented):
Looks good, thanks. But now we have *four* different `useEffect`s in this component -- I wonder if we can consolidate them? Especially since there's so much overlap between the dependancy arrays.

I quickly tried this and it seemed to work, but I haven't put it through its paces:
```tsx
  useEffect(() => {
    // sets prescribed quantity on initial load
    if (
      preferences?.editPrescribedQuantityOnPrescription &&
      prescribedQuantity === null
    ) {
      const selectedItem = items.find(
        prescriptionItem => prescriptionItem.id === item?.id
      );
      const newPrescribedQuantity =
        selectedItem?.lines?.find(
          ({ prescribedQuantity }) =>
            prescribedQuantity != null && prescribedQuantity > 0
        )?.prescribedQuantity ?? 0;
      setPrescribedQuantity(newPrescribedQuantity);
    }

    if (!isAutoAllocated) {
      setIssueUnitQuantity(allocatedUnits);
      return;
    }

    const newIssueQuantity = Math.round(
      allocatedUnits / Math.abs(Number(packSizeController.selected?.value || 1))
    );
    if (newIssueQuantity !== issueUnitQuantity)
      setIssueUnitQuantity(newIssueQuantity);
    setAllocationAlerts([]);

    if (items.find(prescriptionItem => prescriptionItem.id === item?.id))
      setAbbreviation('');
    setDefaultDirection('');
  }, [item?.id, allocatedUnits, packSizeController.selected?.value]);
```

That replaces all four -- what do you think?

GeronimoJohn:
@CarlosNZ, yesss I was in the same mindset that there was a lot. The above makes sense! Testing it out but most likely going to apply it

GeronimoJohn:
~I tried the consolidating everything into one `useEffect`, but found an issue with setting `issueUnitQuantity`. Adding it all in one seems to over ride the added value to the unit quantity input.~

https://github.com/user-attachments/assets/784f9b9d-35c9-46a3-944a-17ad5d953669

~`allocatedUnits` within the dependency array seems to be the culprit. Move out it out to it's own `useEffect` seems to solve the issue.~

**UPDATE**: I found more bugs. Continuing to investigate

GeronimoJohn:
Have tried consolidating into one useEffect. Though found a few hoops to go through. 

Such as: 
- Prescribed Quantity keeping value on initial item create
- Issued Quantity keeping value on initial item create
- Updating Issued Quantity should retain its value
- Navigating between two created Items should retain their prescribed quantity and not the previous items quantity

That being said, I was able to reduce the amount of `useEffect` to two, heres my attempt for a fix that keeps the functionalities mentioned above: https://github.com/msupply-foundation/open-msupply/pull/6622/commits/3f14a7e3159e6854d0205ade4c9ec52d5adb06e6

CarlosNZ:
I've given this a good going over and I think it's behaving itself now, so I think we're good to merge. I still think all the logic here is overly complicated and very difficult to debug, but that's a problem for another day :)

PR #6623, opened 2025-02-24
6521 Remove whitespace from username input

### LINKED ISSUE #6521: Issue logging in with trailing whitespace characters
## What went wrong? 😲

While assisting a client with login issues, it seems that they had entered two space characters by accident at the end of their username. Unfortunately, if the incorrect username is used, it is also cached.

This may cause some headaches for users expecting that the password is the issue and not their username.

This might be more of a feature request, rather than considering this a bug.

## Expected behaviour 🤔

Options:

- Remove any trailing white space characters (This would possibly affect/require change in mSupply)
- Revert to the previously successful username that was last logged into
- Clear the username field if the username is not correct
- Highlight to the user that the username does not exist (Preferred?)

## How to Reproduce 🔨

Log in to an open mSupply store with a 'space' character at the end of the username.

## Your environment 🌱

- Open mSupply Version: 2.3.0
- Legacy mSupply Central Server Version: v7.18.25

- Platform: Android, Browser

- Database type: SQLite

noel-yeldos:
Fixes #6521

# 👩🏻‍💻 What does this PR do?
Remove whitespace from username input field on submission

Change needed as OG trims whitespace so aligning OMS functionality

I'm not sure if the change on the frontend is enough? It seems to be working as expected.

# 🧪 Testing

- [ ] Go to login page
- [ ] Enter username but add whitespace before or after
- [ ] Enter password and login
- [ ] See successful login despite whitespace in username field

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/host/src/components/Login/hooks.ts
+++ b/client/packages/host/src/components/Login/hooks.ts
@@ -47,7 +47,7 @@ export const useLoginForm = (
 
   const onLogin = async () => {
     setError();
-    const { error, token } = await login(username, password);
+    const { error, token } = await login(username.trim(), password);
     setError(error);
     setPassword('');
     if (!token) return;


fergie-nz (approved):
Nice Noel LGTM! 🚀

PR #6626, opened 2025-02-24
Prevent needing to double-click 'Next' when adding items to Internal Orders

### LINKED ISSUE #6079: User has to double-click 'Next' when adding items to Internal Orders
## What went wrong? 😲

Whenever you add an item to an Internal Order and update the quantity requested, the page does a very quick refresh if you click anywhere on the screen. From the user perspective, they will click 'Next' to move onto the next item -> the refresh is still happening, causing the user to double-click the button every time. Quite annoying!

Video ->>> [here](https://drive.google.com/drive/folders/1lsP-fup8cN9DztjpQhKxsoPxpZTYkCKm)

## Expected behaviour 🤔

Maybe the page shouldn't do a quick refresh, so the user can move onto the next action. They shouldn't have to double-click either

## How to Reproduce 🔨

1. Go to Replenishment -> Internal Orders
2. Create a new Internal Order or modify an existing one
3. Add new items to it -> update quantity
4. If you click anywhere on the screen, the page will refresh
5. Move onto the next item -> update quantity -> click 'Next'
6. Because the refresh is still happening, you will need to double-click the button every time
7. Annoying!

## Your environment 🌱

- Database type: PostgreSQL

aimee-mcneil-melville:
Fixes #6079 

# 👩🏻‍💻 What does this PR do?

When a user is entering information to an internal order, the 'Next' button now responds on the first click, instead of the second - and saves any changes to the field as expected.

Each time the user clicked next it would appear to be 'clicked' however the action actually being taken was 'unblur' then 're-render the page' - thus the 'next' action wasn't getting called until the second time it was clicked. This was also occurring with changes to other fields on the same page, and with the 'previous' button.

The response edit page now has a wrapper and an inner, and the wrapper handles loading state/page refreshing, and returns the page inner. This fix is the same as the fix in #5867, though the issues presented differently.

# 🧪 Testing

- [ ] Go to Replenishment -> Internal Orders
- [ ] Create a new Internal Order or modify an existing one
- [ ] Add or edit items -> update quantity
- [ ] Click next - the next item will be selected
- [ ] Click Previous - your changes will be saved
- [ ] No refreshes are happening after editing something and then clicking somewhere

# 📃 Documentation

  1.
  2.

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestEditPage.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestEditPage.tsx
@@ -8,7 +8,7 @@ import {
   useBreadcrumbs,
   useParams,
 } from '@openmsupply-client/common';
-import { useRequest } from '../../api';
+import { RequestFragment, useRequest } from '../../api';
 import { ListItems } from '@openmsupply-client/system';
 import { AppRoute } from '@openmsupply-client/config';
 import { useDraftRequisitionLine, usePreviousNextRequestLine } from './hooks';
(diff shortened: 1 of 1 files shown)


client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestEditPage.tsx
```diff
@@ -52,12 +70,9 @@ export const RequestLineEditPage = () => {
     });
   }, [currentItem]);
 
-  if (isLoading || requestIsLoading) return <BasicSpinner />;
```
lache-melvin, line 55: I think important to note - it was just the `isLoading` here that was causing the issue - unmounting the entire page whenever the `useDraftRequisitionLine` loading state was true, which was every time we saved. However I like the introduction of the `Inner` component, let's keep it 😁

lache-melvin (approved):
LGTM 🙏

PR #6649, opened 2025-02-24
5837 requisition line column and plugin data

### LINKED ISSUE #5837: New column on requisition_line plugin

Second part of #5833, show data form requisition line plugin data in internal order and response requisition lists and line edit pages, this new column is called Area AMC

Would require plugin data to store area amc

andreievg:
Fixes #5837 

# 👩🏻‍💻 What does this PR do?

* Front end changes to add requisition line column, quite a bit of juggling around of types to have RequestLineFragment available in plugin types
* Add plugin data to transform request requisition line

Convinced more that plugin data should be on the record, added more reasons for this in https://github.com/msupply-foundation/open-msupply/issues/2951#issuecomment-2650673745

There is also "update" on request requisition line that will trigger transform, this will cause duplicates for plugin data atm

Oops just realise should really be transform request requisition lines: https://github.com/msupply-foundation/open-msupply/issues/6650

# 🧪 Testing
 
Using this branch of plugin examples: https://github.com/msupply-foundation/open-msupply-plugins/pull/10
Install backend plugin: 
```
cargo run --bin remote_server_cli -- generate-and-install-plugin-bundle -i '../client/packages/plugins/myPluginbundle/backend' --url 'http://localhost:8000' --username admin --password pass
```
Now you when creating new internal order (and add from master list) you should see plugin data being added (in database browser), and you will see new column in request requisition


client/packages/common/src/plugins/types.ts
```diff
@@ -15,9 +18,13 @@ export type Plugins = {
...
     columns: ColumnDefinition<StockLineRowFragment>[];
   };
+  requestRequisitionColumn?: {
+    StateLoader: React.ComponentType<{ requestLines: RequestLineFragment[] }>[];
+    columns: ColumnDefinition<RequestLineFragment>[];
+  };
 };
 
-type PluginData<D> = { relatedRecordId: string; data: D };
+type PluginData<D> = { relatedRecordId?: string | null; data: D };
```
andreievg, line 27: Not sure how this was passing compilation previously (in plugins), didn't want to spend too much time investigating old hooks, the new hook i added to AggrageAmcColumn example plugin did complain (since relatedRecordId was changed to be optional)

client/packages/plugins/backendTypes/generated/TransformRequisitionLineOutput.ts
```diff
@@ -1,4 +1,5 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+import type { PluginDataRow } from "./PluginDataRow";
 import type { RequisitionLineRow } from "./RequisitionLineRow";
 
-export type TransformRequisitionLineOutput = { transformed_lines: Array<RequisitionLineRow>, };
+export type TransformRequisitionLineOutput = { transformed_lines: Array<RequisitionLineRow>, plugin_data: Array<PluginDataRow> | null, };
```
andreievg, line 5: Was hoping plugin_data would be optional rather then null, but this is ok for now

client/packages/requisitions/src/RequestRequisition/api/api.ts
```diff
@@ -282,9 +282,7 @@ export const getRequestQueries = (sdk: Sdk, storeId: string) => ({
 
     throw new Error('Unable to create requisition');
   },
-  insertProgram: async (
-    input: InsertProgramRequestRequisitionInput
-  ) => {
+  insertProgram: async (input: InsertProgramRequestRequisitionInput) => {
```
andreievg, line 285: prettier changes, hmm, is it mine or is it repo config, not sure

client/packages/requisitions/src/RequestRequisition/api/index.ts
```diff
@@ -1,2 +1,7 @@
 export * from './hooks';
 export * from './operations.generated';
+export {
+  RequestLineFragment,
+  ItemWithStatsFragment,
+  ItemInformationFragment,
+} from '@openmsupply-client/system';
```
andreievg, line 7: RequestLineFragment lived in higher level package, didn't want to introduce circular dependency, so moved the fragment generation to system and re-exported here

client/packages/system/src/RequestRequisitionLine/operations.graphql
```diff
@@ -0,0 +1,73 @@
...
+  }
+  item {
+    ...ItemWithStats
+  }
+  reason {
+    ...ReasonOptionRow
+  }
+}
+
+query OnlyHereToAvoidUnusedWarnings {
```
andreievg, line 69: had to add this otherwise wasn't compiling because getSdk was empty and not using parameters

server/service/src/requisition/request_requisition/insert_program.rs
```diff
@@ -158,6 +165,13 @@ fn validate(
     ))
 }
 
+pub(super) struct GenerateResult {
```
andreievg, line 168: noticed that response and request were using the same GenerateResult, updated for them to be separate

server/service/src/backend_plugin/types/transform_requisition_lines.rs
```diff
@@ -15,6 +15,7 @@ pub struct Input {
 #[ts(rename = "TransformRequisitionLineOutput")]
 pub struct Output {
     pub transformed_lines: Vec<RequisitionLineRow>,
+    pub plugin_data: Option<Vec<PluginDataRow>>,
```
andreievg, line 18: Plugin data is now returned by plugin, didn't want to save it during plugin execution, to still have transaction rollback

server/service/src/requisition/request_requisition/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod generate;
+
```
lache-melvin, line 2: 🔥

lache-melvin (approved):
Should R&R form also use this plugin? It is creating request requisition lines...

~~Also flagged the batch request requisition mutations - we now only use them for delete, but all the code is still there (and not using plugin) - even more motivation to remove! Can we add clean up issue?~~ batch mutations not used but service code is 🙈 

Wondering about processors too, and other more fine grained mutations (e.g. `use_suggested_quantity`) - seems fair enough, they're only changing one field, do we really need to run the transform? Same time, if that was a column I wanted to transform, and was getting overwritten by those actions ... do we need to document expectations with each plugin interface about where it can be expected to be called?

lache-melvin:
> Convinced more that plugin data should be on the record, added more reasons for this in https://github.com/msupply-foundation/open-msupply/issues/2951#issuecomment-2650673745

Agree, plugin data on record would be much cleaner, this feels like an awful lot of surface area, easy to miss something

andreievg:
Thanks for review @lache-melvin, closing as no extra diff to the stack top

PR #6664, opened 2025-02-24
6657 Misc prescription/insurance fixes

### LINKED ISSUE #6657: Dispensing follow-up: History bug & more
Will include the following small fixes in this issue, from #6604 :

- [x] The History shown lists all items in the current prescription, even when the prescription is in 'Picked' status, i.e. not actually given to the patient yet. I found that a little confusing: for me, the history shows items prior to this prescription
- [x] There is no translation for `PRESCRIPTION_STATUS_CANCELLED` : need to add `"log.prescription-status-cancelled": "Prescription cancelled",` to the JSON file
- [x] When verified, the PRINT button is disabled but not disabled
- [x] The wording on 'Is it an active insurance?' doesn't read very nicely. Perhaps just a heading of 'Status' would be better
- [x] The title of the modal 'New insurance' isn't quite right. Could you change to 'New insurance policy'?

CarlosNZ:
Fixes #6657

# 👩🏻‍💻 What does this PR do?

I've put each item in a separate commit so can be reviewed separately if you like:

- The History shown lists all items in the current prescription, even when the prescription is in 'Picked' status, i.e. not actually given to the patient yet. I found that a little confusing: for me, the history shows items prior to this prescription 1bc4a80488
- There is no translation for `PRESCRIPTION_STATUS_CANCELLED` : need to add `"log.prescription-status-cancelled": "Prescription cancelled",` to the JSON file 8f454efc17
- When verified, the PRINT button is disabled but not disabled 8d2b10f45b
- The wording on 'Is it an active insurance?' doesn't read very nicely. Perhaps just a heading of 'Status' would be better 1c7fb68f38 (Also, change to Toggle switch: c35101219b)
- The title of the modal 'New insurance' isn't quite right. Could you change to 'New insurance policy'? b306e2ac64
- Move add item to be the first button: df2b86b691

See comments

# 🧪 Testing

 - [ ] Confirm the current prescription's items are excluded from the "History" modal
 - [ ] Go to the "Log" tab of a Cancelled prescription -- confirm the status shows as "Cancelled", not "PRESCRIPTION_STATUS_CANCELLED"
 - [ ] Confirm Print button is never disabled. You will be informed if you lack permissions to generate a report.
 - [ ] Insurance Active/Inactive now just shows "Status" with a Toggle switch
 - [ ] Modals for insurance say "New insurance policy" / "Edit insurance policy" in the Header
 - [ ] Confirm "Add items" is the first button in the App Bar (top of page)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1207,6 +1206,7 @@
   "log.prescription-deleted": "Deleted",
   "log.prescription-status-picked": "Picked",
   "log.prescription-status-verified": "Verified",
+  "log.prescription-status-cancelled": "Cancelled",
```
CarlosNZ, line 1209: Not exactly what was asked for, but is consistent with the other statuses

client/packages/system/src/Patient/Insurance/InsuranceModal.tsx
```diff
@@ -127,12 +128,37 @@ export const InsuranceModal: FC = (): ReactElement => {
               })
             }
           />
-          <InsuranceStatusToggle
-            isActive={draft.isActive}
-            onChange={isActive =>
-              updatePatch({
-                isActive,
-              })
+          <InputWithLabelRow
```
CarlosNZ, line 131: Didn't think this quite justified its own component now.

Also, have left the two checkboxes, but I'm open to reducingn it to just a single toggle if that seems prefereable?

jmbrunskill (approved):
Looks great, just not sure about the toggle always showing active...

PR #6670, opened 2025-02-25
Add functionality to disable reports

### LINKED ISSUE #6300: Add method to deactivate reports
We want to be able to deactivate a report on a sync-system-wide basis. (ie deactivate all reports for a particular code for a central server and all its remote sites).

This also needs to be on a is_custom basis. 

- [ ] Add a new column is_active to reports
- [ ] Setting or method needs to act on all reports of a code and is_custom
- [ ] If is_custom = true is deactivated, all reports of that code where is_custom = true should be deactivated

Some discussion or design required around how this should be implemented for a report. 
Suggesting a cli to central server with arguments of 'code', 'is_custom', and 'active-status' (so that there is a facility to reactivate reports).

If current repository method updates the entire record, we should add a new method to update only the is_active field of the record.

EthanMcQ-TMF:
Fixes #6300 

# 👩🏻‍💻 What does this PR do?

- Add new `is_active` column to `report` table
- Add new `toggle-report` command to cli
  - Takes a `--code` argument and swaps the `is_active` value of that report
  - Accepts an optional `--is_custom` filter value
  - Can be made to hard set the `is_active` value with `--enable` or `--disable` flags
    - Otherwise just NOTs the fields
- Add option to filter by `is_active` to `/reports` path graphql api
  - We can still request deactivated reports, wanted to leave that in since we want to control this from a GUI at some point
- Update client to filter report request for `is_active = true`

# 🧪 Testing

- Testing toggle command
  - [ ] Run `toggle-report`, should reject without code
  - [ ] Run `toggle-report --code [code]`, `is_active` field should be toggled true/false
  - [ ] Run `toggle-report --code [code] --is_custom true`, only `is_custom = true` reports should toggle
  - [ ] Run `toggle-report --code [code] --enable`, `is_active` should be set to `true`
  - [ ] Run `toggle-report --code [code] --disable`, `is_active` should be set to `false`
- Testing disabling reports
  - [ ] Open reports menu in omSupply, observe enabled reports
  - [ ] Disable report with toggle command
  - [ ] Refresh reports menu in omSupply, the disabled report should no longer appear

# 📃 Documentation

- [ ] Added to `standard_reports/README.md`

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/cli/src/cli.rs
```diff
@@ -624,6 +643,36 @@ async fn main() -> anyhow::Result<()> {
...
+            let connection_manager = get_storage_connection_manager(&settings.database);
+            let con = connection_manager.connection()?;
+
+            let mut filter = ReportFilter::new().code(EqualFilter::equal_to(&code));
+            match is_custom {
+                Some(value) => {
+                    filter = filter.is_custom(value);
+                },
+                None => {}
+            }
```
fergie-nz, line 657: This is great thanks Ethan.

One thing I didn't explicitly state in the issue was being able to disable the standard reports while maintaining custom reports. (This is in the case where we have custom reports replacing standard reports for a client, but the custom reports are on a more recent version of OMS than some of their remote sites. The remote sites in this case would show the standard code-equivalent report where the desired behaviour might be to show no reports).

And you've written this in a generic way so we can have a targeted response so thank you for thinking of all cases.

server/repository/src/migrations/v2_06_00/add_report_is_active.rs
```diff
@@ -0,0 +1,20 @@
...
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "add_report_is_active"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                ALTER TABLE report ADD COLUMN IF NOT EXISTS is_active BOOL NOT NULL DEFAULT true;
```
fergie-nz, line 14: I notice we don't usually have this check 'IF NOT EXISTS' in other migrations. Does it add extra safety which is necessary in this case?

standard_reports/README.md
```diff
@@ -174,7 +174,7 @@ This command is used to update standard reports to the current branch in databas
 
 ### Show Report
 
-`show-report --path <path-to-report-dir-containing-report-manifest.json> --config <optional-path-to-dir-containing-test-config.json>
+`show-report --path <path-to-report-dir-containing-report-manifest.json> --config <optional-path-to-dir-containing-test-config.json>`
```
fergie-nz, line 177: nice thank you

server/repository/src/migrations/v2_06_00/add_report_is_active.rs
fergie-nz: In fact  I think this check is not compatible with sqlite, is that right?

fergie-nz (changes requested):
Looks great thanks Ethan.

Just missing migration issue on sqlite and test cases need updating (existing tests broken), also it would be awesome to have new tests cases to cover we are selecting reports appropriately with the version control checking when some are deactivated.

I also wonder if we could add some info in the cli when reports are toggled - I imagine users of this would want to know that the reports they have selected have had the desired effect. 

For example when I toggle reports and use a non existent code, there is no error.

I also think would be useful to clarify exactly what has changed. I think confusions could arise where a selection of reports are being toggled where some of those are currently active and some are inactive. This might not be obvious to the user of the cli.

Was also wondering about a use case where we might want to deactivate a grouping of reports with a certain code prefix. That could be useful if there were many reports in use by a particular client for example and it was inconvenient to deactivate all of them. But I guess that can wait to another issue if required.

server/repository/src/migrations/v2_06_00/add_report_is_active.rs
EthanMcQ-TMF: Whoops. You're right. Sqlite is forever my enemy 🙃

EthanMcQ-TMF:
@fergie-nz
- Fixed up the broken tests
- Added logging during the toggle action
- Upserting records no longer hard sets `is_active = true`, it'll maintain the existing state if it exists

I looked into doing "starts with" or "ends with" filters but that'd require changing the code filter type in the `ReportFilter` and that has some knock on changes required that I truthfully just didn't wanna deal with right now 😅

server/service/src/standard_reports.rs
```diff
@@ -57,11 +57,13 @@ impl StandardReports {
...
         let mut num_std_reports = 0;
         for report in reports_data.reports {
-            let existing_report_count = ReportRepository::new(con).count(Some(
-                ReportFilter::new().id(EqualFilter::equal_to(&report.id)),
-            ))?;
+            let existing_report = ReportRowRepository::new(con).find_one_by_id(&report.id)?;
+            let set_active = match &existing_report {
+                Some(report) => report.is_active,
+                None => true
+            };
```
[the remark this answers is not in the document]
EthanMcQ-TMF: @fergie-nz I've updated this functionality where if a new report is being added and previous versions of that report already exist in the db, it will maintain the deactivated status (only if _all_ the previous versions are disabled)

I didn't think putting that behind an extra argument is entirely necessary. I think it's better if this behaviour is consistent between uploading new reports or reloading embedded reports.

fergie-nz (approved):
Looks great!

PR #6671, opened 2025-02-25
6668 fix context type in report migration

### LINKED ISSUE #6668: Graphql /reports path broken after reinitialising table (Postgres)
## What went wrong? 😲

The graphql api `/reports` path is broken after running the `reinitialise_reports.rs` migration if using postgres.

The `context` column in the table was previously an enum. This was swapped for a text field in the migration. When the API runs a filter on `context` it formats the query to use an enum. Psql cannot run `enum = text`

Image

## Expected behaviour 🤔

Reports should return correctly filtered

## How to Reproduce 🔨

1. Run server with `--features postgres` enabled
2. Make sure `reinitialise_reports.rs` has been run
3. Open the reports screen in the client
4. Observe no reports

## Your environment 🌱

- Open mSupply Version: v2.6.0-RC

- Platform: macOS

- Database type: PostgreSQL

andreievg:
Fixes #6668

# 👩🏻‍💻 What does this PR do?

Update fragment to set context to be enum for postgres.
Changing fragment name will mean it will re-run

Custom report will be lost, but that is ok
For some reason this migration did not run so we had to add it again

https://github.com/msupply-foundation/open-msupply/blob/342394370b71b799f9af617022f128262a9a5427/server/repository/src/migrations/v2_00_00/returns/return_context_types.rs#L8-L9

# 🧪 Testing

Make sure reports work in postgres

### DIFF
--- a/server/repository/src/db_diesel/report_row.rs
+++ b/server/repository/src/db_diesel/report_row.rs
@@ -10,10 +10,12 @@ use diesel_derive_enum::DbEnum;
 
 use serde::{Deserialize, Serialize};
 
-#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ValueEnum)]
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ValueEnum, Default)]
+#[cfg_attr(test, derive(strum::EnumIter))]
 #[DbValueStyle = "SCREAMING_SNAKE_CASE"]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum ContextType {
+    #[default]
     Asset,
     InboundShipment,
     OutboundShipment,
@@ -51,7 +53,7 @@ joinable!(report -> form_schema (argument_schema_id));
 allow_tables_to_appear_in_same_query!(report, form_schema);
 
 #[derive(
-    Clone, Insertable, Queryable, Debug, PartialEq, Eq, AsChangeset, Serialize, Deserialize,
+    Clone, Insertable, Queryable, Debug, PartialEq, Eq, AsChangeset, Serialize, Deserialize, Default,
 )]
 #[diesel(table_name = report)]
 pub struct ReportRow {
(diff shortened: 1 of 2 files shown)


server/repository/src/db_diesel/report_row.rs
```diff
@@ -67,24 +69,6 @@ pub struct ReportRow {
     pub version: String,
     pub code: String,
 }
-
-impl Default for ReportRow {
```
andreievg, line 71: Removed as we are using derive now

server/repository/src/db_diesel/report_row.rs
```diff
@@ -180,3 +164,39 @@ impl Upsert for ReportRow {
         )
     }
 }
+
+#[cfg(test)]
+mod test {
```
andreievg, line 169: For future to make sure we don't break this.

We should have theses test on all enums

server/repository/src/migrations/v2_06_00/reinitialise_reports.rs
```diff
@@ -4,10 +4,26 @@ pub(crate) struct Migrate;
...
+            "context_type"
+        } else {
+            "TEXT"
+        };
+
+        if cfg!(feature = "postgres") {
+            sql!(
+                connection,
+                r#"
+                ALTER TYPE context_type ADD VALUE 'OUTBOUND_RETURN';
```
andreievg, line 21: See PR comment

server/repository/src/db_diesel/report_row.rs
jmbrunskill: Would be good to share this as an example of the wiki maybe? Then we can easily point to it as a how to next time this comes up?

jmbrunskill (approved):
Thanks Andrei.

I'm a little concerned about the will loose custom report but `that's ok` comment though.

How do we know for sure?
I know if will at least affect Alain.

Have we communicated to support that we'll be deleting all custom reports on a 2.6 upgrade? Should be clearly marked as breaking changes in the release notes at the very least.

I feel like we might have slightly broken it on 2.5 as well?

server/repository/src/db_diesel/report_row.rs
andreievg: Thanks, good idea: https://github.com/msupply-foundation/open-msupply/wiki/Unit-Test

andreievg:
> I'm a little concerned about the will loose custom report but that's ok comment though.

Agree, we were planning this from the start though

> I know if will at least affect Alain

Custom reports will be lost on central server yeah, will ask Roxy to msg Alain

> Have we communicated to support that we'll be deleting all custom reports on a 2.6 upgrade? Should be clearly marked as breaking changes in the release notes at the very least.

I'll msg them now thanks : )

> I feel like we might have slightly broken it on 2.5 as well?

I think 2.5 is fine, apart from not having omSupply sync and reports not fully working as expected (is custom not working)

andreievg:
Thanks for reivew @jmbrunskill

PR #6675, opened 2025-02-25
6441 fix add item variant

### LINKED ISSUE #6441: Unable to create item variants
## What went wrong? 😲

When trying to create an item variant, entering the name causes the page to crash

https://github.com/user-attachments/assets/d553456c-a04a-419d-aa55-34425c99e53b

## Expected behaviour 🤔

no crashing, name entered

## How to Reproduce 🔨

1. Choose an item to test on from Catalogue -> Items
2. Click on Variants and create a new one
3. Try entering name -> should lead to the page crashing

## Your environment 🌱

- Open mSupply Version: `V2.6.0-build-test-3`
- Legacy mSupply Central Server Version: `V7.19.03`

- Database type: `PostgreSQL`

jmbrunskill:
Fixes #6441 (maybe???)

# 👩🏻‍💻 What does this PR do?

Change the way we open the modal to be consistent with the other button.
Not sure why it would matter though...

This probably isn't the right fix, I don't understand how we have different behaviour with this change!

# 🧪 Testing

- [ ] Open the Catalogue -> Items -> Variants tab on OMS Central Server
- [ ] Choose an item that has no variants
- [ ] Click the `Create a new one` button

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
+++ b/client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
@@ -50,7 +50,7 @@ export const ItemVariantsTab = ({
         {itemVariants.length === 0 ? (
           <NothingHere
             body={t('messages.no-item-variants')}
-            onCreate={onOpen}
+            onCreate={() => onOpen()}
           />
         ) : (
           itemVariants.map(v => (
--- a/client/packages/common/src/ui/components/errors/NothingHere.tsx
+++ b/client/packages/common/src/ui/components/errors/NothingHere.tsx
@@ -22,7 +22,7 @@ export const NothingHere: React.FC<NothingHereProps> = ({
   const createButtonText = buttonText || t('button.create-a-new-one');
 
   const CreateButton = !!onCreate ? (
-    <Button sx={{ textTransform: 'none' }} onClick={onCreate}>
+    <Button sx={{ textTransform: 'none' }} onClick={() => onCreate()}>
       {createButtonText}
     </Button>
   ) : undefined;


client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
```diff
@@ -50,7 +50,7 @@ export const ItemVariantsTab = ({
         {itemVariants.length === 0 ? (
           <NothingHere
             body={t('messages.no-item-variants')}
-            onCreate={onOpen}
+            onCreate={() => onOpen()}
```
lache-melvin, line 53: Nice find James! I think we should actually fix this in the `NothingHere` component. 

Modal `onOpen` accepts _either_ nothing (create mode) or an existing state (for opening in edit mode.) `NothingHere` is passing the click event into the `onCreate` handler, so before this change, the click event was being passed right through and becoming the draft item variant 😱 

Then when we type in the input, the validation function was throwing because the expected fields obv weren't there.

I think pretty safe to say `onCreate` shouldn't ever want the click event, so we should update this one:

https://github.com/msupply-foundation/open-msupply/blob/a54bbe3485f8d76bba24288fa64694c83791969b/client/packages/common/src/ui/components/errors/NothingHere.tsx#L25

To `onClick={() => onCreate()}`

lache-melvin (changes requested):
Requesting we do this change in the common `NothingHere` component so we don't miss again in future 🙏

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
jmbrunskill: Nice thanks @lache-melvin was not sure what was going on!

lache-melvin (approved):
Epic thanks for changes James 🙏

PR #6684, opened 2025-02-25
6650 rename requisition line transformation and expose plugin data query to plugins

### LINKED ISSUE #6650: Rename plugin transform_requisition_line to requisition_line and add get plugin data method

Should rename transform_requisition_line to transform_request_requisition_line. Also add a method to get plugin data in backend plugin (otherwise have to do raw sql)

andreievg:
Fixes #6650

# 👩🏻‍💻 What does this PR do?

Sorry I did a bit more then specified by the issue. 

Renamed transform_requisition_line to transform_request_requisition_line, while renaming I decided to use PluginType as source of truth, this also included updating types in the backend plugin package.json (see example branch mentioned in testing)

Added get plugin data method that is exposed now to plugins

Exposed request edit line in front end to also have columns shown

I found a bug where plugin data is set for wrong id, because id is overwritten after generating requisition line (since we are using id generated on front end), this was fixed by providing item_id -> id mapping to generate requisition line

Plugin data filter is exposed to plugin, we starting to expose types that are deep in our code, which is very convenient but would require testing strategy, which will be implemented for CIV plugins.

# 🧪 Testing

Use this branch of examples: https://github.com/msupply-foundation/open-msupply-plugins/pull/11

Create internal order, see aggregate AMC value appearing in table and edit lines view. Delete a line with 0 aggregate amc. Create manual response requisition with that item, set amc. Insert that one line again in internal order, see aggregate amc change having value. Aggregate AMC should be sum of all of the latest response requisition line AMC for that item in that store


client/packages/plugins/README.md
```diff
@@ -173,7 +173,7 @@ You can work on plugins as if they were part of the app (types should be shared,
 
 ```bash
 # From server directory
-cargo run --bin remote_server_cli -- generate-plugin-bundle -i ../client/packages/plugins/mynewplugin/frontend -o pluginbundle.json
+cargo run --bin remote_server_cli -- generate-plugin-bundle -i ../client/packages/plugins/myPluginBundle/frontend -o pluginbundle.json
```
andreievg, line 176: decided to change this since submodule example in readme uses myPluginBundle as plugin name

server/repository/src/db_diesel/backend_plugin_row.rs
```diff
@@ -8,10 +8,10 @@ use diesel_derive_enum::DbEnum;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[serde(rename_all = "snake_case")]
 pub enum PluginType {
-    Amc,
-    TransformRequisitionLines,
+    AverageMonthlyConsumption,
```
andreievg, line 13: This was not necessary, but not a huge change, sorry for the extra diff

server/repository/src/db_diesel/backend_plugin_row.rs
```diff
@@ -8,10 +8,10 @@ use diesel_derive_enum::DbEnum;
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+#[serde(rename_all = "snake_case")]
```
andreievg, line 11: snake case now because this maps directly to main plugins type now, made comments in types/mod.rs

server/repository/src/db_diesel/filter_sort_pagination.rs
```diff
@@ -101,16 +102,22 @@ impl StringFilter {
     }
 }
 
-#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Debug, TS, Serialize, Deserialize)]
 pub struct EqualFilter<T>
 where
     T: 'static,
 {
+    #[ts(optional)]
```
andreievg, line 110: Without optional Option<T> is treated as null, with optional treated as undefined

server/service/src/backend_plugin/boajs/call_plugin.rs
```diff
@@ -30,16 +31,22 @@ impl PartialEq for BoaJsPluginError {
...
 }
 
+fn plugin_type_to_string(r#type: &PluginType) -> String {
+    serde_json::to_string(r#type).unwrap().replace("\"", "")
+}
+
 pub(crate) fn call_plugin<I, O>(
     input: I,
-    name: &str,
+    r#type: &PluginType,
```
andreievg, line 40: I like this change even though it's a little bit out of scope, it avoids having to hard code string and makes PluginType source of truth

server/service/src/backend_plugin/boajs/utils.rs
```diff
@@ -19,13 +20,11 @@ pub(super) fn get_string_argument(args: &[JsValue], index: usize) -> Result<Stri
     use GetStringArgumentError as Error;
 
     let closure = move || -> Result<String, GetStringArgumentError> {
-        let Some(arg) = args.get(index) else {
-            return Err(Error::NoArgumentAtIndex(index));
-        };
+        let arg = args.get(index).ok_or(Error::NoArgumentAtIndex(index))?;
```
andreievg, line 23: sorry again refactor, i added new method below which uses ok_or so thought to keep consistent, subjectively no difference in readability ¯\_(ツ)_/¯

server/service/src/backend_plugin/boajs/utils.rs
```diff
@@ -37,3 +36,31 @@ pub(super) fn std_error_to_js_error(error: impl StandardError) -> JsError {
...
+enum GetJsonArgumentError {
+    #[error("No argument at index {0}")]
+    NoArgumentAtIndex(usize),
+    #[error(transparent)]
+    JsError(#[from] JsError),
+    #[error(transparent)]
+    SerdeError(#[from] serde_json::Error),
+}
+
+pub(super) fn get_serde_argument<D: DeserializeOwned>(
```
andreievg, line 50: this is used to get of plugin data filter

server/service/src/backend_plugin/types/amc.rs
```diff
@@ -30,6 +31,10 @@ pub trait Trait: Send + Sync {
 // TODO as macro ? Can do types here too
 impl self::Trait for PluginInstance {
     fn call(&self, input: Input) -> PluginResult<Output> {
-        Ok(call_plugin(input, "average_monthly_consumption", self)?)
```
andreievg, line 33: This shows why PluginType is used now, rather then a string

server/service/src/requisition/request_requisition/add_from_master_list.rs
```diff
@@ -127,6 +129,7 @@ fn generate(
         store_id,
         &requisition_row,
         items_ids_not_in_requisition,
+        HashMap::new(),
```
andreievg, line 132: this is the change for item to new line id mapping, see generate_requisition_lines

server/service/src/backend_plugin/boajs/call_plugin.rs
```diff
@@ -53,28 +60,29 @@ where
...
     let namespace = module.namespace(context);
     let plugins = namespace
         .get(js_string!("plugins"), context)?
         .as_object()
         .cloned()
-        .ok_or_else(|| Error::PluginNamespaceMissing(name.to_string()))?;
+        .ok_or_else(|| Error::PluginNamespaceMissing(r#type.clone()))?;
 
-    let key = js_string!(name);
+    let key = js_string!(r#type.as_str());
```
lache-melvin, line 80: How come this one not using `plugin_type_to_string`?

server/service/src/backend_plugin/boajs/methods/get_plugin_data.rs
```diff
@@ -0,0 +1,35 @@
...
+use repository::{PluginDataFilter, PluginDataRepository, PluginDataRow};
+
+use crate::backend_plugin::{boajs::utils::*, plugin_provider::PluginContext};
+
+pub(crate) fn bind_method(context: &mut Context) -> Result<(), JsError> {
+    context.register_global_callable(
+        JsString::from("get_plugin_data"),
+        0,
+        NativeFunction::from_copy_closure(move |_, args, mut ctx| {
+            // TODO Is this actually safe ? (need to check reference counts after plugin has run)
```
lache-melvin, line 11: Would like to know about this before we merge..?

server/service/src/backend_plugin/boajs/methods/get_plugin_data.rs
```diff
@@ -0,0 +1,35 @@
...
+
+            let filter: PluginDataFilter = get_serde_argument(&mut ctx, args, 0)?;
+
+            let connection = service_provider
+                .connection()
+                .map_err(std_error_to_js_error)?;
+
+            // TODO pagination or restrictions ?
+            let plugin_data: Vec<PluginDataRow> = PluginDataRepository::new(&connection)
+                .query_by_filter(filter)
```
lache-melvin, line 22: Does feel pretty loose, had same thought with `transform_lines` as plugin... would like to hear more about how much we intend to trust the plugins?

server/service/src/requisition/request_requisition/generate.rs
```diff
@@ -46,6 +49,8 @@ pub fn generate_requisition_lines(
     store_id: &str,
     requisition_row: &RequisitionRow,
     item_ids: Vec<String>,
+    // Need for plugins since sometimes we set id passed in as parameter in graphql
```
lache-melvin, line 52: Well, the `insert_request_requisition_line` that does this is only used from batch mutation, which is not used by our frontend? Do you know if there is an integration that we need to support here, or should we remove the batch mutation and then wouldn't need to do this?

lache-melvin:
Looking good, would like to know about the todos you left, will test properly in the morning 🚀

server/service/src/requisition/request_requisition/generate.rs
lache-melvin: I TAKE IT BACK IM BAD AT SEARCHING IT IS INDEED USED

lache-melvin (changes requested):
Requesting changes at least until hear from you what we do now vs added to carryover:

- Currently `transform_request_requisition_lines` is not called by `update_request_requisition_line` but it is by `update_request_requisition` - should be added to line update too I think
  - If plugin interface/capability is as generic as "transform" then I think should be called in all cases
  - As discussed, could pass an ActionType enum into plugin, for plugin to determine if runs on all calls, only create etc.
  - Other places currently not using plugin that probably should in this case: R&R form, processors, fine grained calls like `use_suggested_quantity` (+ maybe more?? see below, I'm worried about this, how do we ensure we don't miss any?)
- Backend plugin helper methods have: ` // TODO Is this actually safe ? (need to check reference counts after plugin has run)` ... can we remove this if you are happy it is safe?
- Back and frontend plugins should have same code in package.json
  - I wonder if common frontend should expose `usePluginData` hook rather than implementing for each plugin
    - This can extract pluginCode from the plugin itself?
  - Query and mutate should only be able to access plugin data for that plugin code
    - restrict `get_plugin_data` 
    - Maybe `call_plugin` method supports returning of plugin_data rows, can then get processed and plugin code applied, before returned to service call for saving? Though probs requires a second kind of `call_plugin_with_data_output` or something....
- For thought - best practices to ensure plugin calls aren't missed in future core dev? Should live in mod file, as generic place as possible... are there better patterns to support the plugin architecture?

FYI at one point you mentioned update would be creating duplicate plugin data, but looks like backend plugin that isn't happening anymore. I wonder about frontend though, whether we should expose upsert plugin data mutation rather than 2 separately

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
```diff
@@ -213,6 +218,10 @@ export const RequestLineEdit = ({
                 label={t('label.amc')}
                 sx={{ marginBottom: 1 }}
               />
+              {line &&
+                plugins.requestRequisitionColumn?.editViewColumns?.map(
+                  (Column, index) => <Column key={index} line={line} />
```
lache-melvin, line 223: I think we should call this `Field`, and `editViewFields` ? `Column` has me thinking I should be defining a column definition with cell/accessor etc..

andreievg:
> I wonder if common frontend should expose usePluginData hook rather than implementing for each plugin
> This can extract pluginCode from the plugin itself?

Thanks, this was added to this one awhile ago: https://github.com/msupply-foundation/open-msupply/issues/2951#issuecomment-2650673745 (third bullet point)

> Query and mutate should only be able to access plugin data for that plugin code

I was thinking about this at the start, I am not sure if that's the case, at least `Query` should allow accessing data from other plugins, maybe mutate can be restrictive, but I don't think we should be doing this automatically, or at least I think it would be quite a bit more complicated (or would require a bit of a refactor in some areas). I thinks it's ok to leave some responsibility for this to plugin

> Backend plugin helper methods have:  // TODO Is this actually safe ? (need to check reference counts after plugin has run) > ... can we remove this if you are happy it is safe?

Doneski here: https://github.com/msupply-foundation/open-msupply/pull/6684/commits/b5483e59f16b07d6a5c504410f6b20a5cf23a2cd

> Currently transform_request_requisition_lines is not called by update_request_requisition_line but it is by update_request_requisition - should be added to line update too I think

I've tried to address this by doing with this commit: 08f318054560f22156a90b5bfdabdbdb81b44a05, can have another catch up and chat about this tomorrow

server/service/src/backend_plugin/boajs/call_plugin.rs
andreievg: it is actually, but naming not great: 70141b844125b750b113cc55d60bc406fb230a57

server/service/src/backend_plugin/boajs/methods/get_plugin_data.rs
andreievg: Please see other comments and this commit: [b5483e5](https://github.com/msupply-foundation/open-msupply/pull/6684/commits/b5483e59f16b07d6a5c504410f6b20a5cf23a2cd)

server/repository/src/db_diesel/backend_plugin_row.rs
```diff
@@ -201,7 +201,10 @@ mod test {
         let repo = BackendPluginRowRepository::new(&connection);
         let id = "backend_plugin_row".to_string();
 
-        let types = PluginTypes(vec![PluginType::Amc, PluginType::Amc]);
+        let types = PluginTypes(vec![
+            PluginType::AverageMonthlyConsumption,
+            PluginType::AverageMonthlyConsumption,
```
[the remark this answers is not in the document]
andreievg: 77179aba06ef6c097315fb97ac876e57e19d4693

server/service/src/backend_plugin/boajs/methods/get_plugin_data.rs
andreievg: As per our convo, they would need to be signed, or have a warning, either way a review for our devs

server/service/src/requisition/request_requisition/generate.rs
andreievg: 🥇

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
andreievg: Change here, sorry lots of other changes https://github.com/msupply-foundation/open-msupply/commit/08f318054560f22156a90b5bfdabdbdb81b44a05

lache-melvin:
From the example repo, could we add to carry over:

Frontend plugin - there is possibility to `mountEvent` but we don't pass in any kind of event type. Core might eventually dispatch multiple kinds of events to plugins (edit, save, blur etc) - plugin event handler should receive event type so they can handle accordingly

lache-melvin (commented):
Looking great thanks for changes Andrei, just waiting on shuffle of that frontend plugin interface - guess this would require change of client plugin too...

Also would be good to add that comment that transform requisition line contexts/actions are not exhaustive 🙏

lache-melvin (approved):
ACTUALLY going to approve this one, functioning well as is - requested changes are all naming and dev experience improvements. I think this can be separate PR, and seeing as I have review context here, I think should merge this, someone else can review next PR while I'm away 🙏 

THOUGH can there plz be a near term carry over - if can't get to supporting multiple backend plugins of same type too soon - to at least warn when uploading a second plugin of same type? I couldn't work out why I wasn't seeing my plugin results, thought was caching issue for a bit

andreievg:
Thanks so much for review @lache-melvin, will make an issue and follow PR shortly.

> ACTUALLY going to approve this one, functioning well as is - requested changes are all naming and dev experience improvements. I think this can be separate PR, and seeing as I have review context here, I think should merge this, someone else can review next PR while I'm away 🙏

> THOUGH can there plz be a near term carry over - if can't get to supporting multiple backend plugins of same type too soon - to at least warn when uploading a second plugin of same type? I couldn't work out why I wasn't seeing my plugin results, thought was caching issue for a bit

Issue #6624, opened 2025-02-24
Refactor Prescriptions' Detail View and Their Queries to Avoid Disabling Queries


zachariah-at-msupply:
## The suggestion

1. Call usePrescription in the detail view, and pass it to the Side Panel. _This avoids calling it multiple times in different parts of the detail view._
2. Disable the 'more' button that launches the sidepanel, and do not render the side panel, until the data has been received from usePrescription. _This is necessary, as otherwise the side panel will generate type errors (if the data is not available when it's rendered) or it will obscure part of the screen with no way to remove it (if the more button is pressed without rendering the sidepanel)._
3. Refactor the queries so that usePrescription returns all the insurance data necessary for the sidepanel. _This is better than having a separate query specially for this information (which we have currently, it is called insuranceById and it is found in client/packages/invoices/src/Prescriptions/api/operations.graphql), because it queries for such a small amount of information, and this information is used throughout the prescription detail view._

## Example use case

N/A

### Why should we invest time in this?

The current design is such that the detail view calls usePrescription, then the sidepanel, when launched re-calls usePrescription.
Based on the output of usePrescription, it also calls useInsurance.
This means that, to cope with the possibility that usePrescription has not provided the necessary data, useInsurance has pass the 'enabled' property to its query, which is best avoided:

<img width="497" alt="Image" src="https://github.com/user-attachments/assets/2c62030b-156a-4a32-9b27-6f38de43796a" />

Whereas the best practice is not to even render the sidebar until we have the data from usePrescription, and to make this query include insurance, which means the sidepanel won't even need to have any queries in it and we don't need to use the enabled property in useInsurance.

### Are there any risks associated with this change?

Updating the query to provide these fields in usePrescription might involve a *lot* of changes to the server side. That could add a day or two of extra work.

### How much effort is required?

This change is pretty easy, it should take half a day, add one or two days if updating the server side is a big challenge.

alainsussol:
Already have been tested as part of CIV carryover 2.6.0 release testing feature

PR #6727, opened 2025-02-27
6719 insurance loader

### LINKED ISSUE #6719: Insurance Loader on Invoice
We need to get the invoice related insurance policies in a report, to do that we need to have access to this data.
We can't add a second query based on the results of a first one, so I think loader or SQL is the only option. 
I think the loader is appropriate.

jmbrunskill:
Fixes #6719 #6624

# 👩🏻‍💻 What does this PR do?

Adds a loader for insurance, which will be used for the CIV prescription receipt.

It also updates the side panel to use this loader (avoiding the need to re-query), removing `useInsurance`
And renames `insurances` to `insurancePolicies` as I think that is a better name

Some side quests...

- [ ] Rename Insurance Scheme to Insurance Policy
- [ ] Make New Insurance Policies Active by default

I wasn't sure how far to go with the re-naming, for example `InsertInsuranceInput` could be renamed. Some backend stuff is still called insurnances, but I think this is enough for now?

# 🧪 Testing

- [ ] Check you can create and edit insurance policies
- [ ] Check you can add insuranace policies to a prescription on Confirm
- [ ] Check insurance policies are active by default

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**
- [X] Technically some Breaking Changes but not expected to impact any integrations
> Only affects the RC releases

**Issue Review**
The refactor issue  #6624 also talks about passing prescription into the side panel rather then re-querying. Would could keep it alive for that part if needed.

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
```diff
@@ -133,7 +133,7 @@ export const PaymentsModal: FC<PaymentsModalProps> = ({
           </Grid>
           <Grid size={{ xs: 12, sm: 6 }}>
             <InputWithLabelRow
-              label={t('label.insurance-scheme')}
+              label={t('label.insurance-policy')}
```
jmbrunskill, line 136: Side Quest

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PricingSection.tsx
```diff
@@ -8,57 +8,56 @@ import {
...
-          <PanelLabel>{t('label.insurance-discount-percentage')}</PanelLabel>
-          <PanelField>{insuranceData.discountPercentage}% </PanelField>
-        </PanelRow>
-        </>
+            <PanelRow>
+              <PanelLabel>
+                {t('label.insurance-discount-percentage')}
+              </PanelLabel>
+              <PanelField>
+                {prescriptionData.insuranceDiscountPercentage ?? 0}%{' '}
```
jmbrunskill, line 57: This is actually a bug fix, using the prescription data which should have been used before

client/packages/invoices/src/Prescriptions/api/operations.graphql
```diff
@@ -79,9 +79,15 @@ fragment PrescriptionRow on InvoiceNode {
...
 
-  nameInsuranceJoinId
   insuranceDiscountAmount
   insuranceDiscountPercentage
+  insurancePolicy {
+    insuranceProviders {
+      providerName
+    }
+    policyNumber
+    policyType
```
jmbrunskill, line 89: Realised I'm not using this, maybe we shoudl either display it or remove?

client/packages/system/src/Patient/apiModern/hooks/useInsurancesPolicies.ts
```diff
@@ -22,13 +22,13 @@ const defaultDraftInsurance: DraftInsurance = {
   policyNumberPerson: '',
   insuranceProviderId: '',
   policyType: '' as InsurancePolicyNodeType,
-  isActive: false,
+  isActive: true,
```
jmbrunskill, line 25: Side Quest

server/graphql/general/src/queries/insurance_policies.rs
```diff
@@ -30,72 +26,21 @@ pub struct InsuranceSortInput {
     desc: Option<bool>,
 }
 
-#[derive(PartialEq, Debug)]
```
jmbrunskill, line 33: Had to move this into the shared types crate

server/graphql/general/src/types.rs
```diff
@@ -216,39 +187,6 @@ pub enum InsuranceProviderResponse {
     Response(InsuranceProviderConnector),
 }
 
-#[Object]
-impl InsuranceProviderNode {
```
jmbrunskill, line 220: This needs to move to as a consequence

client/packages/invoices/src/Prescriptions/api/operations.graphql
GeronimoJohn: This is just referring to `policyType` right? Yea noticed it's not being used as well, wondering if we can just drop it?

GeronimoJohn (approved):
Thanks for this James! Have tested the components where the changes applies. Works well! 🔥 

Just left some comments but just a bunch of NITs

client/packages/invoices/src/Prescriptions/api/operations.graphql
jmbrunskill: Yes Policy Type was what I was thinking of

jmbrunskill:
> Don't forget to change to `v2.6.0-RC` as the head please

Thanks! I just noticed this too

GeronimoJohn:
Thanks for the changes @jmbrunskill, just have to rerun `yarn generate` as we've removed `policyType`

PR #6741, opened 2025-02-27
6733 changing insurance doesnt correctly change

### LINKED ISSUE #6733: Changing insurance doesn't correctly change %
## What went wrong? 😲

If you have two insurances from the same provider which give different percentage discounts.
The modal doesn't update the % correctly.

## Expected behaviour 🤔

Should show the correct % and calculation

## How to Reproduce 🔨

1. Setup 2 valid insurances for the a patient with the same provider but different discounts
2. Create a prescription for that patient with a cost
3. Confirm prescription
4. Change between the different policies, make sure they price updates correctly

## Your environment 🌱

- Open mSupply Version: V2.6.0-RC4

- Platform:

- Database type:

jmbrunskill:
Fixes #6733

# 👩🏻‍💻 What does this PR do?

Insurance discounts are now applied properly when you have two policies from the same provider

# 🧪 Testing

- [ ] Setup 2 valid insurances for the a patient with the same provider but different discounts
- [ ] Create a prescription for that patient with a cost
- [ ] Confirm prescription
- [ ] Change between the different policies, make sure they price updates correctly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [X] Frontend


client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
```diff
@@ -54,15 +52,19 @@ export const PaymentsModal: FC<PaymentsModalProps> = ({
...
   const { plugins } = usePluginProvider();
   const pluginEvents = usePluginEvents({
     isDirty: false,
   });
 
+  const totalAfterTax = prescriptionData?.pricing.totalAfterTax ?? 0;
+  const discountPercentage = selectedInsurance?.discountPercentage ?? 0;
+  const totalToBePaidByInsurance =
+    totalAfterTax * ((selectedInsurance?.discountPercentage ?? 0) / 100);
+  const totalToBePaidByPatient = totalAfterTax - totalToBePaidByInsurance;
```
CarlosNZ, line 66: The value `totalToBePaidByPatient` is *only* used by the plugin, so wouldn't it be better to calculate this value inside the plugin?

client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
```diff
@@ -172,7 +164,7 @@ export const PaymentsModal: FC<PaymentsModalProps> = ({
               label={t('label.discount-rate')}
               Input={
                 <BasicTextInput
-                  value={`${discountRate}%`}
+                  value={`${discountPercentage}%`}
```
CarlosNZ, line 167: This is pretty nitpicky, but I don't think it should show the "%" symbol twice. I.e. should either be "Discount: 10%", or "Discount(%): 10".
Screenshot 2025-02-28 at 10 49 27 AM

CarlosNZ (changes requested):
Looking good thanks, much cleaner without the unnecessary `useEffect` and `useState` value! 

Apart from the comments inline, a couple of other things:

1. We can't currently clear the "Insurance" selection after selecting? Should we be able to? I'd have thought that if they can choose an insurance, they could also choose not to use any insurance and just pay the full price. Is that correct?
2. This is not related to what you've done, but I just noticed the inputs have used a very odd way of making them act disabled (via the CSS `pointerEvents: 'none'` property it seems). Not sure why it was done like this, but can we just get rid of these and pass the `disabled` prop to them all?

client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
jmbrunskill: I wasn't sure what else might be affecting the price, so I'd rather the App told the plugin what it thought the correct amount is, and then the plugin could do something different with it as needed?
Feels like the right thing to me...

client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
jmbrunskill: Yeah I'm with you...

jmbrunskill: [e61ad1f](https://github.com/msupply-foundation/open-msupply/pull/6741/commits/e61ad1f85e70a1ee2d8c4bfd62bbe12d650b56f7)

jmbrunskill:
@CarlosNZ 
Clearable insurance [f338202](https://github.com/msupply-foundation/open-msupply/pull/6741/commits/f338202396a7e5c8b7ccbcddce4cebcb8b30d54d) - that had been bothering me too
Disabled Inputs [44c4317](https://github.com/msupply-foundation/open-msupply/pull/6741/commits/44c4317fa8d1a4d082c643a2b8f13a2520f48244) - not sure how we ended up with that CSS thing! I'd happily copy and pasted it into my plugin though. I think this is better.

client/packages/invoices/src/Prescriptions/DetailView/Payments/PaymentsModal.tsx
```diff
@@ -151,6 +151,8 @@ export const PaymentsModal: FC<PaymentsModalProps> = ({
                   onChange={(_, option) => {
                     if (option) {
                       setInsuranceId(option.value);
+                    } else {
+                      setInsuranceId(null);
                     }
```
CarlosNZ, line 156: Personal preference, but for these short condition branches, I like the simpler:
```ts
if (option) setInsuranceId(option.value)
else setInsuranceId(null);
```

Up to you 🤷‍♂️

CarlosNZ (approved):
Thanks @jmbrunskill -- working nicely now :)

PR #6750, opened 2025-02-27
Disable asset type when creating non-catalogue asset

### LINKED ISSUE #6227: Type field selector shows empty when importing cold chain equipment
## What went wrong? 😲

Type field selector shows empty when importing cold chain equipment!

I assume that's because Category (the field before) is not selected, but then if you switch back to `Use Catalogue Assets`, Catalogue Item fields show even if Category is not selected

https://github.com/user-attachments/assets/ce1aafbb-185c-443a-9fea-35a7fe750c43

## Expected behaviour 🤔

Would be good if both followed the same behaviour - if the field above is not selected, the next one cannot be selected (maybe greyed out instead of showing an empty line like the window is not fully loaded)? Or both fields can be selected even if the previous one is empty!

## How to Reproduce 🔨

2. Click New Asset -> see the edit modal
3. Click on Catalogue Item without selecting Category -> see you are given options
4. Switch to not use the catalogue assets -> click on Type -> see an empty rectangle

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC5

- Database type: PostgreSQL

mark-prins:
Fixes #6227 

# 👩🏻‍💻 What does this PR do?

I tested the idea when looking at the issue and this fixes what I think the problem is, so have just put the PR in. It's a very low priority, just trying to tick off an issue.

The `Type` input is disabled if you are adding a non-catalogue asset and have not yet selected a category.

# 🧪 Testing

- [ ] Click 'New Asset', switch the 'Use catalogue asset' off. Category should be empty and Type should be disabled 
- [ ] Select a Category, Type should be enabled

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/coldchain/src/Equipment/ListView/CreateAssetModal.tsx
+++ b/client/packages/coldchain/src/Equipment/ListView/CreateAssetModal.tsx
@@ -280,6 +280,7 @@ export const CreateAssetModal = ({
                     });
                   }}
                   value={draft.typeId}
+                  disabled={!draft.categoryId}
                 />
               }
             />


roxy-dao (approved):
LGTM~

PR #6752, opened 2025-02-28
2897 navigate to returns

### LINKED ISSUE #6481: I am not notified when a supplier return is created (and it's only in nw status so never gets finalised...)
## What went wrong? 😲

If I return items from an inbound shipment using the new footer UI, I do not get an alert letting me know that it has been created - I just have to assume it has been! 

Then, since I didn't get an alert, I didn't notice that it was created on new status and therefore hasn't actually been sent to the supplier yet...it's 
not intuitive to me that I would need to take further action here. 

I've upped the priority on this a little as I feel like it will be a common support ask. 

## How to Reproduce 🔨

1. Create an inbound shipment 
2. Select some lines and choose to return them (ps the scroll bar is not visible here once the modal shows up (and the input field is at the very right) so it's hard to know what action to take...) 
3. No alert! 

## Your environment 🌱

- Open mSupply Version: 2.5 
- Platform: demo-oms

fergie-nz:
Fixes #6481 

# 👩🏻‍💻 What does this PR do?

Adds navigate to supplier and customer returns when they are created from inbound and outbound returns respectively

# 🧪 Testing

- [ ] Create an inbound shipment
- [ ] Select some lines and choose to return them
- [ ] click 'ok' on the modal
- [ ] See navigated to Supplier Returns
- [ ] Confirm the same works with Outbound shipment and Customer Returns

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/Returns/api/operations.generated.ts
```diff
@@ -543,6 +544,7 @@ export const UpdateCustomerReturnLinesDocument = gql`
     ... on InvoiceNode {
       __typename
       id
+      invoiceNumber
```
fergie-nz, line 547: Needed to update api to find invoice number when updating lines

client/packages/invoices/src/Returns/modals/CustomerReturn/useDraftCustomerReturnLines.ts
```diff
@@ -114,19 +114,17 @@ export const useDraftCustomerReturnLines = ({
...
+    const returnOp = returnId ? await updateLines({
+      customerReturnId: returnId,
+      customerReturnLines,
+    }) : await insert({
+      id: FnUtils.generateUUID(),
+      customerId,
+      outboundShipmentId,
+      customerReturnLines,
+    });
+    return typeof returnOp === "number" ? returnOp : returnOp.invoiceNumber;
```
fergie-nz, line 126: awaiting update / insert and then converting to always return a number to be consumer focussed.

roxy-dao:
Hi Ferg, just wondering if you have had the chance to read the latest comments on the issue?

fergie-nz:
Closing - different solution needed.

Issue #5257, opened 2024-10-30
Better error handling for upserting standard reports


fergie-nz:
## Is your feature request related to a problem? Please describe 👀

Upserting standard reports gives unclear errors.

Better error handling would help diagnosing any issues that arise.

Including but not limited to:
- which file or directory can't be found  (maybe also what dir was searched)

## Describe the solution you'd like 🎁

Not using anyhow error on load_reports.

Could add error enum or at very least use .expect()

### Additional context 💌

Part of #4668 yet untriaged

### Moneyworks Jobcode 🧰

OMS:REPT

lache-melvin:
Related #5135 ?

PR #6757, opened 2025-02-28
6527 change test config default search location

### LINKED ISSUE #6527: Test config file not in location show report command is expecting
## What went wrong? 😲

test-config.json file used in the `show-report` command was moved in [this pr](https://github.com/msupply-foundation/open-msupply/pull/6432/files#diff-4fc153f7f0d72d0e3421800bcbd97acef0ba4d3bd32a716a756b2b6c2d67869c)

## Expected behaviour 🤔

show-report should work without manually passing the config file location

Should also show more descriptive errors on which file/dir cannot be located (ie specifically the test-config file). Otherwise users may assume that it is the report dir which cannot be found.

## How to Reproduce 🔨

1. run `cargo build && ./target/debug/remote_server_cli show-report --path ~"/Documents/Github/open-msupply-reports/clients/Ivory-Coast/stock_levels/latest"`
2. See error of no file found

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #6527 

# 👩🏻‍💻 What does this PR do?

Fixes `show-report` cli command where test-config file was moved in restructure.
Now looks in the correct place.

# 🧪 Testing

You can test this report by trying to show a report with the show command using default test-config file location :

For example:

`cargo build && ./target/debug/remote_server_cli show-report --path ~"/Documents/GitHub/open-msupply-reports/clients/Ivory-Coast/satisfaction_per_program/2_6_0"` 

Where the relative location of the report repo may differ.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/cli/src/cli.rs
+++ b/server/cli/src/cli.rs
@@ -582,7 +582,7 @@ async fn main() -> anyhow::Result<()> {
             let test_config_path = if let Some(config) = config {
                 config
             } else {
-                Path::new("reports").to_path_buf()
+                Path::new("../standard_reports").to_path_buf()
             };
 
             let test_config_file = fs::File::open(test_config_path.join("test-config.json"))


jmbrunskill (approved):
Works for me!

fergie-nz:
> I think it's probably ok but do we need a follow up issue for this?

Thanks James.

Yes good point I didn't add that in.

I've added it to [this issue](https://github.com/msupply-foundation/open-msupply/issues/5257) because I think it falls under the umbrella of other error handling which would be helpful in OMS.

Also I think this one is much less likely to be an annoying blocker in report dev work now that there is a default test.config location, and the user can A B test which dir is missing.

Previously the error received was that the dir couldn't be found, and it was easy to assume this was the report dir which was the argument being passed in (where in actual fact it was the test-config dir). Now that the test-config default is working, there is less ambiguity.

PR #6761, opened 2025-03-02
Dispensing keyboard usability

### LINKED ISSUE #6755: Dispensing usability
A request for usability features in the dispensing UI after a review with Craig

- [x] After clicking 'new prescription' the Patient input should have focus
- [x] Add a hot key for adding a new item (ALT+N the current CTRL+N / CMD+N used in OG will open a new browser window when using a web browser)
- [x] The list of batches should be collapsed by default, so that you see the `Issue` input
- [x] and can then tab to read abbreviation next (currently you have to tab through all of the batch 'units issued' inputs
- [x] change the wording of `You have issued a non-whole number of packs. Are you sure you are able to do this? If not, you will need to issue {{nearestAbove}} units to issue in whole packs.` to `You will have to break a pack in order to issue this number of units. Are you sure you are able to do this? If not, you will need to issue {{nearestAbove}} units to issue in whole packs.`

mark-prins:
Fixes #6755

# 👩🏻‍💻 What does this PR do?
Adds in the keyboard shortcuts and other usability concerns raised on Friday after a review with Craig & Mariya.

* There's a keyboard shortcut (Alt+N) to add a new item. This is implemented as a base button component, so the same button and shortcut can be used for other areas that want a shortcut for the 'new thing' button
* A shortcut (Alt+S) to save
* And a shortcut (Escape) for the cancel button
* Hide the batches by default - these are now in an accordion
* Tab takes you from the number input 
* Focus the patient input after clicking 'new prescription'
* update wording

Screenshot 2025-03-02 at 9 21 43 AM

I've also used the new keyboard shortcut for the scanner shortcut as that was using the same functionality. I've now consolidated into a shared hook

# 🧪 Testing

- [ ] After clicking 'new prescription' the Patient input should have focus
- [ ] Hot key (Alt+N) to add a new item to a prescription
- [ ] The list of batches should be collapsed by default
- [ ] But you can expand and see the batches
- [ ] Pressing TAB from the quantity input takes you to the abbreviation input
- [ ] change the wording to `You will have to break a pack in order to issue this number of units. Are you sure you are able to do this? If not, you will need to issue {{nearestAbove}} units to issue in whole packs.` which is shown if you dispensing a non-whole page number
- [ ] Alt+S will save the prescription line edit
- [ ] Escape will close the prescription line edit view

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
```diff
@@ -136,6 +137,23 @@ export const DialogButton: React.FC<DialogButtonProps> = ({
...
+      case 'cancel':
+        return e.code === 'Escape';
+      default:
+        return false;
+    }
+  };
+
+  useKeyboardShortcut({
+    isKeyValid,
+    onKeyPressed: () => ref.current?.click(),
```
CarlosNZ, line 155: What's the advantage of simulating a click, as opposed to just calling the `onClick()` function here?

If it's just a typing thing (cos the signature of `onClick` takes an event parameter), I had a quick look around and I can't see any DialogButton implementations that actually use this, so we could just remove it or make it optional?

client/packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
```diff
@@ -7,6 +7,7 @@ import {
 } from '@mui/material';
 
 export type BasicTextInputProps = StandardTextFieldProps & {
+  inputReference?: React.RefObject<HTMLInputElement>;
```
CarlosNZ, line 10: I dunno about you, but I associate the term "ref" 100% with React refs, whereas the term "reference" can be much more broadly interpreted. Would it work to keep the prop name as `inputRef` so it's clearer for consumers n(and follows typical convention), and just rename it internally to avoid the name clash with the local `inputRef`?

Bit nitpicky, just a thought.

(Also, I just did a quick search and I couldn't find anywhere that this inputReference prop is actually being set -- am I missing one?)

client/packages/invoices/src/Prescriptions/DetailView/AppBarButton.tsx
```diff
@@ -60,10 +60,9 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   return (
     <AppBarButtonsPortal>
       <Grid container gap={1}>
-        <ButtonWithIcon
+        <AddButton
           disabled={isDisabled}
           label={t('button.add-item')}
-          Icon={<PlusCircleIcon />}
           onClick={() => onAddItem()}
```
CarlosNZ, line 66: Not related to this issue, but just noticed -- there's no need to make an anonymous function here, can just set `onClick={onAddItem}` -- not sure if it's worth changing 🤷‍♂️
 
(And the `onViewHistory()` just below too ;) )

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -352,6 +355,7 @@ export const PrescriptionLineEditForm: React.FC<
                       },
                     },
                   }}
+                  onBlur={() => abbreviationRef.current?.focus()}
```
CarlosNZ, line 358: I don't suppose there is a cleaner way to get "Tab" from the quantity input to focus here? Cos by doing it with `onBlur`, this steals focus regardless of whether you Tabbed or not -- if I intentionally click in the "Presribed Quantity" field (for example) cos I want to change it, I get pulled to the "Abbreviation" field which is not a great user experience.

CarlosNZ (commented):
Looks good, works mostly as intended. My comments are mostly just questions -- you probably had a reason for doing it the way you did. 

My main issue is with not following OS convention keyboard shortcuts for common commands, like "Save" and "New". Are we sure we want to start going down the path of breaking this? That was always one of the most frustrating things about using OG mSupply for me, I'd hate to end up in a similar place for this one.

client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
mark-prins: it's because if you do that, then the onClick becomes a captured enclosure - in the case of the save function, then it will capture the state when the dialog button is rendered, that is, it has no lines to save. 

It may be possible to resolve that and recreate the captured fn, but this is simpler. it has the side effect of removing the type inconsistency where the event parameter is required; adding that in needs a casting `as` or requires a big refactor to remove the parameter.

client/packages/common/src/hooks/useKeyboardShortcut.ts
```diff
@@ -0,0 +1,26 @@
...
+        onKeyPressed();
+      }
+    }
+
+    document.addEventListener('keydown', keyDownHandler);
+
+    return () => {
+      document.removeEventListener('keydown', keyDownHandler);
+    };
+  }, []);
```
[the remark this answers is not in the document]
mark-prins: > Does the command-K utility not support locally-scoped shortcut keys?

well.. it would be possible. You can take the `useKBar` hook and manipulate actions, injecting an action for the current environment; but that's not as clean as implementing a keyboard shortcut hook imho.

client/packages/invoices/src/Prescriptions/DetailView/AppBarButton.tsx
mark-prins: `onAddItem` you can, bc the onClick has no parameter
`onViewHistory` though you can't. That button's `onClick` is the standard and has an optional parameter which differs from the `onViewHistory` optional. you get that pattern throughout the codebase, which is why there are so many of these anonymous functions which look redundant.

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
mark-prins: thanks - that's a good point. Have changed to watch for a `Tab` keypress

client/packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
mark-prins: I suggest that you try your suggestion 😉 

You'll end up with problems in other components, as the type of the `inputRef` is incorrect. Giving the non-standard prop the non-standard name is the simple way to resolve the issue.

that was an initial attempt, and was later removed in the usage - so I've just taken it out of here now. still seems to work..

CarlosNZ:
> can you clarify? I can't see any comments detailing what isn't working, so I don't know what to correct.

Sorry, I should have been clearer about that -- the only thing that didn't "work" properly with respect to the intention was the focus pull into the Directions input, but that's fixed now, thanks 👍

> * your browser is using those conventions, if you press the meta key (CMD on mac) + N on a browser then it opens a new window. If we use the same keyboard shortcut then how does a user open a new window using the accepted keyboard conventions?
>
>  * The browser does not actually pass on those keyboard shortcuts. you'll see that the meta key is pressed, but you do not then get the `KeyN` because the new window is opened

Yeah, it's tricky, I've encountered this issue before. The "N" one is particularly irksome, but for "Save", you can easily intercept the "Meta" key for "Cmd-S" and it works fine, and I'd argue that no one needs their browser's internal "Save" dialog while using OMS (and they can use the menu in the rare occasion they might). (Lots of web apps override this one, including Google Drive apps)

But when you get to the Electron desktop app, it becomes harder to justify using non-standard keyboard commands, IMO. I realise that means more work, and then you've got the issue of inconsistency between browser and Desktop, but I think all these problems are resolvable with some thought and effort.

But that's more of a larger design discussion. I think we've kind of gone round in circles a bit with the keyboard shortcuts, moving away from the single-key commands -- for good reason, but I'm not sure that we've fully moved to a better place (yet). It'd be good to have a sense of the overall philosophy for what we're trying to do with them, and I'd appreciate being part of that discussion.

Obviously that's all beyond this issue. Though the one sticking point for me right now is having shortcuts that are both non-standard *and* not listed in the command palette. What is the intention for communicating this information to the user going forward?

mark-prins:
oki - will have a look and see if it can be integrated into KBar. I take your point on the Save shortcut - not totally convinced about CMD+N as I use that one. It is consistent. 

I was thinking to rely on docs for informing people. The CMD/CTRL + K isn't advertised, so that probably needs to be made more visible.

mark-prins:
I had a play around using KBar and `useRegisterActions` but this is less than ideal because:

* The $mod+N cannot be overridden in chrome (see [this issue](https://issues.chromium.org/issues/41081444) for more)
* Implementing $mod+S would then be inconsistent with whatever we do for 'Add item'
* If the Save button is disabled, or save is not implemented on a page, then the default browser behaviour would work, and if Save was enabled then the custom save would work. In other words our app would behave inconsistently
* Using KBar means that if you press the key combinations when in an input, the keyboard shortcuts are disabled and do not handle. This means that if you try to save when entering a quantity on a prescription, or typing in directions, Save won't work

The last item is [inherent to KBar](https://github.com/timc1/kbar/blob/main/src/utils.ts#L101) and we'd require something pretty funky to work around it

PR #6770, opened 2025-03-03
6656 prescription reference filter

### LINKED ISSUE #6656: Add Reference Filter for Prescriptions

In our client's specification, there is a request to filter by the prescriptions' references. Currently this is not available.

Add a filter to the prescriptions' list view.

Nothing else will do ❤ 

### Additional context 💌

Currently, we can filter prescriptions by name, from date, to date, status and invoice number, we just need to add a filter for reference.

### Moneyworks Jobcode 🧰

OMS:CIDV for absolutely sure this time.

zachariah-at-msupply:
Fixes #6656 

# 👩🏻‍💻 What does this PR do?

This PR adds a reference filter to the prescriptions' list view.

image

I made the reference filter to be shown in the toolbar by default. However, the toolbar is quite full, so I wondered if it might be better to move it to the filters dropdown menu. What do you reckon?

# 🧪 Testing

- [ ] In Open mSupply, go to Dispensary > Prescriptions.
- [ ] Add a prescription. Add something memorable and unique in the reference field.
- [ ] Go back to Dispensary > Prescriptions.
- [ ] Type something in to the 'Reference' filter input that matches none of the prescription's references.
- [ ] Check no references are shown.
- [ ] Type in the reference that you added to your prescription (which was supposed to be unique).
- [ ] Check that only that prescription is shown.
Thanks for testing!

# 📃 Documentation

- [x] **These areas should be updated or checked**: Prescriptions can now be filtered by reference.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


CarlosNZ:
> However, the toolbar is quite full, so I wondered if it might be better to move it to the filters dropdown menu. What do you reckon

Yeah, I reckon leave it out. I think 3 default filters should normally be the maximum.

client/packages/invoices/src/Prescriptions/ListView/ListView.tsx
```diff
@@ -46,6 +46,7 @@ const PrescriptionListViewComponent: FC = () => {
     initialSort: { key: 'prescriptionDatetime', dir: 'desc' },
     filters: [
       { key: 'otherPartyName' },
+      { key: 'theirReference', condition: 'equalTo' },
```
CarlosNZ, line 49: I was wondering why it wasn't working for me, and I think this is the problem -- it's only an `equalTo` filter, but I was typing only part of the reference expecting a "like" filter (like the "Comment" filter has).

I don't think `equalTo` is useful for this, no one would expect to have to type the entire reference exactly to get a match, and it should still show results when typing.

Unfortunately there is no `like` filter already defined for this one:
Screenshot 2025-03-03 at 3 39 48 PM

So we'll need to add that added in on the server I think. I think check with James whether he'd rather do this in a separate issue or as part of this one.

CarlosNZ (changes requested):
Looks good, see my comment about the filter type.

client/packages/invoices/src/Prescriptions/ListView/ListView.tsx
zachariah-at-msupply: Yeah, you're right, how odd, when I tested it before pushing the commit it did the job with just one character. Maybe I overlooked something whe nI tested it...

zachariah-at-msupply: I added it to this PR, so I'll re-request review.

client/packages/invoices/src/Prescriptions/ListView/Toolbar.tsx
```diff
@@ -29,6 +29,12 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
               urlParameter: 'otherPartyName',
               isDefault: true,
             },
+            {
+              type: 'text',
+              name: t('label.reference'),
+              urlParameter: 'theirReference',
+              isDefault: false,
+            },
```
CarlosNZ, line 37: I only have one small comment in that I would think that "Status" should go above "Reference" in the dropdown list, but it's not a biggie and someone else may think otherwise.

CarlosNZ (approved):
Looking good, thanks for the additional filter fix. I just made one little comment about the filter order in the drop-down, if you want to quickly squeeze it in before you merge.

PR #6776, opened 2025-03-03
6768 labels undefined clinician and trunc

### LINKED ISSUE #6768: Labels Clinician printing as "Undefined", also spacing
## What went wrong? 😲

If a clinician is undefined we should print nothing. Also it gets truncated regardless

Image
(Bottom right most is an OG label)

## Expected behaviour 🤔

If no clinician is on the prescription, print nothing. Also fix the spacing by putting clinician on next line.

## How to Reproduce 🔨

1. New prescription
2. No Clinician
3. Add lines
4. print labels
5. Clinician is included at bottom of label as "Undefined"
6. Also truncated as there is not enough room.

## Your environment 🌱

- Open mSupply Version: 2.6.0-rc5

Chris-Petty:
Fixes #6768 

# 👩🏻‍💻 What does this PR do?

- Only print clinician if it exists
- 3 lines for details, 5 line for item directions+warning
- Some layout tweaks

Gist of the layout as per rendering on Labelary previewer:
image

A crude example of what that might look like (just duplicated field content on FE):

# 🧪 Testing

- [ ] Print some labels with clinician. Looks good
- [ ] Print some labels without clinician. Looks good

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/print/label.rs
+++ b/server/service/src/print/label.rs
@@ -79,26 +79,26 @@ pub fn print_prescription_label(
                 ^CI28
 
                 ^A0,25
-                ^FO10,10
-                ^FB570,2,0,C
+                ^FO5,10
+                ^FB550,2,0,C
                 ^FD{item_details}\&^FS
 
                 ^FX Line
-                ^FO10,65
+                ^FO5,65
                 ^GB570,2,2^FS
 
                 ^A0,25
-                ^FO10,75
-                ^FB570,5,0,L
+                ^FO5,75
+                ^FB550,5,0,L
                 ^FD{item_directions}\&{warning}\&^FS
 
                 ^FX Line
-                ^FO10,210
+                ^FO5,210
                 ^GB570,2,2^FS
 
                 ^A0,20
-                ^FO10,220
-                ^FB570,3,0,C
+                ^FO5,220
+                ^FB550,3,0,C
                 ^FD{patient_details}\&{details}\&^FS
 
                 ^XZ
--- a/server/service/src/print/label.rs
+++ b/server/service/src/print/label.rs
@@ -89,16 +89,16 @@ pub fn print_prescription_label(
 
                 ^A0,25
                 ^FO10,75
-                ^FB570,6,0,L
+                ^FB570,5,0,L
                 ^FD{item_directions}\&{warning}\&^FS
 
                 ^FX Line
-                ^FO10,180
+                ^FO10,210
                 ^GB570,2,2^FS
 
                 ^A0,20
-                ^FO10,195
-                ^FB570,2,0,C
+                ^FO10,220
+                ^FB570,3,0,C
                 ^FD{patient_details}\&{details}\&^FS
 
                 ^XZ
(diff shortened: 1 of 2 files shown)


server/service/src/print/label.rs
```diff
@@ -79,28 +79,28 @@ pub fn print_prescription_label(
                 ^CI28
 
                 ^A0,25
-                ^FO10,10
-                ^FB570,2,0,C
-                ^FD{item_details}^FS
+                ^FO5,10
+                ^FB550,2,0,C
+                ^FD{item_details}\&^FS
```
Chris-Petty, line 84: Hot tip from Labelary.com - trailing new line makes alignment work better!

server/service/src/print/label.rs
```diff
@@ -79,28 +79,28 @@ pub fn print_prescription_label(
...
+                ^FO5,210
                 ^GB570,2,2^FS
 
-                ^A0,25
-                ^FO10,220
-                ^FD{patient_details}^FS
+                ^A0,20
+                ^FO5,220
+                ^FB550,3,0,C
+                ^FD{patient_details}\&{details}\&^FS
```
Chris-Petty, line 102: Main change here is putting patient details and details in a text box with word wrap rather than explicit lines.

aimee-mcneil-melville (changes requested):
Looking great, I like the smaller patient/clinician text and the alignment there! Can take a lot of info and still look neat!

Just with the main directions body, there is a tiiiny cutoff on the right hand side - if that can be realigned please
IMG_0193 Large

 I'm also wondering if it is preferable to cut off the directions at the bottom if it gets too long? Then at least five legible lines would be included, rather than four

Chris-Petty:
> Looking great, I like the smaller patient/clinician text and the alignment there! Can take a lot of info and still look neat!
> 
> Just with the main directions body, there is a tiiiny cutoff on the right hand side - if that can be realigned please !
> 

Haha it's so finicky to get text that exhibits this, but yes. I have printed so many labels now... after many tweaks... I concluded the printer was the issue. I've adjusted it now with new configuration sent from OG 😅 

image

width was 70, label left margin was 5. Just tweaked some things and printed (OG sends commands to configure printer on every print it would seem!). Looks better...



> I'm also wondering if it is preferable to cut off the directions at the bottom if it gets too long? Then at least five legible lines would be included, rather than four

Unfortunately that's how the ^FB field box works. There is a ^TB Text box, where you set a height in dots that would just chop excess off, but unfortunately it doesn't support any new line character it seems! So it gets janky with the warning... which we don't use so I'll just remove for now.... We lose the ability to centre align though so I didn't do this for the item details or bottom details.

Beyond that I'm putting that in the too hard for now basket. We'll redo the template here with Tera and do some extra logic in rust and Tera land to work out how things should line up, scale and spread things over multiple labels if too long etc (what a mess if they have multipart labels). Dare I say, it might even be a backend plugin for janky cases (6 part labels haha).

aimee-mcneil-melville (approved):
Beauty! 
I like that it will configure the printer each time - prevent it from mis-aligning 😄

Chris-Petty:
> Beauty! I like that it will configure the printer each time - prevent it from mis-aligning 😄

Haha yea. We'll need to do in OMS once we've finished the printer config work!

PR #6789, opened 2025-03-03
Add insert returns as shipped when inserting from shipments

### LINKED ISSUE #6481: I am not notified when a supplier return is created (and it's only in nw status so never gets finalised...)

fergie-nz:
Fixes #6481

# 👩🏻‍💻 What does this PR do?

Sets return status as 'shipped' by default when a return is created from a shipment.

This happens for both inbound and outbound shipments for supplier and customer returns respectively.

Follow up [issue](https://github.com/msupply-foundation/open-msupply/issues/6790) created for more fleshed out UX.

# 🧪 Testing

- [ ] Create an inbound shipment from store A - B (easy to test this if on the same site)
- [ ] Add an item to the shipment
- [ ] Set the shipment status as delivered
- [ ] Select the line for the shipment, and return the selected line using the footer ui
- [ ] Navigate to Supplier Returns, and see a return is created and that the status is 'Shipped'.
- [ ] Manually create a return, and see it is generated with status 'New'
- [ ] Log in to store B, and navigate to Customer Returns
- [ ] See a transfer return is generated. See that the status 'shipped'.

Then:

- [ ] Repeat the above steps with outbound shipments and customer returns.
>  Note - a customer return will not create a transfer supplier return, so skip this part. 
> Note - status of customer return generated from an outbound shipment will be 'Verified' rather than 'Shipped'

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/Returns/modals/CustomerReturn/useDraftCustomerReturnLines.ts
```diff
@@ -120,6 +121,7 @@ export const useDraftCustomerReturnLines = ({
         customerId,
         outboundShipmentId,
         customerReturnLines,
+        status: InsertCustomerReturnStatusInput.Shipped,
```
fergie-nz, line 124: This hook is only used when creating returns from shipments, so we can use shipped for this case.

server/graphql/invoice/src/mutations/customer_return/insert.rs
```diff
@@ -20,6 +21,13 @@ pub struct InsertInput {
     pub customer_id: String,
     pub outbound_shipment_id: Option<String>,
     pub customer_return_lines: Vec<CustomerReturnLineInput>,
+    pub status: Option<InsertCustomerReturnStatusInput>,
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum InsertCustomerReturnStatusInput {
+    New,
+    Shipped,
```
fergie-nz, line 30: Have not used the 'picked' status option for input. Maybe there is a future use case for this? I figure it is easy to add in future if need be.

server/service/src/vaccination/update/generate.rs
```diff
@@ -94,6 +94,7 @@ pub fn generate(
                     reason_id: None,
                     note: None,
                 }],
+                status: None,
```
fergie-nz, line 97: This None status will default to new, and doesn't change existing insert behaviour from vaccines.

client/packages/invoices/src/Returns/modals/CustomerReturn/useDraftCustomerReturnLines.ts
roxy-dao: This should be `Verified`

client/packages/invoices/src/Returns/modals/CustomerReturn/useDraftCustomerReturnLines.ts
```diff
@@ -120,6 +121,7 @@ export const useDraftCustomerReturnLines = ({
...
       });
     } else {
       await updateLines({
```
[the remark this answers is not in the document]
roxy-dao: Yes remove this since we are finalising customer returns straight away now

client/packages/invoices/src/Returns/modals/SupplierReturn/useDraftSupplierReturnLines.ts
```diff
@@ -63,6 +64,7 @@ export const useDraftSupplierReturnLines = ({
         supplierId,
         inboundShipmentId,
         supplierReturnLines,
+        status: InsertSupplierReturnStatusInput.Shipped,
       });
     } else {
       await updateLines({
```
roxy-dao, line 70: Same here, remove please since it won't be called

server/graphql/invoice/src/mutations/customer_return/insert.rs
```diff
@@ -165,3 +175,13 @@ impl CustomerReturnLineInput {
         }
     }
 }
+
+impl InsertCustomerReturnStatusInput {
+    pub fn to_domain(&self) -> InsertCustomerReturnStatus {
+        use InsertCustomerReturnStatus::*;
+        match self {
+            InsertCustomerReturnStatusInput::New => New,
+            InsertCustomerReturnStatusInput::Shipped => Shipped,
```
roxy-dao, line 184: This should be `Verified`, not `Shipped`

server/service/src/invoice/customer_return/insert/generate.rs
```diff
@@ -47,7 +54,7 @@ pub fn generate(
         name_store_id: other_party.store_id().map(|id| id.to_string()),
         store_id: store_id.to_string(),
         created_datetime: current_datetime,
-        status: InvoiceStatus::New,
+        status: status,
```
roxy-dao, line 57:
```suggestion
        status,
```

Redundant field name since the variable is already the same as the field name

server/service/src/invoice/customer_return/insert/mod.rs
```diff
@@ -18,13 +19,20 @@ use validate::validate;
 
 use super::CustomerReturnLineInput;
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum InsertCustomerReturnStatus {
+    New,
+    Shipped,
```
roxy-dao, line 25:
```suggestion
    Verified,
```

server/service/src/vaccination/update/generate.rs
roxy-dao: We would want to finalise this as well.

roxy-dao (changes requested):
Thanks for the changes Ferg. I just have a few comments below. 

- Customer return only has the following statuses: `New`, `Delivered` and `Verified`, so to change the status to finalised, it should be set to `Verified`
- Supplier returns is the one that finalises on `Shipped` status. 

Did you test your changes Ferg? Adding tests would also help validate your changes.
The validation on insert stock_out_line is not allow finalising of supplier returns, I imagine the same thing will happen with customer return once you change the status to `Verified`

server/graphql/invoice/src/mutations/customer_return/insert.rs
fergie-nz: Nice good catch thank you - I had copied the supplier statuses

roxy-dao (changes requested):
Thanks for the changes Ferg.

Have you tested your changes? 
There's still one more request up top that hasn't been implemented yet

> Adding tests would also help validate your changes.
> The validation on insert stock_out_line is not allow finalising of supplier returns, I imagine the same thing will happen with customer return once you change the status to Verified

fergie-nz:
> We would want to finalise this as well.

I think this is okay because the customer return from vaccination is immediately verified after creation.

server/service/src/vaccination/update/generate.rs
```diff
@@ -94,6 +94,7 @@ pub fn generate(
                     reason_id: None,
                     note: None,
                 }],
+                status: Some(InsertCustomerReturnStatus::New),
```
fergie-nz, line 97: Passing status as New here (it must be new to add lines)

This is subsequently updated to verified in a second DB call.

fergie-nz (commented):
Hi @roxy-dao , I've made a few changes to the front end as I realise the return modal was being used in multiple places.

Where returns are being made and their status is not new, two db calls are being made (and therefore two changelog rows). This is similar to how we are managing new return -> add lines -> set return to desired status.

Transfer returns are being updated to match the status of the generated return.

I've updated the PR test instructions to match these changes.

roxy-dao:
Hi Ferg~

Just to iterate over what was talked about:
We don't need to pass the status through from Frontend
We can conclude which returns have been created from an Inbound/Outbound shipment if it has an originalShipmentId.
We can then in the insert method, do a check for if this exists and if it does, finalise the return

client/packages/common/src/types/schema.ts
```diff
@@ -3498,6 +3498,7 @@ export type InvoiceFilterInput = {
   colour?: InputMaybe<EqualFilterStringInput>;
   comment?: InputMaybe<StringFilterInput>;
   createdDatetime?: InputMaybe<DatetimeFilterInput>;
+  createdOrBackdatedDatetime?: InputMaybe<DatetimeFilterInput>;
```
fergie-nz, line 3501: Unrelated to this PR - I think non generated schema was committed to rc branch

server/service/src/invoice/supplier_return/insert/mod.rs
```diff
@@ -90,6 +97,25 @@ pub fn insert_supplier_return(
...
+                update_supplier_return(
+                    ctx,
+                    UpdateSupplierReturn {
+                        supplier_return_id: supplier_return.id.clone(),
+                        comment: None,
+                        status: Some(UpdateSupplierReturnStatus::Shipped),
+                        colour: None,
+                        on_hold: None,
+                        their_reference: None,
+                        transport_reference: None,
```
roxy-dao, line 111:
```suggestion
                    UpdateSupplierReturn {
                        supplier_return_id: supplier_return.id.clone(),
                        status: Some(UpdateSupplierReturnStatus::Shipped),
                        ..Default::default()
```
Can default the None values instead of declaring each one

server/service/src/invoice/customer_return/insert/mod.rs
```diff
@@ -91,6 +98,25 @@ pub fn insert_customer_return(
...
+                let _ = update_customer_return(
+                    ctx,
+                    UpdateCustomerReturn {
+                        id: customer_return.id.clone(),
+                        comment: None,
+                        status: Some(UpdateCustomerReturnStatus::Verified),
+                        colour: None,
+                        on_hold: None,
+                        their_reference: None,
+                        other_party_id: None,
```
roxy-dao, line 112:
```suggestion
                        id: customer_return.id.clone(),
                        status: Some(UpdateCustomerReturnStatus::Verified),
                        ..Default::default()
```

roxy-dao (changes requested):
Thanks for the changes Ferg. I just have one more request. 

It would be nice if there was a toast to tell the user that a 'verified' or 'shipped' return has been created. Currently you press `Ok` and the user has no idea what has happened.

server/service/src/invoice/supplier_return/insert/mod.rs
```diff
@@ -50,6 +53,10 @@ pub enum InsertSupplierReturnError {
         line_id: String,
         error: UpdateLineReturnReasonError,
     },
+    ErrorSettingNonNewStatus {
+        update_error: UpdateSupplierReturnError,
+    },
+    ManuallyCreatedReturnMustHaveNewStatus,
```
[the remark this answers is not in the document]
fergie-nz: You're right! I added that earlier when giving the front end the capacity to choose status. Not needed now, thank you

client/packages/invoices/src/Returns/api/operations.graphql
```diff
@@ -414,6 +423,9 @@ mutation updateCustomerReturnLines(
     ... on InvoiceNode {
       __typename
       id
+      originalShipment {
+        id
+      }
```
fergie-nz, line 428: This adding to graphql (and for updateSupplierReturnLines) is for type compatibility.

I wonder if it is possible to not return this type and have type compatibility, but I couldn't see how to.

client/packages/invoices/src/Returns/api/api.ts
```diff
@@ -340,7 +340,7 @@ export const getReturnsQueries = (sdk: Sdk, storeId: string) => ({
     const { insertCustomerReturn } = result;
 
     if (insertCustomerReturn.__typename === 'InvoiceNode') {
-      return insertCustomerReturn.invoiceNumber;
+      return insertCustomerReturn;
```
fergie-nz, line 343: returning the full return so that navigation to return from AppBarButtons still works with invoice number

client/packages/invoices/src/Returns/modals/CustomerReturn/CustomerReturn.tsx
```diff
@@ -78,8 +80,9 @@ export const CustomerReturnEditModal = ({
 
   const onOk = async () => {
     try {
-      !isDisabled && (await save());
+      let customerReturn = !isDisabled && (await save());
       onCreate?.();
+      !!customerReturn && customerReturn?.originalShipment?.id && isNewReturn && success(t('messages.customer-return-created-verified'))();
```
fergie-nz, line 85: Type mapping was a little bit messy to make this work where it could be false | invoice node

roxy-dao: Do we need to check original shipment id and is new return? If api returns with a result (!!customerReturn) then it would prompt the message, correct?

server/graphql/invoice/src/mutations/supplier_return/insert.rs
```diff
@@ -100,7 +100,9 @@ fn map_error(error: ServiceError) -> Result<InsertErrorInterface> {
         | ServiceError::OriginalInvoiceNotAnInboundShipment
         | ServiceError::CannotReturnInboundShipment
         | ServiceError::InvoiceAlreadyExists
-        | ServiceError::OtherPartyDoesNotExist => BadUserInput(formatted_error),
+        | ServiceError::OtherPartyDoesNotExist
+        | ServiceError::ManuallyCreatedReturnMustHaveNewStatus
```
roxy-dao, line 104: This should be removed if not used in validation...

client/packages/invoices/src/Returns/modals/SupplierReturn/SupplierReturn.tsx
```diff
@@ -76,8 +78,9 @@ export const SupplierReturnEditModal = ({
 
   const onOk = async () => {
     try {
-      !isDisabled && (await save());
+      let supplierReturn = !isDisabled && (await save());
       onCreate?.();
+      !!supplierReturn && supplierReturn?.originalShipment?.id && isNewReturn && success(t('messages.supplier-return-created-shipped'))();
```
roxy-dao, line 83: See above comment

roxy-dao (changes requested):
Thanks for the changes Ferg. Just have a question...

server/graphql/invoice/src/mutations/supplier_return/insert.rs
fergie-nz: Thank you for catching this, I missed this.

client/packages/invoices/src/Returns/modals/SupplierReturn/SupplierReturn.tsx
fergie-nz: This one is needed because the same insert modal is used both from appbarbuttons in the return list views, and in shipment list views. And in both cases a return node is returned - so we need the additional check that it has an original shipment id.

roxy-dao (approved):
LGTM~ tysm for all the changes Ferg!

PR #6796, opened 2025-03-03
6746 Client Code Checks

### LINKED ISSUE #6746: Remove git commit hook for yarn compile, and replace with github action
As per the title :)

The `yarn compile` step in git commits takes a long time, and slows down developers
It only needs to be checked at PR time.

GeronimoJohn:
Fixes #6746 

# 👩🏻‍💻 What does this PR do?

We initially set up a Typescript compilation and linter hook (`pre-commit-lint`) that runs every time a developer makes a change within the client directory and then commit their changes. It's a very powerful tool but the TS compilation and linter takes a bit of time to run.

To avoid waiting. We've decided to create a Github Action that runs a linter and formatter against the client directory. To ensure that we're still following the code standards that was initially set. And remove husky or the `pre-commit-lint` hook from running.

Screenshot 2025-03-05 at 2 20 59 PM

ESLint had a major update about a few months ago - version 9.0.0 onwards. It's a breaking change and we are required to migrate our old configuration ESLint file to follow their new pattern. Not doing so will not make ESLint work. The interim solution is to downgrade to version 8.50.0. And are to work on the migration at a different time whenever time permits

# 🧪 Testing

DEVS ONLY

- [ ] This shouldn't run the pre-build-lint when doing a commit after a change within the client directory
- [ ] There should be a new Github Action called - `Client Code Checks`
   - [ ] It should pass if it doesn't see any changes that doesn't follow the rules set in our ESLint config
      - [ ] Update any file within the client directory and commit
      - [ ] Remove changes after 
   - [ ] It should fail if it does see any changes that doesn't follow the rules set in our ESLint config
   - [ ] Ensure that the logs returned by the Github action makes sense

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


GeronimoJohn:
**Note on Logs:**

You'll notice a lot of `warnings` in the logs. These are not ideal but currently acceptable. The majority of these warnings come from `useEffect` hooks, specifically their dependency arrays. `useEffect` and ESLint rules can be tricky, as dependency arrays often depend on the specific use case.  

If a `useEffect` is working as expected without the dependencies ESLint suggests, it's fine to ignore the warning using TypeScript's `ts-ignore`, e.g.  
`// eslint-disable-next-line react-hooks/exhaustive-deps`.

client/.husky/pre-commit
```diff
@@ -1,9 +1,13 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-SRC_PATTERN="client/"
+# The following lines have been commented out to disable automatic execution of the pre-commit linting and formatting tasks.
```
jmbrunskill, line 4: Can we just remove husky now? Is it needed still?

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
```diff
@@ -26,7 +26,7 @@ import {
   InventoryAdjustmentReasonRowFragment,
   InventoryAdjustmentReasonSearchInput,
   ItemVariantInputCell,
-  PackSizeEntryCell,
+  packSizeEntryCell,
```
CarlosNZ, line 29: This really should be `PackSizeEntryCell`, as the source is very much a component. I wonder if we just need to correctly type the original component in "PackSizeEntryCell.tsx" as React.FC etc...

client/packages/requisitions/src/RequestRequisition/DetailView/IndicatorEdit/IndicatorLineEdit.tsx
```diff
@@ -123,14 +124,18 @@ export const IndicatorLineEdit = ({
...
+            <InputWithLabel
+              key={column.value?.id}
+              data={column}
+              disabled={disabled}
+              autoFocus={i === 0}
+            />
+          ) : (
+            <></>
+          );
+        })}
```
CarlosNZ, line 138:
```suggestion
        {columns.map(
          (column, i) =>
            column.value != null && (
              <InputWithLabel
                key={column.value?.id}
                data={column}
                disabled={disabled}
                autoFocus={i === 0}
              />
            )
        )}
```
Bit cleaner?

client/packages/requisitions/src/RnRForms/api/hooks/useRnRFormContext.ts
```diff
@@ -18,6 +18,7 @@ export const useRnRFormContext = create<RnRFormContext>(set => ({
     })),
   clearDraftLine: id =>
     set(state => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
CarlosNZ, line 21: This is a bit odd, because we have a lint rule to ignore unused errors that are "_":
```
    '@typescript-eslint/no-unused-vars': [
      'error',
      {
        argsIgnorePattern: '^_',
        caughtErrorsIgnorePattern: '^_',
      },
    ],
```
Maybe have a quick look what's going on there? But probably don't spend too much time on it, this is fine otherwise.

CarlosNZ (commented):
Looking good, @GeronimoJohn, this is going to be a big help. Few little comments.

I think we should probably wait to merge this until after 2.6.0 has been merged into develop. (And then we can do all the re-formatting as well :)

client/.husky/pre-commit
GeronimoJohn: From what I can see, not needed anymore. I was going to suggest that we just remove it now

.github/workflows/client-code-checks.yaml
```diff
@@ -0,0 +1,30 @@
+on:
+  pull_request:
+    branches:
+      - main
+      - develop
```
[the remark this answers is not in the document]
GeronimoJohn: Ah that's a good point, I'll look into that

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
GeronimoJohn: That makes sense actually

GeronimoJohn: This proved to be a bit trickier than expected.

Here are a few things I tried:

1. Since `PackSizeEntryCell` is a HOC, the initial return is a function.
   - Making it start with a capital letter causes ESLint to give the error when we call it: `A function with a name starting with an uppercase letter should only be used as a constructor. eslint(new-cap)`
      - The solution is to disable the error with `// eslint-disable-next-line new-cap`
   - Additionally, since the actual component is in a callback, converting the first character to an uppercase gives another error: hooks cannot be called inside a callback.
   
2. Converting it to just be a normal React function component creates a type mismatch, as shown in the image below, which might require more work:
   - Screenshot 2025-03-06 at 11 50 44 AM

The solution I applied:
- Keep the first character of the component in uppercase.
- Disable the error with `// eslint-disable-next-line new-cap`.
- Convert the function called inside to a standalone component, which we then call.
  - If we stick to lowercase for the first character, we adhere to the rules of React for hooks in callbacks, which is weird to me lol.

GeronimoJohn: This proved to be a bit trickier than expected.

Here are a few things I tried:

1. Since `PackSizeEntryCell` is a HOC, the initial return is a function.
   - Making it start with a capital letter causes ESLint to give the error when we call it: `A function with a name starting with an uppercase letter should only be used as a constructor. eslint(new-cap)`
      - The solution is to disable the error with `// eslint-disable-next-line new-cap`
   - Additionally, since the actual component is in a callback, converting the first character to an uppercase gives another error: hooks cannot be called inside a callback.
   
2. Converting it to just be a normal React function component creates a type mismatch, as shown in the image below, which might require more work:
   - Screenshot 2025-03-06 at 11 50 44 AM

The solution I applied:
- Keep the first character of the component in uppercase.
- Disable the error with `// eslint-disable-next-line new-cap`.
- Convert the function called inside to a standalone component, which we then call.
  - If we stick to lowercase for the first character, we adhere to the rules of React for hooks in callbacks, which is weird to me lol.

client/.husky/pre-commit
GeronimoJohn: Have removed it. Please see changes here: https://github.com/msupply-foundation/open-msupply/pull/6796/commits/5969e2980af3f26e056f8004ef22771b4ec6ae7b

I've dropped these libraries
- Husky
- Lint-staged

client/packages/requisitions/src/RnRForms/api/hooks/useRnRFormContext.ts
```diff
@@ -18,7 +18,6 @@ export const useRnRFormContext = create<RnRFormContext>(set => ({
     })),
   clearDraftLine: id =>
     set(state => {
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const { [id]: _, ...draftLines } = state.draftLines;
```
GeronimoJohn: Updated the rule for it to support this case

```typescript
'@typescript-eslint/no-unused-vars': [
      'error',
      {
        argsIgnorePattern: '^_',
        caughtErrorsIgnorePattern: '^_',
        varsIgnorePattern: '^_',
      },
    ]
```

GeronimoJohn:
Thanks for the review Carl, please see changes here: https://github.com/msupply-foundation/open-msupply/pull/6796/commits/229e61ad545ba4b13a52129f21401e0273c38b55

.github/workflows/client-code-checks.yaml
```diff
@@ -0,0 +1,32 @@
...
+      # This pattern matches branches for release candidates (e.g., v2.6.0-RC)
+      - v*-RC
```
GeronimoJohn, line 7: Supports cases for RC branches. This workflow will run against any PRs that are open and targeting these branches.

As of the time this was written, the GitHub Action is failing because v2.6.0-RC has a console log.

.github/workflows/server-tests.yaml
```diff
@@ -3,8 +3,10 @@ on:
     branches:
       - main
       - develop
+      # This pattern matches branches for release candidates (e.g., v2.6.0-RC)
+      - v*-RC
```
GeronimoJohn, line 7: Side Quest: Enable CI or server tests to run against RC branches

.github/workflows/client-code-checks.yaml
andreievg: Why not on every PR ?

client/.eslintrc.js
```diff
@@ -40,6 +40,7 @@ module.exports = {
       {
         argsIgnorePattern: '^_',
         caughtErrorsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
```
andreievg, line 43: What does this do ?

andreievg:
@GeronimoJohn, in this PR: https://github.com/msupply-foundation/open-msupply/pull/6879, I've added re-compile, which i think should just replace the compile yarn script (it's 3 times as fast), I think it's worth doing that here, what you reckon ?

client/.eslintrc.js
andreievg: Never mind oop i should read all comments found above

.github/workflows/client-code-checks.yaml
GeronimoJohn: Thats a good question actually, it should just be applied to every PR. Will change soon

GeronimoJohn:
> @GeronimoJohn, in this PR: #6879, I've added re-compile, which i think should just replace the compile yarn script (it's 3 times as fast), I think it's worth doing that here, what you reckon ?

Sounds good to me! Also noticed in that PR that we're formatting the graphql generated files 🔥. Big fan

.github/workflows/server-tests.yaml
```diff
@@ -1,10 +1,8 @@
 on:
   pull_request:
-    branches:
-      - main
-      - develop
+    types: [opened, reopened, edited]
     paths:
-      - 'server/**'
+      - "server/**"
```
GeronimoJohn, line 5: Side Quest: As discussed on omSupply Devs, we want server tests to run on every PR. Added opened, reopened, and edited types to this action. This GH action runs whenever a PR affecting the server directory is opened, reopened, or edited.

.github/workflows/server-tests.yaml
```diff
@@ -1,10 +1,8 @@
 on:
   pull_request:
-    branches:
-      - main
-      - develop
+    types: [opened, reopened, edited, synchronize]
     paths:
-      - 'server/**'
+      - "server/**"
```
GeronimoJohn, line 5: Side Quest: Run this on every PR that is opened, reopened, edited, or synchronized. This ensures the action runs on all PR changes.

andreievg:
@GeronimoJohn after updating from develop could you please replace current 'compile' yarn script with 're-compile'

GeronimoJohn:
Thanks @andreievg for the review, 

Updated Github action `Client Code Checks` and `Server Tests` to run on every PR: https://github.com/msupply-foundation/open-msupply/pull/6796/commits/f2a86a36dd9d488c7b64bc4b64461522ee001536

Have also updated the yarn command for `lint-and-format` to use `re-compile`.

Noticed that some of our ColdChain queries are not following camelCase. Added a fix here: https://github.com/msupply-foundation/open-msupply/pull/6796/commits/23bb7ba41122e4c1159b9ab35fe46f0dc1682411

client/package.json
```diff
@@ -16,8 +16,6 @@
     "compile": "lerna run --scope @openmsupply-client/* --parallel tsc --since HEAD -- --noEmit",
     "re-compile": "tsc --project packages/host/tsconfig.json --noEmit",
     "compile-full": "lerna run --scope @openmsupply-client/* --parallel tsc",
```
CarlosNZ, line 18: Just wondering what we want "re-compile" for, as opposed to "compile-full"?

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -118,7 +113,7 @@ export const QuantityTableComponent: FC<TableProps> = ({
       },
     ],
     getColumnLookupWithOverrides('packSize', {
-      Cell: PackUnitEntryCell,
+      Cell: PackSizeEntryCell<DraftInboundLine>,
```
CarlosNZ, line 116: Great work getting this mess sorted! 👍

CarlosNZ (approved):
Looks good @GeronimoJohn, will be good to get this in. I'm happy with the changes, don't suppose @andreievg has anything remaining?

client/package.json
andreievg: We don't want any parallel compilation it's 3 x slower

PR #6797, opened 2025-03-03
6048 fix tooltip misplacement

### LINKED ISSUE #6048: The item appears somewhere else when you hover over it
## What went wrong? 😲

It doesn't look good 😅
Screenshot 2025-01-14 at 3 26 21 pm

## How to Reproduce 🔨

2. Add prescription
3. Add item 
4. Hover over item
5. See the item's name elsewhere

## Your environment 🌱

- Open mSupply Version: V2.5

- Platform:

- Database type:

zachariah-at-msupply:
Fixes #6048 

# 👩🏻‍💻 What does this PR do?

Removes tooltips from the prescriptions' line edit view.

This was achieved by adding a prop to the ListItem component, called 'tooltipDisabled'. If false, the list item is returned within a tooltip tag. If true, it is returned without the tooltip tag. This is all handled by and if/else clause which returns different code depending on the value of tooltipDisabled. I haven't seen this pattern elsewhere in the code, so I'm pointing it out in case it is not best practice for this situation.

# 🧪 Testing

- [ ] In Open mSupply, go to Dispensary > Prescriptionsand click on the button to add a prescription in the top right.
- [ ] Add an item to your prescription.
- [ ] Return to the main view of your prescription and click on the item you just added.
- [ ] Hover your cursor over the item you added.
- [ ] Check that no tooltip is shown.
- [ ] Navigate to another part of the UI and hover over something you would expect to have a tooltip.
- [ ] Check that a tooltip is shown.
Done!

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Item/Components/ListItems/ListItems.tsx
```diff
@@ -50,21 +52,39 @@ export const ListItems = ({
...
+              } else navigate(route.addPart(id).build());
+            }}
+            options={options}
+            enteredLineIds={enteredLineIds}
+            scrollRef={scrollRef}
+          />
+        </Box>
+      </Tooltip>
+    )
+  };
```
GeronimoJohn, line 89: This is on the right path!

Tooltips doesn't show if the [title](https://arc.net/l/quote/mmznnztp) prop is either `null`, `undefined`, `false` , or string have zero length.

We should be able to reduce the code through changing `title` only to `null` (or whatever value is preferred) if `tooltipDisabled` is `true`.

GeronimoJohn (commented):
Thanks @zachariah-at-msupply, just left one comment

client/packages/system/src/Item/Components/ListItems/ListItems.tsx
```diff
@@ -66,5 +68,5 @@ export const ListItems = ({
         />
       </Box>
     </Tooltip>
-  );
+  )
```
GeronimoJohn, line 71: Sorry can we add back the semi colon 🙏🏽

client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
```diff
@@ -172,6 +172,7 @@ export const PrescriptionLineEditView = () => {
             isDirty={isDirty.current}
             handleSaveNew={onSave}
             scrollRef={scrollRef}
+            tooltipDisabled={true}
```
GeronimoJohn, line 175: NIT: `tooltipDisabled={true}` -> `tooltipDisabled`. Can go without true

GeronimoJohn (approved):
LGTM! Just a few NITs. Up to you if you would like to edit those

jmbrunskill (changes requested):
@zachariah-at-msupply can you check if the weird behaviour with the tooltip is affecting the Internal Orders too, it looks like it has the same problem to me.

image


I think you can probably just remove the tooltip entirely from that ListItems component.

GeronimoJohn (approved):

PR #6804, opened 2025-03-04
#5825 Fix contraception method/prescription selection reset

### LINKED ISSUE #5825: Family Planning/Reproductive Health: contraception method selection resets on first attempt
## What went wrong? 😲

Seems when you first select a contraception method, it resets and only from the second attempt you can select it

Does not happen on the second attempt, so only first time is buggy

Video of the behaviour attached ->>> [here](https://drive.google.com/drive/folders/1jDys0w0F1P1sQOtiFlQ-w0bpU_zPBukX)

## Expected behaviour 🤔

Should be able to select contraception method from the first try

## How to Reproduce 🔨

1. Make sure you have Programs and Family Planning module enabled for your store
2. Create an encounter for an existing or a new patient who is enrolled into a family planning module
3. In the encounter, go to contraception selection and try and choose the prescribed contraception method
4. See buggy behaviour

## Your environment 🌱

- Open mSupply Version: `V2.4.1`
- Legacy mSupply Central Server Version: `V7.19.00`

- Database type: `SQLite3`

CarlosNZ:
Fixes #5825

# 👩🏻‍💻 What does this PR do?

As mentioned in [this comment](https://github.com/msupply-foundation/open-msupply/issues/5825#issuecomment-2695827269), the issue has become more severe since originally reported, as it creates an infinite re-render. 

This ended up being due to a change in [this commit](https://github.com/msupply-foundation/open-msupply/issues/5825#issuecomment-2695827269) where the `setIsDirty` function is now being returned from the `useConfirmOnLeaving` hook. This caused `setIsDirty` to be unstable, returning a new instance on every render, which in turn caused the `useFormActions` hook to return a new object, etc. causing a loop in the Prescription component.

I'm not particularly happy with this fix, but it works, and it's small. I tried a `useCallback()` to return the `setIsDirty`, but that didn't work -- and I'm a bit mystified as to why, as it should have as far as my understanding goes.

# 🧪 Testing

- [ ] Confirm problem as described in the issue is fixed
- [ ] Confirm you can still create a prescription, save it, etc.
- [ ] Confirm the "dirty state" behaviour is still correct, i.e. the "Save" button remains disabled until you make at least one change, then it can be clicked. After save, it should revert to disabled until the next change.
- [ ] Please check in a couple of other places for possible regressions (anywhere that uses the "Confirm" modal when trying to leave the page) -- make sure you can still save and that "dirty state" behaviour is correct.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
+++ b/client/packages/common/src/hooks/useConfirmOnLeaving/useConfirmOnLeaving.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { useBeforeUnload, useBlocker } from 'react-router-dom';
 import { create } from 'zustand';
 import { useTranslation } from '@common/intl';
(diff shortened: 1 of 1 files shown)


GeronimoJohn (approved):
Solutions works, small changes and straight to the point! Though when we get time, probs good to dig more into it?

PR #6867, opened 2025-03-06
change background colour to white

### LINKED ISSUE #6795: Prescription side bar inputs should have white background
## What went wrong? 😲

Image

## Expected behaviour 🤔

Background for these inputs should be white (or light grey?) to make it clear they are editiable.

## Your environment 🌱

- Open mSupply Version: V2.6.0-RC

- Platform:

- Database type:

zachariah-at-msupply:
Fixes #6795 

# 👩🏻‍💻 What does this PR do?

This issue gives inputs in the prescription sidebar (the 'more' panel) white backgrounds so that they look different from the normal text fields.

Have a good weekend!

# 🧪 Testing

- [ ] In Open mSupply, go to Dispensary > Prescriptions
- [ ] Select a prescription (or make one if you don't have any, by clicking on the 'add prescription' button in the top right)
- [ ] Click 'more' in the top right.
- [ ] Check that the inputs in the sidebar all have white backgrounds.
- [ ] Navigate to another part of mSupply and check the inputs in the toolbar are as they were before: only the inputs in the prescriptions sidepanel should be changed.
Thanks!

# 📃 Documentation

- [x] **These areas should be updated or checked**: Screenshots of the prescriptions' sidepanel should be updated to be consistent with the UI changes.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
+++ b/client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
@@ -7,6 +7,7 @@ import {
   AutocompleteInputChangeReason,
   AutocompleteProps as MuiAutocompleteProps,
   PopperProps,
+  SxProps,
 } from '@mui/material';
 import {
   AutocompleteOption,
@@ -48,6 +49,7 @@ export interface AutocompleteProps<T>
   popperMinWidth?: number;
   inputProps?: BasicTextInputProps;
   required?: boolean;
+  textSx?: SxProps;
 }
 
 export function Autocomplete<T>({
@@ -75,6 +77,7 @@ export function Autocomplete<T>({
   popperMinWidth,
   inputProps,
   required,
+  textSx,
   ...restOfAutocompleteProps
 }: PropsWithChildren<AutocompleteProps<T>>): JSX.Element {
   const filter = filterOptions ?? createFilterOptions(filterOptionConfig);
@@ -90,6 +93,7 @@ export function Autocomplete<T>({
           disableUnderline: false,
           sx: {
             padding: '4px !important',
+            ...textSx != null ? textSx : {},
           },
           ...props.InputProps,
         },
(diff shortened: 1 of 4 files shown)


client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -90,6 +93,7 @@ export function Autocomplete<T>({
           disableUnderline: false,
           sx: {
             padding: '4px !important',
+            ...textSx != null ? textSx : {},
```
CarlosNZ, line 96: The `textSx` should never be `null` as you've defined it as `textSx?: SxProps`, so either `SxProps` or undefined.
Also, the lack of brackets is confusing, as it's not clear you're spreading the whole thing, not just the first `textSx` (I don't actually know what the default order of operations would be here). 

But I think the cleanest approach is to just have `...textSx` here, and then in the function input give it a default of `{}`.

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PrescriptionDetailsSection.tsx
```diff
@@ -186,6 +186,13 @@ export const PrescriptionDetailsSectionComponent: FC = () => {
             disabled={isDisabled}
             size="small"
             sx={{ width: 250 }}
+            slotProps={{
+              input: {
+                sx: {
+                  backgroundColor: 'white',
```
CarlosNZ, line 192: We generally don't want to refer to colour directly, we use the "theme" colours, which means we can update them at once (e.g. if someone switches to a "Dark" them, this would look completely wrong).

Have a look at some of the other places this is done, usually using the `useTheme` hook. The colour you'd want here is `theme.palette.background.white` :)

CarlosNZ: Same for the other places too

CarlosNZ (changes requested):
Thanks, @zachariah-at-msupply, looking good. Couple of small comments to address, let me know if you need any further explanation.

I'm tempted to suggest making a more general prop for inputs to handle different backgrounds, but let's not do that just yet.

CarlosNZ (approved):
Nice -- Let's get it merged! ✊

PR #6875, opened 2025-03-09
Allow setting config file via command line argument

### LINKED ISSUE #6674: Add config location with cli when omSupply starts

Need a way to specify config file/folder location when starting omSupply. 

This will be very useful when starting multiple omSupply instances (with database in different locations etc..). Could also be useful for starting multiple instances of omSupply as a service

Use clap, default to current location

Keep on using ENV variable overrides, which is way less convenient. Although can probably use env files but this defeats the purpose of having config file

EthanMcQ-TMF:
Fixes #6674

# 👩🏻‍💻 What does this PR do?

Allow passing a `config-path` argument to server code to load separate config.yaml files.

I used `config-path` as the name to avoid confusion with the cargo `--config` arg for loading a `.toml` file

Code will look for `base.yaml` within the same directory as the chosen file.

# 🧪 Testing

- [ ] Run the server, should load `local.yaml` from the default location
- [ ] Create a new `.yaml` configuration file
  - Alter some of the settings (I recommend `database_name` since that's logged during startup)
- [ ] Run the server with `--config-path` pointed to the new configuration file, it should load the adjusted settings

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/server/src/configuration.rs
```diff
@@ -19,12 +20,24 @@ const CONFIGURATION_EXAMPLE_FILE_PATH: &str = "example.yaml";
 const CONFIGURATION_ENVIRONMENT_PREFIX: &str = "app";
 const CONFIGURATION_ENVIRONMENT_SEPARATOR: &str = "__";
 
+/// Config argument
+#[derive(Args)]
+pub struct ConfigArgs {
+    #[clap(long)]
```
andreievg, line 26: Can you add comment here about base.yaml please, so that it's part of --help

server/server/src/configuration.rs
```diff
@@ -119,12 +132,14 @@ pub fn copy_example_environment_configuration_file_if_required() -> Result<(), S
 ///
 /// Individual settings properties can be also manually defined as environment variables: see
 /// `get_configuration_environment` for details.
-pub fn get_configuration() -> Result<Settings, SettingsError> {
+pub fn get_configuration(config_file: Option<PathBuf>) -> Result<Settings, SettingsError> {
```
andreievg, line 135: Should this one get ConfigArgs rather then config_file ? Seems like you've made ConfigArgs, rathern then just config_file: Option<PathBuf>, so that future configuration additions would require less code change (since cli and main executeable share these configs), it feels natural in this sense to pass ConfigArgs to get_configuration

server/server/src/configuration.rs
```diff
@@ -133,6 +148,21 @@ pub fn get_configuration() -> Result<Settings, SettingsError> {
     Ok(settings)
 }
 
+fn derive_file_paths(config_file: Option<PathBuf>) -> Result<ConfigFilePaths, SettingsError> {
+    let file_paths: ConfigFilePaths = match config_file {
```
andreievg, line 152: Just fyi : ConfigFilePaths is inferred here

andreievg (approved):
Looks great, can be merged, but would be great to add a small comment that will appear in --help, see my other comment

andreievg (commented):
Actually another this to do, is to mention this in README.md, and when merged in omSupply chat (not the dev one but the general, are you in that one ?)

Issue #6823, opened 2025-03-04
Compile host only


andreievg:
## Is your feature request related to a problem? Please describe 👀

Compilation takes longer then needed and doesn't provide click through

We are compiling all packages in parallel, if I remove `--since HEAD` I get the following result:

```
   ✔  @openmsupply-client/reports:tsc (60s)
   ✔  @openmsupply-client/invoices:tsc (60s)
   ✔  @openmsupply-client/dashboard:tsc (1m)
   ✔  @openmsupply-client/requisitions:tsc (1m)
   ✔  @openmsupply-client/coldchain:tsc (1m)
   ✔  @openmsupply-client/system:tsc (1m)
   ✔  @openmsupply-client/programs:tsc (1m)
   ✔  @openmsupply-client/inventory:tsc (1m)
   ✔  @openmsupply-client/common:tsc (1m)
   ✔  @openmsupply-client/config:tsc (1m)
   ✔  @openmsupply-client/host:tsc (1m)
```

If I just do @openmsupply-client/host, with `--since HEAD` removed i get

```
 @openmsupply-client/host (18s)
```

If I make a small change in reports, with `--since HEAD` and `@openmsupply-client/*` (small upstream packages), i get the following:

```
   ✔  @openmsupply-client/reports:tsc (20s)
   ✔  @openmsupply-client/host:tsc (21s)
```

Above suggests that we are not getting any benefit from compiling in parallel (at least the way we are using it now), and it's better to just compile host

## Describe the solution you'd like 🎁

Replace `compile` with `re-compile` script as per this comment: https://github.com/msupply-foundation/open-msupply/pull/6796#issuecomment-2712519709

Also check if compile-full is sued anywhere and remove (what about clean and server ?) 

Lastly check if we need lerna in start local and start remote (would like to see if there is some error is it easier to navigate to it, through click through in console, without it)

GeronimoJohn:
> [@GeronimoJohn](https://github.com/GeronimoJohn)
> 
> In this PR i was hoping we would remove 'parallel' compilation as per this comment:
> 
> [#6796 (comment)](https://github.com/msupply-foundation/open-msupply/pull/6796#issuecomment-2712519709)
> 
> I've updated this issue description

Thanks Andrei, picking it up soon

PR #6879, opened 2025-03-10
6308 improve codegen generation

### LINKED ISSUE #6308: Format generated .ts files after `yarn generate`

We use our `yarn generate` script to generate types and graphql query functions, which are outputted to .ts files.

However, these generated files don't comply with the TS file code formatting standards we use throughout the rest of our code.

Usually this is fine (we shouldn't be editing generated files) but often you open the file to look at a type, and hit save - you have format-on-save on and the whole file reformats! 1000+ line PR incoming 😭 

Run a format on the generated files after generation - or maybe we can pass our prettier config into the generate script?

Make the generated files into the same formatting so we don't have PR bloat/noise or annoyingly having to unformat/rerun scripts.

andreievg:
Fixes #6308

Needed to priorities this due to testing that will be added for plugins (to at least make sure that we haven't broken types, which required a reliable graphql type generation)

# 👩🏻‍💻 What does this PR do?

Format codegened files.

Fix occasional typescript errors:

It turns out most of the problem with had is because we didn't exclude node modules, I had different solution at first but when i added a custom hook to fix the export at the bottom, i saw that codegen was passing through every generated file, and a lot of them were from node_modules (actually heaps of them). So after exclude node_modules in codegen.yaml, things seem to work correctly.

I also added re-compile, it is taking 3 times as long to to do our `yarn compile`, also re-compile output is now click through (to vs code file), there is related issue: https://github.com/msupply-foundation/open-msupply/issues/6823

# 🧪 Testing

Run generation a few times, and compilation, i used the new re-compile command: `yarn generate && yarn re-compile`


client/codegen.yml
```diff
@@ -25,3 +27,7 @@ generates:
       scalars:
         DateTime: string
         NaiveDate: string
+        NaiveDateTime: string
+hooks:
+  afterAllFileWrite:
+    - prettier --write
```
jmbrunskill, line 33: Will this break if someone doesn't have prettier installed? Do we need to update some docs?

jmbrunskill (approved):
Thanks Andrei, makes sense to get this in I think!
If there's some operation issues with it we can deal with them as they come up.
Not sure if the prettier thing will help with the occasional issue with yarn generate failing to update correctly? Hopefully it doesn't make it worse anyway.

andreievg:
@jmbrunskill thanks for review 

>Not sure if the prettier thing will help with the occasional issue with yarn generate failing to update correctly? Hopefully it doesn't make it worse anyway.

We were including node_modules in generation, that was the main issue. 
I don't think prettier has any effect, generated files overwritten and hook happens after all files were written.,

client/codegen.yml
andreievg: Hmm, there is something in: https://github.com/msupply-foundation/open-msupply/blob/5417317be042f067e66a13715c0f556c9312352b/CONTRIBUTING.md#L81 about it

PR #6880, opened 2025-03-10
6013 Fix customer Stock On Hand for - request requisitions in Legacy mSupply

### LINKED ISSUE #6013: Internal order `stock_on_hand` not synced to OG correctly
## What went wrong? 😲

Customer current SOH on requisitions OMS -> OG is always 0! 😱 

Screenshot 2025-01-13 at 7 12 50 PM

When creating an internal order, we include our SOH, in this case 150:
Screenshot 2025-01-13 at 7 39 42 PM

In OMS, we save this as `available_stock_on_hand` on requisition line:
Screenshot 2025-01-13 at 7 40 54 PM

Which gets translated to the `stock_on_hand` field in v5 sync to OG:
Screenshot 2025-01-13 at 7 42 20 PM

But OG is expecting that value to be in `cust_prev_stock_balance`, which OMS sets as 0!
Screenshot 2025-01-13 at 7 45 03 PM

Pretty sure in OG the `stock_on_hand` field is used for the SOH of the supplying store in the response requisition

Translations and transfers will need to be reviewed, to correctly handle this value differently between request and response requisitions. 

## Expected behaviour 🤔

See customer SOH in OG requisitions from OMS internal orders.

## How to Reproduce 🔨

1. In OMS Store, create internal order to an OG store, with some SOH for your item
2. Mark as sent and sync
3. View requisition in OG - no customer SOH!

## Your environment 🌱

- Open mSupply Version: 2.5.0 RC2
- Legacy mSupply Central Server Version: 7.18

- Platform:

- Database type:

jmbrunskill:
Fixes #6013

# 👩🏻‍💻 What does this PR do?

Possible fix for translation of stock_on_hand for legacy mSupply.

I've just set the `initial_stock_on_hand_units` in the database when we create a new request requisition.

I don't know if this has some kind of bad effects somewhere else in the code? Looks like it's ok to me, but the naming doesn't make sense, feels like both OMS and the OG names could do with some work.

I also don't know why the name was changed to `initial_stock_on_hand_units` it's not obvious from that name to me that this is referring to customer `SOH` maybe we should explicitly state that?

It also sounds like this might be linked to the previous stock on hand value from the last R&R form based on the name in OG, so maybe it's intended to show a different number entirely?

Maybe mSupply should translate from SOH to `Cust_prev_stock_balance` itself and we shouldn't make this change in OMS?

Do we want this same behaviour for manual requisitions? It feels like maybe we do but not sure how 

# 🧪 Testing

- [ ] Create an internal order from OMS to OG
- [ ] Check that the requisition created in OG has the corect SOH for the customer site.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/requisition/request_requisition/generate.rs
+++ b/server/service/src/requisition/request_requisition/generate.rs
@@ -74,7 +74,7 @@ pub fn generate_requisition_lines(
                 requested_quantity: 0.0,
                 approved_quantity: 0.0,
                 approval_comment: None,
-                initial_stock_on_hand_units: 0.0,
+                initial_stock_on_hand_units: available_stock_on_hand,
                 incoming_units: 0.0,
                 outgoing_units: 0.0,
                 loss_in_units: 0.0,


server/service/src/requisition/request_requisition/generate.rs
```diff
@@ -74,7 +74,7 @@ pub fn generate_requisition_lines(
                 requested_quantity: 0.0,
                 approved_quantity: 0.0,
                 approval_comment: None,
-                initial_stock_on_hand_units: 0.0,
+                initial_stock_on_hand_units: available_stock_on_hand,
```
roxy-dao, line 77: Initial stock on hand is the stock on hand at the start of the period.. so I feel this isn't right either 🤔

jmbrunskill:
Ok general consensus seems to be that this is an mSupply Bug, so I'll close this PR and have raised one for mSupply.

https://github.com/msupply-foundation/msupply/issues/16254

PR #6883, opened 2025-03-10
6599 unable to set prescribed quantity when no stock

### LINKED ISSUE #6599: Unable to set Prescribed Quantity when no stock available
## What went wrong? 😲

Image

## Expected behaviour 🤔

Should be able to save

## How to Reproduce 🔨

1. Create a prescription
2. add an item with no stock
3. Try to add a Prescribed Quantity
4. Try to Save (Save button disabled)

## Your environment 🌱

- Open mSupply Version: V2.6.0-RC1

- Platform:

- Database type:

GeronimoJohn:
Fixes #6599

# 👩🏻‍💻 What does this PR do?

We want to enable saving an item within a prescription even when there is no stock, provided that the user has set a prescribed quantity.

To achieve this, we have introduced a new mutation called `set_prescribed_quantity` within the `batch_prescription` mutation. This focuses on saving an item despite the absence of stock.

# 🧪 Testing

- [ ] Navigate to any prescriptions
- [ ] Add a new item with no stock
- [ ] Enter a prescribed quantity to the selected item
- [ ] Save button should get enabled - click on the save button
- [ ] Item should get saved!

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


GeronimoJohn:
Found a few issues - working on fix

jmbrunskill:
image

jmbrunskill:
image

jmbrunskill:
Not a major but something like this would be nice...
image

jmbrunskill:
After adding stock, the prescribed quantity disappears?
image

(Not a major btw, I think this is an edge case)

jmbrunskill:
> After adding stock, the prescribed quantity disappears? 

Actually this seems to be a general problem...
image

jmbrunskill (changes requested):
Can you remove prescribed quantity from 

`InsertPrescriptionLineInput` and `UpdatePrescriptionLineInput` please?

I think it will be confusing having two ways to update the same thing...

server/service/src/invoice_line/stock_out_line/set_prescribed_quantity/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+
+            let invoice_line_row_repo = InvoiceLineRowRepository::new(connection);
+
+            let existing_lines = InvoiceLineRepository::new(connection).query_by_filter(
+                InvoiceLineFilter::new()
+                    .item_id(EqualFilter::equal_to(&input.item_id))
+                    .invoice_id(EqualFilter::equal_to(&input.invoice_id)),
+            )?;
+
+            if existing_lines.len() > 1 {
```
jmbrunskill, line 55: I think this logic should have some comments to explain the intended behaviour.

E.g.

When we set prescribed quantity for an item, there's 3 scenarios.

1. We don't have any stock lines already create for that item, in which case we need to create an unallocated line for the prescription to record the prescribed quantity against
2. We already have more than one stock line, we need to update the one that already has prescribed quantity, or assign it to a single line if none of them have it.
3. If there's a proper allocated line, and an unallocated one, we need to remove the unallocated one just to clean up

server/service/src/invoice_line/stock_out_line/set_prescribed_quantity/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+                invoice_id: mark_prescription_unallocated_invoice_line().invoice_id,
+                item_id: mark_prescription_unallocated_invoice_line().item_link_id,
+                prescribed_quantity: 10.0,
+            },
+        );
+
+        assert!(result.is_ok());
+
+        let invoice_line = result.unwrap();
+
```
jmbrunskill, line 312: I think you need a test here to see how many lines exist after this function runs. 
We should be deleting the unallocated line right?

jmbrunskill: Would also like a test where there's more than one allocated stockline for the item.
We should check that there's only one line with prescribed quantity set on it in this scenario.

server/service/src/invoice_line/stock_out_line/set_prescribed_quantity/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+            if let Some(existing_line) = existing_line_with_stock {
+                let mut updated_line = existing_line.clone();
+                updated_line.invoice_line_row.prescribed_quantity = Some(input.prescribed_quantity);
+                invoice_line_row_repo.upsert_one(&updated_line.invoice_line_row)?;
+                new_line = updated_line.invoice_line_row
+            } else {
+                let invoice_line = generate(uuid(), item_row, input.clone())?;
+                invoice_line_row_repo.upsert_one(&invoice_line)?;
+                new_line = invoice_line;
+            }
```
jmbrunskill, line 84:
```suggestion
            let new_line = match existing_line_with_stock {
                Some(existing_line) => {
                    let mut updated_line = existing_line.clone();
                    updated_line.invoice_line_row.prescribed_quantity =
                        Some(input.prescribed_quantity);
                    invoice_line_row_repo.upsert_one(&updated_line.invoice_line_row)?;
                    updated_line.invoice_line_row
                }
                None => {
                    let invoice_line = generate(uuid(), item_row, input.clone())?;
                    invoice_line_row_repo.upsert_one(&invoice_line)?;
                    invoice_line
                }
            };
```

Using a match here is more `Rusty` than the if else.
I'm actually quite surprised your code worked without assigning `new_line` as `mut`?

jmbrunskill (commented):
Have you tried this with Edit prescribedQuantity turned off?

image

jmbrunskill:
> Have you tried this with Edit prescribedQuantity turned off?
> 

When I do I get this...

image

I've also noticed that we're sending a lot of duplicate setPrescribed Quantity requests???

image


We should only send 1 per item...

And Ideally when the store pref is off, we wouldn't send any at all

GeronimoJohn:
Thanks for the review @jmbrunskill 

Clean up `set_prescribed_quantity` changes: https://github.com/msupply-foundation/open-msupply/pull/6883/commits/539c90face0f9a7e1167bd19e0a287675c976530

Remove `prescribed_quantity` from `InsertPrescriptionLineInput` and `UpdatePrescriptionLineInput`: https://github.com/msupply-foundation/open-msupply/pull/6883/commits/c5dbe7b0e8aacd072ad4aa5965c6bc1558dbc4f5

NaN issue fix: https://github.com/msupply-foundation/open-msupply/pull/6883/commits/b922c5dbf086d52d2e1aec7092df60516d877fab

GeronimoJohn:
> Have you tried this with Edit prescribedQuantity turned off?
> 
> image

Yup! Prescribed Quantity input gets hidden, or at least it should 😁 

Screenshot 2025-03-11 at 2 48 16 PM

GeronimoJohn:
> image

What would be the expected behaviour here? In my head it kinda sense to not change status since the items are just placeholders

GeronimoJohn:
> After adding stock, the prescribed quantity disappears? image
> 
> (Not a major btw, I think this is an edge case)

Fixed the issue within: https://github.com/msupply-foundation/open-msupply/commit/be9fbc0590674e47d976fe7e20c6a3dd4dadf0b5 (I think 😄)

https://github.com/user-attachments/assets/c0829fc4-c8ab-4e79-bbd3-a8359384e44f

jmbrunskill (approved):
Thanks @GeronimoJohn working well from what I can tell!

Note: I've updated your tests cases, it might be good if you can review that commit before merging.

There were a few problems in the tests and I thought I'd just do a comit rather than suggestion branches explainations etc.

1. You had duplicate ids for the invoice line row, this meant that the 3 lines was actually only 1 due to upsert logic
2.  Updated the tests to query from database rather than rely on mocks which aren't actually updated by the logic
3. Added tested to check the unallocated line is deleted.

GeronimoJohn:
> Thanks @GeronimoJohn working well from what I can tell!
> 
> Note: I've updated your tests cases, it might be good if you can review that commit before merging.
> 
> There were a few problems in the tests and I thought I'd just do a comit rather than suggestion branches explainations etc.
> 
> 1. You had duplicate ids for the invoice line row, this meant that the 3 lines was actually only 1 due to upsert logic
> 2. Updated the tests to query from database rather than rely on mocks which aren't actually updated by the logic
> 3. Added tested to check the unallocated line is deleted.
> 
> [f80676f](https://github.com/msupply-foundation/open-msupply/commit/f80676f1394aa3e664c1da9932f6781e5a1013c4)

Thanks @jmbrunskill! I've reviewed the commit, and that makes sense. I'm kinda facepalming a bit over points 1 and 2—didn't catch that while writing it 😆

PR #6884, opened 2025-03-10
Add soft delete translations

### LINKED ISSUE #5314: No way to remove CCE from central database
## What went wrong? 😲

- When deleting an equipment asset on a remote site, that same asset is not removed from the central server (bug 1) 
- The central user is also unable to delete a CCE from a remote site (bug 2)
- The error seems to be doubled - is it possible to show one error with the contents of both? (bug 3)

Spoke to @adamdewey, he confirmed ideally central server user should be able to remove assets even if they belong to other stores. Not sure if this one is a bug or intended?

Behaviour attached ->>> [here](https://drive.google.com/drive/folders/1hGIkxIMZhSWiqXM7ccjW6RquHZXrdiKM)

## Expected behaviour 🤔

- 1: Equipment should be updated on central server when removed on a remote site
- 2: Should only be one error
- 3: If not intended behaviour, then central server user should be able to remove cold chain assets even when assigned to other stores. 

## How to Reproduce 🔨

2. Have Remote Site & Central Server set up
3. Add an asset for a remote site from your central server -> see it appear on your remote site
4. Attempt to remove that asset from your central server -> unable to, that's `bugs 2 and 3`
5. Now go to your remote site -> remove the asset, see central server does not get updated -> `bug 1`

## Your environment 🌱

- Open mSupply Version: `V2.4.0-test1`

- Platform: Google Chrome browser on laptop, Galaxy Tab A8 tablet (Android)

- Database type: `SQLite3`

fergie-nz:
Fixes #5314 

# 👩🏻‍💻 What does this PR do?

- Adds delete sync for assets
- Remote sites can delete their assets. This is synced to OMS central server.
- Central server can delete assets from any site. This is synced to remote sites.

# 🧪 Testing

- [ ] Have OMS set up on central and remote sites
- [ ] Create an asset from central server for a remote site
- [ ] See the asset is visible on both sites
- [ ] Delete the asset from central server
- [ ] See that this change is made for both sites.

Repeat the above steps for each permutation of creation and deleting:
Create from central - delete from remote
Create from remote - delete from central
Create from remote - delete from remote

In all cases the asset will show up on both when created from either central and remote when created, and will be removed from both when deleted from either.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/asset/delete.rs
+++ b/server/service/src/asset/delete.rs
@@ -1,4 +1,5 @@
 use super::validate::check_asset_exists;
+use crate::sync::CentralServerConfig;
 use crate::{activity_log::activity_log_entry, service_provider::ServiceContext};
 use repository::assets::asset_internal_location_row::AssetInternalLocationRowRepository;
 use repository::{
@@ -55,7 +56,7 @@ pub fn validate(
     };
 
     if let Some(store_id) = &asset_row.store_id {
-        if ctx_store_id != store_id {
+        if ctx_store_id != store_id && !CentralServerConfig::is_central_server() {
             return Err(DeleteAssetError::AssetDoesNotBelongToCurrentStore);
         }
     }
--- a/server/service/src/asset/delete.rs
+++ b/server/service/src/asset/delete.rs
@@ -37,7 +37,7 @@ pub fn delete_asset(ctx: &ServiceContext, asset_id: String) -> Result<String, De
                 .delete_all_for_asset_id(&asset_id);
 
             AssetRowRepository::new(connection)
-                .delete(&asset_id)
+                .mark_deleted(&asset_id)
                 .map_err(DeleteAssetError::from)
         })
         .map_err(|error| error.to_inner_error())?;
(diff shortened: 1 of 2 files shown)


server/service/src/asset/delete.rs
```diff
@@ -55,7 +56,7 @@ pub fn validate(
     };
 
     if let Some(store_id) = &asset_row.store_id {
-        if ctx_store_id != store_id {
+        if ctx_store_id != store_id && !CentralServerConfig::is_central_server() {
```
fergie-nz, line 59: This is to ensure remote sites can't delete equipment from other stores.

I'm now wondering if it would be important to cover this use case with tests?

jmbrunskill: I would argue the change it to allow central server to delete the remote sites data...
I think it's ok to not have a test case for this for now, although I agree it would be a good idea. It's not easy to put that into a test case though, although I guess we can probably fake being a central server somehow?

The whole area needs a review I think as it's confusing for users to have different behaviour on central server vs remote.

jmbrunskill (approved):
Works good for me, thanks @fergie-nz

PR #6893, opened 2025-03-10
6224 create dummy users on sync

### LINKED ISSUE #6224: User that are not active doesn't seems to be migrated on omSupply even though they entered historical transactions
## What went wrong? 😲

I have historical transactions where the "Entered by" is empty, the user who have entered it is now inactive in OG. I am not sure but it might be possible that it was not migrated because the user is inactive.
https://drive.google.com/file/d/1oArIM7bi7Gm0JMzFu5FIhFabDsQYl9PY/view?usp=sharing

on OG

https://drive.google.com/file/d/1ZDKCvAlZqXtFrOCFhQdMnI35sxpP2VpV/view?usp=sharing

The immediate consequence is that the outbound shipment invoice (outbound-shipment-landscape) trigger an error on printing

Image

## Expected behaviour 🤔
All outbound shipment should have an Entered user even though the user is inactive

## How to Reproduce 🔨

Choose a store for which there are transactions done by now inactive users. Migrate that store to omSupply 
1. Go to 'Distribution > outbound shipment'
2. Open an outbound shipment entered by that inactive user 
3. Click on the More button
4. See error. The Entered by field is empty
5. Click on the Print button
6. See Error. There is an error message 

## Your environment 🌱

- Open mSupply Version: v2.5.0RC5
- Legacy mSupply Central Server Version: v8.00dev43

- Platform: Android

- Database type: SQLite3

jmbrunskill:
Fixes #6224

# 👩🏻‍💻 What does this PR do?

If a transaction comes in via sync,and we don't have the associated user_id, a default user is created with blank details

I haven't really tested this with a real scenario, I assume that's ok as @alainsussol  can test his real world scenario?

# 🧪 Testing

See issue #6224 for testing steps

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [X] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/test/test_data/invoice.rs
+++ b/server/service/src/sync/test/test_data/invoice.rs
@@ -155,7 +155,7 @@ fn transact_1_push_record() -> TestSyncOutgoingRecord {
         record_id: TRANSACT_1.0.to_string(),
         push_data: json!(LegacyTransactRow {
             ID: TRANSACT_1.0.to_string(),
-            user_id: None,
+            user_id: Some("MISSING_USER_ID".to_string()),
             name_ID: "name_store_a".to_string(),
             store_ID: "store_b".to_string(),
             invoice_num: 1,
--- a/server/service/src/sync/test/test_data/invoice.rs
+++ b/server/service/src/sync/test/test_data/invoice.rs
@@ -85,7 +85,7 @@ const TRANSACT_1: (&str, &str) = (
       "user2": "",
       "user3": "",
       "user4": "",
-      "user_ID": "",
+      "user_ID": "MISSING_USER_ID",
       "wardID": "",
       "waybill_number": "",
       "om_allocated_datetime": "",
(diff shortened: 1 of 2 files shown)


fergie-nz (approved):
Works great!

I am wondering though why not make reports flexible to not show users in this case to avoid errors? Are there other issues that might arise from having missing users?

Also I believe this same issue could also occur in stocktakes (none of the standard forms use the user currently, but has the same issue of syncing across and saving a possibly inactive user). Maybe worth adding in here also if easy?

jmbrunskill:
> Works great!
> 
> I am wondering though why not make reports flexible to not show users in this case to avoid errors? Are there other issues that might arise from having missing users?

Great point and I agree. I think @andreievg was thinking this would fix it all in one go, and there was a high risk of some future reports being broken for the same reason. The correct fix is probably for mSupply to sync the users?

> Also I believe this same issue could also occur in stocktakes (none of the standard forms use the user currently, but has the same issue of syncing across and saving a possibly inactive user). Maybe worth adding in here also if easy?

I think it's a good point. I think I'll merge this and can come back to stocktake?

PR #6899, opened 2025-03-11
6526 replace number urls with UUID

### LINKED ISSUE #6526: Store merging breaks invoice/requisition URLs by serial number
## What went wrong? 😲

I don't have pics, but this has been reported to me :)

https://cerb.msupply.org/profiles/ticket/LY-9179-FG

OG has implemented the likely cursed feature of merging stores. It's in the TODO for OMS #4468. Nonetheless a problem has arisen when stores are merged and then OMS is initialised - when you view invoice by serial number 1, but there are 2 invoices with serial number 1, what happens? Turns out, you just get one of them reliably but never the other :(. And it has the lines of both 🥴.

As OMS has different requirements and architecture, effectively being a web app, we've done URLs based on resource and serial number to keep them tidy. 4D gets away with a lot being a low code app that only really works well on LAN :) (at least the way we've done it 😉).

## Expected behaviour 🤔

NOT BORKEN.

## How to Reproduce 🔨

1. Merge 2 stores each with invoices in OG
2. Initialise the remaining store on an OMS site
3. View an invoices, see duplicate serial numbers, click either of them and get the same parent invoice, with the lines of both (you can check on OG central what they should look like)

## Your environment 🌱

- Open mSupply Version: 2.5 but certainly effects all versions
- OG: Recent enough to have merge. 7.18 or something.

## Agreed solution

We should rewrite to use UUIDs in the URL instead:

- It fixes this problem with merging
- It fixes this problem when invariably some future bug causes duplicate serial numbers outside of merging
- It simplifies the API

### TODO

- [ ] change invoice GQL to do querying by record.id 
- [ ] change requisition GQL to do querying by record.id 
- [ ] change stock takes GQL to do querying by record.id
- [ ] change invoice_line GQL to do querying by record.invoice_id
- [ ] change requisition_line GQL to do querying by record.requisition_id
- [ ] change stock_take_line GQL to do querying by record.stock_take_id

## Alternatives

1. My presumption is that we can't not have duplicate serial numbers. If they have paper copies of old invoices, they need to be able to look up the serial number from paper in OMS and find it. If I'm wrong, then the easiest solution is for merge logic to change the serial numbers. Need broader input on this. **This needs client feedback - some might veto duplicate serial numbers, some may veto modifying them. I say we leave duplicates those who veto them shan't merge stores lol**
2. We don't URL by serial number and use UUID instead 😢 ugly URLs. Effective though! Less ugly than google URLs still lol. **We might actually just do this**
3. Tangent: Invoice line loader (I looked now) should use the UUID, not the serial number of the invoice. FE can have UUID in state, but in URL perhaps still have serial number ideally if we can.
4. We do 3, and if the backend detects duplicate invoice numbers it redirects FE to a UUID based URL to invoice. This is more complicated but we get to keep serial number URLs for general case. **Why so complicate plssss 😢**

fergie-nz:
Fixes #6526

# 👩🏻‍💻 What does this PR do?

- Replaces byNumber url param queries with byId url param queries
- Adds custom breadcrumbs for numbers where IDs are now used in navigation 

There are a number of byNumber end points that  are no longer no longer being used with this change. I was hesitant to take them out in case they were being used by any report / plugin which I am unaware of (though these will likely soon encounter the same issues once stores are merged).

[Follow up issue](https://github.com/msupply-foundation/open-msupply/issues/6902) to remove byNumber endpoint from our code base.

String() functions linger in some places (and throughout navigation where id was used anyway). Is another way to handle undefined being passed into RouteBuilder.

# 🧪 Testing

- [ ] Navigate in to detail view
- [ ] Confirm that link works as expected
- [ ] Confirm that number is used in breadcrumbs (rather than a UUID)
- [ ] Repeat the above with:
- inboundShipments
- requestRequisitions 
- supplierReturns 
- OutboundShipments 
- ResponseRequisitions 
- SupplierReturns 
- prescriptions 
- stock takes.

Prescriptions has a nested detail view
- [ ] Navigate to a prescription
- [ ] Ensure uuid is not exposed in navigation bar
- [ ] Open an item in the prescription
- [ ] Ensure uuid is not exposed in navigation bar

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/views/mod.rs
```diff
@@ -143,6 +143,7 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
         name,
         invoice.type AS invoice_type,
         invoice.invoice_number AS invoice_number,
+        invoice.id AS invoice_id,
```
fergie-nz, line 146: Adding to this view allows breadcrumbs to expose the invoice id in front end.

client/packages/inventory/src/Stocktake/DetailView/SidePanel.tsx
```diff
@@ -88,6 +88,7 @@ export const SidePanel = () => {
     selectedRows: [stocktake],
     deleteAction,
     messages: {
+      // TODO differentiate betweeen duplicate stocktakeNumber on delete where stores have been merged (https://github.com/msupply-foundation/open-msupply/pull/6789)
       confirmMessage: t('messages.confirm-delete-stocktake', {
         number: stocktake?.stocktakeNumber,
```
fergie-nz, line 93: Niche case where merged stores have two stocktakes of the same number, and delete message is ambiguous. I considered removing the stocktake number from this message but figure it is a rare case where it would actually cause confusion, and would be helpful in most cases.

client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
```diff
@@ -174,7 +138,10 @@ const useGetById = (invoiceId: string | undefined) => {
 
     if (invoice?.__typename === 'InvoiceNode') {
       return invoice;
+    } else {
+      throw new Error('Could not find invoice');
     }
+    // TODO check if this is throwing unweildy errors where program component has id for yet to exist presciptions.
     // Don't throw error for this one if not found -- it's mainly used by
     // Program component (Prescription), which may have stored an id for a
     // prescription that doesn't yet exist
```
fergie-nz, line 147: I wasn't sure if this was a common case where prescriptions would have a stored id. Should we be throwing errors here? Or withholding errors?

Another option to make a prescription specific api which doesn't throw errors if not found. This could help if stored IDs is a common occurrence for prescriptions.

jmbrunskill:
@fergie-nz after discussion with @alainsussol I think we should merge this to develop for 2.7
The argument is thus...

1. This should mostly only affect merged stores, and then only for older records, newly create records should get a new number as far as I know
2. There's a lot to test, as you say and it could delay the 2.6.1 release even further adding to the testing load.

@mark-prins @andreievg please comment if you disagree.

client/packages/invoices/src/Prescriptions/api/hooks/usePrescriptionLines.ts
```diff
@@ -32,7 +32,7 @@ export const usePrescriptionLines = (id?: string) => {
     mutateAsync: updateMutation,
     isLoading: isSavingLines,
     error: saveLineError,
-  } = useSaveLines(data?.id ?? '', data?.invoiceNumber ?? -1);
+  } = useSaveLines(data?.id ?? '', data?.id ?? "");
```
CarlosNZ, line 35: Prettier again -- do you not have auto-format enabled?

client/packages/requisitions/src/RequestRequisition/api/hooks/line/useInsertRequestLine.ts
```diff
@@ -1,14 +1,14 @@
 import { useQueryClient, useMutation } from '@openmsupply-client/common';
-import { useRequestNumber } from '../document/useRequest';
+import { useRequestId } from '../document/useRequest';
 import { useRequestApi } from '../utils/useRequestApi';
 
 export const useInsertRequestLines = () => {
-  const RequestNumber = useRequestNumber();
+  const RequestId = useRequestId();
```
CarlosNZ, line 6: Prob shouldn't be capitalised

CarlosNZ (commented):
Wow, very comprehensive! Thanks for tracking down all these cases.

I haven't tested absolutely everything, but it looks solid and everything I did try seemed to work as expected.

There's a small amount of tiny formatting tweaks, but since you're sick today I can do them for you so we can get this in.

Thanks 🙏

CarlosNZ:
Holding off on merging this one until 2.6.1 branch has been merged into `develop`

CarlosNZ (approved):
Okay, looks good, let's get this in! Thanks for the hard work going through all this.

Can I suggest you post in the omSupply devs chat and explain this change and ask for people to be vigilant about any potential issues caused by it, as it's likely we won't have 100% accounted for every single case in such a large PR?

Thanks

PR #6910, opened 2025-03-12
6889 enable insurance tab even when program module off

### LINKED ISSUE #6889: Patient Insurance isn't available without the `uses Om Program Module` store pref
## What went wrong? 😲

There's a note in the code but we obviously missed it :(

```
      {/* Only show tabs if program module is on and patient is saved.
      TODO: Prescription tab? - would need tab refactoring since also seems useful in programs
      */}
      {!createNewPatient && store?.preferences.omProgramModule ? (
        <DetailTabs tabs={tabs} requiresConfirmation={requiresConfirmation} />
      ) : (
        <PatientDetailView onEdit={setIsDirtyPatient} />
      )}
```

Image

## Expected behaviour 🤔

If insuranceProviders are available, the tab should show for insurance, even if the `Uses Program Module` is turned off

## How to Reproduce 🔨

1. Turn off `Open mSupply: Uses Program Module`
2. Go to Patients
3. Try to add Insurnace

## Your environment 🌱

- Open mSupply Version: 2.6.0

- Platform:

- Database type:

GeronimoJohn:
Fixes #6889 

# 👩🏻‍💻 What does this PR do?

If insuranceProviders are available, the tab should show for insurance, even if the Uses Program Module is turned off.

Side Quest: I've moved `PatientDetailView` to its own file. `PatientView` had a lot going on for it, navigating can be a bit overwhelming

# 🧪 Testing

If you have insurance providers in your store and programs module is on
- [ ] Turn off `Open mSupply: Uses Program Module` within legacy mSupply for your store
- [ ] Navigate to a Patient
- [ ] Should be able to see any Programs related component
   - [ ] E.g. programs tab, encounter button in dropdown of `AddButton` 
- [ ] Should be able to navigate to insurance tab and add insurance

If you don't have insurance providers in your store and programs module is off
- [ ] Turn off `Open mSupply: Uses Program Module` within legacy mSupply for your store
- [ ] Navigate to a Patient
- [ ] Should not be able to see any programs related component
- [ ] Should not be able to add insurance or visit insurance tab

If you don't have insurance providers in your store but programs module is on
- [ ] Navigate to a Patient
- [ ] Should be able to see any Programs related component
   - [ ] E.g. programs tab, encounter button in dropdown of `AddButton` 
- [ ] Should not be able to navigate to insurance tab or add insurance

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Patient/PatientView/AddButton.tsx
+++ b/client/packages/system/src/Patient/PatientView/AddButton.tsx
@@ -72,30 +72,23 @@ export const AddButton: React.FC<AddButtonProps> = ({
     SplitButtonOption<PatientModal> | undefined
   >(options[0]);
 
+  const optionMap: Partial<Record<PatientTabValue, PatientModal>> = {
+    [PatientTabValue.Programs]: PatientModal.ProgramSearch,
+    [PatientTabValue.Vaccinations]: PatientModal.ProgramSearch,
+    [PatientTabValue.Encounters]: PatientModal.Encounter,
+    [PatientTabValue.ContactTracing]: PatientModal.ContactTraceSearch,
+    [PatientTabValue.Insurance]: PatientModal.Insurance,
+  };
+
   useEffect(() => {
     if (options.length === 0) return;
 
-    const optionMap: {
-      [key: string]: SplitButtonOption<PatientModal> | undefined;
-    } = {
-      [PatientTabValue.Programs]: options.find(
-        option => option.value === PatientModal.ProgramSearch
-      ),
-      [PatientTabValue.Vaccinations]: options.find(
-        option => option.value === PatientModal.ProgramSearch
-      ),
-      [PatientTabValue.Encounters]: options.find(
-        option => option.value === PatientModal.Encounter
-      ),
-      [PatientTabValue.ContactTracing]: options.find(
-        option => option.value === PatientModal.ContactTraceSearch
-      ),
-      [PatientTabValue.Insurance]: options.find(
-        option => option.value === PatientModal.Insurance
-      ),
-    };
-
-    setSelectedOption(optionMap[currentUrlTab as string] || options[0]);
+    setSelectedOption(
+      options.find(
+        option => option.value === optionMap[currentUrlTab as PatientTabValue]
+      ) || options[0]
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [currentUrlTab, options]);
 
   const onSelectOption = (option: SplitButtonOption<PatientModal>) => {
--- a/client/packages/system/src/Patient/PatientView/AddButton.tsx
+++ b/client/packages/system/src/Patient/PatientView/AddButton.tsx
@@ -33,7 +33,7 @@ export const AddButton: React.FC<AddButtonProps> = ({
   const { createNewPatient } = usePatientStore();
   const { setModal: selectModal, reset } = usePatientModalStore();
   const {
-    query: { data: InsuranceProvidersData },
+    query: { data: insuranceProvidersData },
   } = useInsuranceProviders();
 
   const options: SplitButtonOption<PatientModal>[] = useMemo(() => {
@@ -58,7 +58,7 @@ export const AddButton: React.FC<AddButtonProps> = ({
       );
     }
 
-    if (InsuranceProvidersData?.length > 0) {
+    if (insuranceProvidersData?.length > 0) {
       baseOptions.push({
         value: PatientModal.Insurance,
         label: t('button.add-insurance'),
@@ -66,7 +66,7 @@ export const AddButton: React.FC<AddButtonProps> = ({
     }
 
     return baseOptions;
-  }, [disableEncounterButton, t, store, InsuranceProvidersData]);
+  }, [disableEncounterButton, t, store, insuranceProvidersData]);
 
   const [selectedOption, setSelectedOption] = useState<
     SplitButtonOption<PatientModal> | undefined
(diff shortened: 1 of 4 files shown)


client/packages/system/src/Patient/PatientView/PatientDetailView.tsx
GeronimoJohn, line 1: Extracted `PatientDetailView` into it's own file as `PatientView` can be a bit overwhelming to navigate

client/packages/system/src/Patient/PatientView/AddButton.tsx
```diff
@@ -5,77 +5,98 @@ import {
...
+        }
+      );
+    }
+
+    if (InsuranceProvidersData?.length > 0) {
+      baseOptions.push({
+        value: PatientModal.Insurance,
+        label: t('button.add-insurance'),
+      });
+    }
```
GeronimoJohn, line 66: Added to support that case where we can add an insurance policy for a patient. Even though Programs Module is off.

client/packages/system/src/Patient/PatientView/AddButton.tsx
```diff
@@ -5,77 +5,98 @@ import {
...
+      ),
+      [PatientTabValue.ContactTracing]: options.find(
+        option => option.value === PatientModal.ContactTraceSearch
+      ),
+      [PatientTabValue.Insurance]: options.find(
+        option => option.value === PatientModal.Insurance
+      ),
+    };
+
+    setSelectedOption(optionMap[currentUrlTab as string] || options[0]);
```
CarlosNZ, line 98: Any particular reason to change from the `switch` block?

I like the "map" approach in general, but in this case it seems unnecessarily complicated (with the `.find()`in every value.

client/packages/system/src/Patient/PatientView/AddButton.tsx
```diff
@@ -92,16 +113,18 @@ export const AddButton: React.FC<AddButtonProps> = ({
 
   return (
     <>
```
CarlosNZ, line 115: I'm not usually a fan of the single node inside a fragment (99% of the time it's a mistake), but this might be cleaner in this case. I'd probably go for a ternary here (`return condition ? <SplitButton> : null`), but you decide which you think is tidier.

CarlosNZ (commented):
Looking good, @GeronimoJohn -- working as intended. Couple of minor stylistic comments, up to you really.

client/packages/system/src/Patient/PatientView/AddButton.tsx
```diff
@@ -5,77 +5,98 @@ import {
...
+          isDisabled: disableEncounterButton,
+        },
+        {
+          value: PatientModal.ContactTraceSearch,
+          label: t('button.add-contact-trace'),
+        }
+      );
+    }
+
+    if (InsuranceProvidersData?.length > 0) {
```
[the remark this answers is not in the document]
GeronimoJohn: woops missed that, usually am for camelCase as well

client/packages/system/src/Patient/PatientView/AddButton.tsx
GeronimoJohn: I initially tried sticking with the existing `switch` statement, but due to the dynamic nature of the options array, which now change based on conditions, accessing the values by their index pretty much was not reliable. Before we could do this:

`const [programOption, encounterOption, contactTraceOption, insuranceOption] = options`

And directly pass values in the `switch` statement. But now, the order of options can change. Meaning `programOption` or index 0 might actually be an insurance option. So I thought map might be a better approach.

Though very open for thoughts/improvements. I do think theres a bit going on in the map

GeronimoJohn:
Thanks @CarlosNZ 😃, worked on those changes: https://github.com/msupply-foundation/open-msupply/pull/6910/commits/6534b426bab36869a76420609257c0adb6d99ba7

client/packages/system/src/Patient/PatientView/AddButton.tsx
CarlosNZ: Maybe it'd be a little nicer to just have a static map between PatientTabValue and PatientModal value, and then just one "find" function, something like:
```ts
const optionMap: {
      [key: PatientTabValue]: SplitButtonOption<PatientModal> | undefined;
    } = {
      [PatientTabValue.Programs]: PatientModal.ProgramSearch,
      [PatientTabValue.Vaccinations]: PatientModal.ProgramSearch,
      [PatientTabValue.Encounters]: PatientModal.Encounter,
      [PatientTabValue.ContactTracing]: PatientModal.ContactTraceSearch ,
      [PatientTabValue.Insurance]: PatientModal.Insurance,
    };

setSelectedOption(options.find(option => option.value === optionMap[currentTab])
```

That's untested, but you get the idea. (Also, I think you can properly type the `key` of the map and avoid having to use `as string`

CarlosNZ: Also, you should be able to define the optionMap outside the component to avoid recreating a new instance every render.

GeronimoJohn: So much cleannerrrr. Thanks @CarlosNZ please see changes here: https://github.com/msupply-foundation/open-msupply/pull/6910/commits/98db54624fdb647b6665265e4dcb656bf87ef26a

I've tested and seems to be running the same. And I also tried moving `optionsMap` outside of the component, but JavaScript threw an error indicating that it couldn't access values in it before they were initialised or populated.

CarlosNZ: > And I also tried moving `optionsMap` outside of the component, but JavaScript threw an error indicating that it couldn't access values in it before they were initialised or populated.

Even if the `optionsMap` is up the top? (just under the `interface`)

CarlosNZ (approved):
LGTM! Thanks @GeronimoJohn 

(P.S. I think you could put the `optionsMap` outside the function if it were at the top, but it's not a big deal either way)

GeronimoJohn:
> LGTM! Thanks @GeronimoJohn
> 
> (P.S. I think you could put the `optionsMap` outside the function if it were at the top, but it's not a big deal either way)

Thanks @CarlosNZ, I gave it a go again of moving `optionsMap` above and under the `interface`. Yea it seems to be `undefined` during initialisation which make sense I think. Then gets populated after.  

Screenshot 2025-03-14 at 11 42 45 AM

Error it returns when defined outside of component

Screenshot 2025-03-14 at 11 31 05 AM

Hoping it's not just my machine kinda thing 😅

PR #6917, opened 2025-03-12
6391 date given does not update when vaccine is changed to not given

### LINKED ISSUE #6391: Date given does not update when vaccine is changed to Not Given
## What went wrong? 😲

Noticed the following: 

https://github.com/user-attachments/assets/580d0e84-9092-44c2-a7fc-84e21729d404

When changing the status of the vaccine - given vs. not given the `Date given` column stays populated even when the vaccine is not given?

And shouldn't 'Vaccine given' field be required before saving?

## Expected behaviour 🤔

Date given column should reset if the vaccine has been marked as `Not given`
And I think Vaccine given field should be compulsory, otherwise what's the point of recording a vaccination?

## How to Reproduce 🔨

1. Make sure you have Immunization Program and Vaccination Card enabled in your store
2. Once you do, create a new encounter for a patient and issue some vaccines
3. Test 1 specific vaccine - mark it as Given and then as Not given and observe the changes in a date given column
4. Also note you can record a vaccination without selecting a vaccination item

## Your environment 🌱

- Open mSupply Version: V2.6.0-build-test-3

- Database type: PostgreSQL

roxy-dao:
Fixes #6391

# 👩🏻‍💻 What does this PR do?
- Migration to recreate table to make vaccination_date nullable for sqlite since it doesn't have an ALTER COLUMN method
- Update update endpoint to accept nullable date input
- Amend hooks on frontend to not save/unset vaccination_date if vaccine not given

# 🧪 Testing

- [ ] Have vax set up
- [ ] Enrol patient in vax program
- [ ] Create an encounter for vax program
- [ ] Set 1 dose to 'Not given'
- [ ] Given date shouldn't save/show
- [ ] Set 1 dose to 'Given', save, then set to 'Not Given' 
- [ ] Given date shouldn't show anymore

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/vaccination/get_vaccination_card.rs
```diff
@@ -84,7 +84,11 @@ pub fn get_suggested_date(
     course_rows: Vec<VaccinationCardRow>,
 ) -> Option<NaiveDate> {
     let suggested_date_by_age = patient_dob
-        .map(|dob| dob.checked_add_signed(Duration::days((row.min_age * NUMBER_OF_DAYS_IN_A_MONTH) as i64)))
+        .map(|dob| {
```
roxy-dao, line 87: Sidequest

server/repository/src/migrations/v2_06_01/change_vaccination_date_to_nullable.rs
```diff
@@ -0,0 +1,49 @@
...
+                );
+                INSERT INTO vaccination SELECT * FROM vaccination_old;
+                DROP TABLE vaccination_old;
+            "#
+            )?;
+        } else {
+            sql!(
+                connection,
+                r#"
+                    ALTER TABLE vaccination ALTER COLUMN vaccination_date DROP NOT NULL;
```
aimee-mcneil-melville, line 42: Just for learning here - my understanding of this is that if there is postgres, we change the table name -> copy the data to a new table with the original name (with the new the column type) -> drop the old table -> use the new one as usual
And if it's sql only, we can just change the column type to 'not null' without needing the new table?

roxy-dao: Other way around 😉  sql doesn't support ALTER COLUMN, so we have to recreate the whole table

jmbrunskill: It would be clearer if we had `if isPostgres { do postgres stuff} else { do sqlite stuff }` rather than `if !isPostgres`

BTW: If the table has foreign keys referencing it, I think we might need to use  `PRAGMA foreign_keys = OFF;` before the changes and then `PRAGMA foreign_keys = ON;` otherwise the foreign keys might not point to the right place. But in this case we're fine as there's not foreign keys pointing the vaccination table.

roxy-dao: There's FK on name, user and vaccine_course_dose 😆

jmbrunskill (approved):
Thanks Roxy, All works for me!

PR #6923, opened 2025-03-13
Program Internal Orders Fixes

### LINKED ISSUE #6148: When creating an program internal order, the number of instances of the same supplier shown in the supplier list is sometimes greater than 1 and the number change when you type to search
## What went wrong? 😲

For example, 
I have 1 supplier that appears 3 times in the supplier list I can choose from when creating an internal order If I start typing the first letters from that supplier the instances shown in the list from that supplier grows to 13. But it happens only to this supplier. 
https://drive.google.com/file/d/1UyGlkIdRk_k7Qeqzti8xHvU9V1t45Vvd/view?usp=sharing

There is another supplier that have 2 instance in the list and it will go between 2 to 3 instances while your typing to search.   
https://drive.google.com/file/d/1VJdGi8FpV_kePyZX4ZcPppUkhdUz34js/view?usp=sharing

It seems to be ok for all other suppliers on the list

## Expected behaviour 🤔
Should see never more than 1 instance of the same supplier at any time.

## How to Reproduce 🔨

Have a store similar to the one in the video
Follow the example video

## Your environment 🌱

- Legacy mSupply Central Server Version: 2.5.0RC4

- Platform: Android 14 UI 6.1

- Database type: SQLite3

### LINKED ISSUE #6922: Internal order create program modal should show full name for options
## Expected behaviour 🤔

While working on CIV, noticed they have long program names and store names, which are not fully visible in internal order creation modal

## How to Reproduce 🔨

Use civ data file, try create internal order. Or just create long program master list names

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #6148 
Fixes #6922 

Re-orders the Program Requisition Selection to first select the Supplier, then filter program by available programs for that supplier.

# 👩🏻‍💻 What does this PR do?

This one got a bit out of hand, and had some confusion...

The main bug fix is de-duplicating the supplier list on program requisitions.
The issue was introduced when we started matching both masterlists and elmis codes, this meant that the same supplier could show up multiple times in the list.

I fixed that with this commit, that I accidentally commited to the RC branch, so doesn't show in the diff.
https://github.com/msupply-foundation/open-msupply/commit/2654541670b0344fd703acfd9ba0b92f5d3b5ed1
If there's any issues with it, then we can fix it here.

I also as requested by @alainsussol I also changed the modal to select the supplier first then the program, as this is consistent with mSupply and easier to understand/test.

HOWEVER, I've done it in a hacky way in the frontend for now, I think if we stick with this, the backend should be re-factored (and hopefully simplified?) to match.
I'll create a re-factor issue if we agree to merge this.

Finally I've remove the `no-wrap` on the program name selection to fix #6922 
This now shows as multi line, which looks good to me, is there somewhere else we need to check, was the no-wrap in for a good reason.

Alain's video showed the list of suppliers matched growing while typing. I can't replicate this after the changes I've made but I'm worried there could be some kind of react state bug...

# 🧪 Testing

- [ ] Setup program requisitions with the elmis code linking the same supplier to multiple programs
- [ ] Create a new internal order
- [ ] Check the correct programs are shown after you select a supplier
- [ ] Check that clearing or changing selections correctly clears the related fields.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/ListView/ProgramRequisitionOptions.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/ListView/ProgramRequisitionOptions.tsx
@@ -70,16 +70,36 @@ const useProgramRequisitionOptions = (
   const [period, setPeriod] = useState<Period | null>(null);
   const t = useTranslation();
 
+  const handleSetSupplier = (value: Supplier | null) => {
+    setSupplier(value);
+    setProgram(null);
+    setOrderType(null);
+    setPeriod(null);
+  };
+
   const handleSetProgram = (value: ProgramSetting | null) => {
     setProgram(value);
     setOrderType(null);
-    setSupplier(null);
     setPeriod(null);
   };
   const handleSetOrderType = (value: OrderType | null) => {
     setOrderType(value);
     setPeriod(null);
   };
+
+  const allSuppliers = programSettings
+    .flatMap(program => program.suppliers)
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .reduce((acc: Supplier[], currentValue: Supplier) => {
+      if (acc.length === 0 || acc[acc.length - 1]?.id !== currentValue.id) {
+        acc.push(currentValue);
+      }
+      return acc;
+    }, []);
+
+  const availablePrograms = programSettings.filter(program =>
+    program.suppliers.some(supplier => supplier.id)
+  );
 
   const allOptions: {
     programs: Common<ProgramSetting>;
@@ -88,7 +108,7 @@ const useProgramRequisitionOptions = (
     periods: Common<Period>;
   } = {
     programs: {
-      options: programSettings,
+      options: availablePrograms,
       value: program,
       set: handleSetProgram,
       label: t('label.program'),
@@ -104,10 +124,9 @@ const useProgramRequisitionOptions = (
       renderOption: getOrderTypeRenderer(),
     },
     suppliers: {
-      options: program?.suppliers || [],
+      options: allSuppliers || [],
       value: supplier,
-      set: setSupplier,
-      disabled: program === null,
+      set: handleSetSupplier,
       labelNoOptions: t('messages.not-configured'),
       label: t('label.supplier-name'),
       renderOption: getNameOptionRenderer(t('label.on-hold')),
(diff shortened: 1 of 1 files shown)


roxy-dao (changes requested):
LGTM~ I'm not sure about moving the suppliers to the top tho... In OG, supplier selection is at the top, but you can still select program before supplier. Maybe something for @mark-prins ? Or @andreievg to answer?

mark-prins:
iirc the suppliers are at the top in mSupply because we used them to determine programs available. I may be back-projecting that.. it was a while ago.

the 'nowrap' was likely because having names wrap is ugly and single line is nice. do you have a screenshot of how it looks?

alainsussol:
> iirc the suppliers are at the top in mSupply because we used them to determine programs available. I may be back-projecting that.. it was a while ago.
> 
> the 'nowrap' was likely because having names wrap is ugly and single line is nice. do you have a screenshot of how it looks?

It is my memory too except tfor the following distinction: for internal orders, it doesn't matter if you choose the program first because it is your version of the programs that you will see (whomever your supplier might be) as you are ordering for yourself. That's probably why you can choose program before supplier in internal orders.

In manual requisition it is another matter as your are basically creating the order that your customer should have sent you through an internal order if he could, or wanted. So you need to choose your customer version of a program but you cannot do that if you don't know the customer you are ordering for. That is why when you create a manual requisition in OG, you cannot choose a program until you have chosen the customer you are ordering for.

image

roxy-dao:
We also have to think about current clients who are already using the program module as well, would it be confusing for them if the fields are swapped around?

alainsussol:
> We also have to think about current clients who are already using the program module as well, would it be confusing for them if the fields are swapped around?

Yes that is also true...

jmbrunskill:
> do you have a screenshot of how it looks?

Sent :)

jmbrunskill:
@roxy-dao Mark's said he's ok with this change. Are you happy to approve?

roxy-dao (approved):
LGTM~ Saw the issue for backend implementation too thanks ⚡

PR #6926, opened 2025-03-13
6915 Target MOS is slightly different after migration

### LINKED ISSUE #6915: Target MOS is slightly different after migration
## Expected behaviour 🤔

Matches exactly after migration from OG

## How to Reproduce 🔨

Create internal order with OG, migrated to omSupply, see the difference

## Suggested fix

Use 30 rather then NUMBER_OF_DAYS_IN_A_MONTH when converting between daysToSupply

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #6915

# 👩🏻‍💻 What does this PR do?
Use 30 instead of days in month const. Min threshold seems to be correct and doesn't need changing

# 🧪 Testing

- [ ] Have a store in OG site
- [ ] Create IO for that store, adjust threshold/target MOS
- [ ] Transfer store to OMS site
- [ ] Sync
- [ ] Check threshold/target MOS match OG record

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/translations/requisition.rs
+++ b/server/service/src/sync/translations/requisition.rs
@@ -7,7 +7,9 @@ use repository::{
 };
 
 use serde::{Deserialize, Serialize};
-use util::constants::{MISSING_PROGRAM, NUMBER_OF_DAYS_IN_A_MONTH};
+use util::constants::{
+    MISSING_PROGRAM, MSUPPLY_NUMBER_OF_DAYS_IN_A_MONTH, NUMBER_OF_DAYS_IN_A_MONTH,
+};
 
 use crate::sync::{
     sync_serde::{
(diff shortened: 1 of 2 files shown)


server/service/src/sync/translations/requisition.rs
```diff
@@ -269,7 +269,7 @@ impl SyncTranslation for RequisitionTranslation {
                 date_and_time_to_datetime(data.date_entered, 0),
                 from_legacy_sent_datetime(data.last_modified_at, &r#type),
                 from_legacy_finalised_datetime(data.last_modified_at, &r#type),
-                data.daysToSupply as f64 / NUMBER_OF_DAYS_IN_A_MONTH,
+                data.daysToSupply as f64 / 30.0,
```
jmbrunskill, line 272: Might be nice to have a `MSUPPLY_NUMBER_OF_DAYS_IN_A_MONTH` constant or something?
Feels slightly sad to use a less accurate number just to align with mSupply.

jmbrunskill (approved):
I'm seeing some rounding issues, which probably isn't a blocker but maybe a follow up?

Image


Also for my example OMS seems to have a target stock  for the second item but legacy doesn't?

image

image


Not sure why that would be?

PR #6933, opened 2025-03-13
6848 upgrade asset labels

### LINKED ISSUE #6848: Modify Open mSupply to be a tool to generate & print 2D DataMatrix labels(for CCE that don't come with a 2D DataMatrix provided by the manufacturer - simple ID only)
Currently when editing viewing a cold chain asset, there's an option to print an Asset Label to a Zebra printer
Image

This currently sends a `^BQ` command to the configured Zebra Printer to print a QR code (see`pub fn print_qr_code(` in the rust code)

To change this to create a datamatrix we need to use the `^BX` command again.

https://support.zebra.com/cpws/docs/zpl/BX_Command.pdf

**Acceptance Criteria**

- [ ] Print Asset Label button prints a 2d barcode instead of a QR code
- [ ] QR code functionality is retained in the code for future use.

zachariah-at-msupply:
Fixes #6848 and #7065

# 👩🏻‍💻 What does this PR do?

This PR upgrades the asset label that open mSupply prints by:
- changing the format of the code printed on an asset label from a QR code to a data matrix barcode.
- adding a date printed field 
- adding the mSupply logo

Before:

label-before

After:

telegram-cloud-photo-size-1-5100522897113460621-y

1. The template literal I created on line 41 of client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx is very long:
    - This seems to be industry best practice for template literals so I left it on one line ([AirBnB](https://github.com/airbnb/javascript#strings--line-length) and [Google](https://google.github.io/styleguide/jsguide.html#features-strings-no-line-continuations) style guides recommend this).
    - It will also cause unwanted new lines in the output label if it is broken onto multiple lines, so we would need to do some further modification (e.g. changing the printing method) to achieve this.

2. Line 31 and 32 of server/service/src/print/label.rs define a hard-coded mSupply logo:
    - The logo is the official Open mSupply logo foudn on our cloud storage. I was unable to find a brand book or other document with our rules for use of the logo, but I think it should be fine.
    - It's easy to update it, so let me know if that would be good, and I have written a procedure for this, but need clarification about where that should go in our docs. 
    - If it is desirable for the code to reference a file rather than a hardcoded value, my research indicates we would probably need to convert the file into ZPL using our code, which will not be straightforward, and is why I used a hardcoded value.

3. There's a comment in server/service/src/print/label.rs about how to modify the code to print a QR code instead of a data matrix barcode. This is my interpretation of what the issue was aiming at when it said to keep the ability to print a QR code - just let me know if something else would be better!

# 🧪 Testing

- [ ] In Open mSupply, go to coldchain > equipment and select an asset (or make one if there are none), and the asset's detail view should appear.
- [ ] Click 'print label' in the top right.
- [ ] If mSupply tells you that you need to set up a printer, the details as of 11th March 2025 are IP = 192.168.1.224 and port = 9100.
- [ ] Check that the label looks like the 'after' picture above and has all the bits it should have:
  - data matrix barcode
  - serial number
  - asset number
  - date printed
  - mSupply logo
- [ ] Check the information on the label matches the information that Open mSupply displays.
- [ ] Check it scans, either by using mSupply android or a 2d barcode scanner, or a barcode scanning website (although that's a bit dodgy as you need to give them permission to use your camera).

# 📃 Documentation

  - Any reference to an asset's QR code should be changed to refer to a data matrix barcode instead.
  - Any reference to the content of assets' labels should be updated to include the new content.
  - Any images of assets' labels should be updated to be of the new label.
  - Any images of assets' QR codes should be updated to be data matrix barcodes instead (they look like this:)

image

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [x] SQLite
- [x] Frontend


adamdewey:
@zachariah-at-msupply - is it possible to put the text information and logo below the data matrix?

That way we can print the data matrix as large as possible

There are concerns that some phone cameras have issues scanning the data matrix when it's small

Thanks

jmbrunskill:
Lets remove serial numbe as it's already included on the device

zachariah-at-msupply:
> @zachariah-at-msupply - is it possible to put the text information and logo below the data matrix?
> 
> That way we can print the data matrix as large as possible
> 
> There are concerns that some phone cameras have issues scanning the data matrix when it's small
> 
> Thanks

Hi Adam, absolutely, I can do that!

client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx
```diff
@@ -32,14 +32,13 @@ export const AppBarButtonsComponent = () => {
...
+    const date = new Date().toLocaleString().split(', ')[0];
     setIsPrinting(true);
     fetch(Environment.PRINT_LABEL_QR, {
       method: 'POST',
       body: JSON.stringify({
         code: data?.id,
-        message: `${t('label.serial')}: ${data?.serialNumber ?? ''}\n${t(
-          'label.asset-number'
-        )}: ${data?.assetNumber ?? ''}`,
+        message: `${t('label.serial')}: ${data?.serialNumber ?? ''}\n${t('label.asset-number')}: ${data?.assetNumber ?? ''}\n${t('label.date-printed')}: ${date}`,
```
jmbrunskill, line 41:
```suggestion
        message: `${t('label.asset-number')}: ${data?.assetNumber ?? ''}\n${t('label.date-printed')}: ${date}`,
```
After discussion with @adamdewey we think it's best just to have the asset number as they're probably going to be set to the same thing, and serial number should already be on the device

jmbrunskill:
> > @zachariah-at-msupply - is it possible to put the text information and logo below the data matrix?
> > That way we can print the data matrix as large as possible
> > There are concerns that some phone cameras have issues scanning the data matrix when it's small
> > Thanks
> 
> Absolutely, I can do that! I think that the way to make the barcode as large as possible may be to keep it on one side however - like this: image
> 
> What do you think?

Looks great

adamdewey:
Yep that looks great @zachariah-at-msupply 🔥

zachariah-at-msupply:
I updated the PR's initial comment with this update:
- The logo is now a big, proud, proper Open mSupply logo - huzzah!
- The label now looks like this:
telegram-cloud-photo-size-1-5100522897113460621-y

client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx
```diff
@@ -32,14 +32,14 @@ export const AppBarButtonsComponent = () => {
   };
 
   const printAssetLabel = () => {
+    const date = new Date().toLocaleString().split(', ')[0];
```
jmbrunskill, line 35:
```suggestion
    const date = new Date().toLocaleDateString();
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

I think this is a safer thing to do, I'm not sure if all languages will work with the split function...

jmbrunskill (changes requested):
Just that one quick fix for the `toLocalDateString()` function, otherwise looking good!

jmbrunskill (approved):
Thanks!

jmbrunskill (changes requested):
Actually I'm not sure we should check in the .idea folder.
I'm open to the idea but want to check if it might cause problems...

server/android/.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
+/shelf/
```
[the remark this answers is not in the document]
jmbrunskill: Actually maybe we just replace this with 
https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
?

I've had a read of https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems which seems to suggest it's ok to check in what you're checking in here...

server/android/.idea/caches/deviceStreaming.xml
jmbrunskill, line 1: Pretty sure we don't want to include anything under `caches` even through it's not mentioned in that article.
I'd ignore the caches directory too, unless you know a good reason it should be included?

server/service/src/print/label.rs
zachariah-at-msupply, line 1: In response to cargo's frantic prompting, I made this change myself - but it seems the post request relies on these fields having the same name as their counterparts in the front end. As a result, this change means that the back end does not receive the asset number or date printed from the front end. What do you think would be best to do?

jmbrunskill (approved):
Looks good to me, made some updates to the docs. Can you double check I haven't said anything dumb...

server/service/src/print/label.rs
```diff
@@ -4,44 +4,53 @@ use crate::settings::LabelPrinterSettingNode;
...
         ^FX CI command parameters:
         ^FX - encoding (28 = UTF-8)
         ^CI28
-        ^FO50,{}
-        ^BQN,2,4
-        ^FDMA,{}^FS        
+        ^FO18,18
+        ^BXN,11,200
+        ^FD{}^FS
+        ^FO277,18^A0,32,25^FDAsset Number: {}^FS
```
jmbrunskill, line 35: One issue I've realised with your change is that we can't translate these fields.
At the moment I think it's ok, but in future we might need to find a way to translate this?

zachariah-at-msupply: Hmm, I refactored that to fix it, and uncovered some additional issues with translations:
- it seems that some translations have much longer ways of saying 'asset number' and 'date printed' than others. They push the asset number itself and the date off the label. 
- It seems that some languages, like Arabic, use characters that the printer cannot print in this configuration.

zachariah-at-msupply:
I updated the label to fix the problem with the translation, documented in the comment above and in issue [7065](https://github.com/msupply-foundation/open-msupply/issues/7065).
As Mark specified in that issue, the labels now look like this (without asset number or date printed labels):

telegram-cloud-photo-size-1-5127633731413191889-y

jmbrunskill (approved):
Looks great thanks @zachariah-at-msupply

zachariah-at-msupply:
And to think it only took 2 weeks in the end 😅

PR #6987, opened 2025-03-18
Ran prettier

GeronimoJohn:
Fixes #

# 👩🏻‍💻 What does this PR do?

Ran `yarn prettier -w .` against client. Successfully removes the big list of warning signs whenever we run `yarn prettier` or `yarn prettier -c .`. And formats the client directory to follow pattern set in `.prettierrc`.

**Removes warnings like these**

Screenshot 2025-03-19 at 12 41 58 PM

# 🧪 Testing

- [ ] Run `yarn prettier -c .` within your terminal for client directory
- [ ] No more warnings returned!

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


roxy-dao (changes requested):
Thanks... I don't like how it reformatted some of the generated files... why are we checking prettier in our actions instead of just using husky?

GeronimoJohn:
> Thanks... I don't like how it reformatted some of the generated files... why are we checking prettier in our actions instead of just using husky?

Good points!

Regarding Husky, we found that what it was initially takes time to finish before allowing a commit: https://github.com/msupply-foundation/open-msupply/issues/6746. The GH issue doesn't cover everything - essentially moved all the tasks from Husky to a GH Action. There's definitely room for improvement, and we could consider extracting parts of the GitHub Action back into Husky. 
Personally, I don't mind when Prettier and ESLint run, as long as it does. 

In terms of generated files. Prettier started getting applied to it at this PR: https://github.com/msupply-foundation/open-msupply/pull/6879. It seems that it might be the direction for how these files should be?

roxy-dao:
> > Thanks... I don't like how it reformatted some of the generated files... why are we checking prettier in our actions instead of just using husky?
> 
> Good points!
> 
> Regarding Husky, we found that what it was initially takes time to finish before allowing a commit: #6746. The GH issue doesn't cover everything, but essentially moved all the tasks from Husky to a GH Action. There's definitely room for improvement, and we could consider extracting parts of the GitHub Action back into Husky. Personally, I don't mind when Prettier and ESLint run, as long as it does.
> 
> In terms of generated files. Prettier started getting applied to it at this PR: #6879. It seems that it might be the direction for how these files should be?

Are there different rules then? Can we confirm that the same rule is being applied throughout... Not sure why it would miss some files whilst regenerating?

roxy-dao (approved):
Happy to get this through but we probably wanna ensure that everyone is using the same prettier rules 👀

GeronimoJohn:
> > > Thanks... I don't like how it reformatted some of the generated files... why are we checking prettier in our actions instead of just using husky?
> > 
> > 
> > Good points!
> > Regarding Husky, we found that what it was initially takes time to finish before allowing a commit: #6746. The GH issue doesn't cover everything, but essentially moved all the tasks from Husky to a GH Action. There's definitely room for improvement, and we could consider extracting parts of the GitHub Action back into Husky. Personally, I don't mind when Prettier and ESLint run, as long as it does.
> > In terms of generated files. Prettier started getting applied to it at this PR: #6879. It seems that it might be the direction for how these files should be?
> 
> Are there different rules then? Can we confirm that the same rule is being applied throughout... Not sure why it would miss some files whilst regenerating?

Apologies just saw this now. I can investigate more on that during cooldown. And agree that need to ensure we're all on the same prettier rules 😃

CarlosNZ:
> And agree that need to ensure we're all on the same prettier rules

They're defined in the repo, so should be. Only problem is some people don't have auto-format on Save enabled. 

When reviewing PRs, I tend to save at least one file and see if anything changes. If it does, I bitch to the developer to fix their Prettier config! 😆

roxy-dao:
> > And agree that need to ensure we're all on the same prettier rules
> 
> They're defined in the repo, so should be. Only problem is some people don't have auto-format on Save enabled.
> 
> When reviewing PRs, I tend to save at least one file and see if anything changes. If it does, I bitch to the developer to fix their Prettier config! 😆

Yeah, can one of you make an announcement for this in the group and also instructions of how to do for those who don't know 😉 cheers~!

PR #7008, opened 2025-03-20
4998 Hide time tab from datepicker

### LINKED ISSUE #4998: Prescription date time tab not clickable
## What went wrong? 😲

https://github.com/user-attachments/assets/b7e0342d-ca8a-4264-a182-dc1a52ede8c0

## Expected behaviour 🤔

If I can't enter a time, don't offer me a time tab 🥲 

## How to Reproduce 🔨

1. Create prescription
2. Click date
3. Try to click time tab in date picker

## Your environment 🌱

- Open mSupply Version: V2.3.0 RC5

- Platform: APK

noel-yeldos:
Fixes #4998 

# 👩🏻‍💻 What does this PR do?
Hide time tab from datetime picker in ```New Prescription``` and ```New Encounter``` modal.  

I made changes based on this: https://mui.com/x/react-date-pickers/custom-components/#tabs

```
By default, the tabs are hidden on desktop, and visible on mobile. This behavior can be overridden by setting the hidden prop
```

I set the ```hidden``` prop to ```true``` for both prescriptions and encounters. 

Thanks @aimee-mcneil-melville for brainstorming! 

Screenshot 2025-03-21 at 11 02 13
Screenshot 2025-03-21 at 11 02 36

I tried to set the ```hidden``` prop to ```true``` for the base ```DateTimePickerInput.tsx``` component but that change wasn't coming through for the new encounter datepicker.

@mark-prins I wasn't quite sure how to implement your suggestion of using the ```showTime``` prop to display either ```DateTimePicker``` or ```DatePicker```

Also are we allowing users to set time on any of our date pickers? My understanding was that this was allowed on encounters but this no longer seems to be the case?

# 🧪 Testing

```Prescriptions```

- [ ]  Go to Dispensary -> Prescriptions
- [ ] Create new prescription
- [ ] Click on date picker
- [ ] Observe that time tab no longer displays

```Encounters```

- [ ]  Go to Dispensary -> Patients
- [ ] Ensure you have a patient enrolled in a program
- [ ] Click on that patient -> navigate to encounters tab
- [ ] Create new encounter
- [ ] Click on date picker
- [ ] Observe that time tab no longer displays

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -147,6 +147,9 @@ export const DateTimePickerInput: FC<
           },
         },
         ...(actions ? { actionBar: { actions } } : {}),
+        tabs: {
+          hidden: true,
+        },
       }}
       views={
         showTime
--- a/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -147,9 +147,6 @@ export const DateTimePickerInput: FC<
           },
         },
         ...(actions ? { actionBar: { actions } } : {}),
-        tabs: {
-          hidden: true,
-        },
       }}
       views={
         showTime
--- a/client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx
+++ b/client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx
@@ -174,11 +174,6 @@ export const NewPrescriptionModal: FC<NewPrescriptionModalProps> = ({
                   onChange={newDate => {
                     if (newDate) setDate(newDate);
                   }}
-                  slotProps={{
-                    tabs: {
-                      hidden: true,
-                    },
-                  }}
                 />
               }
             />
--- a/client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx
+++ b/client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx
@@ -174,6 +174,11 @@ export const NewPrescriptionModal: FC<NewPrescriptionModalProps> = ({
                   onChange={newDate => {
                     if (newDate) setDate(newDate);
                   }}
+                  slotProps={{
+                    tabs: {
+                      hidden: true,
+                    },
+                  }}
                 />
               }
             />
--- a/client/packages/system/src/Patient/Encounter/CreateEncounterForm.tsx
+++ b/client/packages/system/src/Patient/Encounter/CreateEncounterForm.tsx
@@ -112,6 +112,9 @@ export const CreateEncounterForm: FC<{
               day: {
                 highlightedDays,
               } as BadgePickersDayProps,
+              tabs: {
+                hidden: true,
+              },
             }}
           />
         }

### CODE REVIEW

File: client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx [hunk 1]
```diff
@@ -174,6 +174,11 @@ export const NewPrescriptionModal: FC<NewPrescriptionModalProps> = ({
                   onChange={newDate => {
                     if (newDate) setDate(newDate);
                   }}
+                  slotProps={{
+                    tabs: {
+                      hidden: true,
+                    },
+                  }}
```
roxy-dao, line 181: I can't seem to reproduce this, but I imagine what was asked in the original issue would suffice? Did declaring `showTime={false}` still show time selection?

https://github.com/msupply-foundation/open-msupply/issues/4998#issuecomment-2403916371

roxy-dao (changes requested):
Thanks for the changes Noel. Just a question below (:

client/packages/invoices/src/Prescriptions/ListView/NewPrescriptionModal.tsx
noel-yeldos: Yea the only way to hide the time tab on mobile seems to be to explicitly hide tabs altogether. The showTime prop seems to only hide or show the actual time being displayed.

```showTime={true}``` and remove ```tabs: { hidden: true}```
Screenshot 2025-03-26 at 09 51 56

```showTime={false}``` and remove ```tabs: { hidden: true}```
Screenshot 2025-03-26 at 10 01 25

With ```tabs: { hidden: true}``` enabled
Screenshot 2025-03-26 at 10 02 04

**Comment by noel-yeldos** [comment]
@roxy-dao should I also do this in the common component rather than on a case by case basis? Only caveat is in the "Notes for reviewer"

Similar to what you've done here: https://github.com/msupply-foundation/open-msupply/pull/7052/commits/740b598df3c48a719a3322827730f4df246deef7

**Comment by roxy-dao** [comment]
> @roxy-dao should I also do this in the common component rather than on a case by case basis? (Only caveat is in the "Notes for reviewer")
> 
> Similar to what you've done here: [740b598](https://github.com/msupply-foundation/open-msupply/commit/740b598df3c48a719a3322827730f4df246deef7)

Yes, would be nicer in the common component, but will need a way to tell the difference between when it is supposed to be date and when it is supposed to be datetime

File: client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx [hunk 2]
```diff
@@ -147,6 +147,9 @@ export const DateTimePickerInput: FC<
           },
         },
         ...(actions ? { actionBar: { actions } } : {}),
+        tabs: {
+          hidden: true,
+        },
```
noel-yeldos, line 152: Have added this prop in the common component to hide tabs. Hidden by default on desktop, this covers mobile.

client/packages/system/src/Patient/Encounter/CreateEncounterForm.tsx
```diff
@@ -112,6 +112,9 @@ export const CreateEncounterForm: FC<{
               day: {
                 highlightedDays,
               } as BadgePickersDayProps,
+              tabs: {
+                hidden: true,
+              },
```
noel-yeldos, line 117: For some reason the change in the common component does not get applied to CreateEncounterForm, have to explicitly hide here.

client/packages/common/src/ui/components/inputs/DatePickers/DateTimePickerInput/DateTimePickerInput.tsx
roxy-dao: Please revert this... we don't want to hide it all the time... 

Just merge with previous changes please

noel-yeldos: But it's hidden on default for desktop anyway right according to docs so aren't we just applying that change to mobile as well?

roxy-dao: This is a datetime picker... if we want to pick time then it should be available?

noel-yeldos: Yup I'm following. I'm going off the below (linked in PR comment)
Screenshot 2025-03-28 at 10 04 57

By their logic it's always hidden on desktop anyway right even for datetime picker even before my change? So even if we wanted to show tab in desktop we would have to set the prop to false in the common component or apply it individually.

roxy-dao: Visible for mobile means to me that they need the tab to pick time? So what happens if you want a date AND time picker on mobile with the tabs hidden?

noel-yeldos: We would have to override on individual area like so: 
``` 
     <DateTimePickerInput
                  value={DateUtils.getDateOrNull(date)}
                  format="P"
                  onChange={newDate => {
                    if (newDate) setDate(newDate);
                  }}
                  slotProps={{
                    tabs: {
                      hidden: false,
                    },
                  }}
                  showTime
                />
```

I think we need the showTime prop as well to be able to select time. Otherwise we just get a disabled time tab if we don't set the hidden prop at all

noel-yeldos: Agreed to revert to original changes

client/packages/system/src/Patient/Encounter/CreateEncounterForm.tsx
noel-yeldos: Comment no longer relevant as keeping original changes where hidden prop was being set explicitly anyway

PR #7012, opened 2025-03-21
Automatically create a location for new CCE assets

### LINKED ISSUE #7001: Automatically Create a location on Asset Creation
When creating an asset in OMS, we should automatically create a location for it to use.

- [x] Location name will be the asset number
- [x] Cold storage type can be null (for now)
- [x] The Cold storage location should be associated with the asset automatically.

jmbrunskill:
Fixes #7001

# 👩🏻‍💻 What does this PR do?

Automatically creates a location for any newly created assets.

image

Location is named after the assetNumber.

I feel like it would be nice to have an asset `Name` field that we could use here?
Having a name like Fridge 1 would be much nicer than a location named asset ABCE1234 I don't think it would be meaningful.

@mark-prins @adamdewey  what do you think about that idea?

We also create an initial status log for the asset in the frontend code. Should we move this to the backend?
I suspect it might have been done this way to allow using the translations in the comment. I feel like we could probably just leave the comment blank? Or maybe it's ok how it is...

# 🧪 Testing

- [ ] Create a Cold Chain Equipment (test via import and manually)
- [ ] Check that a location is automatically created for the asset.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [x] Postgres
- [x] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/asset/insert.rs
+++ b/server/service/src/asset/insert.rs
@@ -1,4 +1,5 @@
 use super::{
+    location::set_asset_location,
     query::get_asset,
     validate::{check_asset_exists, check_asset_number_exists},
 };
@@ -12,8 +13,10 @@ use repository::{
         asset::{AssetFilter, AssetRepository},
         asset_row::{AssetRow, AssetRowRepository},
     },
-    ActivityLogType, RepositoryError, StorageConnection, StringFilter,
+    migrations::constants::COLD_CHAIN_EQUIPMENT_UUID,
+    ActivityLogType, LocationRow, RepositoryError, StorageConnection, StringFilter, Upsert,
 };
+use util::uuid::uuid;
 
 #[derive(PartialEq, Debug)]
 pub enum InsertAssetError {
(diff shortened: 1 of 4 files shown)


server/service/src/asset/insert.rs
```diff
@@ -70,9 +73,32 @@ pub fn insert_asset(
                 }
                 None => input,
             };
+
             let new_asset = generate(input);
             AssetRowRepository::new(connection).upsert_one(&new_asset)?;
 
+            // Automatically create a location for this asset (if it's a cold chain asset)
+            if new_asset.asset_class_id == Some(COLD_CHAIN_EQUIPMENT_UUID.to_string()) {
+                if let Some(store_id) = new_asset.store_id {
```
fergie-nz, line 82: I'm wondering if user documentation is needed to state that locations are only automatically created when creating an asset at that store given that we can manage assets across all stores from OMS central server 

(Or can we just make locations for different stores automatically anyway without this check?)

fergie-nz (approved):
Works great nice one james!

Only thought I had was wondering about compatibility with our remote creation of CCE on central server. Did you have any thoughts on that?

jmbrunskill:
> Works great nice one james!
> 
> Only thought I had was wondering about compatibility with our remote creation of CCE on central server. Did you have any thoughts on that?

Great question, I think we'll need to test it but maybe shouldn't create locations for remote stores...

adamdewey:
> I feel like it would be nice to have an asset Name field that we could use here?
Having a name like Fridge 1 would be much nicer than a location named asset ABCE1234 I don't think it would be meaningful.

@jmbrunskill - I agree, that would be a good idea. 
Ideally we would default populate it with the model (plus a number if there's more than 1 of the same model in the same store) so that the user can still identify it easily even if they don't enter their own custom name

jmbrunskill:
@fergie-nz I've pushed a fix for the location creation on central server to not create location for remote stores.
Can you check again please?

PR #7029, opened 2025-03-24
#6920 Correct rounding for Currency input

### LINKED ISSUE #6920: Currency rounding should round to the nearest value, not round down
### Update

Below is the original issue as posted by @alainsussol. Following the discussion below, the problem seems to be that the rounding for currency values is rounding down, when it should be rounding to the nearest rounding unit, otherwise it creates the kind of discrepancies @alainsussol has described [below](https://github.com/msupply-foundation/open-msupply/issues/6920#issuecomment-2744179382).

-- @CarlosNZ 

## What went wrong? 😲

Calculation seems to be done properly but it looks strange  

Image

the insurance should pay 3.75=15 * 0.25 so the amount to pay would be 11.25 instead  of 11. If you look at the total amount of 15, 3 to pay by the insurance, 11 to pay by the client and 1 dollar disappearing, adding to the feeling of a global conspiracy. If you put 11 as amount paid, omsupply tells me that I didn`t pay enough and would not confirm the transaction unless I paid 12 which a change to give field showing 0 instead of 0.75.

But I am pretty sure that under the hood the calculations are good otherwise when I put 11 as amount paid, it will let me confirm the transaction but it does not because omsupply knows there is still 0.25 to pay. So I think the only problem is on displaying.

Like in the example below:

## Expected behaviour 🤔
2 digits Decimals needs to be displayed and maybe the currency (XOF) symbol near the amount

## How to Reproduce 🔨

Have the plugin for Ivory Coast installed
Have a patient with an insurance with a rate greater than 0.
1. Go to Dispensary > prescription
2. Click on Add a new prescription
3. Add your patient
4. Add an item that has a sell price > 0 and that Sell price * (rate/100) would give a number with a decimal. 
5. Try to confirm
6. A payment modal appear. Choose your patient insurance with your rate.
8. See error. The decimals are not displayed.

## Your environment 🌱

- Open mSupply Version: 2.6.1
- Legacy mSupply Central Server Version: 8.0.17

- Platform: OpenmSupply desktop

- Database type: SQLite

CarlosNZ:
Fixes #6920

# 👩🏻‍💻 What does this PR do?

This is just a quick fix to add some rounding to the value displayed. Internally the "react-currency-input-field" package we use only truncates, it doesn't round, which can lead to inconsistencies like the one highlighted in this issue.

I'm fairly sure this doesn't break anything else, but it'd be good to have a think and see if there are any places where this doesn't work so well.

# 🧪 Testing

- [ ] Have some Insurance policies connected to a patient, and create a prescription for them
- [ ] Add some items and go to "Verify", which should brind up the "Payment" modal
- [ ] You'll need to set up your prescription/policies in such a way that the "Paid By Insurance" field should be rounded up (e.g. 2.5877). You might need a little trial and error (and some manual calcuation) to find prescription amounts that cause this state.
- [ ] If using the Insurance Payment plugin, check that the "Total to be paid" minus the "Paid By Insurance" is exactly equal to "Amount to be paid", like this:
Screenshot 2025-03-24 at 2 14 03 PM
- [ ] If possible, test with currencies other than dollars (eg XOF which only deals in whole numbers, as per the original isssue)
- [ ] Confirm Currency inputs still behave as expected elsewhere (I think they're only used in Stock Buy/Sell prices)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -4,6 +4,7 @@ import RCInput, {
   CurrencyInputProps as RCInputProps,
 } from 'react-currency-input-field';
 import { useCurrency } from '@common/intl';
+import { NumUtils } from '@common/utils';
 
 interface CurrencyInputProps extends RCInputProps {
   onChangeNumber: (value: number) => void;
@@ -44,6 +45,9 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   const isSymbolLast = options.pattern.endsWith('!');
   const prefix = !isSymbolLast ? options.symbol : '';
   const suffix = isSymbolLast ? options.symbol : '';
+  const defaultValueAsNumber = Number.isNaN(Number(defaultValue))
+    ? undefined
+    : Number(defaultValue);
   const valueAsNumber = Number.isNaN(value) ? 0 : Number(value);
 
   return (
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
Nice Carl LGTM!

Issue #7014, opened 2025-03-21
Can't upload multiple files or photos in CCE status update


fergie-nz:
## What went wrong? 😲

There is not functionality for multiple photos or files in CCE status update.

## Expected behaviour 🤔

Can handle multiple files

## How to Reproduce 🔨

When using drop zone to upload a file where one is already uploaded, it will overwrite the original file or photo.

https://github.com/user-attachments/assets/d65be45e-ee81-43a2-b8db-19197624acc3

When using drop zone to upload multiple files, it will show on the file list in mobile, but only one photo is shown on the logs.

https://github.com/user-attachments/assets/be59820e-64da-47c4-bf04-cc05f8304ad4

## Your environment 🌱

- Open mSupply Version: 2.7.0
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

CarlosNZ:
Same basic problem as #6997 -- should fix both together (and anywhere else this happens)

fergie-nz:
> Same basic problem as [#6997](https://github.com/msupply-foundation/open-msupply/issues/6997) -- should fix both together (and anywhere else this happens)

You're right - I think it is an issue with how we are uploading with dropzone.

marthakeezy:
Same as #6997 testing - able to upload multiple and view them now. Checked with Adam if there should be an option to remove the uploaded photos (like in Documents), but he said we will wait for feedback from the field - so moving to Done ✅

PR #7040, opened 2025-03-24
#6997 #7014 Multiple file upload

### LINKED ISSUE #6997: Equipment document upload doesn't accept multiple file selection
## What went wrong? 😲

When you click the "Browse Files" button in the Documents section for an asset, you can select mulitiple items in the Finder. However, only one file is actually uploaded.

## Expected behaviour 🤔

It should *either*:
- upload multiple selected files, or
- restrict the File explorer to a single selection

Note: would pay to check other places that offer file uploads and make sure this is consistent throughout.

## How to Reproduce 🔨

- Go to Cold Chain -> Equipment -> Select an asset -> Documents tab
- Click "Browse files"
- Select multiple files (using Shift or Cmd), then click "OK"
- Notice only one file uploaded and listed

## Your environment 🌱

- Open mSupply Version: 2.6
- Platform: Browser/Firefox

CarlosNZ:
Fixes #6997 & #7014

# 👩🏻‍💻 What does this PR do?

Modifies the server's `upload_sync_file` endpoint function to handle an array of files rather than just the first one in the `formData` array.

# 🧪 Testing

- [ ] Go to ColdChain -> Equipment -> Select an asset -> Documents tab
- [ ] Drag and drop multiple files onto the Upload box
- [ ] *All* files should appear in the Downloaded documents list below
- [ ] Repeat by clicking the "Browse files" button and selecting multiple files in the Dialog

# 📃 Reviewer Checklist

**Breaking Changes**
- [x] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/server/src/static_files.rs
+++ b/server/server/src/static_files.rs
@@ -150,7 +150,7 @@ async fn upload_sync_file(
 
     let path_inner = path.into_inner();
 
-    let mut static_file_ids: Vec<String> = [].to_vec();
+    let mut static_file_ids: Vec<String> = vec![];
 
     for f in file {
         let static_file =
--- a/server/server/src/static_files.rs
+++ b/server/server/src/static_files.rs
@@ -39,7 +39,7 @@ use crate::middleware::limit_content_length;
 #[derive(Debug, MultipartForm)]
 pub(crate) struct UploadForm {
     #[multipart(rename = "files")]
-    pub(crate) file: TempFile,
+    pub(crate) file: Vec<TempFile>,
 }
 
 // this function could be located in different module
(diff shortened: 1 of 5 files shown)


server/server/src/static_files.rs
```diff
@@ -148,17 +148,26 @@ async fn upload_sync_file(
...
-        })?;
-
-    Ok(HttpResponse::Ok().json(static_file.id))
+    let path_inner = path.into_inner();
+
+    let mut static_file_ids: Vec<String> = [].to_vec();
+
+    for f in file {
+        let static_file =
+            upload_sync_file_inner(&service_provider, &settings, path_inner.clone(), f)
```
CarlosNZ, line 157: Was trying to pass a reference to `path` rather than cloning here, but the called function takes a tuple of values here, so wasn't sure what the correct syntax to pass reference to object that is being destructured.

jmbrunskill: FWIW, I tried this and the syntax is like this:   

 (table_name, record_id): &(String, String),

Then you pass it here as:

```suggestion
            upload_sync_file_inner(&service_provider, &settings, &path_inner, f)
```

However, I don't think it actually make is more efficient or simplier as you then need to call the `to_string()` or `to_owned()` methods on those variables in the `upload_sync_file_inner` function, so I think what you have is totally fine.

jmbrunskill (approved):
Thanks @CarlosNZ works great!

server/service/src/static_files.rs
```diff
@@ -70,11 +70,14 @@ impl StaticFileService {
...
         &self,
-        temp_file: TempFile,
+        temp_file: &TempFile,
         category: &StaticFileCategory,
         file_id: Option<String>,
     ) -> anyhow::Result<StaticFile> {
-        let file_name = temp_file.file_name.context("Filename not provided")?;
+        let file_name = temp_file
+            .file_name
+            .clone()
```
jmbrunskill, line 79:
```suggestion
            .to_owned()
```
It doesn't matter at all, but this might be slightly more correct, as you're using clone here to turn a reference into an `owned` value. The code underneath will probably be exactly the same but thought I'd mention anyway.

server/server/src/static_files.rs
```diff
@@ -148,17 +148,26 @@ async fn upload_sync_file(
...
-            InternalError::new(
-                "Error uploading file, please check server logs",
-                StatusCode::INTERNAL_SERVER_ERROR,
-            )
-        })?;
-
-    Ok(HttpResponse::Ok().json(static_file.id))
+    let path_inner = path.into_inner();
+
+    let mut static_file_ids: Vec<String> = [].to_vec();
```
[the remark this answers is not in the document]
CarlosNZ: 4deb62f62e

PR #7059, opened 2025-03-26
6301 add boajs as alternative report convert data

### LINKED ISSUE #6301: Replace extism with BoaJS

Need some additional context and input from @andreievg 

### Moneyworks Jobcode 🧰

OMS:REPT

andreievg:
Fixes #6301 

# 👩🏻‍💻 What does this PR do?

* Update CLI to to allow bundling either extism or boajs via another field in report_manifest (which defaults to Extism if not specified), i also moved generating convert_data within cli rather then report builder (because I couldn't share the new 'type') and shared Command utility to actiually check for failing command call status (by command i mean calling cli tools like yarn via rust)
* CLI to always yarn install (had issue where node modules were present and CLI doesn't yarn install), using lock file (but not comiting), which makes it fast
* Extracted boajs call_plugin to call_method, which is used within call_plugin and transform_data in reports
* Updated all standard reports to use boajs, moved version up to 2.6.1 for them
* Removed all other versions of standard report that use convert_data (since they don't sync anyways, and they will be present on instances that have them bundled in executable)
* Remove convert_data where it wasn't being used in report (report_manifest didn't specify convert_data
* Added full typescript example of report and updated README.md

Do we need to retest all report ? I don't think so, logically convert_data is the same and can see that in the diff (apart from item-usage report, which I tested)

# 🧪 Testing

When app server start reports should be re-installed, can do a quick check for standard reports, forms don't you convert_data as far as I can see

You will notice that app should start much faster (especially on android, doesn't have to load 27mb of json and parse etc..)

Reports will be much faster (didn't time it but something like from 4 second to instant), and initialisation should be quicker for remote site

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/cli/src/cli.rs
```diff
@@ -45,10 +45,11 @@ use util::inline_init;
 mod backup;
 use backup::*;
 
-mod plugins;
-use plugins::*;
-
-use cli::{generate_report_data, generate_reports_recursive, RefreshDatesRepository, ReportError};
+use cli::{
```
andreievg, line 48: Only changes in this file are the imports, everything else is auto format

server/cli/src/helpers.rs
```diff
@@ -0,0 +1,23 @@
+use std::process::{Command, ExitStatus, Stdio};
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum CommandError {
+    #[error(transparent)]
```
andreievg, line 7: this one was moved from call_plugin to be shared with report cli to actually test for failed status

server/cli/src/lib.rs
```diff
@@ -5,3 +5,9 @@ mod report_utils;
 pub use report_utils::*;
 mod graphql;
 pub use graphql::*;
+
+mod helpers;
+pub use helpers::*;
+
+mod plugins;
+pub use plugins::*;
```
andreievg, line 13: I had issues with imports, since plugins was being used directly in CLI, and once i started sharing the helper method there was some warnings ¯_(ツ)_/¯

server/cli/src/plugins.rs
```diff
@@ -16,11 +8,22 @@ use reqwest::Url;
 use serde::Deserialize;
 use serde_json::json;
 use service::backend_plugin::plugin_provider::PluginBundle;
+use std::{
```
andreievg, line 11: oops joined imports and it sorted on save

server/cli/src/plugins.rs
```diff
@@ -110,7 +113,7 @@ struct ManifestJson {
     om_supply_plugin: PluginDescription,
 }
 
-pub(crate) fn generate_plugin_bundle(
+pub fn generate_plugin_bundle(
```
andreievg, line 116: just one method to yarn install and convert_data, it's actually faster now without using --check-files and using lock files (but not comiting lock file)

server/report_builder/src/build.rs
```diff
@@ -101,8 +99,8 @@ fn make_report(args: &BuildArgs, mut files: HashMap<String, PathBuf>) -> Result<
         header: None,
         footer: None,
         query: vec![],
-        convert_data: None,
-        custom_wasm_function: None,
+        // Convert data is generated outside of this method call
```
andreievg, line 102: Everything will be moved to cli eventually, and i couldn't share ConvertDataType with report_builder, plus wanted to use the utility method to check for bad status

server/report_builder/src/lib.rs
```diff
@@ -41,13 +41,6 @@ pub struct BuildArgs {
     /// Default query type, one of: "invoice" | "stocktake" | "requisition",
     #[clap(long)]
     pub query_default: Option<String>,
-
-    #[clap(long)]
-    pub convert_data: Option<PathBuf>,
```
andreievg, line 46: not needed, not generating convert_data in report_builder anymore

server/server/src/lib.rs
```diff
@@ -149,7 +149,7 @@ pub async fn start_server(
     }
 
     // PLUGIN CONTEXT
-    PluginContext {
+    BoaJsContext {
```
andreievg, line 152: It's BoaJs context now, not plugin specific

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -1,28 +1,23 @@
...
 use thiserror::Error;
 
-use crate::{backend_plugin::boajs, service_provider::ServiceProvider};
-
-use super::boajs::BoaJsPluginError;
+use crate::boajs::{self, BoaJsError};
 
 #[derive(Debug, Error, PartialEq)]
-#[error("Error in plugin {r#type:?}")]
+#[error("Error in plugin {code}")]
```
andreievg, line 12: One thing i've changed in this file, is holding plugin code within PluginInstanceVariant, since i really wanted to show plugin code on error

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -80,13 +57,19 @@ where
...
     O: DeserializeOwned,
 {
-    let result = match plugin {
-        PluginInstance::BoaJs(bundle) => {
-            boajs::call_plugin(input, &r#type, &bundle).map_err(Into::into)
-        }
+    let result = match &plugin.variant {
+        PluginInstanceVariant::BoaJs(bundle) => boajs::call_method(
+            input,
+            vec!["plugins", &plugin_type_to_string(r#type)],
```
andreievg, line 63: This is another change, see call_method definition

server/service/src/boajs/call_method.rs
```diff
@@ -2,93 +2,92 @@ use std::{path::Path, rc::Rc};
...
-        .cloned()
-        .ok_or_else(|| Error::PluginNamespaceMissing(type_string.clone()))?;
-
-    let key = js_string!(type_string.as_str());
-    let plugin = plugins
-        .get(key, context)?
-        .as_callable()
-        .cloned()
-        .ok_or_else(|| Error::PluginMissing(type_string.clone()))?;
+    let callable = find_callable_in_exports(context, module, export_location)?;
```
andreievg, line 66: this is the main change really, you'll notice that previously as_callable was used, which still returns a JsObject (but checks if it's callable, i thought it's ok to error out when we try to call non callable

server/service/src/report/report_service.rs
```diff
@@ -568,11 +579,35 @@ fn transform_data(
...
+            transform_data_extism(connection, data, convert_data).map_err(ConvertDataError::Extism)
+        }
+    }
+}
+
+fn transform_data_boajs(data: ReportData, convert_data: String) -> Result<ReportData, BoaJsError> {
+    call_method(
+        data,
+        vec!["convert_data"],
+        &BASE64_STANDARD.decode(convert_data).unwrap(),
```
andreievg, line 602: could do base64 decode in transform_data, but I didn't because of variable naming, if it was called convert_data_base64 then yeah would have made more sense, on hindsight should have been specific (but thought that we could store it as none base64 in the future ¯_(ツ)_/¯l

standard_reports/README.md
```diff
@@ -56,12 +54,13 @@ Optional fields in the manifest json are marked as '// optional'
     "ui": "argument_schemas/arguments_ui.json"
   },
   // optional
-  // path to custom wasm data conversion function. Having a path to a custom function allows users to generate a function from a language other than JS
-  "custom_wasm_function": "path to custom wasm function",
```
andreievg, line 60: ohh and custom_wasm_function was removed

standard_reports/expiring-items/latest/convert_data_js/src/convert_data.js
```diff
@@ -1,16 +1,13 @@
-import { processStockLines } from "./utils";
+import { processStockLines } from './utils';
 
-function convert_data() {
-  let res = JSON.parse(Host.inputString());
+function convert_data(res) {
```
andreievg, line 3: all of these should be safe, no logical differences, please let me know if you see anything

standard_reports/item-usage/latest/convert_data_js/codegen.yaml
```diff
@@ -0,0 +1,24 @@
+overwrite: true
```
andreievg, line 1: This ones are for codegen of graphql queries

standard_reports/item-usage/latest/convert_data_js/codegenTypes.ts
```diff
@@ -0,0 +1,46 @@
+// Content of this file is generate but then manually pasted here, see comment in codegen.yaml
+export type Maybe<T> = T | null;
```
andreievg, line 2: not sper happy about this, but wanted to do a full typescript example

standard_reports/item-usage/latest/convert_data_js/codegen.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Server must be running and initialised
+schema: 'http://localhost:8000/graphql'
```
andreievg, line 3: decided to go with this, otherwise have to generate types from schema file, which also needs to be generated, and we need to know where server is for that, which makes file structure quite dependent, wanted to avoid dependency we had previously, with report cli/bash etc..

standard_reports/item-usage/latest/convert_data_js/src/convert_data.ts
```diff
@@ -0,0 +1,92 @@
+import { ArrayElement, ConvertData } from '../convertDataType';
+import { Arguments } from './generated-types/arguments';
+import { ItemUsageQuery } from './generated-types/graphql';
+import get from 'lodash/get';
+import orderBy from 'lodash/orderBy';
+
+type SqlResult = { item_id: string; quantity: number }[];
```
andreievg, line 7: yeah quite a bit of type definitions, although i think it's worth it, would have saved time in initial dev i think

standard_reports/item-usage/latest/convert_data_js/src/generated-types/graphql.ts
```diff
@@ -0,0 +1,28 @@
+import * as Types from '../../codegenTypes';
+
+export type ItemUsageQueryVariables = Types.Exact<{
```
andreievg, line 3: I really like this simple output

standard_reports/item-usage/latest/convert_data_js/src/test/input.json
```diff
@@ -0,0 +1,144 @@
+{
```
andreievg, line 1: his file was moved and changed to fully reflect the input data (which I think it should)

standard_reports/item-usage/latest/convert_data_js/src/test/output.json
```diff
@@ -0,0 +1,52 @@
+{
```
andreievg, line 1: Also moved this one, and fully reflect output data, plus checking sort with item names (also note capitilisation made to test, since 'I' < 'i'

standard_reports/item-usage/latest/convert_data_js/src/test/test.ts
```diff
@@ -0,0 +1,11 @@
+import { convert_data } from '../convert_data';
+import inputData from './input.json';
+import outputData from './output.json';
+import { describe, expect, it } from '@jest/globals';
+
+describe('Test all', () => {
```
andreievg, line 6: I think we need more 'overall' test for reports, rather then testing individual methods (unless overall fail and you need to drill down on the error)

server/cli/src/report_utils.rs
```diff
@@ -24,10 +22,10 @@ pub enum ReportError {
     PathDoesNotHaveParent(PathBuf),
     #[error("Failed to find package json {0}")]
     FailedToFindPackageJson(PathBuf),
-    #[error("Failed to find esbuild module {0}")]
-    FailedToFindEsbuildModule(PathBuf),
-    #[error("Failed to install yarn {0}")]
-    FailedToInstallYarn(PathBuf, #[source] std::io::Error),
+    #[error("Failed to yarn install {0}")]
```
roxy-dao, line 25: 😍

standard_forms/prescription-receipt/latest/convert_data_js/src/utils.js
```diff
@@ -1,24 +0,0 @@
-module.exports = {
-  processPrescription,
-  roundValueTwoDp,
-};
-
-const processPrescription = (invoice) => {
```
lache-melvin, line 6: I haven't tested yet but... won't this still need to run??

lache-melvin: Oh I see, maybe it actually wasn't being included in the first place?

standard_reports/item-usage/latest/convert_data_js/codegenTypes.ts
lache-melvin: Hm agree, though seems likes these have been fairly stable - are you sure there isn't a way that we can define this is the root of reports rather than in the individual report though? At least then would only have to update once? I guess not quite sure what that looks like for custom reports... do we have TS examples in those yet?

standard_reports/item-usage/latest/convert_data_js/src/test/test.ts
lache-melvin: Hm, not sure I agree completely, I think both have value.

Little unit tests are much easier to write in test driven development, add a new test for each new piece you want to add. They are also much easier to ensure you have tested every case, different kinds of inputs to each individual method.

Overall tests are definitely crucial, like you say with real-looking data, to ensure it hangs together correctly. But I think this is a happy path test, does the correct report data come back with these one or two key use-case inputs.

Don't think we should be deleting the unit tests though, much easier to find what broke, and have confidence that each piece of the logic holds up in all the ways it should 👀

standard_reports/item-usage/latest/convert_data_js/src/utils.test.ts
```diff
@@ -1,158 +0,0 @@
...
-import outputData from "../output.json" assert { type: "json" };
-
-describe("test item lines", () => {
-  it("end to end item-usage", () => {
-    const result = processItemLines(inputData);
-    expect(result).toEqual(outputData.items.nodes);
-  });
-});
-
-describe("Adds monthlyConsumption correctly from query result", () => {
```
lache-melvin, line 16: As above, I think these unit tests have value, don't see a reason to delete them?

server/service/src/report/report_service.rs
lache-melvin: Hm yeah think fine for now, if we end up adding many more convert data types then maybe worth moving later 🤷‍♀️

lache-melvin (approved):
Looks to be working well for me! Tested with some existing extism reports as well. Didn't test plugins but change looks minimal and compiling - did you smoke test?

Guess we need to push to convert the other new custom reports before upcoming releases?

lache-melvin:
Oh, but I do think we should add back those unit tests, and am curious if you think we can put those codegen types somewhere more shared.

lache-melvin (changes requested):
oh sorry doing as request changes, as is rc branch 🙏

andreievg:
@lache-melvin 

I agree that having unit tests would be helpful, but I think typescript kind of makes those particular tests redundant, I've update that report, making it simpler: c75dc36ed0d2f781deb9e6a2be32768a06ce80b7, tested works as expected
* we shouldn't be doing basic formatting in JS, use template for that
* we the use case with negative number should in place where it's relevant

standard_reports/item-usage/latest/convert_data_js/codegenTypes.ts
andreievg: I wanted it in the root of the reports, but it did mean a lot of reshuffling, and actually using adding custom reports as submodule, or export types as npm packages from core (due to the 'rootDir' nature of TS)

standard_reports/item-usage/latest/convert_data_js/src/test/test.ts
andreievg: You'll see comment below, and refactor. Which I wasn't 100% happy with, as it is a bigger change then what I intended in this PR, however I decided to make the change, since this is an example and we should try to keep examples simple, direct and as patterns that will be copied.

andreievg:
Thanks for review !

standard_reports/expiring-items/latest/report-manifest.json
```diff
@@ -12,5 +12,6 @@
     "schema": "argument_schemas/arguments.json",
     "ui": "argument_schemas/arguments_ui.json"
   },
-  "convert_data": "convert_data_js"
-}
+  "convert_data": "convert_data_js",
+  "convert_data_type": "BoaJs"
```
fergie-nz, line 16: Version was not bumped on this manifest

PR #7060, opened 2025-03-26
6280 fix exporting and importing coldchain equipment

### LINKED ISSUE #6280: CCEI: 'Needs replacement' value `FALSE` treated as `TRUE` when importing CCE
## What went wrong? 😲

When CCE are exported the system generates a .csv with `FALSE` for those CCE that do not have the `Needs replacement` checkbox ticked and `TRUE` for those that do. 

However, when importing the same file, the `FALSE` are treated the same as the `TRUE`

.csv export: 

Image

Import UI: 

Image

Note: there is a side bug where there is an inconsistency in whether `FALSE` or <blank> is populated during the export - looks likely related to whether dates for `Replacement due` are populated or not. 

## Expected behaviour 🤔

The values exported for empty checkboxes should result in empty checkboxes when re-imported. 

## How to Reproduce 🔨

1. Export CCE - some with checked box, some without
2. Reimport the exported .csv

## Your environment 🌱

- Open mSupply Version: 2.5.00
- Legacy mSupply Central Server Version: 7.19.06
- Platform: Windows
- Database type: Postgres

zachariah-at-msupply:
Fixes #6280 

# 👩🏻‍💻 What does this PR do?

This PR makes sure that the needs replacement field of cold chain equipment is correctly imported and exported.
It achieves this by adding a formatting function to the 'needs replacement' field in one of the functions that processes the data imported from the csv data.
You'll find that function in line 336 of client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx.

It seems the side bug of the issue (that sometimes a blank was inserted instead of 'false' in the needs replacement column in the exported CSV file) was resolved as well as, when I checked it, it seemed that false was consistently added to the correct rows in the CSV file.

# 🧪 Testing

- [ ] In open mSupply, go to Coldchain > Equipment and add at least three assets (the button is in the top right hand corner).
- [ ] Open one of those assets and tick the 'needs replacement' box.
- [ ] Back on the Equipment page, click export.
- [ ] Open the exported CSV using a spreadsheet app or similar (it will be in your downloads folder) and check that it has 'true' in the needs replacement column where it should.
- [ ] Back on the Equipment page in open mSupply, delete all the assets you added.
- [ ] Click import (top right hand corner) and select the file you just exported.
- [ ] Once you've imported the assets, make sure each one has the 'needs replacement' box ticked or not ticked as it should be.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
+++ b/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
@@ -55,6 +55,14 @@ const formatDate = (value: string): string | null => {
   return Formatter.naiveDate(DateUtils.getDateOrNull(value, 'dd/MM/yyyy'));
 };
 
+const formatNeedsReplacement = (value: string): boolean | null => {
+  if (value.match(/true/i)) {
+    return true;
+  } else {
+    return false;
+  }
+};
+
 function getImportHelpers<T, P>(
   row: P,
   rows: T[],
@@ -325,7 +333,11 @@ export const EquipmentUploadTab: FC<ImportPanel & EquipmentUploadTabProps> = ({
         'label.status',
         status => parseStatusFromString(status, t) ?? StatusType.Functioning
       );
-      addCell('needsReplacement', 'label.needs-replacement');
+      addCell(
+        'needsReplacement',
+        'label.needs-replacement',
+        formatNeedsReplacement
+      );
       processProperties(properties ?? [], row, importRow, rowErrors, t);
       importRow.errorMessage = rowErrors.join(',');
       importRow.warningMessage = rowWarnings.join(',');
--- a/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
+++ b/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
@@ -56,11 +56,8 @@ const formatDate = (value: string): string | null => {
 };
 
 const formatNeedsReplacement = (value: string): boolean | null => {
-  if (value.match(/true/i)) {
-    return true;
-  } else {
-    return false;
-  }
+  if (value.match(/true/i)) return true;
+  else return false;
 };
 
 function getImportHelpers<T, P>(
--- a/client/packages/coldchain/src/Equipment/utils.ts
+++ b/client/packages/coldchain/src/Equipment/utils.ts
@@ -49,7 +49,7 @@ export const assetsToCsv = (
 
     const status =
       node.statusLog?.status && parseLogStatus(node.statusLog.status);
-
+      
     return [
       node.id,
       ...(isCentralServer ? [node.store?.code] : []),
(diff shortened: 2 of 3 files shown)


zachariah-at-msupply:
To correct my note to the reviewer: it seems that the side issue isn't fixed after all based on some further testing :( Thanks for confirming that @GeronimoJohn 👍 !

client/packages/coldchain/src/Equipment/utils.ts
```diff
@@ -61,7 +61,7 @@ export const assetsToCsv = (
       Formatter.csvDateString(node.warrantyEnd),
       node.serialNumber,
       status ? t(status.key) : '',
-      node.needsReplacement,
+      node.needsReplacement ? node.needsReplacement?.toString() : 'false',
```
GeronimoJohn, line 64: Thanks Zac! Just wondering if they need to be a string. I think we can go without it and it should work the same 

`node.needsReplacement ? node.needsReplacement : false,`

GeronimoJohn (approved):
Thanks for the fix @zachariah-at-msupply, good stuff. Works well for me! Just left a few comments but approving!

client/packages/coldchain/src/Equipment/utils.ts
GeronimoJohn: Note* `Formatter.csv` formats the value that is returned by `assetsToCsv` to a string. So thought we could let it handle the type conversion instead

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -55,6 +55,14 @@ const formatDate = (value: string): string | null => {
   return Formatter.naiveDate(DateUtils.getDateOrNull(value, 'dd/MM/yyyy'));
 };
 
+const formatNeedsReplacement = (value: string): boolean | null => {
+  if (value.match(/true/i)) {
+    return true;
+  } else {
+    return false;
+  }
+};
```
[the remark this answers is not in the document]
CarlosNZ: Or just:
```ts
const formatNeedsReplacement = (value: string) => !!value.match(/true/i)
```

😉

zachariah-at-msupply:
It seems like the suggested fix (increasing the width of the list items) might already be implemented. Here is a screenshot from tablet-size sreen (1080x810) to show you what it is like now:

image

However, as you can see from that screenshot, there are some other problems!
The name is wrapping, but seems to be constrained by the row height and overflowing the column width.
Perhaps a horizontal scrollbar is a good solution, or maybe a different way of wrapping, or maybe something else; what are your thoughts @mark-prins?

(Just for reference for the future, I'm making the note that the files effecting this change are client/packages/system/src/Store/components/StoreSearchInput/StoreSearchInput.tsx and client/packages/system/src/Store/components/StoreSearchInput/StoreOptionRenderer.tsx)

PR #7066, opened 2025-03-26
Open internal order edit modals in read only mode

### LINKED ISSUE #6929: Should be able to open requisition edit modals in read only mode

When requisition (internal order) is sent, it can't be edited and we can't access the edit line modal, we should be able to:

* That modal is more convenient and easer to use then the table for tablet users
* Can't see graphs or other info (like by line calculation table)

Able to use by line edit modal in sent requisition

aimee-mcneil-melville:
Fixes #6929 

# 👩🏻‍💻 What does this PR do?

Allows the user to open the line edit view on any Internal order (instead of only draft status)

If the order is not in Draft status all fields are disabled for a 'read only' view

The button to view stats & graphs is still available, and the user can select the items on the left hand side 

Screenshot 2025-03-27 at 11 13 23 AM

Several of the fields are always disabled, is there a better name to differentiate between these?

I also looked at wrapping the section in a disabled package/context which may have been useful except for the stats button needing to be accessible, and probably more work than it's worth without additional use cases

# 🧪 Testing

- [ ] Go to Replenishment -> Internal Orders -> Create a new order
- [ ] Add some items. Observe the editable fields
- [ ] Close -> Confirm Sent
- [ ] Select a line and be taken to the edit view with the details of that line
- [ ] Observe that no fields are editable, but the stats are available by clicking on the orange icon. 
- [ ] Other items can be selected on the left hand side, but no new items added
- [ ] A finalised order will be the same in this view as a Sent order

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/DetailView.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/DetailView.tsx
@@ -67,12 +67,7 @@ export const DetailView: FC = () => {
   };
   const tabs = [
     {
-      Component: (
-        <ContentArea
-          onRowClick={!isDisabled ? onRowClick : null}
-          onAddItem={onAddItem}
-        />
-      ),
+      Component: <ContentArea onRowClick={onRowClick} onAddItem={onAddItem} />,
       value: 'Details',
     },
     {
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestEditPage.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestEditPage.tsx
@@ -55,6 +55,7 @@ export const RequestLineEditPageInner = ({
         .map(line => line.item.id)
     : [];
   const isProgram = !!requisition.programName;
+  const isDisabled = requisition.status !== 'DRAFT';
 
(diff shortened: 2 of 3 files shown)


roxy-dao:
Can we do the same for requisitions as well (: please and ty

aimee-mcneil-melville:
Done in Requisitions and have updated the main comment 😄

noel-yeldos (commented):
Looks good Amy! Everything is working great for me. I couldn't test the additional requisition fields (for some reason they weren't showing for me even in a program requisition with the pref on). 

I agree that the disabled prop needs a different name to avoid confusion. I wonder, since the check is the status of the requisition/internal order, if it can be related to that?

aimee-mcneil-melville:
I've updated the naming, and investigated the additional fields pref - it appears to be working only after changing stores in OMS
I'll raise an issue, and have added additional screenshots to the main comment with all of the extra fields

noel-yeldos (approved):
Thanks Amy, looks good to me! I'm now able to see all extra fields. I can navigate to line edit view for lines in finalised requisitions and sent internal orders. Cannot edit fields where requisition is finalised or where internal order is sent.

PR #7068, opened 2025-03-27
7057 create new inbound shipment firing twice and creates two shipments

### LINKED ISSUE #7057: Create new inbound shipment firing twice and creates two shipments
## What went wrong? 😲
Clicking on "create new inbound shipment" is creating two inbound shipments instead of one

https://github.com/user-attachments/assets/dcd44cf4-242f-414b-a496-fd3a09b496bb

## Expected behaviour 🤔
Should only create one 

## How to Reproduce 🔨

1. Go to 'Replenishment' -> 'Inbound shipment'
2. Click on 'New Inbound Shipment'
3. Select store
4. Go back to list view and see that two shipments were created instead of one

## Your environment 🌱

- Open mSupply Version: 2.7 (but also found in 2.6-RC4)
- Legacy mSupply Central Server Version: 8.01

- Platform: Mac OS chrome

- Database type:

roxy-dao:
Fixes #7057

# 👩🏻‍💻 What does this PR do?
Memorize createInvoice method and add as dependency to useEffect so it doesn't fire twice. 

# 🧪 Testing

- [ ] Have one store with `link Inbound Shipment to IO` store pref and one without for easy testing
- [ ] Create IS shipment on both 
- [ ] Shouldn't have 2 invoices created

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/ListView/AppBarButtons.tsx
+++ b/client/packages/invoices/src/InboundShipment/ListView/AppBarButtons.tsx
@@ -63,25 +63,11 @@ export const AppBarButtons = ({
     success(t('success'))();
   };
 
-  const createInvoice = async (nameId: string) => {
+  const createInvoice = async (nameId: string, requisitionId?: string) => {
     const invoiceNumber = await onCreate({
       id: FnUtils.generateUUID(),
       otherPartyId: nameId,
-    });
-
-    navigate(
-      RouteBuilder.create(AppRoute.Replenishment)
-        .addPart(AppRoute.InboundShipment)
-        .addPart(String(invoiceNumber))
-        .build()
-    );
-  };
-
-  const onRowClick = async (row: LinkedRequestRowFragment) => {
-    const invoiceNumber = await onCreate({
-      id: FnUtils.generateUUID(),
-      otherPartyId: name?.id ?? '',
-      requisitionId: row.id,
+      requisitionId,
     });
 
     navigate(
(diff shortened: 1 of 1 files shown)


andreievg:
It's a bit hard to figure out what is going on in the app bar logic, I think the main problem here is the use of 'useEffect' to createInvoice,  by listening to name changes. Also using conditions for like `data?.totalCount` is not obvious, so should be a variable defined closer to where it's visible what 'data' is.

Can you please:

1 - move `data?.totalCount !== 0 && manuallyLinkInternalOrder` to a variable describing what it means (and where it's visible what they mean)
2 - remove useEffect to create not linked invoice, use a method (createInvoice in this case does not need to be useCallback, and createInvoice can take optional 'requisitionId' to combine it together with onRowClick)
3 - technically don't need check above `<LinkInternalOrderModal`, if there is a check for it before toggle modal

It's a good idea to avoid using useEffect as much as possible, especially when there is reliance on multiple dependencies.

roxy-dao:
> It's a good idea to avoid using useEffect as much as possible, especially when there is reliance on multiple dependencies.

[6038c5c](https://github.com/msupply-foundation/open-msupply/pull/7068/commits/6038c5cd36806bd7e956cc8fa9e86e835db297d7) & [7ef44bf](https://github.com/msupply-foundation/open-msupply/pull/7068/commits/7ef44bf30c7c71c0b3194995fa506b4d1c7b3ec4)

andreievg (approved):
Looks good thanks !

nit: Could just have one method createInvoice(nameId: string, requisitionId?: string)

PR #7071, opened 2025-03-27
Refactor for awesomeness

zachariah-at-msupply:
Just a follow-up PR to [this one](https://github.com/msupply-foundation/open-msupply/pull/7060), implementing a neat little refactor Carl suggested, for the purpose of more awesomer code.

### DIFF
--- a/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
+++ b/client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
@@ -55,10 +55,7 @@ const formatDate = (value: string): string | null => {
   return Formatter.naiveDate(DateUtils.getDateOrNull(value, 'dd/MM/yyyy'));
 };
 
-const formatNeedsReplacement = (value: string): boolean | null => {
-  if (value.match(/true/i)) return true;
-  else return false;
-};
+const formatNeedsReplacement = (value: string) => !!value.match(/true/i);
 
 function getImportHelpers<T, P>(
   row: P,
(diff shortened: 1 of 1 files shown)


client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
```diff
@@ -55,10 +55,7 @@ const formatDate = (value: string): string | null => {
   return Formatter.naiveDate(DateUtils.getDateOrNull(value, 'dd/MM/yyyy'));
 };
 
-const formatNeedsReplacement = (value: string): boolean | null => {
-  if (value.match(/true/i)) return true;
-  else return false;
-};
+const formatNeedsReplacement = (value: string) => !!value.match(/true/i);
```
CarlosNZ, line 58: I would suggest renaming the function to something a bit more explicit, like `parseStringToBoolean` or something, but up to you.

CarlosNZ (approved):
I'll approve this @zachariah-at-msupply cos I'm about to head off for the weekend, but can you make 100% sure the functionality hasn't changed? Thanks.

Also, one small suggestion, up to you.

client/packages/coldchain/src/Equipment/ImportAsset/UploadTab.tsx
andreievg: This is what i call a stylistic change (up to developer preference), I think what you had originally works and look ok, but yeah agree that shorter version looks better. 

I think the important thing, that should not have been a nit: but change request, is the function name. I think the name should be 'isTruthyString'

zachariah-at-msupply:
Confirming that I have tested it as in the original PR and it works.

PR #7087, opened 2025-03-27
6841 edit prefs

### LINKED ISSUE #6841: Central UI Skeleton

See relevant section of epic.

lache-melvin:
Fixes #6841

# 👩🏻‍💻 What does this PR do?

Forgive me, she's a lil big 😅 

Adds a Manage UI for preferences:

Screenshot 2025-03-28 at 10 21 45 AM

In development, we define a default. All values will be this default. User can set global value - all stores will then use this value, unless you define a store-override.

Values are autosaved.

These are just some example prefs to get the idea, I think we remove all except the `show_contact_tracing` one once happy?

Will add some inline comments too, but have added the store-pref select UI as a rough idea, I think we should hide/remove this for now, it probably needs to be paginated/searchable before it adds much value. Imagine in future we would also want to manage from the other direction (on stores page, select a store, can set all prefs) but I think this is the main one we need for now?

# 🧪 Testing

- [ ] On OMS Central
- [ ] There is now a Manage > Preferences page
- [ ] There is a list of available preferences
- [ ] Selecting one, you can set global pref
- [ ] It auto-saves, refreshing the page, the value persists
- [ ] Global pref now shows as the pref for all stores too
- [ ] Can set a pref per-store

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/preference/preferences/complex_pref.rs
```diff
@@ -1,15 +1,55 @@
...
+#[derive(Clone, Default, Deserialize, Serialize)]
 pub struct ComplexPref {
     pub something_here: i32,
     pub something_else: String,
 }
 
-impl Preference<ComplexPref> for ComplexPref {
+pub struct ComplexOne;
+
+impl Preference for ComplexOne {
```
lache-melvin, line 14: So in theory this is about as hard as it would get for us to define a new preference?!

server/service/src/preference/preferences/mod.rs
```diff
@@ -1,18 +1,35 @@
...
+    pub show_contact_tracing: ShowContactTracing,
+    pub complex_one: ComplexOne,
+    pub months_of_stock: MonthsOfStock,
+}
+
+fn get_preference_registry() -> PreferenceRegistry {
+    PreferenceRegistry {
+        show_contact_tracing: ShowContactTracing,
+        complex_one: ComplexOne,
+        months_of_stock: MonthsOfStock,
```
lache-melvin, line 25: Then you just add your new pref to the registry here, and the `Preferences` struct below

server/service/src/preference/preferences/mod.rs
```diff
@@ -21,45 +38,33 @@ pub fn get_preferences(
...
-        complex: ComplexPref::load(connection, store_id)?,
+        show_contact_tracing: show_contact_tracing.load(connection, store_id)?,
+        complex: complex_one.load(connection, store_id)?,
+        months_of_stock: months_of_stock.load(connection, store_id)?,
     };
 
     Ok(prefs)
 }
 
-pub trait Preference<T: Default + DeserializeOwned> {
```
lache-melvin, line 32: moved to own file

server/service/src/preference/preferences/show_contact_tracing.rs
```diff
@@ -0,0 +1,11 @@
+use super::Preference;
+
+pub struct ShowContactTracing;
+
+impl Preference for ShowContactTracing {
+    type Value = bool;
```
lache-melvin, line 6: I think the majority of prefs will be booleans, so would look just like this :)

Chris-Petty:
Been in the DMs, let me know to come back?

lache-melvin:
@Chris-Petty come backkk

client/packages/host/src/routers/ManageRouter.tsx
```diff
@@ -6,33 +6,29 @@ const NameService = React.lazy(
   () => import('@openmsupply-client/system/src/Name/Service')
 );
 
-const IndicatorsDemographicsService = React.lazy(
-  () =>
-    import(
-      '@openmsupply-client/system/src/IndicatorsDemographics/Service/Service'
-    )
+const ManageService = React.lazy(
+  () => import('@openmsupply-client/system/src/Manage/Service')
```
andreievg, line 10: Can you please comment about this change, why it was done, etc.. (since PR is quite big already), happy to review this but a quick note 'This is renaming, this functionality is now 'common' etc..

client/packages/programs/src/JsonForms/common/components/Boolean.tsx
```diff
@@ -21,7 +21,10 @@ const UIComponent = (props: ControlProps) => {
 
   return (
     <DetailInputWithLabelRow
-      sx={DefaultFormRowSx}
+      // On most json form elements, we want the label at the start as
+      // the input may span multiple lines. However, for a boolean field,
+      // the input is a single line, so we want the label centred.
+      sx={{ ...DefaultFormRowSx, alignItems: 'center !important' }}
```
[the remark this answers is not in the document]
andreievg: What effect does this have on other toggles in json forms ?

client/packages/host/src/routers/ManageRouter.tsx
lache-melvin: Sorry did put comments as above but looks like they became outdated somehow, will add again 🙏

client/packages/host/src/routers/ManageRouter.tsx
```diff
@@ -6,33 +6,29 @@ const NameService = React.lazy(
...
+  const goToManage = useMatch(fullManagePath);
 
   if (gotoFacilities) {
     return <NameService />;
   }
 
-  if (gotoIndicatorsDemographics) {
-    return <IndicatorsDemographicsService />;
+  if (goToManage) {
+    return <ManageService />;
```
lache-melvin, line 31: Demographics was just one page under `Manage` section. Now that we are adding a second one, thought best to refactor "indicatorservice" to be manage service, and demographics just a page within it. Sorry for extra noise, it's all just moving files accordingly 🙏

client/packages/programs/src/JsonForms/common/components/Boolean.tsx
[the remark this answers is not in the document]
lache-melvin: Looks the same :)

Screenshot 2025-03-31 at 4 16 54 PM

server/service/src/preference/upsert.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct UpsertPreference {
+    pub id: String,
+    pub store_id: Option<String>,
+    pub key: String,
+    pub value: String,
+}
+
+pub fn upsert_preference(
```
andreievg, line 13: I wanted this to be type safe, and not at runtime but at compile time. 

The thought that I had was having upsert method for every preference, similar to the query method in Preference for now. And when we do by store preference or machine uuid we would probably need a different type of save, maybe with array of store -> pref objects.

client/packages/system/src/Manage/Preferences/api/operations.graphql
```diff
@@ -0,0 +1,40 @@
+query allPrefs {
```
andreievg, line 1: so yeah the endpoints I was thinking about were like this:

```graphql

query MonthsOfStock {
    preference {
        monthsOfStock
    }
}
# Can query them individually, when needed
query ShowContractTracing {
    preference {
        showContractTracing
    }
}
# This preference has a global default but is also by store
query MonthsOfStockForStore($storeId:  String!) {
    preference {
        monthsOfStock(storeId: $storeId)
    }
}
# Saving
mutate SaveMonthsOfStock($globalDefault: String!) {
    preference {
        saveMonthsOfStock(default: $globalDefault)
    }
}
# Saving technically speaking we would need a way to mass configure these
mutate SaveMonthsOfStockForStores($input: MonthsOfStockForStoreUpsertInput!) {
    preference {
        saveMonthsOfStock(forStores: $input)
    }
}
``` 

Above input would be like this:

{ 
    value: string,
    storeId: string
}[]

client/packages/system/src/Manage/Preferences/api/usePreferencesGraphQL.ts
```diff
@@ -0,0 +1,15 @@
+import {
+  useAuthContext,
+  useGql,
+  useQueryClient,
+} from '@openmsupply-client/common';
+import { getSdk } from './operations.generated';
+
+export const usePreferencesGraphQL = () => {
+  const { client } = useGql();
```
andreievg, line 9: Ias looking at this and I think we just need one generic hook for the whole app, used like:

const { sdk, storeId } = useGraphQL(getSdk); <- and this is used inside the hooks

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
```diff
@@ -0,0 +1,35 @@
...
+    if (key)
+      setCustomBreadcrumbs({
+        1: t(`preference.${key}` as LocaleKey),
+      });
+  }, [key]);
+
+  const selectedPref = data?.find(d => key === d.key);
+
+  return (
+    <LineEditBase currentKey={key ?? ''} prefs={data ?? []}>
```
andreievg, line 29: I honestly think this should be a normal table, where we can see all preferences, and then we go to matrix view for individual preferences for stores etc..

andreievg (changes requested):
Hi, keen to have a chat about this in person.

I was thinking about JSON forms, and here is my position:

We should be using them for things that are not know at compile time, like the dynamic nature of some configurations, programs, custom reports etc.. 

Otherwise I think we should be using normal react.

I like the Preference type definition, and the load method for it (which can be overwritten), but the load method may require more parameters, like machine uuid etc.. I think the main benifit of the Preference type definition is to associate 'type', 'key', and default of the pref and to help with serialization and deserialization. 

Load should really be a bunch of helper methods, new one added when needs arrise.

client/packages/system/src/Manage/Preferences/api/operations.graphql
lache-melvin: At least for the query side... that is how it works? Sorry no frontend example yet, but that's what this node is:

https://github.com/msupply-foundation/open-msupply/blob/8974bc9619a1d709403e8dc4f9c6362f5ad24339/server/graphql/types/src/types/preferences.rs#L13-L25

 `allPrefs` is a special query only used to see full list of prefs in the edit view

I think mutation you have is also very similar to what is there, `input` object just has optional `store_id` to distinguish global and forStore. Are you saying you think its worth distinguishing the two inputs further?

Don't think it would be too hard to change mutation to accept an array of inputs if we wanted to do mass upserts in future

lache-melvin:
Hi thanks for review and chat @andreievg  - per discussion:

- Removed extra example prefs: [025eb49](https://github.com/msupply-foundation/open-msupply/pull/7087/commits/025eb49c7a515ce5f6cb1ec5dd1e93602f37e43e) (I think worth coming back to this diff, with current pattern this is all the code required to add a new pref
- Next 4 commits are just some radical code deletion
  - Remove the per-store piece [631d8ce](https://github.com/msupply-foundation/open-msupply/pull/7087/commits/631d8ceeffb40ecae798cae9d7d73101395a1828)
  - As you thought, didn't need the `PreferenceDescription` trait abstraction, except for the `type Value` on each implementation... Seeing as only have `bool` type for now, I just hard coded value to remove the extra trait, as wanted to remove the extra code. I think we will remove the need for this description concept with more type secure edit UI anyway :) [45533e4](https://github.com/msupply-foundation/open-msupply/pull/7087/commits/45533e42be11cebc1b636b019eec9e79ea5e1cdf)
  - Remove prefs by key query - only doing global prefs for this release, so just loading the pref through main prefs query and displaying. Not sure where we are going with store pref UI, we can drive backend based on frontend requirements at the time :)
- Added TODOs throughout for during cooldown, I am keen to work on this 🔥  Will create issue

client/packages/system/src/Manage/Preferences/EditPage/EditPreference.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <Box>
+      <DetailPanelSection title={t('label.global-preference')}>
+        <Box sx={{ backgroundColor: 'white', padding: 1, borderRadius: 1 }}>
+          <EditField
+            value={value}
+            config={selected}
+            // TODO: backend would generate IDs, just send the value
+            onChange={value => update({ value, id: `${selected.key}_global` })}
```
lache-melvin, line 44: Please note this change, if you have existing global prefs in DB, would need to clear

andreievg:
Thanks @lache-melvin, good to have our chat too

andreievg (approved):

PR #7110, opened 2025-03-30
6823 compile host only

### LINKED ISSUE #6823: Compile host only

GeronimoJohn:
Fixes #6823

# 👩🏻‍💻 What does this PR do?

Client directory script changes.

Removed `yarn compile` script, compilation takes longer than needed and doesn't provide click through. We've decided to just stick with using `yarn re-compile` which only focuses on compiling the `host` directory. 

Have removed the use of `yarn clean` script as it's currently not being used anywhere

Screenshot 2025-03-31 at 9 08 31 AM

Checked if `start-local` and `start-remote` in the client directory need Lerna. There’s only one instance of each in OMS, so running them with Lerna and parallel doesn’t seem necessary? Updated and tested `start-local` to `cd` into `host` and run directly (e.g., `cd packages/host && yarn start-local`). It worked fine without errors. We can likely simplify these to standalone commands but leaving that up for a discussion for now.

Also noticed that `yarn start` and `yarn start-local` in the client directory are essentially the same command. `start` just calls `start-local` internally (`"start": "yarn start-local",`). Wondering if we actually need both?

# 🧪 Testing

Dev specific changes

- [ ] Removed `yarn compile`
- [ ] Removed `yarn clean`

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -12,7 +12,6 @@
     "build-stats": "lerna run --scope @openmsupply-client/* build-stats",
     "build-plugins": "lerna run --scope @openmsupply-client/* build-plugin",
     "serve": "lerna run --scope @openmsupply-client/* --parallel serve",
-    "clean": "lerna run --scope @openmsupply-client/* --parallel clean",
     "re-compile": "tsc --project packages/host/tsconfig.json --noEmit",
     "compile-full": "lerna run --scope @openmsupply-client/* --parallel tsc",
     "test": "jest --config ./jest.config.js --maxWorkers=50% --env=jsdom",
--- a/client/package.json
+++ b/client/package.json
@@ -13,7 +13,6 @@
     "build-plugins": "lerna run --scope @openmsupply-client/* build-plugin",
     "serve": "lerna run --scope @openmsupply-client/* --parallel serve",
     "clean": "lerna run --scope @openmsupply-client/* --parallel clean",
-    "compile": "lerna run --scope @openmsupply-client/* --parallel tsc --since HEAD -- --noEmit",
     "re-compile": "tsc --project packages/host/tsconfig.json --noEmit",
     "compile-full": "lerna run --scope @openmsupply-client/* --parallel tsc",
     "test": "jest --config ./jest.config.js --maxWorkers=50% --env=jsdom",


andreievg (approved):
Looks good thanks @JonnaQuismundo.

Yeah i am not sure if we ever used 'lerna' to our benefit (or at least i havne't quite), so I feel like we technically don't need it now, I think it's worth making a refactor issue, to at least ask that question, and to fully clean up this package.json (like build-plugins script is not needed etc..)

PR #7119, opened 2025-03-31
7007 central server UI for manage assets cce

### LINKED ISSUE #7007: Central Server UI for managing assets (Cold Chain Equipment)

Currently on the central server when managing cold chain equipment, you can see the equipment for all the stores in the system.
However, this causes problems if you want to use the Central server as a remote login for stores they only want to see their own store's data.

The current central server behaviour (showing all stores) should still be available on central server via the path Manage -> Equipment (only visible on the central server).

The normal Cold Chain -> Equipment path should only show the current store's data (same as if it was on a remote site)

### Moneyworks Jobcode 🧰

CCEIDEV

GeronimoJohn:
Fixes #7007 

# 👩🏻‍💻 What does this PR do?

Equipment page to show under `Manage` section. Keeping the functionality to add or edit assets for different stores if you're within the central server and vaccine module is on.

Cold Chain's equipment page have been updated to only focus on adding/editing assets the current store's data and not affect the other stores, especially when on central server.

# 🧪 Testing

Don't forget to turn on the Vaccine Module preference in mSupply 😄 

**Manage**
- [ ] Navigate Manage -> Equipment
- [ ] Click on an Asset -> Should still be under Manage page
- [ ] Should be able to add/edit Assets for other stores

**Cold Chain and in Central Server**
- [ ] Navigate Cold Chain -> Equipment
- [ ] Click on an Asset -> Should still be under Cold Chain page
- [ ] Should be able to add/edit an Asset for the store you're in

**Cold Chain and not Central Server**
- [ ] Navigate Cold Chain -> Equipment
- [ ] Click on an Asset -> Should still be under Cold Chain page
- [ ] Should be able to add/edit an Asset for the store you're in

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/graphql/asset/src/mutations/insert.rs
```diff
@@ -30,29 +30,20 @@ pub fn insert_asset(
...
-                asset_input = {
-                    InsertAssetInput {
-                        store_id: Some(store_id.to_owned()),
-                        ..input
-                    }
-                }
-            }
+    let asset_input = match &input.store_id {
+        Some(input_store_id)
+            if !CentralServerConfig::is_central_server() && input_store_id != store_id =>
```
fergie-nz, line 35: This is cool I didn't know you could chain logic statements in match. 

But I'm also wondering if this refactor is necessary if this is doing the same thing? Am I understanding correctly that the resulting logic is unchanged?

fergie-nz (approved):
Looks great nice one, John 🚀

server/graphql/asset/src/mutations/insert.rs
GeronimoJohn: Thanks @fergie-nz! 

It's a slight bit different. 

Before we don't allocate a store_id if inserting from central server. But now we do.

- Inserting an asset from Cold Chain page should use the store_ID regardless if central server or not
   - it should check if they have the permission to update the store to insert an asset 
- Inserting an asset from Manage page should use the selected store_id from the dropdown

The updated logic almost meets the newly desired behaviour, I noticed that the current logic currently doesn't do the permission check on a central server. Going to update the logic

PR #7150, opened 2025-04-01
Add find_all_active pattern

### LINKED ISSUE #6691: Inactive insurance providers can be selected in omSupply
## What went wrong? 😲

I have an Inactive insurance provider in mSupply

Image

And i am able to see it in a dropdown and select it in omSupply

Image

## Expected behaviour 🤔

Should not be able to see inactive insurance providers

## How to Reproduce 🔨

1. Make sure you have insurance providers setup in mSupply
2. Have one active and one inactive
3. In omSupply, go to Dispensary -> Patients -> Insurance and add an insurance
4. In the provider dropdown, note you can see both active and inactive provider

## Your environment 🌱

- Open mSupply Version: V2.6.0-RC2

- Database type: PostgreSQL

fergie-nz:
Fixes #6691

# 👩🏻‍💻 What does this PR do?

Adds option for filter for is_active in insurance_provider_row.

Replaces insurance provider query to search for find_all_active in stead of find_all

Not a sync related issue after all

I notice that we don't use any filtering for is_active in find_all in any of our row repositories, and would typically use a repository file to handle queries with filtering.

We do do find_active_by_id in row repositories, but never a list.

That being said, I'm not sure if what I've done is an acceptable pattern as it isn't used elsewhere.

I added this because adding a repository layer with query + filtering etc seemed like a lot of overhead if we aren't going to use any of the other functionality (pagination, custom filters, etc).

# 🧪 Testing

- [ ] Add an insurance provider (on mSupply if you wanted to confirm not sync related).
- [ ] See that the insurance provider is visible in omSupply.
- [ ] Set the insurance provider is_active status to be false in mSupply
- [ ] See that the insurance provider is no longer visible in omSupply.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/db_diesel/insurance_provider_row.rs
+++ b/server/repository/src/db_diesel/insurance_provider_row.rs
@@ -63,6 +63,15 @@ impl<'a> InsuranceProviderRowRepository<'a> {
         Ok(result)
     }
 
+    pub fn find_all_active(&self) -> Result<Vec<InsuranceProviderRow>, RepositoryError> {
+        let result = insurance_provider::table
+            .load(self.connection.lock().connection())?
+            .into_iter()
+            .filter(|row: &InsuranceProviderRow| row.is_active == true)
+            .collect();
+        Ok(result)
+    }
+
     pub fn upsert_one(&self, row: &InsuranceProviderRow) -> Result<i64, RepositoryError> {
         diesel::insert_into(insurance_provider::table)
             .values(row)
--- a/server/repository/src/db_diesel/insurance_provider_row.rs
+++ b/server/repository/src/db_diesel/insurance_provider_row.rs
@@ -65,10 +65,8 @@ impl<'a> InsuranceProviderRowRepository<'a> {
 
     pub fn find_all_active(&self) -> Result<Vec<InsuranceProviderRow>, RepositoryError> {
         let result = insurance_provider::table
-            .load(self.connection.lock().connection())?
-            .into_iter()
-            .filter(|row: &InsuranceProviderRow| row.is_active == true)
-            .collect();
+            .filter(insurance_provider::is_active.eq(true))
+            .load(self.connection.lock().connection())?;
         Ok(result)
     }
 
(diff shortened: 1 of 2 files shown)


server/repository/src/db_diesel/insurance_provider_row.rs
```diff
@@ -63,6 +63,15 @@ impl<'a> InsuranceProviderRowRepository<'a> {
...
 
+    pub fn find_all_active(&self) -> Result<Vec<InsuranceProviderRow>, RepositoryError> {
+        let result = insurance_provider::table
+            .load(self.connection.lock().connection())?
+            .into_iter()
+            .filter(|row: &InsuranceProviderRow| row.is_active == true)
+            .collect();
+        Ok(result)
+    }
+
```
fergie-nz, line 74: Also option to remove the now unused find_all if that would not be confusing.

server/repository/src/db_diesel/insurance_provider_row.rs
```diff
@@ -63,6 +63,15 @@ impl<'a> InsuranceProviderRowRepository<'a> {
...
     }
 
+    pub fn find_all_active(&self) -> Result<Vec<InsuranceProviderRow>, RepositoryError> {
+        let result = insurance_provider::table
+            .load(self.connection.lock().connection())?
+            .into_iter()
+            .filter(|row: &InsuranceProviderRow| row.is_active == true)
+            .collect();
+        Ok(result)
+    }
```
jmbrunskill, line 73:
```suggestion
    pub fn find_all_active(&self) -> Result<Vec<InsuranceProviderRow>, RepositoryError> {
        let result = insurance_provider::table
            .filter(insurance_provider::is_active.eq(true))
            .load(self.connection.lock().connection())?;
        Ok(result)
    }
```

This will update the sql query so the database does the filtering.
In this case it's probably not very important as there won't be a lot but I think it's still worth updating.

server/repository/src/db_diesel/insurance_provider_row.rs
jmbrunskill: I think it's ok to leave it.

jmbrunskill (changes requested):
Looks good, but can you change to the database filter instead of filtering in rust please?

jmbrunskill (approved):
Thanks Fergus!

PR #7163, opened 2025-04-01
6312 Align placeholder row packSize column

### LINKED ISSUE #6312: Placeholder line always showing 1
## What went wrong? 😲

not sure if this is intentional, but the placeholder line in outbound shipments is always showing 1

for example, i was testing a brand new item that had no previous stock movement or allocation, with the exception of 1 repack. when adding that item to an outbound shipment, i noticed placeholder said 1 - even though it is not added elsewhere

Image

Image

i then went ahead and added that same item to two different outbound shipments, allocated them and picked them -> they showed up in the item's ledger. however, when i went to create a third outbound shipment with the same item, placeholder still said 1? shouldn't it be 2 if the item is picked in 2 different outbound shipments?

Image

Image

## Expected behaviour 🤔

i think the number should be reflective of how many other places the item is being used in? and if none, it should be zero?

## How to Reproduce 🔨

1. Have an item that you haven't tested with - if it's easier, create a new one :)
2. Go to Inventory -> View Stock and add some stock to it, for example 500 units at pack size 1
3. I then repacked my item, so take 100 units and repack them into 50 packs at pack size 2
4. Go to Distribution -> Outbound Shipments and create a new shipment. Add your item to it
5. When adding, note placeholder lines say '1' even though your item is not used elsewhere
6. Now create 2 other outbound shipments and add the same item to them
7. Confirm Allocated -> Picked, note the outbound shipments now show up in the item's ledger
8. Open your third outbound shipment and try and add your item - note the placeholder still says 1 even though you have 2 other shipments
9. Unclear!

## Your environment 🌱

- Open mSupply Version: V2.6.0-cooldown-build-test

- Database type: PostgreSQL

noel-yeldos:
Fixes #6312 

# 👩🏻‍💻 What does this PR do?
Move column to the right by one so that it aligns with packSize column of the table

Screenshot 2025-04-02 at 12 26 39

# 🧪 Testing

- [ ] Go to Distribution -> Outbound Shipments and create a new shipment. Add an item to it
- [ ] When adding, note placeholder lines say '1' for pack size 
- [ ] Observe that it is now aligned correctly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
@@ -86,7 +86,7 @@ const TotalRow = ({ allocatedQuantity }: { allocatedQuantity: number }) => {
         <TotalCell
           style={{
             textAlign: 'right',
-            paddingRight: 12,
+            paddingRight: 20,
           }}
         >
           {!!NumUtils.hasMoreThanTwoDp(allocatedQuantity)
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
@@ -56,8 +56,13 @@ const PlaceholderRow = ({ line }: { line?: DraftStockOutLine }) => {
       <PlaceholderCell colSpan={3} sx={{ color: 'secondary.main' }}>
         {t('label.placeholder')}
       </PlaceholderCell>
-      <PlaceholderCell style={{ textAlign: 'right' }}>1</PlaceholderCell>
-      <PlaceholderCell colSpan={4}></PlaceholderCell>
+      <PlaceholderCell
+        style={{ textAlign: 'right', paddingRight: '14px' }}
+        colSpan={2}
+      >
+        1
+      </PlaceholderCell>
+      <PlaceholderCell colSpan={3}></PlaceholderCell>
       <Tooltip title={line?.numberOfPacks.toString()}>
         <PlaceholderCell style={{ textAlign: 'right' }}>
           {!!NumUtils.hasMoreThanTwoDp(placeholderBuffer)


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -56,8 +56,13 @@ const PlaceholderRow = ({ line }: { line?: DraftStockOutLine }) => {
...
       </PlaceholderCell>
-      <PlaceholderCell style={{ textAlign: 'right' }}>1</PlaceholderCell>
-      <PlaceholderCell colSpan={4}></PlaceholderCell>
+      <PlaceholderCell
+        style={{ textAlign: 'right', paddingRight: '14px' }}
+        colSpan={2}
+      >
+        1
+      </PlaceholderCell>
+      <PlaceholderCell colSpan={3}></PlaceholderCell>
```
noel-yeldos, line 65: While this has fixed it, I'm thinking that this set up for the placeholder cells is fragile. Surely we don't want to have to come in and make changes each time a column gets added or removed from the data table above the placeholder row?

aimee-mcneil-melville: Thanks, Noel! I agree - it is fragile! 

From what I see with this setup is the column spans create the placement - and when an extra column was added in the outbound rows the placeholder row was misaligned as a result - each count of the colSpans will push that cell one column to the right.

I'll add a suggestion below, it looks to me that the value showing for both the placeholder row and total row under the `Qty issued` column, are actually representative of `Pack qty issued` and are also misaligned - very hard to tell on UI as the 1 could represent either!

A comment on the outbound shipment 'columns' file would probably be helpful, as when making changes to the columns it would be quite easy to miss that the Placeholder/Total rows also need adjusting

aimee-mcneil-melville:
```suggestion
      <PlaceholderCell colSpan={4} sx={{ color: 'secondary.main' }}>
        {t('label.placeholder')}
      </PlaceholderCell>
      <PlaceholderCell style={{ textAlign: 'right' }} colSpan={1}>
        1
      </PlaceholderCell>
      <PlaceholderCell colSpan={2}></PlaceholderCell>
```

aimee-mcneil-melville (changes requested):
Thanks Noel, your solution lines it up nicely! 

If you could adjust the right-hand value in Placeholder and Total  to line up with `Pack qty issued` column as well please as they are also misaligned :)

noel-yeldos:
@aimee-mcneil-melville thanks for the review! :)

From what I'm seeing it appears that the total and placeholder qty cells are where they need to be? I'm basing that off the numbers when looking at the rows in the data table.
e.g 355 + 2145 = 2500 whereas 71 + 2145 = 2216 which doesn't add up

Screenshot 2025-04-03 at 11 47 06
Screenshot 2025-04-03 at 11 45 11

aimee-mcneil-melville:
Thanks for checking into that, I had thought the quantity issued column was representing number of packs - that column is all good then! :)

I do have a concern with the padding fix for the longer term however, in that every time a column is changed (an extra column or a width change) the pack size column could easily become misaligned again. Setting the column span so that the cell falls into the correct column I believe is a cleaner and more robust fix for the placeholder pack size as it will move with the table headers.

Can you update it to reflect that please? Let me know if you've got any questions or concerns!

noel-yeldos:
> Thanks for checking into that, I had thought the quantity issued column was representing number of packs - that column is all good then! :)
> 
> I do have a concern with the padding fix for the longer term however, in that every time a column is changed (an extra column or a width change) the pack size column could easily become misaligned again. Setting the column span so that the cell falls into the correct column I believe is a cleaner and more robust fix for the placeholder pack size as it will move with the table headers.
> 
> Can you update it to reflect that please? Let me know if you've got any questions or concerns!

@aimee-mcneil-melville yup your concern about adding fixed padding to the cell is valid! 

However, the Placeholder cells already all have fixed padding.
Screenshot 2025-04-03 at 17 11 56

Without the additional padding this is what the alignment is like
Screenshot 2025-04-03 at 17 12 13

And with the padding override to the specific cell
Screenshot 2025-04-03 at 17 14 02

Did you still want me to remove the additional padding?

aimee-mcneil-melville (approved):
Ah, sorry I do I see what you mean! 

I had misread some of it thinking the padding was moving it a column over 😅  Your fix is good and it does line up better with the padding there!

All good to merge in :)

PR #7174, opened 2025-04-02
Prevents render of too many years

### LINKED ISSUE #5280: Datetime picker loading incredibly slow when changing the year
## What went wrong? 😲

Datetime picker seems to be incredibly slow now when trying to tap on the dropdown to change the year or when tapping on an actual year

Video of the behaviour ->>> [here](https://drive.google.com/drive/folders/1b7jEflWlNiLubUKn5RRo_fzeKMX0_wXH)

## Expected behaviour 🤔

Loads fast

## How to Reproduce 🔨

2. Click on a D.O.B. filter
3. Try clicking on a dropdown to change the year or on an actual year
4. Notice the loading speed is incredibly slow compared to the rest of datetime picker components

## Your environment 🌱

- Open mSupply Version: V2.3.1-RC2
- Legacy mSupply Central Server Version: V7.18.6

- Platform: Galaxy Tablet A8 (Android)

- Database type: SQLite3

fergie-nz:
Fixes #5280 

# 👩🏻‍💻 What does this PR do?

Renders a select amount of years rather than 10,000

I've arbitrarily chosen the limit of 50 years ahead, and 120 years behind (to capture a reasonable date of birth option).

These values could be customised for a specific filter by passing in custom end ranges if we wanted to be more specific (ie it doesn't necessarily make sense to have a date of birth year option in the future).

This issue only fixes the described issue of render time

# 🧪 Testing

- [ ] open up any filter (ie patients, date of birth filter)
- [ ] Select to change the year - see it renders much more quickly.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
@@ -19,15 +19,14 @@ export type RangeOption = 'from' | 'to';
 export const DateFilter: FC<{ filterDefinition: DateFilterDefinition }> = ({
   filterDefinition,
 }) => {
-  const currentYear = new Date().getFullYear()
   const {
     type,
     urlParameter,
     name,
     range,
     displayAs = type,
-    maxDate = `${currentYear + 10}-12-31`,
-    minDate = `${currentYear - 120}-01-01`,
+    maxDate,
+    minDate,
   } = filterDefinition;
   const { urlQuery, updateQuery } = useUrlQuery();
   const { customDate, urlQueryDate, urlQueryDateTime } = useFormatDateTime();
--- a/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
@@ -19,14 +19,15 @@ export type RangeOption = 'from' | 'to';
 export const DateFilter: FC<{ filterDefinition: DateFilterDefinition }> = ({
   filterDefinition,
 }) => {
+  const currentYear = new Date().getFullYear()
   const {
     type,
     urlParameter,
     name,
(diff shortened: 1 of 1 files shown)


client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
```diff
@@ -19,14 +19,15 @@ export type RangeOption = 'from' | 'to';
...
   const {
     type,
     urlParameter,
     name,
     range,
     displayAs = type,
-    maxDate = '9999-12-31',
-    minDate = '0000-01-01',
+    maxDate = `${currentYear + 10}-12-31`,
+    minDate = `${currentYear - 120}-01-01`,
```
CarlosNZ, line 30: Really good find for speeding this up! 👏

I'm just wondering if we need these default values at all? The DatePicker accepts `undefined` for min and max just fine, and I can't see anything immediately wrong with just removing the defaults altogether. Can you?

fergie-nz: You're right I agree - just looking now  and see the date picker defaults to cover 1900 -> 2099 which seems like reasonable bounds and covers a very early date of birth edge case.

I'll remove the defaults

CarlosNZ (approved):
Very nice -- thanks a lot!

I thought there were other places that slow date pickers had been noticed -- can you remember?

fergie-nz:
> Very nice -- thanks a lot!
> 
> I thought there were other places that slow date pickers had been noticed -- can you remember?

Just had a look, but can't see any other reference in an issue - was it maybe just mentioned in chats?

PR #7178, opened 2025-04-02
Fix error in deserializing sql date for plugins

### LINKED ISSUE #7057: Create new inbound shipment firing twice and creates two shipments

### LINKED ISSUE #7075: Initial SOH is not right when creating a program internal order and there 0 stock at the start of the period.
## What went wrong? 😲
Here is what I have explain in that [comment](https://github.com/msupply-foundation/open-msupply/issues/6558#issuecomment-2754911726)
I am not sure that the start date is respected when calculating the initial SOH or that it uses the right one. I am doing a program internal order for the period of March of 2025 so the start date should be the 1st of March 2025. So I assume that because it is a program internal order, the start date would be the start date of the period ie 1st of March 2025. 

So here is the ledger for one of the item of the internal order "Acide Folique 5 MG CP". So, the SOH was 0 until the 8th of March where it received 50 units and then 5 unit went out in a ci on the 11th of March. which put the current stock on hand at 45. 

https://drive.google.com/file/d/1rkRlohlIxXgUY0STBODhd_2PC--1Y_sK/view?usp=sharing

So the initial stock is the stock at the end of day prior to the start date so it should be 0 but it shows 45. It means that the SOH calculation is not right either as it shows 90 (45+50-5) but should be 0 + (- 5) + 50 and gives a MOS of 54 month while it should be half of that. 
https://drive.google.com/file/d/16yqIBC05SBXGpndnKMvJBeRt_jzPJCrS/view?usp=sharing

OG does the same though but I think it is wrong too. It even calculate a SOH of 100 (needs to be verified but it probably does 100 = 45 - (-5) + 50).  
https://drive.google.com/file/d/1mkrGciTCZyBT7zjeNVsztgZHE_dKQHPG/view?usp=sharing

~~So I wonder if the start day use the today date instead of the 1st day of the period in that case ?~~

I did an internal order for the February period and check an item that had stock before the beginning of the period and in that case it seems to calculate properly when there is stock before the start date of the period otherwise it seems to use current stock. Is it possible ?

## Expected behaviour 🤔
If there is no stock at the start of the period, Initial stock should be 0.
 
## How to Reproduce 🔨

Have an item that has never had stock before the beginning of the period but had stock coming in
1. Go to 'Replenishment > Internal order'
2. Click on 'Create a new Internal order'
3. Choose a program your item is part of and the period the current date is on ie period of March 2025 if the date of today is the 27th March
4. See error. Look at your item and you will see that the Initial stock on hand value will show your current stock instead of 0  

## Your environment 🌱

- Open mSupply Version: 2.6.0RC4
- Legacy mSupply Central Server Version: v8-00-17

- Platform: Android

- Database type: SQLlite3

andreievg:
Fixes #7075

I also re-generated bundle straight into this PR: https://github.com/msupply-foundation/civ-plugins/pull/4

This bundle: https://github.com/msupply-foundation/civ-plugins/blob/584a86fae4f6bf63ba8b7713111b6b8ed9a9bcbf/bundle.json

# 👩🏻‍💻 What does this PR do?

When i was testing the use case in issue I created outbound shipment and inventory adjustment in omSupply then backdated them with SQL browser. I noticed 2024-09-08 04:12:27.398858 was failing to deserialize correctly, which failed ledger check. Fixed helper and re-generated plugin bundle

# 🧪 Testing

Need civ setup, as per the issue, I will leave @alainsussol to test, main point is that extra columns must be visible on requsitions, try create requisition for period, make sure there is zero stock for an item. Create adjustment and outbound shipment, backedate those to be within the period, create requisition, should see 0 stock at the start and end stock = stock at the end of the period.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/plugins/backendCommon/utils.ts
+++ b/client/packages/plugins/backendCommon/utils.ts
@@ -41,5 +41,7 @@ export const localDate = (date: Date) => {
 export const sqlList = (list: string[]) => `('${list.join(`','`)}')`;
 export const fromSqlDateTime = (datetime: string) => {
   // Will map '2023-01-01 10:10:10' to UTC '2023-01-01T10:10:10Z'
-  return new Date(`${datetime.split(' ').join('T')}Z`);
+  // Will also map '2024-09-08 04:12:27.398858' to UTC '2024-09-08T04:12:27Z'
+  const withoutMillisecods = datetime.split('.')[0] || datetime;
+  return new Date(`${withoutMillisecods.split(' ').join('T')}Z`);
 };


alainsussol:
@andreievg This fix is working :)

lache-melvin (approved):
Code works for me :) One day I will get a unit test 😢

andreievg:
Thanks @lache-melvin, can't wait for unit tests too

PR #7191, opened 2025-04-03
6526 into 2.7-RC

CarlosNZ:
This is just a copy of @fergie-nz 's original #6889 PR, which he merged into `develop` soon after I created the `v2.7.0-RC` branch. However, I realised that it'd be good to have that in the 2.7 branch as it touches 127 files, so we'll minimise conflicts if people are working off those changes.


fergie-nz (approved):
Looks great thank you Carl

PR #7194, opened 2025-04-03
Upgrade AutocompleteWithPagination styling

### LINKED ISSUE #5336: Store selector window is not wide enough to fit the full names
## What went wrong? 😲

Store selector window is not wide enough to fit the full names

Example attached ->>> [here](https://drive.google.com/drive/folders/1x3PBB7nuL8cKtXFXzc5W90vsEcGFn-Gv)

## Expected behaviour 🤔

2 solutions: either the window can be a bit wider to fit the store names (but how wide?) or there should be a scroll bar/functionality to see longer names. Currently, the users are not able to scroll so the name simply gets cut off. 

## How to Reproduce 🔨

2. Create a new asset
3. Click on the store selector dropdown and see the long names get cut off
4. To recreate, you need a datafile that has a store with a long name or just rename an existing one for testing and find it when selecting from the dropdown

## Your environment 🌱

- Open mSupply Version: V2.4.0-test1

- Platform: Google Chrome (Browser)

- Database type: SQLite3

zachariah-at-msupply:
Fixes part of #5336 

# 👩🏻‍💻 What does this PR do?

Upgrades the styling of the options for a specific component (AutocompleteWithPagination) to truncate options with ellipses rather than wrapping them. (N.B. This only partially solves the original issue, all the other autocomplete components will need to be fixed too, but that will happen in a subsequent PR.)

I hope it looks good :)

# 🧪 Testing

- [ ] Add a store with a really really really really really really really really really long name.
- [ ] Go to Cold chain -> Equipment
- [ ] Create a new asset
- [ ] Click on the store selector dropdown and see the long names get truncated with ellipses
- [ ] Check other similar dropdowns on other pages have not been mucked up.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
```diff
@@ -152,9 +152,15 @@ export const AutocompleteList = <T,>({
...
-            style: {
+            sx: {
               minHeight: height ? `${height}` : 'auto',
               maxHeight: height ? `${height}` : 'auto',
+              '& .MuiTypography-root': {
+                maxWidth: 497,
+                whiteSpace: 'nowrap',
+                overflow: 'hidden',
+                textOverflow: 'ellipsis',
+              },
```
fergie-nz, line 163: I'm actually very hesitant to approve something like this because Autocomplete is used in so many places, and I don't fully understand the ramifications of these changes.

Particularly unsure about the hard coded values. I wonder would this affect other places Autocomplete is used?

CarlosNZ:
@zachariah-at-msupply can you post some screenshots of it's supposed to look? (preferably a few in different contexts)? It's not looking right for me, just wondering if you're seeing something different?

zachariah-at-msupply:
After the modifications I added because I misinterpreted Mark's comment, here is the screenshot you asked for Carl:
image

zachariah-at-msupply:
It also crosses my mind that we might want to add a maximum length for table columns, with an option to expand it to show the full value? The table is scrollable so no information is hidden, but it might be a bit inconvenient at times.


Uploading Screen Recording 2025-04-04 at 12.31.48.mov…

client/packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
GeronimoJohn: This would affect other places for Autocomplete. Though shouldn't be harmful bur rather helpful.

lache-melvin:
Hi - what's holding up this PR?

client/packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
```diff
@@ -151,9 +151,15 @@ export const AutocompleteList = <T,>({
             },
           },
           listbox: {
-            style: {
+            sx: {
               minHeight: height ? `${height}` : 'auto',
               maxHeight: height ? `${height}` : 'auto',
+              '& .MuiTypography-root': {
+                maxWidth: 497,
```
GeronimoJohn, line 158: I think if possible it would be ideal that we don't set the maxWidth here, rather use up the remaining space left in the div. Not sure if this will work but e.g.

```ts
              {
                width: '100%',
                flex: '1',
                whiteSpace: 'nowrap',
                overflow: 'hidden',
                textOverflow: 'ellipsis',
              }
```

GeronimoJohn: If it doesn't work then static values the way

GeronimoJohn: I tried, doesn't work

zachariah-at-msupply:
I've come across these other issues (listed below) which are very similar. Shoul I include them in this PR, or create new issues for them? Seeing these other issues also makes me feel uneasy about the fix I have made - it seems like more sweeping changes are necessary than these fixes, which feel like putting a plaster on a bigger problem that we could address. What are your thoughts?
Here are the 3 things I found:
1. All the dropdowns that use MUI's autocomplete component also have problems with long names: 
image


3. The tables also have problems with long names:
https://github.com/user-attachments/assets/9b336898-5d15-4bfa-8272-46171049b2e7


4. The fix I've proposed is a trade off and will only work in a limited number of use cases (see my unresolved review comments, above).

CarlosNZ:
Screenshot 2025-04-16 at 10 39 57 AM

Hey @zachariah-at-msupply, I'm still seeing this in the Cold Chain->Equipment Store Selection, which is the one raised in the original issue.

zachariah-at-msupply:
N.B. I reverted the previous change as I did not know it was worth introducing a fixed width to the autocomplete for an issue that no one has actually raised - I think that might require a bit of design input, and is probably linked to the wider issue of fixing styling throughout the application.

CarlosNZ (approved):
Okay, I think we should get this in now, the whole issue has become too confused. It looks good for the Store selector in Assets, and that's the main thing for this specific issue. If further changes are needed we can do them in a new issue (with much more specific requirements!)

Thanks @zachariah-at-msupply !

Issue #7034, opened 2025-03-24
Ability to use vaccination card cross sites


andreievg:
## Is your feature request related to a problem? Please describe 👀

Use case, vaccination can be given at different sites:
* At birth place hospital
* At local clinic
* As outreach program

## Describe the solution you'd like 🎁

Similar to patient prescriptions, encounters and other documents, should sync vaccination cards to all sites where patient is visible.

## Use cases to consider

What if there is no internet at a site where patient previously visited, it is possible to wrongly vaccinate, how do we deal with this ?

## On Technical side

This needs a lot of thought since vaccination card is stored on central server, but the patient might not be available there, so maybe we should sync this through mSupply, but we would need to examine the document I think, to get encounter_id <-> this seems to be how it's linked to the patient.

**quick solution** this solution won't scale very well, but is suitable for smaller country for now

* Sync vaccine cards everywhere, change this in changelog filter for this type
* Make patient sync everywhere (make sure documents and encounters sync everywhere), this should be configuration

This case still has the issue of internet not available and old data present (i.e. might not see previous vaccinations given)

**more thorough solution**

Since our patient related record travel based on name_store_visibility, which may not exist on omSupply central server, along with the patient themselves, we would need:

When pulling patient from OG central (when patient is first 'found' through central lookup) we also:

* Add name_store_join for the patient in omSupply via API call
* Pull related data for that patient (vaccine cards)
* omSupply central will forward all future vaccine cards to places with name_store_join (and make sure we don't sync back vaccine cards from site that just pushed them (we already have this logic)

In more detail, we currently have:

When patient is `found` by omSuppy on mSupply central server, what currently happens:
* Send an api request to add name_store_join in OG, this will also queue all of the records for that patient to omSupply
* Do a full sync cycle, where all patient records are pulled the same way as normal sync.

Vaccine cards are pushed to omSupply central, with store_id being the identifier of which site this record will sync to. During initialisation we would sync those record back to the remote site.

Let say we do something similar to what we do in mSupply and create name_store_join on omSupply for patient when it's request by omSupply remote site from mSupply central, there are a few problems/use cases to consider

Problems/use cases:

1 - We need to know which patient the vaccine card belongs to, we would need to add patient_link_id to vaccine card, but patient does not exist on omSupply central server
2 - We need to add name_store_join to omSupply for it to sync related vaccine cards, but patient would not exist on omSupply central
[...]

PR #7206, opened 2025-04-03
7160 fetch v6 patient records when fetching patient

### LINKED ISSUE #7160: Add include_patient_records to v6 sync

Remote site can add a name_store_join for a patient to OMS central server. From then on, COMS will sync patient records to the new repo. However, we need to send the patient records up until the name_store_join was added.

Add param to v6 sync, from manualSync mutation, to pass the patient id that you want to query records for. Should start from cursor 0 and pull all related records.

lache-melvin:
Fixes #7160

# 👩🏻‍💻 What does this PR do?

When fetching a patient, you want to retrieve all of their historic records.

Adds a `fetch_patient_id` to manual sync, which will return all historic vaccination records for the patient on the fetch_patient sync cycle.

# 🧪 Testing

- [ ] Create a new patient on one site
- [ ] Enrol them in a vaccination program
- [ ] Create an encounter and give a few vaccinations
- [ ] From another site, sync (so we know the cursor is up to date, to test the "getting historical records")
- [ ] Fetch this new patient
- [ ] The fetch succeeds
- [ ] Check vaccination table for this site - the historic vaccination records were retrieved

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/graphql/programs/src/queries/link_patient_to_store.rs
```diff
@@ -75,6 +75,8 @@ fn map_result(
         })),
         Err(err) => {
             let formatted_error = format!("{:#?}", err);
+            println!("Error while linking patient: {}", formatted_error);
```
lache-melvin, line 78: Added this so could see a bit more in the logs when I was getting errors, `ConnectionError` just returns a generic message to the frontend. Happy to leave?

lache-melvin:
Sorry just added one more commit - thought best to filter to check name_store_join exists before sending all the patient records 😅

lache-melvin:
Hi @andreievg - just pushed fix for the ambiguous name_link join, will now work on the patient_pull endpoint 🔥

lache-melvin:
> will now work on the patient_pull endpoint 🔥

@andreievg pushed fix for this, compiles but haven't tested yet - I'm looking into prod issue so wanted to share progress for review/get 2.7 out 🙏

client/packages/system/src/Sync/api/hooks/sync/useManualSync.ts
```diff
@@ -5,7 +5,13 @@ export const useManualSync = () => {
...
   const queryClient = useQueryClient();
 
-  return useMutation(api.manualSync, {
+  const mutation = useMutation(api.manualSync, {
     onSettled: () => queryClient.invalidateQueries(api.keys.syncInfo()),
   });
+
+  return {
+    ...mutation,
+    // map react-query required input of "string | undefined" to optional string
```
andreievg, line 14: Thanks for comment

server/graphql/programs/src/queries/link_patient_to_store.rs
andreievg: James mentioned a few times that our logging for api is not great, added issue: https://github.com/msupply-foundation/open-msupply/issues/7243

server/service/src/sync/central_data_synchroniser_v6.rs
```diff
@@ -195,6 +195,51 @@ impl SynchroniserV6 {
...
+    // Replaces the `pull` step if running manual sync for a fetched patient
+    pub(crate) async fn patient_pull<'a>(
+        &self,
+        connection: &StorageConnection,
+        batch_size: u32,
+        fetch_patient_id: String,
+        logger: &mut SyncLogger<'a>,
+    ) -> Result<(), CentralPullErrorV6> {
+        // Temp cursor for patient pull
+        let mut patient_cursor: u64 = 0;
```
andreievg, line 207: Yay

server/graphql/programs/src/queries/link_patient_to_store.rs
andreievg: Can you use log::error! here please

server/repository/src/db_diesel/vaccination_row.rs
```diff
@@ -38,8 +38,9 @@ table! {
     }
 }
 
+// NOTE: both patient_link_id and facility_name_link_id are foreign keys to name_link
+// so not defining a default joinable here, so as not to accidentally join on the wrong one
```
andreievg, line 42: this makes sense, thanks for comment

server/graphql/programs/src/queries/link_patient_to_store.rs
lache-melvin: log is not available in graphql layer... i can move to log from service call?

server/repository/src/db_diesel/vaccination.rs
```diff
@@ -130,7 +131,11 @@ fn create_filtered_query(filter: Option<VaccinationFilter>) -> BoxedVaccinationQ
     let mut query = vaccination::table
         .left_join(clinician_link::table.inner_join(clinician::table))
         .inner_join(vaccine_course_dose::table)
-        .left_join(name_link::table.inner_join(name::table))
+        .left_join(
```
andreievg, line 134: btw, looks like this was never actually used for sorting, sub query would have worked I think

server/repository/src/db_diesel/changelog/test.rs
```diff
@@ -701,3 +702,62 @@ async fn test_changelog_outgoing_sync_records() {
     assert_eq!(outgoing_results.len(), 1);
     assert_eq!(outgoing_results[0].record_id, asset_class_id);
 }
+
+#[actix_rt::test]
+async fn test_changelog_outgoing_patient_sync_records() {
```
andreievg, line 707: thanks for test, looks good

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -245,6 +246,21 @@ pub struct ChangelogRepository<'a> {
 
 type ChangelogJoin = (ChangelogRow, Option<NameLinkRow>);
 
+impl ChangelogRow {
+    pub fn from_join((row, name_link): (ChangelogRow, Option<NameLinkRow>)) -> Self {
+        ChangelogRow {
```
andreievg, line 251: I really like these helpers, in other repositories we call them `to_domain`, with some variations

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -566,6 +598,50 @@ fn create_filtered_outgoing_sync_query(
...
+// This is a manual sync to fetch all records for a specific patient
+// Managed via own cursor
+fn create_filtered_outgoing_patient_sync_query(
+    earliest: u64,
+    sync_site_id: i32,
+    fetch_patient_id: String,
+) -> BoxedChangelogQuery {
+    let mut query = changelog_deduped::table
+        .left_join(name_link::table)
+        .filter(changelog_deduped::cursor.ge(earliest.try_into().unwrap_or(0)))
```
andreievg, line 610: I was trying to make a petter of these methods, but it hasn't fully taken off:

https://github.com/msupply-foundation/open-msupply/blob/98b80b3bf57e87ee199ac74b5f4bbe0d82d8ed2e/server/service/src/lib.rs#L289-L311

server/repository/src/db_diesel/changelog/changelog.rs
lache-melvin: Yeah, I wasn't quite sure here, I thought `to_domain` was more for mapping to the non `-Row` structs, but could call this the same I suppose

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -566,6 +598,50 @@ fn create_filtered_outgoing_sync_query(
...
+        .into_boxed();
+
+    let active_stores_for_site = store::table
```
andreievg, line 613: I actually agree here with the dry choice, although if active_store_for_site was a new method it would have been good too, but something about this simple active_store_for_site variable makes more sense (even though it appears in two places)

server/repository/src/db_diesel/changelog/changelog.rs
lache-melvin: ohh tbh this was copy and paste, didn't realise it was converting a type - I like these helpers! Can update

andreievg (changes requested):
Below has changed, see comment after this one (we would just sync patients and name store join from omSupply remote to omSupply central)

This looks awesome. I think we need to create name_links when creating name_store_joins, I can't see it here:  https://github.com/msupply-foundation/open-msupply/blob/98b80b3bf57e87ee199ac74b5f4bbe0d82d8ed2e/server/service/src/apis/api_on_central/mod.rs#L69-L80. 
Would that mean we can add name_link_id to change_log ? I think so, but we would need to remove reference from name_link.name_id..

There is also a small change request for error logging too

Loving your changelog filter work, technically could do some consolidation there, I haven't done dev today, so going to spend 20 mins to have a go : )

andreievg:
I've added the following to [this issue:](https://github.com/msupply-foundation/open-msupply/issues/7034)

In more detail, we currently have:

When patient is `found` by omSuppy on mSupply central server, what currently happens:
* Send an api request to add name_store_join in OG, this will also queue all of the records for that patient to omSupply
* Do a full sync cycle, where all patient records are pulled the same way as normal sync.

Vaccine cards are pushed to omSupply central, with store_id being the identifier of which site this record will sync to. During initialisation we would sync those record back to the remote site.

Let say we do something similar to what we do in mSupply and create name_store_join on omSupply for patient when it's request by omSupply remote site from mSupply central, there are a few problems/use cases to consider

Problems/use cases:

1 - We need to know which patient the vaccine card belongs to, we would need to add patient_link_id to vaccine card, but patient does not exist on omSupply central server
2 - We need to add name_store_join to omSupply for it to sync related vaccine cards, but patient would not exist on omSupply central
3 - Name link id should really be added to changelog for vaccine card changes on omSupply central, for ease of filtering, but patient may not exist on omSupply central
4 - If patient is created on remote site, name store join is not available on omSupply central (and we would need to use store_id sync for it)
5 - Have to make sure we only sync patients to site that did not edit them last (we do have mechanism for it, last_sync_site_id on changelog)

Solutions:

a - Use "Names created in other stores not visible in this store" and dummy dispensary store on central, to get all patients, this would require a sync before we integrate vaccine card, and we still have some data missing. I also tried this and got all patients after initialisation, but future patients added on other remote sites didn't come through v5 ¯\_(ツ)_/¯

b - Sync patient and name_store_joins to omSupply central, as well as mSupply, this is quite simple, as per [this branch ](https://github.com/msupply-foundation/open-msupply/compare/develop...develop(Sync-patients-to-omSupply-central))(sorry it's from develop, not 2.7 RC), also this branch does not use proper v6 translations, and actually uses legacy translation (but can easily change to v6 translations in the future by creating a new translator). Add endpoint to create name_store_join when other remote site requests that patient from mSupply. Add name_link_id on changelog and filter by that name_link_id + name_store_join for current sites, for vaccine card records. With this solution don't need to use store_id on vaccine record to sync it back to the original site.

For **b**, since we don't add sync record to the queue, and rely on changelog, we would need to add patient id when requesting sync, and have a temporary cursor starting at 0.

I think **b** would deal with omSupply use cases, but there is another use case that we would need to deal with in near future, and that is if patient was created on mSupply site, for this use case, when name_store_join is requested to be created in omSupply central, we would also need to sync that patient from mSupply (as if omSupply central was requesting that patient from mSupply central)

andreievg:
Here is an example of changelog filter by name_link_id and consolidating part of the query https://github.com/msupply-foundation/open-msupply/compare/7160-include-patient-records-v6...7160-include-patient-records-v6(Use-name-link-filter-on-changelog-and-consolidate)

andreievg (approved):
Approving to merge as is, with follow up PR to make the changes in the comment (updated details in the original issue), which is syncing patient to omSupply Central as well as mSupply central, removing reference removal and doing this change: https://github.com/msupply-foundation/open-msupply/compare/7160-include-patient-records-v6...7160-include-patient-records-v6(Use-name-link-filter-on-changelog-and-consolidate)

PR #7218, opened 2025-04-04
7100 show vax card in non home store

### LINKED ISSUE #7100: Show vaccination card in stores where patient was not initially enrolled

Currently, if viewing a vaccination encounter from a store other than the enrolling store, the UI hides the vaccination card to prevent accidental administration of doses. However, we now to need to use the card cross-sites in special cases.

- Make the vax card visible even if you are not the enrolling store.
- When giving a vaccine, show a warning - have you synced lately? Other stores may not have synced! Ensure you aren't giving the patient an accidental/additional dose 😱 

> Note to self, would be worth testing what the card ends up doing if you do create 2 "second dose" records in 2 different stores as they weren't synced, do we need to handle the duplicates 👀

lache-melvin:
Fixes #7100

# 👩🏻‍💻 What does this PR do?

Enables display of vaccination card on stores other than the one the patient enrolled at.

Also adds warning to the vaccination modal:

(I'm always displaying this for new doses... I don't think there's a reliable way to check if the patient has been fetched by any other store, so we should always show just in case?)

Screenshot 2025-04-04 at 3 40 28 PM

There's definitely some edge cases if the different stores start editing existing vaccinations (theoretically, give in store A, store B pulls, marks it as not given, would sync back to store A... but store A still has a prescription record from that stock going out...)

However, I think we are way far down the edge case rabbit hole, and I don't think implications of that are massive anyway.

# 🧪 Testing

- [ ] Create a new patient on one site
- [ ] Enrol them in a vaccination program
- [ ] Create an encounter and give a few vaccinations
- [ ] From another site, fetch this new patient
- [ ] View the patients vaccination encounters - you can see the card
- [ ] Create a new encounter, give some more vaccinations
- [ ] Sync
- [ ] Go back to first store and sync - the new vaccination records from other store should show

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


aimee-mcneil-melville (approved):
Looks good Lache! Nice cleanups throughout too 🚀 

I found a couple of bugs for the (lack of) linking of clinicians and prescriptions which I have raised here: #7228 and #7232 

Tested with multiple changes of the vax cards across two stores, and all otherwise working well. Noting the lack of stock movements - I would think it to be safer to _not_ record movements from a different store anyway

Just a merge conflict to resolve then good to go!

lache-melvin:
Thanks Aimee, especially for those bug catches 🔥 Will merge this once downstream PR approved :)

Issue #4744, opened 2024-09-05
Table Refactor Placeholder


andreievg:
## The suggestion

Need to investigate a solution that can deal with these issues:

#4043 
#3070 
#4538 
Better CSS management, e.g. https://github.com/msupply-foundation/open-msupply/pull/7220#discussion_r2030440590

andreievg:
While checking performance of basic tables it would be great to use react profiler from react dev tools. I think it would be very helpful to have a look at RnR form commits and find the performance improvement <- use pre performance improvement state of RnR for to understand how profiler works (it should ideally lead you to the underlying problematic text input), ideally this exercise would help in understanding how profile works and how we can use it to bench tables (and identify slow points of current tables)

lache-melvin:
Also #5273

noel-yeldos:
Also just a note to review table styling. On detail view, vertical scroll isn't visible until you are all the way to the right
![telegram-cloud-photo-size-1-5127494741976526434-x](https://github.com/user-attachments/assets/be1cb5a9-79e6-482a-810b-15a7ac28de3b)
![telegram-cloud-photo-size-1-5127501944636681663-x](https://github.com/user-attachments/assets/a3f20469-f428-4943-8d3b-05ed87b1ff29)

lache-melvin:
@andreievg would be cool to link your basic table test here!

andreievg:
https://github.com/andreievg/Example-table-state-and-table-via-composition/blob/main/src/SimpleEditableTable.tsx

andreievg:
Working on this in a linked issue

PR #7220, opened 2025-04-04
6666 Make age column width relative vaccination card

### LINKED ISSUE #6666: Encounters: age column is not wide enough
## What went wrong? 😲

Image

## Expected behaviour 🤔

Wider pls

## How to Reproduce 🔨

1. Make sure you have Immunization Programs and Vaccination Card enabled for your store
2. Create an encounter for a patient enrolled into a program
3. Create a vaccine course with a few doses, possibly for ages 10+
4. Go to the encounter to issue vaccination -> see some values do not fit into the Age column

## Your environment 🌱

- Open mSupply Version: V2.6.0-RC2

- Database type: PostgreSQL

noel-yeldos:
Fixes #6666

# 👩🏻‍💻 What does this PR do?
Change age column width to be relative. It now resizes based on the text in the field. 

Screenshot 2025-04-04 at 17 28 31
Screenshot 2025-04-04 at 17 29 16

I initially just thought of increasing the fixed width but then it looked funky when there was only age in months being displayed. 

The columns are inheriting width which makes them ridiculously wide if you remove the width prop altogether
Screenshot 2025-04-04 at 17 27 30

I tried to go up the chain to see if I could track down where that was being inherited from but I didn't have much luck so I came to this solution. 

# 🧪 Testing

- [ ] Make sure you have Immunization Programs and Vaccination Card enabled for your store
- [ ] Create an encounter for a patient enrolled into a program
- [ ] Create a vaccine course with a few doses
- [ ] For scenario 1 you will need an age that has both years and months e.g 10 years 12 months
- [ ] For scenario 2 you should only have ages in months
- [ ] Go to the encounter to issue vaccination -> see that the age column adjusts to fit the length of the text correctly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -165,7 +165,7 @@ const VaccinationCardComponent = ({
           // Only show age label for first of each "block", when repeated
           return sameAsPrev ? null : getAgeLabel(rowData);
         },
-        width: 140,
+        width: '5%',
```
lache-melvin, line 168: Hm, so looks like `5` of the 5% isn't actually for a specific purpose, just that its smaller than actual width so stretching to min-content... in this case, I think safer to go with 1% (as its more obviously not the actual width) and add a comment:

```suggestion
        // Hack for "min-content"
        width: '1%',
```

I have added this to table refactor issue, something else to consider in new table API definition: https://github.com/msupply-foundation/open-msupply/issues/4744

lache-melvin (changes requested):
Thanks for finding this fix! Just requesting comment around it for clarity

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
noel-yeldos: Resolved here 6cdeb7c49460d0136127b63faa407c204e7fb20a

lache-melvin (approved):
Thank you!

PR #7222, opened 2025-04-04
2.6.1 main to develop

CarlosNZ:
Main back to develop after 2.6.1 release

Should only have changes since the last RC (RC6) I would say.


CarlosNZ:
@roxy-dao -- there's 2 server tests that are failing in this branch, but weren't in either `main` or `develop`. I've had a look and changed the "prescribed_quantity" in the place it was failing, but still no good. Can you have a quick look when you have a moment? (And once you're feeling better, of course)

server/service/src/sync/test/test_data/invoice_line.rs
GeronimoJohn, line 1: Found the issue for the failing test - there are a few `prescribed_quantity: Some(0.0)` that are left in this file. Just have to change to `prescribed_quantity: None`. Passed for me locally when I did that

roxy-dao (approved):
Think ok to get in~?

CarlosNZ:
Not required now after merging #7248

PR #7226, opened 2025-04-06
#7113 Outbound alerts layout

### LINKED ISSUE #7113: Outbound issue modal issue input jumps to left and shows NaN
Image

## What went wrong? 😲

I first noticed an issue where issue input jumps to the right, when info box shows up about expired items, but then on further testing also had NaN show up <- I think this is due to pack size = 0, no idea how this is demo data, but would assume it's data error from pervious footruner

## Expected behaviour 🤔

Input don't jump around, NaN is not shown

## How to Reproduce 🔨

My window size for this was "1912-966" (that's not standard of course but very possible for window to be resized this way)

See screen shot and screen capture, I was on store Kopu on demo server:
https://drive.google.com/drive/u/1/folders/1oIW2Qc3dSqzm1_gZsK9E_CY1NIO0gNDM

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #7113

# 👩🏻‍💻 What does this PR do?

Have shuffled the contents around a bit so that the alerts appear to the right of the "Issue" UI rather than push in in front of it:

Screenshot 2025-04-07 at 11 26 23 AM

Also tweaked padding and top-alignment to make it look okay with multiple alerts

Haven't addressed the "NaN" problem, as it seems to be due to a pack size being set to 0, as discussed in the original issue. Let me know if you find if/where this is not the case.

# 🧪 Testing

- [ ] Create outbound shipment
- [ ] Add an item
- [ ] Issue quantities such that you get an alert appearing -- confirm layout is acceptable
- [ ] Try and get 2 alerts at the same time ("Not enough stock" and "Some are expired", and also confirm layout is acceptable

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditForm.tsx
@@ -210,8 +210,8 @@ export const OutboundLineEditForm: React.FC<OutboundLineEditFormProps> = ({
       {item && canAutoAllocate ? (
         <>
           <Divider margin={10} />
-          <Box display="flex" alignItems="center">
-            <Grid container alignItems="center">
+          <Box display="flex" alignItems="flex-start" gap={2}>
+            <Grid container alignItems="center" pt={1}>
               <ModalLabel label={t('label.issue')} />
               <NumericTextInput
                 autoFocus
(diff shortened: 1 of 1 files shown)


GeronimoJohn (approved):
LGTM! Works well 🚀

PR #7251, opened 2025-04-08
6767 sync item warnings

### LINKED ISSUE #6767: Should sync item warnings and print on prescription labels
## What went wrong? 😲

We don't sync warnings so cannot print them on labels. We should sync and print the warnings.

## Expected behaviour 🤔

When printing labels in dispensary 

## How to Reproduce 🔨

1. Have an item with a warning defined
2. Print labels in OMS
3. No warning

## Your environment 🌱

- Open mSupply Version: 2.6.0-RC6

## Agreed solution

Turns out in OG we have tables with relationships

`warning <- item_warning_link -> item`

And the link/join table has a `priority` field. You can have many warnings associated with an item. The priority field indicates which gets printed. 

I will go with this for now although, ideally multiple warnings should be printed, and ideally we'd spill over to multiple labels. THAT is out of scope though.

aimee-mcneil-melville:
Fixes #6767

# 👩🏻‍💻 What does this PR do?

Warnings are synced from OG, and printed on prescription dispensing labels

warning-labels

Warnings are on the item as set in OG, so have been added to the item and accessed via a loader. The warning data received into the prescription is:

Column Name | Type
-- | --
id | String
itemId | String
warningText | String
code | String
priority | bool

As warnings can only be configured in OG, and at this stage we only want to print one warning on the label, there is no front end or UI for the warnings. The label is generated in the front end, so the warning data is accessed by the prescription graphql call and then the label generator has logic for which warning to print.

Whilst only one warning is being printed, we are selecting the first warning with a priority 'true'. If all are 'false' we select the first of those.

The logic could have probably been done server side instead - however I did this in the front end as it may have a UI in future, and will be easier to update when we make changes for multiple warnings across multiple labels.

Tests added in front end label generation, and for the upsert/sync

- Currently a warning being deleted in OG (either removed from the item, or deleted from the main warnings list) will remain in OMS
- The 'first' warning currently is the most recently created in OG
The decision was made to change naming from `ItemWarningLink` to `ItemWarningJoin` as it was more appropriate to the join structure

# 🧪 Testing

- [ ] Set up warnings in OG:
  - [ ] Go to Special -> Warnings -> Create some warnings
  - [ ] Go to items->  items -> select an item
  - [ ] Misc tab -> Add the warnings
  - [ ] Tick some as priority (take note of which warnings on which items)

- [ ] Create a prescription with items that have warnings configured
  - [ ] Add the line edit view, add some directions of various lengths -> Save
  - [ ] From detail view, go to `Print Options -> Print Labels` - alternatively select some lines and use the `Print labels` button in the footer
  - [ ] See one warning printed on the label. If any warnings with `priority - true` were on the item, it should be one of those that had printed, instead of a `priority - false` warning

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/Prescriptions/DetailView/hooks/utils.tsx
```diff
@@ -33,6 +35,11 @@ export const groupItems = (
     const firstItem = items[0];
     const itemWithNote = items.find(item => item.note) || firstItem;
 
+    // gets the first warning with priority 'true'
+    const priorityWarning =
```
aimee-mcneil-melville, line 39: 'items' is all batches of one item - if there are multiple batches of an item issued, data for the label is from the first batch

client/packages/invoices/src/Prescriptions/api/operations.generated.ts
```diff
@@ -105,6 +113,14 @@ export type PrescriptionRowFragment = {
             itemId: string;
             priority: number;
           }>;
+          warnings: Array<{
```
aimee-mcneil-melville, line 116: Duplication here is needed - a prescription contains both the item and the stockline (which contains the item node). Thus the warning node is needed on both item nodes

server/repository/src/db_diesel/item_warning_join.rs
```diff
@@ -0,0 +1,123 @@
...
+    (item_warning_join_row, (_item_link_row, item_row), warning_row): ItemWarningQueryJoin,
+) -> ItemWarningJoin {
+    ItemWarningJoin {
+        warning_row,
+        item_row,
+        item_warning_join_row,
+    }
+}
+
+type BoxedItemWarningJoinQuery = IntoBoxed<
```
aimee-mcneil-melville, line 96: Joining the relevant tables together - item_link is needed wherever there is a join to item

server/service/src/print/label.rs
```diff
@@ -9,20 +9,19 @@ use super::jetdirect::{Jetdirect, Mode};
 pub struct AssetLabelData {
     code: String,
     asset_number: Option<String>,
-    date_printed: Option<String>
+    date_printed: Option<String>,
```
aimee-mcneil-melville, line 12: A few bits of auto-formatting in this file

server/service/src/sync/translations/item_warning_join.rs
```diff
@@ -0,0 +1,82 @@
...
+pub(super) struct ItemWarningJoinTranslation;
+impl SyncTranslation for ItemWarningJoinTranslation {
+    fn table_name(&self) -> &str {
+        "item_warning_link"
+    }
+
+    fn pull_dependencies(&self) -> Vec<&str> {
+        vec![
+            ItemTranslation.table_name(),
+            WarningTranslation.table_name(),
```
aimee-mcneil-melville, line 37: Ensure the ItemTranslation and WarningTranslation are complete prior to attempting the join - otherwise foreign key errors occur as the referenced ids don't exist

server/service/src/sync/translations/item_warning_join.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::sync::translations::{item::ItemTranslation, warning::WarningTranslation};
+use repository::{ItemWarningJoinRow, StorageConnection, SyncBufferRow};
+
+use super::{PullTranslateResult, SyncTranslation};
+
+#[derive(Deserialize, Serialize)]
+
+pub struct LegacyItemWarningJoinRow {
+    #[serde(rename = "Spare")] // Spare column not used in oms
+    spare: f64,
```
aimee-mcneil-melville, line 12: 'Spare' is a column on the original table in OG. We don't want/need to use it in OMS, however we need to ensure it is handled coming in

Then in line 48 'result' we can use the rest of the keys without using 'spare'

client/packages/invoices/src/Prescriptions/api/operations.graphql
```diff
@@ -529,6 +547,9 @@ fragment historicalStockLine on StockLineNode {
     itemDirections {
       ...ItemDirection
     }
+    warnings {
+      ...Warning
```
andreievg, line 551: Do you need this on StockLineNode ? Would it be enough to just have it on PrescriptionLine ?

server/repository/src/db_diesel/item_warning_join.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use crate::{
+    db_diesel::item_row::item, diesel_macros::apply_equal_filter, item_link,
+    repository_error::RepositoryError, DBType, EqualFilter, ItemLinkRow,
+};
+use diesel::dsl::InnerJoin;
+use diesel::{dsl::IntoBoxed, prelude::*};
+
+#[derive(Clone, Default, PartialEq, Debug)]
+pub struct ItemWarningJoin {
```
andreievg, line 14: > The decision was made to change naming from ItemWarningLink to ItemWarningJoin as it was more appropriate to the join structure

This repository should be item warning, it is 'domain' kind of object so can be named anything (it doesn't need to reflect database table name)

server/repository/src/db_diesel/item_warning_join.rs
```diff
@@ -0,0 +1,123 @@
...
+    ItemWarningJoin {
+        warning_row,
+        item_row,
+        item_warning_join_row,
+    }
+}
+
+type BoxedItemWarningJoinQuery = IntoBoxed<
+    'static,
+    InnerJoin<
```
andreievg, line 98: Nice hopefully this wasn't too hard

server/repository/src/db_diesel/item_warning_join_row.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct ItemWarningJoinRowRepository<'a> {
+    connection: &'a StorageConnection,
+}
+
+impl<'a> ItemWarningJoinRowRepository<'a> {
+    pub fn new(connection: &'a StorageConnection) -> Self {
+        ItemWarningJoinRowRepository { connection }
+    }
+
+    pub fn upsert_one(&self, row: &ItemWarningJoinRow) -> Result<(), RepositoryError> {
```
andreievg, line 43: I think repositories should be made just with the methods that are used, and extra methods added when needed

server/service/src/print/label.rs
andreievg: Thanks for self review

server/service/src/sync/translations/item_warning_join.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    fn try_translate_from_upsert_sync_record(
+        &self,
+        _connection: &StorageConnection,
+        sync_record: &SyncBufferRow,
+    ) -> Result<PullTranslateResult, anyhow::Error> {
+        let data = serde_json::from_str::<LegacyItemWarningJoinRow>(&sync_record.data)?;
+
+        let result = ItemWarningJoinRow {
+            id: data.id,
```
andreievg, line 49: I like destructure and restructure here, easier for compiler to tell you what's wrong

andreievg (commented):
Looks great, nothing to stop approval, but have some suggestion for improvements, will give you a chance to have a look and make changes before approval for next RC, main changes:

Rename the repository
Remove extra filters and other repository things that are not used
Remove the extra warning on item on stockline in operation.graphql

Btw I really like 'consumer' based API, consumer being the code that uses the API, i know we already had code around this, so not asking for refactor, but I would probably shape everything to have less re-mapping. Just food for thought, thanks

client/packages/invoices/src/Prescriptions/api/operations.graphql
aimee-mcneil-melville: Oh yes it can be removed from the stockline - I thought it had needed to be there. Thanks!

server/repository/src/db_diesel/item_warning_join.rs
aimee-mcneil-melville: I got some help with this and learnt a lot! 😁

server/repository/src/db_diesel/item_warning_join.rs
aimee-mcneil-melville: Renaming for repository here: 59c5c6a

server/repository/src/db_diesel/item_warning_join_row.rs
aimee-mcneil-melville: Thanks Andrei! I have updated here: 155d858

server/graphql/general/src/queries/item.rs
```diff
@@ -55,6 +55,7 @@ pub struct ItemFilterInput {
     pub is_active: Option<bool>,
     pub is_vaccine: Option<bool>,
     pub master_list_id: Option<EqualFilterStringInput>,
+    pub item_warning_join_id: Option<EqualFilterStringInput>,
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Ah nice, it's not! I have taken out here: 3aeeaaf

server/service/src/sync/translations/item_warning_join.rs
aimee-mcneil-melville: I agree - I thought the 'Spare' column was preventing this but I had another go at it and got it working. 

Updated here: b4ba6fe

aimee-mcneil-melville:
Thanks for the review Andrei, lots to take on board and learn from for the future 😄

andreievg (approved):
Thanks for changes, looks good !

PR #7253, opened 2025-04-08
7187 record not given vaccination date

### LINKED ISSUE #7187: Date of not-given vaccinations is not recorded
## What went wrong? 😲

#6391 raised that vax card still showed `date given` even when the vaccination was set to "Not given". There was a 2.6.1 change to make `vaccination_date` field nullable, and only save on "given" vaccinations.

However, now there is no date recorded for "not given" vaccinations.

High severity - data loss

I enter a "not given" vaccination record for the 1st of april:
Image

But not date saved, so when i reopen, there is no date, it shows default todays date:
Image

## Expected behaviour 🤔

From what I understand (like 3rd hand though, @roxy-dao spoke to @richardmoizeau) 

- We don't need to show the date "not given" in the card
- But the date should be recorded, both to show in the modal, and for use in dashboards

Solutions:
- Just always save the `vaccination_date`, and only show in card if status is Given
- Maybe name `vaccination_date` is confusing, in relation to not given. Could add another column, "date_not_given" or something
  - More biz logic to add though, if you change from given to not given etc. need to clear these
  - If this is a concern, maybe we just rename `vaccination_date` to `visit_date` or something more generic, but keep only the one date?

## How to Reproduce 🔨

1. Go to vaccination encounter
2. For a vaccination, set date to not today, and mark as not-given
3. Reopen that vaccination - date shows today

## Your environment 🌱

- Open mSupply Version: v2.6.1 RC5

- Platform:

- Database type:

zachariah-at-msupply:
Fixes #7187

# 👩🏻‍💻 What does this PR do?

This PR makes it so that a vaccine's date is always saved, whether it was given or not given.
It also makes sure that the vaccine's date is not displayed unless the vaccine's status is 'given'.

The weather is definitely turning!

# 🧪 Testing

- [ ] Set up vaccinations so that you can assign a vaccine dose to a patient.
- [ ] Go to Dispensary > Patients and create an encounter with a patient.
- [ ] Add a vaccine dose to that encounter.
- [ ] Go to the vaccine tab of the encounter you created and bring up the vaccine modal.
- [ ] Make sure the status of the vaccine is 'not given' and enter a date (any date, as long as it's not today).
- [ ] Close the modal and check that 'N/A' is displayed in the 'date given' column.
- [ ] Bring up the vaccine modal again and check that the date you entered is still displayed, and that it is not today's date (which is the default if it hasn't saved the date you entered).
- [ ] Change the status to given.
- [ ] Exit the modal and check that the date you entered is displayed in the 'date given' column.
- [ ] Save the encounter and mark it as visited.
- [ ] Go to Dispensary > Patients and double click on the patient you gave the vaccine to.
- [ ] In that patient's detail view, select the vaccine tab, and double click on the course that you just gave them a vaccine in.
- [ ] Check the date given is displayed.
- [ ] Change the vaccine to not given and check that date given is recorded as N/A.

# 📃 Documentation

- [x] **These areas should be updated or checked**: I don't think the vaccine documentation includes this sort of detail yet, but it could probably use a few screenshots.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -210,7 +210,13 @@ const VaccinationCardComponent = ({
...
         label: 'label.date-given',
-        accessor: ({ rowData }) => localisedDate(rowData.vaccinationDate ?? ''),
+        accessor: ({ rowData }) => {
+          if (rowData.status === VaccinationCardItemNodeStatus.Given) {
+            return localisedDate(rowData.vaccinationDate ?? '');
+          } else {
+            return t('messages.not-applicable');
+          }
+        },
       },
```
zachariah-at-msupply, line 220: I'm not 100% sure this is appropriate to put in the accessor, but didn't know where else it could go - what do you think?

client/packages/plugins/backendTypes/generated/ApprovalStatusType.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
```
CarlosNZ, line 1: Don't think these files should be in this PR.

zachariah-at-msupply: Ah ok, would you mind explaining how I can spot files like this which shouldn't be committed in future?

CarlosNZ: It's all the files in this folder, they probably came from a different base branch, not sure. But you should have a scan through this PR and get rid of any changes you didn't make.

You should be able to see when you commit which files have changed, so it's a good idea to always have a quick scan for any that shouldn't be here. In this case, it's probably because you started from a branch other than 2.7.0, perhaps?

zachariah-at-msupply: Ok. I was hoping to clarigy what shouldn't be included when I do that scan through before committing - it seemed like these files were a proper part of OMS - how can you tell if a file should be committed or not?

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
lache-melvin: I think blank is fine, it is more consistent with the rest of the table, blank unless there is relevant data. It's quite busy, so nice to keep it clean where we can

lache-melvin (changes requested):
Thanks for the mahi Zachariah, changes looking great - mainly just the migrations to fix up 🙏

server/repository/src/migrations/v2_07_00/mod.rs
```diff
@@ -23,6 +23,7 @@ impl Migration for V2_07_00 {
 
     fn migrate_fragments(&self) -> Vec<Box<dyn MigrationFragment>> {
         vec![
+            Box::new(change_vaccination_date_to_not_nullable::Migrate),
```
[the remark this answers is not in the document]
zachariah-at-msupply: Got it, thank you Laché 👍

lache-melvin (approved):
Sweet, thanks for changes 🚀

Issue #6407, opened 2025-02-03
Unable to type text effortlessly and smoothly in prescription notes


marthakeezy:
## What went wrong? 😲

Noticing some sort of text regression when trying to type in prescription notes

https://github.com/user-attachments/assets/270fc926-5759-499f-9472-4b332b9da800

## Expected behaviour 🤔

Should be able to type text smoothly

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to Dispensary -> Prescriptions
2. Create a new prescription or edit an existing one
3. Try to type some text - either fast or slow, long or short sentences
4. Backspace
5. Notice weird behaviour

## Your environment 🌱

- Open mSupply Version: `V2.6.0-build-test-3`
- Legacy mSupply Central Server Version: `V7.19..03`

- Platform: `Google Chrome (browser for Mac)`

- Database type: `PostgreSQL`

roxy-dao:
Do #6188 as well

DhanyaHerath:
Thanks Roxy and Maria! Am going to close this and suggest the work is done as part of #6188

Issue #5035, opened 2024-10-08
Losing user input when typing fast after backspace


Chris-Petty:
## What went wrong? 😲

If you press backspace and type relatively quickly (as a busy pharmacist might!), the state doesn't keep up and we end up deleting keystrokes after the backspaces. `356 +bsp +bsp + 2 + 5` may output `35` rather than `325`. This is bad as it risks wrong quantities prescribed OR inaccurate stock levels if the user doesn't notice the system messing up their valid input.

https://github.com/user-attachments/assets/e51fe2a4-7e62-4403-a51e-79613d446f42

Also the general jank form some state control flipping the values about after typing...

Note this might be even worse/easier to reproduce on a keyboard just as it's easier to be fast.

## Expected behaviour 🤔

No jank. Doesn't delete intended input.

## How to Reproduce 🔨

1. Prescription edit a line
2. have 3-4 digits
3. quickly press backspace 1 or more times and type in new digits - you may loose some of your input.

(Might be a requirement for the prescription to be backdated, or perhaps to have many transactions affecting the stock line you are attempting to issue)

## Your environment 🌱

- Open mSupply Version: 2.2.03 an 2.3.0-RC5
- Legacy mSupply Central Server Version: `main` 7.17.15 
- Platform: Samsung SM-X200
- Database type: sqlite

lache-melvin:
Agree, when busy and trying to work quickly, this could easily result in errors - as well as general frustration! Setting to normal as can be worked around, if we engage the prescription line peacefully 😁 

Should be investigated how widespread this is though - is it the underlying input or a pattern in the prescription modal?

andreievg:
Triage, just adding dispensing label to flag as something to look at in that bit of work. Can be quite frustrating to bruce lee

DhanyaHerath:
Closing this to be looked at as part of https://github.com/msupply-foundation/open-msupply/issues/6188

PR #7255, opened 2025-04-08
6188-6407-5035 Add debounced value callback for prescription comment field

### LINKED ISSUE #6188: On the internal order Info panel what I am writing disappear after I wrote it
## What went wrong? 😲
As recording

Nb: When working on this issue please fix https://github.com/msupply-foundation/open-msupply/issues/6407 as well 

## Expected behaviour 🤔
As the translation of the french saying says "Spoken words fly away, written words stay". I believe my comment should stay :)
## How to Reproduce 🔨

1. Go to 'Replenishment > Internal order'
2. Open an internal order
3. click on the More button
4. Write a comment in the comment box
5. Wait
6. See error. What you have written disappear

## Your environment 🌱

- Open mSupply Version: 2.5.0RC4
- Legacy mSupply Central Server Version: v8.00RC43

- Platform: omSupply, desktop, android

- Database type: SQLlite3

noel-yeldos:
Multi-issue PR (Main issue is #6188 with #6407 and #5035 as linked issues)

This PR Fixes #6188 (Actually the issue that is fixed is #6407 - unable to replicate #6188 and #5035 - only written this way to satisfy sync labels action)

# 👩🏻‍💻 What does this PR do?
Add debounced value callback for prescription comment field input so that updatePrescription does not fire for each keystroke.

https://github.com/user-attachments/assets/7ad61090-ddc4-40c3-be9c-7901d417a7f5

# 🧪 Testing

- [ ] Go to Dispensary -> Prescriptions
- [ ] Create a new prescription or edit an existing one
- [ ] Click on more and type some text - either fast or slow, long or short sentences into the comment field
- [ ] Observe that entered text does not disappear randomly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
@@ -28,7 +28,11 @@ export const AdditionalInfoSectionComponent: FC = () => {
   const [commentBuffer, setCommentBuffer] = useBufferState(comment ?? '');
   const { localisedDate } = useFormatDateTime();
 
-  const debouncedUpdate = useDebouncedValueCallback(update, [commentBuffer]);
+  const debouncedUpdate = useDebouncedValueCallback(
+    update,
+    [commentBuffer],
+    1500
+  );
 
   if (!createdDatetime) return null;
 
--- a/client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
+++ b/client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
@@ -12,6 +12,7 @@ import {
   InfoTooltipIcon,
   useFormatDateTime,
   UNDEFINED_STRING_VALUE,
+  useDebouncedValueCallback,
 } from '@openmsupply-client/common';
 import { usePrescription } from '../../api';
 
@@ -26,6 +27,8 @@ export const AdditionalInfoSectionComponent: FC = () => {
   const [colorBuffer, setColorBuffer] = useBufferState(colour);
   const [commentBuffer, setCommentBuffer] = useBufferState(comment ?? '');
   const { localisedDate } = useFormatDateTime();
+
+  const debouncedUpdate = useDebouncedValueCallback(update, [commentBuffer]);
 
   if (!createdDatetime) return null;
 
@@ -61,7 +64,7 @@ export const AdditionalInfoSectionComponent: FC = () => {
           disabled={isDisabled}
           onChange={e => {
             setCommentBuffer(e.target.value);
-            update({ comment: e.target.value });
+            debouncedUpdate({ comment: e.target.value });
           }}
           value={commentBuffer}
         />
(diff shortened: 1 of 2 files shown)


client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
```diff
@@ -27,6 +28,8 @@ export const AdditionalInfoSectionComponent: FC = () => {
   const [commentBuffer, setCommentBuffer] = useBufferState(comment ?? '');
   const { localisedDate } = useFormatDateTime();
 
+  const debouncedUpdate = useDebouncedValueCallback(update, [commentBuffer]);
```
roxy-dao, line 31:
```suggestion
  const debouncedUpdate = useDebouncedValueCallback(update, [commentBuffer], 1000);
```

Your change is still causing numerous API calls.. 

https://github.com/user-attachments/assets/45fe62c8-220c-4f31-84b8-0172223b632b

If you look at examples where the comment is not being jumpy, you will see that there is a 1k timeout on the API call.. looks like the 500 timeout on the `useDebouncedValueCallback` is not enough and still causes constant API calls while throttlin' network to 3G....

roxy-dao (changes requested):
Thanks Noel~

See comment below

The issues in the other ones might be caused by networks slower than 3G, so might need to extend the timeout =/...

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/AdditionalInfoSection.tsx
noel-yeldos: @roxy-dao thanks for catching that! For some reason when I tested previously it wasn't behaving strangely. But I'm still finding that using ```useBufferState``` with the ```debouncedValueCallback``` even with 1000 timeout doesn't fully resolve the issue. I'm finding that using ```useState``` instead seems to be working fine. Is it ok to swap?

roxy-dao: If extending the timeout to 1k causes problems then your original fix definitely caused problems...? useBuffer is saving the value into state.... maybe needs some thinking?

noel-yeldos: Oh my original fix did cause problems. It's just for whatever reason when I tested before putting the PR in it was working and so I didn't catch it then. When I test now it is causing problems. UseState isn't but I'll have a look at how it's being used in other places.

client/packages/invoices/src/Prescriptions/DetailView/SidePanel/PrescriptionDetailsSection.tsx
```diff
@@ -24,9 +24,11 @@ export const PrescriptionDetailsSectionComponent: FC = () => {
   const [theirReferenceInput, setTheirReferenceInput] =
     useState(theirReference);
 
-  const debouncedUpdate = useDebouncedValueCallback(update, [
-    theirReferenceInput,
-  ]);
+  const debouncedUpdate = useDebouncedValueCallback(
+    update,
+    [theirReferenceInput],
+    1500
```
noel-yeldos, line 30: Added here as well because the same issue was happening

roxy-dao (approved):
Thanks Noel~

PR #7261, opened 2025-04-08
Make print errors translatable and add translations

### LINKED ISSUE #7235: Translations: when printing a custom report and unable to print, error appears in English instead of selected language
## What went wrong? 😲

Spotted by @Rajajajaja 🏅 

When printing one of custom reports (Facture Client) and if unable to print, the error shows in English instead of selected language (French)

Screenshot uploaded ->>> [here](https://drive.google.com/drive/folders/1acyJ3MMyxoBxFGUoxKkE66kJxSGXIQxa)

## Expected behaviour 🤔

Error should be translated, priority is French

## How to Reproduce 🔨

1. Go to https://raja.msupply.org:8000
2. Find login credentials on the screenshot uploaded
3. Go to Distribution -> Outbound Shipments
4. Create a new one or click on an existing one
5. Go to Print -> select a report and encounter the error
6. Make sure you are testing in another language, not English

## Your environment 🌱

- Open mSupply Version: V2.6.1
- Legacy mSupply Central Server Version: V8.0.11

- Database type: PostgreSQL

zachariah-at-msupply:
Fixes #7235 

# 👩🏻‍💻 What does this PR do?

This PR makes the printing error messages translatable and adds a french translation for them.

# 🧪 Testing

- [ ]     Go to Distribution -> Outbound Shipments
- [ ]     Create a new one or click on an existing one
- [ ]     Change the language to French
- [ ]     Click the print button (labelled 'Imprimer') in the top right corner, select a report to print, and check the error message says 'Impossible d'imprimer le document'.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1637,6 +1637,7 @@
   "messages.no-stock-available": "There is no stock available",
   "messages.no-transaction-other-facility": "Stock transactions are not recorded for vaccinations given in other facilities",
   "messages.no-vaccine-items-configured": "No vaccine items configured for this vaccine course. Stock transactions will not be recorded.",
+  "messages.error-printing-report": "Unable to print report",
   "messages.not-applicable": "N/A",
   "messages.not-configured": "Not configured",
   "messages.not-initialised": "[ Not configured ]",
--- a/client/packages/common/src/intl/locales/fr/common.json
+++ b/client/packages/common/src/intl/locales/fr/common.json
@@ -1531,6 +1531,7 @@
   "messages.error-deleting-reasons_other": "{{count}} raisons n'ont pas pu être supprimées",
   "messages.error-generic_one": "{{count}} erreur",
   "messages.error-generic_other": "{{count}} erreurs",
+  "messages.error-printing-report": "Impossible d'imprimer le document",
   "messages.error-no-file-selected": "Aucun fichier sélectionné",
   "messages.error-not-valid-email": "Veuillez saisir une adresse électronique valide",
   "messages.error-saving-insurances": "Erreur lors de l'enregistrement des assurances 🥺",
--- a/client/packages/system/src/Report/api/hooks/usePrintReport.ts
+++ b/client/packages/system/src/Report/api/hooks/usePrintReport.ts
@@ -7,6 +7,7 @@ import {
   useIntlUtils,
   useMutation,
   useNotification,
+  useTranslation,
 } from '@openmsupply-client/common';
 import { Environment } from '@openmsupply-client/config';
 import { Printer } from '@bcyesil/capacitor-plugin-printer';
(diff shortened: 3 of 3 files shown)


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1637,6 +1637,7 @@
   "messages.no-stock-available": "There is no stock available",
   "messages.no-transaction-other-facility": "Stock transactions are not recorded for vaccinations given in other facilities",
   "messages.no-vaccine-items-configured": "No vaccine items configured for this vaccine course. Stock transactions will not be recorded.",
+  "messages.error-printing-report": "Unable to print report",
```
fergie-nz, line 1640: FYI there is an existing translation key of 
"error.failed-to-generate-report" 
in our common.json

Maybe could use this one instead? It more or less means the same thing to me, and already has existing translations in other languages. Over to you :)

client/packages/common/src/intl/locales/fr/common.json
```diff
@@ -1531,6 +1531,7 @@
   "messages.error-deleting-reasons_other": "{{count}} raisons n'ont pas pu être supprimées",
   "messages.error-generic_one": "{{count}} erreur",
   "messages.error-generic_other": "{{count}} erreurs",
+  "messages.error-printing-report": "Impossible d'imprimer le document",
```
fergie-nz, line 1534: Generally we leave any translations for our team who is fluent in the local language.

I understand that you'd probably want to add this to show that the translation is working as expected when the reviewer does the review. In this case, you can make a 'test' branch, and link it in the PR with a dummy translation so that the reviewer can check this, make sure it is working, then approve the PR without any possibly wrong translation keys.

The issue with a non native speaker adding a translation key is that it won't show up as missing from the translation team's perspective, so they won't have the opportunity to review it. There might be subtle or nuanced parts of even simple phrases that will go missed until in front of a client (especially something like this which should only show when there is an error, so our PM team might not see it during demos etc)

fergie-nz (approved):
Looks great, Zachariah! Nice one.

I would recommend removing the french translation in this case so that the translation team has the opportunity to double check the wording in context, but otherwise looks great

client/packages/common/src/intl/locales/en/common.json
fergie-nz: Also I'm just looking and seeing that we are not propagating any error types from the back end to front end, and handling them selectively (you can see that in the api for generateReport, we either get the report node, OR throw a generic error).  So we aren't getting any information about the error as a user.

I'm wondering if it would be better to call this 'error' when we give a toast rather than 'unable' because unable makes me wonder if it is a permission denied thing rather than an issue with the back end.

client/packages/common/src/intl/locales/fr/common.json
zachariah-at-msupply: (Just to clarify, this translation was given by the translation team in the original issue.)

PR #7276, opened 2025-04-09
2.7 RC2 back into develop

CarlosNZ:
Merging RC branch back to develop after latest RC release


client/packages/common/src/intl/locales/fr/common.json
```diff
@@ -1152,7 +1157,7 @@
   "label.to-date": "À ce jour",
   "label.to-datetime": "jusqu'à la date/heure",
   "label.to-expiry": "Jusqu'à péremption",
-  "label.to-shipped-date": "Jusqu'à la date d'expédition",
+  "label.to-shipped-datetime": "Jusqu'à la date d'expedition",
```
roxy-dao, line 1160: Can you confirm which translation string is correct? Is it with or without é accent on top?

roxy-dao (approved):
tyty~ just gotta check translation string

PR #7305, opened 2025-04-13
6446 nav sync badge after period of error

### LINKED ISSUE #6446: Sync badge usability
Currently there is an error icon shown if sync experiences an error. Otherwise a number is shown, which is the number of records waiting to be synchronised out from the current site.

The display of a number in a badge over the sync navigation has been called into question:
* it reads as if there is an action to be performed; but the user doesn't need to do anything
* the number is a proxy for 'action required' but only when it goes over a certain threshold, and there's no way for the user to interpret what that is
* the value of the number is of little consequence for the user
* the number is shown when you click the sync window - if you really must know what it is!

A more useful indicator for users is to show when action is required. 
To that end:
* create a sync warning threshold (1 day) which is stored as a hard coded value ( ideally this would be a global preference - and could be, depending when this issue is tackled
* and a sync error threshold (3 days) which also could be a global pref
* If sync has an error - show the error icon, as it does now
* if a sync has not been successful within the error threshold show the alert icon as a badge in the theme error colour
* if a sync has not been successful within the warning threshold show the alert icon as a badge  in the theme warning colour

Image

Chris-Petty:
Fixes #6446 

# 👩🏻‍💻 What does this PR do?

No longer show the badge if there are simply pending records

If connection has failed for over a day show a warning icon with warning colour (orange)

image

If connection has failed for over 3 days show a warning icon with error colour (red)

image

If there is any error OTHER than an a connection error (i.e. unreachable due to bad internet) then immediately show a warning icon with error colour 

I've left some scaffolding for the opportunity to implement prefs for the warning/error thresholds. Can be another issue when we have time for it? If we care enough for it? heh.

Frankly, the red and orange colours don't look much different on such a small icon. I think this feature could be simplified as one threshold.

I think due to buttons not having borders the badges being on top right of the button looks a bit disjointed. I'm not sure how best to address this other than having borders, different colour to background or moving the badge to be over the top right of the last character (though this would look odd when you press/hover the button and it shows the badge is in the middle of the button). Nothing to do with this PR though, so something for another time.

# 🧪 Testing

This PR is a bit annoying to test because it turns out we cache the last sync_log. So modifying the DB alone isn't enough to get the `lastSuccessfulSyncDate` to change in the frontend.

I ended up just modifying `or_latest_row` to `return self` at the start and modifying the DB to test this worked fine:
https://github.com/msupply-foundation/open-msupply/blob/5e34e099c87935fa2acd622df085dca1cfdab41a/server/repository/src/db_diesel/sync_log_row.rs#L184-L202

You could also just remove this map but I only just thought of that now:

https://github.com/msupply-foundation/open-msupply/blob/5e34e099c87935fa2acd622df085dca1cfdab41a/server/repository/src/db_diesel/sync_log_row.rs#L171.

Then you can modify `finished_datetime` with -2 or -4 days to reach each of the thresholds:

image

- [ ] OMS sync (I was just testing this with OMS central server sync. would be wise to do a proper remote site too)
- [ ] Modify some records in OMS
- [ ] No badge with any number. YAY.
- [ ] Now either have hacked OMS backend to not cache and modify records to test each threshold or...
- [ ] Turn off OG server
- [ ] Wait 24hrs for first warning in orange colour 😆 
- [ ] Wait 72hrs for second warning (error) in red colour 😭 
- [ ] turn on OG server
- [ ] Sync (clearing the warning/error badge from above)
- [ ] Change the site password on OG central
- [ ] Sync OMS - error immediately in sync modal and error sync badge displayed on nav button immediately.

# 📃 Documentation

  1. The nav bar showing badge with number count of pending records anywhere.
  2. Perhaps some explanation about when the badge will show! 

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -2633,6 +2633,7 @@ export type FrontendPluginMetadataNode = {
 export type FullSyncStatusNode = {
   __typename: 'FullSyncStatusNode';
   error?: Maybe<SyncErrorNode>;
+  errorThreshold: Scalars['Int']['output'];
   integration?: Maybe<SyncStatusWithProgressNode>;
   isSyncing: Scalars['Boolean']['output'];
   lastSuccessfulSync?: Maybe<SyncStatusNode>;
@@ -2643,6 +2644,7 @@ export type FullSyncStatusNode = {
   push?: Maybe<SyncStatusWithProgressNode>;
   pushV6?: Maybe<SyncStatusWithProgressNode>;
   summary: SyncStatusNode;
+  warningThreshold: Scalars['Int']['output'];
 };
 
 export enum GenderType {
@@ -9755,7 +9757,7 @@ export type VaccinationNode = {
   invoiceId?: Maybe<Scalars['String']['output']>;
   notGivenReason?: Maybe<Scalars['String']['output']>;
   stockLine?: Maybe<StockLineNode>;
-  vaccinationDate?: Maybe<Scalars['NaiveDate']['output']>;
+  vaccinationDate: Scalars['NaiveDate']['output'];
 };
 
 export type VaccineCourseConnector = {
--- a/client/packages/host/src/components/AppDrawer/SyncNavLink.tsx
+++ b/client/packages/host/src/components/AppDrawer/SyncNavLink.tsx
@@ -3,10 +3,10 @@ import {
   AlertIcon,
   AppNavLink,
   RadioIcon,
+  SyncErrorVariant,
   useTheme,
   useTranslation,
 } from '@openmsupply-client/common';
-import { getBadgeProps } from '../../utils';
 import { useSync } from '@openmsupply-client/system';
 import { useSyncModal } from '../Sync';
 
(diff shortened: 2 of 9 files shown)


client/packages/common/src/types/schema.ts
```diff
@@ -9755,7 +9757,7 @@ export type VaccinationNode = {
   invoiceId?: Maybe<Scalars['String']['output']>;
   notGivenReason?: Maybe<Scalars['String']['output']>;
   stockLine?: Maybe<StockLineNode>;
-  vaccinationDate?: Maybe<Scalars['NaiveDate']['output']>;
+  vaccinationDate: Scalars['NaiveDate']['output'];
```
Chris-Petty, line 9760: random vax card change?

client/packages/host/src/components/AppDrawer/SyncNavLink.tsx
```diff
@@ -46,7 +60,7 @@ export const SyncNavLink = () => {
       }}
       icon={<RadioIcon fontSize="small" color="primary" />}
       text={t('sync')}
-      badgeProps={badgeProps}
+      badgeProps={badgeProps || undefined}
```
Chris-Petty, line 63: I find typescript... obtuse. Can we just have `falsy` as a type? Lol

I think all the types for `syncStatus` (like everywhere else) a cumbersome too... I think because the type is anything that the API MIGHT return, vs specifically what we're requesting the in gql in this component. If there is a way to just have a type that matches the gql request we're making and cleans up everything MAYBE being undefined | null that'd allow us to have much nicer code while still largely being typesafe?

aimee-mcneil-melville:
I had a go at this too (on central, haven't tested remote) - functionality works great!

I agree about the colours and thought that we could use a filled in icon so either an orange or red background instead. Also as it is not a badge with a count anymore, this could be configured as an 'end icon' to get the vertical alignment. Though the badge makes more sense when the drawer is collapsed 🤷‍♀️ 

If we do want multiple warning thresholds, and bearing in mind how small the icon is (could be easy to get used to it being there and just not notice it!) another idea for a future issue could be a pulsing effect to the icon or button to alert the user to the more urgent type

Chris-Petty:
> I thought I would have a crack at this so I've just done a test of the functionality (thanks @Chris-Petty for helping me set up the tests).
> 
> I agree that it is very hard to distinguish between the warning and error colours. Also agree that the floating badge does look disconnected from the sync nav link when the app drawer is open.
> 
> I'm a little confused as to what action I should be taking as the user when I see either the warning or the error. Is that only because we are manipulating the db for the badges to show? Will an actual user get an error message that indicates what has gone wrong and what they should do?

The cases are basically:

1. Haven't had an internet for a day... their orders aren't syncing and things won't get processed.
2. Haven't synced for 3 days, they should really get out and sort out an internet connection (some places literally have to walk up the nearest hill once a week
3. It's an error to do with something other than connection. e.g. password is wrong, site name not found, some server side error/rejection that needs intervention (likely they'll need to reach out to support either locally or TMF). Hence just showing it immediately.

client/packages/host/src/components/AppDrawer/SyncNavLink.tsx
lache-melvin: Yeh... here's a suggestion - moving the getBadge to separate function, handling the non-existence of the syncInfo data, and simplifying the showWarning/showError logic a bit, date utils helps:

https://github.com/msupply-foundation/open-msupply/compare/6446-sync-badge-only-on-warning-or-error...6446-sync-badge-only-on-warning-or-error(refactor-getbadge)

lache-melvin (changes requested):
Requesting that we simplify the error/warning date logic a bit, otherwise looks great. Thanks for seeding for pref 🙏 

Agree with the others around the badge not looking quite right when app drawer open, but I think we are still optimising for tablet where it would be collapsed, and the badge makes sense there. Can be a future improvement?

lache-melvin (approved):
ship it 🚀

mark-prins:
was going to pop in and suggest reversing the fill to give a more obvious red/orange circle and see this has been suggested already. so there's no need.

when adding the badges I spent a while trying to make them show in the right place, as the disjointed floating out to the right annoyed me too. it was a bit tricky, so we have what we have.

Issue #5237, opened 2024-10-25
Review and replace `useEffect` usage


Chris-Petty:
## The suggestion

The useEffect hook is a common source of bugs and performance issues when misused in react. I've seen a few places where it's been used for things other than synchronising state with some external system such as an API request. As all of our API work is managed through abstractions on top of `useQuery` hook, it appears suspicious that we have >250 calls of `useEffect`.

React has some good docs specifically on when not to use `useEffect` https://react.dev/learn/you-might-not-need-an-effect

At a glance a good amount do seem to be legit, but there are definitely some that should probably use a different approach!

## Example use case

https://github.com/msupply-foundation/open-msupply/blob/f895b95dc9d8c49fd5b8af01ed8726d5ad62506a/client/packages/requisitions/src/RequestRequisition/ListView/ProgramRequisitionOptions.tsx#L41-L55

To avoid unnecessary re-renders, prefer to use event handler functions rather than `useEffect`:

https://github.com/msupply-foundation/open-msupply/blob/01fbcea02ff9d486327262d146ec09590715337c/client/packages/requisitions/src/RequestRequisition/ListView/ProgramRequisitionOptions.tsx#L41-L56

Alternatively `useReducer` might be suitable in certain situations. The main point is to only use the `useEffect` hook if synchronising with an external system.

### Why should we invest time in this?

- unnecessary useEffects create unnecessary re-renders which slows the app down and introduces odd bugs.
- useEffect is asynchronous and typically creates code that is hard to follow, so things should become a bit simpler to follow logically.

### Are there any risks associated with this change?

Other than standard regressions, not much!

### How much effort is required?

There are 250+ `useEffect` calls in OMS. I don't know how many are proper usage or not, this issue is to work that out!

### Agreed Solution

During cool down period/refactor fortnight, systematically review useEffects. Make an issue if problematic? One challenge is distributing the load if multiple people look, and recording what was looked at... tempting to just a comment on each `useEffect` like `// Usage is legit!` haha.

andreievg:
In next omSup all team session, let's look at some less then ideal use cases of this and start the conversation or at least make sure that everyone is aware of anti patterns (proper use of useEffect)

andreievg:
Also add example use cases here, it will be and epic for reference

PR #7308, opened 2025-04-14
7186 Add confirmation on leaving modal when navigating with unsaved changes - View stock

### LINKED ISSUE #7186: Navigating away from view stock after making change doesn't show alert
## What went wrong? 😲

After making a change to a stock item and navigating away, app doesn't warn you that you have unsaved changes.

https://github.com/user-attachments/assets/0218796f-48e9-4f8a-995a-75065b8d7815

## Expected behaviour 🤔
As in the patient view, the app should alert you that you have unsaved changes when you try to navigate away after making a change

## How to Reproduce 🔨

1. Go to Inventory -> view stock
2. Click on an item
3. Make a change
4. Navigate away from page without saving
5. See that alert modal for unsaved changes doesn't appear

## Your environment 🌱

- Open mSupply Version: 2.7
- Legacy mSupply Central Server Version: 8.01.03

- Platform: Chrome Mac OS

- Database type:

noel-yeldos:
Fixes #7186 

# 👩🏻‍💻 What does this PR do?
Display confirmation modal if user navigates away from or refreshes page with unsaved changes

Added a `useEffect` that syncs stockLine `isDirty` state with `useConfirmationOnLeaving` hook's internal `isDirty` state

https://github.com/user-attachments/assets/6db79f22-01b5-4aba-8a21-301a383c5b73

I did some reading and tried to see if I could do this without using the `useEffect`. Unfortunately, I wasn't able to figure that out. Please let me know if I've missed something! 

# 🧪 Testing

- [ ] Go to Inventory -> view stock
- [ ] Click on an item
- [ ] Make a change
- [ ] Navigate away from page or refresh without saving
- [ ] See that alert modal for unsaved changes appears (modal for refresh looks different and has been raised)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Stock/DetailView/DetailView.tsx
+++ b/client/packages/system/src/Stock/DetailView/DetailView.tsx
@@ -14,6 +14,7 @@ import {
   useCallbackWithPermission,
   UserPermission,
   usePluginEvents,
+  useConfirmOnLeaving,
 } from '@openmsupply-client/common';
 import { ActivityLogList } from '@openmsupply-client/system';
 import { AppBarButtons } from './AppBarButtons';
@@ -76,6 +77,14 @@ export const StockLineDetailView: React.FC = () => {
     message: t('messages.confirm-cancel-generic'),
     title: t('heading.are-you-sure'),
   });
+
+  const { setIsDirty } = useConfirmOnLeaving('view-stock');
+
+  useEffect(() => {
+    // Getting isDirty from 'draftStockLine' rather than from 'useConfirmOnLeaving' hook
+    // so need to update it manually to sync external isDirty state with hook's isDirty state
+    setIsDirty(isDirty);
+  }, [isDirty]);
 
   const openInventoryAdjustmentModal = useCallbackWithPermission(
     UserPermission.InventoryAdjustmentMutate,


client/packages/system/src/Stock/DetailView/DetailView.tsx
```diff
@@ -77,6 +78,14 @@ export const StockLineDetailView: React.FC = () => {
     title: t('heading.are-you-sure'),
   });
 
+  const { setIsDirty } = useConfirmOnLeaving('view-stock');
+
+  useEffect(() => {
+    // Getting isDirty from 'draftStockLine' rather than from 'useConfirmOnLeaving' hook
+    // so need to update it manually to sync external isDirty state with hook's isDirty state
```
noel-yeldos, line 85: This is my understanding of why the `useEffect` was needed

lache-melvin: Hm - ideally extra useEffect shouldn't be required to sync the isDirty... but to fix is a bit more work.

I think this is ok for now, I can see we have done it in one or two other places as well. Though would you be up for writing up a refactor issue? We should consider how to solve isDirty well, beyond just this case.

To remove the need for the useEffect, either we use the isDirty provided by useConfirmOnLeaving hook (which is looking like a non-option, given dirty check from usePatchState) or we make the isDirty state returned by `usePatchState` easy to pass into useConfirmOnLeaving.

So isDirty would need to be a ref, so useConfirmOnLeaving gets the latest value, maybe [something like this](https://github.com/msupply-foundation/open-msupply/commit/d00727762fbe358b734da9a85206686c779678df). 

However, I don't want to rush this change in now, I think we need to zoom out, what is our long term plan for the relationship between those draft state management hooks, and useConfirmOnLeaving. Maybe they can sit closer together?

🙏

lache-melvin (approved):
LGTM, if you can write up refactor issue 🙏

client/packages/system/src/Stock/DetailView/DetailView.tsx
roxy-dao: https://github.com/msupply-foundation/open-msupply/issues/5237 useEffect refactor issue

PR #7323, opened 2025-04-14
7285 prevent ungiving vaccinations from another store

### LINKED ISSUE #7285: I can un-give a vaccination from another store
## What went wrong? 😲

Vaccination records can be fetched/synced between stores. If one store marks a vaccination as given, currently the other store can then mark the vaccination as not-given. This shouldn't be possible.

## Expected behaviour 🤔

Stores other than the store that gave the vaccination, shouldn't be able to un-give it. 

It should be noted that if a vaccination is marked as not given (e.g. not enough stock) - other stores should be able to give it!

Given above use case, the store that creates the initial vaccination record may not be the store that gives the vaccination (if it is initially created in not given status.) Therefore, probably should record given_store_id on the vaccination records?

(Maybe also shouldn't be able to edit facility name after this....)

## How to Reproduce 🔨

1. Have a vaccination program enrolled patient syncing between 2 stores
2. Give a vaccination in one store
3. Sync
4. In the other store - mark vaccination as not given
5. Bug : no error

## Your environment 🌱

- Open mSupply Version: v2.7.2 rc2

- Platform:

- Database type:

lache-melvin:
Fixes #7285

# 👩🏻‍💻 What does this PR do?

Disables editing of vaccination data (expect comment) from other stores, once vaccination has been given.

Screenshot 2025-04-15 at 8 39 07 AM

Also adds item_link_id to vaccination, so we can see which vaccine item was given to a patient even when it was given in another store. Before, we only determined item via related stock line/invoice, which only exists for the "giving" store.

Apologies, little bit of frontend refactoring, found it hard to reason about where to add conditions, will add comments

There is breaking API change in that we now expect passed in stock line and item id to match, but i don't think vaccination graphql is used externally...

# 🧪 Testing

- [ ] Have a vaccine-program-enrolled patient visible in 2 stores
- [ ] From store A
  - [ ] Mark one dose as given - select a vaccine item but don't select a related batch
  - [ ] Save and reopen the vaccination - item is still selected
  - [ ] For the second dose - select past date, mark as given - you can select an item even without "record historic transaction"/selecting a batch
  - [ ] For the third dose - select Other Facility, mark as given - you can select an item even though you cannot select a batch
  - [ ] Mark a fourth dose as not given
  - [ ] You can reopen and edit all of these doses (change to not given and back, create transactions or don't)
- [ ] Sync
- [ ] View vax card from other store
- [ ] The `given` vaccinations are not editable, except for comment field
- [ ] You can mark the "not given" vaccination as given 
  - [ ] When you change the given status, facility field is updated to your current store

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -256,33 +229,35 @@ const VaccinationForm = ({
         sx={{ margin: '0 auto' }}
         value={draft.given ?? null}
         onChange={event =>
-          updateDraft({ given: event.target.value === 'true' })
+          updateDraft({
+            given: event.target.value === 'true',
+            // Ensure current facility is selected when changing given status
+            facilityId: isFreeTextFacility ? OTHER_FACILITY : store?.nameId,
```
lache-melvin, line 235: Was a bit too easy to change to given, while leaving the old facility selected. I think this feels like the right behaviour?

client/packages/system/src/Vaccination/Components/getSwitchReason.test.ts
```diff
@@ -5,52 +5,51 @@ import { OTHER_FACILITY } from './FacilitySearchInput';
 
 describe('getSwitchReason', () => {
   describe('no existing vaccination (create)', () => {
-    it('should return "record transaction" if the date is historical & doses configured', () => {
+    it('should return "record transaction" if the date is historical & item selected', () => {
```
lache-melvin, line 8: test changes are just noise due to usage change - instead of there being no dose items configured, should be if none is selected.

client/packages/system/src/Vaccination/api/operations.generated.ts
```diff
@@ -258,17 +250,17 @@ export const VaccinationDetailFragmentDoc = gql`
...
       id
-      itemId
-      itemName
       batch
     }
     invoice {
       id
-      store {
-        id
-      }
```
lache-melvin, line 271: Given store id more reliable than invoice, which would be null everywhere except giving store

client/packages/system/src/Vaccination/api/operations.generated.ts
```diff
@@ -258,17 +250,17 @@ export const VaccinationDetailFragmentDoc = gql`
       lastName
     }
     given
+    givenStoreId
+    item {
+      id
+      name
+    }
```
lache-melvin, line 257: item info should be available on vaccination, even if stock line is not selected/is for different store

server/repository/src/db_diesel/vaccine_course/vaccine_course_dose_row.rs
```diff
@@ -26,6 +30,12 @@ table! {
 
 joinable!(vaccine_course_dose -> vaccine_course (vaccine_course_id));
 allow_tables_to_appear_in_same_query!(vaccine_course_dose, vaccine_course);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, clinician_link);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, clinician);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, name_link);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, name);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, item_link);
+allow_tables_to_appear_in_same_query!(vaccine_course_dose, item);
```
lache-melvin, line 38: These were in the vaccination row file, figured they should probably live with the relevant table...

server/service/src/vaccination/insert/generate.rs
```diff
@@ -73,7 +74,9 @@ pub fn generate(
         clinician_link_id: clinician_id,
         vaccination_date: vaccination_date.unwrap_or(now.date()),
         given,
+        given_store_id: given.then_some(store_id),
         stock_line_id,
+        item_link_id: item_id,
```
lache-melvin, line 79: SO I think if FINALLY have link ids in my head, when it is ok to assign item_link_id = item_id, vs when it is not.

I'll do a wiki write up asap.

But this is fine - when items & item links created, item_link_id is item id, so:



| Link ID  | Item ID |
| -------- | ------- |
| item_a  | item_a   |
| item_b | item_b     |

Then Item A and B get merged, to just be Item A:

| Link ID  | Item ID |
| -------- | ------- |
| item_a  | item_a  |
| item_b | item_a   |

So, conflating item_id to be item_link_id in this direction is fine - both ids continue to exist as link ids and will point back to correct item. Where we need to do lookup is where two records could have different link ids, can't just say where link ID = X, as multiple link ids could actually be pointing to same item

server/service/src/vaccination/insert/validate.rs
```diff
@@ -72,15 +73,32 @@ pub fn validate(
             return Err(InsertVaccinationError::FacilityDoesNotExist);
         }
     }
+    // If not given, reason is required
+    if !input.given {
```
lache-melvin, line 77: For some reason, this check was in the "get_stock_line" function... lol?

server/service/src/vaccination/update/generate.rs
```diff
@@ -73,8 +79,11 @@ fn generate_given(
 
     // apply given status, stock and invoice ids
     let vaccination = VaccinationRow {
+        not_given_reason: None,
         given: true,
+        given_store_id: Some(store_id.to_string()),
```
lache-melvin, line 84: The bit that actually solves the PR, assigning given store ID where we mark as given - and in validate method, checking this matches the calling store id

Chris-Petty:
I've run out of time! Heaven forbid it's still here but I will be back next Tuesday if it still needs review 🥲

server/service/src/vaccination/insert/generate.rs
andreievg: Thanks : https://github.com/msupply-foundation/open-msupply/issues/7429

server/service/src/vaccination/update/validate.rs
```diff
@@ -70,18 +71,40 @@ pub fn validate(
...
     }
+
     // If not given, reason is required
     if input.given == Some(false) {
         if input.not_given_reason.is_none() {
             return Err(UpdateVaccinationError::ReasonNotProvided);
         };
     };
 
+    // If selected item is changing - validate it
```
andreievg, line 82: I was wondering if there is a description somewhere about what happens in the different use case for immunizations/vaccinations ? (like if i ungive vaccination, if vaccination is given, with stock line issued, and then item is changed etc..)

server/service/src/vaccination/update/mod.rs
```diff
@@ -429,6 +499,43 @@ mod update {
...
+                UpdateVaccination {
+                    id: mock_vaccination_a().id,
+                    given: Some(false),
+                    not_given_reason: Some("out of stock".to_string()),
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+
+        assert_eq!(result.vaccination_row.given, false);
```
andreievg, line 520: worth checking that store id was set correct, and unset in the one below

client/packages/system/src/Vaccination/Components/SelectBatch.tsx
```diff
@@ -76,7 +76,13 @@ export const SelectBatch = ({
         <BatchTable
           columns={columns}
           data={data?.nodes ?? []}
-          setStockLine={setStockLine}
+          // Allow un-selecting of stock line, if don't want to record
+          // transaction
+          setStockLine={newStockLine =>
+            setStockLine(
```
andreievg, line 82: "toggleStockLine"

client/packages/system/src/Vaccination/Components/SelectItemAndBatch.tsx
```diff
@@ -1,105 +1,74 @@
...
-  }
-
-  if (isOtherFacility) {
-    return <InfoText>{t('messages.no-transaction-other-facility')}</InfoText>;
-  }
-
-  if (!vaccineItemOptions.length) {
-    return <InfoText>{t('messages.no-vaccine-items-configured')}</InfoText>;
-  }
+  const transactionSwitchReason = getSwitchReason(draft, vaccination);
```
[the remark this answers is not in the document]
andreievg: const somethingThatDescribesThisCase = (!isHistorical || draft.createTransactions || vaccination?.stockLine)

client/packages/system/src/Vaccination/Components/SelectItemAndBatch.tsx
```diff
@@ -1,105 +1,74 @@
...
-                }
-                sx={{ flex: 1 }}
-              />
-            }
+          <SelectItem
+            disabled={givenAtOtherStore}
+            dose={dose}
+            draft={draft}
+            updateDraft={updateDraft}
           />
```
andreievg, line 47: Almost feel like there should be explicit button to change stock line (or delegate error to server, or something like this), allowing SelectBatch to be simpler, and just displaying stock line in a different component (when it's been already set)

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
andreievg: Nice

andreievg (approved):
Thanks Lache, there is nothing in this PR that I can see that would stop approval. I haven't thougth too deeply about use cases here, I think we need to ecourage QA to make try different use case.

> thougth too deeply about use cases here

I deed think of a few use case, and all of them were covered in this PR, and I was impressed

client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
lache-melvin: I think that would still be okay, if you are making edits, we should now associate the record with your facility? There's probably cases where it will end up "wrong" either way. If we are continuing with more vaccination features, I would pretty strongly push that we move towards them being "insert only". So we keep original record from first facility, from first encounter, and create a new one with whatever happens in current encounter. Would be better for tracing and reporting...

server/graphql/types/src/types/program/vaccination.rs
```diff
@@ -67,6 +71,19 @@ impl VaccinationNode {
         }
     }
 
+    pub async fn item(&self, ctx: &Context<'_>) -> Result<Option<ItemNode>> {
+        let loader = ctx.get_loader::<DataLoader<ItemLoader>>();
```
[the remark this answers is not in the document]
lache-melvin: Yeah... or like BasicItemNode, which is only the fields from ItemRow? Bc need to use loader, to get all the fields required ItemNode...

server/service/src/vaccination/update/validate.rs
lache-melvin: ugh not really... public docs do sort of discuss the different things you can do, but not clear set of use cases like you describe here... where would you write that up??

PR #7325, opened 2025-04-14
Fix yarn compilation/item warnings graphql

### LINKED ISSUE #6767: Should sync item warnings and print on prescription labels

aimee-mcneil-melville:
Fixes #6767 

# 👩🏻‍💻 What does this PR do?

Adds the warning node type into the stockline node within the operations.graphql for prescription line and item

This appears to be a duplication as warnings are in the item node, and the stockline node contains the item node so it is required there also

Including warnings in both areas allows correct compilation

# 🧪 Testing

- [ ] run yarn re-compile -> no errors
- [ ] yarn test -> no errors
- [ ] Check warnings print as per original PR testing steps in #7251 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -4146,7 +4146,6 @@ export type ItemFilterInput = {
   isVisible?: InputMaybe<Scalars['Boolean']['input']>;
   /** Items that are part of a masterlist which is visible in this store OR there is available stock of that item in this store */
   isVisibleOrOnHand?: InputMaybe<Scalars['Boolean']['input']>;
-  itemWarningJoinId?: InputMaybe<EqualFilterStringInput>;
   masterListId?: InputMaybe<EqualFilterStringInput>;
   name?: InputMaybe<StringFilterInput>;
   type?: InputMaybe<EqualFilterItemTypeInput>;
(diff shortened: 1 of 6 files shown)


client/packages/common/src/types/schema.ts
```diff
@@ -9761,7 +9760,7 @@ export type VaccinationNode = {
   invoiceId?: Maybe<Scalars['String']['output']>;
   notGivenReason?: Maybe<Scalars['String']['output']>;
   stockLine?: Maybe<StockLineNode>;
-  vaccinationDate?: Maybe<Scalars['NaiveDate']['output']>;
+  vaccinationDate: Scalars['NaiveDate']['output'];
```
aimee-mcneil-melville, line 9763: An update from the regeneration

CarlosNZ (approved):
LGTM, thanks @aimee-mcneil-melville !

PR #7201, opened 2025-04-03
7196 sync http request timeout

Chris-Petty:
Fixes #7196

# 👩🏻‍💻 What does this PR do?

Adds total timeout of 5mins to sync http requests using the api helper. Long duration in case reading big response on low bandwidth, and otherwise covers all bases such as getting stuck writing body in the request.

read_timeout too of 30s. If it stops reading anything from the connection for over 30s we timeout. 

image

You will see an error like this. We can make it pretty by handling it properly later.

## 💌 Any notes for the reviewer?

# 🧪 Testing

- [ ] OG interpreted
- [ ] Init OMS central
- [ ] Sync - all good
- [ ] put breakpoint in `Web Authenticate` method
- [ ] Sync - should timeout after 30s
- [ ] It's quite hard to effectively manually test the 300s overall timeout... need to limit bandwidth really

# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #7340, opened 2025-04-15
7339 Timeout on get_site_status v6 api call

### LINKED ISSUE #7339: Sync freezes on v6 push step
## What went wrong? 😲

Sync freezes on v6 step

Image

Sync log (table) says that push v6 started, but didn't finish and totals for v6 push are 0s.

## Expected behaviour 🤔

Sync does not freeze on v6 step

## Agreed solution

As per below the fix was done by @Chris-Petty but we actually had one other place where this needs to be applied and it's here:

https://github.com/msupply-foundation/open-msupply/blob/27077580e5ebf8814f874758eab58e599404948a/server/service/src/sync/api_v6/core.rs#L131

to be the same as

https://github.com/msupply-foundation/open-msupply/blob/27077580e5ebf8814f874758eab58e599404948a/server/service/src/sync/api_v6/core.rs#L98-L102

It's tempting to also add another safeguard, [here](https://github.com/msupply-foundation/open-msupply/blob/27077580e5ebf8814f874758eab58e599404948a/server/service/src/sync/synchroniser_driver.rs#L94-L97), to use tokio select and another future (as timeout to resolve after something like a day)

I have looked at the code in depth and have pretty good confidence that @Chris-Petty fix actually resolves the issue, since I can't really see anything else that could go wrong, looking at database there has been quite a few timeouts after the fix:

`select started_datetime, error_message, duration_in_seconds from sync_log where duration_in_seconds > 300`
Image

## How to Reproduce 🔨

Similar to what @Chris-Petty was able to achieve with https://github.com/msupply-foundation/open-msupply/pull/7201

try adding this `tokio::time::sleep(std::time::Duration::from_secs(10000000)).await;` [here (on central)](https://github.com/msupply-foundation/open-msupply/blob/27077580e5ebf8814f874758eab58e599404948a/server/server/src/sync_on_central/mod.rs#L67) and sync remote

This creates:

Image

andreievg:
Fixes #7339

# 👩🏻‍💻 What does this PR do?

Adds the same timeout introduced in https://github.com/msupply-foundation/open-msupply/pull/7201 but to get_site_status api call

# 🧪 Testing

As per issue

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/api_v6/core.rs
+++ b/server/service/src/sync/api_v6/core.rs
@@ -1,4 +1,3 @@
-use reqwest::Client;
 use thiserror::Error;
 use url::ParseError;
 use util::{with_retries, RetrySeconds};
@@ -128,7 +127,10 @@ impl SyncApiV6 {
             sync_v6_version: *sync_v6_version,
         };
 
-        let result = Client::new().post(url.clone()).json(&request).send().await;
+        let result = with_retries(RetrySeconds::default(), |client| {
+            client.post(url.clone()).json(&request)
+        })
+        .await;
 
         let error = match response_or_err(result).await {
             Ok(SiteStatusResponseV6::Data(data)) => return Ok(data),


lache-melvin (approved):
LGTM, thanks Andrei. I know they wouldn't block sync, but is it worth adding the same wrapper to the other API calls, file upload/download?

andreievg:
LGTM, thanks Andrei. I know they wouldn't block sync, but is it worth adding the same wrapper to the other API calls, file upload/download?

https://github.com/msupply-foundation/open-msupply/issues/7360

Good idea thanks

PR #7391, opened 2025-04-17
#7327 Prescription loading fix

### LINKED ISSUE #7327: Circle of death when trying to create a prescription from family planning
## What went wrong? 😲

https://github.com/user-attachments/assets/6e8c9e1e-397c-4fa3-9155-063f794d8999

API shows for query prescriptionById:
```
        "invoice": {
            "__typename": "NodeError",
            "error": {
                "description": "Record does not exist",
                "__typename": "RecordNotFound"
            }
        }
```

## Expected behaviour 🤔
Should be able to create a prescription from the family planning module

## How to Reproduce 🔨

1. Have the family planning program set up
2. Try to create a contraceptive prescription
3. See circle of death ❌ 

## Your environment 🌱

- Open mSupply Version: v2.6.2-RC2

- Platform:

- Database type:

CarlosNZ:
Fixes #7327 (sort of)

# 👩🏻‍💻 What does this PR do?

I'm just putting in a temporary workaround to remove the immediate problem (for impending demo). Will need to get to the bottom of the React Query issues at some point

The bug **only appears in production builds**. I've narrowed this down to inconsistent behaviour in React Query -- the `loading` and `error` values are not being updated correctly after the invoice is not found (`loading` stays `true`, so the spinner never goes away).

So for now I've just hidden the "Loading" indicator completely. I tried a couple of other things, namely returning a custom loading property from the hook, but it caused other problems, so leaving that out for now.

I'll post a link to this PR in the Update Tanstack issue.

# 🧪 Testing

Must be tested on Production build:
- run `yarn build` from the client folder
- run the server `cargo run`, and then the built front-end will be served at `localhost:8000` (or whatever port you're using)

Need to be set up with Family Planning module

- [ ] Go to an encounter, or create a new one
- [ ] Click on "Méthode contraceptive"
- [ ] Select something in "Méthode Retenue:" which should enable the "Prescription" line
- [ ] Select an item from the list
- [ ] You should see Prescription form
- [ ] Issue some items, close the modal, then "Save" the enounter
- [ ] Prescription should be created (A toast notification, plus check in Prescriptions)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
+++ b/client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
@@ -135,7 +135,7 @@ const useGetById = (invoiceId: string | undefined) => {
     if (invoice?.__typename === 'InvoiceNode') {
       return invoice;
     } else {
-      console.log('No invoice found', invoiceId);
+      console.error('No invoice found', invoiceId);
       throw new Error(`Could not find invoice ${invoiceId}`);
     }
   };
--- a/client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
+++ b/client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
@@ -135,7 +135,8 @@ const useGetById = (invoiceId: string | undefined) => {
     if (invoice?.__typename === 'InvoiceNode') {
       return invoice;
     } else {
-      throw new Error('Could not find invoice');
+      console.log('No invoice found', invoiceId);
+      throw new Error(`Could not find invoice ${invoiceId}`);
     }
   };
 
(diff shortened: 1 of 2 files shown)


lache-melvin (approved):
Works great, thanks for quick solution and updating the other React Query issue, will push to get that upgrade in ASAP! 🙏

client/packages/invoices/src/Prescriptions/api/hooks/usePrescription.ts
```diff
@@ -135,7 +135,8 @@ const useGetById = (invoiceId: string | undefined) => {
     if (invoice?.__typename === 'InvoiceNode') {
       return invoice;
     } else {
-      throw new Error('Could not find invoice');
+      console.log('No invoice found', invoiceId);
```
lache-melvin, line 138: Though, this fails compilation step. I think you can use `console.error()` instead?

PR #7406, opened 2025-04-17
7093 add sort toggle to prevent report filter dropdowns overlapping

### LINKED ISSUE #7093: In reports, filter dropdowns are stacked/overlapping together
## What went wrong? 😲

Using reports on Android (apk)

Sometimes (not all the time), when clicking on one, second one pops up, resulting in dropdowns stacking together and overlapping - can it be caused by keyboard?

Happens across all reports, but not consistently - if unable to replicate, just click around until you can

https://github.com/user-attachments/assets/79acdacf-8966-45eb-8a6c-ac51394881e5

## Expected behaviour 🤔

Should only be showing the dropdown you clicking on

## How to Reproduce 🔨

1. Go to Reports
2. Click on any report from Stock & Items for example
3. Click on one dropdown -> see if you can get the second one show up

## Your environment 🌱

- Open mSupply Version: V2.6.1-RC4

- Database type: PostgreSQL

aimee-mcneil-melville:
Fixes #7093

# 👩🏻‍💻 What does this PR do?

Replaced the Sort Direction dropdown selection in report arguments with a new SortToggle component (Json forms component), which displays buttons for Ascending and Descending sorting
Screenshot 2025-04-17 at 3 57 06 PM

With some testing the original issue was found to be due to android defaults in relation to the keyboard:
When the keyboard is closed whilst a dropdown is open, the dropdown direction goes down.  Then selecting another dropdown, the keyboard goes up. The dorwdowns then seen to  get confused with the direction changes and overlap. 

Rather than changing default behaviour of keyboards or dropdowns, I thought it would be a nicer user experience to have buttons for the sorting direction - it will only be one of two options, and is easier for the user to access. This also solves the overlapping - if there aren't two dropdowns next to each other, they can't overlap.

The selection behaviour is done through state, and selecting an already selected button will unselect it. I looked into the MUI [ToggleButton](https://mui.com/material-ui/react-toggle-button/#exclusive-selection) as they use an `exclusive` prop - however didn't see any benefit to creating this component when they also used state to manage this. 

First go at the styling - can have another pass with suggestions. I used the flat button so it didn't visually compete with the dialog buttons, and as it is not performing an action visible immediately to the user that made more sense to me.

1. This issue could still arise in the future if more dropdown fields are added to report arguments

2. This icon is the existing one for sorting (used in table headers). MUI doesn't have an icon in a similar design to the one suggested by Roxy - this could be looked at in future to either create custom SVGs or import icons from elsewhere
This icon also does not take a color - black and white only (Unless I missed something there).

3. The more I looked the more inconsistencies I found in the form alignments 😢 

- In all forms the labels are aligned to 'top' instead of horizontally centered like elsewhere in the app.
- There is a mix of left aligned fields and right aligned fields. Most fields have padding-right to adjust widths, except for sort-by and sort direction. This was most apparent in the expiring items report.
- I didn't find any quick fixes for the expiring items alignment, as I got further into it I figured it should be a separate issue.

4. Are there any custom report that are affected by this bug, or could benefit from using this component?

# 🧪 Testing

- [ ] from /server: `cargo build` then run `./target/debug/remote_server_cli build-reports && ./target/debug/remote_server_cli upsert-reports -o `
- [ ] Go to Reports -> for each of the 'Stock & Items' and 'Expiring' reports, open the arguments modal and enter some sorting criteria
- [ ] Check that a selection can be unselected
- [ ] Check the ascending and descending buttons cannot both be selected at once
- [ ] Click Ok -> check that any selections made are correct in the resulting report 

# 📃 Documentation

  1. There is no mention of sorting in the docs - should this be added?
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+    '&:hover': {
+      backgroundColor: alpha(theme.palette.primary.main, 0.3),
+    },
+  });
+
+  const getSelectedSx = (value: SortDirection) =>
+    sortDirection === value ? selectedStyles : undefined;
+
+  const handleClick = (value: SortDirection) => {
+    const newValue = sortDirection === value ? null : value;
```
aimee-mcneil-melville, line 39: This line with `newValue` ensures that only one button can be selected at a time - deselecting the button if it’s already selected, or selecting it if it’s not. This is what the MUI ToggleButton prop`exclusive` achieves, whilst using common components we already have

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <DetailInputWithLabelRow
+      label={label}
+      DisabledInput={!enabled}
+      labelWidthPercentage={FORM_LABEL_WIDTH}
+      sx={DefaultFormRowSx}
+      Input={
+        <>
+          <FlatButton
+            label={t('report.ascending')}
+            onClick={() => handleClick('asc')}
```
aimee-mcneil-melville, line 54: Passing in 'asc' as the value, as FlatButton does not have a value attribute. Works in this scenario where only two options exist

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+const UIComponent = (props: ControlProps) => {
+  const { handleChange, label, path, enabled } = props;
+  const t = useTranslation();
+  const [sortDirection, setSortDirection] = useState<SortDirection>();
+
+  if (!props.visible) {
+    return null;
+  }
+
+  const selectedStyles = (theme: Theme) => ({
```
aimee-mcneil-melville, line 27: custom styling for when the button is the selected option

standard_reports/expiring-items/latest/argument_schemas/arguments.json
```diff
@@ -26,8 +26,8 @@
         },
         "dir": {
           "description": "sort by dir",
-          "type": "string",
-          "enum": ["asc", "desc"]
+          "type": ["string", "null"],
```
aimee-mcneil-melville, line 29: Included 'null' here to prevent console errors when deselecting

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
Chris-Petty, line 1: Looks good. If I could request thinking about one thing, it's that this is pretty hard coded to the sorting use case. A more generic and flexible component could probably be made of this (perhaps in the future)

Chris-Petty (changes requested):
The good news is that the new buttons look good. The bad news is that sorting doesn't work?

image

image

standard_reports/item-usage/latest/argument_schemas/arguments.json
```diff
@@ -35,8 +35,8 @@
         },
         "dir": {
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Yes that is quite a difference between the property names! It's not immediately obvious (at least to me) what 'dir' represents. I'd be happy to update that to 'direction'
The description is used for internal/documentation only atm, though it looks like another use case is for a Tooltip

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
aimee-mcneil-melville: Thanks for bringing that up, I should have covered that in the PR! 
I did consider this when creating it, and went with the most straight forward solution to fix the issue. With sorting direction only having two possible options, and without another use case in mind, it made sense to me to hardcode for this case

There is an existing forms pattern with 'options' like in the 'sort-by' dropdown below - to make that pattern work for a flexible toggle button setup it would need at least:
- logic in the toggle component to get the icon and render to correct icon/label pairing
- logic to get and set the 'value' as it is not built in to the FlatButton
- considerations for control over the number of options that could theoretically be added and/or new logic plus styling if able to add more than two (more like a multi-select)

https://github.com/msupply-foundation/open-msupply/blob/35917abd86adac39640de94bb56297df0958cdc6/standard_reports/item-usage/latest/argument_schemas/arguments_ui.json#L31-L46

aimee-mcneil-melville:
Sorting should be consistent now - it was a gotcha with state. Thanks for picking up on that! 
LMK if I should rename all 'dir's to 'direction' on all of the forms as well

standard_reports/item-usage/latest/argument_schemas/arguments.json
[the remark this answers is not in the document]
Chris-Petty: Honestly the reason I even commented is it took me a second too - I was like wait `directory`?

Nah maybe a refactor issue if anything heh.

Yea I thought description might be a tool tip, but then it'd need localisation! Maybe it should be regardless, but the use case isn't really obvious heh.

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
Chris-Petty: Yea with 1-3 short options (especially if icons and no text) toggle is good, <=~5 radio buttons (especially if text) and beyond that is combobox.

client/packages/programs/src/JsonForms/common/components/SortToggle.tsx
```diff
@@ -38,7 +38,7 @@ const UIComponent = (props: ControlProps) => {
   const handleClick = (value: SortDirection) => {
     const newValue = sortDirection === value ? null : value;
     setSortDirection(newValue);
-    handleChange(path, sortDirection);
+    handleChange(path, newValue);
```
Chris-Petty, line 41: oh hahahaha ya that'd do it

Chris-Petty (approved):
fixed woo thanks

standard_reports/item-usage/latest/argument_schemas/arguments.json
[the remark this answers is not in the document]
aimee-mcneil-melville: Thanks Chris, I have raised in #7546

alainsussol:
@aimee-mcneil-melville 

Hi Aimee,

PR #7425, opened 2025-04-18
7412 add max discount percentage graphql node in ItemNode

### LINKED ISSUE #7412: Add master lists max. discount percentage graphql node in ItemNode
Comoros's Facture client report needs to display max. discount percentage from master list records linked to a particular item in the report. Current `master_lists` graphql node in the `ItemNode` looks up only visible master lists by default. Plus it also further filter based on `master_list_name_join` for given store_id the report is running from.

Something similar to this, https://github.com/msupply-foundation/open-msupply/blob/f1d275cc370949d1c8aa01349c5f0279980ba056/server/service/src/pricing/item_price.rs#L54-L70

raviSussol:
Fixes #7412 

# 👩🏻‍💻 What does this PR do?
Added `maxDiscountPercentage` graphql node in the `ItemNode` which returns max. value of discount percentage from all linked master lists to the given item. Would be used in https://github.com/msupply-foundation/open-msupply-reports/pull/204

# 🧪 Testing

- [ ] Make sure you have a master list setup in the mSupply and linked to any customer in the oms site.
- [ ] Set a discount % in the master list from the mSupply as shown in this image:
Screenshot 2025-03-25 at 12 57 54 PM
- [ ] Sync to the oms site and create an outbound shipment > add items from master list
- [ ] Confirm deliver that outbound shipment.
- [ ] In the graphql editor.
- [ ] Run the below `invoiceLines` graphql query:
```graphql
query invoice_lines(
  $storeId: String!
  $invoiceId: String!
) {
  invoiceLines(
    storeId: $storeId
    filter: { invoiceId: { equalTo: $invoiceId } }
  ) {
    ... on InvoiceLineConnector {
      nodes {
        item {
          maxDiscountPercentage
        }
      }
    }
}
```
- [ ] Now goto the mSupply and goto that oms store
- [ ] Double click it and in the `Master lists` tab check that master list which has discount percentage linked to your item
- [ ] Again the above graphql query and now you should still see the max discount percentage for each invoice line

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/core/src/loader/loader_registry.rs
+++ b/server/graphql/core/src/loader/loader_registry.rs
@@ -403,6 +403,12 @@ pub async fn get_loaders(
         async_std::task::spawn,
     ));
     loaders.insert(DataLoader::new(
+        DiscountMasterListByItemIdLoader {
+            service_provider: service_provider.clone(),
+        },
+        async_std::task::spawn,
+    ));
+    loaders.insert(DataLoader::new(
         ReasonOptionLoader {
             connection_manager: connection_manager.clone(),
         },
--- a/server/graphql/core/src/loader/master_list.rs
+++ b/server/graphql/core/src/loader/master_list.rs
@@ -62,3 +62,38 @@ impl Loader<MasterListByItemIdLoaderInput> for MasterListByItemIdLoader {
         Ok(output)
     }
 }
+
+pub struct DiscountMasterListByItemIdLoader {
+    pub service_provider: Data<ServiceProvider>,
+}
+
+impl Loader<String> for DiscountMasterListByItemIdLoader {
+    type Value = f64;
+    type Error = async_graphql::Error;
+
+    async fn load(&self, item_ids: &[String]) -> Result<HashMap<String, Self::Value>, Self::Error> {
+        let service_context = self.service_provider.basic_context()?;
+        let connection = service_context.connection;
+        let repository = MasterListRepository::new(&connection);
+
+        let mut result = HashMap::<String, Self::Value>::new();
+        for item_id in item_ids {
+            let discount_master_list = repository.query_by_filter(
+                MasterListFilter::new()
+                    .is_discount_list(true)
+                    .include_inactive(true)
+                    .item_id(EqualFilter::equal_to(item_id)),
+            )?;
+
+            // Get the max discount percentage from the discount master lists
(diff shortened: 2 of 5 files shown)


andreievg:
@raviSussol you can use sql statement and map it with convert_data_js. Do you want an example ?

Otherwise we would need to use logic in `get_pricing_for_item` (and refactor it to get prices for items). 

One issue for above would be that discount will be shown in invoice line even when it was not given (it was a patient).

Either way both solutions would go into 2.7

raviSussol:
Hi @andreievg,
> Otherwise we would need to use logic in get_pricing_for_item (and refactor it to get prices for items).
One issue for above would be that discount will be shown in invoice line even when it was not given (it was a patient).

You mean the changes in [here](https://github.com/msupply-foundation/open-msupply/pull/7425/files#diff-39aa258bfe62eb62fc02f000bba360c3fdd30e7a76a5a5bbab16bb695a851d7e) would also affect the price calculation for an item in `get_pricing_for_item` function? The query filter still looks up for active master lists by default. Shouldn't differ what `get_pricing_for_item` function calculates I think.

> you can use sql statement and map it with convert_data_js. Do you want an example ?

You mean to get max. discount percentage from master lists part hmm... yeah can use sql for that. Actually I had thought to refactor facture client report's datasource with sql instead of graphql before but later I realized that might not be worth doing it thinking this discount percentage field can be reused by other clients too. Now that you've said it I'll try to use sql to get max discount percentage in the report itself. 👍

andreievg:
@raviSussol

> you mean to get max. discount percentage from master lists part hmm

Yep and combined that with current graphql query using convert_data_js

> You mean the changes in [here](https://github.com/msupply-foundation/open-msupply/pull/7425/files#diff-39aa258bfe62eb62fc02f000bba360c3fdd30e7a76a5a5bbab16bb695a851d7e) would also affect the price calculation for an item in get_pricing_for_item function

Nope, i mean if we are exposing discounts and prices from master lists, we should use the same logic rather then creating new branch of logic for display. But it would require a bit of refactor and a bit of re-testing.

Will close this PR for now, looking forward to seeing how things look in the report repo

PR #7431, opened 2025-04-21
7347 patients accessible on central without related records

### LINKED ISSUE #7347: Patients can be visible in OMS central without related data
## What went wrong? 😲

#7272 syncs OMS remote patients to OMS central. This means if OMS Central store is a dispensary, you will be able to see all of the patients from remote sites. However, name_store_join between for patient on OMS Central store has not been added, so Legacy mSupply Central won't sync related patient records to the Central store. 

i.e. you wouldn't have access to patient encounters etc. Also won't receive updates to patient record, could have stale data.

## Expected behaviour 🤔

All patient data for visible patients should be available on OMS central.

Legacy should remain aware of all sites where patient is visible.

## Proposed solution

Add a central only processor, if name_store_join records are received on OMS central from OMS remote site.

This should look up whether it has a name_store_join record between said name_link_id, and a store on central site. If not exists, call `link_patient_to_store` (call OG to add name store join for patient to central store)

Patient records would be received on next sync

## How to Reproduce 🔨

1. After #7272 fix merged
2. Create a patient on an OMS remote site
3. Enrol in a program and create an encounter
4. Sync
5. Log into OMS Central
6. You can see the new patient
7. Encounter records don't exist

## Your environment 🌱

- Open mSupply Version: v2.7.0 RC4 + #7272

- Platform:

- Database type:

lache-melvin:
Fixes #7347

# 👩🏻‍💻 What does this PR do?

Adds a processor - if name records are pushed to OMS central, and no store on central has visibility for them, call legacy OMS to add visibility.

will leave inline comments

# 🧪 Testing

- [ ] Create a patient in OMS remote site
- [ ] Enrol them in a program and create an encounter
- [ ] Sync
- [ ] Log in to OMS Central
- [ ] You can see the patient record
- [ ] You can see the encounter

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/apis/api_on_central/patient_name_store_join.rs
```diff
@@ -62,38 +67,7 @@ pub async fn add_patient_name_store_join(
     Ok(())
 }
 
-async fn add_patient_to_central(
```
lache-melvin, line 65: moved to search_central file, for reusability

server/service/src/processors/add_central_patient_visibility/add_central_patient_visibility.rs
```diff
@@ -0,0 +1,153 @@
...
+// patient records and receives appropriate updates via sync.
+// See service/programs/patient/README for more info
+
+// If patient is created on remote site, it is possible the patient won't exist on Legacy yet the
+// first time this processor is triggered (remote site pushes to OMS Central before Legacy Central).
+// Processor would fail after trying to add visibility to  OMS Central via call to Legacy.
+// Retry logic doesn't (yet) exist for processors.
+// Hence, this processor watches for name changelogs as well as name_store_join. Next time patient
+// record is mutated on remote site and pushed to OMS Central, we see we don't have visibility and
+// will retry. Not foolproof, but better than nothing for now.
```
lache-melvin, line 28: Thoughts and feelings about this?

server/service/src/processors/add_central_patient_visibility/add_central_patient_visibility.rs
```diff
@@ -0,0 +1,153 @@
...
+            debug!("Patient already visible on central, skipping");
+            return Ok(None);
+        }
+
+        let patient_id = patient.id.clone();
+
+        // Give us the ability to do an async thing in a sync context
+        // Otherwise we'd have to make all processors async...
+        tokio::task::block_in_place(|| {
+            tokio::runtime::Handle::current().block_on(async {
```
lache-melvin, line 115: So this ain't working, @andreievg tagging you as you have context but anyone send help plz lol

Processors are not async, and they only receive a reference to the service provider. We now want to make an async call from a processor... options are:

- Make all processors async (this is probably best, but breaks current traits, felt like bigger refactor than this PR, though maybe worth it?
- Pass Arc<ServiceProvider> into each processor, so they can clone service provider and spawn async processors as needed (this would result in a LOT of clones of service provider, one for every changelog processed, not ideal)
- Something like this, icky little block the thread moment. It's fine because we spawn processors separately, so if they take a while to run, all good. However, block_in_place requires we're running in multi_thread mode, which it would seem processors are not
  - Sooo we get `Processor terminated (Err(JoinError::Panic(Id(4), "can call blocking only when running on the multi-threaded runtime", ...)))'` and the server crashes lol
  - Idk if there are implications that I haven't considered by making processors multithread, I think fine, maybe good actually??
  - I think this is other valid option, do you have pref @andreievg between multithread and making async?

server/service/src/processors/add_central_patient_visibility/test.rs
```diff
@@ -0,0 +1,129 @@
...
+    ctx.processors_trigger
+        .general_processor
+        .try_send(ProcessorType::AddPatientVisibilityForCentral)
+        .unwrap();
+    log::debug!("await_events_processed");
+    ctx.processors_trigger.await_events_processed().await;
+
+    // ABSOLUTE HACK
+    // Currently not possible to mock the call to central from the processor
+    // So let's just check it errors in the right place :violent_sob:
```
lache-melvin, line 96: *\*cries in please_give_me_a_mocking_framework\**

server/service/src/processors/add_central_patient_visibility/add_central_patient_visibility.rs
andreievg: Would have gone with making general processor async with `#[async_trait]`, might actually not have been that much more work. 

On another note, i still want to make everything async, including service and repository, it's a bit of a dream (one of the oldest issue in repo)

server/service/src/processors/add_central_patient_visibility/test.rs
andreievg: *cries in please_give_me_a_mocking_framework*

I think this would have worked: https://github.com/msupply-foundation/open-msupply/blob/d9bd2ceb6402226b9db41e4ee5ec498b5bfb9cdb/server/service/src/sync/api/get_site_status.rs#L41-L58

andreievg (commented):
Looks great. Yes we do at some point need a way to request re-trigger, without major refactor, this https://github.com/msupply-foundation/open-msupply/blob/d9bd2ceb6402226b9db41e4ee5ec498b5bfb9cdb/server/service/src/processors/mod.rs#L111 can return a request for timeout and we tokio::spawn it

andreievg:
As per our conversation though, i think it might be better to just filter patient by name_store_join for at least one active store on site, sounds like it's less surface area to test and maintain (especially since this is a rare use case), and also makes more sense for user, active Dispensary store on central only sees patient it has created or pulled from central (but... also other active Dispensaries on central see, so there is a small contradiction here about 'sense for user')

server/service/src/processors/general_processor.rs
```diff
@@ -120,15 +131,16 @@ pub(super) trait Processor {
...
         service_provider: &ServiceProvider,
         changelog: &ChangelogRow,
     ) -> Result<Option<String>, ProcessorError> {
-        let result = ctx
-            .connection
-            .transaction_sync(|_| self.try_process_record(ctx, service_provider, changelog))
+        // TODO: should be in a transaction, we need to support transaction_async
+        let result = self
+            .try_process_record(ctx, service_provider, changelog)
+            .await
```
lache-melvin, line 143: had to remove transaction to make async, but I checked, all processors do max of 1 DB write for now, so I think ok, I can make an issue?

server/service/src/processors/add_central_patient_visibility/add_central_patient_visibility.rs
lache-melvin: Sweet thanks, async did require less change than I thought 😁 [9d71176](https://github.com/msupply-foundation/open-msupply/pull/7431/commits/9d7117604bf279dcadd61d2ebfe0efe0b5ff5676)

server/service/src/processors/add_central_patient_visibility/test.rs
lache-melvin: ah yeah nice! i think would still require further test setup and I'm super aware of time right now... would it be ok, do this separately?

lache-melvin:
> Looks great. Yes we do at some point need a way to request re-trigger, without major refactor, this
> can return a request for timeout and we tokio::spawn it

Think I'm following... though not quite sure how re-processing would work, as we have bumped the cursor? Future problem 😁

lache-melvin:
> As per our conversation though, i think it might be better to just filter patient by name_store_join for at least one active store on site, sounds like it's less surface area to test and maintain (especially since this is a rare use case), and also makes more sense for user, active Dispensary store on central only sees patient it has created or pulled from central (but... also other active Dispensaries on central see, so there is a small contradiction here about 'sense for user'), still would like to go with filters solution

I'm on the fence about this, it feels really wrong to me for patient `name` records to exist on OMS central, and Legacy not know about that. I think best solution would be the central ghost store we talked about, so we could always add patient visibility to the central ghost store, ensure patient records are synced there, and then we also have the name_store_join filter you're talking about, so the _actual_ active stores on OMS central only see the patients they should? Don't think we have time for that now though... what are you asking be done before 2.7 merge?

lache-melvin:
Requesting re-review, works now after async change 🙏

server/service/src/processors/add_central_patient_visibility/test.rs
andreievg: Sure, no more tears thought

andreievg (changes requested):
Looks good, although I think that the filter solution would have been more suited to the current timing, I think your solution is already implemented and it's solid, and I am happy with it.
I think we need a TODO: (deprecate when all sites are > 2.7)

I do have one request, we had issues with connections being held by networks within countries, maybe it's internal intranet setup or malicious network nodes. This was fixed with: https://github.com/msupply-foundation/open-msupply/blob/9d7117604bf279dcadd61d2ebfe0efe0b5ff5676/server/service/src/sync/api/core.rs#L169-L170. 

Can you please use that one when remote sites are requesting patient from central (don't want processor to get stuck), i think it's ok for oms central to og central coms, and things should also be ok for roms to coms during normal patient remote fetch (yes api could block, but doesn't stop processors or sync)

lache-melvin:
@andreievg - applied with retries to the name_store_join endpoints, simplifies the API too, not having to new up a client on create... are you okay with this? [381e62a](https://github.com/msupply-foundation/open-msupply/pull/7431/commits/381e62a2eaddfbcce4683cba91de2f6a10c0dbe5)

andreievg (approved):
Yay ! Thanks for all of the changes, looks great

PR #7436, opened 2025-04-22
Update verifying functions

### LINKED ISSUE #7363: The page gets stuck when you try to add item by masterlist in the Inbound shipments that has been already delivered
## What went wrong? 😲

Found while testing: https://github.com/msupply-foundation/open-msupply/issues/7362

The page gets stuck when you try to add item by masterlist in the Inbound shipments that has been already delivered

## How to Reproduce 🔨

1. OMS and mSupply setup
2. Create Inbound shipments
3. Make it delivered status
4. Re-open the inbound shipments
5. Now try to add items by masterlist
6. Click on the alert
7. The page gets stuck (BUG)

https://github.com/user-attachments/assets/a619b6b5-ae0c-490f-83fd-1627384df9ae

## Your environment 🌱

- Open mSupply Version: v2.6.01-RC6
- Legacy mSupply Central Server Version: V8-01-12

- Platform: browser 

- Database type: SQLite3

fergie-nz:
Fixes #7363 

# 👩🏻‍💻 What does this PR do?

Changes verifying check to allow for shipped and delivered inbound shipments to be delivered.

# 🧪 Testing

- [ ] OMS and mSupply setup
- [ ] Create Inbound shipments
- [ ] Make it delivered status
- [ ] Re-open the inbound shipments
- [ ] Now try to add items by masterlist
- [ ] See added successfully

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1985,8 +1985,6 @@
   "warning.nothing-to-supply": "Nothing left to supply!",
   "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix",
   "label.cannot-edit-invoice": "Cannot edit invoice",
-  "label.record-not-found": "Record not found",
-  "label.master-list-not-found": "Master list not found",
   "label.cannot-add-item-to-shipment": "Cannot add items to shipment",
   "error.cannot-add-from-masterlist-if-not-new": "Can only add from master list on 'New' invoice"
 }
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1988,4 +1988,5 @@
   "label.record-not-found": "Record not found",
   "label.master-list-not-found": "Master list not found",
   "label.cannot-add-item-to-shipment": "Cannot add items to shipment",
+  "error.cannot-add-from-masterlist-if-not-new": "Can only add from master list on 'New' invoice"
 }
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1983,5 +1983,9 @@
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm against the patient's physical vaccination card first.",
   "warning.field-not-parsed": "{{field}} not parsed",
   "warning.nothing-to-supply": "Nothing left to supply!",
-  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix"
+  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix",
+  "label.cannot-edit-invoice": "Cannot edit invoice",
+  "label.record-not-found": "Record not found",
+  "label.master-list-not-found": "Master list not found",
+  "label.cannot-add-item-to-shipment": "Cannot add items to shipment",
 }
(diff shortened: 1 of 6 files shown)


fergie-nz:
> Hm ok few things...
> 
> * Looks like 2.6.3 patch is being built now, so if we can rebase this to 2.7.0. Checked with Andrei, should be pretty unlikely use case so okay to fix in 2.7.
> * We need to fix the base issue (the confirmation modal spinning forever after the error.) To do this, need to handle/catch the error.
> 
> Add items from master list is using `onError`, which is doing the toast, but `onError` is just a callback, it still propagates the error back to the caller. We need to catch that error, this would be the change (sorry diff is a bit munty, just adding try catch block with an await):
> 
> [7363-allow-edit-adding-via-masterlist...7363-allow-edit-adding-via-masterlist(handle-api-error)](https://github.com/msupply-foundation/open-msupply/compare/7363-allow-edit-adding-via-masterlist...7363-allow-edit-adding-via-masterlist(handle-api-error))
> 
> And those error messages are hard coded! While there, would be good to update to use translation strings 😳
> 
> _Then_...
> 
> If we allow user to add from master list outside of "New" status, we could end up with placeholder lines in our delivered and verified shipments, which isn't a case we support. I checked with @mark-prins, who said yep, if users are adding an entire master list to an Inbound Shipment after it is in Delivered status, that's a process problem. Should only really be adding an individual item at that stage.
> 
> Instead, he said
> 
> * Keep existing "new" status check in the backend
> * When "Add from master list" button clicked in frontend, before opening the confirmation modal, do a status check. If not in new status, instead have a dialog popup, explaining that you can't add from master list as no longer in New status
> 
> Apologies this wasn't better communicated in the original issue Ferg!

No worries at all!

I'll tackle these changes tomorrow, and rebase from 2.7

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1983,5 +1983,10 @@
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm against the patient's physical vaccination card first.",
   "warning.field-not-parsed": "{{field}} not parsed",
   "warning.nothing-to-supply": "Nothing left to supply!",
-  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix"
+  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix",
+  "label.cannot-edit-invoice": "Cannot edit invoice",
+  "label.record-not-found": "Record not found",
+  "label.master-list-not-found": "Master list not found",
```
lache-melvin, line 1989: these two already have translations, i think messages.record-not-found - probably best to reuse 👀

client/packages/invoices/src/InboundShipment/DetailView/AddButton.tsx
```diff
@@ -67,7 +70,9 @@ export const AddButton = ({
         onAddItem(true);
         break;
       case 'add-from-master-list':
-        masterListModalController.toggleOn();
+        invoice?.status === InvoiceNodeStatus.New
+          ? masterListModalController.toggleOn()
+          : error(t('error.cannot-add-from-masterlist-if-not-new'))();
```
lache-melvin, line 75: Could you change this to use info() rather than error() - we're just letting the user know why the action is unavailable, as opposed to something has gone wrong 🙏

lache-melvin (changes requested):
Looks awesome Ferg thanks for all the changes, just requesting a couple small clean ups and then good to go 🚀

client/packages/common/src/intl/locales/en/common.json
fergie-nz: Ah great thanks Lachè - I missed those

lache-melvin (approved):
ship it 😁

PR #7449, opened 2025-04-22
add Open_mSupply prefix to APK zip files

lache-melvin:
Soz, lil side quest

# 👩🏻‍💻 What does this PR do?

When downloading our installers, they're all co-located because of Open_mSupply prefix, but these zip files were harder to find in downloads folder, just called "Universal_Build"

Adds prefix so all installers are together

Apologies, formatter had a field day - its actually just the names that changed

I think would be good to also put the version in the name? So if you download multiple, you can see which version they are for rather than `Open_mSupply_Universal_Build(2)`

Would require getting version in action outputs though - #7450 

# 🧪 Testing

- [ ] See Artifacts from this action: https://github.com/msupply-foundation/open-msupply/actions/runs/14605776894
- [ ] Output has the updated names for the zipped artifacts
- [ ] Download - if your downloads are sorted alphabetically, the zip files will be chilling with the other installers

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/.github/workflows/build-android-apk.yaml
+++ b/.github/workflows/build-android-apk.yaml
@@ -1,8 +1,8 @@
 on:
- push:
-  tags: 
-   - 'v*'
- workflow_dispatch:
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
 
 name: Android Build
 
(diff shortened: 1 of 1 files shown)


aimee-mcneil-melville (approved):
Great, looks good thanks Laché!

PR #7457, opened 2025-04-23
7370 vaccine emergency order notifications

### LINKED ISSUE #7370: Vaccine Support: Notify users when an urgent order is received
#### Backend
- [ ] Add ability to filter requisitions by the `is_emergency` field
- [ ] Add filter `automatically_created` where linked_requisition_id == null
- [ ] Add function to get statistics for # of new emergency automatically created requisitions under `RequisitionCountService`

#### Frontend
- [ ] Add `emergency_order` boolean toggle to the Requisition page
- [ ] Hook for function to get # of new emergency automatically created requisitions
- [ ] On the dashboard under `Distribution` -> `Requisitions`, display `Emergency Order`
- [ ] When users click on `Emergency Order`, they will be redirected to the Requisition page with the filters:
   - [ ] Status == new
   - [ ] is_emergency == true
   - [ ] automatically_created == true 

### Acceptance Criteria ✅
- [ ] Users can see # new emergency orders on their dashboard
- [ ] Users are redirected to the Requisition page where they will see a list of new emergency orders when they click on the link on the dashboard
- [ ] Users can filter requisitions by emergency/automatically created

### Moneyworks Jobcode 🧰
NIGGF:VAX

fergie-nz:
Fixes #7370

# 👩🏻‍💻 What does this PR do?

Adds requisition filter for 'is_emergency' and 'automatically_created'
Adds emergency requisition statistics to ReqyuisitionCounts
Adds emergency order filter toggle to requisition list view
Adds emergency order count to dashboard widget with redirect

# 🧪 Testing

- [ ] Check dashboard includes emergency order count (response requisitions which are new, is_emergency, and is_automatically_created)
- [ ] Check clicking on this widget redirects to responseRequisition list view with appropriate filters set.
- [ ] Check that changing filters  in requisition list view works as expected with new isEmergencyAndAutomatic filters being set by one toggle.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/dashboard/src/widgets/DistributionWidget.tsx
```diff
@@ -116,6 +116,16 @@ export const DistributionWidget: React.FC = () => {
                     .addQuery({ status: RequisitionNodeStatus.New })
                     .build(),
                 },
+                {
+                  label: t('label.emergency'),
```
fergie-nz, line 120: Have left this as emergency for succinctness, but maybe needs to be more specific to show that it is emergcy AND automatically created AND new?

client/packages/requisitions/src/ResponseRequisition/api/api.ts
```diff
@@ -105,6 +105,17 @@ const responseParser = {
...
     optionId: patch?.reason?.id ?? null,
   }),
+  toListFilter: (
+    filterBy: FilterByWithBoolean | null
+  ): FilterByWithBoolean | null => {
+    Object.assign(filterBy ?? {},  {type: { equalTo: RequisitionNodeType.Response }});
+    if (filterBy?.['isEmergencyAndAutomatic']) {
+      filterBy['isEmergency'] = filterBy['isEmergencyAndAutomatic'];
+      filterBy['automaticallyCreated'] = filterBy['isEmergencyAndAutomatic'];
+    }
```
fergie-nz, line 115: I couldn't see if we have a pattern for multiple filters combined by one toggle already, so have mapped here.

I wanted to add is_emergency as one filter, but it needed to be split to two filters (is_emergency & automatically_created) in the back end.

Maybe there is a better pattern?

Have also moved the type filter up to the parser so it is all in one place.

client/packages/dashboard/src/api/operations.graphql
```diff
@@ -32,6 +32,9 @@ query requisitionCounts($storeId: String!) {
     response {
       new
     }
+    emergency {
+      new
+    }
```
fergie-nz, line 37: Have added emergency query into other requisition count queries.

This seemed a lot more simple than adding it to its own hook - but maybe I'm not aware of a use case where it would need its own hook?

jmbrunskill (changes requested):
Thanks @fergie-nz good implementation of the known requirements!

I've discussed with @mark-prins though and we've agreed we should simplify the dashboard view and the filters to just filter on the is_emergency status.

Adding the `Automatically created` filter is potentially a bit confusing for the users as like you pointed out it's not obvious from the user perspective that additional filters are being applied.

If we were to do a pop up notification in future, then we would need to filter out any emergency orders created manually (or potentially just created by the current user/store) as you don't want to create something and immediately have something pop up to tell you that you have an emergency order, in that scenario we will need addtional filters.

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1986,5 +1986,7 @@
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm with the patient first.",
   "warning.field-not-parsed": "{{field}} not parsed",
   "warning.nothing-to-supply": "Nothing left to supply!",
-  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix"
+  "tooltip.defined-by-gs1-matrix": "Defined by GS1 datamatrix",
+  "label.emergency": "Emergency",
+  "label.new-emergency": "New emergency requisitions"
```
fergie-nz, line 1991: @jmbrunskill FYI have added a more verbose description on dashboard.

I thought this would make more sense because just having 'emergency' doesn't seem clear to me that it would also filter for new given that another category is explicitly called 'New'

jmbrunskill (approved):
Thanks Fergus much simplier and clearer now.
I'm wondering if it would be better to go back to just Emergency on the dashbaoad.

@mark-prins do you have an opinion?

image

vs

image

client/packages/common/src/intl/locales/en/common.json
jmbrunskill: I've updated this to "label.new-emergency-requisitions", will help with translations if we make the key more explicit when we're using it for a specific purpose.

jmbrunskill:
@andreievg I found a bug testing this with the open-msupply-plugins enabled.
The query for requisitionCounts is re-defined in that plugin, and it seems like on a refresh of the page we get the version without the `emergency` total in the request, this is something we need to watch out for and potentially safe guard against...

mark-prins:
> do you have an opinion?

the second option, with just 'Emergency'

but also.. can we have an alert icon / different colour or something plz? this emergency really needs to pop

fergie-nz:
> Member

How does this look?

Shows as red when there is more than 0 emergency requisitions

Screenshot 2025-05-02 at 12 40 21 PM

mark-prins:
still waiting for confirmation.. can you hold off merging till I get this confirmed?
was hoping for a more distinct difference - we have these status chips elsewhere (vax card) and it was proposed to have these on the dashboard:

attention

client/packages/common/src/ui/components/panels/StatusChip.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
```
fergie-nz, line 1: I've moved the status cell component into panels. Wasn't quite sure where best to put it. But have moved it out of the cell component to expose just the chip.

jmbrunskill (approved):
Thanks Fergus, looks great.

I was a big confused by the diff but opening in VScode made it all make more sense.
I also merged develop for you.

PR #7458, opened 2025-04-23
7344 display doses in stocktake

### LINKED ISSUE #7344: Dose Management: display doses in Stocktake
Note all mentions of vials is just the unit name and this can differ

## Tasks 📝
- [ ] Add doses per vial column to the stocktake detail view which shows `-` if the item is not a vaccine

### Acceptance Criteria ✅
- [ ] Users should be able to see the correct unit for each item

### Moneyworks Jobcode 🧰
NIGGF:DSE

roxy-dao:
Fixes #7344

# 👩🏻‍💻 What does this PR do?
### Backend
- Dummy `display_vaccine_in_doses` pref (will need to be refactored once prefs has been refactored)
- Defaulting stock_line.doses to item.doses for now. See note below.

#### Stocktake Detail View
- **Doses per unit**: Only shows if `display_vaccine_in_doses` preference is on, and only for vaccine items. For non-vaccine items, column will show `-`.
- **Difference**
   - Shows dose difference in brackets for vaccine items
   - Had to change this from NumberCell to display the doses... not ideal, but have rounded values to 2 DP since NumberCell isn't being used anymore

#### Stocktake Edit View
- Made the modal a bit bigger
- Function to display column label depending on pref and item type
- 2 **readonly** snapshot columns (pack and unit) and 2 **input** counted columns (pack and unit)
   - These show with/without pref. The counted columns should display the correct value regardless of which one the user edits.
   - Columns default to display `unit` if unit name not provided
- With preference:
   - For vaccine items
      - **readonly** doses per {{ unit name }}. Displays how many doses are in vial etc.
      - `Pack `labels should now show `unit name`
      - `Unit` labels should now show `doses`

Video is without `doses per` column 👀  sorry...

https://github.com/user-attachments/assets/77f283a8-c59c-44aa-8d42-0bda3ce24a7d
  
- Without preference: 
- Should still see the two **readonly** and 2 **input** columns
Screenshot 2025-04-23 at 2 22 49 PM
Screenshot 2025-04-23 at 2 23 04 PM
Screenshot 2025-04-23 at 2 22 56 PM

- Will need an issue to populate stock_line.doses with input if receiving/new batch or default to item.doses

# 🧪 Testing

- [ ] (dev only) Preference is already on for this... can change to false in backend and restart server if you want to see without preferences
- [ ] Set up some vaccine items with different dose counts
- [ ] Probably best to upgrade from 2.7 -> 2.8 so the migration updates stock_line.dose column to match item for existing stock lines
- [ ] Display vaccine in doses preference on:
   - [ ] `Doses per unit` column in detail view and populated for vaccine items
   - [ ] Doses in brackets in `difference` if vaccine item
   - [ ] 2 read only columns for snapshot (pack and doses)
   - [ ] 2 input columns for counted (pack and doses). Entering one should change the other. Make sure calculation is correct

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Item/api/operations.graphql
```diff
@@ -4,6 +4,8 @@ fragment ServiceItemRow on ItemNode {
   code
   name
   unitName
+  isVaccine
+  doses
```
roxy-dao, line 8: Will be using in quite a few places 👀

Chris-Petty:
@roxy-dao pls resolve conflicts

Chris-Petty (changes requested):
Lookin' pret' goo' 2 mysel'

image

image

I'm most worried what's happening with the number of packs counted vs the "tablet Counted". It appears that the number of packs is being labelled "Tablet Counted" but is getting saved as number_of_packs_counted without any consideration for pack_size. This would potentially throw the data outta wack? Perhaps needs to be divided by pack_size before saving?

image


little nits/thoughts:

- The "packs snapshot" and "difference" columns don't render blue for the placeholder lines?
- Wondering if number of packs should still show in the line editing
- get a lotta decimal in number
- Tangent: Many clients won't use the Reason column. I wonder why we don't conditionally render it.

roxy-dao:
> Lookin' pret' goo' 2 mysel'
> I'm most worried what's happening with the number of packs counted vs the "tablet Counted". It appears that the number of packs is being labelled "Tablet Counted" but is getting saved as number_of_packs_counted without any consideration for pack_size. This would potentially throw the data outta wack? Perhaps needs to be divided by pack_size before saving?
> This would be more clear if we still show number of packs (even if disabled).

We are still showing number of packs... It's the `{{item unit}} snapshot}} column ;x. Have made changes so that pack size is taken into consideration, sorry slipped my mind.

> * The "packs snapshot" and "difference" columns don't render blue for the placeholder lines?

That's odd.. I think it should do but have noticed that this is happening throughout our codebase 👀 

> * Wondering if number of packs should still show in the line editing

Still showing as mentioned above

> * get a lotta decimal in number

Yeah... Doses count will result in a lot of decimals... do we want to limit? 

> * Tangent: Many clients won't use the Reason column. I wonder why we don't conditionally render it.

I think we've had the conversation about conditionally hiding things vs showing it all the time a few times in the past... The consensus was to show but disable... but there are a lot of places in our codebase that does both 👀  honestly do like things hidden if they aren't used

Chris-Petty (changes requested):
image
I setup another item with a reasonable doses of 12 😉 

Packsize: 2
Doses per tab: 12
I count 1 tablet: 1
Doses are: 24??? But it tells me 12 per tablet!

The "Tablet count" column is being treated as "number of packs" and calculating as such.

This is more apparent in the item list view where the "packs counted" column is actually labeled as such:

image

Chris-Petty:
To actually request something, this is just a column header discrepancy based on pack size. I get it if your pack size is 1, "Tablets Counted" is more intuitive than "Packs counted" - you're almost certainly counting individual units. In which case we could make it a bit more dynamic - if packsize == 1 then show "Packs counted", else "Tablets Counted"?

roxy-dao:
> To actually request something, this is just a column header discrepancy based on pack size. I get it if your pack size is 1, "Tablets Counted" is more intuitive than "Packs counted" - you're almost certainly counting individual units. In which case we could make it a bit more dynamic - if packsize == 1 then show "Packs counted", else "Tablets Counted"?
> 
> Can't tweak it in the stocktake lines table though, have to just assume "Packs Counted" to stay consistent with typical stock UI in mSupply.

Sorry... requirement changed! Have removed the line edit changes and now only the detail view changes remain >.>

client/packages/common/src/intl/utils/IntlUtils.ts
```diff
@@ -134,6 +139,11 @@ export const useIntlUtils = () => {
     return localeKeySet.has(key);
   };
 
+  const getColumnLabelWithPackOrUnit = useCallback(
```
roxy-dao, line 142: Ignore this file, changed in later PR

Chris-Petty (approved):

PR #7466, opened 2025-04-24
7356 cant edit date filter

### LINKED ISSUE #7356: Date filters cannot be modified
### Cold chain -> Monitoring

When accessing Monitoring, by default the date filter will only show the last 24 hours and when a user tries to change the date filters, the little calendar appears but no date is clickable. When I tried to reset the dates and the filters, the same issue happened. 
I tried this after uploading Fridge-tag data.

https://github.com/user-attachments/assets/f34f4b72-1bf8-49b1-be4d-7c80d679bf18

## Expected behaviour 🤔

A user should be able to modify the dates in the filters to zoom in a specific period. The dates should not change switching tabs between Chart, breaches and Logs. 
Bonus point if the dates could match the imported data period. 

## How to Reproduce 🔨

1. Import new Fridge-tag data (not sure if this is relevant)
2. Go to Cold chain -> Monitoring
3. Click on Filters -> Date and time
4. Try to click a new date in the filters
5. See error
## Your environment 🌱

- Open mSupply Version: 2.7.00-RC2
- Djibouti demo server
- Store : CAMME / User : mSupply Server
- Legacy mSupply Central Server Version: not sure
- Platform: Laptop Mac. Web interface 
- Database type: not sure

This issue seems related to [https://github.com/msupply-foundation/open-msupply/issues/4251](url)

lache-melvin:
Fixes #7356

# 👩🏻‍💻 What does this PR do?

Constrain date filter ranges correctly:
Screenshot 2025-04-24 at 12 54 26 PM

# 🧪 Testing

Unit tests added...

- [ ] On coldchain -> monitoring
- [ ] You can edit the date filters
- [ ] The "from" date can't be higher than the "to" date, or  todays date
- [ ] The "to" date can't be less than the "from" date
- [ ] Could also give a quick check to other date filters (e.g. shipment list views)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
@@ -68,10 +68,9 @@ export const DateFilter: FC<{ filterDefinition: DateFilterDefinition }> = ({
     textFieldProps: {
       sx: FilterLabelSx,
     },
-    minDate: getRangeBoundary(urlValue, range, minDate),
-    maxDate: getRangeBoundary(urlValue, range, maxDate),
     actions: ['clear', 'accept'] as PickersActionBarAction[],
     displayAs,
+    ...getMinMaxDates(urlValue, range, minDate, maxDate),
   };
 
   return displayAs === 'dateTime' ? (
@@ -86,22 +85,40 @@ const getDateFromUrl = (query: string, range: RangeOption | undefined) => {
   return DateUtils.getDateOrNull(value);
 };
 
-const getRangeBoundary = (
+export const getMinMaxDates = (
   query: string | RangeObject<string>,
   range: RangeOption | undefined,
-  limit: Date | string | undefined
+  min: Date | string | undefined,
+  max: Date | string | undefined
 ) => {
-  const limitDate = DateUtils.getNaiveDate(limit);
-  if (typeof query !== 'object' || !range) return limitDate || undefined;
+  const minDate = DateUtils.getDateOrNull(min);
+  const maxDate = DateUtils.getDateOrNull(max);
+
+  if (typeof query !== 'object' || !range) {
+    return {
+      minDate: minDate || undefined,
+      maxDate: maxDate || undefined,
+    };
+  }
+
   const { from, to } = query as RangeObject<string>;
 
-  if (range === 'from')
-    return to
-      ? (DateUtils.minDate(DateUtils.getNaiveDate(to), limitDate) ?? undefined)
-      : (limitDate ?? undefined);
-  else
-    return from
-      ? (DateUtils.maxDate(DateUtils.getNaiveDate(from), limitDate) ??
-          undefined)
-      : (limitDate ?? undefined);
+  if (range === 'from') {
+    const toDate = DateUtils.getDateOrNull(to);
+    return {
+      minDate: minDate || undefined,
+      maxDate: DateUtils.minDate(toDate, maxDate) ?? undefined,
+    };
+  } else {
+    const fromDate = DateUtils.getDateOrNull(from);
+
+    return {
+      minDate:
+        // Only use maxDate if both defined (otherwise the undefined date is considered the max)
+        fromDate && minDate
+          ? (DateUtils.maxDate(fromDate, minDate) ?? undefined)
+          : (fromDate ?? minDate ?? undefined),
+      maxDate: maxDate || undefined,
+    };
+  }
 };
--- a/client/packages/common/src/ui/components/inputs/Filters/dateFilter.test.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/dateFilter.test.tsx
@@ -1,0 +1,114 @@
+import { getMinMaxDates } from './DateFilter';
+
+describe('getMinMaxDates', () => {
+  it('should return provided min and max dates when no query object', () => {
+    const query = '';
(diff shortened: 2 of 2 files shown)


client/packages/common/src/ui/components/inputs/Filters/dateFilter.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const result = getMinMaxDates(query, range, min, max);
+
+    expect(result).toEqual({
+      minDate: new Date('2025-09-01'),
+      maxDate: new Date('2025-11-01'),
+    });
+  });
+
+  describe('when range is "from"', () => {
+    it('should return max date from as to date query when no max provided', () => {
```
andreievg, line 32: Is this line correct ?

lache-melvin: english is actually my first language

client/packages/common/src/ui/components/inputs/Filters/dateFilter.test.tsx
```diff
@@ -0,0 +1,114 @@
+import { getMinMaxDates } from './DateFilter';
+
+describe('getMinMaxDates', () => {
+  it('should return provided min and max dates when no query object', () => {
+    const query = '';
+    const range = 'from';
+    const min = '2025-09-01';
+    const max = '2025-11-01';
```
andreievg, line 8: so the min and max in the "DateFilterDefinition" is a seed ?

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
```diff
@@ -86,22 +85,40 @@ const getDateFromUrl = (query: string, range: RangeOption | undefined) => {
...
+  const maxDate = DateUtils.getDateOrNull(max);
+
+  if (typeof query !== 'object' || !range) {
+    return {
+      minDate: minDate || undefined,
+      maxDate: maxDate || undefined,
+    };
+  }
+
   const { from, to } = query as RangeObject<string>;
```
andreievg, line 104: but why is it a string ? in definition ?

andreievg: in method definition ('string' | RangeObject)

client/packages/common/src/ui/components/inputs/Filters/dateFilter.test.tsx
lache-melvin: Yeah... well that's the widest range you want to allow. Then as the user inputs the range, we further constrain - you shouldn't allow a "to" that's earlier than the "from"

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
```diff
@@ -86,22 +85,40 @@ const getDateFromUrl = (query: string, range: RangeOption | undefined) => {
...
-      : (limitDate ?? undefined);
-  else
-    return from
-      ? (DateUtils.maxDate(DateUtils.getNaiveDate(from), limitDate) ??
-          undefined)
-      : (limitDate ?? undefined);
+  if (range === 'from') {
+    const toDate = DateUtils.getDateOrNull(to);
+    return {
+      minDate: minDate || undefined,
```
andreievg, line 109: Do you reckon we restrict anywhere, minData for 'from field' ?

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
lache-melvin: I thinkk because query comes from URL, and URL query string can be anything... tbh didn't think too deep about existing types was just trying to get the logic going 🙈

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
lache-melvin: Not sure what you mean? Only would be restricted if you defined a min in the date filter definition?

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
andreievg: That's fine, thanks

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
andreievg: Oki, i think we need to review this whole thing at some point lol

lache-melvin: agree... along with other date issues 🥲

andreievg (approved):
Looks great, and thank you for the tests !

There are type issues and requirement issues that I think caused it all. Basically no need for an overall restriction of max/min and ideally 'from' and 'to' range should be mapped to either max or min (the type should just have one or the other)

lache-melvin:
Cool, thanks for thorough review! Currently using the overall max restriction to prevent filtering later than todays date, which I think makes sense (though we could use built in disableFuture prop in that case). Didn't check to see if there were further use cases, agree worth a review at some point 🙏

PR #7468, opened 2025-04-24
7421 breadcrumbs show uuids

### LINKED ISSUE #7421: The prescription will show invalid text when you try to prescribe items with no stocks
## What went wrong? 😲

https://github.com/user-attachments/assets/9efb9874-a7af-41df-a92d-10f94d6ff9c7

Image

## Expected behaviour 🤔

There should be valid prescripton number

## How to Reproduce 🔨

1. OMS dispensary store
2. Create Prescription
3. Add the items that does not have any stocks
4. Click on save
5. See the serial number has been turned into the invalid text (BUG)

## Your environment 🌱

- Open mSupply Version: v2.7.00-RC5

- Platform: SQlite3

- Database type:

lache-melvin:
Fixes #7421

# 👩🏻‍💻 What does this PR do?

Fixes the onAllocate check, was enabling the save button based on _attempted_ allocated number of packs, instead of actual allocated packs. Shouldn't be able to save if num packs is 0 (unless prescribed quan is set)

Also fixes - couldn't backspace the prescribed quan field

And fixes setBreadcrumb, wasn't setting the prescription number if you refreshed/opened on the new item page.

This area realllly needs refactoring, we should just be saving the item we are currently looking at, with a draft state we can easily run validation against before save/disable the save button until valid.

But currently prescription draft contains a draft line for every item/stock line touched (you can switch items, it doesn't remove the previous draft lines) - so it's "valid" to have invalid lines in the draft state! They get ignored in backend, but we should only enable save in valid states 😭 This makes it hard to determine valid states to enable/disable the save button - so we just enable via `isDirty` (semantically incorrectly - the form is dirty! Just not valid...)

Not a 2.7 change tho :/

# 🧪 Testing

- [ ] Add item to prescription with no stock available
- [ ] Try to allocate stock, but none available, so input resets to 0
- [ ] Save button stays disabled
- [ ] Enable editPrescribedQuantityOnPrescription store pref
- [ ] You _can_ save a 0 issue quantity, if you have prescribed quantity > 0
- [ ] If you edit prescribed quantity back to 0, the save button disables
- [ ] After save, you are redirected to item, breadcrumb shows invoice number and item name correctly
- [ ] Click new item, then refresh
- [ ] Invoice number shows in breadcrumb

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -185,7 +185,12 @@ export const PrescriptionLineEditForm: React.FC<
     // this method is also called onBlur... check that there actually has been a
     // change in quantity (to prevent triggering auto allocation if only focus
     // has moved)
-    if (inputUnitQuantity === issueUnitQuantity) return;
+    if (
+      quantityType === 'issue'
+        ? inputUnitQuantity === issueUnitQuantity
+        : inputUnitQuantity === prescribedQuantity
+    )
+      return;
 
     const quantity = inputUnitQuantity === undefined ? 0 : inputUnitQuantity;
     setIssueUnitQuantity(quantity);
@@ -204,9 +209,8 @@ export const PrescriptionLineEditForm: React.FC<
   };
 
   const handlePrescribedQuantityChange = (inputPrescribedQuantity?: number) => {
-    if (inputPrescribedQuantity == null) return;
-    setPrescribedQuantity(inputPrescribedQuantity);
-    handleIssueQuantityChange(inputPrescribedQuantity, 'prescribed');
+    setPrescribedQuantity(inputPrescribedQuantity ?? 0);
+    handleIssueQuantityChange(inputPrescribedQuantity ?? 0, 'prescribed');
   };
 
   const prescriptionLineWithNote = draftPrescriptionLines.find(l => !!l.note);
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -95,6 +95,7 @@ export const PrescriptionLineEditForm: React.FC<
   const { format } = useFormatNumber();
   const { rows: items } = usePrescription();
   const { store: { preferences } = {} } = useAuthContext();
+  preferences.editPrescribedQuantityOnPrescription = true;
 
   const [issueUnitQuantity, setIssueUnitQuantity] = useState(0);
   const [prescribedQuantity, setPrescribedQuantity] = useState<number | null>(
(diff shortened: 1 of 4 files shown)


roxy-dao (changes requested):
Thanks Laché

Tried tracking this down...
On a new prescription, I try to add an item with no batch, click save and it does nothing? There's no upsert call since isDirty is always false...
https://github.com/user-attachments/assets/89ea6cc4-2ed8-4239-8667-a99c5f0a76ea

So far I've gotten 1st time you click save isDirty is always false because prescribedQuantity on allocate is always null

lache-melvin:
Shouldn't allow unallocated without the prescribed quan pref

roxy-dao (approved):
Thanks for the changes Laché~

Can you create a carryover issue to handle this properly? And also have a question about whether the prescription should be automatically changed to picked since they're unallocated lines...

roxy-dao (changes requested):
Soz... trigger happy

client/packages/invoices/src/Prescriptions/LineEditView/hooks/useDraftPrescriptionLines.tsx
```diff
@@ -90,30 +85,15 @@ export const useDraftPrescriptionLines = (
...
-        );
-      } else {
-        // Commented out for now until placeholders are implemented for
-        // prescriptions
-        // rows.push(createStockOutPlaceholderRow(invoiceId, item.id));
-      }
+    const allStockLinesExpired = stockLines.every(
+      stockLine =>
+        stockLine.expiryDate && DateUtils.isExpired(stockLine.expiryDate)
+    );
```
lache-melvin, line 91: So, no longer any different logic in new vs picked. Agree we need to confirm if it should be picked if only a placeholder row, but... future problem 😁 Should now be able to set prescribed quantities for all expired stock lines

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -204,9 +209,8 @@ export const PrescriptionLineEditForm: React.FC<
   };
 
   const handlePrescribedQuantityChange = (inputPrescribedQuantity?: number) => {
-    if (inputPrescribedQuantity == null) return;
-    setPrescribedQuantity(inputPrescribedQuantity);
-    handleIssueQuantityChange(inputPrescribedQuantity, 'prescribed');
+    setPrescribedQuantity(inputPrescribedQuantity ?? 0);
+    handleIssueQuantityChange(inputPrescribedQuantity ?? 0, 'prescribed');
```
roxy-dao, line 213: Are we defaulting null / undefined to 0?

lache-melvin: I just copied what was going on for issue quan, bc that one works as expected 🤷

client/packages/invoices/src/Prescriptions/LineEditView/hooks/useDraftPrescriptionLines.tsx
roxy-dao: Also doesn't show in detail UI that that it's placeholder either...

client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
lache-melvin: Possibly should be nulled out, but I didn't know if that would have further type implications, wanted to move quick 👀

lache-melvin:
> Can you create a carryover issue to handle this properly? And also have a question about whether the prescription should be automatically changed to picked since they're unallocated lines...

#7472

Issue #7472, opened 2025-04-24
Prescription line edit: simplify logic


lache-melvin:
## The suggestion

Hi triage, just wanting to make a placeholder for this work, I think its super important. Ideally should be done in amongst move back from the line edit view....

- Managing state for all items, rather than just the item in view, adds way more complexity
- There's no validation on the save button (just some very hacky places where we set know we've done something invalid, so we set "form not touched, nothing to save"/`isDirty = false`)
- Reusing allocation logic from outbound is adding complexity, use cases that are hard to be aware of
  - Probably if allocation logic was broken down into pieces, we could share elements, and leave specific logic in each area
- Concept of placeholder line/unallocated stock is different than for other invoices
  - You can have picked prescription with placeholder line, for prescribed quantity (though maybe not really anymore? honestly can't tell)
  - When should a prescription actually be in New vs Picked?
- Batch mutations are hard to reason about (one example, `isCreated` seems to be creating some strange behaviour [per this PR comment](https://github.com/msupply-foundation/open-msupply/pull/7495#pullrequestreview-2797772473))
- We need comprehensive tests for this stuff, its way too brittle and edge casey
- There are item-level things stored on prescription, which on backend we actually store against a line (there are potentially many lines to an item)
  - This adds a lot of complexity to frontend logic, having to know about this
  - I think we should go for API that supports this - e.g. can just query for item directions for this item on a prescription, rather than -> query all lines for item -> loop through -> find one with a directions (and on edit, do you update one or all lines etc...)

I did some RnD a while ago but didn't finish: https://github.com/msupply-foundation/open-msupply/pull/6817

lache-melvin:
To review when cleaning up in cooldown: #3252 and all linked issues..

lache-melvin:
Carry over, things to add to write up:

- https://github.com/msupply-foundation/open-msupply/pull/7712#discussion_r2090191029
- https://github.com/msupply-foundation/open-msupply/pull/7712#discussion_r2090365861
- Should prescriptions use pricing?
- Ok & next, when should be disabled
- Scanner behaviour
- Placeholder line - how to edit if allocating manually?

Also to consider, the edit mutation - remove invoice lines not in payload? Map by stock line id to be more robust? https://github.com/msupply-foundation/open-msupply/pull/7712#discussion_r2090541751

lache-melvin:
@jmbrunskill  assigning you to this one for the prescription refactor PR, unless you have another issue assigned for that already?

[...]

PR #7495, opened 2025-04-27
7480 prescription breadcrumb shows uuid on new item

### LINKED ISSUE #7480: Prescription item id shows in breadcrumb
## What went wrong? 😲

Found while testing: https://github.com/msupply-foundation/open-msupply/issues/7421#issuecomment-2829504631

Still the Prescription shows invalid text

https://github.com/user-attachments/assets/260bb58f-b417-4004-b4c2-58eddcb48ec4

Image

## Expected behaviour 🤔

There should be no invalid text

## How to Reproduce 🔨

1. OMS dispensary store
2. Create Prescription
3. Add items with quantaties
4. reopen the prescription
5. Now set the items to 0
6. Save the item
7. Click on `New item`
8. See invalid text (BUG)

## Your environment 🌱

- Open mSupply Version: v2.7.00-RC7

- Platform:

- Database type:

lache-melvin:
Fixes #7480

# 👩🏻‍💻 What does this PR do?

Triggers the breadcrumb useEffect when the URL item id changes, so it runs when we switch to `new` item.

Roxy commented in the issue, about one of the items remaining even after setting issue quantity to 0. This appears to only happen when the item has more than one stock line available 🫠 

Seems like pretty unlikely use case, and I think any fix would be yet another hack. Recommending we leave this one for now and consider the case when we clean up prescriptions properly?

# 🧪 Testing

- [ ] Add an item to a prescription, an item with stock available
- [ ] Issue some amount, save
- [ ] For that item in the prescription, expand the batches dropdown and clear the issued stock to 0
- [ ] Save
- [ ] The item should be deleted, you are redirected to new item view
- [ ] Clicking the `new item` in the left hand list does not change the breadcrumbs (still see prescription serial number)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/LineEditView.tsx
@@ -79,7 +79,7 @@ export const PrescriptionLineEditView = () => {
       1: data?.invoiceNumber.toString() ?? '',
       2: currentItem?.name || '',
     });
-  }, [currentItem, data?.invoiceNumber]);
+  }, [currentItem, data?.invoiceNumber, itemId]);
 
   useConfirmOnLeaving(
     'prescription-line-edit',


noel-yeldos (commented):
Thanks @lache-melvin! I also noticed the issue that @roxy-dao had mentioned where the item would not be deleted after setting quantity to 0. 

I had to change line 86 in the below file to `numberOfPacks > 0` instead of `numberOfPacks >= 0` to test. Doing that deleted the item on save when quantity is set to 0 and navigated me to new line where the id was not being displayed (as expected).

I also wonder if `updatePrescriptionLines` needs to be changed as well to be `numberOfPacks > 0`

https://github.com/msupply-foundation/open-msupply/blob/f763ce3308fc042370221a2486d17d15e61e438f/client/packages/invoices/src/Prescriptions/api/hooks/usePrescriptionLines.ts#L72-L110

Do you think it's worth making that change or leave it for the refactor as you suggested?

lache-melvin:
Thanks Noel - I'm nervous to change any of this logic so close to release, it's a real rabbithole, and I suspect it might have implications for lines that have `prescribedQuantity` but no `numberOfPacks`. Might be okay, but we don't really have time to test, and I think the current issue of a line potentially sticking around is both low likelihood and low impact for the time being.

I have linked your comment in this refactor issue though, I think your change is correct and worth moving towards! #7472

noel-yeldos (approved):
I hear ya! Approved 🚀  

(just noting again in case we need to circle back that I was only able to observe expected functionality by making changes to `usePrescriptionLines` as above)

PR #7503, opened 2025-04-28
Prevent simultaneous installs by removing application id suffix

### LINKED ISSUE #7496: v2.7.0RC7 doesn't upgrade previous installations but want to install a new one on Android
## What went wrong? 😲
I believe that I could upgrade before but now it is installing a new version instead of upgrading so I have 2 versions of omSupply. The video below is longer that it should be. I forgot to stop it :)

https://drive.google.com/file/d/1d604mnHufhu7Y5SZEoG4JDbKUqF5iYNe/view?usp=sharing

Image

I am using the universal apk

## Expected behaviour 🤔
New version should upgrade the old instead of creating one
## How to Reproduce 🔨

1. Have 2.6.1 on Android 
2. Try to upgrade to 2.7.0RC7
3. See error

## Your environment 🌱

- Open mSupply Version: 2.7.0RC7
- Legacy mSupply Central Server Version: NA

- Platform: Android

- Database type: SQLite

aimee-mcneil-melville:
Fixes #7496 

# 👩🏻‍💻 What does this PR do?

Ensures when a new version of the app is installed, the existing app is replaced, regardless of the build type. 

To achieve this, I removed the applicationIdSuffix from the arm64 and universal build flavors. This ensures that all app versions use the same applicationId as defined in the defaultConfig of the build.gradle file

When installing an apk on a device, the applicationId determines if the app is 'new' or not. If the applicationId matches exactly, the app will update. If the applicationId does not match exactly, a new version will be installed. As the versions had different suffixes, android could have each of these apps installed simultaneously

Apks with the same version number eg v2.7.0-RC7-arm64 and v2.7.0-RC7-universal can overwrite each other, however as standard behaviour v2.6.1 (being a lower version number) cannot overwrite a v2.7.0

Tested with 2.6.1 in the video here

https://github.com/user-attachments/assets/7c0ee25b-3623-4e22-bb28-d565b8c87096

I retained the versionNameSuffix - this does not affect the install behaviour but it does allow the version to be seen in Settings -> Apps -> Open mSupply if needed for support & determining which one is installed

Screenshot 2025-04-28 at 1 37 37 PM Medium
Screenshot 2025-04-28 at 1 37 25 PM Medium

# 🧪 Testing

- [ ] Have 2.6.1 installed on Android
- [ ] Download apks with this fix here: https://github.com/msupply-foundation/open-msupply/actions/runs/14697956009
- [ ] Install one of the new apks (universal or arm64). Note the prompt says 'update' rather than 'install'
- [ ] See only one instance of Open mSupply installed
- [ ] Repeat with the the other new apk - still have only one app version installed

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/android/app/build.gradle
+++ b/client/packages/android/app/build.gradle
@@ -65,7 +65,6 @@ android {
     productFlavors {
         arm64 {
             dimension "abi"
-            applicationIdSuffix ".arm64"
             versionNameSuffix "-arm64"
             ndk {
                 abiFilters "arm64-v8a"
@@ -73,7 +72,6 @@ android {
         }
         universal {
             dimension "abi"
-            applicationIdSuffix ".universal"
             versionNameSuffix "-universal"
             ndk {
                 abiFilters "armeabi-v7a", "arm64-v8a"


andreievg (approved):
Looks great, thank you

PR #7510, opened 2025-04-28
Remove db triggers in version 2.0 instead of version 2.2

### LINKED ISSUE #7506: Sync freezes after upgrading from 1.7
We previously had this issue, we thought it was resolved, but I think there was misunderstanding of actual issue or an assumption about the state of other sites with deployed omSupply

## What went wrong? 😲

Changelog points to a record that doesn't exist, after upgrade from 1.7 to 2.6

Image

## Expected behaviour 🤔

Sync works after upgrade

## Investigation

There is a migration between 1.6.2 and 1.7.0 that corrupts database, we tried to fix for name_link_id: [pr](https://github.com/msupply-foundation/open-msupply/pull/2946), [commit](https://github.com/msupply-foundation/open-msupply/commit/719a05526a637eec31fb01625be3b16725e8537c),  [sqlite forum](https://sqlite.org/forum/forumpost/96896f825d). Kudos @jmbrunskill for digging this up when we originally had the issue, and for identifying that we didn't apply this to item_link_id.

However this doesn't seem to cause an issue with sync, although invoice_line becomes invisible, and I think they are not queryable. 

For some reason when updating from 1.7 and 2.6 (this was noticed earlier too, 2.3.3, so probably in migrations before that), we end up with records waiting to be synced to central server, and since we can't query for the invoice_lines, sync is stuck, see reproduction steps and associated data

# Solutions 

1. Ignore changelogs that don't point to records 
2. Ability to easily re-initialise
3. Find why we are re-syncing those records after upgrading

## How to Reproduce 🔨

Start with mobile 1.6.2, initialise, notice outbound shipments and invoice lines

Upgrade to 1.7.0, notice outbound shipments don't have invoice lines, but still able to sync new lines.

Upgrade to 2.6.2, notice that there are lots of records in sync queue and sync is stuck.

Data files where you can check this: https://drive.google.com/drive/u/1/folders/1yfr9iIdC5KnrJVkMuO61LUMdQWmLd0lf

jmbrunskill:
Background. Sites that have been updated to version 1.7 were affected by a bug in SQLite that caused data corruption. 
This is most obviously seen when trying to upgrade past 1.7 as during the upgrade database triggers caused the change log to be updated for all existing invoices, this means that the system tries to re-sync records that are no longer valid.

Fixes #7506

# 👩🏻‍💻 What does this PR do?

Technically this isn't a real fix for that database corruption errors, however I think it should allow the upgrade to continue ok. It works by removing the database triggers on the invoice_line (and other tables) this means the updates to these tables don't get changelog records and are only synced to central server when they're actually updated by users.

It's safe to remove the triggers earlier as we know they're upgrading to a version which doesn't rely on db triggers, and will also reduce the sync traffic required for the upgrade.

Also interested to see if there's a way to fix the database corruption, might try to look at that, otherwise we'll need to schedule in some work for a process to re-sync and fix corrupted records....

# 🧪 Testing

- [ ] Use a database was was upgraded to 1.7 using the broken code.
- [ ] Try to update to version 2.6.4 (this PR)
- [ ] Check you can sync successfully to your 4d server.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/v2_00_00/mod.rs
+++ b/server/repository/src/migrations/v2_00_00/mod.rs
@@ -13,6 +13,7 @@ mod invoice_rename_tax;
 mod linked_shipment;
 mod name_deleted_datetime;
 mod pack_variant;
+mod remove_changelog_triggers;
 mod report_views;
 mod requisition_line_add_item_name;
 mod returns;
@@ -30,6 +31,8 @@ impl Migration for V2_00_00 {
     }
 
     fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        // Remove changelog triggers (time travelling, as the code was actually introduced in 2.2.0, but we want to run it early in a patch to avoid syncing changes updated in this version invoice_line_tax for example)
+        remove_changelog_triggers::migrate(connection)?;
         add_source_site_id::migrate(connection)?;
         central_omsupply::migrate(connection)?;
         assets::migrate_assets(connection)?;
--- a/server/repository/src/migrations/v2_00_00/remove_changelog_triggers.rs
+++ b/server/repository/src/migrations/v2_00_00/remove_changelog_triggers.rs
@@ -1,0 +1,141 @@
+use crate::migrations::*;
+
+#[cfg(not(feature = "postgres"))]
+pub(crate) fn migrate(connection: &StorageConnection) -> anyhow::Result<()> {
+    sql!(
+        connection,
+        r#"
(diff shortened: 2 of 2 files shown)


andreievg (approved):
Looks good to me, thanks @jmbrunskill

PR #7522, opened 2025-04-29
7145 prefs refactor

### LINKED ISSUE #7145: Prefs Implementation Refactor
## The suggestion

Context: https://github.com/msupply-foundation/open-msupply/pull/7087#issuecomment-2767881976

Few issues/questions to resolve with current prefs setup:

- Upsert should be type safe - accept input with each pref optional (store prefs probably an optional array, add for each store?)
  - Each pref implements its own upsert logic?
- Frontend therefore sends stricter types, so needs to know about them... may need to move to defining each new pref in the edit UI, rather than an array?
  - At this point, review amount of code required to add a new pref (not just quantity but developer overhead, things to remember to do rather than compiler driven)
  - Existing set up requires little code addition, but a little less type safety, review trade offs
- Shouldn't load all prefs to get one - return methods to the GQL layer so only the required prefs are loaded
- `load()` should only be for querying, and can probably be helper rather than trait method
  - Add helper methods to load different kinds of prefs - store, global... (merges with defaults etc.)
- Not refactor work, but needs to be considered, what do the different UIs look like (global prefs, prefs by store, stores by pref, machine prefs, user prefs...)

We want to be as confident in our types and our data as we can be, and seeing as we know all the prefs, we can strictly type them, only db layer is serialised. 

However, if a little more generic = way less code, I think this has a lot of value for ongoing maintenance, as well as speed when adding new prefs. Keen to experiment here and for different perspectives.

At least a couple days of experimenting and review?

lache-melvin:
Fixes #7145

# 👩🏻‍💻 What does this PR do?

Refactors the preference implementation. Main goals:

- Ensure strict typing the full way through, particularly on upsert
- `load` needs to support different contexts (global/user/store/machine)
- Remove the use of JSON forms for the edit components - more complexity than required
- GraphQL resolver should query each pref separately, rather than all prefs queried when only selecting one

Here's a diff for adding a new global pref 🚀  https://github.com/msupply-foundation/open-msupply/pull/7536/files

# 🧪 Testing

- [ ] On OMS Central Server
- [ ] Manage > Global Preferences
- [ ] Can still set the show contact tracing setting

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/intl/locales/ar/common.json
```diff
@@ -2195,7 +2195,6 @@
   "placeholder.search-by-name": "البحث حسب الاسم",
   "placeholder.search-by-name-or-code": "البحث حسب الاسم أو الرمز",
   "placeholder.search-by-notes": "البحث حسب الملاحظات",
-  "preferences": "التفضيلات",
```
lache-melvin, line 2198: Cleared this in favour of "Global Preferences" - store prefs will be managed separately, from the facilities list view

client/packages/system/src/Manage/Preferences/EditPage/EditPreference.tsx
```diff
@@ -1,58 +1,66 @@
...
 
-  const { data: prefs, isLoading } = usePreferences();
-  const { mutateAsync: update } = useEditPreference(selected.key);
-
-  if (isLoading) {
-    return <BasicSpinner />;
-  }
+  const getRenderer = () => {
+    switch (valueType) {
+      case PreferenceValueNodeType.Boolean:
```
lache-melvin, line 32: This is quite similar to how we went with asset properties - we have a known set of preference value types, and how to render them. Currently, those are scalar types (i.e. bool) but could also include custom ones (e.g. a special renderer for managing custom translation overrides

client/packages/system/src/Manage/Preferences/EditPage/getPrefKey.tsx
```diff
@@ -0,0 +1,30 @@
+import { PreferencesNode } from '@common/types';
+
+export function getPrefKey(key: string): ClientPrefKey | undefined {
+  // Typescript error if any keys are missing
+  const noMissingKeys: EnsureNoMissingKeys = true;
```
lache-melvin, line 5: This was important to me - there is a mapping below between the pref identifier ("ServerKey") and what the object key is for that pref when served to the frontend. When we add a new pref, I wanted typescript to error until we have handled the new pref type

client/packages/system/src/Manage/Preferences/api/useAvailablePreferences.ts
lache-melvin, line 1: just renamed this to "useGlobalPrefList", think its clearer naming

server/graphql/types/src/types/mod.rs
```diff
@@ -155,3 +155,11 @@ impl IdResponse {
         &self.0
     }
 }
+
+pub struct OkResponse;
+#[Object]
+impl OkResponse {
+    pub async fn ok(&self) -> bool {
+        true
```
lache-melvin, line 163: Lol I always want a response object like this, realise it doesn't exist, so end up returning the node... but there wasn't a relevant pref node to return from the upsert response, so just implemented this. We don't use the mutation responses in the frontend anyway, I think this is okay?

server/service/src/preference/mod.rs
```diff
@@ -1,33 +1,43 @@
...
+    fn get_preference_registry(&self) -> PreferenceRegistry {
+        get_preference_registry()
     }
 
-    fn get_preference_descriptions(&self) -> Vec<Box<dyn Preference<Value = bool>>> {
-        get_preference_descriptions()
+    fn get_preference_descriptions(&self, pref_type: PreferenceType) -> Vec<PreferenceDescription> {
+        let PreferenceRegistry {
+            show_contact_tracing,
+        } = &self.get_preference_registry();
```
lache-melvin, line 22: Destructuring here, so should get an error when a new pref is added, to prompt to append to the all prefs below

server/service/src/preference/preferences/README.md
lache-melvin, line 1: Removed readme, it was out of date, any context is currently available in code comments... I'll add a readme back after creating new pref if I think necessary 🙏

server/service/src/preference/preferences/types.rs
lache-melvin, line 1: moved to new folder

server/service/src/preference/query_preference.rs
```diff
@@ -0,0 +1,96 @@
+use repository::{
+    EqualFilter, PreferenceFilter, PreferenceRepository, PreferenceRow, RepositoryError,
+    StorageConnection,
+};
+
+pub fn query_global(
```
lache-melvin, line 6: separate helpers for handling global prefs, store prefs (would extend when adding other pref types)

These are used as defaults (see the types.rs) but can also implement your own if the pref query was custom for whatever reason

server/service/src/preference/types.rs
```diff
@@ -0,0 +1,215 @@
...
+        Ok(pref)
+    }
+
+    fn default_value(&self) -> Self::Value {
+        Self::Value::default()
+    }
+
+    fn deserialize(&self, data: &str) -> Result<Self::Value, serde_json::Error> {
+        serde_json::from_str::<Self::Value>(data)
+    }
```
lache-melvin, line 79: Exposing methods like deserialize and default as methods on the pref, so we can define custom ones when needed

server/service/src/preference/upsert.rs
```diff
@@ -1,38 +1,37 @@
...
-) -> Result<PreferenceRow, RepositoryError> {
-    // TODO: validation here (i.e. can't set store pref where it is global only?)
-    // more validation would be needed once we allow remote stores to set preferences
+    UpsertPreferences {
+        show_contact_tracing: show_contact_tracing_input,
+    }: UpsertPreferences,
+) -> Result<(), UpsertPreferenceError> {
+    let PreferenceRegistry {
+        show_contact_tracing,
+    } = get_preference_registry();
```
lache-melvin, line 20: Again, destructuring prompts compiler warning, so you remember to handle the new upsert type

server/service/src/preference/upsert_helpers.rs
```diff
@@ -0,0 +1,54 @@
+use repository::{PreferenceRow, PreferenceRowRepository, StorageConnection};
+
+use crate::sync::CentralServerConfig;
+
+use super::UpsertPreferenceError;
+
+pub fn upsert_global(
```
lache-melvin, line 7: Like query, these are the default upsert helpers. If need be, can implement your own upsert, e.g. if you need more validation logic

server/service/src/preference/preferences/show_contact_tracing.rs
```diff
@@ -8,4 +8,12 @@ impl Preference for ShowContactTracing {
     fn key(&self) -> &'static str {
         "show_contact_tracing"
```
lache-melvin, line 9: I did consider, should key be an enum? Would have to create a migration for each new pref, but would maybe feel a bit cleaner than string, especially when referenced on frontend?

server/service/src/preference/preferences/mod.rs
```diff
@@ -1,33 +1,12 @@
...
-) -> Result<Preferences, RepositoryError> {
-    let connection = &ctx.connection;
-
-    let prefs = Preferences {
-        show_contact_tracing: ShowContactTracing.load(connection, store_id)?,
-    };
-
-    Ok(prefs)
+pub struct PreferenceRegistry {
+    pub show_contact_tracing: ShowContactTracing,
```
lache-melvin, line 5: note to self, add a comment here saying to add new prefs to this list lol

client/packages/system/src/Manage/Preferences/EditPage/getPrefKey.tsx
CarlosNZ: 100% agree, but I think I'm starting to see an issue with how we're doing this -- namely, what is the point of a query to "getAvailablePreferences", if all prefs are hard-coded in the front-end anyway?

The problem is we wanted to have a system where we could define new prefs in the back end and the front-end would display them all dynamically, but we're not really doing that any more -- so we don't really need a query to fetch them.

I think that, along with my other comment about fetching and updating one at a time the shape of the GraphQL endpoints could do with a bit of a rethink.

I have a few thoughts, but I won't go into it here. I'll try and get a few ideas down and we can have a chat about it later?

CarlosNZ (commented):
Looking good. I have a few comments/questions, including some thoughts about the general shape of the endpoints -- see inline comments.

Maybe this is good for now and we can continue to refine it, but it'd be good to have a bit of a chat about it before going further.

lache-melvin:
Thanks for feedback @CarlosNZ - have applied per our conversation yesterday 🚀

client/packages/system/src/Manage/Preferences/EditPage/EditPreference.tsx
lache-melvin: I think would only be defined as a plugin if the pref was also plugged in - otherwise I was just imagining another folder of components here?

client/packages/system/src/Manage/Preferences/EditPage/getPrefKey.tsx
lache-melvin: Now handled in backend, with enum key types 🙏

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
```diff
@@ -1,20 +1,37 @@
...
+  const { mutateAsync } = useEditPreference();
+
+  const update = async (input: Partial<UpsertPreferencesInput>) => {
+    try {
+      await mutateAsync(input);
+    } catch (err) {
+      console.error('Error updating preferences:', err);
+      error(t('error.something-wrong'))();
+    }
+  };
```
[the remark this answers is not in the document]
lache-melvin: [cd6bf11](https://github.com/msupply-foundation/open-msupply/pull/7522/commits/cd6bf1112f1d765e2cd7cedd4f73c63a776efd01)

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
```diff
@@ -1,20 +1,37 @@
...
 
 export const EditPreferencesPage = () => {
   const t = useTranslation();
   const { key } = useParams();
   const { setCustomBreadcrumbs } = useBreadcrumbs();
+  const { error } = useNotification();
 
-  const { data } = useAvailablePreferences();
+  const { data: prefOptions } = useGlobalPrefList();
+  const { data: configuredPrefs } = usePreferences();
```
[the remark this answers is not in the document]
lache-melvin: Per yesterday, added value to the PreferenceDescription, so usePreference() is only for when querying a specific pref: [85a22aa](https://github.com/msupply-foundation/open-msupply/pull/7522/commits/85a22aa2a63cd777c5987129ce56457e3f19a331)

server/graphql/types/src/types/preferences.rs
```diff
@@ -1,67 +1,83 @@
...
+        PreferenceValueNodeType::from_domain(&self.pref.value_type)
     }
 }
 
-pub struct PreferenceNode {
-    pub preference: PreferenceRow,
+#[derive(Enum, Copy, Clone, Debug, Eq, PartialEq)]
+pub enum PreferenceNodeType {
+    Global,
+    Store,
```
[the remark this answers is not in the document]
lache-melvin: I think we'll add it once we actually support user prefs? We'll need to add a migration to the prefs table to add optional user id, etc...

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
```diff
@@ -1,36 +1,37 @@
...
-      {selectedPref && (
-        <EditPreference key={selectedPref.key} selected={selectedPref} />
-      )}
-    </LineEditBase>
+    <Box display="flex" justifyContent="center" width="100%" marginTop={2}>
+      <Box width="600px">
+        {preferences.map(pref => (
+          <InputWithLabelRow
+            key={pref.key}
+            label={t(`preference.${pref.key}` as LocaleKey)}
```
fergie-nz, line 26: I wonder if you could use `translateDynamicKey` from IntlUtils.ts with manual fallback string?

This way missing translation key will be a bit more nice than 'preferences.key' showing in UI

client/packages/system/src/Manage/Preferences/api/useEditPreference.ts
```diff
@@ -1,27 +1,47 @@
...
 
-export const useEditPreference = (key: string) => {
+export const useEditPreferences = (prefType: PreferenceNodeType) => {
+  const t = useTranslation();
+  const { error } = useNotification();
+
+  const { data } = useAdminPrefsList(prefType);
+  const { mutateAsync } = useUpsertPref();
+
+  // Probably worth debouncing at some stage :)
```
CarlosNZ, line 21: >  // Probably worth debouncing at some stage :)

Indeed -- definitely once we get a string input pref. Maybe update comment to "Please add debouncing once string prefs are implemented"?

client/packages/common/src/authentication/api/hooks/usePreference.ts
```diff
@@ -0,0 +1,40 @@
+import { gql } from 'graphql-tag';
+import {
+  PreferenceKey,
+  PreferencesNode,
+  useAuthContext,
+  useQuery,
+} from '@openmsupply-client/common';
+import { useAuthApi } from './useAuthApi';
+
+export const usePreference = <T extends PreferenceKey>(pref: T) => {
```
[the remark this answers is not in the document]
CarlosNZ: Also -- what's the purpose of the generic `T`? Can it not just be input type `(pref: PreferenceKey)`?

server/graphql/types/src/types/mod.rs
CarlosNZ: Is that cos you have to return something from GraphQL?

CarlosNZ (approved):
Very nice, thanks heaps!

I can see I'm going to have to re-do my new pref PR now 😆, but that's cool, as it all seems pretty simple now that I understand it :) 

I made a few very minor comments, most don't really need action unless you particularly feel like it

client/packages/common/src/utils/index.ts
```diff
@@ -28,6 +28,8 @@ import isEqual from 'lodash/isEqual';
 import isEqualWith from 'lodash/isEqualWith';
 import isObject from 'lodash/isObject';
 import isString from 'lodash/isString';
+import isBoolean from 'lodash/isBoolean';
+import isNumber from 'lodash/isNumber';
```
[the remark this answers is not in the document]
lache-melvin: Oh wow, bundle size diff is less than that including all the other code changes, so yeah hoping ok, but good to note!

Am exporting through the common package, so replacing lodash methods with our own like that shouldn't be a big job 😁

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
lache-melvin: Hm... I like the idea, but I don't really think prefs without a related translation is a valid case we'd want to support. If you're adding a new pref, it should have a matching translation key. Kind of think the ugly preference.key is good to keep around, reminds you there is something left to do!

client/packages/common/src/authentication/api/hooks/usePreference.ts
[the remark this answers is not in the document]
lache-melvin: This is so we can type the return value from the hook - based on which key you pass in, those are the only fields you get back on the result 👍 

Agree about the array, figured we'd get the base use case in for now 🙏

client/packages/system/src/Manage/Preferences/EditPage/EditPreferencesPage.tsx
fergie-nz: > Hm... I like the idea, but I don't really think prefs without a related translation is a valid case we'd want to support. If you're adding a new pref, it should have a matching translation key. Kind of think the ugly preference.key is good to keep around, reminds you there is something left to do!

Yeah nice good point!

client/packages/system/src/Manage/Preferences/EditPage/EditPreference.tsx
```diff
@@ -1,58 +1,52 @@
...
-  }
-
-  const prefKey = getPrefKey(selected.key);
-  if (!prefs || !prefKey) {
-    return <NothingHere />;
-  }
+  switch (preference.valueType) {
+    case PreferenceValueNodeType.Boolean:
+      if (!isBoolean(value)) {
+        return t('error.something-wrong');
```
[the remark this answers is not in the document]
lache-melvin: Hm great point, I'm not sure - I think if there is an import like that, parsing/coercing should happen at that step? Value not matching type at this point would really only happen if a bad value came from the DB. I suppose in that case, defaulting wouldn't be a terrible idea, and hope that it re-saves correctly, but seems worth flagging that something got out of whack? Especially if it's getting loaded wrong, so every time you open the form, you just see the default again...

Feels like a case we can handle when we run into it 😁

client/packages/system/src/Manage/Preferences/api/operations.graphql
```diff
@@ -12,11 +12,11 @@ query preferences($storeId: String!) {
...
 }
 
-mutation upsertPreference($storeId: String!, $input: UpsertPreferenceInput!) {
+mutation upsertPreferences($storeId: String!, $input: UpsertPreferencesInput!) {
   centralServer {
     preferences {
-      upsertPreference(storeId: $storeId, input: $input) {
-        id
+      upsertPreferences(storeId: $storeId, input: $input) {
+        ok
```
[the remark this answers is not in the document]
lache-melvin: Yeah, need to return something, there's not really a related id to the upsert of potentially many prefs? 🤷‍♀️

client/packages/system/src/Manage/Preferences/api/useEditPreference.ts
lache-melvin: ah yeah nice shout

lache-melvin:
Thanks @CarlosNZ - I did rebase your PR earlier, so hopefully it's okay? There are a couple more commits ahead in this branch now though 😅

PR #7534, opened 2025-04-29
7379 add donor tracking pref

### LINKED ISSUE #7379: Donor: Preference and extra columns
## Tasks 📝
- [ ] Migration to add global preference `Allow tracking of received stock by donor`
- [ ] Add to service layer
- [ ] Return in graphQL
- [ ] Add to Preferences UI
- [ ] Migration to add `default_donor_id` to invoice and `donor_id` to invoice_line tables
- [ ] Sync default_donor_id to OG (note that invoice_id in OG is called donor_default_id

### Acceptance Criteria ✅
- [ ] Central users can view global preference
- [ ] Central users can edit global preference
- [ ] default_donor_id syncs to OG and back

### Moneyworks Jobcode 🧰
NIGGF:VAX

fergie-nz:
Fixes #7379

# 👩🏻‍💻 What does this PR do?

Adds migration for default_donor_id into invoice and invoice line rows

Adds translation to receive and push default_donor_id into mSupply

Adds default values (None) throughout service layer (to be used later)

The vast majority of files changed are adding default 'None' for default_donor_id to be used later.

# 🧪 Testing

|

- [ ] Run SQL to generate dummy global pref:
```
INSERT INTO preference (id, KEY, value)
VALUES (allow_tracking_of_received_stock_by_donor_global, allow_tracking_of_received_stock_by_donor, TRUE);
``` 
- [ ] Navigate to preferences on central server
- [ ] See that new preference is visible and can be toggled
- [ ] Add invoice with default_donor_id - and see this is synced up and down from legacy records

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin:
Ah sorry @fergie-nz - #7522 adds some changes to how we implement global prefs. I'm happy to update whenever that one is approved 🙏

fergie-nz:
> Ah sorry @fergie-nz - #7522 adds some changes to how we implement global prefs. I'm happy to update whenever that one is approved 🙏

No problem, thanks Laché! I'll rebase from your changes

fergie-nz:
I am waiting for [this pr](https://github.com/msupply-foundation/open-msupply/pull/7536/files) to be merged to avoid conflicts 

(Effectively adding the same pattern)

lache-melvin:
Hi @fergie-nz - upstream branches merged, you should be good to go, sing out if any funky conflicts 🙏

server/repository/src/db_diesel/invoice_line_row.rs
```diff
@@ -97,6 +98,7 @@ pub struct InvoiceLineRow {
     pub foreign_currency_price_before_tax: Option<f64>,
     pub item_variant_id: Option<String>,
     pub linked_invoice_id: Option<String>,
+    pub donor_default_id: Option<String>,
```
fergie-nz, line 101: Have left the same naming convention as per OG. I thought changing to 'donor_id' would add confusion unnecessarily

client/packages/common/src/authentication/api/operations.graphql
```diff
@@ -148,6 +148,13 @@ query permissions($storeId: String!) {
   }
 }
 
+query preferences($storeId: String!) {
+  preferences(storeId: $storeId) {
+    showContactTracing
+    allowTrackingOfReceivedStockByDonor
+  }
+}
```
lache-melvin, line 156: Hi, this query shouldn't need to be here, can be removed - usePreference hook will accept the pref you are trying to query for, and build graphql query dynamically, usage like this:

https://github.com/msupply-foundation/open-msupply/blob/69e612a5b9269a704ef5a61cf23d5ed438db8e14/client/packages/system/src/Patient/PatientView/AddButton.tsx#L34

server/service/src/preference/mod.rs
```diff
@@ -26,11 +26,14 @@ pub trait PreferenceServiceTrait: Sync + Send {
...
             show_contact_tracing,
             display_population_based_forecasting,
+            allow_tracking_of_received_stock_by_donor,
         } = self.get_preference_provider();
 
         let all_prefs_descriptions = vec![
             // Add each pref here
             show_contact_tracing.as_description(connection, store_id.clone())?,
+            allow_tracking_of_received_stock_by_donor
+                .as_description(connection, store_id.clone())?,
```
lache-melvin, line 36: Apologies, this section changed again, with merge of store pref support 🫠  Hopefully ok to resolve in merge, hit me up if conflicts are funky 😁

server/service/src/sync/test/test_data/stock_line.rs
```diff
@@ -16,7 +16,6 @@ const ITEM_LINE_1: (&str, &str) = (
       "barcodeID": "",
       "batch": "stocktake_1",
       "cost_price": 5,
-      "donor_id": "",
```
lache-melvin, line 19: OG item_line does have donor_id - we should leave this here

server/service/src/sync/translations/invoice_line.rs
```diff
@@ -90,6 +90,8 @@ pub struct LegacyTransLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     #[serde(rename = "linked_transact_id")]
     pub linked_invoice_id: Option<String>,
+    #[serde(deserialize_with = "empty_str_as_option_string")]
+    pub donor_default_id: Option<String>,
```
lache-melvin, line 94: It's called "donor_id" in OG, if we want to rename to include `default_`, you'll nee the serde rename, like the field above

server/repository/src/migrations/v2_08_00/add_donor_id_to_invoice_and_invoice_lines.rs
```diff
@@ -0,0 +1,27 @@
...
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "add_donor_id_to_invoice_and_invoice_lines"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                ALTER TABLE invoice_line ADD COLUMN donor_default_id TEXT;
```
lache-melvin, line 14: I don't fully understand context of these columns - but it seems like invoice has a default donor ID, but then each line should have donor_id, specified ID for that line rather than the default for the whole invoice?

Also, wondering why `donor_default` and not `default_donor`? second (as per issue) reads a bit nicer to me?

server/service/src/processors/transfer/invoice/common.rs
```diff
@@ -84,6 +85,7 @@ pub(crate) fn generate_inbound_lines(
                     stock_line_id: None,
                     location_id: None,
                     inventory_adjustment_reason_id: None,
+                    donor_default_id: None,
```
lache-melvin, line 88: Imagine would be easiest just to take the donor id destructured above and pass it through here?

lache-melvin (changes requested):
Legacy field name needs to match what it is in OG - did you try sync to OG? Values aren't arriving in OG, reinitialise, value doesn't come back...

server/repository/src/migrations/v2_08_00/add_donor_id_to_invoice_and_invoice_lines.rs
lache-melvin: Yep confirmed with roxy, on invoice_line should just be donor_id 🙏

fergie-nz: > I don't fully understand context of these columns - but it seems like invoice has a default donor ID, but then each line should have donor_id, specified ID for that line rather than the default for the whole invoice?
> 
> Also, wondering why `donor_default` and not `default_donor`? second (as per issue) reads a bit nicer to me?

I agree, I prefer default_donor_id.  However I was reluctant to rename because OG uses 'donor_default_id' and I had found the renaming of fields to be confusing. 

What do you reckon?

fergie-nz: I've gone ahead with the rename to default_donor_id FYI :)

lache-melvin: Yeah I agree, rename it. Mindset being - OMS is the long term plan, we should do our best to make OMS make the most sense it can, even when it means a mapping from OG. Going forward, OMS will be standalone, and we'll no longer have to think about the renamings 🙏

server/service/src/sync/test/test_data/invoice_line.rs
```diff
@@ -188,7 +189,8 @@ const TRANS_LINE_2: (&str, &str) = (
         "user_8_ID": "",
         "vaccine_vial_monitor_status_ID": "",
         "volume_per_pack": 0,
-        "om_item_variant_id": ""
+        "om_item_variant_id": "",
+        "donor_id": ""
```
lache-melvin, line 193: For good measure... could populate one of these donor_id fields, check that it translates to Some value, like you did for the invoice?

lache-melvin (approved):
Thanks for changes Ferg, looks great 🚀

PR #7555, opened 2025-05-01
Fixes postgres migration for removing `encounter_clinician_link_id_fkey` in postgres

jmbrunskill:
Fixes postgres migration for removing `encounter_clinician_link_id_fkey` in postgres

# 👩🏻‍💻 What does this PR do?

Fixes SQL 
Changes the migration fragment 2

Now it should never fail as it uses the `IF EXISTS` command. I think that's ok though

# 🧪 Testing

- [ ] Test upgrade from 2.7 to this branch (or 2.7.1) on postgres
- [ ] Check that `encounter_clinician_link_id_fkey` doesn't exist in the database.
- [ ] Test fresh install of 2.7, 
- [ ] Check that `encounter_clinician_link_id_fkey` doesn't exist in the database.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/v2_07_00/remove_encounter_clinician_constraint.rs
+++ b/server/repository/src/migrations/v2_07_00/remove_encounter_clinician_constraint.rs
@@ -16,7 +16,7 @@ impl MigrationFragment for Migrate {
             let result = sql!(
                 connection,
                 r#"
-                    ALTER TABLE encounter CONSTRAINT encounter_clinician_link_id_fkey;
+                    ALTER TABLE encounter DROP CONSTRAINT encounter_clinician_link_id_fkey;
                 "#
             );
             if result.is_err() {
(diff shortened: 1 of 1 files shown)


jmbrunskill:
> Thank you! Approving for the patch release, though I wonder if this issue argues we should throw the error again after warning?

I think my concern the first time we did this was that there might be a different constraint name for some reason, as we don't explicity name our constraints in postgres. I didn't want the system to fail to start in that situation, but I'm on the fence, seems like PG is consistent about how it names constraints, but my research I did lead me to believe it wasn't a guarentee...

lache-melvin:
> I wonder if this issue argues we should throw the error again after warning?

Like the IF EXISTS means it wouldn't throw when key doesn't exist, but at least we'd get the invalid SQL error 🤣

PR #7556, opened 2025-05-01
7548 sync vvm status

### LINKED ISSUE #7548: Sync VVM_Status Table

The VVM status table has been created and now needs to sync.

### Moneyworks Jobcode 🧰

NIGGF:VAX

zachariah-at-msupply:
Fixes #7548 

# 👩🏻‍💻 What does this PR do?

Sets up the translations for the vvm_table so it can by synchronised.

Hope your day is going well :)

# 🧪 Testing

- [ ] In mSupply, create a new vaccine vial monitoring (VVM) status (you can use these instructions if you want: https://docs.msupply.org.nz/cold_chain_equipment:configure?s[]=vvm&s[]=status)
- [ ] In OMS, attempt to sync.
- [ ] Check that the new VVM status you created is present in OMS (you might have to examine OMS' database table called vvm_status, because I don't know where you can see this in the user interface).
- [ ] In mSupply delete the new VVM status you made (you can use these instructions if you want: https://docs.msupply.org.nz/cold_chain_equipment:configure?s[]=vvm&s[]=status)
- [ ] In OMS, attempt to sync.
- [ ] Check that the new VVM status has been deleted (again, you might have to examine OMS' database directly).

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/test/test_data/vvm_status.rs
```diff
@@ -0,0 +1,70 @@
...
+const VVM_STATUS_1: (&str, &str) = (
+    "VVM_STATUS_1",
+    r#"{
+        "ID":"VVM_STATUS_1",
+        "code":"2",
+        "description":"TEST DESCRIPTION 1",
+        "is_active":true,
+        "level":1,
+        "unusable":false,
+        "option_id":"TEST OPTION ID 1"
```
jmbrunskill, line 16: Lets leave option_id out for now...

Ideally it should be linked to the `reason_option` table I think, and should really have a foreign key connection in the database. I think we can fill that in later when we have a real example?
```suggestion
        "option_id":""
```

server/service/src/sync/translations/vvm_status.rs
```diff
@@ -0,0 +1,95 @@
...
+#[allow(non_snake_case)]
+#[derive(Deserialize)]
+pub struct LegacyVVMStatusRow {
+    ID: String,
+    description: String,
+    code: String,
+    level: i32,
+    is_active: bool,
+    unusable: bool,
+    option_id: String,
```
jmbrunskill, line 16: If we make option optional, it think that will be more correct with older mSupply versions.
At the moment the option_id will be set to "" which isn't correct, it should be `null` in the database
```suggestion
    #[serde(deserialize_with = "empty_str_as_option_string")]
    option_id: Option<String>,
```

server/service/src/sync/translations/vvm_status.rs
```diff
@@ -0,0 +1,95 @@
...
+    ) -> Result<PullTranslateResult, anyhow::Error> {
+        let data = serde_json::from_str::<LegacyVVMStatusRow>(&sync_record.data)?;
+        let result = VVMStatusRow {
+            id: data.ID,
+            description: data.description,
+            code: data.code,
+            level: data.level,
+            is_active: data.is_active,
+            unusable: data.unusable,
+            reason_id: Some(data.option_id),
```
jmbrunskill, line 47: With the change above you'll need this...

```suggestion
            reason_id: data.option_id,
```

jmbrunskill (changes requested):
The test `test_sync_pull_and_push` is failing, you'll need to fix that.

I think we should also make `option_id` optional, as right now there's no way to set it in mSupply so it should definitely be optional.

server/service/src/sync/translations/vvm_status.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[allow(non_snake_case)]
+#[derive(Deserialize)]
+pub struct LegacyVVMStatusRow {
+    ID: String,
+    description: String,
+    code: String,
+    level: i32,
+    is_active: bool,
+    unusable: bool,
```
jmbrunskill, line 15:
```suggestion
    #[serde(default)]
    unusable: bool,
```

This will mean if the value is missing it will default to false. I think it's good idea to do this in case an older version of mSupply is used we want to make sure we don't fail to sync the record.

server/service/src/sync/test/test_data/vvm_status.rs
jmbrunskill: Why did you resolve this one? Did you feel it's not needed to change?

zachariah-at-msupply: I believe I committed it, and then it automatically resolved, but I take it that that is different from what you are seeing?

zachariah-at-msupply: Ah, and I actually think I reversed the change in the latest commit, as it was causing a test to fail.

jmbrunskill (approved):
When I used the old mSupply version, I was seeing these errors:
`missing field `option_id` at line 1 column 99`

I think it's nice to make it work with an older version of mSupply, otherwise we'd need to update the compatibility matrix and make open-mSupply fail to sync at all on an older version which we don't want to do unless we need to.

So I've done a quick update to fix that on your branch. Have also updated the sync test case to handle the missing field as per my code review.

PR #7559, opened 2025-05-01
Fix DateUtils tests

CarlosNZ:
Sorry, no issue here, just a super-quick fix so the front-end tests don't fail. 

They're failing today because it's 1st May, and the DateUtils tests use the current date and there's discrepancies with month calculations as it ticks over to the new month.

# 👩🏻‍💻 What does this PR do?

Have changed the test to use Jest's Mock timers to set a fixed date as the current datetime, which makes the tests stable and not dependent on the current date.

# 🧪 Testing

- [ ] Run front-end tests, or just the "DateUtils" ones with `yarn test date`
- [ ] Confirm all pass

### DIFF
--- a/client/packages/common/src/intl/utils/DateUtils.test.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.test.ts
@@ -15,38 +15,43 @@ describe('useFormatDateTime', () => {
 });
 
 describe('getDisplayAge', () => {
+  jest.useFakeTimers().setSystemTime(new Date('2025-10-01'));
   const hookResult = renderHookWithProvider(() => useFormatDateTime());
   const { getDisplayAge } = hookResult.result.current;
-  const today = new Date();
 
   it('returns age in years when patient is over 1 year or 1 year old', () => {
-    const dob = DateUtils.addYears(today, -9);
+    const dob = new Date('2016-02-01');
     const result = getDisplayAge(dob);
     expect(result).toBe('9');
   });
 
   it('returns age in months and days when patient is less than 1 year old', () => {
-    const threeMonthsAgo = DateUtils.addMonths(today, -3);
-    const dob = DateUtils.addDays(threeMonthsAgo, -2);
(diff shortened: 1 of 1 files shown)


client/packages/common/src/intl/utils/DateUtils.test.ts
```diff
@@ -15,38 +15,43 @@ describe('useFormatDateTime', () => {
...
-    const dob = DateUtils.addYears(today, -9);
+    const dob = new Date('2016-02-01');
     const result = getDisplayAge(dob);
     expect(result).toBe('9');
   });
 
   it('returns age in months and days when patient is less than 1 year old', () => {
-    const threeMonthsAgo = DateUtils.addMonths(today, -3);
-    const dob = DateUtils.addDays(threeMonthsAgo, -2);
+    const dob1 = new Date('2025-09-01');
```
jmbrunskill, line 29: Why not keep the `Date -X` calculation, the makes it easier to read and understand the test to me?
Otherwise I have to do mental arithmetic on the date you've put in vs the fake date time...

jmbrunskill (approved):
Tests seem to pass, I'd prefer to have the datetime offsets as they were though.
If you like this way better I'm ok with it though.

CarlosNZ:
> Tests seem to pass, I'd prefer to have the datetime offsets as they were though. If you like this way better I'm ok with it though.

Yeah, a couple of reasons:
- if there's a bug in the `add` logic and the same  bug in the `difference` logic, then you won't notice it if you're doing add then difference
- I actually found a bug (in my opinion, possibly debateable) in the `differenceInMonths` function (that's in the `date-fns` library, not our code) which means that in some circumstances you can add a month, but then the returned difference in months is still 0 (I can explain it to you in person if you're interested)

So with that in mind, I thought it better (if a little more manual labour) to explictly show the dates. I'd like to fix that month-counting bug at some point, but it's a lot of work for a situation that will rarely occur in practice.

Thanks for the review :)

PR #7562, opened 2025-05-01
7511 Fix unable to save stocktake line for an item

### LINKED ISSUE #7511: Stocktake not letting you enter a stockline with error stockline already exist in stocktake
## What went wrong? 😲

https://github.com/user-attachments/assets/9704233a-4915-4d1f-9c3a-e60e35dfcd17

Right now when you go into stocktake and uncheck the "count this line" it will not let you enter the stock in. The only way to work around is you have to search with item filter and then enter the item in and it will not have the error message "stock line already exist in stocktake".

Not sure why the first way of doing it that it does not allow you to enter and says "stockline already exists in stock" - it should be able to let you do both ways, as users would not know to filter for the item and enter it in.

datafile is on the demo-open server (central tamaki warehouse)

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #7511 

# 👩🏻‍💻 What does this PR do?

The problem was that if some of the stock_lines for the item you are editing are on a separate page in the API response, the UI doesn't know that the stock line has been created and there for 

This fix feels a bit messy, I feel like an upsert from the backend might be a nice solution but it's a big change.
A slightly nicer solution would be to add an item_id filter to stocktake lines rather than relying on the existing itemNameOrCode filter.

Didn't do either of those right now to keep the fix small.

# 🧪 Testing

See testing steps in issue to re-create...

- [ ] Try editing stocktake lines where the lines for item item span across the pages of the stock take.
- [ ] Try adding new lines
- [ ] Try editing existing lines
- [ ] Basically try to find if there's a scenario where this fix doesn't work!

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -4248,6 +4248,7 @@ export type MasterListNode = {
   __typename: 'MasterListNode';
   code: Scalars['String']['output'];
   description: Scalars['String']['output'];
+  discountPercentage: Scalars['Float']['output'];
   id: Scalars['String']['output'];
   linesCount?: Maybe<Scalars['Int']['output']>;
   name: Scalars['String']['output'];
@@ -7709,6 +7710,7 @@ export type StocktakeLineConnector = {
 export type StocktakeLineFilterInput = {
   id?: InputMaybe<EqualFilterStringInput>;
   itemCodeOrName?: InputMaybe<StringFilterInput>;
+  itemId?: InputMaybe<EqualFilterStringInput>;
   locationId?: InputMaybe<EqualFilterStringInput>;
   stocktakeId?: InputMaybe<EqualFilterStringInput>;
 };
--- a/client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useDraftStocktakeLines.ts
+++ b/client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useDraftStocktakeLines.ts
@@ -8,7 +8,7 @@ export const useDraftStocktakeLines = (
 ): [DraftStocktakeLine[], Dispatch<SetStateAction<DraftStocktakeLine[]>>] => {
   const { id } = useStocktake.document.fields('id');
   const { data: stockLines } = useStockLines(item?.id || '');
-  const { lines } = useStocktake.line.rows();
+  const { lines } = useStocktake.line.rows(item?.code);
   const [draftLines, setDraftLines] = useState<DraftStocktakeLine[]>([]);
 
   useEffect(() => {
@@ -24,6 +24,7 @@ export const useDraftStocktakeLines = (
         id,
         stocktakeLines ?? []
       );
+
       setDraftLines(fromStockLines.concat(fromStocktakeLines));
     }
   }, [stockLines, lines, item, id]);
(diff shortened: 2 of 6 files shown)


client/packages/inventory/src/Stocktake/api/hooks/line/useStocktakeLines.ts
```diff
@@ -2,11 +2,23 @@ import { useQuery, useUrlQueryParams } from '@openmsupply-client/common';
...
 import { useStocktakeNumber } from '../document/useStocktake';
 
-export const useStocktakeLines = (id: string) => {
+export const useStocktakeLines = (id: string, itemCode?: string) => {
   const { queryParams } = useUrlQueryParams({
     initialSort: { key: 'itemName', dir: 'asc' },
     filters: [{ key: 'itemCodeOrName' }],
   });
+
+  if (itemCode) {
```
jmbrunskill, line 11: The existance of itemCode changes the behaviour of this hook quite a bit.
Maybe should be a seperate hook? or we could look at one of the other solutions I mention in the issue.

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useDraftStocktakeLines.ts
```diff
@@ -8,7 +8,7 @@ export const useDraftStocktakeLines = (
 ): [DraftStocktakeLine[], Dispatch<SetStateAction<DraftStocktakeLine[]>>] => {
   const { id } = useStocktake.document.fields('id');
   const { data: stockLines } = useStockLines(item?.id || '');
-  const { lines } = useStocktake.line.rows();
+  const { lines } = useStocktake.line.rows(item?.code);
```
jmbrunskill, line 11: Using itemCode here as theres isn't an item_id filter in the backend yet.
Should I add it? Or is there a better overall solution?

client/packages/inventory/src/Stocktake/api/hooks/line/useStocktakeRows.ts
```diff
@@ -15,9 +15,12 @@ const getStocktakeItems = (lines: StocktakeLineFragment[]) =>
     } as StocktakeSummaryItem;
   });
 
-export const useStocktakeRows = () => {
+export const useStocktakeRows = (itemCode?: string) => {
   const { data: stocktake } = useStocktake.document.get();
-  const { data: lineData, isLoading } = useStocktakeLines(stocktake?.id ?? '');
+  const { data: lineData, isLoading } = useStocktakeLines(
```
jmbrunskill, line 20: Maybe we need something like this?
```suggestion
  const { data: lineData, isLoading } = useStocktakeLinesForItem(
```

client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks/useDraftStocktakeLines.ts
jmbrunskill: There's a risk that we item/code filter an item name might match a code or visa-versa, so I think correct would be to use the item_id

jmbrunskill:
> Looks good. Took me a minute to figure out what was going on, but it makes sense. Only thing would be to add a filter by ID, as you suggested. How long do you think that would take? If it's quick I think it'd be worth doing, but otherwise can leave it for now.

Not too long will jump on it now

client/packages/common/src/types/schema.ts
```diff
@@ -4248,6 +4248,7 @@ export type MasterListNode = {
   __typename: 'MasterListNode';
   code: Scalars['String']['output'];
   description: Scalars['String']['output'];
+  discountPercentage: Scalars['Float']['output'];
```
jmbrunskill, line 4251: Unrelated but schema must not have been updated properly before

CarlosNZ (approved):
Nice, LGTM 👍

PR #7567, opened 2025-05-01
5537 supplier return item search autocomplete not scrolling

lache-melvin:
Fixes reopened #5537

# 👩🏻‍💻 What does this PR do?

Adds an `has_stock_on_hand` filter to item search, so we only get the relevant results from backend. Can then remove the problematic frontend filter, which doesn't map well to paginated autocomplete.

Took me a while to track down the issue, we thought it was a problem with autocomplete, but it was actually the relationship with item query. We were querying all visible items (paginated) and then filter on frontend for items with stock.

Issue was happening in the case of:

- There are lots of items visible to your store (let's say 5000)
- You only have stock on hand for a few of them (say 200 items)
- Supplier return modal gets items in pages of 100 items
- Of that initial set of 100 visible items, on average, only 4 would have stock on hand
- Front end filters those results, so only 4 items in the autocomplete
- Because 4 items fit in the view, no need for a scroll bar
- No scroll = no action to trigger the call to get next page of items 😱 

The less noticeable version was when you had in stock enough of the items starting with 'A' (i.e. at the start of the list) to have the scroll bar/trigger pagination, but then would only add 2ish items per scroll, because most weren't visible, which seemed super weird.

Also, querying several thousand items only to show a few is not very efficient! Backend query should return only relevant results, not rely on frontend filtering.

# 🧪 Testing

hit me up if you need db file

- [ ] Have loads of items visible to your store
- [ ] Only have stock for some of them
- [ ] Create a Supplier Return, and click Add Item
- [ ] Autocomplete should be scrollable, with about 100 items per scroll-to-bottom

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Returns/modals/SupplierReturn/ItemSelector.tsx
+++ b/client/packages/invoices/src/Returns/modals/SupplierReturn/ItemSelector.tsx
@@ -40,7 +40,7 @@ export const ItemSelector: FC<ItemSelectorProps> = ({
             disabled={disabled}
             currentItemId={itemId}
             onChange={newItem => newItem && onChangeItemId(newItem.id)}
-            filter={{ isOnHand: true }}
+            filter={{ hasStockOnHand: true }}
             extraFilter={
               disabled
                 ? undefined
--- a/client/packages/invoices/src/Returns/modals/SupplierReturn/ItemSelector.tsx
+++ b/client/packages/invoices/src/Returns/modals/SupplierReturn/ItemSelector.tsx
@@ -40,13 +40,11 @@ export const ItemSelector: FC<ItemSelectorProps> = ({
             disabled={disabled}
             currentItemId={itemId}
             onChange={newItem => newItem && onChangeItemId(newItem.id)}
-            includeNonVisibleWithStockOnHand
+            filter={{ isOnHand: true }}
             extraFilter={
               disabled
                 ? undefined
-                : item =>
-                    item.availableStockOnHand !== 0 &&
-                    !existingItemIds?.some(id => id === item.id)
+                : item => !existingItemIds?.some(id => id === item.id)
             }
           />
         </Grid>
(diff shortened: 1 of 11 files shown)


client/packages/invoices/src/Returns/modals/SupplierReturn/ItemSelector.tsx
```diff
@@ -40,13 +40,11 @@ export const ItemSelector: FC<ItemSelectorProps> = ({
...
             onChange={newItem => newItem && onChangeItemId(newItem.id)}
-            includeNonVisibleWithStockOnHand
+            filter={{ hasStockOnHand: true }}
             extraFilter={
               disabled
                 ? undefined
-                : item =>
-                    item.availableStockOnHand !== 0 &&
-                    !existingItemIds?.some(id => id === item.id)
+                : item => !existingItemIds?.some(id => id === item.id)
```
lache-melvin, line 47: We have a lot of uses of StockItemSearchInput that have an extra filter like this... it might bite us too? (Filter excludes all items that already exist in the invoice.)

E.g. I have 150 items visible, I include 130 in an order... and then, of the first 100, only 7 aren't in the order, so I show those in the item selector - that's not enough to trigger the scroll/pagination, can't access the other 23...

Haven't seen this raised as a bug, seems like an edge case, but possibly we should have an "excludeItems" array on the API filter, rather than frontend filter? Can raise an issue?

server/repository/src/db_diesel/item.rs
```diff
@@ -109,6 +111,11 @@ impl ItemFilter {
     }
 
     pub fn has_stock_on_hand(mut self, value: bool) -> Self {
+        self.has_stock_on_hand = Some(value);
+        self
+    }
+
+    pub fn visible_or_on_hand(mut self, value: bool) -> Self {
```
lache-melvin, line 118: Renamed existing helpful function to what it actually mapped to 😅

roxy-dao (changes requested):
LGTM~ Thanks. Just gotta remove some vars that aren't being used anymore (:

roxy-dao (approved):
LGTM~ Ty Laché

PR #7584, opened 2025-05-05
7569 FIx JSON Form telephone field

### LINKED ISSUE #7569: Regression with the phone field in Patient JSON form
## What went wrong? 😲

Using the most recent Patient JSON Form

Noticing some sort of regression with the field - it doesn't always clear, and when it does, I am unable to save the changes. But, on other occasions, I can save the changes when the field is not populated but the red border & text are still present? See the video below: 

https://github.com/user-attachments/assets/f96464fb-545b-4710-b00d-531dd7ea62ff

## Expected behaviour 🤔

Should be easily clearable and functional, like the phone field in the Next of Kin section

## How to Reproduce 🔨

1. Have the most up-to-date Patient JSON schemas uploaded to your datafile
2. Go to Dispensary -> Patients -> create a new patient
3. In the Detailed view, scroll down and play around with the telephone field
4. Note the regression

## Your environment 🌱

- Open mSupply Version: V2.7.1-RC1

- Database type: PostgreSQL

CarlosNZ:
Fixes #7569

# 👩🏻‍💻 What does this PR do?

This was actually a worse problem than what was described in the issue -- there was an infinite re-render going on due to the customErrors triggering a useEffect that in turn updated customErrors (I think).

I've mitigated this as much as I can for now, but I think this whole JSON Forms area needs a significant re-factor as there's all kinds of anti-patterns going on in it (such as returning un-memoized JSX from a hook, which causes a re-mount of child components on re-render). I think we could make this a lot more robust and less bug-prone with a few days cleanup.

I'm still not 100% sure I've totally fixed the issue, as I can still re-create the occasional bit of weirdness around the `isDirty` state update and the error messaging, but not sure if that's just my hot reloading, as it seems to go away when I do a full refresh. I can't consistently recreate any problems such that I can describe the steps though 🤷‍♂️. So keep an eye out :) 

# 🧪 Testing

You'll want to check the following states as you change the telephone field:

- the "Save" button -- should be enabled if changes have been made, and they are valid
- the "Error message" in the telephone field -- should only appear when the pattern is not matched (but not when it's empty)

With that in mind, check that you can do the following things:
- [ ] Can enter a valid phone number into a patient that doesn't currently have one
- [ ] You get an error and can't save if the entered text is invalid
- [ ] WIth a patient that does already have a phone number, check that you can backspace on the number to get the error state, but if you keep backspacing to nothing it should become valid and save-able
- [ ] Can "clear" a phone number (by selecting and deleting) and save the change
- [ ] Make sure all your changes (above) are saved correctly (i.e. persist after a refresh)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { ControlProps, rankWith, schemaTypeIs } from '@jsonforms/core';
 import { withJsonFormsControlProps } from '@jsonforms/react';
 import {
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Works well for me.
I'm slightly concerned about customer error change in case it some how triggers incorrectly but from a quick test I couldn't break it so I think we're good.

PR #7612, opened 2025-05-06
7605 Add fallback to handle undefined value - fix crash on backspace NewStocklineModal

### LINKED ISSUE #7540: Available SOH shouldn't be on `New Stock` modal
## What went wrong? 😲
When adding new stock the focus is on available packs and when you input, save button doesn't enable.
https://github.com/user-attachments/assets/a287e5b1-7d70-4d54-9b33-d0796404c72f

## Expected behaviour 🤔
Available packs should only be available as a read-only component in the list view and detail view, since it is an indicator of how many packs you have minus any reserved in invoices. 

Remove this input from the new stock modal.

## How to Reproduce 🔨

1. Go to Inventory -> View Stock
2. Click `New Stock` button
3. See focus is on `Available (packs)` 
4. If you enter a number in `Available (packs)` the Save button does not enable

## Your environment 🌱

- Open mSupply Version: 2.7

- Platform:

- Database type:

### LINKED ISSUE #7605: omSupply crashes when backspacing in Inventory Adjustments
## What went wrong? 😲

When creating an Inventory Adjustment and backspacing (without entering any values, just backspacing straight away), the app/browser/apk crash

Error: 

Image

Video: 

Attached [here](https://drive.google.com/drive/folders/17spetJvjiiGtX4Z6yYIA1bjjpp6bB5dI)

## Expected behaviour 🤔

Should not crash

## How to Reproduce 🔨

1. Go to Inventory -> View Stock -> New Stock
2. Choose an item that has 0 stock
3. It will autofocus on Available (packs)
4. Backspace -> the page will crash

## Your environment 🌱

- Open mSupply Version: V2.7.1

- Platform: happens on all platforms - Google Chrome (Mac) and Galaxy Tab A8 (apk)

- Database type: PostgreSQL & SQLite3

noel-yeldos:
Fixes #7605 fixes #7540

# 👩🏻‍💻 What does this PR do?

- Add fallback value for totalNumberOfPacks so that undefined value does not throw error
- Remove available packs input field

EDIT - after requested changes
Available packs field still available on stock detail view but not in new stockline modal

Screenshot 2025-05-07 at 09 25 40

Screenshot 2025-05-07 at 09 25 25

~~I haven't applied the fix to the available packs field -> it seems redundant to add code that will only get removed later. We should also remove the available packs field as requested here #7540 but not sure if that needs to go into 2.7.1. I don't see why not~~

# 🧪 Testing

- [ ] Go to Inventory -> View Stock -> New Stock
- [ ] Choose an item that has 0 stock
- [ ] Autofocus on Pack Qty
- [ ] Enter an amount in Pack Qty
- [ ] Remove amount completely by backspacing 
- [ ] See that app does not crash

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/utils/DateUtils.test.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.test.ts
@@ -15,38 +15,43 @@ describe('useFormatDateTime', () => {
 });
 
 describe('getDisplayAge', () => {
+  jest.useFakeTimers().setSystemTime(new Date('2025-10-01'));
   const hookResult = renderHookWithProvider(() => useFormatDateTime());
   const { getDisplayAge } = hookResult.result.current;
-  const today = new Date();
 
   it('returns age in years when patient is over 1 year or 1 year old', () => {
-    const dob = DateUtils.addYears(today, -9);
+    const dob = new Date('2016-02-01');
     const result = getDisplayAge(dob);
     expect(result).toBe('9');
   });
 
   it('returns age in months and days when patient is less than 1 year old', () => {
-    const threeMonthsAgo = DateUtils.addMonths(today, -3);
-    const dob = DateUtils.addDays(threeMonthsAgo, -2);
+    const dob1 = new Date('2025-09-01');
+    expect(getDisplayAge(dob1)).toBe('1 month, 0 days');
 
-    const dayOffset =
-      DateUtils.getDaysInMonth(today) - DateUtils.getDaysInMonth(dob);
+    const dob3 = new Date('2024-10-02');
+    expect(getDisplayAge(dob3)).toBe('11 months, 29 days');
 
-    const result = getDisplayAge(dob);
-    expect(result).toBe(
-      `3 months, ${2 + (dayOffset > 0 ? dayOffset : 0)} days`
-    );
+    const dob4 = new Date('2025-04-01');
+    expect(getDisplayAge(dob4)).toBe('6 months, 0 days');
   });
 
   it('returns age in days when patient is less than 1 month old', () => {
-    const dob = DateUtils.addDays(today, -10);
+    const dob = new Date('2025-09-21');
     const result = getDisplayAge(dob);
     expect(result).toBe('10 days');
+
+    const dob2 = new Date('2025-09-30');
+    expect(getDisplayAge(dob2)).toBe('1 day');
   });
 
   it('returns an empty string if dob is not defined', () => {
     const dob = null;
     const result = getDisplayAge(dob);
     expect(result).toBe('');
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
   });
 });
--- a/client/packages/system/src/Stock/Components/StockLineForm.tsx
+++ b/client/packages/system/src/Stock/Components/StockLineForm.tsx
@@ -119,6 +119,22 @@ export const StockLineForm: FC<StockLineFormProps> = ({
               />
             }
           />
+          {!isInModal && (
+            <StyledInputRow
+              label={t('label.available-packs')}
+              Input={
+                <NumericTextInput
+                  autoFocus
+                  disabled={!packEditable}
+                  width={160}
+                  value={parseFloat(draft.availableNumberOfPacks.toFixed(2))}
+                  onChange={availableNumberOfPacks =>
+                    onUpdate({ availableNumberOfPacks })
+                  }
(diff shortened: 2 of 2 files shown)


client/packages/system/src/Stock/Components/StockLineForm.tsx
```diff
@@ -108,27 +108,17 @@ export const StockLineForm: FC<StockLineFormProps> = ({
...
+                    ? parseFloat(draft.totalNumberOfPacks.toFixed(2))
+                    : 0
+                }
                 onChange={totalNumberOfPacks =>
                   onUpdate({ totalNumberOfPacks })
                 }
               />
             }
           />
-          <StyledInputRow
```
noel-yeldos, line 118: Addresses #7540

lache-melvin: Available packs still needs to be available on the stock detail page, just not in the new stock line form 🙏

client/packages/system/src/Stock/Components/StockLineForm.tsx
```diff
@@ -108,27 +108,17 @@ export const StockLineForm: FC<StockLineFormProps> = ({
                 autoFocus
                 disabled={!packEditable}
                 width={160}
-                value={parseFloat(draft.totalNumberOfPacks.toFixed(2))}
+                value={
+                  draft.totalNumberOfPacks
+                    ? parseFloat(draft.totalNumberOfPacks.toFixed(2))
+                    : 0
```
lache-melvin, line 114: 🚀

client/packages/common/src/intl/utils/DateUtils.test.ts
```diff
@@ -15,38 +15,43 @@ describe('useFormatDateTime', () => {
 });
 
 describe('getDisplayAge', () => {
+  jest.useFakeTimers().setSystemTime(new Date('2025-10-01'));
```
lache-melvin, line 18: Hm, assuming these changes went in because you branched off develop but now merging into RC? 

No worries for this one, just test changes so not concerned about them going in, but worth double checking when you branch 🤓

lache-melvin (changes requested):

client/packages/common/src/intl/utils/DateUtils.test.ts
noel-yeldos: @lache-melvin I branched off 2.7.1 and saw that tests were failing. Checked in with Carl and realised that the fix for the tests went into develop. Carl suggested just copying the changes across in this PR.

lache-melvin: Ah awesome, thanks for explaining 🚀

client/packages/system/src/Stock/Components/StockLineForm.tsx
noel-yeldos: Thanks for catching that! Updated here cf82fe3f532d30203180daba1ac1d23861376149

lache-melvin: Niiice 🎉  one last request...

This works, but `isInModal` is a display based prop, rather than a use case based prop. Currently we only have the two places this form is used, but it's good to treat it as though it might be used in more places. I think it makes more sense to say "we only show available packs, in cases where the number of packs is not editable. Available packs will always be readonly".

In that case, could we update to:

```tsx
 {!packEditable && (
            <StyledInputRow
              label={t('label.available-packs')}
              Input={
                <NumericTextInput
                  autoFocus
                  disabled
                  ...
```

Then good to go 🙏

lache-melvin (changes requested):
Just requesting the different prop used, then ready to ship 🔥

client/packages/system/src/Stock/Components/StockLineForm.tsx
noel-yeldos: @lache-melvin ah I see, thank you for explaining that! I didn't think about it from the multiple use case perspective. Good learning 🙏 

Updated here 6b994c7d5dd090369727e08c638cc5670d01181f

lache-melvin (approved):
🚀

PR #7618, opened 2025-05-06
Fix GraphiQL playground

CarlosNZ:
Fix for the recently broken GraphQL playground, which we access on `http://localhost:8000/graphql`

# 👩🏻‍💻 What does this PR do?

Problem was that the version of the graphiql package wasn't being specified in the script tag, but the style sheet tag specified version (3.0.6). And so when the library was updated to V4 a few days ago, we now get a mismatch between the script JS and the styles. 

I've removed the version specifier completely -- I think it should be fine to always fetch the latest stable version. Is there any reason not to?

ALSO...

I've fixed something that's been annoying me for a while -- I've wrapped the Login form in proper `<form>` tags, so that we can now submit using the "Enter" key, using default form behaviour.

Some auto-formatting changes in here. I suggest reviewing with the "Hide whitespace changes" enabled for a cleaner view of the actual changes.

I've targeted the 2.7.x branch, as that's going to be merged into develop sooner than the other way round. Can do a quick fix to develop as well if anyone's itching for it (but maybe just merge 2.7 into develop sooner rather than later @jmbrunskill ?)

# 🧪 Testing

- [ ] Check it looks correct
- [ ] Check the login works correctly (with Enter key, as well as button click)
- [ ] Confirm queries still work as normal

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/playground.html
+++ b/server/graphql/playground.html
@@ -36,6 +36,46 @@
   <script>
     // Login plugin for GraphiQL
     function LoginContent() {
+      const [isSubmitting, setIsSubmitting] = React.useState(false);
+
+      function handleSubmit(event) {
+        event.preventDefault();
+
+        if (isSubmitting) return;
+
+        setIsSubmitting(true);
+
+        const formData = new FormData(event.target);
+        const username = formData.get('username');
+        const password = formData.get('password');
+
+        fetch('/graphql', {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: JSON.stringify({
+            query: `query Login {
+  authToken(password: "${password}", username: "${username}") {
+    ... on AuthToken {
+      __typename
+      token
+    }
+  }
+}`,
+          }),
+        })
+          .then((res) => res.json())
+          .then((res) => {
+            const token = res.data.authToken.token;
+            localStorage.setItem('token', token);
+            window.location.reload();
+          })
+          .finally(() => {
+            setIsSubmitting(false);
+          });
+      }
+
       return React.createElement(
         'div',
         {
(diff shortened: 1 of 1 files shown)


server/graphql/playground.html
```diff
@@ -1,145 +1,183 @@
...
+      margin-bottom: 10px;
+    }
+  </style>
+
+  <link rel="stylesheet" href="https://unpkg.com/graphiql/graphiql.min.css" />
+  <link rel="stylesheet" href="https://unpkg.com/@graphiql/plugin-explorer/dist/style.css" />
+</head>
+
+<body>
+  <div id="graphiql">Loading...</div>
```
CarlosNZ, line 28: This is all auto-formatting changes except for the `<script>` tags

server/graphql/playground.html
```diff
@@ -1,145 +1,183 @@
...
+      function handleSubmit(event) {
+        event.preventDefault();
+
+        if (isSubmitting) return;
+
+        setIsSubmitting(true);
+
+        const formData = new FormData(event.target);
+        const username = formData.get('username');
+        const password = formData.get('password');
```
CarlosNZ, line 50: Gets values from "formData" rather than querying the DOM

server/graphql/playground.html
```diff
@@ -1,145 +1,183 @@
...
+            onSubmit: handleSubmit,
+            style: { display: 'flex', flexDirection: 'column', alignItems: 'center' },
           },
-          React.createElement('h1', null, 'Login'),
           React.createElement('input', {
             type: 'text',
             id: 'username',
+            name: 'username',
             placeholder: 'Username',
+            autoFocus: true,
```
CarlosNZ, line 102: autofocus FTW! 🥳

jmbrunskill (approved):
Hooray, enter works on logins now!

Thanks for the fix!

PR #7623, opened 2025-05-06
70 the toggle 'record stock transaction for past vaccination' doesn't work when a stockline is ticked

roxy-dao:
Part of [70](https://github.com/msupply-foundation/open-msupply-internal/issues/70)

# 👩🏻‍💻 What does this PR do?
There was a change to auto-select batch if there is only one, this caused the batch to be sent into the API call and the insert function only checks if there's a stock line before creating an invoice!

This PR sets the batch to 'null' if the user is choosing to not record a stock transaction.

https://github.com/user-attachments/assets/eb7c6dd5-6d82-40ff-bf40-83736d0ece9d

Didn't do this part of the issue: 

```
- outcome of the toggle is unclear (does it put back stock, does it remove stock, does it keep stock unchanged ?
- confusing UI : all text look the same, all text has the same toggle
```

Feels like that needs UX input, but if using the same style/font/component for both recording and reverting is causing confusion, then we should think of ways to lessen that either with different colours, or whatever~

record
revert

# 🧪 Testing

- [ ] Create a vaccination encounter for a patient
- [ ] Click on one of the dose lines
- [ ] Give the patient a vaccination in the **past**
- [ ] **Switch** toggle to say you don't want to record the stock transaction -> Save
- [ ] Check stock line -> stock line for item should **stay the same**
- [ ] Now go back and give another dose in the **encounter**, but this time you want to record the stock transaction
- [ ] Check stock line -> stock line should have decreased
- [ ] Check prescriptions -> Prescriptions should have been created

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1992,8 +1992,5 @@
   "warning.caps-lock": "Warning: Caps lock is on",
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm against the patient's physical vaccination card first.",
   "warning.field-not-parsed": "{{field}} not parsed",
-  "warning.nothing-to-supply": "Nothing left to supply!",
-  "description": {
-    "suggested-quantity": "The number of units which the system has calculated should be ordered"
-  }
+  "warning.nothing-to-supply": "Nothing left to supply!"
 }
--- a/client/packages/system/src/Vaccination/api/useVaccination.ts
+++ b/client/packages/system/src/Vaccination/api/useVaccination.ts
@@ -177,6 +177,8 @@ const useInsert = ({
     if (!encounterId) return;
 
     const isOtherFacility = input.facilityId === OTHER_FACILITY;
+    const shouldUpdateStockLine =
+      input.given && !isOtherFacility && input.createTransactions;
 
     const apiResult = await api.insertVaccination({
       storeId,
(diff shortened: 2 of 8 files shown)


mark-prins:
thanks @roxy-dao for jumping on it so quickly! agree with what you say about the UI aspect.

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1992,8 +1992,5 @@
   "warning.caps-lock": "Warning: Caps lock is on",
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm against the patient's physical vaccination card first.",
   "warning.field-not-parsed": "{{field}} not parsed",
-  "warning.nothing-to-supply": "Nothing left to supply!",
-  "description": {
```
lache-melvin, line 1996: this was definitely not used somehwere in a report right?

client/packages/system/src/Vaccination/api/useVaccination.ts
```diff
@@ -177,6 +177,8 @@ const useInsert = ({
     if (!encounterId) return;
 
     const isOtherFacility = input.facilityId === OTHER_FACILITY;
+    const shouldUpdateStockLine =
```
lache-melvin, line 180: nit: `shouldCreatePrescription`? `shouldCreateStockTransaction` ?

lache-melvin (approved):
Looks great to me! Would also be on board with a quick change to add the different colour toggle for different requests, but if uncertain about that, happy to move to separate issue outside of patch 🙏

client/packages/common/src/intl/locales/en/common.json
jmbrunskill: Pretty confident it's not, on develop I just changed it to `description.suggested-quantity` as I think that was the intention, thought i'd fixed this branch but clearly hadn't...

PR #7638, opened 2025-05-07
7603 [DRAFT] debounce render of rnr row and move item meta to new row

### LINKED ISSUE #7603: RnR Form - Scroll Speed
Epic: https://github.com/msupply-foundation/open-msupply-internal/issues/69

On large master list, when scrolling through items, they load slowly causing white screen, which sometimes becomes permanently white. This is mainly due to render speed of a 'frame' of items, and it seems like when they are not rendered in time, Virtualise Scroll View gets confused

Debounce rendering of all elements apart from item metadata, this way initial render is quick and there is less chance of above problem.

In my testing i've also noticed that this method is jumpy and 'janky' with lots of column and lots of sticky column. Also on related note noticed that with sticky columns (without the debounce change), sometimes when you scroll down and click into initial field keyboard opens but there is left/right scroll and input becomes hidden. Therefore as part of this solution I would also add item metadata to another row (which opens up a lot of space)

Refactoring to inputs, and other slow components

andreievg:
**this should be draft**

Fixes #7603

This PR is a draft, i tried debouncing with item meta in the same row, but got very jumpy UI (I think due to sticky columns)

# 👩🏻‍💻 What does this PR do?

Item meta as row and search in 'ven' column

[This apk](https://www.dropbox.com/scl/fi/46ve50ywig301lpsnwjfs/RnDDemo.apk?rlkey=ux039rxlahsmrv3b9hkeuqlcv&st=ufw7b165&dl=0) is of the base branch and [this apk](https://www.dropbox.com/scl/fi/en3bix0o7ggi9mv5uas0h/RnRDemo-item-row.apk?rlkey=l8agpckzpqcelwagn3c3f3vcd&st=c8j7k3q6&dl=0) is this branch, i like the item met in a new row way, probably needs more design input, from UX of data entry, this seems easier, mainly due to being able to see context of previous entered data. And of course scrolling is fast, you can actually scroll and navigate to item you want.

You can see below, how this looks in the two way of doing this. However I think it's important to be aware of the difference from APK perspective, below was just a fast way for me to share show things look in the two optiosn

https://github.com/user-attachments/assets/8984f6f7-4392-4ca3-8726-a0434210c515

https://github.com/user-attachments/assets/6d6d4441-07bb-4422-be97-61d81859937d

# 🧪 Testing

You can test any of the above APK as client connecting to the server that uses data file provided in the epic. You really need a large master list for this

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/android/app/src/main/java/org/openmsupply/client/NativeApi.java
+++ b/client/packages/android/app/src/main/java/org/openmsupply/client/NativeApi.java
@@ -6,7 +6,9 @@ import android.net.nsd.NsdManager;
 import android.net.nsd.NsdServiceInfo;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.SystemClock;
 import android.util.Log;
+import android.view.KeyEvent;
 import android.webkit.WebView;
 
 import androidx.annotation.MainThread;
@@ -327,6 +329,34 @@ public class NativeApi extends Plugin implements NsdManager.DiscoveryListener {
         }
         call.resolve(response);
     }
+
+    @PluginMethod()
+    public void sendTabKeyPress(PluginCall call) {
+        long downTime = SystemClock.uptimeMillis();
+        long eventTime = SystemClock.uptimeMillis();
+        WebView webView = this.getBridge().getWebView();
+
+        KeyEvent tabDownEvent = new KeyEvent(
+                downTime,
+                eventTime,
+                KeyEvent.ACTION_DOWN,
+                KeyEvent.KEYCODE_TAB,
+                0
+        );
+
+        KeyEvent tabUpEvent = new KeyEvent(
+                downTime,
+                eventTime + 50,
+                KeyEvent.ACTION_UP,
+                KeyEvent.KEYCODE_TAB,
+                0
+        );
+
+        // Dispatch the events
+        webView.post(() -> webView.dispatchKeyEvent(tabDownEvent));
+        webView.post(() -> webView.dispatchKeyEvent(tabUpEvent));
+    }
+
 
     // NsdManager.DiscoveryListener
     @Override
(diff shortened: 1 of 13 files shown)


lache-melvin:
@mark-prins - probs a PO input as much as anything needed here?

mark-prins:
thanks @andreievg it appears that you've spent a lot of time on this 🙏 

I'm still not keen on the two-lines-per-item look; we've spent enough time on this already though. One request before committing to it, can you please pass it by Anisha and ask for a design review?

Some other questions
- Can you do anything about the initial page load? navigating to the R&R form and getting only alternate rows showing is a bit weird ( a simple solution is a skeleton or some other full page overlay which hides the strangeness for a second - alternatively adding an `<InlineSpinner />` to the empty table cell looks ok )
- when you say 'Item meta in the same row' you mean the item name and code right?
- There's a magnifying glass in the `VEN` header cell which looks like is part of the header cell and also doesn't look actionable ( it just looks like a grey icon ) Can this be moved out somewhere like the toolbar and be an input pls?

I did have a quick try and the sticky initial columns works ok for me on desktop - much the same as the version in this PR:

https://github.com/user-attachments/assets/6ac0745c-9036-4696-a978-35ff143a0a09

but as you say, you don't get as many input columns visible on-screen at the same time

andreievg:
@mark-prins I don't really want to make this a distraction and sync your and Anisha's time into this, i think there are more pressing issues at hand, just wanted to run an idea past you. 

Looks like your debounce implementation with the loader looks quite good and the client was happy with the latest one row workflow, so no urgency to change. If there are concerned still with the 'white' screen, even after we've introduced the search, we can do your suggestion with the spin loader, going to close this for now. Thanks for you input

PR #7651, opened 2025-05-07
7606 add doses per unit & vvm type to central server UI

### LINKED ISSUE #7606: Add doses_per_unit & VVM Type to central server UI
Add doses_per_unit to the Central Server UI for editing item_variants

## Tasks 📝
- [ ] Add doses_per_unit to Central Server UI
- [ ] Add VMM_type to Central Server UI
- [ ] Add doses_per_unit & vvm_type to service layer and graphql
- [ ] Update frontend mutations to use new graphql 

### Acceptance Criteria ✅
- [ ] Can see doses_per_unit in the Central Server UI
- [ ] Can add doses_per_unit to a new item_variant
- [ ] Can edit doses_per_unit on an existing item_variant
- [ ] Can add vvm_type to a new item_variant
- [ ] Can edit vvm_type on an existing item_variant

### Moneyworks Jobcode 🧰
NIGGF:DSE

roxy-dao:
Fixes #7606

# 👩🏻‍💻 What does this PR do?
- Adds `doses_per_unit` and `vvm_type` to the `item_variant` table and sync.
- Allows these two columns to be upserted for vaccine items
- Update item variant UI to display these two columns for editing and viewing 
- Have also updated manufacturer and cold storage type to allow these inputs to be cleared

Screenshot 2025-05-08 at 7 46 14 AM
Screenshot 2025-05-08 at 7 46 20 AM

# 🧪 Testing

- [ ] (dev) Run tests
- [ ] If you have any existing variants, can try to update the doses_per_unit or vvm_type (note these two columns can only be updated for vaccine items)
- [ ] Go to `Item` -> `Item Variant`
- [ ] Click `Add Variant`
- [ ] Add variant with doses per unit and vvm type
- [ ] Click save
- [ ] Should save and show up in the tab
- [ ] For non-vaccine items these two inputs should be disabled
- [ ] Try to empty storage type and manufacturer 
- [ ] Should empty

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -985,11 +985,6 @@ export type CannotChangeStatusOfInvoiceOnHold = UpdateErrorInterface &
     description: Scalars['String']['output'];
   };
 
-export type DoseConfigurationNotAllowed = UpsertItemVariantErrorInterface & {
-  __typename: 'DoseConfigurationNotAllowed';
-  description: Scalars['String']['output'];
-};
-
 export type CannotDeleteInvoiceWithLines = DeleteCustomerReturnErrorInterface &
   DeleteErrorInterface &
   DeleteInboundShipmentErrorInterface &
@@ -2277,6 +2272,11 @@ export type DocumentSortInput = {
   desc?: InputMaybe<Scalars['Boolean']['input']>;
   /** Sort query result by `key` */
   key: DocumentSortFieldInput;
+};
+
+export type DoseConfigurationNotAllowed = UpsertItemVariantErrorInterface & {
+  __typename: 'DoseConfigurationNotAllowed';
+  description: Scalars['String']['output'];
 };
 
 export type EncounterConnector = {
@@ -4318,7 +4318,10 @@ export type ItemVariantNode = {
   coldStorageTypeId?: Maybe<Scalars['String']['output']>;
   dosesPerUnit: Scalars['Int']['output'];
   id: Scalars['String']['output'];
+  item?: Maybe<ItemNode>;
+  /** @deprecated From 2.8.0. Use item instead */
   itemId: Scalars['String']['output'];
+  /** @deprecated From 2.8.0. Use item instead */
   itemName: Scalars['String']['output'];
   manufacturer?: Maybe<NameNode>;
   manufacturerId?: Maybe<Scalars['String']['output']>;
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -985,6 +985,11 @@ export type CannotChangeStatusOfInvoiceOnHold = UpdateErrorInterface &
     description: Scalars['String']['output'];
   };
 
+export type DoseConfigurationNotAllowed = UpsertItemVariantErrorInterface & {
+  __typename: 'DoseConfigurationNotAllowed';
+  description: Scalars['String']['output'];
+};
+
 export type CannotDeleteInvoiceWithLines = DeleteCustomerReturnErrorInterface &
   DeleteErrorInterface &
   DeleteInboundShipmentErrorInterface &
@@ -4311,6 +4316,7 @@ export type ItemVariantNode = {
   bundlesWith: Array<BundledItemNode>;
   coldStorageType?: Maybe<ColdStorageTypeNode>;
   coldStorageTypeId?: Maybe<Scalars['String']['output']>;
+  dosesPerUnit: Scalars['Int']['output'];
   id: Scalars['String']['output'];
   itemId: Scalars['String']['output'];
   itemName: Scalars['String']['output'];
@@ -4318,6 +4324,7 @@ export type ItemVariantNode = {
   manufacturerId?: Maybe<Scalars['String']['output']>;
   name: Scalars['String']['output'];
   packagingVariants: Array<PackagingVariantNode>;
+  vvmType?: Maybe<Scalars['String']['output']>;
 };
 
 export type ItemVariantNodeManufacturerArgs = {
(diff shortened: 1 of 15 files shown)


client/packages/system/src/Item/DetailView/DetailView.tsx
```diff
@@ -104,9 +104,7 @@ export const ItemDetailView: FC = () => {
 
   isCentralServer &&
     tabs.push({
-      Component: (
-        <ItemVariantsTab itemId={data.id} itemVariants={data.variants} />
-      ),
+      Component: <ItemVariantsTab item={data} itemVariants={data.variants} />,
```
roxy-dao, line 107: Changed the prop to item, so I can access item's isVaccine

client/packages/common/src/types/schema.ts
```diff
@@ -9587,12 +9597,14 @@ export type UpsertItemVariantErrorInterface = {
 };
 
 export type UpsertItemVariantInput = {
-  coldStorageTypeId?: InputMaybe<Scalars['String']['input']>;
+  coldStorageTypeId?: InputMaybe<NullableStringUpdate>;
```
roxy-dao, line 9600: Making this and manufacturer id save as None if input is cleared

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
```diff
@@ -152,6 +154,34 @@ const ItemVariantForm = ({
             </Box>
           }
         />
+        <InputWithLabelRow
+          label={t('label.doses-per-unit')}
+          labelWidth="200"
+          Input={
+            <NumericTextInput
+              fullWidth
+              disabled={!variant.item?.isVaccine}
```
roxy-dao, line 163: I've disabled the input, but maybe we should just hide if the item isn't a vaccine item?

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantsTab.tsx
```diff
@@ -141,6 +146,28 @@ const ItemVariant = ({
               />
             }
           />
+          <InputWithLabelRow
```
roxy-dao, line 149: Same comment as above, hide if item is vaccine?

server/graphql/item_variant/src/mutations/upsert.rs
```diff
@@ -12,18 +16,19 @@ use service::{
...
 };
 
 #[derive(InputObject)]
 pub struct UpsertItemVariantInput {
     pub id: String,
     pub item_id: String,
     pub name: String,
-    pub cold_storage_type_id: Option<String>,
-    pub manufacturer_id: Option<String>,
+    pub cold_storage_type_id: Option<NullableUpdateInput<String>>,
```
roxy-dao, line 27: Refactor to make inputs nullable

server/graphql/types/src/types/item_variant.rs
```diff
@@ -30,14 +30,22 @@ impl ItemVariantNode {
...
 
+    #[graphql(deprecation = "From 2.8.0. Use item instead")]
     pub async fn item_id(&self) -> &String {
         &self.item.id
     }
-
+    #[graphql(deprecation = "From 2.8.0. Use item instead")]
     pub async fn item_name(&self) -> &String {
         &self.item.name
     }
```
roxy-dao, line 40: We're using the Node as the draft, so can't remove item_id, but wondering if we can remove the item name? 👀

client/packages/system/src/Item/DetailView/Tabs/ItemVariants/ItemVariantModal.tsx
aimee-mcneil-melville: I believe generally fields or buttons are kept in view even if not applicable for a consistent UI however.. as a user it could be confusing with no indication why these were disabled, especially if a store doesn't use vaccines.

Personally I'd prefer these hidden as these fields don't apply to a non-vaccine item at all. No risk of a jumpy UI as the variants are accessed individually rather than by a 'Next' button.

Is this more of a product/design decision though?

server/graphql/types/src/types/item_variant.rs
lache-melvin: We really shouldn't remove any fields from our GQL nodes, at least until major version bump, possibly 2 major versions... in case external integrations, reports etc. depend on these fields.

This might be a bit of a dogmatic case though, seeing as item variants were only added for GAPS features which haven't rolled out anywhere yet...

Why are you wanting to deprecate these though? If the only bits the frontend needed from item was id/name, it would be better to use these fields, rather run the whole item loader?

aimee-mcneil-melville: Is this because isVaccine is now needed from the item as well? 

And therefore would be a decision between adding isVaccine separately, or loading the whole item and depreciating the existing individual ones?

lache-melvin: That would drive the frontend decision - for areas where we need to display in doses, they care the item variant is for a vaccine item. Therefore, those queries should use the full item field, or we could add a separate field like you mentioned.

But it's reasonable to expect that item variants might be queried in different cases, where we don't care if they relate to vaccine items, or any other item info, just the item id, or maybe the item name 👀

roxy-dao: Yes, only asking because item variants is still waiting for pilot. The item variants is running the item loader anyways for isVaccine... so since it's running it anyways, can remove.

aimee-mcneil-melville (changes requested):
Thanks Roxy, the doses and vvm type works great! I can save, edit, remove with no issues 😄 

There's two changes to request:
Remove the visibility of the dose and vvm fields if the item is not a vaccine - hiding them makes more sense to me as they are not relevant to a non-vaccine item at all

And allow the clearing of the Manufacturer input - your changes all look good but the input is set to clearable=false and it doesn't clear!

aimee-mcneil-melville (approved):
Thanks Roxy, it's looking great and working well!

I retried the tests and they have passed so ready to merge in 🚀

PR #7690, opened 2025-05-12
Fix GraphQL playground (Again!)

CarlosNZ:
Lol, I said when I [fixed this last week](https://github.com/msupply-foundation/open-msupply/pull/7618):

> I've removed the version specifier completely -- I think it should be fine to always fetch the latest stable version. Is there any reason not to?

Well, they proved me wrong and went and broke it again with v4.0.3 -- now the Explorer panel doesn't scroll!

So I've just fixed the version to 4.0.2 for now -- I guess we should test each version before upgrading rather than assuming new releases are solid.

### DIFF
--- a/server/graphql/playground.html
+++ b/server/graphql/playground.html
@@ -20,7 +20,7 @@
     }
   </style>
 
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Thanks

PR #7712, opened 2025-05-13
7472 outbound shipment edit refactor

### LINKED ISSUE #7472: Prescription line edit: simplify logic

lache-melvin:
Part of #7472

IS BEAST HEH - plz just grab me im happy to chat through changes 🙈 ❤️ 

# 👩🏻‍💻 What does this PR do?

Only UI change is the removal of pack selector here - discussion with PMs was that it wasn't used, and adds confusion. Given there was a lot of re-implementation here, was easier to leave off for now. If it turns out it is a required feature, we can add back separately:

Screenshot 2025-05-13 at 12 43 36 PM

OTHERWISE - epic rearrangement + simplification of code.

Included:
- Targeted query & mutation for the outbound edit modal (allows removal of batch (isCreated/updated logic)
- Introduces `allocationContext` - all allocation logic and state handled through here, one place
  - This will be where you pass your "lens" (issue in doses or packs, sort by expiry of vvm etc.)
- Handle placeholder entirely separate, rather than always looking for placeholder line in arrays
- Bit of a rearrange of components based on removal of dependencies

Sorry for how large - I think for big refactor like this, hard to do piece by piece without quite a lot of churn... anyway, would probably recommend reviewing by just navigating the code that's there now, than trying to parse the changes?

Haven't refactored prescriptions - would like to do this over cooldown period, to reuse this new AllocationContext. Would be able to delete a lot more code after that, there are a lot of shared dependencies.

Also still to do, maybe in cool down? lots to try get in for 2.8, feel I shouldn't spend too much more time here this week: document expected behaviour - for discussion with product team

# 🧪 Testing

- [ ] Barring the removal of the pack selector, it should work the same. Please test:
- [ ] Auto allocation:
  - [ ] Allocates correct quantity
  - [ ] Splits across many stock lines
  - [ ] Warns if not enough stock (creates placeholder in New status, otherwise just warn)
  - [ ] Rounds up to nearest pack size (e.g. if requesting 50 but you only have packs of 100)
  - [ ] Doesn't allocate expired/on hold
  - [ ] Allocates nearest expiring first
- [ ] Can manually allocate expired lines, but not on hold
- [ ] Manual allocation updates the auto-allocation value
- [ ] Add from scanner button on android (scanned batch is highlighted, others are disabled)

# 📃 Documentation

  1. Screenshots & mentions of packs
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/utils/arrays/SortUtils.ts
```diff
@@ -26,8 +26,13 @@ const parseValue = <T, K extends keyof T>(object: T, key: K) => {
 };
 
 const getSortByExpiry =
-  (desc = false) =>
+  (desc = false, nonExpiringLast = false) =>
```
lache-melvin, line 29: Old way was leaving the lines with no expiry value in place? Probably mostly a dev data problem, but I would think "non-expiring" batches should be allocated last

client/packages/invoices/src/OutboundShipment/DetailView/AddFromScannerButton.tsx
```diff
@@ -13,12 +13,12 @@ import {
...
 import { useOutbound } from '../api';
 import { isOutboundDisabled } from '../../utils';
-import { Draft } from '../..';
+import { ScannedBarcode } from '../../types';
 
 export const AddFromScannerButtonComponent = ({
   onAddItem,
 }: {
-  onAddItem: (draft?: Draft) => void;
+  onAddItem: (scannedBarcode?: ScannedBarcode) => void;
```
lache-melvin, line 21: updated the onOpen logic to explicitly handle opened via barcode vs open via item click

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
```diff
@@ -0,0 +1,152 @@
...
+} from './allocation/useAllocationContext';
+import { sumAvailableQuantity } from './allocation/utils';
+
+interface AllocationProps {
+  itemData: OutboundLineEditData;
+  allowPlaceholder: boolean;
+  scannedBatch?: string;
+}
+
+export const Allocation = ({
```
lache-melvin, line 32: Please have a gander per new component structure - but idea was:

```
- OutboundLineEdit (the modal)
  - SelectItem
  - Allocation
    - AutoAllocation (input + alerts)
    - Table (manual allocation)
```

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
```diff
@@ -0,0 +1,152 @@
...
+}: AllocationProps) => {
+  const { initialise, initialisedForItemId } = useAllocationContext(
+    ({ initialise, initialisedForItemId }) => ({
+      initialise,
+      initialisedForItemId,
+    })
+  );
+
+  useEffect(() => {
+    initialise(
```
lache-melvin, line 45: Wrapper component... one time initialisation of the allocation context per item

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
 
-  const placeholder = draftStockOutLines?.find(
-    ({ type, numberOfPacks }) =>
-      type === InvoiceLineNodeType.UnallocatedStock && numberOfPacks !== 0
+  const { data: itemData, isFetching } = useOutboundLineEditData(
+    invoiceId,
+    itemId
   );
+  const { mutateAsync } = useSaveOutboundLines(invoiceId);
+  const { saveBarcode } = useOpenedWithBarcode(asBarcodeOrNull(openedWith));
```
lache-melvin, line 58: Pulled as much of the barcode-y biz into this hook as i could

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
   const handleSave = async (onSaved: () => boolean | void) => {
+    const confirmZeroQuantityMessage = t('messages.confirm-zero-quantity');
     if (
-      getAllocatedQuantity(draftStockOutLines) === 0 &&
-      !showZeroQuantityConfirmation
+      allocatedUnits === 0 &&
+      !alerts.some(alert => alert.message === confirmZeroQuantityMessage)
     ) {
-      setShowZeroQuantityConfirmation(true);
+      setAlerts([{ message: confirmZeroQuantityMessage, severity: 'warning' }]);
```
lache-melvin, line 103: Didn't want to have to pass the showZeroQuantityConfirmation prop all the way through, and I think this is more semantic - just set the alerts in the context to whatever they need to be, and let the renderer display them.

I think a further improvement would be for the translation of the alert messages to be in the component, so we're just dealing with the logical concept of the alert at this point but... now I'm in the weeds 😁

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
-  const { warning } = useNotification();
-  const [showZeroQuantityConfirmation, setShowZeroQuantityConfirmation] =
-    useState(false);
-  useFocusNumberOfPacksInput(draft);
+  const { next, disabled: nextDisabled } = useNextItem(itemId);
 
-  const placeholder = draftStockOutLines?.find(
-    ({ type, numberOfPacks }) =>
-      type === InvoiceLineNodeType.UnallocatedStock && numberOfPacks !== 0
+  const { data: itemData, isFetching } = useOutboundLineEditData(
```
lache-melvin, line 53: custom query - all the bits we need for the modal in one query - rather than composing several queries, with loads of state updates after each resolves. Much more efficient for slower networks/devices too 🎉

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/allocateQuantities.test.ts
```diff
@@ -0,0 +1,186 @@
+import { FnUtils } from '@openmsupply-client/common';
+import { allocateQuantities } from './allocateQuantities';
+import { DraftStockOutLineFragment } from '../../../api/operations.generated';
+
+describe('allocateQuantities - standard behaviour.', () => {
```
lache-melvin, line 5: test cases are the same as old tests (removed no-longer-relevant cases like placeholder though)

client/packages/invoices/src/OutboundShipment/api/hooks/utils/index.ts
```diff
@@ -5,7 +5,6 @@ import { useAddFromMasterList } from './useAddFromMasterList';
 import { useBarcode } from './useBarcode';
 import { useBarcodeInsert } from './useBarcodeInsert';
 import { useSelectedLines } from './useSelectedLines';
-export * from './useGetItemPricing';
```
lache-melvin, line 8: We just include pricing info in the outbound line itemData query

client/packages/invoices/src/Prescriptions/api/hooks/utils.tsx
```diff
@@ -157,25 +156,10 @@ export interface DraftPrescriptionLineSeeds {
 export const createDraftPrescriptionLineFromStockLine = ({
   invoiceId,
   stockLine,
-  defaultPricing,
```
lache-melvin, line 160: default pricing was never being passed in for prescriptions...

client/packages/invoices/src/StockOut/Components/StockOutAlerts.tsx
```diff
@@ -24,7 +24,15 @@ export const StockOutAlerts = ({
       ? allocationAlerts
       : [];
 
-  if (alerts.length === 0) return null;
+  return <AllocationAlerts allocationAlerts={alerts} />;
```
lache-melvin, line 27: leaving existing component for now, for prescriptions. using the display only one below for outbound

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -177,268 +121,6 @@ export const getAllocatedQuantity = (draftStockOutLines: DraftStockOutLine[]) =>
     3
   );
 
-export const issueStock = (
```
lache-melvin, line 180: `allocation` folder should prob get moved here to be shared with prescription, but for now seeing as only used in outbound, kept it there, along with its utils. prescriptions currently has its own copy of these functions

server/graphql/invoice_line/src/mutations/outbound_shipment_line/save_item_lines.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+fn map_error(error: SaveOutboundShipmentLinesError) -> Result<InvoiceNode> {
+    use SaveOutboundShipmentLinesError::*;
+    let formatted_error = format!("{:#?}", error);
+
+    log::error!("Error: {}", formatted_error);
+
+    // Future TODO: Implement structured errors where needed
+    // (Would only occur if 2 people editing at same time)
```
lache-melvin, line 92: Were returning all the structured errors in the old mutation - but weren't handling them on frontend? So I figured this was simpler, until there are actually errors we want to translate!

server/graphql/types/src/types/draft_outbound_shipment_line.rs
```diff
@@ -0,0 +1,93 @@
...
+    ) -> Vec<DraftOutboundShipmentLineNode> {
+        shipment_lines
+            .into_iter()
+            .map(|shipment_line| DraftOutboundShipmentLineNode { shipment_line })
+            .collect()
+    }
+}
+
+#[Object]
+impl DraftOutboundShipmentLineNode {
```
lache-melvin, line 40: custom node for the modal, just the fields we need

server/graphql/types/src/types/draft_outbound_shipment_line.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[Object]
+impl DraftOutboundShipmentItemData {
+    pub async fn draft_lines(&self) -> Vec<DraftOutboundShipmentLineNode> {
+        DraftOutboundShipmentLineNode::from_vec(self.lines.clone())
+    }
+
+    pub async fn placeholder_quantity(&self) -> Option<f64> {
+        self.placeholder_quantity
+    }
```
lache-melvin, line 21: Returning placeholder quan separately - draft lines is therefore just the StockOut lines. easier on logic that constantly filtering for a placeholder...

andreievg:
> @mark-prins I have one concern here... as discussed, removed the pack based auto-allocation, it's just in units now.
> 
> But if I'm looking at an item where pack size is 1000, and I want to ship 100 packs... I'd have to type 100,000?
> 
> I know the "any" thing was confusing, but I'm wondering if the "if all pack sizes the same, issue in that pack size" was somehting that made sense? Though, I guess its also confusing if sometimes the modal is in packs, and sometimes units.
> 
> Again, as the requirements here seem fuzzy, I don't want to spend too much time on what's not useful, but wanted to raise with you before we deploy 2.8, in case this is something we want to smooth over before then: 🙏

DJ would 100% need either default pack size or something like that, but maybe allocation logic is separate to "calculation aid" (for selecting if it's dose/unit/particular pack that is entere in 'issue'
)

client/packages/common/src/utils/arrays/SortUtils.ts
andreievg: > "non-expiring" batches should be allocated last

Yep that's how it's done on b/e

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -0,0 +1,190 @@
...
+    const sortedLines = draftLines.sort(SorterByStrategy[strategy]);
+
+    // Separate lines here, so only dealing with allocatable lines going forward
+    // Note - expired is still considered allocatable, just not via auto-allocation
+    const [allocatableLines, nonAllocatableLines] = ArrayUtils.partition(
+      sortedLines,
+      line =>
+        scannedBatch
+          ? scannedBatchFilter(sortedLines, line, scannedBatch)
+          : canAllocate(line)
```
andreievg, line 89: canAllocate is also checked in scannedBatchFilter, wonder if it should be lifted up

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -0,0 +1,190 @@
...
+      isDirty: false,
+      draftLines: [],
+      nonAllocatableLines: [],
+      placeholderQuantity: null,
+      initialisedForItemId: null,
+      allocateIn: AllocateIn.Units,
+      alerts: [],
+    })),
+
+  setDraftLines: lines =>
```
andreievg, line 116: I wasn't sure if this is only used internally, if it is could move it up to be a closure method (just cause on initialise lines are sorted, but could set unsorted lines with this one)

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/allocateQuantities.ts
```diff
@@ -0,0 +1,200 @@
...
+ * with a bias towards the first stock line(s) in the list (e.g. first expiring).
+ *
+ * This can take up to 3 steps, let's look at an example with 2 stock lines:
+ *
+ * 200 x pack size 1    (exp 01/26)
+ * 3 x   pack size 100  (exp 03/26)
+ *
+ * And we are requesting 350 units.
+ *
+ * Step 1: from top of list allocate to the nearest (rounded down) pack size
```
andreievg, line 22: We actually need to confirm this allocation, I think OG uses FEFO full stop, with fractional number of packs if below pack size.Which makes sense, imagine you have a pallet of good expiring this week, and a carton expiring this year, system should really suggest that you split the pallet when issuing a carton

client/packages/invoices/src/StockOut/utils.tsx
```diff
@@ -177,268 +121,6 @@ export const getAllocatedQuantity = (draftStockOutLines: DraftStockOutLine[]) =>
...
-      htmlInput: {
-        sx: {
-          backgroundColor: props.isDisabled ? undefined : 'background.white',
-        },
-      },
-    }}
-  />
-);
-
 export const getAllocationAlerts = (
```
andreievg, line 124: I like this as one place to figure see what alerts will trigger and when

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -0,0 +1,190 @@
...
+        placeholderQuantity:
+          placeholderQuantity === null ? null : stillToAllocate,
+      }));
+    }
+  },
+
+  manualAllocate: (lineId, quantity) => {
+    const { draftLines, setDraftLines } = get();
+
+    const updatedLines = issueStock(draftLines, lineId, quantity);
```
andreievg, line 181: With this need allocateIn (or is that in future PR)

lache-melvin: Yah future PR

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -0,0 +1,190 @@
...
+        hasExpired,
+        format,
+        t
+      );
+
+      set(state => ({
+        ...state,
+        alerts,
+        placeholderQuantity:
+          placeholderQuantity === null ? null : stillToAllocate,
```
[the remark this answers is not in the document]
lache-melvin: Shouldn't set the placeholder quantity if not using a placeholder line? After allocated status, we should just under-allocate if we don't have enough stock..

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  draftLines: DraftStockOutLineFragment[],
+  idToIssue: string,
+  packs: number
+) => {
+  const foundRowIdx = draftLines.findIndex(({ id }) => id === idToIssue);
+  const foundRow = draftLines[foundRowIdx];
+  if (!foundRow) return draftLines;
+
+  const newDraftLines = [...draftLines];
+  newDraftLines[foundRowIdx] = {
```
andreievg, line 44: I feel like this is a mutation ? (since previous restructure doesn't copy draftLines, it just create a new array with them), I think for immutablility have to destructure down to primatives for objects that change, so somthing like [...sliceBefore, { ...found, newNumberOfPacks} ...sliceAfter]

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useOpenedWithBarcode.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const useFocusNumberOfPacksInput = (
+  batch: string | undefined,
+  itemId: string | null
+) => {
+  useEffect(() => {
+    if (!batch || !itemId) return;
+    setTimeout(() => {
+      const input = document.getElementById(getPackQuantityCellId(batch));
```
andreievg, line 45: I guess this is copy and paste ? useRef set to zustand state would be the alternative

jmbrunskill (commented):
Nice! Great effort lots of code here.
I'm having setup issues testing it (pretty sure it's nothing to do with your branch) so adding a comment and will try to test and approve later.

jmbrunskill (approved):
Ok did some testing looks good.

Haven't tested on Tablet or with barcodes though. Has that been done?

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
+  const {
+    draftLines,
+    allocatedUnits,
+    placeholderQuantity,
+    alerts,
+    isDirty,
+    setAlerts,
+    clear,
+  } = useAllocationContext(state => ({
+    draftLines: state.draftLines,
```
andreievg, line 69: not sure if this selector is really needed

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
-    if (showZeroQuantityConfirmation && newVal !== 0)
-      setShowZeroQuantityConfirmation(false);
-
-    return newAllocateQuantities;
-  };
-
-  const canAutoAllocate = !!(currentItem && draftStockOutLines.length);
-  const okNextDisabled =
-    (mode === ModalMode.Update && nextDisabled) || !currentItem;
+  const okNextDisabled = (mode === ModalMode.Update && nextDisabled) || !itemId;
```
andreievg, line 95: I was confused about this (although I think it's product decision, this behaviour will need to be listed to confirm that's what we want), in add new can ok next to new item but in update mode cant

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -1,197 +1,130 @@
...
-        setCurrentItem(next);
+        setItemId(next.id);
         return true;
       }
       if (mode === ModalMode.Create) {
-        setCurrentItem(null);
+        setItemId(undefined);
         return true;
       }
       onClose();
```
andreievg, line 130: don't think this ever happens

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
andreievg: This one confused me, Allocation will 'refresh' because loading will render while fetching. It makes sense once you look at it in more detail, but I still think we can improve on that at some point, almost feel like when item is fetched there is a bunch of things to be done like this, and kind of feels like it should be an await.
This is still readable, i'll just put a comment about this useEffect

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  draftLines: DraftStockOutLineFragment[];
+  placeholderQuantity: number | null;
+}) =>
+  NumUtils.round(
+    draftLines.reduce(
+      (acc, { numberOfPacks, packSize }) => acc + numberOfPacks * packSize,
+      0
+    ),
+    3
+  ) + (placeholderQuantity ?? 0);
```
andreievg, line 32: This is another one to add to description of behaviour, if we have placeholder less then what you can supply, and we manually supply that, placeholder stays. We kind of need something similar to prescription, 'prescribed' quantity or it could be placeholder, but the rule is, placeholder is populated with how much you want to issue, but is displayed as placholder quantity - how much you've issued

server/service/src/invoice_line/get_draft_outbound_lines.rs
```diff
@@ -0,0 +1,336 @@
...
+
+    let available_stock_lines = stock_line_repo.query_by_filter(
+        StockLineFilter::new()
+            // For selected item
+            .item_id(EqualFilter::equal_to(item_id))
+            // With available stock in this store
+            .store_id(EqualFilter::equal_to(&store_id))
+            .has_packs_in_store(true)
+            // Exclude stock lines already in the invoice
+            .id(EqualFilter::not_equal_all(existing_stock_line_ids)),
```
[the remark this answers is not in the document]
andreievg: James noticed this while i was passing by, on the first look yeah it seems like it should be available, but since we are generating draft lines, it's actually total (we show total in item selection too), the idea is you have some in stock even though it might not be available, and the user can then figure out if they want to re-locate from other shipment

server/service/src/invoice_line/get_draft_outbound_lines.rs
```diff
@@ -0,0 +1,336 @@
...
+            number_of_packs,
+            stock_line_id,
+            pack_size,
+            batch,
+            expiry_date,
+            location_id,
+            sell_price_per_pack,
+            // Stock already included in the invoice wouldn't be on the stock line available packs,
+            // but should be considered available in the context of this invoice
+            available_packs: available_number_of_packs + number_of_packs,
```
andreievg, line 213: nice

server/service/src/invoice_line/save_outbound_shipment_item_lines/generate.rs
```diff
@@ -0,0 +1,153 @@
...
+    let check_already_exists = |id: &str| {
+        existing_lines
+            .iter()
+            .any(|line| line.invoice_line_row.id == id)
+    };
+
+    let lines_to_add = lines
+        .clone()
+        .into_iter()
+        .filter(|line| line.number_of_packs > 0.0 && !check_already_exists(&line.id))
```
andreievg, line 54: hmm, so for the 'ids'for draft lines, I think we should really go with stock_line_ids (when getting draft lines and here when checking if there is invoice line)

server/service/src/invoice_line/save_outbound_shipment_item_lines/generate.rs
```diff
@@ -0,0 +1,153 @@
...
+                batch: None,
+                pack_size: None,
+                expiry_date: None,
+                prescribed_quantity: None,
+                total_before_tax: None,
+                tax_percentage: None,
+                note: None,
+                location_id: None,
+                cost_price_per_pack: None,
+                sell_price_per_pack: None,
```
andreievg, line 76: This doesn't feel right, I think sell price needs to be set ?

server/service/src/invoice_line/save_outbound_shipment_item_lines/generate.rs
```diff
@@ -0,0 +1,153 @@
...
+                quantity: quantity as u32, // worth cleaning up this input... gets saved as f64?
+            })
+        }
+        _ => ManagePlaceholderLine::NothingToDo,
+    };
+
+    Ok(GenerateResult {
+        lines_to_add,
+        lines_to_update,
+        lines_to_delete,
```
andreievg, line 150: I think we would still need to check (I think it's ok in carry over or cool down):
* Duplicated stock lines
* Stock lines in store
* Stock line belongs to that item
* Trip all invoice_lines that are not in the payload (for that item)

andreievg:
What a massive effort !

Thanks @lache-melvin, i went through it all and did some light testing, i think the only big question was about pricing changes in prescriptions. 

It's a huge improvement, thank you, I feel more confident navigating this part of the code base now : )

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useOpenedWithBarcode.ts
lache-melvin: Yeah this was existing...

server/graphql/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
```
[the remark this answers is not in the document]
lache-melvin: Ah yeah sorry, compilation failed... I've seen this added in a couple places now (it's the recommended fix by the compiler?)

Might raise an issue to review it though, along with other developer experience performance things...

server/service/src/invoice_line/get_draft_outbound_lines.rs
```diff
@@ -0,0 +1,336 @@
...
+    invoice_id: &str,
+) -> Result<
+    (
+        Vec<DraftOutboundShipmentLine>,
+        Option<f64>, /* placeholder_quantity */
+    ),
+    ListError,
+> {
+    let outbound = get_invoice(ctx, Some(&store_id), invoice_id)?.ok_or(
+        ListError::DatabaseError(RepositoryError::DBError {
```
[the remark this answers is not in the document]
lache-melvin: Yeah not found with a message would be good, especially when its otherwise not clear which record wasn't found 😁

client/packages/invoices/src/Prescriptions/api/hooks/utils.tsx
```diff
@@ -184,7 +168,7 @@ export const createDraftPrescriptionLineFromStockLine = ({
     prescribedQuantity: 0,
     location: stockLine?.location,
     expiryDate: stockLine?.expiryDate,
-    sellPricePerPack,
+    sellPricePerPack: stockLine?.sellPricePerPack ?? 0,
```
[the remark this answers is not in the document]
lache-melvin: Yep sellPrice was using logic as though a pricing object was being passed in, but we were never calling this function with a pricing object. So just stripped back, so behaviour was clear

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
lache-melvin: Added to carry over, behaviour to list and confirm

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/utils.ts
lache-melvin: Ooh yeah great point, won't be entirely new objects, this was from existing logic... and it does look like everything is updating ok, maybe as we aren't doing any deeper comparisons anywhere?

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+export const scannedBatchFilter = (
+  allLines: DraftStockOutLineFragment[],
+  selectedLine: DraftStockOutLineFragment,
+  scannedBatch: string
+) => {
+  if (!canAllocate(selectedLine)) return false;
+
+  const linesIncludeScannedBatch = allLines.some(l => l.batch === scannedBatch);
+
+  // If the requested batch is not in the list, we can allocate any line
```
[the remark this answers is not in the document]
lache-melvin: Yes this is same as existing... will document for confirmation though, its not super intuitive to me...

server/service/src/invoice_line/save_outbound_shipment_item_lines/generate.rs
lache-melvin: Sell price isn't editable in outbound, so leaving as none here means the stock_out line will just use whatever the sell price is on the stock line

lache-melvin:
Thanks for review both of you!! I will merge this (FYI I think will do without squashing seeing as we have a few dependent PRs 😅) - am creating a couple of follow up issues to capture the other comments in here 🙏

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
lache-melvin: d3dd097cec

PR #7713, opened 2025-05-13
6238 Pack qty to show [multiple] in prescriptions

### LINKED ISSUE #6238: Pack Qty shows '0' when multiple batches of the same item get dispensed
## What went wrong? 😲

Pack Qty shows '0' when multiple batches of the same item get dispensed

Screenshot attached ->>> [here](https://drive.google.com/drive/folders/1vv3lGwUiFHwN_CPnXnDpx6G2896G81j4)

## Expected behaviour 🤔

Should show [multiple] for consistency

## How to Reproduce 🔨

1. Have an item with multiple batches (easiest way to achieve that is to repack an item)
2. Go to Dispensary -> Prescriptions and create a new prescription
3. Issue multiple batches -> see some fields show [multiple] but Pack Qty shows '0'
4. Can be confusing

## Your environment 🌱

- Open mSupply Version: V2.5.0-RC5

- Database type: PostgreSQL

noel-yeldos:
Fixes #6238 

# 👩🏻‍💻 What does this PR do?

- Refactor implementation of 'Pack Qty' column -> No longer uses ColumnDefinitionSetBuilder
  - This allows for string `[multiple]` to display in the 'Pack Qty' column when issuing from multiple batches

Screenshot 2025-05-13 at 13 00 36

# 🧪 Testing

- [ ] Have an item with multiple batches
- [ ] Go to Dispensary -> Prescriptions and create a new prescription
- [ ] Issue from multiple batches and save
- [ ] Close Prescription line edit view and go back to detail view
- [ ] See that for the line you have just issue that 'Pack Qty' has value `[multiple]`

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
+++ b/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
@@ -16,7 +16,6 @@ import {
   usePreference,
   PreferenceKey,
   getDosesPerUnitColumn,
-  NumberCell,
   BasicCell,
 } from '@openmsupply-client/common';
 import { StockOutLineFragment } from '../../StockOut';
--- a/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
+++ b/client/packages/invoices/src/Prescriptions/DetailView/columns.ts
@@ -9,7 +9,6 @@ import {
   ArrayUtils,
   useTranslation,
   useColumnUtils,
-  NumberCell,
   CurrencyCell,
   ColumnDescription,
   NumUtils,
@@ -275,46 +274,22 @@ export const usePrescriptionColumn = ({
   }
 
   columns.push(
-    [
-      'numberOfPacks',
-      {
-        Cell: NumberCell,
-        getSortValue: row => {
-          if ('lines' in row) {
-            const { lines } = row;
-            const packSize = ArrayUtils.ifTheSameElseDefault(
-              lines,
-              'packSize',
-              ''
-            );
-            if (packSize) {
-              return lines.reduce((acc, value) => acc + value.numberOfPacks, 0);
-            } else {
-              return '';
-            }
-          } else {
-            return row.numberOfPacks;
-          }
-        },
-        accessor: ({ rowData }) => {
-          if ('lines' in rowData) {
-            const { lines } = rowData;
-            const packSize = ArrayUtils.ifTheSameElseDefault(
-              lines,
-              'packSize',
-              ''
-            );
-            if (packSize) {
-              return lines.reduce((acc, value) => acc + value.numberOfPacks, 0);
-            } else {
-              return '';
-            }
-          } else {
-            return rowData.numberOfPacks;
-          }
-        },
-      },
-    ],
+    {
+      label: 'label.pack-quantity',
+      description: 'description.pack-quantity',
+      key: 'numberOfPacks',
+      align: ColumnAlign.Right,
+      getSortValue: row =>
+        getColumnPropertyAsString(row, [
+          { path: ['lines', 'numberOfPacks'] },
+          { path: ['numberOfPacks'], default: '' },
+        ]),
+      accessor: ({ rowData }) =>
+        getColumnProperty(rowData, [
+          { path: ['lines', 'numberOfPacks'] },
+          { path: ['numberOfPacks'] },
+        ]),
+    },
     {
       label: 'label.unit-price',
       key: 'sellPricePerUnit',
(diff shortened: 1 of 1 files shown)


client/packages/invoices/src/Prescriptions/DetailView/columns.ts
```diff
@@ -275,46 +274,22 @@ export const usePrescriptionColumn = ({
...
+        getColumnPropertyAsString(row, [
+          { path: ['lines', 'numberOfPacks'] },
+          { path: ['numberOfPacks'], default: '' },
+        ]),
+      accessor: ({ rowData }) =>
+        getColumnProperty(rowData, [
+          { path: ['lines', 'numberOfPacks'] },
+          { path: ['numberOfPacks'] },
+        ]),
+    },
```
noel-yeldos, line 292: I have had to implement in this way to allow string to be displayed in `Pack Qty` column. If this is the desired outcome then this should be fine? 

However, if we want to capture the cumulative 'Pack Qty' then we will need to undo this and tweak the initial code to display the cumulative quantity instead of just 0. 

Any thoughts @mark-prins? I'm wondering if there was a reason behind not implementing [multiple] for the 'Pack Qty' field?

mark-prins: I suspect the lack of `[multiple]` here is an oversight. It should be showing.

what you've done isn't quite right though. It will show the number of packs only if all stock lines have the same number. that's not the same behaviour as the previous code.

the previous will show the sum of all the 'numberOfPack' values, if the pack size is the same.

what we'd want is:
* check the pack sizes.. if they differ show `[multiple]`
* if they are the same, sum the number of packs

so the only change needed to the original code is 

264: `        Cell: BasicCell,`
293: `              return '[multiple]';`


the use of `getColumnProperty` is very nice and ideally we'd move toward that. you'd still need to check the packSize first though

noel-yeldos: Thanks for clarifying @mark-prins ! 

I've reverted the changes and made tweaks to display translated `[multiple]` text.

Shows `[multiple]` when `packSize` is different and shows cumulative `numberOfPacks` when `packSize` is same
Screenshot 2025-05-14 at 10 38 13

I also want to clarify the below scenario. 

I have an item with 16 batches. 15 of the batches have `packSize=1000` and 1 batch has `packSize=1`. When I only issue 2 units from the one batch where `packSize=1`, the `Pack Qty` field shows as `[multiple]`. In this instance I would have expected to see `Pack Qty` of 2 because I'm only issuing from one batch. I understand this is expected behaviour based on the code because it's just checking for if `packSize` is the same or not. Is this expected behaviour from user perspective?
Screenshot 2025-05-14 at 10 38 33

mark-prins: oh fun. ideally the pack size and the pack quantity columns would look at stock lines where the `numberOfPacks` is greater than zero.
that makes for clumsier code though which I don't like:

packSize column:
```
accessor: ({ rowData }) =>
          getColumnProperty(
            {
              ...rowData,
              filteredLines:
                'lines' in rowData
                  ? rowData.lines.filter(line => line.numberOfPacks > 0)
                  : undefined,
            },
            [{ path: ['filteredLines', 'packSize'] }, { path: ['packSize'] }]
          ),
```

or

```
        accessor: ({ rowData }) =>
          'lines' in rowData
            ? ArrayUtils.ifTheSameElseDefault(
                rowData.lines.filter(line => line.numberOfPacks > 0),
                'packSize',
                '[multiple]'
              )
            : rowData.packSize,
```

numberOfPacks:
```
            const packSize = ArrayUtils.ifTheSameElseDefault(
              lines.filter(line => line.numberOfPacks > 0),
              'packSize',
              ''
            );
```

though for some reason all lines are grouped on by item on the prescription - which is not correct either! we should show all batches, and changing to that reduces all of this complexity. there's no need for the `[multiple]` at all, nor the special `getSortValue` or `accessor`

mark-prins: I've created #7743 to remove the per-item grouping. This issue can probably just be closed as it will be superseded by the other

noel-yeldos: @mark-prins no worries thanks for following up! Will close this PR and linked issue.

PR #7717, opened 2025-05-13
7607 add created datetime and created by to item variant

### LINKED ISSUE #7607: Add created_datetime and created_by to item_variant
NOTE: this is a nice to have so might not do this one...

item_variants will be created by remote stores in future, however it would be useful if we could find out who created them and when.

## Tasks 📝

- [x] Add created_datetime to the item_variant table datetime
- [ ] add created_by (Text FK -> User_id)
- [ ] Update service layer to correctly record these fields

### Acceptance Criteria ✅
- [ ] Table syncs with Open-mSupply Central Server
- [ ] Sync doesn't fail when the remote server doesn't have the created_datetime/by column
- [ ] When adding a new item_variant the `created_datetime` and `created_by` are recorded correctly
- [ ] activity log records all changes to item_variants

### Moneyworks Jobcode 🧰
NIGGF:DSE

roxy-dao:
Fixes #7607

# 👩🏻‍💻 What does this PR do?
- Migrations to add `created_datetime` and `created_by` to `item_variant` table
- Update `ItemVariant` struct to return `name & cold storage type` rows
- Other party error checks for manufacturer

**Logs**
- Migrations to add activity_log enums for logging `Item Variant` changes
- Log item variant delete/creation
- Activity logs for any changes to an existing variant

Screenshot 2025-05-13 at 2 09 01 PM

Are we displaying the logs as a tab next to item variants or even somewhere on the item variants tab itself?

# 🧪 Testing

- [ ] OMS Central > Item > Variants > create variant
- [ ] Add a manufacturer (check that your manufacturer list only includes the manufacturers visible to your store)
- [ ] Save
- [ ] Check DB
   - [ ] item_variant: created_datetime and created_by should be saved
   - [ ] activity_log: should have logged item variant created
- [ ] Update the variant - new manufacturer/vvm status etc.
- [ ] Check DB
   - [ ] item_variant: created_datetime and created_by should be still be the same
   - [ ] activity_log: should have logged a change for each field you updated 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/item_variant/item_variant.rs
```diff
@@ -1,20 +1,23 @@
...
 use diesel::{
-    dsl::{InnerJoin, IntoBoxed},
+    dsl::{InnerJoin, IntoBoxed, LeftJoin},
     prelude::*,
 };
 
 #[derive(Debug, PartialEq, Clone, Default)]
 pub struct ItemVariant {
     pub item_variant_row: ItemVariantRow,
+    pub manufacturer_row: Option<NameRow>,
```
roxy-dao, line 18: Needed to get the names of manufacturer and cold storage type

server/repository/src/migrations/v2_08_00/add_activity_log_enums.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn identifier(&self) -> &'static str {
+        "add_activity_log_enums"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        if cfg!(feature = "postgres") {
+            sql!(
+                connection,
+                r#"
+                    ALTER TYPE activity_log_type ADD VALUE IF NOT EXISTS 'ITEM_VARIANT_CREATED';
```
roxy-dao, line 15: We really need a better way to handle logging ;x

server/service/src/activity_log.rs
```diff
@@ -90,12 +90,12 @@ pub fn system_log_entry(
     message: &str,
 ) -> Result<(), RepositoryError> {
     let sync_site_id =
-        KeyValueStoreRepository::new(&connection).get_i32(KeyType::SettingsSyncSiteId)?;
+        KeyValueStoreRepository::new(connection).get_i32(KeyType::SettingsSyncSiteId)?;
```
roxy-dao, line 93: Side quest while I was here... removing redundants...

server/service/src/item/item_variant/delete.rs
```diff
@@ -44,6 +44,14 @@ pub fn delete_item_variant(
                 })
                 .collect::<Result<Vec<_>, RepositoryError>>()?;
 
+            activity_log_entry(
+                ctx,
+                ActivityLogType::ItemVariantDeleted,
+                Some(input.id.clone()),
```
roxy-dao, line 50: Unsure which to save here... if we want to show it with the overall logs in item... then will need some refactoring to get the item id

server/service/src/item/item_variant/upsert.rs
```diff
@@ -26,6 +29,9 @@ pub enum UpsertItemVariantError {
     DuplicateName,
     ColdStorageTypeDoesNotExist,
     DoseConfigurationNotAllowed,
+    OtherPartyDoesNotExist,
```
roxy-dao, line 32: We weren't checking Other Party?!

lache-melvin (changes requested):
Heh did u test w frontend? Getting error on save of any kind of item variant - FOREIGN KEY constraint failed, need to pass through user context rather than just basic_context from mutation...

Also, seeing as you're checking other party now... we should probably only show visible manufacturers in the list 🫠 
Screenshot 2025-05-14 at 8 31 38 AM
Screenshot 2025-05-14 at 8 31 45 AM

server/service/src/item/item_variant/delete.rs
roxy-dao: No just here me thinks.. If we move all the logs to a 'log page' then this won't be an issue 👀  though ideally would want to save BOTH the item id and the variant id... 

No issue, just one for refactoring enums which I totally agree. Though would love a bigger refactor to make it more informative..

server/service/src/item/item_variant/upsert.rs
```diff
@@ -81,11 +91,20 @@ pub fn upsert_item_variant(
...
+                existing_item_variant,
+                new_item_variant.clone(),
+                cold_storage_type,
+                manufacturer,
+                input,
+            )?;
+
             ItemVariantRepository::new(connection)
                 .query_one(
                     ItemVariantFilter::new().id(EqualFilter::equal_to(&new_item_variant.id)),
```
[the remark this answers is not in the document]
roxy-dao: But what if you wanna know what it was previously 👀  think it's useful to save change_from as well >.>

server/graphql/item_variant/src/mutations/upsert.rs
```diff
@@ -73,7 +73,7 @@ pub fn upsert_item_variant(
         },
     )?;
     let service_provider = ctx.service_provider();
-    let service_context = service_provider.basic_context()?;
+    let service_context = service_provider.context(store_id, "".to_string())?;
```
lache-melvin, line 76: don't we want the user id from validate auth? To populate created_by?

server/service/src/item/item_variant/upsert/generate.rs
```diff
@@ -0,0 +1,118 @@
...
+        item_link_id: item_id,
+        cold_storage_type_id: cold_storage_type_id.map(|l| l.value).unwrap_or_default(),
+        manufacturer_link_id: manufacturer_id
+            .map(|manufacturer_id| manufacturer_id.value)
+            .unwrap_or_default(),
+        deleted_datetime: None,
+        doses_per_unit,
+        vvm_type: vvm_type.map(|vvm_type| vvm_type.value).unwrap_or_default(),
+        created_datetime: Utc::now().naive_utc(),
+        created_by: Some(user_id.to_string()),
```
lache-melvin, line 36: oo also, this in an upsert, so would be overwriting this on every edit? probably we should use existing after creation?

lache-melvin (changes requested):
Looking good - need to pass through user id, I still can't save variants 🫠 

Can you also please update the manufacturer list in the frontend, it is showing me options for non-visible manufacturers

server/service/src/item/item_variant/upsert/generate.rs
lache-melvin: I think should fix this?

lache-melvin (changes requested):
I tried changing all the fields but only get a log for name? Have you tried?

Screenshot 2025-05-16 at 11 31 02 AM

roxy-dao:
> I tried changing all the fields but only get a log for name? Have you tried?

Yeah?
Screenshot 2025-05-16 at 11 46 47 AM

server/service/src/item/item_variant/upsert/generate.rs
```diff
@@ -0,0 +1,127 @@
...
+                ctx,
+                ActivityLogType::ItemVariantUpdatedName,
+                Some(existing_item_variant.id.clone()),
+                Some(existing_item_variant.name.clone()),
+                Some(updated_item_variant.name.clone()),
+            )?;
+        }
+
+        if existing_item_variant.cold_storage_type_id != updated_item_variant.cold_storage_type_id {
+            if let (Some(existing_storage_type), Some(updated_storage_type)) = (
```
lache-melvin, line 69: Ohhh its bc the Some() Some()! Only creating a log when it existed previously, and you're changing it to something else, not removing.

We should also log when changing to or from None 👀

roxy-dao: Opps, don't know how I missed that >.< Thanks!

lache-melvin (approved):
🔥

Issue #7374, opened 2025-04-16
Population based forecasting: Calculations / Display forecast in Internal Order


roxy-dao:
See calculations [here](https://docs.google.com/document/d/1-zRyCDxBQvtzIhjh8J_X1CeXX-SDRZohiBajSLqUYb8/edit?tab=t.0)

## Tasks 📝
- [ ] Use above document to implement calculation for suggested quantity forecasting
- `Display population based forecasting` is on:
  - [ ] This calculated value should populate a new column called `Target Quantity` (both in detail and line view)
- [ ] The **Use Suggested Quantities** button uses the population based forecast value rather than the suggested quantity value if the line is a vaccine item

### Acceptance Criteria ✅
- [ ] User can see `Target Quantity` column in the detail view if they use the preference
- [ ] Internal Order `suggested quantity` shows forecast value
 
### Moneyworks Jobcode 🧰
NIGGF:VAX

CarlosNZ:
I'm a little unclear with the specifics here. The "Tasks" part makes it seem like this value will be a new column, but the "Acceptance criteria" sounds like it should show in the existing "Suggested Quantities" column but using the new formula.

Do you know which it is @roxy-dao ?

roxy-dao:
Have edited issue~

CarlosNZ:
> * This calculated value should populate a new column called `Target Quantity` (both in detail and line view)

Would "forecasted_quantity" be more precise? I find "Target quantity" vague and potentially confusing.

roxy-dao:
> > * This calculated value should populate a new column called `Target Quantity` (both in detail and line view)
> 
> Would "forecasted_quantity" be more precise? I find "Target quantity" vague and potentially confusing.

Yup happy to go with forecasted quantit

roxy-dao:
Closed by https://github.com/msupply-foundation/open-msupply-forecasting-plugins/pull/1

PR #7718, opened 2025-05-13
7374 View for Vaccine course

CarlosNZ:
Preliminary work for #7374 

# 👩🏻‍💻 What does this PR do?

The issue is being re-defined to do the forecasting via a plugin. So while that's getting sorted, here is a View to make the vaccine course data available as a "View" table, along with a Diesel repository to query it via item_id. This will simplify the queries needed for the forecasting.

Will hold off on merging until the rest is done, so consider this a temporary "feature branch".

# 🧪 Testing

- [ ] Check View table is created, and shows all relevant data for a vaccine course:
  - Name, id, etc
  - Doses
  - Associated items
  - Demographic and program names

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/views/mod.rs
```diff
@@ -441,6 +442,47 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
...
+        coverage_rate,
+        vc.is_active,
+        wastage_rate,
+        vc.deleted_datetime AS vaccine_course_deleted_datetime,
+        vcd.id AS vaccine_course_dose_id,
+        LABEL AS dose_label,
+        min_interval_days,
+        min_age,
+        max_age,
+        vcd.deleted_datetime AS vaccine_course_dose_deleted_datetime,
```
jmbrunskill, line 458: Would it make sense to just filter out deleted in the view? Does this need to return deleted rows?

server/repository/src/migrations/views/mod.rs
```diff
@@ -441,6 +442,47 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
...
+        p.name AS program_name
+      FROM
+        vaccine_course vc
+        JOIN vaccine_course_dose vcd ON vc.id = vcd.vaccine_course_id
+        JOIN vaccine_course_item vci ON vci.vaccine_course_id = vc.id
+        JOIN item_link il ON vci.item_link_id = il.id
+        JOIN item ON item.id = il.item_id
+        LEFT JOIN unit ON item.unit_id = unit.id
+        LEFT JOIN demographic d ON d.id = vc.demographic_id
+        LEFT JOIN PROGRAM p ON p.id = vc.program_id;
```
jmbrunskill, line 485: Program can be an inner join, it's not optional

jmbrunskill (approved):
Looks good, I'm not sure what counts and in vs out of this view.
I think there's things in there that won't be needed for the calculations but then again I gather we're trying to create a re-usable resource?
Personally I'd probably thin it down a bit an only add what we need for now, and can extend it later but I'm happy to merge how it is...

jmbrunskill (changes requested):
I take it back postgres tests are failing. Need to fix that!

jmbrunskill (approved):
Looks good thanks!

PR #7720, opened 2025-05-13
Adds a get_preference function to backend

### LINKED ISSUE #7659: Add "get_preference" service method to back-end
With new store prefs format, we need a common mechanism for back-end to check/get the value of a given pref.

Currently have `get_store_preferences` in `store_preference.rs`, but that's for the old format, but something similar is needed.

jmbrunskill:
Fixes #7659

# 👩🏻‍💻 What does this PR do?

Adds comments for how to access prefs from the backend

# 🧪 Testing

- [ ] Try to use this to get a Pref in an example code
- [ ] Dev Testing only 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/preference/mod.rs
+++ b/server/service/src/preference/mod.rs
@@ -81,3 +81,6 @@ fn append_if_type(
     }
     Ok(())
 }
+
+// Note, we don't have a get_preference() function here as preferences can be accessed like this:
+// `let can_manage = ManageVvmStatus.load(connection, store_id);`
--- a/server/service/src/preference/preferences/mod.rs
+++ b/server/service/src/preference/preferences/mod.rs
@@ -30,14 +30,3 @@ pub fn get_preference_provider() -> PreferenceProvider {
         sort_by_vvm_status: SortByVvmStatus,
     }
 }
-
-/// Gets the preference passed into it, use it like this...
-/// get_preference(ctx, store_id, DisplayPopulationBasedForecasting);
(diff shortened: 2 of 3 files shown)


lache-melvin (commented):
Tbh, I thought we'd just load the pref directly in backend, rather than via abstraction?

```rs
let can_manage = ManageVvmStatus.load(connection, store_id);
```

> I don't like how the new re-structure doesn't make it obvious how we'd have a global vs store based pref.
> For example this function will need to be called without a store_id if it's a global pref. Could fix that in the function?

Agree its not apparent, though it does work... this is the query fn that is called by .load() - uses storeId if its store pref, ignores it if global:

https://github.com/msupply-foundation/open-msupply/blob/b8162e7b1cbe90f245fb4cf12f99bd9a30b9d979/server/service/src/preference/types.rs#L75-L89

Though I'm wondering, could backend querying of prefs always use the query context?

When querying from frontend, on central server, we need to explicitly provide the context we want to load the pref for (i.e. I want to look at the store prefs for this store) but maybe not a backend use case?

This is getting long I'm coming to chat haha

jmbrunskill:
Thought this didn't work but it does, thanks!

lache-melvin (approved):
Noooice, thanks for capturing 🙏

PR #7654, opened 2025-05-07
5082 reasons refactor part 2

noel-yeldos:
Part 2 #5082 (sorry this is a bigger one but mostly deletions)

# 👩🏻‍💻 What does this PR do?

- Remove inventoryAdjustmentReason and ReturnReason components and hooks from front end
- Remove old inventory_adjustment_reason_id fields from back end
- Update migration to drop reason fields from invoice_line and stocktake_line as per both postgres and sqlite guidelines
- Update reports using inventoryAdjustmentReason to a new minor version and have them query reasonOption instead 
- Changes from part 1
  - Remove abstraction pattern for  useReasonOptions hook so that it is called directly
  - Rename inventory_adjustment_reason_id field to reason_option_id in schema
  - Add soft_delete in reason_option_row
  - Update migration to insert fields into reason_option table from existing reasons table and also to trigger re-sync of reasons

## 💌 Any notes for the reviewer?
I'm hoping I've captured everything! 

# 🧪 Testing

This is part 2 of #7589 (only to need to test once as a whole)

Ensure you have reasons configured in OG (preferences -> options) before testing

`Adding New Stock`
- [ ] Navigate to 'Inventory' -> 'View Stock'
- [ ] Click on 'Add New Stock' button
- [ ] Select item and enter 'Pack Qty' and 'Available(packs)'
- [ ] Check that reason selector is enabled
- [ ] Check that reason selector only displays positive inventory adjustment reasons
- [ ] Check that selected reason does not disappear when clicking on other parts of the modal
- [ ] Click 'Ok' and ensure there are no errors
- [ ] Check database -> invoice_line table -> ensure that the stock line addition is in the db with the correct reason_option_id

`Adjusting Existing Stock`
- [ ] Navigate to 'Inventory' -> 'View Stock'
- [ ] Select an item 
- [ ] Click on 'Adjust Stock' button
- [ ] Enter a quantity to adjust
- [ ] Check that reason selector is enabled
- [ ] Check that reason selector only displays POSITIVE inventory adjustment reasons when INCREASE QUANTITY is selected
- [ ] Check that reason selector only displays NEGATIVE inventory adjustment reasons when DECREASE quantity is selected
- [ ] Check that selected reason does not disappear when clicking on other parts of the modal
- [ ] Click 'Ok' and ensure there are no errors
- [ ] Check database -> invoice_line table -> ensure that the stock line adjustment is in the db with the correct reason_option_id

`Supplier Returns`
[...]

PR #7730, opened 2025-05-13
Fix rebuilding views so it doesn't run on test cases

jmbrunskill:
Fixes #

# 👩🏻‍💻 What does this PR do?

This is modelled off @noel-yeldos fix for views in his PR https://github.com/msupply-foundation/open-msupply/pull/7654/commits/2299f44b3516206b9d146b6950bf173916be3924

Since this is needed for https://github.com/msupply-foundation/open-msupply/pull/7718 I've split this out into it's own PR.

I've tested with 7718 and postgres and the tests all pass, so I think it's good.

I've added a check for if the views have been droped before re-creating.
If we haven't dropped any, we shouldn't need to recreate, and there's a warning message now too if the don't run view logic triggers we'll have something to trace back on.

# 🧪 Testing

- [ ] Test that the views are only re-created for latest migration
- [ ] Check all test cases pass

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/mod.rs
+++ b/server/repository/src/migrations/mod.rs
@@ -229,7 +229,8 @@ pub fn migrate(
     let last_version_in_migration_vec = migrations.last().unwrap().version();
 
     // Recreate views only if we've migrated to the latest version
-    // If we do migrations to a early version for a test case, views won't be available
+    // Creating Views on an earlier version migration test might fail due to more recent views referencing schema elements that didn't previously exist
+    // Note: When Migration tests run, views won't be available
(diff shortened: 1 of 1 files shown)


CarlosNZ (approved):
LGTM, works nicely. 

Tiny suggesting about the comment, but I'm not sure mine is any clearer, so up to you :)

PR #7760, opened 2025-05-15
Update text input debounce to more moderate value

### LINKED ISSUE #7577: Custom age labels in vaccine course edit modal do not retain value
## What went wrong? 😲

Noticing some sort of regression when typing in values in custom age labels when editing a vaccine course

https://github.com/user-attachments/assets/21decf4f-b1e9-4441-9fe4-5377af81fcdc

This does not happen when deleting values, just when entering?

## Expected behaviour 🤔

Should be entered smoothly with no disappearances

## How to Reproduce 🔨

1. Have Vaccination Card/Immunization Programs setup in your store
2. Go to Programs -> Immunizations and choose a program to test with (if none are visible, go through configuration steps in mSupply)
3. Once selected, add a vaccine course - you will see a detailed edit modal where you can add doses
4. Add a few doses to test with (I have 12 for example)
5. Enter all the details - by default custom age label will be blank
6. Go through and enter values like I did in the video - see the values disappearing

## Your environment 🌱

- Open mSupply Version: V2.7.1-RC1

- Database type: PostgreSQL

fergie-nz:
Fixes #7577

Adds custom debounce prop to text input cell so can manually change.

In this case (as not connected to API calls), have set to 0.

Wondered about making it 0 for all textinputCell, but wasn't sure if somewhere it was connected to an API call which might cause some issues, so just adding a custom prop.

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

- [x] Navigate to vaccine courses
- [x] Open edit vaccine course modal with a number of doses
- [x] Attempt to edit many course labels in quick succession (as seen on the issue)
- [x] See that UI is more response to catch all changes.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [x] Postgres
- [x] SQLite
- [x] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/columns/types.ts
+++ b/client/packages/common/src/ui/layout/tables/columns/types.ts
@@ -18,6 +18,7 @@ export interface CellProps<T extends RecordWithId> {
   autocompleteName?: string;
   localisedText: TypedTFunction<LocaleKey>;
   localisedDate: (date: string | number | Date) => string;
+  debounceTime?: number;
 }
 
 export interface HeaderProps<T extends RecordWithId> {
--- a/client/packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
@@ -18,7 +18,7 @@ export const TextInputCell = <T extends RecordWithId>({
   CellProps<T>
 > => {
   const [buffer, setBuffer] = useBufferState(column.accessor({ rowData }));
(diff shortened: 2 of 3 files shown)


zachariah-at-msupply (approved):
Nice work! Does the job perfectly. 
Interestingly, the first version of the fix (the one without the delay prop) was noticeably more responsive. 
Both solved the issue anyway, so LGTM 👍

CarlosNZ (approved):
LGTM, thanks @fergie-nz .

We should probably make a TextCell that doesn't use the debounce callback at all at some point, but this works for now.

PR #7768, opened 2025-05-15
Use local keystore sync url not the requests

### LINKED ISSUE #7772: Sync v6 auth using wrong url
## What went wrong? 😲

https://github.com/msupply-foundation/open-msupply-internal/issues/79

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

Chris-Petty:
Fixes #7772

# 👩🏻‍💻 What does this PR do?

Munge the sync_settings data with the url from the keystore in the central server DB. Without doing anything clever this is the way to do it I think.

Should stop incorrect auth server occurring

- I am pretty sure I found everywhere... perhaps there should be some middleware or something that does this.
- This adds an extra DB call on most endpoints. Probably negligible considering we're making an HTTP request to anther server to authenticate. One day we'll have the site table in OMS and not have to make that HTTP call. So performance I'm not that bother making this better.
- There should probably be tests? Integration tests?
- I touched a patient name store join API not 100% sure how to test off the top of my head. I have not tested that API.

# 🧪 Testing

1. Have 2 copies of OG central running (A and B) with webservers on different ports. They should be copies of the same datafile with same sync sites. Both should have same OMS central site configured with same URL.
2. OMS Central syncing with OGC A
3. Initialise your remote site with OGC B - it should fail on v6 requests

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/sync_on_central/mod.rs
+++ b/server/service/src/sync/sync_on_central/mod.rs
@@ -5,8 +5,8 @@ use std::{
 
 use actix_multipart::form::tempfile::TempFile;
 use repository::{
-    ChangelogRepository, KeyType, KeyValueStoreRepository, SyncBufferRowRepository,
-    SyncFileReferenceRow, SyncFileReferenceRowRepository,
+    ChangelogRepository, SyncBufferRowRepository, SyncFileReferenceRow,
+    SyncFileReferenceRowRepository,
 };
 use util::format_error;
 
@@ -16,7 +16,7 @@ use crate::{
     settings::Settings,
     static_files::{StaticFile, StaticFileCategory, StaticFileService},
     sync::{
-        api::{validate_site_auth, CommonSyncRecord, SyncApiSettings, SyncApiV5},
+        api::{validate_site_auth, CommonSyncRecord},
         api_v6::SiteStatusV6,
         synchroniser::integrate_and_translate_sync_buffer,
         translations::ToSyncRecordTranslationType,
--- a/server/service/src/apis/api_on_central/mod.rs
+++ b/server/service/src/apis/api_on_central/mod.rs
@@ -1,12 +1,9 @@
-use repository::{KeyType, KeyValueStoreRepository, RepositoryError};
+use repository::RepositoryError;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use util::format_error;
 
-use crate::{
-    service_provider::{ServiceContext, ServiceProvider},
-    sync::api::{SyncApiSettings, SyncApiV5},
-};
+use crate::service_provider::ServiceProvider;
 
 mod patient_name_store_join;
 


server/service/src/sync/sync_on_central/mod.rs
```diff
@@ -62,12 +62,10 @@ pub async fn pull(
...
 
-    // Check credentials again mSupply central server
-    let response = SyncApiV5::new(sync_v5_settings)
-        .map_err(|e| Error::OtherServerError(format_error(&e)))?
-        .get_site_info()
+    let ctx = service_provider.basic_context()?;
+    let response = validate_site_auth(&ctx, &sync_v5_settings)
         .await
-        .map_err(Error::from)?;
+        .map_err(|e| Error::OtherServerError(format_error(&e)))?;
```
Chris-Petty, line 68: Maybe this func should also return `Result<SyncBatchV6, CentralApiError>` rather than `Result<SyncBatchV6, SyncParsedErrorV6>`. Then it'd match `add_patient_name_store_join` returning `Result<_, CentralApiError>`, or vice versa. `validate_site_auth` can only do one or the other?

server/service/src/sync/api/core.rs
```diff
@@ -238,6 +242,30 @@ async fn response_or_err(
     Err(SyncApiErrorVariantV5::from_response_and_status(response.status(), response).await)
 }
 
+// OMS Central does not yet do auth validation for site credentials
+// So we call Legacy central server for this
+// (Use sync API for simplest auth)
+pub async fn validate_site_auth(
```
Chris-Petty, line 248: This little refactor as one commit probably makes these changes hard to cherry pick into old releases' patches. Oh well...

server/service/src/sync/api/core.rs
```diff
@@ -238,6 +242,30 @@ async fn response_or_err(
...
+    ctx: &ServiceContext,
+    sync_v5_settings: &SyncApiSettings,
+) -> Result<SiteInfoV5, CentralApiError> {
+    // We need to ignore the OG server URL provided by the remote and ensure we use the one that the OMS central server is expecting
+    let kv_repo = KeyValueStoreRepository::new(&ctx.connection);
+    let kv_url = kv_repo.get_string(KeyType::SettingsSyncUrl)?.unwrap();
```
Chris-Petty, line 254: This unwrap is probably illegal yes?

server/service/src/sync/api/core.rs
```diff
@@ -238,6 +242,30 @@ async fn response_or_err(
...
+    let sync_v5_settings = sync_v5_settings.clone();
```
Chris-Petty, line 255: A cool kid probably doesn't even do this clone()

server/service/src/sync/api/core.rs
jmbrunskill: I think so but I prefer to leave a comment with justification when using an unwrap...

jmbrunskill (commented):
Looks good to me Chris, but haven't actually tested it out, will leave other reviews to do that.

server/service/src/sync/api/core.rs
Chris-Petty: With renewed energy I have just handled it properly by returning an error. No panics!

server/service/src/sync/sync_on_central/mod.rs
andreievg: Great point, can't wait till v7 where we don't need to stack errors like this : ), a bit hard to know why it was done as CentralApiError now ¯\_(ツ)_/¯

server/service/src/sync/api/core.rs
andreievg: That's ok, i like the refactor, plus this will be the only version it goes to, I've rebased to 2.7.3 (as that's the next release, coming hot)

andreievg (approved):
Very good, thank you !

I tested by:

* Initialising central 
* Change port of OG 
* Try initialise remote (didn't work, connection errors, but worked in main 👍 )

PR #7773, opened 2025-05-15
Fix GraphiQL (again!)

CarlosNZ:
Looks like a mismatch between the GraphiQL version and the explorer plugin (which was updated 2 days ago).

# 👩🏻‍💻 What does this PR do?

I've fixed the version for both the main package and the exporer plugin (previously on did the main package), which should remain stable now.

# 🧪 Testing

Launch server, go to `http://localhost:8000/graphql` and check that it all works correctly

### DIFF
--- a/server/graphql/playground.html
+++ b/server/graphql/playground.html
@@ -20,8 +20,8 @@
     }
   </style>
 
-  <link rel="stylesheet" href="https://unpkg.com/graphiql@4.0.2/graphiql.min.css" />
-  <link rel="stylesheet" href="https://unpkg.com/@graphiql/plugin-explorer/dist/style.css" />
+  <link rel="stylesheet" href="https://unpkg.com/graphiql@4.0.5/graphiql.min.css" />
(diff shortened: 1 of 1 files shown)


server/graphql/playground.html
```diff
@@ -20,8 +20,8 @@
     }
   </style>
 
-  <link rel="stylesheet" href="https://unpkg.com/graphiql@4.0.2/graphiql.min.css" />
-  <link rel="stylesheet" href="https://unpkg.com/@graphiql/plugin-explorer/dist/style.css" />
+  <link rel="stylesheet" href="https://unpkg.com/graphiql@4.0.5/graphiql.min.css" />
+  <link rel="stylesheet" href="https://unpkg.com/@graphiql/plugin-explorer@4.0.4/dist/style.css" />
```
CarlosNZ, line 24: This is the latest version of both of these, and they seem to play nicely together

jmbrunskill (approved):
Thanks once again. Looks like everything is on fixed versions now

PR #7777, opened 2025-05-16
7776 Add plugin processor

### LINKED ISSUE #7776: Create plugin processor

DHIS2 integration of prescription requires a processor that can execute plugin function

Add a general processor that can use plugins for processing

andreievg:
Fixes #7776

# 👩🏻‍💻 What does this PR do?

* Added a new general processor that can deal with plugin processors, this required some changes to processor traits, thus other processors have changed slightly
* Added dynamic cursor, this is for plugins mainly, still key values store but it's stored in text field as json, extended cursor controller to deal with one or the other
* Implemented a plugin processor for general processors
* Add new processor type
* Added example here: https://github.com/msupply-foundation/open-msupply-plugins/pull/15

Currently there is no way for the same plugin bundle to implement multiple processors, however we can work around this by having one processor deal with multiple log types.

There is some backward compatibility issue with new plugin type existing on central but not on remote, added to main issue.

# 🧪 Testing

Add plugin to your running server:
```
rm -rf .gitmodules && rm -rf client/packages/plugins/myPluginBundle && rm -rf .git/modules && git rm --cached -r client/packages/plugins/myPluginBundle
```
```
git submodule add https://github.com/msupply-foundation/open-msupply-plugins.git client/packages/plugins/myPluginBundle
```

check out to the correct branch in submodule

```
cargo run --bin remote_server_cli -- generate-and-install-plugin-bundle -i '../client/packages/plugins/myPluginBundle/backend' --url 'http://localhost:8000' --username 'Admin' --password pass
```

run sync and see logs for invoices

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/plugins/backendCommon/generated/PluginTypes.ts
```diff
@@ -1,13 +1,28 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
-import type { AverageMonthlyConsumptionInput } from "./AverageMonthlyConsumptionInput";
```
andreievg, line 2: not sure why there is formatting changes

client/packages/plugins/backendCommon/types.ts
```diff
@@ -12,6 +12,9 @@ export type BackendPlugins = {
   graphql_query?: (
     _: PluginTypes['graphql_query']['input']
   ) => PluginTypes['graphql_query']['output'];
+  processor?: (
```
andreievg, line 15: This is the only type file i touched after generating types from backend

server/repository/src/db_diesel/backend_plugin_row.rs
```diff
@@ -13,6 +13,8 @@ pub enum PluginType {
     AverageMonthlyConsumption,
     TransformRequestRequisitionLines,
     GraphqlQuery,
+    // TODO backwards compatibility ? When integrating this one via sync
```
andreievg, line 16: There is a bulletpoint for this in the epic

server/repository/src/db_diesel/changelog/changelog.rs
```diff
@@ -230,14 +231,21 @@ pub struct ChangelogRow {
     pub source_site_id: Option<i32>,
 }
 
-#[derive(Default, Clone)]
+#[derive(Default, Clone, Serialize, Deserialize, Debug, TS)]
 pub struct ChangelogFilter {
+    #[ts(optional)]
```
andreievg, line 236: I was hoping this would be automatic, maybe it is but i can't find it in ts-rs docs

server/service/Cargo.toml
```diff
@@ -29,6 +29,7 @@ url = "2.5.1"
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
+strum = { workspace = true }
```
andreievg, line 32: this is needed for Display in general_processor

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -80,29 +81,31 @@ pub struct PluginBundle {
 
 impl PluginInstance {
     pub fn get_one(r#type: PluginType) -> Option<Arc<PluginInstance>> {
-        let plugin_instance = {
```
andreievg, line 83: Sorry slight refactor here to not use scope, since function already scoped

server/service/src/backend_plugin/types/mod.rs
```diff
@@ -33,6 +35,7 @@ mod generate_typescript_types {
 
     #[test]
     fn export_plugin_typescript() {
-        PluginTypes::export_all_to("../../client/packages/plugins/backendTypes/generated").unwrap();
+        PluginTypes::export_all_to("../../client/packages/plugins/backendCommon/generated")
```
andreievg, line 38: hmm somehow i missed this, it should go to backendCommon

server/service/src/backend_plugin/types/processor.rs
```diff
@@ -0,0 +1,37 @@
...
+use repository::{ChangelogFilter, ChangelogRow, PluginType};
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+fn plugin_type() -> PluginType {
+    PluginType::Processor
+}
+
+#[derive(TS, Clone, Deserialize, Debug, Serialize)]
+#[serde(tag = "t", content = "v")]
```
andreievg, line 12: See ProcessorOutput in generated to see why t and v

server/service/src/boajs/methods/log.rs
```diff
@@ -7,7 +7,10 @@ pub(crate) fn bind_method(context: &mut Context) -> Result<(), JsError> {
         JsString::from("log"),
         0,
         NativeFunction::from_copy_closure(move |_, args, _| {
-            info!("from js {:#?}", args);
+            for arg in args {
```
andreievg, line 10: This is much better log

server/service/src/processors/general_processor.rs
```diff
@@ -1,12 +1,18 @@
...
+    ChangelogFilter, ChangelogRepository, ChangelogRow, ChangelogTableName, EqualFilter,
+    PluginType, RepositoryError, TransactionError,
 };
+use strum::Display;
 use thiserror::Error;
 
 use crate::{
-    cursor_controller::CursorController,
+    backend_plugin::{
+        plugin_provider::{PluginError, PluginInstance},
```
andreievg, line 12: quite a few changes in this file, main change, is being able to processor multiple processor for one type (for plugins), this needed a vec to be returned

andreievg: Probably best to view diff ignoring whitespace

server/service/src/cursor_controller.rs
```diff
@@ -18,6 +41,95 @@ impl CursorController {
...
+    ) -> Result<(), RepositoryError> {
+        KeyValueStoreRepository::new(connection).set_i32(key_type.clone(), Some(cursor as i32))
+    }
+    fn update_dynamic(
+        &self,
+        connection: &StorageConnection,
+        cursor_id: &str,
+        cursor: u64,
+    ) -> Result<(), RepositoryError> {
+        let mut json_text = KeyValueStoreRepository::new(connection)
```
jmbrunskill, line 63: I wonder if this should be in a transaction with update lock on postgres?
To avoid potential for two processors to have a race condition with the update?
Feels like low risk for now, happy to make it a follow up issue?

jmbrunskill (approved):
Thanks Andrei, haven't tested it out much but seems good.

Not a big fan of having the processors as a `vec` for all the non-plugin ones.
I feel like there's probably another solution the but haven't taken the time to play around with that.

I also think it would be nice if our Key Value Store actually allowed Keys and Values rather than just Enum Value to avoid the need for the special general cursor code.

That feels like quite a big refactor to fix so I'm ok with the solution for now, although I think having multiple plugin processors could end up with data race issues...

My one other question I had was if we wanted some mechanism to start the cursor from the time the plugin was installed rather than processing every changelog in history. I guess it's pretty fast so it shouldn't really matter but still, might be quite slow in a big site?

jmbrunskill (changes requested):
Looks like postgres tests are failing

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -2,6 +2,7 @@ use std::sync::{Arc, RwLock};
 
 use base64::{prelude::BASE64_STANDARD, Engine};
 
+use log::info;
```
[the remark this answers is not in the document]
andreievg: Thanks this is gone now in on of the commits

server/service/src/cursor_controller.rs
andreievg: Good idea thanks: dc86a2c5871aabcff656332b2865b846698e1576

andreievg:
> Not a big fan of having the processors as a vec for all the non-plugin ones.
> I feel like there's probably another solution the but haven't taken the time to play around with that.

@jmbrunskill, yes agree, it does not look ideal, to me it's fine though, but I wrote it.

I added a comment 412386631277f578fa49060c6fa4daf5622c141f

> I also think it would be nice if our Key Value Store actually allowed Keys and Values rather than just Enum Value to avoid the need for the special general cursor code.

Yep, I would have done that this was going into a release with extensive testing.

> That feels like quite a big refactor to fix so I'm ok with the solution for now, although I think having multiple plugin processors could end up with data race issues...

Even though processors run async to the rest of the code, they should still run in a single thread, so unless those cursors are edited somewhere else, it's pretty safe

> Looks like postgres tests are failing

Fixed, thanks 05d9f1a80ab87b080f3ad25c1b42521fdb8ffe82

> My one other question I had was if we wanted some mechanism to start the cursor from the time the plugin was installed rather than processing every changelog in history. I guess it's pretty fast so it shouldn't really matter but still, might be quite slow in a big site?

Hmm no mechanism but can probably be done with plugin data (store in plugin data latest cursor if plugin data does not yet exist) ?

andreievg:
Thanks for review @jmbrunskill will merge as actions all run

andreievg:
There is a front end linter action that is failing, [looks like the error were fixed in develop](https://github.com/msupply-foundation/open-msupply/pull/7679#issuecomment-2870192407) but not in 2.7.2 (technically first they need to be fixed in a patch branch). 

I hope it's safe to merge

jmbrunskill (approved):
Thanks for the updates, looks good Andrei.
I agree it looks safe to merge.

Issue #7641, opened 2025-05-07
Allocate batches based on VM Status then Expiry


zachariah-at-msupply:
## Is your feature request related to a problem? Please describe 👀

When preference `Sort available batches by VVM status rather than expiry` is enabled.
The auto allocate logic for outbound shipment like should order the batches so that they're allocated based on VVM Status first, then expiry date.

`if you had 2 batches that had VVM status 2, then they should issue the one expiring first`

There is no need to update stock view or stocktake thanks.

## Describe the solution you'd like 🎁
The intention of this is that when you are issuing stock, the normal process is that the stock lines which are closest to expiring are issued first. We do this when auto-allocating stock lines, and also display the lines in increasing expiry date order (all in outbound shipment; when selecting a specific batch for an item)

The change here, is that if you are using VVM status, then most stores will want to issue stock that is in status 2 first! this stock is still usable, but has been exposed to temperatures outside of the desired range. Stores will want to issue that stock regardless of when the expiry date is.

```
When issuing stock, this will be prioritised based on Level 1 to 4 on mSupply. Level 1 will be prioritised first, then level 2 and so forth. You have the flexibility to customize your VVM stages to determine which should be prioritized first.

In this example, we can see that Stage 2 has first level priority for being issued:
```
https://docs.msupply.org/cold_chain_equipment:configure

### Moneyworks Jobcode 🧰

NIGGF:VAX

jmbrunskill:
Mark's comment here...
https://github.com/msupply-foundation/open-msupply/issues/7354#issuecomment-2861798604

jmbrunskill:
I'm thinking this would be good for @lache-melvin ...
If anyone else wants to work on it please talk to me first...

alainsussol:
Tested in Vaccine Support Feature Release testing: see #7783 for results

Issue #7367, opened 2025-04-16
VVM: Outbound & Prescription


roxy-dao:
## Tasks 📝
`Able to specify VVM status when receiving items` preference on:
- [ ] Read-only column next to the vials per pack (in Outbound) and batch (in prescription) which shows the current (latest) VVM status in the line edit

### Acceptance Criteria ✅
- [ ] User should be able to see the latest VVM status of a vaccine item if they have the preference on for their store

### Moneyworks Jobcode 🧰
NIGGF:VAX

alainsussol:
Tested in Vaccine Support Feature Release testing: see #7783 for results

PR #7763, opened 2025-05-15
7382 display donor in stocktake

lache-melvin:
Part 1 of #7382

# 👩🏻‍💻 What does this PR do?

Donor in stocktake detail view:
Screenshot 2025-05-15 at 2 41 07 PM

Donor in edit modal - disabled for now, will add edit in sep PR:
Screenshot 2025-05-15 at 2 41 00 PM

Currently, that's on the locations tab, as that was the roomiest table, seemed like donor and comment were on a similar vibe of "other"? maybe a @mark-prins question?

## 💌 Any notes for the reviewer?

# 🧪 Testing

- [ ] Have a stocktake with a donor (easiest way for now is to add a name id into donor_id field directly in DB)
- [ ] Can see donor name in detail list
- [ ] Can see donor in edit modal

# 📃 Documentation

- [x] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #7780, opened 2025-05-16
7641 sort by vvms status outbound

jmbrunskill:
Half of #7641 & #7367 (outbound shipment only)

# 👩🏻‍💻 What does this PR do?

Adds a new column to see VM Status (if either ManageVVM Status or `Sort available batches by VVM status then expiry`) prefs are turned on...

image

If sort is enabled, any lines with usable vvm statuses will be displayed before any lines with expiry.
If they have both vvm status and expiry, then the we'll sort by vvm status first then expiry second.
This order is also used in the allocation logic.

I've only done outbound shipment here as i was able to use the new refactor for allocation. Now need to decide if we can apply this prescription now, of if we re-do some of this logic there. See next PR for the answer...

# 🧪 Testing

- [ ] TBD will fill in soon

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
```diff
@@ -41,14 +43,18 @@ export const Allocation = ({
...
   useEffect(() => {
     initialise(
       itemData,
-      AllocationStrategy.FEFO,
+      sortByVvmStatus ? AllocationStrategy.VVMStatus : AllocationStrategy.FEFO,
       allowPlaceholder,
       scannedBatch
     );
-  }, []);
+  }, [sortByVvmStatus]);
```
jmbrunskill, line 57: Not sure if this is ok to do? It seems to fix the issue where sometimes the sort order wouldn't apply.
I think it's probably not ideal, so do we need to pass it in to this component rather than checking the pref here? I think it's convient here but maybe unintended consequences?

lache-melvin: #7763 added support for multi pref query, at which point, so have put that in the parent component. Then the pref query only runs once, when the modal first opens, rather than for each item as you click next...

Maybe this is still wise though, in case prefs somehow take a long time to resolve? Would just initialise the modal twice 🤷‍♀️

lache-melvin (approved):
LGTM James, thank you!

Added a couple of commits, mostly thanks to other changes gone into develop 🫠 

Namely, can pass many prefs into pref hook now! And updated the sort to use non-expiring last as well 🙏

PR #7794, opened 2025-05-19
7755 write site configs to files

### LINKED ISSUE #7755: Load test CLI

Creating load test of the sync API

- [ ] action on remote_server_cli
- [ ] create sites in OG
- [ ] make their stores visible to each other
- [ ] output server config files for each site
- [ ] start server process for each site, initialising
- [ ] post an inventory adjustment to each site
- [ ] start sync for each site
- [ ] stop after duration
- [ ] measure measure approx records per second (needs more break down)

### Additional context 💌

https://github.com/msupply-foundation/open-msupply-internal/issues/72

### Moneyworks Jobcode 🧰

LBRCRSRPT

Chris-Petty:
Fixes #7755 

# 👩🏻‍💻 What does this PR do?

For each site create the config file for starting OMS instances from CLI. There are probably some details in the settings that ain't quite right, I plan to iron it out when actually using these configs and seeing what blows up.

from `.../open-msupply/server/` run `cargo build && ./target/debug/remote_server_cli load-test -u http://localhost:8888 -s 5 -i 25 -d 100` (substitute your 4D server in `-u`)

Had to sprinkle `Serialize` derive around Settings structs so that `serde_yml` would work. I was pretty liberal with it, may have got an extra struct but it can't hurt much.

Note, there is a similar crate `serde-yml`, but the author (same guy who made serde, serde-json, anyhow...) stopped maintaining it and marked it as deprecated. `serde_yml` is a fork of `serde-yml` that is maintained. 

# 🧪 Testing

Dev tested 

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -2019,6 +2019,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "serde_yml",
  "server",
  "service",
  "shellexpand",
@@ -4658,7 +4659,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.9",
  "tokio",
  "tower-service",
  "tracing",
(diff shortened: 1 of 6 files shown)


Chris-Petty:
There is a possibility I could just set the database and server properties as environment variables when starting each child process rather than generating files...

lache-melvin (changes requested):
> There is a possibility I could just set the database and server properties as environment variables when starting each child process rather than generating files...

Yeah was wondering about this, especially as most are defaults on the base config? would just be port, db name/path, and some sync settings?

Main reason I could think for keeping config around would be if different sites might end up with different config, and we leave it in the out_dir on error to allow us to investigate.... but looks like they'll all be clones of each other anyway 🤷‍♀️

PR #7816, opened 2025-05-20
7786 auto allocate in packs from outbound shipment

### LINKED ISSUE #7786: Auto allocate in packs

Allocation refactor #7712 removed auto-allocation in packs - I can only issue in units.

But if I'm looking at an item where pack size is 1000, and I want to ship 100 packs... I'd have to type 100,000?

I know the "any" thing was confusing, but the "if all pack sizes the same, issue in that pack size" probably made sense.
 
Screenshot 2025-05-15 at 8 46 15 AM

Needs some design thought:

- It might be confusing if sometimes the input is for packs, and other times its for units. How to make this clear to user?
- Should "in packs" option only turn on if _every_ line is the same pack size?
- Or should it go back to being a drop down, and we use the item default pack size?
- Should there be some kind of indication to the user, which lines will be used in autoallocation? highlighting?

### Additional context 💌

Likely needed for 2.8!

### Agreed solution

For 2.8, agreed with @DhanyaHerath, there are a lot of options, but we don't have time to explore and validate them. We shouldn't change behaviour too much without good reason, as it requires re-training.

So:
- Keep `Units` (or the unit word e.g. `Tablets`) instead of "any", and have this option always available. (this has been existing source of confusion)
- Have a dropdown of the available pack sizes, like the old pack picker (but maybe say "Packs of 10" or something in the dropdown to be a bit clearer)
- Let `Doses` also be an option in the dropdown for vaccine items, rather than the only way to view the modal
- If you change the pack size, clear allocated quantities

2.7:
Image

lache-melvin:
Fixes #7786

# 👩🏻‍💻 What does this PR do?

Adds "Allocate In" selector, for pack sizes/units/doses (if pref on and is vaccine item):
Screenshot 2025-05-20 at 4 53 19 PM

Also adds column to show the lines we will allocate from... thoughts?:
Screenshot 2025-05-20 at 4 52 26 PM

I guess, usability thoughts? Especially - the clearing on change, and the fact that if you reopen the modal, its back to units. What's the best fast thing we can do for now, until we get more feedback?

# 🧪 Testing

- [ ] Modal always opens in units
- [ ] When pref is on and item is vaccine item, you can switch to doses lens (columns change etc.)
- [ ] All available (non expired, non on-hold...) pack sizes are available in the selector
- [ ] Switching pack size, you can see which lines will be allocated from (and they are of the right pack size)
- [ ] All available lines should be allocated from when in doses or units
- [ ] When switching between pack sizes etc. your currently allocated quantity attempts to re-allocated to lines of the selected pack size (placeholder created/updated accordingly)
- [ ] When switching to doses or units, the allocated rows/num packs don't change 
- [ ] Do some allocating - all the totals, placeholders etc. look right in all the different lenses 👀 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -36,15 +39,24 @@ const TotalCell = styled(TableCell)({
...
   const t = useTranslation();
 
   const formattedValue = useFormatNumber().round(quantity ?? 0, 2);
 
   // TODO - maybe should be editable? Can't clear when manually allocating..
   return quantity === null ? null : (
     <tr>
-      <PlaceholderCell colSpan={3} sx={{ color: 'secondary.main' }}>
+      <PlaceholderCell
+        colSpan={4 + extraColumnOffset}
```
lache-melvin, line 57: we're added a bunch of extra columns to the outbound edit modal... this was best way I could think to bump the placeholder/total values over enough. We'll just add to `extraColumnOffset` below based on prefs? Would be better if these values actually just rendered themselves inside the defined columns though? Maybe better as a refactor issue...

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -112,15 +116,10 @@ export const useAllocationContext = create<AllocationContext>((set, get) => ({
       isDirty: false,
       item,
 
-      allocateIn:
-        item.isVaccine && allocateVaccineItemsInDoses
-          ? AllocateIn.Doses
-          : AllocateIn.Units,
```
lache-melvin, line 118: Won't initialise in doses (post convo with dhanya - if you're doing shipments, you're probably thinking in units/packs) but doses option will be available in the dropdown

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/allocation/useAllocationContext.ts
```diff
@@ -131,23 +130,46 @@ export const useAllocationContext = create<AllocationContext>((set, get) => ({
...
       alerts: [],
     })),
 
+  setAllocateIn: (allocateIn: AllocateInOption) => {
+    const { draftLines } = get();
+
+    // Clear the existing allocated stock if changing how we are allocating
+    // This might be annoying... but should validate before spending time on
+    // reallocating on change of allocation type/pack size
+    const result = allocateQuantities(draftLines, 0, { allocateIn });
```
lache-melvin, line 146: Maybe we should move the existing allocated quantity to the placeholder value or something??

GeronimoJohn (approved):
Thanks for the hard work @lache-melvin 🔥. LGTM — happy to approve 😄.  

Just a few usability thoughts that came to mind:

- Part of me was hoping the selected pack size would stick after closing the modal.
- It felt a bit inconvenient having to retype in the issue input after changing the pack size (though makes sense).
- I instinctively hovered over the check icon to figure out what it meant, rather than the info icon.

Side note: 
Also `useAllocationContext` is 🔥 🔥  🔥 , and comments are well written was able to follow 😄

lache-melvin:
> * Part of me was hoping the selected pack size would stick after closing the modal.

Agree... but apparently most places think in units! So I'm thinking lets wait for feedback before putting anything in for this...

> * It felt a bit inconvenient having to retype in the issue input after changing the pack size (but makes sense).

Made a change to try re-allocate on change pack size, and not to clear if just changing to units. What do you think? [5f6b764](https://github.com/msupply-foundation/open-msupply/pull/7816/commits/5f6b7640d3e529c91706bcda575235810a8f58f2)

> * I instinctively hovered over the check icon to figure out what it meant, rather than the info icon.

Oo nice, updated: [d902daa](https://github.com/msupply-foundation/open-msupply/pull/7816/commits/d902daac6bf268d79bc4eb6d7cf3473f1f9d225c)

GeronimoJohn (approved):
LGTM! Thanks for the changes Lache 🔥 🚀

PR #7817, opened 2025-05-20
7804 add campaign to stock line schema

### LINKED ISSUE #7804: Add campaign links to stock line schema (including OG sync)

GeronimoJohn:
Fixes #7804 

# 👩🏻‍💻 What does this PR do?

Added `campaign_id` to be a part of the Stock Line schema. 

Added a loader for the times that we would need to render information about the Campaign in a Stock Line. Whether if we need or not? Not sure. Can easily just remove those changes

# 🧪 Testing

_**Backend Changes Only**_
Need to insert through database (SQL) a campaign to a Stock Line
- [ ] Query a Stock Line with Campaign
- [ ] Should return campaign data within a Stock Line

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/invoice/inbound_shipment/update/generate.rs
```diff
@@ -370,6 +370,7 @@ pub fn generate_lines_and_stock_lines(
                 item_variant_id,
                 donor_id,
                 vvm_status_id: None,
+                campaign_id: None,
```
GeronimoJohn, line 373: I've kept the PR small for now, leaving updates to `campaign_id` to be handled by the relevant issues that need to manipulate its value. Unless we think otherwise - would be more than glad to add it to this PR.

The same applies to any other additions or edits involving `campaign_id`.

server/graphql/core/src/loader/campaign.rs
```diff
@@ -0,0 +1,28 @@
...
+    campaign_row::{CampaignRow, CampaignRowRepository},
+    RepositoryError, StorageConnectionManager,
+};
+use std::collections::HashMap;
+
+pub struct CampaignByIdLoader {
+    pub connection_manager: StorageConnectionManager,
+}
+
+impl Loader<String> for CampaignByIdLoader {
```
GeronimoJohn, line 13: After some discussions and additional input, we've decided to add a `CampaignByIdLoader`, which can be used as needed (currently only used in Stock Line).

That said, while implementing it, I just couldn't visualise where we would render this information. If we decide it's unnecessary, I'm happy to remove it.

server/service/src/sync/translations/stock_line.rs
```diff
@@ -128,6 +128,7 @@ impl SyncTranslation for StockLineTranslation {
             item_variant_id,
             donor_id,
             vvm_status_id,
+            campaign_id: None,
```
jmbrunskill, line 131: Can you create a follow up issue for sync, or implement sync here?
We also need to create an issue in the mSupply repo to create the campaign_id field so that it can be synced

jmbrunskill (changes requested):
Let's wait for tests to pass (which should in base PR)

I think we should probably implement the sync part in anticipation of the mSupply field being added here, or otherwise please create a follow up issue so that's not forgotten?

jmbrunskill (approved):
Looks great thanks John

PR #7826, opened 2025-05-21
Add front-end API

CarlosNZ:
Fixes #7803 (continued)

# 👩🏻‍💻 What does this PR do?

Start of Campaign Admin panel, but just publishing API hook as separate PR so others can use it.

## 💌 Any notes for the reviewer?

Please note that I have made additional refinements in https://github.com/msupply-foundation/open-msupply/pull/7829, so if there's any changes required in here they may be already addressed. If it's easier I can merge these two PRs together for review?

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #7829, opened 2025-05-21
#7805 Campaign Editor

### LINKED ISSUE #7805: UI for Campaign editor (Admin)

CarlosNZ:
Fixes #7805 

# 👩🏻‍💻 What does this PR do?

Implements a (very) basic list viewer and editor modal for "Campaigns". 

Go to "Manage" -> "Campaigns" in Menu (must be central server)

You can:
- see a list of campaigns
- click a campaign to edit it in a modal
- add a new campaign with a button
- delete campaigns by selecting rows

Screenshot 2025-05-21 at 4 50 22 PM

Screenshot 2025-05-21 at 4 51 35 PM

Screenshot 2025-05-21 at 4 50 55 PM

Note that this makes a few changes/improvements to the front-end hooks done in https://github.com/msupply-foundation/open-msupply/pull/7826

See inline comments

# 🧪 Testing

Need to be Central Server, then go to Manage -> Campaigns

- [ ] Can add new campaigns
- [ ] Can see added campaigns in the list
- [ ] Can edit existing campaigns
- [ ] Can delete campaigns
- [ ] Gives an error when you try and create a new campaign with the same name as an existing one
- [ ] Check date functionality:
  - [ ] Can delete a date (i.e. save with an empty date field)
  - [ ] Can change a date
  - [ ] Gives an error when trying to save with invalid dates (i.e. Start is later than End)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Manage/Campaigns/ListView/ListView.tsx
```diff
@@ -0,0 +1,149 @@
...
+      )();
+  };
+
+  const confirmAndDelete = useDeleteConfirmation({
+    selectedRows,
+    deleteAction: async () =>
+      selectedRows.forEach(row => {
+        if (row) {
+          deleteCampaign(row.id);
+        }
```
CarlosNZ, line 70: I don't like doing a separate mutation for each deletion, but the current GraphQL Delete endpoint only accepts a single ID. We should probably change this to accept an array of IDs like some of the other ones. New issue you reckon?

server/service/src/campaign/upsert.rs
```diff
@@ -74,8 +74,11 @@ fn validate(
     input: &UpsertCampaign,
 ) -> Result<(), UpsertCampaignError> {
     // Check for duplicate name
-    let campaigns_with_duplicate_name = CampaignRepository::new(connection)
-        .query_by_filter(CampaignFilter::new().name(StringFilter::equal_to(input.name.trim())))?;
+    let campaigns_with_duplicate_name = CampaignRepository::new(connection).query_by_filter(
+        CampaignFilter::new()
+            .name(StringFilter::equal_to(input.name.trim()))
+            .id(EqualFilter::not_equal_to(&input.id)),
+    )?;
```
CarlosNZ, line 81: This was causing a problem -- it would give an error when updating and exiting row if the name was the same, so I added additional clause to only give this error when *inserting* a new campaign, not updating an existing one.

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -315,6 +317,7 @@
   "error.no-asset-log-reasons": "There are no asset log reasons to display.",
   "error.no-asset-view-permission": "You do not have permission to view assets.",
   "error.no-assets": "There are no assets to display.",
+  "error.no-campaigns": "No campaigns to display",
```
fergie-nz, line 320: Nit: change this to "There are no campaigns to display" for consistency with similar error messages

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1449,6 +1454,10 @@
   "messages.confirm-delete-assets_one": "This will permanently remove this asset.",
   "messages.confirm-delete-assets_other": "This will permanently remove the {{count}} selected assets.",
   "messages.confirm-delete-bundled-item": "This will remove this item from the bundle",
+  "messages.confirm-delete-campaigns_few": "This will permanently remove {{count}} campaign",
+  "messages.confirm-delete-campaigns_many": "This will permanently remove {{count}} campaigns",
+  "messages.confirm-delete-campaigns_one": "This will permanently remove 1 campaign",
+  "messages.confirm-delete-campaigns_other": "This will permanently remove {{count}} campaigns",
```
fergie-nz, line 1460: I have a tangent thought that it would be awesome to have a second dynamic value so we don't need so many similar delete confirmation messages

ie:
`This will permanently remove {{count}} {{name}}`

client/packages/system/src/Manage/Campaigns/ListView/CampaignEditModal.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <DialogButton
+          variant="ok"
+          onClick={async () => {
+            await upsert();
+            onClose();
+            updateDraft(defaultDraftCampaign);
+          }}
+        />
+      }
+      height={1000}
```
fergie-nz, line 58: nit again but could make this smaller as full screen height feels unnecessary to me:

Screenshot 2025-05-22 at 2 15 15 PM

client/packages/system/src/Manage/Campaigns/ListView/ListView.tsx
fergie-nz: Seems very reasonable to me to loop through for now. I can't imagine mass campaign deleting occurring other than from testing?

@jmbrunskill do you have any thoughts?

fergie-nz: Actually seems fine - we use the same looping in a number of other parts of the app (asset delete, or returns rows for example). And this seems less likely to be an area of the app with high traffic.

fergie-nz: However this will give a confirmation of all successfully deleted if one or more of the rows are successfully deleted.


https://github.com/user-attachments/assets/dcdca105-9b1d-47fe-b335-7cbfcf52806b

But I think you can capture success or failure in bulk with the bulk delete patterns like:

```
  const deleteAction = async () => {
    await Promise.all(selectedRows.map(row => mutateAsync(row.id)))
      .then(() => queryClient.invalidateQueries(api.keys.base()))
      .catch(err => {
        throw err;
      });
  };
  ```

fergie-nz (changes requested):
Looks awesome Carl.

Just one change on awaiting success of all deletions.

The other things are UI or naming nits which is just my preference.

client/packages/common/src/intl/locales/en/common.json
CarlosNZ: That's a cool idea, it seemed pretty wasteful to me too, but probably beyond the scope of this issue as I'd need to change it in all the places that use it. Do you want to make a re-factor issue to address this more widely?

client/packages/system/src/Manage/Campaigns/ListView/CampaignEditModal.tsx
CarlosNZ: 100%, this was leftover from the place I copy-pasted it from 😉

Looks much better now: 

Screenshot 2025-05-23 at 2 27 02 PM

Good catch 👍

CarlosNZ:
Thanks for the suggeestions, @fergie-nz, should be all taken care of now :)

client/packages/common/src/intl/locales/en/common.json
fergie-nz: No worries I can make the issue and see if it gains traction.

Did not mean to imply should be added in this issue - was just jotting down my thoughts :)

fergie-nz (approved):
Looks great!

PR #7832, opened 2025-05-21
7831 Expose graphql and repository to plugins

### LINKED ISSUE #7831: Expose graphql and parts of repository to plugins

In order for plugins to create a prescription from a message they need to query and modify a message. They will also need to access service or graphql (to make it easier and safer for plugin to create stock movements, prescriptions etc..), rather then going directly through repository

Expose use_repository to plugin (we can add gradually add repositories here)
Expose use_graphql to plugin, at some point soon plugins would need to validate compatibility against the latest tag.

Expose service rather then graphql, I had a look at batch mutation service it would require a lot of serialize, deserialize and TS derives now and in the future, exposing graphql seems like a more flexible path forward. Also would need to run 'async' code from plugin at some point, so creating a path forward for this would be very useful

andreievg:
Sorry branch name is different to issue

Fixes #7831

This PR is branches from combination of 7809 and 7776, with both of them merged together, will rebase when they are merged

# 👩🏻‍💻 What does this PR do?

* Exposes use_repository for updating and query for messages through repository, there is no service for this one and sometimes it's useful to access repository directly
* Expose use_graphql for access to graphql via plugin, see issue for justification of why this was chosen instead of exposing service
    * There was a big challenge with use_graphql because it's async method that needs to run inside a sync method, a lot of the code is to make this work
    * Changed runtime to tokio to allow for tokio multi thread rather then actix_web modification of this, still actix_web server will run in it's own way, and calling block_in_place and block_on does not work when running within actix_web async context, thus there are two ways to handle blocking method within sync method, there is description of this in code.
    * From my understanding using tokio allow work 'stealing' and tokio is good at spreading tasks across threads, axtix_web avoid swapping tasks between threads to allow for performance gains, this change might seem risky for a patch. Is this change risky for a patch ? According to [this](https://docs.rs/actix-web/latest/actix_web/rt/index.html#running-actix-web-using-tokiomain) it is fine, unless we start using websockets, at this points need to consider modifying this code or using alternative websocket library.
    * Also change loaders to use tokio, which seems to be more appropriate for this tokio/actix runtime.
* Modifies message translations to use serde_json::Value for body
* Adding graphql needed auth, so added user override to RequestUserData <- ideally this would be enum, defined in service and related service method would use it rather then another optional field in RequestUserData

There is a branch in omsupply plugins repository, with link to this PR in readme, also updated main readme to include link to that branch

There was a panic during development that cause the whole app to crash (because it was in processor), due to undefined [not being implemented in boajs](https://github.com/boa-dev/boa/blob/c863b8cdcac6cee4f5feb9683346805d87f351b9/core/engine/src/value/conversions/serde_json.rs#L128) json conversion, [it's fixed here](https://github.com/boa-dev/boa/blob/51d604021d6634af904e3152eb8b06383627ca43/core/engine/src/value/conversions/serde_json.rs#L127), but not in a tagged release yet

The blocking channel needs a test that saturates threads, i am not fully confident that it will wont' freeze, this will be a problem if we use use_graphql within actix_web context

There is also the need to pass along user_id to use_graphql, when it's available at the time plugin is called

[carry over](https://github.com/msupply-foundation/open-msupply/issues/2951#issuecomment-2650673745)

# 🧪 Testing

I've done a test on a tablet, to make sure the blocking async functionality is ok in low hardware setting and that tokio multithread runtime works fine, seems to be ok, also did a [small load test.](https://github.com/msupply-foundation/open-msupply/compare/Bench-sever-initial...Bench-async-blocking-7831?expand=1)

Load plugin [from this branch](https://github.com/msupply-foundation/open-msupply-plugins/tree/Process-messages-to-create-prescription) (note this plugin is just an example)

Use data file [here](https://drive.google.com/drive/u/1/folders/1RZj20Z2Nh9QARPrQOnYB9H1S-9IeuZb9): 

user1 -> user1 for OG related stuff, like registering etc..
test -> pass for site initialisation
test -> pass for omSupply user

Run this code to create messages, then sync and see prescriptions created:

```
$newMsg:=ds.message.new()
$newMsg.ID:="1"
$newMsg.toStoreID:="572B52B26FB54D4896C740F35BCBC75E"
$newMsg.fromStoreID:=""
$newMsg.body:="{\"itemId\": \"2FC2E930CCD543C883D17BCDD0EB1251\", \"patientId\": \"0196f0ec-e3fe-7745-ace7-a15ad08bf9f9\" }"
$newMsg.createdDate:=!2023-01-01!
$newMsg.createdTime:=?02:03:04?
$newMsg.status:="new"
$newMsg.type:="createPrescription"
$newMsg.save()
```

Increase msg id to create more messages, after prescription is generated and subsequent sync, see message status changed to 'processed' or 'error' when there is not stock etc..

Run this graphql to see use_graphql within a plugin and actix web context

```gql
query MyQuery($input: JSON!) {
  pluginGraphqlQuery(
    input: $input
    pluginCode: "plugin_examples"
    storeId: "572B52B26FB54D4896C740F35BCBC75E"
  )
}
```
```json
{
  "input": {
    "type": "doubleEcho",
    "echo": "echome"
  }
}
```

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/plugins/backendCommon/generated/GraphqlQueryInput.ts
```diff
@@ -1,4 +1,4 @@
 // This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
-import type { JsonValue } from "./serde_json/JsonValue";
+import type { JsonValue } from './serde_json/JsonValue';
```
andreievg, line 2: Sorry not sure why this changed, i must have saved it to fix formatting

client/packages/plugins/backendCommon/types.ts
```diff
@@ -27,4 +27,10 @@ declare global {
   var get_plugin_data: (
     _: PluginTypes['get_plugin_data']['input']
   ) => PluginTypes['get_plugin_data']['output'];
+  var use_repository: (
```
andreievg, line 30: Added manually

client/packages/plugins/myPluginBundle
andreievg, line 1: This was somehow committed, removed with git rm --cached client/packages/plugins/myPluginBundle

server/android/src/android.rs
```diff
@@ -75,7 +75,10 @@ pub mod android {
 
         // run server in background thread
         let thread = thread::spawn(move || {
-            actix_web::rt::System::new()
+            tokio::runtime::Builder::new_multi_thread()
```
andreievg, line 78: Android also needs multi thread

server/graphql/core/src/loader/loader_registry.rs
```diff
@@ -35,224 +35,224 @@ pub async fn get_loaders(
         ItemLoader {
             connection_manager: connection_manager.clone(),
         },
-        async_std::task::spawn,
+        tokio::spawn,
```
andreievg, line 38: I am doing some test on a docker image to see if this makes a difference, may switch to actix_rt::spawn since this will run in actix_web context, will comment further

server/graphql/lib.rs
```diff
@@ -423,3 +424,32 @@ pub fn attach_discovery_graphql_schema(
 async fn discovery_index(schema: Data<DiscoverySchema>, req: GraphQLRequest) -> GraphQLResponse {
     schema.execute(req.into_inner()).await.into()
 }
+
+pub struct PluginExecuteGraphql(pub Data<GraphqlSchema>);
```
andreievg, line 428: We have another executer, for reports, this one is slightly different and needs to override user_id, also this one needs to be available in boajs context

server/server/src/discovery.rs
```diff
@@ -11,7 +11,7 @@ const HARDWARE_ID_KEY: &str = "hardware_id";
 const CLIENT_VERSION: &str = "unspecified";
 
 pub(crate) fn start_discovery(protocol: Protocol, port: u16, hardware_id: String) {
-    tokio::task::spawn(async move {
+    tokio::spawn(async move {
```
andreievg, line 14: This one was not required change, i just wanted to align with other (did a global search for spawn)

server/android/src/android.rs
andreievg: Stop signal still work correctly, tested

server/server/src/lib.rs
```diff
@@ -292,7 +298,7 @@ pub async fn start_server(
     info!("Starting discovery graphql server",);
     let closure_service_provider = service_provider.clone();
     // See attach_discovery_graphql_schema for more details
-    actix_web::rt::spawn(
+    tokio::spawn(
```
andreievg, line 301: Needed this otherwise was getting an error for local spawn

server/service/src/auth.rs
```diff
@@ -961,8 +979,7 @@ impl AuthServiceTrait for AuthService {
         };
 
         Ok(ValidatedUser {
-            user_id: validated_auth.user_id,
-            claims: validated_auth.claims,
```
andreievg, line 965: claims were never used and was confusing me a lot

server/service/src/boajs/methods/get_plugin_data.rs
```diff
@@ -10,7 +10,7 @@ pub(crate) fn bind_method(context: &mut Context) -> Result<(), JsError> {
         NativeFunction::from_copy_closure(move |_, args, mut ctx| {
             let filter: PluginDataFilter = get_serde_argument(&mut ctx, args, 0)?;
 
-            // When using PluginContext, it's best to use 'scope' see PluginContext for a link to testing repo
+            // When using BoaJsContext, it's best to use 'scope' see PluginContext for a link to testing repo
```
andreievg, line 13: It used to be PluginContext, updated comment

server/service/src/boajs/methods/use_graphql.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use crate::{boajs::context::BoaJsContext, boajs::utils::*};
+
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+pub(crate) struct UseGraphqlInput {
+    query: String,
+    variables: serde_json::Value,
+}
+
+// TODO this method should accept user id, when call plugin is called it should accept this user id
```
andreievg, line 14: There is a bulletpoint for this in carry over

server/service/src/boajs/methods/use_repository.rs
```diff
@@ -0,0 +1,60 @@
...
+                let service_provider = BoaJsContext::service_provider();
+                let connection = service_provider
+                    .connection()
+                    .map_err(std_error_to_js_error)?;
+
+                use UseRepositoryInput as In;
+                use UseRepositoryOutput as Out;
+
+                match input {
+                    In::GetMessageById(id) => Out::GetMessageById(
```
andreievg, line 40: Can add more repositories here

server/service/src/boajs/utils.rs
```diff
@@ -64,3 +64,55 @@ pub(super) fn get_serde_argument<D: DeserializeOwned>(
...
+        query: &str,
+        variables: serde_json::Value,
+    ) -> Result<serde_json::Value, ExecuteGraphQlError>;
+}
+
+/// Executes an async function in a blocking manner from within an async or sync context.
+/// Required to run async code, like GraphQL queries, from boa js sync context.
+/// Functionality here allows for two examples, one for actix_web server and one for main runtime.
+/// Can be seen in action in open-msupply-plugins repository, through graphql query and message processor
+pub fn do_async_blocking<T, F>(handle: Arc<Handle>, f: F) -> T
```
andreievg, line 90: I've tested this in many ways, one method doesn't work for both actix_web server and for normal runtime things, like processors, it's quite hard to do a normal unit test for this, however this should be tested within plugin repo against docker build of omSupply, on my todo

server/service/src/sync/synchroniser.rs
```diff
@@ -304,6 +304,8 @@ impl Synchroniser {
         ctx.processors_trigger
             .trigger_processor(ProcessorType::ContactFormEmail);
 
+        // This should be before plugin processor below, in case there is a processor error, need to be able
+        // to sync new plugin version to avoid bricking the app
```
andreievg, line 308: There was a panic, that cause app to quit on startup, after sync, PR description

server/service/src/sync/test/test_data/message.rs
```diff
@@ -19,7 +19,7 @@ const MESSAGE_1: (&str, &str) = (
     "ID": "message1",
     "toStoreID": "store_a",
     "fromStoreID": "store_b",
-    "body": "{\"key\": \"value\"}",
+    "body": {"key": "value"},
```
andreievg, line 22: updated translator to use body as serde_json::Value

server/service/src/sync/translations/message.rs
```diff
@@ -120,6 +123,9 @@ impl SyncTranslation for MessageTranslation {
         let created_date = created_datetime.date();
         let created_time = created_datetime.time();
 
+        // "unwrap_or" here would result in a string version of body json
+        let body = serde_json::from_str(&body).unwrap_or(serde_json::json!(body));
```
andreievg, line 127: I was really confused why to_value did not work on 'body', and tests were failing

standard_reports/item-usage/latest/convert_data_js/codegen.yaml
```diff
@@ -8,10 +8,11 @@ generates:
     preset: import-types
     presetConfig:
       # Because we are using typescript-operations only, some pre requisite types are missing, those are usually generated with typescript plugin
-      # Would need to update codegenTypes if new scalars are added, do this by adding typescript plugin below, generating and coping first few lines to codegenTypes 
+      # Would need to update codegenTypes if new scalars are added, do this by adding typescript plugin below (uncomment), generating and coping first few lines to codegenTypes
```
andreievg, line 11: There is a comment about graphql codegen in the readme of the plugin

lache-melvin (changes requested):
Thanks Andrei, looking great, will check in again once plugin user in place :)

server/graphql/core/src/loader/loader_registry.rs
andreievg: makes not difference with actix_rt::spawn or tokio::spawn, for some reason I trust tokio::spawn more. Data loaders had issues for me with 'delay' using future-rs, if you interested: https://github.com/async-graphql/async-graphql/issues/1716

server/service/src/auth.rs
andreievg: Hmm it does seem to come from claims.sub in validate_auth ?

server/service/src/auth.rs
```diff
@@ -882,13 +885,19 @@ impl AuthServiceTrait for AuthService {
         context: &ServiceContext,
         auth_data: &AuthData,
         auth_token: &Option<String>,
+        override_user_id: &Option<String>,
         resource_request: &ResourceAccessRequest,
     ) -> Result<ValidatedUser, AuthError> {
-        let validated_auth = validate_auth(auth_data, auth_token)?;
+        let user_id = if let Some(override_user_id) = override_user_id {
```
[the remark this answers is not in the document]
andreievg: 96a2f25fbeb7969dceb1836e4cc7cf2a6a65aca5, thanks

server/service/src/boajs/methods/use_graphql.rs
andreievg: If this is running through service, user_id is usually available, I thought in that case we should try to use user_id. I have an example of calling use_graphql from graphql_query plugin, there could be a plugin that does something via user UI action.

andreievg: > Kind of think, either execute via usual graphql API

Might need to chat about this again, passing user credentials to plugin/plugin methods, in service layer is very tricky

lache-melvin:
Wanted to reply to comments so far but plz request re-review once all changes ready 🙏

andreievg:
@lache-melvin I've address everything apart from message repository in downstream PR. I agree the message repository is too 'open', the alternative would be to expose service, with restrictions and boundaries. However I ran out of time on this, I'll record in carry over

PR #7866, opened 2025-05-25
7755 simulate load

### LINKED ISSUE #7755: Load test CLI

Chris-Petty:
Fixes #7755 

# 👩🏻‍💻 What does this PR do?

Makes lots of requisitions and syncs them, which are queued for other sites, which create their transfers.

No measurement yet.

# 🧪 Testing

- [ ] OG+OMS centrals
- [ ] Create a site called "test_site" with pass: pass
- [ ] Enable testing API on OG server
- [ ] Run the CLI: `cargo build && ./target/debug/remote_server_cli load-test -u http://localhost:8888 -s 25 -d 600`

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/cli/src/load_test.rs
+++ b/server/cli/src/load_test.rs
@@ -1,7 +1,7 @@
-use std::{io::Write, path::PathBuf};
+use std::{path::PathBuf, time::Duration};
 
 use repository::database_settings::DatabaseSettings;
-use reqwest::Client;
+use reqwest::{Client, Error};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use serde_yml;
@@ -9,8 +9,47 @@ use service::{
     settings::{DiscoveryMode, ServerSettings, Settings},
     sync::settings::{BatchSize, SyncSettings},
 };
+use tokio::{process::Child, time::sleep};
 use util::uuid::uuid;
 const TEST_API: &str = "sync/v5/test";
+
+const SYNC_INFO_QUERY: &str = r#"
+query SyncInfo {
+  latestSyncStatus { isSyncing }
+}
+"#;
+
+const MANUAL_SYNC_QUERY: &str = r#"
+mutation ManualSync {
+  manualSync
+}
+"#;
+
+const STOCK_TAKE_MUTATION: &str = r#"
+mutation InsertStockTake($storeId: String!, $input: InsertStocktakeInput!) {
+  insertStocktake(storeId: $storeId, input: $input) {
+    ... on StocktakeNode {
+      id
+    }
+  }
+}
+"#;
+
+const STOCK_TAKE_LINE_MUTATION: &str = r#"
+mutation InsertStockTakeLine ($storeId: String!, $input: InsertStocktakeLineInput!) {
+  insertStocktakeLine (storeId: $storeId, input: $input) {
+    __typename
+    ... on StocktakeLineNode {
+      id
+    }
+    ... on InsertStocktakeLineError{
+      error {
+        description
+      }
+    }
+  }
+}
+"#;
 
 #[derive(clap::Args)]
 pub struct LoadTest {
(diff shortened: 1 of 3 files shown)


server/cli/src/load_test.rs
```diff
@@ -218,93 +367,250 @@ impl LoadTest {
...
+                    .kill_on_drop(true)
+                    .spawn()
+                {
+                    Ok(child) => child,
+                    Err(e) => {
+                        println!("Failed to spawn process: {}", e);
+                        return;
+                    }
+                };
+                sleep(Duration::from_secs(10)).await; // Let db get created, migrated and initialisation started
```
Chris-Petty, line 408: ah I had this scaled to amount of sites at one point, it may not be reliable at 10s if you start say 1000 sites. Though you're likely pretty crazy to do that on one machine anyway, unless it's some 128core beast or something.

_Ideally_, I'd do something like capture stdout in process here and continue when the server states that it's finished migration and started initialisation.

lache-melvin: Agree, think that would be better... plus a log here, something like "remote sites started, beginning load test..."?

server/cli/src/load_test.rs
```diff
@@ -218,93 +371,226 @@ impl LoadTest {
...
+                            Err(e) => {
+                                println!("insertRequestRequisitionLine request failed: {}", e);
+                                kill(&mut child).await;
+                                return;
+                            }
+                        };
+
+                        let line_gql = json!({
+                            "operationName": "UpdateRequestRequisitionLine",
+                            "query": UPDATE_REQUISITION_LINE_MUTATION,
```
lache-melvin, line 471: curious once we start adding measurements, but adding each line as insert+update rather than adding all lines via one batch mutation... _could_ make the OMS site the bottleneck rather than sync? I mean, if it is, that's good to know - but if we're going for load test against OG, maybe the fewer set up requests we have to do to the remote site the better?

lache-melvin (commented):
Looking good, left a couple questions - how do you want this reviewed?

Per our chat the other day, I still think worth putting load test cli option behind the integration_test feature flag, and agree with other comments you have, around wait for migration rather than 10s wait, and using built code for starting the remote sites etc... but if you're adding more branches before merging all this, happy for that to be upstream?

server/cli/src/load_test.rs
Chris-Petty: Honestly I just found the individual update graphql first and went with it- figured surely it's not significantly slower when the requests are originating from the same machine (i.e. the CLI and spawned processes, not the same as central servers ideally).

I would rather do the batch insert and update but was rolling with it for now :S. If nothing else it'll make hyper's debug logging quieter (25x less in fact) which would be pleasant lol.

Chris-Petty:
> Looking good, left a couple questions - how do you want this reviewed?
> 
> Per our chat the other day, I still think worth putting load test cli option behind the integration_test feature flag, and agree with other comments you have, around wait for migration rather than 10s wait, and using built code for starting the remote sites etc... but if you're adding more branches before merging all this, happy for that to be upstream?

Yea I think park reviewing meaningfully for now will do a final branch with measurements addressing our discussions through the stack 🤪

Chris-Petty:
Thought I closed this. Supersede by stacked branch PR #8072

PR #7871, opened 2025-05-26
7807 add ability to select campaign inbound shipments

### LINKED ISSUE #7807: Add ability to select "Campaign" when receiving stock in Inbound Shipments (same pattern as Donor?)

GeronimoJohn:
Fixes #7807 

# 👩🏻‍💻 What does this PR do?

Added the ability to insert or update a campaign for an inbound shipment line

# 🧪 Testing

- [ ] Navigate Replenishments -> Inbound Shipments -> Add or select an item -> Navigate to Other Tab
- [ ] If you have campaigns already -> Select a campaign from the dropdown
- [ ] Click Ok -> Campaign is now saved!
- [ ] Select on item again and choose a different campaign
- [ ] Click Ok -> Campaign is updated!
- [ ] If you don't have campaigns -> The dropdown should return a useful message

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/translations/invoice_line.rs
```diff
@@ -95,6 +95,9 @@ pub struct LegacyTransLineRow {
     #[serde(deserialize_with = "empty_str_as_option_string")]
     #[serde(rename = "vaccine_vial_monitor_status_ID")]
     pub vvm_status_id: Option<String>,
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_str_as_option_string")]
+    pub campaign_id: Option<String>,
```
fergie-nz, line 100: I am thinking that we don't need legacy translation here because campaign table  isn't in OG (only in OMS).

However I'm not super confident with this part of the codebase. @jmbrunskill, would you be willing to confirm if we need this?

server/graphql/invoice_line/src/mutations/inbound_shipment_line/line/insert.rs
```diff
@@ -118,7 +120,7 @@ impl InsertInput {
             stock_line_id: None,
             barcode: None,
             stock_on_hold: false,
-            campaign_id: None,
+            campaign_id,
```
fergie-nz, line 123: This is getting passed via graphql, but I think need to refactor generate now as campaign_id is getting dropped in service layer, and not actually being inserted when we insert an inbound line.

See behaviour here:


https://github.com/user-attachments/assets/d07ff0fa-9f71-4afc-a8bc-9c794eb16caa

fergie-nz (dismissed):
Hey John looks great! I think just one change required which is refactoring the insert generate. I think it was carried over from previous PR, but is dropping the campaign id on insert.

Also had one other question about sync, but am not certain and am hoping James can provide insights. Or maybe you know it needs/doesn't need legacy sync? In any case I just commented my question below :)

server/service/src/sync/translations/invoice_line.rs
jmbrunskill: Yes, we should add to both invoice_line and stock_line in OMS. As discussed lets do it as an `OMS_fields` where we can extend the JSON in OMS without having to do an mSupply version change again in future.

fergie-nz (approved):
Nice John awesome looks great!

server/service/src/sync/translations/invoice_line.rs
GeronimoJohn: Thanks @jmbrunskill, have created a separate issue for it: https://github.com/msupply-foundation/open-msupply/issues/7882

Issue #7343, opened 2025-04-15
Dose Management: display doses in Internal Order & Requisition


roxy-dao:
Note all mentions of vials is just the unit name and this can differ.

### Internal Order
Display vaccine in doses preference on:
Detail view:
- [ ] Change unit from `{unit name}` to `dose`
- [ ] Re-calculate all fields

Line edit:
- [ ] `Requested packs` column should now display `Requested doses`
- [ ] The switch for units/packs should now display unit/doses
- [ ] Check calculations to convert vials to doses, this is fetched from `item.doses`

### Requisition
Manual Requisition:
Display vaccine in doses preference on:
Detail view:
- [ ] Change unit from `{unit name}` to `dose`
- [ ] Re-calculate all fields

Line edit:
- [ ] Add a new component below requested quantity which displays the requested quantity in units or vials depending on preference
- [ ] Also check that the charts are displaying the correct values (e.g. if it's in doses then the chart should be showing in doses)

For automatic requisition/manual:
Line edit: 
- [ ] Component to supply requested in doses

Carryover:
Both edit views are set to be refactored so that all the fields show in units/doses but also have a packs input field

### Acceptance Criteria ✅
- [ ] Users are able to request doses when making an Internal Order
- [ ] Users should be able to request in packs when requesting a non-vaccine item
- [ ] Users can see the unit displayed correctly in the Requisition detail view (e.g. doses instead of vials if pref is turned on)
- [ ] Users should be able to see the requested quantity in packs or vials in the line edit page
- [ ] User should be able to supply in doses

### Moneyworks Jobcode 🧰
NIGGF:DSE

roxy-dao:
Wait for IO refactor to modal before starting

roxy-dao:
Closed for now. We'll be ordering in units (:

adamdewey:
@roxy-dao , @mark-prins - isn't this one of the main requests from Nigeria (and Niger, Chad etc)?

To be able to order in doses, not units?

roxy-dao:
Use item.doses for except for available which will show available_doses. 
2nd iteration: able to select item_variant to use in calculation

adamdewey:
@roxy-dao , @DhanyaHerath , @mark-prins - if we use item.doses that will basically force Niger to treat each item as a separate variant. 

So, if they're happy structuring their items like that, we might as well just have 1 variant per item, have the variant selected automatically in internal orders and use the doses field on the variant instead?

[...]

PR #7885, opened 2025-05-26
7815 requisition modal

### LINKED ISSUE #7815: Requisition modal

roxy-dao:
Fixes #7815 

# 👩🏻‍💻 What does this PR do?
Turn requisition back into a modal & some refactoring along the way... I moved short expiry to the right column to free up more space... but lemme know what you think.

Have hidden AMC and MOS in manual requisitions. OG doesn't show them... would we want to display them regardless? 👀 Which links into... I think we should be calculating suggested on manual requisitions depending on what the supplying store inputs for the customer if they have the stats.... (which means we shouldn't be hiding AMC & MOS fields)

#### With Extra Fields
extra-vials

extra-packs

#### Manual Requisition
mans

#### Linked Internal Order
linked-IO

#### Disabled
finalised

- Haven't tested with plugin

# 🧪 Testing

- [ ] Make sure requisition flow still works as it's supposed to

# 📃 Documentation

  1. Requisition 
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/requisition_line/response_requisition_line/update.rs
```diff
@@ -126,8 +130,17 @@ fn generate(
         option_id: updated_option_id,
     }: UpdateResponseRequisitionLine,
 ) -> (Option<RequisitionRow>, RequisitionLineRow) {
+    let suggested_quantity = generate_suggested_quantity(GenerateSuggestedQuantity {
```
roxy-dao, line 133: Calculating suggested for manual requisition lines... maybe should be invalidating frontend if available_SOH changes... we don't have enough params in frontend to calculate it there... 👀

lache-melvin (changes requested):
Thanks Roxy, mammoth effort 🔥 Can I request that you add back those fields, keep this to just UI rearrange, that can be separate decision?

Couple issues found...

Our stock is blank when adding items:

https://github.com/user-attachments/assets/6a617863-6599-4363-bf31-429024733029

First OK & Next doesn't work, until field out of focus?

https://github.com/user-attachments/assets/a40b8d5a-3dbb-4d2f-9b0d-25a93ce8a7f0

Nits:

Don't think the singularising is worth it, this looks weird as when you type 1 on your way to typing 12:

https://github.com/user-attachments/assets/4759eff4-ba29-4179-ab3c-bff840f8276e

(also no drop shadow under supply field 🤣)

client/packages/inventory/src/Stocktake/DetailView/ReduceLinesToZeroModal.tsx
```diff
@@ -68,7 +68,7 @@ export const ReduceLinesToZeroConfirmationModal = ({
               value={reason}
               onChange={reason => setReason(reason)}
               reasonOptions={reasonOptions?.nodes ?? []}
-              isLoading={isLoading}
+              loading={isLoading}
```
roxy-dao, line 71: Can't comment up above since I haven't made a change there... but ReasonSearchInput can be refactored further, so that we don't need to pass in reasonOptions and loading... We made a decision to show this everywhere and just have it disabled if it isn't required, so was there a reason why it was hidden here?

roxy-dao (commented):
- I've rearranged the columns as requested by @mark-prins... 
- AMC was added in when doing changes to allow for manual requisitions... Have hidden this for manual requisitions since the supplier probably doesn't have this information (and it wasn't there in the original modal...). Also a decision made by @mark-prins 
- Ok & Next seems to be working for me 👀 
- Have kept the plural in for now but have debounced it so it doesn't change all the time... But note this does not work in other languages apart from English, so would just be singular for other languages. Even if we use the other way of deriving plurals, there's a chance that the user might name the unit `tablets` which will cause it to say `tabletss`... neither good solutions...

Modal with keyboard... was agreed that it's fine to hide a bit of the comment input~
photo_2025-05-29 11 39 27

lache-melvin (changes requested):
Looking great, thanks Roxy, main remaining thing is the charts loading on new items :)

- It still says no stock available in the charts when I am adding an item. When we first put in manual requisitions, we just hid the charts completely when adding item, only show them when editing... feels like a pretty big oversight to me, surely I want to see my stock stats when adding the items? But out of scope, I think raise that separately and hide the "no stock available" for now

Screenshot 2025-05-29 at 5 05 11 PM


- NIT: the different coloured section is doing some slightly funchy biz on full screen modal:

Screenshot 2025-05-29 at 5 06 02 PM


- I'm still getting the ok next issue - [removing the buffer states](https://github.com/msupply-foundation/open-msupply/compare/7815-requisition-modal...7815-requisition-modal(remove-buffer-states)) seemed to fix it?

Other thoughts:
- should the `supply` field autofocus? Currently, manual req auto-focuses the `their SOH` field, and transfers have nothing focused. Supply field is the main one here right?
- when i don't have the extra columns on, this seems like a real waste of space? Fine to leave for now, esp if you have run by Mark, but would like to think we could utilise that better lol

client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseEditModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+    else onClose();
+  };
+
+  useEffect(() => {
+    if (!!draft?.isCreated) {
+      save();
+    } else {
+      if (!!draft?.id) setPreviousItemLineId(draft.id);
+    }
+  }, [draft, setPreviousItemLineId]);
```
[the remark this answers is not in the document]
roxy-dao: onChangeItem would still be working with previous state of the draft, so if it's moved to the onChangeItem would need a separate function to ensure it has the latest draft.. I've added a comment instead, is that alright?

mark-prins:
> when i don't have the extra columns on, this seems like a real waste of space?

I didn't test all the combinations (clearly!) it isn't great with the extra fields not shown..
would it be weird to have comment on the left column in this case?

roxy-dao:
Internal Order:
io
io-extra

Requisition: 
req

There was another request to shuffle the rows... also fixed full screen discolouring... matched reasons component with the one in Internal Orders since we have the space for it...
Screenshot 2025-05-30 at 7 29 23 PM

roxy-dao:
More styling changes to come in #7343

lache-melvin (approved):
Works awesome, thanks Roxy for all the effort, let's get it in and tested 🙏 

I added a comment around that useEffect, to try clarify all the different states... had me thinking, not sure we need the `previousItemLine` state, surely we can just delete line based on `draft` id and then change item??

Also noticed, for some reason Supply field doesn't auto-focus after clicking ok next 👀 

I think these are small things though, we should get this in and QA'd rather than fixing anything further 😅

roxy-dao:
> Works awesome, thanks Roxy for all the effort, let's get it in and tested 🙏
> 
> I added a comment around that useEffect, to try clarify all the different states... had me thinking, not sure we need the `previousItemLine` state, surely we can just delete line based on `draft` id and then change item??
> 
> Also noticed, for some reason Supply field doesn't auto-focus after clicking ok next 👀
> 
> I think these are small things though, we should get this in and QA'd rather than fixing anything further 😅

Will check it out in doses PR. Thanks Laché~

PR #7911, opened 2025-05-27
7868 add running balance total for ledger queries

### LINKED ISSUE #7868: Item ledger calculations not adding correctly
## What went wrong? 😲

Image

8000 - 1000 does not equal -3800...

## Expected behaviour 🤔

Calculate correctly

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #7868 

# 👩🏻‍💻 What does this PR do?

Refactors item_ledger to prevent double counting of ledger items in cases where they occur at the same datetime.
Preserve multiple id query logic without mapping to keep functionality for reports which use this query
Add tests to ensure logic is stable

Note that the ledger balance is always calculated based on dateTime, which leads to strange displaying results if we sort by datetime in the front end in a reverse order from how we calculate the balance.

You can see this when you change the datetime order from ascending to descending in the item ledger tab of items.
Some ways the ledgers are out of order.

Unsure quite how to handle this, but I am thinking this is still better than the alternative of prematurely summing of total values.

# 🧪 Testing

- [ ] Navigate to Items and then ledger tab 
- [ ] See that balance is calculated for each line where more than one line exists for the same datetime.
- [ ] Also should double check the stockouts report [here](https://github.com/msupply-foundation/open-msupply-reports/blob/8b5414fc9ee03de983f1f20da1812b90723c059f/clients/Ivory-Coast/stockouts/latest/src/query.graphql#L2) is still working without regression

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/ledger.rs
+++ b/server/service/src/ledger.rs
@@ -1,11 +1,11 @@
 use std::collections::HashMap;
 
 use repository::{
-    ledger::{self, LedgerFilter, LedgerRepository, LedgerRow, LedgerSort, LedgerSortField},
+    ledger::{LedgerFilter, LedgerRepository, LedgerRow, LedgerSort, LedgerSortField},
     EqualFilter, Pagination, PaginationOption, StorageConnection, StorageConnectionManager,
 };
 
-use crate::{get_default_pagination_unlimited, i64_to_u32, usize_to_u32};
+use crate::{get_default_pagination_unlimited, usize_to_u32};
 
 use super::{ListError, ListResult};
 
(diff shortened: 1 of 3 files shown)


server/service/src/ledger.rs
```diff
@@ -43,7 +50,6 @@ pub fn get_item_ledger(
         .unwrap_or_default()
         .store_id(EqualFilter::equal_to(store_id));
 
-    let rows = repository.query(pagination, Some(filter.clone()), sort)?;
     let all_ledger_items = repository.query(
```
fergie-nz, line 53: I am slightly worried about the consistency of this query. 

Will it always return ledgerRows in the same order if datetime is equal?
Does it depend upon the database?

Would be confusing for users if we calculate in one order for sqlite, and differently for postgres for example.
Would it change order for the same query randomly?

Won't make a difference in terms of final calculation, but different order of operations would look weird for users.

I almost want to add the cursor to the ledgerRow so that we can define what happens first / next, even if arbitrarily, so that it tells a consistent story to users.

PR #7888, opened 2025-05-26
7641 prescriptions vvm allocate

jmbrunskill:
Fixes #7641 #7707 #7799 #7472

Thanks to @lache-melvin who did most of the work on this!

# 👩🏻‍💻 What does this PR do?

Refactors Prescriptions to use new allocation methods
Adds break packs support to new allocation structure
Adds support for sorting by VVM status to Prescriptions

## 💌 Any notes for the reviewer?

Is there any functionality we're lost in this refactor that we need to re-instate?

# 🧪 Testing

- [ ] Test prescriptions thouroghly.
- [ ] Backdated prescriptions have right stock levels available?
- [ ] Correct stock is selected based on VVM Status and/or expiry

# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [X] **These areas should be updated or checked**: 
  1. Prescriptions
  2.

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #7923, opened 2025-05-28
#7901 Fix JSON forms prescription component

### LINKED ISSUE #7901: Fix JSON Forms Prescription component for new allocation refactor
## What went wrong? 😲

After https://github.com/msupply-foundation/open-msupply/pull/7888, the Prescription component for JSON Forms is not working correctly. Need to update it to use the new allocation functionality.

## Expected behaviour 🤔

It should work

CarlosNZ:
Fixes #7901

# 👩🏻‍💻 What does this PR do?

There should be no functional change from the user POV here, it's just a refactor to fix up what broke when doing #7888.

The `useDraftPrescriptionLines` hook was causing problems with a rendering loop in the Prescription component, and now that `useDraftPrescriptionLines` is no longer used anywhere else, I've removed it completely and re-created a much simpler version for this component (`useDraftLines`).

See inline notes.

# 🧪 Testing

- [ ] Ensure functionality of Prescription component is unchanged from what it was

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1259,6 +1260,7 @@
   "label.units-available": "{{unit}} Available",
   "label.units-in-stock": "{{unit}} in Stock",
   "label.units-issued": "{{unit}} Issued",
+  "label.issued": "Issued",
```
CarlosNZ, line 1263: These got changed recently, so the strings in the Prescription component were broken.

client/packages/programs/src/JsonForms/components/Prescription/useDraftLines.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const stockLines = uniqBy(historicalStockLines.nodes, 'id').filter(
+      stockLine => !stockLine.onHold
+    );
+
+    setDraftLines(
+      stockLines.map(stockLine =>
+        createDraftPrescriptionLineFromStockLine({ stockLine, invoiceId: '' })
+      )
+    );
+  }, [historicalStockLines]);
```
CarlosNZ, line 26: I'm not 100% clear on what is being fetched here, and why, but I assume it's correct as it's following the same as the prior `useDraftPrescriptionLines` hook.

lache-melvin: We only show the available stock lines before creating the prescription right? Like we don't need to consider lines that already exist on the prescription here?

lache-melvin: Using historical stock lines is probably surplus to requirements, seeing as we don't support backdated prescriptions in the JSON form component... I suppose we could?

But I think better not to change without explicit decision, user should probably confirm something in the UI before we go backdating their prescriptions, seems out of scope for 2.8 fix!

PR #7937, opened 2025-05-29
7851 support upload

### LINKED ISSUE #7851: Support Database & Logs Extract

Support team need a method to get a sqlite database and logs from a tablet.

Needs some discussion.

Options:
* Sync files to Central Server based on a sync message triggerd rom central?
* Push by end user as part of support request?

GeronimoJohn:
Fixes #7851

# 👩🏻‍💻 What does this PR do?

Initial work for extracting a database and its logs from a tablet. Started out with creating a dedicated table called `support_table`. It holds information as which site is asking for support, what error are they getting, a list of database and logs files and more.

# 🧪 Testing

- [ ] Database only

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/support_upload_row.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use chrono::NaiveDateTime;
+use diesel::prelude::*;
+use diesel_derive_enum::DbEnum;
+use serde::{Deserialize, Serialize};
+
+table! {
+    support_upload(id) {
+        id -> Text,
+        datetime -> Timestamp,
```
jmbrunskill, line 14:
```suggestion
        created_datetime -> Timestamp,
```

server/repository/src/db_diesel/support_upload_row.rs
```diff
@@ -0,0 +1,112 @@
...
+        site_id -> Text,
```
jmbrunskill, line 15: site_id is actually and integer I think, can you check other usages?

jmbrunskill: Bigger questions actually, I think we'll either need to add a store_id to this table (and let them download the database for a store (which will include the entire site) or we'll need to make changes to sync logic to look at this site_id. I think changing to store_id is the simple solution for now...

jmbrunskill (changes requested):
I think we need to add a `store_id` after all, otherwise the central sync logic won't actually know which site to send it too. Would be possible to add `destination_site_id` to change_log but not sure that's a good idea, we already have a source_site_id which is used for a different purpose. Would definitely need a new sync_style I think?

GeronimoJohn:
Closing due to tackling the underlying issue with a different approach

Issue #7859, opened 2025-05-23
Query + cache OMS prefs once on app startup


lache-melvin:
## The suggestion

2.8 adds new OMS prefs, they are currently queried where they are used, with hook like:

```ts
const { data: prefs } = usePreference(PreferenceKey.ManageVaccinesInDoses)
```

However, this hook call does an API query, which takes time to resolve, and means we do lots of pref queries throughout the app. Implications more below.

Suggestion is to do one query for all prefs on app startup, invalidate if prefs are edited, otherwise the usePreference hook just fetches from cache. So usage looks the same, but the pref should be available. 

### Why should we invest time in this?

We have a number of areas in our frontend code that "initialise" i.e. only run once to set themselves up, e.g. table columns.

If prefs query hasn't resolved before this initialisation runs, then the column might not be rendered. Next time you come back to the page, it is there though.

While this can be got around in development (run pref query in parent component, only render child once resolved) this step is inconvenient (I just want to use the pref where I am) and also easily forgotten.  We've already run into it a few times.

### Are there any risks associated with this change?

Not really - hook usage remains same, just the access mechanism changes

### How much effort is required?

Half a day ish?

mark-prins:
triage: sounds like a great idea and a good cooldown task

andreievg:
I have some ideas on this front, like 'do we need to do a dynamic query and manually cast types' -> I think probably don't need to do that logic if we are 'caching' and can get all preferences, keen to chat, I did have a go at looking at preferences without 'registry' but ran out of time.

PR #7940, opened 2025-05-29
7932 outbound shipment modal doesnt load

### LINKED ISSUE #7932: Outbound shipment modal doesn't load for items with some allocated and some placeholder stock
## What went wrong? 😲

I have an item with some available stock. I try to issue more than I have currently available. I'm in new status, so it creates a placeholder for the remaining amount:
Image

I save the line, happy days. (Secondary bug here, not resetting on reopen, need to refresh first) Reopen line:

Image

API says: `DBError { msg: \"Stock line not found\", extra: \"0196c682-4e07-7941-a15e-815d6eb5840c\" }`

Pretty sure something to do with changes for prescriptions...

## Expected behaviour 🤔

I can reopen and see my placeholder line!

## How to Reproduce 🔨

1. Follow steps above

## Your environment 🌱

- Open mSupply Version: v2.8-RC3

- Platform:

- Database type:

lache-melvin:
Fixes #7932

# 👩🏻‍💻 What does this PR do?

Allows outbound shipment/prescription lines with stock lines fully allocated to still be re-opened :)

Screenshot 2025-05-29 at 2 46 09 PM

Issue mentioned the state not reloading when you open, having to refresh the page - I think this was actually just because the query was then failing, so we weren't updating the data, works fine now :)

# 🧪 Testing

- [ ] Add an item - use the auto-allocation input to issue more than the currently available quantity (i.e. so all available stock gets allocated)
- [ ] Save
- [ ] Reopen the modal - lines load correctly
- [ ] Seeing as this touches historical stock code - please do the same for a backdated prescription, and ensure available quantities are correct when you reopen

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
@@ -14,6 +14,8 @@ import {
   useNotification,
   useTableStore,
   useBreadcrumbs,
+  usePreference,
+  PreferenceKey,
 } from '@openmsupply-client/common';
 import { toItemRow, ActivityLogList } from '@openmsupply-client/system';
 import { ContentArea } from './ContentArea';
@@ -33,6 +35,12 @@ const DetailViewInner = () => {
   const t = useTranslation();
   const { info } = useNotification();
   const isDisabled = useOutbound.utils.isDisabled();
+
+  const { data: prefs } = usePreference(
+    PreferenceKey.SortByVvmStatusThenExpiry,
+    PreferenceKey.ManageVaccinesInDoses
+  );
+
   const { entity, mode, onOpen, onClose, isOpen, setMode } =
     useEditModal<OutboundOpenedWith>();
   const {
@@ -111,6 +119,10 @@ const DetailViewInner = () => {
               onClose={onClose}
               status={data.status}
               invoiceId={data.id}
+              prefOptions={{
+                sortByVvmStatus: prefs?.sortByVvmStatusThenExpiry ?? false,
+                manageVaccinesInDoses: prefs?.manageVaccinesInDoses ?? false,
+              }}
             />
           )}
 
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
@@ -36,6 +36,7 @@ interface AllocationProps {
   scannedBatch?: string;
   prefOptions: {
     sortByVvmStatus: boolean;
+    manageVaccinesInDoses: boolean;
   };
 }
 
@@ -44,7 +45,7 @@ export const Allocation = ({
   invoiceId,
   allowPlaceholder,
   scannedBatch,
-  prefOptions: { sortByVvmStatus },
+  prefOptions: { sortByVvmStatus, manageVaccinesInDoses },
 }: AllocationProps) => {
   const { initialise, item } = useAllocationContext(({ initialise, item }) => ({
     initialise,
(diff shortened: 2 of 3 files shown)


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1729,7 +1729,8 @@
   "messages.patients-found_one": "1 patient found matching the supplied details",
   "messages.patients-found_other": "{{count}} patients found matching the supplied details",
   "messages.period-not-available": "No available periods",
-  "messages.placeholder-allocated": "Not enough stock is available to allocate the requested quantity. A placeholder has been added for  {{placeholderQuantity}} units.",
+  "messages.placeholder-allocated-doses": "Not enough stock is available to allocate the requested quantity. A placeholder has been added for {{placeholderQuantity}} doses.",
+  "messages.placeholder-allocated-units": "Not enough stock is available to allocate the requested quantity. A placeholder has been added for {{placeholderQuantity}} units.",
```
lache-melvin, line 1733: Sorry, side quest #1, was testing, noticed that when issuing in doses, warning still said it was placeholder in units.

client/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -33,6 +35,12 @@ const DetailViewInner = () => {
   const t = useTranslation();
   const { info } = useNotification();
   const isDisabled = useOutbound.utils.isDisabled();
+
+  const { data: prefs } = usePreference(
+    PreferenceKey.SortByVvmStatusThenExpiry,
+    PreferenceKey.ManageVaccinesInDoses
+  );
```
lache-melvin, line 42: Moved the pref query out of modal component, into parent, ensure these prefs are available in time load the lines in the right order etc...#7859

client/packages/invoices/src/StockOut/useAllocationContext.ts
```diff
@@ -257,24 +257,24 @@ export const useAllocationContext = create<AllocationContext>((set, get) => ({
...
+    const placeholderInUnits =
+      allocateIn.type === AllocateInType.Packs
+        ? stillToAllocate * allocateIn.packSize
+        : stillToAllocate;
+
     const alerts = getAllocationAlerts(
       quantity,
       allocatedQuantity,
-      stillToAllocate,
+      placeholderInUnits,
```
lache-melvin, line 270: last unrelated fix lol - placeholder is in units, so warning should show in units too 😅

server/service/src/invoice_line/get_draft_outbound_lines.rs
```diff
@@ -157,9 +163,34 @@ fn get_outgoing_invoice_lines(
             .item_id(EqualFilter::equal_to(item_id)),
     )?;
 
+    let mut invoice_stock_lines = existing_invoice_lines
+        .iter()
+        .filter_map(|line| line.stock_line_option.clone())
+        .collect::<Vec<StockLineRow>>();
+
+    // If invoice is backdated, available packs should be based on the historical date
+    if let Some(backdated_datetime) = outbound.backdated_datetime {
```
lache-melvin, line 172: This is the actual change!

Old prescriptions frontend did 2 queries, one for historical stock lines, one for the existing invoice lines and their stock. Now we're doing it on the backend, we need to do the same...

Main historical stock lines query only gets lines with available stock, hence the stock_line_not_found issue when reopening after issuing all available stock. 

Could update it, to get _all_ stock lines, and then filter here to the ones we need - but then we're running all of the historical calculations on every stock line that ever existed for an item... would probably get expensive over time??

Figured best to leave that one as is, and query separately here, getting the historical quantities for stock lines already in the invoice, regardless of whether they have remaining available stock

But keen for thoughts! @jmbrunskill

jmbrunskill (approved):
Looks good thanks for the find!
Haven't had a chance to think through all the implications but I think this looks like a good solution so will merge and get testing underway for 2.8

PR #7946, opened 2025-05-29
#7915(2) Prefs key invalidation fix

### LINKED ISSUE #7915: When Preference Manage VVM status for stock is off, the VVM status field show in the View stock window as well as the Status history tab being visible
## What went wrong? 😲
As title plus the button to add VVM is clickable when clicking on a status history line, the VVM field is empty and the comment box can be interacted with will not display any text you are writing.  
https://drive.google.com/file/d/1nMI5VsYzEvpxIH819JQ1oACvFzqijgxE/view?usp=sharing

~~The VVM status is also appearing in the Outbound shipment;~~ The VVM status doesn't appear on outbound shipment after  I logged out and back but the stock window is till showing VVM status and status history tab
https://drive.google.com/file/d/1YJvnQsCVVH7gW787EAqE3SWi2RZ28Dr9/view?usp=sharing

## Expected behaviour 🤔
I think that when the `Manage VVM status for stock` is off, we should use the same behaviour than for item that are not a vaccine ie doesn't show the Vaccine history tab at all and the VVM status field is hidden. Same thing for Stock line when allocating stock of a vaccine in an outbound shipment

## How to Reproduce 🔨

Have some vaccine in stock that have VVM status  
On OMS central turn off the Preference `Manage VVM status for stock` for your store
Sync
1. Go to 'Inventory > View stock'
2. Click on one of your vaccine
3. See that the VVM status field is showing as well as the Status History tab
4. Create a new inboind shipment
5. Add a vaccine item
6. See error when allocating qty form your stock line. The VVM status is showing. 

## Your environment 🌱

- Open mSupply Version: v2.8RC3
- Legacy mSupply Central Server Version: v8.02

- Platform: Android

- Database type: SqLite3

CarlosNZ:
Fixes #7915 

# 👩🏻‍💻 What does this PR do?

@noel-yeldos already added the preference check to the VVM Status tab, but this PR fixes the preferences query invalidation so that when you change the preference the Tab appears/disappears without a page reload.

There was an inconsistency with the queryKeys array, which is why the prefs weren't being re-fetched. However, I've also simplified the Prefs hooks to use the same API hook and not rely on the Auth API. This makes it a bit clearer and not quite so intertwined with the Auth API.

# 🧪 Testing

- [ ] Go to "View Stock" and click on a vaccine item
- [ ] Note whether the Status History tab (and the VVM status input) is available or not
- [ ] Open your store prefs by clicking "Edit" in the Footer (must be central server)
- [ ] Toggle the "Manage VVM status for stock" pref, and close
- [ ] Note the Status History tab and VVM input have switched (either on or off)
- [ ] Try it the other way round (turn it off/on)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/authentication/api/hooks/usePreference.ts
```diff
@@ -29,7 +27,7 @@ export const usePreference = <T extends PreferenceKey>(...prefs: T[]) => {
   };
 
   return useQuery({
-    queryKey: [keys.preferences(), ...prefs],
```
CarlosNZ, line 32: This was the source of the bug -- `keys.preferences()` returns an array, so they key array was `[ [ "preferences"], "etc..." ]` when it should have been `[ "preferences"], "etc..." ]`.

The simplification to use the `usePreferencesGraphQL` API helps reduce this potential confusion.

client/packages/system/src/Manage/Preferences/api/useEditPreference.ts
```diff
@@ -54,8 +52,7 @@ const useUpsertPref = () => {
     },
     {
       onSuccess: () => {
-        queryClient.invalidateQueries(keys.preferences());
-        queryClient.invalidateQueries(PREFERENCE_DESCRIPTIONS);
+        queryClient.invalidateQueries(PREFERENCES);
```
CarlosNZ, line 55: Now we only need to invalidate one set of keys :)

client/packages/system/src/Manage/Preferences/api/usePreferencesGraphQL.ts
```diff
@@ -4,12 +4,13 @@ import {
...
+import { PREFERENCES } from './keys';
 
 export const usePreferencesGraphQL = () => {
   const { client } = useGql();
   const queryClient = useQueryClient();
   const { storeId, store } = useAuthContext();
   const api = getSdk(client);
 
-  return { api, queryClient, storeId, store };
+  return { api, queryClient, storeId, store, client, PREFERENCES };
```
CarlosNZ, line 15: Just re-exporting key from here to avoid two ugly long imports in `usePreference`

PR #7961, opened 2025-05-29
7031 open internal order in new tab

### LINKED ISSUE #7031: Cannot open internal order in new tab
## Expected behaviour 🤔

I want to open internal order in a new tab, this is useful when searching, maybe for other operations too.
In my use case i want to find a particular internal order for a period, there are 5 entries for this period, i scroll down by sorting by period and find the order, i can't open in new tab, so have to open order in current window, at which point i loose my place in the sort when i come back to it (similar to this workflow: https://github.com/msupply-foundation/open-msupply-internal/issues/58, which we thought is ok on web version, but it doesn't look like it is).

Check other areas, this should be possible.

Things to consider: When making changes (in another tab), and going back to the list (in original tab), the list will not be updated

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #7031 

# 👩🏻‍💻 What does this PR do?

Allows right clicking on the internal orders list view to open an internal order in a new tab

This is done via an optional prop on the common data table component. It can be defined for specific tables to use, without affecting other areas.

Screenshot 2025-05-30 at 11 07 46 AM

I also created an APK and tested that changes didn't break anything on a tablet - I can share if needed

Only on Internal Orders for now to get functionality in. Should be straight forwards to add to other tables through the prop as required

There were a lot of iterations of this... challenges of this were as follows:
- Current implementation of OnClick does not support native browser behaviours such as context menu & right click
- `Link` supports these features needed however:
- `TableRow` and `TableCell` cannot be directly wrapped with a `Link` - causes invalid HTML
- `Link` needs to be wrapped around the content
- The content does not fill the cells, therefore much of the row was not clickable. Also any cells that are empty, do not have a link at all. This was hard to navigate and showing row highlighting for the whole row (even though only ~half was clickable) - would be confusing for the user
This is what was clickable with the content wrapped in `Link`
Screenshot 2025-05-30 at 11 18 33 AM

Solution: 
Wrap each `TableCell`'s content in a `Link` that fills the cell - this is the `ContentWrapper`

I styled the `ContentWrapper` so that it could fill the cell. There were styling implications to this which affected other areas:
- Text content could be aligned easily, but not other types of content. The comment popover being one, also the icon in 'Supplier' and vaccine labels etc. Needed to use 'flex' instead of 'block'
- With flex column alignment changed - number columns need to be right aligned (passed through alignment prop to justify content)

Puppeteer would've been great to use with this! For reference here's a screenshot from develop to check consistency

Note sidequest - changing cell type for Program to `TooltipTextCell` for long program names
develop

# 🧪 Testing

- [ ] Go to Internal Orders -> List View -> have a few internal orders created
- [ ] Right click on an internal order -> see the context menu appear
- [ ] Select `open in new tab` -> see the correct internal order open in a new tab within the same browser window
- [ ] From the list view, press ctrl/cmd and click - this will also open in a new tab
- [ ] From the list view, left click on an internal order -> see it navigate within the same window
- [ ] Go to another list view eg Requisitions -> See the right click behaviour is not enabled
- [ ] See consistency with styling between the Internal Orders list view and other list views
  - [ ] column headers are aligned with contents
  - [ ] column header sort correctly
  - [ ] images (eg comment popover) are centered within the row
  - [ ] tooltips, popovers and select checkboxes work

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/requisitions/src/RequestRequisition/ListView/ListView.tsx
```diff
@@ -93,6 +95,7 @@ export const RequestRequisitionListView: FC = () => {
         sortable: true,
         width: 150,
         defaultHideOnMobile: true,
+        Cell: TooltipTextCell,
```
aimee-mcneil-melville, line 98: Sidequest: make this a tooltip (which also adds ellipses) for long program names

PR #7971, opened 2025-05-30
7959 backend allocate in doses

### LINKED ISSUE #7959: Allocate selected lines needs to support doses

Creating an OS from requisition creates with placeholder values. From here, want to select all lines and auto-allocate stock. If my requisition was in doses, allocation should act accordingly, taking variants into consideration.

### Moneyworks Jobcode 🧰

NIGGF:DSE

lache-melvin:
Fixes #7959

# 👩🏻‍💻 What does this PR do?

Updates backend `allocate` method to consider item variants when allocating in doses.

Note number of units when in placeholder - is based on default doses per unit:
Screenshot 2025-05-30 at 5 43 10 PM

After allocating, its based on the item variant of the stock line - correct doses allocated, so fewer units:
Screenshot 2025-05-30 at 5 43 21 PM

Had to update frontend allocation as well - was treating placeholder unallocated stock as 1 dose per unit, but impossible for backend to tell if we requested the dose or unit lens. We should use default doses per unit, so then placeholder quantity is always in units (e.g. 10 doses = 2 units... placeholder is now for 2 units regardless of lens, rather than switching between 10 and 2 🫠).

## Send help

@DhanyaHerath @mark-prins - wanting to confirm - if the doses pref is on, this should _always_ meet the dose quantity requested, rather than unit quantity?

I.e. per above, goal should always be to issue 110 doses? Or is there a use case for, I have the pref on, but I really do want 22 units, regardless of variant?

We don't currently store anything on the record to indicate if requested quantity was in doses or units... we store in units and display in doses if the pref is on. If they want to sometimes order in units, regardless of dose-per-unit variations, things gon get more complicated 😨 

I did find one issue, was pre-existing though... 

[Frontend allocation has a 3 step cycle](https://github.com/msupply-foundation/open-msupply/blob/v2.8.0-RC/client/packages/invoices/src/StockOut/allocateQuantities.ts), to try to allocate the exact requested quantity.

Backend does always FEFO, will round up to nearest pack size. So [this test fails](https://github.com/msupply-foundation/open-msupply/compare/7959-backend-allocate-in-doses...7959-backend-allocate-in-doses(prevent-overallocation-test))... 

I think should be a decision from @mark-prins / team, whichever way we go, it should probably be consistent no matter which button you click!

# 🧪 Testing

For now I think your easiest way to set up might be:
- [ ] Have Manage Vaccines in doses pref on
- [ ] Have a vaccine item with no stock on hand - it should have at least one variant defined that has a different number of doses than the default doses per unit on the item (e.g. default is 10 doses per unit, variant is 20 doses)
- [ ] Create a manual Outbound Shipment, and add this vaccine item
- [ ] Modal should be issuing in doses - add some amount of placeholder quantity (e.g. 80)
- [ ] In the detail view, you can see the num units vs num doses in placeholder (for this example, would be 8 units, 80 doses)
- [ ] Now go add a stock line for your vaccine item, and give it the variant
- [ ] Back in outbound shipment, select the line and click `Allocate placeholder lines`
- [ ] Correct number of _doses_ has been allocated, based on the variant (for this example, 4 units, 80 doses)

# 📃 Documentation

  1. Mention this with the allocation button docs
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -54,18 +56,20 @@ const PlaceholderRow = ({
...
+        colSpan={5 + extraColumnOffset}
         sx={{ color: 'secondary.main' }}
       >
         {t('label.placeholder')}
       </PlaceholderCell>
-      <PlaceholderCell
-        style={{ textAlign: 'right', paddingRight: '14px' }}
-        colSpan={2}
-      >
+      <PlaceholderCell style={{ textAlign: 'right', paddingRight: '14px' }}>
```
lache-melvin, line 64: bumping the above placeholder cell one col wider, rather than having this one (pack size) span 2 cols

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -54,18 +56,20 @@ const PlaceholderRow = ({
...
         1
       </PlaceholderCell>
-      <PlaceholderCell colSpan={3}></PlaceholderCell>
+      {!!dosesPerUnit && (
+        <PlaceholderCell style={{ textAlign: 'right', paddingRight: '14px' }}>
+          {dosesPerUnit}
+        </PlaceholderCell>
```
lache-melvin, line 70: thought it was helpful to show the default doses per unit in the placeholder, to help see the relationship between that and units

client/packages/invoices/src/StockOut/api/operations.graphql
```diff
@@ -27,6 +27,12 @@ fragment StockOutLine on InvoiceLineNode {
     doses
   }
 
+  itemVariant {
```
lache-melvin, line 30: added item variant to stock out line - so the correct num doses is shown in the detail view!

mark-prins:
> if the doses pref is on, this should always meet the dose quantity requested, rather than unit quantity?

yes thanks. that's the underlying assumption I'm working with. for distribution, the EPI teams are thinking in terms of dose numbers. given the complexity of trying to support a hybrid approach, and the uncertainty of need, let's keep to dose numbers.

> Backend does always FEFO

I like the over-engineered front end approach, but concede that it is, for current purposes, over engineered. the most important thing is the consistency, and I would use the backend approach. though it does need to be updated to respect VVM not always FEFO

roxy-dao (changes requested):
Thanks Laché. Works great, code looks good too. Though...

Silly me... thought I had added stock to my store... went to allocate, got the generic `2 lines not allocated` message 🤦‍♀️

Could we make the generic not allocated message clearer?
`{{ count }} lines not allocated because.....??`

> Backend does always FEFO, will round up to nearest pack size. So [this test fails](https://github.com/msupply-foundation/open-msupply/compare/7959-backend-allocate-in-doses...7959-backend-allocate-in-doses(prevent-overallocation-test))...
I think should be a decision from @mark-prins / team, whichever way we go, it should probably be consistent no matter which button you click!

Also, my understanding is that OG does FEFO, so frontend logic should also follow this unless a pref trumps it 🤔

lache-melvin:
There is a separate issue for frontend to be updated to do FEFO: https://github.com/msupply-foundation/open-msupply/pull/7971#pullrequestreview-2895735821

Also think the error messages are a bit out of scope for this PR, to support doses - agree they would be helpful but prob not urgent for 2.8?

roxy-dao (approved):
Can create another issue for above request

PR #7975, opened 2025-05-31
7960 st final touches

### LINKED ISSUE #7960: ST dhis2 final touches
This is mainly carry over from PR reviews to
* Use plugin user
* Rename messages to sync messages

Also final provision for plugin to:
* Get active stores on site

And finally I had issues in testing built app, processors were stuck, this was replicated in actix_web runtime or tokio multithread runtime with 2 worker threads

andreievg:
Fixes #7960 

# 👩🏻‍💻 What does this PR do?

Have a look at commit message for summary.

Back to actix-web runtime. I had issues with use graphql within, when testing a built version on ST demo server. It was blocking processors, and was blocking sync operations. I tried various things, what worked was adding spawn_blocking in plugin processor, around boajs call. This all seemed too complicated, and I ended up just starting a new runtime in a new thread, this seemed like the safest solution. So reverted most of the change from parent PR, apart from using tokio::spawn for loaders, I think this is still better then async-std which is now removed from deps.

I was very worried about 'undefined' as JsValue causing a panic, used boajs from this PR: https://github.com/boa-dev/boa/pull/4212, required changed to use new API and bumping serde (couldn't figure out why it didn't auto bump and find the right version). The commit that i used is not available in boaJS version yet, so had to use it.

Other changes are described in the commit description and in self review comments

# 🧪 Testing

There have been quite a lot of testing for this in the built versions on ST demo server. Going to do a couple more changes and add more testing instructions (at least for the example plugin, in repo), also will link actual ST dhis integration plugin here

[Example plugin was updated](https://github.com/msupply-foundation/open-msupply-plugins/tree/Process-messages-to-create-prescription), also has test cases now

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/plugins/backendCommon/generated/ActiveStoresOnSite.ts
```diff
@@ -0,0 +1,4 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+import type { Store } from './Store';
```
andreievg, line 2: All these type changes due to adding get active stores on site and renaming message to sync_message

client/packages/plugins/backendCommon/utils.ts
```diff
@@ -45,3 +45,8 @@ export const fromSqlDateTime = (datetime: string) => {
   const withoutMillisecods = datetime.split('.')[0] || datetime;
   return new Date(`${withoutMillisecods.split(' ').join('T')}Z`);
 };
+
+export const toNaiveDateTime = (date: Date) => {
```
andreievg, line 49: This one is needed for creating message, we are using repository so need to provide correct format for sync message row

server/Cargo.toml
```diff
@@ -28,7 +28,7 @@ reqwest = { version = "0.12.12", features = [
   "multipart",
   "blocking",
 ], default-features = false }
-serde = "1.0"
```
andreievg, line 31: As per PR description

server/android/src/android.rs
```diff
@@ -76,10 +76,7 @@ pub mod android {
         // run server in background thread
         let thread = thread::spawn(move || {
             // This code is from expanding macro in main.rs
-            tokio::runtime::Builder::new_multi_thread()
-                .enable_all()
-                .build()
-                .expect("Failed building the Runtime")
+            actix_web::rt::System::new()
```
andreievg, line 79: This is a revert from changes in base PR, see PR description

server/repository/src/migrations/v2_07_04/create_sync_message_table.rs
```diff
@@ -5,34 +5,35 @@ pub(crate) struct Migrate;
 
 impl MigrationFragment for Migrate {
     fn identifier(&self) -> &'static str {
-        "create_message_table"
+        "create_sync_message_table"
```
andreievg, line 8: Changing identifier would mean migration will re-run (in case devs testing this already ran the previous migration). Don't need to delete possible message table

server/server/src/lib.rs
```diff
@@ -261,8 +261,6 @@ pub async fn start_server(
     BoaJsContext::new(
         &service_provider,
         PluginExecuteGraphql(graphql_schema.clone()),
-        // See explanation in BoaJsContext::new and in do_async_blocking
-        tokio::runtime::Handle::current(),
```
andreievg, line 265: This is not needed anymore, we are running async in sync method with a new thread, as per PR description

server/service/src/boajs/call_method.rs
```diff
@@ -50,9 +52,9 @@ where
 
     // Wait for module to load
     let promise = module.load_link_evaluate(context);
-    context.run_jobs();
+    context.run_jobs()?;
     match promise.state() {
-        PromiseState::Fulfilled(JsValue::Undefined) => {}
+        PromiseState::Fulfilled(v) if v == JsValue::undefined() => {}
```
andreievg, line 57: Api changed for JsValue

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "client/packages/plugins/stPlugins"]
```
jmbrunskill, line 1: Do we want to check this in?

andreievg: Thanks: c4a504cfc538e8b292556749f7362dcbadc70496

client/packages/plugins/backendCommon/generated/ChangelogTableName.ts
```diff
@@ -63,4 +63,4 @@ export type ChangelogTableName =
   | 'FormSchema'
   | 'PluginData'
   | 'Preference'
-  | 'Message';
+  | 'SyncMessage';
```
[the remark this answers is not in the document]
andreievg: I'll double check in plugin aswell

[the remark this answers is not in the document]
lache-melvin: Oh lol, this was in your migration branch, maybe its fine now. All good thanks sorry for noise

lache-melvin (commented):
Code looks good, haven't manual tested/looked at plugin yet though

lache-melvin:
OH actually I forgot about the message syncing - looks like it won't work to me, will validate in the morning 🙏

client/packages/plugins/backendCommon/generated/SyncMessageRow.ts
```diff
@@ -0,0 +1,14 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+import type { SyncMessageRowStatus } from './SyncMessageRowStatus';
+import type { SyncMessageRowType } from './SyncMessageRowType';
+
+export type SyncMessageRow = {
+  id: string;
+  to_store_id: string | null;
+  from_store_id: string | null;
```
lache-melvin, line 8: I guess it would be a breaking change at this point, but would be nice if these got mapped to camel case in TS conversion 👀

jmbrunskill (commented):
I've had a quick read of this (not tested) keen to understand a bit more about sync although I think that's primarily a legacy concern? If we do sync_messages OMS <-> OMS then we probably need more of a discussion.

Also would like to understand plugin user a bit more, I think you said you'd document that @andreievg ?

andreievg:
> huh ok I take it back, the sync messages are getting synced back... can you explain to me how though?? I think this should mean, if table_name is for Remote sync style, should only get synced if store id is set on changelog??

`create_filtered_outgoing_sync_query` use used in `outgoing_sync_records_from_central` so for v6 sync.

For v5 we go with this one: 

https://github.com/msupply-foundation/open-msupply/blob/6086afa3a3494c3bd99ec44a78be3d96e772da27/server/service/src/sync/mod.rs#L52-L56

lache-melvin:
OHhh ok thanks, sorry thought all the central ones had `_central` suffix, guess not all 😁

andreievg:
> Do they have release version scheduled

I couldn't find an issue for that, based on their release page it's overdue, but it's a breaking change, so not sure, the commit i pointed to is in main

andreievg:
> Hm - long term, seems frustrating to have to do this in every plugin implementation that wants to send a date? What point does it error? Wondering if its in our core code that we can handle it or almost an sdk if developing OMS plugins in TS

I've done something like this with sql helpers, although there are still helpers that need to be used in date conversions there.

This is also a helper that's available to all plugins, but yep we have to use it explicitly for the sync message repository in plugin

server/service/src/sync/translations/sync_message.rs
```diff
@@ -1,25 +1,25 @@
...
+    SyncMessageRowStatus, SyncMessageRowType,
 };
 use serde::{Deserialize, Serialize};
 
 use super::PushTranslateResult;
 
 /// Message from mSupply Central Server
 #[derive(Deserialize, Serialize, Debug)]
-pub struct LegacyMessageRow {
+pub struct LegacySyncMessageRow {
```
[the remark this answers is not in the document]
andreievg: Agree, and it was an easy change: [ab0c065](https://github.com/msupply-foundation/open-msupply/pull/7975/commits/ab0c0653301b0e538b7241820d94de6028de4191)

andreievg:
> I've had a quick read of this (not tested) keen to understand a bit more about sync although I think that's primarily a legacy concern? If we do sync_messages OMS <-> OMS then we probably need more of a discussion.

Yep sync_message is really a legacy sync idea, thing, I am having trouble finding wiki docs for it. But basically it will sync to to_store_id site, if not specified will sync to central. Can be created anywhere, but editing only allowed on to_store_id. Empty to_store_id or from_store_id means central server site. 

Jonna found a doc for me,[ it's early stage of the spec](https://wiki.sussol.net/doku.php/msupply:specifications:sync:messages)

> Also would like to understand plugin user a bit more, I think you said you'd document that @andreievg ?

Sorry no documentation beyond what's in this PR description and previous PR. Basically graphql queries done through plugin are overwritten with plugin_user, plugin_user goes through auth without authentication, with all permissions for now (apart from context permissions for things like programs), the idea for future is to set those permissions in configuration.  Plugin user override happens only in call where graphql is called from plugin, there is no way to sneak that user override in from normal graphql queries over http.

PR #7976, opened 2025-05-31
New status for og migrated inbound shipment to be considered as shipped

### LINKED ISSUE #7765: "New" inbound shipment not being updated to "Shipped" when OG customer invoice is finalised
## What went wrong? 😲

A user is reporting they have an inbound shipments that are "New", even thought the customer invoice in OG is finalised. NOTE: v2.6.4

In my inspection the first of the inbound shipments is also missing its only line.

There are no errors in system log, suggesting to me that it's just early returning from the processor if it is running at all.

https://cerb.msupply.org/profiles/ticket/KN-1794-MK

## Expected behaviour 🤔

A finalised customer invoice in OG should always be "Shipped" in OMS

## How to Reproduce 🔨

Someone figure it out 😉 I'd have guessed a migration issue where there was a preexisting new invoice and the processor exclusively does "PICKED" invoices:

https://github.com/msupply-foundation/open-msupply/blob/23aef4551440a8189906dcdae5d7a4f23f69e2d4/server/service/src/processors/transfer/invoice/update_inbound_invoice.rs#L62-L65

However checking data the ID is in the OMS format, so the record was generated in OMS.

## Your environment 🌱

- Open mSupply Version: 2.6.4
- Legacy mSupply Central Server Version: 8.0.17

- Platform: Unsure

- Database type: Unsure

andreievg:
Fixes #7765 

# 👩🏻‍💻 What does this PR do?

When migrating og data for si that has new status we should be translating status to shipped, including settings shipped datetime to created datetime

Added a translation test, which was a bit cumbersome, I think because of number of fields that need to be specified, so I used existing data, which still required a lot of work (but in the process i noticed the shipped date needs to be updated)

# 🧪 Testing

[This datafile](https://drive.google.com/drive/u/1/folders/1R-Z6CwJ7s099jYtne_Nta7I7Y906KMAy) has a transfer create in og, ready to be synced on initialisation. (user1 -> user1 for OG related stuff, like registering etc.. test -> pass for site initialisation test -> pass for omSupply user). You will need to check inbound shipment in trasnfertest store

Otherwise, create si in new status as transfer, and change the site for the receiving store.

Before this change after taking invoice on hold you cannot change status, with this change you can change status, and you should see shipped status too

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/test/test_data/invoice.rs
+++ b/server/service/src/sync/test/test_data/invoice.rs
@@ -98,119 +98,125 @@ const TRANSACT_1: (&str, &str) = (
       "om_expected_delivery_date": ""
   }"#,
 );
-fn transact_1_pull_record() -> TestSyncIncomingRecord {
-    TestSyncIncomingRecord::new_pull_upsert(
-        TABLE_NAME,
-        TRANSACT_1,
-        InvoiceRow {
-            id: TRANSACT_1.0.to_string(),
-            user_id: Some("MISSING_USER_ID".to_string()),
-            store_id: "store_b".to_string(),
-            name_link_id: "name_store_a".to_string(),
-            name_store_id: Some("store_a".to_string()),
-            invoice_number: 1,
-            r#type: InvoiceType::InboundShipment,
-            status: InvoiceStatus::Delivered,
-            on_hold: false,
-            comment: None,
-            their_reference: None,
-            transport_reference: None,
-            created_datetime: NaiveDate::from_ymd_opt(2021, 7, 30)
+
+fn transact_1_pull_row() -> InvoiceRow {
+    InvoiceRow {
+        id: TRANSACT_1.0.to_string(),
+        user_id: Some("MISSING_USER_ID".to_string()),
+        store_id: "store_b".to_string(),
+        name_link_id: "name_store_a".to_string(),
+        name_store_id: Some("store_a".to_string()),
+        invoice_number: 1,
+        r#type: InvoiceType::InboundShipment,
+        status: InvoiceStatus::Delivered,
+        on_hold: false,
+        comment: None,
+        their_reference: None,
+        transport_reference: None,
+        created_datetime: NaiveDate::from_ymd_opt(2021, 7, 30)
+            .unwrap()
+            .and_hms_opt(0, 0, 0)
+            .unwrap()
+            + Duration::seconds(47046),
+        allocated_datetime: None,
+        picked_datetime: None,
+        shipped_datetime: None,
+        delivered_datetime: Some(
+            NaiveDate::from_ymd_opt(2021, 7, 30)
                 .unwrap()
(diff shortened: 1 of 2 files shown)


server/service/src/sync/test/test_data/invoice.rs
```diff
@@ -98,119 +98,125 @@ const TRANSACT_1: (&str, &str) = (
       "om_expected_delivery_date": ""
   }"#,
 );
+
+fn transact_1_pull_row() -> InvoiceRow {
```
andreievg, line 102: wanted to re-use this for the test, so I don't have to re-define a lot of fields, probably would have been quicker if i copied it

server/service/src/sync/translations/invoice.rs
```diff
@@ -755,7 +768,9 @@ fn invoice_status(invoice_type: &InvoiceType, data: &LegacyTransactRow) -> Optio
         // inbound
         InvoiceType::InboundShipment | InvoiceType::CustomerReturn => match data.status {
             LegacyTransactStatus::Sg => InvoiceStatus::New,
```
lache-melvin, line 770: Should we not map this one as well? Per OG docs, `sg = suggested (has been opened since first creation, still editable, stock not yet added to mSupply)` ... just because I opened the invoice at one stage, would assume it should still show up as shipped so can edit after migration to OMS 👀

andreievg: > OK POST DIGGING: actually sg is not a thing for SI's, we've removed from docs now... however, if by some munging you end up with an SG transferred invoice, we get the same issue:

I've added a comment for now 2689081422c54281c6aba7d8a2a688d064752407, I hope that's enough

server/service/src/sync/test/test_data/invoice.rs
```diff
@@ -1466,6 +1472,58 @@ fn cancelled_prescription_push_record() -> TestSyncOutgoingRecord {
...
+    let json_body = TRANSACT_1
+        .1
+        .to_string()
+        .replace(TRANSACT_1.0, TRANSACT_MIGRATE_OG_SI_STATUS_ID)
+        .replace(r#""status": "cn""#, r#""status": "nw""#)
+        .replace(
+            r#"confirm_date": "2021-07-30"#,
+            r#"confirm_date": "0000-00-00"#,
+        )
+        .replace(r#"confirm_time": 47046"#, r#"confirm_time": 0"#);
```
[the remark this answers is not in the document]
andreievg: > Hm yeah, all good now but I do think the duplication is better for these tests, makes it easier to tell exactly what data is present for a record when a test is failing

Sounds good, I think a two phased test might have been better, serde and then out translations (from legacy row to row)

PR #7988, opened 2025-06-03
7617 Display item variant detail in outbound shipment

### LINKED ISSUE #7617: Allow viewing item_variant detail when creating/editing outbound invoices.
Allow viewing item_variant detail when creating/editing outbound invoices.

## Tasks 📝

- [ ] If an item_variant is assigned to the stockline/shipment_line, Add an info icon to the Batch field in out bound shipments
- [ ] When the info icon is clicked a hover over (similar to store selector with close button shows the item_variant information)

### Acceptance Criteria ✅
- [ ] Can see item variant details when selecting a batch in an outbound shipment

### Moneyworks Jobcode 🧰
NIGGF:DSE

Image

noel-yeldos:
Fixes #7617

# 👩🏻‍💻 What does this PR do?
For stocklines in outbound shipment that are item variants, display an icon that opens a popover which displays the details for the variant

- Create a new Cell component that conditionally renders the detail icon if stockline is a variant
- Create new variant details component -> Loosely modelled off variants tab in items catalogue
- Update graphql to query necessary fields
- Update batch column in outbound shipments to use new Cell 

`Line edit view`
Screenshot 2025-06-03 at 12 15 30

`Variant with details where information is filled in`
Screenshot 2025-06-03 at 12 15 59

`Variant with details where no information is filled in`
Screenshot 2025-06-03 at 12 16 11

`Item variants tab for comparison`
Screenshot 2025-06-03 at 12 16 54

The layout of the popover still needs a bit of work so some feedback to improve that would be great.

# 🧪 Testing

- [ ] Have an item with variants
- [ ] Ensure you have stock for the variants and also for the non-variant
- [ ] Open Outbound Shipments and create a new shipment
- [ ] Add the item with variants
- [ ] In the line edit view observe that variant stockline has an icon in the batch column and non-variant stockline does not have icon
- [ ] Click on the icon and observe a popover which has details of the variant
- [ ] Compare variant details with item catalogue to ensure that data is correct

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/components/Cells/BatchCell/BatchWithVariantCell.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
noel-yeldos, line 1: New cell to be able to display the icon popover if stockline is a variant

client/packages/common/src/ui/layout/tables/components/Cells/BatchCell/VariantPopoverDetails.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    {t('label.pack-size')}:{' '}
+                    {item.packSize ?? UNDEFINED_STRING_VALUE}
+                  </Typography>
+                  <Typography>
+                    {t('label.volume-per-unit')}:{' '}
+                    {item.volumePerUnit ?? UNDEFINED_STRING_VALUE}
+                  </Typography>
+                </Box>
+              ))}
+            </Stack>
```
noel-yeldos, line 87: Not sure if this needs to be a table like in the variants tab in Item detail view. Could I repurpose `ItemPackagingVariantsTable`?

lache-melvin (changes requested):
Ah heck, sorry Noel, should have had a chat about this first.

In general, it's a good idea to try to reuse existing components where possible - so we add as little extra surface area as we can, and so if we change something it gets applied everywhere.

In this case, we have existing item variants selector, used in Stock edit, inbound etc - I think more consistent to reuse this one, just disabled.

Packaging variants, and all that other data, is currently only used for GAPS, for now just needing variant name, and maybe doses/vvm if pref on (per existing selector)

Here is example branch, taking same concept you had, with batch and info button, but using the selector 👀 

https://github.com/msupply-foundation/open-msupply/compare/7617-view-item-variant-outbound-line-edit...7617-view-item-variant-outbound-line-edit(reuse-item-variant-selector)

lache-melvin:
Ah yeah, nice shout - after all that, it's sounding like item variants might be evolving a bit in the next release, so I'm apprehensive to spend more time here. That's also making me think, let's go with the table, least amount of code added - if it turns out we keep them and need more detail, we can add that then. 

Or, also happy with - take the component you have, but instead of defining new content structure, reuse the existing item variant section from the Item detail page. Think that would be more work though, wouldn't want edit buttons, bundles section...

noel-yeldos:
> Ah yeah, nice shout - after all that, it's sounding like item variants might be evolving a bit in the next release, so I'm apprehensive to spend more time here. That's also making me think, let's go with the table, least amount of code added - if it turns out we keep them and need more detail, we can add that then.
> 
> Or, also happy with - take the component you have, but instead of defining new content structure, reuse the existing item variant section from the Item detail page. Think that would be more work though, wouldn't want edit buttons, bundles section...

Thanks @lache-melvin I've gone with your suggested change :)

lache-melvin (approved):
🚀

PR #7990, opened 2025-06-03
7910 Show campaign in Outbound Shipment

### LINKED ISSUE #7910: Show campaign when selecting outgoing stock
When doing an outbound shipment or prescription,
if the server has any campaigns created
Show a new column with the campaign in the stock selection window

CarlosNZ:
Fixes #7910

# 👩🏻‍💻 What does this PR do?

Adds a column to display the "Campaign" in the Outbound Shipment "Line Edit" modal:

Screenshot 2025-06-03 at 2 08 24 PM

Only displays if there are campaigns defined in the system

Was quite a bit more work than originally anticipated, as I had to add to the invoice line mutations to set the campaign ID when they are added or updated. The "campaign" can be fetched from the Stock line when the item is first being added to the invoice, but subsequent displays need to fetch it from the invoice line (that's the way the current queries work, so makes sense to include it in existing functionality).

# 🧪 Testing

- [ ] Add some campaigns to the system (using Mange -> Campaigns (on central server)
- [ ] Make sure some vaccine item stock lines have campaigns associated with them (via Inventory -> View Stock)
- [ ] Go to an Outbound Shipment
- [ ] Add one of the vaccine items
- [ ] Note the "Campaign" column in the edit modal
- [ ] When adding an item that is *not* part of a campaign, note the table cell is blank
- [ ] Delete all campaigns and confirm "Campaigns" column not visible at all any more

# 📃 Documentation

  1. Campaigns

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Manage/Campaigns/api/hooks/useCampaigns.ts
+++ b/client/packages/system/src/Manage/Campaigns/api/hooks/useCampaigns.ts
@@ -103,7 +103,7 @@ const useGetList = (queryParams?: ListParams) => {
 
 const toSortInput = (sortBy?: SortBy<CampaignRowFragment>) => ({
   desc: sortBy?.isDesc,
-  key: sortBy?.key as CampaignSortFieldInput,
+  key: (sortBy?.key as CampaignSortFieldInput) ?? CampaignSortFieldInput.Name,
 });
 
 const useUpsertCampaign = () => {
--- a/server/service/src/invoice_line/stock_out_line/update/generate.rs
+++ b/server/service/src/invoice_line/stock_out_line/update/generate.rs
@@ -86,6 +86,8 @@ fn generate_line(
         foreign_currency_price_before_tax,
         sell_price_per_pack: invoice_line_sell_price_per_pack,
         cost_price_per_pack: invoice_line_cost_price_per_pack,
+        donor_link_id,
+        campaign_id,
         ..
     }: InvoiceLineRow,
     ItemRow {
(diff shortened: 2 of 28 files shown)


client/packages/system/src/Manage/Campaigns/api/hooks/useCampaigns.ts
```diff
@@ -103,7 +103,7 @@ const useGetList = (queryParams?: ListParams) => {
 
 const toSortInput = (sortBy?: SortBy<CampaignRowFragment>) => ({
   desc: sortBy?.isDesc,
-  key: sortBy?.key as CampaignSortFieldInput,
+  key: (sortBy?.key as CampaignSortFieldInput) ?? CampaignSortFieldInput.Name,
```
CarlosNZ, line 106: This hook was giving an error when called with no args, but it shouldn't.

server/service/src/invoice_line/stock_out_line/update/generate.rs
```diff
@@ -135,9 +137,9 @@ fn generate_line(
         item_variant_id,
         vvm_status_id,
         linked_invoice_id: None,
-        donor_link_id: None,
+        donor_link_id,
```
CarlosNZ, line 140: The donor_link_id needs to be passed through otherwise it gets reset to "Null" by repository layer (unrelated to this issue, but discovered while debugging this one with @jmbrunskill )

PR #8001, opened 2025-06-03
7992 Add population percentage to demographic table

### LINKED ISSUE #7992: Need to add current "population_percentage" to "demographic" table
## What went wrong? 😲

The "demographic_indicator" table is currently not synced as it was envisaged as a GAPS feature and therefore only needed on the central server. However, with vaccination courses, we are (sometimes, in a plugin) using the population percentage for a given demographic (which is accessed via the "vaccination_course" View), and this is stored on the "demographic_indicator" table.

So any stores that use a plugin that needs the population percentage won't work if it's not a central server.

## Expected behaviour 🤔

Stores should have access to population_percentage values.

There are two possible solutions:

1. Sync the demographic_indicator table as well
2. Add an additional "population_percentage" field to the "demographic" table, and update this value from the associated demographic_indicator whenever it is updated. We don't currently support multiple indicators per demographic in the UI, so it'll be a one-to-one relationship at this point. In the future, if we allow projecting multiple different "indicator" scenarios in the UI, we will also add the option to specify a particular indicator as the "current" one, which is what would go into the "population_percentage" field on the main demographic record.

After discussion with @jmbrunskill, we thought option 2 would be best, as it keeps the idea of GAPS functionality being a central server feature only, but allows access to current values from remote stores. It will also simplify the Vaccination_course view, as it can get the associated demographic from just the "demographic" table, rather than having to drill down via the "demographic_indicator" table as it does currently.

CarlosNZ:
Fixes #7992

# 👩🏻‍💻 What does this PR do?

- Migration to add "population_percentage" to "demographic" table
- Funcionality to automatically update that field whenever a "demographic_indicator" is added/edited
- Change "vaccination_course" view to use percentage from this field rather the demographic_indicator

# 🧪 Testing

Not much for QA to test here, other than confirming that the forecasting plugin works on a non-central system

Devs, check database is modified as expected

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/v2_08_00/add_population_percentage_to_demographic.rs
+++ b/server/repository/src/migrations/v2_08_00/add_population_percentage_to_demographic.rs
@@ -11,17 +11,32 @@ impl MigrationFragment for Migrate {
         sql!(
             connection,
             r#"
-                ALTER TABLE demographic ADD COLUMN population_percentage {DOUBLE} NOT NULL DEFAULT 0;
+                ALTER TABLE demographic ADD COLUMN population_percentage {DOUBLE} NOT NULL DEFAULT 0;  
+            "#,
+        )?;
 
+        // Populate the new column with the existing values from demographic_indicator if they exist
+        // They should only exist for the central server, so this will only affect the central server
+        sql!(
+            connection,
+            r#"
                 UPDATE demographic SET population_percentage = (
                     SELECT population_percentage FROM demographic_indicator
                     WHERE demographic_id = demographic.id
                 ) WHERE id in (
                     SELECT demographic_id FROM demographic_indicator
-                    WHERE population_percentage IS NOT NULL
                 );
             "#,
         )?;
+
+        // Create changelog entries for all the existing demographic records, so they'll be synced with their new population_percentage
+        // We limit it to records that have a corresponding demographic_indicator, so this actually only affects the central server, otherwise we sync 0 records to central!
+        sql!(
+                connection,
+                "INSERT INTO changelog (table_name, record_id, row_action) SELECT 'demographic', id, 'UPSERT' FROM demographic WHERE id in (
+                    SELECT demographic_id FROM demographic_indicator
+                );"
+            )?;
 
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


server/repository/src/migrations/v2_08_00/add_population_percentage_to_demographic.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                ALTER TABLE demographic ADD COLUMN population_percentage {DOUBLE} NOT NULL DEFAULT 0;  
+            "#,
+        )?;
+
+        // Populate the new column with the existing values from demographic_indicator if they exist
```
jmbrunskill, line 18: I found the old code was failing on remote site as the `demographic_indicator` records were missing.

jmbrunskill (approved):
@CarlosNZ there's quite a few edge cases to consider here for the new field and sync.

I had issues initialising my remote site, which highlighted it to me, have pushed some commits, have a look to see if you're happy with them?

I think I've covered all the cases but definitely worth having a think to see if there's something else to consider....

server/repository/src/migrations/v2_08_00/add_population_percentage_to_demographic.rs
```diff
@@ -0,0 +1,54 @@
...
+        sql!(
+                connection,
+                "INSERT INTO changelog (table_name, record_id, row_action) SELECT 'demographic', id, 'UPSERT' FROM demographic WHERE id in (
+                    SELECT demographic_id FROM demographic_indicator
+                );"
+            )?;
+
+        // If we've update and synced the records on central, the remote site might not have had that column yet.
+        // So we need to re-integrate the records on the remote site on upgrade, this should be safe as only the remote site should have data in the sync_buffer.
+        // Would be nice to limit this so Central server doesn't run this code at all but it doesn't know it's central till first sync.
```
jmbrunskill, line 43: Tried to add inline comments for the justifcations, I think it will help people understand what happened if it comes up with a problem and also if someone wants to copy a pattern it's good to have extra context I think?

CarlosNZ:
@jmbrunskill thanks for the updates!

Issue #7998, opened 2025-06-03
Internal order Forecast Amount is empty instead of displaying 0


alainsussol:
## What went wrong? 😲

The column is empty instead of having a value like 0 when there is nothing to display.

![Image](https://github.com/user-attachments/assets/8a1201db-d065-462a-8e83-cf98a46f6270)

## Expected behaviour 🤔
The column should not be empty it should display 0 when there is nothing to display.

## How to Reproduce 🔨

Steps to reproduce the behaviour:
Setup for forecasting calculation
1. Create a program internal order
2. Choose the vaccine program you have configured
3. See error. the forecast Amount column doesn't show any value.

## Your environment 🌱

- Open mSupply Version: 2.8.0RC6
- Legacy mSupply Central Server Version: v8.2.0

- Platform: omSupply Desktop

- Database type: SQLite

alainsussol:
I agree that it is probably better as you say to put a symbol like "-", or "N/A" rather than just empty. The problem I find with empty is that it is not quite clear that it is intentionally empty :)

jmbrunskill:
`UNDEFINED_STRING_VALUE` is defined in the frontend for this purpose.

CarlosNZ:
Closed by #8010

alainsussol:
# 🧪 Testing



- [ ] Install latest version of plugins from the above branch (for front-end can use either development src, or compiled `bundle.json`
- [ ] Go to Internal Orders -> Click an order -> Click a line 
- [ ] Observe "Forecasting quantity" info is in the correct place (as in screenshot above)
- [ ] When there is no forecast quantity (e.g. non-vaccine items, or not configured), both the table column and the modal field should display "-" for the value
- [ ] Check the forecast value appears correctly in list when adding an item (or deleting and re-adding), as per #8024

alainsussol:
However, I found this #8086

PR #8010, opened 2025-06-04
#7993 Change Forecasting field to match new UI

### LINKED ISSUE #7993: "Forecasting" plugin field appears incorrectly in new Internal Order detail modal
## What went wrong? 😲

Since the Internal Order modal has been given a UI facelift, the additional fields added via plugin are not appearing in the correct place:

image

## Expected behaviour 🤔

Plugin field should be part of the main data fields section

## How to Reproduce 🔨

1. Install "Forecasting plugins"
2. Go to Internal Order, click an order, open a line in modal -- note incorrect field placement

## Your environment 🌱

- Open mSupply Version: 2.8.0
- Platform: Browser

CarlosNZ:
Fixes #7993 (& #7998 & #8024)

# 👩🏻‍💻 What does this PR do?

Fixes the "Forecasting" front-end plugin to show the detail field in the new-look UI:

Screenshot 2025-06-04 at 12 27 55 PM

You will need the latest version of the plugin on branch "handle-doses-and-units-correctly" in the [open-msupply-forecasting-plugins](https://github.com/msupply-foundation/open-msupply-forecasting-plugins/pull/3) repo.

I'm assuming we're only going to show the forecast amount in Units (for now), which is what the back-end is working in, so I've hard-coded those values into the `ValueInfoRow` props.

# 🧪 Testing

- [ ] Install latest version of plugins from the above branch (for front-end can use either development src, or compiled `bundle.json`
- [ ] Go to Internal Orders -> Click an order -> Click a line
- [ ] Observe "Forecasting quanity" info is in the correct place (as in screenshot above)
- [ ] When there is no forecast quantity (e.g. non-vaccine items, or not configured), both the table column and the modal field should display "-" for the value
- [ ] Check the forecast value appears correctly in list when adding an item (or deleting and re-adding), as per #8024

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/plugins/api/hooks/usePluginData.ts
```diff
@@ -26,7 +34,7 @@ export const usePluginData = ({ pluginCode, filter }: PluginDataProps) => {
   };
 
   const { data, isError, isLoading } = useQuery({
-    queryKey: [PLUGIN_DATA, storeId, pluginCode],
+    queryKey: [PLUGIN_DATA, storeId, pluginCode, queryKey],
```
CarlosNZ, line 37: I realised that we needed a way to invalidate the query for individual items, so added a `queryKey` prop to this hook. I'm using it to pass in the line ID for the individual item modal view. Otherwise, the query wasn't fetching the new value when going directly to the next item with "OK and Next".

jmbrunskill (approved):
Looks good. Field's displaying correctly.

However...

I'm still not getting my remote store forecasts. Will re-test when you finish the other plugin updates. Maybe still something wrong with my setup...

jmbrunskill (approved):
Looks good, just check my merged changed before merging

PR #8027, opened 2025-06-04
8000 Outbound shipments make total row sticky

### LINKED ISSUE #8000: Placeholder line when issuing in Outbound Shipments does not clear when switching between measurements and there is enough stock
## What went wrong? 😲

As per title - for example, I am issuing in packs of different sizes, doses or units

When switching between all the options, there sometimes can be not enough stock -> so a placeholder line is added. But let's say I switch to another measurement - there is enough stock to issue however much I want, but the placeholder line does not clear.

https://github.com/user-attachments/assets/c8345ec6-183a-4260-bf31-6dfd9c795ff9

## Expected behaviour 🤔

If there is enough stock to issue, placeholder line should clear. Especially because from that modal there is no way to remove it (it's added automatically), so it can be not intuitive for the user to click OK before they are able to remove it on the next screen

## How to Reproduce 🔨

1. Have Dose Management on for your store and have an item with different measurements setup from different batches, for example different Pack Sizes - in my example, there are 3 pack sizes, doses and units (so 5 in total)
2. Go to Distribution -> Outbound Shipments and create a new shipment
3. Add your item in -> try an issue more stock than you have in some packs of 5 for example (so only one batch line is used)
4. See placeholder line is added
5. Now switch to something like units, where every batch line can be used
6. Note you have enough quantity to issue -> but placeholder line is still there
7. No way to remove it from that model unless you click OK and remove it from the next screen

## Your environment 🌱

- Open mSupply Version: V2.8.0-RC6

- Database type: PostgreSQL

noel-yeldos:
Fixes #8000

# 👩🏻‍💻 What does this PR do?
Make total quantity row in `OutboundLineEditTable` sticky 

The change was made to reduce confusion when placeholder line gets added as user switches between issuing units.

https://github.com/user-attachments/assets/3138dadb-cc28-4c6e-934d-2a8f9b2ad2ec

I've currently left the TotalRow as an additionalRow. An alternative solution is to move the TotalRow outside the DataTable (however table styles won't be able to the row in this case)

# 🧪 Testing

- [ ] Navigate to Distribution -> Outbound Shipment
- [ ] Create a shipment and add an item that has enough batches to enable vertical scroll
- [ ] Observe that when you scroll the total quantity row is always shown (i.e is not hidden away below the last batch)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
@@ -33,11 +33,14 @@ const PlaceholderCell = styled(TableCell)(({ theme }) => ({
   color: theme.palette.secondary.main,
 }));
 
-const TotalCell = styled(TableCell)({
+const TotalCell = styled(TableCell)(({ theme }) => ({
   fontSize: 14,
   padding: '4px 12px 4px 12px',
   fontWeight: 'bold',
-});
+  position: 'sticky',
+  bottom: 0,
+  background: theme.palette.background.white,
+}));
 
 const PlaceholderRow = ({
   quantity,


lache-melvin (approved):
Looks great thanks Noel! Just pushed an extra couple bits:

When we have the extra columns on, sometimes the total row wasn't spanning the full width:

Screenshot 2025-06-05 at 11 58 52 AM

And replaced the divider with a border, so its always visible 🙏

client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEditTable.tsx
```diff
@@ -191,7 +196,10 @@ export const OutboundLineEditTable = ({
     );
 
   let extraColumnOffset = 0;
-  if (prefs?.manageVvmStatusForStock || prefs?.sortByVvmStatusThenExpiry) {
+  if (
+    item?.isVaccine &&
```
noel-yeldos, line 200: Add isVaccine check so that non-vaccine items are not affected by the extraColumnOffset

lache-melvin (approved):
ship ittt

PR #8031, opened 2025-06-05
7999 rename vaccine course is active

### LINKED ISSUE #7999: `Include in GAPS calculations` for vaccine courses is confusing/wrong
## What went wrong? 😲

The label `Include in GAPS calculations` for vaccine courses actually affects the is_active field in the database. This means it's excluded from the forecasting plugin currently.

`  "label.calculate-demand": "Include in GAPS calculations",`
```
        <Row label={t('label.calculate-demand')}>
          <Checkbox
            checked={draft?.isActive ?? true}
            onChange={e => updatePatch({ isActive: e.target.checked })}
          ></Checkbox>
        </Row>
```

## Expected behaviour 🤔

Maybe should rename the field so it's just defining active or not. Or potentially align the label with `calculate demand` which more intuatively enables it for forecasting.

I'm not sure/don't think we should use all the `in_active` ones for our forecasting as we don't want sandbox ones affecting ordering.

We probably need to separate out GAPS stuff from non-GAPS anyway in future, so a longer term solution will need to be found (probably a scenario planner that would allow different setups to be compared)

## How to Reproduce 🔨

1. Install forecasting plugin
2. Setup forecasting requirements
3. Uncheck `Include in GAPS calculations` for your vaccine course
4. See that now forecasts get generated for that vaccine

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

lache-melvin:
Fixes #7999

# 👩🏻‍💻 What does this PR do?

Have opened this issue to update GAPS dashboards https://github.com/msupply-foundation/msupply-dashboard/issues/249

There is a breaking change to the graphQL API... I think changing the insert/update inputs is okay, should be central only mutation, and I'm not aware of reports/integrations against vaccine_course, definitely not with is_active taken into consideration... so probably safe?

# 🧪 Testing

- [ ] On OMS central, with at least one immunisation program set up from OG
- [ ] You can still create a vaccine course for the program
- [ ] You can edit the `use in GAPS calculations` field
- [ ] Have a remote OMS site using old version
- [ ] Changes from OMS central still sync successfully to the remote site (you can see "use in GAPS calculations" value from OMS central in the "is_active" field in remote OMS database)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/types/src/types/vaccine_course.rs
+++ b/server/graphql/types/src/types/vaccine_course.rs
@@ -40,8 +40,8 @@ impl VaccineCourseNode {
         self.row().coverage_rate
     }
 
-    pub async fn is_active(&self) -> bool {
-        self.row().is_active
+    pub async fn use_in_gaps_calculations(&self) -> bool {
+        self.row().use_in_gaps_calculations
     }
 
     pub async fn wastage_rate(&self) -> f64 {
--- a/server/repository/src/db_diesel/vaccine_course/vaccine_course_row.rs
+++ b/server/repository/src/db_diesel/vaccine_course/vaccine_course_row.rs
@@ -17,7 +17,7 @@ table! {
         program_id -> Text,
         demographic_id -> Nullable<Text>,
         coverage_rate -> Double,
-        is_active -> Bool,
+        use_in_gaps_calculations -> Bool,
         wastage_rate -> Double,
         deleted_datetime -> Nullable<Timestamp>,
     }
@@ -34,7 +34,8 @@ pub struct VaccineCourseRow {
     pub program_id: String,
     pub demographic_id: Option<String>,
     pub coverage_rate: f64,
-    pub is_active: bool,
+    #[serde(rename = "is_active")] // To prevent breaking change in v6 sync API
+    pub use_in_gaps_calculations: bool,
     pub wastage_rate: f64,
     pub deleted_datetime: Option<chrono::NaiveDateTime>,
 }
--- a/client/packages/common/src/types/schema.ts
+++ b/client/packages/common/src/types/schema.ts
@@ -282,6 +282,7 @@ export type AllocateOutboundShipmentUnallocatedLineNode = {
   issuedExpiringSoonStockLines: StockLineConnector;
   skippedExpiredStockLines: StockLineConnector;
   skippedOnHoldStockLines: StockLineConnector;
+  skippedUnusableVvmStatusLines: StockLineConnector;
   updates: InvoiceLineConnector;
 };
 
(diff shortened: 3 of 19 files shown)


server/graphql/types/src/types/vaccine_course.rs
```diff
@@ -40,8 +40,8 @@ impl VaccineCourseNode {
         self.row().coverage_rate
     }
 
-    pub async fn is_active(&self) -> bool {
-        self.row().is_active
+    pub async fn use_in_gaps_calculations(&self) -> bool {
+        self.row().use_in_gaps_calculations
     }
```
lache-melvin, line 45: I feel like this is a safe-enough breaking API change - but could deprecate instead if we think this might be more widely queried?

CarlosNZ:
> @jmbrunskill / @CarlosNZ just confirming the is_active filter on forecasting plugin has been removed right? By which I mean, I checked the repo, just want to ensure I'm not somehow missing something haha

The vaccine_course `is_active` field is being used to filter in the "vaccination_course" course View. So inactive courses are getting filtered out there, not in the plugin.

CarlosNZ:
> > @jmbrunskill / @CarlosNZ just confirming the is_active filter on forecasting plugin has been removed right? By which I mean, I checked the repo, just want to ensure I'm not somehow missing something haha
> 
> The vaccine_course `is_active` field is being used to filter in the "vaccination_course" course View. So inactive courses are getting filtered out there, not in the plugin.

Which I see is now causing the tests to fail:

```
Caused by:
    UNKNOWN ("column vc.is_active does not exist      14573"))
```

server/graphql/types/src/types/vaccine_course.rs
Chris-Petty: This will all get more fun with the Cambrian explosion of backend plugins

server/repository/src/db_diesel/vaccine_course/vaccine_course_row.rs
```diff
@@ -34,7 +34,8 @@ pub struct VaccineCourseRow {
     pub program_id: String,
     pub demographic_id: Option<String>,
     pub coverage_rate: f64,
-    pub is_active: bool,
+    #[serde(rename = "is_active")] // To prevent breaking change in v6 sync API
+    pub use_in_gaps_calculations: bool,
```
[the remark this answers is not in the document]
lache-melvin: that would be great, its hard to remember to consider all cases in development!

PR #8039, opened 2025-06-05
#7881 Inbound Shipment - round up to next whole packs

### LINKED ISSUE #7881: Number input issues in Inbound Shipments for Doses
## What went wrong? 😲

Issues: 

- When entering certain values in `Units Received` column, these values (and subsequently `Doses Received` values) show a lot of decimal points unprompted
- When backspacing these decimal points, the value of the cell changes to something like 9,999,999 and you need to remove the whole value before being able to type again

https://github.com/user-attachments/assets/bfbf6224-e51f-4763-9bad-1b18e04a7e0e

I am unsure why this happens for some values only (e.g. 55 and 56) but not for 54 or 25 for example

## Expected behaviour 🤔

No decimal points should be showing if they were not typed out by the user, also possibly limit these to 2 decimal points anyway? And backspacing should work as expected, e.g. backspacing the original value entered and not 9,999,999

## How to Reproduce 🔨

As I mentioned above, I am not sure why this happens for some values and not the others, so possibly best to replicate my entire setup and see if it happens on your end. 

1. Have an item that is a vaccine (in my case, it's a Covid Pfizer Vaccine) and is marked as a vaccine in OG mSupply
2. In your omSupply store, create three item variants for this item - I also have different doses per unit set for each of them - 1, 2 and 5
3. Create an Inbound Shipment with this item added -> add all three item variants as different batches to the shipment -> copy the exact values I have on the video
4. Once copied, play around with Units received - e.g. enter 25, 54, 55 and 56 and see if you get weird decimal points in the cells of Units Received and Doses Received

## Your environment 🌱

- Open mSupply Version: V2.8.0-RC2

- Database type: PostgreSQL

aimee-mcneil-melville:
Fixes #7881

# 👩🏻‍💻 What does this PR do?

When entering stock in inbound shipments, packs are rounded up to the nearest whole pack.

This resolves decimal issues from the issue and we want to support only whole packs for an inbound shipment.

Shows an alert message to inform the user that the units have been rounded up.

https://github.com/user-attachments/assets/35135a12-33e7-44d6-ac00-60d04d8351b2

I had some issues with NaN during development, looks to be resolved but something to keep an eye out for

Should the alert persist when adding another batch or changing other values eg packs received will also change the units received, but the alert no longer matches the inputs

# 🧪 Testing

- [ ] Go to Inbound Shipments -> select a shipment or create one -> add a vaccine item to the shipment
- [ ] Enter a`Pack Size` of more than 1
- [ ] Change the `Units Received` received to a number that would not equal a whole number of packs (eg 8 when pack size is 5) -> see `Units Received` round up to the next whole pack
- [ ] See the alert appear detailing the difference between entered units and rounded up units
- [ ] Enter `Units Received` as a multiple of pack size -> see the alert disappear
- [ ] See that changing values in `Pack Size` and `Packs Received` results in correct calculations
- [ ] Check that when `Units Received` has rounded up, that the rounded value is saved

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


noel-yeldos (changes requested):
Thanks Aimee! It works great on first user input. 

Subsequent inputs don't seem to re-calculate the rounded value and so the alert persisting is confusing. See video for unexpected behaviour:

https://github.com/user-attachments/assets/d08c5aa8-9a85-4c06-a330-86afc726bd02

aimee-mcneil-melville:
Thanks Noel!
I've added some additional state updates to clear the message, and returned the rounded value to the UI so that it will 're-round' when needed.
I also added a longer debounce time, as if a user is trying to type (especially on a tablet) we don't want this rounding too fast for them to input the correct value

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -125,6 +136,7 @@ export const QuantityTableComponent = ({
       unit: pluralisedUnitName,
     }),
     width: 100,
+    cellProps: { debounce: 1000 },
```
noel-yeldos, line 139: Just from testing this feels a bit long - 500 seems smoother to me. What do you think?

noel-yeldos (changes requested):
Thanks for the changes Aimee! It's working smoothly now :) 

Just one thing that may still cause some confusion. When click 'add batch' the message still persists. What do you think about hiding the previous message when we click 'add batch'?

client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabLayout.tsx
```diff
@@ -97,7 +101,10 @@ export const TabLayout = ({
             disabled={isDisabled}
             color="primary"
             variant="outlined"
-            onClick={addDraftLine}
+            onClick={() => {
+              addDraftLine;
```
noel-yeldos, line 105: @aimee-mcneil-melville just need to call the function here :) `addDraftLine();`

noel-yeldos (changes requested):
@aimee-mcneil-melville sorry just left a comment to call the `addDraftLines` function!

noel-yeldos (approved):
Thanks for the changes Aimee, LGTM! 🚀

PR #8046, opened 2025-06-06
Handle residual stock in stock lines, e.g. take all stock if nearly enough

### LINKED ISSUE #7964: When backdating prescriptions, Doses show as 1 even though it should be 0
## What went wrong? 😲

I have an item (Pfizer Vaccine) which was only added to the system on 26th of May, via Inventory Adjustments.

Image

Now, I am trying to issue some quantity for the item for 1st of May - which should be impossible because no stock existed. 

Prescription date:

Image

Doses available show 1 (should be 0): 

Image

And when trying to Save, I see the following error: 

Image

Once error is acknowledged, you try and save again and nothing happens :)

## Expected behaviour 🤔

I think it should prevent the user earlier in the process from saving the prescription - Doses should show 0 if none were available, and there should be a prompt saying no stock is available on this date for the item

## How to Reproduce 🔨

1. Find an item you want to test with, and note the first time ledger had an update, e.g. stock being added
2. Have Dose Management enabled in your store
3. Go to Dispensary -> Prescriptions and create a new prescription
4. In datetime picker, set a date to before your item's first ledger entry
5. Try and add that item -> it will allow you. Enter quantity in doses
6. Click Save -> encounter the error

## Your environment 🌱

- Open mSupply Version: V2.8.0-RC4

- Database type: PostgreSQL

jmbrunskill:
Fixes #7964

# 👩🏻‍💻 What does this PR do?

# 🧪 Testing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
@@ -254,9 +254,11 @@ const dosesSummary = (
     0
   );
 
+  const roundedDoses = NumUtils.round(totalDoses);
+
   const unitWord = t('label.doses-plural', {
-    count: NumUtils.round(totalDoses),
+    count: roundedDoses,
   });
 
-  return `${totalDoses} ${unitWord}`;
+  return `${roundedDoses} ${unitWord}`;
 };
--- a/server/service/src/invoice_line/stock_out_line/validate.rs
+++ b/server/service/src/invoice_line/stock_out_line/validate.rs
@@ -3,6 +3,17 @@ use repository::{
     StockLineFilter, StockLineRepository, StorageConnection,
 };
 
+const LAST_PACK_THRESHOLD: f64 = 0.001;
+
+pub fn adjust_for_residual_packs(available_packs: f64, requested_number_of_packs: f64) -> f64 {
+    let residual_stock = available_packs - requested_number_of_packs;
+    if residual_stock.abs() < LAST_PACK_THRESHOLD {
+        // there is nearly enough, or almost everything requested, we'll take everything...
+        return requested_number_of_packs + residual_stock;
+    }
+    requested_number_of_packs
+}
+
 pub fn check_batch_exists(
     store_id: &str,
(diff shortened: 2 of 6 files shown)


client/packages/invoices/src/Prescriptions/LineEditView/PrescriptionLineEditForm.tsx
```diff
@@ -254,9 +254,11 @@ const dosesSummary = (
     0
   );
 
+  const roundedDoses = NumUtils.round(totalDoses);
+
   const unitWord = t('label.doses-plural', {
-    count: NumUtils.round(totalDoses),
+    count: roundedDoses,
```
jmbrunskill, line 260: Feels like there's probably other places similar to this, outbound shipments for example? Other rounding needed?

server/service/src/invoice_line/stock_out_line/validate.rs
```diff
@@ -3,6 +3,17 @@ use repository::{
     StockLineFilter, StockLineRepository, StorageConnection,
 };
 
+const LAST_PACK_THRESHOLD: f64 = 0.001;
+
+pub fn adjust_for_residual_packs(available_packs: f64, requested_number_of_packs: f64) -> f64 {
+    let residual_stock = available_packs - requested_number_of_packs;
+    if residual_stock.abs() < LAST_PACK_THRESHOLD {
+        // there is nearly enough, or almost everything requested, we'll take everything...
+        return requested_number_of_packs + residual_stock;
```
andreievg, line 12: I think it reads better to use available_packs here

andreievg (approved):
This looks great, thanks, however I am not sure about all the use cases, would need some time to look at it in depth, happy for RC build, but before release want to consider:
* Auto-allocation logic, would we always skip a line that's just below threshold ? 
* As you've mentioned, other areas, like outbound shipments
* Would decimals like this compound at some point to be more then threshold ¯\_(ツ)_/¯

PR #8050, opened 2025-06-06
#6167 Reorder cold storage types to appear in desc order

### LINKED ISSUE #6167: Reorder item variant list
Request from @adamdewey 

Please order this list in descending 'from' temperature order

Image

noel-yeldos:
Fixes #6167

# 👩🏻‍💻 What does this PR do?
Reorder cold storage types to appear in desc order (based on minimum temperature)

Screenshot 2025-06-06 at 16 41 16

The issue had only specified item variant cold storage type selector but I've made api return the list in descending so it will appear that way in all places that cold storage type selector is used. Seems like that's ok to me but happy to be told otherwise. 

# 🧪 Testing

Ensure you have cold storage location types set up in OG - Item -> show location types:
  - your store will need to be on OG server to be able to edit location types
  - I just moved a random store to OG server -> logged in as that store to og -> added the types -> and then moved store back to OMS server
- [ ] Navigate to Catalogue -> Item
- [ ] Go to item variant tab and create a variant
- [ ] Open cold storage type dropdown and see that types are ordered in descending order (i.e higher temperatures will appear first as per screenshot above)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/general/src/queries/cold_storage_type.rs
+++ b/server/graphql/general/src/queries/cold_storage_type.rs
@@ -38,7 +38,6 @@ pub struct ColdStorageTypeSortInput {
 pub struct ColdStorageTypeFilterInput {
     pub id: Option<EqualFilterStringInput>,
     pub name: Option<EqualFilterStringInput>,
-    pub min_temperature: Option<EqualFilterBigFloatingNumberInput>,
 }
 
 #[derive(SimpleObject)]
@@ -85,16 +84,11 @@ pub fn cold_storage_types(
 
 impl ColdStorageTypeFilterInput {
     pub fn to_domain(self) -> ColdStorageTypeFilter {
-        let ColdStorageTypeFilterInput {
-            id,
-            name,
-            min_temperature,
-        } = self;
+        let ColdStorageTypeFilterInput { id, name } = self;
 
         ColdStorageTypeFilter {
             id: id.map(EqualFilter::from),
             name: name.map(EqualFilter::from),
-            min_temperature: min_temperature.map(EqualFilter::from),
         }
     }
 }
--- a/server/graphql/general/src/queries/cold_storage_type.rs
+++ b/server/graphql/general/src/queries/cold_storage_type.rs
@@ -1,6 +1,6 @@
 use async_graphql::*;
 use graphql_core::{
-    generic_filters::{EqualFilterBigFloatingNumberInput, EqualFilterStringInput},
+    generic_filters::EqualFilterStringInput,
     pagination::PaginationInput,
     standard_graphql_error::{validate_auth, StandardGraphqlError},
     ContextExt,
--- a/server/graphql/general/src/queries/cold_storage_type.rs
+++ b/server/graphql/general/src/queries/cold_storage_type.rs
@@ -1,6 +1,6 @@
 use async_graphql::*;
 use graphql_core::{
-    generic_filters::EqualFilterStringInput,
+    generic_filters::{EqualFilterBigFloatingNumberInput, EqualFilterStringInput},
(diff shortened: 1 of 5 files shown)


aimee-mcneil-melville (changes requested):
Thanks Noel, those testing steps were super helpful for getting this set up! I do also like your approach and think the consistency will be great 🚀 

Unfortunately my locations have not sorted in order - they have appeared in the order of the random ID, which in my case looks like this: 
Screenshot 2025-06-09 at 3 40 49 PM

It looks as though minTemperature will need to be added into the Sorting options [here](https://github.com/msupply-foundation/open-msupply/blob/a4c8c241851214b0dc63cc09f62a9c5e0dd27306/server/graphql/general/src/queries/cold_storage_type.rs#L22) so this can be passed through as the key instead of id

server/repository/src/db_diesel/cold_storage_type.rs
```diff
@@ -20,12 +20,14 @@ pub struct ColdStorageType {
 pub struct ColdStorageTypeFilter {
     pub id: Option<EqualFilter<String>>,
     pub name: Option<EqualFilter<String>>,
+    pub min_temperature: Option<EqualFilter<f64>>,
```
aimee-mcneil-melville, line 23: And min_temperature also not needed in the filter here

server/graphql/general/src/queries/cold_storage_type.rs
```diff
@@ -37,6 +38,7 @@ pub struct ColdStorageTypeSortInput {
 pub struct ColdStorageTypeFilterInput {
     pub id: Option<EqualFilterStringInput>,
     pub name: Option<EqualFilterStringInput>,
+    pub min_temperature: Option<EqualFilterBigFloatingNumberInput>,
```
aimee-mcneil-melville, line 41: min_temperature is not needed in the filter here

aimee-mcneil-melville (changes requested):
Awesome Noel, this works really well! 🎉 

I noticed that min_temperature has also been included in the filter - can you remove the struct/implementation in the files where I've commented please? 

Filtering operates independently from the sorting, so though the Sort and Filter structs often match, in this case min_temperature is not needed in the filter (the way the autocomplete is set up for the component does allow searching by temperature without this filter set up)

server/graphql/general/src/queries/cold_storage_type.rs
noel-yeldos: Updated here 3e20c1efc54c1365fcce80909d79e2acd5806192

server/repository/src/db_diesel/cold_storage_type.rs
noel-yeldos: Updated here 3e20c1efc54c1365fcce80909d79e2acd5806192

noel-yeldos:
> Awesome Noel, this works really well! 🎉
> 
> I noticed that min_temperature has also been included in the filter - can you remove the struct/implementation in the files where I've commented please?
> 
> Filtering operates independently from the sorting, so though the Sort and Filter structs often match, in this case min_temperature is not needed in the filter (the way the autocomplete is set up for the component does allow searching by temperature without this filter set up)

Ah thanks @aimee-mcneil-melville! I had misunderstood the code and thought that we needed filtering for sorting 😅 Have removed now.

aimee-mcneil-melville (approved):
Awesome, good to go! Thanks for the quick changes :)

PR #6844, opened 2025-03-05
6712 safely deserialise invoice_line.item_variant_id

Chris-Petty:
Fixes #6712 

# 👩🏻‍💻 What does this PR do?

OG/Mobile remote sites munge null fields to empty values e.g. strings to `""`. This breaks inbound shipment invoice line FK constraints in OMS when a OG supplier remote site makes a CI to a OMS site. Unless the FK field has     `#[serde(deserialize_with = "empty_str_as_option_string")]`!

## 💌 Any notes for the reviewer?

Suppliers on OG central server will not exhibit this bug, it must be a remote site OG supplier (or old mobile)

# 🧪 Testing

- [ ] OG or mobile supplier, OMS customer on remote (central OMS probably OK for this test)
- [ ] Make a CI to OMS customer.
- [ ] Sync both 
- [ ] Inbound shipment should have all the lines from the CI
- [ ] For completeness repeat with a CI generated from a customer requisition that was from OMS internal order (as per the original issue)

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #8150, opened 2025-06-12
Handle empty om_variant_id on stock_line

### LINKED ISSUE #8142: Stock_lines are not imported for some store during omSupply integration
## What went wrong? 😲

It seems that when integrating certain stores, there are ForeignKeyViolation constraint for the item lines which prevent them to be imported into OMS. It doesn't seem to happen to all the store. Saying that after multiple test the only store where I can see my stock is a site I have migrated before with 2.7.0. If I delete my database and re-initalise again the stock is there but for all other sites I have tried (4 sites) have no stock. I have tried multiple build going back to 2.7.0RC1 and I am encountering the same problem. The only constant in all my test is that oms Central is on 2.8.0RC9 and my mSupply central is on 8.2.12

I think it might be due to this new field in mSupply 

Image

It was probably working for my original store because it was already migrated when this new field as added. And maybe it is why it is still working when I re-initialise this store but not the other ones.

https://drive.google.com/file/d/1lJtcpepu5EIlKjI6pmevh0Rl0HaTw6ea/view?usp=sharing

Here is what I can see in the log. It seems to happen for all stock lines
Image

Image

@AnushaUp had a similar problem last night when testing but could not replicate. @marthakeezy had also a similar problem on Monday.

Here is the log file: https://drive.google.com/file/d/12PSVcrDFsnZQGTKphijPxi_kvc5f1GKp/view?usp=sharing 

the oms datafile: https://drive.google.com/file/d/19kXNscSToAQ8cJnTMrjBtA3hF5uhUz3Z/view?usp=sharing

the OG central server datafile: https://drive.google.com/file/d/1JTg7xUgAzI5x5vn9wOrfe9ysClTCN6tn/view?usp=sharing

Oms central server datafile: https://drive.google.com/file/d/15d2xHe0k5fVd9NdJ1ETjOKvOj6VjWNbE/view?usp=sharing

## Expected behaviour 🤔
Stock line are integrated into oms
## How to Reproduce 🔨

Use the OG datafile provided. Message me for the store to pick
1. Initialise your oms remote with that store
2. Login to omSupply
3. Navigate to Inventory > View stock

## Your environment 🌱

- Open mSupply Version: v2.8.0 or possibly 2.7.6
- Legacy mSupply Central Server Version: 8.2.12

- Platform: Android, Desktop

- Database type: SQLlite

jmbrunskill:
Fixes #8142

# 👩🏻‍💻 What does this PR do?

Handles om_variant_id coming though as `"om_item_variant_id": "",`  when translating `item_line`.

Looks like we had this right at one point in history...

image

We also failed to have an sync test case for this :(

I think what happened was that this field has only just made it into a deployed mSupply even though it's been in OMS for a while, still surprised it wasn't picked up on any dev testing.

Was a bit hard to setup and test, but once I did was obvious fix.
We need to make sure we're diligent to check for  `#[serde(deserialize_with = "empty_str_as_option_string")] ` on any optional fields from Legacy. 
`OMS_Fields` is a bit of an exception as being an option it actually comes through as `null`

Will merge to 2.7.7 + cherry pick for 2.8.1

# 🧪 Testing

- [ ] Use up to date mSupply `main` or  `8.2.12`
- [ ] Try to migrate an existing mSupply to an open-mSupply one
- [ ] See stock doesn't show up...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/test/test_data/stock_line.rs
+++ b/server/service/src/sync/test/test_data/stock_line.rs
@@ -140,7 +140,7 @@ const ITEM_LINE_2: (&str, &str) = (
       "user_8_ID": "",
       "volume_per_pack": 0,
       "vvm_status": "",
-      "weight_per_pack": 0
+      "weight_per_pack": 0,
       "om_item_variant_id": ""
   }"#,
 );


jmbrunskill:
> This [commit](https://github.com/msupply-foundation/open-msupply/pull/5236/commits/d00efd0856304f24bcb136105b01c1d08dee938f) also has it for invoice lines

Fixed here...
https://github.com/msupply-foundation/open-msupply/pull/6844

server/service/src/sync/test/test_data/stock_line.rs
```diff
@@ -141,6 +141,7 @@ const ITEM_LINE_2: (&str, &str) = (
       "volume_per_pack": 0,
       "vvm_status": "",
       "weight_per_pack": 0
```
andreievg, line 143: Comma

jmbrunskill:
@andreievg Will fix OMS fields on 2.8 branch, it's not in 2.7.7

PR #8185, opened 2025-06-15
#8085 Fix error when going back after changing customer

### LINKED ISSUE #8085: Infinite loading when clicking back arrow from outbound shipment line detail modal after changing the outbound shipment customer
## What went wrong? 😲

See #8059 for context. Opening a outbound shipment line after changing the customer of an outbound shipment is working. However Clicking on the back arrow while the line is in detail view results in a infinite loading situation. See video. You can see the navigation bar show "Outbound Shipments/" instead of "Outbound Shipments" so it has probably something to do with the problem.

https://drive.google.com/file/d/11r_Rgh5xxm0ziaMBe6bqRVwIS9OlqOyW/view?usp=sharing

## Expected behaviour 🤔
Normally if you click the back button for a line detail when you didn't change the customer before, you are brought back to the Outbound shipments list window. It should have the same behaviour.

## How to Reproduce 🔨

See the steps in #8059 
1. When you have open your line after changing the customer name of the outbound shipment 
2. Click on the back button of your tablet 
3. See error as per video

## Your environment 🌱

- Open mSupply Version: v2.8.0RC8 
- Legacy mSupply Central Server Version: v8.2.11

- Platform: Android

- Database type: SQLite

CarlosNZ:
Fix #8085

# 👩🏻‍💻 What does this PR do?

The error was caused by the "back" button attempting to go to the old invoice URL, which no longer exists. This fix prevents that by "replacing" the url in the history when navigating after changing customer. However, if you go "back" from the line edit modal now, you'll be taken to the list view, which is not ideal, but it's consistent with how it works in a normal detail view (when not changing customer).

It doesn't really make sense to go "back" from a modal like this, as the URL hasn't actually changed. I think, ultimately, it would be good to add a url parameter to the url when opening modals (e.g. `?editItem=<id>`), so that you *can* go "back", but that's a much larger change and needs further discussion/planning.

The behaviour now should be exactly the same as it is when you haven't changed the customer at all.

# 🧪 Testing

- [ ] In an Outbound shipment, change the customer, noting the re-direct to a new invoice URL
- [ ] Open a line
- [ ] Click "Back" -- you should be taken to the Outbound shipments list
- [ ] Confirm this behaviour is the same when *not* changing the customer 

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
@@ -62,7 +62,8 @@ export const Toolbar: FC = () => {
                         RouteBuilder.create(AppRoute.Distribution)
                           .addPart(AppRoute.OutboundShipment)
                           .addPart(newId)
-                          .build()
+                          .build(),
+                        { replace: true }
                       );
                     }}
                   />


jmbrunskill (approved):
Looks good, agree it shouldn't be possible to go back to a URL that no longer exists.

PR #8187, opened 2025-06-15
Dockerise

### LINKED ISSUE #7297: Toggle omSupply central via env variables or config

In the work I was doing with demo with pre configured data file, it would be good to start omSupply as central server, for extra configurations that are only available in omSupply central to appear, without connecting to mSupply

For DEV mode or special DEMO build (behind feature) it should be possible to set omSupply as 'central' server, via config (that can also be overwritten with env variables)

A dummy api that omSupply central can query that mimics 4d omSupply site endpoint, to tell the site that it's central (a bit too much setup)

andreievg:
(third time lucky)

Fixes #7297 

Majority of the changes is from update export.json to load correct data, as it's needed for docker image

# 👩🏻‍💻 What does this PR do?

I think to understand the changes it's best to look through the readme in [docker hub](https://hub.docker.com/r/msupplyfoundation/omsupply)

And then README.md file in this PR

On top of change needed to Dockerise omSupply, a tag action to do this automatically and push this to dockerhub, I also added some changes that are require for automating plugin testing:

* Add a way to override is central server with config
* Added changes from develop to allow generating typescript types for backend plugins (modified this to run without needing to test)

I've been using a fork to build and push images to dockerhub.

Added docker credentials to github secrets

There is also an action to trigger plugin test with latest docker build, this is in trigger-plugin-test workflow, can be triggered manually or via dockerise workflow, see https://github.com/msupply-foundation/sao-tome-plugins/blob/main/.github/workflows/test.yml (currently targeting 'Tests' branch). dockerise will call triger-plugin-tests which in turn will call test workflow in the repositories listed in triger-plugin-tests. I am way behind general plugin docs, will add one about these workflows in the plugin docs, sorry.

Need to add ORG_WORKFLOW_TOKEN, i used my personal one for this in the fork

# 🧪 Testing

Try things in the dockerhub readme.

If you want to test dockerisation locally, may want to run the build script for dev-client and server, when on break : ), take time and a bit of resources.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/data/reference1/export.json
andreievg, line 1: Most changes in this PR are in this file, need strategy to handle this for future

server/cli/src/cli.rs
```diff
@@ -209,11 +212,18 @@ enum Action {
         #[clap(short, long, action = ArgAction::SetTrue, conflicts_with="enable")]
         disable: bool,
     },
-    /// Generate TypeScript Types for backend plugins and format with Prettier
-    GenerateTypeScriptTypes,
-    /// Run load test
-    #[cfg(feature = "integration_test")]
-    LoadTest(LoadTest),
+    GeneratePluginTypescriptTypes {
```
jmbrunskill, line 215: This rename makes good sense, however could it break any existing workflows or docs that need updating?

jmbrunskill (approved):
Looks fine to me. I haven't actually run and tested it out, but I can't see if breaking any thing, and it's a good starting point to iterate from if we need to make improvements.

.github/workflows/trigger-plugin-tests.yaml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  trigger-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger tests in plugin repositories
+        run: |
+          # List of repositories to trigger tests in
+          REPOS=(
+            "msupply-foundation/sao-tome-plugins"
+            # Add more repositories as needed
```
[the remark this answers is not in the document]
andreievg: Yep this will follow : )

server/cli/src/cli.rs
andreievg: I think it would be ok, i couldn't find any docs that mention this, it's not mentioned here anyways: https://github.com/andreievg/open-msupply-actions-test/blob/2bebddeaa46ed7baac1c6556435507bd018941d3/server/README.md#L332-L351

I have started using this now in the plugin: https://github.com/msupply-foundation/sao-tome-plugins/blob/43236f3a9401e4592aedfadb6e2649942abec89d/test/src/test_lib/build_plugin_bundle.rs#L52-L63 so hope it's ok to keep.

I am glad you made a comment here since the line above removes load test as a variant, and compiler doesn't complain because it's behind integration test, reinstated here
2bebddeaa46ed7baac1c6556435507bd018941d3

Issue #6898, opened 2025-03-11
Item ledger filters


mark-prins:
## Is your feature request related to a problem? Please describe 👀
In the requirements we have the line "Recreate the functionality of the  tab “Ledger” present in mSupply Desktop"

However I didn't explicitly state that we should include sort / filter / search in the requirements which were used to build

## Describe the solution you'd like 🎁

Add a filter for
* Type
* Status
* Date

roxy-dao:
Would need a bit of a rewrite on how we calculate the balance.

DhanyaHerath:
Duplicate of https://github.com/msupply-foundation/open-msupply/issues/4885 - we can keep this one as it has more detail

DhanyaHerath:
Triage: @mark-prins to confirm with @Noelkouassi - the filtering function should address most of the needs that the sort function does, if we add in the filter function, is it enough without adding in sort as well

mark-prins:
Have confirmed that we can add this in v2.8 as long as we communicate that to the client

jmbrunskill:
Added Ruru and Piwakawaka so we don't forget about this. Sounds like it's a bit of work @roxy-dao ?

DhanyaHerath:
@mark-prins can I check - does this still need to be in V2.8.0, or can we push this out to a later release where we collate more of the work from this particular UAT?

mark-prins:
yes, fine to push out into a release which has the UAT response work

alainsussol:
This will be built tested as part of the 2.9.0 CIV UAT build testing. See #8367

PR #8200, opened 2025-06-16
7905 refactor item ledger query

### LINKED ISSUE #7905: Refactor item ledger query
**Sub-issues for this are just related issues not children issues**

### Tasks 🗒 
- [x] Repository layer should only filter by item and start date
- [ ] Change return type of the item_ledger query to sort by `item_id`
   - [ ] Can be done as item_ledgers { item_id, ledgers } or in a HashMap, or up to you~ 🥇 
- [x] Filtering _type, status & date_ @ the service layer separate from repository filter to keep balance integrity
- [x] Remove sorting for item ledgers in the service layer, probably repository layer too expect for datetime
- [ ] datetime should be calculated to server time for compatibility with OG
- [x] Currently, we are forward calculating ledgers which can cause querying for ledgers to be really slow. Refactor the current item_ledger query to back calculate based balances.

Same datetime balance calculation is currently using UUID as per [this PR](https://github.com/msupply-foundation/open-msupply/pull/7911). Whist UUID7 is time-sortable since it's time-based, other systems we used might not use UUID7 which will break this (namely, OG). Instead, 
- [x] If there are two ledger entries with the same datetime (e.g. Finalised Inbound vs Finalised Outbound), we calculate the incoming stock first, then the outgoing. 
- [x] Picked Prescriptions vs Cancelled, calculate the cancelled prescription first then the adjustment back
- [x] For items with multiple stock lines from same invoice, we can keep that calculate the balance as a total like OMS does

### Acceptance Criteria ✅
- [ ] Ledger's balance is accurate
- [ ] Ledger performance should be an improvement compared to the old item ledger (benchmark?)
- [ ] Item ledger endpoint should allow the filtering by: item_id, start_datetime, type, status and/or date
- [ ] Item ledger should match up with what OG has 👀 

### Moneyworks Jobcode 🧰
OMS:CIDV

roxy-dao:
Fixes #7905

# 👩🏻‍💻 What does this PR do?
- Created a separate item ledger view that piggybacks off the stock movement view to calculate running balance with a window function... this reduces the code needed in rust for calculating balances for all ledgers between when filters are used... SQL does seem to be doing some kind of caching / optimising with the window function and have done some local benchmarking, but would be curious to see how this performs in a cloud server 👀 
- Doing it in the view also allows for each line to be calculated without having to add extra code in rust for this 👀 
- Have also added the type and status filters as requested in #6898, and removed sorting.
- The datetime for the ledgers are already being parsed to server time in the frontend...
- I've doubled checked with OG, seems generally correct... though I don't get how sorting works in OG cause once I sort date asc the balances go out of wack
- Also didn't change the item_ledger to return { item_id, ledgers }... thought nice to keep generic since we're already allowing item_id as a filter
- Also staggered balancing prescription by a few seconds to get the correct balance when prescriptions are cancelled

# 🧪 Testing

- [ ] Have an item with some transactions (having multiple stock lines in/out would be nice for testing too)
- [ ] Go to item -> ledger
- [ ] Check balance entries to ensure they're calculating properly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/item_ledger.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn query(
+        &self,
+        pagination: Pagination,
+        filter: Option<ItemLedgerFilter>,
+    ) -> Result<Vec<ItemLedgerRow>, RepositoryError> {
+        let mut query = create_filtered_query(filter);
+        query = query
+            .order(item_ledger::datetime.desc())
+            .then_order_by(item_ledger::id.desc())
+            .then_order_by(item_ledger::type_precedence.asc());
```
roxy-dao, line 115: Need sorting to match with view so that the balances to match with the correct entry 👀

server/repository/src/migrations/views/mod.rs
```diff
@@ -153,7 +154,12 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
         invoice.status AS invoice_status,
         invoice_line_stock_movement.total_before_tax AS total_before_tax,
         invoice_line_stock_movement.pack_size as pack_size,
-        invoice_line_stock_movement.number_of_packs as number_of_packs
+        invoice_line_stock_movement.number_of_packs as number_of_packs,
+        CASE
+          WHEN invoice.type IN ('INBOUND_SHIPMENT', 'CUSTOMER_RETURN', 'INVENTORY_ADDITION') THEN 1
+          WHEN invoice.type IN ('OUTBOUND_SHIPMENT', 'SUPPLIER_RETURN', 'PRESCRIPTION', 'INVENTORY_REDUCTION') THEN 2
+          ELSE 3
+        END AS type_precedence
```
roxy-dao, line 162: `type_precedence` was added to get stock_ins to calculate before stock_outs in case of somehow inbound and outbound ends up with the same datetime....

server/service/src/invoice/prescription/update/mod.rs
```diff
@@ -118,7 +118,11 @@ pub fn create_reverse_prescription(
     new_invoice.id = uuid();
     new_invoice.linked_invoice_id = Some(orig_invoice.id.clone());
     new_invoice.is_cancellation = true;
-    new_invoice.verified_datetime = Some(Utc::now().naive_utc());
+    new_invoice.created_datetime = orig_invoice.created_datetime + Duration::seconds(10);
+    new_invoice.picked_datetime = orig_invoice
+        .picked_datetime
+        .map(|dt| dt + Duration::seconds(10));
+    new_invoice.verified_datetime = Some(Utc::now().naive_utc() + Duration::seconds(10));
```
roxy-dao, line 125: Staggering times so that the adjustment stock is calculated after...

Though, I still find it confusing that the prescription that reduces stock is 'Cancelled' and the one that adjusts stock back is 'Verified'... Should be keeping the original as 'Verified' and then creating a 'Cancelled' one to adjust stock back....

server/service/src/invoice/prescription/update/mod.rs
```diff
@@ -143,8 +147,7 @@ pub fn create_reverse_prescription(
         if let Some(stock_line_id) = &line.invoice_line_row.stock_line_id {
             let stock_line_repo = StockLineRowRepository::new(connection);
             let mut stock_line = stock_line_repo
-                .find_one_by_id(stock_line_id)
-                .map_err(RepositoryError::from)?
```
roxy-dao, line 147: Redundant conversion..

Chris-Petty (approved):
Approving however I'd appreciate changing that `quantity` field in the view to any name other than that 😉 

Works good on my little DB (pg, m3 macpro in dev)

image

Also tested on about 200k trans_lines and just as speedy.

server/repository/src/db_diesel/item_ledger.rs
```diff
@@ -0,0 +1,149 @@
...
+use diesel::prelude::*;
+
+table! {
+    item_ledger (id) {
+        id -> Text,
+        name -> Text,
+        item_id -> Text,
+        store_id -> Text,
+        quantity -> Double,
+        datetime -> Timestamp,
```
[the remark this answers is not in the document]
roxy-dao: Yeah, I agree... had to go through some of the movement views to find out what it was

server/repository/src/migrations/views/mod.rs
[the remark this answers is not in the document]
roxy-dao: Okay, have just dupped it 👀 but will keep it same for now...

server/repository/src/migrations/views/mod.rs
```diff
@@ -165,6 +171,35 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
...
+      cost_price_per_pack,
+      sell_price_per_pack,
+      invoice_status,
+      total_before_tax,
+      pack_size,
+      number_of_packs,
+      type_precedence,
+      SUM(quantity) OVER (
+        PARTITION BY stock_movement.store_id, stock_movement.item_id
+        ORDER BY stock_movement.datetime, stock_movement.id, type_precedence
```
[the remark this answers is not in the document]
roxy-dao: stock_movement.id is the invoice_line id, which is also a timestamped queryable 👀

server/service/src/invoice/prescription/update/mod.rs
roxy-dao: Yup that's the suggestion... it's confusing for cancelled to be negative and verified to be positive 👀

PR #8251, opened 2025-06-19
8237 add oms fields deserialiser

### LINKED ISSUE #8237: Not handling empty string sync of oms_fields as mSupply object
## What went wrong? 😲

Some tables in OMS including invoice_line (transline), and program_requisition_settings use an object for sync. 

OMS assumes that these objects make assumptions that aren't necessarily correct:
- invoice_line doesn't necessarily handle object cases at all and is only tested with `null` or `""` values
- program_requisition_settings assumes an object is returned, but cannot handle a `""` value

There may be other tables which make similar assumptions about mSupply object sync types.

## Expected behaviour 🤔

custom oms fields from mSupply should be able to handle deserialisation of `null`, `""`, or `{}` values as None.
Should attempt to deserialise to the oms object if the value received from oms is an object - and then handle the nested deserialisation on a case by case basis.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #8237 

# 👩🏻‍💻 What does this PR do?

Adds new deserialisation to object fields from OG.

Captures possible case of objects being returned from OG as an empty string, null, or empty object without throwing error.

I was wondering when writing this if there are any cases where we deserialise an object into a OMS struct where the struct is NOT optional on the the legacy row? 

These cases might yet be missed if the type is not nullable object in mSupply, but there might be lingering empty strings in our sync record.

# 🧪 Testing

You can see how the sync might have been failing by changing sync values to other cases and see failure.

You can view how we are handling cases on branch `8237-without-new-deserialisation` which adds more comprehensive test cases but without the new deserialiser

- [ ] Change to branch `8237-without-new-deserialisation` and see tests fail
- [ ] Change to `8237-add-oms_fields-deserialiser` which includes deserialiser, and see tests now pass

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/test/test_data/program_requisition_settings.rs
```diff
@@ -119,6 +119,88 @@ const MASTER_LIST_WITH_PROGRAM_2: (&str, &str) = (
...
 
+// const MASTER_LIST_WITH_EMPTY_OBJECT_PROGRAM_SETTINGS: (&str, &str) = (
+//     "program_test_3",
+//     r#"{
+//     "ID": "program_test_3",
+//     "description": "Program Test 03",
+//     "date_created": "2017-08-17",
+//     "created_by_user_ID": "0763E2E3053D4C478E1E6B6B03FEC207",
+//     "note": "note 4",
+//     "gets_new_items": false,
```
fergie-nz, line 130: Added this test data in case it could be used to test deserialising of program_requisition_setting - but I think in this case it is OK as the deserialising handles having no program settings in the generate function, and throws a specific error.

server/service/src/sync/translations/special/oms_object_fields.rs
```diff
@@ -0,0 +1,90 @@
+use serde::{Deserialize, Serialize};
+
+use crate::sync::sync_serde::object_fields_as_option;
+
```
fergie-nz, line 4: This file is generic handling of all possible null cases including empty string by the new deserialiser

server/service/src/sync/translations/name.rs
```diff
@@ -136,6 +137,7 @@ pub struct LegacyNameRow {
     #[serde(deserialize_with = "zero_date_as_option")]
     #[serde(serialize_with = "date_option_to_isostring")]
     pub date_of_death: Option<NaiveDate>,
+    #[serde(deserialize_with = "object_fields_as_option")]
```
fergie-nz, line 140: Handles generic Value here also to turn `""` into `None`

jmbrunskill: Lets take this one out as don't want to risk breaking the SAGE integration (low chance but still...)

server/service/src/sync/test/test_data/program_requisition_settings.rs
jmbrunskill: I think we'll jsut take these out for now. It's only managed on central server so not likely to be messed up to an empty object...

PR #8268, opened 2025-06-22
#8243 Purchase order detail view -- sort and filter lines

### LINKED ISSUE #8243: Sorting/Filtering of Detail View lines
~~Front-end needs to use a separate query for PO lines, as fetching them as part of the main PO entity doesn't allow for sorting/filtering/pagination~~

It looks like other invoices (Outbounds/Inbounds/Prescriptions) all use front-end sorting/filtering of lines, so I guess we'll just go with that.

CarlosNZ:
Fixes #8243

# 👩🏻‍💻 What does this PR do?

Implements front-end sort/filter of PO lines so UI can Sort via clicking column header and filter using the "Filter Items" input in the Toolbar.

Put the logic in the `usePurchaseOrder` hook, in its own subhook within that. It just exposes the line data, and the value/setter for the "filter" string.

# 🧪 Testing

- [ ] Make a PO and add multiple lines (will have to manually edit the database now, no UI yet)
- [ ] Confirm that each column in the detail view can be sorted by clicking the header (and click again to swap order)
- [ ] Confirm text input into the "Filter items" input correctly filters by partial matching to any item "code" or "name"

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/purchase_orders/src/DetailView/Toolbar.tsx
```diff
@@ -70,26 +71,12 @@ export const Toolbar = ({ isDisabled }: ToolbarProps) => {
...
-            }}
-            debounceTime={0}
-          />
-        </Grid>
+        <SearchBar
+          placeholder={t('placeholder.filter-items')}
+          value={itemFilter ?? ''}
+          onChange={newValue => setItemFilter(newValue)}
+          debounceTime={0}
+        />
```
fergie-nz, line 79: I was wondering if it would be helpful to say that it is filtering by code or name?

But I can see that this is the standard string for this filter, so good as gold.

fergie-nz (approved):
LGTM thanks Carl

client/packages/purchase_orders/src/DetailView/Toolbar.tsx
CarlosNZ: Yeah, okay, I've added a comment in the actual filter implentation now, I don't think it matters here

1754798cb5

PR #8271, opened 2025-06-23
Default to allocating in packs if there's only one packsize

### LINKED ISSUE #8258: Outbound Shipment > Add item page: implement a preference for user to choose whether issuing is in units or in packs
Bug cause this feature is to replicate a behaviour that existed in previous versions of Open mSupply. 

## What went wrong? 😲

Some countries may decide to do most of their issuing in quantity of packs, and not in quantity of units. 
However, when adding an item to an Outbound Shipment, the default is to dispense in units, even when there is a single stock line with a pack size of >1

Meaning that user, for each item added to the O/S, will have to select to dispense in packs. In previous versions of OMS, system would automatically suggest to dispense in the pack size of the available stockline. 

## Expected behaviour 🤔

Dropdown should default to the available pack size, where there's only one. 

Another solution would be to have a preference to specify whether default should be units or packs. Similar to what is suggested in https://github.com/msupply-foundation/open-msupply/issues/7954 for ordering. 

## Your environment 🌱

- Open mSupply Version: 2.8.1

- Platform: browser (Chrome)

- Database type:

jmbrunskill:
Fixes #8258

# 👩🏻‍💻 What does this PR do?

When creating an outbound shipment, if there's only one existing packsize in stock, default to allocating in that packsize

Feels like this maybe should be done in Allocation Context somewhere, however this limits this just to the Outbound shipment page.

# 🧪 Testing

- [ ] Try allocating stock for an item with only one packsize in stock, the sysmte should default to packs of size X where X is the size of the existing stock lines
- [ ] If there's more than one packsize, it should default to units (or doses if vaccine item and `manageVaccinesInDoses` is selected
- [ ] Check I haven't broken anything else.

# 📃 Documentation

  1. Allocation docs
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
@@ -25,6 +25,7 @@ import {
   sumAvailableUnits,
   AllocateInSelector,
   useOutboundLineEditData,
+  AllocateInOption,
 } from '../../../StockOut';
 import { useOutbound } from '../../api';
 import { CurrencyRowFragment } from '@openmsupply-client/system';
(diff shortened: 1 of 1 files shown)


client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
```diff
@@ -62,6 +63,23 @@ export const Allocation = ({
...
 
+      const packsizes = [
+        ...new Set(data.draftLines.map(line => line.packSize)),
+      ];
+
+      // if there is only one packsize, we should auto-allocate in packs for that size
+      let allocateInPacksize: AllocateInOption | undefined = undefined;
+      if (packsizes.length === 1) {
+        const packSize = packsizes[0];
+        if (packSize) {
```
jmbrunskill, line 74: These multiple `if` statements are a bit messy but Typescript didn't seem to trust that packsize might not be defined. Pretty sure if can be cleaned up but also this works?

GeronimoJohn (approved):
LGTM! 🔥

PR #8272, opened 2025-06-23
8056 nightly automate builds gh action create tags

### LINKED ISSUE #8056: Nightly automated builds

Currently there are a number of steps we need to do to create a build for the QA team, or for PMs to review.
For example we need to update package.json, create a tag, then push it, then we need to create a github release, wait for the android and or jenkins builds to finish download the files then upload again to github.

This all takes effort and could be more automated

Every night there should be an automated build from both develop and (if active) any v2.X-RC branches.

The version number should be automatically generated based on todays date and RC or Dev branches

e.g. v2.8.0-dev-20250509 (develop branch) or v2.8.0-rc-20250509 (rc branch)

Both the android action and jenkins should push their completed files to a cloud storage location that is accessible by QA team. A Telegram bot will announce when a file is ready for a release and provide a link.

Current process but more automated e.g. create a github release.
However, this feels like it will clutter up github too much so we should just use it for

### Additional context 💌

Need to discuss best place to upload and navigate to the files. Potentially something like a public folder in backblaze and maintain a webpage which links to the files as I don't think there's a normal file browser view by default?

### Moneyworks Jobcode 🧰

OMS:JAN24
Team leads agreed this was worth the effort

GeronimoJohn:
Partially Fixes #8056

# 👩🏻‍💻 What does this PR do?

Wrote a Github Action that is triggered at 6:30 at night. The action generates two tags. One for develop and the other for the latest RC branch

The GH Action will only take effect if its on the default branch of a repo. Meaning while its a PR. It theoretically should never run. 

From Github docs: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule
```
This event will only trigger a workflow run if the workflow file is on the default branch.

Scheduled workflows will only run on the default branch.
```

# 🧪 Testing

- [ ] Created a dedicated repo to test out the GH action
- [ ] Navigate and access the repo https://github.com/GeronimoJohn/gh-action-tags
- [ ] View create-nightly-tags.sh file and uncomment schedule code
- [ ] Navigate to the Repo's actions tab
- [ ] Wait for GH action to run
- [ ] Tags should get created if doesn't exist yet

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/.github/workflows/create-nightly-tags.sh
+++ b/.github/workflows/create-nightly-tags.sh
@@ -1,16 +1,17 @@
 #!/bin/bash
 set -e
 
+# Tag format example: v1.2.3-rc-1025201530
 create_tag_for_branch() {
     local branch_name=$1
     
     git checkout "$branch_name"
     
-    PKG_VERSION=$(cat ./package.json | grep 'version":' | sed 's/.*"version": *"\([^"]*\)".*/\1/')
+    PKG_VERSION=$(cat ./package.json | grep 'version":' | sed 's/.*"version":[ \t]*"\([^"]*\)".*/\1/')
     COMMIT_DATE=$(git log -1 --format=%cd --date=format:%m%d%H%M)
     
     if [[ -n "$PKG_VERSION" && -n "$COMMIT_DATE" ]]; then
-        TAG_NAME="$PKG_VERSION-$COMMIT_DATE"
+        TAG_NAME="v$PKG_VERSION-$COMMIT_DATE" 
         echo "Creating tag for $branch_name: $TAG_NAME"
         git tag "$TAG_NAME"
         git push origin "$TAG_NAME"
--- a/.github/workflows/create-nightly-tags.sh
+++ b/.github/workflows/create-nightly-tags.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -e
+
+create_tag_for_branch() {
+    local branch_name=$1
+    
+    git checkout "$branch_name"
+    
+    PKG_VERSION=$(cat ./package.json | grep 'version":' | sed 's/.*"version": *"\([^"]*\)".*/\1/')
+    COMMIT_DATE=$(git log -1 --format=%cd --date=format:%m%d%H%M)
+    
+    if [[ -n "$PKG_VERSION" && -n "$COMMIT_DATE" ]]; then
+        TAG_NAME="$PKG_VERSION-$COMMIT_DATE"
+        echo "Creating tag for $branch_name: $TAG_NAME"
+        git tag "$TAG_NAME"
+        git push origin "$TAG_NAME"
+    else
+        echo "Failed to generate tag for $branch_name (PKG_VERSION: '$PKG_VERSION', COMMIT_DATE: '$COMMIT_DATE')"
+    fi
+}
+
+# Get all RC branches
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-automated-builds.yaml
```diff
@@ -0,0 +1,59 @@
+on:
+  schedule:
+    - cron: "0 0 * * *" # Runs daily at midnight UTC
```
GeronimoJohn, line 3: Just a starting point, can change time to whatever we think is best

.github/workflows/nightly-automated-builds.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  create-tags:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.PAT_WITH_WORKFLOW_SCOPE }}
```
GeronimoJohn, line 16: Needs to be created, the general github_token doesn't have the right permissions

.github/workflows/nightly-automated-builds.yaml
```diff
@@ -0,0 +1,26 @@
+name: Nightly Automated Builds
+on:
+  workflow_dispatch:
+  # schedule:
+  #   - cron: "0 6 * * *" #  Every day at 6AM UTC which is 6PM NZDT
```
GeronimoJohn, line 5: Still getting it to work

GeronimoJohn: Have tested on a separate repo and works: https://github.com/GeronimoJohn/gh-action-tags/actions/runs/15840970494

.github/workflows/nightly-automated-builds.yaml
GeronimoJohn: Have tested to just use the general github_token and works with it

.github/workflows/create-nightly-tags.sh
```diff
@@ -0,0 +1,29 @@
...
+    PKG_VERSION=$(cat ./package.json | grep 'version":' | sed 's/.*"version":[ \t]*"\([^"]*\)".*/\1/')
+    COMMIT_DATE=$(git log -1 --format=%cd --date=format:%m%d%H%M)
+    
+    if [[ -n "$PKG_VERSION" && -n "$COMMIT_DATE" ]]; then
+        TAG_NAME="v$PKG_VERSION-$COMMIT_DATE" 
+        echo "Creating tag for $branch_name: $TAG_NAME"
+        git tag "$TAG_NAME"
+        git push origin "$TAG_NAME"
+    else
+        echo "Failed to generate tag for $branch_name (PKG_VERSION: '$PKG_VERSION', COMMIT_DATE: '$COMMIT_DATE')"
```
jmbrunskill, line 19: This will fail if tag already exists, would be nice if we actually handled that a bit better, maybe can be part of the telegram notificaiton PR, as that will need to only send the message if there's a new tag version.

GeronimoJohn: Yup thought of that too, handling it in later pr

PR #8279, opened 2025-06-24
8278 Fix load test hang

### LINKED ISSUE #8278: Load test not ending due to stuck child processes
## What went wrong? 😲

For a 5hr (18000s) test, it ran faaaar longer due to stuck processes, which eventually release themselves about 12hrs after the test was started!

Image

Should investigate how it gets so stuck on sync as to push 100k records...

## Expected behaviour 🤔

The load test CLI should end after the specified duration (approximately at least)

## How to Reproduce 🔨

1. Ran the test with 20 sites on admittedly underpowered, not isolated test machines with a duration of 18000s.
2. It ran way past 18000s, 18/20 children had been correctly killed, the last 2 finally finished their last sync that duration >20000s. It was about 12hours after the test start, however.

You could reproduce this in dev by putting an infinite loop within the inner spawned processes that the LoadTest `run` function creates.

## Your environment 🌱

- Open mSupply Version: dev

Chris-Petty:
Fixes #8278 

# 👩🏻‍💻 What does this PR do?

- Use channels to return metrics throughout test
- Parent process kills children after duration of test if they haven't terminated themselves
- include the child site_id in kill log message

I copiloted 90% of this

# 🧪 Testing

- [ ] OMS and OG central
- [ ] a site called `test_site` with at least 1 store on it
- [ ] add an infinite loop to the code to ensure that the child processes are stuck
- [ ] `cargo build --features integration_test && ./target/debug/remote_server_cli load-test --msupply-central-url http://localhost:8888 --oms-central-url http://localhost:8000 -s 10 -d 10`

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/cli/src/load_test.rs
+++ b/server/cli/src/load_test.rs
@@ -337,8 +337,10 @@ impl LoadTest {
             all_metrics
         });
 
-        // Wait for either all tasks to complete or timeout
-        let timeout_duration = Duration::from_secs(duration + 30); // Extra 30 seconds for cleanup
+        // Wait for either all tasks to complete or timeout. We delay by a significant amount here as the child processes don't start their timers based
+        // on duration until after they've initialised, where this timer will start essentially immediately, before the children have initialised.
+        // The more sites/children spawned, the longer we should expect initialisation to take.
+        let timeout_duration = Duration::from_secs(duration + (60 * 2 * num_sites as u64));
 
         let handles_for_timeout = handles.iter().map(|h| h.abort_handle()).collect::<Vec<_>>();
 


server/cli/src/load_test.rs
```diff
@@ -327,26 +312,70 @@ impl LoadTest {
...
+                        println!("Task failed: {}", e);
+                    }
+                }
+            } => {
+                println!("All tasks completed normally");
             }
         }
 
+        // Collect all metrics that were sent via channels
+        let results = match tokio::time::timeout(Duration::from_secs(5), results_handle).await {
```
Chris-Petty, line 367: If I'd written this rather than claude 4.0 I'd probably not bothered with the timeout of 5s here, however it should be a bit more robust if for some bizarre reason the `results_handle` has not completed (which would only happen if the channel didn't close properly. But they said rust had _fearless_ concurrency!)

roxy-dao: No harm right~?

PR #8283, opened 2025-06-24
8211 Create Clinician From Within Search

### LINKED ISSUE #8211: Create new clinician from within the search component

In the context of prescription, the plus button more looks like "add" clinician to the prescription, rather than "create a new clinician because I'm not in the existing options":
Image

I want the create button to be within the select component. Ideally the last option in the list. 

I.e. third option here, should be `Create new clinician`:
Image

It should look like the create patient option, except the create option should always be visible, when scrolling or typing to start filtering (in patient, you have to type enough of the name that there are no matching patients before the create button appears)
Image

aimee-mcneil-melville:
Fixes #8211

# 👩🏻‍💻 What does this PR do?

Can create a clinician from within the clinicians search input
Is always visible - regardless of filtering

Screenshot 2025-06-25 at 9 47 31 AM
Screenshot 2025-06-25 at 9 47 41 AM

I first created this as a specific component, and then pulled out the logic to the common autocomplete - can see in commit history.

To add this functionality to another search component, the following will be needed:
- define the CreateOption object with an id, label, and _isAddOption flag
- define a handler for what needs to happen when this option is selected (open modal)
- pass both of these to the Autocomplete component
- conditions can be set eg createOption only to be passed through if there are no results from filtering
- popover open/closed state may need to be managed here as well

Changes to a common component - it should be set up so as to not interfere with any existing functionality

Both createOption and onCreateOptionClick need to be provided to the autocomplete in order for this to work properly. Omitting createOption - it doesn't render.
Omitting onCreateOptionClick - it selects the createOption as value.

Is there a way to ensure that both are provided? I considered adding the handler as a required key of the createOption object

# 🧪 Testing

Have some clinicians already visible

- [ ] Go to Prescriptions -> New Prescription -> click on the clinicians search area
- [ ] See list of existing clinicians, and a `Create Clinician` button at the bottom of the list
- [ ] Search in the clinicians, see this button always visible regardless of search results
- [ ] Click on `Create Clinician` -> See the side modal open where you can create a clinician
- [ ] Enter clinician details -> OK -> new clinician name will populate
- [ ] Repeat in Prescriptions -> open an existing prescription -> search in clinicians. Should be the same, except selecting `Create Clinician` will open a modal

As this is a change on a common component, check other autocompletes throughout the app and that they work as expected

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [x] Postgres
- [x] SQLite
- [x] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
+++ b/client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
@@ -1,4 +1,4 @@
-import React, { PropsWithChildren } from 'react';
+import React, { PropsWithChildren, useMemo } from 'react';
 import {
   Autocomplete as MuiAutocomplete,
   AutocompleteRenderInputParams,
@@ -9,6 +9,9 @@ import {
   PopperProps,
   SxProps,
   Theme,
+  Box,
+  Typography,
+  FilterOptionsState,
 } from '@mui/material';
 import {
   AutocompleteOption,
@@ -19,6 +22,13 @@ import { BasicTextInput, BasicTextInputProps } from '../TextInput';
 import { StyledPopper } from './components';
 import { useOpenStateWithKeyboard } from './utils';
 import { useTranslation } from '@common/intl';
+import { PlusCircleIcon } from '@common/icons';
+
+export interface AddOption {
+  label: string;
+  id: string;
+  _isAddOption: boolean;
+}
 
 export interface AutocompleteProps<T>
   extends Omit<
@@ -52,6 +62,9 @@ export interface AutocompleteProps<T>
   inputProps?: BasicTextInputProps;
   required?: boolean;
   textSx?: SxProps<Theme>;
+  /** Create Option requires both createOption and onCreateOptionClick */
+  createOption?: AddOption;
+  onCreateOptionClick?: (option: AddOption) => void;
 }
 
 export function Autocomplete<T>({
@@ -80,11 +93,32 @@ export function Autocomplete<T>({
   inputProps,
   required,
   textSx,
+  createOption,
+  onCreateOptionClick,
   ...restOfAutocompleteProps
 }: PropsWithChildren<AutocompleteProps<T>>): JSX.Element {
   const t = useTranslation();
-  const filter = filterOptions ?? createFilterOptions(filterOptionConfig);
   const openOverrides = useOpenStateWithKeyboard(restOfAutocompleteProps);
+
+  const isCreateOption = (option: unknown): option is AddOption => {
+    return (
+      typeof option === 'object' &&
+      option !== null &&
+      '_isAddOption' in option &&
+      (option as any)._isAddOption === true
+    );
+  };
+
+  const filter = useMemo(() => {
+    const filterType = filterOptions ?? createFilterOptions(filterOptionConfig);
+    return (options: T[], state: FilterOptionsState<T>) => {
+      const filtered = filterType(options, state);
+      const addOptions = options.filter(
+        option => isCreateOption(option) && !filtered.includes(option)
+      );
+      return [...filtered, ...addOptions];
+    };
+  }, [filterOptions, filterOptionConfig]);
 
   const defaultRenderInput = (props: AutocompleteRenderInputParams) => (
     <BasicTextInput
@@ -125,6 +159,55 @@ export function Autocomplete<T>({
   );
   const popper = popperMinWidth ? CustomPopper : StyledPopper;
 
+  const createOptionRenderer = (createOption: AddOption) => (
+    <Box
+      display="flex"
+      justifyContent="space-between"
+      alignItems="center"
+      gap={1}
+      height={25}
+      width="100%"
+    >
+      <Typography
+        overflow="hidden"
+        fontWeight="bold"
+        textOverflow="ellipsis"
+        sx={{
+          whiteSpace: 'nowrap',
+        }}
+      >
+        {createOption.label}
+      </Typography>
+      <PlusCircleIcon color="secondary" />
+    </Box>
+  );
+
+  const optionsWithAdd = createOption
+    ? [...options, createOption as T]
+    : options;
+
+  const customRenderOption = (props: any, option: any, state: any) => {
+    if (isCreateOption(option)) {
+      return (
+        <li {...props} key={option.id}>
+          {createOptionRenderer(option as AddOption)}
+        </li>
+      );
+    }
+
+    if (renderOption) {
+      return renderOption(props, option, state);
+    }
+
+    return (
+      <li {...props} key={option.id}>
+        {(getOptionLabel || defaultGetOptionLabel)(option)}
+      </li>
+    );
+  };
+
+  const shouldUseCustomRenderOption = createOption || renderOption;
+
   return (
     <MuiAutocomplete
       {...restOfAutocompleteProps}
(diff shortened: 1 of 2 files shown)


client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -125,14 +159,70 @@ export function Autocomplete<T>({
...
+    if (renderOption) {
+      return renderOption(props, option, state);
+    }
+
+    return (
+      <li {...props} key={option.id}>
+        {(getOptionLabel || defaultGetOptionLabel)(option)}
+      </li>
+    );
+  };
```
Chris-Petty, line 207: Can have types:

```suggestion
  const customRenderOption = (
    props: React.HTMLAttributes<HTMLLIElement>,
    option: T,
    state: AutocompleteRenderOptionState
  ) => {
    if (isCreateOption(option)) {
      return (
        <li {...props} key={option.id}>
          {createOptionRenderer(option)}
        </li>
      );
    }

    if (renderOption) {
      return renderOption(props, option, state);
    }

    return (
      <li {...props} key={option.id}>
        {(getOptionLabel || defaultGetOptionLabel)(option)}
      </li>
    );
  };
```

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -125,14 +159,70 @@ export function Autocomplete<T>({
...
+        </li>
+      );
+    }
+
+    if (renderOption) {
+      return renderOption(props, option, state);
+    }
+
+    return (
+      <li {...props} key={option.id}>
```
Chris-Petty, line 203: Idk what MUI's internal rendering for this list is, however it doesn't seem to need a key prop for list elements (normally react will complain in console about it, but it's not). There is no guarantee `option.id` exists even if you did type this with `AutocompleteOption` anyway.

client/packages/system/src/Clinician/ClinicianSearchInput.tsx
```diff
@@ -100,6 +112,9 @@ export const ClinicianSearchInput = ({
   return (
     <Box width={`${width}px`} display={'flex'} alignItems="center">
       <Autocomplete
+        open={open}
+        onOpen={() => setPopoverOpen(true)}
```
Chris-Petty, line 116: I don't follow why this is required, it seems to work fine in browser without this. Is it an android thing? Otherwise `Autocomplete` already seems to manage this state internally (as it did before this PR).

Chris-Petty (changes requested):
Testing in browser an works well! However got some improvements suggested

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -80,12 +93,33 @@ export function Autocomplete<T>({
...
 
+  const isCreateOption = (option: unknown): option is AddOption => {
+    return (
+      typeof option === 'object' &&
+      option !== null &&
+      '_isAddOption' in option &&
+      (option as any)._isAddOption === true
+    );
+  };
+
```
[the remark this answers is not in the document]
aimee-mcneil-melville: I like this! I did look at using the createOption as `AutocompleteOption`, but had issues with typing - though with this change and removing the key, the createOption only needs to be a string for the label. Much simpler!

If we do need to extend this later (to allow a `remove` action for example) we would need an object to pass through an icon, or if as above we pass the `onCreateOptionClick` in the object..

client/packages/system/src/Clinician/ClinicianSearchInput.tsx
aimee-mcneil-melville: Strange, I can remove this now with no issues 👀 
I added it as the popover was staying open & above the slider modal here: 
Screenshot 2025-06-25 at 5 03 54 PM

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
aimee-mcneil-melville: Have removed the key :)

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -80,12 +93,33 @@ export function Autocomplete<T>({
...
+  const filter = useMemo(() => {
+    const filterType = filterOptions ?? createFilterOptions(filterOptionConfig);
+    return (options: T[], state: FilterOptionsState<T>) => {
+      const filtered = filterType(options, state);
+      const addOptions = options.filter(
+        option => isCreateOption(option) && !filtered.includes(option)
+      );
+      return [...filtered, ...addOptions];
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Ah this makes sense! Clinicians is likely small but others in the future may be quite a bit bigger so good to get this in now

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -93,32 +92,29 @@ export function Autocomplete<T>({
...
-      '_isAddOption' in option &&
-      (option as any)._isAddOption === true
-    );
+  const isClickableOption = (
+    option: unknown
+  ): option is ClickableOptionConfig => {
+    return option === clickableOption;
   };
 
-  const filter = useMemo(() => {
```
Chris-Petty: Heh I was skeptical the useMemo would really have any benefit here so happy to see it go

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -93,32 +92,29 @@ export function Autocomplete<T>({
...
-  }, [filterOptions, filterOptionConfig]);
+  const filterType = filterOptions ?? createFilterOptions(filterOptionConfig);
+  const filter = (options: T[], state: FilterOptionsState<T>) => {
+    const filtered = filterType(options, state);
+    if (
+      clickableOption &&
+      !filtered.some(option => isClickableOption(option))
+    ) {
+      return [...filtered, clickableOption as T];
+    }
```
Chris-Petty, line 115: better than what I had suggested, thanks!

client/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -238,8 +232,8 @@ export function Autocomplete<T>({
         shouldUseCustomRenderOption ? customRenderOption : undefined
       }
       onChange={(_event, option, reason, details) => {
-        if (isCreateOption(option) && onCreateOptionClick) {
-          onCreateOptionClick(option as AddOption);
+        if (isClickableOption(option) && clickableOption?.onClick) {
+          clickableOption?.onClick();
```
Chris-Petty, line 236: silly nit but typescript should be happy to drop the optional chain cause the if condition guarantees it is defined hehe

```suggestion
          clickableOption.onClick();
```

But like... same same lol

Chris-Petty (approved):
Sweeeeeeet thanks @aimee-mcneil-melville !

Issue #7304, opened 2025-04-13
DHIS2 Vaccine Card -> Aggregate


andreievg:
Currently two countries are interested in integrating vaccine card data with DHIS2. 

This issue is for the simple 'aggregate' data element type and requirements on omSupply end

## Is your feature request related to a problem? Please describe 👀

Data needs to be available in mSupply for this integration, but vaccine related data exists on omSupply end only. 

~~There is a number of 'links' that needs to be created between data in mSupply in DHIS2, these links need to be maintained if configurations change.~~ No linking in open mSupply

The main integration will be done by mSupply

## Describe the solution you'd like 🎁

Sync vaccine related data to mSupply, related issue here -> @JonnaQuismundo can you add link to mSupply issue please (as per our conversation last week)
https://github.com/msupply-foundation/msupply/issues/16422

~~Expose a new field for plugin in below view, make sure it's central only plugin and sync plugin data to mSupply (the plugin will allow user to enter DHIS2 data elements related to this vaccine course). The main reason for this linking is vaccine course spans multiple items, and here would be the most direct way for this linking to be configured.~~

### Describe alternatives you've considered 💭

Use prescriptions to deduce vaccination statistics.

~~Add configuration file in mSupply~~ this is how it will work

~~Above would not be as accurate or as direct of a data source and configurations file might be hard to maintain and create a lot of overhead for support and/or Jonna.~~

### Additional context 💌

~~The 'linking' change needs to be available on omSupply central only.~~

The sync change needs to be available on both omSupply central and omSupply remote, unless we add a way to proxy remote omSupply data through omSupply to mSupply (technically possible, and would be part of v7 at some point).

Doing change in omSupply means it's less urgent, since it's easier to upgrade omSupply central then remote site.

There is a **strong leaning towards** doing this on omSupply central

jmbrunskill:
Needed for July can go into 2.8.1?

JonnaQuismundo:
hi, new tables and fields are now created in og V8-02-06 but we haven't updated the syncV5 max version.

jmbrunskill:
Changed to 2.9.0 as should be ready mid July, however should be sone ASAP so we can start testing as early as possible.

Chris-Petty:
Sounds sick is a chunk of current thinking for v7 sync on the OMS central side 👍

PR #8284, opened 2025-06-25
8209 sync vaccination course

### LINKED ISSUE #8209: Sync Vaccination table to Legacy mSupply via OMS Central
### Possible Solution

2 Translators for the table

1. For OMS -> OMS
2. For OMS Central to Legacy

2 should only on OMS Central (normally this would not include records created from sync, how do we deal with that?)

jmbrunskill:
Fixes #8209 and completes #7304

# 👩🏻‍💻 What does this PR do?

Creates a new approach  for the central server to sync records to Legacy.
Instead of limiting to only sending records from active stores, it sends all records that haven't come from the mSupply Central Server. This means that it should be setup for sync v7 (when all oms remote sites only sync to oms central)

NOTE: Requires mSupply with this PR included (e.g. latest master branch) https://github.com/msupply-foundation/msupply/pull/16772/files

I'm fairly confident about the `vaccination` being sync'd properly not but not sure if there's some edge case with the central OMS sync that will go wrong, e.g. a record not being synced that should have, or an infinite sync loop.
Pretty sure we won't get infinite sync but could potentially miss records?

Not 100% if we should merget this or just do a build for testing at this point.
Not sure if we'll have the mSupply patch version out in time for 2.9, is it save to require a newer sync version?

# 🧪 Testing

- [ ] Upgrade mSupply to latest master (or new enough version)
- [ ] Create a remotes site eg. APK Perhaps? and central open mSupply server
- [ ] Setup vaccination card (or similar)
- [ ] Check that records for vaccinations sync from remote sites and from oms central to the Legacy mSupply om_vaccination table
- [ ] Check that no other weird stuff happens with sync e.g. Records that don't appear in mSupply but should have...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/v2_09_00/resync_existing_vaccination_records.rs
```diff
@@ -0,0 +1,21 @@
...
+    fn identifier(&self) -> &'static str {
+        "resync_existing_vaccination_records"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                INSERT INTO changelog (table_name, record_id, row_action, name_link_id, store_id, is_sync_update, source_site_id)
+                SELECT table_name, record_id, row_action, name_link_id, store_id, is_sync_update, source_site_id FROM changelog WHERE table_name = 'vaccination' and source_site_id IS NOT NULL;
```
jmbrunskill, line 15: Copying change log gets the correct site id, rather than the approach for vaccine_course like this...
```
                INSERT INTO changelog (table_name, record_id, row_action)
                SELECT 'vaccine_course', id, 'UPSERT' FROM vaccine_course; 
```
I think these are still fine, but just the remote records need to include the right source site...

server/service/src/sync/central_data_synchroniser.rs
```diff
@@ -35,6 +37,15 @@ impl CentralDataSynchroniser {
 
         let cursor_controller = CursorController::new(KeyType::CentralSyncPullCursor);
 
+        let msupply_central_server_id = KeyValueStoreRepository::new(connection)
+            .get_i32(KeyType::SettingsSyncCentralServerSiteId)?;
+
+        log::info!(
```
jmbrunskill, line 43: Not strictly needed, maybe should make this a debug, but thought it was a helpful message to have in the logs?

server/service/src/sync/sync_buffer.rs
```diff
@@ -65,7 +65,7 @@ impl<'a> SyncBuffer<'a> {
                     .action(action.equal_to())
                     .integration_datetime(DatetimeFilter::is_null(true))
                     .source_site_id(match source_site_id {
-                        Some(site_id) => EqualFilter::equal_to_i32(site_id),
+                        Some(site_id) => EqualFilter::equal_any_or_null_i32(vec![site_id]),
```
jmbrunskill, line 68: Need to integration `null` `source_site_id` on remote sites, as records from OMS Central still come through as null.
Should I also do something for OMS Central, I kind of think I should but wanted to get feedback first....

server/service/src/sync/synchroniser.rs
```diff
@@ -279,7 +295,7 @@ impl Synchroniser {
                 false => Some(logger),
                 true => None,
             },
-            None,
+            central_sync_server_id, // Also include OMS Central as source site_id is null, should probably be passing a vec here...
```
jmbrunskill, line 298: As per earlier comment, should we make this a vec, and have a OMS Central Site ID set too?

server/service/src/sync/translations/vaccination_legacy.rs
```diff
@@ -0,0 +1,203 @@
...
+            .find_one_by_id(&changelog.record_id)?
+            .ok_or(anyhow::Error::msg(format!(
+                "Vaccination row ({}) not found",
+                changelog.record_id
+            )))?;
+
+        let legacy_row = LegacyVaccinationRow {
+            ID: row.id.clone(),
+            created_datetime: row.created_datetime.and_utc().to_rfc3339(),
+            user_id: row.user_id.clone(),
```
jmbrunskill, line 87: Not sure if I need all these clones?

jmbrunskill (commented):
Added some notes here. Also need to fix tests...

server/service/src/sync/mod.rs
```diff
@@ -47,6 +47,18 @@ pub(crate) struct ActiveStoresOnSite {
 pub(crate) fn get_sync_push_changelogs_filter(
     connection: &StorageConnection,
 ) -> Result<Option<ChangelogFilter>, GetActiveStoresOnSiteError> {
+    if CentralServerConfig::is_central_server() {
+        // If this is a central server, we want to send everything that that wasn't from legacy site
+        let msupply_central_server_id = KeyValueStoreRepository::new(connection)
+            .get_i32(repository::KeyType::SettingsSyncCentralServerSiteId)
+            .map_err(|_| GetActiveStoresOnSiteError::CentralSiteIdNotSet)?
```
andreievg, line 54: This is note GetActiveStoreOnSiteError, we might need another error type for get_sync_push_changelogs_filter rather then re-using get active store on site error

andreievg: Also for this one in particular there is repository error variant, but the one below yep, something else

server/service/src/sync/sync_buffer.rs
andreievg: I think this integration by 'source_site_id' was done for one use case, when omSupply Central has just received data as a push from omSupply remotes, and needs to integrate those records (just those, not the rest of the sync_buffer because the rest of the sync buffer might have unfinished sync batches). I don't think we need to change this

server/service/src/sync/synchroniser.rs
andreievg: The intention for this was to filter just by one site and the only requirement was for remote sites when they push to central (integration on omSupply central). Now there is this requirement i think we should go to this:

https://github.com/msupply-foundation/open-msupply/compare/8209-sync-vaccination-course...8209-sync-vaccination-course(get-source-site-id-from-sync-buffer)

Interesting enough you were on your way there in this PR: https://github.com/msupply-foundation/open-msupply/pull/4392, but my suggestion changed it to what it is now: https://github.com/msupply-foundation/open-msupply/pull/4392/commits/ba0055445b6fb0062f5487e43ebc3ee57afeb448

I think i was thinking of one use case and also wasn't super happy with mutating integration operation

server/service/src/sync/translations/vaccination_legacy.rs
```diff
@@ -0,0 +1,203 @@
...
+};
+
+/*
+    This translator is only used to push Vaccination rows to the legacy mSupply server.
+    It should run from the Central Server
+*/
+
+#[allow(non_snake_case)]
+#[derive(Serialize)]
+pub struct LegacyVaccinationRow {
```
andreievg, line 17: Worth checking please https://github.com/msupply-foundation/msupply/issues/16422

andreievg: I would suspect name_id should be from name_link table, same with item link

server/repository/src/migrations/v2_09_00/resync_existing_vaccination_records.rs
```diff
@@ -0,0 +1,21 @@
+use crate::migrations::*;
+
+pub(crate) struct Migrate;
+
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "resync_existing_vaccination_records"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
jmbrunskill: I was thinking it effectively is a central server only migration as source_site_id should be null, although if they have re-initialsed that won't be the case...
Do we have a way to make a central server only migration yet? We're not storing central server state right?

server/service/src/sync/translations/vaccination_legacy.rs
jmbrunskill: Good pick up, had somehow just copied over, probably planning to come back to it and forgot...
I don't think we need to do our own name_link mapping right, it's implied that name_ID is a link in mSupply right?

We do have a problem with two name related  fields in OMS and only 1 in mSupply. I've pinged @JonnaQuismundo  about that...

server/repository/src/migrations/v2_09_00/resync_existing_vaccination_records.rs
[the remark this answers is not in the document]
andreievg: So true, we don't have a way to know if server is central at this stage. I might be missing something but this migration should run on remote as it is now ? 
I get it now with the source_site_id is not NULL, can you please add a small comment: "source_site_id" on vaccinations will only be 'not null' on oms central, because it's only set when records are pushed to oms central.

server/service/src/sync/synchroniser.rs
```diff
@@ -167,6 +169,20 @@ impl Synchroniser {
         let site_info = self.remote.sync_api_v5.get_site_info().await?;
         CentralServerConfig::set_central_server_config(&site_info);
 
+        let central_sync_server_id = KeyValueStoreRepository::new(&ctx.connection)
+            .get_i32(KeyType::SettingsSyncCentralServerSiteId)?;
+
+        if central_sync_server_id.is_none() {
```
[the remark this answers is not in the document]
andreievg: > Q: Is there a chance it could change? Should we somehow detect this and adjust records? Haven't considered that use case...

Central server site id would only change if there is transition to-from mirror sync (I think id 1 and 2 are magic numbers). I mean we could have a magic number for mSupply central site id (something that would never be possible for a site to be), I think the current solution should work pretty well

server/service/src/sync/translations/vaccination_legacy.rs
JonnaQuismundo: @jmbrunskill oh if it serves a different purpose? then create the other field in mSupply. Thanks!

andreievg: @jmbrunskill @JonnaQuismundo there is no name_link or item_link concept in mSupply

server/service/src/sync/sync_buffer.rs
jmbrunskill: I think we need to integrate Sync V5 and Sync V6 records together as one block. Should be safe to do Sync V5 then Sync V6 I think? Could be an option instead of doing this?

server/service/src/sync/synchroniser.rs
jmbrunskill: Changed to this approach, but kept the integrating null source site id records from OMS Central.
Other option is to do two integrations on central sync (one for Legacy and one for OMS Central records?)

I also wondered about a magic number for OMS Central site_id like `-1` otherwise we'll need mSupply to provide use with a site_id for OMS Central.
I think this can all be solved when moving to sync v7 though.  I think it's ok to keep OMS Central records with `null` source_site_id

server/repository/src/migrations/v2_09_00/resync_existing_vaccination_records.rs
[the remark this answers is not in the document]
jmbrunskill: Added a comment but realised something else.
`NOTE: This will miss any vaccination records that were created on the central server.`

I think that's ok right, are any clients using Central OMS for vaccinations?

Other option is to just re-sync all vaccinations and just accept a bit of extra sync traffic?

[6c38c77](https://github.com/msupply-foundation/open-msupply/pull/8284/commits/6c38c77b8ef73ba64ce73d7b377299bb2b95c6f5)

server/service/src/sync/translations/vaccination_legacy.rs
jmbrunskill: Updated here [4b3e670](https://github.com/msupply-foundation/open-msupply/pull/8284/commits/4b3e670e44c25251c786e1e210b083379efb9d9e)

server/service/src/sync/synchroniser.rs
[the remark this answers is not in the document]
jmbrunskill: I think even in that case, unless the site_id is re-used for some reason we'll probably be ok keeping the source_site_id as the old one right?

jmbrunskill:
@andreievg I think I've covered the requested changes. Please re-review

server/service/src/sync/synchroniser.rs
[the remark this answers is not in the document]
andreievg: yeah the worst that can happen is re-sync to og central, from current cursor ?

server/repository/src/migrations/v2_09_00/resync_existing_vaccination_records.rs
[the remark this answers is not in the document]
andreievg: > Other option is to just re-sync all vaccinations and just accept a bit of extra sync traffic?

Hmm i like the migration solution you have, i think it should work well thanks

server/service/src/sync/mod.rs
```diff
@@ -47,6 +47,18 @@ pub(crate) struct ActiveStoresOnSite {
...
+            .ok_or(GetActiveStoresOnSiteError::CentralSiteIdNotSet)?;
+
+        return Ok(Some(ChangelogFilter::new().source_site_id(
+            EqualFilter::not_equal_to_i32(msupply_central_server_id),
```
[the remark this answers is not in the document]
andreievg: Thanks @Chris-Petty, @jmbrunskill pointed out that we won't push unless translator exists, which i missed, so we should be fine, won't overload sync buffer with integration error

andreievg:
@jmbrunskill thanks for changes, I've made another suggestion:

* Fix tests (had to revert back to central server = false, this still could have concurrency issues, I did extra test with nextest a few times and seems to work)
* We do want to restrict by source site id only when integrating v6 records on central, to respect the 'atomic batch' and make sure we don't integrate any other "null" source site id, although this seems impossible now that we would capture source site id for v5, I still wanted to be explicit so added source_site_id_filter to integration, checked all usages of integration method and seems correct. You can comment on commit if you think i need to adjust something
* Also updated the error in get sync changelog filter, I do want error to be specific.
* I also deleted extra logic for only setting central sync site id when it's not set (in synchroniser)

https://github.com/msupply-foundation/open-msupply/compare/8209-sync-vaccination-course...8209-sync-vaccination-course(Fix-tests-and-add-explicit-source-site-id-filter)

andreievg (approved):
Thanks @jmbrunskill looks great !

PR #8286, opened 2025-06-25
8280 Prescription: Can't input correct amount of units & input losing focus

### LINKED ISSUE #8280: Prescription: Can't input correct amount of units & input losing focus
## What went wrong? 😲
Prescription: when I try to issue all units available for certain batch lines, the input restricts me to the pack quantity not the unit quantity... In the video I have 45 tablets available but I can only enter 9....

https://github.com/user-attachments/assets/c636f913-d828-4770-b58d-895c56a47d67

BUT if I save the input, in the detail view it says I've issued 9 units  = 1.8 packs 👀 

Also the input keeps losing focus as soon as I stop typing. Allocation allows me to bypass this so the max on individual inputs is probably set to packs.

## How to Reproduce 🔨

1. Go to prescription
2. Have a pack size of > 1 for an item
3. Try to issue all units
4. See that it only allows you to issue max pack quantity instead of unit quantity

## Your environment 🌱

- Open mSupply Version: 2.9, but possibly in 2.8

- Platform:

- Database type:

GeronimoJohn:
Fixes #8280 

# 👩🏻‍💻 What does this PR do?

Updated the max calculation for `NumericTextInput` to use `availablePacks` * `packSize`.

Chose not to use `NumberInputCell`—it has some built-in calculations that trigger right when the value changes. We could debounce the update for longer, but even then we’d run into the same issue: it loses focus and causes the UI to jump around.

Instead, saving now happens on blur. It gives a much smoother UI experience. The only trade-off is that the `Save` button won’t appear clickable until the user clicks out of the input.

# 🧪 Testing

- [ ] Go to prescription
- [ ] Have a pack size of > 1 for an item
- [ ] Try to issue all units
- [ ] Should let you issue all of the units
- [ ] Click out of the input
- [ ] Save button is now clickable

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/columns.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/columns.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import {
   CellProps,
   ColumnAlign,
@@ -6,8 +6,9 @@ import {
   ExpiryDateCell,
   Formatter,
   NumberCell,
-  NumberInputCell,
+  NumericTextInput,
   PreferenceKey,
+  useBufferState,
   useColumns,
   useIntlUtils,
   usePreference,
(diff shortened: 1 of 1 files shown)


CarlosNZ (dismissed):
Hi @GeronimoJohn, 

I don't really like having to click outside the input before the "Save" button becomes enabled -- we usually expect that users can go straight from the input field to the Save button.

So I spent a bit of time trying to solve the problem of the focus loss, since it works fine elsewhere. It seems the problem is something to do with the "allocate" function, since it's doing some side effects rather than updating the value directly. 

This should be possible to fix though, since it's working in the overall "Prescribed" and "Issue" quantity inputs above. Can I sugggest having a look in the `AutoAllocatePrescribedQuantityField` component and trying to copy how it's working there? It looks like that their is a `debounceAllocate` function that should probably help us out.

 Not totally sure though, so if I'm barking up the wrong tree please let me know and we'll try and come up with another solution.

I guess we probably don't want to spend too long on this, so let's just keep in touch.

client/packages/invoices/src/Prescriptions/LineEditView/columns.tsx
```diff
@@ -150,9 +150,10 @@ export const usePrescriptionLineEditColumns = ({
...
   <NumberInputCell
     {...props}
-    max={props.rowData.availablePacks}
+    max={props.rowData.availablePacks * props.rowData.packSize}
     decimalLimit={2}
     min={0}
+    debounce={1000}
     slotProps={{ htmlInput: { sx: { backgroundColor: 'white' } } }}
   />
 );
```
GeronimoJohn, line 159: Findings: 

`UnitQuantityCell` is part of columns which uses `draftLines`. Draft lines is updated through the `manualAllocate` method or the setter that is called on change for this column. But because of this the table will always re-render on update. Losing focus after typing/debounce times out. 

Attempted solutions like memoizing, followed patterns from `AutoAllocatePrescribedQuantity`, and additional debouncing don't work because draftLines fundamentally changes on every allocation update, forcing a re-render.

`AutoAllocatePrescribedQuantityField` doesn't experience this issue because prescribedQuantity is an independent state value that isn't affected by external allocation changes

We could use `useRef`, but feels like smelly code doing so

GeronimoJohn: Solution: Just make it debounce for 1000, which seems to be enough time to type

GeronimoJohn: The actual issue of losing focus is coming from DataRow. Getting fixed in a separate issue

GeronimoJohn:
I've branched off develop for this fix. But the fix was meant for 2.8.3. Changing heads brings out a major chunk of difference. Since its a one liner fix. Easier to close and just branch off 2.8.3. That being said

Closing PR and will be recreated

PR #8318, opened 2025-06-30
4890 received status

### LINKED ISSUE #4890: Can't mark an Inbound Shipment as 'Delivered' without it affecting current stock levels
## What went wrong? 😲

The current workflow does not allow a user to confirm receipt of a delivery without triggering stock levels to be updated, regardless of whether they have had time to verify the contents of the delivery or not. 

These steps are separate in medium to large facilities as it takes time to verify the contents. 

It would be preferable to wait until the delivery has been marked as 'Verified' to update stock levels. 

As soon as this is marked as 'Delivered', the stock levels will be increased by 800 and can be distributed to customers (even if the stock doesn't actually exist):

telegram-cloud-photo-size-1-5163525153903193193-y

## Expected behaviour 🤔

Stock levels should not be updated until the delivery is 'Verified'

## How to Reproduce 🔨

1. Open a new Inbound Shipment
2. Mark the shipment as 'Delivered'
3. Note stock levels have been updated

## Your environment 🌱

- Open mSupply Version: 2.2.2
- Legacy mSupply Central Server Version: 7-17-10

jmbrunskill:
Fixes #4890

# 👩🏻‍💻 What does this PR do?

Adds a new status for inbound shipments called `Delivered`, and moves the old `Delivered` status to `Received`.
Adds test to check delivered status does not update stock and received status does
Customer Returns now go straight to `Received` status, skipping `Delivered`.

I think there might be an enum issue when transferring between older and new sites as the new `Received` status will fail when translating the sync record.I think it should be ok as it should just jump from delivered to verified,but we need to test.

Supplier Returns haven't been updated, but probably need to be, should skip delivered too?

# 🧪 Testing

- [ ] Test creating inbound shipments, via as many different methods as possible (e.g. based on internal order from other store, manual, remote stores vs local stores, mSupply vs OMS)
- [ ] Make sure you can set inbound shipment status to `Delivered`
- [ ] Make sure no stock updates happen at that time
- [ ] Make sure you can set the inbound shipment to received
- [ ] Make sure stock is updated when the inbound shipment is marked as delivered or verified
- [ ] Test Customer and Supplier Returns still work correctly

# 📃 Documentation

  1. Inbound Shipments
  2. Customer Returns

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -2087,5 +2087,6 @@
   "warning.check-before-vaccinating": "Vaccination records from other stores may not be up to date. Ensure you have synced recently, and always confirm against the patient's physical vaccination card first.",
   "warning.field-not-parsed": "{{field}} not parsed",
   "warning.nothing-to-supply": "Nothing left to supply!",
+  "label.invoice-received": "Received",
```
jmbrunskill, line 2090: I thought invoice-received might be clearer for the meaning.
But I'm actually now wondering about `invoice-status-received` or something would be better?

client/packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -51,15 +57,21 @@ const getStatusOptions = (
...
     options[3].isDisabled = false;
     options[4].isDisabled = false;
+    options[5].isDisabled = false;
   }
 
-  // When the status is delivered, only verified is available to select.
+  // When the status is Delivered, only verified & Delivered re available to select.
   if (currentStatus === InvoiceNodeStatus.Delivered) {
     options[4].isDisabled = false;
+    options[5].isDisabled = false;
```
jmbrunskill, line 70: Would be nice to refactor to not rely on the indexes here, as it's really messy to update and to verify if it's correct

CarlosNZ:
Getting a migration error (using Postgres):

```sql
INSERT INTO changelog (record_id, row_action, store_id)
    SELECT id, 'UPSERT', store_id
    FROM invoice
    WHERE received_datetime IS NOT NULL
    AND store_id in (SELECT store_id FROM store WHERE site_id in (SELECT value_int FROM key_value_store WHERE id = 'SETTINGS_SYNC_SITE_ID'))
```

Gives: **NOT NULL constraint failed: changelog.table_name**

server/repository/src/migrations/views/mod.rs
```diff
@@ -102,12 +102,12 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
         quantity_movement as quantity,
         item_id,
         store_id,
-        delivered_datetime as datetime
+        received_datetime as datetime
```
CarlosNZ, line 105: Do you think we should have both `delivered` and `recieved` fields in the View?

CarlosNZ (changes requested):
Looking good, quite a big job this one. Hope we've covered everything. Couple of minor suggestions/tweaks.

One thing I wondered more generally -- if we've subtly changed what "Delivered" means for Inbounds, have we implied something about what "Delivered" means for Outbounds? Does it matter?

client/packages/invoices/src/OutboundShipment/DetailView/Footer/Footer.tsx
```diff
@@ -33,6 +33,7 @@ const createStatusLog = (invoice: OutboundFragment) => {
     [InvoiceNodeStatus.Delivered]: null,
     [InvoiceNodeStatus.Verified]: null,
     [InvoiceNodeStatus.Cancelled]: null,
+    [InvoiceNodeStatus.Received]: null,
```
andreievg, line 36: Should this be here, if yep should it be moved between delivered/verified with correct datetime mapped ?

andreievg: Same in prescriptions

client/packages/invoices/src/Returns/CustomerDetailView/Footer/Footer.tsx
```diff
@@ -44,7 +45,7 @@ const createStatusLog = (invoice: CustomerReturnFragment) => {
     statusLog[InvoiceNodeStatus.Shipped] = invoice.shippedDatetime;
   }
   if (statusIdx >= 4) {
-    statusLog[InvoiceNodeStatus.Delivered] = invoice.deliveredDatetime;
+    statusLog[InvoiceNodeStatus.Received] = invoice.receivedDatetime;
```
andreievg, line 48: What about deliveredDatetime ?

client/packages/invoices/src/utils.ts
```diff
@@ -37,19 +37,22 @@ export const inboundStatuses: InvoiceNodeStatus[] = [
...
   InvoiceNodeStatus.Delivered,
+  InvoiceNodeStatus.Received,
   InvoiceNodeStatus.Verified,
 ];
 
 export const nextStatusMap: { [k in InvoiceNodeStatus]?: InvoiceNodeStatus } = {
   [InvoiceNodeStatus.New]: InvoiceNodeStatus.Delivered,
   [InvoiceNodeStatus.Shipped]: InvoiceNodeStatus.Delivered,
-  [InvoiceNodeStatus.Delivered]: InvoiceNodeStatus.Verified,
+  [InvoiceNodeStatus.Delivered]: InvoiceNodeStatus.Received,
```
andreievg, line 54: It feels like the idea of all of those, is for them to be use in the StatusChange Toolbars, you are right about refactor

client/packages/common/src/intl/locales/en/common.json
jmbrunskill: Ok, but I think we should probably refactor them all to better names. I think not having proper context is bad for translations...

server/service/src/invoice/customer_return/update/generate.rs
```diff
@@ -112,24 +112,31 @@ fn set_new_status_datetime(customer_return: &mut InvoiceRow, patch: &UpdateCusto
     let current_datetime = Utc::now().naive_utc();
 
     match (&customer_return.status, new_status) {
-        // From New/Picked/Shipped to Delivered
+        // From New/Picked/Shipped to Received (Always skip Delivered for Customer Returns)
         (
             InvoiceStatus::New | InvoiceStatus::Picked | InvoiceStatus::Shipped,
```
andreievg, line 117: One more case here, to 'Delivered'

server/service/src/processors/transfer/invoice/update_outbound_invoice_status.rs
```diff
@@ -76,7 +76,7 @@ impl InvoiceTransferProcessor for UpdateOutboundInvoiceStatusProcessor {
         let updated_outbound_invoice = InvoiceRow {
             // 7.
             status: inbound_invoice.invoice_row.status.clone(),
-            delivered_datetime: inbound_invoice.invoice_row.delivered_datetime,
+            received_datetime: inbound_invoice.invoice_row.received_datetime,
```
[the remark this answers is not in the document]
jmbrunskill: Yes, I think Louisa found that bug!

server/service/src/sync/translations/invoice.rs
```diff
@@ -725,7 +739,7 @@ fn to_legacy_confirm_time(
     InvoiceRow {
         r#type,
         picked_datetime,
-        delivered_datetime,
+        received_datetime: delivered_datetime,
```
andreievg, line 742: there is only two fields to change, I think should be explicit here

server/repository/src/migrations/v2_08_03/invoice_received_status.rs
```diff
@@ -0,0 +1,55 @@
...
+                received_datetime = delivered_datetime
+                WHERE delivered_datetime is not null
+
+                -- Set all existing invoices for this site that were Delivered to Received
+                UPDATE invoice
+                SET status = 'RECEIVED'
+                WHERE status = 'DELIVERED';
+
+                --Create changelogs to resync all the invoices that were Delivered before and are for this sync site
+                INSERT INTO changelog (record_id, table_name, row_action, store_id)
```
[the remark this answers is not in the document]
jmbrunskill: Yes unfortunately!

server/repository/src/migrations/views/mod.rs
jmbrunskill: I think because it's stock movement, we only want the stock movement ones,which means it doesn't include delivered_datetime? Unless I'm missing something?

server/service/src/invoice/customer_return/update/generate.rs
jmbrunskill: Yes, I agree! I changed to that in this PR as I think this approach is prone to errors...

server/service/src/sync/translations/invoice.rs
jmbrunskill: Comment still relevant?

CarlosNZ (approved):
Functionality appears correct from my POV. I was going to comment on the failing tests, but I see you're fixing those.

All good from me then, just wait for Andrei to have a final look I guess

server/service/src/sync/translations/invoice.rs
```diff
@@ -174,6 +216,11 @@ pub struct LegacyTransactRow {
     #[serde(deserialize_with = "empty_str_as_option")]
     pub delivered_datetime: Option<NaiveDateTime>,
 
+    // To be added when available in mSupply
+    // #[serde(default)]
+    // #[serde(rename = "om_received_datetime")]
+    // #[serde(deserialize_with = "empty_str_as_option")]
+    // pub received_datetime: Option<NaiveDateTime>,
     #[serde(default)]
```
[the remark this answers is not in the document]
jmbrunskill: Leaving this for now, as we need a follow up issue to implement in mSupply

server/service/src/sync/translations/invoice.rs
andreievg: It does, it's a minor though

server/service/src/sync/translations/invoice.rs
```diff
@@ -815,31 +883,35 @@ fn legacy_invoice_type(_type: &InvoiceType) -> Option<LegacyTransactType> {
...
-            InvoiceStatus::Cancelled => LegacyTransactStatus::Fn, // TODO enable cancelled status to sync with mSupply https://github.com/msupply-foundation/open-msupply/issues/6495
+            InvoiceStatus::Cancelled => LegacyTransactStatus::Fn,
         },
         InvoiceType::Prescription => match status {
             InvoiceStatus::New => LegacyTransactStatus::Nw,
             InvoiceStatus::Allocated => LegacyTransactStatus::Cn,
             InvoiceStatus::Picked => LegacyTransactStatus::Cn,
             InvoiceStatus::Shipped => LegacyTransactStatus::Fn,
-            InvoiceStatus::Delivered => LegacyTransactStatus::Fn,
+            InvoiceStatus::Delivered => LegacyTransactStatus::Sg,
```
andreievg, line 913: Probably should be 'InvoiceStatus::New ' ? I am not sure about this 'Sg' status now

PR #8327, opened 2025-06-30
8315 input cell loses focus for datatable

### LINKED ISSUE #8315: Input cell loses focus after updating value
## What went wrong? 😲

This is a continuation of #8286. We have found that the input cell is losing focus on update. This is happening due to the table getting re-rendered once a value gets updated. This is Zustands default way of running things due to my findings. 

## Expected behaviour 🤔

Keep focus even after typing within the input cell.

## How to Reproduce 🔨

1. Navigate to Dispensary -> Prescriptions -> Select a prescription with an item
2. Open quantity section
3. Add specific value for item line
4. Watch it lose focus

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

GeronimoJohn:
Fixes #8315 

# 👩🏻‍💻 What does this PR do?

Previously, `ColumnContent` and `ContentWrapper` were defined inside `DataRowComponent`, causing them to be recreated on every render. Because variables like `isDisabled` and others from the parent scope change, this led to new component identities and unnecessary rerenders. Moving these components outside ensures stable references and better rendering performance

# 🧪 Testing

- [ ] Navigate to any page that uses DataTable and have an input (e.g. Prescriptions Edit Page, Stocktake Edit page, Inbound Shipment Edit page, etc)
- [ ] Add value to the Input cell, (if there's more than one - add to all of them)
- [ ] Input shouldn't lose focus!

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/client/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -71,66 +71,8 @@ const DataRowComponent = <T extends RecordWithId>({
 
   useEffect(() => {
     if (isFocused) onRowClick();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [keyboardActivated]);
-
-  interface ColumnContentProps<T extends RecordWithId> {
-    column: Column<T>;
-    columnIndex: number;
-    isError: boolean | undefined;
-  }
-
-  const ColumnContent = ({
-    column,
-    columnIndex,
-    isError,
-  }: ColumnContentProps<T>) => (
-    <column.Cell
-      isDisabled={isDisabled || column.getIsDisabled?.(rowData)}
-      rowData={rowData}
-      columns={columns}
-      isError={isError}
-      column={column}
-      rowKey={rowKey}
-      columnIndex={columnIndex}
-      rowIndex={rowIndex}
-      autocompleteName={column.autocompleteProvider?.(rowData)}
-      localisedText={localisedText}
-      localisedDate={localisedDate}
-      dense={dense}
-      rowLinkBuilder={rowLinkBuilder}
-      {...column.cellProps}
-    />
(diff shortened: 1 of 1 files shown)


roxy-dao:
Should this not go into 2.8.3 since the precursor did? 👀

CarlosNZ (approved):
Thanks @GeronimoJohn -- nice work tracking this down.

I have some reservations about the whole ColumnWrapper/ColumnContent design, and why it's necessary, but that's not for this issue.

CarlosNZ:
> I have some reservations about the whole ColumnWrapper/ColumnContent design, and why it's necessary, but that's not for this issue.

I feel like at the very least "ColumnContent" is misnamed. Should it be "CellContent"? And "ContentWrapper" can be "CellContentWrapper"?

GeronimoJohn:
> Thanks @GeronimoJohn -- nice work tracking this down.
> 
> I have some reservations about the whole ColumnWrapper/ColumnContent design, and why it's necessary, but that's not for this issue.

Same here, would be nice to clean up during cooldown. This was a nice quick and somewhat clean fix for now

GeronimoJohn:
Hey @CarlosNZ, can I please get another review. I got carried away while changing the component names 😁 . I've added tests and move `CellContentWrapper` into its own file. I ended up not making `CellContent` as its own thing, it feels like just a thin wrapper for `column.Cell` . 

If we think these changes are a bit much I can go back and just change the names as initially suggested 😅

CarlosNZ (approved):
Nice, thanks @GeronimoJohn -- tests as well!!

There's a failing action, but that seems to be the base build, which this PR is fixing. Not sure how that slipped through though :/

PR #8332, opened 2025-06-30
8299 distributed quantity field gets automatically replaced by minimum pack size before the user finishes typing

### LINKED ISSUE #8299: Distributed quantity field gets automatically replaced by minimum pack size before the user finishes typing
## What went wrong? 😲

When entering a quantity for distribution, the input field automatically replaces the entered number with the minimum pack size value as soon as the typed number is temporarily lower than the minimum pack size — even if the user hasn't finished typing.

Example Scenario:
Stock counted in doses, minimum pack size = 20 doses (one vial = 20 doses)
User wants to distribute 1000 doses
Starts typing:
1 → field sees 1 < 20 → replaces with 20
Or
10 → field sees 10 < 20 → replaces with 20
As a result, unless the user types very quickly, they can't input numbers like 1000, because the system intervenes after the first few digits.

## Expected behaviour 🤔

The system should only validate or auto-correct the number after the user finishes typing — either:
After a short delay or After the user confirms the entry (presses enter/tab or moves to another field)

## How to Reproduce 🔨

1. Go to Outbound shipment → New shipment
2. Select a Client and a Master list (preferably with vaccine items)
3. Select an item that has a defined pack size (e.g., 20 doses per pack)
4. In the quantity field, try to enter a number that is higher than the pack size, but whose initial digits are smaller than the pack size
Example: Try to enter 1000 when the pack size is 20
5. See error: It becomes impossible to enter the desired quantity unless you type very fast
As you type 1 or 10, the system replaces the number with 20 before you can finish typing

https://github.com/user-attachments/assets/262dc4d4-57ab-42d7-b029-8c24646ed352

## Your environment 🌱

- Open mSupply Version: demo server 2.8.2. and live server 2.8.2 

- Platform: Tablets (Samsung Galaxy A8) and Mac 15.3

- Database type:

@Cyrilbrites 
@richardmoizeau

GeronimoJohn:
Fixes #8299 

# 👩🏻‍💻 What does this PR do?

Updates debounce time from 500 to 1000 (1 sec) after a user types. 

Took the time to think about how long our users actually take to type something. 1 second seems after a user type seems to be enough, but some might be new to typing and can be slower with typing. 

Anyways thought could start off with bumping debounce to 1 sec and wait for any user to mention anything about it

# 🧪 Testing

- [ ] Go to Outbound shipment → New shipment
- [ ] Select a Client and a Master list (preferably with vaccine items)
- [ ] Select an item that has a defined pack size (e.g., 20 doses per pack)
- [ ] In the quantity field, try to enter a number that is higher than the pack size, but whose initial digits are smaller than the pack size. Example: Try to enter 1000 when the pack size is 20
- [ ] Should be able to enter values without the system overriding it straight away
- [ ] System should override value 1 second after typing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/StockOut/Components/AutoAllocateIssueField.tsx
+++ b/client/packages/invoices/src/StockOut/Components/AutoAllocateIssueField.tsx
@@ -27,29 +27,18 @@ export const AutoAllocateField = ({
     allocatedQuantity: getAllocatedQuantity(state),
     allocateIn: state.allocateIn,
   }));
-
-  // Using buffer state with the allocated quantity, so gets pre-populated with existing
-  // quantity, and updated when the user edits the individual lines
   const [issueQuantity, setIssueQuantity] = useBufferState(allocatedQuantity);
 
-  // using a debounced value for the allocation. In the scenario where
-  // you have only pack sizes > 1 available, and try to type a quantity which starts with 1
-  // e.g. 10, 12, 100.. then the allocation rounds the 1 up immediately to the available
-  // pack size which stops you entering the required quantity.
-  // See https://github.com/msupply-foundation/open-msupply/issues/2727
-  // and https://github.com/msupply-foundation/open-msupply/issues/3532
   const debouncedAllocate = useDebounceCallback(
     quantity => {
       const allocated = autoAllocate(quantity, format, t, allowPartialPacks);
       setIssueQuantity(allocated);
     },
     [],
-    500
+    1000
   );
 
   const handleIssueQuantityChange = (quantity: number | undefined) => {
-    // this method is also called onBlur... check that there actually has been a change
-    // in quantity (to prevent triggering auto allocation if only focus has moved)
     if (quantity === issueQuantity) return;
 
     // Set immediate value to the input


client/packages/invoices/src/StockOut/Components/AutoAllocateIssueField.tsx
```diff
@@ -27,29 +27,18 @@ export const AutoAllocateField = ({
...
       setIssueQuantity(allocated);
     },
     [],
-    500
+    1000
   );
 
   const handleIssueQuantityChange = (quantity: number | undefined) => {
-    // this method is also called onBlur... check that there actually has been a change
-    // in quantity (to prevent triggering auto allocation if only focus has moved)
```
GeronimoJohn, line 52: Not really being called in onBlur? Removed due to that reason

client/packages/invoices/src/StockOut/Components/AutoAllocateIssueField.tsx
```diff
@@ -27,29 +27,18 @@ export const AutoAllocateField = ({
...
-  // Using buffer state with the allocated quantity, so gets pre-populated with existing
-  // quantity, and updated when the user edits the individual lines
   const [issueQuantity, setIssueQuantity] = useBufferState(allocatedQuantity);
 
-  // using a debounced value for the allocation. In the scenario where
-  // you have only pack sizes > 1 available, and try to type a quantity which starts with 1
-  // e.g. 10, 12, 100.. then the allocation rounds the 1 up immediately to the available
-  // pack size which stops you entering the required quantity.
-  // See https://github.com/msupply-foundation/open-msupply/issues/2727
-  // and https://github.com/msupply-foundation/open-msupply/issues/3532
```
GeronimoJohn, line 40: The comments makes sense. But the code seems to be straight forward enough to understand without the comment?

CarlosNZ (approved):
Thanks, I think this is fine for now.

We need a more robust solution eventually.

PR #8356, opened 2025-07-03
Remove filter that masterList exists on store

### LINKED ISSUE #7705: Masterlist removed from store - items with stock not visible in view stock screen
## What went wrong? 😲

When a masterlist is removed from a store but there is still stock in the store you can no longer see it in the View Stock screen. However you can see it on the items screen and outbound shipment.

This will affect our future deployments since we have tried to reduce the masterlist here and clients are having trouble seeing their inventory with stock but not on a masterlist.

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #7705

# 👩🏻‍💻 What does this PR do?""

Show stock regardless of if masterList exists on store.

Note - this query is only used on the stock ListView component - so restrictions on adding stock on a masterList not visible on the store remain

# 🧪 Testing

- [ ] Add stock to a store where master list exists on store
- [ ] Remove masterlist from store
- [ ] See stock remains visible

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Stock/api/hooks/useStockList.ts
+++ b/client/packages/system/src/Stock/api/hooks/useStockList.ts
@@ -37,10 +37,6 @@ export const useStockList = (queryParams: StockListParams) => {
     const filter = {
       ...filterBy,
       hasPacksInStore: true,
-      masterList: {
-        existsForStoreId: { equalTo: storeId },
-        ...filterBy?.masterList,
-      },
     };
     const query = await stockApi.stockLines({
       storeId,


jmbrunskill (approved):
Looks great Fergus. I'm wondering if we should be put this into 2.9 since it's such a small change, and will make it available much sooner...

PR #8363, opened 2025-07-03
7958 refactor migration column adds

### LINKED ISSUE #7958: Potential data corruption risk with legacy migrations
## What went wrong? 😲

When diagnosis was added to prescription, we re-integrated the sync buffer, however if changes had been made to the invoices since mSupply sync this would cause corruption of the data.
```
        // Retranslate all invoices on the next sync, to make sure we capture the diagnosis_id if populated. DON'T DO THIS!
        sql!(
            connection,
            r#"
            UPDATE sync_buffer SET integration_datetime = NULL WHERE table_name = 'invoice';
        "#,
        )?;
```

## Expected behaviour 🤔

I think we've been lucky so far but should remove all remote data re-integration from older migraitions, and make sure this doesn't get though furture code reviews...

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

fergie-nz:
Fixes #7958 

# 👩🏻‍💻 What does this PR do?

- initial commit (f17c46a77ea9e0b7bf7a848396d141912016bce2) moves serialisation functions into utils file. This is so that they can be used in migrations to more closely mimic translation migration behaviour.

- Removes resetting of integration datetime for data which is editable on remote sites to prevent corruption on post migration edits before sync.

- Adds adding missed sync values in migration for next of kin id on name table

- Does not add missed sync value for diagnosis_id on invoice. This is because diagnosis id references diagnosis which won't yet exist in DB, so will result in foreign key violation.

The previous method which reset integration datetime would work and would add diagnosis id to invoice, however still risks data corruption where data is edited on remote site after migration but before syncing.

Some possible solutions to this:

a - Add a similar migration of dependency table data in the sync buffer into the db during migrations. A simplification of this approach could be to add dependency table with default values other than the id, so that we don't need to implement a full translation. This might then be cleaned up later from the sync_buffer.

b - leave it be an accept old invoices won't have diagnosis id attached (which is what this PR does)

Other cases where integration datetime is set to null are fine - they are used as central data only. (does this need to be reexamined if central data later becomes remote editable?)

# 🧪 Testing

Run test cases

Manual testing would involve:

- [ ] run OMS on a version pre 2.5.0 where OG has names with next_of_kin_id
- [ ] sync
- [ ] Migrate to 2.5.0 or later
- [ ] Confirm that next_of_kin_id has been added to name

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/v2_06_00/add_name_next_of_kin_id.rs
```diff
@@ -16,14 +39,104 @@ impl MigrationFragment for Migrate {
...
+            diesel::update(name::table)
+                .filter(name::id.eq(name_id))
+                .set(name::next_of_kin_id.eq(legacy_row.next_of_kin_id))
+                .execute(connection.lock().connection())?;
+        }
 
         Ok(())
     }
 }
+
```
fergie-nz, line 64: FYI this test was primarily written by AI

server/repository/src/migrations/v2_05_00/diagnosis_add_to_invoice.rs
```diff
@@ -15,14 +15,6 @@ impl MigrationFragment for Migrate {
         "#
         )?;
```
jmbrunskill, line 17:
```suggestion
// We don't re-translate existing invoices here, as they could have been changed since the sync buffer record and could be overwritten
```

jmbrunskill (approved):
Looks good Fergus. Have you run all the tests in the PR Description yourself?
I haven't done all the testing, but I'm happy with the code.

server/repository/src/migrations/v2_05_00/diagnosis_add_to_invoice.rs
```diff
@@ -14,15 +14,7 @@ impl MigrationFragment for Migrate {
             ALTER TABLE invoice ADD COLUMN diagnosis_id TEXT REFERENCES diagnosis(id);
         "#
         )?;
-
-        // Retranslate all invoices on the next sync, to make sure we capture the diagnosis_id if populated.
-        sql!(
-            connection,
-            r#"
-            UPDATE sync_buffer SET integration_datetime = NULL WHERE table_name = 'invoice';
```
fergie-nz, line 22: invoice table name in sync buffer is 'transact' - so this migration wasn't active in the first place 

(Probably a good thing as we would have risked corrupting data)

PR #8369, opened 2025-07-03
Ordered vs received variance

roxy-dao:
Part of https://github.com/msupply-foundation/open-msupply-reports/issues/218

# 👩🏻‍💻 What does this PR do?
Stops the helper text from expanding the date filters~
Screenshot 2025-07-04 at 9 52 05 AM

# 🧪 Testing

- [ ] Report not available yet for testing...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/fr/common.json
+++ b/client/packages/common/src/intl/locales/fr/common.json
@@ -2090,6 +2090,6 @@
   "warning.caps-lock": "Avertissement: Verrouillage majuscule activé",
   "warning.check-before-vaccinating": "La carte de vaccination éléctronique provenant d'autres sites de vaccination peuvent ne pas être à jour ! Assurez-vous d'avoir effectué une synchronisation récente et vérifiez toujours la validité de la carte de vaccination papier du patient.",
   "warning.field-not-parsed": "{{field}} non analysé",
-  "warning.manual-store-requisition": "Saisir une réquisition pour ce dépôt peut  donner lieu à un doublon si le dépôt crée également une commande interne.",
+  "warning.manual-store-requisition": "Attention - La saisie d'une réquisition client pour un dépôt peut entraîner des doublons si ce dépôt soumet aussi une commande interne.",
   "warning.nothing-to-supply": "La quantité demandée à été fournie !"
 }
--- a/client/packages/common/src/intl/locales/fr/common.json
+++ b/client/packages/common/src/intl/locales/fr/common.json
@@ -1095,7 +1095,7 @@
   "label.patient-type": "Type de patient",
   "label.patients": "Patients",
   "label.payment-amount-outstanding": "Montant restant à payer",
-  "label.payment-amount-paid": "Montant payé",
+  "label.payment-amount-paid": "Montant reçu",
   "label.payment-amount-to-be-paid": "Montant à payer",
   "label.payment-change": "Monnaie à rendre",
   "label.payment-type": "Type de paiement",
--- a/client/packages/programs/src/JsonForms/common/components/DateTime.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/DateTime.tsx
@@ -52,7 +52,7 @@ const UIComponent = (props: ControlProps) => {
   const inputFormat = !dateOnly ? 'P p' : 'P';
 
   const onChange = (e: Date | null) => {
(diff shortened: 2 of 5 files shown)


client/packages/programs/src/JsonForms/common/components/DateTime.tsx
```diff
@@ -52,7 +52,7 @@ const UIComponent = (props: ControlProps) => {
   const inputFormat = !dateOnly ? 'P p' : 'P';
 
   const onChange = (e: Date | null) => {
-    if (!e) handleChange(path, null);
+    if (!e) handleChange(path, undefined);
```
roxy-dao, line 55: Wasn't clearing properly~

client/packages/common/src/intl/locales/fr/common.json
```diff
@@ -1095,7 +1095,7 @@
   "label.patient-type": "Type de patient",
   "label.patients": "Patients",
   "label.payment-amount-outstanding": "Montant restant à payer",
-  "label.payment-amount-paid": "Montant payé",
+  "label.payment-amount-paid": "Montant reçu",
```
roxy-dao, line 1098: @Cyrilbrites requested for this change

lache-melvin (changes requested):
I can't quite tell why, but the styles only apply for me when I apply them from here? I think its okay though, this should probs always be the case, can't see a reason why date picker should ever be super wide 👀 

https://github.com/msupply-foundation/open-msupply/compare/Ordered-vs-Received-Variance...Ordered-vs-Received-Variance(move-styles)

client/packages/programs/src/JsonForms/components/DateRange.tsx
```diff
@@ -78,7 +78,9 @@ const UIComponent = (props: ControlProps) => {
             onChange={date => updateDate('afterOrEqualTo', date)}
             disabled={!props.enabled}
             actions={actions}
-            maxDate={DateUtils.getDateOrNull(data?.to) ?? undefined}
+            maxDate={
+              DateUtils.getDateOrNull(data?.beforeOrEqualTo) ?? undefined
```
roxy-dao, line 82: Fix because this actually isn't working... Also makes me question if the encounters report is actually filtering out correctly? Doesn't appear to be.

client/packages/common/src/intl/locales/fr/common.json
```diff
@@ -2090,6 +2090,6 @@
   "warning.caps-lock": "Avertissement: Verrouillage majuscule activé",
   "warning.check-before-vaccinating": "La carte de vaccination éléctronique provenant d'autres sites de vaccination peuvent ne pas être à jour ! Assurez-vous d'avoir effectué une synchronisation récente et vérifiez toujours la validité de la carte de vaccination papier du patient.",
   "warning.field-not-parsed": "{{field}} non analysé",
-  "warning.manual-store-requisition": "Saisir une réquisition pour ce dépôt peut  donner lieu à un doublon si le dépôt crée également une commande interne.",
+  "warning.manual-store-requisition": "Attention - La saisie d'une réquisition client pour un dépôt peut entraîner des doublons si ce dépôt soumet aussi une commande interne.",
```
roxy-dao, line 2093: Cyril asked for this to change~

client/packages/programs/src/JsonForms/components/DateRange.tsx
lache-melvin: Oh rip, how was that working 👀 

I think DateRange has value - we pretty often need a date range filter on a report. The component handles the max/min logic in one place, so we don't need to keep implementing on the JSON schema

lache-melvin (approved):
Pushed fix for the value fields on that DateRange component too lol, they weren't repopulating when you reopen the filter modal 🫠 

Thanks Roxy 🙏

client/packages/programs/src/JsonForms/components/DateRange.tsx
roxy-dao: Hmm, it's only used in the Encounters though... I've used the DateTime one for this report, works just the same~

Issue #8137, opened 2025-06-11
Label Printing is Not Working and Making Printer Unaccessible


kahn-msupply:
> UPDATE - some logs around this have gone into 2.9 but keeping issue open, further work needed

## What went wrong? 😲

When printing from Open mSupply:

- The test printer configuration says 'Connected successfully to printer!'
- Going to a prescription and printing it says 'Label printed successfully.'
- Check printer, no label has been printed
- Try printing, and it fails with the message 'Error printing label.: connection timed out'
- Before printing a prescription, you can visit the printer webui
- After printing the prescription, you can no longer visit the printer webui
- The printer has to be power-cycled to get it to respond again

Logs set to 'Debug', and only the following is recorded:

`2025-06-18 12:57:25.959502000 [INFO ] <actix_web::middleware::logger:424>:127.0.0.1 "POST /print/label-prescription HTTP/1.1" 200 13 "http://localhost:3003/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36" 1.008868
2025-06-18 12:57:51.753221000 [INFO ] <actix_web::middleware::logger:424>:127.0.0.1 "POST /print/label-prescription HTTP/1.1" 500 20 "http://localhost:3003/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36" 5.005547`

Tested with two Zebra label printers and two different networks

@louisaw123 has requested to try and have this fixed before August for a deployment

## Expected behaviour 🤔

Connection Test:
![Image](https://github.com/user-attachments/assets/51b67909-6bfa-4a9d-a279-f150a3d4ad34)

Label Printed (Which it isn't)
![Image](https://github.com/user-attachments/assets/3ac6caf2-a47f-49df-bc76-19463b4514c9)

Try printing again, get error:
![Image](https://github.com/user-attachments/assets/facb94a9-eba8-441e-bc15-37cf0c4d102f)

Before and after printing (Printer webui):

![Image](https://github.com/user-attachments/assets/e8a5d556-f865-438a-81bb-9c35902c3901)
![Image](https://github.com/user-attachments/assets/2d717254-eb99-4eca-bdb8-19a1e8907779)

## How to Reproduce 🔨

[...]

Issue #7243, opened 2025-04-07
Improve API logging


andreievg:
## Is your feature request related to a problem? Please describe 👀

@jmbrunskill mentioned a few times that our logs are not necessarily helpful when it comes to API. I agree.

I think we should log all errors in API to console/file.

For starters it can be all standard errors (this can be done through async graphql logging middleware, but we should also log structured errors

andreievg:
https://github.com/msupply-foundation/open-msupply/pull/7206#discussion_r2027772576

andreievg:
Encourage logging in 'info' mode for other things like 'processors steps'.

PR #8418, opened 2025-07-09
8137 Move HTTP logs to debug level, add graphql logs

lache-melvin:
Another part of #8137 

# 👩🏻‍💻 What does this PR do?

Was raised in discussion, that the HTTP logs are noisy, and not particularly helpful.

Moves them to `DEBUG` log level, and strips some of the user agent etc. noise from the format:

Screenshot 2025-07-09 at 4 57 34 PM

Our request logs are especially unhelpful, because the majority of our requests are graphql, so always to `/graphql`, and even when graphql returns an error, its still HTTP 200 status. We don't have access to the request/response body in the logger, just the headers, so this is about all the info we can show...

So logging the requests at all only really adds value for the REST requests we have. Maybe its helpful to see that requests are being received/responded to?

I'm wondering about something [like this](https://docs.rs/actix-web/latest/actix_web/middleware/struct.Logger.html#method.custom_response_replace) - and we leave as info level, but only log non-graphql requests?

I think with #7243 - we'd get more value by logging in each query/mutation, the name of the query/mutation (+ maybe params? Just not login 😁), and then the result before returning.

# 🧪 Testing

- [ ] Restart server - with default Info log level
- [ ] Do some things in the UI, no HTTP logs
- [ ] Restart server - with Debug log level
- [ ] Do some things in the UI, see the request logs, but they're simpler than before 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/logger.rs
+++ b/server/graphql/logger.rs
@@ -1,0 +1,161 @@
+use async_graphql::{
+    extensions::{
+        Extension, ExtensionContext, ExtensionFactory, NextExecute, NextParseQuery, NextValidation,
+    },
+    parser::types::{ExecutableDocument, Selection},
+    Response, ServerError, ServerResult, ValidationResult, Variables,
+};
+use chrono::{DateTime, Utc};
+use rand::Rng;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+pub struct GraphQLRequestLogger;
+
+impl ExtensionFactory for GraphQLRequestLogger {
+    fn create(&self) -> Arc<dyn Extension> {
+        Arc::new(LoggerExtension {})
+    }
+}
+
+struct LoggerExtension {}
+
+#[derive(Debug, Default)]
+pub struct QueryLogInfo {
+    inner: Arc<Mutex<QueryInfoInner>>,
+}
+impl QueryLogInfo {
+    pub fn new() -> QueryLogInfo {
+        let mut rng = rand::thread_rng();
+        QueryLogInfo {
+            inner: Arc::new(Mutex::new(QueryInfoInner {
+                id: rng.gen_range(100_000_000..=999_999_999),
+                start_time: Utc::now(),
+            })),
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct QueryInfoInner {
+    id: u64,
+    start_time: DateTime<Utc>,
+}
+
+#[async_trait::async_trait]
+impl Extension for LoggerExtension {
+    // Log any invalid queries received
+    async fn validation(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        next: NextValidation<'_>,
+    ) -> Result<ValidationResult, Vec<ServerError>> {
+        let res = next.run(ctx).await;
+
+        match res {
+            Ok(_) => res,
+            Err(ref errors) => match ctx.data_opt::<QueryLogInfo>() {
+                Some(info) => {
+                    let info = info.inner.lock().await;
+                    for e in errors {
+                        log::info!(
+                            target: "gql_logger",
+                            "[QueryID: {}] Received invalid query: {}",
+                            info.id,
+                            e.message
+                        )
+                    }
+                    res
+                }
+                None => res,
+            },
+        }
+    }
+
+    // Log the received request
+    async fn parse_query(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        query: &str,
+        variables: &Variables,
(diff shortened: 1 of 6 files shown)


server/server/Cargo.toml
```diff
@@ -25,7 +25,7 @@ util = { path = "../util" }
 machine-uid = { version = "0.5.1", optional = true }
 
 actix-cors = "0.7.0"
-actix-web = { version = "4.7.0", features = ["rustls-0_23"] }
+actix-web = { version = "4.11.0", features = ["rustls-0_23"] }
```
lache-melvin, line 28: Log level only available in v4.11

server/server/src/middleware/mod.rs
```diff
@@ -1,12 +1,16 @@
...
 
 pub fn compress() -> actix_web::middleware::Compress {
     actix_web::middleware::Compress::default()
 }
 
 pub fn logger() -> actix_web::middleware::Logger {
-    actix_web::middleware::Logger::default()
+    actix_web::middleware::Logger::new(r#""%r" FROM: "%{Referer}i" STATUS: %s"#)
+        .log_level(Level::Debug)
+        .log_target("request_log")
```
lache-melvin, line 13: Supposedly this would replace:

`2025-07-09 16:57:23.039229000 [DEBUG] actix_web::middleware::logger:"POST...`

with 

`2025-07-09 16:57:23.039229000 [DEBUG] request_log:"POST...`

It doesn't do it in the in debug build, didn't build a release target, figured could check after a nightly build?

jmbrunskill:
> I'm wondering about something [like this](https://docs.rs/actix-web/latest/actix_web/middleware/struct.Logger.html#method.custom_response_replace) - and we leave as info level, but only log non-graphql requests?

Makes sense to me.

> I think with https://github.com/msupply-foundation/open-msupply/issues/7243 - we'd get more value by logging in each query/mutation, the name of the query/mutation (+ maybe params? Just not login 😁), and then the result before returning.

Also keen for this. Although I wonder if we're consistent with logging inside our services if we really need this. Query/Muation name would be much much better than current state :)

lache-melvin:
> Issue #8137 doesn't seem right?

Ah yeah sorry, was just that it came up when adding logs for that issue, that the HTTP ones are noisy and unhelpful 😁

jmbrunskill (approved):
Nice thanks Lache, can refine as we work on the other related logging issues.

lache-melvin:
@jmbrunskill got excited and wanted to try... what do you think?

Usual info level, logs incoming requests (see password masked) only log error responses:
Screenshot 2025-07-10 at 3 20 52 PM

Debug shows responses:
Screenshot 2025-07-10 at 3 20 11 PM

server/server/src/middleware/mod.rs
```diff
@@ -1,12 +1,17 @@
...
 pub fn compress() -> actix_web::middleware::Compress {
     actix_web::middleware::Compress::default()
 }
 
 pub fn logger() -> actix_web::middleware::Logger {
-    actix_web::middleware::Logger::default()
+    actix_web::middleware::Logger::new(r#""%r" FROM: "%{Referer}i" STATUS: %s"#)
+        .log_level(Level::Info)
+        .log_target("request_log")
+        .exclude("/graphql")
```
lache-melvin, line 14: excluding completely... maybe we still want to show the request metadata though? like you say, might become clear as we work more on logging, get clearer on requirements...

server/graphql/logger.rs
```diff
@@ -0,0 +1,161 @@
...
+    Response, ServerError, ServerResult, ValidationResult, Variables,
+};
+use chrono::{DateTime, Utc};
+use rand::Rng;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+pub struct GraphQLRequestLogger;
+
+impl ExtensionFactory for GraphQLRequestLogger {
```
lache-melvin, line 15: Taken mostly from here https://github.com/kurotych/async-graphql-logger - but a bit simpler

jmbrunskill:
Absolutely love it, thanks Lache. Only concern is if it creates a performance problem on android or something?

lache-melvin:
Nice shout - I'll try on a tablet today and double check 🙏

lache-melvin:
Tried on android, no observable difference for me... though I think worth keeping tabs on in wider review of android performance?

PR #8424, opened 2025-07-10
8403 donor not visible

### LINKED ISSUE #8403: DonorNotVisible Error when inserting stockline
## What went wrong? 😲

Image

## Expected behaviour 🤔

If the UI shouldn't allows you to select a non-visible donor, these should be filtered out.
If this error comes because of a transfer, not sure what the correct behaviour should be. Should we remove any non-visible donors from stocklines?

## How to Reproduce 🔨

TBD

## Your environment 🌱

- Open mSupply Version: 2.8.3

- Platform:

- Database type: SQLite

jmbrunskill:
Fixes #8403

# 👩🏻‍💻 What does this PR do?

Allows newly added stocklines to go inserted with invisible donors.
Very edge case situation but is affecting a client with migrations so was considered worth fixing for this usecase.

I'll raise a follow up issue to what should happen when donors are made invisible.

I haven't changed the `update.rs` logic in the same way. The justification for now is that its a nullable update, so it should only be set if they change the donor and then they should only see valid donors? But maybe I should change it as well to be consistent?

# 🧪 Testing

See bug issue for testing steps

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/invoice_line/stock_in_line/insert/mod.rs
+++ b/server/service/src/invoice_line/stock_in_line/insert/mod.rs
@@ -134,10 +134,9 @@ mod test {
         barcode::{BarcodeFilter, BarcodeRepository},
         mock::{
             mock_customer_return_a, mock_customer_return_a_invoice_line_a, mock_inbound_shipment_a,
-            mock_inbound_shipment_c, mock_inbound_shipment_e, mock_item_a, mock_name_b,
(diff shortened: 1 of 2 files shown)


jmbrunskill:
Just fixing tests...

PR #8433, opened 2025-07-10
Add no_equal_to_or_null filter option & use in central server sync

### LINKED ISSUE #8432: omSupply Central Server doesn't sync vaccination_course, dose, or item to Legacy
## What went wrong? 😲

When upgrading to 2.9.0 from 2.8.1, open-mSupply Central server should sync `vaccination_course`, `vaccination_dose` and `vaccination_item` to Legacy.

It also doesn't sync `vaccination` records created on 

## Expected behaviour 🤔

We're now going to be supporting using Central Server as a oms Store, so we should also support it for using recording vaccinations.
We at least need to sync any vaccination setup records to Legacy which will be on OMS Central

## How to Reproduce 🔨

1. Initialise or restore a central OMS server on version 2.8.X
2. Setup vaccination courses
3. Upgrade to 2.9
4. Check that`vaccination_course`, `vaccination_dose` and `vaccination_item`  sync to Legacy mSupply

## Your environment 🌱

- Open mSupply Version: 2.9.0-RC

- Platform:

- Database type:

jmbrunskill:
Fixes #8432 

# 👩🏻‍💻 What does this PR do?

Adds a new filter option to allow not_equal_to_or_null
The problem with recently changes to allow central server to sync on behalf of remote sites was that it's own changes were being ignored as they were set to null.

Basically when we say not equal to something in SQL, null records are excluded.

`SELECT * FROM changelog WHERE source_site_id <> 1 ` 
image

`SELECT * FROM changelog WHERE source_site_id <> 1 or source_site_id is null`
image

I think we don't need to worry about effects on non-central sites as it doesn't use the same filter now.
However, could it cause an issue with Central server?
What about the edge case of first sync after upgrading. Does it know that it's a central server fast enough?

Rather than adding this to equal filter should we create something special for changelog only?

# 🧪 Testing

- [ ]  Try creating records on OMS Central, make sure they sync to Legacy mSupply (especially the records mentioned in the issue)
- [ ] Try to create some kind of sync loop or issue with syncing.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/core/src/generic_filters.rs
+++ b/server/graphql/core/src/generic_filters.rs
@@ -60,6 +60,7 @@ where
             not_equal_to,
             equal_any_or_null,
             not_equal_all: None,
+            not_equal_to_or_null: None,
             is_null: None,
         }
     }
--- a/server/graphql/core/src/lib.rs
+++ b/server/graphql/core/src/lib.rs
@@ -144,6 +144,7 @@ macro_rules! map_filter {
             equal_any: $from
                 .equal_any
                 .map(|inputs| inputs.into_iter().map($f).collect()),
+            not_equal_to_or_null: None,
             equal_any_or_null: None,
             not_equal_all: None,
             is_null: None,
--- a/server/graphql/general/src/queries/inventory_adjustment_reason.rs
+++ b/server/graphql/general/src/queries/inventory_adjustment_reason.rs
@@ -121,6 +121,7 @@ fn map_inventory_adjustment_reason_filter(
                             }),
                             equal_any_or_null: None,
                             not_equal_all: None,
+                            not_equal_to_or_null: None,
                             is_null: None,
                         }
                     }),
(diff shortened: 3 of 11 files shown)


jmbrunskill:
> Thanks James, looks great, this is the same gotcha as that disabled store filter.
> 
> I guess we already found this out in other area like the ` .is_sync_update(EqualFilter::equal_or_null_bool(false)),`

Yup exactly. Should have been caught in code review and/or dev testing. Cost a bunch of time in both Jonna and QA team so it's definitely something we (e.g. me) can improve on! I think the diesel context obfuscates some of these kinds of things, in pure SQL I feel like I might have caught it. But either way I should have tested properly both remote and central stores syncing.

PR #8437, opened 2025-07-11
Remove disabled option from report selector

### LINKED ISSUE #8431: Unable to print labels when prescription is verified
## What went wrong? 😲

Should be able to still print a prescription label of a verified/finalised prescription but the option is grayed out right now. 

Please turn the option on.

** Update **
The main print button has been enabled, but the row/table action is disabled

Image

## Expected behaviour 🤔
Both print options should be enabled for Verified prescriptions

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

jmbrunskill:
Fixes #8431

# 👩🏻‍💻 What does this PR do?

Removes the option to disable the report selector, as we generally always want to allow report or printing if the use can view.

Is there some reason why we do want to disable this? Examples seemed to be ok to me...

# 🧪 Testing

- [ ] Check reports/printing isn't disabled in prescriptions when verified.
- [ ] Check we haven't broken any reports or other functionality

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
+++ b/client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
@@ -50,7 +50,6 @@ export const RepackModal: FC<RepackModalControlProps> = ({
   const { Modal } = useDialog({ isOpen, onClose });
 
   const [invoiceId, setInvoiceId] = useState<string | undefined>(undefined);
-  const [reportDisabled, setReportDisabled] = useState<boolean>(false);
   const [isNew, setIsNew] = useState<boolean>(false);
 
   const { data: logData } = useActivityLog(stockLine?.id ?? '');
@@ -71,7 +70,6 @@ export const RepackModal: FC<RepackModalControlProps> = ({
 
   const onRowClick = (rowData: RepackFragment) => {
     setInvoiceId(rowData.id);
-    setReportDisabled(false);
     setIsNew(false);
   };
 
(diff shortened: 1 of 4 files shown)


client/packages/system/src/Stock/Components/Repack/RepackModal.tsx
```diff
@@ -50,7 +50,6 @@ export const RepackModal: FC<RepackModalControlProps> = ({
   const { Modal } = useDialog({ isOpen, onClose });
 
   const [invoiceId, setInvoiceId] = useState<string | undefined>(undefined);
-  const [reportDisabled, setReportDisabled] = useState<boolean>(false);
```
jmbrunskill, line 53: From what I can see this was always set to false?

lache-melvin (changes requested):
thanks James, will need to handle the repack case 🫠 🚀

lache-melvin:
@jmbrunskill this one still on your radar? If needed I'm happy to pick up 🙏

jmbrunskill:
@lache-melvin does this change cover the use case you were referring too? I can't actually seem so to figure out how to reliably replicate. I've tried delaying the requests, which I managed to trigger once. Since this change I can't seem to see a disabled button, so maybe I'm doing something wrong in my testing?

jmbrunskill:
Also snuck in change after discussing with @mark-prins  to default to label printing for Prescriptions.
Most people aren't using the prescription receipt so will be less annoying for people to have to change for now.

jmbrunskill:
Ah right I was always selecting a line then printing... Makes sense noW!

PR #8456, opened 2025-07-13
8345 make the confirm finalise button in requisitions smaller   users are complaining clicking it and making errors

### LINKED ISSUE #8345: Make the Confirm Finalise button in requisitions smaller - Users are complaining clicking it and making errors
## What went wrong? 😲

Receiving more user complaints about the "Confirm finalise" button - everyone keeps clicking it instead of create outbound shipment and it is messing up their orders...

Image

Suggest make that button into a checkbox and make it smaller like OG msupply so it does not keep getting clicked... way too much user errors happening and complaints being given.

Image

reference - Demo screenshots 

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #8345

# 👩🏻‍💻 What does this PR do?
- Move the `Create Shipment` button to be within split buttons
- Users will have to manually click arrow and click finalise button if they want to finalise
- Allow different icons for split buttons

https://github.com/user-attachments/assets/4b951d96-0daa-40f6-a1b7-58775651440e

# 🧪 Testing

- [ ] Make sure the `Create Shipment` and `Finalise` buttons in Requisitions still works as it's supposed to
- [ ] To finalise the user will have to click the arrow button and change the button to 'Finalise'

# 📃 Documentation

  1. Update the behaviour of the buttons in Requisition documentation
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/buttons/standard/SplitButton.tsx
+++ b/client/packages/common/src/ui/components/buttons/standard/SplitButton.tsx
@@ -16,6 +16,7 @@ export interface SplitButtonOption<T> {
   label: string;
   value?: T;
   isDisabled?: boolean;
+  Icon?: ButtonWithIconProps['Icon'];
 }
 
 export interface SplitButtonProps<T> {
@@ -82,6 +83,7 @@ export const SplitButton = <T,>({
     },
     label: buttonLabel,
   };
+  const icon = selectedOption.Icon || Icon;
 
   return (
     <>
@@ -98,11 +100,11 @@ export const SplitButton = <T,>({
           {isLoadingType ? (
             <LoadingButton
               isLoading={isLoading}
-              startIcon={Icon}
+              startIcon={icon}
               {...sharedButtonProps}
             />
           ) : (
-            <ButtonWithIcon Icon={Icon} {...sharedButtonProps} />
+            <ButtonWithIcon Icon={icon} {...sharedButtonProps} />
           )}
 
           <ShrinkableBaseButton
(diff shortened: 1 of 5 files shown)


lache-melvin (changes requested):
Oo niiice yah def makes more sense given our other flows 🤩 

There was one more requirement in the issue, should default to confirm finalised after supplying:

```if status is New AND there are some lines which have supply as non-zero AND all lines have remaining as zero then we show Confirm finalised rather than Create shipment```

Also, big diff is making me wonder if it could be simpler 👀  There's really only one decision (the check above) to show create shipment or confirm verified - once verified we don't show the button... maybe easier to write like that, then trying to stick w the array, having to force and then filter 'create-shipment' as a status everywhere?

lache-melvin:
oh lol had to pull again

roxy-dao:
> Oo niiice yah def makes more sense given our other flows 🤩
> 
> There was one more requirement in the issue, should default to confirm finalised after supplying:
> 
> `if status is New AND there are some lines which have supply as non-zero AND all lines have remaining as zero then we show Confirm finalised rather than Create shipment`
> 
> Also, big diff is making me wonder if it could be simpler 👀 There's really only one decision (the check above) to show create shipment or confirm verified - once verified we don't show the button... maybe easier to write like that, then trying to stick w the array, having to force and then filter 'create-shipment' as a status everywhere?

We don't show the button once finalised though...

lache-melvin (approved):
Looks great, thanks Roxy 🙏 

Def feels like more logic that should be extracted and tested tho... sad the component hasn't been set up in a very testable way, would need a bigger refactor i think 🫠

PR #8463, opened 2025-07-14
8131 stocktake list flickers on load from ascending to descending order

### LINKED ISSUE #8131: Stocktake list flickers on load from ascending to descending order
## What went wrong? 😲

The stocktake list flickers from ascending to descending order of "created" when you click on the button in nav.

https://github.com/user-attachments/assets/77beb114-5f49-461f-9382-454ef70e104b

## Suggested solution

We are making 2 identical requests on navigation to the page, bar the "desc" variable being false in the first and true in the second.
Image

Image

We should agree on what the default direction is (I think desc) and then only request that once.

I assume refreshing the page also makes 2 requests, but the full app render is too slow for you to see the order flickering.

## Expected behaviour 🤔

No flickering just load correct order straight away

## How to Reproduce 🔨

1. Just click the stocktakes button in navigation and watch it show ascending order for a moment before rerendering descending order (with the order arrow in column header animating as such)

## Your environment 🌱

- Open mSupply Version: dev 2.9 I assume prod tho
- Legacy mSupply Central Server Version: irrelevant

- Platform:irrelevant

- Database type:irrelevant

GeronimoJohn:
Fixes #8131 

# 👩🏻‍💻 What does this PR do?

This fixes the double API calls and flickering by ensuring the correct searchParams are initialized proactively rather than patched reactively. We now set initialSort directly to `useSearchParams` within `useUrlQuery` the source of 'params truth' for our codebase.

Might need to test other places too, I might have missed out a specific use case :D

# 🧪 Testing

- [ ] Navigate to any list where we define initialSort for (e.g. Outbound Shipment, Stocktakes)
- [ ] Click on navigation button within the navigation drawer for the page that you're in
- [ ] It shouldn't cause any flickers
- [ ] Check the network tab
- [ ] There should only be _**ONE**_ api call on refresh
- [ ] And since the variables have not change -> clicking again on navigation button should _**NOT**_ call another api call
- [ ] Should only call api call when filters are added or params changed

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
+++ b/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
@@ -54,7 +54,7 @@ export const useUrlQueryParams = ({
   const initialParams: URLSearchParamsInit = initialSort
     ? {
         sort: initialSort?.key ?? '',
-        dir: initialSort?.dir ?? 'asc',
+        dir: initialSort?.dir === 'desc' ? 'desc' : '',
       }
     : {};
 


CarlosNZ (approved):
Looks great @GeronimoJohn -- excellent troubleshooting, always great to remove another `useEffect`. 

Have tested it in several places, I can't see any problems 👍

PR #8498, opened 2025-07-16
Update BasicSpinner.tsx

### LINKED ISSUE #2758: App is unresponsive while data is loading
## What went wrong? 😲

UPDATE: it's any query! Affecting any page! Can't interact with app while loading

While data for coldchain monitoring graph is rendering, the date range of the graph can't be changed.

This is a little bit annoying if you select a wide range by mistake, and need to wait for it to load.

This might be a low priority issue, but could be very annoying if there are users with very low performance machines which take a long time to load.

This is partially related to issue #2757 because it might solve it getting stuck. 

https://github.com/msupply-foundation/open-msupply/assets/90807420/278d98b1-2799-4cb6-9d6c-0df3ee71f59d

## Expected behaviour 🤔
Date range can be edited to reset query even in loading state

## How to Reproduce 🔨

Go to coldchain -> monitoring in omSupply
Select a date range
Attempt to change the selection while the graph is rendering

fergie-nz:
Fixes #2758

After some investigation, found that our basic spinner by default renders in a fixed position over the top of our other components.
This means whenever basic spinner is rendered - everything in the main part of the app is unresponsive as it is behind the spinner overlay.

The AppDrawer wasn't affected and was still able to be interacted with as it was a sibling of where the spinner was rendered and therefore not overlayed. At least this is my understanding of what was happening.

This is fixed by setting it inline (and not in a fixed position).

# 👩🏻‍💻 What does this PR do?

See current behaviour where when loading - main DOM area can't be interacted with:

https://github.com/user-attachments/assets/dc3035fc-39ba-482d-bb7e-b46cf0650c86

Then after changes - can be interrupted:

https://github.com/user-attachments/assets/0f334e57-64f5-4f37-9e02-d8ebc7c7568c

Question for reviewer - is there any case where not having the spinner in a fixed position, will it break the app? ie spinner is off centre or positioned weirdly? 

Secondly - are there other loading cases or suspended states where we create a similar overlay? I am thinking that these things aren't typically noticeable in a dev environment and would only show up with slow internet.

# 🧪 Testing

- Use a part of the app which has a slow loading time. Cold chain chart is a good option because it has a slow enough load time with lots of data and when throttling with dev tools
- Change the date - and try to interact with the main part of the app when loading
- See that you can still on this branch (but not on develop)

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/loading/BasicSpinner/BasicSpinner.tsx
+++ b/client/packages/common/src/ui/components/loading/BasicSpinner/BasicSpinner.tsx
@@ -29,7 +29,7 @@ const StyledText = styled(Typography)(({ theme }) => ({
 
 export const BasicSpinner = ({
   messageKey = 'loading',
-  inline = false,
+  inline = true,
   style = {},
 }: BasicSpinnerProps) => {
   const t = useTranslation();


CarlosNZ (approved):
LGTM -- nice fix @fergie-nz 👍

Issue #8415, opened 2025-07-09
Reports excel export and printing take too long (10mins+)


alainsussol:
## What went wrong? 😲
On desktop it is more than 10 minutes to Export. The export finished just after I stopped the recording and I am not sure if it is because I clicked somewhere or if it has genuinely finished. 

It would be good if there is a way to speed that up a little and make sure that the user knows that omSupply is working on it because otherwise you would not know that something is happening

Exporting "Ruptures de stock" on desktop platform ~11 minutes video
https://drive.google.com/file/d/152DhuXQvOs7GCDoZEj1eADX7vWYZL_RX/view?usp=sharing

It was also long on Android  but the interval between the start date and the long date was less and the button export or print stayed highlighted during the process.

Exporting "Ruptures de stock" on Android platform ~2 minutes 
https://drive.google.com/file/d/1ei7zPGFnjb1zbd7TmgRbvVPQndjV_IPm/view?usp=sharing

Printing  "Ruptures de stock" on Android platform ~ 1.5 minutes
https://drive.google.com/file/d/1caW5760rT4aUnVqoRHlZ7fdoUcG7s0sG/view?usp=sharing

## Expected behaviour 🤔

Faster if possible but somehow a message or a progression bar to let the user know that something is happening

## How to Reproduce 🔨

Steps to reproduce the behaviour:
 
1. Go to 'Report > Ruptures de stock (Stockout report)
2. Choose a start date and an end date which are a couple of months apart
3. Click OK and wait
4. You will see nothing happening but if you wait long enough your report will be ready to save as an excel
 
## Your environment 🌱

- Open mSupply Version: 2.9.0-RC-07081818
- Legacy mSupply Central Server Version: 8.3.0

- Platform: Desktop and Android (Tablet)

- Database type: SQLite

Chris-Petty:
Ya not ideal! Maybe 2.9.1? Excel export basically unusable on any significant system for many reports

PR #8500, opened 2025-07-16
8415 reports print and excel export loading spinner

Chris-Petty:
#8415
I refuse to say this "fixes" 8415 - 10 minutes to excel export 4000 rows of data is outrageous. A loading spinner doesn't cut it! The actual react page takes about 12s. Blasting that same data into excel (i.e. XML) should _very_ generously take less than 12s.

I used just the normal `Item Usage` report with 4000 items and it takes forever. For warehouses this would just be the tip of the iceberg! E.g. multiply that several times if talking about reports on 

# 👩🏻‍💻 What does this PR do?

image

Use the LoadingButton. If you click either print or export it will spin both. LoadingButton expands width in loading state which looks bad, however the original suggested solution would just make the buttons disabled (which isn't the correct user feedback)

# 🧪 Testing

- [ ] Item usage report
- [ ] click export or print
- [ ] Both buttons turn to loading spinners regardless of which you pressed
- [ ] Once print or export completes they stop spinning

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/reports/src/DetailView/AppBarButton.tsx
+++ b/client/packages/reports/src/DetailView/AppBarButton.tsx
@@ -6,6 +6,7 @@ import {
   EnvUtils,
   FilterIcon,
   Grid,
+  LoadingButton,
   Platform,
   PrinterIcon,
   SwipeIcon,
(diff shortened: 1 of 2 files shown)


roxy-dao (approved):
Thanks~ we really need to speed up the reports somehow 🤔

PR #8509, opened 2025-07-17
#8505 Add "Enrolment info" button to Encounter

### LINKED ISSUE #8505: Add button to bring up "Program Enrolment" form from Encounter page
This is a requirement for part of the budgeted Djibouti RH improvements.

They will move more data from the Encounter schemas into the Enrolment form (to avoid having to re-enter it each time). To make this more accessible (it's currently pretty buried and you don't really look at it again after initial enrolment), we'll add a button to the Encounter page to launch this as a modal:

Image

This provides a mechanism to quickly look up "standard" patient program information as part of an Encounter.

### Acceptance Criteria ✅

Button as described, but also need to move some Encounter schema fields to Enrolment form (Details TBD)

### Moneyworks Jobcode 🧰

**DJIFPCDV**	Open mSupply Family Planning Development

CarlosNZ:
Fixes #8505

# 👩🏻‍💻 What does this PR do?

Add "Enrolment info" button to Encounter toolbar to make the "common" program-related info for a Patient easier to access during encounter.

Screenshot 2025-07-17 at 12 31 04 PM

See inline comments

# 🧪 Testing

- [ ] Have some "Programs" configured
- [ ] Go to an Encounter (or create a new one)
- [ ] Click on the new "Encounter info" button in the Toolbar
- [ ] Should have the original Program Enrolment modal form appear
- [ ] Make changes and save -- confirm it saves correctly
- [ ] Confirm Enrolment form behaves correctly (loads and saves) when used in normal context (Patient -> Programs tab) (this is because I've made a small change to how the form accesses the Patient ID)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/programs/src/api/operations.graphql
```diff
@@ -269,9 +274,6 @@ query encounterById($storeId: String!, $encounterId: String!) {
       __typename
       nodes {
         ...Encounter
-        programEnrolment {
-          isImmunisationProgram
-        }
```
CarlosNZ, line 274: Don't need these explicitly here, as they're in the "Encounter" fragment definition (above)

client/packages/programs/src/api/operations.graphql
```diff
@@ -227,6 +227,11 @@ fragment Encounter on EncounterNode {
   }
   programEnrolment {
     id
+    isImmunisationProgram
+    document {
+      type
+      name
```
CarlosNZ, line 233: Fetch enrolment doc info as part of the Encounter query to save having to make an additional query

client/packages/system/src/Patient/ProgramEnrolment/Components/DetailModal.tsx
```diff
@@ -48,11 +48,17 @@ const useUpsertProgramEnrolment = (
...
 };
 
-export const ProgramDetailModal: FC = () => {
+export const ProgramDetailModal = ({
+  patientId: inputPatientId,
+}: {
+  patientId?: string;
+}) => {
   const t = useTranslation();
-  const patientId = usePatient.utils.id();
```
CarlosNZ, line 53: Had to pass patientId into component when used here, as it can no longer fetch patientID from URL.

Still uses `usePatient.utils.id()` if it can (below), so I don't have to update the other usage(s) of the Form.

Personally, I think it's a bit dodgy to rely on the presence of a URL parameter in a Modal, as they can easily be used in places other than their original implementation (as is the case here!)

jmbrunskill (approved):
Works great, thanks @CarlosNZ 

I'm getting a bit of different button placement but doesn't seem to critical?
image
I wonder if next the `More` button would be a more reliable positioning?

PR #8516, opened 2025-07-17
8415 improve report excel export performance

### LINKED ISSUE #8415: Reports excel export and printing take too long (10mins+)

Chris-Petty:
Fixes #8415 

# 👩🏻‍💻 What does this PR do?

Before:
100 rows: 0.12s
1000 rows: 9.5s
10000 rows: 1123s 
flamegraph_before

It appears the vast majority of time is `Worksheet::insert_new_row()`. Perhaps quadratic if not exponential big O as for every row added I am guessing it executes `rebuild_map_and_indicies`, as the sheet grows this gets slower. Oof. Soooo I removed it:

After: 10000 rows 1.6s
flamegraph

(note our `apply_data_rows` is now a very minor player, and most of the time is in `scraper` code parsing the html rather than `umya` building the spreadsheet)

This appears to work correctly for item usage report which just does default direct mapping of html to excel, however I expect the current change to break footers if using a xlsx template. The intent with the insert rows is to push the footer rows of the template down. 

Turns out you can just `insert_new_row` with second parameter being how many rows to add, so you only really need to call it once. So now:

10000 rows: 1.6s (crudely timed maybe 50ms slower than without adding rows at all)
flamegraph

Virtually identical to not adding rows at all, but respects the excel template footer by pushing it down.

So ya like flamegraphs? I spent the morning figuring out how to get a cli tool working on a particular test so you don't have to:

```sh
cargo install flamegraph
```
https://github.com/flamegraph-rs/flamegraph

```sh
cargo flamegraph --dev --unit-test service -- report::convert_to_excel::report_to_excel_test::test_generate_excel_perf
```

(Note, I'm using --dev cause otherwise it compiles release which takes forever. If you were flamegraphing the actual OMS server while running you might benefit from running release binary.)

# 🧪 Testing

- [ ] OMS site with 4000+ items
- [ ] item usage report takes <1min (definitely not 5mins+)

# 📃 Documentation

  1.
  3.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [x] Postgres
- [x] SQLite
- [x] Frontend

### DIFF
--- a/server/service/src/report/convert_to_excel.rs
+++ b/server/service/src/report/convert_to_excel.rs
@@ -327,6 +327,8 @@ fn apply_known_styles(cell: &mut Cell, cell_type: &str) {
 
 #[cfg(test)]
 mod report_to_excel_test {
+    use std::time::Duration;
+
     use super::*;
     use scraper::Html;
 
@@ -554,8 +556,8 @@ mod report_to_excel_test {
         );
     }
 
-    #[test]
-    fn test_generate_excel_perf() {
+    #[tokio::test]
+    async fn test_generate_excel_perf() {
         // We want to ensure that excel export takes a sensible amount of time.
         // Many reports may have several thousand rows with dozens of columns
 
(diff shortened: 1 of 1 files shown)


lache-melvin:
LEGENDARY RESEARCH SIR

Chris-Petty:
@lache-melvin it is ready! If the comments didn't give it away, the new test for the excel export with template was 90% vibe coded and then fixed haha. The tokio timeout is artisanally crafted on the other hand.

lache-melvin (approved):
- ✅ so fast
- ✅ footer still there
- ✅ has tests

party town

PR #8542, opened 2025-07-18
Fix vaccine card returns that were over-ridding stock and packsize

### LINKED ISSUE #8530: When using the vaccination program, reverting a vaccine that has been given change the packsize of the corresponding stockline to 1 without converting the number packs to match the new packsize
## What went wrong? 😲

Context

I have a vaccine where each unit is one dose. 
https://drive.google.com/file/d/12aYlLJaKOv5BJCu_bCkgzFhmt5qLFs_e/view?usp=sharing

And the Store preference has the "Set pack to one to all visible items in this store"
https://drive.google.com/file/d/1nUanD00c2PbxPcAxtc1-rUsny_TfCSrR/view?usp=sharing

**Video**
https://drive.google.com/file/d/1mOB8FzFG6zLQzxlXmrFbhxdu9gKctHhO/view?usp=sharing

You can see in the example in the video that I have a stock line of 35 packs with a packsize of 20 for a total of units of 700 units/doses

If you then create an encounter for a patient and give a vaccine from this stock line. Everything is fine, omSupply substrack 0.05 from a pack which is the equivalent of one dose on a pack of 20. The numbers are ok. 

If you mark the encounter as visited then go back to the vaccination encounter and set the vaccination has not given and toggle the button to revert back the stock transaction. Once the reverting is done if you go back to stock line, the packsize of the stockline had becomes 1 but my number of packs still stayed the same so instead of having 34.95 packs of 20 units for 699 doses, I have 34.95 packs of 1 for 34.95 doses.
 

## Expected behaviour 🤔
When reverting a transaction, the current packsize of the stock_line involved needs to stay the same.   

## How to Reproduce 🔨
See in description 

## Your environment 🌱

- Open mSupply Version: v2.7.3 but still replicable in v2.9.0-RC-07171624
- Legacy mSupply Central Server Version: v8.03

- Platform: omsupply desktop 

- Database type: SQLite

jmbrunskill:
Fixes #8530

# 👩🏻‍💻 What does this PR do?

Previously there was an exception for packsize to 1 preference were it wasn't applied using the add stock mechanism.
This what's applied to customer returns though, so even though the invoice said to return a single dose to the line, it actually over-wrote the packsize breaking the ledger..

Now we're using a canceled prescription to reverse the existing prescription instead of customer returns for vaccinations

I've left in a change to the customer return with only over-writes customer returns when there's no existing stockline.
That is an extra db request though, so maybe better remove it?

# 🧪 Testing

- [ ] Turn on packsize to 1 pref
- [ ] Add vaccine stock to the store with packsize > 1
- [ ] Create a vaccine encounter for a patient and give them the stock line created above
- [ ] (one dose should now have been removed from available stock)
- [ ] Update encounter to say the vaccine wasn't given and select the revert stock transation option
- [ ] Check that the single dose gets returned back to the stockline.
- [ ] Check that stock is update correctly with normal customer returns and inbound shipments.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/invoice_line/stock_in_line/insert/generate.rs
+++ b/server/service/src/invoice_line/stock_in_line/insert/generate.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use repository::{
     vvm_status::vvm_status_log_row::VVMStatusLogRow, BarcodeRow, InvoiceLineRow, InvoiceLineType,
-    InvoiceRow, ItemRow, RepositoryError, StockLineRow, StorageConnection,
+    InvoiceRow, ItemRow, RepositoryError, StockLineRow, StockLineRowRepository, StorageConnection,
 };
 
 use super::InsertStockInLine;
@@ -36,11 +36,20 @@ pub fn generate(
 
     let mut new_line = generate_line(input.clone(), item_row, existing_invoice_row.clone()); // include vvm status here
 
+    // Check if the stock line already exists, if it does we may need to update it rather than replacing it
+    let old_stock_line = match &input.stock_line_id {
+        Some(stock_line_id) => {
+            StockLineRowRepository::new(&connection).find_one_by_id(stock_line_id)?
+        }
+        None => None,
+    };
+
     let should_overwrite_stock_levels = match &input.r#type {
         // Even though we're `inserting` here, if a stock line already exists, we want to add to the existing quantity rather than replace it.
-        // for inventory adjustments and customer returns, the invoice only records the new additional stock to add to the stock line
+        // for inventory adjustments and customer returns
         StockInType::InventoryAddition => false,
-        StockInType::CustomerReturn => false,
+        // For customer returns, we only want to overwrite stock levels if the stock line does't already exist
+        StockInType::CustomerReturn => old_stock_line.is_none(),
         // For inbound shipments, we always create a new stock line, never update an existing one, so shouldn't overwrite stock levels based on the invoice when adding stock
         StockInType::InboundShipment => true,
     };
--- a/server/service/src/invoice_line/stock_in_line/insert/generate.rs
+++ b/server/service/src/invoice_line/stock_in_line/insert/generate.rs
@@ -50,7 +50,7 @@ pub fn generate(
         StockInType::InventoryAddition => false,
         // For customer returns, we only want to overwrite stock levels if the stock line does't already exist
         StockInType::CustomerReturn => old_stock_line.is_none(),
-        // For inbound shipments, we always create a new stock line, never update an existing one, so shouldn't overwrite stock levels based on the invoice when adding stock
+        // For inbound shipments, we always create a new stock line, never update an existing one, so should overwrite stock levels based on the invoice when adding stock
         StockInType::InboundShipment => true,
     };
 
--- a/server/service/src/invoice_line/stock_in_line/insert/generate.rs
+++ b/server/service/src/invoice_line/stock_in_line/insert/generate.rs
@@ -36,7 +36,16 @@ pub fn generate(
 
     let mut new_line = generate_line(input.clone(), item_row, existing_invoice_row.clone()); // include vvm status here
 
-    if StockInType::InventoryAddition != input.r#type && store_preferences.pack_to_one {
+    let should_overwrite_stock_levels = match &input.r#type {
+        // Even though we're `inserting` here, if a stock line already exists, we want to add to the existing quantity rather than replace it.
+        // for inventory adjustments and customer returns, the invoice only records the new additional stock to add to the stock line
+        StockInType::InventoryAddition => false,
+        StockInType::CustomerReturn => false,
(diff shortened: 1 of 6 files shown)


lache-melvin:
tests failing @jmbrunskill ?

jmbrunskill:
Thanks @lache-melvin updated tests, however I'm not entirely sure this is what we want...
This is a change in behaviour that now we have both Customer Returns and Inventory Adjustments that bypass the packsize to 1 pref. Is that really what we want? Probably ok for now.

However, I feel like we maybe should force inventory adjustments to align with packsize to 1 pref as well as customer returns, and only skip the packsize to one step on customer returns if they have an existing stockline selected? Doing an `insert` with an existing stockline does feel like an anti pattern, so would be nice to clean that up too. Maybe the cancel prescription mechanism rather than doing a customer return for vaccine card?

BTW: Still waiting to confirm if we can deploy 2.9.1 to fix this so won't merge yet anyway.

lache-melvin:
> Maybe the cancel prescription mechanism rather than doing a customer return for vaccine card?

I actually really like this idea - do you have a sense if it would be much work to replace now? I didn't love the customer return solution at the time, now we have cancelling prescriptions, that makes way more sense!

> only skip the packsize to one step on customer returns if they have an existing stockline selected? 

I like this as a next best option - introducing new lines should respect pack to one, but let's keep any existing stock lines aligned

lache-melvin:
Happy to approve this as the "it works" option for patch, if capacity for the better solution is low. However would prefer the cancelling prescription, and leave customer returns respecting pack to 1 🙏

andreievg:
> and leave customer returns respecting pack to 1

Yep agree with Laché either way this will be kept and needs a test please.

> Maybe the cancel prescription mechanism rather than doing a customer return for vaccine card?

Yes i think we should do it now, sounds like pretty easy to do ? @lache-melvin is it ok to assume that only one batch/line will existing on prescriptions generated by vaccination ?

https://github.com/msupply-foundation/open-msupply/blob/9e983fb13107b9158dd6351eb64a4d9facdaaee2/server/service/src/vaccination/update/generate.rs#L288

> This is a change in behaviour that now we have both Customer Returns and Inventory Adjustments that bypass the packsize to 1 pref. Is that really what we want? Probably ok for now.

That sounds ok, although technically speaking Customer Return should never have stock_line_id input (always generates new batch, apart from this current edge case, I think that would be restricted in the input at single entry point for stock movements, when we improve it).

> However, I feel like we maybe should force inventory adjustments to align with packsize to 1 pref as well as customer returns, and only skip the packsize to one step on customer returns if they have an existing stockline selected? Doing an insert with an existing stockline does feel like an anti pattern, so would be nice to clean that up too. Maybe the cancel prescription mechanism rather than doing a customer return for vaccine card?

There are so many use cases to consider with pack size to one:

* If stock line exists with non 1 pack size and we turn the pref on then we should not adjust existing stock line
* If there is inbound shipment or customer return in progress with non 1 pack size, and then we switch to pack size to 1, we should only adjust if stock line is not already used (let user know this straight away ?)

I think it's worth writing up a all the use cases when we do the single point of entry for all stock movement.

andreievg (changes requested):
As per comment above:

Tests for existing changes.

Use cancel prescription and updates tests for that, thanks.

lache-melvin:
> Yes i think we should do it now, sounds like pretty easy to do ? @lache-melvin is it ok to assume that only one batch/line will existing on prescriptions generated by vaccination ?

Yep that's how it is for now 🙏  If we can't just use "cancel prescription", which should handle all lines by default, then I think would be good to at least flag in a comment near the vaccination creation - so if we add multi-line in the future, we know to fix the reversing as well

server/service/src/invoice_line/stock_in_line/insert/generate.rs
```diff
@@ -36,7 +36,25 @@ pub fn generate(
...
+        None => None,
+    };
+
+    let should_overwrite_stock_levels = match &input.r#type {
+        // Even though we're `inserting` here, if a stock line already exists, we want to add to the existing quantity rather than replace it.
+        // for inventory adjustments and customer returns
+        StockInType::InventoryAddition => false,
+        // For customer returns, we only want to overwrite stock levels if the stock line does't already exist
+        StockInType::CustomerReturn => old_stock_line.is_none(),
+        // For inbound shipments, we always create a new stock line, never update an existing one, so shouldn't overwrite stock levels based on the invoice when adding stock
```
lache-melvin, line 53:
```suggestion
        // For inbound shipments, we always create a new stock line, never update an existing one, so should overwrite stock levels based on the invoice when adding stock
```

lache-melvin (approved):
Awesome thanks James, this is a way nicer solution 🙏 

Did make me think about a carry over, when you "give" the vaccine it shows linked prescription at the top. Maybe we should keep that there, link to the cancelled prescription, rather than it seemingly disappearing...

PR #8628, opened 2025-07-25
8593 move edit into store popper & move packaging level from gaps initialisation

### LINKED ISSUE #8593: Move `edit` into store popper & move packaging level from GAPS initialisation
- [ ] Move `Packaging level` out from GAPS properties and into `General properties` to be shown all the time
- [ ] The `Edit` button next to store in footer should be available always regardless of store level
- [ ] Change item variant packaging to be cubic meters (m3) instead of litres

### Acceptance Criteria ✅
- [ ] Users should not be able to see the `Edit` button next to other stores if they are not on central server
- [ ] Users should always be able to click the `Edit` button next to their store even if they are not a central user
- [ ] Users on central server with central edit permission should be able to edit store preferences
- [ ] Users on remote servers should not be able to edit store preferences
- [ ] Users can edit `Packaging level` regardless of server
- [ ] When the user edits the `Packaging level`, the new value should save

### Moneyworks Jobcode 🧰
OMS:NGUCF

roxy-dao:
Fixes #8593

# 👩🏻‍💻 What does this PR do?
- [ ] Migration for packaging
- [ ] Remove check around `Edit` button
- [ ] The other request to have the UI changed for this will be moved to another issue since it needs to take into account both multi-stores or single stores in the design
- [ ] Changed the pref query endpoint permission so that remote stores can view even if they don't have edit permissions
- [ ] Disabled preference editing if user doesn't have permission

# 🧪 Testing

- [ ] Initialise a fresh database
- [ ] Edit button next to store should always be available
- [ ] Changing languages to french allows you to see the packaging level in french
- [ ] If you do not have central edit permission or are not central server then you cannot edit the preferences

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


aimee-mcneil-melville (approved):
Looking great Roxy!  Good to go 🚀 

Is the follow up issue created for the UI design?

client/packages/system/src/Name/ListView/Stores/StoreProperties.tsx
```diff
@@ -125,3 +129,17 @@ const Row = ({
...
+
+const translatePropertyLabels = (
+  t: TypedTFunction<LocaleKey>,
+  label: string
+) => {
+  // For now, the rest of the properties have to be initialised, and are saved
+  // in french so have not included translations yet
+  switch (label) {
+    case 'Packaging Level':
+      return t('label.packaging-level');
```
lache-melvin, line 141: Yo just popping in, was a bit confused what you were meaning in stand up about only translating some of the properites, but I think this makes sense in that it should always ship with OMS, rather than only if GAPS is initialised...

What if we just store `label.packaging-level` in the DB, and call `t()` on it here? `t` would return the base label if no translation exists (for the existing property keys) but that way we can see from data in the DB, its a translation string, rather than looking like english but showing in french?

roxy-dao: That would work too... but the other GAPS properties will also need to be changed..

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1362,7 +1363,7 @@
   "label.visit-start": "Start",
   "label.volume-per-outer-pack": "Volume per outer pack",
   "label.volume-per-pack": "Volume per pack",
-  "label.volume-per-unit": "Volume per unit (L)",
+  "label.volume-per-unit": "Volume per unit (m³)",
```
jmbrunskill, line 1366: How will we transition existing setups? Won't think break all the GAPS datasets?

PR #8692, opened 2025-07-31
Extra localised string for Encounters

CarlosNZ:
From @Rajajajaja 's [requests](https://docs.google.com/document/d/1TfkpMN1STVLOjbNm9MsO7OBDRZq1rvtknZS1TI_0jlw/edit?tab=t.cbq26y41gbmz).

# 👩🏻‍💻 What does this PR do?

Adds additional "Save progress" button for use in Encounters

I think this makes sense, and it can be put back to just plain old "Save" using custom translations if people don't think it should be different to normal "Save".


jmbrunskill (approved):
Looks good.

PR #8695, opened 2025-07-31
Fix shipped_number_or_packs migration

### LINKED ISSUE #8653: Intermittent issues with shipped_number_of_packs on 2.9.1ish upgrades
## What went wrong? 😲

```
Caused by:
0: Error during fragment time migration (2.9.0) (add_shipped_number_of_packs_to_invoice_line)
1: Sql error
ALTER TABLE invoice_line ADD COLUMN shipped
_number_of
_packs DOUBLE PRECISION;
2: UNKNOWN ("column \"shipped_number_of_packs\" of relation \'invoice_line\" already exists
0") '
```

## Expected behaviour 🤔

No error

## How to Reproduce 🔨

Unclear at this stage, but on one server we deleted the column and re-started the service. It appeared that the column had come back and the migration fragment hadn't been recorded in the database.

## Your environment 🌱

- Open mSupply Version: 2.9.1-RC

- Platform: Windows

- Database type: PostgreSQL

jmbrunskill:
Fixes #8653 

# 👩🏻‍💻 What does this PR do?

Splits the migration into two parts (one for sync and one for adding the columns)
It also add serde_default to json parsing shouldn't fail.

This was milestones for 2.10 but since this has broken 2 demo server so far, I'm worried about 2.9.1 going out without it might result in having to re-initialise tablets when upgraded.
I think it should be in 2.9.1

As a follow up issue I think all migration fragments should run in a transaction. So that if part of it fails the other parts get rolled back... 

- [ ] Issue to be created...

Q: I think this would still fail on something like `sentQuantity: ""` as it's not a float. Should we just not error if we can't parse the json?

# 🧪 Testing

- [ ] Install 2.8.X or earlier
- [ ] Have a sync record for an invoice with no `sentQuantity` field
- [ ] Try to update to 2.9.1 in this branch
- [ ] Check that there aren't any migration errors

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/v2_09_00/add_shipped_number_of_packs_to_invoice_line.rs
```diff
@@ -30,29 +10,10 @@ impl MigrationFragment for Migrate {
         sql!(
             connection,
             r#"
-                ALTER TABLE invoice_line ADD COLUMN shipped_number_of_packs DOUBLE PRECISION;
+                ALTER TABLE invoice_line ADD COLUMN shipped_number_of_packs {DOUBLE};
```
andreievg, line 13: Nice, but what was the effect of 'DOUBLE PRECISION' before ?

server/repository/src/migrations/v2_09_00/add_shipped_number_of_packs_to_invoice_line_legacy.rs
```diff
@@ -0,0 +1,52 @@
+use crate::migrations::{
+    templates::add_data_from_sync_buffer::{sync_buffer, SyncAction},
```
andreievg, line 2: We are not meant to be having any dependencies like this in migrations, i guess it would be fine in this case

andreievg (changes requested):
> Q: I think this would still fail on something like sentQuantity: "" as it's not a float. 

It's a 'real' in mSupply so should always be parseable, i have no idea how this error could happen (seems impossible)

> Should we just not error if we can't parse the json?

Yes i think we shouldn't error, treat this like sync (we don't stop sync if failed to deserialize)

> As a follow up issue I think all migration fragments should run in a transaction

I agree, yes please

> Unclear at this stage, but on one server we deleted the column and re-started the service

This sounds like a different issue, because if it was deserialization, fragment will fail to run again, and even with splitting it will fail to run.

One change request, make sure serde doesn't cause migration fail.

I think in general in migrations we should be doing this for any logic things like this.

server/repository/src/migrations/v2_09_00/add_shipped_number_of_packs_to_invoice_line_legacy.rs
roxy-dao: Because..?

andreievg: Blanket rule, to avoid us using 'repository' or 'service' layer which would have current data types that won't match migrations.

This case should be fine i guess because data type in migration/template won't change

server/repository/src/migrations/v2_09_00/add_shipped_number_of_packs_to_invoice_line.rs
jmbrunskill: Seems to work ok in SQLite, but thought better to keep it consistent with current patterns.

PR #8696, opened 2025-07-31
8688 Create scheduler for ledger fix

### LINKED ISSUE #8688: Schedule and trigger for ledger
Epic: https://github.com/msupply-foundation/open-msupply/issues/8632

1. A way to run ledger fixes on a schedule (once a day), without affecting user activity

2. A way to run ledger fix after migration to a version where ledger fix exists

3. A way to trigger ledger fix after initialisation.

A scheduler that runs approximately once a day, in the background. 
Something basic for now, can be extended to be more elaborate in the future (Carry over, being able to set time of day to run etc..).

Basic would be just storing the last date time (key value) when this scheduler ran, triggering the scheduler every X interval and checking if it should run based on last date time. (this and above deals with 1)

Manual trigger should force run it. (should deal with 3, alongside calling trigger after initialisation)

On startup should start the schedule with a small delay (on migration last date time will not be set so will run after migration, this should deal with 2)

andreievg:
Fixes #8688

# 👩🏻‍💻 What does this PR do?

Add a scheduler for ledger fix, to run 5 seconds after startup (after going to this version) and then every day (first check if needs to run is 5 seconds after startup, then every hour, will run max once a day, unless manually triggered, for now this is only manually triggered after initialisation).

If there are errors, should be logged in oms system log and in console/file log, i also added

# 🧪 Testing

On startup after upgrading to this version should log "Performing ledger fix.." after 5 seconds, after that, on restart etc.. would only show this once a day, you can modify to the following to see this in action more frequently (also check key_value_store for date updates)
```
const FIRST_RUN_DELAY: Duration = Duration::from_secs(5);
// This trigger is not to re-run ledger fix but to check if ledger fix is needed again
// Will check against key values store LAST_LEDGER_FIX_RUN and LEDGER_FIX_INTERVAL
const RE_TRIGGER_DELAY: Duration = Duration::from_secs(2); // 1 hour
const LEDGER_FIX_INTERVAL: TimeDelta = TimeDelta::seconds(10);
```

After initialisation "Performing ledger fix.." should be triggered.

You can change date in key value store to see an error logged in console and added to system_log, but ledger fix message should still be called, and date updated

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/db_diesel/system_log_row.rs
+++ b/server/repository/src/db_diesel/system_log_row.rs
@@ -5,6 +5,7 @@ use chrono::NaiveDateTime;
 use diesel::prelude::*;
 use diesel_derive_enum::DbEnum;
 use serde::{Deserialize, Serialize};
+use strum::Display;
 
 table! {
     system_log (id) {
@@ -17,21 +18,29 @@ table! {
     }
 }
 
-#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(DbEnum, Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Display)]
+#[cfg_attr(test, derive(strum::EnumIter))]
 #[DbValueStyle = "SCREAMING_SNAKE_CASE"]
 pub enum SystemLogType {
+    #[default]
     ProcessorError,
+    LedgerFixError,
+    LedgerFix,
 }
 
 impl SystemLogType {
     pub fn is_error(&self) -> bool {
         match self {
             SystemLogType::ProcessorError => true,
+            SystemLogType::LedgerFixError => true,
+            SystemLogType::LedgerFix => false,
         }
     }
 }
 
-#[derive(Clone, Queryable, Insertable, AsChangeset, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(
+    Clone, Queryable, Insertable, AsChangeset, Default, Debug, PartialEq, Serialize, Deserialize,
+)]
 #[diesel(treat_none_as_null = true)]
 #[diesel(table_name = system_log)]
 pub struct SystemLogRow {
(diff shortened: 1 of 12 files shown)


server/repository/src/db_diesel/system_log_row.rs
```diff
@@ -17,21 +18,29 @@ table! {
...
 
-#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(DbEnum, Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Display)]
+#[cfg_attr(test, derive(strum::EnumIter))]
 #[DbValueStyle = "SCREAMING_SNAKE_CASE"]
 pub enum SystemLogType {
+    #[default]
     ProcessorError,
+    LedgerFixError,
+    LedgerFix,
```
andreievg, line 28: LedgerFix would be used when we actually record info about the ledger fix we are doing

server/service/src/activity_log.rs
```diff
@@ -105,6 +107,19 @@ pub fn system_log_entry(
     Ok(())
 }
 
+// Will also log in file/console
+pub fn system_error_log(
```
andreievg, line 111: I thought it was worth adding this

server/repository/src/db_diesel/system_log_row.rs
```diff
@@ -17,21 +18,29 @@ table! {
     }
 }
 
-#[derive(DbEnum, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(DbEnum, Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Display)]
```
andreievg, line 21: strum:Display to test and used when logging error

server/service/src/ledger_fix/ledger_fix_driver.rs
```diff
@@ -0,0 +1,134 @@
...
+#[derive(Clone)]
+pub struct LedgerFixTrigger {
+    sender: Sender<Option<String>>,
+}
+
+const FIRST_RUN_DELAY: Duration = Duration::from_secs(5);
+// This trigger is not to re-run ledger fix but to check if ledger fix is needed again
+// Will check against key values store LAST_LEDGER_FIX_RUN and LEDGER_FIX_INTERVAL
+const RE_TRIGGER_DELAY: Duration = Duration::from_secs(60 * 60); // 1 hour
+const LEDGER_FIX_INTERVAL: TimeDelta = TimeDelta::days(1);
```
Chris-Petty, line 26: Was having some thoughts about trying to run in off hours (say 12am) but particularly smaller sites may turn their desktops or tablets off over night

jmbrunskill (commented):
Did you consider using the existing `scheduled_task_runner`?
Feels like it would be nice to have it as part of that, it's where I'd expect to see it anyway...

andreievg:
@jmbrunskill 

> Did you consider using the existing scheduled_task_runner?
> Feels like it would be nice to have it as part of that, it's where I'd expect to see it anyway...

I didn't see that one, thanks for pointing it out, just had a quick look, scheduled_task_runner doesn't quite have enough 
It needs to run via trigger too (and shouldn't have more then one instance running). 

I think we would eventually consolidate all schedulers, at least to have one helper to run them (there is some similarity between all 3 (sync, ledger fix, email)

andreievg:
> Was having some thoughts about trying to run in off hours (say 12am) but particularly smaller sites may turn their desktops or tablets off over night

It would be nice to have a more granular control over when schedules run (like if sync fails we should retry it more often, but if success maybe we should wait for a bit etc..), there is a carry over to think about scheduler further.

Thanks for review, I'll have to move migrations to 2.9.2, i'll do it in the downstream branch

PR #8711, opened 2025-08-03
8632.1 Adjust historic incoming invoice

### LINKED ISSUE #8632: Ledger Fix
Related issue: https://github.com/msupply-foundation/msupply/issues/16738

## Tasks

- [ ] https://github.com/msupply-foundation/open-msupply/pull/8717
- [ ] https://github.com/msupply-foundation/open-msupply/pull/8716
- [ ] https://github.com/msupply-foundation/open-msupply/pull/8715
- [ ] https://github.com/msupply-foundation/open-msupply/pull/8713
- [ ] https://github.com/msupply-foundation/open-msupply/pull/8712
- [ ] https://github.com/msupply-foundation/open-msupply/pull/8711
- [ ] https://github.com/msupply-foundation/open-msupply/issues/8698
- [x] #8688 
- [ ] flesh out the remainder use cases https://docs.google.com/spreadsheets/d/1JKqZjXJPxK5SIwBKNItPenv1WJPKPk9hcPS2jiaJdr8/edit?gid=1953147700#gid=1953147700 @andreievg 
- [ ] Internal docs

## Carry Over

KeyValue store to have NaiveDateTime column, to be used in scheduler.

Scheduler to be more advanced, and easily re-useable (advanced meaning have the ability to run at certain time etc..)

One optimisation to think about, ledger has been fixed for a line at a date and we know that, how do we avoid looking at it again (diss allow back date at this point etc..).  

Another one, we know we had a problem and we've fixed it, do we need to check that use case all the time. 

We should be able to pass along ledger summary from the ledger fix (we are calling that one too much).

When totals are adjusted (resurrected, how this affects user, especially when stock can't be adjusted because it's reserved) ? (link to the internal doc)

Don't have to re-check if ledger was fixed, if no action was performed by previous fix

Find ledger discrepancy does not need stock_line filter, it just needs store_id filter (need to add to view), which i think will make it more performant.

## What went wrong? 😲

Historic ledger discrepancies when migrating from OG mobile sites causing negative numbers in:
* R&R forms
* Program Requisition with extra columns 
* In item ledger, Ledger 

See this doc for [examples](https://drive.google.com/drive/u/1/folders/1MeNljsanO3WiC4j8eoEmiLPo1BLmxua2).

## Expected behaviour 🤔

There is no historic ledger discrepancy

## Solution

After talking with @craigdrown @npendergrast @JonnaQuismundo @louisaw123 about this, we concluded that we need:

1. Identify and notify about ledger discrepancy in timely manner
2. Able to fix ledger discrepancy on remote site
3. Investigate reason for ledger discrepancy

This issue is about 2, although I do want to mention that 1 would need to be done and we want to deal with 2 before 3 (it also seems that most of the issues are from OG mobile sites, however we have seen omSupply ledger issue already)

## 2. Able to fix ledger discrepancy on remote site

1. Support team can request (via mSupply central) ledger discrepancy fix for item/store.

2. User can see if items have ledger discrepancies and able to manually run ledger discrepancy fix.

3. User is informed that stocktake needs to be done for items that had ledger discrepancy fix. 

Ledger discrepancy fix will consist of looking through ledger, finding negative historic balances and deploying a strategy to fix those balances, for every stock line with negative historic ledger discrepancy:

If there is a balancing incoming invoice but it's dated in the future, bring it back (change date). Otherwise create inventory adjustment.

1. This is done via message, created in mSupply, for item/store.
2. Some UI to show a table of possible ledger discrepancies (should there be a notification in form of dashboard stat ? would imply running this on schedule)
3. Not sure this might need a design, or stocktake option, thinking dashboard ?

This solution will still need a bit of refinement.

## How to Reproduce 🔨

See this doc for [example](https://drive.google.com/drive/u/1/folders/1MeNljsanO3WiC4j8eoEmiLPo1BLmxua2).

Setup omSupply site with store in data file, sync, check item ledger as per example.

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

andreievg:
Fixes #8632 (partially)

# 👩🏻‍💻 What does this PR do?

* Adds skeleton for ledger fix functions
* Improves speed overall view for ledger discrepancy check (slightly)
* Adds by stock line in rust ledger discrepancy check

For testing instruction and [full documentation](https://github.com/msupply-foundation/open-msupply/blob/ab87308017b509addfb1a0813c42d6ce47eb1613/server/service/src/ledger_fix/README.md) see downstream [PR/branch](https://github.com/msupply-foundation/open-msupply/pull/8717)


server/service/src/ledger_fix/fixes/adjust_historic_incoming_invoices copy.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::ledger_fix::fixes::LedgerFixError;
+
+const MAX_ITERATIONS: i32 = 100;
+
+pub(crate) fn adjust_historic_incoming_invoices(
+    connection: &StorageConnection,
+    operation_log: &mut String,
+    stock_line_id: &str,
+) -> Result<(), LedgerFixError> {
+    operation_log.push_str("Starting adjust_historic_incoming_invoices\n");
```
andreievg, line 16: there is extra condition added to this in upstream doc branch

server/service/src/ledger_fix/mod.rs
```diff
@@ -1,2 +1,60 @@
...
+
+    // Unwrap is safe stock_line must exist at this pont
+    let stock_line = StockLineRowRepository::new(connection)
+        .find_one_by_id(stock_line_id)?
+        .unwrap();
+
+    let available = stock_line.available_number_of_packs * stock_line.pack_size;
+    let total = stock_line.total_number_of_packs * stock_line.pack_size;
+
+    if available > 0.0 || total > 0.0 && ledger_lines.is_empty() {
```
andreievg, line 47: these conditions have been changes slightly in upstream branches

PR #8721, opened 2025-08-04
Add spacer to re-instate padding

CarlosNZ:
Follow up for #8652 to restore the NavLink arrow padding

# 👩🏻‍💻 What does this PR do?

There was a change in #8652 to remove a `flexGrow` in order to make the sync badge attach to the icon rather than floating out on its own. That inadvertantly also made the Nav link arrows (chevron) no longer push out to the end of the button.

This fixes the latter, while keeping the intended spacing for the actual sync badge.

It may seem a bit hacky adding an empty `span` in order to create padding. However, I would argue that it's less so than what was there originally -- it was using the `Badge` container to add the flex-grow, even when there was no badge, which was confusing and unclear. At least it's now explicit where the extra padding is coming from.

I tried setting the width of the container to "100%" (or slightly less), but it wasn't exactly the same as it used to be, and it would require more potentially regressive tweaks to wrangle it into place.

# 🧪 Testing

- [ ] Sync badge is still snug with the icon:
  Screenshot 2025-08-04 at 2 37 06 PM
- [ ] Padding for menu item arrows is as it used to be -- arrows are pushed to the right
  Screenshot 2025-08-04 at 2 38 25 PM

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
+++ b/client/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
@@ -172,7 +172,6 @@ export const AppNavLink: FC<AppNavLinkProps> = props => {
                       color: 'gray.dark',
                     }
                   : undefined,
-              // flexGrow: 1,
 
               '& .MuiBadge-badge:not(.MuiBadge-invisible)': drawer.isOpen
                 ? {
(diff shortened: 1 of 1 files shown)


client/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
```diff
@@ -193,11 +192,10 @@ export const AppNavLink: FC<AppNavLinkProps> = props => {
                     selected || isSelectedParentItem ? 'bold' : 'normal',
                   color: isSelectedParentItem ? 'primary.main' : undefined,
                 },
-                flexGrow: 0,
               }}
             />
           </Badge>
-
+          <span style={{ flexGrow: 1 }} />
```
CarlosNZ, line 198: Now we can clearly see there's a "spacer" here, as opposed to it being hidden inside the Badge component, which is invisible most of the time.

PR #8725, opened 2025-08-04
Add authorised sync

### LINKED ISSUE #8639: Sync is_authorised to purchase orders via status
In mSupply we have an `is_authorised` boolean.
When syncing we should set that if the status is `Authorised` or `Finalised`

When translating the purchase order status, we currently miss the authorised status, we need to include the boolean in this mapping...
```
fn from_legacy_status(status: &LegacyPurchaseOrderStatus) -> PurchaseOrderStatus {
    let oms_status = match status {
        LegacyPurchaseOrderStatus::Nw => PurchaseOrderStatus::New,
        LegacyPurchaseOrderStatus::Sg => PurchaseOrderStatus::New,
        LegacyPurchaseOrderStatus::Cn => PurchaseOrderStatus::Confirmed,
        LegacyPurchaseOrderStatus::Fn => PurchaseOrderStatus::Finalised,
        LegacyPurchaseOrderStatus::Others => PurchaseOrderStatus::New, // Default to New for
    };
    oms_status
}
```

jmbrunskill:
Fixes #8639 

# 👩🏻‍💻 What does this PR do?

Translates the is_authorised boolean to a new `Authorised` status in OMS.
Add status to oms_fields

# 🧪 Testing

- [ ] Try initialising a OMS Site from a Legacy site with purchase orders
- [ ] Check that ones that have been authorised show up as Authorised or Finalised (depending)
- [ ] Check nothing it broken...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/translations/purchase_order.rs
+++ b/server/service/src/sync/translations/purchase_order.rs
@@ -330,7 +330,8 @@ impl SyncTranslation for PurchaseOrderTranslation {
             .and_then(|oms_field| oms_field.finalised_datetime);
 
         let status = oms_fields
-            .and_then(|oms_field| oms_field.status)
+            .clone()
+            .map(|oms_field| oms_field.status)
             .unwrap_or_else(|| from_legacy_status(&status, is_authorised));
 
         let result = PurchaseOrderRow {
--- a/server/service/src/sync/translations/purchase_order.rs
+++ b/server/service/src/sync/translations/purchase_order.rs
@@ -524,7 +524,7 @@ fn to_legacy_status(status: &PurchaseOrderStatus) -> LegacyPurchaseOrderStatus {
 fn check_is_authorised(status: &PurchaseOrderStatus) -> bool {
     matches!(
         status,
-        PurchaseOrderStatus::Authorised | PurchaseOrderStatus::Finalised // Assuming Finalised is always authorised, but might be skipped in some cases
+        PurchaseOrderStatus::Authorised | PurchaseOrderStatus::Finalised // Assuming Finalised is always authorised, but the action might be skipped if authorisation is not required due to global preference
     )
 }
 
--- a/server/service/src/sync/translations/purchase_order.rs
+++ b/server/service/src/sync/translations/purchase_order.rs
@@ -223,6 +223,8 @@ pub struct LegacyPurchaseOrderRow {
     #[serde(serialize_with = "date_option_to_isostring")]
     pub received_at_port_date: Option<NaiveDate>,
     #[serde(default)]
+    pub is_authorised: bool,
+    #[serde(default)]
     #[serde(deserialize_with = "object_fields_as_option")]
     pub oms_fields: Option<PurchaseOrderOmsFields>,
 }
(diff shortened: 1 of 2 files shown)


server/service/src/sync/translations/purchase_order.rs
```diff
@@ -479,13 +491,22 @@ impl SyncTranslation for PurchaseOrderTranslation {
...
-        LegacyPurchaseOrderStatus::Fn => PurchaseOrderStatus::Finalised,
-        LegacyPurchaseOrderStatus::Others => PurchaseOrderStatus::New, // Default to New for
+        LegacyPurchaseOrderStatus::Cn => {
+            if is_authorised {
+                PurchaseOrderStatus::Authorised
+            } else {
+                PurchaseOrderStatus::Confirmed
+            }
+        }
+        LegacyPurchaseOrderStatus::Fn => PurchaseOrderStatus::Finalised, // authorised might or might not be true in this case...
```
zachariah-at-msupply, line 508: If we're assuming that finalised <=> authorised, then, similar to copilot's suggestion, I thought maybe it might be most consistent to make it just confirmed if it's not authorised?
```suggestion
LegacyPurchaseOrderStatus::Fn => {
            if is_authorised {
                PurchaseOrderStatus::Finalised
            } else {
                // We're assuming a finalised order is also one that is authorised atm, so for now, if not authorised, default it to confirmed.
                PurchaseOrderStatus::Confirmed
            }
}
```

server/service/src/sync/translations/purchase_order.rs
```diff
@@ -479,13 +491,22 @@ impl SyncTranslation for PurchaseOrderTranslation {
...
-        LegacyPurchaseOrderStatus::Cn => PurchaseOrderStatus::Confirmed,
-        LegacyPurchaseOrderStatus::Fn => PurchaseOrderStatus::Finalised,
-        LegacyPurchaseOrderStatus::Others => PurchaseOrderStatus::New, // Default to New for
+        LegacyPurchaseOrderStatus::Cn => {
+            if is_authorised {
+                PurchaseOrderStatus::Authorised
+            } else {
+                PurchaseOrderStatus::Confirmed
+            }
+        }
```
zachariah-at-msupply, line 507: Just flagging in case it's important: there's a comment on line 518 of this same file saying that legacy status Cn `// should also set is_authorised to true` it seems this might not be happening here?

zachariah-at-msupply (commented):
Looks great, tests pass, good to go I reckon! I added a couple of suggestions but it seems to do the job except for those details, so see if you think they are relevant - you'll know more than me! Then I'll approve once that's confirmed either way.

server/service/src/sync/translations/purchase_order.rs
jmbrunskill: I think if it was finalised before it should stay finalised. I think this scenario is when mSupply isn't using authorisation so it doesn't need to be set...

server/service/src/sync/translations/purchase_order.rs
jmbrunskill: I've updated the comment, hopefully that makes it clearer. It's just meant to be a note to the reader that we'll also set that in another part of the code.

zachariah-at-msupply (approved):
Done and done! LGTM :)

Issue #8240, opened 2025-06-19
Add Missing Fields to Purchase Order Form


zachariah-at-msupply:
This is a follow up issue of https://github.com/msupply-foundation/open-msupply/issues/8182

## Is your feature request related to a problem? Please describe 👀

- Add button in the purchase order detail view to print purchase order forms and reports.

- Generate the form's json, so it will be upserted into databases on startup.

- Add fields to the purchase order print-out. It should have the fields in [this](https://drive.google.com/file/d/1EwCxiIcxbxRmb6rnDWMmsM3kQ4eiC79l/view), mSupply's purchase order print-out. Some are already added, but we still need to add these. They should have the same layout as mSupply's purchase order print-out:
- [x] a watermark reading 'authorised' or 'unauthorised' depending on its status
- [ ] total cost in invoice currency
- [x] conversion rate between invoice currency and local currency (= the number of units of local currency per one unit of invoice currency)
- [ ] total cost in local currency
And each row of the table needs to have:
    - [ ] supplier code (N.B. is the supplier's code for the inidividual item e.g. an SKU, not the code that identifies the supplier)
    - [ ] cost
    - [ ] line total cost in local currency
    - [ ] comment

- Those things are here in mSupply's example:
![Image](https://github.com/user-attachments/assets/8ea4acf9-9fbd-4b91-85aa-2cd549411854)

### Acceptance Criteria ✅

- [ ] This is a standard report: it should be available in a standard installation of OMS 2.9.
- [ ] The print button in a purchase order's detail view should allow you to print that purchase order.
- It should include these fields (this is just a list of the fields in mSupply's example report, linked above):
- [x] name of the store that made the purchase order
- [x] name of the supplier that the purchase order is made to
- [x] order number
- [x] expected date
- [x] date confirmed
- [x] date printed
- [ ] a watermark reading 'authorised' or 'unauthorised' depending on its status
- [ ] total cost in invoice currency
- [ ] conversion rate between invoice currency and local currency (= the number of units of local currency per one unit of invoice currency)
- [ ] total cost in local currency
- [ ] and for each item:
    - [ ] line number
    - [ ] supplier's item code
    - [ ] item code
    - [ ] item name
    - [ ] qty
    - [ ] pack
    - [ ] unit
    - [ ] total qty
    - [ ] cost
    - [ ] line total cost in local currency
    - [ ] comment

### Moneyworks Jobcode 🧰

[...]

PR #8777, opened 2025-08-07
8645 purchase order print button doesnt work

### LINKED ISSUE #8645: Purchase Order Print button doesn't work
## What went wrong? 😲

```
    "data": {
        "generateReport": {
            "__typename": "PrintReportError",
            "error": {
                "__typename": "FailedToFetchReportData",
                "description": "Failed to query data required for the report",
                "errors": [
                    {
                        "locations": [
                            {
                                "column": 7,
                                "line": 5
                            }
                        ],
                        "message": "Unknown field \"expectedDeliveryDatetime\" on type \"PurchaseOrderNode\". Did you mean \"expectedDeliveryDate\"?"
                    },
                    {
                        "locations": [
                            {
                                "column": 11,
                                "line": 9
                            }
                        ],
                        "message": "Unknown field \"packSize\" on type \"PurchaseOrderLineNode\"."
                    },
                    {
                        "locations": [
                            {
                                "column": 11,
                                "line": 11
                            }
                        ],
                        "message": "Unknown field \"numberOfPacks\" on type \"PurchaseOrderLineNode\"."
                    },
```
## Expected behaviour 🤔

Should print!

This was introduced when I changed some names of fields in the database I think.
number_of_packs is no longer available for example, it's now `stock_on_hand_in_units` or `requested_number_of_units` in the database....

## How to Reproduce 🔨

1. Create a purchase order
2. Add a line
3. Click Print button at top of the screen

## Your environment 🌱

- Open mSupply Version:2.10-DEV

- Platform:

- Database type:

zachariah-at-msupply:
Fixes #8645

And #8240 

# 👩🏻‍💻 What does this PR do?

Renders all the right fields in purchase order forms.
Adds some of those fields to the GraphQL API.
Modifies some of those field's database definitions.
This should fulfill [point 10.3.2 in the PRD](https://docs.google.com/document/d/1asQdDMpUINwoML_yXerlR6ID6ZOrdGiu_mjHWSxTp_8/edit?pli=1&tab=t.0#heading=h.v70iotrp17hr).
So now PO forms look like this:

image

Some things to scrutinise:
- [x] It was decided to make total cost = the orderTotalAfterDiscount field that's stored in the purchase order table.
- [ ] I made foreign exchange rate not null and default to 1 in the database. To make this work I had to update some of the test data, including adding curr_rate to some legacy records that were testing how OMS would respond to having a blank curr_rate fields. OMS will absolutely cope with having blank curr_rate fields but I couldn't think of a time-effective way to get the tests passing with blank fields being inputted. I don't think this harms the tests, but it is a bit of a hacky workaround.
- [ ] I assumed local currency = currency x 'foreign_exchange_rate' in the DB is this correct?
- [ ] How will we know to update the form template if the authorisation logic changes? Right now it prints 'authorised' if status = authorised or finalised, and prints 'unauthorised' otherwise. Might this need to be changed in the future and if so, if we change the logic elsewhere in the code, how can we ensure we update the template logic to match it? I added a comment to the logic in the translations that affects this, but maybe there's other logic that I don't know about...

# 🧪 Testing

- [ ] Add a purchase order with a few lines.
- [ ] Go to its detail view and click print > purchase order.
- [ ] Check that all the fields that should be included are included (list below).

### List of fields to be included in a PO form:
(this is just a list of the fields in mSupply's example report, linked [here](https://drive.google.com/open?id=1EwCxiIcxbxRmb6rnDWMmsM3kQ4eiC79l&usp=drive_fs)):

- [ ] name of the store that made the purchase order
- [ ] name of the supplier that the purchase order is made to
- [ ] order number
- [ ] expected date
- [ ] date confirmed
- [ ] date printed
- [ ] a watermark reading 'authorised' or 'unauthorised' depending on its status
- [ ] total cost in invoice currency
- [ ] conversion rate between invoice currency and local currency
- [ ] total cost in local currency
- and for each item:
    - [ ] line number
    - [ ] item code
    - [ ] item name
    - [ ] qty
    - [ ] pack
    - [ ] unit
    - [ ] total qty
    - [ ] supplier's item code
    - [ ] cost
    - [ ] line total cost in local currency
    - [ ] comment

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


zachariah-at-msupply:
Just adding a comment in response to your question @jmbrunskill:
It seems that when making a PO in OG, curr_rate is never null and defaults to 1. The way it works is that you select which currency you want the PO to be in, and that currency's exchange rate seems to default to 1. Here's a video:

https://github.com/user-attachments/assets/d620270b-4507-4299-9972-65a4619c1c31

server/repository/src/migrations/v2_10_00/add_purchase_order_tables.rs
```diff
@@ -89,7 +89,8 @@ impl MigrationFragment for Migrate {
                     stock_on_hand_in_units {DOUBLE} NOT NULL DEFAULT 0.0,
                     supplier_item_code TEXT,
                     price_per_unit_before_discount {DOUBLE} NOT NULL DEFAULT 0.0,
-                    price_per_unit_after_discount {DOUBLE} NOT NULL DEFAULT 0.0
+                    price_per_unit_after_discount {DOUBLE} NOT NULL DEFAULT 0.0,
+                    comment TEXT
```
GeronimoJohn, line 93: Personally think that we should write another migration for this. Though it's only in develop and no user should have used this yet so its fine to change this? Thoughts @jmbrunskill

server/service/src/sync/translations/purchase_order.rs
```diff
@@ -525,6 +525,7 @@ fn check_is_authorised(status: &PurchaseOrderStatus) -> bool {
     matches!(
         status,
         PurchaseOrderStatus::Authorised | PurchaseOrderStatus::Finalised // Assuming Finalised is always authorised, but the action might be skipped if authorisation is not required due to global preference
+                                                                         // N.B. if this logic changes, update the Purchase Order form's logic (the 'AUTHORISED/UNAUTHORISED' watermark in this file: /Users/zachariah/code/open-msupply/standard_forms/purchase-order/latest/src/template.html)
```
GeronimoJohn, line 528:
```suggestion
// N.B. if this logic changes, update the Purchase Order form's logic (the 'AUTHORISED/UNAUTHORISED' watermark in this file: .../open-msupply/standard_forms/purchase-order/latest/src/template.html)
```

standard_forms/purchase-order/latest/convert_data_js/package.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "description": "",
+  "devDependencies": {
+    "@jest/globals": "^29.7.0",
+    "webpack": "^5.76.2",
+    "webpack-cli": "^5.0.1"
+  }
```
GeronimoJohn, line 16: Thinking that this should also need: 
```
    "jest": "^29.7.0",
    "ts-jest": "^29.2.6",
```

standard_forms/purchase-order/latest/convert_data_js/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "convert_data_js",
+  "version": "1.0.0",
+  "main": "convert_data.js",
+  "scripts": {
+    "build": "webpack"
+  },
```
GeronimoJohn, line 7: would be good to add test script: 
```
    "test": "jest"
```

Can we please try this out and get the test to run, I think it'll require a few more dev dependencies

GeronimoJohn (approved):
Working well! @zachariah-at-msupply. I've left a few things to note before merging to main. If we could please address them before merging then that would be lovely. 

Also noticed that for a PDF the image shows as the placeholder/undefined one when there's none set? 


Screenshot 2025-08-08 at 11 32 32 AM

zachariah-at-msupply:
> Screenshot 2025-08-08 at 11 37 36 AM
> 
> RE: image above. Is that only for tests? I don't see any other changes relating to that. I global searched `local_currency`

This is the calculation done to get the value for the 'Total cost in local currency' field on the form.
We query the database for order_total_after_discount and foreign_exchange_rate fields and then multiply them together.
I was wondering if that was correct.

zachariah-at-msupply:
> Working well! @zachariah-at-msupply. I've left a few things to note before merging to main. If we could please address them before merging then that would be lovely.
> 
> Also noticed that for a PDF the image shows as the placeholder/undefined one when there's none set?
> Screenshot 2025-08-08 at 11 32 32 AM

I'll fix that now, to make it blank.

server/repository/src/migrations/v2_10_00/add_purchase_order_tables.rs
jmbrunskill: Yes I think a new migration for comment would be better. Changing to not null is ok if we have always had values there should be fine, but a  new column everyone will miss, so better to do a new migration

zachariah-at-msupply:
> > Screenshot 2025-08-08 at 11 37 36 AM
> > RE: image above. Is that only for tests? I don't see any other changes relating to that. I global searched `local_currency`
> 
> Sorry, perhaps my writing wasn't the clearest it could have been! This is the calculation done to get the value for the 'Total cost in local currency' field on the form: we query the database for order_total_after_discount and foreign_exchange_rate fields and then multiply them together. My original comment was asking to check if that is correct @jmbrunskill @GeronimoJohn ?

Based on this PO print out from OG, local currency = PO currency * foreign exchange rate, so that's what this PR will use:
image

PR #8796, opened 2025-08-10
8785 viewforpuchaseorderpricing

### LINKED ISSUE #8785: Create View for total price and discounts
- [ ] Create a view that calculates total price  based on sum of costs from `purchase_order_line`. (Needs to work in both SQLite and postgres
- [ ] Add migration for the new view 
- [ ] Create a join table with the `purchase_order_id`, `total_cost_before_discount`
- [ ] Remove order_total_before_discount and order_total_after_discount from `purchase_order` including service layer
- [ ] Update insert and update services
- [ ] Change service layer to use domain object containing the joined price row
- [ ] Update tests
- [ ] Update sync

jmbrunskill:
Fixes #8785

# 👩🏻‍💻 What does this PR do?

Calculates the total order price for a purchase order based on a view.
Removes unused/calculated fields from the DB

Note: Small change from mSupply (I think) where the discounted price was showing just the difference between the whole order discount, I've currently translated this to relate to the pre-line discount price as I think that's more correct but won't display quite right in mSupply? 

# 🧪 Testing

- [ ] Create purchase orders with different prices/lines/discounts
- [ ] Make sure totals add up correctly
- [ ] Sync to mSupply
- [ ] Re-initialise your site
- [ ] Check all totals add up correctly still and no numbers have changed

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


fergie-nz:
"Note: Small change from mSupply (I think) where the discounted price was showing just the difference between the whole order discount, I've currently translated this to relate to the pre-line discount price as I think that's more correct but won't display quite right in mSupply?"

Am I understanding this to mean:

mSupply showed discounted price as "total of lines after the line discount" to "total of lines after line discount AND total order discount"

Whereas in OMS we will discounted price as difference between "total of lines before discount" to "total of lines after line discount AND total order discount"?

server/repository/src/migrations/views/mod.rs
```diff
@@ -640,6 +641,35 @@ CREATE VIEW vaccination_course AS
...
+            ), 0) AS line_total_after_discount,
+            COALESCE(SUM(
+                CASE 
+                    WHEN pol.authorised_number_of_units IS NOT NULL 
+                    THEN pol.authorised_number_of_units * pol.price_per_unit_after_discount
+                    ELSE pol.requested_number_of_units * pol.price_per_unit_after_discount
+                END
+            ), 0) * (1-(COALESCE(po.supplier_discount_percentage, 0)/100)) AS order_total_after_discount
+        FROM
+            purchase_order po JOIN purchase_order_line pol on po.id = pol.purchase_order_id
```
fergie-nz, line 669: I was thinking this also needs a WHERE clause for deleted datetime - and I realise we don't have it for purchase order lines. 

I guess that is fine? Just flagging here as I don't remember discussing.

server/service/src/sync/translations/purchase_order.rs
```diff
@@ -458,8 +463,8 @@ impl SyncTranslation for PurchaseOrderTranslation {
             communications_charge,
             insurance_charge,
             freight_charge,
-            supplier_discount_amount,
-            purchase_order_number,
+            supplier_discount_amount: line_total_after_discount
+                * (supplier_discount_percentage.unwrap_or(0.0) / 100.0),
```
fergie-nz, line 467: This makes sense - but I was wondering is there any risk of weird floating point errors resulting from different calculation methods between sql vs rust?

fergie-nz:
> > "Note: Small change from mSupply (I think) where the discounted price was showing just the difference between the whole order discount, I've currently translated this to relate to the pre-line discount price as I think that's more correct but won't display quite right in mSupply?"
> > Am I understanding this to mean:
> > mSupply showed discounted price as "total of lines after the line discount" to "total of lines after line discount AND total order discount"
> > Whereas in OMS we will discounted price as difference between "total of lines before discount" to "total of lines after line discount AND total order discount"?
> 
> I'm not sure if follow so I think an example might help!
> 
> In mSupply we are displaying the total cost, and discounted code for the whole invoice
> 
> image However we are also taking into account the line level discount before calculating the total discount. image
> For mSupply in this case I'm sending the non-discounted total as $50,000 (500 ajt packs X $100) Not the discounted price of $25,000, which then gets further discounted to $22,500
> 
> Now that I'm seeing those numbers I'm worried if the prices we're getting is in packs or units when syncing need to double check I think!

Great thank you, I understand and looks great

fergie-nz (approved):
Nice James looks good to me.

I had some thoughts pop up while reviewing which I have commented - but nothing blocking and looks awesome

server/graphql/types/src/types/purchase_order.rs
```diff
@@ -47,7 +48,12 @@ impl PurchaseOrderNode {
     }
 
     pub async fn order_total_after_discount(&self) -> f64 {
-        self.row().order_total_after_discount
+        let order_total_after_discount = match &self.stats {
+            Some(stats) => stats.order_total_after_discount,
+            None => 0.0,
```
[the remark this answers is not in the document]
jmbrunskill: I think it is, if there's no lines then there will be not total and therefore $0 price? Feels nicer than optionals to me but I'm not sure if everyone agrees...

server/repository/src/migrations/views/mod.rs
```diff
@@ -640,6 +641,35 @@ CREATE VIEW vaccination_course AS
         GROUP BY
 	        invoice_line.invoice_id;
 
+    CREATE VIEW purchase_order_stats AS
+        SELECT
+            po.id AS purchase_order_id,
+            COALESCE(SUM(
+                CASE 
+                    WHEN pol.authorised_number_of_units IS NOT NULL 
+                    THEN pol.authorised_number_of_units * pol.price_per_unit_before_discount
```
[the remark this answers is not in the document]
jmbrunskill: This is in the follow up PR... https://github.com/msupply-foundation/open-msupply/pull/8812

server/repository/src/migrations/views/mod.rs
jmbrunskill: Good point we don't have `deleted_datetime` on `purchase_order_line` but if we did this would be important...

PR #8798, opened 2025-08-10
8759 prevent old transfers by months glob pref

### LINKED ISSUE #8759: Prevent ancient picked inbound shipments being generated
## What went wrong? 😲

- Some users never quite get round to finalising their work. They'll have confirmed customer invoices/outbound shipments going back years. When you initialise OMS, transfers will be generated in `picked` status for all of them. This is annoying and editing is locked so the user can't even remove them.
  - https://github.com/msupply-foundation/open-msupply-internal/issues/67
- There is an edge case where old request requisitions don't have `request.linked_requisition_id` populated, but OG central server does have a log indicating that a response was generated once upon a time. The belief here is the simplest explanation - that a user probably deleted the response. There is a migration in OG to populate these IDs where we have that log. However we want some more prevention in OMS transfer processor rules
  - https://github.com/msupply-foundation/open-msupply-internal/issues/94

Lets prevent transfers generating based on X months, where X is a global pref.

- outbound shipment in `picked` status based on `picked_datetime`
- request requisition in `finalised` status based on `finalised_datetime`

## Expected behaviour 🤔

Mostly don't get ancient inbound shipments or customer requisitions :)

## How to Reproduce 🔨

Chris ought to put a good test datafile here...

1. A site with store data where the store has pending "confirmed" customer invoices and finalised request requisitions that haven't had transfer generated yet, and their status date is more than 12 months ago
2. Initialise OMS
3. Now have locked "picked" inbound shipments from more than 12 months ago
4. Now have response requisitions from orders made more than 12 months ago

## Your environment 🌱

- Open mSupply Version: <=2.9.3
- Legacy mSupply Central Server Version: ~8.3 and earlier definitely, later maybe

Chris-Petty:
Fixes #8759 

# 👩🏻‍💻 What does this PR do?

Prevents `picked invoice` and `requisition` transfers if their status dates were X months preference before initialisation of the site.

- Adds global pref
- Adds to processor logic to prevent the transfers

# 🧪 Testing

Incoming test file...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/translations/preference.rs
```diff
@@ -42,11 +42,10 @@ impl SyncTranslation for PreferenceTranslator {
         row: &ChangelogRow,
         r#type: &ToSyncRecordTranslationType,
     ) -> bool {
-        match r#type {
-            ToSyncRecordTranslationType::PullFromOmSupplyCentral => {
-                self.change_log_type().as_ref() == Some(&row.table_name)
-            }
-            _ => false,
+        if let ToSyncRecordTranslationType::PullFromOmSupplyCentral = r#type {
```
lache-melvin, line 45: I think this would just be `if` rather than `if let`

Though maybe best to leave this as match statement, I imagine pushing at least some variants of prefs from remote sites back to oms central will happen at some point 👀

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -46,15 +50,48 @@ impl RequisitionTransferProcessor for CreateResponseRequisitionProcessor {
...
+                    msg: e.to_string(),
+                    extra: "".to_string(),
+                })?;
+            if pref_months > 0 {
+                let sort = Sort {
+                    key: SyncLogSortField::DoneDatetime,
+                    desc: None,
+                };
+
+                let latest_log_sorted_by_finished_datetime = SyncLogRepository::new(connection)
```
lache-melvin, line 77: is this actually `first_fully_processed_log`? `latest` makes it sound like most recent, but I think that's not right if we're trying to find initialisation date?

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -359,3 +396,91 @@ fn generate_response_requisition_indicator_values(
...
+        let requisition_transfer_old = RequisitionTransferProcessorRecord {
+            other_party_store_id: "store_a".to_string(),
+            requisition: requisition_old.clone(),
+            linked_requisition: None,
+        };
+
+        let requisition_row_new = RequisitionRow {
+            id: "requisition_row_new".to_string(),
+            sent_datetime: Some(
+                NaiveDate::from_ymd_opt(2025, 8, 7)
```
lache-melvin, line 442: will this test fail in a few months time?

server/service/src/processors/transfer/requisition/mod.rs
```diff
@@ -159,6 +159,24 @@ pub(crate) fn process_requisition_transfers(
 #[error("Database error in processor ({0}) {1:?}")]
 pub(crate) struct ProcessorError(String, RepositoryError);
 
+#[derive(Debug)]
+enum RequisitionTransferOutput {
```
Chris-Petty, line 163: The story here (and for invoice processor equivalent) is that when I was testing my tests I realised that I had a false pass because a different check processor was preventing the transfer, rather than the one I had added. With this we can test precisely why the processor skipped. No substantial use for it in actual runtime code though.

Also tempted to do something funky like a hierarchy of enums where each processor has it's own enum of failure modes, but there is a fair bit of overlap between them and 1-big-enum is sorta easier to throw around anyway.

server/service/src/processors/transfer/requisition/create_response_requisition.rs
Chris-Petty: That's completely fair. In my defence I copy pasted it 😁

server/service/src/processors/transfer/invoice/mod.rs
```diff
@@ -271,6 +271,24 @@ fn get_delete_operation(
 
     Ok(Operation::Delete { linked_invoice })
 }
+#[derive(Debug)]
+#[allow(dead_code)] // The "Wrong" variants with fields don't use the fields yet. Be warned allowing dead_code here silences when we never use a particular variant as well.
+enum InvoiceTransferOutput {
```
lache-melvin, line 276: I really like this! Great to get it in debug logs too 🔥 

Future request would be to have put it in a sep PR than for the patch tho, for speed of review 🥲

Chris-Petty: Fair sorry- I did it for the reasons here: https://github.com/msupply-foundation/open-msupply/pull/8798#discussion_r2265484467

Was quite easy to write dud tests accidentally hitting the `store.created_date` check

server/service/src/processors/transfer/requisition/mod.rs
lache-melvin: siiick love the clearer skipping types

server/service/src/processors/transfer/invoice/mod.rs
andreievg, line 276: This is great ! Although as a patch (even though it's feature), would probably do it in two parts, one for the actual fix (2.9.4) and one for adding the error types (can do to 2.9.4 then to dev when that's merged)

server/service/src/processors/transfer/requisition/create_response_requisition.rs
lache-melvin: Oh I see, it won't, bc earliest sync log was 2024...

Wondering if better to insert a mock sync log here, than the base mock data you added - so we can see how it lines up with the dates of your test invoices? Kind of hard to see how it relates 👀 

Extra excuse is if someone adds a super old sync log to the shared mock data at some point, this test would fail 😱

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
+                return Ok(InvoiceTransferOutput::InvoiceCreatedBeforeStore);
+            }
+        }
+
+        // 6.
+        if outbound_invoice.invoice_row.status == InvoiceStatus::Picked {
+            if let Some(picked_date) = outbound_invoice.invoice_row.picked_datetime {
+                let pref_months = PreventTransfersMonthsBeforeInitialisation {}
+                    .load(connection, None)
+                    .map_err(|e| RepositoryError::DBError {
```
andreievg, line 114: This is probably the quick fix, for a patch, thanks, but follow up to extend transfer processor error to be an enum

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
+                            log.sync_log_row.integration_finished_datetime
+                        {
+                            if let Some(cutoff_date) = initialisation_date
+                                .checked_sub_months(Months::new(pref_months as u32))
+                            {
+                                if picked_date < cutoff_date {
+                                    return Ok(InvoiceTransferOutput::BeforeInitialisationMonths);
+                                }
+                            }
+                        }
```
andreievg, line 139: I think at this point i would make a separate method, just to reduce cognitive load of this processor, still keeping it somewhat 'high' level

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
+                        msg: e.to_string(),
+                        extra: "".to_string(),
+                    })?;
+                if pref_months > 0 {
+                    let sort = Sort {
+                        key: SyncLogSortField::DoneDatetime,
+                        desc: None,
+                    };
+
+                    let latest_log_sorted_by_finished_datetime = SyncLogRepository::new(connection)
```
andreievg, line 124: 'first' finished log i think, not latest ?

server/service/src/processors/transfer/requisition/create_response_requisition.rs
Chris-Petty: >Wondering if better to insert a mock sync log here, than the base mock data you added - so we can see how it lines up with the dates of your test invoices/requisitions? Kind of hard to see how "new" and "old" are defined 👀

Yea I generally hate the base mock data for this reason 😆 

>Extra excuse is if someone adds a super old sync log to the shared mock data at some point, this test would fail 😱

That's a THEM problem

Jokes I'll just copy pasta the logs into this test so they're safe

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
Chris-Petty: Yea a little dirty!

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
+                        .query(Pagination::one(), None, Some(sort))?
```
andreievg, line 125: First log might not have integration_finished_datetime I think, should filter by that one being 'not null'.

We also may decide to trim here, I think we should look at new key_value 'value' with initialisation date <- but can do that in develop

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
Chris-Petty: Laché said it first!!! I copied your variable name from 3 years ago 😛 

https://github.com/msupply-foundation/open-msupply/blob/deeda7e1d032e7f9c522bdec66b3649940c92b9f/server/service/src/sync/sync_status/status.rs#L219-L221

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
+                        .pop();
+
+                    if let Some(log) = latest_log_sorted_by_finished_datetime {
+                        if let Some(initialisation_date) =
+                            log.sync_log_row.integration_finished_datetime
+                        {
+                            if let Some(cutoff_date) = initialisation_date
```
andreievg, line 132: btw this is where you can do the .and on integration_finished_datetime and on latest_log_sorted_by_finished_datetime to reduce indentation and have just one Some statement (also reducing indentation)

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
Chris-Petty: Yea you're right the nice

server/service/src/sync/translations/preference.rs
andreievg: I agree, I think these refactors should be avoided in general (reduce diff reduce stylistic discussions), unless we've decided to only go with one pattern (but even then, patch update probably not the best time to apply those new patterns)

andreievg (changes requested):
Looks great in general, one note is about general targeted changes, especially in patch release, mainly about the new return type (and lots of files touched, that are not direcly related to the patch issues), there is a comment about another way to deal with this (still doing it but in a follow up PR, happy if it doesn't have an issue, as long as it's separated and goes into 'develop').

One change request:
* Filter by integration_datetime not null.

A couple of follow ups (for 'develop'):
* Error type on processor to be enum
* Keep processor method 'high level'
* Initialisation Date key_value (for develop)

server/service/src/processors/transfer/requisition/create_response_requisition.rs
andreievg: I feel like base mock data is something that would be 'shared' amongst test, do you reckon these two mock records would be useful in future tests somewhere ?

server/service/src/processors/transfer/requisition/mod.rs
andreievg: > The story here (and for invoice processor equivalent) is that when I was testing my tests I realised that I had a false pass because a different check processor was preventing the transfer, rather than the one I had added. With this we can test precisely why the processor skipped. No substantial use for it in actual runtime code though.

I think in this case checking that record is created first, and then adjusting picked date/pref/initialisation date and, then doing another test would be the way to go

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
Chris-Petty: But I did flip it from desc to asc so that's my bad lol

server/service/src/sync/translations/preference.rs
Chris-Petty: reverted

server/service/src/processors/transfer/requisition/create_response_requisition.rs
Chris-Petty: Not sure, maybe if someone needs something quick and dirty? Can just nuke them.

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
Chris-Petty: Turns out when sorted ascending, nulls are at the end of the sql result rather than start so this pagination/pop ensures that we get the earliest log. Regardless I've added a filter to be safe!

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -60,40 +65,80 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
...
             .find_one_by_id(&record_for_processing.other_party_store_id)?
             .ok_or(RepositoryError::NotFound)?;
-
         if let Some(created_date) = store.created_date {
-            let store_created_datetime = NaiveDateTime::new(
-                created_date - Duration::days(30),
-                NaiveTime::from_hms_opt(0, 0, 0).unwrap_or_default(),
-            );
+            let store_created_datetime =
+                NaiveDateTime::new(created_date - Duration::days(30), Default::default());
```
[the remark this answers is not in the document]
andreievg: Actually i think this is good, sorry i glossed over this previously

server/service/src/processors/transfer/invoice/mod.rs
andreievg: I think a bit too much work to de-couple, in general please

server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -252,3 +296,128 @@ fn generate_inbound_invoice(
...
+                ..Default::default()
+            },
+        )
+        .await;
+
+        let processor = CreateInboundInvoiceProcessor {};
+        let result = processor
+            .try_process_record(&connection, &invoice_transfer_old)
+            .unwrap();
+        assert!(
```
andreievg, line 408: there is assert_matches! test util helper : )

andreievg (approved):
Looks great, thanks for the change ! I think we will be on 2.9.4-RC now, so rebase (or even branch creation) would be needed.

It's worth capturing the remainder in another issue or a draft PR (to then go into develop when this one is merged there : ). If you are adding InitialisationDate to key value at this point worth adding new key value column for datetime (the ledger check would also benifit from that one)

PR #8812, opened 2025-08-11
8787 adjusted quantity status changes

### LINKED ISSUE #8787: Adjusted Quantity & Status changes
From the mSupply Docs...
When a purchase order (PO) is first created it has status “sg” (for “suggested”)
POs with status sg can be modified in almost every way, including changing the supplier.
Once there are no more lines to add to a PO, it can be confirmed and sent to the supplier. A PO is confirmed by clicking on the Confirm button.
Once a PO is confirmed, it has status “cn” (for “confirmed”) and an appropriate log entry is created.
Confirmed POs have the order amount added to the “on order” field for each item, so that if another PO is created for the same item, this quantity already on order will be taken into account.
Confirmed POs can have new lines added to them but only with an adjusted quantity, not an original quantity. This is to allow for amendments made by the supplier after they have received the order.
Details of existing confirmed PO lines can be amended. Their adjusted quantity can be amended (to reflect changes made by the supplier) original quantity cannot be amended.
Lines can be deleted from a confirmed purchase order as long as nothing has been received against them on a goods received note. If any part of a line has been received on a goods received note then mSupply will show an alert and will not allow you to delete the line.
Once no more stock is to be received against a PO, it can be finalised. An order becomes finalised by clicking on the Finalise order tick-box at the bottom left of the PO window.
POs that have been finalised have status “fn” (for “finalised”) and an appropriate log entry is created.
Finalised POs can no longer be modified in any way.
No more goods can be received against a finalised PO. The `unsupplied' lines will not appear when you look for outstanding PO lines (Supplier>Show outstanding PO lines) and the `unsupplied' quantities are ignored in any of mSupply's subsequent Stock on order calculations.

Tasks
- [x] Rename authorised -> adjusted 
- [x] Change line edit view to only allow editing of adjusted after confirmed
- [x] New lines can be added after confirmed but only have an adjusted quantity no original order qty (can be 0.0)

jmbrunskill:
Fixes #8787

# 👩🏻‍💻 What does this PR do?

Renames `authorised_number_of_units` to `adjusted_number_of_units`
Makes sure that the `adjusted_number_of_units` is what's available to adjust after the PO is confirmed.

Making this draft as I need to merge #8247 once it's merged into develop...

# 🧪 Testing

- [ ] After confirming a PO, make sure you can only adjust the adjusted quantity, not the requested quantity
- [ ] Make sure a new line can be added after confirmed, but that it will only receive an adjusted qty, requested should be 0...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


CarlosNZ:
Additional changes done by me:

- Add missing language strings
- Show "Adjusted quantity" in line edit view once status = "Confirmed"
- Re-order line edit columns so that quantity input(s) appear first (as that field is initially in focus -- so you just type a number and press "Enter" without moving to another field)
- Add pricing fields and discount percentage to line edit (the discount percentage is not actually stored, but we keep it as part of the draft so we can work with it and display it in UI, but it gets pulled out before saving)
- Logic to update price fields and discount percentage based of each other.
- Had to make a small tweak to Numeric Text input as it wasn't correctly formatting with decimal place limits if modified externally

GeronimoJohn (commented):
Feels much nicer! Definitely gives a better flow!

I noted down some minor NITs. But found two major things 👀 

1. When creating a new line, the Before Price or After Price values aren’t being saved. I think we just need to update the Create mutation to support these fields.

https://github.com/user-attachments/assets/841f363a-25bb-4fc5-b48f-063337a47dff


2. When a PO is confirmed and a new line is added, you can’t finalise the PO because that new line remains a placeholder line—the requested packs aren’t updated. The validation happens in StatusChangeButton I think

https://github.com/user-attachments/assets/12747f82-a58b-4f1e-a543-fd0e3494f73e

GeronimoJohn:
We could also write a follow up issue for the findings above? @jmbrunskill @CarlosNZ

PR #8848, opened 2025-08-12
8602 stocktake list flickers on load from ascending to descending order (take 2)

### LINKED ISSUE #8602: Stocktake list flickers on load from ascending to descending order (take 2)
## What went wrong? 😲

**DUPLICATE OF** #8131. _**Duplicated because the initial fix didn't work well.**_ 

The stocktake list flickers from ascending to descending order of "created" when you click on the button in nav.

https://github.com/user-attachments/assets/77beb114-5f49-461f-9382-454ef70e104b

## Suggested solution

We are making 2 identical requests on navigation to the page, bar the "desc" variable being false in the first and true in the second.
Image

Image

We should agree on what the default direction is (I think desc) and then only request that once.

I assume refreshing the page also makes 2 requests, but the full app render is too slow for you to see the order flickering.

## Expected behaviour 🤔

No flickering just load correct order straight away

## How to Reproduce 🔨

1. Just click the stocktakes button in navigation and watch it show ascending order for a moment before rerendering descending order (with the order arrow in column header animating as such)

## Your environment 🌱

- Open mSupply Version: dev 2.9 I assume prod tho
- Legacy mSupply Central Server Version: irrelevant

- Platform:irrelevant

- Database type:irrelevant

roxy-dao:
Fixes #8602

# 👩🏻‍💻 What does this PR do?
Refactors the list (I didn't change in all the places calling the old API only in the list view tho), delete and hasStocktake hooks.

Not sure if we want the initial sort in the URL params? Personally think it's fine to not have it there on initial load =/ but 🤷🏻.. plus doing it this way doesn't call the api again if you just click on stocktake! 

# 🧪 Testing

- [ ] Go to stocktakes
- [ ] Shouldn't see the list flickering from asc to desc sorting

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


aimee-mcneil-melville:
I noticed the same issue appears in other areas as well (shipments, prescriptions, sensors etc), and just noting here as from what I understand the resolution is using the new hooks pattern + removing the initial sort in the query params (specifically ones that sort by 'desc')

If removing the initial sort from query params is ok from a product perspective, should this be applied through the other areas that are already using the new hooks pattern to resolve the flickering and multiple api calls?

aimee-mcneil-melville (approved):
Sweet, this is working well, great to have some hooks refactoring in 🚀 

I just added a comment about a wider implementation but looking good to me!

roxy-dao:
> I noticed the same issue appears in other areas as well (shipments, prescriptions, sensors etc), and just noting here as from what I understand the resolution is using the new hooks pattern + removing the initial sort in the query params (specifically ones that sort by 'desc')
> 
> If removing the initial sort from query params is ok from a product perspective, should this be applied through the other areas that are already using the new hooks pattern to resolve the flickering and multiple api calls?

Yup would be good to removed everywhere where possible~ can be another issue thanks Aimee

PR #8857, opened 2025-08-13
7859 query and cache prefs

### LINKED ISSUE #7859: Query + cache OMS prefs once on app startup

lache-melvin:
Fixes #7859

# 👩🏻‍💻 What does this PR do?

Render `usePreferences` hook on mount, querying for all preferences. Use infinite cache (expect for when invalidated on edit, on sync, on store cache).

Cleans up all the areas where we had queried the pref in a parent component and passed through to the children, now can use directly, pref will be available.

# 🧪 Testing

- [ ] Go to an area of the app that is configured by a pref, some examples:
  - [ ] Prescription/outbound shipment allocation - should default to doses for vaccine items when `manage vaccine in doses` is on
  - [ ] Stocktake/inbound shipment edit modals - `Donor` column enabled by global pref
- [ ] Ensure the pref in question is enabled - then refresh the page. The pref-enabled feature should be visible immediately
- [ ] Without refreshing, switch stores to a store where the pref is off - feature is not visible
- [ ] Without refreshing, edit the pref - feature enables/disables accordingly
- [ ] On OMS central, edit a global or store pref
- [ ] On remote site, sync, and then without refreshing, see that feature has been enabled/disabled accordingly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/authentication/api/hooks/index.ts
```diff
@@ -6,4 +6,6 @@ export * from './useLogin';
 export * from './useGetUserPermissions';
 export * from './useIsCentralServer';
 export * from './useUpdateUser';
-export * from './usePreference';
+export * from './usePreferences';
+
+export const PREFERENCES_QUERY_KEY = 'preferences';
```
lache-melvin, line 11: This was in `system/Manage` - shouldn't be importing from system into common, so moved here, importing into system now 🙏

Chris-Petty:
Weeeeird. I too can't reproduce now. Wonder if it had anything to do with changing branches last time.

Chris-Petty (approved):

lache-melvin:
let us cross our fingers

PR #8871, opened 2025-08-14
8408 freezing part of the screen when scrolling the internal order

### LINKED ISSUE #8408: Freezing part of the screen when scrolling the internal order
When creating an internal order, if you scroll on the right hand side to see all the columns, you can't the item code or item description anymore.
It would be more user-friendly if that information stayed on screen

Image

Solution: like on the R&R form, ensuring that item code and item description are visible at all time when entering an order

roxy-dao:
Fixes #8408

# 👩🏻‍💻 What does this PR do?
Make the item and name columns sticky, so that they're always on the screen

https://github.com/user-attachments/assets/acee93b9-3940-424f-ada5-45ebc0982c84

# 🧪 Testing

- [ ] Best tested on small screens
- [ ] Go to any of the tables 
- [ ] Scroll
- [ ] Item name and code should always show

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/client/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -39,7 +39,26 @@ interface RenderRowsProps<T extends RecordWithId> {
   additionalRows?: JSX.Element[];
   /** will ignore onRowClick if defined. Allows opening in new tab */
   rowLinkBuilder?: (row: T) => string;
+  stickyColumnPositions?: Map<string | keyof T, number>;
 }
+
+const calculateStickyPositions = <T extends RecordWithId>(
+  columns: Column<T>[]
+): Map<string | keyof T, number> => {
+  const positions = new Map<string | keyof T, number>();
+  let currentPosition = 0;
+
+  columns.forEach(column => {
+    if (column.isSticky) {
+      positions.set(column.key, currentPosition);
+      const columnWidth = Number(column.width) || 120;
+      currentPosition += columnWidth;
+    }
+  });
+
+  return positions;
+};
+
 const RenderRows = <T extends RecordWithId>({
   mRef,
   data,
@@ -52,6 +71,7 @@ const RenderRows = <T extends RecordWithId>({
   isRowAnimated,
   additionalRows,
   rowLinkBuilder,
+  stickyColumnPositions,
 }: RenderRowsProps<T>) => {
   const t = useTranslation();
   const { localisedDate } = useFormatDateTime();
@@ -75,6 +95,7 @@ const RenderRows = <T extends RecordWithId>({
             localisedDate={localisedDate}
             isAnimated={isRowAnimated}
             rowLinkBuilder={rowLinkBuilder}
+            stickyColumnPositions={stickyColumnPositions}
           />
         ))}
         {additionalRows}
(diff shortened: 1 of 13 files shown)


client/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -89,9 +91,10 @@ const DataRowComponent = <T extends RecordWithId>({
                 ? theme => alpha(theme.palette.secondary.main, 0.1)
                 : null,
               '&.MuiTableRow-root': {
-                '&:nth-of-type(even)': {
-                  backgroundColor: 'background.toolbar',
-                },
+                backgroundColor:
```
roxy-dao, line 94: Forcing the colour to be one or the other so it's consistent throughout the app =/

lache-melvin:
Just noting this from Mark in the issue, did that recommendation change?

> Recommend also only fixing the code column, if there is limited space then the checkbox+code+name takes up half the screen!

Long names do take up about half the screen in tablet view:

Screenshot 2025-08-14 at 1 55 27 PM

Wondering if checkbox should be sticky instead? I think always being able to select the row would be helpful 👀

client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -141,6 +147,16 @@ export const HeaderCell = <T extends RecordWithId>({
     child
   );
 
+  const stickyStyles = isSticky
+    ? {
+        position: 'sticky' as const,
+        left: stickyPosition,
+        backgroundColor: 'white',
+        // When position is at 0, need to add some padding to avoid overlap
+        zIndex: `calc(${stickyZIndex} + 200)`,
```
lache-melvin, line 156: How does the padding apply to z index?

client/packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -144,10 +166,16 @@ const DataTableComponent = <T extends RecordWithId>({
...
-
     return cols.every(c => c.key === 'selection') ? [] : cols;
   }, [columns, columnDisplayState]);
 
+  const stickyColumnPositions = React.useMemo(() => {
+    return calculateStickyPositions(columnsToDisplay);
+  }, [columnsToDisplay]);
+  const sumStickyColumnWidths = Array.from(
+    stickyColumnPositions.values()
+  ).reduce((acc, width) => acc + width, 0);
```
lache-melvin, line 177: is there a reason this isn't all inside calculateStickyPositions?

lache-melvin (changes requested):
Oo v nice 🤩  requesting we go with fixed checkbox + code unless there has been a decision otherwise?

client/packages/common/src/ui/layout/tables/components/Header/Header.tsx
roxy-dao: Cause sum for position will be 0 if there's only 1 column since we want left = 0~ so just added some extra px to make sure the text in the header doesn't overlap

roxy-dao: Oh yeah true XD

lache-melvin (approved):
🔥

PR #8881, opened 2025-08-14
8787 add pack size

### LINKED ISSUE #8787: Adjusted Quantity & Status changes

jmbrunskill:
Fixes #8787

# 👩🏻‍💻 What does this PR do?

Adds number of packs, and packsize and changes the adjusted and requested packs to be calculated.

image

Could make the unit quantity trigger the calculations too, but feels unnecessary for now.

# 🧪 Testing

- [ ] Check that changing packsize and number of packs updates the adjusted or requested quantity
- [ ] adjusted should only change after confirmed
- [ ] requested should only change before confirmed
- [ ] finalised shouldn't allow any changes.
- [ ] Reloading/reopening purchase order lines should show correct number of packs

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
+++ b/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
@@ -62,6 +62,9 @@ export const usePurchaseOrderLineEditColumns = ({
                 setter: (patch: Partial<DraftPurchaseOrderLine>) => {
                   updatePatch({ ...patch });
                 },
+                cellProps: {
+                  isDisabled: true, // Edited via number of packs for now
+                },
               },
             ]
           : []),
@@ -73,7 +76,7 @@ export const usePurchaseOrderLineEditColumns = ({
             updatePatch({ ...patch });
           },
           cellProps: {
-            isDisabled: status === PurchaseOrderNodeStatus.Confirmed,
+            isDisabled: true, // Edited via number of packs for now
           },
         },
 
--- a/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
+++ b/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
@@ -9,7 +9,7 @@ import {
   PurchaseOrderNodeStatus,
   useColumns,
 } from '@openmsupply-client/common/src';
-import { calculatePricesAndDiscount } from './utils';
+import { calculatePricesAndDiscount, calculateUnitQuantities } from './utils';
 
 export const usePurchaseOrderLineEditColumns = ({
   draft,
@@ -23,6 +23,36 @@ export const usePurchaseOrderLineEditColumns = ({
   const columnDefinitions: ColumnDescription<DraftPurchaseOrderLine>[] =
     useMemo(
       () => [
+        {
+          Cell: NumberInputCell,
+          key: 'numberOfPacks',
+          label: 'label.requested-packs',
+          setter: patch => {
+            // Adjust the requested and adjusted number of units based on the number of packs
+            const adjustedPatch = calculateUnitQuantities(
+              'numberOfPacks',
+              status,
+              patch,
+              draft
+            );
+            updatePatch({ ...patch, ...adjustedPatch });
+          },
+        },
+        {
+          Cell: NumberInputCell,
+          key: 'requestedPackSize',
+          label: 'label.pack-size',
+          setter: patch => {
+            // Adjust the requested and adjusted number of units based on the new pack size
+            const adjustedPatch = calculateUnitQuantities(
+              'packSize',
+              status,
+              patch,
+              draft
+            );
+            updatePatch({ ...patch, ...adjustedPatch });
+          },
+        },
         ...(status === PurchaseOrderNodeStatus.Confirmed
           ? [
               {
@@ -46,6 +76,7 @@ export const usePurchaseOrderLineEditColumns = ({
             isDisabled: status === PurchaseOrderNodeStatus.Confirmed,
           },
         },
+
         {
           Cell: NumberInputCell,
           key: 'pricePerUnitBeforeDiscount',
--- a/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
+++ b/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
@@ -1,6 +1,7 @@
 import { ItemStockOnHandFragment } from '@openmsupply-client/system/src';
 import { DraftPurchaseOrderLine } from '../../api/hooks/usePurchaseOrderLine';
 import { FnUtils } from '@common/utils';
+import { PurchaseOrderNodeStatus } from '@common/types';
 
 export const createDraftPurchaseOrderLine = (
   item: ItemStockOnHandFragment,
@@ -19,6 +20,7 @@ export const createDraftPurchaseOrderLine = (
     pricePerUnitAfterDiscount: 0,
     // This value not actually saved to DB
     discountPercentage: 0,
+    numberOfPacks: 0,
   };
 };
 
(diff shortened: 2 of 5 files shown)


client/packages/purchasing/src/purchase_order/api/hooks/usePurchaseOrderLine.ts
```diff
@@ -18,6 +18,7 @@ export type DraftPurchaseOrderLine = Omit<
   purchaseOrderId: string;
```
CarlosNZ, line 18: Just a thought, though not strictly related to the PR (but we may want a followup?) -- should the "requestedPackSize" default to the item's default pack size?

And if not, I think it should at least default to `1` -- it's weird seeing a pack size of `0` as that's not actually a valid value.

Image

client/packages/purchasing/src/purchase_order/api/hooks/usePurchaseOrderLine.ts
```diff
@@ -61,12 +63,20 @@ export function usePurchaseOrderLine(id?: string) {
         100
       : 0;
 
+  // Number of packs is not in the DB, so we derived it from the draft
+  const initialNumberOfPacks = data?.nodes[0]?.requestedNumberOfUnits
+    ? (data?.nodes[0]?.adjustedNumberOfUnits ??
+        data?.nodes[0]?.requestedNumberOfUnits ??
+        0) / (data?.nodes[0]?.requestedPackSize ?? 1)
+    : 0;
```
CarlosNZ, line 71: So we're saying use the "adjusted" value if there is one, otherwise use the "requested" value?

In which case, isn't this a bit clearer?
```ts
  const initialNumberOfPacks =
    (data?.nodes[0]?.adjustedNumberOfUnits ??
      data?.nodes[0]?.requestedNumberOfUnits ??
      0) / (data?.nodes[0]?.requestedPackSize);
```

Or, to be clearer still:
```ts
  const adjustedUnits = data?.nodes[0]?.adjustedNumberOfUnits;
  const requestedUnits = data?.nodes[0]?.requestedNumberOfUnits ?? 0;
  const requestedPackSize = data?.nodes[0]?.requestedPackSize ?? 1;

  const initialNumberOfPacks =
    (adjustedUnits ?? requestedUnits) / requestedPackSize;
```

(I know, I probably should have done this for the `initialDiscountPercentage` too ;)

client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
```diff
@@ -86,3 +88,39 @@ export const calculatePricesAndDiscount = (
...
   }
 };
+
+type UnitCalcFields = 'numberOfPacks' | 'packSize';
+
+export const calculateUnitQuantities = (
+  changingField: UnitCalcFields,
+  status: PurchaseOrderNodeStatus,
+  data: Partial<DraftPurchaseOrderLine>,
+  draft: DraftPurchaseOrderLine | null | undefined
```
CarlosNZ, line 98: You shouldn't need both `data` and `draft` -- don't they contain the same info?

client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
```diff
@@ -32,6 +62,9 @@ export const usePurchaseOrderLineEditColumns = ({
                 setter: (patch: Partial<DraftPurchaseOrderLine>) => {
                   updatePatch({ ...patch });
                 },
+                cellProps: {
+                  isDisabled: true, // Edited via number of packs for now
+                },
```
CarlosNZ, line 67: Since this is always disabled, should probably remove the "setter" function -- it never gets used, the value is updated via the other fields)

client/packages/purchasing/src/purchase_order/DetailView/LineEdit/columns.ts
```diff
@@ -43,9 +76,10 @@ export const usePurchaseOrderLineEditColumns = ({
             updatePatch({ ...patch });
           },
           cellProps: {
-            isDisabled: status === PurchaseOrderNodeStatus.Confirmed,
+            isDisabled: true, // Edited via number of packs for now
```
CarlosNZ, line 79: Same for this one

client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
```diff
@@ -86,3 +88,39 @@ export const calculatePricesAndDiscount = (
...
+  switch (changingField) {
+    case 'numberOfPacks': {
+      numberOfPacks = data.numberOfPacks || 0;
+      break;
+    }
+    case 'packSize': {
+      requestedPackSize = data.requestedPackSize || 0;
+      break;
+    }
+  }
```
CarlosNZ, line 112: Don't think you need the switch here -- you're multiplying these two values together, it doesn't matter which one has changed in this case (it's only when there was three that it mattered)

CarlosNZ (changes requested):
Cool, couple of tweaks to make it easier to read/understand I think :)

client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
jmbrunskill: This is just the draft vs data thing I agree probably can remove it by the solution I mentioned above...

client/packages/purchasing/src/purchase_order/api/hooks/usePurchaseOrderLine.ts
CarlosNZ: Okay, should probably make a follow-up issue for that.

In the meantime, there's just this little linter error, then we should be good to go:

>   ./open-msupply/client/packages/purchasing/src/purchase_order/DetailView/LineEdit/utils.ts
>   97:7  error  'numberOfPacks' is never reassigned. Use 'const' instead      prefer-const
>   98:7  error  'requestedPackSize' is never reassigned. Use 'const' instead  prefer-const

PR #8890, opened 2025-08-14
8880 custom translations not saving

### LINKED ISSUE #8880: Custom translations are not saving
## What went wrong? 😲

1. Custom translations for some strings are not saving. 

2. For the login button, I saved a string and then deleted it. Even after deleting, the translated button which is `hello!` is still showing. I tried saving the default string -> it didn't bring the original text back (as shown on the video)

Also cleared my cache to make sure it's not interfering with the feature, no luck :(

https://github.com/user-attachments/assets/d0f4f2ea-066a-4d06-a54a-4223799bae08

## Expected behaviour 🤔

1. Strings should show if saved
2. Strings should go back to default if deleted

## How to Reproduce 🔨

1. On your Central Server, go to `Manage` -> `Global Preferences` -> `Custom Translations`
2. Add a few strings that will show on your login screen - for example, login.body + login.heading + button.login
3. See if any of them show (logout, login, sync, logout again - are they showing?)
4. Clear your cache just in case - are they showing? 
5. If one shows, try deleting it -> note the custom translation stays even if you remove it 

## Your environment 🌱

- Open mSupply Version: `2.10.00-dev-develop-08130349`

- Database type: `PostgreSQL`

lache-melvin:
Fixes #8880

# 👩🏻‍💻 What does this PR do?

Clears local storage cache of custom translations in production mode. 

Also restores the correct update call for translations, so they're applied when you navigate to the page in question - shouldn't need to refresh.

# 🧪 Testing

- [ ] With built version (or for devs - set `isDevelopment` in `IntlContext.ts` to `false`)
- [ ] On OMS central, edit a translation key
- [ ] Without refreshing, navigate to the page with that translation string
- [ ] It should show up immediately

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Manage/Preferences/Components/CustomTranslations/CustomTranslationsModal.tsx
+++ b/client/packages/system/src/Manage/Preferences/Components/CustomTranslations/CustomTranslationsModal.tsx
@@ -82,6 +82,11 @@ export const CustomTranslationsModal = ({
       )
       .forEach(key => localStorage.removeItem(key));
 
+    // Clear from i18next cache (specifically for when we delete a translation)
+    for (const lang of i18n.languages) {
+      i18n.removeResourceBundle(lang, CUSTOM_TRANSLATIONS_NAMESPACE);
+    }
+
     // Then reload from backend
     // Note - this is still requires the components in question to
     // re-render to pick up the new translations


aimee-mcneil-melville (changes requested):
Thanks Laché, this is working really well for saving! 🚀  All automatically update other than the App Bar buttons

I did get some strange behaviour with deleting though, can you look into this so that the deletion is consistent?
- Deleted one & the translation remained in the cache but wasn't in the translations table (also sometimes not in the selector)
- Deleted one and it reappeared in the custom table (not in cache)



https://github.com/user-attachments/assets/f48272c0-78af-4362-864a-fb97f8837eb0




https://github.com/user-attachments/assets/3d6eec72-e344-4bf3-86ff-17fca10236d7

lache-melvin:
Ah great finds thank you Aimee! I have pushed an update, so we clear the internal i18n cache as well on deletes 🙏

PR #8907, opened 2025-08-17
Kdd floating point prescision

### LINKED ISSUE #999: Secure cors for discovery
Epic: https://github.com/openmsupply/open-msupply/issues/996

Comment: https://github.com/openmsupply/open-msupply/pull/802/files/bd8f3fc5a6196285234e3a8e47cb5d07a303ab8d#r1025902024

Basically we ignore cors for certain user agents (to allow graphql during discovery), this is potentially problematic, and perhaps the client (electron and capacitor web view) should allow cors vs server.

aimee-mcneil-melville:
Fixes #999

# 👩🏻‍💻 What does this PR do?
KDD for decision on the handling of floating point numbers

# 🧪 Testing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


decisions/2025-08-15_floating_point_precision_stock_calculation.md
```diff
@@ -0,0 +1,262 @@
+# Handling Floating Point Precision in Stock Calculations
+
+- _Date_: 2025-08-15
+- _Deciders_:
+- _Status_: DRAFT
+- _Outcome_: Pending
+
+## Context
+
+Our system is a stock management system, and a key part of this is ensuring accurate representation of stock quantities for the users.
```
roxy-dao, line 10: ❤️

CarlosNZ:
Correct me if I'm wrong, but I would have thought that if the expected level of precision we'd want in out system (maybe 6dp?) is several orders of magntude larger than precision errors introduced by floating point arithmetic, then it should be fairly easy to identify and ignore them[^1] ? As long as we round the results of any calculation to a value that is *more* precise than our system requirement, but *less* precise than floating point errors (which only happen in the last 1 or 2 dp), then the floating point errors will nicely just "go away".

It would also deal with the "re-combining" problem: when you have a unit quanity divided in 3, you get 0.333333 of each, which then re-sums to 0.999999, which obviously can cause problems with equality checks, etc. But if you store results that are 2-3 dp beyond our guaranteed precision level, and then round the result of adding them (to our guaranteed precision level), you'll get correct results.

The hard part is really remembering where and when to implement this kind of rounding which, I guess, is one of the reasons why these "Number" libraries exist. But I don't think any particular representation is going to account for all possible issues, and so care and vigiliance in all calulations and roundings will still be required. I do think we could establish a set of "best practices" that would account for most of the scenarios we're likely to face. (The worst scenarios for handling precision are when you have extremely large numbers interacting with extremely small numbers, but I don't think that's something we really have to face in stock management)

I think a good place to start is to specify a level of precision we can guarantee to capture meaningfully, so then anything beyond that can be treated as the garbage that it probably is (looking at you `...00000000004 packs`)

[^1]: I realise that's not always the case, due to the "floating" part of floats, but would be reliable across several orders of magnitude

CarlosNZ:
Just adding a comment here to note another consideration we should factor into any agreed solution/rule structure:

https://github.com/msupply-foundation/open-msupply/pull/11219#discussion_r3084427617

CarlosNZ:
Here are Claude's thoughts on the matter:

https://claude.ai/share/96134d23-d499-4ef6-b1de-02fd12a9f16f

Interesting points -- definitely need to address this sooner rather than later I think, as it will be a big refactor. But probably worth doing if we want OMS to be mathematically accurate and consistent going forward.

Key takeaways:
- Decimal libraries the way to go
- Allocate, never divide (e.g. 100/3 => 33.34 + 33.33 + 33.33)
- Postgres okay, storing decimals in SQLite trickier, probably have to be string
- Always round line totals before summation (which we do in invoice lines, but not purchase order lines, so we're not consistent)
- Have a clear policy on discounts/taxes -- discount rounded lines then sum, or sum rounded lines then discount total?

craigdrown:
For the record, doing KDDs like this leaves a series of comments and no way to tell which ones are acted on or chosen to ignore. Please keep me looped in when there's a final draft thanks.

Chris-Petty:
@craigdrown I think there is fear to commit any which way cause it's actually a big endeavour to correct course and will generally make things harder going forward - a change likely needs training through the whole dev team to effect it. It's simultaneously a problem that shouldn't and can be deferred indefinitely while resources are low, as it's a slow burn background pain point compared to high urgency client deliverables.

Tangentially and without having reread any of this, I'm sure there is probably a middle ground for making a our own monetary type that still uses f64 but makes sure we do arithmetic in a consistent way. I.e. make sure "Allocate, never divide (e.g. 100/3 => 33.34 + 33.33 + 33.33)" happens robustly.

adamdewey:
Hey, not sure how relevant to the discussion but I saw mention of 6 dp above. 

Here is a screenshot from a client datafile that requires 7 dp for volume management (due to using m^3 for small items):


image

Chris-Petty:
@adamdewey Hahahaha nice. I'd say yes relevant thanks. FWIW, implementation wise (without checking what was done) it might be more favourable to have storage volumes stored in mL or cm^2 and hopefully end up with a whole number in many cases, but right there you have 19.2mL. Then just shift the dp in presentation up to m^3 as needed.

In either case the tangent I really wanted to say is that it could be automatically presented in mL easily (say if every value in a column is < 1L). 7dp is ugly and getting users to input so many 0s accurately seems prone to error (and tedium).

Also... I'm definitely poking the bull here but if there's a different channel for it; the vial storage volume might be inaccurate for warehouse management if it's not including outer cartons!

adamdewey:
Definitely all for making it easier for users to enter volumes 😁 

Yeah, that is a vial size, but it's averaged from a pack of 100 or so, so it's actually a 'packaged per-vial' size

(they are using pack size 1)

craigdrown:
@adamdewey  In OG you can enter "1ml" or "1l" into volume fields and it will store the value in m3 in the back end. OMS has a way to go ;-)

craigdrown:
@Chris-Petty Can I propose we distill all these comments down to a proposal in a google doc? (Later it will be saved as .md and added to our specs folder)

adamdewey:
@craigdrown - I saw that! 

Yes something like that would be good - or being able to set the volume granularity in the item master data so the appropriate unit is applied everywhere relevant in the UI for that item.

jmbrunskill:
closing, re-opened in internal repo

PR #8929, opened 2025-08-19
8916 remove location type restriction from items

### LINKED ISSUE #8916: When location restriction is removed in OG mSupply, omSupply does not update (still restricts items to certain location type)
## What went wrong? 😲

I used to have a location restriction set for an item (restricted to Cold Storage) -> therefore in omSupply, only 2 locations which were Cold Storages were showing. 

I hence removed that restriction and set it to None in OG mSupply -> however omSupply still restricts that item to 2 locations which are Cold Storages.

https://github.com/user-attachments/assets/ffa8eabf-a481-46bc-848f-75b8955de80e

## Expected behaviour 🤔

If a location restriction is removed in OG mSupply, omSupply should update and show all locations available in the store

## How to Reproduce 🔨

1. Have a few locations setup in omSupply, and an item you are testing with
2. Go to OG mSupply -> find your item -> restrict your location to a certain type (for example, Cold Storage or whatever you have set in your store)
3. Make sure you have a location with that type to test with
4. Go to Inventory -> View Stock -> your item -> note in the location dropdown you will only see the restricted types ✅ 
5. Now in OG mSupply, remove that restriction and set it to None
6. Go back to OMS -> sync a few times -> note your item is still restricted to the type you set before

## Your environment 🌱

- Open mSupply Version: `2.10.00-RC-08151658`

- Database type: `PostgreSQL`

aimee-mcneil-melville:
Fixes #8916 

# 👩🏻‍💻 What does this PR do?

Removes restricted location type id from `item` if it is removed in OG -> allow update to None

The test passed prior to the change as it checks the synced information - unsure if the test actually adds value in this case

# 🧪 Testing

Setup in OG on your store:

- [ ]  Set some locations types: Item -> Show location types
- [ ]  Set some locations: Item -> Show locations -> add a location type to several locations (can also set in OMS)
- [ ]  Add some restricted location types to some items in OG -> Item -> Storage -> Restricted to
- [ ] In OMS confirm that the restricted location type is shown in Item -> Detail View -> Storage
- [ ] In OG, remove the restricted location type from the item (None)
- [ ] In OMS -> see the restricted location type is no longer there

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin (changes requested):
Awesome thanks Aimee!

> The test passed prior to the change as it checks the synced information - unsure if the test actually adds value in this case

Yeah great question - the issue was with the DB saving, and the test you've added only checks the conversion between the sync record and how it translates to our rust struct.

To add a test that actually validates your change, you could instead add a test directly to the item_row file - it could do an insert with all fields defined, and then an update to set the nullable fields to null, and assert they have saved correctly 🙏

aimee-mcneil-melville:
Thanks Laché, I've added a new test and removed the previous one 😄

lache-melvin (approved):
🚀

PR #8948, opened 2025-08-20
8533 date picker use dates in correct timezone

### LINKED ISSUE #8533: The selection from the date filters doesn't include records that are on the `To Date`
## What went wrong? 😲

For example,

If I have 3 records on the 10th and 2 on the 14th and I am filtering from the 10th to the 14th, I would expect to see the three of them but I only see the one from the 10th. I feel I should the 3 of them.

https://drive.google.com/file/d/1p8gFtf5B_ZNHZ_IUyx6ddkLDjT_t0u-t/view?usp=sharing

There is also this strange behaviour where I can set the To date to be the same date as the From date but I cannot set the From date to be the same  date as the To date. So if i choose the from date to be the 10th, I can choose the To date to be the 10th but if I choose the To date to be the 10th then I cannot choose the From date to be the 10th, manually it will change to the 9th automatically and from the date picker the 10th will be unavailable so I cannot choose it.

https://drive.google.com/file/d/1LrfHPCyn060lA-sDslKDldCYVUXARX0R/view?usp=sharing 

## Expected behaviour 🤔
I believe the selection should include records thaat are on the To date
## How to Reproduce 🔨

1. Go to 'Distribution > Outbound shipments'
2. Click on 'Filters' and choose "Created date"
3. Choose a from date and make sure the To date is the same a some outbound shipment created date

## Your environment 🌱

- Open mSupply Version: v2.09-RC-07111524
- Legacy mSupply Central Server Version: 8.03

- Platform: Android

- Database type: SQLite

aimee-mcneil-melville:
Fixes #8533

# 👩🏻‍💻 What does this PR do?

Filters with a date correctly use either date or dateTime filter types

Filters with dateTime (often a created date or similar) works with timezones, so that the resulting records (with localised dates) match the filtered date selection. Several filters that were using date have now been updated to dateTime

Filters where only a date is needed (eg expiry, DOB) use 'date' only as they have no time element

Some refactoring for more robust date type handling, and clarity on handling for showTime. Comments to help developers choose the correct date type when implementing or updating filters 

For the second part of the issue, selecting the dates all worked correctly for me and I couldn't reproduce the behaviour shown in the second video. Possibly this is also related to timezones? 

# 🧪 Testing

- [ ] Create some outbound shipments. These will need to be created in timeframe where the timezone will create a difference between the local date and the UTC date. 
- Eg in NZ create them in the morning - a shipment saved on 20/8 at 10.46am local time is recorded in the database in UTC time, 19/08 10.48pm. 
This can be checked in the graphql or database to confirm the UTC date 
Screenshot 2025-08-20 at 6 41 56 PM

- [ ] Filter by the local date created, all shipments where the Created date matches the filter will be in the list (also showing local date)
- [ ] Test that other list view date filters also behave correctly:
- Inbound shipment
- Prescriptions
- Stocktakes
- Purchase orders
- Encounter
- Item -> Item Ledger
- Temperature Breach & Temperature Log (these two also have time picker)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/Toolbar.tsx
+++ b/client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/Toolbar.tsx
@@ -40,12 +40,14 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
               elements: [
                 {
                   type: 'dateTime',
+                  showTimepicker: true,
                   name: t('label.from-start-datetime'),
                   urlParameter: 'datetime',
                   range: 'from',
                 },
                 {
                   type: 'dateTime',
+                  showTimepicker: true,
                   name: t('label.to-start-datetime'),
                   urlParameter: 'datetime',
                   range: 'to',
--- a/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -49,7 +49,6 @@ export const DateTimePickerInput = ({
   actions,
   dateAsEndOfDay,
   disableFuture,
-  displayAs,
   error,
   required,
   textFieldSx: inputSx,
@@ -68,7 +67,6 @@ export const DateTimePickerInput = ({
   actions?: PickersActionBarAction[];
   dateAsEndOfDay?: boolean;
   disableFuture?: boolean;
-  displayAs?: 'date' | 'dateTime';
   required?: boolean;
   textFieldSx?: SxProps;
 }) => {
@@ -79,7 +77,6 @@ export const DateTimePickerInput = ({
   const t = useTranslation();
   const format =
     props.format === undefined ? (showTime ? 'P p' : 'P') : props.format;
-  const displayDt = displayAs === 'date';
 
   const isDesktop = useMediaQuery('(pointer: fine)');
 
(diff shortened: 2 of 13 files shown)


client/packages/coldchain/src/Monitoring/ListView/TemperatureBreach/Toolbar.tsx
```diff
@@ -40,12 +40,14 @@ export const Toolbar: FC<{ filter: FilterController }> = () => {
               elements: [
                 {
                   type: 'dateTime',
+                  showTimepicker: true,
```
aimee-mcneil-melville, line 43: Turning timePicker on as an option now instead of by default - only used in a couple of places

client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
```diff
@@ -79,7 +77,6 @@ export const DateTimePickerInput = ({
   const t = useTranslation();
   const format =
     props.format === undefined ? (showTime ? 'P p' : 'P') : props.format;
-  const displayDt = displayAs === 'date';
```
aimee-mcneil-melville, line 82: use the showTime bool instead

client/packages/common/src/ui/components/inputs/Filters/DateFilter.tsx
```diff
@@ -44,10 +42,15 @@ export const DateFilter = ({
         return;
       }
 
-      const date =
-        range === 'to' && displayAs === 'date' // if no time picker, set "TO" field to end of day
-          ? DateUtils.endOfDay(selection)
-          : selection;
+      let date = selection;
+
+      if (!showTimepicker && type === 'dateTime') {
```
lache-melvin, line 47: this one might be helped by a little comment... something like "when user is only selecting the date, ensure our search filter is inclusive of start and end day" ?

PR #8958, opened 2025-08-21
Make Enum filters (status, type) clearable

### LINKED ISSUE #8531: The item ledger filters should have the X button to stop the filtering
## The suggestion
Currently when using a filter in the item ledger, there is no way to stop the filter unless you select another value to filter by or click on the button remove all filter. It will be good to have a X button to stop filtering like in filters in the Item or view Stock windows. Maybe it is because the filters have values to choose by instead of free text like in the Item or View Stock windows. 

https://drive.google.com/file/d/1Ch0oFVcmKhMdYoEAGiUgGjqV1hdkbNdG/view?usp=sharing

When using the type filter for example, I can never stop the filtering. I can only go for one value to the other. I can only stop filtering by going to the Filters box and click on "Remove all filters".  It would be good if I could click on a X in the active filter box to stop filtering by that filter.

Image

I think it makes the experience more enjoyable to be able to easily return to the state that I could see all records after using the filtering

I don't think so

I don`t know

aimee-mcneil-melville:
Fixes #8531

# 👩🏻‍💻 What does this PR do?

Allow any enum filters (status, type etc) clearable without removing the filter. This applies throughout the app

Screenshot 2025-08-21 at 1 12 52 PM

The issue suggested adding `X` in the corner, the base component that the enum selection uses already had functionality for the `Clear selection` option. This is also consistent design with the filter menu 

I noticed that in the Item Ledger Filter, after clearing I was unable to immediately scroll in the list. This was not an issue elsewhere. Had a look into fixing it but didn't find a solution so far - likely cause it that it is in a tab?

# 🧪 Testing

- [ ] Go to Items -> select an item -> Ledger tab
- [ ] Use the Status and Type filters
- [ ] Be able to remove selections from these filters

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/Filters/EnumFilter.tsx
+++ b/client/packages/common/src/ui/components/inputs/Filters/EnumFilter.tsx
@@ -39,6 +39,7 @@ export const EnumFilter: FC<{
       label={name}
       value={value ?? ''}
       onChange={e => handleChange(e.target.value)}
+      clearable
     />
   );
 };


lache-melvin (approved):
TY!

> I noticed that in the Item Ledger Filter, after clearing I was unable to immediately scroll in the list. This was not an issue elsewhere. Had a look into fixing it but didn't find a solution so far - likely cause it that it is in a tab?

Hm yeah looks like there is a minuscule amount of scroll on the table and that stops the overall tab scrolling or something 👀 Ah well, I think probably a rabbit hole not worth going down for now :)

PR #8964, opened 2025-08-21
8952 volume per pack not updating

### LINKED ISSUE #8939: Changing 'Received Pack Size' in Inbound Shipments does not trigger a change in Volume per pack (m³)
## What went wrong? 😲

When adding an item to an Inbound Shipment, you can add per item variant. These item variants can have different pack sizes - and different volumes associated with them. Changing the pack size does not trigger a change in volume per pack as per video 

https://github.com/user-attachments/assets/3d8db2ee-b22d-4ba9-a239-59117abbe3a7

## Expected behaviour 🤔

Should be updating the volume per pack field if pack size has changed to match the item variant's one

## How to Reproduce 🔨

1. Have an item set up in omSupply with the default pack size = 1 

Image

2. For that item, have two different item variants set up. One item variant should have a matching pack size (e.g. 1), whereas another item variant should have a pack size = 2. Have volumes set up for them both too

Image

3. Go to Replenishment -> Inbound Shipments and create a new Inbound Shipment for the item you are testing with. `When selecting an item variant, volume will only populate if the pack size is matching the item variant's pack size.` This is important!
4. Add an item variant to the shipment. Select item variant with a pack size 1 = volume populated. 
5. Select an item variant with a pack size 2 = volume resets to 0. This is expected as the pack size is not matching the item variant's pack size. 
6. However, change the pack size to 2 -> volume field is still sitting at 0. This needs to be updating!

## Your environment 🌱

- Open mSupply Version: `2.10.00-RC-08181654`

- Database type: `PostgreSQL`

### LINKED ISSUE #8952: Volume per pack (m³) does not populate when creating a return from an outbound shipment
## What went wrong? 😲

I do not know what is causing the issue so just including all the details here: 

- I have an item and its default pack size is **1**
- I have 2 item variants for this item
     - Item variant 1: pack size 1 - **matching the item**
     - Item variant 2: pack size 2 - **does not match the item**

I created an Outbound Shipment with an item's batch that has an item variant assigned with a pack size of 2. I then create a return from that Outbound Shipment, and the volume per pack does not populate even though volume per pack is defined for that item variant. 

https://github.com/user-attachments/assets/59186da4-13f0-440e-a7b2-30aca12b9cc9

## Expected behaviour 🤔

I think, if our item variant has a pack size defined and a volume per pack that is matching that pack size, the field should populate when creating a return?

## How to Reproduce 🔨

1. Have an item and its default pack size is **1**
2. Have 2 item variants for this item
     - Item variant 1: pack size 1 - **matching the item**
     - Item variant 2: pack size 2 - **does not match the item**
3. Go to `Distribution` -> `Outbound Shipments` and create a new shipment with the batch and the item variant that is different to your item. I was testing with **item variant 2 (pack size 2)**
4. Change the status of the shipment to `Verified`
5. Now, select that same item in the shipment and in the footer click on `Return selected lines`
6. You will see a modal for the return - but the total volume per pack is empty
7. Even if you change around the item variants and include matching pack sizes, the field will still stay empty and can only be manually populated

## Your environment 🌱

- Open mSupply Version: `V2.10.00-RC-08201651`

- Database type: `PostgreSQL`

lache-melvin:
Fixes #8952

Fixes #8939

# 👩🏻‍💻 What does this PR do?

Now, when updating pack size anywhere that also has a volume per pack input:

- If there is a selected item variant
- If that variant has a packaging variant which matches the new pack size
- The volume per pack will update according to that packaging variant
- Otherwise volume per pack is cleared

When changing the variant, if no pack size matches, volume per pack should be cleared.

# 🧪 Testing

- [ ] In stocktake line edit
- [ ] In customer return line edit
- [ ] In inbound shipment line edit
- [ ] In view stock:

The following should occur:
- If there is a selected item variant
- If that variant has a packaging variant which matches the new pack size
- The volume per pack will update according to that packaging variant
- Otherwise volume per pack is cleared

When changing the variant, if no pack size matches, volume per pack should be cleared.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnQuantitiesTable.tsx
+++ b/client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnQuantitiesTable.tsx
@@ -12,6 +12,7 @@ import {
   useTranslation,
 } from '@openmsupply-client/common';
 import {
+  getVolumePerPackFromVariant,
   ItemVariantInputCell,
   useIsItemVariantsEnabled,
 } from '@openmsupply-client/system';
@@ -55,23 +56,12 @@ export const QuantityReturnedTableComponent = ({
         key: 'itemVariantId',
         label: 'label.item-variant',
         width: 170,
-        setter: patch => {
-          const { packSize, itemVariant } = patch;
-
-          if (itemVariant) {
-            const packaging = itemVariant.packagingVariants.find(
-              p => p.packSize === packSize
-            );
-            // Item variants save volume in L, but it is saved in m3 everywhere else
-            updateLine({
-              ...patch,
-              volumePerPack:
-                ((packaging?.volumePerUnit ?? 0) / 1000) * (packSize ?? 1),
-            });
-          } else {
-            updateLine(patch);
-          }
-        },
+        setter: patch =>
+          updateLine({
+            ...patch,
+            volumePerPack:
+              getVolumePerPackFromVariant(patch) ?? patch.volumePerPack,
+          }),
         Cell: props => (
           <ItemVariantInputCell {...props} itemId={props.rowData.item.id} />
         ),
@@ -95,7 +85,12 @@ export const QuantityReturnedTableComponent = ({
       ],
       getColumnLookupWithOverrides('packSize', {
         Cell: PackSizeEntryCell<GenerateCustomerReturnLineFragment>,
-        setter: updateLine,
+        setter: patch =>
+          updateLine({
+            ...patch,
+            volumePerPack:
+              getVolumePerPackFromVariant(patch) ?? patch.volumePerPack,
+          }),
         getIsDisabled: () => isDisabled,
         label: 'label.pack-size',
       })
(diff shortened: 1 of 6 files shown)


client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnQuantitiesTable.tsx
```diff
@@ -131,7 +124,7 @@ export const QuantityReturnedTableComponent = ({
         key: 'volumePerPack',
         label: t('label.volume-per-pack'),
         Cell: NumberInputCell,
-        cellProps: { decimalLimit: 2 },
+        cellProps: { decimalLimit: 10 },
```
lache-melvin, line 127: Fix for vol per pack not showing in returns modal!

PR #8970, opened 2025-08-21
8963 outbound shipment report

### LINKED ISSUE #8963: Outbound shipment report
- numeric columns should be right-aligned
- if you run for all customers, there is no column showing the customer
- if you do choose a customer, run the report, then click the filter again
   a. the customer you chose isn't shown as the default
   b. there is no way to run for all customers (leaving the field blank still runs for the previously chosen customer)

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #8963

# 👩🏻‍💻 What does this PR do?
- Align numeric columns to the right
- Add a column showing what customer the line is for 
- Populate the filter selection & handle clearing properly
- Total should be grouped by item AND other party

https://github.com/user-attachments/assets/5ad1a86e-2994-4df2-bb02-24657b009be4

# 🧪 Testing

- [ ] Go to reports
- [ ] Click on Outbound report
- [ ] All numbers should be right aligned
- [ ] Click to change filters
- [ ] If you have selected a customer then the customer name should be shown in this modal by default
- [ ] Clear the customer name
- [ ] Reports should be regenerated with all customers
- [ ] select customer
- [ ] report should be generated with info only from that customer

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


standard_reports/outbound_shipments/latest/src/template.html
```diff
@@ -26,23 +21,23 @@ <h3>{{data.otherPartyName}}</h3>
     <tbody>
       {% for line in data.lines %}
       <tr>
+        <td>{{line.otherPartyName}}</td>
```
lache-melvin, line 24: nit: seeing as it is sorted by item and then customer, I think I'd expect this column after the item name 👀

client/packages/system/src/Name/Components/CustomerSearchInput/CustomerSearchInput.tsx
```diff
@@ -17,23 +17,36 @@ export const CustomerSearchInput = ({
...
+  clearable = false,
+  currentId = undefined,
 }: NameSearchInputProps) => {
   const t = useTranslation();
   const { data, isLoading } = useName.document.customers();
   const [buffer, setBuffer] = useBufferState(value);
   const NameOptionRenderer = getNameOptionRenderer(t('label.on-hold'));
 
+  // For use in JSON forms
+  useEffect(() => {
```
lache-melvin, line 29: if you refresh on the generated report page, then open the filter modal, customer isn't there, I think this one needs dependency on `data`?

Though seeing as you had to duplicate this for both customer and supplier... I'm tempted to say that NameSearch json form component could just do the name query? And if it gets the result, pass that in here as `value`?

PR #8972, opened 2025-08-21
Clear if received pack size changed not shipped

### LINKED ISSUE #8971: When updating the default pack size in an inbound shipment, default sell price does not reset
## What went wrong? 😲

When an item has both default pack size and default sell price defined, adding it to an inbound shipment will result in both cells being pre-populated. However, when changing the default pack size to something else, default sell price does not reset (it should)

https://github.com/user-attachments/assets/cd25b681-e323-43fe-a568-6842fbf921f5

## Expected behaviour 🤔

Default sell prices should reset if the default pack size is changed to something else

## How to Reproduce 🔨

1. On the video above, I have an item - in OG mSupply, default pack size is set to 2 and default sell price is set to $5
2. In omSupply, add that item to an Inbound Shipment - you will see both cells pre-populated with the values you have defined
3. Change the default pack size to something else - note default sell price does not reset (it should)

## Your environment 🌱

- Open mSupply Version: `V2.10.00-RC-08211216`

- Database type: `PostgreSQL`

roxy-dao:
Fixes #8971

# 👩🏻‍💻 What does this PR do?
Was clearing in shipped pack size but with received pack size variable in the check... have moved this to received pack size. Not we're not re-populating the default sell price back if the pack size matches default... not sure if this is something we want and probably very edge case

# 🧪 Testing

- [ ] Have default sell price set for an item
- [ ] Create an Inbound Shipment 
- [ ] Add item to inbound shipment 
- [ ] Change received pack size
- [ ] Sell price should clear

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -112,7 +112,6 @@ export const QuantityTableComponent = ({
       width: 100,
       Cell: PackSizeEntryCell<DraftInboundLine>,
       setter: patch => {
-        setPackRoundingMessage?.('');
         updateDraftLine(patch);
       },
       getIsDisabled: rowData => !!rowData.linkedInvoiceId,
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -113,19 +113,7 @@ export const QuantityTableComponent = ({
       Cell: PackSizeEntryCell<DraftInboundLine>,
       setter: patch => {
         setPackRoundingMessage?.('');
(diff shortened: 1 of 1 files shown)


client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -146,10 +134,15 @@ export const QuantityTableComponent = ({
     getColumnLookupWithOverrides('packSize', {
       Cell: PackSizeEntryCell<DraftInboundLine>,
       setter: patch => {
-        setPackRoundingMessage?.('');
```
lache-melvin, line 149: should this have been removed? probably makes more sense here than on the shipped pack size right?

lache-melvin (changes requested):
works gooood - just a q about the warning clearing..

lache-melvin (approved):
🙏

PR #8999, opened 2025-08-25
8993 cant save stocktake with on hold location

### LINKED ISSUE #8993: Once a stocktake was attempted to be finalised with a location that is On Hold, error persists even when the location is updated
## What went wrong? 😲

So, running into the following error: 

Image

Image

Image

This has happened when a stockline was allocated to a location that was on hold -> I have since removed the location from hold, but the error persisted. I updated the location to a new one all together but still couldn't save my stocktake. 

~~Also, related bug number 2 - if a location is on hold, it should be removed from the `Available` column so it doesn't accidentally get selected~~

Image

## Expected behaviour 🤔

Even if the user is accidentally allocating stock to a location that is on hold, they should be able to either remove the location from hold or update it - and the stocktake in this case should be saved.

~~Also, on hold locations should be removed from the Available column :)~~

## How to Reproduce 🔨

1. Have a location that is on hold
2. Go to `Inventory` -> `Stocktakes` and create a new stocktake
3. Add a new item -> in `Other` tab, note you can choose your location that is on hold from the `Available` column
4. Proceed to select it -> then, try to Save the stocktake -> you will see an error (expected ✅)
5. Now, remove that location off hold -> sync -> refresh the page -> the error is still there (shouldn't be 🚫 )
6. In your stocktake, change the location to something else entirely or remove the location
7. You are still unable to save the Stocktake 🚫 

## Your environment 🌱

- Open mSupply Version: `2.10.00-RC-08221718`

- Database type: `PostgreSQL`

lache-melvin:
Fixes #8993

# 👩🏻‍💻 What does this PR do?

When doing a stock reduction - we were doing the location validation based where the stock line _currently_ was, rather than where we are moving it to in the stocktake.

# 🧪 Testing

Note - this only applies to stock reduction - matching count and increasing stock should work successfully either way.

- [ ] Have stock which is NOT in an on hold location
- [ ] Create a stocktake including that stock line
- [ ] Cause a stock reduction - set counted packs to less than snapshot
- [ ] Set the location to an on hold location
- [ ] Save & finalise stocktake - you get an error that location is on hold
- [ ] Have stock in an on hold location
- [ ] Create a stocktake including that stock line
- [ ] Cause a stock reduction - set counted packs to less than snapshot
- [ ] Clear the location/set to a not-on-hold location
- [ ] Save & finalise stocktake - saves successfully

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/invoice_line/stock_out_line/insert/mod.rs
+++ b/server/service/src/invoice_line/stock_out_line/insert/mod.rs
@@ -123,7 +123,7 @@ pub fn insert_stock_out_line(
 mod test {
     use repository::{
         mock::{
-            mock_item_a, mock_item_b_lines, mock_name_store_a,
+            mock_item_a, mock_item_b_lines, mock_location_on_hold, mock_name_store_a,
             mock_outbound_shipment_a_invoice_lines, mock_outbound_shipment_c,
             mock_outbound_shipment_c_invoice_lines, mock_patient, mock_prescription_a,
             mock_stock_line_a, mock_stock_line_location_is_on_hold, mock_stock_line_on_hold,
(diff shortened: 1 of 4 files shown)


roxy-dao (approved):
Thanks. Not part of your changes, but there's double toast showing up for the error
Screenshot 2025-08-26 at 8 35 08 AM

lache-melvin:
Hm yeah would like to look at that, maybe with usability work, if there is some capacity for handling errors caught in different places and reporting them better!

PR #9000, opened 2025-08-25
Fix Inbound Form Displaying Broken Logo

### LINKED ISSUE #8613: Reports showing broken logo
## What went wrong? 😲

Image

There is no requirement to set an image as the logo for a store in mSupply, so it is optional. All our report/form templates should conditionally include the img html.

## Expected behaviour 🤔
[ MP edited ]
The following standard forms need to be updated
- [ ] inbound
- [ ] internal-order
- [ ] outbound-invoice-landscape-with-logo
- [ ] outbound-invoice-portrait-with-logo
- [ ] repack
- [ ] requisition
- [ ] stock-take-detail-view
- [ ] stock-take-variance
- [ ] stock-take-with-quantity
- [ ] stock-take-without-quantity

## How to Reproduce 🔨

1. New store
2. OMS
3. Print stocktake form (or presumably many other reports)

## Your environment 🌱

- Open mSupply Version: 2.9.1-RC dev but seen it a couple weeks ago too

zachariah-at-msupply:
Fixes #8613 

# 👩🏻‍💻 What does this PR do?

Adds a filter to the template so that if there is no logo, the logo section is not included in the printed report.
Also adds some spacing so that the report will have consistent spacing whether or not the logo is displayed.

# 🧪 Testing

- [ ] Download [this database](https://drive.google.com/file/d/1j7UcTLoCVj4aNJ0eWA2xCX_CwXx8iOmm/view?usp=drive_link) and put it in open-msupply/server.
- [ ] Download [this file](https://drive.google.com/file/d/16Sg7pnYvP7ziXgRetYe2Z1Rk-DtsYy_k/view?usp=drive_link) and put it in open-msupply/standard_reports. If you already have a file of the same name in there, replace it with this one.
- [ ] Run the server.
- [ ] With open-msupply/server as your working directory, run this command (with the file paths that make sense for your environment) `./target/debug/remote_server_cli show-report --path /Users/zachariah/code/open-msupply/standard_forms/purchase-order/latest --config /Users/zachariah/code/open-msupply/standard_reports`
- [ ] The report should display in your browser. Check it has no broken images displaying and that the spacing looks good.

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/standard_forms/inbound/latest/src/header.html
+++ b/standard_forms/inbound/latest/src/header.html
@@ -1,48 +1,65 @@
-{% macro dateOrNA(datetime) %}
-{% if datetime %}{{ datetime | date(format="%d/%m/%Y", timezone=arguments.timezone) }}{% else
-%}{{t(k="messages.not-applicable", f="N/A")}}{% endif %}
-{% endmacro input %}
+{% macro dateOrNA(datetime) %} {% if datetime %}{{ datetime |
+date(format="%d/%m/%Y", timezone=arguments.timezone) }}{% else
+%}{{t(k="messages.not-applicable", f="N/A")}}{% endif %} {% endmacro input %}
 
-<table class="header_image_section" style="width: 100%;height: 98%;">
-    <tr>
-        <th style="text-align:left; width:10%;height:70%;">
-            <img class="logo" src="{{data.store.logo}}" />
-        </th>
-        <th style="text-align:left;width:20%">
-            <span style="font-size: 12pt; font-weight: bold;">{{ data.store.storeName }}</span><br>
-            <span style="font-size: 10pt;">{{ data.store.name.address1 }}</span><br>
-            <span style="font-size: 10pt;">{{ data.store.name.address2 }}</span><br>
-            <span style="font-size:6pt;">
-                {{t(k="label.phone",f="Telephone")}}: {{ data.store.name.phone }}
-            </span><br>
-            <span style="font-size:6pt;">
-                {{t(k="label.email",f="Email")}}: {{ data.store.name.email }}
-            </span>
-        </th>
-        <th style="text-align:right;width:70%;font-size:28pt;">{{ t(k="report.inbound-form", f="Inbound Shipment Form")
-            }}</th>
-    </tr>
+<table class="header_image_section" style="width: 100%; height: 98%">
+  <tr>
+    {% if data.store.logo %}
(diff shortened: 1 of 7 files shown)


zachariah-at-msupply:
I will wait for a review of the first three to check whether the pattern is sound before applying it to other reports (it seems like the same pattern can be used for all the other reports).

jmbrunskill (changes requested):
@zachariah-at-msupply I think we can leave this as a modification to the `latest` report.
We only need to bump the major version to say 2.10 if we're relying on something in graphql or frontend code that's only available in 2.10.
For this logo fix, i should be backwards compatiable right? it is just a tera check for the logo to exist?

standard_forms/inbound/latest/src/header.html
```diff
@@ -1,48 +1,65 @@
...
-        <th style="text-align:right;width:70%;font-size:28pt;">{{ t(k="report.inbound-form", f="Inbound Shipment Form")
-            }}</th>
-    </tr>
+<table class="header_image_section" style="width: 100%; height: 98%">
+  <tr>
+    {% if data.store.logo %}
+    <th style="text-align: left; width: 10%; height: 70%">
+      <img class="logo" src="{{data.store.logo}}" />
+    </th>
+    {% endif %}
```
jmbrunskill, line 11:
```suggestion
    <th style="text-align: left; width: 10%; height: 70%">
        {% if data.store.logo %}
      <img class="logo" src="{{data.store.logo}}" />
          {% endif %}
    </th>
```

Why not this?

jmbrunskill: I'm guessing this is ok as we're not using any table rows here, I take it back...

zachariah-at-msupply:
Ah yes, my apologies I forgot that the format-on-save option can make diffs difficult.
That's correct thought, the only changes are:
- adding the {% if %} {% endif %} block
- adding two <br/> tags later on to preserve spacing when the logo's not present.

Just wondering, if that makes things clear? Or still best to re-do the changes without formatting? (I think it may take about 30 mins in case that's helpful info).

jmbrunskill (approved):
Thanks @zachariah-at-msupply looks good to me!

PR #9024, opened 2025-08-26
8928 ledger fix invoice status from log

### LINKED ISSUE #8928: Find and correct invoice status for ledger fix process
## What went wrong? 😲

Due to a bug in the OG central server some invoice would revert status. E.g.

1. Inbound shipment created, shipped, stock adjusted, and all synced to central server
2. OG central bug causes inbound shipment in OMS to revert back to "new"

Now the stock has been adjusted, but the invoice_lines for that adjustment are no longer in the running balance for that stock as their in a "new' invoice.

This applies for all invoice types, but the solution is much the same. Check the logs and correct the status.

## Expected behaviour 🤔

Our ledger fix processor also has a fix/check that looks up the invoices of the lines that affected stock_line being assessed. If the invoice's status does not match the latest status recorded in the activity log then update the record:

Image

## How to Reproduce 🔨

This is a bit tricky as we've already fixed the the root cause of the problem in OG, however there may records that are still affected.

1. Finalise an inbound shipment
2. Edit DB forcing the status back to "new"
3. Check ledger - bad. SOH > running balance.

## Your environment 🌱

- Open mSupply Version: <2.10
- Legacy mSupply Central Server Version: <8.3.15 (though old data may be effected so version isn't 100% reliable)

Chris-Petty:
Fixes #8928 

# 👩🏻‍💻 What does this PR do?

- our discrepancy sql now includes stocklines that have no associated invoice_lines
- The ledger fix will use activity log to update the status of an invoice to eliminate discrepancies
  - if users have reprocessed a record, moving it's status forward again, then the fix will skip
  - will also update the datetime fields of the record based on the corresponding status change logs

# 🧪 Testing

- [ ] in OMS on a clean store, add a inbound shipment with some stock of itemA and verify it
- [ ] sync
- [ ] delete your OMS db
- [ ] in OG revert the status of the invoice back to "new" or "allocated"
- [ ] reinitialise your OMS site
- [ ] The invoice is updated back to "verified"
- [ ] Ledger is nice
- [ ] system_log reports the applied ledger fix

Here is a 4D script that can revert the status of the last invoice added to the database (incl via sync)

```4dm
var $eInvoice : cs.transactEntity
var $esInvoices : cs.transactSelection
$esInvoices:=ds.transact.all()
For each ($eInvoice; $esInvoices.slice(-1))
	$eInvoice.status:="nw"
	$eInvoice.confirm_date:=Null
	$eInvoice.confirm_time:=Null
	$eInvoice.ship_date:=Null
	$eInvoice.arrival_date_estimated:=Null
	$eInvoice.arrival_date_actual:=Null
	$eInvoice.om_allocated_datetime:=Null
	$eInvoice.om_picked_datetime:=Null
	$eInvoice.om_shipped_datetime:=Null
	$eInvoice.om_delivered_datetime:=Null
	$eInvoice.om_verified_datetime:=Null
	$eInvoice.om_status:="NEW"
	$eInvoice.save()
End for each 
```

# 📃 Documentation

Great question...

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
+++ b/server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
@@ -192,9 +192,9 @@ pub(crate) fn fix(
                                 continue;
                             }
                         };
+                        operation_log.push_str(&format!("Adjusting invoice {invoice_id} status {invoice_status:?} to match latest log status: {log_status:?}.\n", invoice_id = invoice.id));
+                        InvoiceRowRepository::new(connection).upsert_one(invoice)?;
                     }
-                    operation_log.push_str(&format!("Adjusting invoice {invoice_id} status {invoice_status:?} to match latest log status: {log_status:?}.\n", invoice_id = invoice.id));
-                    InvoiceRowRepository::new(connection).upsert_one(invoice)?;
                 }
                 _ => continue,
             }
@@ -219,7 +219,8 @@ mod test {
         InvoiceStatus, KeyValueStoreRepository, StockLineRow,
     };
 
-    fn mock_data() -> MockData {
+    #[actix_rt::test]
+    async fn test_ledger_fix_adjust_invoice_status() {
         // No status change so should not be fixed by this ledger fix
         let total_does_not_match = StockLineRow {
             id: "total_does_not_match".to_string(),
@@ -254,10 +255,15 @@ mod test {
                 .iter()
                 .flat_map(invoice_generate_logs)
                 .collect();
-        an_invoice_where_status_changed_movements.invoices[0].status = InvoiceStatus::Allocated;
-        an_invoice_where_status_changed_movements.invoices[0].delivered_datetime = None;
-        an_invoice_where_status_changed_movements.invoices[0].received_datetime = None;
-        an_invoice_where_status_changed_movements.invoices[0].verified_datetime = None;
+        let invoice = &mut an_invoice_where_status_changed_movements.invoices[0];
+        let invoice_status_changed_id = &invoice.id.clone();
+        let old_delivered_datetime = invoice.delivered_datetime.clone();
+        let old_received_datetime = invoice.received_datetime.clone();
+        let old_verified_datetime = invoice.verified_datetime.clone();
+        invoice.status = InvoiceStatus::Allocated;
+        invoice.delivered_datetime = None;
+        invoice.received_datetime = None;
+        invoice.verified_datetime = None;
 
         // Invoice changed status, but user changed status again creating duplicate in logs. Should not get fixed.
         let an_invoice_where_status_changed_duplicate_status = StockLineRow {
@@ -274,17 +280,19 @@ mod test {
             .iter()
             .flat_map(invoice_generate_logs)
             .collect();
-        duplicate_status_movements.invoices[0].status = InvoiceStatus::Delivered;
-        duplicate_status_movements.invoices[0].received_datetime = None;
-        duplicate_status_movements.invoices[0].verified_datetime = None;
+        let invoice = &mut duplicate_status_movements.invoices[0];
+        let invoice_duplicate_status_id = &invoice.id.clone();
+        invoice.status = InvoiceStatus::Delivered;
+        invoice.received_datetime = None;
+        invoice.verified_datetime = None;
         duplicate_status_movements.activity_logs.push({
             ActivityLogRow {
                 id: "duplicate_delivered_status".to_string(),
-                ..duplicate_status_movements.activity_logs[1].clone() // Should be the delivered log from the first time
+                ..duplicate_status_movements.activity_logs[1].clone() // Should be the delivered log from the first time.
             }
         });
 
-        MockData {
+        let mock_data = MockData {
             stock_lines: vec![
                 total_does_not_match.clone(),
                 an_invoice_where_status_changed.clone(),
@@ -294,15 +302,12 @@ mod test {
         }
         .join(total_does_not_match_movements)
         .join(an_invoice_where_status_changed_movements)
-        .join(duplicate_status_movements)
-    }
+        .join(duplicate_status_movements);
 
-    #[actix_rt::test]
-    async fn adjust_total_to_match_ledger_test() {
         let ServiceTestContext { connection, .. } = setup_all_with_data_and_service_provider(
-            "adjust_total_to_match_ledger",
+            "test_ledger_fix_adjust_invoice_status",
             MockDataInserts::none().names().stores().units().items(),
-            mock_data(),
+            mock_data,
         )
         .await;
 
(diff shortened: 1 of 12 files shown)


server/service/src/test_helpers.rs
```diff
@@ -131,30 +132,43 @@ pub(crate) fn make_movements(stock_line: StockLineRow, date_quantity: Vec<(i64,
...
-                    created_datetime: date,
-                    allocated_datetime: Some(date),
-                    picked_datetime: Some(date),
-                    shipped_datetime: Some(date),
-                    delivered_datetime: Some(date),
-                    received_datetime: Some(date),
-                    verified_datetime: Some(date),
+                    r#type: InvoiceType::InboundShipment,
+                    status: InvoiceStatus::Verified,
+                    created_datetime: date.with_hour(1).unwrap(),
```
Chris-Petty, line 143: These dates generate the corresponding logs. So it is important that they're in the right order so that the latest activity log is actually the latest status in the sequence

server/service/src/ledger_fix/fixes/fix_cancellations.rs
```diff
@@ -119,6 +119,12 @@ pub(crate) mod test {
         with_cancellation.invoices[3] = inline_edit(&with_cancellation.invoices[3], |mut u| {
             u.status = InvoiceStatus::Cancelled;
             u.r#type = InvoiceType::Prescription;
+            u.cancelled_datetime = Some(u.created_datetime.clone());
```
Chris-Petty, line 122: This will create conflicts with develop just need to refactor into whatever syntax copilot did here

server/service/src/ledger_fix/ledger_fix_driver.rs
```diff
@@ -82,7 +82,8 @@ async fn ledger_fix(service_provider: Arc<ServiceProvider>) {
 
     for (index, stock_line_id) in stock_line_ids.iter().enumerate() {
         let mut operation_log = format!(
-            "{index}/{} Fixing stock line {stock_line_id} {}\n",
+            "{}/{} Fixing stock line {stock_line_id} {}\n",
```
Chris-Petty, line 85: if there were 2 lines our logs were

`0/2...`
`1/2...`

I prefer

`1/2...`
`2/2...`

server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
```diff
@@ -0,0 +1,376 @@
...
+    },
+    stock_line_ledger::{StockLineLedgerFilter, StockLineLedgerRepository},
+    ActivityLogRow, ActivityLogType, EqualFilter, InvoiceLineFilter, InvoiceLineRepository,
+    InvoiceRow, InvoiceRowRepository, InvoiceStatus, Pagination, StockLineRowRepository,
+    StorageConnection,
+};
+
+use crate::ledger_fix::{fixes::LedgerFixError, ledger_balance_summary, LedgerBalanceSummary};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)] // Wanted to use InvoiceStatus, but need a good "other" variant
```
Chris-Petty, line 13: Note derive `Ord` and `PartialOrd` here means the order of the variants in the enum definition here is important. Without extensive unit tests it'd be pretty easy to break if innocently adding a status...

server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
```diff
@@ -0,0 +1,376 @@
...
+            InvoiceStatus::Allocated => Status::Allocated,
+            InvoiceStatus::Picked => Status::Picked,
+            InvoiceStatus::Shipped => Status::Shipped,
+            InvoiceStatus::Received => Status::Received,
+            InvoiceStatus::Delivered => Status::Delivered,
+            InvoiceStatus::Verified => Status::Verified,
+            InvoiceStatus::Cancelled => Status::Cancelled,
+        }
+    }
+}
```
Chris-Petty, line 55: Lemme know if there is an elegant alternative!

server/service/src/test_helpers.rs
```diff
@@ -182,3 +196,85 @@ pub(crate) fn make_movements(stock_line: StockLineRow, date_quantity: Vec<(i64,
...
+    let mut logs = Vec::new();
+
+    let created_log = ActivityLogRow {
+        id: format!("{}_created", invoice.id),
+        r#type: ActivityLogType::InvoiceCreated,
+        user_id: Some("user_account_a".to_string()),
+        store_id: Some(invoice.store_id.clone()),
+        record_id: Some(invoice.id.clone()),
+        datetime: invoice.created_datetime.clone(),
+        changed_to: None,
```
andreievg, line 210: Should we use ..Default instead of 'None' here ? 

I've been thinking that all test should use ..Default (to reduce number of places to change when new fields is added)

server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
```diff
@@ -0,0 +1,376 @@
...
+        operation_log.push_str(&format!(
+            "Ledger does not match use case as for adjust_invoice_status- there are no related invoice lines\n",
+        ));
+        return Ok(());
+    }
+
+    let mut invoices: Vec<InvoiceRow> = invoice_lines.into_iter().map(|l| l.invoice_row).collect();
+    invoices.sort_by(|a, b| a.id.cmp(&b.id));
+    invoices.dedup_by(|a, b| a.id == b.id);
+
```
andreievg, line 110: InvoiceRepository actually has stock_line_id filter: 

https://github.com/msupply-foundation/open-msupply/blob/c40979b9ff453952ab70af14fa3f7eb258647331/server/repository/src/db_diesel/invoice.rs#L291-L297

server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
```diff
@@ -0,0 +1,376 @@
...
+    'invoice_loop: for invoice in &mut invoices {
+        let logs = ActivityLogRepository::new(connection).query(
+            Pagination::all(),
+            Some(ActivityLogFilter::new().record_id(EqualFilter::equal_to(&invoice.id))),
```
andreievg, line 114: You can add a filter to equal_any for statuses below

server/service/src/ledger_fix/fixes/adjust_invoice_status.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        KeyValueStoreRepository::new(&connection)
+            .set_i32(
+                repository::KeyType::SettingsSyncSiteId,
+                Some(mock_store_a().site_id),
+            )
+            .unwrap();
+
+        let mut logs = String::new();
+        let invoice_id = &total_does_not_match.id;
```
andreievg, line 322: stock_line_id

andreievg (changes requested):
I got a bit stuck on all of the mappings:

1 - We've previously used status 'order' and used that as comparison in validate/generate logic in service, this was super seeded by doing 'match' statement.
2 - There is a check a few times for correct ActivityLogType, I don't think it's needed, we can filter out by ActivityLogType in the repo filter. 
3 - Can filter invoice by stock line

> Had a thought while cycling home yesterday...

I think that should be ok as per your points (we should try to change status, even if invoice was changed, it's rare enough)

4 - Skip check if outbound is not in new or allocated
5 - Skip check if inbound is in received or verified

Lastly, I am worried about the complexity here, to do the duplicate check 'properly'

`outbound`
* two 'picked' status logs
* two 'verified' status logs
* verified status log datetime < picked status datetime
* picked status in invoice > verified or picked status in log

Similar for inbound.

Since this issue was fixed, and it's pretty rare, is it ok to ignore this use case completely ?

PR #9046, opened 2025-08-26
9045 local date fix

### LINKED ISSUE #9045: localDate doesn't allow for correct sort and comparison order in plugins
When plugins use the common function `localDate` it returns a date like this...

`2025-7-1`
(1st of July 2025)
`2025-7-31`
(31st of July 2025)

However if you want to say sort the dates in the month, or compare them you'll get strange behaviour.

e.g.
```
2025-7-1
2025-7-2
2025-7-3
2025-7-30
2025-7-31
2025-7-4
```
I this also means we can't easily compare which date is later, e.g. `2025-7-4` is after `2025-7-31`

To fix this we need to pad the dates so they look like this...
```
2025-07-01
2025-07-04
2025-07-31
...
```

jmbrunskill:
Fixes #9045
Fixes Internal Issue for Plugin Translations

# 👩🏻‍💻 What does this PR do?

See issue for description of the problem, but the plugin was and is assuming that dates can be sorted and compared, however the format we were using didn't work well for that.

Any reason why this might cause a problem for an existing plugin?

# 🧪 Testing

- [ ] Test with CIV Plugins

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin (approved):
This is pretty standard formatting so I'm happy with this change, makes sense to me!

PR #9057, opened 2025-08-27
8868 sync passwords and permissions

### LINKED ISSUE #8868: Sync passwords and permissions

Corresponding OG issue: https://github.com/msupply-foundation/msupply/issues/17028

When deploying sites it is bad that we're heading towards an SOP that for each device:

1. Initialise (often at a central location with good internet)
2. (new SOP) Login with each user that will use that site (as to make sure their password/permissions are available offline
3. Go on site (often bad internet/offline) for deployment

This had problems, particularly step 2:

- Annoying for internal teams doing deployments to risk hiccup if they don't correctly do a tedious task. Initialising dozens of devices is tedious enough!
- Annoying for external teams for same reasons -> annoying for support too when they raise tickets
- Both internal and external teams have to know their user's passwords to do this BOOOO
- We can fix it!! It's not even hard (potentially famous last words).

- Sync user passwords from OG
- Sync user permissions from OG (user_store)
- OMS integrate the legacy user.password_hash field
- OMS integrate the user permissions from OG 

Suffering until:

- We implement a whole new fan dangled user management system either on the OMS central server or on remote
- Allowing this painful SOP to login every user proceed into existence 

### Additional context 💌

I don't like features that are objectively worse than OG, and in this case OG sets a pretty low bar.

### Acceptance Criteria ✅

- [ ] After initialising OMS, you can immediately go offline and login with any of the users that have access to stores on the site.

### Moneyworks Jobcode 🧰

OMS:DFGEN

Chris-Petty:
Fixes #8868

Requires: https://github.com/msupply-foundation/msupply/pull/18198 in OG

# 👩🏻‍💻 What does this PR do?

- Adds saving of `password_hash` field that is now sent from OG. It is a bcrypt value compatible with OMS.
- Adds `user_store` translator that maps legacy `user_store` records to the appropriate `user_permission` and `user_store_join` records.

# 🧪 Testing

Note: we're constantly turning off the webserver because if it is on, OMS will query the old login API like it used when you login or synchronise, updating the user info and permissions. Half way through writing these tests I switch to having a separate window open with admin user just syncing as it's easier.

<Details>

- [ ] OG and OMS central servers
- [ ] Have 2 stores active on OMS central server site
- [ ] Have an admin user that can login everywhere and do everything
- [ ] Have a new user that can login to both OMS stores
- [ ] (re)initialise the OMS server
- [ ] Turn off the OG web server
- [ ] Login either user, you can (New behaviour)
- [ ] Turn on the OG web server
- [ ] Login admin to OMS
- [ ] OG make a new user2 with login rights to both of the OMS stores (no permissions)
- [ ] Sync OMS (as admin)
- [ ] Turn off OG web server
- [ ] Login with the new user, you can (New behaviour), can't do anything (no permissions), webapp doesn't crash (if it does perhaps raise new issues lol. I saw some oddness when testing with no permissions)
- [ ] Change stores and see the same
- [ ] Turn on the web server
- [ ] Login with admin
- [ ] On OG give user permissions for all the things on the 2 stores
- [ ] Sync OMS
- [ ] Turn off web server
- [ ] Login with new user2 - can and now you have all the permissions (New behaviour)
- [ ] Login to admin in a private window/different browser
- [ ] Turn on web server
- [ ] Remove all login rights for user 2
- [ ] Using admin user window, sync
- [ ] User2 refresh page - all permissions are gone (New behaviour)
- [ ] Turn off web server
- [ ] User2 logout and login - can't (New)
- [ ] Turn on web server
- [ ] Give user2 login rights to both OMS stores
- [ ] Admin user window sync
- [ ] Turn off web server
- [ ] User2 window login - can (New)
- [ ] User2 can change stores
- [ ] Turn on web server
- [ ] Remove login rights to the store that user2 is NOT logged into
- [ ] Admin user window sync
- [ ] User2 window can change store, but they have NO permissions
- [ ] User2 window change back to store they are allowed in
- [ ] User2 can no longer change stores
- [ ] Admin add login rights to both stores again and sync
- [ ] User2 refresh - both stores in selection again
- [ ] GROUPS. Groups control the permissions a user for stores, but login rights is still controlled by the login rights of the user.
- [ ] Create a group with no permissions
- [ ] Add user2 to the group
- [ ] Admin sync
- [ ] User2 now has no permissions again
- [ ] Add some permissions to the group
- [ ] Admin sync
- [ ] User2 now has some permissions
- [ ] Remove user2 from the group
- [ ] Admin sync
- [ ] User2 permissions now back to their own permissions, not that of the group
- [ ] User2 in group again
- [ ] reinitialise
- [ ] turn off sync
- [ ] User2 login - has group permissions

</Details>

# 📃 Documentation

  1. Probably need to explain that users sync now if we explicitly mention having to login with a connection to get user details?

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use repository::{mock::MockDataInserts, test_db::setup_all};
+
+    #[actix_rt::test]
+    async fn test_user_permission_translation() {
+        use crate::sync::test::test_data::user_permission as test_data;
```
Chris-Petty, line 161: This test is copy pasta without modification - I need to write appropriate tests still!

(user_permission test data is for legacy `om_user_permission` table which controls a couple document permissions)

server/service/src/sync/translations/user.rs
```diff
@@ -71,6 +75,8 @@ impl SyncTranslation for UserTranslation {
             None => ("".to_string(), None),
         };
 
+        let hashed_password = password_hash.map_or(hashed_password, |p| p);
```
Chris-Petty, line 78: This was all I thought I needed to do until I remembered permissions 🤣🤣🤣

raviSussol (commented):
@Chris-Petty Do you think we should force logout when a user password is updated (or in the case of wrong authentication) with central server (msupply) from the oms app at the time of syncing or when a user do force syncing?

It looks like the force synchroniser worked successfully but due to failed authentication nothing gets pulled or pushed from the oms - which looks confusion to me...

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,181 @@
...
+        let mut integration_operations: Vec<IntegrationOperation> = Vec::new();
+
+        // Login code may hit OG API if online. If it does, it drops all permissions and regenerates them with new PKs.
+        // There should only be one join per user and store, so we just match on the user and store ids rather than relying on the PK.
+        // If it doesn't exist just upsert a new one using the PK from OG Central.
+        let user_store_join_row = UserStoreJoinRowRepository::new(connection)
+            .find_one_by_user_and_store(&user_id, &store_id)?
+            .map_or_else(
+                || UserStoreJoinRow {
+                    id: uuid(), // generating a new id. The incoming ID might be a user_group_id, OMS does not support these as such so OG may send it multiple times, for each user in group.
```
andreievg, line 76: Would is_default still be for the user ?

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,181 @@
...
+                    is_default,
+                },
+                |r| UserStoreJoinRow { is_default, ..r },
+            );
+
+        let existing_permissions = UserPermissionRepository::new(connection).query_by_filter(
+            UserPermissionFilter::new()
+                .user_id(EqualFilter::equal_to(&user_id))
+                .store_id(EqualFilter::equal_to(&store_id))
+                .has_context(false),
```
andreievg, line 88: Context permissions already sync right ? Via oms permisisons ? Would be good to comment

andreievg (commented):
Looks good to me, I wonder if we can use this translator in the login service.

A little bit unsure about group/user permission sync, assuming if user is in group, permissions from group are synced but with user and store id, with any user specific fields populated, will check mSupply code now.

Next step would be making some time for this in 2.11 ?

server/service/src/sync/translations/user.rs
Chris-Petty: I think it is too much, in the sense it's defensive code against something that isn't likely, and if it was happening I have to ask how can the sync system be relied on? Or rather, anything be relied on?

Either 4D has a bcrypt implementation bug (I'm skeptical they wouldn't be using a standard c++ opensource lib for it), I'd wager that's a major issue in 4D. 4D should patch their LTS 20.1. 

Or, payloads are getting broken (A) in our sync code in Legacy, (B) broken during transmission, or (C) broken in OMS. A or B we should absolutely be fixing, and C then the internet is literally broken 😁. If any are the case, then there is no end to the amount of validation we should be doing: what about UUIDs? We could have regexes to make sure they match either 4D's UUID format, mobile's or UUIDv7 that OMS uses. I think in those cases it's more obvious that regexing to validate them all is unnecessary overhead, and that reflects my view on regexing a bcrypt hash as well. We have to assume strings reliably transmitted or else our whole sync system is doomed anyway.

Chris-Petty:
Minor tweaks here and it's working nicely based on `feature-sync` branch. I've opened a new branch and PR for the OG side https://github.com/msupply-foundation/msupply/pull/18198. It's substantially simpler than the original one as the routing is boiled down to simply always sending OMS Central both `user` and `user_store` records when they're updated.

There a few scenarios to account for when thinking about when and which records are synced (not sure if best here or in legacy repo):

## New OMS central server, initialising

We're queuing all the site and store records from scratch. This does _not_ include all users and user_store records, only for the users that are active on the stores in OMS central and the users referenced in records that would be sent, e.g. invoices, logs, etc. (but not their user_stores)

As store incremental migration is done over time, the rest of users and their user_store records will be included.

**Could be tempted to change this - for COMS initialisation specifically just queue all the users, just we specifically queue all the site table records.** Would be less risk of a FK constraint being sad, though if one did because of existing logic, then it's an existing problem right now in production 😅.

## Existing OMS central server, Legacy Central Server just got upgraded with this new code

When legacy migrates, it populates the new user.password_hash field with a bcrypt hash. This queues all of the users for OMS central. When OMS central next syncs it will simply get all users at that point.

user_store records would need to be resynced at least for active stores (I'll make sure that happens in the OG PR now). The rest of user_store records should come through as stores are incrementally synced.

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,181 @@
...
+            permissions,
+            is_default,
+            can_login,
+            can_action_replenishments: _,
+        } = serde_json::from_str::<LegacyUserStorePermissionTable>(&sync_record.data)?;
+        if StoreRepository::new(connection)
+            .query_one(StoreFilter::new().id(EqualFilter::equal_to(&store_id)))?
+            .is_none()
+        {
+            return Ok(PullTranslateResult::NotMatched);
```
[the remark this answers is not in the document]
Chris-Petty: I think you're on point here, happy to nix it!

server/service/src/sync/translations/user_store_permissions.rs
Chris-Petty: Yes is_default is still from the user's record, but the permissions are from the user's group. I'm not sure why I have this comment, the UUID should still be the user's user_store. The [COGS code](https://github.com/msupply-foundation/msupply/blob/295bd568267fba01b91f45b5f8f8caa0268b3ab7/Project/Sources/Methods/syncOutGetV5RelatedData.4dm#L63) only updates the permission array to be the group's permissions rather than whatever is on the user's user_store already.

Makes me realise I think it'll be sending user groups now. Probably need to ignore them in sync till we decide to do something, least maybe it's possible to use the group's username and an empty password to login lol.

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,181 @@
...
+            ));
+        }
+
+        integration_operations.push(IntegrationOperation::upsert(user_store_join_row));
+
+        let new_permissions = map_api_permissions(permissions);
+        let mut new_permission_set = permissions_to_domain(new_permissions);
+        new_permission_set.insert(PermissionType::StoreAccess);
+
+        // Similar to user_store_join, the login functionality will drop literally all the user's permissions and recreate them with new PKs.
```
Chris-Petty: I think this comment is invalidated by the below comments 🤔 good job 2025 me

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,181 @@
...
+                    store_id: Some(store_id.clone()),
+                    permission: permission.clone(),
+                    context_id: None,
+                }));
+            }
+        }
+
+        // Some prefs come from om_user_permission! They should be preserved so the `user_permission` translator can handle them
+        existing_permissions.remove(&PermissionType::DocumentQuery);
+        existing_permissions.remove(&PermissionType::DocumentMutate);
```
Chris-Petty: Got real worried for a second that initialisation code might neglect `om_user_permission`, but double checked and they're already covered.

Chris-Petty:
I'm going to make some changes as request and I still need to make the tests useful!

server/service/src/sync/translations/user_store_permissions.rs
Chris-Petty: added comment

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,281 @@
...
+        // and are managed by the user_permission translator — filter them out here so we don't
+        // touch them.
+        let existing_permissions = UserPermissionRepository::new(connection).query_by_filter(
+            UserPermissionFilter::new()
+                .user_id(EqualFilter::equal_to(user_id.to_owned()))
+                .store_id(EqualFilter::equal_to(store_id.to_owned()))
+                .has_context(false),
+        )?;
+
+        if !can_login {
```
andreievg, line 87: I think for the first pass it would be fine, but I would like a way to reduce chatter in sync cause of permissions

Can we think of a way to standardise IDs for user permissions ? i.e. hash of everything unique about user permission, or even the whole row ? And then we delete only the ones that are not matching anymore ? And delete all when not active ?

I think that would especially be needed for v7 sync, where we would forward these permissions to active stores etc..

andreievg: Oops i just saw that's only for !can_login

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,281 @@
...
+        let mut existing_permissions: HashMap<PermissionType, UserPermissionRow> =
+            existing_permissions
+                .into_iter()
+                .map(|p| (p.permission.clone(), p))
+                .collect();
+
+        for permission in new_permission_set {
+            if existing_permissions.remove(&permission).is_none() {
+                integration_operations.push(IntegrationOperation::upsert(UserPermissionRow {
+                    id: uuid(),
```
andreievg: IF we hash user_id, store_id, permission, context_id and permission we would have a pretty standard uuid for user permission row. This means in translations here we would need to update the uuid of existing matching row, which would be quite simple I think

andreievg (commented):
Looks great keen to approve, just the whole standardise ids thing, which I think should be trivial, at least here.

server/service/src/sync/translations/user_store_permissions.rs
Chris-Petty: Honestly in retrospect I kind of wish we did it the OG way! Just a field with either a bit array, or array (i.e. Set) of the ENUMs permissions, or a JSON object with true/false... would save a lot of records! Tradeoff is that it's way less storage and indexing, easy enough for retrieving a user's permissions, but significantly slower for checking what users have a particular permission (basically have to sequentially scan, but if infrequent that's fine eh?).

Definitely an another day problem right?

andreievg: Yeah ! A record with boolean fields would work, for each permission, should be fast to check too

Chris-Petty: True, so long as you don't hit column limits heheh

server/service/src/sync/translations/user_store_permissions.rs
Chris-Petty: Should we use a composite key at the sql level?

Chris-Petty: claude:

Short answer: hash-based deterministic ID is the right call here. Don't switch to a SQL composite key.

**Why not composite key (even though both PG and SQLite support it):**
- The whole sync/changelog pipeline in this repo is built around a single `TEXT id` per row — `record_id: String` everywhere, FK references, `find_one_by_id`, the upsert-on-conflict pattern in `user_store_join_row.rs`. Switching one table to a composite PK means either special-casing it everywhere or serialising the tuple back into a string for the changelog — at which point you've reinvented the hash but messier.
- No other table in `server/repository/src/db_diesel/` uses a composite key (grep — zero matches for `primary_key(`). Doing it for just this one breaks the pattern.
- Diesel works with composite keys but the ergonomics drop (tuple PKs, no `.eq()` shorthand, `find()` takes tuples). Worth it if you needed it; you don't.

**Why hash (deterministic UUID):**
- Solves exactly the bug the `user_store_permissions.rs` comment is working around: OG regenerates PKs on login, so we can't trust the legacy `ID`. If `id = uuid_v5(NAMESPACE, user_id + ":" + store_id)` then every site computes the same ID for the same pair, and the `find_one_by_user_and_store` lookup becomes unnecessary.
- Fits the existing `TEXT` column with zero schema/sync/FK churn.
- `uuid v5` (SHA-1, name-based) is the standard for this — produces a real UUID, deterministic, collision-resistant for our input space.

**Concrete suggestion:** use `uuid::Uuid::new_v5(&Uuid::NAMESPACE_OID, format!("{user_id}:{store_id}").as_bytes()).to_string()` — pick one namespace constant and stash it in `util::uuid` next to the existing `uuid()` helper so it's reusable (the same trick will apply to `user_permission.id`, which has the same problem).

server/util/src/uuid.rs
```diff
@@ -8,3 +8,10 @@ pub fn uuid() -> String {
 pub fn small_uuid() -> String {
     uuid().split('-').next().unwrap().to_string()
 }
+
+/// Stable UUID v5 derived from `name` under `namespace`. Same input always produces
+/// the same UUID — use for IDs that must be reproducible across sites so the same
+/// logical row computes to the same id wherever it's created.
+pub fn deterministic_uuid(namespace: &Uuid, name: &str) -> String {
+    Uuid::new_v5(namespace, name.as_bytes()).to_string()
```
andreievg, line 16: Should we put namespace here too ? Happy with where it is no, just food for thought

server/service/src/sync/translations/user_store_permissions.rs
```diff
@@ -0,0 +1,295 @@
...
+        let mut integration_operations: Vec<IntegrationOperation> = Vec::new();
+
+        // Login code may hit OG API if online. If it does, it drops all permissions and regenerates them with new PKs.
+        // There should only be one join per user and store, so we just match on the user and store ids rather than relying on the PK.
+        // If it doesn't exist just upsert a new one using the PK from OG Central.
+        let user_store_join_row = UserStoreJoinRowRepository::new(connection)
+            .find_one_by_user_and_store(&user_id, &store_id)?
+            .map_or_else(
+                || UserStoreJoinRow {
+                    id,
```
andreievg, line 69: if id is already deterministic, find one by id ?

server/repository/src/migrations/v3_00_00/migrate_user_permission_to_deterministic_id.rs
```diff
@@ -0,0 +1,175 @@
...
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "migrate_user_permission_to_deterministic_id"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        // Read `permission` as TEXT so we don't have to map the Postgres enum
+        // type into the live Rust `PermissionType`, which may evolve later.
+        let select_sql = if cfg!(feature = "postgres") {
+            "SELECT id, user_id, store_id, permission::text AS permission \
```
andreievg, line 28: could do "CAST(permission AS TEXT)"

andreievg (approved):
Looks great thank you, some nits nothing stopping merge.

server/service/src/sync/translations/user_store_permissions.rs
Chris-Petty: TRUE

server/util/src/uuid.rs
Chris-Petty: I thought at first glance the same but decided to roll with claude on it, doesn't hurt to have separate name spaces for the two contexts but maybe over kill heh

PR #9089, opened 2025-08-29
9087 integration test compilation

### LINKED ISSUE #9087: Fix integration test compilation
## What went wrong? 😲

Integration_test feature flag won't compile. I can't run load test.

- Open mSupply Version: 2.11 dev

Chris-Petty:
Fixes #9087

# 👩🏻‍💻 What does this PR do?

Fix compiler errors

I added default- bite me!

I did not actually test integration_tests. I just want the feature flag to build so I can use the `load_test` cli!

# 🧪 Testing

```sh
cargo check --features integration_test
```

### DIFF
--- a/server/cli/src/load_test.rs
+++ b/server/cli/src/load_test.rs
@@ -807,6 +807,8 @@ impl TestSite {
             .await?)
     }
 
+    // Repeatedly starts sync until local db sync info confirms that integration of sync buffer finished
+    // Consider consolidating with similar `sync_omsupply_central` in `server/service/src/sync/test/integration/omsupply_central/mod.rs`
     async fn do_sync_until_integrated(&self) -> Result<SyncInfo> {
         loop {
             self.do_sync().await?;


server/service/src/sync/test/integration/mod.rs
```diff
@@ -154,6 +154,8 @@ pub(crate) fn integrate_with_is_sync_reset(
     let cursor = changelog_repo.latest_cursor().unwrap();
     // Need to reset is_sync_update since we've inserted test data with sync methods
     // they need to sync to central (if is_sync_update is set to true they will not sync to central)
-    integrate(&connection, integrations, None).unwrap();
+    let integrations: Vec<(Option<_>, IntegrationOperation)> =
+        integrations.iter().cloned().map(|i| (None, i)).collect();
+    integrate(&connection, &integrations).unwrap();
```
Chris-Petty, line 159: All the cloning trait shenanigans was so I could do this cause otherwise I was refactoring chains of functions to take `&[(Option<i32>, IntegrationOption)]` becoming a mess.

Idk how to trait good, copilot on claude 4.0 helped a lot.

Chris-Petty:
Ah I got carried away and committed some changes to load_test procedure rather than make a new issue/branch 🤦

server/repository/src/lib.rs
```diff
@@ -63,6 +63,32 @@ pub trait Upsert: DebugTrait {
     }
 }
 
+pub trait UpsertAndClone: Upsert {
```
jmbrunskill, line 66: Spent WAAY too much time trying to find a nicer solution but it turns out it's harder than it looks.
I found a rust crate https://crates.io/crates/dyn_clone that solves this problem but is an extra dependancy and uses unsafe code so not sure how much I can trust it.

I also tried the use `Arc` instead of box approach which works apart from the use mutable downcast in test cases. I'm not clear why we can get a mutable ref for downcast from a box but I can't get. non-mutable ref in an arc. If we make upsert `Send + Sync` we can clone in the arc but I'm not sure it's safe to do that either.

Maybe just add a quick comment here to explain why the trait is needed?

```
Because upsert is used as a `Dyn trait` we can't ensure that the types also implement `Clone` without creating an additional Trait like this, or us a similar trait from a library such as `dyn-clone`
```

jmbrunskill (approved):
Looks all good to me, thanks @Chris-Petty just would like an explanation for the extra Trait.

andreievg (changes requested):
I would suggest this, instead of new trait: https://github.com/msupply-foundation/open-msupply/compare/9087-integration_test-compilation...9087-integration_test-compilation(New-trait-alternative)

Chris-Petty:
Yep that works thanks @andreievg!

server/service/src/sync/test/mod.rs
```diff
@@ -25,7 +25,7 @@ impl TestSyncIncomingRecord {
         upsert: U,
     ) -> TestSyncIncomingRecord
     where
-        U: Upsert + 'static,
+        U: Upsert + Clone + 'static,
```
andreievg, line 28: I think this is still remainder from the trait stuff ? If so :nit to delete please

andreievg (approved):
Looks great, thanks

PR #9096, opened 2025-08-29
9088 backend-managed feature flags

### LINKED ISSUE #9088: Configure feature flags via server config

Feature flags currently need to be available to the frontend at build time, making it difficult to share builds for different features with stakeholders.

Add a `features` section to the local.yaml configuration file. All a user would need to do is enable the feature, and restart the service. 

These flags would be exposed to the frontend and enable/disable the feature accordingly.

### Acceptance Criteria ✅

- [ ] Upon setting a `table_usability_improvements` feature to true in local.yaml, and restarting server
- [ ] Outbound shipment detail view shows "INCOMING: NEW TABLE" rather than the usual data table

### Moneyworks Jobcode 🧰

OMS:GEDSI

lache-melvin:
Fixes #9088

# 👩🏻‍💻 What does this PR do?

Add feature:
Screenshot 2025-08-29 at 5 22 29 PM

Restart server.

See feature:
Screenshot 2025-08-29 at 5 22 47 PM

Per issue, means PMs and other stakeholders can easily enable different in-progress features, regardless of the build version they receive!

_Does_ mean, if we deploy with feature flags exposed, user _could_ go into local.yaml and turn things on?

I think risk of this is low, usually only our support team touching those config files... there's the same risk in them touching sync settings, db config, and I'd say that's worse?

# 🧪 Testing

- [ ] Upon setting a `table_usability_improvements` feature to true in local.yaml, and restarting server
- [ ] Outbound shipment detail view shows "INCOMING: NEW TABLE" rather than the usual data table

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/README.md
```diff
@@ -26,10 +26,6 @@ To see it in action, check out the [demo server](https://demo-open.msupply.org/)
 
 `yarn start API_HOST='http://localhost:8001'` (using custom API url, see [config.ts for more info](./packages/config/src/config.ts))
 
-- If there are feature flags in use, they are set as environment variables. For example:
-
-`yarn start FEATURE_INVENTORY_ADJUSTMENTS=true`
```
lache-melvin, line 31: have replaced the old implementation, as you'll see below

client/packages/host/src/Admin/Settings.tsx
```diff
@@ -77,17 +75,7 @@ export const Settings: React.FC = () => {
         <LabelPrinterSettings />
         <ElectronSettings />
       </SettingsSection>
-      {Environment.FEATURE_PRINTER_SETTINGS && (
```
lache-melvin, line 80: Ability to add printers was behind a feature flag, but it was just a dummy component. Hasn't been touched for 6mo. 

I think fair to just take out at this stage? Can be added if/when we actually implement this?

client/packages/host/src/useInitPlugins.ts
```diff
@@ -40,7 +41,7 @@ export const useInitPlugins = () => {
   useEffect(() => {
     if (
       process.env['NODE_ENV'] === 'production' ||
-      Environment.LOAD_REMOTE_PLUGINS
+      featureFlags.loadRemotePluginsInDev
```
andreievg, line 44: I think this won't quite work featureFlags won't be available at this stage yet ?

I haven't really used this feature that much, it was useful when originally doing plugin work, not so much since. 

I think it's ok to remove this for now (the alternative is to build the front end and access it through 'server', which is ok.

andreievg (commented):
Keen to approve, to get this work in now.

However there is a view that we have lots of places to make changes when adding/modifying things, so I am wondering if in:
* .yaml features is an array
* useFeatureFlags exposes a method to which you can pass a string for the features name (or just access array element directly with !!).

What do you think ?

zachariah-at-msupply (approved):
Looks great Laché!

Really cool feature, let's merge it in!

client/packages/host/src/useInitPlugins.ts
lache-melvin: Ohh true, I didn't try that sorry. Ok, will remove for this case - good thing to be aware of if we need flags available on startup in future though!

lache-melvin:
> * .yaml features is an array
> * useFeatureFlags exposes a method to which you can pass a string for the features name (or just access array element directly with !!).
> 
> What do you think ?

> we have lots of places to make changes when adding/modifying things

What do you mean by lots of places, in this case? The only place to define a new flag would be the `useFeatureFlags`, just to map the string key to a known variable name. I have a preference for this, so we can easily find references (especially for features that enable something in several places.)

I'm not sure how the array helps? Means you'd have to know how many features there are, and set the 3rd one to true?

client/README.md
lache-melvin: Ah great shout, thanks! I think as feature flags were not widely used, I'm okay with not commenting on the old pattern, but have added a note to the frontend readme about how they work now!

lache-melvin:
@andreievg per above, I'm going to merge this one. I think low risk if we need to change, if there's something I'm missing with your explanation I'm keen to chat. Otherwise keen to get this in :)

andreievg:
@lache-melvin i think it's fine, just wanted your opinion. Also I didn't go through the full review and see that we don't need to change anything in settings.rs when feature flag is added. 

Btw for mobile we would need to make separate builds still, but that should be quite straight forward, I think you will need to do this for table rc build, I'll forward a msg on how to build with data

lache-melvin:
> Btw for mobile we would need to make separate builds still, but that should be quite straight forward, I think you will need to do this for table rc build, I'll forward a msg on how to build with data

@andreievg Thank you! I flagged in comment above, and in code - I think if wanting to view feature flagged code on android, just need to run in client mode, and connect to a desktop server with the flags set. I'm not even sure how to set local env vars on android?

PR #9104, opened 2025-09-01
9007 Fix invalid url in custom translations on Electron

### LINKED ISSUE #9007: New installation of omSupply doesn't get to the initialisation screen
## What went wrong? 😲
**Note Could be on my side too as I have the same problem with version Open_mSupply_Server_v2.10.00-RC-08221718**

Starting a new installation omSupply with an empty datafile prevent omsupply to get to the initialisation screen.

https://github.com/user-attachments/assets/8c3cab3d-90a2-438f-9ce9-691733a201f7

No message error but log says that that it was not able get any active site but I didn't try to initialise yet.  

log: 

[omsupply.log](https://github.com/user-attachments/files/21975248/omsupply.log)

 I had a similar problem when trying to upgrade my previous installation of 2.9.2 but the log was different. That was why I tried to re-install with a new datafile

Image

If I ignore the warning and proceed to connect to omSupply
Image

When launching the test-connectioncli.exe

Image

My configuration files are there.

Image

and my certificates too

Image
 

## Expected behaviour 🤔
Should probably get the initialisation :)

## How to Reproduce 🔨

Install RC installer: Open_mSupply_Server_v2.10.00-RC-08251530.exe
1. Make sure the service is installed
2. Try to connect using the Browser or open mSupply Desktop
3. See error. You never get to the login or initialisation page

## Your environment 🌱

- Open mSupply Version: v2.10.00-RC-08251530
- Legacy mSupply Central Server Version: 8.3.15

- Platform: Desktop or Browser

- Database type: SQLite

jmbrunskill:
Fixes #9007

# 👩🏻‍💻 What does this PR do?

It turns out this was custom translations after all.
I've tried to add some comment for the next poor soul who ventures into this code.

Basically the problem was that one electron when we load the inital version of the app for discovery, we don't have a valid API Host. The electron app is loaded via a `file://` path which is what we used to try to figure out the API host.

Even though we didn't need to access this url, when initialising the translation context if the url isn't even seen as a valid url, it fails to initialise with an error.

Probably should have found this much sooner, but learned a lot in this process at least!

Maybe it's better to detect `file:` path in the ENVIRONMENT codes and mark API_HOST as null in this case, as it's potentially confusing to have it set to something that won't work...

# 🧪 Testing

- [ ] Build electron with `yarn electron:build`
- [ ] Try to start the exe

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/intl/context/IntlContext.tsx
```diff
@@ -35,10 +35,11 @@ export function initialiseI18n({
 
   // Served with frontend bundle
   // Electron `main` window translations should be served with relative path
+  // for electron, the preloaded script path is `file://:` we don't get a valid API_HOST url until we connect to the server and re-initialise the window
   const defaultTranslationsLoadPath = `${!!isElectron ? '.' : ''}/locales/{{lng}}/{{ns}}.json`;
 
-  // Served from backend
-  const customTranslationsLoadPath = `${Environment.API_HOST}/custom-translations`;
+  // Served from backend, on electron we use a dummy but valid url https://localhost:8000 which shouldn't actually be used.
+  const customTranslationsLoadPath = `${!!isElectron ? 'https://localhost:8000' : Environment.API_HOST}}/custom-translations`;
```
lache-melvin, line 42: This means custom translations wouldn't be available on Desktop if you run your server on any other port?

lache-melvin (changes requested):
Thanks James for all the investigation! I think we need a solution that works on other ports though? 

Think I agree, would be better solution for `Environment` to only expose valid paths in the first place, if its not too much extra work 👀

client/packages/common/src/intl/context/IntlContext.tsx
jmbrunskill: I don't think it does? Wait let me test...

jmbrunskill: Ok further followup I did get a chance to yarn build then restart but it seems the yarn build version isn't loading custom translations in either a browser or electron. Will need to investigate further tomorrow.

lache-melvin (approved):
Epic! Also made a test build off this branch, Maria has just tested and working well for her too 🙏  Thanks heaps James ❤️

PR #9126, opened 2025-09-03
Inital table usability changes -- crude prototype

CarlosNZ:
This is just a crude prototype so the new potential table replacement can be tested by various stakeholders and tested in different enviroments.

**It is very crude at this stage and lot's doesn't work properly, so please don't give bug reports -- just general usability experience**

I have implemented for:
- Outbound Shipments list view 
- Outbound Shipment detail view

These cover our two main types of tables:
- paginated table, where sort/filter/pagination changes makes calls to back-end API to get new data
- "pre-loaded" table, with all rows present in front-end, so sorting/filtering done locally. (No pagination) -- recommend testing this one with a data file that has thousands of items in a single master list, so you can push the table responsiveness to its limits -- it should be properly virtualised, so after the initial load changes to sort/filter should be almost instant.

**Requires feature flag enabled** in local.yaml: 
```
features:
  table_usability_improvements: true
```

## What is working and what isn't:

**List view**:
 - Columns should show correct data, formatted for type (currency, status, etc) (except the color blob with Name)
 - Column sorting works (though there's something a bit funky going on with Invoice Number when you click it multiple times)
 - Pagination works (though I've hidden the "Rows per page" dropdown, since it's not actually connected to the rest of the state properly, even in the current version)
 - Filtering is NOT working fully -- you can enter values into the inline filters and should see changes, but state is not properly synced between UI/url and table
 - Row selection UI is working, and should see Footer action, but doesn't do anything on submission
 - Column show/hide working

**Detail view**:
- Sort/FIlter/Search all working (as it's done completely withing Table component library)
- Column adjustable width is working
- Column show/hide is working
- Placeholder rows NOT formatted differently yet
- Row disabling not implemented
- Row selection UI available, but no footer or action
- Group by item not implemented yet

## Next Steps

These examples are just hard-coded in place at the moment, so we'll want to generalise the components now I understand them a bit better.

Next step is to write up list of issues to get this implemented properly and re-usable in a general way.

# 🧪 Testing

- [ ] Outbound shipment list view -- play with the features :) 
- [ ] Outbound shipment detail view -- load a table with thousands of items and have a play

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin:
it IS feature flagged 🎉

Issue #9129, opened 2025-09-03
Create utilities for mapping sorting, filtering and pagination between table state and our API hooks in a reactive and general purpose way


CarlosNZ:
Will do "sort" first, see how that goes, and decide if "filter" and "pagination" should be separate issues

**UPDATE:** The "sort" and "pagination" handling is fairly straightforward (notwithstanding a couple of weird bugs), but the filtering will require a little more work. 

The filter definitions will have to be defined in the column definitions themselves, so the key points will be:
- "auto-defining" filter definitions based on column type (i.e. if the column is "status", it should be a "Select" with each of the status values in it -- we don't want to have to define this manually every time)
- mapping the "values" from the filters (in the table) to the requred shape to insert into the URL query (and back again)

This will completely replace the filter definitions from the Toolbar, but we can probably re-use some of the utilities in that infrastructure.

marthakeezy:
This will be tested as a part of #9250 in V2.11 Release Testing - so moving to Done ✅

PR #9134, opened 2025-09-03
#9128 Generalise new table hooks

### LINKED ISSUE #9128: Make generalised components for the two main types of tables (API-backed/front-end)
I think we might just be able to make two hook variants and use the `MaterialReactTable` component as-is. Because the standard pattern is to use MRT's own `useMaterialReactTable` table, whic is where all the configuration props for the Table go. So I suggest we make two hooks, which are just wrappers around `useMaterialReactTable` with most of the required props pre-configured for our purposes: 

- `usePaginatedMaterialReactTable`
- `useNonPaginatedMaterialReactTable`

It's important though, that we still pass through *all* of the `useMaterialReactTable` props so, even though we'll be pre-configuring the tables for our purpose, we can still override any prop on a case-by-base basis.

We might also want to make a third variant for the line edit (batch edit) tables, but let's see how these go first

CarlosNZ:
Fixes #9128

# 👩🏻‍💻 What does this PR do?

Very simple abstraction of the two main table types we'll use. Wraps around the `useMaterialReactTable` and injects expected props for the table type (though any can be overwritten).

I've still got hard-coded filter values in the "Paginated" variant, but will work on that next in #9129 

WIP, as we know...

# 🧪 Testing

- [ ] Should look and behave the same as #9126, just been refactored into generic hooks

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin (approved):
LGTM, nice base to build from, thanks :)

PR #9135, opened 2025-09-03
9082 Hide other order quantity in requisitions

### LINKED ISSUE #9082: "Other requested stat" in My Store Tab show as Negative even though there is no pending requisition for that item in my store
## What went wrong? 😲

My finalised requisition
Image

As title. But should not show negative anyway because we are adding qty and I didn't see in the database any Negative number that would make a sum giving a negative result.

Manual says that other requested show the qty being requested for this item  in other pending requisitions  from all customer to your store. So I am assuming by pending requisition we mean all not finalised requisitions for which store_id is my store. 

Image

In database: All requisitions lines (Response and Request) for that item (in green the internal order where there was a qty requested greater than 0 and in red the corresponding requisition. The 1st line is request requisition from another store and the 4th line is a response requisition that has been finalised and for another store)
Image

In this specific case both the customer and the supplier store are present on my site. I wonder if it could affect the result ?

## Expected behaviour 🤔
Other requested should never be a negative number

## How to Reproduce 🔨

1. Use the datafile provided log 
2. Log into Slytherin clinic
3. go to Distribution > Requisition
4. Open requisition no. 2
5. Click on line of item  Salbutamol scored 2mg tabs
6. See error. The Other requested stat under My Store tab section at the bottom show a negative number
 
## Your environment 🌱

- Open mSupply Version: v2.10.00-RC-08281607
- Legacy mSupply Central Server Version:8.3.15

- Platform:Browser

- Database type: SQLite

raviSussol:
Fixes #9082 

# 👩🏻‍💻 What does this PR do?
Normalizes negative other requested quantity to 0.

# 🧪 Testing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseStats/ResponseStoreStats.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/DetailView/ResponseStats/ResponseStoreStats.tsx
@@ -174,47 +174,43 @@ export const ResponseStoreStats = ({
             </span>
           </Typography>
         ) : (
-          (formattedRequested !== null || formattedOtherRequested !== null) && (
-            <>
-              <Typography style={{ textAlign: 'start' }} variant="h6">
-                {t('label.requested')}
-              </Typography>
-              <Box
-                display="flex"
-                alignItems="flex-start"
-                flexDirection="column"
-                width={`${requestedPercent}%`}
-              >
-                <Box display="flex" width="100%">
-                  <NewValueBar
-                    value={formattedRequested}
-                    total={totalRequested}
-                    colour="primary.main"
-                  />
-                  <NewValueBar
-                    value={formattedOtherRequested}
-                    total={totalRequested}
-                    colour="primary.light"
-                  />
-                </Box>
-                <Box paddingTop={1}>
-                  {formattedRequested !== null &&
-                    statsDisplay(
-                      'label.requested-quantity',
-                      formattedRequested,
-                      'primary.main'
-                    )}
-                  {requisitionStatus !== RequisitionNodeStatus.Finalised &&
-                    formattedOtherRequested !== null &&
-                    statsDisplay(
-                      'label.other-requested-quantity',
-                      formattedOtherRequested,
-                      'primary.light'
-                    )}
-                </Box>
+          <>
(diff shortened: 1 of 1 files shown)


jmbrunskill (changes requested):
Hi Ravi, I don't think this is the correct fix.

It would work for this example,but if there was another pending requisition with say 100 requested, the number would say 25, which isn't correct.

I think it would be better to exclude the current requisition when finding lines in `other requisitions`
I see this code in the `response_line_state.rs`

```
    let response_requisition_lines = RequisitionLineRepository::new(connection).query_by_filter(
        RequisitionLineFilter::new()
            .store_id(EqualFilter::equal_to(store_id))
            .item_id(EqualFilter::equal_to(&requisition_line.item_row.id))
            .r#type(RequisitionType::Response.equal_to())
            .status(RequisitionStatus::Finalised.not_equal_to()),
    )?;
```

I think we should probably rename the variable to be `other_response_requisition_lines` to make it clear that it excludes the current one.

Then filter out the current requisition line from this query so it's not included.

Something like this (but I haven't tested it)
```
    let other_request_requisitions = RequisitionLineRepository::new(connection).query_by_filter(
        RequisitionLineFilter::new()
            .id(EqualFilter::not_equal_to(&requisition_line.id))
            .store_id(EqualFilter::equal_to(store_id))
            .r#type(RequisitionType::Request.equal_to())
            .item_id(EqualFilter::equal_to(&requisition_line.item_row.id))
            .status(RequisitionStatus::Sent.equal_to()),
    )?;
```

Then you can remove the `sum_of_lines - current_line_quantity` code and just return the `sum_of_lines`.

That way we'll be sure we're not accidentally adjusting for the current line.
I think the problem is that the requisition is finalised, so we are not actually getting the currently line back in the query above, this means we're adjusting for a line that we don't have in out calculation. 
The current code works fine when it isn't finalised but once it's finalised 

@alainsussol @mark-prins is the intention to show a snapshot value of other requests at the time of the requisition? Right now it looks like it's calculating that value even for finalised requisitions.

Pretty sure that's a new issue, but wanted to flag that might be the user's expected behaviour?

PR #9146, opened 2025-09-04
#9129 Table sorting and filtering

### LINKED ISSUE #9129: Create utilities for mapping sorting, filtering and pagination between table state and our API hooks in a reactive and general purpose way

CarlosNZ:
Fixes #9129

# 👩🏻‍💻 What does this PR do?

Implements a common mechanism for handling the various types of filters we have. Currently just handles text, enum and date range, but that covers most of the cases I think.

I created a new hook to parse the column info, but it allows us to just specify a general filter type in *our* column definitions, and the hook converts that into the more detailed shape required by MRT, as well as providing "updater" functions for updating the URL query correctly.

Also implemented the reverse, where the URL query is converted back into the "filter state" requried by MRT so they can stay in sync with each other.

As far as pagination and sorting goes, I didn't really change anything, I just extracted them into named handler functions inside the hook.

I'm not sure I've got the use of `useMemo` and `useCallback` quite right, some of them may be redundant. Please point out any obvious problems, but also let's keep an eye on that going forward. :)

# 🧪 Testing

- Sorting: All columns should be sortable, and should no long get "stuck" when repeatedly toggling a column
- Pagination: should work correctly
- Filtering: Have defined filters for Name, Status and Created, which should work correctly. Other fields should have no filters (for now)

Note: there is a bug you might find in the Date range filter, but I've confirmed that's actually a pre-existing bug -- if you apply both a start and end date to the filter, then delete just the start date, you'll get a query error. I'll make an issue to fix this.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/material-react-table/useMaterialTableColumns.tsx
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/useMaterialTableColumns.tsx
@@ -1,0 +1,114 @@
+/**
+ * Hook to map column settings (as defined by us, with a handful of common
+ * filter types) to the exact column structure required by MaterialReactTable
+ */
+import { useMemo } from 'react';
+import { MRT_ColumnDef, MRT_RowData } from 'material-react-table';
+import {
+  DateUtils,
+  useFormatDateTime,
+  useUrlQuery,
+} from '@openmsupply-client/common';
+
+type FilterType = 'none' | 'text' | 'number' | 'enum' | 'dateRange';
+
+interface EnumOption {
+  value: string;
+  label: string;
+}
+
+type ColumnDefinition<T extends MRT_RowData> = MRT_ColumnDef<T> & {
+  filterType?: FilterType;
+  filterValues?: EnumOption[];
+};
+
+export const useMaterialTableColumns = <T extends MRT_RowData>(
+  columns: ColumnDefinition<T>[]
+) => {
+  const { urlQuery, updateQuery } = useUrlQuery();
+  const { customDate, urlQueryDate, urlQueryDateTime } = useFormatDateTime();
+
+  const { mrtColumnDefinitions, filterUpdaters } = useMemo(() => {
+    const mrtColumnDefinitions: MRT_ColumnDef<T>[] = [];
+    const filterUpdaters: Record<string, (value: any) => void> = {};
+
+    columns.forEach(
+      ({ filterType = 'none', filterValues, ...mrtProperties }) => {
+        const filterKey = (mrtProperties.accessorKey ||
+          mrtProperties.id) as string;
+
+        switch (filterType) {
+          case 'none':
+            mrtColumnDefinitions.push({
+              ...mrtProperties,
+              enableColumnFilter: false,
+            });
+            break;
+          case 'text':
+            mrtColumnDefinitions.push(mrtProperties);
+            filterUpdaters[filterKey] = (value: string) => {
+              updateQuery({ [filterKey]: value });
+            };
+            break;
+
+          case 'enum':
+            mrtColumnDefinitions.push({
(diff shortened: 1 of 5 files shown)


client/packages/common/src/ui/layout/tables/material-react-table/useMaterialTableColumns.tsx
```diff
@@ -0,0 +1,118 @@
...
+  value: string;
+  label: string;
+}
+
+type ColumnDefinition<T extends MRT_RowData> = MRT_ColumnDef<T> & {
+  filterType?: FilterType;
+  filterValues?: EnumOption[];
+};
+
+export const useMaterialTableColumns = <T extends MRT_RowData>(
```
lache-melvin, line 29: This is a real nice abstraction!

client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.ts
```diff
@@ -1,98 +1,155 @@
...
-        const newPaginationValue = pagination(current);
-        // console.log('Pagination changed:', newPaginationValue);
+    [sortBy, updateSortQuery]
+  );
+
+  const filterState = getFilterState();
+
+  const handlePaginationChange = useCallback(
+    (paginationUpdate: MRT_Updater<MRT_PaginationState>) => {
+      if (typeof paginationUpdate === 'function') {
```
lache-melvin, line 91: I'm finding these a little hard to follow with the comments + dense logic... are these functions considered "done" or still a bit first pass?

client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.ts
```diff
@@ -1,98 +1,155 @@
...
     enableRowSelection: true,
     onRowSelectionChange: setRowSelection,
     rowCount: totalCount,
     state: {
       sorting: [{ id: sortBy.key, desc: !!sortBy.isDesc }],
       pagination: { pageIndex: pagination.page, pageSize: pagination.first },
       showProgressBars: isLoading,
-      columnFilters,
+      columnFilters: filterState,
       rowSelection,
```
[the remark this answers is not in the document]
CarlosNZ: Yes, it does, but I'm utilising that state in the next PR: #9174

client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.ts
CarlosNZ: Yes, they're quite ugly, but I can't really see a way to simplify them. The main problem is that argument  for the "update" functions provided by MRT is a function, which means you have to *call* it to get the current state. Plus I've found a couple of odd behaviours/bugs, which require a bit of a hacky workaround. Let's see if we can't find a way to hide or deal with this a bit better going forward.

PR #9167, opened 2025-09-07
9156 Disable Manual Returns with permission

### LINKED ISSUE #9156: When the permission Disable Manual Return is off, there is an error message instead of the lack of permission like clicking on the New Supplier return
## What went wrong? 😲

When I click on the button New Return, I received immediately a blue message that I can't. But when there is no return I can click on the Create New Return link. When I do, it will let me a choose supplier and only I will have a different permission lacking message but also a red undefined message. See video below. 

https://github.com/user-attachments/assets/53cb36bc-340c-4e4d-a8b4-eeb49f20e177

 

## Expected behaviour 🤔
I think I should not be able to get to the choose a supplier from the supplier modal. I should have a blue permission warning instead and no red flag error message

## How to Reproduce 🔨

log to a store that has no Supplier returns
set the omsupply server disable store preference on or set the user preference Create supplier return to off

Image

1. Go to 'Replenishment > Supplier return'
2. Click on 'Create New supplier return' link
3. See that you can choose a supplier.
4. See red flag error after you do

## Your environment 🌱

- Open mSupply Version: 2.10.1
- Legacy mSupply Central Server Version: v8-04-04

- Platform: omSupply Desktop

- Database type: SQlite

aimee-mcneil-melville:
Fixes #9156

# 👩🏻‍💻 What does this PR do?

Prevents creating a new supplier or customer return from the `Nothing here` area in list view

These are controlled by the `Disable manual returns` store pref in OMS, and the following prefs on OG - `Return stock from customer invoices` and `Return stock from supplier invoices`

This is not directly related to the `Create supplier credits` pref mentioned in the issue :)

# 🧪 Testing

- [ ] Have both the OG and OMS permissions turned off
- [ ] Have no supplier returns existing on your store
- [ ] Attempt to create a supplier return from the 'Nothing here' link -> get toast advising that manual returns are disabled for your store
- [ ] Will get the same message if attempting to create form the `New return` button in the Toolbar
- [ ] Test that if one permission is off and the other on, cannot create a return
- [ ] If allowed in both, a return is able to be created
- [ ] Repeat for Customer returns

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/Returns/CustomerListView/AppBarButtons.tsx
+++ b/client/packages/invoices/src/Returns/CustomerListView/AppBarButtons.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react';
+import React from 'react';
 import {
   DownloadIcon,
   PlusCircleIcon,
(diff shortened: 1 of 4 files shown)


roxy-dao (approved):
Thanks Aimee~

PR #9174, opened 2025-09-08
#9148 Replace store for row state

### LINKED ISSUE #9148: Replace Table store with local state to handle row selection
Don't need a store for this, just use MRT and keep the value in the parent and pass to Footer. For:
  - [ ] Row selection
  - [ ] Row disabled state

CarlosNZ:
Fixes #9148

# 👩🏻‍💻 What does this PR do?

Handles the selected row state, and makes it available to the Footer actions in the new table structure.

I've hoisted the actual `useMaterialTable` hook up a level, so it can share the selected row state with the Footer. At first I was resistant to doing this, but I actually think it's fine as the actual rendering of the `<MaterialTable>` component still happens in the ContentArea component. One additional advantage of this is that when you switch tabs, the table state is persisted, which I think could be handy in several other table implementations. Performance-wise I think it's about the same, but we might be able to add some additional optimisations once we've done a bit more.

I've also added an unrelated prop for handling the table width -- `forceFullWidth`, see inline comment for explanation.

The conditional logic due to the feature flag is getting a bit gnarly in places, so it'll nice to simplify this once we get rid of the feature flag.

# 🧪 Testing

- [ ] Row selection and actions should work as before in both list view and detail view

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
@@ -10,7 +10,7 @@ import {
   CheckboxIndeterminateIcon,
 } from '@common/icons';
 
-interface NonPaginatedTableConfig<T extends MRT_RowData>
+export interface BaseTableConfig<T extends MRT_RowData>
   extends MRT_TableOptions<T> {
   onRowClick?: (row: T) => void;
   isLoading: boolean;
@@ -21,7 +21,7 @@ export const useBaseMaterialTable = <T extends MRT_RowData>({
   onRowClick,
   state,
   ...tableOptions
-}: NonPaginatedTableConfig<T>) => {
+}: BaseTableConfig<T>) => {
   const table = useMaterialReactTable<T>({
     enablePagination: false,
     enableColumnResizing: true,
--- a/client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.ts
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.ts
@@ -15,7 +15,7 @@ import {
   MRT_ColumnFiltersState,
 } from 'material-react-table';
 import { useCallback, useMemo, useRef, useState } from 'react';
-import { useBaseMaterialTable } from './useBaseMaterialTable';
+import { BaseTableConfig, useBaseMaterialTable } from './useBaseMaterialTable';
 
 type FilterType = 'none' | 'text' | 'number' | 'enum' | 'dateRange';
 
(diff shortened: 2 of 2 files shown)


client/packages/common/src/ui/layout/tables/material-react-table/MaterialReactTable.tsx
```diff
@@ -15,14 +15,20 @@ import { Box } from '@openmsupply-client/common';
...
-      <MaterialReactTable table={table} />
-    </Box>
-  );
+  if (forceFullWidth) {
+    return (
+      <Box sx={{ width: '100%' }}>
+        <MaterialReactTable table={table} />
+      </Box>
+    );
+  }
```
CarlosNZ, line 31: It'd be nice to find a way around this, but I haven't been able to set `width: 100%` on the MaterialTable component's outermost div. 

The reason I've made it conditional here is that, without it, tables with fewer columns don't fill the full available horizontal space. But when this is on, the "Detail" table (with row virtualisation) ends up with either a double-vertical scroll, or single vertical scroll around the wrong div. 

I don't think this is right yet, but we should probably re-address once all the functionality is in.

client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.ts
```diff
@@ -36,6 +42,12 @@ export const useNonPaginatedMaterialTable = <T extends MRT_RowData>({
         verticalAlign: 'bottom',
       },
     },
+    muiTableContainerProps: {
+      sx: {
+        maxHeight: '100%',
+        overflow: 'auto',
+      },
+    },
```
CarlosNZ, line 50: This keeps the *content* part of the tablel scrollable, but leaves the headers fixed

client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.ts
```diff
@@ -12,13 +17,27 @@ export const useNonPaginatedMaterialTable = <T extends MRT_RowData>({
   onRowClick,
   ...tableOptions
 }: NonPaginatedTableConfig<T>) => {
+  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
```
lache-melvin, line 20: can you explain why the separate state is needed?

lache-melvin (changes requested):
need clear selected wired up in the detail view too...

and do you think we should just disable the selected rows count in the table itself?

lache-melvin (approved):
LGTM!

PR #9176, opened 2025-09-08
9131 add table styling

### LINKED ISSUE #9131: Table Styling
...(this is a biggie, and will require UI designs), will probably break into multiple issues when we know more

Here's a few things to remember to cover:
- [ ] Remove hard-coded style values and replace with palette/theme

lache-melvin:
Fixes #9131

# 👩🏻‍💻 What does this PR do?

Initial table styling, to align closer to original designs:

Screenshot 2025-09-08 at 1 09 04 PM

Didn't spend too much time on anything too hard, and haven't looked at stying for things like placeholder/disabled rows yet, I think those are separate

Only frustration is the enum filter - there is a massive area that you can't click on to open the dropdown, it's where the `clear` icon will go once an option is selected. But to open the dropdown, you have to click the text, the arrow down doesn't work? It's the same in their demos, I might come back to it later and see if we can append the onClick handler somehow..

# 🧪 Testing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+  CheckboxIndeterminateIcon,
+} from '@common/icons';
+
+interface NonPaginatedTableConfig<T extends MRT_RowData>
+  extends MRT_TableOptions<T> {
+  onRowClick?: (row: T) => void;
+  isLoading: boolean;
+}
+
+export const useBaseMaterialTable = <T extends MRT_RowData>({
```
lache-melvin, line 19: shared props didn't like working with the generic type, so instead using this as base, and passing in overrides from the paginated table hook

client/packages/invoices/src/Prescriptions/DetailView/AppBarButton.tsx
```diff
@@ -64,7 +64,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   useRegisterActions([
     {
       id: 'print',
-      name: `${t('button.print')} (${ALT_KEY}+L)`,
+      name: `${t('button.print-prescription-label')} (${ALT_KEY}+L)`,
```
lache-melvin, line 67: side quest :)

client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.tsx
CarlosNZ: Cool, should we put `useBaseMaterialReactTable` into a separate file?

client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+      },
+    }),
+
+    ...tableOptions,
+  });
+  return table;
+};
+
+// Export a renamed version for clarity
+export const useNonPaginatedMaterialTable = useBaseMaterialTable;
```
CarlosNZ, line 109: Oh I see, the non-paginated version is just the base version renamed. That relies on the non-paginated version having no settings that are not also in the paginated version, which I'm not totally sure is the case (`enableRowVirtualization` for one). I guess you can just "unset" them in the paginated version, but I think it would be slighly cleaner to have the two variants both use the base and only put *common* settings in the base one (especially if we move "Base" to its own file, then it would be really clear to see which properties are common, and what are the unique properties for each variant)

lache-melvin: Yep fair call, especially with the virtualisation, hadn't noticed that, will update now 🙏

client/packages/common/src/ui/layout/tables/material-react-table/index.ts
```diff
@@ -1,4 +1,6 @@
-export * from './MaterialReactTable';
 export * from './usePaginatedMaterialTable';
 export * from './useNonPaginatedMaterialTable';
 export * from './useMaterialTableColumns';
+
+// Re-exporting so all imports come from common folder
+export { MaterialReactTable as MaterialTable } from 'material-react-table';
```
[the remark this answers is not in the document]
lache-melvin: Our pattern with the MUI components has been to re-export. I don't think it's a hard and fast rule, but does give us a bit of dependency ring-fencing if things change in future upgrades.

I didn't think the types were being used outside of the common folder, but I see the detail view is using the ColumnDef type, I'll add that here :)

client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.tsx
lache-melvin: [8365919](https://github.com/msupply-foundation/open-msupply/pull/9176/commits/8365919f6b11ccea9a51178568f0b4959f4143ab)

client/packages/common/src/ui/layout/tables/material-react-table/index.ts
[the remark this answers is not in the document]
lache-melvin: [f6397d6](https://github.com/msupply-foundation/open-msupply/pull/9176/commits/f6397d660aafcee9fe83781e1210c39a59bc894b)

Issue #9073, opened 2025-08-28
Exclude items which have ignore for orders


roxy-dao:
Exclude items which have `ignore for orders` from being added to manual Purchase Orders.

- [ ]  Add `ignore_for_orders` boolean to the `item_store_join` table
- [ ]  Items with this property checked cannot be added to manually created Purchase Orders

### Moneyworks Jobcode 🧰
OMS:DPOF. DFAT resource.

PR #9189, opened 2025-09-08
9073.1 exclude items which have ignore for orders

roxy-dao:
Part 1 of #9073

# 👩🏻‍💻 What does this PR do?
Tryin' to keep PRs small. This is part 1 of excluding items which have ignore for orders. 
- Migrations to add `ignore_for_order` to the `item_store_join` table
- Exposing it in the graphql
- Backend changes on purchase line insert/update and purchase order update to check if the item has ignore for orders and if it does then prompt error. 
- Showing it in item's store tab

Screenshot 2025-09-09 at 9 29 47 AM

Part 2 will contain hookin' up errors in frontend and filtering out items that have this property so that the user can't add it!

# 🧪 Testing

- [ ] For now all you can do is set an item to ignore for order in OG. Item > Order Options
- [ ] Sync
- [ ] Go the item in OMS and click store tab
- [ ] See that ignore for order is checked

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/system/src/Item/DetailView/Tabs/Store.tsx
```diff
@@ -43,7 +44,19 @@ export const StoreTab = ({ item }: { item: ItemFragment }) => {
         flexDirection="column"
         style={{ maxWidth: 500 }}
         gap={4}
-      ></Grid>
+      >
+        <DetailSection title={t('title.ordering')}>
+          <DetailInputWithLabelRow
```
aimee-mcneil-melville, line 49: Nit: this input section is slightly larger than the one in the other column - something causing it to expand?

aimee-mcneil-melville (approved):
Looking great Roxy - heaps of work gone in here! 

There's a styling nit in the Store tab, otherwise all 🔥 🔥

client/packages/system/src/Item/DetailView/Tabs/Store.tsx
roxy-dao: Maybe something happening in General tab too. It's the same code 👀 We'll just get this in first and worry about the slight difference later

PR #9192, opened 2025-09-09
#9175 Persist table settings in local storage

### LINKED ISSUE #9175: Persist ALL user-customisation in local storage (column sizes, column order, visibility, rows-per-page, etc)

CarlosNZ:
Fixes #9175

# 👩🏻‍💻 What does this PR do?

Tried a couple of different approaches here, but this is the most robust. The problem with the event handlers (e.g. `onColumnSizingChange`) is that by invoking them you have to manage the state for those properties as well, which adds a lot of additional complexity. You can just use them as a "listener".

So the simplest solution for our case is to monitor the changes with `useEffect`, which is a solution suggested by the Tanstack table developers: https://github.com/TanStack/table/discussions/4238

I've abstracted this "listener" into its own hook, so it doesn't interact with the core functionality at all -- it just listens and updates local storage accordingly.

There is a simple "getSavedTableState" method for parsing the local storage -- this has to exist outside the above hook, as it needs to run *before* the table is instantiated (as it passes the `initialState` to it), whereas the hook needs to be called *after* the table definition, as it needs to calll `table.` methods.

See comments.

TO-DO: I think we probably want a "Reset all" button somewhere, but will do that in a later issue, as I think it needs a bit of discussion where to put it/what it should look like)

# 🧪 Testing

You should be able to change any of the following table properties, and have the settings persist when reloading the table or refreshing the page:
- [ ] Table density
- [ ] Column show/hide
- [ ] Colum ordering
- [ ] "Pinned" (sticky) columns
- [ ] Column widths

(There is one tiny bug -- see if you can find it, and is it important enough to try and fix? (I don't think so for now))

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
@@ -9,19 +9,23 @@ import {
   CheckboxEmptyIcon,
   CheckboxIndeterminateIcon,
 } from '@common/icons';
+import { useTableLocalStorage } from './useTableLocalStorage';
+import { useEffect } from 'react';
 
-interface NonPaginatedTableConfig<T extends MRT_RowData>
+export interface BaseMRTableConfig<T extends MRT_RowData>
   extends MRT_TableOptions<T> {
+  tableId: string; // key for local storage
   onRowClick?: (row: T) => void;
   isLoading: boolean;
 }
 
 export const useBaseMaterialTable = <T extends MRT_RowData>({
+  tableId,
   isLoading,
   onRowClick,
   state,
   ...tableOptions
-}: NonPaginatedTableConfig<T>) => {
+}: BaseMRTableConfig<T>) => {
   const table = useMaterialReactTable<T>({
     enablePagination: false,
     enableColumnResizing: true,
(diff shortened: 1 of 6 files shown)


client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
```diff
@@ -9,19 +9,29 @@ import {
...
 }
 
 export const useBaseMaterialTable = <T extends MRT_RowData>({
+  tableId,
   isLoading,
   onRowClick,
   state,
   ...tableOptions
 }: BaseTableConfig<T>) => {
+  const initialState = useRef(getSavedTableState(tableId));
```
CarlosNZ, line 33: Made it a ref as it never changes after initial load. Could also memoise, but I think this is slightly more efficient and robust.

client/packages/common/src/ui/layout/tables/material-react-table/useTableLocalStorage.ts
```diff
@@ -0,0 +1,98 @@
...
+    ),
+    columnPinning: pinned,
+    columnSizing,
+    columnOrder,
+  };
+
+  return tableState;
+};
+
+export const resetSavedTableState = (tableId: string) => {
```
CarlosNZ, line 96: Not currently used, but we should implement at some point

client/packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -161,6 +161,8 @@ const OutboundShipmentListViewComponent: FC = () => {
           header: '',
           enableColumnActions: false,
           enableSorting: false,
+          enableResizing: false,
```
CarlosNZ, line 164: This column leaves a weird floating handle (since the header string is empty), so easiest to just turn off this property.

client/packages/common/src/ui/layout/tables/material-react-table/useTableLocalStorage.ts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+export const getSavedTableState = (tableId: string) => {
+  const savedString = localStorage.getItem(
+    `@openmsupply-client/tables/${tableId}`
+  );
+  const savedData = savedString ? JSON.parse(savedString) : {};
+
+  const {
+    density = 'comfortable',
```
lache-melvin, line 76:
```suggestion
    density = 'compact',
```

lache-melvin (approved):
Niiiice!! Feels clean, works great 🔥

PR #9204, opened 2025-09-09
8771 ledger fix no linked invoice line

### LINKED ISSUE #8771: Ledger discrepancy check and fix when no invoice_lines
This has been noted in TL

We've missed a use case where stock_line exists with positive total/available but no invoice_lines exists at all. This would lead to this stock missing from reports (i.e stock date calculated with item or stock_line ledger would not take that batch into account)

Add this check to ledger discrepancy view and make a new fix use case. 
Create a fix. Update doc.

Would suggest that the fix deletes those stock lines

How far back to go for inventory adjustment ? Or do we just nullify stock_line (since it was not used yet, is it going to be) ?

Chris-Petty:
Fixes #8771 

# 👩🏻‍💻 What does this PR do?

Adds ledger fix that will delete orphan stock_lines

# 🧪 Testing

- [ ] WELL... you could install OG central server 8.4.1 with whatever OMS central version and [replicate this bug](https://github.com/msupply-foundation/msupply/issues/17137).
- [ ] Then upgrade your OMS central to this version and see if it fixes ledger issues.
- [ ] For completeness, repeat step 1 then issue some of the bad stock line out in an outbound shipment or prescription before upgrading. The upgrade should make an inventory adjustment adding the stock rather than deleting it.
- [ ] For completeness, repeat step 1 then add the bad stock line out to a stocktake (don't finalise it or count it, just add it) before upgrading. The upgrade should make an inventory adjustment adding the stock rather than deleting it.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/views/mod.rs
```diff
@@ -195,57 +195,74 @@ pub(crate) fn rebuild_views(connection: &StorageConnection) -> anyhow::Result<()
...
-                  AND current_balance.t_q = current_balance.a_q
-              )
-              OR (
-                  allocated_not_picked.q IS NOT NULL
-                  AND current_balance.a_q + allocated_not_picked.q = current_balance.t_q
-              )
-          )
-      );
+  SELECT
+    current_balance.stock_line_id
```
Chris-Petty, line 246: The key here is selecting from current_balance and joining on running_balance instead of the other way around. And the `OR running_balance.q IS NULL;` at the end

andreievg: There has been a change in 2.11 to move view to their own files, this change has the potential of being lost in conflicts resolution

Chris-Petty: I was hoping to just resolve the conflicts proper? I would say that the tests should fail but I didn't actually add a test for this - I should!

Chris-Petty: now we have https://github.com/msupply-foundation/open-msupply/pull/9204/commits/0dabdf46ddfc702e95f22a72603136151bb0f944

andreievg:
@Chris-Petty I am keen to approve this one, there is just that use case which I saw in testing, where CI is not yet finalised, but there is stock already in the system. 

How is that one dealt with for future use cases (would all OG central servers be upgraded and this is a no case ? Do we need to zero out or delete these ones ?).

Secondary what is the scope of this use case for existing sites (where item_line exists for SI transact for status that shouldn't have that item_line).

Both of the above can be done separately to this PR, one this for this PR is to update README.md with the use case and status for that use case

Chris-Petty:
> @Chris-Petty I am keen to approve this one, there is just that use case which I saw in testing, where CI is not yet finalised, but there is stock already in the system.
> 
> How is that one dealt with for future use cases (would all OG central servers be upgraded and this is a no case ? Do we need to zero out or delete these ones ?).
> 

As it is this fix deals with lines after the CI is finalised. OG central upgrade would prevent future occurrences, but any current "confirmed" invoices that have had items added could still fall victim as the orphaning step is when OMS deletes all invoice_lines and recreates them in the invoice transfer processors.

If we do nothing and install with this PR, then these lines will have a ledger fix inventory adjustment introducing the stock created, the same fallback as when this stock error occurs but the user has already issued it (so no longer an orphan).

We could get a bit finicky to try get rid of the stock lines still, though. We could delete them if:

1. They have only one related trans_line
2. That one trans_line is from a "picked" inbound shipment

> Secondary what is the scope of this use case for existing sites (where item_line exists for SI transact for status that shouldn't have that item_line).
> 

As above it's pretty niche - would you like me to investigate?

> Both of the above can be done separately to this PR, one this for this PR is to update README.md with the use case and status for that use case

I've updated readme. Apparently I hadn't pushed removing the extra message text from the log [17d8b0a](https://github.com/msupply-foundation/open-msupply/pull/9204/commits/17d8b0a9cc8565673e6e29f742e96ea17febd7cf) sorry, is there now!

Chris-Petty:
> Looks great, thanks @Chris-Petty.
> 
> I guess next step is to think about how we would notify affected users about these stocklines (btw do we record item name, batch name, expiry, quantity in the log ? probably should for that task)

The log has the whole StockLineRow struct in debug format

PR #9216, opened 2025-09-09
#9205 Pagination for new tables

### LINKED ISSUE #9205: Pagination -- re-implement "Rows per page" functionality
Also need to persist this value to local storage (or should it be in URL query?)

CarlosNZ:
Fixes #9205

# 👩🏻‍💻 What does this PR do?

Basically just replicated the existing functionality, except for the "rows-per-page" value, which is now included in the URL query.

Screenshot 2025-09-10 at 11 26 15 AM

Screenshot 2025-09-10 at 11 26 25 AM

See comments

# 🧪 Testing

- [ ] Pagination works as before
- [ ] Changing rows per page works correctly and updates the URL query (except for the default value (20))
- [ ] Looks correct compared with previous

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
```diff
@@ -49,21 +45,16 @@ export const useUrlQueryParams = ({
...
   const { urlQuery, updateQuery } = useUrlQuery({
     skipParse,
   });
 
-  const [storedRowsPerPage] = useLocalStorage(
-    '/pagination/rowsperpage',
-    DEFAULT_RECORDS_PER_PAGE
-  );
-  const rowsPerPage = storedRowsPerPage ?? DEFAULT_RECORDS_PER_PAGE;
-
```
CarlosNZ, line 66: Removed -- don't think it'll affect current functionality, just means will always use 20 until using the new tables (don't think current one was working properly anyway)

client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.tsx
```diff
@@ -119,6 +127,53 @@ export const usePaginatedMaterialTable = <T extends MRT_RowData>({
...
     manualPagination: true,
     manualSorting: true,
+    enableBottomToolbar: true, // required for pagination
+    enablePagination: true,
+    paginationDisplayMode: 'pages',
+    muiBottomToolbarProps: {
+      sx: {
+        '& .MuiInputLabel-root': {
+          fontSize: '0.9em',
+        },
```
CarlosNZ, line 137: The `muiPaginationProps` targets the pagination component only, not the rows selector, so have to target this via the outer Toolbar component

client/packages/common/src/ui/layout/tables/material-react-table/usePaginatedMaterialTable.tsx
```diff
@@ -119,6 +127,53 @@ export const usePaginatedMaterialTable = <T extends MRT_RowData>({
...
+          </Typography>
+          <Typography sx={{ fontWeight: 'bold', marginRight: '4px' }}>
+            {xToY}
+          </Typography>
+          <Typography sx={{ marginRight: '4px' }}>{t('label.of')}</Typography>
+          <Typography sx={{ fontWeight: 'bold', marginRight: '4px' }}>
+            {totalCount}
+          </Typography>
+        </Box>
+      );
```
CarlosNZ, line 174: This just copied from existing "PaginationRow.tsx"

client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
```diff
@@ -134,16 +129,21 @@ export const useUrlQueryParams = ({
     onClearFilterRule: key => updateFilterQuery(key, ''),
     filterBy: getFilterBy(),
   };
+
+  const pageSize =
+    urlQuery['pageSize'] && typeof urlQuery['pageSize'] === 'number'
+      ? urlQuery['pageSize']
+      : DEFAULT_RECORDS_PER_PAGE;
```
lache-melvin, line 136: storedRowsPerPage definitely was working before, just tested on develop. I'd default to that in here where you've been using DEFAULT_RECORDS_PER_PAGE... we don't want to break ability to change page size in the interim

lache-melvin (changes requested):
Not quite working with selected rows:
https://github.com/user-attachments/assets/12e166de-d03b-4822-82f0-e7c01f4537a5

First noticed that we end up with 2 footers - which maybe not quite ideal, and then it looked like you'd be able to select across multiple pages? Alas it's just highlight the second item on any page... seeing as we only have this page of data from backend, probably makes sense to stick with our existing pattern, only show pagination toolbar when nothing selected?

CarlosNZ:
Screenshot 2025-09-10 at 2 48 02 PM

CarlosNZ:
> Not quite working with selected rows: https://github.com/user-attachments/assets/12e166de-d03b-4822-82f0-e7c01f4537a5
> 
> First noticed that we end up with 2 footers - which maybe not quite ideal, and then it looked like you'd be able to select across multiple pages? Alas it's just highlight the second item on any page... seeing as we only have this page of data from backend, probably makes sense to stick with our existing pattern, only show pagination toolbar when nothing selected?

Oh, okay. I've updated it to only show the pagination footer if nothing is selected -- is that what you meant?

Looks like this:

client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
```diff
@@ -81,9 +84,13 @@ export const useUrlQueryParams = ({
     [updateQuery]
   );
 
-  const updatePaginationQuery = (page: number) => {
+  const updatePaginationQuery = (page: number, pageSize?: number) => {
```
lache-melvin, line 87: Do you think we should keep the save of rows per page? Could persist to local storage here..?

lache-melvin (approved):
Looks great, thanks Carl 🚀

client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
CarlosNZ: Yeah, okay 👍

e56b4d8a7b

PR #9224, opened 2025-09-10
9203 default hide on mobile

### LINKED ISSUE #9203: Re-implement "defaultHiddenOnMobile" column setting (and any other properties that may have slipped through in this process)

lache-melvin:
Fixes #9203

# 👩🏻‍💻 What does this PR do?

Implements column visibility for hidden columns (e.g. behind pref) and default hidden on mobile.

A number of todos in the code, am working on part two with the reusable column definitions, but this was big enough as it was

# 🧪 Testing

- [ ] Have simplified mobile UI pref on (include pack to 1 etc.) on a small screen
- [ ] Go to outbound shipment list view
- [ ] Customer name, reference, comment columns don't show
- [ ] On a larger screen, they do show
- [ ] Back on the smaller screen, you can enable these columns via the column manager in top right
- [ ] Navigate away and back, the enabled columns are still there
- [ ] With ManageVVMStatusForStock pref off, there is no VVM status column in the detail view (including column manager)
- [ ] Turn the pref on - VVM status column shows (between expiry date and location)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
+++ b/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
@@ -55,9 +55,7 @@ export const useUrlQueryParams = ({
   // usually a 'startsWith'
   const skipParse = filters.length > 0 ? filters.map(f => f.key) : ['filter'];
 
-  // TO-DO: Remove this after all tables have been replaced with the new table
-  // library
-  const [storedRowsPerPage] = useLocalStorage(
+  const [storedRowsPerPage, setStoredRowsPerPage] = useLocalStorage(
     '/pagination/rowsperpage',
     DEFAULT_RECORDS_PER_PAGE
   );
@@ -84,13 +82,17 @@ export const useUrlQueryParams = ({
     [updateQuery]
   );
 
-  const updatePaginationQuery = (page: number, pageSize?: number) => {
+  const updatePaginationQuery = (
+    page: number,
+    pageSize: number = rowsPerPage
+  ) => {
     // Page is zero-indexed in useQueryParams store, so increase it by one
     updateQuery({
       page: page === 0 ? '' : page + 1,
       pageSize:
         pageSize && pageSize !== DEFAULT_RECORDS_PER_PAGE ? pageSize : '',
     });
+    setStoredRowsPerPage(pageSize);
   };
 
   const updateFilterQuery = (key: string, value: string) => {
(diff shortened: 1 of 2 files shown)


client/packages/common/src/ui/layout/tables/material-react-table/useTableLocalStorage.ts
```diff
@@ -41,18 +42,31 @@ export const useTableLocalStorage = <T extends MRT_RowData>(
...
       `@openmsupply-client/tables/${tableId}`,
       JSON.stringify({
         density,
-        hidden,
-        pinned: columnPinning,
         columnSizing: debouncedColumnSizing,
-        columnOrder,
+
+        // These can change with different preferences, or versions of OMS
+        // We should only save them when the user actually customises something
```
lache-melvin, line 58: When nothing in local storage, we have some defaults which populate table `initialState` - in previous version, these defaults were then being saved as the user customisation.

If you then enabled different prefs, we were still respecting that customisation. This logic instead determines that if the table state hasn't changed from whatever was it's initial state, we don't need to save anything (so if the defaults change, user will see that)

client/packages/common/src/ui/layout/tables/material-react-table/useMaterialTableColumns.tsx
```diff
@@ -7,95 +7,100 @@
...
-        }
-
-        // Add alignment styling
-        if (align) {
-          columnDefProperties.muiTableBodyCellProps = params =>
-            mergeCellProps(
+  const simplifiedMobileView = useSimplifiedTabletUI();
+  const tableDefinition = useMemo(() => {
+    const columns = omsColumns
+      .filter(col => col.includeColumn !== false)
```
lache-melvin, line 27: if `include` is false, don't include in the column definitions passed into the table at all. Essentially does the same thing as all the column.push() we had before, but in one place :)

client/packages/common/src/ui/layout/tables/material-react-table/useMaterialTableColumns.tsx
```diff
@@ -7,95 +7,100 @@
...
             );
+          };
         }
+        return col;
+      });
+
+    const defaultHiddenColumns = simplifiedMobileView
+      ? columns
+          .filter(col => col.defaultHideOnMobile)
+          .map(col => String(col.id ?? col.accessorKey))
```
lache-melvin, line 55: These will be passed as the default hidden columns, if no pref is stored in localstorage for this table yet

Issue #9165, opened 2025-09-07
Statuses and editing


roxy-dao:
**Note:** The `Authorise purchase orders` permission should be used for privileges down below regardless of Auth preference

| Internal System Status | Status as shown for users | Can edit lines? | Comment |
|-----------------------|--------------------------|-----------------|---------|
| new                   | New                      | Yes             | If authorisation preference is not on, “Ready for sending” button should be available and the rest should be read-only (ie. “Order Sent”).<br>If authorisation preference is on, “Ready for approval” button should be available and the rest should be read-only. |
| request_approval (rename authorised)      | Ready for approval       | Yes             | Only applies when the global preference to authorise POs is on; otherwise skip straight to confirmed.<br>Only users with auth privileges can change status to confirmed.<br>Should show a prompt to ask them if they are ready to confirm, and therefore send to the supplier.<br>Log who has approved. |
| Confirmed             | Ready for sending        | Only quantity    | Can not confirm a PO if any line has item_store_join.do_not_order set to true (alert which lines need removing).<br>Changing quantities is privileged, and only for people with auth permissions.<br>Must log change. |
| sent                  | Order sent               | Only adjusted quantity, and finalised status | Changing adjusted quantities is privileged, and only for people with auth permissions.<br>Can not adjust below the amount already received.<br>Changing original quantities is not allowed. |
| finalised             | Closed for receiving     | No              | On change: Pop-up displayed noting there are items with outstanding quantities will be marked as closed for receiving? |

PR #9251, opened 2025-09-15
9165 Purchase Order statuses and editing - Pt 1 

aimee-mcneil-melville:
Part 1 of #9165

Splitting to keep PRs smaller..
Not to be merged without part 2 front end changes - will break things!

# 👩🏻‍💻 What does this PR do?

Back end changes for statuses and editing permissions/validation

- Update activity log statuses
- Update PO status enum & rename throughout
- Renamed `authorised_datetime` to `request_approval_datetime`
- Renamed `CannotAdjustRequestedQuantity` to `CannotEditRequestedQuantity` and added `CannotEditAdjustedQuantity` errors/checks
- Check user permissions when updating PO status from `Request approval` -> `Confirmed`
- Check user permissions when updating PO lines `adjusted quantity`
- Tests!

Received units not yet populated from the goods received - value will always be 0 for now but the validation and test is in place

# 🧪 Testing

- [ ] Tests pass
- [ ] Manual testing to be done with part 2

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/purchase_order_line/update/mod.rs
+++ b/server/service/src/purchase_order_line/update/mod.rs
@@ -99,7 +99,7 @@ pub fn update_purchase_order_line(
                         currency_id: None,
                         foreign_exchange_rate: None,
                         shipping_method: None,
-                        sent_datetime: None,
+                        sent_datetime: Some(NullableUpdate { value: None }),
                         contract_signed_date: None,
                         advance_paid_date: None,
                         received_at_port_date: None,


GeronimoJohn (approved):
LGTM! Some solid hard work coming through here. Thanks Aimee. 

I’ve tested the new status functionality in the GraphQL Playground, and everything is working well — both querying and updating statuses! 🙏🏽

One thing I noticed: I was able to update the status from "Request Approval" to "Confirmed" — PO Status Request Approval → PO Status Confirmed. I believe this is allowed because the preference isn’t turned on yet. Once it's enabled, the validation check should kick in.

Screenshot 2025-09-16 at 4 34 07 PM

It’s something we might want to support. That said, since the “Requires Approval” button will be hidden in the UI? It shouldn’t be an issue — would only matter if someone tries to update the status via an API call haha

Left a few notes too 😁 🙏🏽

server/service/src/purchase_order/update/validate.rs
```diff
@@ -15,20 +15,22 @@ pub fn validate(
...
 ) -> Result<PurchaseOrder, UpdatePurchaseOrderError> {
     let purchase_order = PurchaseOrderRepository::new(connection)
         .query_by_filter(PurchaseOrderFilter::new().id(EqualFilter::equal_to(&input.id)))?
         .pop()
         .ok_or(UpdatePurchaseOrderError::PurchaseOrderDoesNotExist)?;
 
-    if input.status == Some(PurchaseOrderStatus::Authorised) {
-        let is_authorised = AuthorisePurchaseOrder
+    // check user has permission to authorise purchase order, if authorisation is required
+    if input.status == Some(PurchaseOrderStatus::Confirmed) {
```
GeronimoJohn, line 26: We should also add support for the transition to `Request Approval` as per the issue: 

_Only applies when the global preference to authorise POs is on; otherwise skip straight to confirmed._

GeronimoJohn (commented):
Apologies, have went back on my review. After our conversation realised we should consider the case mentioned below

server/service/src/purchase_order/update/validate.rs
aimee-mcneil-melville: Great thank you, have added to part 3!

server/repository/src/migrations/v2_11_00/rename_authorised_on_purchase_order_status_enum.rs
```diff
@@ -0,0 +1,29 @@
+use crate::migrations::*;
+
+pub(crate) struct Migrate;
+
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "rename_authorised_on_purchase_order_status_enum"
```
roxy-dao, line 7: nit can't this all be done in one migration? It's confusing to have multiple migrations for the same time in different files

server/repository/src/migrations/v2_11_00/update_purchase_order_activity_log_type_enum.rs
```diff
@@ -0,0 +1,36 @@
...
+                    ALTER TYPE activity_log_type RENAME VALUE 'PURCHASE_ORDER_AUTHORISED' to 'PURCHASE_ORDER_REQUEST_APPROVAL';
+                "#
+            )?;
+
+            sql!(
+                connection,
+                r#"
+                    ALTER TYPE activity_log_type ADD VALUE 'PURCHASE_ORDER_SENT';
+                "#
+            )?;
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Thanks for the thorough review John!

I have updated the sent date and refined migrations + moved them to V2.12

GeronimoJohn (approved):
GOOOD STUFFF. Thanks for putting in the hard yards @aimee-mcneil-melville. This is looking good to go.

Was able to change statuses - with/without Authorisation, adjust number of units changes status back to Confirmed from Sent. And yea just awesome stuff! 

Just left a few nits. But the error one might be worth supporting?

GeronimoJohn:
Just have to update package.json. Running the server failed for me cause mismatch versions haha
Screenshot 2025-09-22 at 5 32 15 PM

PR #9253, opened 2025-09-16
8915 purchase order line insert refactor

### LINKED ISSUE #8915: Purchase orders insert refactor
Don't need separate methods to insert / insert from csv, and also don't even need the insert method. Should all be refactor to upsert for frontend consumption. Also default::default should only be used in mocks and not for normal operations since we can't guarantee that we won't forget to add fields! 

Or maybe you do need the insert method, but would merge insert / insert from csv

GeronimoJohn:
Fixes #8915 

# 👩🏻‍💻 What does this PR do?

Refactored the purchase order line insert methods. Previously, there were two—one for standard inserts and one for CSVs—with only minor differences like handling None values and validating via item_id vs. item_code. These have been merged into a single method to reduce duplication.

I initially started with logic to generate input based on the mode (e.g. CSV or Standard), since each had a different shape. But there's an open issue to align the CSV template with the line edit, which makes their structure the same. So I’ve removed the mode-specific logic that generates the input and simplified the code to handle both cases.

P.S. Merging this to develop will effectively break the processing logic of inserting a CSV for a Purchase Order. Branching off this PR to solve that issue

# 🧪 Testing

- [ ] Navigate to your preferred GraphQL Playground

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/purchase_order_line/insert/mod.rs
```diff
@@ -62,23 +77,33 @@ pub fn insert_purchase_order_line(
...
+                item_id: Some(item.id.clone()),
+                item_code: None,
+                ..input
+            };
+
             let new_purchase_order_line =
-                generate(connection, &ctx.store_id.clone(), item, input.clone())?;
+                generate(connection, &ctx.store_id.clone(), item, generate_input)?;
+
+            // Only log activity for standard mode (not CSV bulk imports)
```
roxy-dao, line 96: Why won't there be a long for CSV? Could be a 'bulk created from CSV log'...

server/service/src/purchase_order_line/insert/mod.rs
```diff
@@ -62,23 +77,33 @@ pub fn insert_purchase_order_line(
             let validate_input = validate::ValidateInput {
```
roxy-dao, line 77: This doesn't need to be here anymore no? Can just pass the whole input into the validate method...

server/service/src/purchase_order_line/insert/mod.rs
```diff
@@ -62,23 +77,33 @@ pub fn insert_purchase_order_line(
...
-                // TODO amend default value if we extend standard insert line input
-                requested_pack_size: input.requested_pack_size.unwrap_or_default(), // Default value
+                item_id: input.item_id.clone(),
+                item_code: input.item_code.clone(),
+                requested_pack_size: input.requested_pack_size.unwrap_or_default(),
                 manufacturer_id: input.manufacturer_id.clone(),
             };
             let item = validate(&ctx.store_id.clone(), &validate_input, connection)?;
+
+            let generate_input = InsertPurchaseOrderLineInput {
```
roxy-dao, line 87: Same here... don't need this the item is being returned from validate method, so we use that to get id and code

roxy-dao (dismissed):
Thanks John~ some comments below and what about the frontend?

server/service/src/purchase_order_line/insert/mod.rs
GeronimoJohn: I was about to ask about that. Before refactor the insert_csv didn't have log. But makes a lot of sense to have it as well

server/graphql/purchase_order_line/src/mutations/insert.rs
```diff
@@ -8,20 +8,30 @@ use graphql_types::types::IdResponse;
...
 #[graphql(name = "InsertPurchaseOrderLineInput")]
 pub struct InsertInput {
     pub id: String,
     pub purchase_order_id: String,
-    pub item_id: String,
+    pub mode: InsertModeInput,
+    // Standard mode uses item_id
+    pub item_id: Option<String>,
+    // CSV mode uses item_code
+    pub item_code: Option<String>,
```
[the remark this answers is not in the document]
GeronimoJohn: Applied changes! https://github.com/msupply-foundation/open-msupply/pull/9253/commits/70c672206e2efee2b5b22aa39facc53e7b54aff8. Can find changes in there for other comments as well

PR #9271, opened 2025-09-17
9264 url state for tables

### LINKED ISSUE #9264: Use URL state for all new tables
Non-paginated table filters are not being updated via URL like they used to be (sort & filter - don't send this to backend though). Also fix `clear sort` for paginated tables.

lache-melvin:
Fixes #9264

# 👩🏻‍💻 What does this PR do?

Moves table filter/sorting state to always be in the URL. 

`usePaginatedMaterialTable` expects the passed in data to be sorted/filtered (manual). `useNonPaginated` handles this internally, just using the URL to hold the state. This means, when using non-paginated table, we shouldn't pass those URL params into the backend query.

# 🧪 Testing

- [ ] Clicking column headers - sorts desc, asc, then off
- [ ] Sort icon only shows when actively sorting
- [ ] Sort and filtering persists on page refresh, on both list and detail view

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
```diff
@@ -82,6 +83,8 @@ export const useUrlQueryParams = ({
     [updateQuery]
   );
 
+  const clearSort = () => updateQuery({ sort: undefined, dir: undefined });
```
lache-melvin, line 86: We didn't currently have a way to clear sorting - also had to change the above useEffect to only reset the initial sort on page mount, rather than any time we don't have a currently applied sort.

client/packages/common/src/ui/layout/tables/material-react-table/getTableDisplayOptions.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { MRT_RowData, MRT_TableOptions } from 'material-react-table';
+import {
+  CheckboxCheckedIcon,
+  CheckboxEmptyIcon,
+  CheckboxIndeterminateIcon,
+} from '@common/icons';
+import { MenuItem, Typography } from '@mui/material';
+import { ColumnDef } from './types';
+
+export const getTableDisplayOptions = <T extends MRT_RowData>(
```
lache-melvin, line 11: Pulled all this out of `useBase` hook, so that hook focuses on functionality

client/packages/common/src/ui/layout/tables/material-react-table/getTableDisplayOptions.tsx
```diff
@@ -0,0 +1,178 @@
...
+  muiTableHeadCellProps: ({ column, table }) => ({
+    sx: {
+      fontWeight: 600,
+      fontSize: table.getState().density !== 'spacious' ? '0.9em' : '1em',
+      lineHeight: 1.2,
+      verticalAlign: 'bottom',
+      justifyContent: 'space-between',
+      '& .MuiTableSortLabel-root': {
+        display: column.getIsSorted() ? undefined : 'none',
+      },
```
lache-melvin, line 60: this was the only addition here

client/packages/common/src/ui/layout/tables/material-react-table/useTableFiltering.ts
```diff
@@ -108,28 +93,30 @@ export const useManualTableFilters = <T extends MRT_RowData>(
   };
 
   return {
-    filterState,
-    filterHandlers: { onColumnFiltersChange: handleFilterChange },
+    columnFilters: filterState,
+    onColumnFiltersChange: handleFilterChange,
```
lache-melvin, line 97: this hook is a little simpler now, that we always sort the filter state in the URL - the `manualFiltering` only controls whether table will use its internal filter methods, or expect filtered data to be passed in

CarlosNZ:
Haven't had a proper look at the code yet, but there's a couple of quirks I noticed in an initial test:

- The "select" filter for "Unit name" isn't working -- it neither updates the URL nor changes the data
- I can't clear the values for the Date filters -- when clicking the "X", the value disappears from the input, but doesn't update the URL (it works if I select "Clear filter" from the drop-down menu)

CarlosNZ (changes requested):
Looking really good, I like the tidy-ups.

Here's the diff to fix the two problems I mentioned:

https://github.com/msupply-foundation/open-msupply/commit/45c02366278911b3a4c1fc61b423d57a3ce5026f#r165999812

lache-melvin:
Thanks for that investigation Carl! 🙏 Have merged

PR #9277, opened 2025-09-18
9252 shipping method table

### LINKED ISSUE #9252: Shipping method table
- [ ] Migrations for a new shipping_method table:
id: text
method: text
deleted_datetime: Optional<Datetime>
- [ ] Query to get all shipping methods
- [ ] Pull-only from OG for now
- [ ] Autocomplete component to select shipping method
- [ ] Purchase order's shipping_method field should reference this
- [ ] Change side-panel component from text to the above autocomplete

### Moneyworks Jobcode 🧰
OMS:DPOF. DFAT resource

GeronimoJohn:
Fixes #9252

# 👩🏻‍💻 What does this PR do?

Created a table for Shipping Method, added translations work - only pulls data from OG, and created an Autocomplete that can be reused for ShippingMethod. 

The ShippingMethodAutocomplete have replaced the Shipping Method input for the Side Panel of Purchase Order

Screenshot 2025-09-18 at 3 37 41 PM

Focusing on the `shipping_method` field in `PurchaseOrder`: ideally, for an Autocomplete, this should be an ID field—`shipping_method_id`.

The nice thing about using an ID is that the frontend Autocomplete gets way cleaner. Instead of trying to match by text or `method name`, we can just look up the selected shipping method by its `ID`. 

At first, switching the field to be `shipping_method_id` felt natural. But after checking OG, `delivery_method` / `shipping_method` is just plain text:

Screenshot 2025-09-18 at 3 44 08 PM

Changing it to an `ID` (or `shipping_method_id`) could add complexity at the translation layer: OG -> OMS. In OG, it seems you can just write anything for delivery_method since it’s just a text input. If that value doesn’t match anything in the `shipping_method` table in OMS, it may appear as blank there.

Kept it simple and just used the `method name` to get the shipping method through the UI (Check out the `OtherSection` component if you need a reference)

# 🧪 Testing

- [ ] In OG, click on Customer within the toolbar (top bit of the screen)
- [ ] Click on `Show Shipping Methods` and click on Plus Icon to add some shipping methods. 
- [ ] After creating Shipping Methods, run sync in OMS
- [ ] Navigate to a Purchase Order (or create one)
- [ ] Click on the more button and see the Shipping Method Autocomplete under Other Section
- [ ] Select your shipping method and it should instantly save (refresh screen to sanity check)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/shipping_method/mod.rs
```diff
@@ -0,0 +1,20 @@
+use repository::shipping_method::{ShippingMethod, ShippingMethodFilter};
+
```
roxy-dao, line 2: Nit: I would remove all the white space between the imports... no one really pays attention to them so don't need to separate them like this 👀

server/graphql/types/src/types/shipping_method.rs
```diff
@@ -0,0 +1,53 @@
...
+impl ShippingMethodNode {
+    pub async fn id(&self) -> &str {
+        &self.row().id
+    }
+
+    pub async fn method(&self) -> &str {
+        &self.row().method
+    }
+
+    pub async fn deleted_datetime(&self) -> &Option<NaiveDateTime> {
```
roxy-dao, line 21: Use UTC datetime please 

`.map(|d| Datetime::<Utc>::from_naive_utc_and_offset(d, Utc))`

client/packages/system/src/ShippingMethod/api/useShippingMethodGraphQL.ts
```diff
@@ -0,0 +1,15 @@
+import {
+  useAuthContext,
+  useGql,
+  useQueryClient,
+} from '@openmsupply-client/common';
+import { getSdk } from './operations.generated';
+
+export const useShippingMethodGraphQL = () => {
```
roxy-dao, line 8: Should really make this common at some point ;--;

roxy-dao (changes requested):
Thanks John! Just a few changes please (:

server/repository/src/migrations/v2_11_00/add_shipping_method_table.rs
```diff
@@ -0,0 +1,24 @@
...
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "add_shipping_method_table"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                CREATE TABLE shipping_method (
```
roxy-dao, line 14: And please move migrations to 2.12~ tyty

GeronimoJohn: Thanks Roxy, please see changes here: https://github.com/msupply-foundation/open-msupply/pull/9277/commits/edba4b1480d030ec3857ff7199b7010f15541b95

PR #9293, opened 2025-09-19
9031 - Include Adjusted Units value when purchase order is confimed

### LINKED ISSUE #9031: The Adjusted column does not gets updated for the first time when you add new items in confirmed purchase orders
## What went wrong? 😲

https://github.com/user-attachments/assets/f3bc472e-36b3-4001-bf52-06288fe1a09b

## How to Reproduce 🔨

1. Open any confirmed PO
2. Add new line
3. Enter the adjusted quantity
4. The quantity does not gets saved and cannot be seen adjusted window (BUG)

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #9031 

# 👩🏻‍💻 What does this PR do?

The adjusted units will now be automatically generated when a purchase order status is changed to Confirmed

Not quite as per issue.. Lines can no longer be added at Confirmed stage. However still needed fixing because if a quantity is not updated after reaching confirmed status, the adjusted amount will never be populated!

Simplified the utils function as adjusted units is only editable from confirmed status onward (sent status coming in 9165), and requested units only editable before confirmed

# 🧪 Testing

- [ ] Create a purchase order and add a line with a requested quantity. See no adjusted units value in the detail table
- [ ] Proceed to Confirmed status -> see the adjusted units populate with the value in requested column
- [ ] Update adjusted units in line edit view -> see only the adjusted value change

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/purchase_order/update/generate.rs
+++ b/server/service/src/purchase_order/update/generate.rs
@@ -186,6 +186,13 @@ fn update_lines(
                     PurchaseOrderStatus::Confirmed => {
                         line.purchase_order_line_row.status =
                             repository::PurchaseOrderLineStatus::Sent;
+
+                        // Insert requested units value to adjusted units field when purchase order changes to Confirmed status
+                        // Requested units is a fallback as status can change backwards from Sent to Confirmed, so value for adjusted units may already exist
+                        line.purchase_order_line_row.adjusted_number_of_units = line
+                            .purchase_order_line_row
+                            .adjusted_number_of_units
+                            .or(Some(line.purchase_order_line_row.requested_number_of_units))
                     }
                     PurchaseOrderStatus::Finalised => {
                         line.purchase_order_line_row.status =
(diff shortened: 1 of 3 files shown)


server/service/src/purchase_order/update/generate.rs
```diff
@@ -186,6 +186,13 @@ fn update_lines(
                     PurchaseOrderStatus::Confirmed => {
                         line.purchase_order_line_row.status =
                             repository::PurchaseOrderLineStatus::Sent;
+
+                        // Insert requested units value to adjusted units field when purchase order changes to Confirmed status
+                        // Requested units is a fallback as status can change backwards from Sent to Confirmed, so value for adjusted units may already exist
+                        line.purchase_order_line_row.adjusted_number_of_units = line
+                            .purchase_order_line_row
+                            .adjusted_number_of_units
+                            .or(Some(line.purchase_order_line_row.requested_number_of_units))
```
aimee-mcneil-melville, line 195: Coming with issue #9165 is purchase order statuses can go BACK from Sent to Confirmed, so the value shouldn't be replaced every time the new status is Confirmed  - if Adjusted value exists already it remains (couldn't test in this branch though 😢 )

roxy-dao (approved):
Thanks! Let's get this in and a separate issue can be created to double check with new statuses (:

PR #9334, opened 2025-09-23
9298 Fix negative balance for RnR Forms

### LINKED ISSUE #9298: Negative 'Initial Balance' (-1) in an R&R form
## What went wrong? 😲

Seeing this which prevents me from finalising an R&R form: 

Image

This is Item's Ledger so I am not sure where the -1 is coming from?

Image

This is the same item's ledger but in OG mSupply

Image

## Expected behaviour 🤔

Should not show negative Initial Balance in an R&R form

## How to Reproduce 🔨

1. I am unsure, sorry... Here are the details I know
2. I was creating an R&R form for a Master List for the period of June-August 2025
3. The R&R form was created today (22/09/2025)
4. Item Ledger is not showing any weird transactions as per the screenshot above
5. I just created an R&R form and was prevented from saving it as the initial balance was showing a negative number
6. Ask me for login details if you need to play around with the item :)

backup of my datafile - https://mega.nz/fm/0RQVmb6Q

## Your environment 🌱

- Open mSupply Version: `V2.11.00-develop-09190330`

- Database type: `PostgreSQL`

jmbrunskill:
Fixes #9298

# 👩🏻‍💻 What does this PR do?

Previously we weren't including any stock allocated from today (utc time) in the stock we were removing from the pre-saved SOH view, now we use current time, which should include everything unless we somehow have future dated invocies.

Note:I feel like I've fixed this before, but maybe in different context? Makes sense to use some kind of shared logic for these kinds of operations that are being done in a few places now.
Was there another reason I'm not aware of for trying to use `today` rather than `now`?

There's a few ways to skin this cat now. Currently went for the smallest code diff, however could also do things like.

- [ ] Call getHistoricalStockLines for that date (does similar code)
- [ ] Use new `stock_ledger` view 

Consolidating all these things is probably a good refactor issue that I should create? Feel like we probably don't need/want it in a patch though?

# 🧪 Testing

- [ ] Remove some stock for an item in an R&R form now.
- [ ] Make sure that `Today` in utc doesn't include this new stock transaction
- [ ] Try creating an R&R form for a masterlist/period that doesn't have a previous RnR form
- [ ] Check that initial balance is correct, could be too high or too low depending on the transactions not included.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/ledger_fix/fixes/delete_unused_orphan_stock_lines.rs
+++ b/server/service/src/ledger_fix/fixes/delete_unused_orphan_stock_lines.rs
@@ -229,7 +229,7 @@ pub(crate) mod test {
             is_ledger_fixed(&connection, "legacy_stock_line_with_invoice_line"),
             Ok(false)
         );
-        assert!(logs.contains("stock_line referenced in another table:"));
+        assert!(logs.contains("stock_line referenced in another table"));
 
         // No invoice lines but referenced in stocktake line
         let mut logs = String::new();
@@ -251,6 +251,6 @@ pub(crate) mod test {
             is_ledger_fixed(&connection, "legacy_stock_line_with_stock_take_line"),
             Ok(false)
         );
-        assert!(logs.contains("stock_line referenced in another table:"));
+        assert!(logs.contains("stock_line referenced in another table"));
     }
 }
--- a/server/service/src/rnr_form/generate_rnr_form_lines.rs
+++ b/server/service/src/rnr_form/generate_rnr_form_lines.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, ops::Neg};
 
-use chrono::{Duration, NaiveDate};
+use chrono::{Duration, NaiveDate, Utc};
 use repository::{
     AdjustmentFilter, AdjustmentRepository, ConsumptionFilter, ConsumptionRepository, DateFilter,
     DatetimeFilter, EqualFilter, MasterListLineFilter, MasterListLineRepository, Pagination,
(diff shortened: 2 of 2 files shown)


lache-melvin (changes requested):
> Was there another reason I'm not aware of for trying to use today rather than now?

TBH, bet I imported that method assuming it would behave like TS Date.now(), rather than as start_of_day 🫠  Nice catch!


> Consolidating all these things is probably a good refactor issue that I should create? Feel like we probably don't need/want it in a patch though?

Agree, if not consolidating now, it should definitely be done. This is complex enough as it is, don't need it implemented 3 different ways!


Just requesting changes as it looks like a test is failing - not sure if related to these changes or other ledger fix pieces?

jmbrunskill:
Yeah looks like ledger fix tests fail... Will have a chat to chris

server/service/src/ledger_fix/fixes/delete_unused_orphan_stock_lines.rs
```diff
@@ -251,6 +251,6 @@ pub(crate) mod test {
             is_ledger_fixed(&connection, "legacy_stock_line_with_stock_take_line"),
             Ok(false)
         );
-        assert!(logs.contains("stock_line referenced in another table:"));
+        assert!(logs.contains("stock_line referenced in another table"));
```
lache-melvin, line 254: rough 🤣

PR #9348, opened 2025-09-24
#9340 Tables -- Don't show "No results" when refetching

### LINKED ISSUE #9340: When sorting or filtering, should show old results until the new ones are available (currently says no results)
Originally had this just as part of #9181, but I think it needs to be its own issue, as it's not quite as simple as I first thought.

Currently, when the data is re-loading, the MRT hooks are getting `[ ]` for data. Old tables received `undefined`, which made it easy to distinguish between data still loading and empty results. We can use the `isLoading` flag, but we still have the problem of the empty data array.

I think we need some kind of mechanism to make sure `data` stays unchanged until `isLoading` is complete.

CarlosNZ:
Fixes #9340

# 👩🏻‍💻 What does this PR do?

Not 100% sure this is the best solution, but it works. Open to alternative approaches, but I think this is fairly important -- we dont' want to show the "Nothing to display" component every time we change the sort or filter settings on a table.

Here's the difference:

Before:

After:

Basic mechanism is a small hook that returns the data, but if the data is in the `isLoading` state, it returns the previously fetched data until loading is complete.

See comments for further detail.

# 🧪 Testing

- [ ] Check sorting and filtering behaves correctly, but without showing "Nothing to display" in between each changes
- [ ] Check non-paginated tables still good
- [ ] Confirm the "Nothing to display" component does actually appear when there's actually no results

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/material-react-table/useNonPaginatedMaterialTable.ts
```diff
@@ -1,14 +1,21 @@
 import { MRT_RowData } from 'material-react-table';
 import { BaseTableConfig, useBaseMaterialTable } from './useBaseMaterialTable';
 
+const VIRTUALIZATION_THRESHOLD = 250;
```
CarlosNZ, line 4: I noticed that when the table is virtualized, there's no indication of how many total rows there are, as the scroll bars are only in proportion to the currently virtualized list, not the full table. So it seemed like a good idea to only virtualize when we need to -- is 250 a good threshold you think?

CarlosNZ: Ignore this -- have restored to original behaviour. Turns out that the `enableRowVirtualization` setting only considers the first load, and the `data` is usually undefined on first load, so we don't get a meaningful value. There's things that could be done to handle this, but it's overly complex, and doesn't really matter now that I've fixed that scroll behaviour I showed you.

client/packages/common/src/ui/layout/tables/material-react-table/usePersistDataOnRefetch.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ *
+ */
+
+import { useEffect, useRef } from 'react';
+import { MRT_RowData } from 'material-react-table';
+
+export const usePersistDataOnRefetch = <T extends MRT_RowData>(
```
lache-melvin, line 8: I think the correct way to do this is to lean on tanstack react query instead - can use `keepPreviousData` in the query, and `isFetching` for loading state.

Then shouldn't need any of your other changes:

https://github.com/msupply-foundation/open-msupply/compare/9340-keep-data-on-refetch...9340-keep-data-on-refetch(use-tanstack)

lache-melvin (changes requested):
thanks carl - nice solution, but I think we should get that behaviour from our query rather than reimplementing ourselves 😁

CarlosNZ:
> thanks carl - nice solution, but I think we should get that behaviour from our query rather than reimplementing ourselves 😁

Agree, but I thought it would be too big a job to change every list query! I guess if we do it for this one, then we'll need to make a note to also update the queries when implementing the new table?

lache-melvin:
eh, we'll be going through every view regardless, and its only a one liner change. I think that's okay as a one time sweep improvement, than to be managing this ourselves going forward

CarlosNZ:
Thanks @lache-melvin, have used most of your suggestion -- good to know about the Tanstack query options, will definitely need to make sure they get used going forward (have made a note to add it to docs)!

The only think I changed was the conditional rendering of the "NothingHere" element, as without that you still get the "Nothing Here" showing on first load, which is not great. Come have a chat if you think that can be handled differently.

lache-melvin (approved):
thank you!

PR #9352, opened 2025-09-24
9302 align dashboard numbers

### LINKED ISSUE #9302: Align dashboard numbers and filtered items
Follow up to #9119 

## What went wrong? 😲

Catalogue>items and the dashboard filter items in different ways.
So e.g. the dashboard shows there are 0 items with more than 3 months of stock the dashboard, but when you follow the link to the items page, you see 24 items.
Only the items with more than / less than X months of stock are affected.

## Expected behaviour 🤔

The number of items in the dashboard and the item page are the same, or at least clearly related.

## Solution

Laché and I came up with this (see line 46 of server/service/src/dashboard/item_count.rs for context):
1. Refactor the logic to find the root cause: abstract the logic below line 46 into their own functions, so the logic can be tested without all the database and test data setup. Instead each test and function can assume the database is all good.
2. Write tests for finding items with more than 6 MOS: when items have some AMC, item is included in result. When no AMC, item is excluded from result. Do same tests for low stock items (<3 MOS).
3. Update from only using `is_visible` to is visible or on hand, so all items with stock are included - use one of the existing tests to validate this
4. Write test for is visible or on hand logic: just arrange a few items with masterlist entries, and add one without a master list entry BUT give it a stockline entry instead, it should be included because it is therefore on hand.
5. Go back to the item query filtering, and filter out any items with 0 AMC (update tests accordingly).

zachariah-at-msupply:
Fixes #9302 

# 👩🏻‍💻 What does this PR do?

Modifies bits of the backend and frontend so that the counts of items in the dashboard line up with what's shown by the filters in catalogue> item.

It's spring! How nice to have more sun :D

Also, now the stock that has 0 AMC is excluded from months of stock related filters, because we assume 0 AMC means that there is no consumption data, so we cannot tell how many months of stock there are.

# 🧪 Testing

- [ ] Go to dashboard.
- [ ] Note the number of each statistic in the stock view widget.
- [ ] Click on each statistic's link and check that the expected number of items is shown.
- [ ] Thank you!

# 📃 Documentation

- [x] **These areas should be updated or checked**: Now the stock that has 0 AMC is excluded from months of stock related filters, because we assume 0 AMC means that there is no consumption data, so we cannot tell how many months of stock there are.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/stock_on_hand.rs
```diff
@@ -3,6 +3,7 @@ use super::StorageConnection;
 use crate::{diesel_macros::apply_equal_filter, item_link, EqualFilter, RepositoryError};
 use diesel::prelude::*;
 
+// This is actually a view in our database, not a table, but diesel treats them both the same.
```
lache-melvin, line 6: ❤️

client/packages/system/src/Item/api/hooks/useItems.ts
```diff
@@ -46,11 +46,13 @@ export const useVisibleOrOnHandItems = (queryParams: ItemParams) => {
...
     // if using the hasStockOnHand filter, replace the isVisibleOrOnHand filter with hasStockOnHand
-    if (filterBy?.['hasStockOnHand'] !== undefined) {
+    let hasStockOnHand = filterBy?.['hasStockOnHand'];
+    if (hasStockOnHand !== undefined) {
       filter = {
         ...filter,
         isVisibleOrOnHand: undefined,
-        hasStockOnHand: !!filterBy?.['hasStockOnHand'],
+        hasStockOnHand: !!hasStockOnHand,
+        isVisible: !hasStockOnHand ? true : undefined,
```
lache-melvin, line 55: Hm, this is starting to seem a bit unwieldy to me... what do you think of extracting like this?

https://github.com/msupply-foundation/open-msupply/compare/9302-align-dashboard-numbers...9302-align-dashboard-numbers(on-hand-filter-suggestion)

lache-melvin (approved):

PR #9376, opened 2025-09-25
4840 invoice translate legacy finalised date/time

### LINKED ISSUE #4840: Invoice translate legacy finalised date/time

For DHIS2 LMIS integration on Legacy Central Server we push customer invoices (outbound shipments) to DHIS2 using the confirmed date/time at the moment. It really should be the finalised date time, but OMS doesn't translate populating the finalised date field in Legacy.

This should probably map based on invoice type i.e. either `shipped_datetime` or `verified_datetime`

- [ ] In invoice sync translation, populate `finalised_date` of the legacy row from the `shipped_datetime` from OMS
- [ ] In invoice sync translation, populate `finalised_time` of the legacy row from the `shipped_datetime` from OMS. ~Note OG msupply doesn't have this field yet, but pushing it should be fine.~ It does now

### Moneyworks Jobcode 🧰

COMSCUST

roxy-dao:
Fixes #4840

# 👩🏻‍💻 What does this PR do?
Implement finalised date/time mapping to / from OG for invoices

# 🧪 Testing

- [ ] Create some inbound / outbound shipments in OG or OMS
- [ ] Sync them
- [ ] Check finalised datetimes have been populated correctly. iOG status = fn maps to Outbound, Prescription and Supplier Return -> shipped datetime OR inbound shipment, inventory add/reduct, repack, customer return = verified datetime

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/translations/invoice.rs
+++ b/server/service/src/sync/translations/invoice.rs
@@ -727,7 +727,8 @@ fn map_legacy(
 
     let finalised_datetime = data
         .finalised_date
-        .map(|finalised_date| NaiveDateTime::new(finalised_date, data.finalised_time));
+        .map(|finalised_date| NaiveDateTime::new(finalised_date, data.finalised_time))
+        .or(confirm_datetime);
 
     // Try to figure out if a legacy record was backdated
     let backdated_datetime = map_backdated_datetime(&mapping.created_datetime, &confirm_datetime);
--- a/server/service/src/sync/translations/invoice.rs
+++ b/server/service/src/sync/translations/invoice.rs
@@ -1,15 +1,10 @@
+use super::{PullTranslateResult, PushTranslateResult, SyncTranslation};
 use crate::sync::translations::{
     clinician::ClinicianTranslation, currency::CurrencyTranslation,
     diagnosis::DiagnosisTranslation, name::NameTranslation,
     name_insurance_join::NameInsuranceJoinTranslation, store::StoreTranslation, to_legacy_time,
 };
-
 use anyhow::Context;
-use util::sync_serde::{
-    date_from_date_time, date_option_to_isostring, date_to_isostring, empty_str_as_option,
-    empty_str_as_option_string, naive_time, zero_date_as_option, zero_f64_as_none,
-};
-
 use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
 use repository::{
     ChangelogRow, ChangelogTableName, CurrencyFilter, CurrencyRepository, EqualFilter, Invoice,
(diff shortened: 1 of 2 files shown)


server/service/src/sync/translations/invoice.rs
```diff
@@ -176,21 +173,20 @@ pub struct LegacyTransactRow {
     /// time in seconds
     #[serde(deserialize_with = "naive_time")]
     pub entry_time: NaiveTime, // e.g. 47046,
-    /// shipped_datetime
-    #[serde(deserialize_with = "zero_date_as_option")]
-    #[serde(serialize_with = "date_option_to_isostring")]
-    pub ship_date: Option<NaiveDate>, // "0000-00-00",
```
roxy-dao, line 182: Not sure why we're populating ship_date and arrival_date_actual with shipped and delivered... we generate these times and I'm guessing actual shipping time and arrival times will differ and we don't have that in OMS UI anyway...

Chris-Petty (changes requested):
Looks generally good, but I think may need to be blocked until we sort this out: https://github.com/msupply-foundation/msupply/issues/17242

server/service/src/sync/translations/invoice.rs
```diff
@@ -733,6 +725,10 @@ fn map_legacy(
         .confirm_date
         .map(|confirm_date| NaiveDateTime::new(confirm_date, data.confirm_time));
 
+    let finalised_datetime = data
+        .finalised_date
+        .map(|finalised_date| NaiveDateTime::new(finalised_date, data.finalised_time));
```
Chris-Petty, line 730: I'm wondering if we should fallback to confirmed time if the invoice is finalised? I think we can't guarantee that the date is populated :S.

Aside, the ledger fixes essentially use `confirmed_date` anyway, so I'm not sure anything really breaks if finalised date is wrong or None on a finalised invoice.

server/service/src/sync/translations/invoice.rs
```diff
@@ -575,12 +569,10 @@ impl SyncTranslation for InvoiceTranslation {
             linked_transaction_id: linked_invoice_id,
             entry_date: created_datetime.date(),
             entry_time: to_legacy_time(created_datetime),
-            ship_date: shipped_datetime
-                .map(|shipped_datetime| date_from_date_time(&shipped_datetime)),
-            arrival_date_actual: delivered_datetime
-                .map(|delivered_datetime| date_from_date_time(&delivered_datetime)),
```
Chris-Petty, line 581: I kind of think these were valid mappings... not sure though. Could you ask around the old guard a bit to see if they actually make sense?

ship_date being "shipped_datetime" kinda tracks. 

Also I haven't checked how these might play out to the transfer processors, any thoughts there?

Chris-Petty (changes requested):

server/service/src/sync/translations/invoice.rs
roxy-dao: OG has it as an editable field on itself, right? 
Screenshot 2025-10-02 at 1 24 58 PM

Idk can add it back, but what if the user is entering the data on a later day? Then it'll show the wrong dates? 👀

Chris-Petty (approved):
🌸

PR #9383, opened 2025-09-25
Tables -- Add filter to VVM status

CarlosNZ:
Sorry, just noticed this was missing when writing up table documentation. Just adds a "Select" filter to the VVM status column.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
@@ -135,6 +135,8 @@ export const useBaseMaterialTable = <T extends MRT_RowData>({
     onColumnFiltersChange,
     onSortingChange,
 
+    filterFromLeafRows: true,
+
     initialState: {
       density: density.initial,
       columnSizing: columnSizing.initial,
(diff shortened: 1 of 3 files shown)


lache-melvin (changes requested):
Hm, I think we need to be wary with this select variant one... couple of issues:

Firstly, the (16) for blank lines looks weird to me? I think for fields where the value might be empty, we're probably better off defining filter options explicitly? The `[multiple]` also shouldn't be in the list 🙈 
Screenshot 2025-09-26 at 9 51 19 AM

Secondly - maybe should raise a separate issue to look into this, I think this applies for all filters when we have group by on... our current implementation doesn't appear to allow for filtering, when the parent is a `[multiple]`...

e.g. I have a vaccine item with two lines, one stage 1 and one stage 2:
Screenshot 2025-09-26 at 9 52 03 AM

Despite the `Stage 1` showing in the filter options, it isn't included in the results?
Screenshot 2025-09-26 at 9 52 14 AM

Shows when we turn off group by item:
Screenshot 2025-09-26 at 9 51 43 AM

lache-melvin:
Ah, just saw a `filterFromLeafRows` option - I think that should solve!

PR #9384, opened 2025-09-25
Table developer documentation

CarlosNZ:
I actually intended this to go in the [wiki](https://github.com/msupply-foundation/open-msupply/wiki), but I realise there's no way to create PRs for that repo, so I'll just put it here initially and copy it to the wiki once approved.

I also notice we're putting dev docs in (at least) 2 different places — this folder, and the wiki, so it'd be good to standardise that. (Personally I think the wiki is better — maybe I'll bring it up in stand-up this afternoon)


doc/patterns/tables.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+> [!NOTE]
+> - Usage is the same for all three hooks
+> - The hook call is **typed** with a generic parameter, which should correspond to the type of the data **row**s
+
+### Input props
+
+**Required**:
+- `tableId`: Must be unique throughout the whole app — this is used as part of the data key when the table config is saved to local storage
```
lache-melvin, line 79: This makes me wonder - do you think we should make an enum or something for this? So it's easy to see which table IDs have already been used, and one click to the definition with a big comment saying each table must have unique id??

Issue #9369, opened 2025-09-25
Propagate new tables throughout the codebase


lache-melvin:
Catch all issue to associate our PRs with!

There's 90 or so tables, let's get to work 🔥 

As of 20/10/2025, these tables remain to be migrated:

*Tables that are considered important for 2.12 release are **italicised and bold***

### Replenishment
- **Purchase Orders**
	- [x] List View (R)
	- [x] Detail View (R)
- **Goods Received**
	- [ ] List View
	- [ ] Detail View
	- [ ] Line Edit View
- **Inbound Shipments**
	- [x] _**List View**_ (R)
	- [x] _**Detail View**_ (R)
- **Supplier Returns**
	- [ ] Detail View
	- [ ] Line Edit View
- **RnR Forms**
	- [ ] List View
	- [ ] Detail View (this one probably needs to be handled differently as it's not using the exisiting `DataTable` component)
- **Suppliers**
	- [ ] List View
	- [ ] Purchase Orders List (Tab)
	- [ ] Contacts (Tab)

### Inventory
- **Stock**
	- [x] _**List View**_ - James
	- [x] Ledger (tab)

### Distribution
- **Requisitions** (in progress with #9498)
	- [x] _**List View**_ (R)
	- [x] _**Detail View**_
- **Customer Returns**
	- [ ] List View
	- [ ] Detail View
	- [ ] Line Edit View

### Dispensary
- **Patients** (R)
	- [x] _**List View**_
	- [x] _**Programs (Tab)**_
	- [x] _**Encounters (Tab)**_
	- [x] _**Vaccinations list (Tab)**_
	- [ ] Insurances (Tab)
- **Prescriptions** (R)
	- [x] _**List View**_
	- [x] _**Detail View**_
	- [x] _**Quantity (line edit) table (inner component)**_
- **Encounters**
	- [x] _**List View**_

### Cold chain
- **Monitoring**
	- [ ] Breaches List (Tab)
- **Sensors**
	- [ ] List View

### Programs
- **Immunizations**
	- [ ] List View
	- [ ] Detail View
	- [ ] Line Edit (Doses table)

### Catalogue
- **Assets**
	- [ ] List View
- **Items**
	- [ ] List View
	- [ ] Variants (check for multiple tables in this page)
- **Master Lists**
	- [ ] List View
	- [ ] Detail View

### Manage
- **Demographics**
	- [ ] Main table

> [!NOTE]
[...]

PR #9387, opened 2025-09-26
9369 stocktake table

lache-melvin:
Part 1 of many for #9369

# 👩🏻‍💻 What does this PR do?

Implements new table for Stocktake detail view

# 🧪 Testing

- [ ] Go to stocktake detail view - new table is there :)
- [ ] Filter/sorting of enabled columns behaves as expected
- [ ] Added back group by item, seeing as we are no longer paginated - validate grouped rows display the right info
- [ ] Create a stocktake, add a stock line
- [ ] Go make a prescription or something the reduce that stockline
- [ ] Then try finalise your stocktake
- [ ] You'll get an error, snapshot doesn't match current. Snapshot packs column should be highlighted in red for the culprit rows.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/material-react-table/utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    expect(result).toHaveLength(1);
+    expect(result[0]!.status).toBe('active'); // Should preserve equal value
+    expect(result[0]!.category).toBe('A'); // Grouping field should be preserved
+  });
+
+  it('should handle object values correctly', () => {
+    const data = [
+      { category: 'A', details: null },
+      { category: 'A', details: { type: 'small', color: 'red' } },
```
lache-melvin, line 67: had a bug with the grouping when a column is an object type - we were only using the first item in the array to determine if the field is object type... when `null` wasn't an object, the field wasn't treated as object.

So added this test to prove out the change, and other tests as a base for the grouping behaviour, if we need to iterate on it in future!

client/packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -1,133 +1,102 @@
...
-import {
-  StocktakeFragment,
-  StocktakeLineFragment,
-  useStocktakeOld,
-} from '../api';
+import { StocktakeLineFragment, useStocktakeOld } from '../api';
 import { StocktakeLineErrorProvider } from '../context';
-import { isStocktakeDisabled } from '../../utils';
-
-const StocktakeTabs = ({
```
lache-melvin, line 34: this was broken into 3 components, found it easier just to have one now that everything is a bit simpler, and so the file reads a bit more in the order of the component tree

client/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -89,26 +90,34 @@ export const DetailView = () => {
...
+    initialSort: { key: 'itemName', dir: 'asc' },
+    onRowClick: !isDisabled ? onRowClick : undefined,
+    getIsPlaceholderRow: row =>
+      !!(
+        isPlaceholderRow(row) ||
+        // Also mark parent rows as placeholder if any subRows are placeholders
+        row.subRows?.some(isPlaceholderRow)
       ),
-    });
+    renderEmptyRowsFallback: () => (
```
CarlosNZ, line 113: Should this be `noDataElement` as per other usage?

lache-melvin: ah might have been missed in merge!

client/packages/inventory/src/Stocktake/DetailView/columns.tsx
```diff
@@ -133,9 +133,7 @@ export const useStocktakeColumns = () => {
         id: 'difference',
         header: t('label.difference'),
         columnType: ColumnType.Number,
-        Cell: ({ cell, row }) => (
-          <UnitsAndDosesCell cell={cell} item={row.original.item} />
-        ),
+        Cell: UnitsAndDosesCell,
```
CarlosNZ, line 136: Much nicer 🙏

client/packages/invoices/src/OutboundShipment/DetailView/columns.ts
```diff
@@ -69,6 +73,8 @@ export const useOutboundColumns = () => {
         id: 'itemUnit',
         accessorKey: 'item.unitName',
         header: t('label.unit-name'),
+        enableColumnFilter: true,
+        filterVariant: 'select',
```
CarlosNZ, line 77: I thought I already did this last week 🤔

CarlosNZ (approved):
Very nice, thanks 🙏

client/packages/invoices/src/OutboundShipment/DetailView/columns.ts
lache-melvin: I think this diff actually is from merging your changes... 🤷

PR #9393, opened 2025-09-26
9389 custom bg color

### LINKED ISSUE #9389: Custom formatting for Excel Report exports

Some reports need custom colour coding that isn't determined directly by the cell in excel.

For example in on report we want to show the colour of the cell as `Pink` when the current stock is higher than the forecast needed for the year.

I think it would be nice to be able to set a custom format on a cell as part of the HTML template

It might be possible to do the custom formatting with cell formatting rules and hidden columns describing the colour in Excel, but not sure how wide spread support would be for this in other tools like libre office....

### Moneyworks Jobcode 🧰

OMS:NGUCF

lache-melvin:
Fixes #9389

# 👩🏻‍💻 What does this PR do?

Set BG color in report:
Screenshot 2025-09-26 at 2 09 40 PM

Would probably recommend, doing the logic for which colour to set, in the convert_data method. Then can just pass the hex code direct into both style prop for HTML, and excel-bg-color attribute.

Considered just pulling the `style` attribute off the cell, looping through it and seeing if there was any we knew how to parse... but that parsing is a bit more involved. Could be worth re-considering in future? Or happy to give it more of a play now if it would feel valuable?

# 🧪 Testing

- [ ] Create a report html template, add `excel-bg-color="some-hex-code"` to a td element
- [ ] Run show-report with the `--format excel` flag
- [ ] See cells have colour

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/report/convert_to_excel.rs
+++ b/server/service/src/report/convert_to_excel.rs
@@ -130,9 +130,7 @@ fn apply_header(sheet: &mut Worksheet, header_html: &str) -> u32 /* rows used */
 
         sheet_cell.set_value(inner_text(el));
 
-        if let Some(cell_type) = el.attr("excel-type") {
-            apply_known_styles(sheet_cell, cell_type);
-        };
+        apply_known_styles(sheet_cell, el);
 
         let (_, row, _, _) = index_from_coordinate(coordinate);
         let row_index = row.unwrap_or(0);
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Nice works great!

I don't think it's worth worrying about auto discovery of the styles for now.

PR #9394, opened 2025-09-26
9356 purchase order ui label navigation issues

### LINKED ISSUE #9368: Purchase Orders: PO sent date is not display in the PO list view / Side panel
## What went wrong? 😲

PO sent date is not shown in the PO list view/and Side panel before the confirmation even you have saved the PO sent date in to the DB.

## Expected behaviour 🤔

Should display PO sent date without considering the PO state if customer set the PO sent date.

## How to Reproduce 🔨

1. Create a PO and set the PO set date via the Side panel
2. Check the purchase_order table for Sent_datetime colum
3. Check the PO sent date from the PO  List view, or Side panel

## Your environment 🌱

- Open mSupply Version: 2.11.00-develop-09181725

- Platform: Chrome

- Database type: PostgreSQL

GeronimoJohn:
Fixes #9368

# 👩🏻‍💻 What does this PR do?

Have updated the text when you try to search for an item within the CSV import modal. And ensures that it opens the General tab after importing a CSV or adding a new line/item

Have done a few clean up of imports and have remove React fragments that are not needed

# 🧪 Testing

- [ ] Navigate to a purchase order (or create a new one)
- [ ] Click on the dropdown within Add Item and select Import Lines
- [ ] Upload a CSV file (or download the template and fill with values)
- [ ] Within the 1st tab try search for an item
- [ ] Text should say: `Purchase order lines not found`
- [ ] Remove search and click on Ok&Next
- [ ] If you're on a different tab should be redirected to `General` tab (try in a different language as well)
- [ ] -------------------------------------------------
- [ ] Within the same purchase order -> click on Add Item
- [ ] If you're on a different tab, should be redirected to `General` tab (try in a different language as well)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -444,6 +444,7 @@
   "error.provide-reason-stock-adjustment": "A reason must be provided before adjusting stock",
   "error.provide-valid-reason": "Adjustment reason does not match adjustment direction",
   "error.purchase-order-line-already-exists": "Purchase order line already exists",
+  "error.purchase-order-line-not-found": "Purchase order line not found",
   "error.purchase-order-not-finalised": "Purchase order is not finalised",
   "error.purchase-order-not-found": "Purchase order not found",
   "error.reasons-not-provided-program-requisition": "Reasons must be provided when suggested quantity differs from requested quantity.",
--- a/client/packages/purchasing/src/purchase_order/DetailView/AppBarButtons/AddButton.tsx
+++ b/client/packages/purchasing/src/purchase_order/DetailView/AppBarButtons/AddButton.tsx
@@ -1,19 +1,19 @@
 import React, { useEffect, useMemo, useState } from 'react';
-import { SplitButton, SplitButtonOption } from '@common/components';
-import { useTranslation } from '@common/intl';
-import { AddFromMasterListButton } from './AddFromMasterListButton';
-import { useToggle } from '@common/hooks';
-import { PlusCircleIcon } from '@common/icons';
-import { PurchaseOrderFragment } from '../../api';
-import { UserPermission } from '@common/types';
-import { LineImportModal } from '../ImportLines/LineImportModal';
 import {
   NonEmptyArray,
   useCallbackWithPermission,
   useQueryClient,
+  useToggle,
+  PlusCircleIcon,
+  UserPermission,
+  SplitButton,
+  SplitButtonOption,
+  useTranslation,
 } from '@openmsupply-client/common';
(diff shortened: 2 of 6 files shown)


client/packages/common/src/intl/locales/en/common.json
```diff
@@ -444,6 +444,7 @@
   "error.provide-reason-stock-adjustment": "A reason must be provided before adjusting stock",
   "error.provide-valid-reason": "Adjustment reason does not match adjustment direction",
   "error.purchase-order-line-already-exists": "Purchase order line already exists",
+  "error.purchase-order-line-not-found": "Purchase order line not found",
```
aimee-mcneil-melville, line 447: Based on issue comments my understanding is this should be `Purchase Order line not found`
Having said that, even just in this diff display there are two more with `Purchase order` and it probably needs a dedicated issue to overhaul everywhere 👀

aimee-mcneil-melville (approved):
Thanks John this is great, awesome to have those clean ups in too! 

Good to go, just a couple of things noted:

I think the casing is going to need an overhaul everywhere and should probably be a dedicated issue, I also noticed like here in the modal - if this was to be cased as per comment it would read `Import Purchase Order lines` with only one lowercase word

Image

There also was mentioned in the issue having no `Cancel` or `X` on the master lists modal. The current implementation is consistent with elsewhere on the app where master lists and suppliers are selected in that modal type. I don't see `X` being widely used in modals, but many do have a cancel button.  Did you have any further info on if  this needs any changes?

GeronimoJohn:
> Thanks John this is great, awesome to have those clean ups in too!
> 
> Good to go, just a couple of things noted:
> 
> I think the casing is going to need an overhaul everywhere and should probably be a dedicated issue, I also noticed like here in the modal - if this was to be cased as per comment it would read `Import Purchase Order lines` with only one lowercase word
> 
> Image
> There also was mentioned in the issue having no `Cancel` or `X` on the master lists modal. The current implementation is consistent with elsewhere on the app where master lists and suppliers are selected in that modal type. I don't see `X` being widely used in modals, but many do have a cancel button. Did you have any further info on if this needs any changes?

Thanks @aimee-mcneil-melville! And yep for the casing - was thinking the same thing too, create a dedicated issue for it. Just have to make that consistent throughout the app

In terms of the modal with the close button, forgot to leave my thoughts in the issue. IMO things have to be consistent and if the other master list modals don't have it then we don't as well. That was probably an expected behaviour set from beginning but will leave that as a question

PR #9395, opened 2025-09-26
9381 not possible to enter decimals in pricing fields in stock line

### LINKED ISSUE #9381: Not possible to enter decimals in pricing fields in stock line
## What went wrong? 😲

In a stock line details view, it is not possible to type in a decimal value. In the screenshot below, in the 'cost price' field, the $4 dollars was typed and I couldn't enter 4.5 (for example). In the 'sell price' field, the 6.5 value was successfully added via a copy and paste but I couldn't manually type 6.5. It would be worth it checking other modals if the behaviour is similar. 🙂

Image

## Expected behaviour 🤔

We would expect to be able to manually add decimals in pricing fields.

## How to Reproduce 🔨

1. Go to 'View Stock'
2. Click on a stock line
3. Go to 'Details' tab
4. See error: try to manually enter a decimal value in cost price or sell price field.

## Your environment 🌱

- Open mSupply Version: 2.10.1
- Legacy mSupply Central Server Version: V8-04-09 (Nigeria MEBS Training Server --> temporary used only for training) --> issue noticed on desktop macOS and browser (didn't try others)

- Platform:

- Database type:

roxy-dao:
Fixes #9381

# 👩🏻‍💻 What does this PR do?
The currency component was changed to a controlled component to get the sell price to update... which was messing with the decimals... have changed this back to an uncontrolled component and forced the currency input to re-render if the sell price has changed

# 🧪 Testing

- [ ] Go to stock 
- [ ] Add new stock -> try to enter prices in decimals

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -47,8 +47,10 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   const isSymbolLast = options.pattern.endsWith('!');
   const prefix = !isSymbolLast ? options.symbol : '';
   const suffix = isSymbolLast ? options.symbol : '';
-  const val = value !== undefined ? value : defaultValue;
-  const valueAsNumber = Number.isNaN(Number(val)) ? 0 : Number(val);
+  const defaultValueAsNumber = Number.isNaN(Number(defaultValue))
+    ? undefined
+    : Number(defaultValue);
+  const valueAsNumber = Number.isNaN(value) ? 0 : Number(value);
 
   return (
     <StyledCurrencyInput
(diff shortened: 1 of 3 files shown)


jmbrunskill (approved):
Works good for me, Thanks for the explanation was very helpful.
Basically it looks like we have what we had before?
I presume there was a reason for trying to change it the first time?
I'm ok with it how it is but wondering how we avoid this kind of issue again? I wonder if we should add a warning about issues with making this component controlled again?

roxy-dao:
> Works good for me, Thanks for the explanation was very helpful. Basically it looks like we have what we had before? I presume there was a reason for trying to change it the first time? I'm ok with it how it is but wondering how we avoid this kind of issue again? I wonder if we should add a warning about issues with making this component controlled again?

Yes, the uncontrolled component wasn't re-rendering, which it is doing now. Special use-case here I think with price changing in the same screen if pack size changes. In other places they're on separate screens, so it does re-render

PR #9423, opened 2025-09-29
9172 Jenkins builds intermittently don't include frontend

### LINKED ISSUE #9172: Jenkins build fails silently when client build fails
## What went wrong? 😲

#9007 raised a few issues, so I'm leaving that one closed and opening a new one, just for this problem. 

Jenkins `yarn build` to build the frontend has been occasionally failing, but then continuing to build the backend. We get successful installers, but after install, you only see this screen:

Image

#9067 went in to try fix this (Jenkins build should exit due to there being an error, but it would seem it didn't work, it happened again. Here's a log from Jenkins, you can see client build failing and then continuing:

Image

## Expected behaviour 🤔

1. Client build doesn't intermittently fail (this is the second time in about 2 weeks)
2. If it does fail for whatever reason, the whole installer build fails, and ideally alerts, so we know something has gone wrong!

## How to Reproduce 🔨

1. View logs for `v2.11.00-develop-09050416` build on Jenkins - install this installer to see the `cannot find index.html` error.

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #9172

# 👩🏻‍💻 What does this PR do?
Using blocks instead of a single-line if statement... could be possible that the single-line statement is doing something creative before its fully terminated 👀 . Can't say, I've tried building on my windows with fake failures and both single/block statements work for me 🤷🏻 

Are we needing alerts...? I saw that #9067 attempted telegram alerts which got removed?

# 🧪 Testing

- [ ] Could recreate with windows...
- [ ] cd into the folder and run build.bat 👀 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/build/windows/omsupply-android.bat
+++ b/build/windows/omsupply-android.bat
@@ -2,4 +2,7 @@
 @REM copy the keystore and local.properties for apk signing
 copy c:\android\local.properties client\packages\android\ /y && copy c:\android\release.keystore client\packages\android\app\ /y
 cd "client" && yarn android:build:release && copy packages\android\app\build\outputs\apk\release\*.apk ..\omSupply
-@exit /b %errorlevel%
+@if %errorlevel% neq 0 (
+    @ECHO ERROR: Failed to build android apk
+    exit /b %errorlevel%
+)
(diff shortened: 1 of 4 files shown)


GeronimoJohn (approved):
Lets give this one a go! 🙏🏽

PR #9427, opened 2025-09-29
9419 Fix Date of Birth Translations

### LINKED ISSUE #9419: Patient DOB can be over written due to OMS Central & Legacy Central interaction & formatting.
## What went wrong? 😲

The date of birth format between OMS and Legacy are different. This causes problems when translating the records.
We also shouldn't be updating `name` records from OMS Central unless they are from the OMS Central server itself.

## Expected behaviour 🤔

Fix translator (error in translator rather then defaulting to null, don't send "T" with time, and allow "T" with time to be parsed)

Fix sync direction (don't forward name and name store join records to central and don't forward records to ROMS that arrived to COMS from CMS.

Fix HSH_code compatibility (and other nullable fields that were added during that time)

## How to Reproduce 🔨

- With OMS central setup (one store as dispensary).
- And OMS remote setup (as dispensary).
- Add patients, edit patients, have OMS central off, and then turn it on, at some point DOB will be missing (more exact steps when can replicate)
- 
## Your environment 🌱

- Open mSupply Version: 2.9.4 (remote site)

- Platform:

- Database type:

jmbrunskill:
Fixes #9419 (Part 1)

# 👩🏻‍💻 What does this PR do?

Changes `zero_date_as_option` deserializer to handle datetimes like `2021-01-01T00:00:00` and `2021-01-01 00:00:00` as well as `2021-01-01`

This appears to stop the DOB overwriting in my test setup when applied to Central server even with remote store running `v2.9.4`

Note: there is a change of behaviour here. `zero_date_as_option` now returns an error.
Previously if there was a parsing issue we'd just return `None` this might have been a conscious decision at the time, or might have been a mistake?

I think it's safer to error than return null, as that will potentially loose data.
If we have some truely unimportant fields we would have two versions of this serializer, one that return None on error and one that actually errors?

# 🧪 Testing

- [ ] Setup a 2.9.4 client device
- [ ] Update oms central to this version
- [ ] Update a Patient Date of Birth
- [ ] Sync
- [ ] Wait a few minutes
- [ ] Sync Again
- [ ] Check date of birth hasn't been overwritten...
- [ ] Check that we haven't broken something else...

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/util/src/sync_serde.rs
+++ b/server/util/src/sync_serde.rs
@@ -333,7 +333,7 @@ mod test {
         assert_eq!(d.unwrap().option_t, None);
     }
 
-    #[derive(Deserialize, Serialize, Debug, PartialEq)]
+    #[derive(Deserialize, Serialize, PartialEq)]
     pub struct LegacyRowWithOptionDate {
         #[serde(rename = "ID")]
         pub id: String,
@@ -357,7 +357,7 @@ mod test {
         assert!(a.is_ok());
         assert_eq!(
             a.unwrap().date_of_birth,
-            Some(NaiveDate::from_ymd(2022, 1, 1))
+            Some(NaiveDate::from_ymd_opt(2022, 1, 1).unwrap())
         );
 
         // case with 00-00-0000 (null date)


server/service/src/sync/translations/name.rs
```diff
@@ -127,27 +136,42 @@ pub struct LegacyNameRow {
...
     #[serde(default)]
+    #[serde(rename = "om_date_of_death")]
     #[serde(deserialize_with = "zero_date_as_option")]
     #[serde(serialize_with = "date_option_to_isostring")]
     pub date_of_death: Option<NaiveDate>,
+
+    #[serde(default)]
     pub custom_data: Option<serde_json::Value>,
+
+    #[serde(default)]
```
jmbrunskill, line 157: Added default to all these fields incase they're missing...

Chris-Petty: If it's optional doesn't it get deserialised as `None` if it's missing anyway? How is default different?

server/service/src/sync/translations/name.rs
```diff
@@ -443,7 +467,13 @@ mod tests {
...
             let translation_result = translator
                 .try_translate_from_upsert_sync_record(&connection, &record.sync_buffer_row)
-                .unwrap();
+                .expect(
+                    format!(
+                        "Error translating from upsert sync record {:?}",
+                        record.sync_buffer_row.record_id
+                    )
+                    .as_str(),
+                );
```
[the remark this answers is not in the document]
jmbrunskill: Looks like an external crate https://crates.io/crates/expect_with

server/util/src/sync_serde.rs
```diff
@@ -32,8 +32,22 @@ pub fn ok_or_none<'de, T: Deserialize<'de>, D: Deserializer<'de>>(
...
+        // also treat 0000-00-00 as None
+        return Ok(None);
+    }
+
+    // Try parsing into date even if it has a time component
+    let response = NaiveDate::parse_from_str(&s, "%Y-%m-%d")
+        .or_else(|_| NaiveDate::parse_from_str(&s, "%Y-%m-%dT%H:%M:%S"))
+        .or_else(|_| NaiveDate::parse_from_str(&s, "%Y-%m-%d %H:%M:%S"))
+        .map(Some)
+        .map_err(Error::custom)?;
```
Chris-Petty, line 48: Noice

Huh I've never done `.map(Some)`, of course it's a function!

jmbrunskill: I admit this was a copilot suggestion and I just went with it :)

server/service/src/sync/translations/name.rs
jmbrunskill: At least the HSH_CODE one breaks. Maybe it's if we're using a custom translator? I'll leave it how it is for now.
```
thread 'sync::translations::name::tests::test_name_translation' panicked at service/src/sync/translations/name.rs:469:18:
Error translating from upsert sync record "ef232e6c-93cf-4d02-a95a-c563abccc24f": missing field `HSH_code` at line 33 column 1
```

PR #9461, opened 2025-10-01
9369 location list table

lache-melvin:
Part of #9369

# 👩🏻‍💻 What does this PR do?

Replaces location table:
Screenshot 2025-10-01 at 4 15 56 PM

# 🧪 Testing

- [ ] Sort still works
- [ ] Name and on hold filters are now within the table

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Location/ListView/ListView.tsx
+++ b/client/packages/system/src/Location/ListView/ListView.tsx
@@ -6,7 +6,6 @@ import {
   useUrlQueryParams,
   UNDEFINED_STRING_VALUE,
   InlineProgress,
-  Box,
   usePaginatedMaterialTable,
   ColumnDef,
   ColumnType,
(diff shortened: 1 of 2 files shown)


client/packages/system/src/Location/ListView/ListView.tsx
CarlosNZ, line 87: Do we need this Box wrapper? Can't we just set a width on the column? (Particularly since the InlineProgress component already has block-level containers -- if we really need it, I'd be inclined to pass the width as a prop and apply it in the component's already existing "Box" wrapper)

CarlosNZ (commented):
Great, just a question about the Box around the InlineProgress component.

CarlosNZ (commented):
Sweet, just one comment/suggestion.

PR #9463, opened 2025-10-01
#9374 Fix future encounter creation

### LINKED ISSUE #9374: Creating a future Encounter leads to endless spinning
## What went wrong? 😲

> [!NOTE]
> THIS IS IN V2.12 DEVELOP

I have Djibouti schemas installed for my datafile

And creating an Encounter leads to endless spinning

https://github.com/user-attachments/assets/4b3c30f1-dde0-4b8a-9a23-e38ff9118cbc

Please note - refreshing the page or navigating to another tab and back shows that the encounter was in fact created, it just is not showing straight away and instead loads endlessly

This is what I am seeing in GraphQL: 

Image

## Expected behaviour 🤔

Should be showing the created encounter straight away, no spinning

## How to Reproduce 🔨

1. Upgrade to the latest installer
2. You will need to have the EIR setup ready in your store
3. Easiest way would be to follow the instructions in #4642
4. Make sure you get the latest schemas from the repository
5. Once the setup is done, go to Dispensary -> Patients and click on a patient or create a new one
6. Enrol them into the program -> create an encounter
7. This will lead to spinning
8. Refresh the page or navigate to another tab and back
9. The encounter now added to the list view

## Your environment 🌱

- Open mSupply Version: `V2.12.00-develop-09241424`

- Database type: `PostgreSQL`

CarlosNZ:
Fixes #9374

# 👩🏻‍💻 What does this PR do?

Problem was when re-directing back to the Encounters list view, the query hook was getting an invalid `sortBy` value, as it hadn't been populated by the "initialSort" on the first render.

This fixes the `urlQueryParams` hook so that the sort by will fall back to the "initialSort" value if it can't get it from the URL (which will be populated on next render).

# 🧪 Testing

- [ ] Go to a Patient and select their "Encounters" tab
- [ ] Select "Add Encounter" from the button at the top
- [ ] Select a date in the future for the encounter, then submit
- [ ] You should be redirected back to the encounter list, with no errors (Encounters in the past or present redirect to the actual encounter, which is why the problem didn't show up)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


lache-melvin (changes requested):
Hm, this breaks sorting on our new tables, which need the "sorting off" option - e.g. on outbound shipment list, can't change sort direction 👀 

I wonder if actually we should update the .graphql file, to handle no sort being passed in? Should be fine as far as the API is concerned...

Or update the redirect, to add the sort param, same as the tabs do?

Issue #3969, opened 2024-05-21
Can we have some shortcut like 'Reset/clean cache' for server exe?


nisha-dangol:
## What went wrong? 😲
In desktop exe, `Help>Clear Data` is supposed to clean all cache data. But I do not find it quite working since I still see some `open mSupply` folder remaining under AppData even after `Help>Clear Data`. 

https://github.com/msupply-foundation/open-msupply/assets/59193490/6ecc1fda-c1c2-4eaf-abb1-1f6cef713d9f

Also, it'd to good to have this feature for server exe as well (for testers, it'd be very helpful). Since server exe may not have UI like `Help>Clear Data`, I suggest create some shortcut that cleans the cache data.

## Expected behaviour 🤔
`Reset/clean cache` shortcut (same like we have for mSupply) should delete any `open mSupply` folder in below locations:
- `C:\Users\<username>\AppData\Roaming\`
- `C:\Users\<username>\AppData\Local\`
- `C:\Users\<username>\AppData\Local\Temp\`
- the temp folder:  `<%temp%>\open mSupply\`

## How to Reproduce 🔨
Steps to reproduce the behaviour:
Sometimes (especially when multiple datafiles has been loaded in the same PC), we normally see the desktop app not responding or not connecting. On manually cleaning cache (from above mentioned locations), it works. So good to have a feature to reset/clean cache? 

## Your environment 🌱

- Open mSupply Version: v2.0.0 exe
- Legacy mSupply Central Server Version: v7-14-06

- Platform: server exe, desktop-exe

- Database type: sqlite

andreievg:
@nisha-dangol, what would be the expected behaviour for clearing server.exe data ?

nisha-dangol:
> @nisha-dangol, what would be the expected behaviour for clearing server.exe data ?

I expect; on installation of server.exe there will be a shortcut sth like `Reset/clear cache data` installed in the desktop as well, which when run will prompt with some warning/question sth like:
<img width="982" alt="image" src="https://github.com/msupply-foundation/open-msupply/assets/59193490/81cb7826-cdae-4feb-bf3d-163aae7ea5f8">
On selecting Y/Yes, what actually happens:
- Delete folder: `C:\Users\<username>\AppData\Roaming\open mSupply`
- Delete folder: `C:\Users\<username>\AppData\Local\open mSupply`
- Delete folder: `C:\Users\<username>\AppData\Local\Temp\open mSupply`
[...]

PR #9467, opened 2025-10-01
3969 name & clinician tables

lache-melvin:
Part of #3969

# 👩🏻‍💻 What does this PR do?

Replaces supplier, customer and clinician list view tables

# 🧪 Testing

- [ ] Can still sort by name & code for supplier/customer
- [ ] Can still sort by name/code/initials for clinicians
- [ ] Pagination still working

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Clinician/ListView/ListView.tsx
+++ b/client/packages/system/src/Clinician/ListView/ListView.tsx
@@ -1,47 +1,36 @@
 import React from 'react';
 import {
-  TableProvider,
-  DataTable,
-  createTableStore,
   NothingHere,
   useTranslation,
   useUrlQueryParams,
+  usePaginatedMaterialTable,
+  MaterialTable,
 } from '@openmsupply-client/common';
 import { useClinicianListColumns } from './columns';
-import { ClinicianFragment, useClinicians } from '@openmsupply-client/programs';
+import { useClinicians } from '@openmsupply-client/programs';
 
-const ClinicianListComponent = () => {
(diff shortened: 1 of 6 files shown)


CarlosNZ (approved):
LGTM!

One small thing though -- we probably want a new issue for it if it's not a quick fix: These tables have no filters enabled, so it would be good if we could hide the "Filter" button on the tool bar in this case. It's slightly clunky UI to be able to click it and see a small little movement in the layout, but nothing actually changing.

lache-melvin:
Nice thanks! Have added hiding when enable columns set to false 🙏

client/packages/system/src/Clinician/ListView/ListView.tsx
```diff
@@ -29,6 +29,7 @@ export const ClinicianListView = () => {
     isError,
     isLoading: isFetching,
     enableRowSelection: false,
+    enableColumnFilters: false,
```
CarlosNZ, line 32: Oh, I was thinking of some internal logic that checks the column definitions and turns this off if there are no filters enabled, so we don't have to explicitly add this to our table definition. Is that tricky to do?

lache-melvin: Ohh, maybe not, I'll have a quick look in my current PR

Issue #7678, opened 2025-05-09
Significant problems with JSON Forms custom error handler


CarlosNZ:
## What went wrong? 😲

This is a result of an investigation I did after [this follow-up comment](https://github.com/msupply-foundation/open-msupply/issues/7569#issuecomment-2864382509) from a recent fix.

The current bug (that really only became obvious after that above fix) is that it is possible to "Save" a JSON Form with some of the elements in an "invalid" state -- for example the phone number pattern in the above example. 

The reason why is that the `useJSONFormsCustomError` hook is fundamentally broken -- what it's *trying* to do is insert custom errors (i.e. errors generated by a custom validation function inside one of our renderers) into JSONForms global "errors" array. It does this by calling a JSONForms internal `dispatch` method to update the error array.

On the face of it, this seems like it should work (otherwise why else would there be a method to dispatch errors to the state?), but in fact it is *not* the correct way to go about it. Because what happens is that JSONForms own validation process (which can use custom Ajv validation if implemented) performs a validation pass and resets the Error array on each render. So what ends up happening is that the custom errors hook updates the error array, but it immediately gets reset on the next render. From what I could tell, there's some code in that hook that looks like it *might* be trying to deal with this problem by *re-* injecting the current error if it persists, but this actually ended up causing an infinite render loop on custom errors, which is the thing I fixed in the above issue.

The basic problem is that JSONForms (in conjuction with Ajv) provides a built-in validation mechanism, which works well and is the intended pattern for error handling in JSONForms -- any attempt to circumvent this is an anti-pattern and causes all kinds of issues.

## How to fix?

This will require some further discussion, but ther are two main things that can be done:

1. Use JSON Schemas properly -- this will fix the telephone validation issue above, more on that below
2. Create a separate context for *Custom* errors independent of JSONForms use `additionalErrors` prop to handle them in JSONForms

In more detail...

### 1. Use JSON Schemas properly

[...]

PR #9470, opened 2025-10-02
#9372 JsonForm error handler fix

### LINKED ISSUE #9372: JSON forms: able to save the fields even when they don't match the requirements
## What went wrong? 😲

> [!NOTE]
> THIS IS IN V2.12 DEVELOP

I have Djibouti schemas installed for my datafile

And seems like I am able to save information in fields with requirements even when entered information does not match those requirements

https://github.com/user-attachments/assets/dfa214ee-2769-4a9a-a91e-9e3927bc4afe

## Expected behaviour 🤔

If there is a requirement - e.g. code should be a 6-digit string of numbers, then I should not be able to save anything that does not match that

## How to Reproduce 🔨

1. You will need to have the EIR setup ready in your store
2. Easiest way would be to follow the instructions in #4642
3. Make sure you get the latest schemas from the repository
4. Once the setup is done, go to Dispensary -> Patients and click on a patient or create a new one
5. Fields like Code and Phone Number have a specific requirement to them - e.g. a string of numbers
6. Enter anything that does not match that
7. Click Save -> you are able to save

## Your environment 🌱

- Open mSupply Version: `2.12.00-develop-09241424`

- Database type: `PostgreSQL`

CarlosNZ:
Fixes #9372

# 👩🏻‍💻 What does this PR do?

Another one of those "ends up being a tiny fix, but takes a lot of digging to pin it down" issues! (Which we seemed to get a lot of in JSONForms, due to the fragile house of useEffect cards we've managed to builde).

Something was causing the `JsonForm` component to re-render and lose it's "error" values that it was passing to its `onChange` handler. Eventually tracked it down to this `handleDefaultsAjv` object -- if I removed it the problem went away. Looks like the original mistake was defining this inside the component, which makes a new instance on every render, when it's just a static object that should be defined once.

So now I've moved that definition *outside* the component, and that seems to have fixed the problem. Hopefully it doesn't affect anything else, I've really only tested it in the Patient form where the original bug was noted.

I'm aware that we still have [this issue](https://github.com/msupply-foundation/open-msupply/issues/7678) outstanding, which is related, but not exactly the same.  Would be good to address that sometime soon too.

# 🧪 Testing

- [ ] You'll need a Patient Json schema that has some Regex validation in it (not sure if the default one does)
- [ ] Go to a specific patient
- [ ] Enter some invalid data in a field that has strict pattern validation (e.g. telephone no, or ID number)
- [ ] Confirm that the "Save" button remains disabled (and you see the error UI around the input)
- [ ] Correct the data
- [ ] Confirm you can now Save
- [ ] Also check that if you make some valid changes (which makes "Save" enabled) and *then* enter invalid data that the Save button becomes disabled again

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


CarlosNZ:
> It works on the first invalid change... but not the second 😢

Okay @lache-melvin, can confirm that the second problem is a result of [this bug](https://github.com/msupply-foundation/open-msupply/issues/7678) -- if I move the `pattern` field for the ID and Telephone fields to the schema rather than the UI schema (which is where it should be), it works as expected. 

For now I'll update the Djibouti patient schema, will have to see when we have time to fix the main issue.

PR #9482, opened 2025-10-03
9369 `manage` list view tables

lache-melvin:
Part of #9369

# 👩🏻‍💻 What does this PR do?

replaces stores, campaigns, reports, and equipment list views.

Adds custom `filterKey` handling, for cases where display key/column accessor/sort key (e.g. `categoryName`) doesn't match the filter key (`categoryId`)

# 🧪 Testing

- [ ] Sort/filter/pagination on each page (especially check filtering on equipment columns)
- [ ] Select to delete works for campaigns and equipments

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/material-react-table/useGetColumnDefDefaults.tsx
```diff
@@ -89,6 +91,17 @@ export const useGetColumnTypeDefaults = () => {
...
+      case ColumnType.Boolean:
+        return {
+          align: 'center',
+          Cell: ({ cell }) => {
+            const value = cell.getValue<boolean | null>();
+            return value ? (
+              <CircleIcon sx={{ transform: 'scale(0.5)' }} />
+            ) : null;
+          },
+        };
```
CarlosNZ, line 103: You think this UI should be the default for all Booleans? Is that how it is currently?

CarlosNZ (approved):
Looking good, just a couple of quick questions, but ready to merge when ready 👍

lache-melvin:
> Is this necessary? Can't we just change the id?

Not really - in this case, backend expects sort key = `categoryName`, and filter input of `categoryId`. Now that we're populating from the same column, we do need a way for these two to be mapped separately...

client/packages/common/src/ui/layout/tables/material-react-table/useGetColumnDefDefaults.tsx
lache-melvin: Yeah we have used `DotCell` for most booleans 👀

PR #9503, opened 2025-10-07
#7678 JSONForms Additional Errors

### LINKED ISSUE #7678: Significant problems with JSON Forms custom error handler

CarlosNZ:
Fixes #7678

# 👩🏻‍💻 What does this PR do?

Removed the dodgy `useJSONFormsCustomError` and replaced it with a new, simpler one `useAdditionalErrors`, which more closely follows the approach reccommended by JSON Forms: https://jsonforms.io/docs/validation#external-validation-errors

Have also replaced the usage in the renderer components that are using it, except for "Text" where I've removed it completely as it doesn't need to do any pattern validation at all within the component if the pattern is correctly defined in the schema rather than the uischema options. 

The way it works is the new hook `useAdditionalErrors` just holds the `additionalErrors` array state, and provides a couple of methods to add/update or remove errors from this array. These methods and state array are acquired by the main `JsonForm` component, where the methods are added to the global `config` object so they can be called by any component. The `additionalErrors` are passed to the main component via the `additionalErrors` property -- a bonus of this is that they are merged together with the "internal" validation errors, so components only need to access the main "errors" prop rather than also checking for separate "customErrors" now.

See inline comments.

# 🧪 Testing

I suggest using the latest Djibouti programs schemas, which have had the "pattern" field correctly migrated from the ui-schema to the schema.

- [ ] Confirm input errors show up correctly for these components, and that the form can't be submitted when in and error state:
  - [ ] Text (with Regex pattern validation) e.g. Patient ID and Telephone fields
  - [ ] BloodPressure (in HIV Care Encounter -> Physical Exam), gives error when entering only one of the two values
  - [ ] IDGenerator (in Program Enrolment form) 
  - [ ] Date/DateTime and DateOfBirth components -- shows error when entering (via keyboard) invalid dates
- [ ] If you can be bothered, try it with a schema that still has the "pattern" field in the uischema -- you should see a "Bad input configuration" error

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/programs/src/JsonForms/common/components/CategorizationTabLayout.tsx
```diff
@@ -140,7 +140,10 @@ const containsErrors = (layout: Layout, errorPaths: string[]): Set<string> => {
         ''
       );
       for (const errorPath of errorPaths) {
-        if (scopePath == errorPath) {
+        // Match on the parent as well, as some components have internal "child"
+        // elements that set the error rather than the component itself
+        const errorPathParent = errorPath.split('/').slice(0, -1).join('/');
+        if (scopePath == errorPath || scopePath === errorPathParent) {
```
CarlosNZ, line 146: I simplified the BloodPressure component by removing a messy `useEffect` update, but the consequence is that *only* the child elements (two data leaves below the component level) show in error state, so this change just loosens the error matching here in order to display the "number of errors" badge on the Category tile.

Image

client/packages/programs/src/JsonForms/common/components/Date.tsx
```diff
@@ -59,15 +59,13 @@ const UIComponent = (props: ControlProps) => {
               path,
               !e ? undefined : formatDateTime.customDate(e, 'yyyy-MM-dd')
             );
-            if (customError) setCustomError(undefined);
+            customErrors.remove(path);
           }}
           format="P"
           disabled={!props.enabled}
-          error={customError ?? props.errors ?? zErrors ?? ''}
+          error={props.errors ?? zErrors ?? ''}
```
CarlosNZ, line 66: As mentioned above, the `additionalErrors` are now included in the `props.errors` field

client/packages/programs/src/JsonForms/common/hooks/useAdditionalErrors.ts
```diff
@@ -0,0 +1,81 @@
...
+          instancePath: path,
+          message: message,
+          schemaPath: '',
+          keyword: 'custom',
+          params: {},
+        },
+      ]);
+
+      // This updates the form's overall "error" state, if it's defined
+      if (setError) setError(message);
```
CarlosNZ, line 52: This is what allows the parent component to keep the "Save" button disabled even for the custom/additional errors. Previously, this didn't do this, as it wouldn't actually save the invalid values, but I think it's better still to not even allow the "Save" button to be clicked while there are any type of errors present.

client/packages/programs/src/JsonForms/components/BloodPressure.tsx
```diff
@@ -28,18 +27,7 @@ export const UIComponent = (props: ControlProps) => {
...
-    'BloodPressure'
-  );
-
-  useEffect(() => {
-    if (core) {
-      const getChildErrors = subErrorsAt(path, schema);
-      const errors = getChildErrors(core);
-      setCustomError(errors[0]?.message);
-    }
-  }, [core]);
```
CarlosNZ, line 42: This is the nasty `useEffect` I mentioned above. It wouldn't play nicely at all with the custom errors and the child errors both changing, so it's easier to remove it completely. This means that *only* the child data nodes are in "error" state, but I think that's fine as it's technically more correct. Just means we need that slight relaxation of the "matching" in the Category tile components.

client/packages/programs/src/JsonForms/common/hooks/useAdditionalErrors.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    [additionalErrors]
```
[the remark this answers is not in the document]
CarlosNZ: 0e82d66254

client/packages/programs/src/JsonForms/common/components/Select.tsx
```diff
@@ -258,36 +261,36 @@ const useFilteredItems = (
       !filtered.includes(currentSelection)
     ) {
       setVisibleItems([currentSelection, ...filtered]);
-      setError('Please select a valid option');
+      customErrors.add(path, 'Please select a valid option');
```
lache-melvin, line 264: should this be translated?

client/packages/programs/src/JsonForms/common/components/Text.tsx
```diff
@@ -10,18 +10,14 @@ import { z } from 'zod';
...
-import { useJSONFormsCustomError } from '../hooks/useJSONFormsCustomError';
 import { usePrevious } from '../hooks/usePrevious';
 import { PreviousValueDisplay } from '../utilityComponents';
 
 const Options = z
   .object({
     /**
-     * Additional pattern to be matched that can be defined in ui schema
-     */
-    pattern: z.string().optional(),
```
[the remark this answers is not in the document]
lache-melvin: Seems like a good call - maybe add to [known issues before deployment](https://docs.google.com/spreadsheets/d/1VmviJcym4owNtLXq2uQRG4k4PHhN3_BJA004m8l3iao/edit?gid=0#gid=0) just to flag that schemas should be reviewed/updated before upgrading ?

client/packages/programs/src/JsonForms/components/DateOfBirth.tsx
```diff
@@ -94,7 +91,7 @@ const UIComponent = (props: ControlProps) => {
     setDoB(dob);
     handleChange(dobPath, formatDateTime.customDate(dob, 'yyyy-MM-dd'));
     handleChange(estimatedPath, true);
-    setCustomError(undefined);
+    customErrors.remove(path);
```
lache-melvin, line 94: If I change another field, the date value resets to previous valid value, save button enables, but it's still showing error: 

https://github.com/user-attachments/assets/743db670-c89a-4f58-a762-6d91954a7297

lache-melvin (changes requested):
> latest Djibouti programs schemas, which have had the "pattern" field correctly migrated

Maybe there's no errors configured at all on some of these schemas, but I was expecting this to show error? It doesn't save..

Screenshot 2025-10-08 at 11 42 00 AM

PR #9508, opened 2025-10-07
7575 item scrolling

### LINKED ISSUE #7575: Items might not be in item selector after adding many items to an invoice
We have a lot of uses of StockItemSearchInput that have an extra filter like this... it has the possibility to bite us? (Filter excludes all items that already exist in the invoice.)

E.g. I have 150 items visible, I include 130 in an order... and then, of the first 100, only 7 aren't in the order, so I show those in the item selector - that's not enough to trigger the scroll/pagination, can't access the other 23...

Haven't seen this raised as a bug, seems like an edge case, but possibly we should have an "excludeItems" array on the API filter, rather than frontend filter? Can raise an issue?

_Originally posted by @lache-melvin in https://github.com/msupply-foundation/open-msupply/pull/7567#discussion_r2070928316_

See PR for more context

aimee-mcneil-melville:
Fixes #7575

# 👩🏻‍💻 What does this PR do?

When many items are on an order and therefore filtered out from the item selector, all remaining items are accessible in the paginated results.

Added a 'notEqualAll' filter type to filter these items out in the API call (sent from front end), so the database results don't include the items we don't want visible.

I also tried two other approaches:

If results > pagination limit -> call next page in the api
- simple to implement in the autocomplete component & universal
- resulted in multiple back-to-back calls if needing to get further in the list ie only a few items left

Backend filtering by invoice/requisition/stocktake id -> check lines for item id and exclude those
- different cases to implement, id vs invoiceId in params etc. More fragile
- marginally faster if an order has large data eg 500+ items (compared to implemented approach)
- not flexible

This approach - sending items to exclude from front end through api
- need to call data in each component to get the existing items and provide that to the filter. Very similar to existing implementation that did the same, but only prevented the items from rendering in the UI
- marginally faster on orders with less items (compared to backend approach)
- flexible per component implementation

# 🧪 Testing

- [ ] Easiest to test in areas where you can use a masterlist to add all items to an order eg requisition, stocktake, inbound shipment
- [ ] Note: pagination limit is 20 on requisitions, 100 elsewhere. Will need to test with more than 100 items
- [ ] Create a new order -> check you can scroll the item list, pagination works as expected when reaching the end of the scroll
- [ ] Add all items to an order -> should have no results if attempting to add another item
- [ ] Delete a few lines, ensure that some are from towards the end and past the pagination limit above
- [ ] Go to item selector -> see the items from the just deleted lines
- [ ] Delete more items to have 2-3 x the pagination limit not on the order
- [ ] Go to item selector -> see pagination working correctly when scrolling
- [ ] Check searching by item name or code works correctly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditForm.tsx
```diff
@@ -33,6 +33,8 @@ export const StocktakeLineEditForm: FC<StocktakeLineEditProps> = ({
   const t = useTranslation();
   const disabled = mode === ModalMode.Update;
 
+  const existingItemIds = items.map(item => item.item?.id ?? '');
```
aimee-mcneil-melville, line 36: Declaring this as a variable I find keeps the filter easier to read - happy to update them to inline though if that's preferred

client/packages/system/src/Item/api/api.ts
```diff
@@ -55,6 +55,7 @@ export const getItemQueries = (sdk: Sdk, storeId: string) => ({
           ...filterBy,
           type: { equalTo: ItemNodeType.Stock },
           isActive: true,
+          isVisible: true,
```
aimee-mcneil-melville, line 58: Added this here since it was common across all filters

client/packages/system/src/Item/Components/StockItemSearchInputWithStats/StockItemSearchInputWithStats.tsx
```diff
@@ -42,7 +44,10 @@ export const StockItemSearchInputWithStats = ({
   const { data, isLoading, fetchNextPage, isFetchingNextPage } =
     useStockItemsWithStats({
       rowsPerPage: ROWS_PER_PAGE,
-      filter: { ...filter, isVisible: true },
```
aimee-mcneil-melville, line 45: isVisibleOrOnHand: true is in the api, so isVisible not needed here

PR #9551, opened 2025-10-13
#8969(1) Config and UI for "skip dose" option

CarlosNZ:
Part 1 of #8969 

# 👩🏻‍💻 What does this PR do?

This just adds the migration to add the extra field "can_skip_dose" to the vaccine_course table, and implements the UI and database operations to configure this value when inserting or updating a Vaccine course.

Screenshot 2025-10-13 at 2 02 11 PM

Next up -- use this option when adminstering vaccines

## 💌 Any notes for the reviewer?

Clarified requirement with @mark-prins and @adamdewey -- have written them up in the "Agreed rules" section of the original issue.

# 🧪 Testing

- [ ] Go to Programs -> Immunizations
- [ ] Add a new Vaccine course
- [ ] Note the option for "Can skip dose"
- [ ] Ensure that whatever setting you selected for this option is persisted after creating the course
- [ ] Open a course for editing, toggle the option, ensure it saves correctly.

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #9558, opened 2025-10-13
8969(2) Allow skipping of vaccine doses

### LINKED ISSUE #8969: Allow administering later doses when first dose not given
## What went wrong? 😲

The system restricts users from administering a dose if the previous dose in the vaccine course has not been administered. 

Image

However, in the real-world they do this. 

For example, if a patient doesn't receive a bOPV vaccination at birth (bOPV0) they are considered to have missed it and they skip to the next dose when they next visit (bOPV1)

Here is an example on paper

Image

## Expected behaviour 🤔

System should only warn user about giving doses out of order, not restrict the user. 

### Agreed rules

(Adding this here for clarity around expectations)

- Will be a per-vaccine-course preference
- Will receive UI warning dialog when administering a dose that is not the next in the sequence
- Will automatically apply "Not given" to all skipped doses
- Will only be able to edit the most recently applied (given or not given) dose (same as current behaviour)

## How to Reproduce 🔨

1. Mark the 1st dose of a vaccine course as not given
2. Attempt to mark the next dose as given

## Your environment 🌱

- Open mSupply Version: 2.9.2

Money Works Code: SSUNI2OMD

CarlosNZ:
Fixes #8969 -- follows on from PR #9551 

# 👩🏻‍💻 What does this PR do?

A few moving parts here...

### Back-end
- Only give `InsertVaccinationError::VaccineIsNotNextDose` error if the vaccine course does not allow skipping doses (as configured/defined in previous PR)
- When a dose is skipped, create vaccinations for all the skipped ones, with a "Not given" status

### Front-end
- Allow later vaccine rows to be clicked (opened) if "can skip dose) is enabled
- Add additional warning when skipping dose
- Created a new hook `useConfirmModalSequence` which provides a mechanism for multiple confirmation steps for an action. The "save vaccination" action already had (potentially) 2 modal warnings, and we're adding a third, and the logic for this is very messy and inter-dependent. This new hook provides a cleaner mechanism, and allows us to remove 2 additional hooks for handling specific cases.

See inline comments

# 🧪 Testing

Have a vaccination course where "skip dose" is NOT allowed
Create a vaccination encounter and go to the vaccination card view
- [ ] Can only click on the next dose or earlier doses
- [ ] Dose administering behaves the same as before

Now create a vaccination course (or edit the same one) where "skip dose" IS allowed:
- [ ] Now you can click ALL vaccination rows
- [ ] Skipping a dose (i.e. marking as "given") with empty earlier doses:
  - [ ] You get a warning (modal) that you are about to skip doses
  - [ ] Creates "Not given" vaccinations for all the earlier (missing) ones

Confirm the other warnings still work when appropriate:
- [ ] Giving a vaccination with no stock selected
- [ ] Giving a vaccination outside the suggested date range

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/vaccination/insert/mod.rs
+++ b/server/service/src/vaccination/insert/mod.rs
@@ -62,7 +62,7 @@ pub struct InsertVaccination {
     pub not_given_reason: Option<String>,
     /// Internal flag to unnecessary recursion when creating missing "Not Given"
     /// vaccinations
-    pub skip_dose_backfill: bool,
+    pub create_not_given_records_for_skipped_doses: bool,
 }
 
 pub fn insert_vaccination(
@@ -91,7 +91,7 @@ pub fn insert_vaccination(
 
             // If can_skip_dose is enabled and we're not already doing backfill,
             // create "Not given" vaccinations for any missing previous doses
-            if !input.skip_dose_backfill {
+            if !input.create_not_given_records_for_skipped_doses {
                 if let Ok(vaccine_course_dose) =
                     check_vaccine_course_dose_exists(&input.vaccine_course_dose_id, connection)
                 {
@@ -208,7 +208,7 @@ fn create_missing_not_given_vaccinations(
                 item_id: None,
                 stock_line_id: None,
                 not_given_reason: Some("Dose skipped (Later dose administered)".to_string()),
-                skip_dose_backfill: true, // Prevent recursive backfill creation
+                create_not_given_records_for_skipped_doses: true, // Prevent recursive backfill creation
 
                 // Copy all other properties from the original input
                 encounter_id: original_input.encounter_id.clone(),
--- a/server/graphql/programs/src/mutations/vaccination/insert.rs
+++ b/server/graphql/programs/src/mutations/vaccination/insert.rs
@@ -57,7 +57,7 @@ impl From<InsertVaccinationInput> for InsertVaccination {
             item_id,
             stock_line_id,
             not_given_reason,
-            skip_dose_backfill: false, // Default to false for API inputs
+            create_not_given_records_for_skipped_doses: false, // Default to false for API inputs
         }
     }
 }
--- a/server/service/src/vaccination/insert/generate.rs
+++ b/server/service/src/vaccination/insert/generate.rs
@@ -40,7 +40,7 @@ pub fn generate(
         stock_line_id,
         not_given_reason,
         item_id,
-        skip_dose_backfill: _, // Ignore this field as it's only used for control flow
+        create_not_given_records_for_skipped_doses: _, // Ignore this field as it's only used for control flow
     } = insert_input;
 
     let now = Utc::now().naive_utc();


client/packages/system/src/Vaccination/Components/VaccinationModal.tsx
```diff
@@ -100,18 +105,43 @@ export const VaccinationModal = ({
...
+        title: t('heading.are-you-sure'),
+        message: t('messages.no-batch-selected'),
+        condition: () =>
+          getShouldShowConfirmation(
+            draft,
+            !!dose?.vaccineCourse.vaccineCourseItems?.length,
+            store?.nameId ?? ''
+          ),
+      },
+    ],
```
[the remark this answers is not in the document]
jmbrunskill: Much nicer!

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -123,7 +117,9 @@ const VaccinationCardComponent = ({
 
   const isEncounter = !!encounterId;
 
-  useStyleRowsByStatus(data?.items, isEncounter);
+  const canSkipDose = data?.items.some(item => item.canSkipDose) ?? false;
```
jmbrunskill: I think this logic is too generic. If i have a vaccine course with 2 vaccines 1 skippable and one not skippable, then they're both show as clickable in the ui but you get an error on the non-skipable one...

Image

Image

jmbrunskill: I think we should at least translate this error properly but probalby better to tweak the logic to avoid it in th UI...

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
```diff
@@ -169,9 +165,7 @@ const VaccinationCardComponent = ({
       {
         key: 'status',
         label: 'label.status',
-        accessor: ({ rowData }) =>
-          // Only show label for existing vaccinations and the next editable row
-          includeRow(true, rowData, data?.items) ? rowData.status : null,
+        accessor: ({ rowData }) => rowData.status,
```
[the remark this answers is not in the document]
jmbrunskill: Yeah I think this is a low priority one to update.

server/service/src/vaccination/insert/mod.rs
```diff
@@ -79,6 +89,27 @@ pub fn insert_vaccination(
             // Create the vaccination
             VaccinationRowRepository::new(connection).upsert_one(&vaccination)?;
 
+            // If can_skip_dose is enabled and we're not already doing backfill,
+            // create "Not given" vaccinations for any missing previous doses
+            if !input.skip_dose_backfill {
+                if let Ok(vaccine_course_dose) =
+                    check_vaccine_course_dose_exists(&input.vaccine_course_dose_id, connection)
```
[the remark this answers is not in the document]
jmbrunskill: That's assuming the error type is the same (or can go `into` it) which I haven't checked...

client/packages/common/src/intl/locales/en/common.json
```diff
@@ -1830,6 +1830,7 @@
   "messages.confirm-delete-vaccine-courses_one": "This will permanently remove this vaccine course",
   "messages.confirm-delete-vaccine-courses_other": "This will permanently remove the {{count}} selected vaccine courses",
   "messages.confirm-early-vaccination": "The minimum recommended date for this vaccine is {{ date }} (based on patient age and the two-dose interval). Are you sure you can administer this vaccine?",
+  "messages.confirm-skip-dose": "You are about to administer a dose even though earlier doses were not given. This will make earlier doses no longer editable. Are you sure you want to continue?",
```
jmbrunskill: For me earlier doses are still editable? Is this a setting or is the message wrong?

jmbrunskill (changes requested):
Looks great nice work @CarlosNZ.

Just one logic bug to fix.

I think we need to handle some vaccines as skippable and some not in the same Immunisation Program better. Let me know if it's going to be a big pain to implement as I don't think it's likely to be a common scenario but It's definitely a valid setup even if it's likely to be a mistake...

jmbrunskill (approved):
Changed to approved pending fix for that issues with multiple vaccines with differnet skip doses statuses. If needed you can merge this and raise an issue to fix the other. Would be good to get this in to tonight's build so we can get the QA started.

client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
CarlosNZ: > I think this logic is too generic. If i have a vaccine course with 2 vaccines 1 skippable and one not skippable, then they're both show as clickable in the ui but you get an error on the non-skipable one...

Got it, yeah, I actually over-complicated it since `canSkipDose` is already on the "row", so no need to pass it in the `canClickRow` function. 

I've fixed this in the next PR though, since I modified some of the other logic around it anyway.

PR #9578, opened 2025-10-15
8969 update vaccination tables

### LINKED ISSUE #8969: Allow administering later doses when first dose not given

CarlosNZ:
Fixes #8969 -- migrates the table to the new table pattern

# 👩🏻‍💻 What does this PR do?

Replicates the VaccinationCard table using the new table pattern:

**PREVIOUS**:
Screenshot 2025-10-15 at 2 56 39 PM

**NEW:**
Screenshot 2025-10-15 at 4 11 47 PM

Fairly straightforward but matching the styling was a bit of a challenge -- I had to extend the table infrastructure to handle proper merging of props -- so we could add row and cell styling without overriding the table default styles.

# 🧪 Testing

- [ ] Vaccination Card table looks and behaves the same as it did before, but with the new table functionality

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/ui/layout/tables/material-react-table/utils.ts
```diff
@@ -20,24 +20,28 @@ type MuiTableBodyCellPropsParams<TData extends Record<string, any>> = {
...
+    | ((params: MuiTableBodyCellPropsParams<TData>) => TableCellProps),
   params: MuiTableBodyCellPropsParams<TData>
 ): TableCellProps => {
   const { table } = params;
   const globalPropsOption = table.options.muiTableBodyCellProps;
 
-  // Resolve global props if it's a function
+  const resolvedCustomProps =
+    typeof customProps === 'function' ? customProps(params) : customProps || {};
+
```
CarlosNZ: Changed to allow customProps to be the functional type as well (which they are in the VaccinationCard table, as the style is conditional on the column)

jmbrunskill (approved):
Looks good

PR #9580, opened 2025-10-15
#9577 Fix for JSON Forms telephone save problem

### LINKED ISSUE #9577: Patient JSON schemas: follow-up
## What went wrong? 😲

This issue is a follow-up from #7569 

### Cannot clear the address field

But only if this is the only field filled out. To recreate: create a new patient - only name, last name, nationality and code should be filled out. Go straight to address -> fill it out -> Save. Attempt to remove - save button is disabled.

https://github.com/user-attachments/assets/4263efef-51c9-41e1-8fd4-db913780461d

### Cannot clear the region field

But only if this is the only field filled out. To recreate: create a new patient - only name, last name, nationality and code should be filled out. Go straight to region -> fill it out -> Save. Attempt to remove - save button is disabled.

https://github.com/user-attachments/assets/dfd0e12f-0564-44c7-a2c7-87d67e6a3796

### Cannot clear the phone field

Only the 'Contact' one, not the 'Next of Kin' one. Once filled out, you cannot remove it from the field

https://github.com/user-attachments/assets/a3b350ad-7acc-4fd7-a6fd-152f1e658504

## Expected behaviour 🤔

Should be able to clear all the fields

## Your environment 🌱

- Open mSupply Version: 2.12.00-develop-10101345 + latest patient JSON schemas
- Legacy mSupply Central Server Version: V8.05.08

- Database type: PostgreSQL

CarlosNZ:
Fixes #9577

# 👩🏻‍💻 What does this PR do?

Makes the "Telephone" field for the Djibouti Patient contact info clearable.

It was a bit bewildering why the Telephone field in for the patient contact had this problem but the Telephone field for the next of kin (on the same) form didn't. Turns out the one for the patient is actually structured in an *array* of contact info data in the schema. But we're only using one, so we use the "FirstItemArray" component. The problem was in that component -- the "onChange" function wasn't being called in the component when it's child's onChange passed up an empty value, so the parent data didn't get updated.

I've replaced the child "JsonForm" component in that component with the "JsonFormsDispatch" component, which is the correct way to do nested Json Forms, and is consistent with how the other Array components do it, which fixes the problem (I could have just modified the `onChange` method, but this is more thorough).

# 🧪 Testing

- [ ] Confirm that the problem reported in the issue has been remedied

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/Text.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/Text.tsx
@@ -58,7 +58,6 @@ const UIComponent = (props: ControlProps) => {
   const error = !!errors || !!zErrors;
   const onChange = (value: string | undefined) =>
     handleChange(path, !!value ? value : undefined);
-
   const { text, onChange: onDebounceChange } = useDebouncedTextInput(
     data,
     onChange


client/packages/programs/src/JsonForms/common/components/Array/FirstItemArray.tsx
```diff
@@ -88,19 +88,13 @@ const FirstItemArrayComponent = (props: FirstItemArrayControlCustomProps) => {
...
         key={childPath}
         schema={schemaWithDefs}
         uischema={childUiSchema || uischema}
-        data={data?.[0] ?? {}}
-        config={config}
-        readonly={!enabled}
-        renderers={renderers ?? []}
-        onChange={({ data }) => {
-          if (Object.values(data).filter(it => !!it).length > 0) {
-            handleChange(childPath, data);
```
CarlosNZ: This was the culprit -- Falsy values were being filtered out so the "handleChange" function didn't run

jmbrunskill (approved):
Seems all good to me, I haven't tested but I'll leave @marthakeezy  to do that!

PR #9583, opened 2025-10-16
Properties kdd

andreievg:
One way to deal with custom item catalogue (pricing) for SAGE integration -> Internal Order is through item properties, which aligns with OMS central server direction for having extendable fields for most entries.

This KDD attempts to suggest a design for these properties.

@craigdrown @mark-prins to verify requirements.

Similar to name properties, for the very first iteration we can just upsert item_property for the price via a script. The property value however needs to be updatable via API so that SAGE integration can update it, and graphql should be available to query for item property when plugin is implemented.

Btw the 'shape' of item property when consumed is a JSON containing all the properties (similar to name_property), but proposed storage is quite different.


decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,79 @@
...
+1. User can configure a new property:
+  a. type can be given one of Number (int), Text, Date, Real (Float), Option (user defined categories/options) 
+  b. name can be given 
+  c. translation can be given (would default to translation if one is found for current langauge, otherwise would use general name) 
+  d. options can be hierarchical, i.e current item categories in mSupply or organisation units in DHIS2 (to be confirmed)
+  e. tables which will have this property are specified, property can be re-used across many tables. (to be confirmed)
+
+2. User is able to set properties based on configurations, properties should all be clearable. 
+
+3. User can change property configurations, however migrations are too difficult to implement and maintain, it's well communicated that updates to configurations (like removing options or or changing option type) will not update existing value property values
```
roxy-dao:
```suggestion
3. User can change property configurations, however migrations are too difficult to implement and maintain, it's well communicated that updates to configurations (like removing options or changing option type) will not update existing value property values
```

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,79 @@
...
+
+Also adding property configuration functionality to omSupply is slightly easier with "Option 1", however syncing properties attached to remote records is slightly harder (although properties for remote records would only be syncable with V7)
+
+### Option 3
+
+Have table specific property and/or property_value tables. The idea is to build for future performance in mind and reduce dynamic structured (usually safer and more readable/disoverable). However a pattern like would be very common and well understood, negating any effects this abstraction should have on readability/discoverability. As for performance, there is an assumption to be validated that table_name index together with record_id index will only speed up insert time (for index building), not query (query is the the performance concern, not insert), this goes in favour ot "Option 1".
+
+### Option 4
+
+Store value as JSON rather then different typed fields. This would be easier in general but if we need to do any calculations, it would be slower. Also if configuration is changed for the property it would require higher congnitive load and likely translations (to make sure existing values work)
```
roxy-dao:
```suggestion
Store value as JSON rather than different typed fields. This would be easier in general but if we need to do any calculations, it would be slower. Also if configuration is changed for the property it would require higher cognitive load and likely translations (to make sure existing values work)
```

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,79 @@
...
+item categories
+asset categories/class
+name properties
+any other ?
+
+## Further consideration
+
+Reports - We can be quite generic with properties for most things, having generic filter/display in tables etc.. However in reports it would be hard to be generic with properties, especially at the time of display. Similar for report filters. 
+
+Multiple values - Do we need an array type element when multiple values can be associated for the same property ? This sounds more like tags, which is also something generic that should exist in OMS central
```
roxy-dao: In what use cases were you thinking an array would be needed? 🤔 I think should be easy enough to add if it ever pops up?

mark-prins (commented):
I favour option 1 - keeping the current property table and extending its usage to other linked tables.

Other thoughts:
Would be good to have provision for default values

A possible future requirement is validation beyond typing i.e. having a regex validation option, doesn't impact this KDD though, that can come later for any of the options.

a consideration which you've touched on is the item categories which we've implemented and not properly shown in the UI yet. Should we consider migration of OG categories as part of the property implementation? it would be helpful to bring transaction categories (etc) across and map values into properties

Chris-Petty:
Also an Option 1 backer, seems like the cleanest data model to get head around as opposed to JSON hell haha.

The comment above `property_table` table suggests it'd be less code and alleviates index concerns, but it seems like more code and at least 3 indexes, as opposed than just having `table_name` indexed field on `property` table?

Functionally it'd allow you to have 1 property used for many tables satisfying requirement 1e, though that's yet to be confirmed. Seems a little wild trying to imagine where it'd be useful 🤔 you just want a "priority" property that's an integer and it's easier to add it to several tables, rather than configuring it a few times?  Ah as suggested in requirements, something like DHIS2_ID might be handy to add around the place.

An idea to add that could be left as an extension another day - some kind of render option field on table `property`?  For a simple example, an integer or float value you may want it presented as a percentage or perhaps currency.

Side note - it's gonna be real hard for support if properties are loaded up with no trace of English, especially if not even a latin alphabet. Nick might have to learn French, maybe Arabic too 😁. That is to say I don't imagine we're restricting what goes in the property.name field at all, maybe the translation is redundant? Or will TMF support strictly configure these? I don't have any good alternative to that, just something to keep in mind!

Chris-Petty:
Requirements 2 and 4 might be better to hash out what the UI will be for properties. Like would any line modal have a special tab just for property entry? Is there anywhere that'd not work as a blanket rule for input?

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,79 @@
...
+### Requirements
+
+1. User can configure a new property:
+  a. type can be given one of Number (int), Text, Date, Real (Float), Option (user defined categories/options) 
+  b. name can be given 
+  c. translation can be given (would default to translation if one is found for current langauge, otherwise would use general name) 
+  d. options can be hierarchical, i.e current item categories in mSupply or organisation units in DHIS2 (to be confirmed)
+  e. tables which will have this property are specified, property can be re-used across many tables. (to be confirmed)
+
+2. User is able to set properties based on configurations, properties should all be clearable.
```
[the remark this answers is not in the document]
andreievg: f3f2a8f82c72302185afa7eee98fb2f04cc4bfb5

decisions/2025-10-16_properties.md
andreievg: > In what use cases were you thinking an array would be needed? 🤔 I think should be easy enough to add if it ever pops up?

It's hard to query, especially in SQLite, thus would prefer a tag for those (like name tags), although could be property with lots of property_values right ? And then don't really need tags, hmm but tags are so well known by user, vs setting up more involved 'property', ¯\_(ツ)_/¯

andreievg:
@Chris-Petty @mark-prins great points, I think they are linked, wanted to capture them, added to KDD see commit below

> A possible future requirement is validation beyond typing i.e. having a regex validation option, doesn't impact this KDD though, that can come later for any of the options.

> An idea to add that could be left as an extension another day - some kind of render option field on table property? For a simple example, an integer or float value you may want it presented as a percentage or perhaps currency.

6ed4cbfac2b207364c29256ecdfb597375ddd5de

andreievg:
> Requirements 2 and 4 might be better to hash out what the UI will be for properties. Like would any line modal have a special tab just for property entry? Is there anywhere that'd not work as a blanket rule for input?

Yeah agree, i was only really thinking of new tables, where it seems super standard

andreievg:
@mark-prins 

> a consideration which you've touched on is the item categories which we've implemented and not properly shown in the UI yet. Should we consider migration of OG categories as part of the property implementation? it would be helpful to bring transaction categories (etc) across and map values into properties

I think that would be great, and custom fields too, I've add a TODO for that: 3821c1d88d9c2bb847a535980e53bff87d81be77, I think most migrations would be pretty straight forward, and I would love for us to have just one 'properties' type for existing properties/categories/asset class etc.. There is no design/plan for migration of those, but it all seems possible, I was hoping to have this refined enough to see the horizon (end goal), while taking our next steps (just the minimum things needed for property of catalogue price on item)

Chris-Petty:
> @Chris-Petty
> 
> > Side note - it's gonna be real hard for support if properties are loaded up with no trace of English, especially if not even a latin alphabet. Nick might have to learn French, maybe Arabic too 😁. That is to say I don't imagine we're restricting what goes in the property.name field at all, maybe the translation is redundant? Or will TMF support strictly configure these? I don't have any good alternative to that, just something to keep in mind!
> 
> I've added translation_key for that reason in (1. translation user requirement), would that be sufficient ? (i.e. core 'name' is probably in the 'base' language, and then there is translation_key for whatever language we want, if we have a standard set of categories we add through properties we can have standard translation keys for them

My concern is if it's being configured by a 3rd party other than us they might very well not bother setting translation keys. Again I don't really think we can help it that much in that scenario.

jmbrunskill (commented):
I'm on the fence about which option i like,

Having the json on the record as per name (Option 2 I think) has a lot of benefits in terms of `sync`  there's not extra complex code to manage figuring out how and where to send out a property `value`. If it was all central data it would be simpler but I think we need to consider things like invoice and invoice line properties being defined.

It also would make things like backend sorting would I think be easier as trying to dynamically join to multiple properties tables sounds really complex.

(With option 1: Frontend sorting will be easy though and with the new tables maybe there's not a need to do backend sorting?)

I'd like to see a prototype done with Option 1 so we can test out things like.
filtering and sorting, performance on large data files (records can be created using a script)

I'd also like to do a little more refining of our requirements as I mentioned some things like defining if data is remotely editable for example which is a current requirement in our existing system.

I'd like to confirm as some examples, if we really need to have translatable options (wouldn't they be defined once for the country?), Do we really need hierarchical options, if so what would that look like in practice, with queries etc?) Do we need to be able to rename options and have it apply everywhere? (Currently in gaps we don't have that functionality)

Which specific things we currently have are we looking to replace? Does plugin data still exist for example? 

What kinds of user configurations can we expect to define? What customisations are we aiming to be able to avoid?

Chris-Petty:
> I'm on the fence about which option i like,
> 
> Having the json on the record as per name (Option 2 I think) has a lot of benefits in terms of `sync` there's not extra complex code to manage figuring out how and where to send out a property `value`. If it was all central data it would be simpler but I think we need to consider things like invoice and invoice line properties being defined.
> 

Agree that it's pretty easy for sync! Though provided it's not abused - if someones putting images or something else heinously large it may hang sync up. I think our options might be limited resolving that when using a record JSON field other than mitigating with some restrictions?

> It also would make things like backend sorting would I think be easier as trying to dynamically join to multiple properties tables sounds really complex.
> 

We should probably add indexed sorting as a key functional requirement? Or at least performant sorting as an operational requirement...

Backend sorting through the gql API is essential (more on that below) and with a JSON string field it's very expensive - you'd need to query for all records, deserialise the field, then sort them. Where as with properties and joins you could feasibly index them. Postgres has jsonb that can be indexed, but the vast majority of installations are using sqlite... huh turns out sqlite 3.45 has jsonb? What's more [diesel 2.3.1 has support for both pg an sqlite jsonb fields](https://docs.diesel.rs/master/diesel/sql_types/struct.Jsonb.html) 😮. Just learnt 2 new exciting things.

That said, it might be a bit of work to start usage of jsonb in our codebase, one of the few references to it in our code is a comment from yourself saying it's hard to use with diesel. Could you elaborate your thoughts on that (albeit your wrote that comment 17 months ago 😁)?

What's probably more cursed regardless of storage option is having a dynamic gql API based on customised properties? Including sorting and filtering? I think we need every option to make it's case for how it might address this? Also an overarching concern a frontend that has dynamic gql queries that consumes this dynamic API! (I'd argue we should have this already to an extent but this comment is already too big 😉.)

> (With option 1: Frontend sorting will be easy though and with the new tables maybe there's not a need to do backend sorting?)
> 

I've heard this mentioned a couple times regarding the new tables. I think that with new tables any fallback to frontend sorting if there isn't a backend API for sorting is a flawed implementation that should be removed. If the data set has more than your current pagination (or the max page size constraint of the backend) then there is no magic, the sorting is just broken. Most tables in OMS will return more records than pagination limits. Do you know if there is an existing issue regarding this?

> I'd like to see a prototype done with Option 1 so we can test out things like. filtering and sorting, performance on large data files (records can be created using a script)
> 

💯 

> I'd also like to do a little more refining of our requirements as I mentioned some things like defining if data is remotely editable for example which is a current requirement in our existing system.
> 

I think this is implicit but should be declared explicitly, I agree. If it's on store data tables like invoices and requisitions then it should be remote editable values. Configurations should be central though or it'll be chaos.

> I'd like to confirm as some examples, if we really need to have translatable options (wouldn't they be defined once for the country?), Do we really need hierarchical options, if so what would that look like in practice, with queries etc?) Do we need to be able to rename options and have it apply everywhere? (Currently in gaps we don't have that functionality)
> 
> Which specific things we currently have are we looking to replace? Does plugin data still exist for example?
> 
> What kinds of user configurations can we expect to define? What customisations are we aiming to be able to avoid?

Examples of the problems we're trying to solve with this decision would be good 💯.

Currently we just have `One way to deal with custom item catalogue (pricing) for SAGE integration -> Internal Order is through item properties` but that is fairly heavily leaning on prerequisite knowledge of what that means 😉

craigdrown:
I had a chat with James just before I left, and not sure where to write further comments, so here they are (and sorry if wrong)
a. Use cases
- the primary use case is "reporting" - both formal reports (e.g. someone wants to see the item usage report, but only for antibiotics, which we've added a property for), but also allowing users to find a set of items to look at ("We made a basket of goods we want to monitor every day- we labelled them all "Priority" : "high-2025") - and we make an SOP that someone checks stocks for them daily.
- storing data that we retrieve via API (e.g. Universal Code properties on an item). These should be read-only for the user
- a place to store data that a plugin can use for custom behaviours. e.g. "Add an extra 10% margin to all items with property "dangerous-goods-category":"1") - so the plugin doesn't have to handle a UI for setting this stuff.

b. I think the current model is too complex - I would give up having the same property apply to different tables, as supporting this adds complexity
c. I don't think the current name property functionality will cut it once we get millions of names (and we will). 
aside: I see Diesel now [supports JSONb](https://github.com/diesel-rs/diesel/releases/tag/v2.3.0) which would reduce most of the overhead, but still... 
e. I had mainly thought of properties for central data, which simplifies the syncing. Not averse to supporting remote data too, but if having properties that are edited on multiple sites breaks our sync rules, I would choose sync rule consistency by making separate properties tables for remote data (Every time we break a rule we add developer overhead in having to remember where it's broken?)

andreievg:
Another thing, a requirement to have hierarchy for 'district', like organisation unit in DHIS2

GaryWilletts:
@andreievg there's nothing in the model to make properties mandatory. This is something that will be necessary e.g. Chemonics Haiti already have mandatory properties in OG (we had to customise that).

I think we'll also need something to choose where they appear in the UI. At the moment they all appear on a separate tab on the transaction in OG but there are also 4 original ones that appear on the main window of a transaction (below the comments field). Users cope with the fact that the custom fields are all on a separate tab because they have no choice, but they often really want some of them on the main window.

adamdewey:
Should this kdd be renamed 'Item properties kdd' ?

Or perhaps the requirement has morphed from the initial post to cover more than just item properties?

jmbrunskill:
> Should this kdd be renamed 'Item properties kdd' ?
> 
> Or perhaps the requirement has morphed from the initial post to cover more than just item properties?

The second one. We want a way to add properties to different kinds of records not just items. Although you're right that does expand the complexity of the task

andreievg:
A small document with potential [use cases for properties](https://docs.google.com/document/d/15_YfqKhRUwt4oF1NV0BBn85btLKEfTPGF4BQ8l_bpq8/edit?tab=t.0), kudos @DhanyaHerath

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,81 @@
...
+### Option 5
+
+Instead of options, extend and use category table, see KDD, although it wasn't implemented yet to support full and extendable system, there is foundation for it (or change category to be properties) -> Combining properties with categories would be worth it if we consolidate all 'custom/dynamic' configurations, we have too many now:
+
+item categories
+asset categories/class
+name properties
+any other ?
+
+## Further consideration
```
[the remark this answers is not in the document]
jmbrunskill: I think this can come as an update later

decisions/TEMPLATE.md
```diff
@@ -0,0 +1,74 @@
+<!--
+KDD (Key Decision Document) template.
```
jmbrunskill, line 2: snuck this is as it seems we didn't have a template and that was creating some inconsistency (e.g there was no decision section to fill out.

jmbrunskill (approved):
I'm conformfortable where this has got too. Could probably clean up a little bit of AI cruft but would be good to get this in.

@andreievg @ria8651 happy with it now?

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,177 @@
...
+
+Considerations for performance:
+
+- Can apply broad indexes across the whole `property_value` table
+- Harder to create ideal indexes - more indirection through joins
+
+Considerations for sync:
+
+- More complex then JSON, need to track and lookup attached records to know where to sync to
+- Handles Requirement 8 well if we choose to tackle that
```
ria8651, line 52: Should say requirement 7

decisions/2025-10-16_properties.md
```diff
@@ -0,0 +1,177 @@
...
+
+- Whenever parsing json in DB (for sorting/filtering) use binary JSON columns - there's no downside
+- Slower than Relational but not by a crazy amount - even with 60 properties per row
+- With a targeted index for a particular property achieves better performance than relational
+
+Considerations for sync:
+
+- Slots into existing model cleanly (already have examples of synced json categories/properties in codebase), just follows along with attached row
+- More data sent per update - no row partial updates in sync
+- Will require filtering/merging on sync if we want to tackle Requirement 8
```
ria8651, line 71: Same here, should be requirement 7

ria8651 (approved):
LGTM besides wrong requirement reference. Could do with cleanup but oh well.

PR #9595, opened 2025-10-21
#9593 Fix table keyboard navigation

### LINKED ISSUE #9593: Tabbing in tables go to read-only fields and have to tab twice on inputs
## What went wrong? 😲

Tabbing in tables go to read-only fields and have to tab twice on inputs...

https://github.com/user-attachments/assets/4e020998-57d8-4edf-9c3b-ee862f7ea0a3

## Expected behaviour 🤔
When I tab it should go to the next input? And I shouldn't have to tab twice to get input working?

## How to Reproduce 🔨

1. Go to somewhere with input in tables, like Outbound Shipment
2. Enter something in first input then tab
3. See that it goes to the read-only field
4. Tab until you get to next input
5. Try to enter text
6. Can't. Tab again, input is now enabled

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #9593

# 👩🏻‍💻 What does this PR do?

Turns out this was actually a quick fix. By disabling MRT's internal keyboard navigation, we restore our normal tab navigation, where focus will go to the next or previous *input*, not every cell.

Any downsides to this?

# 🧪 Testing

- [ ] Go to a table that has some input cells (e.g. Line Edit table in Outbound Shipment)
- [ ] Confirm that Tab/Shift-tab navigates between input cells, not all cells
- [ ] Check for any possible regressions or undesirable behaviours as a result of this change

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
+++ b/client/packages/common/src/ui/layout/tables/material-react-table/useBaseMaterialTable.tsx
@@ -152,7 +152,9 @@ export const useBaseMaterialTable = <T extends MRT_RowData>({
     enableStickyHeader: true,
     // We want tab navigation to follow our normal behaviour of moving to the
     // next INPUT, not move through every table cell. If we need specific Table
-    // keyboard navigation in future, we can enable this in a more granular way.
+    // keyboard navigation in future, we can enable this in a more granular way
+    // using our own custom shortcuts:
+    // https://www.material-react-table.com/docs/guides/accessibility#custom-keyboard-shortcuts
     enableKeyboardShortcuts: false,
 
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Downside is that you can't use the arrow keys to go up and down and side to side in the cells. Feels like it would be kind of useful in some cases, looks like we can add custom shortcuts though. https://www.material-react-table.com/docs/guides/accessibility#custom-keyboard-shortcuts

I think this is a lot better than before!

PR #9601, opened 2025-10-23
9496 requisition auto finalise

### LINKED ISSUE #9496: Requisitions - Automatically finalised when fully supplied
Assume close customer requisition when all requested quantities have been shipped on outbound shipments. Needs a store pref I think

Money Works Code: SSUNI2OMD

Chris-Petty:
Fixes #9496 

# 👩🏻‍💻 What does this PR do?

Adds a processor that is triggered after sync and finalising a customer invoice that will check if the new "Requisition auto finalise" pref is on for the store. If so it checks if all requisition lines' supply quantities are fulfilled by the related shipped outbound shipments.

# 🧪 Testing

- [ ] Central OMS server/store
- [ ] Go to Manage > Stores > your store > preferences, and turn on "Auto finalise fulfilled requisitions"
- [ ] (A) Create a manual requisition for a customer
- [ ] Add a requisition line and set a supply quantity
- [ ] (B) Create an outbound shipment (from the requisition)
- [ ] Add a shipment line that **equals** the supply quantity
- [ ] "Ship" the outbound shipment
- [ ] The requisition is automatically finalised
- [ ] **Repeat A to B above**
- [ ] Add a shipment line that is **less** than supply quantity
- [ ] "Ship" the outbound shipment
- [ ] The requisition is still "New"
- [ ] Create another shipment and add a shipment line that fulfils the supply amount
- [ ] "Ship" the outbound shipment
- [ ] The requisition is auto finalised
- [ ] **Repeat A to B above**
- [ ] Add a second line to fulfil to the requisition, with 0 to supply
- [ ] Create an outbound shipment that just fulfils the first requisition line
- [ ] "Ship" the outbound shipment 
- [ ] The requisition is automatically finalised
- [ ] **Repeat A to B above**
- [ ] Turn the preference off
- [ ] Add a shipment line that is **more than** the supply quantity
- [ ] "Ship" the outbound shipment
- [ ] The requisition is not automatically finalised

There are about a million permutations of statuses and orderings of events, but the above seem like a good breadth of how this will be used. Feel free to go further.

# 📃 Documentation

  1. Preferences page should explain this preference
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/preference/upsert.rs
+++ b/server/service/src/preference/upsert.rs
@@ -21,6 +21,7 @@ pub struct UpsertPreferences {
     pub prevent_transfers_months_before_initialisation: Option<i32>,
     pub show_contact_tracing: Option<bool>,
     pub sync_records_display_threshold: Option<i32>,
+    pub warning_for_excess_request: Option<bool>,
 
     // Store preferences
     pub manage_vaccines_in_doses: Option<Vec<StorePrefUpdate<bool>>>,
@@ -56,6 +57,7 @@ pub fn upsert_preferences(
         use_simplified_mobile_ui: use_simplified_mobile_ui_input,
         disable_manual_returns: disable_manual_returns_input,
         requisition_auto_finalise: requisition_auto_finalise_input,
+        warning_for_excess_request: warning_for_excess_request_input,
     }: UpsertPreferences,
 ) -> Result<(), UpsertPreferenceError> {
     let PreferenceProvider {
@@ -78,6 +80,7 @@ pub fn upsert_preferences(
         use_simplified_mobile_ui,
         disable_manual_returns,
         requisition_auto_finalise,
+        warning_for_excess_request,
     }: PreferenceProvider = get_preference_provider();
 
     ctx.connection
@@ -113,6 +116,10 @@ pub fn upsert_preferences(
 
             if let Some(input) = sync_records_display_threshold_input {
                 sync_records_display_threshold.upsert(connection, input, None)?;
+            }
+
+            if let Some(input) = warning_for_excess_request_input {
+                warning_for_excess_request.upsert(connection, input, None)?;
             }
 
             // Store preferences, input could be array of store IDs and values - iterate and insert...
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -2186,6 +2186,7 @@
   "preference.syncRecordsDisplayThreshold": "Sync records display threshold",
   "preference.useProcurementFunctionality": "Use procurement functionality",
   "preference.useSimplifiedMobileUi": "Use simplified mobile UI",
+  "preference.warningForExcessRequest": "Warn when requested quantity exceeds suggested quantity in requisitions",
   "prescription": "Prescriptions",
   "prescriptions": "Prescriptions",
   "programs": "Programs",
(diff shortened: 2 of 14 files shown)


server/service/src/processors/requisition_auto_finalise/requisition_auto_finalise.rs
```diff
@@ -0,0 +1,178 @@
...
+                    equal_to: Some(InvoiceLineType::StockOut),
+                    ..Default::default()
+                }),
+        )?;
+        if invoice_lines.len() == 0 {
+            return Ok(None);
+        }
+
+        let mut invoice_line_item_units: HashMap<String, f64> = HashMap::new();
+        invoice_lines
```
Chris-Petty: Probably more efficient for execution to make an SQL view for this, however I think our architecture should be reviewed if every slightly beyond basic data access requires adding a view and diesel repo.

server/service/src/processors/requisition_auto_finalise/requisition_auto_finalise.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let invoice_lines = InvoiceLineRepository::new(connection).query_by_filter(
+            InvoiceLineFilter::new()
+                .invoice_id(EqualFilter::equal_any(
+                    invoices.into_iter().map(|i| i.invoice_row.id).collect(),
+                ))
+                .r#type(EqualFilter {
+                    equal_to: Some(InvoiceLineType::StockOut),
+                    ..Default::default()
+                }),
```
Chris-Petty: Whenever I try to do `EqualFilter::equal_to` the compiler throws a wobbly that the inner type needs to be `String` but I'm giving it `InvoiceLineType`. The generics are getting lost? Anyway rather than fight it I just instantiated the struct instead, which does the generic fine 🤷 .

server/service/src/processors/requisition_auto_finalise/requisition_auto_finalise.rs
```diff
@@ -0,0 +1,178 @@
...
+        requisition.finalised_datetime = Some(chrono::Utc::now().naive_utc());
+
+        requisition_row_repo.upsert_one(&requisition)?;
+
+        system_activity_log_entry(
+            connection,
+            ActivityLogType::RequisitionStatusFinalised,
+            &requisition.store_id,
+            &requisition.id,
+        )?;
```
Chris-Petty: Probably better to use `update_response_requisition`? The validation there might be important and if more statuses are added it'll probably handle setting multiple datetime fields (and this file neglected).

That said, I'd like the log to say something along the lines of "Requisition auto finalised" in its detail. I haven't done that here, and it's even harder with `update_response_requisition`... well at least not without making the function parameters ugly.

server/service/src/preference/upsert.rs
```diff
@@ -185,3 +147,14 @@ pub fn upsert_preferences(
 
     Ok(())
 }
+
+fn upsert_store_input<P: Preference>(
```
jmbrunskill: Nice lil refactor :)

server/service/src/processors/requisition_auto_finalise/requisition_auto_finalise.rs
jmbrunskill: Agreed!

jmbrunskill (approved):
Looks great Chris, nice work!

I haven't tested all the scenarios just some basic ones with the pref turned on. I figure tesing all the combinations that's what the QA team is for haha :)

@alainsussol just a heads up that would be good to work through all Chris' test scenario mentioned in this PR.

I'd like to change to using the `update_response_requisition` function but otherwise I think everything look good to me...

Did you remember a PG enum for the new Activity Log Type?

jmbrunskill (changes requested):
Need to fix the tests

server/service/src/processors/requisition_auto_finalise/requisition_auto_finalise.rs
Chris-Petty: I think going full generic sounds great! I've been kinda confused what's going on in that file for a while. I suspect if we needed specific handling for a particular type the compiler would be smart enough to use that if implemented rather than generic (I presume the compiler generates functions for every type used with the generic anyway, we'd just be manually implementing instead)

PR #9608, opened 2025-10-23
#9534 Stocktake price input fix

### LINKED ISSUE #9534: Text regression while typing in Pack Sell & Cost Price when adding an item to a Stocktake
## What went wrong? 😲

> [!NOTE]
> THIS IS IN V2.12 DEVELOP

Seeing this: 

https://github.com/user-attachments/assets/1554c6fa-718b-4d59-91fb-5ba48d9b2fa5

It seems like when I try and type in a decimal value into the cells, the system backspaces not only the digit after the `.` but also the `.`  and the digit before that, resulting in wrong whole numbers. If the user is typing fast and does not notice the regression, they end up typing in huge numbers, in which case the $NaN.00 value is showing

## Expected behaviour 🤔

No regression - can type in the decimal values effortlessly

## How to Reproduce 🔨

1. Upgrade to the latest installer
2. Go to Inventory -> Stocktakes -> New Stocktake
3. Add any item -> go to tab 2
4. In Pack Sell Price and Pack Cost Price, attempt to type in decimal values a few times
5. Type in fast, type in slow, backspace, type in again - should be able to replicate the issue

## Your environment 🌱

- Open mSupply Version: `V2.12.00-develop-10081537`

- Database type: `PostgreSQL`

CarlosNZ:
Fixes #9534

# 👩🏻‍💻 What does this PR do?

It's pretty hard to be sure this handles every case, but it's definitely a big improvement to just remove the debouncing on the CurrencyInputCell. The debouncing was causing the internal value and the "draft state" value to get out of sync in certain circumstances. I think we need to be careful with this and make sure we do debouncing at the right level -- and we don't need it at all when it's just updating draft state, only when it's actually triggering server requests.

The new table's "CurrencyInputCell" is only used in this one table so far, so there's no risk of causing problems elsewhere. The old "CurrencyInputCell" doesn't include debouncing, so I think it should stay out of the cell logic -- if we need it, it can go in the "updater" function that is provided to it.

# 🧪 Testing

- [ ] Follow the input patterns demonstrated in the original issue video and confirm it's no longer a problem

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


jmbrunskill (changes requested):
I'm not sure if it's actually related the same issue, however, I'm still getting NANs generated. Seems like the combination of a comma plus a decimal e.g. `1,000.00` causes problems.

I started to have a quick look at it but quickly getting into the weeds so hopefully you'll faster to look into it? Feels like we need a separate display value to what's calculated as the actual number perhaps?

https://github.com/user-attachments/assets/7123f2f6-8686-413e-938d-af077a33cff3

PR #9615, opened 2025-10-24
9571 Days out of stock dynamic query

### LINKED ISSUE #9570: Performant DOS and Stock on Date calculation with SQL

We have multiple ways of calculating Days out of Stock in plugins, report and in the core, ideally we would have one performant way to do this.

Also stock on date can be very fast if done using stock movements table.

We can extend diesel to allow parameters to be added to more complex queries, beyond normal diesel dsl, I've done a tutorial for this here: https://github.com/andreievg/diesel-rs-dynamic-queries

This combined with window functions, we can have a very performant Days out of Stock calculation:

<details>

<summary> SQL </summary>

```SQL
WITH starting_stock AS (
  SELECT
    item_id,
    store_id,
    SUM(quantity) AS running_balance,
    '2019-10-03 13:00:29' AS datetime
  FROM
    stock_movement
  WHERE
    datetime <= '2019-10-03 13:00:29'
  GROUP BY
    item_id,
    store_id
),
ending_stock AS (
  SELECT
    item_id,
    store_id,
    SUM(quantity) AS running_balance,
    '2019-11-03 13:00:29' AS datetime
  FROM
    stock_movement
  WHERE
    datetime <= '2019-11-03 13:00:29'
  GROUP BY
    item_id,
    store_id
),
ledger AS (
  SELECT
    *,
    DATE(datetime) AS date
  FROM
    starting_stock
  UNION
  SELECT
    *,
    DATE(datetime) AS date
  FROM
    ending_stock
  UNION
  SELECT
    item_id,
    store_id,
    running_balance,
    datetime,
    DATE(datetime) AS date
  FROM
    item_ledger
  WHERE
    datetime > '2019-10-03 13:00:29'
    AND datetime < '2019-11-03 13:00:29'
),
daily_stock AS (
  SELECT DISTINCT
    item_id,
    store_id,
    date,
    MAX(running_balance) OVER (PARTITION BY store_id,
      item_id,
      date) AS max_stock,
    FIRST_VALUE(running_balance) OVER (PARTITION BY store_id,
      item_id,
      date ORDER BY datetime DESC) AS running_balance
  FROM
    ledger
),
with_lag AS (
  SELECT
    *,
    LAG(running_balance) OVER (PARTITION BY store_id,
      item_id ORDER BY date) AS pr,
    LAG(date) OVER (PARTITION BY store_id,
      item_id ORDER BY date) AS pd
  FROM
    daily_stock
  ORDER BY
    store_id,
    item_id
)
SELECT
  item_id,
  store_id,
  sum(julianday(date) - julianday(pd))
FROM
  with_lag
WHERE
  pr = 0
GROUP BY
  1,
  2
ORDER BY
  store_id,
  item_id
```

</details>

Explanation, we want to get a rows like this

`item_id, store_id, ledger_date, previous_ledger_date, previous_ledger_balance`

notice this is grouped by date

And then we sum up all of the differences between ledger_date and previous_ledger_date where previous_ledger_balance = 0.

Since we are checking for DOS in a particular period, we also need to have an entry for stock on starting date of the period and ending date of the period, thus the union.

So the query really contains these parts.

get starting date stock and ending stock, union it to ledger for the period.

aggregate by date (rather then date time), taking last entry for the date, and max entry for the date (the latter is not used yet, but i know that IC uses that, we can make the end query output both stats, and it should).

after all of that we can create the rows described above and then aggregate the final result. 

this is super fast on my 2 gig sqlite date file.

After creating this generic repository we should expose it to plugins and reports, and use in plugins and report where possible

Also the stock on date repository can be added, which should be trivial (and can even be done without diesel extension, using group by and filtering.

</details>

aimee-mcneil-melville:
Fixes #9570 

# 👩🏻‍💻 What does this PR do?

Get days out of stock calculations with a dynamic query

Use this query in AMC calculations for requisitions

Also added a readme about dynamic queries as an implementation example

**Definition of days out of stock:** the number of days where the item had a balance of 0 stock on hand on the previous day

Example item with 2 days out of stock and adjustment preference off:
Screenshot 2025-10-24 at 4 57 25 PM

And with adjustement preference on:
Screenshot 2025-10-24 at 4 57 10 PM

Unsure if the readme & query (and maybe repository) should be in their own folder? Happy to move if makes sense structurally, but is the only dynamic query implementation for now

# 🧪 Testing

Dev testing:
- [ ] Have some items with out of stock days - balance of 0 in store. It must have some days with stock and some without, NOT including today! 
- [ ] Run the test in days_out_of_stock_query.rs - change dates to suit out of stock days if needed
- [ ] Paste the sql from test output into DB Browser
- [ ] Replace ? placeholders with appropriate values
- [ ] See your items with out of stock days in the result & total_dos should be what you expect

UI testing: 
- [ ] Go to an item in the catalogue with a few outbound transactions in the ledger over last 3 months (outbound shipments/prescriptions with picked status or higher.)
- [ ] The item should have had days out of stock within the last 3 months, NOT including today
- [ ] Take note of the AMC value
- [ ] Go to Global preferences -> turn `Adjust for days out of stock` ON
- [ ] Go back to the item -> see the AMC increase by the expected amount. eg if 10 days were out of stock over the 1 month amc period (~30.4 days, or change the days in month pref to an even value) then the AMC will be calculated at 1/3 higher with the dos pref on

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/mock/mod.rs
+++ b/server/repository/src/mock/mod.rs
@@ -82,7 +82,6 @@ pub use asset::*;
 pub use asset_log::*;
 pub use barcode::*;
 pub use campaign::*;
-use chrono::Utc;
 pub use clinician::*;
 use common::*;
 pub use contact_form::*;
@@ -1734,64 +1733,5 @@ impl MockData {
         self.goods_received.append(&mut goods_received);
         self.goods_received_line.append(&mut goods_received_line);
         self
-    }
-}
-
-pub fn make_movements(stock_line: StockLineRow, date_quantity: Vec<(i64, i64)>) -> MockData {
-    let (invoices, invoice_lines) = date_quantity
-        .into_iter()
-        .map(|(date, quantity)| {
-            let invoice_id = format!("invoice_{}_{}_{}", stock_line.id, date, quantity);
-            let date = Utc::now().naive_utc() + chrono::Duration::days(date - 30);
-
-            (
-                InvoiceRow {
-                    id: invoice_id.clone(),
-                    store_id: stock_line.store_id.clone(),
-                    name_link_id: mock_name_a().id.clone(),
-                    r#type: if quantity > 0 {
-                        InvoiceType::InboundShipment
-                    } else {
-                        InvoiceType::OutboundShipment
-                    },
-                    status: if quantity > 0 {
-                        InvoiceStatus::Verified
-                    } else {
-                        InvoiceStatus::Shipped
-                    },
-                    created_datetime: date,
-                    allocated_datetime: Some(date),
-                    picked_datetime: Some(date),
-                    shipped_datetime: Some(date),
-                    delivered_datetime: Some(date),
-                    received_datetime: Some(date),
-                    verified_datetime: Some(date),
-                    ..Default::default()
-                },
-                InvoiceLineRow {
-                    id: format!("line_{invoice_id}"),
-                    invoice_id,
-                    item_link_id: stock_line.item_link_id.clone(),
-                    stock_line_id: Some(stock_line.id.clone()),
-                    pack_size: stock_line.pack_size,
-                    number_of_packs: quantity.abs() as f64,
-                    r#type: if quantity > 0 {
-                        use InvoiceLineType;
-
-                        InvoiceLineType::StockIn
-                    } else {
-                        InvoiceLineType::StockOut
-                    },
-                    ..Default::default()
-                },
-            )
-        })
-        .unzip();
-
-    MockData {
-        invoices,
-        invoice_lines,
-
-        ..Default::default()
     }
 }
--- a/server/service/src/test_helpers.rs
+++ b/server/service/src/test_helpers.rs
@@ -1,7 +1,11 @@
 use std::sync::Arc;
 
 use actix_rt::task::JoinHandle;
-pub(crate) use repository::mock::make_movements;
+use chrono::Utc;
+use repository::{
+    mock::mock_name_a, InvoiceLineRow, InvoiceLineType, InvoiceRow, InvoiceStatus, InvoiceType,
+    StockLineRow,
+};
 use repository::{
     mock::{MockData, MockDataInserts},
     test_db::setup_all_with_data,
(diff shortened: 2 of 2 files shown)


GeronimoJohn (commented):
Thanks @aimee-mcneil-melville, seems like a fun piece of feature to write! As per discussion, will literally wait for a day to test DOS 

I've adjusted an item, that have AMC, to be out of stock - hopefully see results tomorrow

Just left a few things below 😃 

Wondering if we should get Roxys thoughts as well on this at it seems to be a new pattern?

aimee-mcneil-melville:
Thanks John, I have also made some changes so that postgres can handle the datetimes and days difference

server/repository/src/mock/mod.rs
```diff
@@ -1735,3 +1736,62 @@ impl MockData {
         self
     }
 }
+
+pub fn make_movements(stock_line: StockLineRow, date_quantity: Vec<(i64, i64)>) -> MockData {
```
GeronimoJohn: Just wondering if we should move this back to test_helpers? Though not aware of context as to why it got moved here

GeronimoJohn (approved):
Dynamic Query! Looks like a fun one to learn and write. Thanks for the hard yards. 

Working well for me! Would say pretty happy to go in, but since new pattern thought might be good to have another set of eyes on it 👀.  Also left one question

Pref turned off:
Image

Pref turned on:
Image

server/repository/src/mock/mod.rs
```diff
@@ -1735,3 +1736,62 @@ impl MockData {
...
+                    ..Default::default()
+                },
+            )
+        })
+        .unzip();
+
+    MockData {
+        invoices,
+        invoice_lines,
+
```
roxy-dao: Such a nit but whitespace lul

roxy-dao (approved):
Thanks Aimee! We went through this together initially as well! Thanks for the lovely docs too <3

pokemon-pokémon

Can you create another issue for consolidation unless this is already done?

Issue #8486, opened 2025-07-15
Have to click `sync` to get transfer records


roxy-dao:
## What went wrong? 😲
I created an outbound shipment to a store on the same site but had to click sync to get the corresponding inbound shipment

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create an `Outbound Shipment` to another OMS store
2. Add some lines and change to shipped
3. Go to receiving store
4. See record not there
5. Sync -> see record there

## Your environment 🌱

- Open mSupply Version: 2.10
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

jmbrunskill:
@roxy-dao do you know if this is in 2.9 as well? Mainly would affect training I think rather than most client but maybe should be high priority to fix...

mark-prins:
triage: I've just tried on 2.9 and could not replicate - is this on develop? If so can you update the version in the issue? if this is a develop issue then we'd need to fix in current release

andreievg:
I've also seen this, not sure how to consistently replicate

andreievg:
Would suggest next task `needs investigation` in next cooldown

roxy-dao:
It's a bit hard to recreate... and happens intermittently 👀

PR #9667, opened 2025-11-03
9666 fix inbound shipment to be created when added in picked or shipped outbound shipment

### LINKED ISSUE #9666: Adding new line in picked/shipped outbound shipment stock transfer doesn't create corresponding line in its transfer inbound shipment
## What went wrong? 😲
As title says, all those newly added lines in the confirmed CI (mSupply) or picked/shipped outbound shipment (OMS) are not being created to its transfer inbound shipment.

## Expected behaviour 🤔
Should be creating newly added line(s) to its transfer inbound shipment (or SI in mSupply) even when CI or outbound shipment is confirmed/picked/shipped. Note, there is a global pref in the OMS _Prevent transfers from this many months before initialisation_ which prevents creating lines if exceeds the given value from when the site has been initialized.

## How to Reproduce 🔨

1. Create stock transfer outbound
2. Add a single line and then picked or shipped the shipment
3. Sync it to the central server
4. In the other remote site, do force sync and then open inbound shipment
5. You'll have that single item appeared in the shipment
6. Now again go back to that outbound shipment and then add new line to it and then just sync
7. In the inbound shipment, do force sync and see error

- Open mSupply Version: v2.11.xx
- Legacy mSupply Central Server Version: V8-05-XX

- Platform: All

- Database type: Both

raviSussol:
Fixes #9666 

# 👩🏻‍💻 What does this PR do?
Modified inbound shipment line update check when updated from picked outbound shipment

# 🧪 Testing

- [ ] Create stock transfer outbound
- [ ] Add a single line and then picked or shipped the shipment
- [ ] Sync it to the central server
- [ ] In the other remote site, do force sync and then open inbound shipment
- [ ] You'll have that single item appeared in the shipment
- [ ] Now again go back to that outbound shipment and then add new line to it and then just sync
- [ ] In the inbound shipment, do force sync and see error

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
```diff
@@ -90,7 +90,7 @@ impl InvoiceTransferProcessor for CreateInboundInvoiceProcessor {
 
         // 4.
         if linked_invoice.is_some() {
-            return Ok(InvoiceTransferOutput::NoLinkedInvoice);
+            return Ok(InvoiceTransferOutput::AlreadyLinked);
```
Chris-Petty: heh nice spotting!

server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
```diff
@@ -125,7 +128,7 @@ impl InvoiceTransferProcessor for UpdateInboundInvoiceProcessor {
 
         let updated_inbound_invoice = InvoiceRow {
             // 6.
-            status: InvoiceStatus::Shipped,
+            status: outbound_invoice_row.status.clone(),
```
Chris-Petty: I like this, arguably should have already been this!

server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
```diff
@@ -67,7 +67,10 @@ impl InvoiceTransferProcessor for UpdateInboundInvoiceProcessor {
             ));
         }
         // 5.
-        if outbound_invoice.invoice_row.status != InvoiceStatus::Shipped {
+        if !matches!(
+            &outbound_invoice.invoice_row.status,
+            InvoiceStatus::Shipped | InvoiceStatus::Picked,
```
Chris-Petty: To preface here is my test setup (which should at least work, and I think generally sufficient for testing this cause it's the same logic being triggered)

1. OMS site with 2 stores A and B
2. Make an outbound shipment with 1 line and "Picked" to B
3. Check inbound shipment in B
4. Add a line to the outbound shipment 
5. Check inbound on B (generally this works)
6. 4-5 again, sometimes doesn't work?
7. Try delete one (or all) lines in outbound shipment (not always reflected in B)

This works, sometimes? I was excited 'cause the first time I tried it worked great, but it seems it isn't always triggering still. From a quick skim I think the only reason we even get a changelog for the invoice to process is because the "picked_datetime" gets updated here:

https://github.com/msupply-foundation/open-msupply/blob/f4e1f2c551f7d64c4f2d462c7b48bffe00888f6e/server/service/src/invoice_line/stock_out_line/update/mod.rs#L90-L95

That fn in turn tries not to spam updates, skipping if already done in the last minute:

https://github.com/msupply-foundation/open-msupply/blob/f4e1f2c551f7d64c4f2d462c7b48bffe00888f6e/server/service/src/invoice/update_picked_date.rs#L61-L69

That said, if I observe the change log it does appear to pretty reliably update the outbound shipment (`019a4d15-ab2c-799c-929a-5cc0a28a319f`).

Image

But, the _last_ changelog shows that the inbound shipment (`019a4d15-d467-7053-a1f6-7395ac43ff70`) _isn't_ being updated, it should have a cursor number higher than the outbound shipment if this processor ran afterwards. Or at least it's lines should!

...Hmm I checked again a few minutes later and it eventually did run! Maybe relates to https://github.com/msupply-foundation/open-msupply/issues/8486 ? (I still have an older invoice from an earlier test of this that hasn't gotten updated, so updates are definitely still not quite happening reliably)

Image

Funnily enough that last picture highlights what I believe me and @andreievg discussed in the past when I previously highlighted that updating a picked invoice doesn't get reflected in the other site. What happens is that every time it's triggered it literally deletes all lines of the inbound shipment and recreates them. If this shipment has thousands of lines, all potentially being updated and all potentially triggering this process, it could churn an exponential amount of database updates (and sync messages via changelog)! 

So I'm hesitant to suggest going further with this PR as it is as it could get really bad DB performance (not to mention blow up the changelog table haha). However that's not to say the issue isn't valid! I think the approach is to carefully think through how we'd rather do this instead of the current brutish way. 

What I think Drei and I previously discussed was implementing a new processor that looks at the `invoice_line` changelog and handles individual line updates on inbound shipments in `picked` status. I'd probably the existing processor here as it was, only operating on `SHIPPED`, that way it'd have a final pass. Maybe, I'd need to see if there is a guaranteed order of processors. Up for discussion/investigation a bit! I'd just definitely not want the inbound getting auto updates after it's technically already become editable (even if those updates are the same as the previous processor, it may override user inputs if we don't control the window or order the processors run in)

Does that sounds reasonable, would you be happy to go a bit deeper and broader writing a new processor for I guess `update_inbound_invoice_line.rs`?

Chris-Petty (changes requested):

server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
raviSussol: @Chris-Petty sorry to get back on this lately. That's a thorough review/testing! I also managed to test with few more items (adding new lines) - doing upsert/delete back and forth and could see slow update in inbound shipment (like you've mentioned due to 60 second cool down to prevent spam). Plus really slow performance (even editing a single line - especially when you add lines from master list). I pretty much agree everything you said and sounds like a plan to me to have line-level processor rather than invoice level processor for update which not only boost performance but also reduce unnecessary sync traffics.

I'll update this pr as per line-level processor idea/suggestion.

Chris-Petty: No issue with the 60s throttling! I think it's a good idea

PR #9677, opened 2025-11-04
2 Item Sell Price plugin oms changes

aimee-mcneil-melville:
Fixes #[2](https://github.com/msupply-foundation/congo-plugins/issues/2)

# 👩🏻‍💻 What does this PR do?

Use a global sell price for an item is on an essential medicines master list

No changes that affect non-plugin users

Defined the sell price component so that it can be imported into the plugin and appear the same, could also create it as a new file if that's cleaner

The plugin components are separate in the existing `Store` tab - PRD not up to date on this. Allows flexibility if new fields are added (either with or without the plugin)

# 🧪 Testing

- [ ] Before installing the plugin, see that Item -> Detail view -> Store tab has a default sell price field and it is disabled
- [ ] Add a sell price to an item (in OG) that is on some master lists
- [ ] Rest to be tested with the plugin installed - testing steps in linked issue

# 📃 Documentation

- [] **Part of an epic**: documentation will be completed for the feature as a whole
  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/plugins/types.ts
+++ b/client/packages/common/src/plugins/types.ts
@@ -12,7 +12,7 @@ export type Plugins = {
   prescriptionPaymentForm?: React.ComponentType<PrescriptionPaymentComponentProps>[];
   itemProperties?: {
     ItemSellPrice?: React.ComponentType<{ item: ItemFragment }>[];
-    ItemFooter?: React.ComponentType<{ item: ItemFragment }>[];
+    ItemFooter?: React.ComponentType<{ itemId: string }>[];
   };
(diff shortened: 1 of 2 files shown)


andreievg (changes requested):
The issue says:
`"should overwrite original 'sell price' from OG"`.

I would like to change that to say 'catalogue price' as extra column/input, rather the overwriting display of original sell price.

This would mean this would be an extension item field, similar to the type for stock line.

PR #9705, opened 2025-11-06
#9693 Fix for requsition "reason" column error

### LINKED ISSUE #9693: Navigating to Requisition DetailView shows error
## What went wrong? 😲

Navigate to a Requisition Detail View results to an error saying 

`getError is not a function`

After some investigation, we found that theres a method from our old context that is being used within the `reason` column. 

Reason columns shows if items within the requisition are part of a program and store preferences extraFieldsInRequisition is true

## How to Reproduce 🔨

1. Navigate to a Requisition with items that are part of a program
2. See error `getError is not a function`

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

CarlosNZ:
Fixes #9693 

# 👩🏻‍💻 What does this PR do?

The Provider was only wrapped around the returned JSX, so the columns hook didn't have access to it. Have now wrapped the whole component, so `getError` is defined again.

Also took the opportunity to remove the TableProvider, which is no longer required with the new Material tables.
# 🧪 Testing

- [ ] See issue, confirm problem is no longer present

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/requisitions/src/ResponseRequisition/context/responseRequisitionLineError.tsx
+++ b/client/packages/requisitions/src/ResponseRequisition/context/responseRequisitionLineError.tsx
@@ -39,13 +39,18 @@ export type UseResponseRequisitionLineErrors = ReturnType<
   typeof useResponseRequisitionLineErrors
 >;
 
-const ResponseRequisitionLineErrorContext =
-  createContext<UseResponseRequisitionLineErrors>({} as any);
+const ResponseRequisitionLineErrorContext = createContext<
+  UseResponseRequisitionLineErrors | undefined
+>(undefined);
 
 export const useResponseRequisitionLineErrorContext = () => {
   const context = useContext(ResponseRequisitionLineErrorContext);
 
-  if (!context) throw new Error('Context does not exist');
+  if (!context) {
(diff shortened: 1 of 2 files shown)


client/packages/requisitions/src/ResponseRequisition/context/responseRequisitionLineError.tsx
```diff
@@ -39,13 +39,18 @@ export type UseResponseRequisitionLineErrors = ReturnType<
...
 
 export const useResponseRequisitionLineErrorContext = () => {
   const context = useContext(ResponseRequisitionLineErrorContext);
 
-  if (!context) throw new Error('Context does not exist');
+  if (!context) {
+    throw new Error(
+      'useResponseRequisitionLineErrorContext must be used within a ResponseRequisitionLineErrorProvider'
+    );
+  }
```
CarlosNZ: This was actually incorrect too, as the context was initialised with `{}`, which is truthy so would never throw error like it's supposed to.

PR #9716, opened 2025-11-08
9696 warn stocktake complex pref

### LINKED ISSUE #9696: Prompt the user to perform stocktake when creating an internal order

As part of their standard procedures, the it is best practice to perform a stocktake before placing their monthly order

We did not want to implement a hard restriction within Open mSupply however. There is also a request to make this a general feature.

Tasks:

Create a store preference for Warn if creating an internal order when stocktake is older than this many days
Create a store preference for Warn if creating an internal order: minimum items on stocktake 
When a user creates an internal order:
If the ‘days’ preference is > 0, and
There are no finalised stocktakes with a finalised date within the previous x days (given by the store preference), where
Any stocktakes with fewer than the minimum items are excluded, then
Show a confirmation modal prior to the supplier selection modal. The confirmation says “You have not completed a stocktake within the last [x] days. Your stock levels may not be accurate. Are you sure that you want to continue?”
Example of option should be:
Yes, continue
No, go to stocktake page
Cancel

jmbrunskill:
Fixes #9696 (adding prefs)

# 👩🏻‍💻 What does this PR do?

This is an attempt to create a complex/structured preference, it's still a bit messy but I think better than having lots of ungroup-ed prefs?

Basically uses a couple of styling components to improve consistency, and moves the logic inside the existing switch statement so if feels more natural to have complex prefs using different display components.

With this change we'll need to update the frontend query everytime we add a pref, I think that's ok as it's what we do for every other query?

# 🧪 Testing

- [ ] Check both global and store pref still save correclty.
- [ ] Check you can save and set the new complex pref...

# 📃 Documentation

  1.
  3.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Manage/Preferences/EditPage/PreferenceAccordion.tsx
+++ b/client/packages/system/src/Manage/Preferences/EditPage/PreferenceAccordion.tsx
@@ -25,6 +25,9 @@ export const PreferenceAccordion = ({
         borderColor: 'grey.400',
         borderRadius: 1,
         boxShadow: 'none',
+        '&::before': {
+          display: 'none', // Common use case - hide the default border
+        },
         ...sx,
       }}
     >


client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
```diff
@@ -0,0 +1,113 @@
...
+  Typography,
+  NumericTextInput,
+  Switch,
+  WarnWhenMissingRecentStocktakeDataNode,
+  InputWithLabelRow,
+  useTranslation,
+} from '@openmsupply-client/common';
+
+// Backend returns snake_case, but TypeScript types expect camelCase
+type WarnWhenMissingRecentStocktakeDataSnakeCase = {
```
jmbrunskill: This is the messiest thing... Not sure why it's `snake_case` in one context and `camelCase` in the other...

client/packages/common/src/authentication/api/hooks/usePreferences.ts
```diff
@@ -13,17 +13,37 @@ type PreferencesQuery = {
...
   const { client } = useGql();
   const { storeId } = useAuthContext();
 
+  // Build field list with subfields for complex types
+  const fields = Object.values(PreferenceKey)
+    .map(key => {
+      const subfields = COMPLEX_PREFERENCE_FIELDS[key];
+      return subfields ? `${key} { ${subfields} }` : key;
+    })
+    .join('\n');
```
Chris-Petty: Ah of course, hadn't considered what pain this might cause for GQL. An alternative would be having this discoverable from the API through gql introspection? Makes an extra round trip for displaying the UI though (if we're not already doing that, I haven't looked at this part deeply).

client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const handleMaxAgeChange = (newMaxAge?: number | undefined) => {
+    createUpdateValue({ maxAge: newMaxAge });
+  };
+
+  const handleMinItemsChange = (newMinItems?: number | undefined) => {
+    createUpdateValue({ minItems: newMinItems });
+  };
+
+  return (
```
Chris-Petty: As in my last comment, if we're making a specific component anyway then maybe not worth all the effort?

server/service/src/preference/preferences/warn_when_missing_recent_stocktake.rs
```diff
@@ -0,0 +1,26 @@
+use crate::preference::{PrefKey, Preference, PreferenceType, PreferenceValueType};
+
+pub struct WarnWhenMissingRecentStocktake;
+
+#[derive(Default, Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
```
Chris-Petty: My reckons is maybe here you need to give `serde` or `graphql` an attribute to serialise/rename fields to camelCase?

Chris-Petty (changes requested):
As you mentioned this morning... hmm is this worth the effort and is it really better than just having 4 prefs? We do end up writing in react the pref interface anyway, so not really avoiding frontend changes currently!

I think for usage of this pref in business logic of front end, it is nice having the group structure (though in gql you could perhaps still achieve one gql query to get several pref records without issue). Similar for backend (if other prefs used this approach), rather than having to get 4 separate prefs just get one which is somewhat convenient!

client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
CarlosNZ: Agree, the inner prefs UI should be auto-rendered like the root layer, otherwise not much point have them dynamically queried, right?

CarlosNZ: I think it should be reasonably straightforward to create a generic "ComplexPref" wrapper that renders the inner prefs with the same code as the root prefs?

client/packages/system/src/Name/ListView/Stores/EditStorePreferences.tsx
```diff
@@ -37,6 +42,43 @@ export const EditStorePreferences = ({
...
+          sx={{
+            marginTop: 1,
+            border: '1px solid',
+            borderColor: 'grey.400',
+            borderRadius: 1,
+            boxShadow: 'none',
+          }}
+        >
+          <AccordionSummary
+            expandIcon={<ExpandIcon />}
```
CarlosNZ: Needs the opposite icon ("Collapse" I think) when open -- currently it's just spinning the same icon 180 degrees (which looks identical! 😆)

CarlosNZ: But I thought this was our more standard Accordion UI:

Image

Is there precedent for your one?

client/packages/common/src/authentication/api/operations.graphql
```diff
@@ -184,3 +184,40 @@ query lastSuccessfulUserSync {
     ...UpdateUser
   }
 }
+
+query preferences($storeId: String!) {
```
jmbrunskill: Standard query for preferences now...

client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
jmbrunskill: I've tried to capture this idea by creating the `PreferenceAccordion` component, so that can contain what ever groupings or custom components are needed.

client/packages/system/src/Manage/Preferences/EditPage/PreferenceAccordion.tsx
```diff
@@ -0,0 +1,40 @@
...
+        marginTop: 1,
+        border: '1px solid',
+        borderColor: 'grey.400',
+        borderRadius: 1,
+        boxShadow: 'none',
+        ...sx,
+      }}
+    >
+      <AccordionSummary
+        expandIcon={<ChevronDownIcon />}
```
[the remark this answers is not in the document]
jmbrunskill: Hopefully consistent enough for your @CarlosNZ Do we have some other existing component I should be using?

client/packages/system/src/Manage/Preferences/EditPage/PreferenceLabelRow.tsx
```diff
@@ -0,0 +1,27 @@
...
+  label,
+  Input,
+  isLast = false,
+}: PreferenceLabelRowProps) => {
+  return (
+    <InputWithLabelRow
+      labelWidth={'100%'}
+      label={label}
+      Input={Input}
+      sx={{
```
jmbrunskill: This just keeps styling consistent.
I've now made both store and global prefs show the label on the left...

client/packages/common/src/authentication/api/operations.graphql
Chris-Petty: I agree fine for now... 

This critique is true of most of our app's API consumption but particular here I believe we make a query for each preference specified so every time we call this we do a good 30 DB queries to get the response. (I haven't looked, maybe there is some loader shenanigans where we query once for all of them into a hashmap?)

Would be nice to have something like `usePreferences(storeId: String, keys: String[])` that makes the actually appropriate gql query. As it is across the app we pay for the complexity of implementing gql and then don't bother actually making specific gql queries 😅.

CarlosNZ (commented):
I think this is basically fine for now, though it'd be good to look at a more consistent overall handling of groups in both Global and Store prefs.

I made one suggestion to reuse the "EditPreference" component inside your custom group wrapper -- would be more consistent and simpler code if it'll work?

CarlosNZ:
There's also a compilation error in `EditGlobalPreferencesPage`, due to the `sx` prop passed to `EditPreference` no longer being valid.

jmbrunskill:
> There's also a compilation error in `EditGlobalPreferencesPage`, due to the `sx` prop passed to `EditPreference` no longer being valid.

I thought I'd removed but obviously not, thanks!

client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <PreferenceLabelRow
+          label={t('preference.warnWhenMissingRecentStocktake.enabled')}
+          Input={
+            <Switch
+              disabled={disabled}
+              checked={enabled}
+              onChange={(_, checked) => createUpdateValue({ enabled: checked })}
+            />
+          }
+        />
```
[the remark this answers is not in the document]
jmbrunskill: Ok, this gets messy because the `EditPreference` provides the debouncing code and accepts an `update` method, `EditWarningWhenMissingRecentStocktakeData` needs to receive a handleChange method idealy but would need to pass the update method again, and we also have to fake new preferenceNodes, needs a rethink how we do the whole thing I think. Will leave it till we build a proper complex pref wrapper component I think? Although feel free to add a suggestion branch?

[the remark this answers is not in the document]
CarlosNZ: I think I can make this work, and make it a bit cleaner overall -- I'll push a suggestion in an hour or so if that's okay?

client/packages/system/src/Manage/Preferences/EditPage/PreferenceAccordion.tsx
```diff
@@ -0,0 +1,40 @@
...
+}: PreferenceAccordionProps) => {
+  return (
+    <Accordion
+      sx={{
+        marginTop: 1,
+        border: '1px solid',
+        borderColor: 'grey.400',
+        borderRadius: 1,
+        boxShadow: 'none',
+        ...sx,
```
CarlosNZ: To fix the inconsistent top border, you can add:
```
    '&::before': {
          display: 'none', // Common use case - hide the default border
        },
```

Changes it from this:
Image
to this:

Image

client/packages/system/src/Manage/Preferences/Components/EditWarningWhenMissingRecentStocktakeData.tsx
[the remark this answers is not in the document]
CarlosNZ: Okay, never mind -- I can make it work, but it doesn't really save anything in terms of code complexity or amount. Can revisit later if we need to.

CarlosNZ (approved):
Okay, I think this is fine now. Can you you just make that small style tweak to the top border I've suggested?

I'm a bit concerned that we now have two different ways to "group" prefs, but it'll need a bigger refactor to consolidate it all, so we should revisit at some point.

PR #9719, opened 2025-11-10
#9712 Table inputs debouncing fix

### LINKED ISSUE #9712: Text regression while typing a batch number or quantities pack and packsize when adding an item to an inbound shipment
## What went wrong? 😲
See [video](https://drive.google.com/file/d/1VfwoJlmT39fNqpkuGpzJkU2w0xWHCocH/view?usp=sharing). You can see that when your typing fast the number doesn`t have to register before being replaced by the next number you have typed.

It happens for the Batch field, all pack size and pack qty fields and volume

## Expected behaviour 🤔
The cursor should not back when typing
## How to Reproduce 🔨

Using latest 2.12 
1. Go to Replenishment > Inbound shipment
2. Click on 'New Inbound shipment'
3. Click on button 'Add item' and choose an item
4. Start typing fast a batch number and quantities and volume
5. See error

## Your environment 🌱

- Open mSupply Version: 2.12.00-develop-11061529
- Legacy mSupply Central Server Version: mSupply 8.5.13

- Platform: Desktop

- Database type: Postgres

CarlosNZ:
Fixes #9712

# 👩🏻‍💻 What does this PR do?

For the most part, just removing the debouncing is the best solution -- if debouncing is needed it should be done at the hook level, not within the component itself.

The exception is the "Allocation" logic for the NumberInputCell (in OutboundShipments) -- there's some special-case logic for that and I didn't want to start unpicking that here, so have made that component take an optional "debounce" time, but with a default of 0. Seems to fix the immediate problem, but we might want to look at this again at some point.

# 🧪 Testing

- [ ] Confirm text and number inputs as described in issue work as expected
- [ ] Confirm OutboundShipment line input still working okay (with debouncing)
- [ ] Maybe check Prescriptions as well? (the batch table in line edit)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
@@ -51,7 +51,7 @@ interface QuantityTableProps extends TableProps {
   removeDraftLine: (id: string) => void;
 }
 
-export const QuantityTableComponent = ({
+export const QuantityTable = ({
   lines,
   updateDraftLine,
   removeDraftLine,
@@ -362,8 +362,6 @@ export const QuantityTableComponent = ({
 
   return <MaterialTable table={table} />;
 };
-
-export const QuantityTable = React.memo(QuantityTableComponent);
 
 export const PricingTableComponent = ({
   lines,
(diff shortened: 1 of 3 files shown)


client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -363,8 +363,6 @@ export const QuantityTableComponent = ({
   return <MaterialTable table={table} />;
 };
 
-export const QuantityTable = React.memo(QuantityTableComponent);
```
CarlosNZ: Seems unnecessary now with the new tables

jmbrunskill (approved):
Thanks Carl, that's working well. At first it didn't appear to work but after restarting my frontend it's working correctly

PR #9749, opened 2025-11-13
9690 auto finalise inbound shipments pref

### LINKED ISSUE #9690: Auto-Finalise Inbound Shipments

When outbound shipments are sent to a site, some clients want to assume this has arrived.
To do this we'll need to automatically finalise the inbound shipment on the remote store, and update stock assuming the shipment has arrived.

Add a store preference for Auto verify stock transfer inbound shipments  which has the same behaviour as the mSupply preference Finalise stock transfer supplier invoice automatically

When the inbound shipment is automatically created based on the outbound, if the store preference is enabled in the customer store:
Status is updated to verified
The verified date is set to the current date
The intermediate status dates also set to the current date
Items on the shipment are brought into stock

### Moneyworks Jobcode 🧰

`OMS:NECSR`

Chris-Petty:
Fixes #9690 

# 👩🏻‍💻 What does this PR do?
image

- Adds preference to auto verify inbound shipments
- Does exactly that when receiving "Shipped" shipments

I was going to try rethinking through errors in this PR as the processor returns a "RepositoryError" even though it may be a `PreferenceError` or a `UpdateInboundShipmentError` (which itself has a `DatabaseError(RepositoryError)` variant). It's all sorta a bit confusing but maybe I just don't get it. 

# 🧪 Testing

- [ ] Turn the pref on
- [ ] Between 2 OMS stores on the same site create an outbound shipment to the other and mark it as Picked
- [ ] The other store gets an inbound shipment in "Picked" status
- [ ] "Ship" the outbound shipment
- [ ] The other store's inbound shipment is changed to "Verified" and stock is added
- [ ] Make another outbound shipment and immediately set it to shipped, skipping picked.
- [ ] The other store received an inbound shipment and it is already "verified"
- [ ] There is an activity log indicating that `omsupply_system` verified the invoice
- [ ] Test the above with the two stores on 2 separate OMS sites and sync between the steps
- [ ] Test the above but the supplying store is an OG store

# 📃 Documentation

  1. Prefs list
  2. Inbound shipments perhaps

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
+++ b/server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
@@ -276,7 +276,7 @@ mod test {
 
         // Setup for the test with the preference on
         let preference = PreferenceRow {
-            id: PrefKey::InboundShipmentAutoVerify.to_string() + "_" + mock_store_a().id.as_str(),
+            id: "preference_on".to_string(),
             key: PrefKey::InboundShipmentAutoVerify.to_string(),
             value: "true".to_string(),
             store_id: Some("store_a".to_string()),
--- a/server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
+++ b/server/service/src/processors/transfer/invoice/create_inbound_invoice.rs
@@ -555,7 +555,7 @@ mod test {
         );
 
         let preference = PreferenceRow {
-            id: PrefKey::InboundShipmentAutoVerify.to_string() + "_store_a",
+            id: "preference_on".to_string(),
             key: PrefKey::InboundShipmentAutoVerify.to_string(),
             value: "true".to_string(),
             store_id: Some("store_a".to_string()),


server/service/src/invoice/inbound_shipment/update/generate.rs
```diff
@@ -385,36 +388,34 @@ pub fn generate_lines_and_stock_lines(
             ..
         }: InvoiceLineRow = invoice_line;
 
-        if number_of_packs > 0.0 {
```
Chris-Petty: Ah yea this was a minor side quest - we were skipping it in the end so why do all the work earlier in the loop?

server/service/src/processors/transfer/invoice/common.rs
```diff
@@ -146,3 +153,48 @@ pub(crate) fn convert_invoice_line_to_single_pack(
...
+
+    if should_auto_verify {
+        update_inbound_shipment(
+            ctx,
+            UpdateInboundShipment {
+                id: inbound_shipment.id.to_string(),
+                status: Some(UpdateInboundShipmentStatus::Verified),
+                ..Default::default()
+            },
+            Some(&inbound_shipment.store_id),
```
Chris-Petty: All the context changing was so I could call this method with a ctx.

Unfortunately the ctx started at the beginning of the call chain can't know what store_id and user_id we'll need, so just tacked a store_id in here.

That said... I need to do user_ID too in some capacity, as right now it's not making an activity log correctly!

server/service/src/service_provider.rs
```diff
@@ -330,6 +331,19 @@ impl ServiceProvider {
         })
     }
 
+    pub fn system_context(
```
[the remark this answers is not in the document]
jmbrunskill: Yeah probably better not to have more code for rust to try to evaluate even though it doesn't do anything...

server/service/src/processors/transfer/invoice/mod.rs
```diff
@@ -299,11 +299,12 @@ trait InvoiceTransferProcessor {
...
         &self,
-        connection: &StorageConnection,
+        ctx: &ServiceContext,
         record: &InvoiceTransferProcessorRecord,
     ) -> Result<Option<String>, ProcessorError> {
-        let output = connection
-            .transaction_sync(|connection| self.try_process_record(connection, record))
+        let output = ctx
+            .connection
+            .transaction_sync(|_| self.try_process_record(ctx, record))
```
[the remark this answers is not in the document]
jmbrunskill: Nice, thanks for the summary

server/service/src/invoice/inbound_shipment/batch.rs
```diff
@@ -177,11 +177,21 @@ pub fn batch_inbound_shipment(
...
-            let (has_errors, result) = mutations_processor
-                .do_mutations_with_user_id(input.update_shipment, update_inbound_shipment);
-            results.update_shipment = result;
-            if has_errors && !continue_on_error {
-                return Err(WithDBError::err(results));
+            if let Some(update_shipment_inputs) = input.update_shipment {
+                let output: Vec<_> = update_shipment_inputs
+                    .into_iter()
+                    .filter_map(|input| {
+                        let result = update_inbound_shipment(ctx, input.clone(), None);
```
Chris-Petty: All these changes are because I changed the `update_inbound_shipment` fn signature with the new 3rd parameter, so it no longer fit the signature `do_mutations_with_user_id` which expects the callback to have 2 parameters

Chris-Petty: It's worth noting that functionally it is slightly different, but I suspect that the old generic method is doing it a less than ideal way:

https://github.com/msupply-foundation/open-msupply/blob/73029fd38a3e4076dddcd59d08f54fc7a3096456/server/service/src/lib.rs#L228-L250

After the first error is encountered, it'll include all subsequent results even if they don't contain errors. The implementation I have done here for `update_inbound_shipment` only includes results that contain errors.

server/service/src/processors/transfer/invoice/common.rs
Chris-Petty: I made another commit after my comment after realising the logging wasn't working correctly! I ended up just making the ctx `system_context` (so actually used now) rather than `basic_context` which includes the OMS user ID constant.

server/service/src/processors/transfer/invoice/update_inbound_invoice.rs
```diff
@@ -164,3 +170,166 @@ impl InvoiceTransferProcessor for UpdateInboundInvoiceProcessor {
...
+    async fn test_update_inbound_invoice_auto_finalise() {
+        fn get_invoice_row(connection: &StorageConnection, id: String) -> InvoiceRow {
+            InvoiceRepository::new(connection)
+                .query_one(InvoiceFilter::new().id(EqualFilter::equal_to(id)))
+                .unwrap()
+                .unwrap()
+                .invoice_row
+        }
+
+        let first_half_row = InvoiceRow {
```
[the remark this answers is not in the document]
Chris-Petty: In irl discussions I find we often revert to saying first and second half so I leant on that here for brevity

server/service/src/activity_log.rs
```diff
@@ -63,6 +63,34 @@ pub fn activity_log_entry(
     Ok(())
 }
 
+/// Use instead of `activity_log_entry()` to provide the store_id if the ctx doesn't include it (e.g. system or basic context used in changelog processors)
+pub fn activity_log_entry_with_store(
```
Chris-Petty: I left the old method name alone to avoid updating a bajillion files. This name seems clear enough including the doc comment?

PR #9602, opened 2025-10-23
9528 Configurable supply level name properties

GeronimoJohn:
Fixes #9528

# 👩🏻‍💻 What does this PR do?

Added the ability to configure supply levels. And for it to be used within Stores.

Screenshot 2025-10-29 at 12 11 29 PM

Screenshot 2025-10-29 at 12 13 23 PM

## 💌 Any notes for the reviewer?

# 🧪 Testing

This is assuming that you have Sys Admin permission and on a central server
- [ ] Navigate to Settings Page
- [ ] Click on Configuration Section
- [ ] Click on Edit button for `Configure Supply Levels`
- [ ] **_Click on the Text Input to add a Supply Level_**
- [ ] Configure Levels (add levels or remove levels)
- [ ] **_Navigate to the Stores Page and select any store_** 
- [ ] Should be able to set Supply Level based on what was configured

# 📃 Documentation

- [ ] **Part of an epic**: documentation will be completed for the feature as a whole
- [ ] **No documentation required**: no user facing changes or a bug fix which isn't a change in behaviour
- [ ] **These areas should be updated or checked**: 
  1.
  2.

# 📃 Reviewer Checklist

The PR Reviewer(s) should fill out this section before approving the PR

**Breaking Changes**
- [ ] No Breaking Changes in the Graphql API
- [ ] Technically some Breaking Changes but not expected to impact any integrations

**Issue Review**
- [ ] All requirements in original issue have been covered
- [ ] A follow up issue(s) have been created to cover additional requirements

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

PR #9769, opened 2025-11-14
9750 Add same supply level transfer logic

### LINKED ISSUE #9750: Add same supply level transfer logic

Currently in AMC calculation transfers are determined as invoice having linked_invoice_id (inter store). 

There is client requirement to extend this to be based on supply level name property (same level = transfer). 

In order to extend core in generic way, we've made transfer logic as a method [get_transfer_consumption_map](https://github.com/msupply-foundation/open-msupply/blob/b493f3b0378d00fd4e3077fe0b2f4d7e404112df/server/service/src/item_stats.rs#L101).

Above should be extended as plugin interface. 

As alway for plugins input should be a struct (can be extended in the future to take extra params).

Also would like consumption view to have invoice type and name id.

After above is done add plugin to filter out consumption map based on name id supply level matching current store. May need to also pass along store_id to the input. Need some thoughts about querying for names associated with a property.

roxy-dao:
Fixes #9750

# 👩🏻‍💻 What does this PR do?
Support for excluding same supply level from consumption values plugin' & consolidation of consumptions maps... There were 3 methods and... like we really only need one 👀 

## Note
Review in conjunction with https://github.com/msupply-foundation/congo-plugins/pull/16

# 🧪 Testing

- [ ] Get plugin' bundle from https://github.com/msupply-foundation/congo-plugins/pull/16
- [ ] Install plugin
- [ ] Make sure you have stores set up with same supply level, and ones with different supply level. See #9602 for instructions
- [ ] Toggle the global preference `Exclude transfers` under the AMC accordion
- [ ] If plugin' installed correctly then AMC amount should not include consumption from same level stores
- [ ] can toggle global preference on and off to see diff or do the calculation yourself

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/views/consumption.rs
```diff
@@ -26,25 +26,27 @@ impl ViewMigrationFragment for ViewMigration {
...
+            CREATE VIEW consumption AS
+                SELECT
+                    'n/a' as id,
+                    items_and_stores.item_id AS item_id,
+                    items_and_stores.store_id AS store_id,
+                    {absolute}(COALESCE(stock_movement.quantity, 0)) AS quantity,
+                    date(stock_movement.datetime) AS date,
+                    stock_movement.invoice_type AS invoice_type,
+                    stock_movement.name_id AS name_id,
+                    stock_movement.name_properties AS name_properties,
```
roxy-dao: Thought we could just return the property here since it's available 👀

server/service/src/item_stats.rs
```diff
@@ -219,31 +204,21 @@ impl amc::Trait for DefaultAmc {
...
-    consumption_rows: &Vec<ConsumptionRow>,
-) -> Result<HashMap<String /* item_id */, f64 /* transfer consumption */>, RepositoryError> {
-    let mut transfer_consumption_map = HashMap::new();
-    for consumption_row in consumption_rows.into_iter() {
-        if consumption_row.is_transfer == true {
-            let item_transfer_consumption = transfer_consumption_map
-                .entry(consumption_row.item_id.clone())
-                .or_insert(0.0);
-            *item_transfer_consumption += consumption_row.quantity;
+        if exclude_transfers && consumption_row.is_transfer {
```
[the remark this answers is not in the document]
andreievg: I agree tbh, it was in the original spec thought so, but I guess it was outdated. I'll ask PO

andreievg (changes requested):
Looks good thanks Roxy.

Sorry I only realised when looking at the PR that passing through consumption rows to plugin could lead to a lot of serialisation/deserialisation for plugins that could lead to performance issues, especially if there are a lot of transactions (prescriptions) and lookback period is big.

For that reason I think we should have this input to the plugin:

store_id
item_ids
start_date
end_date

And it should return transfer_map (as per before).

I think we will have generic transfer logic at some point, should it be based on invoice having linked transaction id ? (not sure, maybe the intention was to have generic logic that looks at supply level). Since we don't have agreed generic logic, I think it's ok to remove the pref for now and just use plugin.

The logic in AMC calculation I think should still be somewhat sequential (don't really think plugin should be substracting transfers, it should provide a map of item_id and transfer quantity, and then the core logic can substract it).

Also the plugin name should be generic (as per another comment)

server/service/src/backend_plugin/types/get_consumption.rs
```diff
@@ -0,0 +1,31 @@
...
+#[derive(TS, Clone, Deserialize, Serialize)]
+#[ts(rename = "GetConsumptionInput")]
+pub struct Input {
+    pub store_id: String,
+    pub item_ids: Vec<String>,
+    pub start_date: String,
+    pub end_date: String,
+}
+
+pub type Output = HashMap<String /* item_id */, f64 /* transfer consumption */>;
```
andreievg: Comment here, consumption rather then transfer consumption

andreievg (approved):
Looks great, thanks Roxy, a couple of minor things:

I think backend types need to be re-generated ? Looks like there are remainders from consumption rows (invoice types etc..).

Comment in the Output

PR #9806, opened 2025-11-18
9785 Threshold input for expiring items

### LINKED ISSUE #9785: Threshold for expiring items field issues
## What went wrong? 😲

1. First threshold can be set higher than second threshold (if both start at 0). Also it possibly shouldn't be higher than 30?

Image

2. If you set first threshold high first, then set second to something lower, you are able to save 

Image

3. If you type a value and then backspace it (without saving in-between, just type and delete), you run into the following error: 

Image

## Expected behaviour 🤔

Values possibly shouldn't be able to be set higher than 30, and no errors when entering and backspacing values in the same go

## How to Reproduce 🔨

1. Go to Edit -> Preferences and make sure both fields are sitting at 0
2. Set first threshold to something high - e.g. 60
3. You are able to save
4. Set second threshold to something lower - e.g. 20
5. You are able to save
6. Type something in a cell and backspace it - running into an error

## Your environment 🌱

- Open mSupply Version: `V2.13.00-RC-11130357`
- Legacy mSupply Central Server Version: `V8.06.00`

- Database type: `PostgreSQL`

aimee-mcneil-melville:
Fixes #9785

# 👩🏻‍💻 What does this PR do?

- Show a notification if either the first or second threshold is over 30 days
- Show an error border on numeric input preference types if the update didn't save
- Always able to close the modal (saves automatically, not on Ok)
- Fix numeric inputs being backspaced/cleared - sends 0 to graphql, instead of empty which causes graphql error

https://github.com/user-attachments/assets/ea4ecd71-9d25-4ea9-a298-b12eed405102

These changes will still allow the first threshold to be higher than the second threshold, which when used to open the expiring items view it will have no results. This is not a breaking change and the user can now go back to edit the thresholds (which they couldn't before)

Due to the setup of preferences where the inputs are not individually defined, and save automatically after entering, decided to simplify the validation rather than block the user from being able to edit these inputs.

The error border will apply to any numeric inputs in both global and store prefs where the update was not successful

Looked at adding an alert in the modal to notify the user that the value reverted to the previously saved one - timeboxed it and would require a fair bit more reworking to get the state to the correct level so left that out. The error border provides more feedback to the user that something went wrong than before where it just reverted

On a sidenote global preferences just don't update if the value is empty - have to explicitly type 0. Hence the graphql error only showed in store preferences when trying to update a required value to undefined.

# 🧪 Testing

- [ ] Open store preferences modal -> preferences tab
- [ ] See first and second threshold inputs
- [ ] Try to save a value in either of them that is more than 30 -> see toast warning and error border on the relevant input
- [ ] Enter a value less than 30 -> see it saves and error border is gone
- [ ] Backspace a saved value -> see no error -> close and reopen modal -> see 0 is saved

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Manage/Preferences/api/utils.ts
+++ b/client/packages/system/src/Manage/Preferences/api/utils.ts
@@ -3,20 +3,16 @@ import {
   TypedTFunction,
   UpsertPreferencesInput,
 } from '@openmsupply-client/common';
-import { Dispatch, SetStateAction } from 'react';
 
 export const inputValidation = (
   input: Partial<UpsertPreferencesInput>,
   t: TypedTFunction<LocaleKey>,
-  warning: (msg: string) => () => void,
-  setActionValid?: Dispatch<SetStateAction<boolean>>
+  warning: (msg: string) => () => void
 ): boolean => {
   const thresholdResult = thresholdValidation(input, t, warning);
 
   // Combine results
   const isValid = thresholdResult; // && otherResult
-
-  setActionValid?.(isValid);
 
   return isValid;
 };
@@ -30,7 +26,7 @@ const thresholdValidation = (
   const inputSecondThreshold =
     input?.secondThresholdForExpiringItems?.[0]?.value;
 
-  // Second threshold should not exceed 30 days
+  // Thresholds should not exceed 30 days
   if (
     (inputSecondThreshold && inputSecondThreshold > 30) ||
     (inputFirstThreshold && inputFirstThreshold > 30)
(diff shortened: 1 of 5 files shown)


client/packages/system/src/Manage/Preferences/api/utils.ts
```diff
@@ -1,63 +1,37 @@
...
   if (
-    firstThreshold != null &&
-    secondThreshold != null &&
-    secondThreshold !== 0 &&
-    secondThreshold < firstThreshold
+    (inputSecondThreshold && inputSecondThreshold > 30) ||
+    (inputFirstThreshold && inputFirstThreshold > 30)
   ) {
-    warning(t('label.second-threshold-is-less-than-first-threshold'))();
+    warning(t('label.threshold-exceeds-days'))();
```
GeronimoJohn: Halo, apologies, I know its still this is still in draft. But I peeked and noticed this. 

I left a comment on the issue: https://github.com/msupply-foundation/open-msupply/issues/9785#issuecomment-3549512831. But basically first threshold should be lower than the second

PR #9808, opened 2025-11-19
#9700 Custom store colour (footer)

### LINKED ISSUE #9700: Differentiate the current store

When a user is required to log into various stores on the same site,  well as perform normal stock management functions in their own stores.

To make it clear which store you are logged into, have a mechanism such as a store property to define a unique colour for each store. If a colour is defined then set the background colour of the site footer.

If the colour is too dark we will need a complimentary colour for the text in the footer: would recommend storing the colour as hex and using the mui function:  theme.palette.getContrastText(colour) to get the text colour.

We'll need a mechanism/place to edit these colours as well as storage, perhaps name properties or a new oms_field for store?

CarlosNZ:
Fixes #9700

# 👩🏻‍💻 What does this PR do?

- Add "Store Custom Colour" as a store pref
- Display the Footer using that colour as the background

Screenshot 2025-11-19 at 12 57 46 PM
Screenshot 2025-11-19 at 12 55 38 PM
Screenshot 2025-11-19 at 12 55 07 PM
Screenshot 2025-11-19 at 12 54 52 PM

This is the first "string" store pref, so double-check the Rust implementation is solid.

# 🧪 Testing

- [ ] When no custom colour is set, footer displays normally (orange for central server, grey for standard)
- [ ] Set custom colour, can be #nnn, #nnnnnn, rgb(), rgba(), hsl(), hsla(), color() -- try a range of different ones
- [ ] Footer bar updates immediately with the new colour
- [ ] Text colour is appropriately contrasted with the background colour (i.e. dark for light backgrounds and vice versa)
- [ ] Deleting colour value from pref restores default footer colour
- [ ] Invalid colour values (e.g. misformated colour strings) show error in console but otherewise fail silently (TO-DO: would be good to validate directly in the prefs input, but that would require a custom component just for this pref)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Name/ListView/Stores/EditStorePreferences.tsx
+++ b/client/packages/system/src/Name/ListView/Stores/EditStorePreferences.tsx
@@ -33,6 +33,7 @@ export const EditStorePreferences = ({
 
     return (
       <EditPreference
+        key={pref.key}
         disabled={
           !isCentralServer || !userHasPermission(UserPermission.EditCentralData)
         }


client/packages/common/src/ui/components/portals/AppFooter/AppFooter.tsx
```diff
@@ -29,14 +34,31 @@ export const AppFooter: FC = () => {
 
   const hideFooter = fullScreen || keyboardIsOpen;
 
+  // Colours for the Footer bar, if specified in Store prefs
+  let customColour: string | undefined;
+  let textColour: string | undefined;
+  if (storeCustomColour) {
+    // Try/catch essentially validates the colour string -- if it's invalid, the
```
jmbrunskill: Thanks great having comment like this! Better in the code than just in the code review..

jmbrunskill (approved):
Works great for me. Would be very nice to allow the store to edit their own colour.
It's great if you're on central and immediately see the colour change.

Rust side looks fine to me.

PR #9840, opened 2025-11-21
9755 Allow processing master list and editing plugin data in processor plugin

### LINKED ISSUE #9755: Is essential master list field integration through processor

Currently we have a very specific plugin for integration one field from master list, plugin interfaces should be generic.

Deal with this change request affecting the field name: https://github.com/msupply-foundation/msupply/issues/17425.

Plugin should be generic, we can do this integration with processor. Would need to add master list to changelogs, should ideally insert all master lists in changelogs when doing this. 

Processor would need to listen to changelogs on master list and then access sync_buffer, extracting relevant field and populating pluginData. Plugin data update needs to happen with sync in mind so we may need to expose plugin data repository (like sync message repository) to plugins

No serialisation of is_essential in translators.

andreievg:
Fixes #9755 

See plugin PR here

# 👩🏻‍💻 What does this PR do?

* Add master list to changelog
* Remove master list delete (this was always meant to be soft delete)
* Add update pluging data to use repository exposed to plugin
*  Remove existing sync interface for plugin and master list

I considered adding another method to upsert plugin data rather then using repository (because get plugin data is separate method). I think alternatively we should expose get plugin data in use repository rather then having it as separate method, this would break some things so keep for now, until we have plugin versioning.

# 🧪 Testing

See plugin issue

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/db_diesel/master_list_row.rs
+++ b/server/repository/src/db_diesel/master_list_row.rs
@@ -82,7 +82,7 @@ impl<'a> MasterListRowRepository<'a> {
 impl Upsert for MasterListRow {
     fn upsert(&self, con: &StorageConnection) -> Result<Option<i64>, RepositoryError> {
         let change_log_id = MasterListRowRepository::new(con).upsert_one(self)?;
-        Ok(Some(change_log_id)) // Table not in Changelog
+        Ok(Some(change_log_id))
     }
 
     // Test only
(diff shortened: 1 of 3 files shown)


andreievg (commented):
I did a self review, pretty thorough one, will merge as we need this for 2.13 testing tonight/tomorrow

PR #9846, opened 2025-11-21
9812 ccei / gaps] cannot find store to transfer cce when there are many stores

### LINKED ISSUE #9812: [CCEI / GAPS] Cannot find store to transfer CCE when there are many stores
## What went wrong? 😲

Client has 1000 stores, the search function in the Equipment page cannot handle this many and so the user cannot locate the store they need to make the transfer of the CCE

Image

But store exists: 

Image

## Expected behaviour 🤔

All stores should be retrievable in search function

## How to Reproduce 🔨

1. Have 1000 stores
2. Attempt to search for a store in the 2nd 500 results

## Your environment 🌱

- Open mSupply Version: 2.12.00

roxy-dao:
Fixes #9812

# 👩🏻‍💻 What does this PR do?
Have kept the infinite query for now... but wondering if we should implement something similar to patients for the stores as well? Was there a reason why we want paginated autocomplete? @mark-prins 

This PR keeps the AutoCompleteWithPagination in place so allows users to scroll through large store lists, or if they start a search then that api will be refetched since it's possible that the search result might not exist in the 'cached result'. Would like to change this to do something similar to what patients is doing if there isn't a valid reason to keep AutocompleteWithPagination =/

# 🧪 Testing

- [ ] Have a datafile with thousands of stores
- [ ] Create an asset
- [ ] Search for a store at the bottom of your list
- [ ] Store should show up as a result

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/system/src/Store/api/hooks/document/useStores.ts
+++ b/client/packages/system/src/Store/api/hooks/document/useStores.ts
@@ -1,5 +1,5 @@
 import {
-  FilterController,
+  FilterBy,
   useInfiniteQuery,
   useQuery,
   useQueryParamsStore,
@@ -21,7 +21,7 @@ export const useStores = () => {
 };
 
 interface useStoresProps {
-  filter?: FilterController;
+  filter: FilterBy | null;
   rowsPerPage: number;
 }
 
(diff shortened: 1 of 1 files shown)


GeronimoJohn (commented):
Was about to approve because the code looks good and functionality still works. But the environment I tested it in didn't have much store. Wondering if you have a datafile that I can use @roxy-dao. 

That said, I am interested in the solution that was implemented for patients (haven't had much time to look into patients). Am coming from a place where if something works already and if clean can just reapply?

PR #9859, opened 2025-11-25
#9841 Fix bug when changing skipped dose from Not Given to Given

### LINKED ISSUE #9841: Dose Management: when a dose is issued with skipping previous ones, updating it to 'Not Given' and then to 'Given' again leads to an error
## What went wrong? 😲

I enrol a patient from scratch (no vaccines have been issued prior to testing). I have two vaccine courses (S = skipping, NS = no skipping), and three doses for each course (so 6 doses in total). 

I jump straight into S3 (which is a third dose in a course that allows skipping). 

- Issue it - works ✅ 
- Ungive it - also works ✅ 
- Try to issue is again - leads to an error 🚫 

Image

Image

## Expected behaviour 🤔

Shouldn't be an error

## How to Reproduce 🔨

1. Have an immunization program set up with a vaccine course that allows skipping doses, and three doses for that course
2. Enrol a patient (fresh) into that program
3. Go to that patient -> create a new encounter -> you should be able to administer vaccines to them now
4. Click on Dose 3 -> skipping the first two. You are able to issue it ✅ 
5. Click on the same dose -> ungive it -> you are able to ungive it ✅ 
6. Click on that same dose again -> issue -> running into 'Oops something gone wrong' error 🚫 

## Your environment 🌱

- Open mSupply Version: `2.13.00-RC-11201551`
- Legacy mSupply Central Server Version: `V8.06.00`

- Database type: `PostgreSQL`

CarlosNZ:
Fixes #9841

# 👩🏻‍💻 What does this PR do?

A bit of an edge case, but the back-end checks were only letting a vaccination be changed from "Not Given" to "Given" when no doses had been skipped. We missed this when changing the logic recently, as now it's possible to skip a dose and therefore this check doesn't need to be quite as strict.

Have relaxed the validation logic to only consider this check when the course is NOT skippable (i.e. as it was previously).

# 🧪 Testing

- [ ] Repeat steps described in the issue -- confirm can mark as "Given" with no error.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/rnr_form/tests/insert.rs
+++ b/server/service/src/rnr_form/tests/insert.rs
@@ -341,10 +341,10 @@ mod insert {
                 .program_requisition_settings()
                 .full_master_list(),
             MockData {
-                // make supplier store B visible in store A
+                // make supplier store C visible in store A
                 name_store_joins: vec![NameStoreJoinRow {
-                    id: String::from("name_store_a_join_b"),
-                    name_link_id: String::from("name_store_b"),
+                    id: String::from("name_store_a_join_c"),
+                    name_link_id: String::from("name_store_c"),
                     store_id: String::from("store_a"),
                     name_is_customer: false,
                     name_is_supplier: true,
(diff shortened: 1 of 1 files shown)


jmbrunskill:
@CarlosNZ @mark-prins  is this the desired logic?

If a previous dose is marked as not given, should we always allow the new dose to be given? Regardless of skip status?
I thought we decided the skipping logic was just a short cut, and you could effectively just skip doses by manually setting them to not given.

PR #9866, opened 2025-11-25
9441 add a supplier return form

### LINKED ISSUE #9441: Print a Supplier Return Form (in Cote d'Ivoire: "Bordereau de retour")
## What went wrong? 😲

When trying to print the "bordereau de retour" (return slip), it says No Forms Available

Image

## Expected behaviour 🤔
The Bordereau de retour should be similar to the one in OG. Note that the Reference (on the print out and on screen) should say "Retour créé depuis la facture fournisseur XXX" with XXX being the number of the original outbound shipment for that particular item.

This is on the CIV demo server.

See below the form in OG

Image

adrianwb:
Fixes #9441

# 👩🏻‍💻 What does this PR do?

There was no supplier return form => added one based on the outbound shipments one

This is a clone of the landscape outbound shipment form, but I had to patch up the date and price formatting => should I back-port the same changes to the outbound shipment form(s) too?

# 🧪 Testing

- [ ] Server with client running this PR
- [ ] `cargo run --bin remote_server_cli build-reports --path ../standard_forms/supplier-return/latest` and then `cargo run --bin remote_server_cli upsert-reports --path ../standard_forms/supplier-return/latest/generated/reports.json --overwrite`
- [ ] Open/create a supplier return with some lines and print/export the report; check layout and formatting
- [ ] Open an outbound shipment; check that "Supplier return" is not in the report choices

# 📃 Documentation

  1. Supplier return form is new

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/Returns/CustomerDetailView/AppBarButtons.tsx
```diff
@@ -37,8 +37,7 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
           onClick={() => onAddItem()}
         />
         <ReportSelector
-          context={ReportContext.InboundShipment}
-          subContext="CustomerReturn"
+          context={ReportContext.InboundReturn}
```
roxy-dao: Would have this as CustomerReturn, and below as SupplierReturn, so these both need to be added to ContextType in backend... and would probably do a migration to drop InboundReturn and OutboundReturn 👀

standard_forms/supplier-return/latest/report-manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "is_custom": false,
+  "version": "2.6.5",
+  "code": "supplier-return",
+  "context": "OUTBOUND_SHIPMENT",
+  "sub_context": "SupplierReturn",
```
roxy-dao: Would change version to match app version and context should be supplier return (: else it shows up in the Outbound selection~

```suggestion
  "version": "2.13.0",
  "code": "supplier-return",
  "context": "SUPPLIER_RETURN",
```

standard_forms/supplier-return/latest/src/style.css
```diff
@@ -0,0 +1,116 @@
...
+  margin-left: 8px;
+  border: 0.1px solid black;
+}
+
+.pageNumber::after {
+  font-family: Tahoma;
+  font-size: 8pt;
+  counter-increment: page;
+  content: "Page " counter(page);
+}
```
roxy-dao: Not sure about keeping this in right now since it currently doesn't work 👀

standard_forms/supplier-return/latest/src/template.html
```diff
@@ -0,0 +1,60 @@
...
+            <td class="expiry" style="width: 80px;">{{self::dateOrNA(datetime=line.expiryDate)}}</td>
+            <td class="sell_price" style="width: 50px;">{{self::formatCurrency(amount=line.sellPricePerPack)}}</td>
+            <td class="cost_price" style="width: 50px;">{{self::formatCurrency(amount=line.costPricePerPack)}}</td>
+            <td class="total_extension" style="width: 60px;">{{self::formatCurrency(amount=line.numberOfPacks * line.sellPricePerPack)}}</td>
+        </tr>
+        {%- endfor %}
+    </table>
+</body>
+
+<table class="body_total_section" cellpadding="2" cellspacing="0">
```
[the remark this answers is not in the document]
roxy-dao: In a return we want to know how much is cost us to return right? Cause we're returning it to the supplier? I think the total cost is just getting sell price but I could be wrong....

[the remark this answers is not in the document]
roxy-dao: Line total is 12 bucks but it shows 17.64 which I'm assuming is the sell price (:
Screenshot 2025-12-02 at 10 05 33 AM

roxy-dao (changes requested):
Thanks Adrian just a request to fix the total ;x

roxy-dao (approved):
Thanks Adrian~ LGTM

andreievg:
> as you can often do the same thing in either?

@adrianwb we've run into problems with tera when doing data manipulation, it's too foreign and cumbersome for that. There is also ability to query again (sql, graphql etc..) in JS too, which can't be done in tera.

> FYI I've farmed this out into a new refactor issue (https://github.com/msupply-foundation/open-msupply/issues/9920) now so I'll link to this from there

Oki, great, thanks, i'll try to have a looks today

PR #9873, opened 2025-11-27
9697 show indicative unit price in requisitions

### LINKED ISSUE #9697: Show an indicative Unit price on internal orders and requisitions

The Unit price is a fixed national rate per item which is  identical at every supply level.

This price should be displayed with the label Prix unitaire / Unit price.

The [national](https://docs.msupply.org.nz/items:pricing?s[]=price#default_price_list) price list could work for this.

- [x] Add store pref and gql
- [x] Add `requisition_line.price_per_unit` and gql
- [x] Pivot to global pref
- [x] Apply in requisition transfer processor
- [x] Handle in new internal orders 
- [x] Show in internal orders UI
- [x] Show in distribution/requisition UI

Chris-Petty:
Fixes #9697 

# 👩🏻‍💻 What does this PR do?

- Adds `requisition_line.price_per_unit`
- Displays it in internal orders image
- And response requisitions 
image

I wrote tests for the the create_response_requisition processor, but not for the service layer changes. No particular reason other than short of time :S.

I haven't added it to the modal editing a line...

# 🧪 Testing

- [ ] Have 2 OMS stores on different sites
- [ ] Global pref off
- [ ] Make an internal order from one store to the other and finalise
- [ ] No prices shown in either
- [ ] Turn global pref on - prices are 0.0 for both requisitions
- [ ] Make sure there are prices in the "default price list" master list 
- [ ] Make another internal order form one store to the other, prices are shown for the items that had prices in the master list. The "Indicative Price" column is requested * unit price 
- [ ] Add some items that don't have a price in the master list
- [ ] Add the price in the master list, the items added **don't** get a price update
- [ ] Keep the priceless lines and finalise
- [ ] Update the master list price for one of the lines with a price
- [ ] Finalise 
- [ ] The response requisition uses prices from original requisition if available. If there isn't a price on the line, the response uses the price from the master list
- [ ] Create a manual requisition
- [ ] Add a line with some requested quantity
- [ ] Indicative price column shown and calculated correctly

# 📃 Documentation

  1. Global prefs and requisitions
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/sync/translations/requisition_line.rs
```diff
@@ -16,10 +16,9 @@ use super::{PullTranslateResult, PushTranslateResult, SyncTranslation};
 
 #[derive(Deserialize, Serialize, PartialEq)]
 pub struct RequisitionLineOmsFields {
-    #[serde(deserialize_with = "empty_str_as_option_string")]
-    pub rnr_form_line_id: Option<String>,
-    #[serde(deserialize_with = "empty_str_as_option")]
```
Chris-Petty: I'm quite certain these annotations are not needed for properties of objects coming from OG, just cargo cult from the legacy record translation. OG's more primitive field types will set null to empty values (e.g. null String -> "", null number -> 0, bool -> false...). This is not true of object field's content other than null Object -> empty Object `{}`. Beyond that there is no schema.

server/service/src/requisition_line/response_requisition_line/insert/generate.rs
Chris-Petty: Used in literally one place and just inlining it seemed easier than changing the function signature!

server/repository/build.rs
Chris-Petty: Stopping errors from Glob trying to process cargo build temp files coming in and out of existence

server/repository/src/mock/mod.rs
```diff
@@ -596,7 +596,7 @@ impl MockDataInserts {
         self
     }
 
-    pub fn full_master_list(mut self) -> Self {
+    pub fn full_master_lists(mut self) -> Self {
```
Chris-Petty: The rest are plural, the property it modifies is plural... sorry it touched many files 🫨

server/repository/src/migrations/v2_14_00/requisition_line_add_price_per_unit.rs
```diff
@@ -0,0 +1,20 @@
...
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "requisition_line_add_price_per_unit"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                ALTER TABLE requisition_line ADD COLUMN price_per_unit {DOUBLE};
```
Chris-Petty: Made this nullable cause initially this was a store pref and seemed that knowing whether the ordering party in a requisition transfer had the pref on and populated the values was useful, i.e. use the national list instead. However changed it to global pref anyway.

Still optional 🤔. Figure if we leave it and need it fine, need nullable later it's easy to switch. But switching from not null to nullable and having ambiguity of 0.0 being intentional or not might be hard. I could swing either way on this though still haha.

As the sync layer uses `oms_fields` object nullable is actually preserved on sync through OG unlike normal fields though! Tangentially the `master_list_line.price_per_unit` field is Optional, but OG is likely to always send 0.0 regardless. Maybe worth leaving as OMS Full Central could benefit?

roxy-dao:
Haven't gone through it yet, but is this compatible with the indicative price shown for Congo?

client/packages/requisitions/src/RequestRequisition/DetailView/columns.tsx
```diff
@@ -216,9 +227,11 @@ export const useRequestColumns = () => {
       warningForExcessRequest,
       showExtraColumns,
       usesRemoteAuthorisation,
+      showIndicativeUnitPriceInRequisitions,
       maxMonthsOfStock,
       plugins.requestRequisitionLine?.tableColumn,
       errors,
+      t,
```
jmbrunskill: @CarlosNZ told me we're not adding this in  as a dependancy on our new tables as it should available. Not sure the downside though. @CarlosNZ care to comment?

server/repository/src/migrations/v2_14_00/requisition_line_add_price_per_unit.rs
jmbrunskill: I think nullable seems better to me, it's annoying in the rust code but having a distinction between 0.0 and null feels relevant to some degree.

server/repository/src/mock/full_master_list.rs
```diff
@@ -145,3 +146,34 @@ pub fn mock_master_list_master_list_line_filter_test() -> FullMockMasterList {
         ],
     }
 }
+
+pub fn mock_master_list_default_price_list() -> FullMockMasterList {
```
jmbrunskill: I checked, looks like we (I) just created a master list in the test case for item pricing (default_price_list).
This test case doesn't include master lists just `MockDataInserts::none().items().names().stores(),`

I think we have undefined behaviour if we have more than one `is_default_price_list` right? It will just pick the first one it finds.

All that is to say I think we're ok to add this but wanted to think about implications.
Maybe should even have a comment here saying there should only ever be one?

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -350,13 +337,57 @@ fn generate_response_requisition_lines(
...
-                approval_comment: None,
+                price_per_unit,
             },
-        )
-        .collect();
+        item_row: ItemRow { id: item_id, .. },
+        requisition_row: _,
+    } in request_lines
+    {
+        let price_per_unit = {
```
jmbrunskill: Why not call `get_pricing_for_item` here? I was thinking that's where we'd put the variation for different setups and possibly where we'd add a plugin interface. This approach saves a db lookup but looking at the code now, I'm wondering why we don't just always populate the price, the  pref just controls if you see it?

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -350,13 +337,57 @@ fn generate_response_requisition_lines(
...
+            {
+                // It's tempting to use a hashmap lookup here to avoid potentially making a query for each line. However, if the global pref is on
+                // then all sites should be sending through prices anyway (i.e. price_per_unit/new_price_per_unit is Some value!) and making the
+                // Hashmap would be a waste. What this then covers is when you've just turned the global preference on and there are prior requisitions
+                // that don't have a value. We could however make the hashmap on demand though, might still be better than this!
+                // Note transfers from mobile/OG remote sites will have request_line.price_per_unit = Some(0.0) regardless of the pref...
+                new_price_per_unit = get_item_price_per_unit(
+                    connection,
+                    &item_id,
+                    &default_price_list.as_ref().unwrap().id,
```
jmbrunskill: Not a fan of the unwrap, although it is safe in this case...

How about:
```
        let price_per_unit = {
            let mut new_price_per_unit = price_per_unit;
            if populate_price_per_unit && new_price_per_unit.is_none() {
                match &default_price_list {
                    Some(pricelist) => {
                        // It's tempting to use a hashmap lookup here to avoid potentially making a query for each line. However, if the global pref is on
                        // then all sites should be sending through prices anyway (i.e. price_per_unit/new_price_per_unit is Some value!) and making the
                        // Hashmap would be a waste. What this then covers is when you've just turned the global preference on and there are prior requisitions
                        // that don't have a value. We could however make the hashmap on demand though, might still be better than this!
                        // Note transfers from mobile/OG remote sites will have request_line.price_per_unit = Some(0.0) regardless of the pref...
                        new_price_per_unit =
                            get_item_price_per_unit(connection, &item_id, pricelist.id)?
                    }
                    None => {}
                };
            }
            new_price_per_unit
        };
```

client/packages/requisitions/src/RequestRequisition/DetailView/columns.tsx
Chris-Petty: IDK if other don't have linters running or ignore them but throughout our code base basic react lint rules complain about missing dependencies in memoization/callbacks not to mention useEffects. I went down the rabbit hole to check if t is actual generally a stable reference but it's a bit of typescript hell. Wish instead the type system actually made it clear, that'd be so sweet.

jmbrunskill (commented):
Looking good Chris.

I'm just adding a comment rather than approve or request changes, as I'd like to get a bit of production direction.

@mark-prins another one for you to comment on?

Currently the pref created by Chris is very specific to the national price list approach.

Image

I'm wondering if we want to be more generic, e.g just `show indicative price` on Internal Orders?

For example we have a item_pricing structure that currently looks up National Price list, but can apply a customer discount (for comoros).

I was thinking of this as generic functionality that we could extend as needed. We also have a default sell price in mSupply. would it be useful if that was an optional way to populate this field too?

Image

I think we've met the immediate requirements but wondering about how generic we should be thinking about this feature...

server/repository/build.rs
```diff
@@ -8,14 +8,26 @@ use crate::constants::{TEMPLATE_MARKER_FILE, TEST_OUTPUT_DIR};
...
-        match entry {
-            Ok(path) => {
-                fs::File::create(Path::new(&path).join(TEMPLATE_MARKER_FILE)).unwrap();
+    // Only search in specific directories, in particular avoiding ../target/ which has temporary files that may cause glob to error
+    let patterns = [
+        format!("../{TEST_OUTPUT_DIR}"),
+        format!("../repository/**/{TEST_OUTPUT_DIR}"),
+        format!("../cli/**/{TEST_OUTPUT_DIR}"),
+        format!("../graphql/**/{TEST_OUTPUT_DIR}"),
+        format!("../service/**/{TEST_OUTPUT_DIR}"),
```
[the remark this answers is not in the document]
Chris-Petty: Yea was tempted to just silence it but worried the same. Another option is to use a different lib to do the directory traversal ourselves and just ignore `target/` directory (until we for some reason name a directory that ourselves haha).

Another - just put test_output from through our tests all in one folder at root rather than all over the code base. Then just have to put markers in that one folder. TBH I'd probably find that easier for finding the particular test DB for inspection anyway haha. I guess at the moment we have rather duplicate test template DBs around the place too, so this would reduce that duplication (not that it particularly matters.

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -350,13 +337,57 @@ fn generate_response_requisition_lines(
...
-                supply_quantity: 0.0,
-                approved_quantity: 0.0,
-                approval_comment: None,
+                price_per_unit,
             },
-        )
-        .collect();
+        item_row: ItemRow { id: item_id, .. },
+        requisition_row: _,
+    } in request_lines
```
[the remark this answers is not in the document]
Chris-Petty: Heh you're right, I think earlier on I had the pricing query in the loop and handling Results in iterator function closures is a pain. Gonna leave it!

server/service/src/sync/translations/requisition_line.rs
```diff
@@ -87,7 +86,6 @@ pub struct LegacyRequisitionLineRow {
     pub oms_fields: Option<RequisitionLineOmsFields>,
 }
 // Needs to be added to all_translators()
-#[deny(dead_code)]
```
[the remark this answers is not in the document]
Chris-Petty: I've always wondered why it was there, I removed it 'cause it's bothered me for a long time haha. I also misinterpreted that `deny` silences the dead_code warning but it makes the build error - that's a `allow` lol. We don't tend to accept warnings in the build anyway though...

server/repository/build.rs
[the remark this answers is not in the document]
jmbrunskill: I like the central test_output folder idea actually but lets leave that for a future refactor!

server/service/src/processors/transfer/requisition/create_response_requisition.rs
```diff
@@ -296,67 +299,97 @@ fn generate_response_requisition(
...
+                },
+            item_row: ItemRow { id: item_id, .. },
+            requisition_row: _,
+        },
+    ) in request_lines.into_iter().enumerate()
+    {
+        let price_per_unit = {
+            if price_per_unit.is_none() {
+                if let Some(price_list) = &price_list {
+                    price_list[i].calculated_price_per_unit
```
Chris-Petty: Should be safe to index in like this with the current implementation of `get_pricing_for_items`. If we optimise `get_pricing_for_items` we'll may want to be careful that the order is maintained.

jmbrunskill (approved):
Thanks! I think the hashmap is better!

Issue #9715, opened 2025-11-07
Test and explanation for days out of stock dynamic diesel query


andreievg:
## Is your feature request related to a problem? Please describe 👀

Days out of stock query done with dynamic diesel is complex and foreign, lack validation

## Describe the solution you'd like 🎁

Add tests, explanation and link to dynamic sql tutorial (check original issue for explanation)

Bonus to remove if let Some in the DaysOutOfStockRepository.query

aimee-mcneil-melville:
A PR has been made with some improvements. May need a part 2 or another follow up

Some questions still open:

1. Getting the correct date (local vs UTC time) **Affects AMC and DOS calculations**

date_now() returns UTC time - so could be yesterday, today, or tomorrow depending on time of day and local timezone

We want the item stats filter to use the end of day current local date (if no period end supplied). Records are stored in UTC. So if the user is in NZT eg 28/11 with a +12hr difference, the dates provided in the item_stats filter should be 27/11 12:00 UTC - 28/11 12:00 UTC

Usually this is handled client side with the filters, I made a start at implementing a similar approach to (item counts)[https://github.com/msupply-foundation/open-msupply/blob/babdf29e95f3cf28514c7e9ec6997beb0e026df2/server/service/src/dashboard/invoice_count.rs] with passing through a timezone, but quickly expanded into passing this through many areas so I'm not sure is the best approach.

Other options to look at: 
- Local storage
- Store timezone on ServiceContext


2. DOS query doesn't determine full days out of stock - it determines if the day ends on a balance of 0. Could amend the query to check if start and end balance is 0 (would still allow an edge case of stock in and out on the same day). Is the desired outcome of this query days ending on 0 or full days out of stock?

PR #9875, opened 2025-11-27
9715 Test dos dynamic query

aimee-mcneil-melville:
Partial #9715

# 👩🏻‍💻 What does this PR do?

Add tests to the days out of stock repository to check the number of days returned

Some improvements in the repository query

Add test_helpers to repository layer and moved make_movements there so it can be accessed. This is where a bunch of file changes come from sorry!

TBC expected behaviour here but have some findings from testing:

1. date_now() returns UTC time - so could be yesterday, today, or tomorrow depending on time of day and local timezone
- If amc or dos are queried at NZT 27/11 10am the date_now() is 26/11 - yesterday
- If amc or dos are queried at NZT 27/11 3pm the date_now() is 27/11 - current

    Update: we do want this to be localised to the end of current date (if no period end supplied). Records are stored in UTC. So if the user is in NZT eg 28/11 with a +12hr difference, the UTC query should be 27/11 12pm - 28/11 12pm
    
    Usually this is handled client side with the filters, I made a start at implementing a similar approach to (item counts)[https://github.com/msupply-foundation/open-msupply/blob/babdf29e95f3cf28514c7e9ec6997beb0e026df2/server/service/src/dashboard/invoice_count.rs] with passing through a timezone, but quickly expanded into many areas so I'm not sure is the best approach.
    
    Will add comments to the issue about next steps! 

2. I removed the endDate + 1 from the filter in dos and amc (kept it in the start date calculation).  I'd expect the amc/dos calulations to want the end date to be 'today' or maybe EOD yesterday - UTC might not be the best for determining that but would need to be converted to UTC before the queries to get the correct records from database

3. DOS query doesn't determine full days out of stock - it determines if the day ends on a balance of 0. Could amend the query to check if start and end balance is 0 (would still allow an edge case of stock in and out on the same day). Is the desired outcome of this query days ending on 0 or full days out of stock?

# 🧪 Testing

- [ ] Tests pass
- [ ] Dates make sense with stock movements

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/days_out_of_stock.rs
```diff
@@ -48,64 +48,265 @@ impl<'a> DaysOutOfStockRepository<'a> {
...
+        else {
+            return Ok(vec![]);
+        };
+
+        // Convert start_date and end_date to NaiveDateTime
+        // TODO: Can the query be changed to NaiveDate instead?
+        let start = start_date.and_time(chrono::NaiveTime::from_hms_opt(0, 0, 0).unwrap());
+        let end = end_date.and_time(chrono::NaiveTime::from_hms_opt(23, 59, 59).unwrap());
+
+        let dos_query = Dos {
```
aimee-mcneil-melville: Some of this is indentation moving... hard to differentiate that in here though 😢

server/service/src/item_stats.rs
```diff
@@ -551,11 +552,8 @@ mod test {
...
 
-        // With DOS adjustment: 30 days / (30 - 16 dos) = 30/14 = 2.142857
-        // AMC = 6.0 * 2.142857 = 12.857142857142858
-        assert_eq!(
-            item_stats[0].average_monthly_consumption,
-            12.857142857142858
-        );
+        // With DOS adjustment: 30 days / (30 - 15 dos) = 30/15 = 2
+        // AMC = 6.0 * 2 = 12
+        assert_eq!(item_stats[0].average_monthly_consumption, 12.0);
```
aimee-mcneil-melville: Another change as a result of removing the +1 from end date

server/service/src/item_stats.rs
```diff
@@ -76,7 +76,7 @@ pub fn get_item_stats(
     let filter = ConsumptionFilter {
         item_id: Some(EqualFilter::equal_any(item_ids.clone())),
         store_id: Some(EqualFilter::equal_to(store_id)),
-        date: Some(DateFilter::date_range(&start_date, &offset_end_date)),
+        date: Some(DateFilter::date_range(&start_date, &end_date)),
```
aimee-mcneil-melville: Using just end date (today or end of period) instead of end date + 1

jmbrunskill:
> DOS query doesn't determine full days out of stock - it determines if the day ends on a balance of 0.

Haven't looked at the PR, but I think normally we do if the day ends out of stock. CIV want a whole day out of stock though, which is currently in their plugin.

> Timezones 

I agree we should probably think about timezones somehow. UTC midnight probably works pretty well most of the time, however it might be confusing to users if the `end of day` doesn't align with their own view of things. We're lucky that (most?) african countries are pretty close to UTC time :)

In a previous role, we had a custom start of day time, we started our day at 7am. Lots of systems didn't know how to deal with that. Personally I think this is ok to leave for now until we have a customer requirement?

> I removed the endDate + 1 from the filter in dos and amc (kept it in the start date calculation). I'd expect the amc/dos calulations to want the end date to be 'today' or maybe EOD yesterday - UTC might not be the best for determining that but would need to be converted to UTC before the queries to get the correct records from database

I'm not aware of why this was added, perhaps to handle the timezone problem, should have been a comment explaining it. I'd be hesitant to remove this, hopefully @andreievg has some more context on it...

PR #9900, opened 2025-12-01
#9891 Store can't choose the correct destination customer when creating an internal from a requisition if the Store category is set to "Customer" only

### LINKED ISSUE #9891: [Congo] Store can't choose the correct destination customer when creating an internal from a requisition if the Store category is set to "Customer" only
Andrei: Update, @rukmal-msupply found this to happen when store.name is not a supplier but just customer, so it sounds like filter needs updating

Congo demo server 2.13RC
HR TALANGAI PHARMACIE has got the "can create internal order from a requisition" preference
Received a requisition from  HR TALANGAI-CTA
Can create an internal order from the requisition but cannot select HR TALANGAI-CTA as the destination customer

Bad configuration? Or is it because their the same supply level?

Image

roxy-dao:
Fixes #9891

# 👩🏻‍💻 What does this PR do?
Have separate invalidation keys for internal supplier & customer api calls since they both call name under the hood.

# 🧪 Testing

- [ ] Have a customer that is customer only (and not a supplier)
- [ ] Have the store preference `Select destination store...` on
- [ ] Go to an Internal Order
- [ ] Click on `Destination Customer`, should be able to see the customer only name

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
@@ -11,6 +11,7 @@ import {
   Autocomplete,
   useConfirmationModal,
   usePreferences,
+  NameNodeType,
 } from '@openmsupply-client/common';
 import {
   CustomerSearchInput,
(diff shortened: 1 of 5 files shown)


andreievg:
Thanks for the change @roxy-dao and for review @GeronimoJohn, merging

PR #9995, opened 2025-12-09
9990 update sync translation and fields for vaccination

### LINKED ISSUE #9990: translate and sync vaccination fields to og msupply

We are working on DHIS2 vaccine aggregated data and we need the data the vaccines data

OG already have some of the vaccination fields, but is missing the ff fields:
- program_enrolment_id
- encounter_id
- given_store_id (what is this field and what difference between the store_id?)

as above. 

### Additional context 💌

og msupply issue https://github.com/msupply-foundation/msupply/issues/17508

### Moneyworks Jobcode 🧰

[DJIGINTN] Djibouti GAVI World Bank Countrywide Expansion Phase 2: Customisations - Integration work (SAGE & DHIS2)

raviSussol:
Fixes #9990 

# 👩🏻‍💻 What does this PR do?
- Added 3 missing fields of vaccination table for OG mSupply
- Added v2.14.0 migration to resync all vaccination records to mSupply

# 🧪 Testing

- [ ] Take a Djibouti and login to the dispensary store.
- [ ] Create a new patient (or just open an existing patient)
- [ ] Goto a Programs tab - add a vaccination program
- [ ] Then add an encounter for that program
- [ ] Give some vaccination and then marked them as visited, this will create vaccination card for that patient
- [ ] Now sync it to the OG mSupply
- [ ] In the mSupply, there's no such interface to show these data. So you have to open the record browser (from the Sussol admin window which you can open from the about mSupply window and click Authorisation).
- [ ] Select `om_vaccination` table from the dropdown and then click query to query specific vaccination.
- [ ] Similarly, select `om_encounter` table from the dropdown to check all encounters for that patient which should be synced from the OMS.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/v2_14_00/resync_existing_vaccination_records.rs
```diff
@@ -0,0 +1,21 @@
...
+            connection,
+            r#"
+                INSERT INTO changelog (table_name, record_id, row_action, name_link_id)
+                SELECT 'vaccination', id, 'UPSERT', patient_link_id FROM vaccination;
+            "#
+        )?;
+
+        Ok(())
+    }
+}
```
Chris-Petty: @andreievg something for v7 (or sooner?), it'd be good to be able to tell we're on a central server in migrations to avoid remote sites updating change log and resyncing if they don't need to. James explained to me that we just have an Arc Mutex that contains the central server state and only gets that state after first sync authentication, so a bit hard right now. I figure just storing in keystore would be pretty straightforward? I'd generally keep accessing the in memory state though as the state doesn't really change and don't want to make a DB call every time.

Also this would also resync all vaccination records from central to all remote sites (where patient is visible) correct? Not just from OMS central to OG central? Not to mention all remote sites sending to central first given not being able to deduce and avoid that in this migration. Harebrained suggestion is to mingle queue and feed ideas here: add a `destination_site_id: Option<i32>` field that can be used to target a specific site in sync. I.e. on this migration on just the central server, we add change logs with specific intent of only site_id <whatever OG central is> getting them in push. OMS remotes pulling would ignore any records that have a `destination_site_id` that isn't equal to their own ID.

The main user of vaccination features has just half a dozen remote sites with decent connectivity afaik, so will probably cope with this migration at the moment.

server/repository/src/migrations/v2_14_00/resync_existing_vaccination_encounter_records.rs
```diff
@@ -0,0 +1,40 @@
...
+                INSERT INTO changelog (table_name, record_id, row_action, name_link_id, store_id)
+                SELECT 
+                    'encounter',
+                    id,
+                    'UPSERT',
+                    patient_link_id,
+                    store_id
+                FROM encounter;
+            "#
+        )?;
```
Chris-Petty: Thought (I presume thought is even lower request than nit? haha): Doesn't matter but you could do one `sql!` expression with both sql statements in it. Though maybe it's better this way as it'll be completely unambiguous which statement failed!

server/repository/src/migrations/v2_14_00/resync_existing_vaccination_records.rs
Chris-Petty: We could be crazy in this PR and nab something like the code from sync:

https://github.com/msupply-foundation/open-msupply/blob/26247eca2a482543e64608225205ca2c9d2c708e/server/service/src/sync/synchroniser.rs#L168-L170

But if the web request fails, just proceed and run regardless (I think this is the safest most predictable fall back). If the request succeeds then conditionally run based on being the central server or not.

Chris-Petty (approved):
Tested through and works well!

PR #10079, opened 2025-12-17
Stock to date report 

roxy-dao:
Fixes https://github.com/msupply-foundation/open-msupply-reports/issues/323

# 👩🏻‍💻 What does this PR do?
Translation label & allow stock ledger to be filtered by stock properties

# 🧪 Testing

- [ ] Testing through GraphQL

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -758,6 +758,7 @@
   "label.asset-notes": "Notes",
   "label.asset-number": "Asset Number",
   "label.asset-properties": "Asset Properties",
+  "label.at": "At",
   "label.atc-category": "ATC category",
   "label.auth-status": "Approval Status",
   "label.authorised-quantity": "Authorised Quantity",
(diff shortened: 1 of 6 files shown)


andreievg (changes requested):
Thanks Roxy,

I am not sure how relevant these changes will be, after this PR review:
https://github.com/msupply-foundation/open-msupply-reports/pull/324#pullrequestreview-3589985678

Although I think the filter is useful for future. Ohh man I wish I was reviewing the PR that added 'connection' to stock line create filtered query, really not fan of it (and there is easy and I think cleaner way to not need it)

We shouldn't be adding any translation for client specific reports (like the one that is dependant on this PR), those translations should just be hard coded in the report.

Happy to approve, without the translation. If you can refactor stock line create filtered query to not need connection passed on, would be even happier to merge : )

PR #10157, opened 2026-01-04
#10065 Fix auto-allocation bug for auto-created shipments

### LINKED ISSUE #10065: Expired stock are getting allocate to Auto generated internal orders
## What went wrong? 😲

When there are multiple expiry dates and multiple pack sizes for an item, they system automatically assign expired stocks to the internal orders.
i.e. I have stocks for paracetamol in different packings. i.e. TABs and Pack size 1, Pack size 10 etc. and these are having different expiry dates.

Expiry threshold dates
Image

Multiple packing /expiry

Image

Auto allocation

Image

## How to Reproduce 🔨

1. Create Stock line with different expiry dates, which are having less days than the expiry threshold dates
2. Create an internal order from different site
3. Create requisition, shipment

Image

## Your environment 🌱

- Open mSupply Version: 2.14.2

- Platform: Chrome

- Database type: PG

CarlosNZ:
Fixes #10065 

# 👩🏻‍💻 What does this PR do?

Turns out it wasn't actually a back-end issue, as the allocation logic for the new shipment lines is still in the front-end. It's just that when I did #10051 I missed this case as the initial values of the auto-created shipments are called separately to when you put the issue amounts in by the user.

I've refactored the `useAllocationContext` to take the expiry threshold value in its `initalise` method, so it's always available to all other methods in the context. Now *all* auto-allocation logic should have the correct value for this threshold.

# 🧪 Testing

- [ ] See steps in issue - confirm that stock within the expiry threshold is not auto-allocated (if pref is on)
- [ ] Confirm that changing the issue quanity yourself (in the UI) still behaves correctly (i.e. respects this threshold)
- [ ] Maybe check a prescription too, to confirm the change applies everywhere?

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
+++ b/client/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/Allocation.tsx
@@ -42,7 +42,12 @@ export const Allocation = ({
 }: AllocationProps) => {
   const t = useTranslation();
   const { format } = useFormatNumber();
-  const { manageVaccinesInDoses, sortByVvmStatusThenExpiry } = usePreferences();
+  const {
+    manageVaccinesInDoses,
+    sortByVvmStatusThenExpiry,
+    expiredStockPreventIssue,
+    expiredStockIssueThreshold,
+  } = usePreferences();
 
   const { initialise, item } = useAllocationContext(({ initialise, item }) => ({
     initialise,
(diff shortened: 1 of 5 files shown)


jmbrunskill (approved):
Nice, I think this a much cleaner...

PR #10161, opened 2026-01-05
9501 plugin versions

### LINKED ISSUE #9501: Plugin Version Management

OMS 2.12 adds new table implementation, so old plugins that add table columns will bork. However, if you update your plugin for 2.12, there is no version management for plugins, so older sites will get the new plugin version, and break 😢

Only solution is to upgrade all sites at once, but this is often not feasible.

Same kind of version handling that we use for reports, should be used for plugins.

Plugin manifests to be updated to have version number, similar to reports, if version number is missing, version number should be 2.16.

Plugins can be installed with whatever version.

When plugin is selected to be server (by backend) check versioning.

I did a rough start on this here: https://github.com/msupply-foundation/open-msupply/commit/ab2e223faafb52d383d9aa63d9814d8cdbb6fc36

## Testing scenario

Remote site 1 = 2.16

Remote site 2 = 2.15

Central = 2.17 (would not be released yet but can be mocked in code)

Central will have a new interface added or existing extended for backend and frontend plugin.

A backend and fronted plugin should exists that works with existing version, and a new version will be added for this plugin that uses the new interface.

Update central with new plugin bundles, make sure they work on remote and central

### Additional context 💌

For reports our versioning strategy is as follows:

* Use latest plugin version
* If plugin version is > current version, exclude this plugin from selection

Patches are not considered when evaluating "plugin version".

There is a use case where we might 'break' api/schema for reports, we deal with this use case by making sure new version exists and is selected correctly when upgrading, for standard reports it's automatic, however for custom reports we need coordination with support to make sure new reports are uploaded before upgrade.

On a similar note, when table API was broken, existing plugins were breaking a lot of demo/training/testing servers. 

So if we implement plugin version that is the same as report, we will have to coordinate mass plugin updates if future versions are broken. Therefore we will make plugin api backwards compatible, making sure to only add optional fields to the interface (with the table API, we should have added a new interface, and deprecate the old one, removing it overtime)

ria8651:
Fixes #9501

# 👩🏻‍💻 What does this PR do?

Adds a plugin version field in the frontend and backend database that gets populated from the plugin manifest. This is used to select plugins based on the app version with the latest compatible version being selected.

# 🧪 Testing

- [ ] Install lots of versions of the same plugin and check the correct version is selected.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/backend_plugin/plugin_provider.rs
+++ b/server/service/src/backend_plugin/plugin_provider.rs
@@ -97,7 +97,11 @@ impl PluginInstance {
         let plugins = PLUGINS.read().unwrap();
 
         for plugin in plugins.iter() {
-            log::info!("Plugin loaded: {} (version {})", plugin.instance.code, plugin.instance.version);
+            log::info!(
+                "Plugin loaded: {} (version {})",
+                plugin.instance.code,
+                plugin.instance.version
+            );
         }
 
         plugins
@@ -135,14 +139,15 @@ impl PluginInstance {
         }
 
         // Get existing plugin with same code in the plugin provider
-        let plugins = PLUGINS.read().unwrap();
-        if let Some(existing_plugin) = (*plugins).iter().find(|p| p.instance.code == code) {
-            if existing_plugin.instance.version > version {
-                // Existing plugin is higher version, skip (still install if same version)
-                return;
+        {
+            let plugins = PLUGINS.read().unwrap();
+            if let Some(existing_plugin) = (*plugins).iter().find(|p| p.instance.code == code) {
+                if existing_plugin.instance.version > version {
+                    // Existing plugin is higher version, skip (still install if same version)
+                    return;
+                }
             }
-        }
-        drop(plugins);
+        } // Drop read lock
 
         // Prepare plugin bundle
         let plugin_bundle = BASE64_STANDARD.decode(bundle_base64).unwrap();
@@ -157,7 +162,7 @@ impl PluginInstance {
         let instance = Arc::new(plugin);
 
         let mut plugins = PLUGINS.write().unwrap();
- 
+
         // Remove existing plugins of older versions with same code
         (*plugins).retain(|p| p.instance.code != code);
 
--- a/server/service/src/plugin/mod.rs
+++ b/server/service/src/plugin/mod.rs
@@ -159,14 +159,15 @@ pub trait PluginServiceTrait: Sync + Send {
         }
 
         // Get the existing cached version of the plugin
-        let plugins = ctx.frontend_plugins_cache.0.read().unwrap();
-        if let Some(existing_plugin) = (*plugins).get(&code) {
-            if existing_plugin.metadata.version > version {
-                // Existing cached plugin is higher version, skip (still install if same version)
-                return;
+        {
+            let plugins = ctx.frontend_plugins_cache.0.read().unwrap();
+            if let Some(existing_plugin) = (*plugins).get(&code) {
+                if existing_plugin.metadata.version > version {
+                    // Existing cached plugin is higher version, skip (still install if same version)
+                    return;
+                }
             }
-        }
-        drop(plugins);
+        } // Drop read lock
 
         let mut files_content = HashMap::new();
         // Prepare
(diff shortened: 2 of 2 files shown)


server/cli/src/plugins.rs
```diff
@@ -228,20 +230,23 @@ fn bundle_backend_plugin(
...
+
     bundle.backend_plugins.push(BackendPluginRow {
-        // TODO for now id = code in the future id = code + version (similar to reports)
-        id: format!("backend_{code}"),
+        id: format!("backend_{code}_{version_id}"),
         bundle_base64: bundle_base64,
         variant_type,
         types,
         code,
+        version, // TODO parse version to check format is valid?
```
jmbrunskill: This is a nice idea, would probably be a good idea to catch this error here rather than in production

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -80,31 +101,26 @@ pub struct PluginBundle {
...
-        plugins
-            .iter()
+        compatible_plugins_with_current_app_version()
+            .into_iter()
             .find(|p| p.has_type(&r#type))
-            .map(|p| p.instance.clone())
+            .map(|p| p.instance)
     }
 
     pub fn get_all(r#type: PluginType) -> Vec<Arc<PluginInstance>> {
```
jmbrunskill: This needs to de-dup by code, but that's covered in your next PR so that's fine.

server/service/src/backend_plugin/types/transform_request_requisition_lines.rs
```diff
@@ -48,25 +48,28 @@ impl self::Trait for PluginInstance {
...
     pub(crate) fn transform_request_requisition_lines(
         context: Context,
-        lines: Vec<RequisitionLineRow>,
+        mut lines: Vec<RequisitionLineRow>,
         requisition: &RequisitionRow,
     ) -> PluginResult<(Vec<RequisitionLineRow>, Vec<PluginDataRow>)> {
-        let Some(plugin) = PluginInstance::get_one(plugin_type()) else {
-            return Ok((lines, Vec::new()));
-        };
+        let plugins = PluginInstance::get_all(plugin_type());
```
andreievg: Nice !

ria8651:
Would it be worth having a combined frontend and backend plugin cache? Seem as they implement the same logic. Would also be good to rename the plugin modules and types to clearly differentiate between frontend and backend functionality.

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -113,26 +122,45 @@ impl PluginInstance {
...
 
+        // Get existing plugin with same code in the plugin provider
+        let plugins = PLUGINS.read().unwrap();
+        if let Some(existing_plugin) = (*plugins).iter().find(|p| p.instance.code == code) {
+            if existing_plugin.instance.version >= version {
+                // Existing plugin is same or higher version, skip
+                return;
+            }
+        }
+        drop(plugins);
```
andreievg: Drop all plugins 😱

server/service/src/backend_plugin/plugin_provider.rs
```diff
@@ -113,26 +122,45 @@ impl PluginInstance {
...
+        if !version.is_compatible_by_major_and_minor(&app_version) {
+            return;
+        }
 
+        // Get existing plugin with same code in the plugin provider
+        let plugins = PLUGINS.read().unwrap();
+        if let Some(existing_plugin) = (*plugins).iter().find(|p| p.instance.code == code) {
+            if existing_plugin.instance.version >= version {
+                // Existing plugin is same or higher version, skip
+                return;
```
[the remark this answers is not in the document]
jmbrunskill: Agreed

server/service/src/backend_plugin/plugin_provider.rs
jmbrunskill: Good catch, I thought I had two running at once but maybe I didn't!

PR #10169, opened 2026-01-06
Use base database when creating a fresh install

### LINKED ISSUE #8703: Improve migration speed and use transactions
See comments below, actual milestone set to 2.16.0 for speeding up migrations with migration dump

## The problem

Currently our migrations don't run in transactions which means that part of migration can succeed while other parts fail. This leaves the database in an unknown state which can result in tablets unable to upgrade and require re-intialisation (If it wasn't synced would it result in lost data) plus additional effort from QA, Support and Dev teams to resolve this issue.

## Suggestion

Wrap every version upgrade migration in a transaction.

NOTE: There are problems with doing this particularly with SQLite, as we have a number of migrations that use `PRAGMA FOREIGN_KEYS: OFF`

When inside a transaction this statement silently does nothing which may break existing migrations that were working.
https://sqlite.org/foreignkeys.html
https://sqlite.org/pragma.html#pragma_defer_foreign_keys

Suggestions...

1. Pick a version of oms that doesn't rely on lots of our old migrations and update this migration step to a single set of create initial tables. Any older versions would get an error and ask the user to upgrade to an intermediate version. This would also allow us to remove diesel migrations dependancy.
2. Build a mechanism to work around PRAGMA FOREIGN_KEYS: OFF for future migrations `PRAGMA defer_foreign_keys` should work in many cases, or perhaps we'll just need to do more re-creating tables to avoid these issues?

There is a an increasing risk particularly with tablets with poor internet.
It would be good to at least be able to reinstall the old version and keep working if there was a migration issue with a client device.

Yes, one issue moving to transactions has been identified for SQLite but there could be others.
Will require lots of migration update testing to be sure we've got a good solution

I'd estimate we should allocate 3-4 days to allow proper testing and migration updates

jmbrunskill:
Fixes #8703

# 👩🏻‍💻 What does this PR do?

With this PR, when a database is created by OMS, on start up by default it will get a snapshot of the migration based on the latest dump.

If a db version is specified e.g. during migration tests, we start form the earliest migration version.

ALSO: The `memory` feature has been removed, it wasn't much faster than the template system @clemens-msupply designed and wasn't working anyway.

To create these dumps I've used the template databases we use for tests, I'm pretty sure that's fine but maybe theres some nuance there that might affect things?

I haven't removed diesel migrations yet, will do in the next PR...

# 🧪 Testing

- [ ] Test that the app functions normally after creating a fresh database, and initialising.
- [ ] Test Android, Windows and Mac

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/base_migration.rs
+++ b/server/repository/src/migrations/base_migration.rs
@@ -1,8 +1,7 @@
 use crate::diesel::connection::SimpleConnection;
 use crate::diesel::RunQueryDsl;
 use crate::diesel_helper_types::Count;
-use crate::StorageConnection;
-use diesel::result::Error;
+use crate::{RepositoryError, StorageConnection};
 
 // This file defines the base database migrations that gets run when creating new database from scratch
 // This means new installs shouldn't need to run as many migrations, as long as this is kept up todate.
@@ -39,17 +38,17 @@ const EMPTY_DB_QUERY: &str = "
     WHERE table_schema = 'public';
 ";
 
-pub fn is_empty_db(conn: &StorageConnection) -> Result<bool, Error> {
+pub fn is_empty_db(conn: &StorageConnection) -> Result<bool, RepositoryError> {
     let result: Count = diesel::sql_query(EMPTY_DB_QUERY).get_result(conn.lock().connection())?;
     Ok(result.count == 0)
 }
 
-pub fn initialize_latest_db(conn: &StorageConnection) -> Result<(), Error> {
+pub fn initialize_latest_db(conn: &StorageConnection) -> Result<(), RepositoryError> {
     conn.lock().connection().batch_execute(BASE_SCHEMA_LATEST)?;
     Ok(())
 }
 
-pub fn initialize_earliest_db(conn: &StorageConnection) -> Result<(), Error> {
+pub fn initialize_earliest_db(conn: &StorageConnection) -> Result<(), RepositoryError> {
     conn.lock()
         .connection()
         .batch_execute(BASE_SCHEMA_EARLIEST)?;
--- a/server/repository/src/migrations/mod.rs
+++ b/server/repository/src/migrations/mod.rs
@@ -114,8 +114,6 @@ pub enum MigrationError {
         version: Version,
         identifier: &'static str,
     },
-    #[error("Error initializing base database schema: {0}")]
-    InitializeDatabaseError(anyhow::Error),
     #[error(transparent)]
     DatabaseError(#[from] RepositoryError),
 }
@@ -177,19 +175,17 @@ pub fn migrate(
     ];
 
     // Check if the database has been initialised, if not run the base sql to kick start the process
-    if is_empty_db(connection).map_err(|e| MigrationError::InitializeDatabaseError(e.into()))? {
+    if is_empty_db(connection)? {
         log::info!("Empty database detected, creating base schema...");
 
         if to_version.is_some() {
             log::info!("Target version specified, initializing earliest base schema");
             // We always use the earliest base schema when migrating to a specific version
             // This is the easiest way to makes sure migration tests can still run.
-            initialize_earliest_db(connection)
-                .map_err(|e| MigrationError::InitializeDatabaseError(e.into()))?;
+            initialize_earliest_db(connection)?;
         } else {
             log::info!("No target version specified, initializing latest base schema");
-            initialize_latest_db(connection)
-                .map_err(|e| MigrationError::InitializeDatabaseError(e.into()))?;
+            initialize_latest_db(connection)?;
         }
         log::info!("Base schema...installed");
     }


server/repository/src/db_init.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::diesel::RunQueryDsl;
+use crate::diesel::connection::SimpleConnection;
+
+// This file defines the base database definition that gets run when creating new database from scratch
+// This means new installs shouldn't need to run many migrations, as long as this is kept up todate.
+// I've used the test db templating feature to get a `base` database to use for tests and initialisation of a new db.
+
+
+// To refresh run tests to create template databases, then run...
+//  pg_dump -h localhost -U postgres -d ___template_1.0.4 --inserts  --no-owner --no-privileges --schema=public | sed -E '/^\\(un)?restrict [A-Za-z0-9]+$/d' > repository/src/new_db_init/postgres_earliest.sql
```
jmbrunskill: Took a bit of trial and error + AI to help me get this to work...
Hopefully is repeatable now though

server/repository/src/database_settings.rs
```diff
@@ -1,21 +1,19 @@
...
 use serde::{Deserialize, Serialize};
-use std::time::Duration;
 
 // Timeout for waiting for the SQLite lock (https://www.sqlite.org/c3ref/busy_timeout.html).
 // A locked DB results in the "SQLite database is locked" error.
 #[cfg(not(feature = "postgres"))]
 const SQLITE_LOCKWAIT_MS: u32 = 30 * 1000;
 
-#[cfg(all(not(feature = "postgres"), not(feature = "memory")))]
+#[cfg(all(not(feature = "postgres")))]
```
ria8651:
```suggestion
#[cfg(not(feature = "postgres"))]
```

Theres a few more of these

ria8651 (commented):
I tried running this with some new migrations and it seems like all the newer migration fragments ran fine off of `postgres_latest.sql` without the old ones needing to be run.

server/repository/src/migrations/mod.rs
```diff
@@ -169,15 +168,31 @@ pub fn migrate(
         Box::new(v2_13_01::V2_13_01),
         Box::new(v2_14_00::V2_14_00),
         Box::new(v2_15_00::V2_15_00),
-        Box::new(v2_16_00::V2_16_00),
     ];
 
+    // Check if the database has been initialised, if not run the base sql to kick start the process
+    if is_empty_db(connection).expect("Unable to check for empty db on startup: Nothing returned") {
```
andreievg: Although I think expect works well here, it's not consistent with the rest of the method, I think it's worth keeping it consistent

server/repository/src/diesel_helper_types.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct Count {
+    #[diesel(sql_type = diesel::sql_types::Integer)]
+    pub count: i32,
+}
+
+#[cfg(feature = "postgres")]
+#[derive(QueryableByName, Debug)]
+pub struct Count {
+    #[diesel(sql_type = diesel::sql_types::BigInt)]
+    pub count: i64,
```
andreievg: Why would these be different structs ?

server/repository/src/db_init.rs
```diff
@@ -0,0 +1,63 @@
+use diesel::result::Error;
+use crate::StorageConnection;
+use crate::diesel_helper_types::Count;
+use crate::diesel::RunQueryDsl;
+use crate::diesel::connection::SimpleConnection;
+
+// This file defines the base database definition that gets run when creating new database from scratch
```
andreievg: Should this file really live in migration rather then test_db ?

I like the comment here but I think it's worth adding something about this in the README.md of migrations

server/repository/src/diesel_helper_types.rs
jmbrunskill: Hmm, I added this as it didn't work for postgres with i32, but you're probably right I think i64 will probably work for sqlite too...

server/repository/src/db_init.rs
jmbrunskill: Not sure what you mean by test_db, but I think you're right it's kind of a base migration.
I think I got a bit lost as I was originally trying to add this to the create database logic but actually it's just a base migration.

server/repository/src/migrations/base_migration.rs
```diff
@@ -7,13 +7,14 @@ use diesel::result::Error;
 // This file defines the base database migrations that gets run when creating new database from scratch
 // This means new installs shouldn't need to run as many migrations, as long as this is kept up todate.
 // I've used the test db templating feature to get a `base` database to use for tests and initialisation of a new db.
+// Note: It's probably best to dump the database from the earliest version of postgres we support, but have tried to create clean up commands for dumping from v18
```
andreievg: What is v18 ?

andreievg:
@jmbrunskill we don't typically use diesel::Error, there is repository error and it becomes quite easy to map errors at this stage: https://github.com/msupply-foundation/open-msupply/compare/8703-migrations...8703-migrations(Error-map).

I'll talk about errors in rust at some point in [sharing sessions](https://trello.com/c/TzSKYSxK/17-errors-in-rust) (specifically thiserror, why we use it etc..)

Chris-Petty:
> ALSO: The `memory` feature has been removed, it wasn't much faster than the template system @clemens-msupply designed and wasn't working anyway.

Hell yea removing unused 🔥 (soz clemens)

server/repository/src/migrations/base_migration.rs
Chris-Petty: PG v18 - I presume there is a minor risk that the SQL commands dumped from v18 aren't supported on v12 (our current default that the installer downloads)

jmbrunskill:
> @jmbrunskill we don't typically use diesel::Error, there is repository error and it becomes quite easy to map errors at this stage: [8703-migrations...8703-migrations(Error-map)](https://github.com/msupply-foundation/open-msupply/compare/8703-migrations...8703-migrations(Error-map)).
> 
> I'll talk about errors in rust at some point in [sharing sessions](https://trello.com/c/TzSKYSxK/17-errors-in-rust) (specifically thiserror, why we use it etc..)

Yes please, @Chris-Petty and I were trying to figure out what `thiserror` exists not long ago!

Will update based on your branch, although we already have a mapping for repository error, do you care if it says initialisation error or are you ok if we just had the error log message a generic error passed back? To me this is all busy work as we're not actually able to do anything useful for the error, I guess we could potenitally provide a translated error message for the upgrade state, but I think that's a lot more work that we wanted to spend now?

server/repository/src/migrations/base_migration.rs
jmbrunskill: Does the installer install 12??? I didn't realise that, was assuming we just had a few old versions to clean up....

It's not a minor risk at all! In fact I've done quite a few sed shenanigans to try to make it backwards compatible, my AI now says it's should be compatible but I've actually only tested back till 13 as that's the earliest I can get post app to install. I can do a test with docker now that I think about it. I think or CI should at least test against our earliest supported version btw, I've checked and the mac mini is running postgres v16 currently for the tests....

PR #10172, opened 2026-01-06
Remove legacy diesel migrations

### LINKED ISSUE #8703: Improve migration speed and use transactions

jmbrunskill:
Fixes #8703

# 👩🏻‍💻 What does this PR do?

Removed diesel migrations as we now had a method to define the `base` database...

#8703 mentions transactions, maybe we can do this next?

# 🧪 Testing

- [ ] Check upgrades and new dbs both still work

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/mod.rs
+++ b/server/repository/src/migrations/mod.rs
@@ -196,9 +196,6 @@ pub fn migrate(
 
     let to_version = to_version.unwrap_or(Version::from_package_json());
 
-    // Historic diesel migrations
-    run_db_migrations(connection).unwrap();
-
     // Rust migrations
     let starting_database_version = get_database_version(connection);
 


server/repository/Cargo.toml
```diff
@@ -21,7 +21,6 @@ diesel = { version = "2.2.12", default-features = false, features = [
     "with-deprecated",
 ] }
 diesel-derive-enum = { version = "2.1.0", default-features = false }
-diesel_migrations = "2.2.0"
```
andreievg: yay

andreievg (approved):
Looks great thanks

Issue #9824, opened 2025-11-20
Rethink name_link join in name repository (and possibly item as well)


jmbrunskill:
## Background

Currently in the name repository, we're joining from `name` to `name_link` then left join to `name_store_join`.

This allows us to use the results of the query to return information such as `is_customer` on the name graphql query without extra lookups.

The same query also used for ProgramSupplier and CustomerSupplier requests that summaries available programs, this requires currently joining more tables...

## The problem

Any time we naively join `name` to `name_link` (it doesn't matter if its left or inner join) any merged names will return multiple database records.

E.g. if you have 

`Name B` is merged into `Name A`
Querying 
```
SELECT  n.id,n.name, nl.id as name_link_id, name_link.name_id
FROM name n 
JOIN name_link nl ON n.id = nl.name_id
WHERE n.id =  'name_a_id'
```
Returns 2 rows like this....

| n.id  | name  | name_link_id.id | name_id | 
| -------- | ------- | ------- | ------- |
| name_a_id | Name A | name_a_id | name_a_id|
| name_a_id | Name A | name_b_id | name_a_id|

This appears as duplication in side the app, and requires external filtering.

What we really want in this situation is to return just the remaining name.

| n.id  | name  |
| -------- | ------- | 
| name_a_id | Name A | 

However it's not as simple as that!

If there are other related entities in the system such as `name_store_join` or `master_list_name_join`, we want to associate the union of the records across the `Name A` and `Name B`

For example Name A might have a name store_join to Store X,
But only Name B has a `master_list_name_join` and we want to return both the `name_store_join` or `master_list_name_join` records for a name, we'd need to merge the results across the different tables.

### Possible solutions

1. Situational specific merging or inclusion of records in rust code.
One example is this PR, which looks for the db record that matches the correct name store_join row to return.

https://github.com/msupply-foundation/open-msupply/pull/9807

If this approach and for other too perhaps, we can and perhaps should clean up how we query program & supplier combinations?

2. Create a db view that de-dups and aggregates these joins.
For example if we just need the `name` & `name_store_join` and not associated data we can do this...
`SELECT DISTINCT n.id, name_is_customer, name_is_supplier, nsj.store_id
FROM name n 
JOIN name_link nl ON n.id = nl.name_id
LEFT JOIN name_store_join nsj ON n.id = nsj.name_link_id `

Question is how do we scale this approach to other `name_link` joins? 

[...]

PR #10181, opened 2026-01-06
2937 feat(repository): Abstract name_link_id from invoice table public API (AI Generated)

### LINKED ISSUE #2937: Make item_link_id, name_link_id and clinician_link_id private to the repository layer
This can be done by not export the underlying diesel rows but having a "proper" domain objects that contains the resolved `item_id`, `name_id`, `clinician_id`.

Please see new `architecture.md` docs for more details.

jmbrunskill:
## Summary

  This PR applies the entity linking abstraction pattern to the `invoice` table, hiding the internal `name_link_id` column from the public API and exposing only the resolved `name_id`.

  Fixes #2937

  ### Background

  When names are merged (e.g., "Name B" merged into "Name A"), the `name_link` table tracks this relationship. Currently, the `name_link_id` implementation details leak throughout the codebase. This PR follows Andrei's proposed pattern from #9824 to abstract this complexity.

  ### Changes

  **Database Layer:**
  - Created `invoice_view` that joins `invoice` with `name_link` to resolve `name_link_id` → `name_id`
  - Defined dual Diesel table schemas for SELECT (view) and INSERT/UPDATE (core)

  **Repository Layer:**
  - Updated `InvoiceRow` struct to use `name_id` instead of `name_link_id`
  - Updated `upsert_one()` to translate `name_id` → `name_link_id` when writing

  **Service Layer:**
  - Updated all invoice-related code to use `name_id`

  **Documentation:**
  - Added comprehensive documentation at `server/docs/content/docs/entity-linking/`

  ### Remaining Tables

  Same pattern has been applied to other tables linking to name, including donor, and other optional links


server/graphql/types/src/types/purchase_order.rs
```diff
@@ -107,8 +107,8 @@ impl PurchaseOrderNode {
     pub async fn currency_id(&self) -> &Option<String> {
         &self.row().currency_id
     }
-    pub async fn foreign_exchange_rate(&self) -> &Option<f64> {
-        &self.row().foreign_exchange_rate
+    pub async fn foreign_exchange_rate(&self) -> f64 {
```
jmbrunskill: Not sure how these foreign_exchange_rate changes were affected by this PR. Is seems in the db, this is a not null but was being treated as nullable in the code??? 

See:
https://github.com/msupply-foundation/open-msupply/blob/fcd2a31dc435a86da7db0bf3f0cfa50a4206e262/server/repository/src/migrations/v2_10_00/add_purchase_order_tables.rs#L47

Why these changes caused tests to fail though. AI fixed in this way which I agree with, but still, a definitely out of the main scope...

server/service/src/invoice_line/stock_in_line/generate.rs
```diff
@@ -45,7 +45,7 @@ pub fn generate_batch(
         location_id,
         note,
         item_variant_id,
-        donor_link_id,
+        donor_id: donor_link_id,
```
jmbrunskill: This should just be donor_id all the way through, but I think it's ok?? Should we update or ask AI too @andreievg seems ok to me? Just a clean up task...

server/repository/src/migrations/views/link_views.rs
```diff
@@ -0,0 +1,446 @@
...
+                SELECT
+                    indicator_value.*,
+                    name_link.name_id as customer_name_id
+                FROM
+                    indicator_value
+                JOIN
+                    name_link ON indicator_value.customer_name_link_id = name_link.id;
+
+                CREATE VIEW stock_line_view AS
+                SELECT
```
andreievg: a lot of overhead

server/repository/src/migrations/views/link_views.rs
```diff
@@ -0,0 +1,446 @@
...
+                    stock_line.id,
```
andreievg: @copilot use * here like previous views, also for other views in this file, keep with pattern that doesn't need extra changes in the future when fields are added

.claude/settings.local.json
```diff
@@ -0,0 +1,23 @@
+{
+  "permissions": {
```
jmbrunskill: @copilot  Remove this file and make it ignored in gitignore

jmbrunskill:
## Merge with develop — resolved conflicts

Merged latest develop into this branch. There were **6 merge conflicts**, all resolved:

| File | Conflict | Resolution |
|------|----------|------------|
| `server/graphql/core/src/loader/name.rs` | Develop still had `NameByNameLinkIdLoader`; PR removed it | Kept PR's removal — the loader is fully replaced by `NameByIdLoader` |
| `server/repository/src/db_diesel/name.rs` | Develop reformatted `NameLinkRow` in test; PR removed the field entirely | Kept PR's removal of `name_link_row` from `Name` struct |
| `server/repository/src/db_diesel/item.rs` | Develop moved `create_filtered_query` into `impl ItemRepository`; PR had it as standalone with `name_id` changes | Removed duplicate standalone function, updated the impl block version to use `store::name_id` and `master_list_name_join::name_id` |
| `server/service/src/invoice/inbound_shipment/update/mod.rs` | PR: `donor_id == None` vs develop: `donor_link_id.is_none()` | Used `donor_id.is_none()` (PR's field name + develop's idiomatic style) |
| `server/service/src/sync/translations/vaccination_legacy.rs` | PR used direct assignment (IDs already resolved by view); develop had old name_link lookup | Kept PR's clean version — IDs are already resolved by the view |
| `client/packages/dashboard/src/api/operations.generated.ts` | Generated file conflicts | Accepted develop's version — will need `yarn generate` to regenerate |

**Note:** The `operations.generated.ts` file should be regenerated with `yarn generate` after this merge.

server/repository/src/db_diesel/name.rs
```diff
@@ -374,7 +363,7 @@ impl NameFilter {
         self
     }
 
-    pub fn name_link_id(mut self, filter: StringFilter) -> Self {
+    pub fn name_link_id(mut self, filter: EqualFilter<String>) -> Self {
```
andreievg: @Copilot can this be deleted ? along with name_link_id on the filter ?

server/repository/src/db_diesel/vaccination.rs
```diff
@@ -110,44 +107,30 @@ impl<'a> VaccinationRepository<'a> {
...
         .left_join(clinician_link::table.inner_join(clinician::table))
         .left_join(item_link::table.inner_join(item::table))
         .inner_join(vaccine_course_dose::table)
-        .left_join(
-            name_link::table
-                .on(vaccination::facility_name_link_id.eq(name_link::id.nullable()))
-                .inner_join(name::table),
-        )
-        .into_boxed();
+        .left_join(name::table.on(vaccination::facility_name_id.eq(name::id.nullable())))
```
andreievg: @Copilot does this have to be an on join ? I think join can be derived here

andreievg:
@copliot there is still some things to change:

document.rs uses .left_join(name_link::table.inner_join(name::table)) to resolve the document owner's name, and purchase_order.rs uses a name_link::table.inner_join(name::table) subquery to filter by supplier name string

can these be abstracted away using methods in PR

Chris-Petty:
I've definitely not read this deep enough or a test whirl to give a proper review, though wanted to wrap some thoughts up I had from IRL discussion. I think this looks great and seems like a massive improvement!

While it touches many files (largely cause the struct field changed) it's really a pretty simple change. Eliminating some of the nasty loaders like `NameByNameLinkIdLoaderInput` is nice.

I do wonder about graphql backwards compatibility earlier but had neglected that we don't really expose the name_link_id in the API, and it definitely is confusing abstracting it so high in the stack. 

For future unmerge of name records (i.e. unmerge patients in particular), at a glance I don't think this strictly restricts it - in the worst case we can make a special repo that uses the actual table right?

PR #10186, opened 2026-01-07
10158 fix client crash when server down

### LINKED ISSUE #10113: Runtime error when re-initialising
## What went wrong? 😲

If you've previously initialised, and then logged in, and then cleared database, you'll have runtime error until you've cleared your cookies

## Expected behaviour 🤔

No runtime error

Suggested branch to fix this: https://github.com/msupply-foundation/open-msupply/compare/main...main(Clear-token-if-initialisation-status-is-not-initialised)

## Context

Introduce in 2.14.2 while adding a preference that affects original navigation, this preference needs to load before app is loaded, thus suspense was used on usePreference. However usePreference has previously thrown an error (during initialisation) that did not affect the app, as this error wasn't handled. Now with suspense it's handled. We though that using token check for storeId to enable/disable fetching of preference would work, but realised we are not clearing app token unless there is auth error or logout action, which doesn't happen in described scenario

## How to Reproduce 🔨

As per what went wrong. This is frequent thing in dev, not common at all in prod. For android clearing data should remove all data, including any cookies. On client/server setup, including web browser and client app, this would happen if server is re-set.

## Your environment 🌱

- Open mSupply Version: 2.14.2

- Platform:

- Database type:

### LINKED ISSUE #10158: Client crashes if it can't access the server
## What went wrong? 😲

When running the client without the server up it freezes till a timeout and crashes. Looks like this issue was introduced by #10116 here:

https://github.com/msupply-foundation/open-msupply/blob/e3cb804e74c85990a402ea06e74b8be3f8465ca5/client/packages/host/src/Host.tsx#L74

## Your environment 🌱

- Open mSupply Version: develop
- Platform: macOS, chrome
- Database type: postgres

ria8651:
Fixes #10158

# 👩🏻‍💻 What does this PR do?

Even before #10116 when there are cookies in the cache and the server is unaccessible the client crashes with a runtime error. This PR fixes #10113 and #10158 by using option 1 of the following:

1. (this pr) remove suspense from initialisation queries. This does mean that preferences and is initialised will have placeholder values to begin with but I haven't been able to find any cases where this causes a problem. While waiting for the queries I return an empty preferences object and return isInitialised as false. Some more rigorous testing would be required to make sure this doesn't cause problems.
2. Add error boundary for initialisation. This would let you see the page but would still freeze the page load until the initialisation queries time out due to the suspense.
3. Just revert #10116. This would at least let the page load when you don't have cookies.

# 🧪 Testing

# 📃 Documentation

  1.
  4.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/api/hooks/useInitialisationStatus.ts
+++ b/client/packages/common/src/api/hooks/useInitialisationStatus.ts
@@ -3,7 +3,6 @@ import { getSdk } from '../operations.generated';
 
 export const useInitialisationStatus = (
   refetchInterval: number | false = false,
-  shouldSuspend?: boolean
 ) => {
   const { client } = useGql();
   const sdk = getSdk(client);
@@ -16,7 +15,6 @@ export const useInitialisationStatus = (
     },
     {
       cacheTime: 0,
-      suspense: shouldSuspend,
       refetchInterval,
     }
   );
--- a/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts
+++ b/client/packages/common/src/authentication/api/hooks/useIsCentralServer.ts
@@ -15,7 +15,7 @@ export const useIsCentralServerApi = () => {
       refetchOnMount: false,
       cacheTime: Infinity,
       staleTime: Infinity,
-      suspense: true,
+      suspense: false,
       enabled: !!storeId,
     }
   );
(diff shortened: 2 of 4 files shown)


andreievg:
@ria8651 i was looking at this a bit, and have alternative solution: https://github.com/msupply-foundation/open-msupply/pull/10196, I would like to deal with the issue at the root, especially since we do need preference before router is loaded to navigate to gaps page if that preference is set. Also we used to get glitched with login page showing before initialisation page, which was fixed.

@mark-prins the issue name and some of the description says that we are fixing a client crash when server is not accessible, I would have thought that's the behaviour we want, an error message when serve is not accessible

PR #10188, opened 2026-01-07
9651 users are able to edit line item quantity for auto generated customer returns

### LINKED ISSUE #9651: Users are able to edit line item quantity for auto generated Customer returns
## What went wrong? 😲

I am getting a LineDeleteError when returning items in customer return orders, specifically if I edit a line item's quantity.
For example: I created an internal order for Amoxicillin 20 packs. When I received it, I noticed an issue with the medicine and returned the whole quantity to the supplier (I created a supplier return).
The supplier receives the customer return, and initially, I cannot edit the received quantity in the customer return note. However, I can add an additional line item after confirming the customer return. At that point, I am also able to edit the original line item quantities (ref: video recording 4 min onwards).
https://www.loom.com/share/dab14a965dad4589a10d0d3fa446d7d4

Afterwards, my customer return note allows me to edit any item, creating a flaw that could allow for manipulating the order.

https://www.loom.com/share/01373d587d484078a5e00671866017e2

## How to Reproduce 🔨

1. An internal order for 20 packs of Amoxicillin was created.
2. The full quantity was returned to the supplier via a supplier return order due to a quality issue.
3. The customer return is received.
Initially, the received quantity cannot be edited in the customer return note.
4. After confirming the customer return: I am able to add an additional line item.
Simultaneously: I can also edit the quantities of the original line items (ref: video recording 4 min onwards).
Result: The system allows me to edit any item within the customer return note after this point, which is a flaw that could lead to unauthorized order manipulation.

## Your environment 🌱

- Open mSupply Version: 2.12.00-develop-10301855

- Platform:

- Database type:

roxy-dao:
Fixes #9651

# 👩🏻‍💻 What does this PR do?
- Show sell price in customer returns
- Disable quantity inputs for customer returns if customer return was created from a supplier return
- When transferring from supplier return, save correct cost / sell prices

# 🧪 Testing

- [ ] Create a supplier return to Store B
- [ ] Go to Store B, quantities shouldn't be editable
- [ ] Cost price should show in the detail view

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/processors/transfer/invoice/common.rs
+++ b/server/service/src/processors/transfer/invoice/common.rs
@@ -40,7 +40,7 @@ pub(crate) fn generate_inbound_lines(
                     invoice_id: _,
                     stock_line_id: _,
                     location_id: _,
-                    cost_price_per_pack: _,
+                    cost_price_per_pack,
                     total_after_tax: _,
                     linked_invoice_id: _,
                     reason_option_id,
@@ -79,12 +79,12 @@ pub(crate) fn generate_inbound_lines(
 
                 let supplier_id = &source_invoice.store_row.name_link_id;
 
-                let cost_price_per_pack = sell_price_per_pack;
+                let trans_cost_price = sell_price_per_pack;
 
                 let total_before_tax = match r#type {
                     // Service lines don't work in packs
                     InvoiceLineType::Service => total_before_tax,
-                    _ => cost_price_per_pack * number_of_packs,
+                    _ => trans_cost_price * number_of_packs,
                 };
 
                 let default_price_per_default_pack = item_properties
@@ -101,13 +101,8 @@ pub(crate) fn generate_inbound_lines(
                 let adjusted_sell_price_per_pack = if default_price_for_inbound_pack > 0.0 {
                     default_price_for_inbound_pack
                 } else {
-                    get_cost_plus_margin(
-                        connection,
-                        cost_price_per_pack,
-                        item_properties,
-                        supplier_id,
-                    )
-                    .unwrap_or(cost_price_per_pack)
+                    get_cost_plus_margin(connection, trans_cost_price, item_properties, supplier_id)
+                        .unwrap_or(trans_cost_price)
                 };
 
                 InvoiceLineRow {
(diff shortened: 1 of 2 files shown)


server/service/src/processors/transfer/invoice/common.rs
```diff
@@ -121,7 +121,10 @@ pub(crate) fn generate_inbound_lines(
                     pack_size,
                     total_before_tax,
                     total_after_tax: calculate_total_after_tax(total_before_tax, tax_percentage),
-                    cost_price_per_pack: sell_price_per_pack,
+                    cost_price_per_pack: match invoice_row.r#type {
+                        InvoiceType::SupplierReturn => cost_price_per_pack,
```
roxy-dao: Probably didn't need to do this since supplier return cost price is the sell price from the Inbound Shipment 👀

client/packages/invoices/src/Returns/modals/CustomerReturn/ReturnSteps.tsx
```diff
@@ -85,7 +83,7 @@ export const ReturnSteps = ({
       </TabPanel>
       <TabPanel value={Tabs.Reason}>
         <ReturnReasonsTable
-          isDisabled={inputsDisabled}
+          isDisabled={isDisabled}
```
ria8651: Just checking you are supposed to be able to change the return reason for an auto-generated return? Might be good to change the variable names as disabled and isDisabled is a bit confusing.

ria8651 (commented):
Looks good, can verify the changes. I'll leave it for someone else to approve as I'm not confident in how the price mapping should work.

aimee-mcneil-melville (changes requested):
Awesome the quantity part is great! I do have the same question as @ria8651 though - is there a use case for the reasons being editable?

For the pricing, I still see the suppliers sell price = customers sell price instead of cost price - can you update this please? 

Maybe some renaming would be beneficial for clarity in the processor as the customers cost_price_per_pack isn't used, and then a new cost_price_per_pack is defined with the value of sell price.

For the supplier, is their cost price expected to be the same as their sell price?

roxy-dao:
Opps, yeah the transfer was getting sell price instead of cost price ;x

mark-prins:
> should reasons be editable? 

happy to have this as not editable - thanks @roxy-dao

PR #10192, opened 2026-01-07
Fixes #4918: avoid bogus ellipsis from float noise

### LINKED ISSUE #4918: Total doesn't display the correct value

ksawaneh:
Fixes #4918

The ellipsis ("...") marker for >2dp values was triggering for values like `4.40000000000006`, causing totals/currency cells to show `4.40...`.

- Make `NumUtils.hasMoreThanDp/hasMoreThanTwoDp` robust to floating-point rounding noise
- Add unit tests covering the regression

Tests:
- `docker run --rm -v "$PWD":/work -w /work/client node:20-bullseye bash -lc "yarn lint-and-format"`
- `docker run --rm -v "$PWD":/work -w /work/client node:20-bullseye bash -lc "yarn test"`

### DIFF
--- a/client/packages/common/src/utils/numbers/NumUtils.test.ts
+++ b/client/packages/common/src/utils/numbers/NumUtils.test.ts
@@ -41,8 +41,12 @@ describe('NumUtils', () => {
     });
 
     it('handles very large numbers', () => {
-      const veryLarge = 1e15 + 0.12;
-      expect(NumUtils.hasMoreThanDp(veryLarge, 2)).toBe(false);
+      // At this magnitude, doubles can only represent fractional steps of 0.125.
+      const veryLargeWithTwoDp = 1e15 + 0.25;
+      const veryLargeWithMoreThanTwoDp = 1e15 + 0.375;
+
+      expect(NumUtils.hasMoreThanDp(veryLargeWithTwoDp, 2)).toBe(false);
+      expect(NumUtils.hasMoreThanDp(veryLargeWithMoreThanTwoDp, 2)).toBe(true);
     });
 
     it('handles very small numbers', () => {
--- a/client/packages/common/src/utils/numbers/NumUtils.ts
+++ b/client/packages/common/src/utils/numbers/NumUtils.ts
@@ -55,7 +55,11 @@ export const NumUtils = {
     if (!Number.isFinite(value)) return false;
 
     const multiplier = 10 ** dp;
-    return !isNearlyInteger(value * multiplier);
+    // For very large values, `value * multiplier` can lose the fractional part due to IEEE-754
+    // precision limits (e.g. `1e15 + 0.12` becomes `1e15 + 0.125`). Only inspect the fraction.
+    const abs = Math.abs(value);
+    const fraction = abs - Math.trunc(abs);
+    return !isNearlyInteger(fraction * multiplier);
   },
   /**
    * This constant should be used for values that are potentially send to a backend API that expects


ksawaneh:
Added additional unit tests covering very large numbers, very small numbers, and negative numbers for `hasMoreThanTwoDp`/`hasMoreThanDp` in `NumUtils.test.ts` as requested. Rebasing is now up to date with `develop`. Verified with `cd client && yarn test --testPathPattern NumUtils.test.ts` and `cd client && yarn lint-and-format`.

client/packages/common/src/utils/numbers/NumUtils.test.ts
```diff
@@ -20,4 +20,55 @@ describe('NumUtils', () => {
...
+    expect(NumUtils.hasMoreThanDp(4.4, 2)).toBe(false);
+    expect(NumUtils.hasMoreThanDp(4.40000000000006, 2)).toBe(false);
+    expect(NumUtils.hasMoreThanDp(4.41, 2)).toBe(false);
+    expect(NumUtils.hasMoreThanDp(4.411, 2)).toBe(true);
+
+    expect(NumUtils.hasMoreThanDp(0.1234, 3)).toBe(true);
+    expect(NumUtils.hasMoreThanDp(0.123, 3)).toBe(false);
+  });
+
+  it('handles very large numbers', () => {
```
esmehm: Thanks for adding these tests! As these changes grow the size of the file, structure is more important - having flat `it(...)` blocks are hard to scan and would be better grouped under one `describe` block per function. 
For example,

```
describe('hasMoreThanTwoDp', () => {
    it('detects values with > 2 decimal places', () => {
      expect(NumUtils.hasMoreThanTwoDp(4.4)).toBe(false);
      expect(NumUtils.hasMoreThanTwoDp(4.40000000000006)).toBe(false);
      expect(NumUtils.hasMoreThanTwoDp(4.401)).toBe(true);
      expect(NumUtils.hasMoreThanTwoDp(0.01)).toBe(false);
      expect(NumUtils.hasMoreThanTwoDp(0.001)).toBe(true);
    });

    it('handles large numbers', () => {
      const large = 123456789.12;
      const largeWithNoise = large + 1e-8;

      expect(NumUtils.hasMoreThanTwoDp(large)).toBe(false);
      expect(NumUtils.hasMoreThanTwoDp(largeWithNoise)).toBe(false);
....
```

client/packages/common/src/utils/numbers/NumUtils.test.ts
```diff
@@ -20,4 +20,55 @@ describe('NumUtils', () => {
...
+    expect(NumUtils.hasMoreThanDp(0.0001, 4)).toBe(false);
+  });
+
+  it('handles negative numbers', () => {
+    expect(NumUtils.hasMoreThanTwoDp(-4.4)).toBe(false);
+    expect(NumUtils.hasMoreThanTwoDp(-4.40000000000006)).toBe(false);
+    expect(NumUtils.hasMoreThanTwoDp(-4.401)).toBe(true);
+
+    expect(NumUtils.hasMoreThanDp(-0.1234, 3)).toBe(true);
+    expect(NumUtils.hasMoreThanDp(-0.123, 3)).toBe(false);
```
esmehm: Please also add negative near zero case and test both sides of the boundary for it, eg:
```
expect(NumUtils.hasMoreThanDp(-0.0001, 4)).toBe(false);
expect(NumUtils.hasMoreThanDp(-0.00001, 4)).toBe(true);
```

esmehm (changes requested):
Generally good test coverage! 
However as this test file expands it becomes harder to scan and so please review the structure using describe and it statements to group by function as discussed in the comments. This will also allow you to have a comprehensive test suite for `hasMoreThanDp` and jusrt test that `hasMoreThanTwoDp` is working as a wrapper of the former

ksawaneh:
Thanks! I’ve restructured `NumUtils.test.ts` to group tests by function using `describe(...)`. The edge-case coverage (large/small/negative) now lives under `hasMoreThanDp`, and `hasMoreThanTwoDp` is only tested as a wrapper via `hasMoreThanDp(value, 2)`. Also added the negative near-zero boundary cases and updated the large-number stress test to use `1e15 + 0.12`.

PR #10193, opened 2026-01-07
Fixes #9695: disable editing for cancelled prescriptions

### LINKED ISSUE #9695: JS Exception when Saving a cancelled Prescriptions.
## What went wrong? 😲

This can be seen for both programs and non programs, where I can see the save button is enabled when you edit Item qty/Direction of the cancelled prescriptions. Although, the records are not saved to the database, we should disable the save button or disable the qty and description fields?

## Expected behaviour 🤔

Should not be able to click cancel button or edit qty or description for cancelled prescriptions.

## How to Reproduce 🔨

1. Go to Prescriptions >> click more and click cancel button.
2. Click on cancelled prescription and click on Item qty or Direction fields and change the values.
3. You can notice that the save button is getting enabled.
4. See error on the console logs.
https://drive.google.com/file/d/1ubDubIZQYftAmBtiSNWk0yLIw8nYYlQv/view?usp=drive_link

## Your environment 🌱

- Open mSupply Version: 2.12.00-develop-1105174

- Platform:

- Database type:

ksawaneh:
Fixes #9695

Cancelling a prescription could leave a brief window where the UI still allowed editing (quantity/directions), enabling Save and leading to a console error when attempting to persist changes.

- Update the cached prescription on successful update so the cancelled status (and `isDisabled`) is reflected immediately
- Ensure all LineEditView inputs respect the disabled state (including allocate-in + prescribed quantity)
- Disable Save (and creation via nav) when the prescription is cancelled/disabled

Tests:
- `docker run --rm -v "$PWD":/work -w /work/client node:20-bullseye bash -lc "yarn lint-and-format"`
- `docker run --rm -v "$PWD":/work -w /work/client node:20-bullseye bash -lc "yarn test"`

### DIFF
--- a/client/packages/invoices/src/Prescriptions/LineEditView/AllocationSection.tsx
+++ b/client/packages/invoices/src/Prescriptions/LineEditView/AllocationSection.tsx
@@ -25,7 +25,9 @@ export const AllocationSection = ({
 
   return (
     <>
-      {showPrescribedQuantity && <AutoAllocatePrescribedQuantityField />}
+      {showPrescribedQuantity && (
+        <AutoAllocatePrescribedQuantityField disabled={disabled} />
+      )}
       <Grid display="flex" alignItems="center" gap={1}>
         <AutoAllocateField
           inputColor="white"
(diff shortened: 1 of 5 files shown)


roxy-dao:
Sorry you had to do this but this issue isn't a problem anymore so I will close this PR ;x

PR #10196, opened 2026-01-08
10183 Fix login in APK issue

### LINKED ISSUE #10183: Cannot log in to Android application : Authentication Error
## What went wrong? 😲

I am block for log in to the android application after upgraded to the v 2.16.0 apk.

https://drive.google.com/file/d/1dY4AoaG4fXj-d4SuLB9M-tEf2sZgGqEd/view?usp=drive_link

This error prevents user login even if you delete the cache data.

## How to Reproduce 🔨

1. Upgrade to the v 2.16.0
2. 
## Your environment 🌱

- Open mSupply Version:

- Platform: Android TAB/ Phone

- Database type: N/A

andreievg:
Fixes #10183

# 👩🏻‍💻 What does this PR do?

In an attempt to fix #10183 and while reviewing related PRs, I wanted to fix this at the root.

I found that a preference and another query is using previous token in the authorisation header, after login, while other queries before it and after are using new token. 

It looks like our wrapper around graphql client is loosing this somehow, I've changed to:

* When extending use extend rather then composition 
* Use setHeader directly from client, otherwise we are loosing this binding

# 🧪 Testing

In APK (seems to work fine in normal browser ¯\_(ツ)_/¯).

Login. The close app, then try login again (should see error without this fix)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/api/GqlContext.tsx
+++ b/client/packages/common/src/api/GqlContext.tsx
@@ -94,7 +94,6 @@ const shouldSaveRequestTime = (documentNode?: DocumentNode) => {
 };
 
 class GQLClient extends GraphQLClient {
-  private client: GraphQLClient;
   private emptyData: object;
   private skipRequest: SkipRequest;
   private lastRequestTime: Date;
@@ -105,7 +104,6 @@ class GQLClient extends GraphQLClient {
     skipRequest?: SkipRequest
   ) {
     super(url, options);
-    this.client = new GraphQLClient(url, options);
     this.emptyData = {};
     this.skipRequest = skipRequest || (() => false);
     this.lastRequestTime = new Date();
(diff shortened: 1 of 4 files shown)


ria8651:
I've not been able to reliably reproduce the login issue on develop so I can't tell if this has fixed it, do you have more detailed instructions on reproducing it?

mark-prins (approved):
🙏  I'm not sure about the argument of how it fixes things - I think it more likely to be a captured enclosure, with the GQL context not being updated when the new token is fetched

whatever the reason - this is cleaner than the previous code and is working for me on android. let's get it in and hand over to Rukmal

andreievg:
@mark-prins thanks for review, it was hard to figure out what was going on, more of a trial an error. I was seeing this binding issue at the last step, when I was destructuring useGql as { client { setHeader }}, so I agree that the original cause wasn't well described in my comment

PR #10201, opened 2026-01-08
9240 create specific keys for statuses (e.g. label.status new)

### LINKED ISSUE #9240: Create specific keys for statuses (e.g. label.status-new)
## Translation Context Split Request

It would be great if ALL statuses had their own key pliz. In french what can be used in a status might not be reused somewhere else. Example the status 'New' cannot be reused in the dashboard as we need to take into account gendered and plural forms in French (yes. We. Always. Do. lol).
Same for the word 'Created' in column headers for which we wanted to use 'Creation Date' in French (but we realised the column header was also used for the status 'Created'.. and we were ending up with a status named 'Creation Date')

At the same time would suggest to rename the key status to make them easily identifiable: i.e. label.status-new instead of label.new (which can be used somewhere else) ♥️🙏

The strings can be kept as they are atm (we will update the statuses that need to be changed using Weblate) but it would be great to have specific keys for all status where they exist (inbound shipments, orders, outbound shipments, requisitions, ..)

The list of status include (hopefully didn't forget any):
- New
- Allocated
- Picked
- Shipped
- Delivered
- Received
- Verified
- Finalised
- Draft
- Sent

roxy-dao:
Fixes #9240

# 👩🏻‍💻 What does this PR do?
Seperate translation key for statuses

# 🧪 Testing

- [ ] Go to any invoice page
- [ ] Change language to french
- [ ] Statuses should mostly still be in English since translators haven't added their translations yet except for a few like new

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -2487,7 +2487,6 @@
   "status.finalise": "Finalise",
   "status.unserviceable": "Unserviceable",
   "status.verified": "Verified",
-  "status.visited": "Visited",
   "stock": "Stock",
   "stocktake.all-master-list-items": "All items (include out of stock items)",
   "stocktake.comment-and-template": "{{start}} and {{end}}",


client/packages/inventory/src/utils.ts
```diff
@@ -29,16 +28,6 @@ export const getNextStocktakeStatus = (
   return nextStatus ?? null;
 };
 
-export const getStocktakeTranslator =
```
roxy-dao: Side quest: Same as getStatusTranslation so removed and refactored

client/packages/requisitions/src/utils.ts
```diff
@@ -42,24 +42,12 @@ export const getRequisitionTranslator =
   (currentStatus: RequisitionNodeStatus): string =>
     t(getStatusTranslation(currentStatus));
 
-export const getNextRequestStatus = (
-  currentStatus: RequisitionNodeStatus
+export const getNextRequisitionStatus = (
```
roxy-dao: Side quest: Refactor to use one function to get next status

client/packages/requisitions/src/utils.ts
```diff
@@ -176,7 +163,6 @@ export const shouldDeleteLine = (
   draftId?: string,
   isDisabled?: boolean
 ): boolean => {
-  if (mode === ModalMode.Create) return true;
-  if (!draftId || isDisabled || mode === ModalMode.Update) return false;
-  return false;
+  // Only delete lines when creating new entries
+  return mode === ModalMode.Create && !!draftId && !isDisabled;
```
roxy-dao: 😉

ria8651 (approved):
Looks good to me, some good bug catches there. I can't tell if this covers everything in the issue but it covers most of it so its worth merging.

roxy-dao:
> Looks good to me, some good bug catches there. I can't tell if this covers everything in the issue but it covers most of it so its worth merging.

The encounter statuses are already covered by 'encounter-status'

PR #10219, opened 2026-01-12
Fix barcode scanning imports and Electron API null-safety

### LINKED ISSUE #10223: Coldchain scanner: avoid deep import of BarcodeFormat; Electron API should be optional
In the client, `AddFromScannerButton` deep-imports `BarcodeFormat` from `@capacitor-mlkit/barcode-scanning/dist/esm/definitions`, which is brittle (depends on internal package layout).

Also, some code assumes `window.electronNativeAPI` always exists, which can break web builds / non-electron contexts.

Proposed fix: import `BarcodeFormat` from the package root and treat `electronNativeAPI` as optional + guard access.

PR: #10219

ksawaneh:
This PR makes barcode scanning code more robust across web/Electron/native.

Changes:
- Avoid deep-importing `BarcodeFormat` from `@capacitor-mlkit/barcode-scanning/dist/...`.
- Treat `window.electronNativeAPI` as optional for web builds.
- Guard `setScannerType` against missing Electron API.
- Keep React hooks usage valid in Electron settings while handling optional API.

Checks run:
- `cd client && yarn lint-and-format`
- `cd client && yarn test`

Fixes #10223

### DIFF
--- a/client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
+++ b/client/packages/coldchain/src/Equipment/ListView/AddFromScannerButton.tsx
@@ -21,7 +21,7 @@ import {
 import { AppRoute } from '@openmsupply-client/config';
 import { useAssets } from '../api';
 import { DraftAsset } from '../types';
-import { BarcodeFormat } from '@capacitor-mlkit/barcode-scanning/dist/esm/definitions';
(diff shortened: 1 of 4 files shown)


andreievg (approved):
Thanks looks good !

PR #10221, opened 2026-01-12
Test: ignore convert_data_js in Jest to avoid haste-map collisions

### LINKED ISSUE #10225: Reports: make convert_data_js package names unique to avoid jest-haste-map collisions
Multiple `standard_reports/**/convert_data_js/package.json` files use the same `name` (e.g. `"name": "convert_data_js"`).

This can trigger `jest-haste-map` module name collisions in tooling that scans packages.

Proposed fix: give each `convert_data_js` package a unique name (e.g. derived from report/form path).

PR: #10221

ksawaneh:
This addresses `jest-haste-map` "Haste module naming collision" warnings caused by multiple `package.json` files using the same `name` ("convert_data_js") under `standard_forms/*/convert_data_js` and `standard_reports/*/convert_data_js`.

Change:
- Ignore those `convert_data_js` directories in Jest's module discovery (no package renames).

Rationale:
- Avoids adding overhead when creating/updating reports/forms.
- Removes noisy warnings from test output.

Checks run:
- `cd client && yarn test`

Fixes #10225

### DIFF
--- a/client/jest.config.js
+++ b/client/jest.config.js
@@ -20,5 +20,9 @@ module.exports = {
   transformIgnorePatterns: [
     '/node_modules/(?!uuid|@mui/x-date-pickers|@babel)',
   ],
+  modulePathIgnorePatterns: [
+    '[/\\\\]standard_reports[/\\\\].*[/\\\\]convert_data_js[/\\\\]',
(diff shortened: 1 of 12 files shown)


andreievg (approved):
Hi @ksawaneh sorry for late reply.

This looks great, thanks (although it looks like we only really support mac/linux for the regex, as per node_modules single `/`, but happy as it is now

Issue #4746, opened 2024-09-05
Update migrations <2.2.0 to use Migration Fragments


lache-melvin:
## The suggestion

Since v2.2.0 we have changed the migration strategy to [migration fragments](https://github.com/msupply-foundation/open-msupply/blob/develop/server/repository/src/migrations/README.md#migration-fragments)

Previous migrations should be updated to use migration fragments instead.

### Why should we invest time in this?

- Consistency - and clear for new devs, we only have one pattern to follow
- Then no longer need to maintain the old `migrate` method
- Breaking migrations up into fragments gives us more visibility of what has/hasn't run if a migration were to fail

### Are there any risks associated with this change?

All fragments that have run are written to the fragment log - if migrations have already run with the old pattern then need to ensure they aren't run again as fragments!

Otherwise should be low risk, we aren't changing any implementation of the migrations, just how they are run.

### How much effort is required?

Probably a couple of days to update everything?

andreievg:
> Probably a couple of days to update everything?

I think it's a few hours (2)

roxy-dao:
Closing. Earlier migration fragments should be put into init script

PR #10237, opened 2026-01-13
4746.1 update migrations 2.2.0 to use migration fragments

roxy-dao:
Part 1 of #4746

# 👩🏻‍💻 What does this PR do?
Part 1 of refactoring to use migration fragments

# 🧪 Testing

- [ ] Make sure fields still in db in their respective tables
- [ ] Tests still pass

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/migrations/v1_01_05/add_api_verson_incompatible.rs
+++ b/server/repository/src/migrations/v1_01_05/add_api_verson_incompatible.rs
@@ -1,0 +1,18 @@
+use crate::migrations::*;
+
+pub(crate) struct Migrate;
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "add_api_verson_incompatible"
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+               ALTER TYPE sync_api_error_code ADD VALUE IF NOT EXISTS 'API_VERSION_INCOMPATIBLE';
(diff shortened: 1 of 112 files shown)


server/repository/src/migrations/v1_01_05/add_api_verson_incompatible.rs
```diff
@@ -0,0 +1,18 @@
+use crate::migrations::*;
+
+pub(crate) struct Migrate;
+impl MigrationFragment for Migrate {
+    fn identifier(&self) -> &'static str {
+        "add_api_verson_incompatible"
+    }
+
```
ria8651:
```suggestion

    #[cfg(feature = "postgres")]
```

This migration needs to only run on postgres

ria8651 (changes requested):
The tests really aren't enough to validate migrations as they run without any data in the database. I found a few issues with the migrations in this PR that need fixing but didn't want to have figure out how to fix all of them. For my testing I loaded up v1.0.4, sync from a datafile with data that covers as many features as possible (to fill as many tables as possible) then migrated the database it created. I had to use rust 1.79 for it to compile. However, given the difficulty of properly validating these changes and the risk involved I'm not sure if it's worth it? Another solution is to wait until we can update the earliest sql file and drop for migrations before fragments entirely.

PR #10254, opened 2026-01-14
Fix: tolerate transient Postgres outages (retry + no panics)

### LINKED ISSUE #9574: Open mSupply service should wait if postgres connection is lost

We have had a few occasions where a postgres server goes down (in one case because the VM the pg instance was hosted on was restarted). Open mSupply Central will then shut itself down if it can no longer connect to postgres.
In one case postgres was down for less than 1 minute while the VM it was hosted on was restarted and in that time the Open mSupply service stopped 

This service shutdown is by design and generally considered good practice, however for the environments Open mSupply is operated in, it would perhaps be good if Open mSupply had a little more tolerance built in

Add a wait time of 5-10 minutes for the Open mSupply service to continue trying to reconnect to postgres before it completely shuts off.

Add a browser screen to show database is offline if a user tries to login to the server during the wait time

Possibly sending an email to support just prior to shutting down so we know that the reason it has died is due to loss of pg connection (url monitoring doesn't always tell us the Open mSupply server is down as it can often be merely an internet issue preventing the api ping). However, adding an email alert may be too much additional development when we already have other tools to assess service down

We need to test if the windows service failure settings will actually restart the service if it fails from no postgres connection. As this may be sufficient to not need to change the Open mSupply code if it works.

ksawaneh:
Fixes #9574.

What this does:
- Adds `database.connection_retry_seconds` (default 600s, `0` = fail fast) and retries the initial Postgres connect attempt for transient failures (VM restart, brief network flap, max_connections spikes).
- Prevents background tasks from panicking when the DB is temporarily unavailable (sync scheduling, file sync, ledger fix, scheduled tasks). These now log + skip and will recover once Postgres is back.
- Fixes a small test flake (unique sqlite db name) and makes `util::assert_matches!` evaluate its expression once (avoids misleading output/flaky asserts when `$left` has side effects).

Notes:
- Does not implement a dedicated “DB offline” UI page yet; this focuses on keeping the service running and recovering automatically.

Checks:
- `cd server && cargo test` (passed locally on 2026-01-14)
- `cd server && cargo check -p repository --no-default-features --features postgres` (passed locally on 2026-01-14)


mark-prins:
I updated to resolve conflicts and keep the dream alive

ria8651 (approved):
Looks good overall! Haven't been able to crash the server with your PR. On startup I think it would be better to crash immediately, should be able to get user intervention at that stage. For the long running tasks I think we still want the server to crash after say 5 mins of no postgres.

I think I'll merge this into a branch and make the changes myself instead of getting you to update a 3 month old PR 😅

PR #10261, opened 2026-01-15
update base migration database to v1.3.0

ria8651:
# 👩🏻‍💻 What does this PR do?

In #10169 the base database version was v1.0.4. It looks like there are 7 more diesel migrations that were added after v1.0.4. #10172 then removed all the diesel migrations including the ones added after v1.0.4. As such the 7 newer diesel migrations never get run causing later migrations to fail. As the last of these 7 migrations was added in v1.2.0 this version should be the base database version. It's also worth noting that even before #10169 and #10172 I had migration problems from v1.0.4. I suspect this is because all the diesel migrations are run before any rust migrations but the diesel and rust migrations need to run in lockstep. I suspect this issue wasn't caught as when doing incremental updates the migrations do run in lockstep.

This PR changes the base database version to v1.3.0 where all the diesel migrations have already been run and removes the older migrations that don't work.

# 🧪 Testing

- [ ] Initialise a v1.3.0 Open mSupply instance from mSupply
- [ ] Load the database created by v1.3.0 (try with both sqlite and postgres) in the latest version and check migrations work
- [ ] Initialise an empty database in the latest version
- [ ] Run the tests to initialise databases in every version

# 📃 Documentation

- [x] **These areas should be updated or checked**: Somewhere it should be noted that the earliest supported oms version for migrations is v1.2.0 (there will also be an error when trying to migrate from an earlier version)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/migrations/base_migration.rs
```diff
@@ -8,8 +8,8 @@ use crate::{RepositoryError, StorageConnection};
 // I've used the test db templating feature to get a `base` database to use for tests and initialisation of a new db.
 // Note: It's probably best to dump the database from the earliest version of postgres we support, but have tried to create clean up commands for dumping from v18
 
-// To refresh run tests to create template databases, then run...
-//  pg_dump -h localhost -U postgres -d ___template_1.0.4 --inserts  --no-owner --no-privileges --schema=public | sed -E '/^\\(un)?restrict [A-Za-z0-9]+$/d' | sed '/^CREATE SCHEMA public;$/d' | sed "s/SELECT pg_catalog.set_config('search_path', '', false);/SET search_path = public;/" > repository/src/base_migrations/postgres_earliest.sql
+// To refresh run tests to create template databases, then run:
+// pg_dump -h localhost -U postgres -d ___template_1.0.4 --inserts  --no-owner --no-privileges --schema=public | sed -E '/^\\(un)?restrict [A-Za-z0-9]+$/d' | sed '/^CREATE SCHEMA public;$/d' | sed "s/SELECT pg_catalog.set_config('search_path', '', false);/SET search_path = public;/" > repository/src/migrations/base_migrations/postgres_earliest.sql
```
jmbrunskill:
```suggestion
// pg_dump -h localhost -U postgres -d ___template_1.2.0 --inserts  --no-owner --no-privileges --schema=public | sed -E '/^\\(un)?restrict [A-Za-z0-9]+$/d' | sed '/^CREATE SCHEMA public;$/d' | sed "s/SELECT pg_catalog.set_config('search_path', '', false);/SET search_path = public;/" > repository/src/migrations/base_migrations/postgres_earliest.sql
```

Should be 1.2.0 now right?

PR #10280, opened 2026-01-19
5525 item_variants_configured is inefficient

### LINKED ISSUE #5525: item_variants_configured is inefficient
## The suggestion

We have a query run from the frontend to check if item_variants are configured.
```
    pub async fn item_variants_configured(
        &self,
        ctx: &Context<'_>,
        store_id: String,
    ) -> Result<bool> {
        let user = validate_auth(
            ctx,
            &ResourceAccessRequest {
                resource: Resource::QueryItems,
                store_id: Some(store_id.clone()),
            },
        )?;

        let service_provider = ctx.service_provider();
        let service_context = service_provider.context(store_id.clone(), user.user_id)?;

        let item_variants = service_provider
            .item_service
            .get_item_variants(&service_context, None, None, None)
            .map_err(StandardGraphqlError::from_list_error)?;

        Ok(item_variants.count > 0)
    }
```

This should use a db count or similar query to avoid querying all item_variants every time.

It's inefficient, espeically with a lot of item_variants which gaps does have (around 300 by default I think?)

Minimal

Maybe 1 hour to plumb a `count` query through?

roxy-dao:
Fixes #5525

# 👩🏻‍💻 What does this PR do?
Only query for first variant if it doesn't exist then item variants aren't configured

# 🧪 Testing

- [ ] Don't have item variants set up 
- [ ] Go add an item in inbound shipment -> Shouldn't see variants selection
- [ ] If you have variants then you should see the selection
- [ ] Set up 300 + variants and try loading inbound shipment (check before and after loading times.. should be faster)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/item_variant/src/lib.rs
+++ b/server/graphql/item_variant/src/lib.rs
@@ -3,6 +3,7 @@ use graphql_core::{
     standard_graphql_error::{validate_auth, StandardGraphqlError},
     ContextExt,
 };
+use repository::PaginationOption;
 use service::auth::{Resource, ResourceAccessRequest};
 
 mod mutations;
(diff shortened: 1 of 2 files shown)


jmbrunskill (approved):
Thanks! Nice lil improvement

Issue #37, opened 2022-05-04
Inbound shipment : Duplicated item_lines


mark-prins:
Raised by Bijay in https://github.com/openmsupply/openmsupply-client/issues/1206

The problem is that mSupply ends up with duplicated `[item_line]` records when the status changes from `Delivered` to `Verified`

To reproduce:

- From Web client, Click on Replenishment -> Inbound Shipment
- Create a new Inbound Shipment clicking New Shipment
- Add at least one batch line
- Confirm by changing status to 'Delivered'
- Sync and then change status to 'Verified'

The issue is that when the second sync takes place ( update status to verified ) the `item_line` updates have an incorrect `record_ID`. This means that these lines are hit: https://github.com/sussol/msupply/blob/master/Project/Sources/Methods/sync_process_sync_record.4dm#L106-L108
and the status changes from `U` to `I` which creates a new `[item_line]` for each `[stock_line]`

I cannot see where in the remote-server database the `[item_line]ID` is being retained. Perhaps it isn't? Regardless, in the update, only the `[transact]` needs updating, so perhaps limit the update to that table?

clemens-msupply:
Yes it looks like a new stock_line is generated on every update:
`service/src/invoice/inbound_shipment/update/generate.rs`

PR #10293, opened 2026-01-21
37 Dashboard plugin to highlight new Requisitions pt 2

aimee-mcneil-melville:
Part 2 of [#37](https://github.com/msupply-foundation/congo-plugins/issues/37)

# 👩🏻‍💻 What does this PR do?

Allow dashboard statistic plugins to define core stats to hide

Using a hiddenStat property, one or many core stats can be hidden when adding a plugin stat or stats

The list will be in a default order of core stats followed by plugin stats, this does not allow replacing a plugin stat in the place of a core stat

`Today` and `This week` are hidden here, and `Active`, `Inactive` are added
Screenshot 2026-01-22 at 9 29 50 AM

Have also added capability to hide any core stat, panel or widget on the dashboard if the items context is provided in a plugin component of the same type

Stats can be hidden from _any_ widget or panel, so long as the context matches in the hiddenStats array.

# 🧪 Testing

- [ ]  Install the [plugin examples](https://github.com/msupply-foundation/open-msupply-plugins/pull/19)
- [ ]  See the following stats in the Replenishment widget (compared to this example with no active plugins)
- Hidden stats: Today, This Week
- Additional stats: Active, Inactive
Screenshot 2026-01-22 at 11 20 45 AM

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/domain/Dashboard/StatsPanel.tsx
+++ b/client/packages/common/src/ui/components/domain/Dashboard/StatsPanel.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Paper, Tooltip, Typography } from '@mui/material';
+import { Paper, SxProps, Theme, Tooltip, Typography } from '@mui/material';
 import { InlineSpinner, StockIcon } from '../../../';
 import { useTranslation } from '@common/intl';
 import { ApiException, isPermissionDeniedException } from '@common/types';
@@ -15,6 +15,7 @@ export type Stat = {
   link?: string;
   extraMessage?: string;
   alertFlag?: boolean;
+  labelSx?: SxProps<Theme>;
 };
 export interface StatsPanelProps {
   error?: ApiException;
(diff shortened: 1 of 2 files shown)


client/packages/dashboard/src/DashboardService.tsx
```diff
@@ -1,13 +1,20 @@
...
   StockWidget,
 } from './widgets';
+import { useDashboardWidgets } from './utils';
 
 const Dashboard: React.FC = () => {
-  const { plugins } = usePluginProvider();
+  const coreWidgets = [
+    <ReplenishmentWidget key="replenishment" widgetContext={'replenishment'} />,
+    <DistributionWidget key="distribution" widgetContext={'distribution'} />,
+    <StockWidget key="stock" widgetContext={'stock'} />,
```
Chris-Petty: With the widget contexts I am 50/50 if it's worth making an enum of the values so that the plugins can import the enum type for a bit of type safety.

PR #10307, opened 2026-01-27
bug(10306): only comment if not a forked repo and check bundle size i…

### LINKED ISSUE #10306: In forked PRs, comment workflow breaks
## What went wrong? 😲

In a PR based off a fork (eg used by devs external to the mSupply organisation), check bundle size fails. 
The comment step requires write permissions to post PR comments. GitHub's GITHUB_TOKEN doesn't grant write access to PRs from forks for security reasons, causing the action to fail.

Example can be seen at https://github.com/msupply-foundation/open-msupply/actions/runs/20783521838/job/59687220722?pr=10192

Note: This workflow only runs when files within /client are changed

## Expected behaviour 🤔

We need a way for the workflow to pass AND for us to protect against forks increasing bundle size by __too much__
->
Proposed solution:
- Only run comment step on non-fork PRs (otherwise we'd need to give access to GITHUB_TOKEN which is a security concern)
- Continue including bundle size in json output (so it is available for contributing devs and maintainers to see)
- Add a new step just for forks to check bundle size diff - pass if below "acceptable" threshold

## How to Reproduce 🔨

1. Fork the repo
2. Make a change within client
3. Make a PR
4. See workflow fail

## Your environment 🌱
NA as occurs on github

esmehm:
Fixes #10306 

# 👩🏻‍💻 What does this PR do?

Previously, all forked repos with changes within /client would fail CI because changes within client triggered client-bundle-sie github workflow and the comment step within that failed as it required GITHUB_TOKEN which the fork would not have access to.
This PR therefore only runs the comment step if the PR is not coming from a fork. 
FOr forks, it instead checks the bundle diff and fails if the diff is above a conservative threshold (currently 3% but very open to feedback on this). The stats are still available in the workflow's uploads. This acts as a failsafe so significant bundle size increases are not accidentally allowed from forks (as we dont have the comment informing the reviewer). 
My working assumption is that large bundle size differences will primarily come from within the mSupply org and so occur on the base repo and not on forks. If this assumption is incorrect or becomes incorrect in the future then this workflow will fail and the approach should be reconsidered

This is designed to unblock external contributors and help us encourage more of them. I am very open to feedback - especially while this is my first PR and I am missing a lot of surrounding context

# 🧪 Testing

This should be tested by pushing a commit which increasing the bundle size a small amount and seeing it pass and another which increases it a large amount and seeing it fail. 

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/host/src/Host.tsx
+++ b/client/packages/host/src/Host.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import Bugsnag from '@bugsnag/js';
+// Import large library to test bundle size increase
+import * as lodash from 'lodash';
 import {
   Routes,
   Route,
@@ -42,6 +44,10 @@ import { useInitPlugins } from './useInitPlugins';
 import { ScreenOrientation } from '@capacitor/screen-orientation';
 
 const appVersion = require('../../../../package.json').version; // eslint-disable-line @typescript-eslint/no-var-requires
+
+// Test bundle size increase - using lodash to trigger significant bundle change
+const testBundleSize = lodash.debounce(() => console.log('test'), 1000);
+testBundleSize();
 
 const queryClient = new QueryClient({
   defaultOptions: {


andreievg:
As much as I like the change, I did turn of action completely for external devs that are not part of org or had explicit permissions to avoid actions misuse. 

Our devs can always approve a workflow right ? Does that make it run with our token ?

esmehm:
Closing in favour of https://github.com/msupply-foundation/open-msupply/pull/10359

PR #10312, opened 2026-01-27
10297 Performance improvements

### LINKED ISSUE #10297: Slow to load internal order list and many other pages on 2.15.2
## What went wrong? 😲

2.15.2 takes a painful time to load data compared to 2.11.3

Same datafile: 

https://github.com/user-attachments/assets/7b8978e8-0c66-457d-a4d5-0038370ead0c

## Expected behaviour 🤔

2.15.2 should be fast and fresh

## How to Reproduce 🔨

Try loading a larger datafile on 2.15.2 and compare to same datafile on 2.11.3

## Your environment 🌱

- Open mSupply Version: 2.15.2
- Platform: Android

andreievg:
Fixes #10297

# 👩🏻‍💻 What does this PR do?

* Delete prescription in footer was getting a max paginated number of prescriptions, which is quite slow for Android on large data files, just using delete prescriptions, removed delete from userPrescriptionList
* I noticed in develop we get two queries for prescription, due to useUrlQueryParams race condition in hook, this was changed to use simple useQuery sort or initial sort or default "empty" sort. @CarlosNZ can you please check useUrlQueryParams change, it should work unless our initial sort is dynamically set somehow
* The it was refactoring dashboard counts, the biggest issue was with the 'having' count clause, which froze up my local sqlite browser with the same datafile (linked in issue), changed that to distinct
* Each stats panel should have a separate query or have them all combined, I made separate queries (requisitions were combined before), while doing that decided to refactor to streamlined api hook, sorry if that pollutes the PR
* Next was improving counts performance by using inner queries rather then item ids, this required exposing item filter to stock line and consumption repositories, since consumption repository is open to plugins, had to add TS support for item filter

I also noticed we had a lot of field we don't use in prescription list, i don't think we need to select lines (and a few other things), but that's too much to refactor and performance looks adequate, equivalent to previous versions.

Also we need to think a bit more about cache key control, I think they need to be in common methods somewhere, and unique. 

I've made a branch for ease of debug build, this really needs to go in at some point: https://github.com/msupply-foundation/open-msupply/compare/10297-Performance-improvement...10297-Performance-improvement(Debug-build-for-android)

For most of the work i used, `yarn android:build:server:debug` (which is quite fast), and then deploy with android studio, and then access from my mac dev front end with `yarn start -- -- --env API_HOST='{iphere}:8000'`, i did have to go to https://{iphere}:800 first to accept certificate

I think requisition and invoice counts should really be from normal filters of invoice and requisition, with first = 0 (if possible, or 1, not selecting node and just selection totalCount), didn't want to go there though.

# 🧪 Testing

Try using 2.11 apk with data from [issue in drive](https://drive.google.com/drive/u/1/folders/19NcTReG-fO1-Uv3SMNL5sYRIg1Dw5Vv7), then update to 2.15 and see performance issues. Then upgrade to release APK in drive

Areas to try is dashboard (this will be pretty evident) and prescriptions

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
+++ b/client/packages/common/src/hooks/useUrlQuery/useUrlQueryParams.ts
@@ -71,6 +71,7 @@ export const useUrlQueryParams = ({
     if (urlQuery['sort']) return;
 
     updateQuery(initialUrlSort);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [initialUrlSort?.sort, initialUrlSort?.dir]);
 
   const updateSortQuery = useCallback(


server/repository/src/db_diesel/item.rs
```diff
@@ -1,3 +1,5 @@
+use std::time::SystemTime;
+
```
aneeshthunga26: nit: remove unused imports

server/service/src/dashboard/item_count.rs
```diff
@@ -1,3 +1,5 @@
+use std::time::SystemTime;
```
aneeshthunga26: nit: unused import?

CarlosNZ:
> * I noticed in develop we get two queries for prescription, due to useUrlQueryParams race condition in hook, this was changed to use simple useQuery sort or initial sort or default "empty" sort. @CarlosNZ can you please check useUrlQueryParams change, it should work unless our initial sort is dynamically set somehow

@andreievg -- this seems to have broken something -- the sort *direction* is not being respected, it seems to list "descending" no matter what is selected or shown.

I think you're right to remove what you did though, so I can have a look into this if you like?

andreievg:
@CarlosNZ thanks so much for checking, I've updated, simplified. We can get default/initial sort from the start, in the same shape as it would come from url query.
I think the core for my confusion was treating 'sort' and 'dir' separately, rather then one object and the use case of mapping 'asc' to empty string, I think i dealt with both (I am prepared to argue that mapping 'asc' to empty string has no negative effect for user or devs)

aneeshthunga26 (approved):
I am happy with the new changes. Looks good to me, approving.

PR #10313, opened 2026-01-27
37 dashboard plugins pt 3

aimee-mcneil-melville:
Fixes #[37](https://github.com/msupply-foundation/congo-plugins/issues/37)

# 👩🏻‍💻 What does this PR do?

Allow styling of a dashboard stat label via sx prop

Added for the purpose of a plugin stat to have customisations, however can also be used for core stat styling if needed in the future

# 🧪 Testing

- [ ] Test https://github.com/msupply-foundation/congo-plugins/pull/40

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/dashboard/src/README.md
+++ b/client/packages/dashboard/src/README.md
@@ -1,16 +1,45 @@
-## dashboard
+# OMS Dashboard
 
-### Overview
+## Overview
 
 The 'home page' of `omsupply-client`
 
-### Intentions
+The OMS dashboard consists of nested widgets, panels, and statistics to provide key information to the user.
 
-Contain a dashboard page. This involves displaying widgets from multiple different packages - rather than importing various functionality from different packages, packages export a dashboard widget, which this dashboard can import.
+## Dashboard Items Structure
 
-### Tips & Things to keep in mind
+```
+Dashboard
+├── Widget A
+│   ├── Panel A
+│   │   ├── Statistic A
+│   │   └── Statistic B
+│   └── Panel B
+│       ├── Statistic A
+│       └── Statistic B
+└── Widget B
+    └── Panel A
+        └── Statistic A
+```
 
-### Future considerations
+## Dashboard Context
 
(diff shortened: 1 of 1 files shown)


client/packages/common/src/ui/components/domain/Dashboard/StatsPanel.tsx
```diff
@@ -75,6 +77,7 @@ export const Statistic = ({
             fontSize: '12px',
             fontWeight: 500,
             marginInlineStart: '8px',
+            ...labelSx,
```
Chris-Petty: s t u n n i n g
Image

Easy to tweak!

client/packages/dashboard/src/README.md
```diff
@@ -1,16 +1,45 @@
...
+
+```
+Dashboard
+├── Widget: distribution
+│   ├── Panel: distribution-customer-requisitions
+│   │   └── Statistic: distribution-customer-requisitions-new
+└── Widget: stock
+  └── Panel: stock-summary
+    └── Statistic: stock-summary-expiring
+```
```
Chris-Petty: Just an architectural/approach thought: We've got a faiirly complex system of coding structures, exporting hidden tags, add tags as props to core components etc.

For a completely different approach, would it have been possible to make all the panel and statistic (primarily the statistic really!) components that could be imported into the plugin so that we can arrange the core components arbitrarily along side the plugin defined components? i.e. composability 🙂. Then the plugin would essentially replace the whole page but reuse the core components if desired. Doing it this way feel like it'd do away with having to managing all the hiding via filtering etc which all adds little overheads at runtime. Downside though is if you just want to change the style of one thing you have to make sure the whole dashboard is declared, but ideally doing so would be trivial 😁

Issue #20, opened 2022-05-12
Local network server


mark-prins:
Need to investigate and have a solution for running the remote-server on a LAN and connecting to it securely and easily.

Two problems:
1. Discovery. How does a client (browser and packaged app) know where the server lives?
2. Security. Local connections should be secure (for HTTPS via browser requires verified cert)

Two solutions:

1. Local network setup. Configure router to have a static IP and to resolve a domain (discovery) to that IP. Can get a cert (secure) for that domain and distribute it privately or setup cert bot to acquire it locally.
2. Package web app in somethings (electron etc..), this would allow manually setting up secrete connection or trusting self signed cert (secure) and remote server can broadcast itself and packaged app should be able to listen to the broadcast on local network (discovery)

This first and foremost needs some research, looking at the simplest solution (local network setup) as a priority and of course we can check out the other solution (but shouldn't spend too much time on it).

PR #10340, opened 2026-01-29
Hide graphs

roxy-dao:
Part of [forecasting-plugin-#20](https://github.com/msupply-foundation/open-msupply-forecasting-plugins/issues/20)

# 👩🏻‍💻 What does this PR do?

Adds the ability for plugins to selectively hide information (charts and stats) in the Request Requisition line edit view through plugin configuration (only the bottom section!!!)

Currently hiding it for all items =/ not sure if this is the correct choice to make or not, and unsure how graphs are useful to other users right now too so thought it would be okay 🤷🏻 ...

# 🧪 Testing

- [ ] Load [this PR](https://github.com/msupply-foundation/open-msupply-forecasting-plugins/pull/21)
- [ ] Graphs should no longer show in the line edit page

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/programs/src/JsonForms/common/components/CategorizationTabLayout.tsx
+++ b/client/packages/programs/src/JsonForms/common/components/CategorizationTabLayout.tsx
@@ -195,6 +195,7 @@ const UIComponent: FC<LayoutProps & AjvProps> = ({
   uischema,
   visible,
   cells,
+  config: _,
 }) => {
   const [activeCategory, setActiveCategory] = useState<number | undefined>();
   const categorization = uischema as Categorization;
--- a/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
+++ b/client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
@@ -111,6 +111,7 @@ export const RequestLineEdit = ({
             sx,
             endAdornmentOverride,
             displayVaccinesInDoses: showDoses,
+            roundUp,
           }) => (
             <ValueInfoRow
               key={label}
@@ -124,6 +125,7 @@ export const RequestLineEdit = ({
               displayVaccinesInDoses={showDoses ?? displayVaccinesInDoses}
               dosesPerUnit={currentItem?.doses}
               decimalLimit={0}
+              roundUp={roundUp}
             />
           )
         )}
--- a/client/jest.config.js
+++ b/client/jest.config.js
@@ -20,5 +20,9 @@ module.exports = {
   transformIgnorePatterns: [
     '/node_modules/(?!uuid|@mui/x-date-pickers|@babel)',
   ],
+  modulePathIgnorePatterns: [
+    '[/\\\\]standard_reports[/\\\\].*[/\\\\]convert_data_js[/\\\\]',
+    '[/\\\\]standard_forms[/\\\\].*[/\\\\]convert_data_js[/\\\\]',
+  ],
   roots: ['../client', '../standard_reports', '../standard_forms'],
 };
--- a/client/packages/common/src/intl/number/IntlNumber.ts
+++ b/client/packages/common/src/intl/number/IntlNumber.ts
@@ -62,6 +62,16 @@ export const useFormatNumber = () => {
       });
       return intl.format(newVal ?? 0);
     },
+    roundUpToWholeNumber: (value?: number): string => {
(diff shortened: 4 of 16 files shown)


client/packages/programs/src/JsonForms/common/components/CategorizationTabLayout.tsx
```diff
@@ -195,14 +195,13 @@ const UIComponent: FC<LayoutProps & AjvProps> = ({
...
   cells,
-  config,
 }) => {
   const [activeCategory, setActiveCategory] = useState<number | undefined>();
   const categorization = uischema as Categorization;
 
   const categories = categorization.elements.filter(
     (category: Category | Categorization): category is Category =>
-      isVisible(category, data, '', ajv, config) && category.type === 'Category'
+      isVisible(category, data, '', ajv) && category.type === 'Category'
```
roxy-dao: Probably caused by fixing jsonforms to version 3.5.1, and then havin' yarn.lock files updated 👀  this version only takes in 4 variables instead of 5

client/packages/requisitions/src/RequestRequisition/DetailView/utils.ts
```diff
@@ -14,12 +14,13 @@ export const buildIndicatorEditRoute = (
     .addPart(indicatorId)
     .build();
 
-export const buildItemEditRoute = (
-  requisitionId?: string,
-  itemId?: string
-) =>
+export const buildItemEditRoute = (requisitionId?: string, itemId?: string) =>
```
Chris-Petty: Can someone for the love of god write standard formatter and linter rules that everyone uses? 😆

client/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEdit.tsx
```diff
@@ -282,22 +293,24 @@ export const RequestLineEdit = ({
...
-          >
-            <StockDistribution
-              availableStockOnHand={line.itemStats?.availableStockOnHand}
-              averageMonthlyConsumption={
-                line.itemStats?.averageMonthlyConsumption
-              }
-              suggestedQuantity={line.suggestedQuantity}
-            />
-          </Box>
+          {isInfoVisible(STOCK_DISTRIBUTION_INFO) && (
```
Chris-Petty: Very minute, you could useMemo each of these boolean func return values... however `.includes()` execution might be faster than memomisation in this context given there's only a few elements to check at worst and 0 elements at best!

client/packages/requisitions/src/RequestRequisition/DetailView/utils.ts
```diff
@@ -14,12 +14,13 @@ export const buildIndicatorEditRoute = (
...
   RouteBuilder.create(AppRoute.Replenishment)
     .addPart(AppRoute.InternalOrder)
     .addPart(String(requisitionId))
     .addPart(String(itemId))
     .build();
+
+export const STOCK_DISTRIBUTION_INFO = 'stock-distribution';
+export const CONSUMPTION_HISTORY_INFO = 'consumption-history';
+export const STOCK_EVOLUTION_INFO = 'stock-evolution';
```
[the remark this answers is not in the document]
Chris-Petty: @aimee-mcneil-melville @andreievg seeing as @roxy-dao is close enough the enum camp... which we we wanna go cause we should be consistent? I like doing the enum...

Chris-Petty (changes requested):
Looks all good to me but we'd likely benefit from consensus on how to share the keys for "hiding" elements between core and plugins.

To that end I'd rather an enum over the constants over the raw strings.

client/packages/requisitions/src/RequestRequisition/DetailView/utils.ts
roxy-dao: This is from the prettier set up in the codebase 👀

client/packages/requisitions/src/RequestRequisition/DetailView/utils.ts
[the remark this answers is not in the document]
Chris-Petty: Huh true! However if you declare it as `const enum` it completely compiles the code away, so no runtime or bundle cost:

image

Chris-Petty:
Now it's your PR stung by CategorizationTabLayout 😛

PR #10345, opened 2026-01-30
10202 table error display

### LINKED ISSUE #10202: Table error state
## What went wrong? 😲

I have an error in the server and the new table gave me the following result:

Image

which is a modification of the 'Nothing here' component

Image

I have some concerns with this implemenation:

* The 'new shoot' icon is being used to display an error - the intention of the icon was to show that there are no results and that this is an opportunity for the user to add / create items. Using it as an error icon changes the intention of the icon - it would be better to have an error icon
* The text without the title row looks strange
* There is no way to display additional details about the error which is the pattern that we've been introducing with the `BoxedErrorWithDetails` and `ErrorWithDetails` components
* The change has resulted in conditional logic in the `NothingHere` component which is not necessary - I think better to use an error component 

## Expected behaviour 🤔
Can we evaluate the use of the standard `NothingHere` component as an error display state please.
Here are some suggestions using existing components:

Image

Image

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

aimee-mcneil-melville:
Fixes #10202 

# 👩🏻‍💻 What does this PR do?

Add a component to display an error message & icon, and implement it as the error fallback on tables

Consistent style with the `Nothing Here` component, but with relevant symbol and colours. I looked at using the existing ErrorWithDetails components and they seemed a bit empty in the space + extra styling needed around them anyway to center nicely

Screenshot 2026-01-30 at 3 28 05 PM

The issue mentions being able to display details of the error, I do think this is a good idea (detail is in the toast but it disappears quickly). Happy to add a tooltip or extra text in this PR or a follow up if we want this functionality in. 
To get the query error into the message will take a bit of extra work as it's currently handled at the Site level in [QueryErrorHandler.tsx](https://github.com/msupply-foundation/open-msupply/blob/e3b7b9286d01c9c275317da8988a6ca5a8c1e937/client/packages/host/src/QueryErrorHandler.tsx)

# 🧪 Testing

- [ ] Go to a list view page with a table
- [ ] Change the url - specifically misconfiguring the sort parameter or add a select filter and change that parameter
- [ ] eg /localhost:3003/inventory/stocktakes?status=FINALISD has the 'E' in FINALISED removed -> crashes
- [ ] See error component

The error page is used on many list views, and some detail views, though I couldn't trigger it on many of them. If the page refreshes, try a changing a different parameter, some crash it and some don't

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/errors/DataError.tsx
+++ b/client/packages/common/src/ui/components/errors/DataError.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import { useTranslation } from '@common/intl';
+import { useAppTheme } from '@common/styles';
+import { Box, AlertIcon, Typography, Stack } from '@openmsupply-client/common';
+
+export type DataErrorProps = {
+  error?: string;
+  title?: string;
+};
+
+export const DataError = ({ error, title }: DataErrorProps) => {
+  const t = useTranslation();
+  const theme = useAppTheme();
+  const heading = title || t('error.no-data');
+
+  return (
+    <Stack
+      flex={1}
+      justifyContent="center"
+      alignItems="center"
+      height="100%"
+      padding={1}
+      sx={{ color: theme.palette.background.error }}
+    >
+      <AlertIcon sx={{ fontSize: 120 }} />
+      <Box justifyContent="center"></Box>
+      <Typography fontSize={18} fontWeight={700} sx={{ color: 'error.main' }}>
+        {heading}
+      </Typography>
+      <Typography fontSize={14} sx={{ color: 'error.main' }} display="inline">
+        {error ? error : t('error.unable-to-load-data')}
+      </Typography>
+    </Stack>
+  );
+};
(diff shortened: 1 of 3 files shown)


client/packages/common/src/ui/components/errors/DataError.tsx
```diff
@@ -0,0 +1,35 @@
...
+      padding={1}
+      sx={{ color: theme.palette.background.error }}
+    >
+      <AlertIcon sx={{ fontSize: 120 }} />
+      <Box justifyContent="center"></Box>
+      <Typography fontSize={18} fontWeight={700} sx={{ color: 'error.main' }}>
+        {heading}
+      </Typography>
+      <Typography fontSize={14} sx={{ color: 'error.main' }} display="inline">
+        {error ? error : t('error.unable-to-load-data')}
```
aneeshthunga26: nit: This can be simplified to `error ?? t('error.unable-to-load-data')`

Issue #10328, opened 2026-01-28
Improve table filter usability


mark-prins:
## Is your feature request related to a problem? Please describe 👀
There are concerns that the change from an always-available default filter input to a small filter button is not as intuitive for users.

We need to change the filter UI so that it is obvious and easy to use.

Issues identified
* To filter by the most commonly used filters requires an additional click - to reveal the filters
* The button to enable filters is not obvious - it is small, located mid-page, and is an unexpected icon. It is also located alongside other icons and therefore does not stand out
* Partly due to the above, it isn't clear that filtering of the table data is possible

Considerations
* the previous filters took up a lot of vertical space
* it was possible to have some filters enabled by default which is helpful but also meant also that these could not be removed by the user
* adding more filters required two additional clicks

## Describe the solution you'd like 🎁

Just reinstate the previous filters - test to ensure that they work on every table and that the default filters (noted below or refer to the comparison table in [this document](https://docs.google.com/document/d/1FvTf2I1Geo6qdiqWUg6EZ8-N2smtWbjrNeQh0ABsSJU/edit?tab=t.5gd25crjhpdc)) are available

**Replenishment**
* [ ] Internal orders list
* [ ] Purchase orders list
  * Supplier 
* [ ] Inbound shipments list
* [ ] Supplier returns list

**Inventory**
* [ ] Stock list
  * Search
* [ ] Stocktakes
* [ ] Locations

**Distribution**
* [ ] Requisitions list
* [ ] Outbound shipments list
* [ ] Customer returns list

**Dispensary**
* [ ] Patients list
  * First name
  * Last name
  * Patient ID
* [ ] Prescriptions list
  * Name
  * From date
  * To date
* [ ] Encounters list

**Cold chain**
* [ ] Equipment list
  * Functional status
  * Category
  * Asset number
  * Serial number

PR #10352, opened 2026-02-02
10328 reinstate toolbar filters part 1

aimee-mcneil-melville:
Part 1 #10328 

# 👩🏻‍💻 What does this PR do?

Hide filtering options on the MRT tables - this includes hiding filter/remove filter options from the column dropdown, and hiding the icon that shows/hides filters for all columns in the table header

Screenshot 2026-02-02 at 2 51 24 PM

Reinstate filters in the Toolbar for the following List Views: 

**Replenishment**

- [x] Internal orders list
	- On simplified mobile UI: all filters are hidden
- [x] Purchase orders list
    - Default filter: Supplier 
- [x] Inbound shipments list
	- Reordered statuses to match the Inbound shipment status order
	- On simplified mobile UI: name/created date/status is replaced by a Search
- [x] Supplier returns list
	- Updated invoice status options - delivered is now received

**Inventory**

- [x] Stock list
    - Default filter: Search
    - VVM Status filter is available if Store pref `Manage VVM status for stock` is on 
- [x] Stocktakes list
- [x] Locations list

Any Search filter for numbers (eg Inbound Shipments) is an equals number search - eg typing `10` will not show results for `101`, `105` etc. It only filters when typing numbers (as the invoice number should be) - you can type letters but it doesn't do any filtering on them
Not something to address in this issue but noting it for testing

MRT filtering - intentionally just hiding this capability for now so that filters aren't competing or duplicated. TBD whether it gets removed or kept long term

# 🧪 Testing

- [x] Check each list view above for the following: 

- [x] Filters options within the table are not visible - this includes not being an option when clicking on column headers, and no filter icon in the table toolbar 
Screenshot 2026-02-02 at 3 38 26 PM

- [x] Filters menu dropdown is available in the Toolbar area
- [x] Filters can be selected from the filter menu, and each one filters by the correct information
- [x] Filter values can be cleared
- [x] If a default filter(s) is mentioned, it is visible when the page loads and cannot be removed via the menu

# 📃 Documentation

  1. Table documentation/mentions
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [x] Frontend

### DIFF
--- a/client/packages/purchasing/src/purchase_order/DetailView/DetailView.tsx
+++ b/client/packages/purchasing/src/purchase_order/DetailView/DetailView.tsx
@@ -50,6 +50,7 @@ const DetailViewInner = () => {
 
   const {
     query: { data, isFetching, isLoading },
+    lines: { filteredLines: lines },
     draft,
     handleChange,
     invalidateQueries,
@@ -62,8 +63,6 @@ const DetailViewInner = () => {
     entity: lineId,
     isOpen,
   } = useEditModal<string | null>();
-
-  const lines = React.useMemo(() => data?.lines.nodes ?? [], [data]);
 
   const onRowClick = useCallback(
     (line: PurchaseOrderLineFragment) => {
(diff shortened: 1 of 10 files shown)


aneeshthunga26 (changes requested):
I can still see filters in the column drop-downs for other languages

Image

See my other comments for a possible fix

PR #10359, opened 2026-02-02
10306 - comment using workflow run

### LINKED ISSUE #10306: In forked PRs, comment workflow breaks

esmehm:
Fixes #10306 

# 👩🏻‍💻 What does this PR do?
Fix bundle size workflow for external contributors
Problem: The bundle size check fails to post comments on PRs from external contributors because the workflow runs in the fork's context without write permissions to post comments.

Why not pull_request_target?
While pull_request_target was suggested as it runs with base repo permissions, it's a security risk. All jobs would execute with repository secrets and write access, meaning malicious PRs could run yarn install with postinstall scripts that exfiltrate secrets or compromise the repository.

Solution: Split into two workflows using the workflow_run pattern:

Build workflow (pull_request): Runs in untrusted fork context, builds bundles, uploads artifacts as JSON
Comment workflow (workflow_run): Runs in base repo context, downloads artifacts (safe static files), posts comment
This approach is secure because:

No PR code execution happens with repository permissions
Only static artifacts (JSON files) cross the trust boundary
Malicious code can't access secrets or write permissions

References for further reading:
GitHub Security Lab: [Preventing pwn requests](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
This is the recommended pattern for workflows that need to comment on external contributor PRs

Open to other ideas!

# 🧪 Testing

I cannot test from my own fork because I have permissions, so would be best tested from an external fork

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -2,6 +2,7 @@
 
 Welcome! This is the front end application for open mSupply.
 
+
 This is a multi-lingual, responsive web application making use of the open mSupply [remote-server API](https://github.com/open-msupply/server/README.md).
 
 It is built using React, with typescript, and a heavy reliance on hooks. The component framework is material-ui and we're using [storybook](https://storybook.js.org/) to demonstrate the usage and functionality of custom components.


.github/workflows/client-bundle-size-diff.yaml
```diff
@@ -91,23 +89,35 @@ jobs:
...
         uses: NejcZdovc/bundle-size-diff@v1
         with:
           base_path: "./base/stats.json"
           pr_path: "./pr/stats.json"
 
-      - name: Comment
-        uses: NejcZdovc/comment-pr@v1
+      - name: Save comparison data
+        run: |
+          echo "${{ github.event.pull_request.number }}" > pr_number
```
mark-prins: @esmehm I found having the comparison branch mentioned in the comment useful (yes, even though it is always 'main'!)

Image

also preferred the simplicity of the previous layout.. do we have to change the layout to implement the functional change?

PR #10368, opened 2026-02-03
10390: rename useIsGapsStoreOnly to useIsExtraSmallScreen

### LINKED ISSUE #10390: Rename useIsGapsStoreOnly to useIsExtraSmallScreen and isGaps to isExtraSmallScreen
## The suggestion

The hook useIsGapsStoreOnly is actually only checked the screen size breakpoint -> rename it to useIsExtraSmallScreen to better reflect the logic
Rename the uses of it (often called isGaps) to isExtraSmallScreen

## Example use case

Change `const isGaps = useIsGapsStoreOnly();` to `const isExtraSmallScreen = useIsExtraSmallScreen();`

The function itself is 
```export const useIsGapsStoreOnly = (): boolean => {
  const theme = useAppTheme();
  const isGapsStoreOnly = useMediaQuery(theme.breakpoints.down(Breakpoints.sm));

  return isGapsStoreOnly;
};
``` 

and should be renamed to useIsExtraSmallScreen so it is clear what it is doing.

This function is used for some things directly related to screen size eg row formatting and for other things like default landing page which made the assumption small screens would only be used for gaps and that assumption no longer holds with the addition of barcode scanner -> this will be addressed in follow up tickets around mobile nav bar and mobile landing pages

This ticket is for developer clarity and to enable the follow up work around mobile landing pages and mobile nbar bar preferences in #10389 and #10392 

Low risk

This change is low effort - investigating potential approaches was the majority of the work
~half a day

### Agreed Solution
Only a rename will take place for this ticket and all functionality shall remain as is

esmehm:
Fixes #10390 

# 👩🏻‍💻 What does this PR do?

Previously we had the hook `useIsGapsStoreOnly` which was reading the screen size and also had preferences about gaps store. The naming of the two was not clear when each logic was being used throughout the app.

This PR renames the hook which uses the screen breakpoint to `useIsExtraSmallScreen` for clarity and renames uses of this from `isGaps` to `isExtraSmallScreen`. -> it is clear which logic is being used. 
This logic applies to things like how the navigation bar appears for small screen

This PR is primarily intended for developer readability and clarity -> it sets us up for changing the navigation logic in the sibling issues #10389 and #10392 
However there is one place where the logic has changed and that is around the root path (both where you go when you switch store and when you first login) - I have extracted this to a new hook useRootNavigationPath and made it so
- desktop always goes to dashboard
- if mobile and gaps and central server -> manage/equipment
- If mobile and gaps and not central server -> cold-chain-equipment
- if mobile and gaps not enabled -> /replenishment/inbound-shipment
This therefore has the working assumption that desktop do not use gaps
This was done as aq uick win, and #10390 is still intended to follow it and expand the logic inside this hook

There were some other places where the small breakpoint was also being checked directly to differentiate XS screens and I have also refactored to reuse the useIsExtraSmallScren hook to reduce the number of imports and improve maintainability. 

No changfes to logic - only to clarity / readability. The necessary logic changes around nav items and landing pages will be addressed in other tickets

# 🧪 Testing

Check all behaviour on mobile looks as expected

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/ListView/ListView.tsx
+++ b/client/packages/invoices/src/InboundShipment/ListView/ListView.tsx
@@ -38,7 +38,7 @@ export const InboundListView = () => {
     queryParams: { first, offset, sortBy, filterBy },
   } = useUrlQueryParams({
     initialSort: { key: 'invoiceNumber', dir: 'desc' },
-    ...isMobile && { initialFilter: [{ id: 'status', value: 'NEW,DELIVERED,RECEIVED' }] },
+    ...isExtraSmallScreen && { initialFilter: [{ id: 'status', value: 'NEW,DELIVERED,RECEIVED' }] },
     filters: [
       { key: 'invoiceNumber', condition: 'equalTo', isNumber: true },
       { key: 'otherPartyName' },


client/packages/host/src/Site.tsx
```diff
@@ -135,15 +132,15 @@ export const Site: FC = () => {
...
                 <Box
                   flex={1}
                   display="flex"
                   flexDirection="column"
                   overflow="hidden"
                 >
-                  {isGapsStore && <MobileNavBar />}
-                  {!isGapsStore && <AppBar />}
+                  {isExtraSmallScreen && <MobileNavBar />}
+                  {!isExtraSmallScreen && <AppBar />}
```
jmbrunskill:
```suggestion
                  {isExtraSmallScreen? <MobileNavBar />: <AppBar />}
```
I feel like this might be more readable?

client/packages/host/src/Site.tsx
```diff
@@ -110,16 +110,13 @@ export const Site: FC = () => {
...
-    // but is required on small devices
-    if ((isGaps || isGapsStore) && isCentralServer) {
+    // This is a temporary solution until we consider preferences further for small devices
+    if ((isGaps || isExtraSmallScreen) && isCentralServer) {
       return RouteBuilder.create(AppRoute.Manage)
         .addPart(AppRoute.Equipment)
         .build();
     }
-    if (isGaps || isGapsStore) {
+    if (isGaps || isExtraSmallScreen) {
```
jmbrunskill: If gaps is turned off can we navigate to inboundshipment instead of equipment on extra small screens.

jmbrunskill (commented):
Just adding a couple of comments first, but overall I think it looks good.
Ironically we did used to have isExtraSmallScreen, but John removed in favour of media queries rather than hiding the logic behind a hook if that's all it;s doing. Personally I'm not worried either way...

PR #10371, opened 2026-02-03
10208 Inbound shipment (external) detail view

### LINKED ISSUE #10208: Inbound shipment (external) Detail View

Workflows
I am able to view the detail of an existing IS(E)
I can edit a IS(E) if I have permission to. The mSupply permission is Add/Edit goods received

Existing functionality 

Below is a wireframe of the detail screen:

Image

While there is an existing detail page there are some changes needed for the external version

Detail view
The following columns are shown in the table

- Code
- Item name
- Batch
- Expiry date
- Number of packs
- Pack size
- Units
- 

The following buttons are available

- Add PO Line : which allows adding a single line to the IS(E)

The following properties can be edited

- Received date
- Supplier reference
- Comment
- Donor
- Campaign
- 

The following are shown as read-only

- Shipment number
- Purchase order number
- Created by
- Created date
- Status

New functionality
The following functionality changes apply to the IS(e) version of the IS detail page.

Detail view

The supplier name, PO number and PO reference are read only

Date is filled with creation date. 
It is editable, but can’t be set in the future.

Table actions

- Delete - deletes the selected line(s)
- Approve - sets line(s) to approved status (Permissioned activity)
- Reject - sets line(s) to rejected status (Permissioned activity)
- Note there is an approver’s note if you edit an IS(E) line - if approvers want to add notes, they shouldn’t use the bulk action
- 

The following additional columns are need to be shown in the table
Approval (if the preference Inbound Shipment (External) lines must be authorised  is enabled)
Comment

The following buttons are available
Add PO Line : which allows adding a single line to the IS(E)
If the line is already on the  IS(E), it is not added and user is alerted
Print - this prints the ‘Goods received form’ which is in mSupply ([example](https://drive.google.com/open?id=1quV1pm8stwyA5PDPWMR6YK82Xr8PE-hT&usp=drive_fs)) The new form will need to have a more suitable name though such as Inbound shipment external
The form will show:

- Store name
- Supplier name
- Purchase order number
- Supplier reference
- Date created
- Date received
- Then for each item
- Location
- Item name
- Item code
- Pack size
- Quantity
- Total quantity
- And and entry field (just a line to write on) for the quantity received

The following additional properties can be edited
Received date
Campaign

The following are shown as read-only
Purchase order number (which links to the PO)
Total price for the IS(E)

Financial tab
This displays the financial details and is only visible if the user has the OG add/edit goods received permission enabled

Image

Notes
PO price per pack is read-only
Invoice price per pack is editable - the text shows in red if different to PO price
2 columns with (local) suffix only show if PO currency != local currency
Sell price is calculated based on rules for the product/supplier/item configured in OMS/OG central to determine a markup or sell price.

Currency tab
Also includes inbound shipping charges

Image

Where 
A is the sum of charges on the PO and is in the currency of the PO
B is the sum of any local charges and is in the local currency
C is the sum of all costs on the PO converted to the local currency

Read-only fields:
PO currency
Total charges: which is calculated as ( A/exchange rate )+B 
Note that exchange rate is the number of foreign currency units in the base currency: e.g. if base is NZD, USD would be about 0.6. If A is USD100 and B is NZD200, total charges are 100/0.6 + 200 = 366.67
% Cost adjustment - calculated as (A+B)/C

Currency rate is filled in when the PO is chosen for this IS(E). It is a lookup of the current currency rate, but is editable.

Delivery status tab

Image

Columns shown in the table are:
Previous deliveries: Total on previous finalised  IS(E)s belonging to this PO for this item
This Delivery: total for the current IS(E)
Total delivered: sum of the above 2
In transit: Total on IS(E)s belonging to this PO for this item that have status “sent” or “arrived” but excluding items on this IS(E).
PO quantity: adjusted quantity from PO
Remaining: PO Quantity - Total delivered - In transit

Documents tab
Allows the uploading and downloading of documents to the IS(E)

A button to upload a document is shown within the tab

Documents can be deleted and viewed / downloaded from this table

Lists documents in a table with the columns

- Filename
- Created by
- Created date

Log tab
This is a standard log tab displaying log entries showing

- Goods received created
- Goods received finalised
- Line Approvals & Rejections

ria8651:
Fixes #10208

# 👩🏻‍💻 What does this PR do?

Screenshot 2026-02-03 at 3 22 03 pm

Implements a specialised detail view for inbound shipments (external). Main changes:

- Extra toolbar inputs of the created date and purchase order number and reference
- Updated API for changing an invoice's created date to backend
- Updated API for changing an invoice line's status
- Added line status column to table
- Added bulk line status update to footer
- Wont allow changing status to received until there are no pending lines
- Don't bring lines into stock that are rejected

For deciding when to show the approval column and buttons I deviated from the spec by checking the lines for any non null statuses. If the preference is used instead you could end up with pending lines in an inbound shipment that can't be approved if the preference is turned off after making the inbound shipment.

The spec doesn't talk about whether lines should be locked after being authorised so I didn't implement anything along those lines but in order to reduce implementation complexity line status can't be changed after being brought into stock. So when received and not verified lines can still be edited, added and deleted but line status can't change meaning new lines will be stuck at pending and can't go into sock. Once more thought has been put into how line statuses should work I think this should be changed.

# 🧪 Testing

- [ ] Make an external inbound shipment from a purchase order
- [ ] Confirm the UI changes are applied for it and not a normal inbound shipment
- [ ] Check the PO number reference and date changing works
- [ ] Check the PO nubmer shows in the related documents tab
- [ ] Approve and reject some lines by selecting them and hitting approve or reject in the footer

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/invoice/inbound_shipment/update/generate.rs
+++ b/server/service/src/invoice/inbound_shipment/update/generate.rs
@@ -2,8 +2,8 @@ use chrono::{NaiveDateTime, Utc};
 
 use repository::vvm_status::vvm_status_log_row::VVMStatusLogRow;
 use repository::{
-    EqualFilter, InvoiceLineFilter, InvoiceLineRepository, InvoiceLineType, LocationMovementRow,
-    Name, RepositoryError,
+    EqualFilter, InvoiceLineFilter, InvoiceLineRepository, InvoiceLineStatus, InvoiceLineType,
+    LocationMovementRow, Name, RepositoryError,
 };
 use repository::{
     InvoiceLineRow, InvoiceLineRowRepository, InvoiceRow, InvoiceStatus, StockLineRow,
@@ -83,7 +83,6 @@ pub(crate) fn generate(
 
     // Already validated in validate
     if let Some(delivered_datetime) = patch.delivered_datetime {
-        log::info!("Setting delivered datetime to: {}", delivered_datetime);
         update_invoice.delivered_datetime = Some(NaiveDateTime::from(delivered_datetime));
     }
 
@@ -354,6 +353,10 @@ pub fn generate_lines_and_stock_lines(
 
     for invoice_line in lines.into_iter() {
         if invoice_line.number_of_packs <= 0.0 {
+            continue;
+        }
+        // Rejected lines should not create stock entries
+        if invoice_line.status == Some(InvoiceLineStatus::Rejected) {
             continue;
         }
         let mut line = invoice_line.clone();
--- a/server/service/src/invoice/inbound_shipment/update/validate.rs
+++ b/server/service/src/invoice/inbound_shipment/update/validate.rs
@@ -41,14 +41,18 @@ pub fn validate(
             },
         )?;
 
-        // If changing to Received, verify no pending lines
+        // All pending lines must be resolved (accepted or rejected) before the invoice can be
+        // received or verified, otherwise stock would be created for lines that haven't been
+        // reviewed yet.
         use UpdateInboundShipmentStatus::*;
         if matches!(patch.status, Some(Received | Verified)) {
             check_no_pending_lines(&invoice.id, connection)?;
         }
     }
 
-    // Delivered datetime check
+    // Delivered datetime is only editable for external inbound shipments (those created from a
+    // purchase order). It must not be in the future and must not be after the received datetime,
+    // as the goods can't have been delivered after they were received.
     if let Some(delivered_datetime) = patch.delivered_datetime {
         if invoice.purchase_order_id.is_none() {
             return Err(CanOnlyChangeDateOfExternalInboundShipments);
--- a/server/service/src/invoice/inbound_shipment/update/mod.rs
+++ b/server/service/src/invoice/inbound_shipment/update/mod.rs
@@ -237,8 +237,9 @@ mod test {
         test_db::setup_all_with_data,
         vvm_status::vvm_status_log::{VVMStatusLogFilter, VVMStatusLogRepository},
         ActivityLogRowRepository, ActivityLogType, EqualFilter, InvoiceLineFilter, InvoiceLineRow,
-        InvoiceLineRowRepository, InvoiceLineType, InvoiceRow, InvoiceRowRepository, InvoiceStatus,
-        NameRow, NameStoreJoinRow, StockLineRowRepository,
+        InvoiceLineRowRepository, InvoiceLineStatus, InvoiceLineType, InvoiceRow,
+        InvoiceRowRepository, InvoiceStatus, InvoiceType, NameRow, NameStoreJoinRow,
+        StockLineRowRepository,
     };
 
     use crate::{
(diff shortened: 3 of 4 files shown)


server/service/src/invoice/inbound_shipment/update/generate.rs
```diff
@@ -81,6 +81,10 @@ pub(crate) fn generate(
     update_invoice.currency_id = patch.currency_id.or(update_invoice.currency_id);
     update_invoice.currency_rate = patch.currency_rate.unwrap_or(update_invoice.currency_rate);
 
+    if let Some(created_datetime) = patch.created_datetime {
```
jmbrunskill: @ria8651 I don't think we should support changing created_datetime.

We do have a backdated_datetime we can use if you want to adjust stock movements etc.
I've asked a question in the spec with @mark-prins Sorry I didn't read this section closely enough!

client/packages/common/src/ui/layout/tables/useGetColumnDefDefaults.tsx
```diff
@@ -139,11 +138,7 @@ export const useGetColumnTypeDefaults = () => {
           enableResizing: false,
           enableColumnActions: false,
           // Comment popover is pretty narrow, show icon rather than full label
-          Header: () => (
-            <Tooltip title={t('label.comment')} placement="top">
-              <MessageSquareIcon fontSize="small" />
-            </Tooltip>
-          ),
+          Header: () => <MessageSquareIcon fontSize="small" />,
```
jmbrunskill: @ria8651 was this deliberately removed? I think the tool tip was helpful...

jmbrunskill (changes requested):
@ria8651 I'm getting an issue where approval status doesn't load on first time entering the page:

Image
If I click on Documents then click back I see it....

Image

Also I notice that the ledger is showing the rejected lines as if they were included in stock. 

You can see Amoxicilllin has been rejected so shouldn't be in stock but it shows in the stock view.

Image

That is just waiting on this PR right? **https://github.com/msupply-foundation/open-msupply/pull/10450**

client/packages/invoices/src/InboundShipment/DetailView/SidePanel/RelatedDocumentsSection.tsx
```diff
@@ -34,26 +35,42 @@ export const RelatedDocumentsSectionComponent = () => {
       username: user?.username ?? UNDEFINED_STRING_VALUE,
     })}`;
   }
+  const showRequisition = requisition && !orderedFromDifferentStore;
 
   return (
     <DetailPanelSection title={t('heading.related-documents')}>
       <Grid flexDirection="column" gap={0.5}>
-        {!requisition || orderedFromDifferentStore ? (
+        {!showRequisition && !purchaseOrder ? (
```
jmbrunskill: I think it might be more maintainable & I think fine to always show this `Related Documents` panel.
If there's nothing it just says there's nothing right? Seems fine?

Not neccesary change but I think might be worth keeping it simple.

client/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -130,7 +129,9 @@ const DetailViewInner = () => {
     setMode(ModalMode.Update);
   };
 
-  const columns = useInboundShipmentColumns();
+  const external = data?.purchaseOrder !== null;
+  const showLineStatus = data?.lines.nodes.some(line => line.status != null) ?? false;
```
jmbrunskill: I'm guessing this is the one that needs fixing to get it to load correctly? We could just the pref here but I do like this as it should nicely navigate between before and/or after the pref for authorisation is added.

However, It might be better to just always show the column if the authorise pref is turned on?
Then we can show a default status for null something like `Authorisation Not Required`

That would leave the issue if the pref is turned off but we have Pending lines, so we probably still need this check.

@mark-prins do you prefer this column to always show if we've turned on the authorise lines pref, even for older lines that don't need authorisation?

server/service/src/invoice_line/stock_in_line/insert/generate.rs
```diff
@@ -33,8 +35,17 @@ pub fn generate(
...
+        ExternalInboundShipmentLinesMustBeAuthorised
+            .load(connection, Some(existing_invoice_row.store_id.clone()))
+            .unwrap_or(false);
+
+    let mut new_line = generate_line(
+        input.clone(),
+        item_row,
+        existing_invoice_row.clone(),
+        external_inbound_shipment_lines_must_be_authorised,
+    ); // include vvm status here
```
jmbrunskill: Is this vvm status line still required? Should be followed up on or removed? I know you didn't add it but it I wonder if we need to follow to create an issue or something? Do you mind investigating a little bit?

ria8651: The VVM status is passed through from the request so I think the comment isn't needed anymore. I've removed it

ria8651:
I've resolved most of the comments but left the ones with design decisions.

server/service/src/invoice/inbound_shipment/update/validate.rs
```diff
@@ -74,12 +96,7 @@ pub fn validate(
         return Err(CannotIssueForeignCurrencyForInternalSuppliers);
     }
 
-    // Delivered datetime check
-    if patch.delivered_datetime.is_some() {
-        if invoice.purchase_order_id.is_none() {
-            return Err(CanOnlyChangeDateOfExternalInboundShipments);
-        }
-    }
+    // Don't put validation here, there is an early return above
```
jmbrunskill: Thank you!

jmbrunskill (approved):
Thanks @ria8651 much better comments now, and the tests are really good for verifying the expectations.

Issue #10213, opened 2026-01-09
Barcode Scanner Configuration


jmbrunskill:
## Describe the solution you'd like 🎁

Scanner configuration
Workflows
On a Honeywell handheld scanner device I want to use the built-in barcode scanner
Existing functionality
Barcode scanners are configurable in the desktop version. On android we use the device camera.

New functionality
Integrate the scanner plugin
For the zapp app we have made use of the Honeywell capacitor plugin.

This will need to be added to the Open mSupply android package.

Scanner selection
In the Admin > Devices section add the configuration option to select between using the camera and the built-in barcode scanner.

### Acceptance Criteria ✅

- [ ] honeywell scanner is included in the apk build
- [ ] Configuration to select the scanner or maybe auto detect honeywell scanner exists (display in configuration page?)

### Moneyworks Jobcode 🧰
OMS:NGUCF

PR #10393, opened 2026-02-04
#10213 Handle barcode input (Inbound shipment)

CarlosNZ:
Part of #10213 

# 👩🏻‍💻 What does this PR do?

Adds functionality to input data directly from scanner on an inbound shipment. Uses a separate, simplified modal with only the current input line shown:

Screenshot 2026-02-05 at 12 14 09 PM

Access to all batches for an can be done via the existing edit line modal, by clicking on a line in the Shipment -- this new modal is purely for scanned input.

There are four possible states the form can be in, reflected in the different "Message" alert boxes:
1. Scanned barcode is not recognized, and is not valid GS1, so no barcode data will be kept after input
2. Scanned barcode is not recognized, but IS valid GS1, so use must select an item, and this item will be associated with the barcode's GTIN for all future inputs
3. Barcode is associated with an existing item, and the batch/packsize matches a line already in the invoice -- the input quantity will be *added* to the existing line
4. Barcode is associated with an item, but batch or packsize is different to existing lines -- a new line will be created

Additional changes:
- modified the BarcodeScannerContext to decouple "startListening" from the "registerCallback" functionality. The "handleScan" callback can now be registered by passing it directly to `useBarcodeScannerContext`. This means the component that starts the scanner listening (e.g. the Scan button) doesn't have to also hold the definition for the scan handler function.
- consolidated `AddFromScannerButton` to a single general component and used this for Inbounds and Equipment (which previously had their own version with mostly duplicated code)

See comments

# 🧪 Testing

Either enable the mock scanner (settings), or use an actual scanning device
- [ ] Going to an Inbound Shipment (Detail) page automatically sets the scanner in "listening" mode ("Ready to scan" button active, pop-up shows if mock scanner in use)
- [ ] Listening can be togged off/on by clicking the "Ready to scan" button
- [ ] Scan a barcode for a known item, complete with batch, expiry date, pack size and quantity included -- Scan input modal should pop up with all this data populated
- [ ] Scan an unknown GS1 barcode. You should be required to select an item, after which the barcode should be recognized next time
- [ ] Scan a barcode with just GTIN data. Pack size should default to the item's default pack size
- [ ] *With the input modal already open*, scan a barcode that only has additional data (quantity, expiry, etc). This should *update* the modal with the new values while leaving the item selection (and other unchanged values) alone
- [ ] Check the "Alert" messages are all correct for each situation and makes sense
- [ ] Submitting the form adds or updates the invoice line
  - [ ] When adding a new batch, it creates a new line
  - [ ] When adding to an existing batch, the quantity should be *added* to the existing line
- [ ] Try and find other ways to break the functionality

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/utils/BarcodeScannerContext.tsx
+++ b/client/packages/common/src/utils/BarcodeScannerContext.tsx
@@ -307,7 +307,6 @@ export const BarcodeScannerProvider: FC<PropsWithChildrenOnly> = ({
   const stopScan = useCallback(async () => {
     setHideApp(false);
     setIsListening(false);
-    callbackRef.current = null;
     if (mockScannerEnabled) {
       await MockScanner.stopListening();
     }
@@ -382,6 +381,11 @@ export const BarcodeScannerProvider: FC<PropsWithChildrenOnly> = ({
           if (data && 'barcode' in data) {
             if (callbackRef.current) {
               callbackRef.current(parseResult(data.barcode));
+            } else {
+              console.error(
+                'No scan callback registered to handle barcode:',
+                data.barcode
+              );
             }
           } else if (data && 'error' in data) {
             console.error('Honeywell scanning error:', data.error);
@@ -404,6 +408,11 @@ export const BarcodeScannerProvider: FC<PropsWithChildrenOnly> = ({
           const barcode = parseBarcodeData(data);
           if (callbackRef.current) {
             callbackRef.current(parseResult(barcode));
+          } else {
+            console.error(
+              'No scan callback registered to handle barcode:',
+              barcode
+            );
           }
         });
       } catch (e) {
@@ -423,6 +432,11 @@ export const BarcodeScannerProvider: FC<PropsWithChildrenOnly> = ({
         const result = parseResult(barcode);
         if (callbackRef.current) {
           callbackRef.current(result);
+        } else {
+          console.error(
+            'No scan callback registered to handle barcode:',
+            result
+          );
         }
       };
       await MockScanner.startListening(scanHandler);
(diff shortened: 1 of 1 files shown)


jmbrunskill:
> Hey @CarlosNZ , @jmbrunskill - has any thought been given for how barcodes relate to item variants?
> 
> The concept of a variant basically maps to a GS1 barcode
> 
> E.g.
> 
> > Scan a barcode for a known item, complete with batch, expiry date, pack size and quantity included -- Scan input modal should pop up with all this data populated
> 
> Ideally this would select the appropriate variant (which includes packsize) and populate the batch, expiry and quantity
> 
> > Scan an unknown GS1 barcode. You should be required to select an item, after which the barcode should be recognized next time
> 
> Ideally this would create a new variant and the GS1 stored against it
> 
> > Scan a barcode with just GTIN data. Pack size should default to the item's default pack size
> 
> If the variant for this GTIN exists then the correct packsize would be retrieved, no need for guessing default. Otherwise prompt to create variant.

We haven't considered this yet. Right now item variant isn't even recorded against the barcode.

adamdewey:
> We haven't considered this yet. Right now item variant isn't even recorded against the barcode.

Hmm, yes this is where I think we have conceptual issues - the barcode should be recorded against the variant, not the other way around :)

jmbrunskill:
@CarlosNZ I found I can't scan on the tablet view. Will create a Branch with a fix shortly.

PR #10407, opened 2026-02-09
move the border of vax card groups

### LINKED ISSUE #10402: [Vax card] Visual table grouping not applied correctly for doses
## What went wrong? 😲

I have an immunization program set up with the following 3 doses to be administered at birth. 

Each has the same `From age`, `To age` and `Custom age label`

Image

Image

Image

But the doses are then visually grouped incorrectly in the table of the patient vaccination card: 

Image

Image

## Expected behaviour 🤔

The table should behave as a normal table, with the box in the left column mapping to the boxes in the right column

BCG 1, HepB1 and bOPV0 should all be grouped in `Birth` 

For example: 

Image

## How to Reproduce 🔨

1. Configure vax card using schemas in https://github.com/msupply-foundation/programschemas/tree/main/immunisation_program
2. Set up an immunization program with 3 different vaccines with same from, to and custom age label as above
3. View patient vaccination card
## Your environment 🌱

- Open mSupply Version: 2.15.03

mark-prins:
Fixes #10402

# 👩🏻‍💻 What does this PR do?
The table cell bottom border was previously hidden if the row's Age is the same as the previous row's Age.
This will show ages grouped incorrectly though - the bottom border should be hidden if the next row is the same Age value / part of the current group. It should show if the next row is a new group. when the next row differs, we want to show a border which indicates that a new group is coming up.

Note that in the table display options, we add a bottom border only - so the adjustment in this table does have to be for the bottom border

This PR moves the border hiding to group correctly - like this:

Screenshot 2026-02-09 at 4 35 19 PM

# 🧪 Testing

- [ ] Create more than one vaccine course and create Doses for these courses which are the same. check the grouping is correct

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
+++ b/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
@@ -295,7 +295,8 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
                 borderRadius: '8px',
               }
             : {
-                borderBottom: '1px solid rgba(224, 224, 224, 1)',
+                borderBottom: '1px solid',
+                borderColor: 'border',
               }),
         },
       };
--- a/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
+++ b/client/packages/system/src/Vaccination/Components/VaccineCardTable.tsx
@@ -74,6 +74,13 @@ export const VaccineCardTable = ({
     const index = data?.items.findIndex(item => item.id === row.id) ?? 0;
     return row.minAgeMonths === data?.items?.[index - 1]?.minAgeMonths;
   };
+  const isAgeSameAsNextRow = (row: VaccinationCardItemFragment) => {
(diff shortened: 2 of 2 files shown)


client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
```diff
@@ -295,7 +295,8 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
                 borderRadius: '8px',
               }
             : {
-                borderBottom: '1px solid rgba(224, 224, 224, 1)',
+                borderBottom: '1px solid',
+                borderColor: 'border',
```
mark-prins: offroad: wasn't keen on the hard coded colours in this file so I've changed one of them

PR #10408, opened 2026-02-09
feat(migrations): Show migrations in progress message in client (#4745)

### LINKED ISSUE #4745: Indicate that server is unavailable due to migrations being in progress

If a migration touches a lot of data it could take a long time to run, and the web server is down during this time. As far as the user is concerned, the service isn't responding - they could hard reset/restart everything and screw up the data rather impressively 😱 

Of particular concern when running as windows service

There should be some kind of indication to the user when the server is down due to migrations in progress.

aneeshthunga26:
Fixes #4745 

# 👩🏻‍💻 What does this PR do?

Shows a migration indication to the end user when migrations are running in the server

image

- Introduces caching in KeyValueStore
- Token secret and saving log level to database will not be persisted on initial server run if key_value_store table does not exist (new setup)
- GraphQL now has 3 schemas — MigratingDatabase, Initialisation, Operational
- Actix server starts up before migration and sync initialisation
- GraphQL schema goes from MigratingDatabase -> Initialisation -> Operational

# 🧪 Testing

- [ ] Use an older datafile like 2.11 or older and migrate using this branch. Open the client as soon as server starts up and view the migrating message
- [ ] Run the server without having to upgrade — verify if normal behaviour is maintained
- [ ] Verify with different devices 

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -209,7 +209,7 @@
   "customers": "Customers",
   "dashboard": "Dashboard",
   "description.already-issued": "The number of units already issued in shipments.",
-  "migration-info.migrations-in-progress": "Database migrations in progress",
+  "migration-info.migrations-in-progress": "Server startup in progress",
   "description.annual-target-doses-calculation": "Target Population × Number of Doses × (Coverage Rate / 100) × Loss Factor",
   "description.available-months-of-stock": "Available months of stock",
   "description.available-soh": "Available stock. $t(description.doses-quantity)",


server/graphql/lib.rs
```diff
@@ -305,8 +310,9 @@ impl Mutations {
 pub struct GraphqlSchema {
     pub(crate) operational: OperationalSchema,
     initialisation: InitialisationSchema,
+    migration: MigrationSchema,
     /// Set on startup based on InitialisationStatus and then updated via SiteIsInitialisedCallback after initialisation
-    is_operational: RwLock<bool>,
+    operational_status: Data<RwLock<OperationalStatus>>,
```
aneeshthunga26: moving from bool to enum to capture new MigratingDatabase state

server/graphql/lib.rs
```diff
@@ -365,27 +371,43 @@ impl GraphqlSchema {
...
-            self.operational.execute(req.data(user_data)).await
-        } else {
-            self.initialisation.execute(req).await
+        match &*self.operational_status.read().await {
+            OperationalStatus::Operational => {
+                // auth_data is only available in schema in operational mode
+                let user_data = auth_data_from_request(&http_req);
+                self.operational.execute(req.data(user_data)).await
+            }
+            OperationalStatus::MigratingDatabase => self.migration.execute(req).await,
```
aneeshthunga26: Minimal schema for MigratingDatabase status

server/server/src/configuration.rs
```diff
@@ -205,8 +205,8 @@ impl From<VarError> for SettingsError {
 pub fn get_or_create_token_secret(connection: &StorageConnection) -> String {
     //Get Token Secret from DB if available
     let kv_repo = KeyValueStoreRepository::new(connection);
-    let token_secret_option = kv_repo.get_string(KeyType::SettingsTokenSecret).unwrap();
-    if let Some(token_secret) = token_secret_option {
+    let token_secret_option = kv_repo.get_string(KeyType::SettingsTokenSecret);
+    if let Ok(Some(token_secret)) = token_secret_option {
```
aneeshthunga26: Previously only case being handled was if there was no value stored in db for token_secret. Now handling the case where the table doesn't exist (since this happens before migration)

server/graphql/general/src/lib.rs
```diff
@@ -635,6 +646,17 @@ impl InitialisationMutations {
     }
 }
 
+/// Auth is not checked during migration stage
+#[derive(Default, Clone)]
+pub struct MigrationQueries;
```
jmbrunskill: I wonder if it would be better to call this something like `Startup` or similar.
I feels like we're being a bit too specific targeting migrations. For example we also do things like inserting standard reports at startup. I just wonder if theres a more generic term we should use that we can extent with other server status in the future? This is fine though as we can do that change later...

server/graphql/general/src/queries/migration_status.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(SimpleObject)]
+pub struct MigrationStatusNode {
+    pub in_progress: bool,
+    pub version: Option<String>,
+}
+
+pub(crate) async fn migration_status(ctx: &Context<'_>) -> Result<MigrationStatusNode> {
+    // Get the migration status from context
+    let in_progress = if let Some(status) = ctx.data_opt::<Data<RwLock<OperationalStatus>>>() {
```
andreievg: So far we've been using helpers to access Data as per https://github.com/msupply-foundation/open-msupply/blob/37764a99b3d3c80b7504faac05c8a18e7bbde59a/server/graphql/core/src/lib.rs#L59-L61, I think good to continue (it's unchecked but would only be missing if it's not bound correctly and would be visible straight away), also for locks, it's nice to have an abstraction

server/graphql/lib.rs
```diff
@@ -365,27 +371,43 @@ impl GraphqlSchema {
         .data(service_provider.clone())
         .extension(GraphQLRequestLogger);
 
+        // Migration schema shows migration status
+        let operational_status_ref = Data::new(RwLock::new(operational_status.clone()));
```
andreievg: new endpoint is available in other schemas, I think we should add this one to other schemas too

server/service/src/log_service.rs
```diff
@@ -91,7 +91,11 @@ pub trait LogServiceTrait: Send + Sync {
             Level::Trace => "trace",
         };
 
-        key_value_store.set_string(KeyType::LogLevel, Some(log_level.to_string()))?;
+        if let Err(e) = key_value_store.set_string(KeyType::LogLevel, Some(log_level.to_string())) {
+            log::warn!(
```
andreievg: Some of these methods don't need to return result anymore

server/server/src/lib.rs
```diff
@@ -348,10 +251,6 @@ pub async fn start_server(
     // START SERVER
     info!("Initialising http server..",);
```
andreievg: Even though these won't run until they are `driven`, they don't need to be initialised here and I think it's better if they are moved after migration, closer to tokio select

client/packages/common/src/api/hooks/useMigrationStatus.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Hook to query database migration status from the server.
+ * @param refetchInterval - Interval in ms to refetch, or false to disable polling
+ * @param suspense - true to enable react suspense mode, false otherwise
+ * @returns Query result with migration status (inProgress and version)
+ */
+export const useMigrationStatus = (
+  refetchInterval: number | false = false,
+  suspense = false
```
andreievg: when is suspense not true ? also when is refetchInterval not a number ?

aneeshthunga26:
> > I have addressed most of the comments.
> 
> Thanks @aneeshthunga26. Can you mentioned which ones were not dealt with (we typically try to make it super easy for reviewer, by replying on every single comment, otherwise reviewer has to look at commits and comments and trying to determine what's been address)

Apologies. I meant I addressed all others apart from the one I mentioned in my comment — the renaming MigrationQueries to something else

andreievg:
Thanks @aneeshthunga26, after our conversation:

> Use a datafile that requires initial sync — does not show migrating message

This one may lead to confusion from QA

* Would be great to make an issue for the things i've mentioned in the original PR comment
* We should bind the data for operations status to all schemas if using unchecked
* Would be great to save the token to database after migration

Issue #10372, opened 2026-02-03
UI improvements for the variants selection in ST


mlenignol:
## What went wrong? 😲

Please flag if I should open 3 different issues :)

When using variants in the stocktake there are a couple of annoying UI things :(
- Variant name is not truncated when too long and overlaps the field

<img width="1245" height="631" alt="Image" src="https://github.com/user-attachments/assets/eb80abde-0917-4b80-9be3-7d74460fa5c3" />

- Confusion between the VVM status between variant (actually VVM type) and the stock line

<img width="1040" height="491" alt="Image" src="https://github.com/user-attachments/assets/bd0cd021-01c0-41e1-a622-41e9685df864" />

Note: it is VVM type on the variant tab (in Item):

<img width="1019" height="527" alt="Image" src="https://github.com/user-attachments/assets/8a6be635-2a74-44a2-905b-4a727a042232" />

And overall the interface when choosing the variant is a bit buggy, it is slow, not really practical when filling multiple lines and needing to close the window everytime or wait (maybe a drop down would suit better?)

https://github.com/user-attachments/assets/1e632fe6-1446-4bdd-9793-8cb2f13b774b

## Expected behaviour 🤔

- Fit between the field and the variant name
- VVM Type for the variant
- UI tweak for the variant selection

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to New Stock Take
2. Click on a Line for an item with a variant
3. Add or update a variant

## Your environment 🌱

- Open mSupply Version: 2.15.02 (https://tchad.msupply.org)

Tagging @richardmoizeau

roxy-dao:
VVM status has already been renamed to VVM type (: rest is valid

Issue #10331, opened 2026-01-28
Update table icons


mark-prins:
## Is your feature request related to a problem? Please describe 👀
This is part of a larger request - to update our icon set from the current, [feather icons](https://feathericons.com), to [hugeicons](https://hugeicons.com/icons?style=Stroke&type=Rounded)

( feather isn't being actively updated and is a small icon set, we've had to take random icons from noun project for some time to fill the gaps. changing to a full icon set will improve consistency in our iconography )

## Describe the solution you'd like 🎁

Update the filter/search, density, show/hide columns, refresh and full screen as per the mock up below:

<img width="2282" height="1062" alt="Image" src="https://github.com/user-attachments/assets/d389abe5-2cf5-4e6c-8dc1-77c7ff6b0cfb" />

jmbrunskill:
Closing — this is specific to the current front end, which the rewrite replaces. Part of the icon set replacement tracked in #10378. These specific icons are slots on the old table library, which the new front end doesn't use, and the prescribed icon set has already been adopted in the rewrite.

If the same problem turns up in the new front end, please raise it fresh — the screen this was reported against won't be there.

If we've got this wrong, just reopen it!

<sub>Part of the Sep 2026 backlog review (698 open issues), labelled `x. closed - Sep 2026 review` so the whole set stays findable and reversible. The review used automated code analysis — mistakes are possible.</sub>

PR #10415, opened 2026-02-09
10327 table settings icon pt2

### LINKED ISSUE #10327: Add table settings icon
Similar to Marie's example:

Image

Add a settings icon and add to it
* Unpin all columns
* Reset column order
* Show all columns
* Table density
* Reset table to defaults

Additional request: on mobile if you long press a column header it shows `Column Actions` can this instead please show the  column label

as above

aimee-mcneil-melville:
Fixes #10327

# 👩🏻‍💻 What does this PR do?

Add a settings menu to tables with options to reset parts or all of the table display. 

- Options are disabled if there is no saved state for that option
- Toggle density is always available as it cycles through the three density options (does not auto close)
- Reset table to defaults button prompts a confirmation modal to confirm that the user wants to reset all settings

This mimics the column options menu, including using the mui icon set to match icons to ones that are used elsewhere in the table. Icons can be easily updated with the new icon set in #10372 and #10331

Table with display changes made -> table at default (reset options disabled)

Screenshot 2026-02-10 at 8 30 48 AM
Screenshot 2026-02-10 at 8 30 56 AM

No all tables have the display customisations enabled (vax card, line edit view tables etc). It looks as though these cases also have the top toolbar turned off - this should be ok as the settings menu won't render & isn't needed as the tables don't allow customisations. 
There may be tables in future where one or all of the options in the settings menu is not available as an action on the table. In that case, the option should remain disabled as the local state won't be changed. They could potentially be changed to hide if the table doesn't allow that action, but probably not necessary for now

Resetting state for pinned columns resets to any columns that are pinned by default. The issue mentions 'unpin all columns', and this action is available in the column menu, but I realised that on some tables 'default' and 'unpin all' are not the same (the select column is always excluded from this/is always pinned and cannot be changed). Eg Internal orders detail view has some columns pinned by default
 

# 🧪 Testing

- [ ] Go to a line or detail view table - should see a toolbar with a settings icon (not all tables have one)
- [ ] Open settings menu -> see all disabled except toggle density
- [ ] See toggle density changes between three density settings in a loop + icon reflects the change
- [ ] Test changing each setting in the table -> see option is now enabled
- [ ] Resetting the option will reset only that one, not other settings 
- [ ] Test changing multiple settings and then resetting the whole table - all will revert to default settings after selecting Ok in the confirmation

# 📃 Documentation

  1. Table documentation (internal and/or public)
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
+++ b/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
@@ -86,8 +86,15 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
         ...internalColumnMenuItems,
       ];
     },
+    // Use column header as action button tooltip
+    muiColumnActionsButtonProps: ({ column }) => {
+      const { header } = column.columnDef as ColumnDef<T>;
+      return {
+        title: header,
+      };
+    },
 
-    // Add reset state button to toolbar
+    // Add grouping toggle and settings menu to toolbar
     renderToolbarInternalActions: ({ table }) => (
       <>
         {toggleGrouped && !isMobile && (


client/packages/common/src/ui/layout/tables/components/SettingsMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+      return <DensityLargeIcon fontSize="small" />;
+    } else {
+      return <DensityMediumIcon fontSize="small" />;
+    }
+  };
+
+  return (
+    <>
+      <IconButton
+        icon={<SettingsIcon />}
```
mark-prins:
```suggestion
        icon={<SettingsIcon fontSize="small" />}
```

the icon looks too big - inconsistent with the other icons:

Image

with change:

Image

PR #10425, opened 2026-02-10
Use MRT built in column grouping

### LINKED ISSUE #10505: Tables: group by item icon issues
## What went wrong? 😲

1. When the table density is set to the lowest setting (e.g. least dense), group by item icon is cut off and cannot be adjusted to be showing fully

Image

2. When manually ungrouping items (either one or all), the icon does not update

https://github.com/user-attachments/assets/ff096415-b83a-4433-bc68-6f8e4a3df4c7

## Expected behaviour 🤔

Should be visible and should update if manual grouping/ungrouping has been done

## How to Reproduce 🔨

1. Go to Inventory -> Stocktakes and generate a stocktake
2. Make sure you have items included with multiple batches
3. Group items and set the table to the lowest density -> icon does not show fully
4. Manually ungroup/group items -> see the icon does not update

## Your environment 🌱

- Open mSupply Version: `V2.16.01-develop-02121710`
- Legacy mSupply Central Server Version: `V8.06.12`

- Database type: `PostgreSQL`

ria8651:
# 👩🏻‍💻 What does this PR do?

Uses the built in column grouping of Material React Table instead of our custom grouping logic. Looks almost the same as before but with a nicer API. I also fixed #10505, cleaned up the padding at different densities, made the expand arrow rotate in the right direction and locked the expand column (#10525).

Screenshot 2026-02-19 at 2 19 51 pm

<details><summary>Old screenshots</summary>
<p>

Screenshot 2026-02-10 at 4 11 40 pm

image

</p>
</details> 

# 🧪 Testing

- [ ] There's quite a few affected tables due to the api changes, basically every table with grouping. For at least some of them check:
  - [ ] Grouping works and is saved in local storage
  - [ ] Aggregation works as expected

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/useMaterialTableColumns.tsx
+++ b/client/packages/common/src/ui/layout/tables/useMaterialTableColumns.tsx
@@ -120,7 +120,7 @@ export const useMaterialTableColumns = <T extends MRT_RowData>(
       });
 
     return { columns };
-  }, [getColumnTypeDefaults, omsColumns, t]);
+  }, [omsColumns]);
 
   return tableDefinition;
 };


ria8651:
As far as I can tell it was a bug in tanstack as it happens from just having grouping without sorting in the table. Fixed it by deduping the rows in the row model.

client/packages/common/src/ui/layout/tables/useBaseMaterialTable.tsx
```diff
@@ -188,14 +165,52 @@ export const useBaseMaterialTable = <T extends MRT_RowData>({
...
+        rows.push(row);
+
+        if (row.subRows?.length > 1 && row.getIsExpanded()) {
+          row.subRows.forEach(handleRow);
+        }
+      };
+
+      rowModel.rows.forEach(handleRow);
+
+      // We can't pass rowModel.flatRows directly as for some reason rows come in duplicated when there's grouping and no sorting applied
```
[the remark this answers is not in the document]
esmehm: Hmm 
Both the supplier return and customer return does have an initial sort
```
export const CustomerReturnListView = () => {
  const t = useTranslation();
  const {
    filter,
    queryParams: { sortBy, first, offset },
  } = useUrlQueryParams({
    initialSort: { key: 'createdDatetime', dir: 'desc' },
    filters: [
      { key: 'otherPartyName' },
      { key: 'status', condition: 'equalTo' },
    ],
  });
  ```
https://vscode.dev/github/msupply-foundation/open-msupply/blob/use-mrt-column-grouping/client/packages/invoices/src/Returns/CustomerListView/ListView.tsx#L32-L33

File: client/packages/common/src/ui/layout/tables/useMaterialTableColumns.tsx [hunk 2]
```diff
@@ -63,9 +72,56 @@ export const useMaterialTableColumns = <T extends MRT_RowData>(
...
+          columnId,
+          _leafRows,
+          childRows
+        ) => {
+          // if all child rows have the same value, return that value, otherwise return '[multiple]'
+          const firstValue = childRows[0]?.getValue(columnId);
+          if (childRows.every(row => row.getValue(columnId) === firstValue)) {
+            return firstValue;
+          }
+          return '[multiple]';
```
CarlosNZ: Should localise this value -- key already exists: `mutliple`

client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
```diff
@@ -167,6 +187,15 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
...
         justifyContent: 'space-between',
         opacity: 1,
+        padding: padding(table.getState().density, column.getSize()),
+        // needed to get resize handle to show in the right place (idk why it's hard coded in MRT)
+        paddingRight: column.getCanResize()
+          ? table.getState().density === 'compact'
+            ? '8px'
+            : table.getState().density === 'comfortable'
+              ? '16px'
+              : '24px'
```
CarlosNZ: Can you explain what difference this makes? I can't quite see it in the UI.

CarlosNZ: Also -- can we use `rem`s like used above?

client/packages/common/src/ui/layout/tables/aggregationFns.ts
```diff
@@ -0,0 +1,19 @@
+import { MRT_RowData } from "material-react-table";
+import { Row } from "@tanstack/table-core"; // MRT_Row doesn't work, is this fine?
+
+export const weightedAverage = <T extends MRT_RowData & { packSize?: number; numberOfPacks?: number }>(
```
CarlosNZ: Also, wondering why we'd want to use the weighted average for the cost columns, as opposed to just showing `[multiple]` like the other columns do?

I think it's possibly a bit confusing without any UI indicating it's an average, and arguably not that helpful.

If you think it's important, I'd create an issue to suggest it, but I don't think we should add it without running it by project team.

CarlosNZ (changes requested):
Looking good thanks, @ria8651, much cleaner overall.

I have a few questions/concerns/comments, as well as:

Image

- I think we should remove the option to "Group by" in individual columns, as it's confusing and kind of meaningless for columns like this. The single "Group by" buttom in the Toolbar is adequate, I'd say, and I think we want to maintain existing functionality, not add more for now.

- There's a small bug -- with an unmodified table, if you click the "Group By" button, then `columnOrder` is saved to local storage as well, and it remains after "ungrouping". You have to reset column order to get rid of it.

Image

client/packages/common/src/ui/layout/tables/aggregationFns.ts
ria8651: This was how it was done previously, didn't want to change things.

client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
```diff
@@ -238,6 +257,7 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
         },
         sx: {
           backgroundColor: 'inherit',
+          minHeight: table.getState().density === 'compact' ? '32px' : '40px',
```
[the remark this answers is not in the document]
CarlosNZ: I guess it's fine. I remember when doing this originally, there was a lot of these kind of things that were quite finicky about where they were applied -- if it was wrong it would fix it for one context and mess it up for another. Hence I'm cautious whenever these styles get tweaked. But this seems okay, but should keep our eyes open for any odd quirks.

client/packages/common/src/ui/layout/tables/useMaterialTableColumns.tsx
```diff
@@ -63,9 +72,56 @@ export const useMaterialTableColumns = <T extends MRT_RowData>(
           };
         }
 
+        const defaultAggregationFn: MRT_AggregationFn<T> = (
+          columnId,
+          _leafRows,
+          childRows
+        ) => {
```
CarlosNZ: I guess move this to your new `aggregationFns` file now?

client/packages/invoices/src/InboundShipment/DetailView/columns.ts
```diff
@@ -110,62 +109,41 @@ export const useInboundShipmentColumns = () => {
...
-            return (
-              ArrayUtils.getUnitQuantity(row.subRows ?? []) *
-              (row.item.doses ?? 1)
-            );
-
           return row.packSize * row.numberOfPacks * (row.item.doses ?? 1);
         },
+        header: t('label.doses'),
+        columnType: ColumnType.Number,
+        aggregationFn: 'sum',
```
CarlosNZ: Can you put `aggregationFn: 'sum'` in the default column definitions for `ColumnType.Number`? In `useGetColumnDefDefaults`

PR #10439, opened 2026-02-10
10173 reduce needless log entries

### LINKED ISSUE #10173: Reduce needless log entries
## What went wrong? 😲

I tried to find the cause of a problem on one of our test servers and on viewing the log had to wade through MB of this:

Image

## Expected behaviour 🤔
Do we have to log these empty results? they don't provide any useful information to me and are just filling up the log.

If they must remain could the log level be something more verbose than `WARN` please - for warnings I would only want to see entries which are of concern. 

## How to Reproduce 🔨

1. Configure your log settings to display warnings
2. Sync
3. Check logs

You should see entries like these

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #10173

# 👩🏻‍💻 What does this PR do?
Reduces number of logs from sync. Results should only log if there is a result, and warning is separated from info logs.

# 🧪 Testing

- [ ] Run server
- [ ] Don't see endless integration result logs
- [ ] Don't see countless push / pull logs

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/translation_and_integration.rs
+++ b/server/service/src/sync/translation_and_integration.rs
@@ -247,7 +247,7 @@ impl TranslationAndIntegrationResults {
         Default::default()
     }
 
-    pub(crate) fn log_if_has_results(&self, operation_name: &str) {
+    pub(crate) fn log(&self, operation_name: &str) {
         let has_results = !self.0.is_empty()
             && self
                 .0
@@ -261,6 +261,11 @@ impl TranslationAndIntegrationResults {
                     log::info!("{operation_name} Integration result for {table_name}: {result:?}");
                 }
             }
+        } else {
+            log::debug!(
+                "{operation_name} Integration result: No records integrated or errored {:?}",
+                self.0
+            );
         }
     }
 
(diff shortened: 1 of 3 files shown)


server/service/src/sync/translation_and_integration.rs
```diff
@@ -247,6 +247,23 @@ impl TranslationAndIntegrationResults {
         Default::default()
     }
 
+    pub(crate) fn log_if_has_results(&self, operation_name: &str) {
```
jmbrunskill: I wonder if we can extend this to log all these operations but at debug level? It can be helpful to see that an action has occured but has nothing to do, rather than just not logging? Just a general comment feels like its helpful to be able to turn on debug and see that it at least tried to to integrate records, or tried to sync and got nothing etc.

Chris-Petty: I tend to agree! Can then just chop the fn name to `log` then I think.

PR #10443, opened 2026-02-10
10442: Add recommended extensions for vs code to repo

### LINKED ISSUE #10442: Recommended VS code extensions for devs
We are going to assume every dev uses VS code for the purpose of this ticket - if they dont then their experience will not be impacted 

As a new developer (internal or external contributor)
I want to know the best tools for the job
So that I can contribute most effectively!

- Therefore, we should share the extensions which the majority of the team are using as part of the repo :-) 
- Obviously, some niche people will use their own in addition to these, but we should include the extensions which most of our devs are using.

I also considered vs code extension pack but that would be a lot more opinionated and public facing

### Additional context 💌

- Developers can choose to ignore the vs code prompt - it does not force them to comply
- Gives capability to list both recommended extensions and unwated extensions - which may be useful for recommending against conflicting formatters for example
- In the future you could go further and add a script `pnpm dev:setup` which runs `code --install-extension esbenp.prettier-vscode` to bootsctrap for new hires, but this is less necessary

### Acceptance Criteria ✅

- When a developer who does not have all the recommended extensions installed, opens the repo in a new vs code window, then vs code prompts them to install the missing recommended extensions

This could be :
- a new dev
- an existing dev who doesnt have them all already installed
- - for testing purposes then you can create a new blank profile on vs code without any extensions and open the repo in that profile

esmehm:
Fixes #10442 

# 👩🏻‍💻 What does this PR do?
This PR adds a list of recommended extensions for developers contributing to this repo using vs code to the code base. This allows us to discuss the extensions like code (whereas other solutions are more static) and easily ensures any new devs have the best tools that the rest of the team uses from the get go. 

If you are missing extensions yuou will get this prompt new time you open a new window with this repo open
image

If you click see recommendations
image

Please add any others which you use - either by commenting or pushing directly to the PR
Equally please shout if these are not what you're using - I am not the right person to know what we actually do use seeing as this was inspired by me missing extensions myself

# 🧪 Testing

If you wish to test, you can create a blank profile in VS code (cmd shift p, creare new profile) without any extensions and then open a new window as that profile with the repo in it, and then see the prompt to install the extensions

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -16,8 +16,5 @@
     // General
     "streetsidesoftware.code-spell-checker",
     "github.vscode-pull-request-github",
-
-    // Testing
-    "orta.vscode-jest"
   ]
 }


.vscode/extensions.json
```diff
@@ -0,0 +1,22 @@
...
+    // TypeScript/React
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    
+    // GraphQL
+    "graphql.vscode-graphql",
+    "graphql.vscode-graphql-syntax",
+    
+    // General
+    "eamodio.gitlens",
```
jmbrunskill: I think at least Andrei and I don't like gitlens as its too heavy, and most of the important features are already in vscode like blame. I think we should probably remove that one...

ria8651: Yea not a fan of gitlens

PR #10486, opened 2026-02-12
9186 item category join's item id should be item link

### LINKED ISSUE #9186: item_category_join's item_id should be item link id
## The suggestion

Change item_category_join's item_id to item link id and make sure that it references the table.

roxy-dao:
Fixes #9186

# 👩🏻‍💻 What does this PR do?
Change item_id -> item_link_id in item_category_join & update joins

# 🧪 Testing

- [ ] Make sure tests still pass

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/sync/translations/item.rs
+++ b/server/service/src/sync/translations/item.rs
@@ -2,8 +2,8 @@ use chrono::Utc;
 use repository::{
     item_category::{ItemCategoryFilter, ItemCategoryRepository},
     item_category_row::ItemCategoryJoinRow,
-    ChangelogRow, ChangelogTableName, EqualFilter, ItemLinkRowRepository, ItemRow, ItemRowDelete,
-    ItemRowRepository, ItemType, StorageConnection, SyncBufferRow, VENCategory,
+    ChangelogRow, ChangelogTableName, EqualFilter, ItemRow, ItemRowDelete, ItemRowRepository,
+    ItemType, StorageConnection, SyncBufferRow, VENCategory,
 };
 use serde::{Deserialize, Serialize};
 


roxy-dao:
> Just one tidy up in the migration, otherwise looks legit.
> 
> I am wondering if we should have this in 2.16... merge not working correctly I'd say is in the category of high priority bugs at least...
> 
> Fortunately, if I'm thinking straight, the migration should "fix" merges that have already taken place (i.e. categories that appeared lost after merge would be recovered).
> 
> That makes me think though, that's a good way to have an item appear twice in a category! Maybe worth opening a follow up issue and maybe we need to think about how we address this...
> 
> [`fdd84ab`/server/service/src/sync/translations/special/item_merge.rs](https://github.com/msupply-foundation/open-msupply/blob/fdd84abf1ec664695591fb999e95c86796a20a74/server/service/src/sync/translations/special/item_merge.rs)
> 
> [`fdd84ab`/server/service/src/sync/translations/special/name_merge.rs](https://github.com/msupply-foundation/open-msupply/blob/fdd84abf1ec664695591fb999e95c86796a20a74/server/service/src/sync/translations/special/name_merge.rs)
> 
> name_merge.rs handles double visibility of patients by deleting the effectively identical `name_store_join` records. Upside, tidies things up a bit. Downside, sort undermines the ability in the future to "unmerge" names, and would be the same case if we did this for item/item_category_join. Perhaps just a DISTINCT in our sql queries would suffice...

Will create issue for it soon, but also wanna see how this is being handled https://github.com/msupply-foundation/open-msupply/pull/10181

Chris-Petty:
Can you figure out why the tests are failing? At a glance maybe it's freaking out about the apostrophe in the branch name

roxy-dao:
> Can you figure out why the tests are failing? At a glance maybe it's freaking out about the apostrophe in the branch name

yes that's exactly why it's failing lul

PR #10495, opened 2026-02-13
10468 filter by requisition/invoice number in list views

### LINKED ISSUE #10468: It would be good to be able to filter requisition/inbound/outbound shipments by invoice numbers
## What went wrong? 😲

I feel it is useful when you have a list with lot of records and you look for a particular requisition/invoice which happens quite a lot for me when trying to pinpoint the outbound shipment linked to a specific inbound shipment. Now we can mostly filter by either name or status. It can also be a search field, if it is more convenient. 

Image

I believe that it might be useful to clients that look for a specific invoice/requisition the same way we can do it for requisitions.

Image

## Expected behaviour 🤔
A way of finding a specific invoices/requisition by it's number

## How to Reproduce 🔨

1. Have a store that has a lot of inbound shipments or Outbound shipments or Requisition/Internal orders 
2. Go to the list of any one of those
3. Click on Filters
4. There is no filter on invoice or requisition number. There is no search field either

## Your environment 🌱

- Open mSupply Version: 2.16.00-RC1-develop-02111053
- Legacy mSupply Central Server Version: v8.08

- Platform: All plateform

- Database type: Postgres and SQLite

aimee-mcneil-melville:
Fixes #10468

# 👩🏻‍💻 What does this PR do?

Adds a filter to the toolbar to filter by `Invoice Number` in list view for Inbound and Outbound shipments
Adds a filter to the toolbar to filter by `Requisition Number` in list view for Request and Response requisitions

Screenshot 2026-02-13 at 1 54 32 PM

Added width as requisition labels were cut off even when the filter is selected and using the small label. Could remove this and instead abbreviate number to #, which we do infrequently

I think it's more user friendly to not require an exact equal number match - maybe a 'starts with' condition? To change to this would require changing the filter definition - would need to confirm that this won't affect queries/transformations using the requisition/invoice number by getting unintended records in results

# 🧪 Testing

- [ ] Test for each of the areas List View: Request requisitions, response requisitions, inbound shipments, outbound shipments
- [ ] Go to the relevant list view 
- [ ] Select from the filter menu - `Requisition number` or `Invoice number`
- [ ] Type a number in the field - see only exact match results
- [ ] Typing a letter or symbol does nothing
- [ ] Clearing the field restores all results

# 📃 Documentation

  1. Filter mentions for each of request requisitions, response requisitions, inbound shipments, outbound shipments
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -2277,7 +2277,6 @@
   "placeholder.search-by-name": "Search by name",
   "placeholder.search-by-name-or-code": "Search by name or code",
   "placeholder.search-by-notes": "Search by notes",
-  "placeholder.search-by-requisition-number": "Search by requisition number",
   "plugin.amd": "AMD",
   "plugin.catalogue-price": "Catalogue price",
   "plugin.civ.final-soh": "Final SOH",


jmbrunskill (approved):
Looks good to me.

```
I think it's more user friendly to not require an exact equal number match - maybe a 'starts with' condition? To change to this would require changing the filter definition - would need to confirm that this won't affect queries/transformations using the requisition/invoice number by getting unintended records in results
```
Agreed! Should be a follow up issue though, this is enough for now.

BTW: Noticed requisition seems to sort by createdDatetime asc by default think that needs to be descending will raise an issue or PR

Issue #10242, opened 2026-01-14
UX/UI issues noted during the V2.16 Release Testing


marthakeezy:
### Creating one big issue for triage to go through instead of many 'noisy' ones, please feel free to open or let me know which ones need to be opened as separate issues to be resolved quicker than others :)

_____

See below for images & recreation steps ⬇️

~~- [ ] 1. Catalogue -> Items: old filters are still there~~ as they should be
- [x] 2. Inventory -> Stock: ascending sort is not pre-selected in the column header
- [x] 3. Monitoring -> "un-acknowledge" Label is not displayed properly
- [x] 4. Internal orders: Warning alert - > Cancel button should be placed at the last
- [ ] 5. On Android, when initialising, once client or server option is selected, there is no way to get back to the initial screen to change options
- [ ] 6. Initialisation error associated with the wrong central server URL on Android leaks beyond the error box
- [ ] 7. Can't switch languages easily when pulling or integrating records during initialisation
- [x] 8. 'No matching patients' message is a bit too close to the header
- [ ] 9. Demographics -> cannot cancel once attempting to save with an error, refreshing causes unnecessary toast
- [ ] 10. On login screen, Central Server indicator disappears upon refresh
- [ ] 11. Patient creation module - funky layout
- [x] 12. Stock details unit value is not in line (a bit low) with the label and the scan button should be to the right of the barcode field instead of below it. 
- [ ] 13. Unclear which fields are required when updating a status for Cold Chain Equipment
- [ ] 14. 'Total' line hangs around at the bottom even when all lines have been removed
- [x] 15. When no items are added to a prescription, wrong wording appears
- [x] 16. 'On hold' does not show fully when issuing an item in a prescription
- [ ] 17. When toggling table density, comment icon disappears on the highest setting and no way to increate the width of the column for comments
- [ ] 18. In requisitions, placeholder line's doses are not showing in blue
- [ ] 19. Wrong icon used when downloading an Internal Order PDF
- [x] 20. When adding items to an Inbound Shipment, almost none of the column titles show up fully
- [ ] 21. Too many different ways of showing decimal points when adding an item
- [x] 22. Weird 'Close' button when viewing Vaccine Courses
- [x] 23. GAPS is not capitalised under Global Preferences
- [ ] 24. Population figures reset to 0 when filling out % growth in demographics for the first time
- [x] 25. 'Creating a new document' should probably be 'uploading a new document'
- [x] 26. Spacing in Replenishment??
[...]

Issue #9646, opened 2025-10-31
Wrong icon for import / export in CC equipments


mlenignol:
## What went wrong? 😲

Heya :) the icons for import and export in the cold chain equipment seem to be reversed (arrow going out for import instead of going in, and vice-versa for export)

<img width="1908" height="865" alt="Image" src="https://github.com/user-attachments/assets/4aafaf04-ca9c-4cc4-9ce2-e1af720b2f18" />

## Expected behaviour 🤔

Arrows going the other way around (out for export and in for import)
Buttons work as expected --> only the icon seems wrong

## How to Reproduce 🔨

Steps to reproduce the behaviour: Go to Cold Chain --> Equipment --> import / export buttons

## Your environment 🌱

- Open mSupply Version: 2.11
- Legacy mSupply Central Server Version: browser (msupply demo central server: demo-open.msupply.org)
- Platform:

- Database type:

mark-prins:
triage: yes, they should be the other way around! adding to cooldown

PR #10515, opened 2026-02-15
10242 uiux issues v2.16

esmehm:
Fixes several points raised in #10242 - breakdown below

# 👩🏻‍💻 What does this PR do?

I am not addressing the android specific problems at this time for speed - and have broken down the bullet points raised - either addressing or deprioritising / deferring each
The reviewer can ignore points which are not ticked - I simply included them to fully correspond to the ticket. We are only reviewing the ticked changes here.

~~- [ ] 1. Catalogue -> Items: old filters are still there~~  - as they should be

- [x] 2. Inventory -> Stock: ascending sort is not pre-selected in the column header 
Already resolved elsewhere

- [x] 3. Monitoring -> "un-acknowledge" Label is not displayed properly
Already resolved

- [x] 4. Internal orders: Warning alert - > Cancel button should be placed at the last
Resolved in https://github.com/msupply-foundation/open-msupply/pull/10337/commits (into develop, oops!), cherry picked to this branch. 
James and Mark, pointed out it would better on the left. So added additional commit to move it to the left -> fixed to left on RC, and change will merge smoothly into develop in future
image

- [ ] 5. On Android, when initialising, once client or server option is selected, there is no way to get back to the initial screen to change options
Not addressing android issues  in this PR

- [ ] 6. Initialisation error associated with the wrong central server URL on Android leaks beyond the error box
Not addressing android issues  in this PR

- [ ] 7. Can't switch languages easily when pulling or integrating records during initialisation
Deprioritising as can set before

- [x] 8. 'No matching patients' message is a bit too close to the header
Addressing in this PR
After this PR, more space between the two
image

- [ ] 9. Demographics -> cannot cancel once attempting to save with an error, refreshing causes unnecessary toast - 
Deprioritising

- [ ] 10. On login screen, Central Server indicator disappears upon refresh - expected because isCentralServer uses authenticated endopoint
Deprioritised

- [ ] 11. Patient creation module - funky layout - uses JSOn form and is different depending on how set up. Deprioritised

- [x] 12. Stock details unit value is not in line (a bit low) with the label and the scan button should be to the right of the barcode field instead of below it.
Addressing
After this PR, they are indeed side by side (future work to consider if the barcode should be editable directyl and how you may not be able to see the whole length)
image

- [ ] 13. Unclear which fields are required when updating a status for Cold Chain Equipment
Should be addressed in a follow up

~~[ ] 14. 'Total' line hangs around at the bottom even when all lines have been removed~~
@mark-prins confirmed we should keep this behaviour - follow up ticket to be created for totalling other rows beyond just volume

- [x] 15. When no items are added to a prescription, wrong wording appears
Addressed in this PR
used to say no prescriptions -> after this PR it correctly says no items
image

- [x] 16. 'On hold' does not show fully when issuing an item in a prescription 
Addressed in this PR
After this PR:
image
image

Tablet is no worse than before (on hold still isnt showing properly) but to address that furthwere we'd need to reasses the design and probably not have the new item take up so much width, as there's not enough width overall
image

- [ ] 17. When toggling table density, comment icon disappears on the highest setting and no way to increate the width of the column for comments - deprioritised for now

- [ ] 18. In requisitions, placeholder line's doses are not showing in blue
Cannot replicate - have asked

- [ ] 19. Wrong icon used when downloading an Internal Order PDF
Previous issues also tried to address this and there is debate online about which arrow direction should be used when -> separated into devoted sub issue and will consult designer. Not addressing in this PR

- [x] 20. When adding items to an Inbound Shipment, almost none of the column titles show up fully
Addressed in this PR
Sadly the columns on all these old style tables ar ehard coded so dont auto adjust which would be better (or dragable like new style tables), so left as hard coded (but larger ) values
image
A longer lang like french is still truncated on some but would be a bigger change to resolve as it doesnt have horizontal scroll

- [ ] 21. Too many different ways of showing decimal points when adding an item
Struggling to repolicate the problematic 15 dp - shall repro with Maria in the office and address later

- [x] 22. Weird 'Close' button when viewing Vaccine Courses 
Already addressed

- [x] 23. GAPS is not capitalised under Global Preferences
Addressed here
After this PR:
image
Other latin character langs already look ok

- [ ] 24. Population figures reset to 0 when filling out % growth in demographics for the first time
deprioritised

- [x] 25. 'Creating a new document' should probably be 'uploading a new document'
Before this PR:
image

After this PR:
image

- [x] 26. Spacing in Replenishment??
Addressed
The nav bar children were out of line - this aligns them and does a small refactor to make it less likely for this to occur in future (TLDR the problem was forgetting `end` prosp which was expected on every child, now we just expect `isParent` on the parents, of which there are fewer). Thanks to Mark who had already got an idea ofr this
After the PR:
image

- [ ] 27. On Android, Export/Print button flickers with every screen interaction
Android out of scope of this PR

- [x] 28. Inbound Shipment detailed view de-selects the menu item
Appears to already be resolved
image

- [x] 29. Pack volume  of an item is not visible
I think this problem was already resolved by the dp ticket
image

I will update the ticket with what is in once merged for QA/ PM to assess remaining

# 🧪 Testing

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.test.tsx
+++ b/client/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.test.tsx
@@ -86,6 +86,7 @@ describe('AppNavLink', () => {
                   to="/outbound-shipment"
                   icon={<TruckIcon />}
                   text="Distribution"
+                  isParent={true}
                 />
               </Box>
             }


client/packages/invoices/src/OutboundShipment/DetailView/columns.tsx
```diff
@@ -196,21 +200,21 @@ export const useOutboundColumns = () => {
...
-                      (subRow.stockLine?.volumePerPack ?? 0) *
-                        subRow.numberOfPacks,
-                    0
-                  )
-                : (row.original.stockLine?.volumePerPack ?? 0) *
-                  row.original.numberOfPacks;
-              return sum + rowVolume;
-            }, 0);
+          const rows = table.getFilteredRowModel().rows;
+          if (!rows.length) return null;
```
jmbrunskill: I don't like how we have to put these checks in multple places, I think I'd prefer we always show the total row even without any lines? @mark-prins do you think the old behaviour was acceptable?
I think we should probably also add other totals to this bar to make it more useful, volume is the only thing totaled right now... Prices at least make sense

client/packages/system/src/Patient/CreatePatientModal/PatientResultsTab.tsx
```diff
@@ -210,7 +211,9 @@ export const PatientResultsTab: FC<
     data,
     columns,
     onRowClick: handleRowClick,
-    noDataElement: t('messages.no-matching-patients'),
+    noDataElement: (
+      <Box paddingTop={3}>{t('messages.no-matching-patients')}</Box>
```
jmbrunskill: This isn't displaying well on tablet, not sure if its realted to this change though?

Image

Note: this is adding a patient from the Create Prescriptoin workflow..

jmbrunskill: Also if it fails to save a patient it still tries to use them in the prescription workflow...
I probalby need to raise an issue specifically for this!

client/packages/common/src/ui/icons/Download.tsx
```diff
@@ -16,9 +16,10 @@ export const DownloadIcon = (props: SvgIconProps): JSX.Element => {
 
   return (
     <SvgIcon {...combinedProps} viewBox="0 0 24 24">
-      <path d="M21 15v4a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2v-4"></path>
-      <polyline points="17 8 12 3 7 8"></polyline>
-      <line x1="12" y1="3" x2="12" y2="15"></line>
+      <path d="M21 15v4a2 2 0 0 1-2 2H5a2 2 0 0 1-2-2v-4" />
+      {/* Arrow DOWN */}
```
jmbrunskill: Suprised these we always wrong. @mark-prins any ideas why?

client/packages/system/src/Patient/CreatePatientModal/PatientResultsTab.tsx
esmehm: Also broken on the release branch - shall fix and add to this PR

client/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -40,11 +41,23 @@ export const ConfirmationModal = ({
   buttonLabel,
   cancelButtonLabel,
   otherButtons,
+  placeCancelButtonLast = false,
```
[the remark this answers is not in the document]
esmehm: This discussion is around this point
image

Thoughts @mark-prins ?

client/packages/invoices/src/OutboundShipment/DetailView/columns.tsx
mark-prins: I don't mind having the total bar show - though I would prefer to have the actual totals (`0`) shown for the columns which have totals enabled. to me this indicates that the bar is deliberate and will do something if rows are added to the table

client/packages/common/src/ui/icons/Download.tsx
jmbrunskill: @mark-prins do we want the other icons for importing/exporting? Was this issue just wrong? https://github.com/msupply-foundation/open-msupply/issues/9646

Sounds like you said you wanted this way here: 
`triage: yes, they should be the other way around! adding to cooldown`

Esme's change seems right to me... but still...

esmehm: So it turns out there is a lot of online debate about this
To summarise, we will all agree there are 2 possible arrow directions (upwards, out the box) and downwards (into the box)
And there are 4 words we could use between them:
- upload (I think we can all agree upwards arrow)
- download (I hope we can all agree downwards arrow)
- export - debatable
- import - debatable

Different people appear to feel differently about whether export should use up / down arrow, depending on the semantics, and vice versa for import
We have been using download icon for exporting, but this article for example suggests we should be using the opposite https://www.linkedin.com/pulse/breaking-down-misleading-icons-upload-download-import-nikita-kurnosov-sbawe/ . Aware it's not the best source though - struggling to find a good one with a definitive answer

So we basically need to decide whch direction for these 4 words, and then I will put it into the codebase to reduce this confusion in future

client/packages/system/src/Patient/CreatePatientModal/PatientResultsTab.tsx
esmehm: I have separated this into its own subissue and PR so separating it from this scope
https://github.com/msupply-foundation/open-msupply/pull/10573

It is not impacted by this branch and the two will npot have any conflicts

client/packages/invoices/src/OutboundShipment/DetailView/columns.tsx
esmehm: Agreed that prices would also make sense to total @jmbrunskill but should be its own ticket (if high enough priority to do)

PR #10518, opened 2026-02-15
#10480 Scanner input -- persist expiry date for existing lines

### LINKED ISSUE #10480: The system does not correctly detect differences in expiry dates when scanning barcodes.
## What went wrong? 😲

When an item is first scanned with Batch A and a valid expiry date (24/07/04), the line item is created correctly. However, when the same item is scanned again using a different barcode where the expiry date is missing (null), the system:

Displays a message indicating that batch and pack size are different
Adds the scanned item to the existing line item
Overrides the previously stored expiry date with a blank value

## Expected behaviour 🤔
It is okay to display this warning, but user should be informed to select the expiry date/ or should not over ride the expiry date.

Not sure whether there are medicines with different expiry date in same batch,

## How to Reproduce 🔨

Image

This barcode has no expiry date, but same batch number and GTIN

Image

## Your environment 🌱

- Open mSupply Version: 2.16.0

- Platform: Desktop

- Database type:

CarlosNZ:
Fixes #10480

# 👩🏻‍💻 What does this PR do?

As part of the draft state update, we now also check if the updated draft values match an existing line, and populate draft expiryDate with existing value if it does.

There's a tiny inefficiency in that `lines.find()` is run both in the update function and in the component itself, but this is preferable to the alternative which is to add `useEffect` to watch for changes to `existingLine` ;) 

# 🧪 Testing

- [ ] Scan a barcode where the batch/item matches an existing line (but without the expiryDate in the barcode itself) -- confirm that the expiry date from the existing line is kept
- [ ] Scan a barcode with a matching item, but no batch. Manually enter the batch number of the existing line -- confirm that the expiry date is immediately populated with the existing value
- [ ] Scan a barcode where the expiry date is in the barcode,  but different to the one in the existing line -- confirm that existing expiry date is overwritten with the new one (I don't think this case should occur IRL, just want to check behaviour is as expected)
- [ ] Check nothing else is broken

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/DetailView/ScanInputModal.tsx
+++ b/client/packages/invoices/src/InboundShipment/DetailView/ScanInputModal.tsx
@@ -12,6 +12,8 @@ import {
   TypedTFunction,
   LocaleKey,
   useNotification,
+  LoadingButton,
+  CheckIcon,
 } from '@openmsupply-client/common';
 import { FnUtils, ScanResult, useBarcodeScannerContext } from '@common/utils';
 import React, { useCallback, useState } from 'react';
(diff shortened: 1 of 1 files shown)


jmbrunskill (approved):
Works well for me.

Just wait for Esme to finish here review on the base branch

jmbrunskill:
Oops accidentally fixed Esme's feedback on this branch. Will merge!

PR #10527, opened 2026-02-16
#10474 Fix confusion with simultaneous modals

### LINKED ISSUE #10474: Barcode Scanner: Confusion when scanning duplicate batches while on the Item Edit screen
## What went wrong? 😲

Scanning the duplicate items while working in the "Item Edit Screen", the existing batch entries disappear from the screen.  It confuses the user, and thinking I lost the entry.

## How to Reproduce 🔨

1. Go Inbound shipment of Desktop app
2. Configure Zebra scanner
3. Scan the item with two types of batches
4. Click on the Item line and navigate to the Item Line Edit screen
5. Scan the item which has duplicate batch
6. Insert qty and click ok
---
Shipping line already exists

Image

---

User navigates to Item Edit screen

Image

---

Scanning the same batch

Image

---

User insert the qty and click Ok button, causes item to disappear from the screen.

Image
## Your environment 🌱

- Open mSupply Version: 2.16.0

- Platform: Desktop

- Database type:

CarlosNZ:
Fixes #10474

# 👩🏻‍💻 What does this PR do?

The LineEdit modal didn't really have any proper `isDirty` tracking, so just went with the simpler option here -- if the Line Edit modal is open, we simply suppress any scanning until it's closed. If the user attempts to scan, they'll get a warning like so:

Screenshot 2026-02-16 at 3 29 38 PM

Was a bit tricky since these two modals don't have any knowledge of each other's state. There was a convoluted approach suggested by Co-pilot, but it had way to much interdependence between the modals and was quite hard to follow. So I've just updated the BarcodeScannerContext so it can temporarily change the scan handler callback. 

So the Edit modal registers a new callback when it opens (which toggles on the warning), and calls `revertCallback()` on dismount so the original callback becomes active again.

I'm not 100% convinced this is rock solid, so we should check it on different devices.

# 🧪 Testing

- [ ] Scan items as normal
- [ ] Click on an item line and bring up the Edit modal
- [ ] Now try scanning again (can just click button on mock scanner, but try with real scanner if possible)
- [ ] Should see the warning shown above
- [ ] Close the modal (either save or cancel)
- [ ] Now scan again, the normal scan input modal should come back

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -2641,6 +2641,7 @@
   "warning.high-number-records-to-sync": "A high number of records are being synced. This may take longer than usual.",
   "warning.insufficient-recent-stocktake-items": "You have not completed stocktake(s) containing at least {{ minItems }} items within the last {{maxAge}} days. Your stock levels may not be accurate. Are you sure that you want to continue?",
   "warning.manual-store-requisition": "Entering a requisition for this store may result in duplicates if the store also submits an Internal Order.",
+  "warning.no-scanning-while-line-editing": "Please save or cancel the line edit modal before scanning",
   "warning.nothing-to-supply": "Nothing left to supply!",
   "warning.requested-exceeds-suggested": "Your requested quantity is higher than the suggested",
   "settings.enable-mock-barcode-scanner": "Enable manual barcode input (for testing purposes only)",
--- a/client/packages/common/src/utils/BarcodeScannerContext.tsx
+++ b/client/packages/common/src/utils/BarcodeScannerContext.tsx
@@ -66,6 +66,7 @@ interface BarcodeScannerControl {
   setMockScannerEnabled: (enabled: boolean) => void;
   supportsContinuousScanning: boolean;
   registerCallback: (callback: ScanCallback) => void;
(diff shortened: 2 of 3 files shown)


jmbrunskill (changes requested):
Copilot is worried and I am too, about the revert callback thing.
Depending on rendering orders or repeats etc it's hard to know it will do the right thing.

How about this alertnative approach? https://github.com/msupply-foundation/open-msupply/pull/10532

CarlosNZ:
> Depending on rendering orders or repeats etc it's hard to know it will do the right thing.
> 
> How about this alertnative approach? #10532

Yeah I initially tried something similar, but was trying to get both modals to know the state of each other which was unnecessarily complicated. Using a snackbar instead means only one modal needs to know the other state, so that should work nicely.

Only downside is, if we want to update it later to only prevent scanning when Edit modal is in a `isDirty` state, we won't really be able to. Probably doesn't matter though, right?

CarlosNZ:
Replacing with https://github.com/msupply-foundation/open-msupply/pull/10532

PR #10554, opened 2026-02-18
10530 Use ordered float

### LINKED ISSUE #10530: Item Usage report occasionally fails to generate
## What went wrong? 😲

Noticing this behaviour - item usage report fails to produce data about 50% of the time

https://github.com/user-attachments/assets/787d96cb-5274-40eb-a103-b37e5650c932

GraphQL error: 

Image

## Expected behaviour 🤔

Should be able to produce a report

## How to Reproduce 🔨

1. Go to Reports -> Item Usage
2. Generate a report a few times
3. See if you can run into the issue
4. Possibly: log out, log back in again, refresh pages in-between to see if you can recreate the issue

## Your environment 🌱

- Open mSupply Version: `V2.16.00-RC-02130838` & happening in V2.17 develop also
- Legacy mSupply Central Server Version: `V8.06.12`

- Database type: `PostgreSQL`

ria8651:
Fixes #10530

# 👩🏻‍💻 What does this PR do?

It seems that the custom hash function using `to_bits` was unstable causing the async graphql dataloader to fail due to the broken assumptions of Eq, Ord and Hash. Replacing the custom hash function with the wrapper from `ordered-float` seems to have fixed the issue for me. Was quite annoying to track down and would argue that custom hash functions shouldn't be needed in oms. Could be the kind of thing to add to the rust developer practices wiki page.

# 🧪 Testing

- [ ] Generate the Item Usage report lots of times

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/core/src/loader/available_volume_on_requisition.rs
+++ b/server/graphql/core/src/loader/available_volume_on_requisition.rs
@@ -1,4 +1,3 @@
-use crate::loader::IdPair;
 use actix_web::web::Data;
 use async_graphql::dataloader::Loader;
 use async_graphql::*;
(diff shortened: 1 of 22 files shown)


jmbrunskill:
BTW: When I say it handles it for item_stats, it explicitly errors on the case you mentioned where you have two store_ids in the request. I think this is probably fine to keep the complexity down but otherwise yes you'd also need to group by store_id as well.

jmbrunskill:
Also given your description if graphql loader is aggrating across multiple grapqhl requests haveing a bug due to the item_stats loader having 2 different store_ids is just a matter of time on a busy multi store mutli user system...

PR #10556, opened 2026-02-18
Remove `IdPair`

ria8651:
# 👩🏻‍💻 What does this PR do?

`IdPair` obfuscates what the actual IDs are for and made for a confusing developer experience. I couldn't see the advantage of it over having separate input types for each loader given the disadvantages.

Have merge base as my other PR branch because I didn't wanna merge or rebase. I'll merge this once #10554 is merged into 2.16 and 2.16 is merged into dev.

# 🧪 Testing

- [ ] Check the DataLoaders still work (Should be covered by tests)

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/graphql/core/src/loader/mod.rs
+++ b/server/graphql/core/src/loader/mod.rs
@@ -134,15 +134,6 @@ pub use vvm_status::*;
 pub use vvm_status_log::*;
 pub use warning::*;
 
-/// Helper function to get unique ids from a list of inputs. Could be replaced with Itertools' unique if we wanted to add that as a dependency, but this is simple enough for now.
-pub fn unique_ids<T: Eq + std::hash::Hash + Clone, I: IntoIterator<Item = T>>(inputs: I) -> Vec<T> {
-    use std::iter::FromIterator;
-    // Using HashSet to get unique values, then converting back to Vec
-    std::collections::HashSet::<T>::from_iter(inputs)
-        .into_iter()
-        .collect()
-}
-
 #[derive(Clone)]
 // Using struct instead of () to avoid conflicting new implementations
 pub struct EmptyPayload;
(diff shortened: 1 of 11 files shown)


server/graphql/core/src/loader/mod.rs
```diff
@@ -134,8 +134,10 @@ pub use vvm_status::*;
 pub use vvm_status_log::*;
 pub use warning::*;
 
+/// Helper function to get unique ids from a list of inputs. Could be replaced with Itertools' unique if we wanted to add that as a dependency, but this is simple enough for now.
```
jmbrunskill: `Helper function to get unique ids from a list of inputs.`

should be 

`Helper function to get unique values from an iterator as a vec`

server/graphql/core/src/loader/mod.rs
```diff
@@ -158,3 +160,21 @@ impl RequisitionAndItemId {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_unique_ids() {
```
jmbrunskill: Thanks @ria8651 this makes a lot more sense now. I was thrown by the name `unique_ids` and the call signature that it was trying to do something special. I if it was just called it `unique` or `dedup` that might be clearer. I know rust's vec `dedup` requires sorting though which is probably a good reason to not call it dedup even though that would be natural for me.. 

Btw: I notice we have this code
```
    let mut codes: Vec<String> = reports_with_compatible_versions
        .iter()
        .map(|r| r.code.clone())
        .collect();
    codes.dedup();
```

Which I suspect is a potential but if we don't sort by code.

Other example is:
```
    // There can be duplicates in the program settings due to name tags
    // which we don't care about here, so we dedup here
    period_schedule_ids.sort_unstable();
    period_schedule_ids.dedup();
```

This could be cleaned up by your function...

I'm thinking we should move it to the `utils` package so it's available to be used elsewhere in the code base?

jmbrunskill (approved):
Looks great, thanks for the updates @ria8651 

I haven't verified all the graphql, or reviewed if payloads are currently being miss-used. It seems like there are a number that are making assumptions about the payload that might not always hold true but are probably okish in our usage currently. Andrei's concerns still stand that loaders might not be properly grouping all variations of queries when there's more complex inputs to the loader.

This change makes that easier to review and check though in my opinion!

PR #10587, opened 2026-02-20
10553 dashboard link alignment 2

### LINKED ISSUE #10553: Dashboard: Links do not return the correct information
## What went wrong? 😲

If you click on the following Links from the dashboard, the search returns incorrect values:

Image

For example, it says "4 items with no stock" but on hyperlink it returns only 1 item line

## How to Reproduce 🔨

1. Go to 'dashboard
2. Click on 'inventory management' 
3. Click on the hyperlinks 

## Your environment 🌱

- Open mSupply Version:  2.17.00-develop-02170444

- Platform: browser

jmbrunskill:
Fixes #10553

# 👩🏻‍💻 What does this PR do?

Fixes two miss alignments between dashboard counts and item filters.

**Batches expiring in between 30 days and 90 days**
Dashboard query was treating between 30 and 90 days as exclusive of the 90days, now that's aligned in the link they both show between 30 and 89 days (excluding the 90th day)

**Expired batches**
Dashboard query was looking for items which were available but the stock filter was looking for items that have `packs_in_store` causing a miss match when stock was assigned to outbound shipments

Have briefly reviewed the logic for the other two that were reported...

**Items with no stock**
**Items with less than 3months stock**

They now appear to be working correctly on the test server and code looks ok to me, so will leave it here unless there's an issue replicated again so I can investigate better...

# 🧪 Testing

- [ ] Check dashboard links return the same row count as we see on the dashboard.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/service/src/dashboard/stock_expiry_count.rs
+++ b/server/service/src/dashboard/stock_expiry_count.rs
@@ -112,7 +112,7 @@ mod stock_count_test {
         let expired_stock_a: StockLineRow = {
             let mut updated = expired_stock_a();
             updated.available_number_of_packs = 0.0;
-            updated.total_number_of_packs = 20.0;
+            updated.total_number_of_packs = 0.0;
             updated
         };
 


andreievg (changes requested):
Looks good but would like a comment about 89

PR #10649, opened 2026-02-25
Use mui popover for popovers

ria8651:
# 👩🏻‍💻 What does this PR do?

Been annoyed by the popover quirks for a while so switched to using the mui popover component instead of a popper with lots of wrappers. Makes popovers work on mobile. Also removed `useDisabledNotificationPopover.tsx` as it doesn't fit with the current pattern of using toasts.

# 🧪 Testing

- [ ] Check all the popovers you can think off.

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/components/popover/PaperPopover/PaperPopover.tsx
+++ b/client/packages/common/src/ui/components/popover/PaperPopover/PaperPopover.tsx
@@ -72,7 +72,7 @@ export const PaperPopover: React.FC<PropsWithChildren<PopoverProps>> = ({
               height: height ? `${height}px` : 'auto',
               width: width ? `${width}px` : 'auto',
               borderRadius: '16px',
-              transform: getTranslation(placement),
+              ...getMargin(placement),
               ...paperProps?.style,
             },
           },
@@ -108,12 +108,8 @@ const getTransformOrigin = (placement: PopoverOrigin): PopoverOrigin => ({
         ? 'left'
         : 'center',
 });
-const getTranslation = (placement: PopoverOrigin) => {
-  const y =
-    placement.vertical === 'top'
-      ? '-8px'
-      : placement.vertical === 'bottom'
-        ? '8px'
-        : '0';
-  return `translateY(${y})`;
+const getMargin = (placement: PopoverOrigin): React.CSSProperties => {
+  if (placement.vertical === 'top') return { marginTop: '-8px' };
+  if (placement.vertical === 'bottom') return { marginTop: '8px' };
+  return {};
 };


client/packages/common/src/ui/components/popover/PaperPopover/PaperPopover.tsx
```diff
@@ -0,0 +1,154 @@
...
+  //   }) as DOMRect;
+  const handlePopoverShow = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handlePopoverHide = () => {
+    setAnchorEl(null);
+  };
+
+  useImperativeHandle(actions, () => ({
```
ria8651: @CarlosNZ is it ok to use `useImperativeHandle`? It's not used anywhere else in the codebase but seems to be the recommended way to do what was happening in `usePopover`.

client/packages/coldchain/src/common/Monitoring/TemperatureChart/BreachPopover.tsx
```diff
@@ -26,6 +25,7 @@ import {
   TemperatureBreachFragment,
   useTemperatureBreach,
 } from '../../../Monitoring/api/TemperatureBreach';
+import { Popover } from '@mui/material';
```
CarlosNZ: I think we're supposed to import this in common and re-export, as per all the other Material components

client/packages/coldchain/src/Equipment/DetailView/AppBarButtons.tsx
```diff
@@ -46,7 +43,6 @@ export const AppBarButtonsComponent = () => {
           variant="outlined"
         />
       </Grid>
-      <DisabledNotification />
```
CarlosNZ: Can you explain why we don't need this any more?

client/packages/common/src/ui/components/popover/PaperPopover/PaperPopover.tsx
CarlosNZ: Okay, I see one in StoreSelector. But is `show` ever called from another component?

CarlosNZ (changes requested):
Okay, looking pretty good I think. It *seems* to be working properly, although I'm not 100% clear on the mechanism in a few places. 

Would it be possible to add some documentation comment of how it works in the `PaperPopover` file?

Also, there's one little bug I had hoped this would fix, but it doesn't seem to be the case: In "Mobile" view, the Status popover in and Inbound Shipment becomes "onClick" triggered (rather than hover), though I'm not sure how it changes -- maybe built into the Popover component for small screens? Anyway, once you've clicked this on mobile, clicking outside it no longer closes it. 

Image

It works okay when you set `mode="click"`. See if you can figure out what's going on there. You might need to conditionally set the `mode` based on `isMobile`, so we control it explicitly.

CarlosNZ (changes requested):
Thanks @ria8651, working well.

Just one last fix please, regarding the Status Popover on mobile -- see latest comment

CarlosNZ (approved):
Okay, seems to be a Firefox-only problem with the mobile popover, so let's leave it as it for now.

PR #10657, opened 2026-02-25
10656 upgrade to yarn 4

### LINKED ISSUE #10656: Upgrade Yarn to 4.9.2
## The suggestion

We currently use yarn 1.19. Or at least I do, 'cause that's what's installed globally. The modern approach for several years now is to specify what package manager your project is using, from yarn 2.0.0 through to the latest yarn 4.9.2.

We've had one external contributor bamboozled by this archaic yarn version use and it'd be good for us to modernise. I'm also hoping that it's a starting point to further tech debt/complexity reductions such as [removing lerna](https://github.com/msupply-foundation/open-msupply/issues/7129), all towards faster build times and ~less cognitive load~ smaller agent contexts 😉.

Mainly breaking dev workflows, breaking CI/CD, breaking docker builds. No big deal 😆. Nothing that shouldn't be ultimately fixable, though.

I've already nudged claudecode along doing it in the background while I reviewed other PRs and OG work, so at this point very little 😉.

### Agreed Solution

Whatever claude figures out...

Chris-Petty:
Fixes #10656

# 👩🏻‍💻 What does this PR do?

Claudecode migrating code base from yarn 1.19 to yarn 4.9.2.

One key shift that it decided to make that I'm on the fence about is that it's shifted the yarn settings to the root of the repo. I think I'm alright with this as we do indeed use yarn there too, and having one consolidated config seems good? Though if we have many different frontends that wanted their own package management that'd be a spanner. Fortunately that's not in roadmap! The node_modules also move to the root which is a little jarring, though it's gitignored anyway...

Further, my personal interest is to remove `lerna` and possibly consolidate the frontend packages into one normal package built by webpack. This would remove all the various `package.json` files and reduce the `client/` code base complexity considerably - I don't foresee any chance of building combinations of the frontend with random parts of what is currently core functionality. Obviously plugins are plugins, but I think the complexity of that should be limited to that 😉.

# 🧪 Testing

- [ ] `yarn start` from `/` works (starting server and client)
- [ ] `yarn start` from `/client/` works (just starts client)
- [ ] `yarn test` from `/client/`
- [ ] `yarn build` works
- [ ] actions work
- [ ] Docker builds work
- [ ] OMS apk build works
- [ ] OMS installer build works
- [ ] build plugins works
- [ ] build reports/forms works

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


Chris-Petty (commented):
Hate it when github ties up my comments as a review and I don't notice. So here's my thought from yesterday!

PR #10664, opened 2026-02-25
filter the PO table when creating an external inbound shipment

### LINKED ISSUE #10647: Can filter POs when creating inbound shipment external

When creating an external inbound shipment, you select a purchase order , like so:

Image

Imagine we had loads of POs - we should be able to filter the POs so we can find the right one -> implement filtering by 
- supplier
- PO number
- reference

Although the origina ticket #10209 suggested filtering - search may make more sense for reference and PO number

esmehm:
Fixes #10647 

# 👩🏻‍💻 What does this PR do?

This PR adds filters to each of the supplier, PO number and reference columns on the table of purchase orders when choosing a PO to link a new external inbound shipment to.
This is necessary to find the correct PO when the list is longer

image

The table is unpaginated and I have applied client side filtering so it does not need to reload as you type
If the table were especially large, then we should transition to server side filtering

# 🧪 Testing

Clic new inbound shipment external
See the PO mtable
Test out the filters together and apart, including no resolts
Ensure no problem with realistically large numbers of POs

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/invoices/src/InboundShipment/ListView/LinkPurchaseOrderModal.tsx
+++ b/client/packages/invoices/src/InboundShipment/ListView/LinkPurchaseOrderModal.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo } from 'react';
 import { InboundShipmentPurchaseOrderLineFragment } from '../api';
 import {
   useWindowDimensions,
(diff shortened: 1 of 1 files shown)


ria8651 (approved):
LGTM for now though there is a wider question of filtering in table vs filter dropdown. I also pushed a commit to remove the custom selection logic, I think it was just left over from when I was using `useSimpleMaterialTable`.

PR #10672, opened 2026-02-25
Use rust-analyzer-lsp plugin for claude

ria8651:
# 👩🏻‍💻 What does this PR do?

Automatically enables the rust and typescript language server protocols for Claude. Not sure if this is something that should be in the repository or just locally but seem as anyone working on this repo with Claude will want these it might make sense.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,6 @@
+{
+  "enabledPlugins": {
+    "rust-analyzer-lsp@claude-plugins-official": true,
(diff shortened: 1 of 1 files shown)


Chris-Petty (approved):
LGTM. I was doing some reading around the `.claude` and the docs seem to intend it to be committed and shared. I think these settings at least are completely appropriate to be standard. That said, I just opened an issue to gitignore `.claude/worktrees`

I imagine these settings will help cloud based runners too!

PR #10690, opened 2026-02-26
10577 Admin global configuration of table defaults

### LINKED ISSUE #10577: Global column placement
[from this trello card](https://trello.com/c/8NFd1jGa/376-civ-uat-follow-up)

There is a requirement to align internal order column to be in a particular order

On central server, can 'save' table configuration as 'global'. This will be the default view of the table on all installations (configuration needs to sync). Current 'app_data' configurations should overwrite global ones (should merge them ?)

Do it as a plugin: https://github.com/msupply-foundation/civ-plugins/issues/29 -> sounds more like a general functionality

Do it by store or store tag -> future

### Moneyworks Jobcode 🧰

OMS:CIDE

CarlosNZ:
Fixes #10577

## Summary

- Adds a new `GlobalTableConfigs` global preference that allows admin users on the central server to save table viewing configurations (column visibility, ordering, sizing, density, pinning) as defaults for all users
- When a non-admin user visits a table with no custom config, the admin-configured defaults are applied instead of hard-coded defaults
- Admin users see a "Save table config as global default" button in the table Settings menu (requires `EditCentralData` permission)
Screenshot 2026-02-27 at 11 02 41 AM

- Resetting a table restores to global defaults for non-admin users, and to hard-coded defaults for admins
- Individual column resets (order, visibility, sizing, pinning) also respect global defaults
- Saving an empty config (no customizations) removes the table entry from the preference
- Snackbar notifications on save success/failure

### Backend
- New `GlobalTableConfigs` preference (Rust trait impl, `PrefKey` variant, `PreferenceProvider` registration)
- GraphQL resolver on `PreferencesNode` + mutation input on `UpsertPreferencesInput`
- Hidden from admin Preferences UI

### Client
- GraphQL query field + mutation for `globalTableConfigs`
- `useGlobalTableDefaults(tableId)` hook reads global defaults from preferences
- `useSaveGlobalTableConfig()` hook saves current table state to the preference
- All 6 table state hooks fall back to global defaults when no local config exists
- `SettingsMenu` shows save button for admin users, passes global defaults to reset handlers

### Bug fix
- Fixed `useColumnOrder` comparing against `state` instead of `initial` in `differentOrUndefined`, which caused reset to write the default value to localStorage instead of clearing it
- Also fixed the column order reset, which was being called at the wrong time when resetting all table properties

## Testing

- [ ] Log in as admin on central server, configure a table, click "Save table config as global default" — verify success snackbar
- [ ] Log in as a non-admin user — verify the table loads with the admin's saved defaults
- [ ] As non-admin, customize a column — verify the user's change takes priority
- [ ] As non-admin, reset table — verify it resets to admin's global defaults
- [ ] As admin, reset table — verify it resets to hard-coded defaults
- [ ] As admin, save with no customizations — verify the table entry is removed
- [ ] Verify the button does NOT appear for users without `EditCentralData` permission
- [ ] On a remote server (after sync), verify non-admin users see global defaults

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/authentication/api/operations.generated.ts
```diff
@@ -270,6 +270,7 @@ export type PreferencesQuery = {
     itemMarginOverridesSupplierMargin: boolean;
     showIndicativePriceInRequisitions: boolean;
     isGaps: boolean;
+    globalTableConfigs: any;
```
CarlosNZ: The `any` is auto-generated, presumably because the back-end type is arbitrary JSON. That okay?

client/packages/common/src/authentication/api/operations.graphql
```diff
@@ -225,5 +225,19 @@ query preferences($storeId: String!) {
...
+  $storeId: String!
+  $input: UpsertPreferencesInput!
+) {
+  centralServer {
+    preferences {
+      upsertPreferences(storeId: $storeId, input: $input) {
+        ok
+      }
+    }
   }
```
CarlosNZ: I asked Claude why we could't just use the existing mutation for saving global prefs, this was the explanation:

> Yes, there's an existing upsertPreferences mutation in [operations.graphql](vscode-webview://0vqoa02ht1jdcp6lmklj98u5hq8u5cslj7a8pqujvr97ie51ea7c/client/packages/system/src/Manage/Preferences/api/operations.graphql) in the system package — it's identical in shape to our saveGlobalTableConfigs. However, we can't directly reuse it because:
> 
> - common cannot import from system (dependency goes system → common, not the other way)
> - The useSaveGlobalTableConfig hook lives in common (because the table state hooks that call it are in common)
> - So the hook needs a generated SDK function available within common
> 
> The two options would be:
> 
> 1. Keep the mutation in common/operations.graphql (current approach) — minimal duplication, each package has the SDK it needs
> 1. Move the entire save hook to system and pass onSaveAsGlobalDefault down from a higher level — but useBaseMaterialTable in common is where it's wired up, so this would require significant restructuring
> 
> The current approach is the pragmatic one. The duplicated mutation definition is just 8 lines and lives next to the code that uses it, which is the standard pattern in this codebase (each package defines the operations it needs).

client/packages/common/src/ui/layout/tables/tableState/useColumnOrder.ts
```diff
@@ -61,17 +64,14 @@ export const useColumnOrder = <T extends MRT_RowData>(
             ? updaterOrValue(prev)
             : updaterOrValue;
 
-        const savedColumnOrder = differentOrUndefined(
-          newColumnOrder,
-          state ?? initial
-        );
+        const savedColumnOrder = differentOrUndefined(newColumnOrder, initial);
```
CarlosNZ: This change ensures the "hasSavedState" gets correctly reset when resetting column order, previously it was just saving the "initial" value as the new value, since it differs from the current state.

CarlosNZ:
> Can you outline what testing you've done?

As expected, it fails gracefully with realistic bad data (realistic being missing or misnamed columns, invalid values) -- it just ignores them.
With completely bad data (such as primitive values when expecting arrays), it fails properly, but that's no different to how it is already.

client/packages/common/src/authentication/api/operations.generated.ts
jmbrunskill: Find in genreated file IMHO

client/packages/common/src/ui/layout/tables/tableState/useColumnSizing.ts
```diff
@@ -6,12 +6,16 @@ import {
 } from 'material-react-table';
 import { getSavedState, updateSavedState, differentOrUndefined } from './utils';
 import { useDebounceCallback } from '@common/hooks';
+import { useGlobalTableDefaults } from './useGlobalTableConfig';
 
 export const useColumnSizing = (tableId: string) => {
+  const globalDefaults = useGlobalTableDefaults(tableId);
```
jmbrunskill: I think it's good having column sizing but I can't help but think maybe it will create problems for tablets when they setup the sizes for desktop. Almost want different settings for different devices but that's getting complicated!

client/packages/common/src/ui/layout/tables/tableState/useGlobalTableConfig.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { mutateAsync } = useMutation(
+    async ({
+      tableId,
+      state,
+    }: {
+      tableId: string;
+      state: ManagedTableState;
+    }) => {
+      const currentConfigs =
+        (globalTableConfigs as Record<string, ManagedTableState>) ?? {};
```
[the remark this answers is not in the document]
jmbrunskill: I agree a comment here would be nice

jmbrunskill:
> > Can you outline what testing you've done?
> 
> As expected, it fails gracefully with realistic bad data (realistic being missing or misnamed columns, invalid values) -- it just ignores them. With completely bad data (such as primitive values when expecting arrays), it fails properly, but that's no different to how it is already.

Validated this too, if it's invalid JSON you get a page crash but should only happen if we manually hack the json.

PR #10727, opened 2026-03-04
#10707 export all assets (not just visible)

### LINKED ISSUE #10707: Export assets only export the visible lines on the current page
## What went wrong? 😲

When I click Export Assets I can only export what is visible on the current page.

A work around is to display more on the page - but it is not intuitive that this step is required prior to export.

https://github.com/user-attachments/assets/a2050d4e-e48f-4658-8f8f-fb14376facb0

Image

## Expected behaviour 🤔

Export function exports all assets by default

## How to Reproduce 🔨

1. Go to Assets
2. Click on Export
3. See only currently visible lines on the page are exported

## Your environment 🌱

- Open mSupply Version: https://demo-open.msupply.org: (v.2.15.2)

ria8651:
Fixes #10707

# 👩🏻‍💻 What does this PR do?

The asset catalogue export was only exporting the rows visible on the current page, because `AppBarButtons` received the already-paginated `assets` prop from the list view.

This PR adds a `useAssetCatalogueListAll` hook that lazily fetches all asset catalogue items (without pagination) when the user clicks Export. `AppBarButtons` now calls this hook directly, so the exported CSV always contains all matching assets regardless of the current page size.

- The new hook respects the current sort and filter params from the URL (so if the user has filtered the list, the export reflects those filters). Is this the expected behaviour?
- Made a new api hook in order to make it async - gives a spinner while loading for lots of assets

# 🧪 Testing

- [ ] Go to Catalogue → Assets (asset catalogue list)
- [ ] Set page size to a small number (e.g. 10) so there are multiple pages
- [ ] Click Export
- [ ] Verify the downloaded CSV contains all asset catalogue items, not just the current page

# 📃 Documentation

- [x] **No documentation required**: bug fix, no change in expected behaviour

### DIFF
--- a/client/packages/system/src/Asset/api/hooks/useAssetList.ts
+++ b/client/packages/system/src/Asset/api/hooks/useAssetList.ts
@@ -7,7 +7,6 @@ import {
   useInfiniteQuery,
   useMutation,
   useQuery,
-  useUrlQueryParams,
 } from '@openmsupply-client/common';
 import { useAssetGraphQL } from '../useAssetGraphQL';
 import { ASSET } from './keys';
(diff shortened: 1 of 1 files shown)


mark-prins:
> The new hook respects the current sort and filter params from the URL (so if the user has filtered the list, the export reflects those filters). Is this the expected behaviour?

personally I don't think it is important as the export is usually for viewing in excel which already has sort and filter functionality!
I think it is better to export everything - but looking at some other implementations, we are not consistent!
Given the list below - please change to all rather than filtered, let's go with the majority 🙏 

page | exports
-- | --
stocktake | all 
locations | filtered
stock | filtered
internal orders | all
inbound shipments | all
supplier returns | all
requisitions | all
Outbound shipments | all
Customer returns | all

mark-prins (changes requested):
lovely ty! can you make it export all plz and then it is good to go

PR #10743, opened 2026-03-04
10608: use goods receiving permissions for external inbound shipments

### LINKED ISSUE #10608: Use goods received permissions for inbound shipments (external)

Implement and test access/modify/create permissions for inbound shipments. Currently we use the same set of permissions as inbound shipments but it might make more sense to use the goods received permissions.

esmehm:
Fixes #10608 

# 👩🏻‍💻 What does this PR do?

Use the pre existing user permissions from OG for goods receiving to govern access to external inbound shipments in OMS.
Internal inbound shipments already use's OG supplier invoice permissions, so this copies that logic over.

- The queries and mutations will only run if you have the relevant permissions
- Inbound shipments only appears in the nav bar if you have access to either internal or external or both
- You can only see the external / internal tabs if you have the right permissions
- Can only create new itnernal / external with the right permissions
- On dashboard have separated out the widget so have 1 for itnernal and 1 for external and can only see the relevant ones based on permissions
I chose to leave the "new inbound shipment" button on dashboard as internal only (and only visible if internal) because the modal for linking to PO is not yet exported, and I think we should leave this unless we have a product feedback to make it worth adding to dashboard. 

If have access to both
image
image

If have access to neither
image

If have access to only external
image
image

if have access to only internal
image
image

I chose to use separate queries instead of client side filtering as that was too leaky

# 🧪 Testing

Test a variety of supplier invoice and goods received permissions combinations to ensure that editing / creating / viewing are allowed and disallowed as expected

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


jmbrunskill:
@esmehm have a look at this PR. https://github.com/msupply-foundation/open-msupply/pull/10758 It needs more clean up and I think the permissions check thing in inboundshipment should be using some standard code somewhere (or moved to a standard place, but interested to see what you think. I'm most worried about the complexity of the batch_inbound which is quite complicated logic in your branch.

I think we can work around the need to give MutateInbound/QueryInbound permissions by creating a new type which validates the user has either `InboundQuery` or `ExternalInboundQuery`, then the graphql can fail quickly if they have neither but the service layer can check the specifics as part of the validation function.

PR #10765, opened 2026-03-05
10703 when i select the same date for the from/to dates in a report filter, i get no results

### LINKED ISSUE #10703: When I select the same date for the from/to dates in a report filter, I get no results
## What went wrong? 😲

Coming from issues #10274 and #10047 

Specifying a to/from date when running a report, if both dates are the same then no results are returned.
When checking - I can see that for the filters, no time is specified in the filter. that results in the current time being appended to both the from and to dates; being the same time for both, no results are returned.

This affects the following standard reports:
* inventory adjustment 
* Outbound shipments
* Inbound shipments (any from/to date results in no results)

## Expected behaviour 🤔
I would expect that the start of day (`00:00`) would be used for the from date, and the end of day (`23:59`) would be used for the to date.
It would be helpful to have a shared function which provides this functionality so that the next report does not repeat this buggy behaviour.

## How to Reproduce 🔨

1. Go to Reports, run one of the above
2. Check to see you have an inbound/invad/outbound for a given date
3. enter that date in the to and from fields for the report filter

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #10703

# 👩🏻‍💻 What does this PR do?
Allows same date filtering in reports

Would prefer if DateRange component got merged with generic DatePicker... Yes, will be more options to declare, but it'll stop dup code...

# 🧪 Testing

- [ ] Have some outbound / inbound shipments on a certain date
- [ ] Go to reports and generate either outbound / inbound report. Filtering by same date
- [ ] Should show results

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


aimee-mcneil-melville (changes requested):
The date filtering is working well 😁 

The issue also mentions that inbound shipment report isn't getting results with any dates - this is still the case for me. Can you address this part please?

PR #10766, opened 2026-03-05
10702 when importing a fridge sensor log file, the location specified for that sensor is not display in the breaches tab loacation column

### LINKED ISSUE #10702: When importing a fridge sensor log file, the location specified for that sensor is not display in the breaches tab loacation column
## What went wrong? 😲
The location filter also doesn't seem to work. It seems that when importing a Sensor fridge file, the location specified before the import is not added to temperature_logs records , same thing for temperature_breach records during importation. 

Sensor window

Image

Monitor window log
Image

Table Temperature breach

Image

Table Temperature log

Image

## Expected behaviour 🤔
If we are asked to specify a location, I assume it is to add that location to the temperature_log and temperature_breach records. 

## How to Reproduce 🔨

Have locations created
1. Go to 'Cold chain > Monitoring'
2. Click on 'Import Fridge Sensor' and use this file

[130400295025_202508060756.txt](https://github.com/user-attachments/files/25689441/130400295025_202508060756.txt)

3. When asked if you want to add a location, Click the OK button.
4. In the next modal, choose a location from the list and click OK. Click again to confirm.
5. Check the breaches tab. 
6. See error. Listed breaches will not have a location displayed. Same thing in the log tab
 
## Your environment 🌱

- Open mSupply Version:  2.16.00-RC-02260440
- Legacy mSupply Central Server Version:  v8.08

- Platform: Desktop

- Database type: SQLite

## Additonnal Notes
Found while testing #10440

roxy-dao:
Fixes #10702

# 👩🏻‍💻 What does this PR do?
Temperature log & breaches get their locations populated when they're created (from the sensor), so this PR updates the logs and breaches when location is changed on the sensor. P.S. should it only populate empty locations with new locations? I imagine you wanna see the latest location on breaches and logs right? Even if they were past breaches and logs? But that might be an expensive update 👀 

# 🧪 Testing

- [ ] Download file in issue
- [ ] Make sure you have mobile: Uses Vaccine Module on in OG store preferences
- [ ] Go to Cold Chain -> Monitoring > Import Fridge Sensor 
- [ ] Import file and click add location 
- [ ] Get redirected to modal
- [ ] Select location
- [ ] Go back to breaches
- [ ] See location populated

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [x] Postgres
- [x] SQLite
- [x] Frontend

### DIFF
--- a/server/repository/src/db_diesel/temperature_breach_row.rs
+++ b/server/repository/src/db_diesel/temperature_breach_row.rs
@@ -145,18 +145,6 @@ impl<'a> TemperatureBreachRowRepository<'a> {
             .filter(temperature_breach::id.eq_any(ids))
             .load(self.connection.lock().connection())?)
     }
-
-    pub fn update_location_id_by_sensor_id(
-        &self,
-        sensor_id: &str,
-        location_id: &str,
-    ) -> Result<(), RepositoryError> {
-        diesel::update(temperature_breach::table)
-            .filter(temperature_breach::sensor_id.eq(sensor_id))
-            .set(temperature_breach::location_id.eq(Some(location_id)))
-            .execute(self.connection.lock().connection())?;
-        Ok(())
-    }
 }
 
 impl Upsert for TemperatureBreachRow {


server/repository/src/db_diesel/temperature_log_row.rs
```diff
@@ -83,6 +83,33 @@ impl<'a> TemperatureLogRowRepository<'a> {
...
+            .execute(self.connection.lock().connection())?;
+
+        if rows_updated == 0 {
+            return Ok(());
+        }
+
+        let logs = temperature_log::table
+            .filter(temperature_log::sensor_id.eq(sensor_id))
+            .filter(temperature_log::location_id.eq(location_id))
+            .load::<TemperatureLogRow>(self.connection.lock().connection())?;
```
Chris-Petty: nit/thought - can't `diesel::update` return the rows updated in the one query rather than just the number of rows updated? Could doge this query :)

server/repository/src/db_diesel/temperature_log_row.rs
```diff
@@ -83,6 +83,33 @@ impl<'a> TemperatureLogRowRepository<'a> {
...
+
+        for log in &logs {
+            self.insert_changelog(log, RowActionType::Upsert)?;
+        }
```
Chris-Petty: My one reservation with this approach is that the `TemperatureLogRowRepository` connection doesn't demand enforce that it's in a transaction, so some chance that this gets used one day and fails half way through and doesn't sync all the changes _and_ doesn't roll back. I suspect we wouldn't want to constrain the repo to transaction based connections though.

That said the service using this does use a transaction connection so seems all good as far as I can see!

Issue #5564, opened 2024-11-26
Upgrade react-query


lache-melvin:
## The suggestion

We rely heavily on react-query - we are now 2 major versions behind, latest is `@tanstack/react-query@v5.61` - we are v3

Docs for v5 show features/improvements for a few of our patterns - would be good to upgrade

### Why should we invest time in this?

Today I've run into:

- all useQuery calls accept an object, rather than the current "sometimes an object, sometimes a query function with options" setup
  - Saves us missing fields
  - Saves cognitive overhead of different implementations
- useInfiniteQuery improvements 
  - More params available for this - would be great to outsource pagination management, would be easier with this hook

Probably other optimisations too! It's good to keep on top of dependency updates as much as we can!

### Are there any risks associated with this change?

I don't think there are incompatibilities with react versions etc., but there will be a bit of work munging all our existing hooks into the correct structure. Will need to happen at some point though!

### How much effort is required?

A couple of hours? Maybe a little more, if things are finnicky to map. And a bit of smoke testing required afterwards!

andreievg:
We want to discuss upgraded (how much effort, what risk etc..) also react needs upgrading

andreievg:
@lache-melvin to create guidelines for upgrades, removing needs refinement

CarlosNZ:
There's a possible bug in our current version that is causing problems in production build.

From my research:

> One known issue in React Query v3 is that under certain conditions, error states don't properly reset loading indicators, especially in production builds where error handling might be optimized differently than in development.

>This specific issue with isLoading remaining true after errors has been reported by others using React Query v3, especially in production builds. It's possible that production build optimizations are affecting how the state machine in React Query transitions between states when errors occur.

This is what was causing [this issue](https://github.com/msupply-foundation/open-msupply/issues/7327), and it was particularly hard to debug because I could only reproduce it on a production build.

Whoever does this issue -- please re-instate the code I temporarily disabled in [this PR](https://github.com/msupply-foundation/open-msupply/pull/7391/files#diff-87e85657225ee309409be671f8f2c198ee5158a33dfd1f6c547f2f4234458c56) and check if the upgrade fixes the problem

marthakeezy:
Tested in #5985 - moving to Done ✅

PR #10770, opened 2026-03-05
Upgrade React to v19 and migrate react-query to @tanstack/react-query v5

### LINKED ISSUE #5985: Update to React 19
We are curerntly using react 18, but there is now a new version. Update to React 19 to keep things up to date.

ria8651:
Fixes #5985 and #5564

## What does this PR do?
Upgrades two major client dependencies that were significantly out of date:

- **React 18.3 → 19.2.4** across all client packages
- **react-query v3 → @tanstack/react-query v5** across ~240 files

## Staged commits (review commit-by-commit)

1. **Stage 1: Update dependencies** (14 files) — package.json + yarn.lock bumps for react 19, @tanstack/react-query v5
2. **Stage 2: Update import paths** (2 files) — `react-query` → `@tanstack/react-query`
3. **Stage 3: Migrate query hooks to v5 object syntax** (257 files) — Convert `useQuery`/`useMutation`/`useInfiniteQuery` from positional args to single-object API; `cacheTime` → `gcTime`, `keepPreviousData` → `placeholderData`, updated `getNextPageParam`/`initialPageParam`
4. **Stage 4: Rename isLoading to isPending** (28 files) — TanStack Query v5 mutation status rename
5. **Stage 5: React 19 type fixes** (9 files) — `RefObject<T | null>`, `FC` removal, `react-jsx` type shims
6. ~~**Stage 6: Fix infinite render loops**~~ — split out into #11443
7. **Stage 7: Remaining fixes & cleanup** (26 files) — `fetchNextPage` API, `bundledWebRuntime` removal, jest config, capacitor config
8. **Stage 8: Review fixes** — small cleanups from review feedback (see [Review fixes](#review-fixes) below)

---

## React upgrade
Updated react, react-dom, @types/react, and @types/react-dom in the root client/package.json and all 12 sub-package package.json files to ^19.2.4.

A React 19 strict-mode violation in `Host.tsx`'s `PreInit` (calling `logout()` directly during render — previously masked by `suspense: true`, now reachable since v5 dropped `suspense:`) is **fixed in #11443** alongside the other render-loop / console-warning cleanup. That PR is intended to land together with this one.

## react-query migration
The package has been renamed and restructured between v3 and v5. Key API changes applied across the codebase:

- Package: `react-query` → `@tanstack/react-query`
- `useQuery(key, fn, opts)` → `useQuery({ queryKey, queryFn, ...opts })`
- `useMutation(fn, opts)` → `useMutation({ mutationFn: fn, ...opts })`
- `useInfiniteQuery(key, fn, opts)` → `useInfiniteQuery({ queryKey, queryFn, initialPageParam: 0, getNextPageParam, ...opts })`
- `cancelQueries(key)` / `invalidateQueries(key)` → `cancelQueries({ queryKey: key })` / `invalidateQueries({ queryKey: key })`
- `cacheTime` → `gcTime`
- `suspense: true` removed (not supported in v5 useQuery)
- `notifyOnChangeProps: 'tracked'` removed (automatic tracking is now the default in v5)
- `queryKey` must always be an array in v5
- `keepPreviousData: true` → `placeholderData: keepPreviousData` (21 list query hooks; silently ignored in v5, causing list views to lose the "keep previous data while fetching" behaviour)
- `useMutation` result: `isLoading` → `isPending` (25+ hook files; `isLoading` was removed from the mutation result object in v5)
- `useQuery` `onError` callback removed (no longer supported in v5 — moved to global QueryClient error handling or `throwOnError`)
- `fetchNextPage({ pageParam })` → `fetchNextPage()` — pageParam argument removed in v5; page param is now derived exclusively from `getNextPageParam`
- Global error handling: `onError` in QueryClient defaults removed in v5; replaced with `QueryCache.subscribe()` listening for `event.action.type === 'error'` events

A jscodeshift codemod was written and run to handle the bulk of the mechanical API signature changes across 223 files. The remaining files (those importing hooks indirectly via `@openmsupply-client/common`, and the `keepPreviousData`/`isLoading` issues not covered by the codemod) were caught through code review and iterative browser testing.

### Root cause of persistent loading spinner
`QueryErrorHandler.tsx` previously set `notifyOnChangeProps: 'tracked'` as a global query default. In v5, `'tracked'` is no longer a valid value — it gets spread into a Set as individual characters (`Set { 't', 'r', 'a', 'c', 'k', 'e', 'd' }`), none of which match any real query result property name. This silently prevented all query subscriptions from ever notifying components, so `isFetching` was permanently stuck at `true` on every list view. Fixed by removing the setting entirely (v5 performs automatic prop tracking by default via `useSyncExternalStore`).

### Broken *All list mutation hooks
Several `use*All` hooks (used to fetch all records for CSV export) had corrupted `mutationFn` entries after the codemod pass — the `queryKey` array was used as the `mutationFn` value, with the actual fetch function spread into the options object (invalid syntax). Fixed in:

- `useAssetsAll.ts`, `usePatientsAll.ts`, `useStocktakesAll.ts`
- `useOutboundsAll.ts`, `useCustomerReturnsAll.ts`, `useSupplierReturnsAll.ts`
- `useResponsesAll.ts`, `useRequestsAll.ts`

### Infinite query pagination fix
Also fixed a correctness bug in all 6 `useInfiniteQuery` hooks: the previous `getNextPageParam` implementation used `lastPage.data?.nodes?.length === rowsPerPage` to decide whether there was a next page. This causes an extra empty network request when the total number of items is exactly divisible by `rowsPerPage`. Changed to use the server-provided `totalCount`:

```ts
getNextPageParam: lastPage =>
  (lastPage.pageNumber + 1) * rowsPerPage < lastPage.data.totalCount
    ? lastPage.pageNumber + 1
    : undefined,
```

Applied to:
- `packages/coldchain/src/Equipment/api/hooks/document/useInfiniteAssets.ts`
- `packages/system/src/Asset/api/hooks/useAssetList.ts`
- `packages/system/src/Item/api/hooks/useItemStockOnHand/useItemStockOnHand.tsx`
- `packages/system/src/Item/api/hooks/useStockItemsWithStats/useStockItemsWithStats.tsx`
- `packages/system/src/Store/api/hooks/document/useStores.ts`
- `packages/programs/src/api/hooks/usePeriodList.ts`

## React 19 compatibility fixes
- **Global JSX namespace removed:** `@types/react@19` no longer provides a global JSX namespace. Added a shim at `client/typings/react-jsx/index.d.ts` that re-exports `React.JSX.*` into the global namespace, picked up automatically via tsconfig `typeRoots`.
- **`useRef<T>(null)` type change:** In React 19, `useRef<T>(null)` returns `RefObject<T | null>` instead of `MutableRefObject<T | null>`. Updated all function signatures that receive such refs to accept `RefObject<T | null>`.
- **`FC` no longer directly assignable to plain function types:** `FC<P>` carries extra optional properties (`displayName`, `defaultProps`) that aren't present on `(props: P) => ReactNode`. Removed `FC<P>` annotations from `ItemInput` in `AutocompleteMultiList.tsx` and `DefaultRenderOption` in `AutocompleteWithPagination.tsx`.
- **`ReactElement.props` is now `unknown`:** In React 19, `ReactElement.props` changed from `any` to `unknown`. Cast in `dashboard/src/hooks.tsx` where widget/panel props are accessed by name.
- **Capacitor config:** Removed deprecated `bundledWebRuntime` property (removed in `@capacitor/cli` v4+).

## React 19 "Maximum update depth exceeded" — fixed in #11443
Three `MaterialReactTable` instances on the inbound shipment line edit view triggered an infinite re-render loop under React 19 concurrent mode. Root cause was unstable function refs from `useFormatNumber()` propagating through `useMemo` deps into `useColumnOrder`'s `setState` effect. **Fixed in #11443** (`useCallback`-wrapped formatters, deduped `useMemo` deps, `React.memo` on `QuantityTable`). #11443 is intended to land together with this PR.

A related symptom — `ItemLedger.tsx` carries an `eslint-disable react-hooks/exhaustive-deps` to work around the same instability in `useFormatDateTime` — will be cleared up in #11443 once `useFormatDateTime` gets the same `useCallback` treatment as `useFormatNumber`.

<a id="review-fixes"></a>
## Review fixes
Small cleanups from [Carlos's review](https://github.com/msupply-foundation/open-msupply/pull/10770#issuecomment-4349831420):

- Dropped the now-dead `_shouldSuspend` parameter from `useInitialisationStatus` and the matching `true` arg in `Host.tsx`.
- `useStoresAll.ts`: removed stray blank line in the options object.
- `useDraftPurchaseOrderInboundLines.ts`: dropped redundant `isPending: isLoading` alias to match `useDraftInboundLines`.

Other items from the review are addressed in dedicated follow-up PRs:
- **#11443** — `Host.tsx` `PreInit` `useEffect` wrap, `useFormatDateTime` stabilization, `ItemLedger` eslint-disable removal.
- **#11446** — `useReportList` permission-error suppression (introduces a proper `silentOnPermissionDenied` per-query opt-out via `meta` flag, replacing the v3 `onError` shim).

The remaining review notes (Stage 7 scope, `as unknown as ApiException` casts, lost `mutationKey`s on `*All` hooks) were verified as non-issues — see the inline review reply for details.

## Notes for the reviewer
- The migration is largely mechanical but touches a large surface area (~240 files). The PR is split into **8 staged commits** for easier review — see the list above.
- The safest way to review is to focus on files that were **not** transformed by the codemod — primarily the `packages/common` hooks, `Host.tsx`, the infinite query hooks, and `QueryErrorHandler.tsx`, all of which required manual intervention.
- The `suspense: true` option has been removed from `useIsCentralServer` and `usePreferences`. These queries previously suspended their component trees; they now return `undefined` data on first render and components must handle the loading state themselves. Existing `enabled` guards and `?? {}` / `?? []` fallbacks already handle this correctly.
- `useInitialisationStatus` and `useMigrationStatus` also lost `suspense: true`; consumers fall back to safe defaults (`?? true` for migration status, manual `data?` checks for init status), so no behaviour change.
- This PR is intended to land together with **#11443**. Without #11443, the inbound-shipment line-edit view will hit "Maximum update depth exceeded" under React 19.
- No GraphQL schema or server changes.
- No user-visible UI changes — this is a pure dependency upgrade and internal refactor.

## Testing
- [ ] Log in as a store user — verify login page loads without console errors
- [ ] Navigate through main sections (Inventory, Invoices, Requisitions, Dashboard) — verify data loads correctly
- [ ] Open a list view that uses infinite scroll (e.g. Items, Assets, Stores) — verify pagination loads additional pages and stops at the last page without an extra empty fetch
- [ ] Can use #11093 for automated testing

## Documentation
- [x] No documentation required: no user facing changes (hopefully)

### Breaking Changes

### Issue Review

### Tests Pass
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/api/hooks/useInitialisationStatus.ts
```diff
@@ -8,7 +8,7 @@ import {
 
 export const useInitialisationStatus = (
   refetchInterval: number | false = false,
-  shouldSuspend?: boolean
+  _shouldSuspend?: boolean
```
CarlosNZ: Not used, so why not just drop it from the input? (I see it's being called in `Host.tsx`, which is doing nothing, so can remove from there too?

CarlosNZ:
# Code Review: PR #10770 — React 19 + TanStack Query v5 upgrade

*Generated by Claude*

## Overview
A large but mostly-mechanical migration: React 18.3 → 19.2.4 and `react-query` v3 → `@tanstack/react-query` v5 across ~240 files. A jscodeshift codemod handled 223 files; the rest were touched manually. The migration is mechanically thorough — I verified there are no leftover `cacheTime`, `suspense:`, `notifyOnChangeProps: 'tracked'`, `keepPreviousData: true`, positional `useQuery`/`useMutation`, or `from 'react-query'` patterns anywhere in `client/`.

That said, I found a handful of real concerns, plus some scope creep that should be teased apart from the upgrade.

---

## 🔴 Significant issues

### 1. `Host.tsx` — described fix is *missing*
The description says: *"Fixed by moving the logout() call into a useEffect."* It hasn't been. The only change to `Host.tsx` is removing `notifyOnChangeProps: 'tracked'`. `PreInit` still calls `logout()` directly during render:

```tsx
if (data.data.status == InitialisationStatusType.Initialised)
  return children;
// Server is not initialised — clear token
logout();        // ← still during render
return null;
```

With `useInitialisationStatus`'s `suspense: true` removed, this path is now reachable on first render and will trip React 19's "Cannot update a component while rendering a different component" warning (or, in concurrent mode, double-fire the side effect). Either apply the documented `useEffect` wrap or update the PR description to acknowledge the gap.

### 2. PR description's "Maximum update depth exceeded" fixes are split out, but the description still describes them as if they're in this PR
Stage 6 was split out as a follow-up (per the strikethrough), but the prose under *"React 19 'Maximum update depth exceeded' fix"* in the description still describes `IntlNumber.ts` and `TabTables.tsx` fixes. I confirmed neither file is touched in this PR.

If this lands without the follow-up, the inbound-shipment line-edit view will hit the update-depth crash in React 19. Worth either:
- merging the two PRs as a unit, or
- making the description match what's actually here.

Same root cause is already biting: `ItemLedger.tsx` had to drop `localisedTime` from its `useMemo` deps with an `eslint-disable react-hooks/exhaustive-deps` comment. The proper fix is to stabilize `useFormatDateTime`'s returns (same pattern as the deferred `useFormatNumber` fix), not to silence the lint rule one site at a time.

### 3. `useReportList` silently lost its per-query `onError` suppression
In `useReportList.ts`, v3 had:

```ts
onError: (e: Error) => {
  if (/HasPermission\(Report\)/.test(e.message)) return null;
  return [];
}
```

This was suppressing the global `setErrorMessage` from firing on permission-denied errors. In v5, `useQuery.onError` is gone and the new global handler in `QueryErrorHandler.tsx` subscribes via `QueryCache` — there's no per-query opt-out anymore. The `queryFn` does have a `try/catch` that swallows thrown errors, so in *most* paths this won't manifest. But if `reportApi.reports()` ever throws (e.g. a network/auth flavour of `HasPermission(Report)`), a user without report permissions will now see a global error message where previously they saw nothing. Worth either:
- filtering `HasPermission(Report)` in the `QueryCache.subscribe` callback, or
- adding an explicit catch in the query's `queryFn` for that error pattern.

### 4. `useInitialisationStatus` / `useMigrationStatus` lost suspense behaviour silently
PR notes call out `useIsCentralServer` and `usePreferences` for the `suspense: true` removal but don't mention these two:

- `useInitialisationStatus.ts` — kept the second arg as `_shouldSuspend?: boolean` (now ignored). `Host.tsx` still passes `true`. This is dead.
- `useMigrationStatus.ts` — `suspense: true` removed; the consumer falls back to `?? true` so it's safe, but worth noting.

Recommend deleting the unused `_shouldSuspend` parameter so callers stop passing dead arguments, and either documenting the suspense removal or wiring it up via React `Suspense` + a thrown promise if it was actually load-bearing.

---

## 🟡 Smaller concerns

### 5. Scope creep in Stage 7 — unrelated to the React/RQ upgrade
- `GrowthRow.tsx` replaces imports of `Column` and `HeaderRow` from `@openmsupply-client/common` with `TableRow`, and defines a **local** `Column<T>` type — sandwiched mid-imports — that's much looser than whatever the original was (`Partial<T>` makes everything optional). Why were `Column` and `HeaderRow` removed from `common`? Either intentional (separate concern → separate PR) or accidental masking of a type break.
- `client/packages/dashboard/src/index.ts` flips from a single named export to `export *` — widens the public surface unrelated to the upgrade.
- Test-fixture additions in `Prescriptions/**/utils.test.{ts,tsx}` and `StockOut/utils.test.ts` (`isVaccine`, `doses`, `volumePerPack`, `priority`, `isCancellation`, `store`) are GraphQL-fragment additions, not RQ migration. Bundling them in Stage 7 obscures their origin.
- `LineEditView/hooks/index.ts` drops `export * from './useDraftPrescriptionLines';`. The file doesn't exist on either branch, so the cleanup is correct — just unrelated to the upgrade.

### 6. `as unknown as ApiException` double-casts in dashboard widgets
In v5, `error` is `Error | TError` rather than `unknown`/`any`, so the previous `error as ApiException` cast needs a double cast now (`DistributionWidget.tsx`, `ReplenishmentWidget.tsx`, etc.). It compiles, but if errors aren't actually `ApiException` shape any more, the consumer (`StatsPanel`) may misread them. Worth one quick look at `StatsPanel` to confirm it's defensive.

### 7. Mutation keys dropped in `*All` hooks
The 8 corrupted `*All` hooks lost their `mutationKey` entirely (e.g. `useAssetsAll.ts`). v5 still accepts `mutationKey` in the options object — used by `useIsMutating` / `useMutationState`. Not a bug if nothing depends on those, but the data is silently gone.

### 8. Cosmetic / minor
- `useStoresAll.ts` has a stray blank line between `queryKey` and `queryFn`.
- Style inconsistency: `useDraftInboundLines.tsx` returns `isPending` directly, while `useDraftPurchaseOrderInboundLines.ts` aliases `isPending: isLoading` internally and exposes `isPending`. Same external shape; different internals. Pick one.

---

## 🟢 Looks good

- No stale v3 patterns leak through anywhere I checked (`cacheTime`, `suspense:`, `notifyOnChangeProps:`, `keepPreviousData: true`, positional `useQuery`/`useMutation`, `from 'react-query'`).
- All 6 `useInfiniteQuery` callers have `initialPageParam` set; pagination math improved to use server-side `totalCount` (eliminates the spurious empty-fetch on exact-page-divisor totals).
- 22 list hooks consistently apply `placeholderData: keepPreviousData`.
- The `QueryCache.subscribe` replacement for the global `onError` is implemented cleanly and unsubscribes correctly.
- The `*All` mutation hooks were fixed correctly after the codemod corrupted them.
- `RefObject<T | null>` and FC-removal changes for React 19 are minimal and well-scoped.

---

## Summary

The mechanical migration is solid. The three things I'd block on:
1. Apply the missing `useEffect` fix in `PreInit`, *or* note in the description that it's deferred.
2. Make sure the `IntlNumber`/`TabTables` follow-up PR lands together with this — without it, the inbound-shipment line-edit will crash in React 19.
3. Decide what to do about `useReportList`'s lost permission-error suppression (probably handle it in `QueryCache.subscribe`).

Everything else is cleanup that can happen in a follow-up — but consider pulling the Stage 7 unrelated changes (`GrowthRow`, dashboard index, test-fixture additions) out into separate PRs so the diff for *this* one is purely the upgrade.

CarlosNZ (commented):
Looks good @ria8651. I've eyeballed each file, made a couple of comments and done spot testing throughout and it all seems good.

Please see the [comment generated by Claude](https://github.com/msupply-foundation/open-msupply/pull/10770#issuecomment-4349831420) with some other considerations -- some of them aren't relevant, but I think there's a couple you might want to re-visit.

Cheers 👍

ria8651:
Thanks @CarlosNZ

**🔴 1. `Host.tsx` `PreInit` `useEffect` wrap** — you're right, the description claimed this fix but it wasn't actually in the diff. Moved to #11443 (which is themed exactly around this kind of React 19 render-loop / console-warning fix). Will land alongside this.

**🔴 2. `IntlNumber.ts` / `TabTables.tsx` deferred fixes** — covered in #11443; will land together with this PR. Updated this PR description to flag the dependency.

**🔴 3. `useReportList` permission suppression** — already planned in #11446, which introduces a proper `silentOnPermissionDenied` per-query opt-out (`meta` flag, checked in the cache-level handler) and rewrites `useReportList` to use it. Much cleaner than a regex match in the global handler. Will coordinate with that PR.

**🔴 4. `_shouldSuspend` parameter** — fixed in this PR. Dropped from `useInitialisationStatus` signature and the dead `true` arg in `Host.tsx`.

**🟡 5. Stage 7 scope creep** — fair point. Going to leave them in here rather than re-splitting; the cost of reshuffling commits outweighs the review benefit at this stage.

**🟡 6. `as unknown as ApiException` casts** — checked `StatsPanel`: it goes through `isPermissionDeniedException(error)`, which uses a structural `'message' in e && e.message === ...` check. Both `Error` and `ApiException` satisfy that, so the casts are safe in practice. Leaving as-is.

**🟡 7. `*All` `mutationKey` loss** — checked: nothing in the codebase calls `useIsMutating` or `useMutationState` against those keys, so they were dead state on `develop` too. Leaving the hooks keyless rather than restoring inert data.

**🟡 8. Cosmetic** — both fixed in this PR (blank line in `useStoresAll`, `isPending: isLoading` alias dropped from `useDraftPurchaseOrderInboundLines`).

`ItemLedger`'s `eslint-disable` will go away once #11443 stabilises `useFormatDateTime` (the same way it did `useFormatNumber`) — flagged that as a remaining task on the follow-up.

CarlosNZ (approved):
Okay, I'm happy, though before we merge we should:
- Get your follow-up PR in here (I'll look at that now)
- Run a test build with Android and Electron apps (if you haven't already?)

ria8651:
I've done the android test on #11443 which includes all the changes from here too.

PR #10778, opened 2026-03-06
#6487 Fix create patient in low internet

### LINKED ISSUE #6487: In low internet situations, when I lose internet while creating a patient, the entire patient record is not saved :(
## What went wrong? 😲

This will be hard to replicate. 
When I had low internet, I created a new patient (test patient, not real data) - went through the first modal where it searches for duplicates, and then got through to the second window: 

Image

At this point, I entered in details and tried to save - I then got a network error. The page went blank, and nothing happened. When I left the page to see what was happening, the patient I had created was not in the patient list :(((( after all my hard work! 

## Expected behaviour 🤔
Even if there is a network error, there should only be the error - it shouldn't lose all my work? Or should save after the first modal?

## How to Reproduce 🔨

Hard to repro sorry: 

1. Patchy internet required 
2. Create a new patient 
3. Try to save on the second window

## Your environment 🌱

- Open mSupply Version: 2.5.0
- Platform: browser, demo-djibouti

ria8651:
Fixes #6487

# 👩🏻‍💻 What does this PR do?

Adds error handling to all `useJsonFormsHandler` `saveData` call sites so that save failures (e.g. server down) show an error toast notification without closing their modal.

### Changes:

**Patient modals:**
- `CreatePatientModal.tsx` — modal no longer closes on save error; shows error toast
- `CreatePatientSlider.tsx` — same fix for the slider variant
- `EditPatientModal.tsx` — modal stays open on error; shows error toast
- `PatientDetailView.tsx` — save confirmation shows error toast on failure

**Program enrolment:**
- `ProgramDetailModal.tsx` — `saveData()` wrapped in try/catch; modal only closes (`reset()`) on success

**Encounter:**
- `Encounter/DetailView.tsx` — Footer save and delete flow both wrapped in try/catch with error toast
- `useSaveWithStatusChange.tsx` — all 3 `onSave()` call sites wrapped in try/catch

**Contact trace:**
- `ContactTrace/DetailView.tsx` — `onSave={saveData}` replaced with async wrapper + try/catch

**Translations:**
- Added `error.failed-to-save-program-enrolment`, `error.failed-to-save-encounter`, `error.failed-to-save-contact-trace` to `common.json`

- `useJsonFormsHandler`'s `saveData` has try/finally (no catch), so errors propagate to callers. All callers now handle these errors.
- `useSaveWithStatusChange` calls `onSave()` synchronously (fire-and-forget) — the try/catch will catch synchronous throws but not async rejections. This matches the existing calling pattern where `onSave` (`saveData`) returns a promise that is not awaited. A deeper refactor to make this fully async would be a larger change.

# 🧪 Testing

- [ ] Start the app with the server running
- [ ] Create a patient and get to the last page of the modal
- [ ] Close server
- [ ] Press create to submit the form
- [ ] The modal shouldn't close and a toast should be shown

# 📃 Documentation

- [ ] **No documentation required**: bug fix — no change in intended behaviour


mark-prins (changes requested):
Nice! works beautifully 😁 
I've toyed around with refactoring and bringing the notification into the useJsonFormsHandler - which reduces the caller code a bit. not convinced that it's worth it.

just needs the compilation fix and all good

PR #10784, opened 2026-03-06
update reports to show page numbers

### LINKED ISSUE #6952: Report page number not displaying
## What went wrong? 😲

Our standard reports have a page number in the footer. Two issues:
* The stocktake forms do not have the required css to display the number
* The number display does not work

After adding the style to the stocktake and generating a 1,000+ item stocktake, I get this:

Image

We have this in the footer

```
    </span>{{t(k="report.page",f="Page")}}: <span class="pageNumber"></span>
```

and this as the require css

```
.pageNumber::after {
    font-family: Tahoma;
    font-size: 8pt;
    counter-increment: page;
    content: "Page " counter(page);
}
```

which is making use of the css [counter](https://developer.mozilla.org/en-US/docs/Web/CSS/counter) function. However the element in the footer is only being rendered once and the counter isn't incremented.

## Expected behaviour 🤔
* The page number should show `Page: [incrementing number]`
* The stocktake forms should have page number displayed
* The additional `Page` needs to be removed from the css 

css needs to be 
```
.pageNumber::after {
    font-family: Tahoma;
    font-size: 8pt;
    counter-increment: page;
    content: counter(page);
}
```

## How to Reproduce 🔨

1. Go to inbound/outbound/returns
2. Create one of the above, with more than a single page of rows
3. Print

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

mark-prins:
Fixes #6952

# 👩🏻‍💻 What does this PR do?
Fixes the issue where page numbers are always `1` on standard forms.
I'd asked copilot to have a go in order to clear the old issue - and the attempt did not work.
Claude however, had no problem.

I've kept the footer.html files in place even though they are no longer used.

**style.css changes** 
— Added `@bottom-center` with `content: counter(page)` inside the existing `@page` block, and removed the old `.pageNumber::after` CSS rule (where it existed):

Form | Had .pageNumber::after | Had broken @page counter
-- | -- | --
outbound-invoice-portrait-with-logo | Yes, removed | -
outbound-invoice-landscape-with-logo | Yes, removed | -
repack | Yes, removed | -
inbound | - | Yes, replaced with @bottom-center
supplier-return | - (no CSS for it) | -
stock-take-with-quantity | - | -
stock-take-without-quantity | - | -
stock-take-detail-view | - | -

**footer.html changes**
Removed `<span class="pageNumber"></span>` (and the `{{t(k="report.page",f="Page")}}:` text in stock-take variants) from all 8 footer files.

# 🧪 Testing

- [ ]  Print a multi-page inbound shipment — footer should show Page: 1, Page: 2, etc. (not Page: Page 1)
- [ ] Print a multi-page outbound shipment (portrait and landscape with logo) — same check
- [ ] Print a multi-page stocktake (with quantity, without quantity, detail view) — page number should now appear in footer
- [ ] Print a repack report — page number displays correctly

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


ria8651 (approved):
Good improvement, most of the reports didn't have numbering in dev and the ones that did didn't work. Now most of the reports I tried seem to have working page numbers. 

Screenshot 2026-03-09 at 1 38 40 pm

Is supposed to say 'Page: #'? I'm only seeing the number. Though even if it is supposed to say 'Page: ' given that the page counters didn't work at all before this is a good improvement anyway.

PR #10834, opened 2026-03-11
Handle inactive indicators in requisitions

jmbrunskill:
Fixes [131](https://github.com/msupply-foundation/open-msupply-internal/issues/131)

# 👩🏻‍💻 What does this PR do?

Fixes bugs reported by client due to handling of inactive (aka deleted) program indicators.

The behaviour in this PR changes as such...

If you open an existing requisition with a disabled indicator, you'll see a message like this and the inputs will be disabled.

image

If you create a new requisition/internal order after the indicator has been disabled, the indicator won't show up.

This work was confusingly complicated by what is basically a bug from mSupply.

mSupply does check the `IndicatorRow` equivalent `is_active` (so it only shows newly activated indicator rows) 
However it doesn't check if the  associated `IndicatorColumn` rows are active, in the demo server these are all set to false.

Apparently this is because older versions of the code had default values for these added which default to `false`
Updating all existing columns to `true` would create a lot of sync out records so @JonnaQuismundo isn't keen to do that.

BTW, mSupply doesn't have any UI for editing enabling or creating columns, they're just hard coded to just `Value` and `Comment`, unless it's an HIV program which get some extra hardcoded stuff apparently.

So for now I'm aligning the mSupply and open-mSupply behaviour. So the is_active status for Columns is effectively ignored in both frontend and backend.

I've just linked the code to Disable the input for removed indicators even on in-progress requisitions, so they're effectively blocked from entry. 
I think this is ok as it's an edge case and shouldn't happen a lot, as long as it doesn't somehow block them from submitting their internal order.

DemoOMS database available here: https://drive.google.com/drive/folders/1FU84hfkzySfTglcTng4tC1dPJ7EKwboN

# 🧪 Testing

- [ ] Create a program using indicators in mSupply

image

- [ ] Create an internal order using this program
- [ ] Check that you see and can edit all your newly added indicators
- [ ] Remove an indicator
- [ ] Sync
- [ ] Try to open your internal order  again, check that the removed one is disabled (can't be edited but can still be seen)
- [ ] Try to create a new internal order , check that the deleted indicator doesn't show up at all.
- [ ] Repeat for response requisitions (try with sync too)

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite

### DIFF
--- a/server/service/src/requisition/program_indicator/query.rs
+++ b/server/service/src/requisition/program_indicator/query.rs
@@ -186,18 +186,14 @@ mod query {
         // Only active lines: line_a, line_b (not line_d_inactive)
         assert_eq!(lines_a.len(), 2);
 
-        // Only active columns: column_a, column_b (not column_c_inactive)
-        // 2 lines x 2 columns = 4
+        // Columns are NOT filtered by is_active (all columns are always included).
+        // This is intentional: mSupply historically defaults column is_active to false,
+        // so filtering would break existing data. Only lines are filtered when include_inactive = false.
+        // 2 active lines x 3 columns (column_a, column_b, column_c_inactive) = 6
         let columns_a_count = lines_a.iter().flat_map(|line| line.columns.iter()).count();
-        assert_eq!(columns_a_count, 4);
+        assert_eq!(columns_a_count, 6);
 
         // Verify no inactive lines are present
         assert!(lines_a.iter().all(|line| line.line.is_active));
-
-        // Verify no inactive columns are present
-        assert!(lines_a
-            .iter()
-            .flat_map(|line| line.columns.iter())
-            .all(|col| col.is_active));
     }
 }


.gitignore
```diff
@@ -1,5 +1,4 @@
 .DS_Store
-.claude/
```
jmbrunskill: THis shouldn't have been there I think?

.gitignore
```diff
@@ -19,5 +18,7 @@ client/dist
 ## Claude
 # `claude -w "feature-name"` will create "/.claude/worktrees/feature-name" for your worktree. Back at root though, git picks up this whole directory copy of the repo!
 worktrees
-.claude/*.local.json
+**/.claude/*.local.json
```
jmbrunskill: My claude started creating local files in backend (I think because I opened backend and frontend separtely) I think probably better not to do this but just in case people do I think this doesn't hurt

server/graphql/core/src/loader/program_indicator_value.rs
```diff
@@ -60,6 +60,9 @@ impl Loader<IndicatorValueLoaderInput> for IndicatorValueLoader {
             } else {
                 return Ok(HashMap::new());
             };
+        if loader_inputs.len() > 1 {
```
jmbrunskill: Sadly I've seen this. probably need to create a new issue for it, but don't think it's relevent to this directly...

jmbrunskill:
@copilot can you please fix the failing tests?

server/graphql/core/src/loader/program_indicator_value.rs
```diff
@@ -60,6 +60,9 @@ impl Loader<IndicatorValueLoaderInput> for IndicatorValueLoader {
...
+            log::error!("Multiple loader inputs provided to IndicatorValueLoader, but only one combination of period_id, store_id, and customer_name_id is supported. Only the first input will be used.");
```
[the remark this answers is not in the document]
esmehm: I can't see copilot's issue - can you? Surprised it didnt comment a link back here

PR #10836, opened 2026-03-11
V7 prototype

andreievg:
Relates to: https://github.com/msupply-foundation/open-msupply-internal/issues/133

There is example of tight coupling: https://github.com/msupply-foundation/open-msupply/pull/11253, which atm i prefer

This prototype is meant to show how a sync v7 system can work. There was a big focus on reducing boilerplate, with minimum updates, still using visitor pattern. A few macros to help with some diesel, including creating dynamic queries (should be able to have a way to expose them in graphql too, would probably need to extend the macro to mark them as objects).

I've asked claude to explain how it work with this prompt, you can find result in [v7_sync_prototype.md](https://github.com/msupply-foundation/open-msupply/blob/7fa8afcf4a09b00966fa47d52bdaecb914631e77/server/v7_sync_prototype.md#L1), I think it's very well explained.

Prompt:
```
I would like to create a PR explaining changes done in the last commit. The changes were done to have an example of new v7 sync version. It has changes beyond that scope and they should be explained too. I want it to be easily readable and split into sections, one section should be about macros that were added to help with diesel (and changing from using db enum to rust enum with extra 'other' type), one section should be about how the traits are structure to acheive minimal changes when new type needs to sync (this can reference the new create that was added to demonstrate those type working), etc..
Make an MD file for this
```

And then 
```
that's great, can you comment somewhere that a crate was added to demonstrate the new trait structure, also talk about the need for CLI changes (I am not sure now and forgot why they are needed), talk about how sync error is much easier to save and map now (previously there was a lot of mapping in the other sync_log)
```

I did some performance testing, after syncing a few stores, you can find [postgres backup here ](https://drive.google.com/drive/u/1/folders/13NNN505cnCIl3r2aNcof3cn1UgOEoh-c)

## How to test performance

Restore postgres backup, make sure you've place backup in the right folder based on central.yaml, and unzipped it:

`cargo run --bin remote_server_cli --features postgres -- --config-path './configurations/central.yaml' restore --backup-name D2026_03_11T03_57_59`

#### For central

`cargo run --features postgres -- --config-path ./configuration/central.yaml`
`yarn start -- -- --env API_HOST='http://localhost:8002' --port 3005`

#### For remote (remove sqlite database)

`cargo run`
`yarn start`

Now you can enter

http://localhost:8002 
test
pass

`note` production build is much faster 

there is [postgres config](https://github.com/msupply-foundation/open-msupply/pull/10836/changes#diff-4daf961533d8b09bd0ca16694a2422697563a9b94593c77df07a84c8a0bd6675) here for optimised performance


server/graphql/core/src/lib.rs
```diff
@@ -135,6 +138,15 @@ pub fn auth_data_from_request(http_req: &HttpRequest) -> RequestUserData {
...
 
+#[derive(Serialize, Deserialize, TS)]
+struct TStypes {
+    sync_lov_v7: SyncLogV7Row,
+}
+
+pub fn export_graphql_typescript(path: PathBuf) {
+    TStypes::export_all_to(path).unwrap();
+}
+
```
ria8651: Not related to sync v7 but surely the json serialisation of rust errors with generated ts types could be used instead of graphql unions for typed error handling in the frontend in general? Ideally the errors would be in the response like with unions but could use the path to try put then back into the response though might make a type mess as it isn't defined what errors can go where. At least it would be nice for the generic graphql errors.

server/service/src/sync_v7/serde.rs
```diff
@@ -0,0 +1,17 @@
...
+};
+
+pub fn serialize(
+    connection: &StorageConnection,
+    changelog: &ChangelogTableName,
+    id: &str,
+) -> Result<Option<serde_json::Value>, SyncRecordSerializeError> {
+    for visitor in translators() {
+        if let Some(value) = visitor.serialize(connection, changelog, id)? {
+            return Ok(Some(value));
```
Chris-Petty: Software pattern semantics, but I'm pretty sure if you return on the first visitor then it's not a visitor pattern 😁. It's pretty much a match statement.

Sorry this visitor/match discussion is pretty sporadic, but I'm going to put comment about it on the `translators()` method code...

server/service/src/sync_v7/prepare.rs
```diff
@@ -0,0 +1,26 @@
...
+};
+
+use super::serde::serialize;
+use crate::sync_v7::sync::SyncRecordV7;
+
+pub(crate) fn prepare(
+    connection: &StorageConnection,
+    changelog: ChangelogRow,
+) -> Result<SyncRecordV7, SyncError> {
+    let Some(data) = serialize(connection, &changelog.table_name, &changelog.record_id)? else {
```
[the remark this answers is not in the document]
Chris-Petty: Last but not least - what we have now probably won't prohibit doing these optimisations later. The API will still be push(batch)s, integrate on central, pull(batch)s, integrate on remote. So reiterating the question of whether now is the time, or if performance of the current approach is good enough for now?

Chris-Petty (commented):
Haven't tested but read through the bulk of it, at least the sync API stuff. The `Conditional` stuff and macros I kind of just got the gist and moved on as I can't add much there I think

Issue #10812, opened 2026-03-09
Facility level coverage / wastage rate for population based forecasting


andreievg:
**Note:** this should include coverage rate as well as wastage rate

## Is your feature request related to a problem? Please describe 👀

[From spec here](https://docs.google.com/document/d/1p-whb4WIZ0cG5XjQZPse-0Rwu-aIiEambcpDHnBuCuc/edit?tab=t.0#heading=h.5snauon6lsf8)

Facility based wastage rate is required for population based forecasting. 

The basic example, consider two facilities, both order monthly, one with target doses (based on population) is 3000 and the other with 60. Lets deal with BCG vaccine, where a vial of 20 is much cheaper per dose then a vial of 5 or 1, and once you open the vial it needs to be thrown out within a day. We can see that there will be a big difference in wastage rates for these facilities. 3000 / 30 = 100 patients per day = 100 / 20 = 5 vials used, no wastage. 60 / 30 = 2 patients par day = 20 - 2 = 18 out of 20 doses are wasted. So when supplying the first facility was don't need a wastage multiplier (3000 / 20 = 150 vials + some buffer will do), whereas we don't take wastage into account for the second site, we would order 60 / 20 = 3 vials, whereas we really need a vial per day (1 - (18/20) * 100) = 10 wastage rate multiplier = 3 * 10 = 30 vials.

We would need to make sure, Facility based wastage rate:
- can be set at vaccine course level on central server
- if exists for facility, use in population based forecasting calculations
- show in the calculations in internal order line edit view

## Describe the solution you'd like 🎁

Three questions:
- ~~Core or plugin -> I feel like this is core as per example above, it looks like a common use case (why have we not heard about it before ?), doing this as plugin will require a lot of work, for something that I think will be used once or then added to core~~ To be done in core
- ~~Where to edit -> on central server when configuring a [vaccine course](https://docs.msupply.foundation/docs/programs/immunisations/#:~:text=Vaccine%20course%20name%2C-,wastage%20rate,-%2C%20dose%20schedule.), it's a modal already @mark-prins can it be a modal within a modal, and then a simple table of facility (store), store is searchable, wastage rate value can be set or cleared ?~~ Central server under vaccine course
[...]

PR #10856, opened 2026-03-13
10812 facility level wastage rate for population based forecasting

roxy-dao:
Part 1 of #10812

# 👩🏻‍💻 What does this PR do?
This PR adds support for `store level` wastage rates on vaccine courses allowing stores to overwrite the global wastage rate used in population based forecasting calculations. Changes include: 
-  New `vaccine_course_store_wastage` table that stores wastage rates against store and vaccine course
- Allow wastage rates to be recorded when inserting/updating vaccine course
- Translations to pull records to remote sites
- Vaccine store wastage is exposed on vaccine course node to show in frontend
- Updates forecasting calculations to take store specific wastage rates into account

Frontend coming soon...

# Questions
- Should we also store item_id? What happens when a vaccine course has more than one vaccine and each has a different wastage rate? Can do if that's something we should consider at this point (:
- Do we need to indicate in the forecasting calculations that the calculation is using store wastage rate instead of global? 

# 🧪 Testing

- [ ] Would need to use GraphQL to insert vaccine course store wastage or wait until frontend PR is done
- [ ] Make sure tests pass

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/repository/src/db_diesel/vaccine_course/vaccine_course_store_wastage_row.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fn insert_changelog(
+        &self,
+        row_id: String,
+        action: RowActionType,
+        changelog_store_id: String,
+    ) -> Result<i64, RepositoryError> {
+        let row = ChangeLogInsertRow {
+            table_name: ChangelogTableName::VaccineCourseStoreWastage,
```
aimee-mcneil-melville: A migration is needed to add the VaccineCourseStoreWastage table to the changelog_table_name enum for postgres

server/repository/src/migrations/v2_17_00/vaccine_course_store_wastage.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fn migrate(&self, connection: &StorageConnection) -> anyhow::Result<()> {
+        sql!(
+            connection,
+            r#"
+                CREATE TABLE vaccine_course_store_wastage (
+                    id TEXT PRIMARY KEY NOT NULL,
+                    vaccine_course_id TEXT NOT NULL REFERENCES vaccine_course(id),
+                    store_id TEXT NOT NULL REFERENCES store(id),
```
aimee-mcneil-melville: Should this reference name_link(id) instead of store(id)?

server/service/src/vaccine_course/update.rs
aimee-mcneil-melville: If a store wastage rate is cleared, the vaccine course update should delete the associated row or set it to Null. Deleting the row is likely tidier and more similar to how doses are handled (soft delete as they have related records)

server/service/src/vaccine_course/update.rs
```diff
@@ -126,6 +148,11 @@ pub fn update_vaccine_course(
                 dose_repo.upsert_one(&dose.to_domain(input.clone().id))?;
             }
 
+            let store_wastage_repo = VaccineCourseStoreWastageRowRepository::new(connection);
+            for wastage in input.clone().store_wastage_rates {
+                store_wastage_repo.upsert_one(&wastage.to_domain(input.clone().id))?;
+            }
```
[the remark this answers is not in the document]
aimee-mcneil-melville: Also raised by @aneeshthunga26 about clearing a field

There are no specs about this, but I would expect the store wastage rate to be able to be cleared so that the global default can be used instead. Distinct from setting to 0 which = 0% wastage, it would require Nulling the field or deleting the row

server/service/src/vaccine_course/update.rs
aimee-mcneil-melville: I had tested only this pr in graphql - looks to be working in part 3 with front end, thanks Roxy 😁

aimee-mcneil-melville:
Will approve once other PRs merged in and tests passed - changes to fix postgres are in part 2

Issue #10755, opened 2026-03-05
PO : Not allow adding a new line that is for the same item


rukmal-msupply:
## What went wrong? 😲

As per the spec, it should allow to add a new line that is already exists in the PO to facilitate multiple deliveries .

<img width="879" height="564" alt="Image" src="https://github.com/user-attachments/assets/59a32cbc-03ea-404e-b27d-3a7329f55144" />

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Create a PO
2. Add an item to the PO line
3. Click again and search the same item in the item search
4. Requirement is not satisfied.

## Your environment 🌱

 
- Open mSupply Version: 2.17.0
- Legacy mSupply Central Server Version:

- Platform:

- Database type:

jmbrunskill:
@mark-prins  @craigdrown I assume this is fucntinoality we want to keep? Same item multiple times?
I feel lik that 's what we implemented originally so probably got changed in subsquent updates? Or maybe I'm just remembering wrong...

mark-prins:
yes please - as per the spec. We'd like to be able to add multiple lines of the same item.

PR #10900, opened 2026-03-17
#10338 track purchase order line and invoice line relationship

### LINKED ISSUE #10338: Track relationship between purchase order lines and invoice line

When creating an inbound shipment from a purchase order it would be good to keep track of which lines in the inbound shipment are associated with lines from the purchase order. This could be done by adding a `purchase_order_line_id` to the `invoice_line` table.

ria8651:
Fixes #10338 and #10755

# 👩🏻‍💻 What does this PR do?

Adds a `purchase_order_line_id` column to the `invoice_line` table to directly track which purchase order line each invoice line corresponds to. This is the foundation for supporting POs with multiple lines of the same item.

When I did inbound shipments external I assumed POs could only have one line for each item. As such the link between invoice lines and PO lines was inferred via item matching in SQL views. Now the relationship is explicit and set up when the invoice line is added.

### Backend (API) Changes

**Database / Migration**
- New column: `invoice_line.purchase_order_line_id TEXT REFERENCES purchase_order_line(id)`
- `invoice_line_stats` view emptied (kept as stub) — `purchase_order_line_id` now lives on the row itself
- `purchase_order_line_stats` view simplified — joins on `il.purchase_order_line_id = pol.id` instead of item-based matching

**GraphQL Schema**
- `InsertInboundShipmentLineInput` — new optional field: `purchaseOrderLineId: String`
  - **Required** when the parent invoice has a `purchase_order_id` (external inbound shipment)
  - Validated: returns `PurchaseOrderLineIdRequired` or `PurchaseOrderLineDoesNotExist` errors
- `InvoiceLineNode.purchaseOrderLine` — now resolves from `invoice_line_row.purchase_order_line_id` instead of the stats view

**Service Layer**
- `InsertStockInLine` input: new field `purchase_order_line_id: Option<String>`
- Validation: when invoice has `purchase_order_id`, the field is required and the referenced PO line must exist
- Generate: passes the field through to `InvoiceLineRow`
- New error variants: `PurchaseOrderLineIdRequired`, `PurchaseOrderLineDoesNotExist`
- All non-inbound-shipment callers (customer returns, inventory adjustments, stocktake, etc.) pass `purchase_order_line_id: None`

### Frontend Changes

Screenshot 2026-03-19 at 10 23 11 am

**Unified `InboundLineEdit` modal** — the single modal now handles both PO-based and non-PO inbound shipments:
- Detects whether the shipment has an associated purchase order (`hasPurchaseOrder`)
- When PO-based: `InboundLineEditForm` renders a `PurchaseOrderLineSelect` autocomplete showing `#lineNumber ItemName (Code) - qty units`
- When non-PO: renders the standard `StockItemSearchInput` as before
- `useDraftPurchaseOrderInboundLines` — dedicated hook for draft line management keyed by PO line, pre-populates from PO line data (pricing, pack size, remaining quantity)
- `useDraftInboundLines` — unchanged, handles item-based draft lines for non-PO shipments
- OK & Next button cycles through available PO lines (PO mode) or items (non-PO mode)

**Detail view changes:**
- External inbound shipments now group by PO line number instead of item code
- New "PO Line #" column (external only)
- Row click passes `purchaseOrderLine.id` to the edit modal for correct PO line resolution

There are two `// REVIEW:` comments in the code:

1. **`validate.rs:116`** — Repository call in validation step to check PO line exists. Is this acceptable or should it be structured differently?

2. **`DetailView.tsx:176`** — Grouping by `purchaseOrderLine.lineNumber` for external shipments could be confusing since there's no UI feedback about what field is being grouped. But grouping by item doesn't make sense for external IS when multiple PO lines can share the same item.

**Breaking GraphQL changes:** `purchaseOrderLineId` is a new optional field on `InsertInboundShipmentLineInput`. It's only required when the invoice has a `purchase_order_id`, so existing integrations that don't use external inbound shipments are unaffected.

# 🧪 Testing

- [ ] Create a purchase order with multiple lines, including multiple lines for the same item
- [ ] Create an inbound shipment from that PO
  - [ ] With add with all lines
    - [ ] Should include a IS line for each PO line
    - [ ] Quantities should be correctly calculated
  - [ ] With add with no lines there shouldn't be anything in the IS
- [ ] Verify the detail view groups by PO line number and shows the "PO Line #" column
- [ ] Click a row to edit — verify the correct PO line is pre-selected in the modal
- [ ] Use "Add Item" — verify the PO line picker shows available (unassigned) PO lines with line numbers
- [ ] Use OK & Next to cycle through PO lines
- [ ] Verify a normal (non-PO) inbound shipment still works as before (item grouping, item picker)
- [ ] Verify `purchase_order_line_stats` view returns correct shipped/received/in-transit numbers using the new join
  - [ ] Check the IS tab on the purchase order
  - [ ] Check the delivery status tab on the IS

# 📃 Documentation

  1. GraphQL API docs: new `purchaseOrderLineId` field on `InsertInboundShipmentLineInput`
  2. External inbound shipment line editing workflow

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/invoices/src/InboundShipment/api/hooks/useDraftPurchaseOrderInboundLines.ts
```diff
@@ -0,0 +1,230 @@
...
+      setIsDirty(true);
+    }
+  }, [existingLines, purchaseOrderLine, invoiceId, isDirty, setIsDirty]);
+
+  const addDraftLine = (initialPatch?: Partial<DraftInboundLine>) => {
+    if (!purchaseOrderLine) return;
+    const newLine = createDraftLine(purchaseOrderLine, invoiceId, initialPatch);
+    setIsDirty(true);
+    setDraftLines(prev => [...prev, newLine]);
+  };
```
esmehm: `addDraftLine` is not wrapped in `useCallback`, creating a new reference on every render. Wrap it:

```suggestion
  const addDraftLine = useCallback(
    (initialPatch?: Partial<DraftInboundLine>) => {
      if (!purchaseOrderLine) return;
      const newLine = createDraftLine(
        purchaseOrderLine,
        invoiceId,
        initialPatch
      );
      setIsDirty(true);
      setDraftLines(prev => [...prev, newLine]);
    },
    [purchaseOrderLine, invoiceId, setIsDirty]
  );
```

client/packages/invoices/src/InboundShipment/api/hooks/useDraftPurchaseOrderInboundLines.ts
```diff
@@ -0,0 +1,230 @@
...
+      setDraftLines(prev => prev.filter(line => line.id !== lineId));
+    } else {
+      setDraftLines(prev =>
+        prev.map(line =>
+          line.id === lineId ? { ...line, isDeleted: true } : line
+        )
+      );
+      setIsDirty(true);
+    }
+  };
```
esmehm: `removeDraftLine` is not wrapped in `useCallback` and closes over `draftLines` directly. Since it's in the `columns` useMemo deps in `InboundLineEditCards`, every keystroke causes columns to recalculate → all Cell components remount → input focus lost.

Use `useCallback` with the functional updater to avoid the `draftLines` closure:

```suggestion
  const removeDraftLine = useCallback(
    (lineId: string) => {
      setDraftLines(prev => {
        const batch = prev.find(line => line.id === lineId);
        if (!batch) return prev;
        if (batch.isCreated) {
          return prev.filter(line => line.id !== lineId);
        }
        setIsDirty(true);
        return prev.map(line =>
          line.id === lineId ? { ...line, isDeleted: true } : line
        );
      });
    },
    [setIsDirty]
  );
```

esmehm (commented):
Focus loss bug: `addDraftLine` and `removeDraftLine` are not wrapped in `useCallback`, so they get new references on every render. Since `removeDraftLine` is a dependency of the `columns` useMemo in `InboundLineEditCards`, each keystroke triggers a full column recalculation, which recreates all inline `Cell` components, remounting the inputs and stealing focus after the first digit.

Compare with `useDraftInboundLines` where `removeDraftLine` IS wrapped in `useCallback` — that's why the bug only appears for external shipments.

Both functions should be wrapped in `useCallback` using the functional updater form of `setDraftLines` to avoid closing over `draftLines`.

^ above written by claude. However, the essence is there: there is a bug when typing in the modal for external IS that you loose focus after each keystroke, and this is my proposed resolution (I tested locally)

The rest LGTM! Although it doesnt need to be part of this PR, I would recommend picking up or flagging the above product question so we don't leave loose ends hanging afterwards

ria8651:
Added followup issue #10942

PR #10919, opened 2026-03-19
#10915 Combined inbound shipment list view

### LINKED ISSUE #10915: Combined list view

Currently in inbound shipments we have an internal and external tab. This was causing confusion with things like internal shipments from external suppliers showing in the internal tab as the external tab was actually linked PO tab.

To avoid a slurry of new tabs for the different kinds of possible inbound shipments we should instead have a single filterable list view with different colours and icons to distinguish the different kinds of shipments.

Image

This covers the different kinds of inbound shipments I know of:

- **Linked to requisition**
  - Orange house icon to show its an internal shipment
  - Linked text underneath to show the requisition ID
- **Manual inbound with internal store**
  - Orange house icon to show its an internal shipment
  - No linked text to show its manually created
- **Manual inbound with external supplier**
  - Blue truck icon to show its an external shipment
  - No linked text to show its manually created
- **Linked to purchase order**
  - Blue truck icon to show its an external shipment
  - Linked text underneath to show the PO ID

For creating shipments we'll use a header section with a New shipment combo button which also has New external shipment as an option. The external option is only available if the procurement preference is enabled.

As for permissions the shipment types you don't have access to will be hidden/filtered out. Whether to and how to show user feedback for this is up for discussion but here's some options I thought of:

1. No feedback - Could be confusing if you were expecting to see something
2. Info box to show which shipments were filtered - Visual noise for users to don't care or don't use other shipment types
3. Show in filters menu - Not too in the way but confusing if you try to change filter to something you don't have permission for

Tab spam

ria8651:
Fixes #10915

> [!WARNING]
> Claude

# What does this PR do?

Combines the internal and external inbound shipment list views into a single unified list. Previously these were split across two tabs; now they appear together in one table.

image

Key changes:
- **GraphQL `invoices` query**: The `type` parameter now accepts a list (`[InvoiceTypeInput!]`) instead of a single value, so callers can request multiple invoice types in one query. GraphQL list coercion means passing a single value still works.
- **Per-type auth**: The server validates permissions for each requested type individually rather than requiring the broad `QueryInvoice` permission (which demands all invoice type permissions). This lets users with only `InboundShipmentQuery` or only `InboundShipmentExternalQuery` still use the query.
- **Permission-based type selection**: The client builds the type list based on which permissions the user has, so only authorized types are requested.
- **Removed redundant client-side filters**: The hardcoded `type: { equalTo: INBOUND_SHIPMENT }` and `purchaseOrderId` filters in the frontend are removed since the server-side `type` parameter now handles this logic (including the InboundShipment vs InboundShipmentExternal purchase order distinction).
- **Combined list view**: Removed the internal/external tabs from `InboundListView`, replacing them with a single list. Row click navigation uses `purchaseOrder` presence to determine whether to route to the external or internal detail view. Added a `SupplierCell` component and `inboundType` field to distinguish shipment origins in the UI.

## Any notes for the reviewer?

- Go through the server changes first to understand the context for the client changes
- The `apply_type_filters` function in `invoice_queries.rs` handles the InboundShipment/InboundShipmentExternal distinction: both map to the same underlying `InvoiceType::InboundShipment`, differentiated by `purchase_order_id`. When both are requested together, no `purchase_order_id` filter is applied. When only one is requested, the appropriate null/not-null filter is set.
- `Resource::QueryInvoice` requires ALL invoice type permissions combined. The new code avoids this by only validating the specific types requested.
- This is a GraphQL non breaking change: `type` went from `InvoiceTypeInput` (single enum) to `[InvoiceTypeInput!]` (list). However, GraphQL list coercion means existing callers passing a single value will still work without changes.

# Testing

- [ ] Log in as a user with both `InboundShipmentQuery` and `InboundShipmentExternalQuery` permissions. Navigate to the inbound shipment list. Verify both internal and external shipments appear in a single combined list.
- [ ] Log in as a user with only `InboundShipmentQuery` (no external permission). Verify only internal inbound shipments appear.
- [ ] Log in as a user with only `InboundShipmentExternalQuery` (no internal permission). Verify only external inbound shipments appear.
- [ ] Click on an internal shipment row and verify it navigates to the internal detail view.
- [ ] Click on an external shipment row (one with a purchase order) and verify it navigates to the external detail view.
- [ ] Test the GraphQL `invoices` query directly: pass `type: INBOUND_SHIPMENT` (single value) and verify list coercion works.

# Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


mark-prins:
> Also, @mark-prins said he was going to update a ticket around the position of the link to the PO in the list view as an outcome of GR meeting

Yes sorry, have only just written up the minutes.
The request was to remove the PO number from where it is now and put it into a column all by itself.
We'd like it to be a clickable link still - that was a nice touch 😊 

@ria8651 will leave it to you if you want to adjust this PR or make a new issue for what is a minor change

PR #10925, opened 2026-03-19
Update to use store preferences

### LINKED ISSUE #10922: Homepage dashboard has hardcoded month values
The `Stock levels` widget displays two statistics which are use hard coded values:

* Items with less than 3 months of stock
* Items with more than 6 months of stock

Ideally these should come from the store preferences for 

* threshold for understock
* threshold for overstock

rather than using the fixed values of 3 and 6 months.
If the store does not have the preference specified, revert to these default values.

Update the text to indicate the correct number of months.

### Moneyworks Jobcode 🧰

`AFUCFOLMS`

mark-prins:
Fixes #10922

# 👩🏻‍💻 What does this PR do?
A cheeky PR which is technically out of scope for 2.17

However, this has been a long-standing quirk in our site and Adam is concerned about it - and planning to work with a country prior to the next release.

* Replace the hard-coded values for 3 and 6 months, using store preferences instead
* Renamed the hard coded property on the `ItemCounts` response from `moreThanSixMonthsStock` to `highStock` in line with the current `lowStock`
* Updated tests
* Defaulted to 3 & 6 if the store pref isn't set (though not actually possible in reality!)
* Updated the wording in the reports side panel
* Used AI translations - I thought that clumsy was better than incorrect

The defaults of 3 & 6 months for the thresholds will never realistically be used as the code is checking for nulls and OG does not allow nulls, defaulting to `0` if you don't enter a value. 

Which means that you can configure zero for upper and lower thresholds which is bonkers but allowable.

# 🧪 Testing

- [ ] Change the understock/overstock store preferences in OG
- [ ] Sync, log out and in again (required for OG prefs) and observe the labels on the homepage dashboard stock indicators
- [ ] Issue some stock, so that you have AMC for products, and check the stats for under/over stock
- [ ] Adjust the store prefs and check that the indicators change

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


client/packages/common/src/types/schema.ts
```diff
@@ -7102,10 +7102,14 @@ export type PurchaseOrderLineError = {
...
 
 export type PurchaseOrderLineFilterInput = {
+  expectedDeliveryDate?: InputMaybe<DateFilterInput>;
   id?: InputMaybe<EqualFilterStringInput>;
+  itemName?: InputMaybe<StringFilterInput>;
   purchaseOrderId?: InputMaybe<EqualFilterStringInput>;
+  purchaseOrderNumber?: InputMaybe<EqualFilterBigNumberInput>;
   receivedLessThanAdjusted?: InputMaybe<Scalars['Boolean']['input']>;
   status?: InputMaybe<EqualFilterPurchaseOrderLineStatusInput>;
+  supplierName?: InputMaybe<StringFilterInput>;
```
ria8651: I assume someone didn't run generate in their pr

PR #10932, opened 2026-03-20
10889: Fix CannotEditCostPrice error on inbound shipment line edit

### LINKED ISSUE #10889: Disable Pack Cost Price for IS(E)

Image

Following feedback from @craigdrown during GR meeting on 17 March - spec for GR is out of dtae and we have the following requirements for when the pack cost price should be enabled / disabled from changing:
- If external supplier and linked to a PO -> disabled (this is inbound shipment external tab)
- If internal suppler -> disabled (these live in inbound shipment external tab)
- If  external supplier and not linked to a PO -> enabled (these also live in inbound shipment internal tab)

Make the field editable / disabled according to the above rules 

There was debate about what happens if there was a mistake, to which Craig suggested they should make a new PO or new itnernal order
He highlights that this is important for fraud protection

### Moneyworks Jobcode 🧰

Part of DFAT work for goods receviing

esmehm:
Fixes #10889 (well resolves the bug from there)

## Summary

- Fixes a bug introduced in #10906 where editing **any** field on an inbound shipment line (e.g. packs received) linked to a purchase order or internal supplier would fail with a `CannotEditCostPrice` error
- The backend validation rejected the update whenever `cost_price_per_pack` was present in the input, even if the value was unchanged — the frontend always sends the current price back as part of the update payload
- The fix changes the validation to only reject when the price value **actually differs** from the existing line, allowing unchanged prices to pass through

## Test plan

- [ ] Open an inbound shipment (external) linked to a purchase order
- [ ] Edit a line and change packs received (without touching cost price)
- [ ] Confirm the update saves without error
- [ ] Edit a line and attempt to change the cost price — confirm it is still blocked

### DIFF
--- a/server/service/src/invoice_line/stock_in_line/update/validate.rs
+++ b/server/service/src/invoice_line/stock_in_line/update/validate.rs
@@ -128,10 +128,12 @@ pub fn validate(
     }
 
     // Cost price is read-only for internal suppliers and external suppliers linked to a PO
-    if input.cost_price_per_pack.is_some()
-        && (invoice.name_store_id.is_some() || invoice.purchase_order_id.is_some())
-    {
-        return Err(CannotEditCostPrice);
+    if let Some(new_cost_price) = input.cost_price_per_pack {
+        if (invoice.name_store_id.is_some() || invoice.purchase_order_id.is_some())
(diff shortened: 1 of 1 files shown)


ria8651 (approved):
I should have noticed that in my review 😅

PR #10938, opened 2026-03-20
8337 common function to define logic for when stock updates

### LINKED ISSUE #8337: Common function to define logic for when stock updates
A status was recently added to OMS that doesn't update stock.
However have a lot of different logic paths that try to identify if stock should be updated.
I think there should all be consolidated into one function that defines the logic.
This will make updating easier and can be tested independantly if logic changes are needed

> I think we need a common function to map when Stock Updates and when it shouldn't so there's only one place to update in future.

_Originally posted by @jmbrunskill in https://github.com/msupply-foundation/open-msupply/issues/8336#issuecomment-3021674767_

Andrei: Just want to add a couple of things to capture:

Above is both for front end and back end, one common place to define statuses and how they affect UI. Also notice that an option for status change that is always disabled and never possible appears in some places (like verify and delivered as an option in drop down in outbound). There is also an issue somewhere I think to remove 'Confirm {status}' (the confirm) from the button.

roxy-dao:
Fixes #8337

# 👩🏻‍💻 What does this PR do?
- Creates a function in backend that acts as the source of truth for all stock effect based on invoice type & status change
   - Outbounds and Supplier returns should reduce stock once status has been changed to Picked or Shipped.
   - Prescription should reduce stock once status has been changed to picked or verified
   - Prescriptions has an extra "Cancelled" status that reverts stock levels
   - Inbounds & Customer Returns should introduce new stock when status has been changed to Received or Verified.
- Consolidate all status declarations & finding next statuses into one file. 
- Remove always disabled statuses from status change buttons in all areas

# 🧪 Testing

- [ ] Make sure statuses changes are working properly in the following areas: Inbound Shipment, Outbound Shipment, Customer Return & Supplier Return
- [ ] Make sure that changing status to Received in Inbound Shipment and Customer Return introduces stock only once. Any statuses changes after that shouldn't create new stock lines
- [ ] Make sure that changing status to Picked in Outbound, Prescription and Supplier Return only reduces stock once. Any status changes after that shouldn't reduce stock furtherer
- [ ] All areas shouldn't show extra disabled statuses anymore e.g. Outbound shouldn't show any Inbound statuses in the status change button, the status mappin' on the left side should remain the same

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/invoice/stock_effect.rs
```diff
@@ -0,0 +1,293 @@
...
+    ReverseStock,
+}
+
+/// Single source of truth: returns the effect a status change has on stock.
+pub fn stock_effects(
+    invoice_type: &InvoiceType,
+    from_status: &InvoiceStatus,
+    to_status: &InvoiceStatus,
+) -> StockEffect {
+    let reduce_stock = to_status.index() >= InvoiceStatus::Picked.index()
```
andreievg: I think we ended up removing use of index(), because it was clearer to be more explicit with matches.

i.e.

match invoice_type {
    OutboundShipment | SupplierReturn => {
        match (from_status, to_status) {
            
            (Picked, Picked) => None,
            (_, Picked) => StockEffect::ReduceStock
        }
         
    }
}

aneeshthunga26 (approved):
Everything looks good to me, the only area which I am not sure of is about how the diagram that @andreievg  posted as guide shows only one type of `InboundShipment` and `CustomerReturn` but `statuses.ts` has multiple. Perhaps the diagram needs to be updated?

andreievg (changes requested):
Looks good thanks Roxy, I like it a lot. I have one change request, for backend status check to not use index, in fact to remove index from enums/variatns, see comment about using match statement.

And food for though, we could define one structure for statuses, like below, i think _NEXT can also be deduced from it

```
const statuses = [
  {
    status: InvoiceNodeStatus.New,
    datetimeField: 'createdDatetime',
    /* includes - if empty applies to all */
  },
  {
    status: InvoiceNodeStatus.Allocated,
    datetimeField: 'allocatedDatetime',
    includes: [{ type: InvoiceNodeType.OutboundShipment }],
  },
  {
    status: InvoiceNodeStatus.Picked,
    datetimeField: 'pickedDatetime',
    includes: [
      { type: InvoiceNodeType.OutboundShipment },
      { type: InvoiceNodeType.Prescription },
      { type: InvoiceNodeType.SupplierReturn },
      {
        type: InvoiceNodeType.InboundShipment,
        // Straight from ops passed to get statuses
        opts: { type: InboundShipmentType.Internal },
      },
    ],
  },
  // ...
  {
    status: InvoiceNodeStatus.Cancelled,
    datetimeField: 'cancelledDatetime',
    includes: [
      // Notice isCancelled, instead of having specific condition for this, it's better to just show it based on opt :https://github.com/msupply-foundation/open-msupply/blob/b58a182b28d16e3b218e3376b07438dab72826dc/client/packages/invoices/src/Prescriptions/DetailView/Footer/Footer.tsx#L96-L100
      { type: InvoiceNodeType.Prescription, opts: { isCancelled: true } },
    ],
  },
];
```

server/service/src/invoice/stock_effect.rs
```diff
@@ -14,60 +14,22 @@ pub fn stock_effects(
...
-        }
+    use InvoiceStatus::*;
+    use InvoiceType::*;
+
+    match (invoice_type, from_status, to_status) {
+        (Prescription, Verified, Cancelled) => StockEffect::ReverseStock,
+        (
+            OutboundShipment | Prescription | SupplierReturn,
+            New | Allocated,
+            Picked | Shipped | Delivered | Received | Verified | Cancelled,
```
andreievg: I think it's worth just specifying that only 'Picked | Shipped' is possible (just have those two), similar to the Inbound Shipment, only Received and Verified'

And for the Inbound only really 'New' | 'Shipped' | 'Delivered' for 'from_status'.

Since what you have works well, this is a nit: not required before approval. Thanks

andreievg (approved):
Looks great, thanks Roxy, one nit: up to you if you want to change or not

server/log/remote_server.1774594437
Chris-Petty, line 1: @roxy-dao did you mean to commit this server log file? Maybe for a snapshot test but I don't see it referenced anywhere

roxy-dao: Opps nope. Have removed it from develop.. Good catch!

PR #10959, opened 2026-03-23
#10946 PO: requested delivery date editable until Sent, auto-fill expected date

### LINKED ISSUE #10946: POs: minor bugs w v2.17 RC
## What went wrong? 😲

a. the requested delivery date should be editable until PO status is "Sent"
b. if the expected delivery date is unfilled, entering requested delivery should also fill expected delivery.

## Your environment 🌱
https://release-demo-open.msupply.org:8888/replenishment/inbound-shipment

CarlosNZ:
Fix #10946

## Summary
- Requested delivery date is now editable until PO status reaches "Sent" (was previously locked at "Confirmed")
- Setting requested delivery date auto-fills expected delivery date on lines where it is empty (both at header and line level)

## Test plan
- [ ] Create a PO in "New" status — verify requested delivery date is editable
- [ ] Move PO to "Confirmed" — verify requested delivery date is still editable
- [ ] Move PO to "Sent" — verify requested delivery date is now disabled
- [ ] On a PO with lines that have no expected delivery date, set the requested delivery date in the toolbar — verify expected delivery dates are auto-filled
- [ ] On a PO line with no expected delivery date, set the requested delivery date in the line edit — verify expected delivery date is also set
- [ ] Verify lines that already have an expected delivery date are NOT overwritten


esmehm:
I make a new PO without a requested del / expected del date and add an item to it, also without requested / expected del date
When I then select a requested delivery date at the PO level, it auto fills the expected del date at the PO level to be the same - fair enough
It also auto fills the expected del date on the item line (which rpeviously had no expected del date) to be the same - fair enough
But then it feels like a strange UX to me that the requested delivery date at the line item level remains blank and is not auto filled? What do you think of this flow? 

https://github.com/user-attachments/assets/a877244e-9dcb-4360-90bf-73de108af772

esmehm (approved):

Issue #10711, opened 2026-03-03
External Inbound Shipments: Product Feedback


jmbrunskill:
## Is your feature request related to a problem? Please describe 👀
Feedback on work in progress

## Describe the solution you'd like 🎁

- [x] If all items are already added to the shipment from the purchase order and the user tries to add an item a user friendly message should be displayed something like `All items from the PO have already been added to this shipment`
- [x] The financials tab table should show currency in it.
- [x] The financials tab table should show totals for  line total, line total (local)and sell price per pack (https://github.com/msupply-foundation/open-msupply/issues/10653#issuecomment-4132603630)
- [x] The financials tab table should show adjusted total (local)  - Cost adjustments need to be Weighted by Line total Value
- [x] PO Quantity should be the last column in the delivery tab
( Mark has edited 😉 )
- [x] Rename `Requested` to `Requested units` in PO 
- [x] Show the price per pack in the detail view table

### Additional context 💌
> Cost adjustments need to be Weighted by Line total Value

Apply the charges across lines in a pro-rata value based on the line total

`charge per line = total charges * (line total / total cost)`

e.g. total charges = $100 and total cost is $1000
A line that has a total of $500 would get $50 added

### Moneyworks Jobcode 🧰

OMS:DPOF

esmehm:
Assigning to you @jmbrunskill to address the remaining with #10654

rukmal-msupply:
Bugs reported #11101

Issue #10654, opened 2026-02-25
Currency tab for inbound shipment external


esmehm:
## Is your feature request related to a problem? Please describe 👀

Create the Currency tab for inbound shipment external

Currency tab
Also includes inbound shipping charges

<img width="618" height="347" alt="Image" src="https://github.com/user-attachments/assets/18609c88-a4e1-4a15-9f48-5fc132aaa9d1" />

Where
A is the sum of charges on the PO and is in the currency of the PO
B is the sum of any local charges and is in the local currency
C is the sum of all costs on the PO converted to the local currency

Read-only fields:
PO currency
Total charges: which is calculated as ( A/exchange rate )+B
Note that exchange rate is the number of foreign currency units in the base currency: e.g. if base is NZD, USD would be about 0.6. If A is USD100 and B is NZD200, total charges are 100/0.6 + 200 = 366.67
% Cost adjustment - calculated as (A+B)/C

Currency rate is filled in when the PO is chosen for this IS(E). It is a lookup of the current currency rate, but is **editable.**

PR #10982, opened 2026-03-24
10211, 10132, 10944, 10979: ISE prod feedback - card view improvements

### LINKED ISSUE #10211: Editing a IS(E)

Editing a IS(E)
Workflows
I’m able to add and remove lines on the IS(E)
I can edit existing lines if the IS(E) does not have the Finalised status
I can add new batches and duplicate existing batches

Existing functionality
Adding a line
Clicking the Add Line button give  ability to add lines from the linked PO

Shows the following columns in a table that you use to find the item you want to add: 
Item code
Item name
New functionality

Adding a line
Clicking the Add Line button give  ability to add lines from the linked PO

The following additional columns are also required to be shown
Pack size
Outstanding number of packs (this is exclusive of any quantities on the current shipment: only on already received shipments)

Editing a line
Clicking on a line in the table will open the edit screen:

On opening the window, all IS(E) lines for that item are loaded

If you have the financial editing permission, but not the inventory editing permission, then inventory fields are shown but read-only.
Comment is always editable, regardless of the IS(E) status
There is a button to duplicate a batch
The Donor and Campaign fields are auto-filled with the donor/campaign from the parent IS(E) (if set there), but can be overridden by the user
Location defaults to the item’s default receiving location, unless that conflicts with the quarantine rules (see below), in which case it is blank
VVM status only shows if item is a vaccine 
Authorisation drop-down and Note are only shown if user has Authorisation permissions and auth is turned on
The grey financial fields are only shown if you have the financial editing permission (They should not be coloured grey- this is just to identify them for this document’s descriptions)
Cost price: this is populated by the price on the PO and shouldn’t be edited. If however the cost price is changed, then the difference should be indicated to the user. Display the text in red to show that the value is different from the PO price.

Note that there isn’t a Delete button, as you can delete from the list view for the IS

Location restrictions
Add the ability to restrict the goods receiving to specific locations. 
A new boolean global preference to be added: Restrict goods receipts to specific locations
If enabled, two other preferences can be set:
Property: which is a location property, selected from a list of available properties. If selected, the property cannot be deleted and an error message is shown if the user tries to delete this property
Value: One of the possible values for the given property e.g. if this is a boolean property such as Quarantine then the values could be true or false

Note: the proposal here is to use location properties rather than location types, as the latter are used for temperatures, and this could overlap. I.e. have a quarantine coolstore location.

Permissions
The following permissions are carried across from mSupply into Open mSupply:
View goods received (mSupply) => Query Inbound Shipment (External)
Add/edit goods received (mSupply) => Mutate Inbound Shipment (External) inventory
Authorise goods received (mSupply) => Required to authorise a Inbound Shipment (External) line
Finalise goods received (mSupply) => Mutate Inbound Shipment (External) financials

The combination of these permissions results in the following matrix:

inventory
financial
Result
✔︎
✔︎
Can see and edit all info- this is the default for a small store that doesn’t want a complex process
✔︎

Only see non-financial information
Cannot update status

✔︎
Inventory fields are read-only - the user can see but not update them
Financial fields are shown

Status changes
To be Decided: For IS(I), the received status has the goods in stock but the IS is editable.
For IS(E), should we allow this same flexibility for the received status?

Changing to Verified

User must have the permission to edit financial info (see d above)

If Authorisation is enabled and there are lines that still have status Pending (i.e. haven’t been accepted or rejected yet), prevent the status from changing and show an alert modal stating “There are still X lines needing to be checked by an Authoriser before this shipment can be verified”

Otherwise, attempting to Verify show a confirmation dialog stating:

X lines will be received into stock
Y lines will be rejected and not received
NOTE: Verifying can not be undone

If the user clicks OK:
All items are brought into stock
If authorisation is enabled, only process items that have status Accepted; otherwise all lines are brought into stock.

### LINKED ISSUE #10944: FC and local currency showing the same in IS(E) edit item
## What went wrong? 😲
In inbound shipment external
If you create a PO in KMF and your instance currency is USD and you have a currency rate sert not equal to 1, then the Pack cost price and FC Cost price are showing as the same number whereas they should be different numbers

## Expected behaviour 🤔
Different cost price to FC cost price, relfecting the currency rate

## How to Reproduce 🔨

Image

Image

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

### LINKED ISSUE #10979: Inbound shipments : Cost price is not display in inbound shipments when it is created manually (without internal orders) and it is not enabled.
## What went wrong? 😲
Pack cost price not editable / dsiabled according to the expected conditions

## Expected behaviour 🤔

Pack cost price in the edit line modal on inbound shipments should be editable / disabled accoring to:

Case | Enabled?
-- | --
External supplier + purchase order | Disabled
Internal supplier + linked to outbound/IO | Disabled
External supplier, created manually | Enabled
Internal supplier, created manually | Enabled

## How to Reproduce 🔨

1. Set default item price in the master list via OG (i.e. Item A)
2. Create an Inbound shipment for that item
3. Check the cost price
#OG default price

Image
#OMS I/S
Image

## Your environment 🌱

- Open mSupply Version: 2.17.0 RC

- Platform:

- Database type:

esmehm:
Fixes #10211
Towards #10711 (adjusted totals left to do)
Fixes #10944
Fixes #10979 

# 👩🏻‍💻 What does this PR do?

A collection of improvements to the Inbound Shipment Edit (ISE) modal card view, based on product and PM feedback:

- **Card view layout**: Reorganised fields into two groups (stock line details + more info) with icons, reordered fields for better workflow
- **Narrower inputs**: Reduced minimum column widths in card grid for better use of space
- **Packs received validation**: Show error state (red border + helper text) on packs received field when value is zero, explaining why the OK button is disabled. Uses shared `isInboundPlaceholderRow` validation logic
- **Auto-focus**: Packs received field is auto-focused on initial load and when a new line is added/duplicated
- **Shipped vs received warnings**: Show shipped number of packs / pack size warnings only after blur, and only when values differ from received (used to be dedicsated fields). Only show the fields all the time for manual ISs
- **Outstanding packs**: Display outstanding packs from purchase order below packs received field
- **Card summary**: Show received packs count and batch in card heading
- **Duplicate/delete buttons**: Moved to card footer as icon buttons
- **Simplified tablet UI**: Support for `labelsAbove` layout without group icons
- **Financial tab**: Added line total footer for ISE financial tab and sell price per pack
- **Currency fixes**: Store values in local currency on IS, show PO currency on FE - further work on this in https://github.com/msupply-foundation/open-msupply/issues/10654
- **PO integration**: Toast notification if all PO lines already added to shipment
- **Pack cost price** - disabled when linked to a PO / outbound shipment and editable for manually created shipments (#10979 )

Screenshots:
image
External inbound shipment on desktop
|

If enter received vals diff to shipped vals:
image
image

After duplicating a batch, the field is in focus:
image

Error when no packs received (on manual IS)
image

Simplified tablet UI
image

Normal tablet:
image
On scanner:
image

- The `isInboundPlaceholderRow` utility is now shared between the OK button disabled logic and the card field error display
- Card grid minimum widths changed from 200px/140px to 175px/125px (portrait/landscape)
- Backend change in `add_from_purchase_order.rs` for inbound shipment line creation

# 🧪 Testing

- [ ] Open an Inbound Shipment linked to a Purchase Order
- [ ] Edit a line item — verify card view shows fields in two groups with icons
- [ ] Verify packs received is auto-focused
- [ ] Leave packs received as 0 — verify red error border and helper text appears, OK button is disabled
- [ ] Enter a value for packs received — verify error clears and OK button enables
- [ ] Duplicate a batch line — verify new card appears and packs received is focused
- [ ] Verify shipped packs/pack size warnings only appear after leaving the field, and only when different from received
- [ ] Check outstanding packs displays correctly below packs received
- [ ] Test on tablet in both portrait and landscape orientations
- [ ] Test with simplified tablet ui
- [ ] Verify financial tab shows line totals correctly

# 📃 Documentation

- [ ] **No documentation required**: UX improvements and bug fixes, no change in core behaviour

### DIFF
--- a/client/packages/common/src/intl/locales/en/common.json
+++ b/client/packages/common/src/intl/locales/en/common.json
@@ -1126,6 +1126,7 @@
   "label.global-preference": "Global preference",
   "label.gps-coordinates": "GPS coordinates",
   "label.group-by-item": "Group by Item",
+  "label.group-by-po-line": "Group by PO Line",
   "label.growth-on-previous-year": "Growth on previous year",
   "label.gtin": "GTIN",
   "label.have-not-shipped": "Have not been shipped",
--- a/client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
+++ b/client/packages/common/src/ui/components/buttons/standard/DialogButton.tsx
@@ -201,11 +201,6 @@ export const DialogButton: React.FC<DialogButtonProps> = ({
       label={customLabel ?? t(labelKey)}
       tabIndex={variant === 'cancel' ? 1 : 0}
       type={type}
-      onKeyDown={e => {
-        if (e.key === 'Enter') {
-          onClick(e);
-        }
-      }}
       sx={
         disabled
           ? {
--- a/client/packages/common/src/ui/components/buttons/standard/ShrinkableBaseButton.tsx
+++ b/client/packages/common/src/ui/components/buttons/standard/ShrinkableBaseButton.tsx
@@ -73,9 +73,6 @@ export const ShrinkableBaseButton = React.forwardRef<
             isRtl={isRtl}
             aria-label={label}
             onClick={onClick}
-            onKeyDown={(event: React.KeyboardEvent<HTMLButtonElement>) => {
-              if (event.code === 'Enter' && !!onClick) onClick({} as any);
-            }}
             startIcon={shrink ? null : startIcon}
             {...props}
           >
--- a/client/packages/common/src/ui/components/panels/SlidePanel/SlidePanel.tsx
+++ b/client/packages/common/src/ui/components/panels/SlidePanel/SlidePanel.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useCallback } from 'react';
 import {
   Box,
   Typography,
@@ -7,6 +7,7 @@ import {
   styled,
   ClickAwayListener,
 } from '@mui/material';
+import { FocusTrap } from '@mui/base';
 
 const Backdrop = styled(Box)({
   position: 'fixed',
@@ -39,6 +40,16 @@ export const SlidePanel = ({
   width = '100%',
   preventClickAway = true,
 }: SlidePanelProps) => {
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent) => {
+      if (e.key === 'Escape') {
+        e.stopPropagation();
+        onClose();
+      }
+    },
+    [onClose]
+  );
+
   const panel = (
     <Box
       sx={{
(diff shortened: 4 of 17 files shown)


client/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEditCards.tsx
```diff
@@ -121,166 +124,157 @@ export const InboundLineEditCards = ({
...
+                  {`${t('label.shipped-number-of-packs')}: ${shippedPacks}`}
+                </Typography>
+              )}
+              {!!purchaseOrderId && poOutstandingPacks != null && (
+                <Typography
+                  variant="caption"
+                  color="text.secondary"
+                  sx={{ mt: 0.5, display: 'block' }}
+                >
+                  {`${t('label.outstanding-packs')}: ${poOutstandingPacks}`}
```
CarlosNZ: Image

Do you think this label should just be "Outstanding packs: <num>"  so it fits on one row? Would also be more consistent with other labels such as "Outstanding units":

Image

CarlosNZ (approved):
Cool, thanks for the changes @esmehm

PR #10990, opened 2026-03-25
10931 add imprest stock history requisition types

### LINKED ISSUE #10931: Accept requisition types "imprest" and "customer stock history" in sync

OG has 2 requisition types that actually predate the ol' "internal order" and "customer requisition" flow by many years. These are:

- [Customer stock histories](https://docs.msupply.org.nz/issuing_goods:stock_control_methods#customer_stock_history_method): used for small customers not using mSupply, e.g. a health centre solely supplied by your store. This overlaps somewhat with currently supported manual customer requisitions and R&R forms.
- [Imprest](https://docs.msupply.org.nz/issuing_goods:stock_control_methods#imprest_method): Super simple method often for topping up a "customer" like a ward cupboard to a desired stock level once a day.

These types come through sync as "sh" and "im" respectively and OMS sync currently rejects them as invalid requisition types, along with all their lines failing due to FK constraint from the parent not integrating. This can be millions of lines for some stores! The OG feature wasn't very good at pruning...

All these errors end up slowing integration down a whole lot. It'd also be nice in the near future to a: include them in dashboard based on OMS and B: actually support these supply methods 😃 

For now we just want to stop all the FK constraint errors! Provided we want to support these one day, may as well just integrate them. Add the types to the `requisition.type` enum; `IMPREST`, `STOCK_HISTORY`.

### Additional context 💌

Note, as requisitions in general are treated as transfer records, the "Other Party" receives these. So if you have an OMS store and an OG store makes an imprest for your store (for what reason? Probably shouldn't be allowed TBH) then that will sync to your OMS site/store. Perhaps we should prevent this, as it's not particularly useful for anything afaik, due to no transfer being created. (similarly, if you manually create a supplier invoice to another store, that store doesn't create a corresponding outbound shipment)

### Acceptance Criteria ✅

- [ ] OG store makes a customer stock history and a imprest to an OMS store
- [ ] Both requisitions and their lines sync and are safely integrated into OMS (check the DB, there is no UI for this)
- [ ] Move the OG store to the OMS remote site
- [ ] Reinitialise the site
- [ ] Both requisitions and their lines sync and are safely integrated into OMS
 
### Moneyworks Jobcode 🧰

OMS:DFGEN

Chris-Petty:
Fixes #10931 

# 👩🏻‍💻 What does this PR do?

AI'd and tidied up

- Adds imprest and stock history requisition types, treating them more or less the same as response requisitions as far as the translators are concerned (bar their type ofc).
- There is a test for "wp" status requisitions, they currently fail and should continue to do so until we figure out what to do with them. Fortunately they are extremely rare 😉.
- The migrations are targeted at 2.17 - we may want to sneak this in to 2.17 to smooth out the many integration errors that can occur. 

# 🧪 Testing

- [ ] Initialise OMS central with a store that has some imprest and customer stock history requisitions 
- [ ] All the records integrate with no issue (unless miraculously of status "wp")
- [ ] They don't somehow show up in customer requisitions or internal orders
- [ ] Bonus: gql you can query for them if you so wanted...

# 📃 Documentation

- [x] **No documentation required**: We aren't adding UI for these, it's just accepting them under the hood.
  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


andreievg (approved):
Looks great thanks ! 

I think it's worth putting into 2.17, thanks for considering that.

I think re-integrating existing stores that migrated and have these in error state in sync_buffer could be possible, we would need to clear sync_buffer for requisition and requisition line that had error in sync_buffer and that don't have associated requisition/line <- this can be issue for later

PR #11002, opened 2026-03-25
RnD graphql sync

### LINKED ISSUE #10934: Setup graphql subscriptions
## The suggestion

Currently in the codebase the frontend uses polling for several APIs like
 - DB migration status
 - Sync
 - Initialisation status
 - Temp sensor notifications

I would like to see these APIs replaced with graphQL subscriptions (might use websockets, server sent events or long polling depending on implementation). This would reduce unnecessary requests from the client to the server and also show information as its made available.

Need to investigate compatibility with caddy

Medium

### Agreed Solution
TBD

aneeshthunga26:
Fixes #10934

# 👩🏻‍💻 What does this PR do?

- Sets up graphql subscriptions.
- Optimised to reduce db queries
- sync status & info are migrated to use subscriptions
- push queue is migrated to use subscriptions
- initialisation status is migrated to use subscriptions
- fallback mechanisms in place when websocket functionality is unavailable (in edge case deployment scenarios)

Tested with Caddy

# 🧪 Testing

- [ ] Initialise omsupply site
- [ ] View the initialisation screen, Verify initialisation steps are updated correctly
- [ ] Add a stock line from Inventory -> Stock -> New Stock
- [ ] Verify, next to Sync in the sidebar, badge is updated is updated immediately
- [ ] Make changes (either by adding inbound shipments, new stock, stocktakes, stock adjustment) within 30 seconds of each other
- [ ] Verify badge updates are delayed by 30 seconds or less
- [ ] Click on Sync to open sync modal and click on sync now
- [ ] Verify sync steps update correctly
- [ ] Verify badge next to sync in the sidebar disappears
 
 #### Checking if websockets are being used
 
- [ ] Sync status updates are received via subscription in the client
  - [ ] Open dev tools (right click -> inspect)
  - [ ] Navigate to network tab
  - [ ] Click on the last ws entry
  - [ ] Start sync manually
  - [ ] See messages come in on the right pane
- [ ] Check websocket activity as described above during site initialisation

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


jmbrunskill (dismissed):
Haven't reviewed, just capturing we definitely need these steps before merging.

Testing implementation with Caddy
Caddy config to allow websockets
+ Documentation shared with support team of any changes required

server/graphql/general/src/queries/sync_status.rs
```diff
@@ -62,19 +62,93 @@ impl SyncStatusWithProgressNode {
...
-    integration: Option<SyncStatusWithProgressNode>,
-    pull_central: Option<SyncStatusWithProgressNode>,
-    pull_v6: Option<SyncStatusWithProgressNode>,
-    pull_remote: Option<SyncStatusWithProgressNode>,
-    push: Option<SyncStatusWithProgressNode>,
-    push_v6: Option<SyncStatusWithProgressNode>,
-    last_successful_sync: Option<SyncStatusNode>,
-    warning_threshold: i64,
-    error_threshold: i64,
+    pub is_syncing: bool,
```
andreievg: Can you comment please about changes in this file, why are they needed

client/packages/common/src/api/GqlContext.tsx
```diff
@@ -90,13 +90,15 @@ class GQLClient extends GraphQLClient {
   private emptyData: object;
   private skipRequest: SkipRequest;
   private lastRequestTime: Date;
+  private _url: string;
```
aneeshthunga26: adding url as a field so `SubscriptionClient` can access it

server/repository/src/db_diesel/storage_connection.rs
```diff
@@ -67,6 +70,27 @@ impl StorageConnection {
...
+    pub fn notify_changelog_insert(&self) {
+        if self.on_changelog_insert.is_some() {
+            self.has_pending_changelog_notification
+                .store(true, Ordering::Relaxed);
+        }
+    }
+
+    /// Fire pending changelog notification if any. Called after transaction commit.
+    fn flush_changelog_notification(&self) {
+        if self
```
aneeshthunga26: if there was an insert to the changelog, fire notification.
This is called after the outermost transaction commits

server/graphql/general/src/subscriptions/sync_status.rs
```diff
@@ -0,0 +1,51 @@
...
+    sync::sync_status::status::FullSyncStatus,
+};
+use tokio::sync::broadcast;
+
+use crate::queries::sync_status::FullSyncStatusNode;
+
+pub fn sync_status_stream(
+    service_provider: Data<ServiceProvider>,
+) -> impl Stream<Item = Option<FullSyncStatusNode>> {
+    let rx = service_provider.sync_status_broadcast.subscribe();
```
aneeshthunga26: receive `SyncLogRow` via tokio broadcast receiver

server/graphql/general/src/subscriptions/initialisation_status.rs
```diff
@@ -0,0 +1,56 @@
...
+            // for PreInitialisation/Initialising. Only Initialised needs site_name.
+            let status = if row.finished_datetime.is_some() {
+                let site_name = match &cached_site_name {
+                    Some(name) => name.clone(),
+                    None => {
+                        // Query site name once and cache it
+                        let name = sp
+                            .basic_context()
+                            .ok()
+                            .and_then(|ctx| sp.settings.sync_settings(&ctx).ok())
```
andreievg: You can directly get this from key_value_store, which is cached anyways

server/graphql/general/src/subscriptions/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+};
+
+use crate::queries::initialisation_status::InitialisationStatusNode;
+use crate::queries::sync_status::FullSyncStatusNode;
+
+use initialisation_status::initialisation_status_stream;
+use push_queue_count::push_queue_count_stream;
+use sync_status::sync_status_stream;
+
+fn get_service_provider(ctx: &Context<'_>) -> Result<Data<ServiceProvider>> {
```
andreievg: this exist on context

andreievg (commented):
Thanks Aneesh, very cool.

I would like to see an alternative that aligns more with the layer boundaries where graphql is only a mapping layer. 

The idea is that whenever an event occurs that would cause a subscription update, a service method is called, which would do the required queries/calculations and then call `rx` on the subscription. That method can be called within it's own thread, you can have general 'subscription resolver' worker for this. 

The main reason I would like a bit more effort on this and would want to see an alternative that would call existing methods, and potentially remove need for any additional caching.

andreievg:
Basically when event occurs, it's processed by one shared worker and then sends a message with resolved service layer type to subscriber. 

Worker would be similar to processors

```rust
loop {
    match rx.recv().await {
       SubscriptionTrigger::InitStatus => // do query for init status and send notification to actual subscriber via another enum ResolvedSubscription::InitStatus
    // etc
    }

}
```

And the in subscriber

```rust
        |(mut rx)| async move {
            let Ok(ResolvedSubscription::InitStatus(status)) =rx.recv().await else {
               return None;
            };
            Some(InitialisationStatusNode::from_domain(status))
```

Btw for sync status sending through SyncStatusRow would be great, and we wouldn't need caching in SyncStatusRow repository. 

Also the processor can easily do debounce (sorry if not accurate, but should be pretty readable:

```rust

let mut last_log_size_time = now();
let mut is_last_log_queued = false;


loop {
    match rx.recv().await {
       SubscriptionTrigger::LogSize => {
          if last_log_size_time > now() + 30seconds {
             if !is_last_log_queued {
                is_last_log_queued =  true
               tokio::task::spawn(  a task at  last_log_size_time + 30seconds to call SubscriptionTrigger::LogSize   )
             }
          } else {
             is_last_log_queued = false;
            // do query and resolve
          
          }
       }
    // etc
    }

}


```

server/graphql/general/src/subscriptions/initialisation_status.rs
```diff
@@ -0,0 +1,25 @@
...
+use crate::queries::initialisation_status::InitialisationStatusNode;
+
+pub fn initialisation_status_stream(
+    broadcast: Data<broadcast::Sender<ResolvedSubscription>>,
+) -> impl Stream<Item = InitialisationStatusNode> {
+    let rx = broadcast.subscribe();
+
+    stream::unfold(rx, |mut rx| async move {
+        loop {
+            match rx.recv().await {
```
aneeshthunga26, line 15: I prefer to use match here as there is a difference between None and continue. `None` signals stream to close and `continue` just skips that message and waits to process the next one

server/service/src/subscription.rs
```diff
@@ -0,0 +1,200 @@
...
+                    last_successful = Some(status.clone());
+                }
+
+                let _ = tx.send(ResolvedSubscription::SyncStatus {
+                    status: status.clone(),
+                    last_successful: last_successful.clone(),
+                });
+
+                // Derive initialisation status from the same row.
+                if !initialised {
```
andreievg, line 138: Please re-query and re-use logic from: https://github.com/msupply-foundation/open-msupply/blob/e3363e523186a45f2b5f097ed384517d9dc8491a/server/graphql/general/src/queries/initialisation_status.rs#L22

server/service/src/sync/sync_status/status.rs
```diff
@@ -289,9 +301,16 @@ fn get_latest_successful_sync_status(
...
+    *LATEST_SUCCESSFUL_SYNC.write().unwrap() = Some(result.clone());
 
-    Ok(result)
+    Ok(Some(result))
+}
+
+/// Update the cached last successful sync status.
+/// Called by the subscription stream when it detects a successful sync completion.
+pub fn cache_successful_sync_status(status: FullSyncStatus) {
+    *LATEST_SUCCESSFUL_SYNC.write().unwrap() = Some(status);
```
andreievg: Can you remove this cache please

server/service/src/subscription.rs
```diff
@@ -0,0 +1,200 @@
...
+                }
+            }
+            SubscriptionTrigger::PushQueueChanged => {
+                if last_push_queue_query.elapsed() >= PUSH_QUEUE_DEBOUNCE {
+                    resolve_push_queue_count(
+                        &service_provider,
+                        &tx,
+                        &mut last_push_queue_query,
+                    );
+                } else {
```
andreievg: I think the bulk of de-bounce logic can go here, we also don't want to de-bounce for other queries

server/service/src/subscription.rs
```diff
@@ -0,0 +1,200 @@
...
+    service_provider: &ServiceProvider,
+    tx: &broadcast::Sender<ResolvedSubscription>,
+    last_query: &mut Instant,
+) {
+    let count = match service_provider.basic_context() {
+        Ok(ctx) => service_provider
+            .sync_status_service
+            .number_of_records_in_push_queue(&ctx)
+            .unwrap_or(0),
+        Err(_) => return,
```
andreievg: At least log here please

server/server/src/lib.rs
```diff
@@ -186,6 +196,9 @@ pub async fn start_server(
     let validated_plugins = ValidatedPluginBucket::new(&settings.server.base_dir).unwrap();
     let validated_plugins = Data::new(Mutex::new(validated_plugins));
 
+    let (_subscription_task, subscription_broadcast) =
```
andreievg: use join handle in tokio select, similar to other tasks defined here

server/repository/src/db_diesel/storage_connection.rs
```diff
@@ -59,6 +60,8 @@ impl<'a> LockedConnection<'a> {
 
 pub struct StorageConnection {
     raw_connection: Mutex<DBConnection>,
+    on_changelog_insert: Option<Arc<dyn Fn() + Send + Sync>>,
```
andreievg: can we just pass through tx here ? Rather then callback

server/service/src/subscription.rs
```diff
@@ -0,0 +1,200 @@
...
+const PUSH_QUEUE_DEBOUNCE: Duration = Duration::from_secs(30);
+
+// ── Triggers (inbound to worker) ──
+
+#[derive(Clone, Debug)]
+pub enum SubscriptionTrigger {
+    /// A sync log row was updated (step start/done, progress, error, completion)
+    SyncStatus(SyncLogRow),
+    /// Changelogs were inserted (mutations created/modified data)
+    PushQueueChanged,
```
andreievg, line 22: Can add Option<u64> here and call it with 0 in synchroniser.rs after push has finished. Don't need this if badge updates after syncing, check timing = badge updated, quick sync (less then 30 seconds)

server/graphql/lib.rs
```diff
@@ -418,10 +441,57 @@ pub fn attach_graphql_schema(
...
+/// mirroring how the HTTP handler routes requests.
+async fn graphql_ws(
+    schema: Data<GraphqlSchema>,
+    req: HttpRequest,
+    payload: web::Payload,
+) -> Result<HttpResponse, actix_web::Error> {
+    let on_connection_init = |value: serde_json::Value| async move {
+        let mut data = async_graphql::Data::default();
+        // Client sends { "Authorization": "Bearer <token>" } in connectionParams
+        if let Some(token) = value.get("Authorization").and_then(|v| v.as_str()) {
```
andreievg, line 464: We should already have a method to handle auth

client/packages/system/src/Sync/api/hooks/utils/useSyncInfo.ts
```diff
@@ -1,24 +1,76 @@
...
+      const existing = queryClient.getQueryData<SyncInfoQuery>(queryKey);
+      return {
+        __typename: 'Queries' as const,
+        numberOfRecordsInPushQueue: data.pushQueueCountUpdated,
+        syncStatus: existing?.syncStatus ?? null,
+      };
+    },
+  });
+
+  const isSubscribed = isSyncStatusSubscribed || isPushQueueSubscribed;
```
andreievg: && would be better

client/packages/system/src/Sync/api/hooks/utils/useSyncInfo.ts
```diff
@@ -1,24 +1,76 @@
...
-  const { token } = getAuthCookie();
+  const { token } = useAuthContext();
+  const queryClient = useQueryClient();
+
+  const isEnabled = !!token && enabled;
+  const queryKey = api.keys.syncInfo();
+
+  // Subscribe to real-time sync status updates via WebSocket.
+  // Merges into existing cache data so we don't clobber numberOfRecordsInPushQueue.
+  const { isSubscribed: isSyncStatusSubscribed } = useSubscription<
```
andreievg: I think better to have these two subscriptions in 1, for better consumability in front end

client/packages/system/src/Sync/api/hooks/utils/useSyncInfo.ts
```diff
@@ -1,24 +1,76 @@
...
+  });
+
+  const isSubscribed = isSyncStatusSubscribed || isPushQueueSubscribed;
 
-  // manually adding the token and setting the authorization header
-  // there were instances where the token was not included in the request
-  // even though the auth cookie existed with a valid token
-  // the query is only enabled if there's a token -
-  // no need to check the sync status if there's no token
   const { data, ...rest } = useQuery(
```
andreievg: How de we prevent a re-fetch after cache key is updated ? In our conversation you mentioned setQueryData inside useSubscription would update data (and not invalidate query). 

I would like to see an alternative to cache control to get data out of subscription and reduced into front end for reactive update (that's a mouth full, but I mean stateful update of data being subscribed to)

client/packages/common/src/api/hooks/useSubscription.ts
```diff
@@ -0,0 +1,124 @@
...
+        },
+        complete: () => {
+          if (!disposed) setIsSubscribed(false);
+        },
+      }
+    );
+
+    setIsSubscribed(true);
+
+    return () => {
```
andreievg, line 105: If we are not using this, we shouldn't have it in code (i think it's needed, when doing manual sync)

client/packages/common/src/authentication/api/hooks/usePreferences.ts
```diff
@@ -25,6 +29,13 @@ export const usePreferences = (): Partial<PreferencesNode> => {
     staleTime: Infinity,
     suspense: true,
     enabled: !!storeId,
+    retry: 3,
```
andreievg: If this is replicatable, can you try clearing auth in initialisation, it's likely due to storeId being present in user context (in develop, when re-initialising this could happen).

server/graphql/lib.rs
aneeshthunga26: Existing method uses http headers to extract the auth token. Websocket on_connection_init callback receives a json payload with the auth token. I thought of a few ways to refactor the existing method but I think that only makes it more complex. This is simpler

client/packages/common/src/api/hooks/useInitialisationStatus.ts
```diff
@@ -8,16 +13,30 @@ export const useInitialisationStatus = (
   const { client } = useGql();
   const sdk = getSdk(client);
 
-  return useQuery(
-    'initialisationStatus',
+  const queryKey = 'initialisationStatus';
+
+  const { isSubscribed, data: subData } = useSubscription<
```
andreievg: With the type inference I was thinking something like this: https://github.com/msupply-foundation/open-msupply/compare/rnd-graphql-sync...rnd-graphql-sync(Infer-types-on-useSubscription)

client/packages/common/src/api/hooks/useInitialisationStatus.ts
```diff
@@ -8,16 +13,30 @@ export const useInitialisationStatus = (
...
+    InitialisationStatusUpdatedSubscription,
+    InitialisationStatusUpdatedSubscription['initialisationStatusUpdated']
+  >({
+    document: InitialisationStatusUpdatedDocument,
+    enabled: true,
+    requireAuth: false,
+    select: data => data.initialisationStatusUpdated,
+  });
+
+  const { data: queryData, ...rest } = useQuery(
```
andreievg, line 27: Comment here to say we are using fallback

client/packages/common/src/api/SubscriptionClient.ts
```diff
@@ -0,0 +1,59 @@
+import { createClient, Client } from 'graphql-ws';
+import { getAuthCookie } from '../authentication/AuthContext';
+
+let subscriptionClient: Client | null = null;
+let currentUrl: string | null = null;
+
+/**
+ * Get or create a shared graphql-ws subscription client.
```
andreievg, line 8: how is it shared ? The function is shared but we seem to create it every time vs GqlClient stored in context. 

Can you make a carry over issue to extend our GqlClient with this, I think it would also mean that we don't need to expose url from GqlClient

server/service/src/subscription.rs
```diff
@@ -0,0 +1,192 @@
...
+
+                if status.summary.finished.is_some() && status.error.is_none() {
+                    last_successful = Some(status.clone());
+                }
+                last_status = Some(status.clone());
+
+                let _ = tx.send(ResolvedSubscription::SyncInfo {
+                    status,
+                    last_successful: last_successful.clone(),
+                    push_queue_count: (row.push_progress_total.unwrap_or(0)
```
andreievg: A carry over issue for badge to update when sync happens

andreievg: I think we can just fire off:

```
   let _ = tx.send(ResolvedSubscription::SyncInfo {
                            status: status.clone(),
                            last_successful: last_successful.clone(),
                            push_queue_count: count,
                        });
```

andreievg (approved):
Looks great, thanks Aneesh, a couple of carry over things to be captured, what happens now when doing manual sync, does the badge update ?

server/service/src/subscription.rs
aneeshthunga26: @andreievg Badge does update when doing manual sync. Did it not update in your local testing?

client/packages/common/src/api/hooks/useInitialisationStatus.ts
aneeshthunga26: I merged your branch, thank you. This is nice and simple

andreievg:
@jmbrunskill can you approve please, Aneesh has fallback for polling if 'ws' is blocked. 

@aneeshthunga26 can you make sure to merge with a squash to easily cherry pick into 2.16

Issue #11008, opened 2026-03-26
Inbound shipment Ext : Text input field is not responsive, noticeable delay when typing quickly.


rukmal-msupply:
## What went wrong? 😲

The text field does not respond instantly to user input.
When typing at a normal or fast speed, there is a delay (few milliseconds to noticeable lag) before characters appear in the field.
i..e Type fast in customer reference filed...

## Expected behaviour 🤔

Text input should be real-time and responsive
Cursor should move instantly as user types
No visible delay between key press and character rendering

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Type fast in inbound shipment text fields.
https://drive.google.com/file/d/12hia7MFxWBVcd_esvVbkcUHqrNUR47fH/view?usp=sharing
## Your environment 🌱

- Open mSupply Version: 2.17.0
- Legacy mSupply Central Server Version:

- Platform: Web/Desktop
- 

- Database type:

jmbrunskill:
@esmehm can you check if you fixed thisin your PR?

PR #11010, opened 2026-03-26
#10974 Fix PO details not editable when sent & debounced save losing fields

### LINKED ISSUE #10974: Purchase order : Permission issues when status is changed.
## What went wrong? 😲

1. PO should not allow to edit/update the PO details if the PO status >= Sent
2. and not allow to add documents or delete existing documents if the PO status >= Sent

## Expected behaviour 🤔

 Spec:

Image

## How to Reproduce 🔨

1. Create a PO and mark as Sent
2. Log in as a normal user who doesn't have authorised permissions
3. Edit PO details
4. Edit PO document details

## Your environment 🌱

- Open mSupply Version: 2.17.0 RC

- Platform:

- Database type:

esmehm:
Fixes #10974 and #11008

# 👩🏻‍💻 What does this PR do?

- Prevents editing of PO details fields and document upload/delete when a PO is in Sent status
- Fixes a bug where changing two detail fields in quick succession would cause the first field to blank out, because the debounced save was replacing (not accumulating) pending changes
- Also fixes an otherwise unidentified bug that supplier discount percentage was not saving if total cost was 0, giving strange behaviour where it went back to zero (now remembers the discount even if you have no cost, yet)
image
- Also allows authorised users to edit the adjusted num of packs for sent POs (and unauthorised cannot)

The debounced save previously passed each partial change independently to the server. Since the debounce replaces the pending callback, only the last field change within the debounce window would be sent. Now pending changes are accumulated in a ref and flushed together.

Note: the `invalidateQueries` in `useUpdate`'s `onSuccess` still causes a brief flash on save — this is a pre-existing issue not addressed here.

# 🧪 Testing

- [ ] Open a Purchase Order in Sent status
- [ ] Verify that detail fields are not editable and document upload/delete is disabled
- [ ] Open a Purchase Order in New/Confirmed status
- [ ] Change two detail fields in quick succession (within 1 second)
- [ ] Verify the first field retains its value after the second field is changed
- [ ] Refresh the page and verify both fields are persisted

# 📃 Documentation

- [ ] **No documentation required**: bug fix which isn't a change in behaviour

### DIFF
--- a/server/service/src/purchase_order/update/mod.rs
+++ b/server/service/src/purchase_order/update/mod.rs
@@ -65,32 +65,64 @@ pub struct UpdatePurchaseOrderInput {
 }
 
 impl UpdatePurchaseOrderInput {
-    pub fn has_non_status_field_changes(&self) -> bool {
-        self.supplier_id.is_some()
-            || self.confirmed_datetime.is_some()
-            || self.supplier_discount_percentage.is_some()
-            || self.supplier_discount_amount.is_some()
-            || self.donor_id.is_some()
-            || self.reference.is_some()
-            || self.currency_id.is_some()
-            || self.foreign_exchange_rate.is_some()
-            || self.shipping_method.is_some()
-            || self.sent_datetime.is_some()
-            || self.contract_signed_date.is_some()
-            || self.advance_paid_date.is_some()
-            || self.received_at_port_date.is_some()
-            || self.requested_delivery_date.is_some()
-            || self.supplier_agent.is_some()
-            || self.authorising_officer_1.is_some()
-            || self.authorising_officer_2.is_some()
-            || self.additional_instructions.is_some()
-            || self.heading_message.is_some()
-            || self.agent_commission.is_some()
-            || self.document_charge.is_some()
-            || self.communications_charge.is_some()
-            || self.insurance_charge.is_some()
-            || self.freight_charge.is_some()
-            || self.freight_conditions.is_some()
+    pub fn is_status_only_change(&self) -> bool {
+        let Self {
+            id: _,
+            status: _,
+            // Comment is editable on sent POs, only disabled when finalised
+            comment: _,
+            // All other fields must be None
+            supplier_id,
+            confirmed_datetime,
+            supplier_discount_percentage,
+            supplier_discount_amount,
+            donor_id,
+            reference,
+            currency_id,
+            foreign_exchange_rate,
+            shipping_method,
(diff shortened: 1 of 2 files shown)


server/service/src/purchase_order/update/mod.rs
```diff
@@ -63,6 +64,37 @@ pub struct UpdatePurchaseOrderInput {
...
+            || self.heading_message.is_some()
+            || self.agent_commission.is_some()
+            || self.document_charge.is_some()
+            || self.communications_charge.is_some()
+            || self.insurance_charge.is_some()
+            || self.freight_charge.is_some()
+            || self.freight_conditions.is_some()
+    }
+}
+
```
ria8651: Forces devs to update function when adding new field.

```suggestion
impl UpdatePurchaseOrderInput {
    pub fn is_status_only_change(&self) -> bool {
        let Self {
            id: _,
            status: _,
            // All other fields must be None
            supplier_id,
            confirmed_datetime,
            comment,
            supplier_discount_percentage,
            supplier_discount_amount,
            donor_id,
            reference,
            currency_id,
            foreign_exchange_rate,
            shipping_method,
            sent_datetime,
            contract_signed_date,
            advance_paid_date,
            received_at_port_date,
            requested_delivery_date,
            supplier_agent,
            authorising_officer_1,
            authorising_officer_2,
            additional_instructions,
            heading_message,
            agent_commission,
            document_charge,
            communications_charge,
            insurance_charge,
            freight_charge,
            freight_conditions,
        } = self;

        supplier_id.is_none()
            && confirmed_datetime.is_none()
            && comment.is_none()
            && supplier_discount_percentage.is_none()
            && supplier_discount_amount.is_none()
            && donor_id.is_none()
            && reference.is_none()
            && currency_id.is_none()
            && foreign_exchange_rate.is_none()
            && shipping_method.is_none()
            && sent_datetime.is_none()
            && contract_signed_date.is_none()
            && advance_paid_date.is_none()
            && received_at_port_date.is_none()
            && requested_delivery_date.is_none()
            && supplier_agent.is_none()
            && authorising_officer_1.is_none()
            && authorising_officer_2.is_none()
            && additional_instructions.is_none()
            && heading_message.is_none()
            && agent_commission.is_none()
            && document_charge.is_none()
            && communications_charge.is_none()
            && insurance_charge.is_none()
            && freight_charge.is_none()
            && freight_conditions.is_none()
    }
}

```

Issue #10513, opened 2026-02-13
Uniformising invoice number description among its columns in inbound/outbound shipments and customer/supplier returns


alainsussol:
## What went wrong? 😲

The Invoice number column description in inbound shipment is number
<img width="249" height="108" alt="Image" src="https://github.com/user-attachments/assets/a53ea2eb-582a-4874-8985-af3be56c31a3" />

but for Outbound shipments, it is Invoice Number/Shipment number
<img width="333" height="144" alt="Image" src="https://github.com/user-attachments/assets/cee19075-67e8-4bae-989e-807bf8c7ae50" />

## Expected behaviour 🤔
Should probably be the same everywhere for transactions 

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Go to outbound shipment, inbound shipment, customer and supplier return 
2. Cclick on the invoice number column
3. compare the description
4. Inbound shipment column description is different from the 3 others

## Your environment 🌱

- Open mSupply Version: 2.16.00-RC-02131657
- Legacy mSupply Central Server Version: v8.08

- Platform: Android , omSupply desktop

- Database type: SQLite

jmbrunskill:
THis is low severity to me, but would be nice to fix up. Next cooldown? AI?

PR #11020, opened 2026-03-26
11000 Table headers do not wrap lines

### LINKED ISSUE #11000: Table headers do not wrap lines
## What went wrong? 😲

The previous implementation of tables wrapped the headers over multiple lines if necessary.

We now have long heading text doing this:

Image

## Expected behaviour 🤔
Header text allowed to wrap over multiple lines. Ideally restrict to two lines - in the previous tables we had 3+ lines of text and this takes up space and is not easily readable.

I don't think a two line limit is technically possible and if that's the case then an editorial approach is fine - we change the header text itself.

## How to Reproduce 🔨

1. Go to most tables ( the example above is internal orders ) and view the headers
2. If needed resize the columns so that the text no longer fits

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

roxy-dao:
Fixes #11000 & #10513 & #11020

# 👩🏻‍💻 What does this PR do?
- Remove nowrap so text warps in table headers
Screenshot 2026-03-26 at 12 30 14

- Use same header & description for invoice tables

# 🧪 Testing

- [ ] Resize columns to test the effects. If column size becomes too small should still show `...`

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
+++ b/client/packages/common/src/ui/layout/tables/useTableDisplayOptions.tsx
@@ -204,6 +204,9 @@ export const useTableDisplayOptions = <T extends MRT_RowData>({
         '& .MuiTableSortLabel-root': {
           display: column.getIsSorted() ? undefined : 'none',
         },
+        '& .Mui-TableHeadCell-Content-Wrapper': {
+          whiteSpace: 'normal',
+        },
         // replace the action button with a full-width invisible button to make
         // the whole header clickable to display the action menu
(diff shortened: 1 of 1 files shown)


roxy-dao:
This is the best I can do 👀 

Screenshot 2026-03-27 at 08 48 39

PR #11034, opened 2026-03-27
Dedup before push on changelog instead of dedup on the fly

jmbrunskill:
Fixes issues with performance on the changelog deduplication when there's a lot of records in changelog

# 👩🏻‍💻 What does this PR do?

Switched from changelog dedup to a just in time de-duplication that happens just before sync kicks off (pushing to sync v6 or sync v5. This is required because otherwise if a record is added and then deleted before the sync happens, you'll get an error.

I'm deduping in both the sync_v6 and sync v5 push as if you are a central server you won't run sync v6 push, but we still need to dedup there too. Can possibly get away without it for this PR as it's only for 1 server for now.
Final PR might need more clean up...

# 🧪 Testing

- [ ] Create an invoice and invoice lines
- [ ] Delete the invoice
- [ ] Try to sync, make sure it works
- [ ] test performance?

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend

### DIFF
--- a/server/repository/src/db_diesel/changelog/changelog.rs
+++ b/server/repository/src/db_diesel/changelog/changelog.rs
@@ -29,22 +29,22 @@ table! {
     }
 }
 
-table! {
-    changelog_deduped (cursor) {
-        cursor -> BigInt,
-        table_name -> crate::db_diesel::changelog::ChangelogTableNameMapping,
-        record_id -> Text,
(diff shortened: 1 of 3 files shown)


Chris-Petty:
Merging with intent to fix up tests later

PR #11035, opened 2026-03-27
Add Docker support for Postgres

### LINKED ISSUE #10999: Export postgres dump
We currently export sqlite data files but we don't do the same for postgres dumps. This isn't an issue on mobile since it uses sqlite but web/desktop can use postgres

### Acceptance Criteria ✅
Allow support to export postgres dump

CarlosNZ:
Fixes #10999, but also updates Docker functionality

Best way to review is probably just go through the instructions in the README and check they match reality and actually work as described in all cases.

## Summary
- Add Postgres database export via `pg_dump` in the support endpoint
- Add Docker targets (`postgres`, `postgres-dev`) that run an embedded PostgreSQL server inside the container
- Support importing a `pg_dump` database dump on container launch via volume mount
- Update CI workflow to build and push all four image variants (sqlite, dev, postgres, postgres-dev) in parallel

## Changes
- **Dockerfile**: Restructured into `base`, `sqlite`, `postgres`, `dev`, `postgres-dev` targets. SQLite and Postgres binaries are built to separate directories (`target/` and `target-postgres/`)
- **docker/entry-postgres.sh**: New entrypoint that starts embedded Postgres, creates the database, imports a dump file if mounted at `/database/import.dump`, then hands off to the shared `entry.sh`
- **docker/local.postgres.yaml**: New Postgres config for in-container use
- **docker/README.md**: Updated with build instructions for both SQLite and Postgres (Docker and native Linux), and usage examples
- **.github/workflows/dockerise.yaml**: Parallel SQLite and Postgres build jobs, pushes all four image tags
- **server/**: Postgres database export endpoint and client UI integration

## Test plan
- [ ] Build SQLite binary and Docker image (`--target sqlite` or default)
- [ ] Build Postgres binary with `--target-dir target-postgres` and Docker image (`--target postgres`)
- [ ] Run Postgres image with a mounted `pg_dump` file and verify database is imported
- [ ] Run Postgres image without a dump and verify clean database is created
- [ ] Verify SQLite image still works as before
- [ ] Validate CI workflow with `act push -W .github/workflows/dockerise.yaml --dryrun` ([act](https://github.com/nektos/act) runs GitHub Actions locally)


docs/content/docker/_index.md
```diff
@@ -14,47 +14,218 @@ Dockerise github action is made to fire when a new tag is created, this tag need
...
+```bash
+cd server && cargo build --release --bin remote_server --bin remote_server_cli
+```
+
+### Postgres
+
+Via Docker (for macOS or other non-Linux hosts — uses `rust:1.94` non-slim because it includes `libpq-dev`):
+
+```bash
+docker run --rm --user "$(id -u)":"$(id -g)" -v "$PWD":/usr/src/omsupply -w /usr/src/omsupply/server rust:1.94 cargo build --release --bin remote_server --bin remote_server_cli --no-default-features --features postgres --target-dir target-postgres
```
ria8651, line 42: Is there a reason it isn't slim this time?

docker/local.postgres.yaml
```diff
@@ -0,0 +1,10 @@
+server:
+  danger_allow_http: true # allow http in production mode
```
ria8651, line 2: Why's this needed?

server/server/src/support/database.rs
```diff
@@ -37,14 +35,89 @@ pub async fn get_database(
...
+
+    let pg_dump_cmd = PathBuf::from(&pg_bin_dir).join("pg_dump");
+
+    let export_dir = PathBuf::from(&settings.server.base_dir);
+    std::fs::create_dir_all(&export_dir).map_err(|e| {
+        actix_web::error::ErrorInternalServerError(format!(
+            "Failed to create export directory: {e}"
+        ))
+    })?;
+    let export_path = export_dir.join("db_export.dump");
```
ria8651, line 76: Worth noting that this is left on disk forever, probably fine but would be good to have in readme somewhere.

docker/entry-postgres.sh
```diff
@@ -0,0 +1,37 @@
...
+# Create the database if it doesn't exist
+if ! gosu $PG_USER psql -lqt | cut -d \| -f 1 | grep -qw "$PG_DB"; then
+    echo "Creating database '$PG_DB'..."
+    gosu $PG_USER createdb "$PG_DB"
+fi
+
+# Import dump file if it exists
+if [ -f "$IMPORT_DUMP" ]; then
+    echo "Importing database dump from $IMPORT_DUMP..."
+    gosu $PG_USER pg_restore --clean --if-exists --no-owner --dbname "$PG_DB" "$IMPORT_DUMP" || true
```
ria8651: I think the `|| true` ignores all errors, maybe good to fail container start if the import fails?

ria8651 (commented):
Haven't finished testing yet but had some questions. Was also wondering why two build steps are required? This was the previous behaviour so its probably fine to keep but I would've thought one container that does the build and deployment would be preferable. Can still do volume mounts to cache the build or just rely on the layer caching built into docker. Though now I'm thinking about it thats probably better in a separate PR.

ria8651:
I tried running the dockerise workflow but it failed due to the async-graphql patch being outdated. So I removed the patch as it seems to have been made obsolete by https://github.com/async-graphql/async-graphql/pull/1780. That PR was merged and released in v7.1.0 and we're now using v7.2.1.

CarlosNZ:
> Was also wondering why two build steps are required?

I think it's so it can cross-compile for a different architecture. So the first step uses a containerised version of Rust to compile for the correct target architecture (arm/AMD), and then the second step just bundles that compiled binary into the image we're building.

> Can still do volume mounts to cache the build or just rely on the layer caching built into docker. Though now I'm thinking about it thats probably better in a separate PR.

Yeah, can probably be refined in a separate PR, just wanted to get this building for postgres now

docs/content/docker/_index.md
CarlosNZ: According to Claude:

> The SQLite version uses rust:1.94-slim because SQLite is bundled — the libsqlite3-sys crate compiles SQLite from source, so no system libraries are needed.
> 
> The Postgres version uses rust:1.94 (non-slim) because it needs libpq-dev to compile the diesel postgres driver, and libpq-dev is included in the full image but not in slim.
> 

Have clarified in doc: d09a698d0c

docker/local.postgres.yaml
CarlosNZ: That's just copied over from the existing SQLite .yaml file.

But it's telling it to accept http, cos the container doesn't need to deal with https, as the proxy on the host system deals with external https and passes http to the container.

server/server/src/support/database.rs
CarlosNZ: It gets overwritten with the same filename every time, so it's not accumulating.

docker/entry-postgres.sh
CarlosNZ: Thanks.  a02ad07a26

> The reviewer is right. The || true was there because pg_restore can emit non-fatal warnings (e.g. "table doesn't exist" when using --clean on a fresh database) which set a non-zero exit code even though the import succeeded. But it also swallows genuine failures like a corrupt dump file.
> 
> A better approach is to use --exit-on-error so pg_restore stops on real errors, and drop --clean since we're restoring into a freshly created empty database anyway

CarlosNZ:
> I tried running the dockerise workflow but it failed due to the async-graphql patch being outdated. So I removed the patch as it seems to have been made obsolete by [async-graphql/async-graphql#1780](https://github.com/async-graphql/async-graphql/pull/1780). That PR was merged and released in v7.1.0 and we're now using v7.2.1.

Cool, thanks.

How did you test run the workflow, BTW?

server/server/src/support/database.rs
ria8651: Oh yea, I just meant there'd be a single couple gig file sitting on the server forever. Not problematic.

CarlosNZ: Couple of gig?! My current dump is 1.5MB :)
(The pg_dump outputs seem to be about 1/10th the size of the full SQLite file)

docker/local.postgres.yaml
ria8651: Ah I see, makes sense. Is that in the docker README? I imagine most docker deployments would have proxies but good to mention somewhere.

ria8651:
> How did you test run the workflow, BTW?

Just pushed a tag on the branch. Workflows run using the branches workflow files so it used your updated workflow file. Though for some triggers like `issue_comment`, they can only run on the default branch.

https://github.com/msupply-foundation/open-msupply/actions/runs/24323343134

CarlosNZ:
@ria8651 Made some improvements, but done more in follow-up PR where I've (well Claude) created an interactive script to make the builds simpler: https://github.com/msupply-foundation/open-msupply/pull/11283

docs/content/docker/_index.md
jmbrunskill: I thought we used bundled sqlite with diesel?

CarlosNZ: @jmbrunskill yes, correct, hence the "slim" image is adequate for the SQLite build

ria8651 (approved):
All seems to work. Worked with podman instead of docker too so thats nice. The sqlite and postgres bind/import worked too. The first section of the docs for building the binaries is repeated in the later parts of the doc which was a bit confusing but once I realised it was just a repeat it made sense.

CarlosNZ:
> The first section of the docs for building the binaries is repeated in the later parts of the doc which was a bit confusing but once I realised it was just a repeat it made sense.

Thanks, I've improved the docs quite a bit in the [next PR](https://github.com/msupply-foundation/open-msupply/pull/11283)

PR #11046, opened 2026-03-27
10364 Store properties playing up when adding supply levels and re-initialising properties

### LINKED ISSUE #10364: Store properties playing up when adding supply levels and re-initialising properties
## What went wrong? 😲

I struggled to find a logic for what was happening but when adding supply levels in the Chad live server I've noticed that the store properties fields were moving around. When re-initialising they were still moving around as well. I think the video is better to explain what is going on as I tried multiple sequences of action and I couldn't work out what was happening.. For example, initially when I added the level "Regional" the order didn't change and when I added "Services de santé", the order started playing up.

https://github.com/user-attachments/assets/18f4c0a3-f2cc-4319-973a-8cabe195ce37

## Expected behaviour 🤔

I would expect the store properties to stay in the same order. (not sure if it is solved in 2.16? - there was an issue related to the order opened in May 2025 https://github.com/msupply-foundation/open-msupply/issues/7725)

## How to Reproduce 🔨

Steps to reproduce the behaviour (doesn't always do that though.. :( ):

1. Go to a store and configure a new supply level
2. See the supply level changing position in the store properties
3. Re-initialise GAPS properties
4. See the supply level changing position

## Your environment 🌱

- Open mSupply Version: 2.15.02 (https://tchad.msupply.org/)

roxy-dao:
Fixes #10364

# 👩🏻‍💻 What does this PR do?
Sort store properties before displaying them so they don't jump around

# 🧪 Testing

- [ ] Initialise store properties 
- [ ] Go view them
- [ ] Re-initialise
- [ ] View again - components shouldn't jump around

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [x] Frontend

### DIFF
--- a/client/packages/system/src/Name/ListView/Stores/StoreProperties.tsx
+++ b/client/packages/system/src/Name/ListView/Stores/StoreProperties.tsx
@@ -51,6 +51,7 @@ export const StoreProperties = ({
         .filter(
           p => p.property.key !== 'latitude' && p.property.key !== 'longitude'
         )
+        .sort((a, b) => a.property.name.localeCompare(b.property.name))
         .map(p => (
           <Row
             key={p.id}


aneeshthunga26 (approved):
Looks good, shows up alphabetically sorted. No jumping around

PR #11093, opened 2026-04-01
Add Playwright e2e smoke tests for all sections

ria8651:
> [!WARNING]

# 👩🏻‍💻 What does this PR do?

image

While updating react I ran into many infinite render issues due to more strict dependency updating. I found this playwright test very useful for testing what pages had issues and giving claude full console context for iteration.

Adds comprehensive Playwright e2e smoke tests that navigate through every main section of the app, clicking into detail views and all tabs, checking for infinite render loops and runtime errors.

**Sections covered:** Dashboard, Distribution, Replenishment (including R&R Forms), Inventory, Catalogue, Dispensary, Cold Chain, Manage (Central Server), Programs, Reports, Settings.

**Also:**
- Auth setup project logs in once and shares session across all workers
- `yarn e2e` command added (use `--headed`, `--workers`, `--ui` flags as needed)
- `BASE_URL` env var support for testing against different servers
- Refactored existing custom translations test to use shared auth
- Added `.eslintrc.js` in playwright dir to disable `no-console`
- Updated README with usage examples

- Sections run in **parallel** across Playwright workers, but tests within a section run **sequentially** (sharing one browser page)
- Uses `expect.soft()` so a failure in one tab doesn't abort the rest of the section
- Central-server-only and dispensary-only pages pass gracefully when not applicable (they just don't find rows to click)
- The `RENDER_SETTLE_MS = 300` wait after each action is the only manual wait — just enough for React to hit an infinite loop if one exists

# 🧪 Testing

- [ ] Run `cd client && yarn e2e --headed smoke --workers 1` with the dev server running
- [ ] All smoke test sections should pass
- [ ] Try `BASE_URL=http://localhost:9000 yarn e2e smoke` against a different instance

# 📃 Documentation

- [ ] **No documentation required**: internal test infrastructure (README updated within the PR)


client/playwright/e2e/smoke-all-sections.spec.ts
```diff
@@ -0,0 +1,245 @@
+/**
+ * Smoke tests: Navigate through all main sections, list views, and detail views
+ * to check for infinite renders, runtime errors, and crashes.
```
esmehm: Currently the test only fails on infinite rerenders - runtime errors are logged but never asserted on. 
Is it a deliberate choice to only fail for infinite rerender? To me it would make more sense to catch what we can

esmehm:
Hey @ria8651 , I have pushed commits to test the other tabs in external inbound shipments and to test the line edit modal (in both inbound and outbound shipments) to check for infinite rerenders too - please check you're happy with these :-) 
As we add more through tests, we could reduce these smoke tests but I think they add value for now - especially as we had the inifnite rerender bug in the line edit modal so much last week

Have also added an improvement to close any other modals which were open when going to the pages, as I got flakey results before that change

My above questions around why dont we assert on other errors beyond infinitie rerenders still holds when you circle back to this PR

Thanks!

esmehm:
I have also changed the waiting - it was not catching infinite rerenders as well as uit could do - so now it waits longer when it spots an error and short when it does not

Have raised a PR into here to resolve all the warnings / improve on the practices raised through these tests
https://github.com/msupply-foundation/open-msupply/pull/11171
It is important to resolve these so the above logic does not extend the test timings unnecessarily off relatively minor warnings

client/playwright/e2e/smoke-all-sections.spec.ts
```diff
@@ -0,0 +1,583 @@
...
+if (!fs.existsSync(screenshotDir)) {
+  fs.mkdirSync(screenshotDir, { recursive: true });
+}
+
+// How long to watch for infinite-loop warnings after navigation/click.
+// Loops usually fire within ~50ms, but async loops can take longer.
+// Clean pages exit early; pages with console activity get a growth check.
+const RENDER_SETTLE_MIN_MS = 300;
+const RENDER_SETTLE_MAX_MS = 2000;
+const RENDER_SETTLE_POLL_MS = 100;
```
ria8651: I don't think this will be very reliable. Relying on lots of console messages to detect infinite renders won't always work. If we need to wait 2 secs for each to catch infinite renders then we probably should do this every time and just run lots of chrome instances at once to speed it up.

If we upgrade to react 19 this issue should go away with much stricter infinite render protections so probably don't need anything too fancy till then.

ria8651 (commented):
> Have also added an improvement to close any other modals which were open when going to the pages

Why would a modal be open? Seems like a problem that should be inspected and the test to fail.

> My above questions around why dont we assert on other errors beyond infinitie rerenders still holds when you circle back to this PR

Yea that'd be great, just need to have a PR that fixes the warnings on all the pages.

I'll make the tests fail on warnings and errors now.

client/playwright/e2e/smoke-all-sections.spec.ts
esmehm: What we had before did not catch the infinite rerender we currently have on develop most of the time (just ocasionally did it fail) and so we need something stronger. This caught it all the time, but is not guaranteed to catching everything all of the time I agree. 
If we can wait longer then that would be betterr yes
And react 19 even better

ria8651: But if we just wait longer on every test?

esmehm: Yes if we can afford to wait longer on every test then great!

mark-prins:
automated UI tests!

esmehm (commented):
I think you have got this to a good place now - should merge and we can work from there on the issues it raises. Thanks for your work on this!

esmehm (approved):
I think you have got this to a good place now - should merge and we can work from there on the issues it raises. Thanks for your work on this!

Issue #11072, opened 2026-03-30
Auth: Users are getting logged out inconsistently


roxy-dao:
## Is your feature request related to a problem? Please describe 👀
Users are getting logged out inconsistently. 

## Describe the solution you'd like 🎁
Fix the authentication method so that this doesn't happen and also provide detailed diagrams and docs about how it works

mark-prins:
> provide detailed diagrams and docs about how it works
Out of curiosity - I thought to document the current process..


# Authentication Workflow

This document describes the authentication and authorization architecture of Open mSupply, covering the full lifecycle from login through token management, permission enforcement, and session termination. It is intended for developers working on the codebase and security reviewers assessing the system.

Open mSupply uses a **JWT-based access/refresh token pair** model. The Rust server (actix-web + async-graphql) issues tokens and enforces permissions per resolver. The React/TypeScript client manages tokens, handles refresh, and protects routes.

---

## Table of Contents

- [Architecture Overview](#architecture-overview)
- [Login Flow](#login-flow)
- [Token Lifecycle](#token-lifecycle)
- [Permission Model](#permission-model)
- [Session Management](#session-management)
- [Cookie and Storage Configuration](#cookie-and-storage-configuration)
- [Security Analysis](#security-analysis)
- [Key File References](#key-file-references)

---

## Architecture Overview

```mermaid
flowchart TB
    subgraph Client ["Client (React/TypeScript)"]
        LoginForm["Login Form"]
        AuthCtx["AuthContext"]
        RefreshHook["useRefreshToken"]
        GqlClient["GQLClient"]
        RouteGuard["RequireAuthentication"]
    end

    subgraph Server ["Server (Rust / actix-web)"]
        GqlEndpoint["GraphQL Endpoint<br/>authToken query"]
        AuthMiddleware["Auth Middleware<br/>validate_auth()"]
        TokenService["Token Service<br/>JWT issue / verify"]
        TokenBucket["TokenBucket<br/>(in-memory HashMap)"]
        PermEngine["Permission Engine<br/>ResourceAccessRequest"]
    end

    subgraph External ["External"]
        CentralServer["Central mSupply Server<br/>LoginApiV4"]
        Database["PostgreSQL / SQLite<br/>key_value_store"]
    end

    LoginForm -->|"username + password"| GqlEndpoint
    GqlEndpoint -->|"sync credentials"| CentralServer
    CentralServer -.->|"fallback to local"| Database
[...]

Issue #2832, opened 2024-01-26
Storing auth in cookie can reach cookie limit


andreievg:
I was looking at where to put is central server setting, and noticed auth cookie containing the following structure:

<details>
<summary>'auth' section of Cookie header</summary>

```JSON
{
    "store": {
        "__typename": "UserStoreNode",
        "code": "GEN",
        "id": "572B52B26FB54D4896C740F35BCBC75E",
        "name": "General",
        "storeMode": "STORE",
        "preferences": {
            "id": "572B52B26FB54D4896C740F35BCBC75E",
            "responseRequisitionRequiresAuthorisation": false,
            "requestRequisitionRequiresAuthorisation": false,
            "packToOne": true,
            "omProgramModule": false,
            "vaccineModule": true
        }
    },
    "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJleHAiOjE3MDYyMjg0NjEsImF1ZCI6IkFwaSIsImlhdCI6MTcwNjIyNDg2MSwiaXNzIjoib20tc3VwcGx5LXJlbW90ZS1zZXJ2ZXIiLCJzdWIiOiI4QjAwMTBDQkJDQzM0MzlGQkM4RTI5NzRGRDcxMkQwOSJ9._zHrFogB2rFM0ntL7ba6H90wF8lsguUcyl6ZoX3b6Ms",
    "user": {
        "id": "8B0010CBBCC3439FBC8E2974FD712D09",
        "name": "test",
        "permissions": [
            "INBOUND_SHIPMENT_QUERY",
            "SENSOR_QUERY",
            "REQUISITION_MUTATE",
            "ITEM_MUTATE",
            "REQUISITION_SEND",
            "STOCKTAKE_MUTATE",
            "SENSOR_MUTATE",
            "ITEM_NAMES_CODES_AND_UNITS_MUTATE",
            "STOCK_LINE_MUTATE",
            "COLD_CHAIN_API",
            "CREATE_REPACK",
            "INBOUND_SHIPMENT_MUTATE",
            "PATIENT_MUTATE",
            "PRESCRIPTION_MUTATE",
            "PATIENT_QUERY",
            "REQUISITION_QUERY",
            "STOCK_LINE_QUERY",
            "LOG_QUERY",
            "STOCKTAKE_QUERY",
            "REPORT",
            "LOCATION_MUTATE",
            "STORE_ACCESS",
[...]

PR #11095, opened 2026-04-01
11094 Auth cookies collide when running multiple instances on different ports

### LINKED ISSUE #11094: Auth cookies collide when running multiple instances on different ports
## Problem

When running two instances of open-mSupply on different ports in the same browser (e.g. `localhost:3003` and `localhost:3009`), users get logged out of both instances more frequently than expected.

### Root cause

Cookies are scoped by domain, not by port. Both instances use hardcoded cookie names:
- `auth` — set by the frontend via `js-cookie`, stores JWT token, user info, and selected store
- `refresh_token` — set by the server as an `HttpOnly` cookie

Since both instances run on `localhost`, they share the same cookie jar. Logging into one instance overwrites the other's cookies, causing:
1. The overwritten instance's token refresh check fails (wrong token)
2. The server rejects the mismatched `refresh_token`, triggering logout
3. The cycle repeats as each instance fights over the shared cookies

The `auth` cookie is also read server-side by `validate_cookie_auth` for file upload/download endpoints, so those break too.

Note: `localStorage` is scoped by origin (scheme + host + port), so it is not affected.

### Affected cookies

| Cookie | Set by | Name | Scoped by port? |
|--------|--------|------|----------------|
| `auth` | Frontend (`js-cookie` in `AuthContext.tsx`) | `"auth"` (hardcoded) | No |
| `refresh_token` | Server (`Set-Cookie` header in `login.rs`) | `"refresh_token"` (hardcoded) | No |

## Solution

Suffix cookie names with the port when running on a non-standard port. Standard port deployments (80/443) continue to use the original names — zero impact on production.

| Scenario | Cookie names |
|----------|-------------|
| Production (standard ports) | `auth`, `refresh_token` — **unchanged** |
| `localhost:3003` | `auth_3003`, `refresh_token_3003` |
| `localhost:3009` | `auth_3009`, `refresh_token_3009` |

### Files changed

**Frontend:**
- `client/packages/common/src/authentication/AuthContext.tsx` — Use `window.location.port` to build a port-aware cookie name

**Server:**
- `server/graphql/core/src/lib.rs` — Add `host_port` to `RequestUserData`, extract from `Host` header, use for `refresh_token` cookie lookup
- `server/graphql/general/src/queries/login.rs` — Port-aware `set_refresh_token_cookie()`
- `server/graphql/general/src/queries/logout.rs` — Pass `host_port` when clearing cookie
- `server/graphql/general/src/queries/refresh_token.rs` — Pass `host_port` when setting new cookie
- `server/graphql/lib.rs` — Add `host_port: None` for internal plugin requests
- `server/server/src/authentication.rs` — Port-aware `auth` cookie lookup for file endpoints

### Multi-instance isolation (the main fix)
- [ ] Run two server instances on different ports (e.g. 3003 and 3009)
- [ ] Open both in the same browser
- [ ] Log in to instance A (port 3003)
- [ ] Log in to instance B (port 3009)
- [ ] Verify instance A is still logged in — navigate around, check no unexpected logout
- [ ] Wait 2+ minutes (past a token refresh cycle) and confirm both sessions remain active
- [ ] Log out of instance B — confirm instance A remains logged in
- [ ] Inspect cookies in browser dev tools: confirm you see port-suffixed names (e.g. `auth_3003`, `refresh_token_3003`, `auth_3009`, `refresh_token_3009`)

### Single instance still works
- [ ] Run a single instance on any port
- [ ] Log in, select a store, navigate around
- [ ] Wait 2+ minutes to confirm token refresh works (no unexpected logout)
- [ ] Log out and confirm you are redirected to the login page

### File upload/download (uses `auth` cookie server-side)
- [ ] Log in to an instance on a non-standard port
- [ ] Upload a file (e.g. attach a document or import data)
- [ ] Download a file
- [ ] Confirm both operations succeed without auth errors

### Token refresh
- [ ] Log in and stay idle for ~58-59 minutes (or temporarily lower `INACTIVITY_TIMEOUT_MINUTES` for faster testing)
- [ ] Confirm the token refreshes automatically and you are not logged out
- [ ] Check browser dev tools: the `refresh_token_<port>` cookie should be updated with a new value after refresh

## Testing to be done by QA:
- [ ] Test document upload
- [ ] The rest shall be covered by general QA use - eg login / out,

esmehm:
Fixes #11094

# 👩🏻‍💻 What does this PR do?

Replaces the port-based auth cookie naming approach with a configurable `cookie_suffix` server setting, so that multiple instances on the same domain no longer overwrite each other's cookies.

Cookies are scoped by domain, not port. When two instances run on `localhost` (e.g. ports 3003 and 3009), they share the same `auth` and `refresh_token` cookies, causing each instance to overwrite the other's auth state and trigger frequent logouts.

The previous approach derived cookie names from the HTTP Host header port, but this was brittle behind reverse proxies (the frontend's `window.location.port` and the server's Host header could disagree, breaking auth in production).

Instead, the cookie name suffix is now an explicit server setting (`server.cookie_suffix` in yaml or `APP__SERVER__COOKIE_SUFFIX` env var). The server exposes it to the frontend via the `initialisationStatus` GraphQL query, so the frontend automatically uses the correct cookie name — no frontend config needed.

**Example usage:**
```bash
# Instance 1 — no suffix needed (uses bare 'auth' / 'refresh_token')
cargo run

# Instance 2 — set a suffix to avoid collision
APP__SERVER__COOKIE_SUFFIX=central ~/run-central.sh
```

## Changes

**Server:**
- `settings.rs` — add `cookie_suffix: Option<String>` to `ServerSettings`
- `initialisation_status.rs` — expose `cookie_suffix` in the GraphQL response
- `login.rs` / `logout.rs` / `refresh_token.rs` — read suffix from settings instead of Host header
- `authentication.rs` — `validate_cookie_auth` accepts suffix parameter
- All file upload/download/label/support endpoints — pass suffix from settings
- Remove `port_from_host` and `host_port` plumbing entirely

**Frontend:**
- `AuthContext.tsx` — replace static `AUTH_COOKIE_NAME` with dynamic `authCookieName` set via `setAuthCookieSuffix()`
- `CookieConfigProvider.tsx` (new) — fetches `initialisationStatus` with suspense, sets cookie name before `AuthProvider` mounts
- `Host.tsx` — wrap `AuthProvider` with `CookieConfigProvider`

- When `cookie_suffix` is unset (the default), bare `auth` and `refresh_token` names are used — fully backward compatible, no one-time logout
- The suffix can be set via yaml (`server.cookie_suffix`) or env var (`APP__SERVER__COOKIE_SUFFIX`)
- The frontend gets the suffix from the server automatically — only the backend needs configuration
- `CookieConfigProvider` suspends (via `useInitialisationStatus`) so the cookie name is always set before `AuthProvider` reads any cookies

# 🧪 Testing

### Multi-instance isolation (the main fix)
- [ ] Run two server instances on different ports, with different `APP__SERVER__COOKIE_SUFFIX` values
- [ ] Open both in the same browser
- [ ] Log in to both instances
- [ ] Verify both sessions remain active independently
- [ ] Log out of one — confirm the other remains logged in
- [ ] Inspect cookies in browser dev tools: confirm you see suffixed names (e.g. `auth_central`, `refresh_token_central`)

### Single instance still works (no suffix configured)
- [ ] Run a single instance without setting `cookie_suffix`
- [ ] Log in, select a store, navigate around
- [ ] Wait 2+ minutes to confirm token refresh works (no unexpected logout)
- [ ] Log out and confirm redirect to login page
- [ ] Inspect cookies: should see bare `auth` and `refresh_token` names

### File upload/download
- [ ] Log in to an instance with a cookie suffix set
- [ ] Upload a file (e.g. attach a document or import data)
- [ ] Download a file
- [ ] Confirm both operations succeed without auth errors

### Token refresh
- [ ] Log in and wait ~58-59 minutes (or temporarily lower `INACTIVITY_TIMEOUT_MINUTES`)
- [ ] Confirm the token refreshes automatically
- [ ] Check browser dev tools: the refresh token cookie should be updated

# 📃 Documentation

- [x] **No documentation required**: config option is documented in `base.yaml` and `example.yaml`


esmehm:
> I think some production servers run on port 8000 but maybe it doesn't matter if you get logged out on updating?

Thanks for flagging!
Assuming theyre directly on 8000 they'll have a one time logout. But I think this minor one time inconvenience is fine

If they were behind a reverse proxy then they wouldnt get logged out, but as far as I can tell it's not configured that way?

ria8651 (changes requested):
I think claude has some good points, the way cookies are matched between client and server seem quite brittle especially with proxies involved that can easily change ports in transit and might not update the host header. As such this could cause the pretty bad problem in production of just not being able to log in. If you can think of a less brittle approach that would be better but otherwise I'm not sure if this is worth it.

ria8651:
One solution that I though would be interesting to look into was from marks comment [here](https://github.com/msupply-foundation/open-msupply/issues/11072#issuecomment-4158727419):

> 2. Access Token Stored in JS-Accessible Cookie
> 
> Risk: High | Category: Cross-Site Scripting (XSS)
> 
> The auth cookie is set via js-cookie and is not HttpOnly. Any JavaScript running on the page -- including injected scripts from an XSS vulnerability -- can read the access token, user data, and store information.
> 
> Impact: An XSS attack can exfiltrate the access token and impersonate the user for up to 60 minutes.
> 
> Recommendation: Store the access token in memory only (e.g., a React context or closure variable) rather than in a cookie. If a cookie is required for SSR or cross-tab persistence, mark it HttpOnly and use a backend-for-frontend pattern.

This would make it easy to solve this problem by using client logic to pick the cookie to send.

esmehm:
Thanks very much for you feedback both @ria8651 and @jmbrunskill ! Have changed to use suffix set in env var / yml instead of relying on the ports. Please let me know what you think :-)

ria8651:
I would much rather have the server set the access token cookie as HttpOnly (like it already does for refresh_token), and expose user info / permissions via a separate authenticated query. Theres already #2832 related to this and would be nice to stop the cookie being extractable from JS. It would be a larger refactor but at least it would solve all the problems at once and make things simpler instead of more complex.

ria8651:
I'll have a go at the http only cookie, seems interesting

ria8651:
I had a go at a deeper change to the auth flow doing http only cookies in #11223

esmehm:
CLosing in favour of #11223 or further auth work in cooldown

PR #10907, opened 2026-03-18
10929 - authorising a line in external inbound shipment

esmehm:
Fixes #10929 and fixes #11015

# 👩🏻‍💻 What does this PR do?

Shows the authorisation status of a line on the inbound line card when applicable (ie on external inbound shipments `Inbound shipment (external) lines must be authorised:` store pref turned on).
Allows the status to be edited in the card via a dropdown and the footer actions, according to the conditions below

### Enforce user perms:
- Can only approve / reject lines if user has authorisation permissions (authorise goods received user permission in legacy mSupply)
- Can only edit details on approved / rejected lines if user has auth perms (otherwise need to change status to back to pending in order to edit)
- Line status can be changed between approved, pending and rejected in any order (if have sufficient permissions) until the shipment is received. Once the shipment is received, the line status cannot be changed. (note a shipment can only be received once all lines are either approved or rejected, no lines can be pending at this point)
- Users with auth permissions can still update line details after the shipment is received, but cannot add / delete / duplicate a batch as this adds a line. Users without auth permssions cannot edit line details after the shipment is received (as the status is approved / rejected so they dont have perms to change the details anymore).
- Once the shipment is verified, nobody can change the line details (this is #11015 )

### Approval status in card view
- Status field added to card view
- Styled to match other card inputs
- Only shows allowed statuses in the dropdown according to shipment stage and user perms

### Pending action
- Added a "Pending" footer action alongside Approve/Reject, allowing lines to be set back to pending status
- New confirmation modal and translations for the pending action

### Bug fixes
- **Rejected lines not updatable**: The `updateLines` filter only included `StockIn` type lines, but rejected lines have type `UnallocatedStock` — so status changes on rejected lines were silently dropped. Now includes `UnallocatedStock` in the update filter.
- **Rejected lines missing from PO line edit modal**: The existing lines filter in `useDraftPurchaseOrderInboundLines` excluded `UnallocatedStock` lines, so opening a rejected line's edit modal showed a blank draft instead of the actual line.

### Stock safety
Line status transitions (Pending/Passed/Rejected) are freely allowed at Delivered status. The validation blocks status changes once the invoice reaches Received, so no stock lines exist when users are editing approval status — all transitions are safe.

Screenshots:
image
Just the pill icon for received shipment -> status cannot be changed

image
Only pending option coz user doesnt have permission to make apporved / rejected lines / approve/reject preexisting lines (chose to only show what they can do as clearer for user and simpler implementation)

All 3 options
image

[...]

PR #11106, opened 2026-04-01
Fix inbound line edit: cell remount and CannotEditCostPrice

### LINKED ISSUE #11105: [Inbound Shipment] `Location` or `Manufacturer` don't get selected on first click
## What went wrong? 😲

The user is unable to select values in the `Location` or `Manufacturer` fields easily: 

https://github.com/user-attachments/assets/79566371-72cf-48ec-bfee-78c54952be54

https://github.com/user-attachments/assets/f2143c82-0c44-49f7-8fef-86d37a5ff417

## Expected behaviour 🤔

Value should be selected on first click. 

## How to Reproduce 🔨

1. Open a line on an Inbound Shipment
2. Attempt to select `Location` or `Manufacturer` value

## Your environment 🌱

- Open mSupply Version: 2.17.00-RC-03310809
- Platform: Windows standalone

esmehm:
As raised here: https://msupply-workspace.slack.com/archives/C0AH1KVC07K/p1775018927869939
Closes #11105

## Summary

Fixes two regressions in the Inbound Shipment line edit modal (InboundLineEditCards):

### Bug 1: Input cells remounting on every keystroke
- **Symptom**: Typing in any input field (batch, number of packs, etc.) was unusable — each keystroke caused the cell to lose focus and reset, making it impossible to enter more than one character.
- **Cause**: PR #10907 added `statusMap` (from `useInvoiceLineStatusMap()`) to the columns `useMemo` dependency array. `useInvoiceLineStatusMap` depends on `useAppTheme()`, which creates a new theme object on every render (no memoization). This made `statusMap` a new reference every render → columns `useMemo` invalidated → new `Cell` function references → `flexRender` unmounted and remounted every cell → all local state (buffer, focus) destroyed.
- **Fix**: Store `statusMap` in a ref (`statusMapRef`) and read from `statusMapRef.current` inside Cell functions. Remove `statusMap` from the `useMemo` dependency array so columns stay stable across renders.

### Bug 2: `CannotEditCostPrice` error when saving line edits
- **Symptom**: Editing any field on an inbound line (for PO-linked or internal supplier shipments) and clicking OK produced a `CannotEditCostPrice` error, even though the cost price field was disabled and unchanged.
- **Cause**: The `toUpdateLine` parser always includes `costPricePerPack` in the update mutation input. The server validation (added in PR #10906, refined in PR #10932) compares the sent value with the stored value using exact `!=` on f64. The value can drift slightly through the JSON serialization round-trip (Rust f64 → GraphQL JSON → JS Number → GraphQL JSON → Rust f64), causing a false positive.
- **Fix**: Use epsilon comparison (`(new - old).abs() > f64::EPSILON`) instead of exact `!=` in the server validation, tolerating insignificant float drift while still rejecting actual cost price edits.
I am not sure why this problem didnt come up in that PR itself - perhaps a combination of changes - needs further thought later

## Test plan
- [ ] Open an inbound shipment linked to a PO
- [ ] Edit a line — verify typing in batch, number of packs, and other fields works normally without focus jumping or resetting
- [ ] Save the line edit — verify no `CannotEditCostPrice` error
- [ ] For a manual (non-PO) inbound shipment, verify cost price can still be edited and saved
- [ ] For a PO-linked shipment, verify that deliberately changing cost price (e.g. via API) is still rejected

### DIFF
--- a/server/service/src/invoice_line/stock_in_line/update/validate.rs
+++ b/server/service/src/invoice_line/stock_in_line/update/validate.rs
@@ -134,10 +134,12 @@ pub fn validate(
         }
     }
 
-    // Cost price is read-only for internal suppliers and external suppliers linked to a PO
+    // Cost price is read-only for internal suppliers and external suppliers linked to a PO.
+    // Use epsilon comparison to allow unchanged values that may have drifted
+    // through floating point serialization (Rust f64 → JSON → JS Number → JSON → f64).
     if let Some(new_cost_price) = input.cost_price_per_pack {
         if (invoice.name_store_id.is_some() || invoice.purchase_order_id.is_some())
-            && new_cost_price != line_row.cost_price_per_pack
+            && (new_cost_price - line_row.cost_price_per_pack).abs() > f64::EPSILON
         {
             return Err(CannotEditCostPrice);
         }


esmehm:
Cause analysis for retro:
Bug 1 was caused when I addressed what appeared to be simple review feedback in https://github.com/msupply-foundation/open-msupply/pull/10907 on my last commit, and I checked it loaded properly but didnt go through the edit and submit process which would have shown this bug. -> check the flow in future after last commit (and even better if e2e tests could catch this too)
Bug 2 - this is a floating point error - when floating points are sent backwards and forwards Rust f64 → JSON → JS Number → JSON → f64 and so we didnt surface it in dev / PR review as it only occurs on particularly obtuse numbers and not the round ones we used until I battle tested this PR, something I should have thought of further when writing the initial PR and hopefully would have otherwise surfaced later in testing.

Issue #7948, opened 2025-05-29
Logged out in a loop from omSupply on a browser


marthakeezy:
## What went wrong? 😲

Reported first in our QA channel, just bringing it to GitHub :)

I had 3 tabs opened, one in Prescriptions, one in View Stock and one in Catalogue -> Items -> Ledger. The system booted me out on one tab, I tried to log in again at which point I was being logged out in a loop: 

https://github.com/user-attachments/assets/f11652b3-f0cc-4a48-bf76-2037672401ae

Personally, the loop behaviour I've encountered for the first time even though I have multiple tabs opened at all times, but what I have encountered before (and so did Mark and Adam) is when you are logged out -> log in -> logged out again -> log in and it works on the second try - never the repeated prompts

## Expected behaviour 🤔

It is very likely users will have multiple tabs opened, so we should allow for that behaviour and also just ask to have them log in once. 

Notes from Mark: 

`The standard 'please login again, again' problem seems like it would be easy to resolve but I've never looked into it. surely it's a matter of on the login page clearing the local storage (/error/auth) and then reloading state? or something like that?`

Notes from Andrei: 

`The login issues i've seen like this is when you have multiple tabs open, let the token expire, i think it's like an hour, and then re-connect one tab, the second tab will be in error like this (refresh token is saved in cookie by the server, but it's not valid anymore as the other tab has used it)`

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Have multiple tabs of omSupply open and log on every one
2. Let them sit for a while until you are logged out
3. Try and log in on one, keep another one logged out? I assume?
4. Sorry, no concrete replication steps beyond that - but it seems like everyone knows what I am talking about :)

## Your environment 🌱

- Open mSupply Version: V2.8.00-RC3
- Legacy mSupply Central Server Version: V8.02.02

- Platform: Google Chrome (browser for Mac)

- Database type: PostgreSQL

jmbrunskill:
I suggest we add this too cool down as we are al experiencing this a lot as Devs...

mark-prins:
triage: agree that this is a good cooldown candidate; and an annoyance which has been observed sporadically in the dev team

Issue #5210, opened 2024-10-23
While navigating filters from reports, it logged me out


regotaina:
## What went wrong? 😲

While using the filters on the Reports section today, OMs logged me out for no reason, with message 'Authentication error'.
It was unexpected as the screen was active (browser).

Maybe logged me out due to sync? @alainsussol also saw this one

## Expected behaviour 🤔

App oes not logout user while performing tasks

## How to Reproduce 🔨

I'm not sure if it's easy to reproduce this one.

## Your environment 🌱

- Open mSupply Version: V2.3.00
- Legacy mSupply Central Server Version: V7.17

- Platform: Browser (chrome)

andreievg:
Would be keen to keep this open to figure out when this happens (is the system being used before hands ? or was the server recently restarted ?) and how frequent this happens @regotaina @alainsussol @marthakeezy would be great if you can add observation about this log out when you face it

marthakeezy:
Is it the same one as in #3967?

regotaina:
> #3967

This happened while we were using the reports filter. We checked a few reports , exploring them and when selected the filter, the system logged us out, in the middle of the operation. 

No restart at the time, we were just navigating the reports module. It doesn't seem easy to replicate, but I can check the logs to see if something came up. Alain had a theory that it was trying to sync ?

Good thing is that when we logged back, the app re-opened in the same window we were seeing before.

marthakeezy:
being endlessly logged out on `2.10.00-RC-08181654`

https://github.com/user-attachments/assets/6e53e53c-9173-45af-a961-17c3451b0b40

jmbrunskill:
> We have reports of this in production environments (v2.11) - do we have a separate issue for this? Can we have an update on this?
> 
> I imagine there could be multiple reasons for something like this to happen, so it might make the issue harder to identify?

@kahn-msupply there's been a few attempts to fix this but like you say there might be a few things going on. I have asked @claude about it and it suggested a change but I'm not convinced it's the source of these errors. Do you have any more details from production that might help narrow down the issue. Feel free to dm me if so!

Issue #3967, opened 2024-05-21
Android: not logging user out after inactivity


mark-prins:
## What went wrong? 😲

Issue raised by Nisha via telegram, validated by Maria & Adam, as noted:

**Nisha**
If we just leave the tablet with omsupply open then this kind of wrong alert arises in every page which is Authentication error permission denied. 

https://github.com/msupply-foundation/open-msupply/assets/9192912/8c17ca68-fe8a-4540-9964-251d564cddac

**Maria**
I noticed that too! Sometimes when I leave it open on the synchronisation screen for too long, that message pops up but the sync is still happening in the background. The name of the store and the admin area disappears though

![telegram-cloud-photo-size-2-5345975398997024008-y](https://github.com/msupply-foundation/open-msupply/assets/9192912/439745b3-3b15-4050-bc9e-6c5f2e862edd)

![telegram-cloud-photo-size-2-5345975398997024009-y](https://github.com/msupply-foundation/open-msupply/assets/9192912/9fb88304-fca9-4cbd-a866-c9026fa8237e)

**Adam**
I've definitely noticed that the tablet doesn't actually log you out when the timeout is reached, so it lets you do some actions and then suddenly says "oh, actually you're not logged in" - it's a bit of a frustrating UX to be honest (and possibly negates the security reason for a timeout if you can still access some info if the screen is left on?)

**Nisha**
this one, I’ve also seen many times…but the steps are not really traceable. 
At times, it logs me out after few mins of inactivity, while sometimes, the session is still active even after leaving the screen inactive for more than 30 mins. 🤡

**Maria**
I’ve only experienced the timeout when leaving the tablet on & inactive in the admin section - has it been happening in other areas of omSupply?

**Adam**
Yeah, it's happened in random places for me - I last touched this over 30 mins ago on this screen
So will probably let me do a little stuff if I start playing with it and then I'll run into an issue and it will log me out

## Expected behaviour 🤔
I think there should be 'You're not logged in....' message instead of 'Permission denied' message in that alert

## How to Reproduce 🔨

Steps to reproduce the behaviour:

1. Log in to the android app
2. Leave alone for an hour+

## Your environment 🌱

- Open mSupply Version:
- Legacy mSupply Central Server Version: Android

- Platform:

- Database type: SQLite3

andreievg:
daily triage: Setting as normal because the app is still quite usable, but does need improvement.

[...]

PR #11107, opened 2026-04-01
11072 Auth: Users are getting logged out inconsistently

### LINKED ISSUE #11072: Auth: Users are getting logged out inconsistently

roxy-dao:
Fixes #11072, #7948 & #5210 & #3967

# 👩🏻‍💻 What does this PR do?
- User activity isn't tracked. It was tracked by api calls. Now refresh token won't be fetched if the user has been inactive for more than an hour based on their interaction with the app
- Moved checking of how long the token should live to backend so we don't have 3 places in frontend racing against each other and causing expected log outs.
- Reduced refresh token lifetime from 6 hours to 2 hours... let me know if I should bump it back up to 6 hours. It's still outliving the auth token lifetime, so thought would be ok? 
- Upped the expiry check to 5 minutes to allow leeway
- Have added basic docs too~ Can make a more complicated diagram with permissions and all that jazz~ just thought a simpler one was easier to digest

This doesn't solve any of the security issues. It only solves the inconsistent logout

# 🧪 Testing

- [ ] For devs, you can shorten the times so you can test easier 😉 
- [ ] Login 
- [ ] Leave the app idle for over 60 mins and confirm that the user is logged out
- [ ] Use the app and test that it doesn't log you out while you're using it even after 60 minutes has passed
- [ ] Also test on android that the user is being logged out after 60 mins of inactivity

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


aneeshthunga26 (changes requested):
User activity has to be synced between tabs, maybe using local storage. Otherwise as long as one tab is inactive, it could trigger logout

roxy-dao:
> User activity has to be synced between tabs, maybe using local storage. Otherwise as long as one tab is inactive, it could trigger logout

Mine seem to persist as long as one tab is active. Are you perhaps on different ports?

mark-prins:
woo! thanks @roxy-dao !!! 🫶  This is great ty

Re the trigger of activity: the use of API calls to indicate user activity was deliberate, it is actually quite difficult to use the site _without_ causing an API call. 

we are now relying on browser event handlers in place of a definitive action which is potentially less reliable and requires adding a number of event handlers. I wondered if these are firing reliably so did a quick test: I don't see any scroll events at all.. and probably just the _mousedown_ and _touchstart_ are enough tbh. we don't need to be updating on every keystroke

I really like the need to no longer exclude certain system generated API calls though! 

Not sure what the refresh token is giving us now. can you explain that for me? as far as I can tell this only extends the inactivity timeout to 2h which is more than intended. inactivity of 60minutes is already pretty long.

what do you think about a config / preference for inactivity? I'm sure that it would be desired.. some RFPs have very short and some require very long 🙄

aneeshthunga26:
> > User activity has to be synced between tabs, maybe using local storage. Otherwise as long as one tab is inactive, it could trigger logout
> 
> Mine seem to persist as long as one tab is active. Are you perhaps on different ports?

I missed a location when reducing the time intervals. It works

aneeshthunga26 (approved):

roxy-dao:
> Not sure what the refresh token is giving us now. can you explain that for me? as far as I can tell this only extends the inactivity timeout to 2h which is more than intended. inactivity of 60minutes is already pretty long.

I'll set inactivity lower for now so it doesn't extend logout to 2 hours and make a follow up for below (: 

> what do you think about a config / preference for inactivity? I'm sure that it would be desired.. some RFPs have very short and some require very long 🙄

Yeah! I thought about that at first but thought to get something out to get feels :D Can create a follow up for this!

PR #11171, opened 2026-04-07
fix: resolve React warnings surfaced by smoke tests

esmehm:
## Summary

The Playwright smoke tests check for React console warnings and report them alongside test results. This PR fixes all the warnings that were surfaced, improving React correctness across the codebase.

### Why this is needed for test performance

The smoke tests use a polling settle mechanism to detect infinite render loops: after each navigation, they wait and watch for React's "Maximum update depth exceeded" warning. **Clean pages** (no console errors) exit after 300ms, but **pages with warnings** trigger extended polling for up to 2s — because any error *could* be the start of an infinite loop. This means every unfixed warning slows down the test suite. Clearing these warnings lets the tests run at full speed.

### What was changed

- **Wrapped all custom SVG icon components with `React.forwardRef`** — MUI components like `Tooltip` and `IconButton` pass refs to their children. Plain function components can't hold refs, which triggered warnings like _"Function components cannot be given refs"_. All ~80 custom icon components in `client/packages/common/src/ui/icons/` were updated.
- **Wrapped `BufferedTextInput` with `React.forwardRef`** — Same issue, this component is used inside `Tooltip` in purchase order and internal order toolbars.
- **Widened `Icon` prop type to `React.ComponentType`** — `SettingsSection` and `ReportWidget` typed their `Icon` prop as a plain function, which didn't accept `forwardRef`-wrapped components.
- **Added `container` prop to `Grid` with `direction`** — MUI's `Grid2` ignores the `direction` prop unless `container` is also set. Fixed in `SideBarComponents.tsx` (related documents section).
- **Coalesced nullable string fields to empty strings** — Fields like `theirReference`, `comment`, and `transportReference` can be `null` from the GraphQL fragment, but React warns when `null` is passed as the `value` prop on an `<input>`. Added `?? ''` fallback in supplier return detail views.
- **Moved React `key` prop to outermost element** in `splitTranslatedLines` — React requires keys on the top-level element returned from `.map()`.
- **Filtered invalid sort columns from URL params** — Prevented crashes when URL sort parameters reference columns that don't exist in the table.

### Why these warnings matter beyond test speed

These warnings indicate real issues — refs that silently fail, props that are silently ignored, and controlled inputs with inconsistent values. Fixing them:
- Prevents subtle bugs (e.g. a `Tooltip` that can't position itself because it couldn't attach a ref to its child)
- Eliminates noise in the console, making genuine new warnings easier to spot
- Aligns with React best practices for ref forwarding and controlled components

## Test plan
- [ ] Run `yarn e2e smoke` — all 52 tests should pass with zero warnings
- [ ] Spot-check UI in purchase orders, internal orders, and supplier returns to confirm no visual regressions


client/packages/common/src/ui/layout/tables/useBaseMaterialTable.tsx
```diff
@@ -95,7 +95,20 @@ export const useBaseMaterialTable = <T extends MRT_RowData>({
...
+  // in a detail view). Ignore sort entries that reference columns this table
+  // doesn't have, to avoid TanStack Table "Column does not exist" errors.
+  const columnIds = useMemo(
+    () => new Set(columns.map(c => c.id ?? c.accessorKey)),
+    [columns]
+  );
+  const sorting = useMemo(
+    () => urlSorting.filter(s => columnIds.has(s.id)),
+    [urlSorting, columnIds]
+  );
```
ria8651: I'd rather use `noUrlFiltering` (renamed) for both sorting and filtering

client/packages/common/src/ui/icons/Alert.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
 import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
 
-export const AlertIcon = (
-  props: SvgIconProps & { fill?: string; double?: boolean }
-): JSX.Element => {
+export const AlertIcon = React.forwardRef<
```
ria8651: Would be nice to have a `displayName`. React.forwardRef creates components that show as ForwardRef (or Anonymous) in React DevTools, making debugging harder.

On the other hand almost all of our component functions would need forwardRef if we go down this path... would it be better to just wrap things a Box if we want to use them in a tooltip or other component that needs forwardRef? Seems easier than having forwardRef on every FC.

ria8651 (changes requested):
Just the comments bellow

ria8651:
Pushed some changes. Still getting some infinite render fails but I've fixed them in #10770 and had trouble fixing here so I'll just leave it till then.

ria8651 (approved):
LGTM, just check my changes

ria8651:
I don't think we want local state only for the detail tables, they should keep their state upon refreshing the page or pressing back on the browser.

ria8651:
With stacked PRs I find it nicer to merge higher PRs after the PRs they depend on are merged cause now the playwright PR has the diffs from this PR as well making it harder to reveiw.

PR #11172, opened 2026-04-07
#11063 Backdate inventory adjustments

### LINKED ISSUE #11063: Backdating: Support entering backdated inventory adjustments
## Is your feature request related to a problem? Please describe

Part of the Backdating epic (#11056). Users need the ability to specify a historical date for inventory adjustments to handle missed data entry scenarios.

### Date Input
- Allow the specification of the **date for the inventory adjustment**, defaulting to the current date.
- Add a global preference for **"Allow entry of inventory adjustment date"** which enables the option.

### Date Validation (same restrictions as mSupply)
- **Issued date check**: When adjusting, prevent the adjustment if the stock has been issued (via outbound or prescription) after the date selected.
- **Negative adjustment**: Same as prescriptions — check that the ledger would **never go below 0** at any point in time after the selected date.
- The date must respect the **Maximum days before the current date** global preference.

## Describe alternatives you've considered

N/A — this is the required approach as specified.

## Additional context

Linked to epic: #11056
Uses the same restriction patterns as mSupply for date validation.

## Acceptance Criteria

- [ ] "Allow entry of inventory adjustment date" global preference exists and can be toggled on/off
- [ ] Date input field is available on inventory adjustments, defaulting to the current date
- [ ] Date field is only editable when the global preference is enabled
- [ ] Cannot select a date if stock has been issued after that date
- [ ] For negative adjustments, validates that the ledger never goes below 0 after the selected date
- [ ] Maximum days before current date preference is respected
- [ ] Follows the same restriction patterns as mSupply

## Moneyworks Jobcode

OMS:COGD

ria8651:
> [!WARNING]

Fixes #11063

image

# 👩🏻‍💻 What does this PR do?

Adds the ability to specify a historical date when creating inventory adjustments.

### Preference (commit 1)
- New global preference `AllowBackdatingOfInventoryAdjustments` in the Backdating group
- `MaximumBackdatingDays` is now enabled when **either** backdating preference is on (OR logic, was previously only tied to shipments)
- Render order: shipments toggle, adjustments toggle, max days

### Backend (commit 2)
- `backdated_datetime` field on `CreateInventoryAdjustmentInput`
- Validation:
  - Cannot set future date (tz-safe: UTC date + 1 day)
  - Respects `MaximumBackdatingDays` preference
  - For reductions: validates the ledger never goes below 0 at any point after the backdated date (uses `get_historical_stock_line_available_quantity` which computes the minimum available packs in the time range)
- Sets `verified_datetime` and `backdated_datetime` on the invoice
- Stock movements automatically use the correct date because the `inventory_adjustment_stock_movement` SQL view uses `verified_datetime` as its datetime

### Frontend (commit 2)
- Date picker in AdjustmentForm (shown when preference enabled)
- `maxDate` = today, `minDate` = MaximumBackdatingDays preference
- `maximumBackdatingDays` added to the frontend preferences query

### Combined Backdating Preference (review feedback)
- Merged `AllowBackdatingOfShipments`, `AllowBackdatingOfInventoryAdjustments`, and `MaximumBackdatingDays` into a single `Backdating` preference with `shipmentsEnabled`, `inventoryAdjustmentsEnabled`, and `maxDays` fields
- Custom `EditBackdating` component in the preference UI
- Removed the preference dependency system (max days disabling handled inline by the component)

### Historical Stock Display (review feedback)
- When a backdated date is selected, the adjustment stats update to show **historical** values:
  - **Available packs** → `min_available` (safe reduction limit)
  - **Packs on hand** → historical total at that date
  - "Current" label changes to "Historical"
- `HistoricalQuantities` struct — `get_historical_stock_lines_available_quantity` now returns both `min_available` and `total` (previously `total` was computed but discarded)
- Loading state with `keepPreviousData` for smooth transitions

### DateTime<Utc> for backdating inputs (review feedback)
- Inbound `received_datetime` and inventory adjustment `backdated_datetime` changed from `NaiveDate`/`NaiveDateTime` to `DateTime<Utc>` — proper UTC timezone handling, no `+ Duration::days(1)` hacks
- Frontend sends `.toISOString()` — midnight local time becomes the correct UTC moment
- Added `BackdatingNotEnabled` validation check for both inbound and inventory adjustments

### Differences from issue

- **"Issued date check"** (prevent if stock issued after date) — **Not implemented as a separate check.** The "ledger never goes below 0" validation (requirement #2) supersedes this. If stock was issued after the backdated date and you try to reduce at that date, the minimum-available check will catch it because the ledger would dip below 0 at the point of the outbound/prescription. For positive adjustments, the issued-date check is irrelevant. So requirement #2 covers #1.

- **"Allow entry of inventory adjustment date"** — implemented as `inventoryAdjustmentsEnabled` field on the combined `Backdating` preference

- The `add_new_stock_line` variant of inventory adjustments does **not** have backdating support yet — only `adjust_existing_stock`. New stock lines don't have historical movements to validate against.

### Combined preference migration
Users who had separate `allow_backdating_of_shipments` and `maximum_backdating_days` preference rows will need a data migration to the new `backdating` JSON struct format.

### Base branch
Based on `11062-backdating-received-date`.

# 🧪 Testing

- [ ] Enable "Backdating" → toggle on inventory adjustments in global preferences
- [ ] Set "Maximum days" to e.g. 30
- [ ] Open a stock line detail view and create an inventory adjustment
- [ ] Verify date picker appears, defaulting to today
- [ ] Select a past date — verify stats update to show "Historical" label with historical values
- [ ] Make a positive adjustment with a past date — verify it succeeds
- [ ] Make a negative adjustment that would cause ledger to dip below 0 — verify it's rejected
- [ ] Make a valid negative adjustment with a past date — verify it succeeds
- [ ] Check the stock ledger — verify the movement appears at the backdated date
- [ ] Try a date beyond the max days limit — verify it's blocked by minDate
- [ ] Disable the preference — verify the date picker disappears
- [ ] `cargo test insert_inventory_adjustment_backdated_ledger_below_zero`
- [ ] `cargo test update_inbound_shipment_backdate`

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


esmehm (approved):
Generally looks good!
Some thoughts:
- It would be safer to check server side whether backdating if allowed rather than relying on the UI as this allows direct stock changes by users who dont have access to change the pref, if bypassing our front end - but I know you also didnt do this on the other backdating so if that's our general approach then
- Would be great to add test for ledger gojng below zero

ria8651:
Added a test, did the TZ stuff from inbound and added backend validation. Can't believe I missed the backend validation, good catch.

ria8651:
Added one more commit to show the historical stock levels in the UI

image

esmehm:
Minor - on adjustments the copy shows as current for today when. you first open the moda, if you backdate it says historic but if you change the date back to today it still says history (I would expect current)

https://github.com/user-attachments/assets/cd7d13b3-a829-4f97-a761-0a4c68acd355

Only a nit in case you hadnt noticed - but I think it's acceptable to leave as is

client/packages/system/src/Stock/Components/InventoryAdjustment/AdjustmentForm.tsx
```diff
@@ -78,6 +87,30 @@ export const AdjustmentForm = ({
...
+              value={
+                draft.backdatedDatetime
+                  ? new Date(draft.backdatedDatetime)
+                  : new Date()
+              }
+              format="P"
+              onChange={date =>
+                setDraft(state => ({
+                  ...state,
+                  backdatedDatetime: date ? date.toISOString() : null,
```
esmehm: We pass this all the way through - if you select an old date but then change your mind and select today's date then it backdatedDateTime will have a value and the whole downstream will treat it under the backdating procedure:
- cosmetic `historical` vs `current` label i highlighted in the video is just the tip of the ice berg with this one I think
- if the time the datetime pivcker picks is earlier than another same day movement then this could produce somewhat unexpected `LedgerGoesBelowZero` errors
- new error variants dont come through properly so show as could not save

I think the simplest fix would be to check whether the date chosen is todsays date here and pass as null if so

esmehm (changes requested):
I have also added some more tests - please check you agree with them :)

esmehm (approved):
LGTM :)Thanks for your persistence

PR #11176, opened 2026-04-07
#11175 Fix infinite rerender on purchase order detail view

### LINKED ISSUE #11175: Infinite rerender on purhcase order details tab
## What went wrong? 😲

On develop , purhcase order details tab is infininitely rerendering

Image

## Expected behaviour 🤔
No infinite rerender

## How to Reproduce 🔨

## Your environment 🌱

- Open mSupply Version:

- Platform:

- Database type:

esmehm:
## Summary
- `useFormatCurrency` returned a new function reference on every render, cascading through `usePurchaseOrderColumns` → `useColumnOrder`, triggering a `setState` → rerender loop when no column order was saved in localStorage
- Memoized currency options and `formatCurrency` with `useMemo`/`useCallback` in `useCurrency`/`useFormatCurrency`
- Memoized error context callbacks and provider value in `usePurchaseOrderLineErrors` to prevent unnecessary consumer rerenders

Closes #11175

## Test plan
- [ ] Navigate to a purchase order detail view — confirm no infinite rerender / frozen UI
- [ ] Clear localStorage for the table (`@openmsupply-client/tables/purchase-order-detail-view`) and reload — confirm no infinite rerender
- [ ] Verify currency formatting still works correctly on PO line costs
- [ ] Verify column reordering and persistence still works
- [ ] Verify PO line error highlighting still works after a failed status change

### DIFF
--- a/client/packages/common/src/ui/layout/tables/tableState/useColumnOrder.ts
+++ b/client/packages/common/src/ui/layout/tables/tableState/useColumnOrder.ts
@@ -6,6 +6,7 @@ import {
   MRT_StatefulTableOptions,
   MRT_TableOptions,
 } from 'material-react-table';
+import { isEqual } from '@common/utils';
 import { getSavedState, updateSavedState, differentOrUndefined } from './utils';
 import { ColumnDef } from '../types';
 import { useGlobalTableDefaults } from './useGlobalTableConfig';
@@ -46,10 +47,13 @@ export const useColumnOrder = <T extends MRT_RowData>(
 
   // If initial state changes (due to plugin column loading, for example) and no
   // custom column order has been saved, update the column order to the new
-  // default
+  // default. Use functional updater to avoid rerender when content is unchanged.
   useEffect(() => {
     if (!getSavedState(tableId)?.columnOrder)
-      setState(globalDefaults?.columnOrder ?? initial);
+      setState(prev => {
+        const next = globalDefaults?.columnOrder ?? initial;
+        return isEqual(prev, next) ? prev : next;
+      });
   }, [initial, enableExpanding]);
 
   const update = useCallback<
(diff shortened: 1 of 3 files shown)


client/packages/common/src/ui/layout/tables/tableState/useColumnOrder.ts
```diff
@@ -46,10 +47,13 @@ export const useColumnOrder = <T extends MRT_RowData>(
...
   // custom column order has been saved, update the column order to the new
-  // default
+  // default. Use functional updater to avoid rerender when content is unchanged.
   useEffect(() => {
     if (!getSavedState(tableId)?.columnOrder)
-      setState(globalDefaults?.columnOrder ?? initial);
+      setState(prev => {
+        const next = globalDefaults?.columnOrder ?? initial;
+        return isEqual(prev, next) ? prev : next;
+      });
```
ria8651: I'd rather fix these kind of issues at the source then put on a bandaid that'll move the problems down the road

ria8651 (changes requested):
Nice to see this kind of thing being fixed, I'd rather remove the column order change though - we shouldn't be passing unstable references in in the first place.

Also not really related but we're getting console errors from the main detail view being mounted while the other tabs tables are being shown. When the url query parms change to match the tabs table the main detail view tab throw column doesn't exist error as it try's to sort by the other tabs columns. If you can find a nice solution would be nice to fix that here too otherwise we can fix it another time.

client/packages/common/src/ui/layout/tables/tableState/useColumnOrder.ts
esmehm: Agreed, reverted this thanks

esmehm:
> Also not really related but we're getting console errors from the main detail view being mounted while the other tabs tables are being shown. When the url query parms change to match the tabs table the main detail view tab throw column doesn't exist error as it try's to sort by the other tabs columns. If you can find a nice solution would be nice to fix that here too otherwise we can fix it another time.

Yes! I have already addressed this in my other PR here https://github.com/msupply-foundation/open-msupply/pull/11171/changes#diff-97f05c196af6e52afd531e500e098e279ff0a591e1f8c56af93d7875ef6e8c6aR105

ria8651 (approved):

PR #11168, opened 2026-04-06
#11062 Backdate inbound shipment received date

ria8651:
> [!WARNING]
> This PR was largely authored by Claude

Fixes #11062

# 👩🏻‍💻 What does this PR do?

image

image

Adds the ability to backdate the **Received date** on inbound shipments. The date can only be moved **earlier**, never forward — this prevents ledger inconsistencies without requiring line deletion.

### Backend
- `received_datetime` field on `UpdateInboundShipmentInput`
- Server-side validation:
  - Shipment must be in Received or Verified status
  - Date can only move earlier, never forward
  - Cannot be in the future (timezone-safe: UTC date + 1 day)
  - Respects `MaximumBackdatingDays` preference
- When backdating, **shipped**, **delivered**, and **created (new)** dates are automatically moved back if they are after the new received date
- Updates `location_movement.enter_datetime` for stock lines when backdating (single query, not N+1)
- Creates an `InvoiceDateBackdated` activity log entry with old and new received dates
- Postgres migration for the new activity log type (v2.18)
- Tests for error cases and success (including date auto-adjustment, location movement updates, and activity log)

### Frontend
- `ReceivedDateInput` component in the inbound shipment toolbar (visible for all inbound shipments)
- Always visible, disabled with contextual tooltip when:
  - "Date editing on shipments is not enabled" (preference off)
  - "Can only be changed once shipment has been received" (wrong status)
  - "Already beyond the maximum backdating limit" (max days exceeded)
- One-way warning confirmation showing the selected date before saving
- Stocktake warning (with date) if a stocktake exists after the selected date
- `maxDate` = current received date, `minDate` = backdating limit from preference
- Uses `useBufferState` to stay in sync after mutations
- Activity log displays "Received date backdated" with from/to dates

### Bug fix
- `DateTimePickerInput`: `onBlur` handler no longer fires `handleDateInput` when disabled or when value hasn't changed, which was causing ghost date changes on disabled/unchanged pickers

### Removed
- **Delivered date picker** removed from toolbar and GraphQL input — superseded by received date. Delivered is auto-adjusted when received is backdated before it.
- **Delivered date validation** removed from backend — no longer user-settable

### Differences from issue
- **No line deletion**: The issue describes deleting lines when changing the date. Instead, this implementation only allows moving the date **earlier** (never forward), which avoids ledger problems without needing to delete anything. Location movement timestamps are updated instead.
- **"Cannot set a date after stock has been distributed"** can't happen with one-way movement.
[...]

PR #11183, opened 2026-04-08
Fix DateTimePicker firing mutation on intermediate year/month selection

### LINKED ISSUE #11179: DateTimePicker fires mutation on intermediate year/month selection
## What went wrong? 😲

Clicking to the right of the **year number** in the delivered date picker on an inbound shipment (external) triggers a `CannotPutDeliveredDateAfterReceivedDate` server error. Clicking the calendar icon or a day number works fine.

### Suspected Root cause

MUI's `DateTimePicker` with `views: ['year', 'month', 'day']` fires `onChange` on **each intermediate view selection**. When you click a year:

1. MUI fires `onChange` with a date that has the new year but the same month/day (e.g. `2026-01-15`)
2. `DateTimePickerInput` propagates this immediately to the parent `onChange`
3. In `Toolbar.tsx`, the `onChange` handler calls `update()` which fires a GraphQL mutation with the intermediate date
4. The server validates `delivered_datetime <= received_datetime` and rejects the intermediate value

This affects any `DateTimePickerInput` usage where `onChange` triggers a mutation — not just the delivered date field.

### Relevant code

- `client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx` — the shared component; `onChange` fires on every intermediate selection (line 107)
- `client/packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx` — delivered date field calls `update()` directly in `onChange` (line 141)
- `server/service/src/invoice/inbound_shipment/update/validate.rs` — server-side validation (line 80)

### Suggested fix

Use MUI's `onAccept` callback (which only fires when the user completes the full selection) instead of `onChange` for propagating the value to the parent, or defer the external `onChange` until the picker closes. There is already a similar pattern in `ExpiryDateInput` which tracks `pickerOpen` state.

## How to Reproduce 🔨

1. Go to an inbound shipment (external) that has a received date
2. Click to the right of the **year number** in the delivered date picker
3. See `CannotPutDeliveredDateAfterReceivedDate` error

Image

## Expected behaviour 🤔

Selecting a year in the date picker should not trigger a server mutation until the full date (year + month + day) has been selected.

Identified on develop on 8th APril

esmehm:
Fixes #11179

# 👩🏻‍💻 What does this PR do?

Prevents `DateTimePickerInput` from propagating intermediate date values to the parent `onChange` when the user is still navigating views in the picker popup (e.g. clicking a year before selecting month and day).

MUI's `DateTimePicker` fires `onChange` on each view transition. Previously this propagated immediately, causing mutations with incomplete dates — triggering server validation errors like `CannotPutDeliveredDateAfterReceivedDate`.

Now:
- When the picker is open, `onChange` only updates internal state (`setValue`)
- The final value is propagated via MUI's `onAccept` callback, which fires only when the full selection is complete
- Keyboard input continues to propagate immediately as before
- Validation error state is properly cleared on accept, so stale errors from prior keyboard input don't persist

This fixes the issue at the `DateTimePickerInput` component level, so all ~25 consumers benefit — not just the inbound shipment delivered date field.

The `ExpiryDateInput` wrapper already tracked `pickerOpen` via the `setIsOpen` prop for its own logic. That continues to work — `onAccept` fires before `onClose`, so `ExpiryDateInput`'s `pickerOpen` ref is still `true` when its `onChange` callback runs from `onAccept`.

# 🧪 Testing

- [ ] Open an inbound shipment (external) that has a received date
- [ ] Click to the right of the year number in the delivered date picker
- [ ] Confirm no server error fires — picker navigates to month view
- [ ] Complete the full date selection (year → month → day)
- [ ] Confirm the delivered date is updated correctly
- [ ] Test keyboard input still works (type a date directly, confirm it propagates)
- [ ] Test ExpiryDateInput (e.g. in stock line edit) still sets last day of month when using picker

# 📃 Documentation

### DIFF
--- a/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 import {
   DateTimePicker,
   DateTimePickerProps,
@@ -72,6 +72,7 @@ export const DateTimePickerInput = ({
   const [internalError, setInternalError] = useState<string | null>(null);
   const [value, setValue] = useBufferState<Date | null>(props.value ?? null);
   const [isInitialEntry, setIsInitialEntry] = useState(true);
+  const pickerOpen = useRef(false);
   const t = useTranslation();
   const format =
     props.format === undefined ? (showTime ? 'P p' : 'P') : props.format;
@@ -105,6 +106,15 @@ export const DateTimePickerInput = ({
       <DateTimePicker
         format={format}
         onChange={(date, context) => {
+          // When the picker is open, intermediate view selections (e.g.
+          // clicking a year) fire onChange before the user has finished
+          // selecting. Only update internal state here; the final value is
+          // propagated via onAccept.
+          if (pickerOpen.current) {
+            setValue(date);
+            return;
+          }
+
           const { validationError } = context;
 
           if (validationError) {
@@ -117,11 +127,9 @@ export const DateTimePickerInput = ({
             setValue(date);
             return;
           }
-          if (!validationError) {
-            setIsInitialEntry(false);
-            setInternalError(null);
-          }
 
+          setIsInitialEntry(false);
+          setInternalError(null);
           handleDateInput(date);
         }}
         label={label}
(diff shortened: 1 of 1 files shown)


ria8651:
I'm not sure if I can reproduce the original issue, I always get the error when setting the delivered date to the received date just with this PR the error only happens after I close the picker. It won't let me set a date after the received date even on develop. I think moving the date when selecting the month is probably a bad idea but thats a MUI datetime picker behaviour issue that this PR doesn't cover. I don't think we should just delegate the update till the picker closes. For the specific broken delivered date picking this is replaced in #11168.

ria8651:
This should be covered by #11200, can probably close unless there was something else you wanted to add

ria8651:
Closing as #11200 is merged

PR #11200, opened 2026-04-09
#11197 Date picker: navigate on month/year select

### LINKED ISSUE #11197: (date picker) Day updating when selecting month
Image

## What went wrong? 😲

When selecting a date from the date time picker the date will be immediately moved upon selecting a year/month from the year dropdown.

I think this is the wrong behaviour. In particular this came up with #11168 where you can only move the date backwards. Having the date movement triggered when selecting the month or year opens the conformation popup without being able to select the day.

## Expected behaviour 🤔

For the year/month selection to behave like the month arrow where it navigates without selecting the date. Will have to make sure the date is still updated properly when selecting the month in the expiry date picker (no day selection).

ria8651:
Fixes #11197

# 👩🏻‍💻 What does this PR do?

Prevents the date picker from updating the date value when selecting a month or year from the dropdown. Instead, month/year selections now just navigate the picker (like the month arrows do).

Pickers without a day view (e.g. ExpiryDateInput) still correctly set the date on month selection since month is their final selection step.

**How it works:** Tracks the current picker view via `onViewChange` + `useState`. When `onChange` fires, React's state batching means `currentView` still holds the *previous* view — so we can detect that the change came from a month/year navigation and suppress it.

- Only one file changed: `DateTimePickerInput.tsx`
- The `hasDayView` check ensures ExpiryDateInput (and JsonForms Date/DateTime with `noDay`/`monthOnly`) still commit on month selection since they pass `views={['year', 'month']}` which omits `'day'`
- The `date !== null` guard ensures the Clear button still works from any view
- No changes needed to ExpiryDateInput or any other consumers

# 🧪 Testing

- [ ] Open any date picker → click the month/year header → select a different month → should navigate to that month's day view **without** changing the date → select a day → date should update
- [ ] Open an expiry date picker → select a month → date should update to last day of that month (existing behaviour preserved)
- [ ] Type a date via keyboard in any date picker → should update immediately as before
- [ ] Open a date picker, navigate to a different month, then click outside to dismiss → date should not change
- [ ] Click the Clear button while viewing month/year selection → should clear the date

# 📃 Documentation

### DIFF
--- a/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
+++ b/client/packages/common/src/ui/components/inputs/DateTimePickers/DateTimePickerInput/DateTimePickerInput.tsx
@@ -72,9 +72,14 @@ export const DateTimePickerInput = ({
   const [internalError, setInternalError] = useState<string | null>(null);
   const [value, setValue] = useBufferState<Date | null>(props.value ?? null);
   const [isInitialEntry, setIsInitialEntry] = useState(true);
+  const [currentView, setCurrentView] = useState<string | null>(null);
   const t = useTranslation();
   const format =
     props.format === undefined ? (showTime ? 'P p' : 'P') : props.format;
+
+  // Month/year selections are intermediate only when a day view exists.
+  // Default views always include 'day'; only explicit overrides (e.g. ExpiryDateInput) omit it.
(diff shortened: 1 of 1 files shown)


esmehm (approved):
I hadn't noticed before that you cannot select a day when using the date picker for expiry but can change the day when typing - would love to understand this

But the code changes here are a good improvement and look good across multiple places! :-)

PR #11208, opened 2026-04-10
Fix Status History dates in timezones ahead of UTC

### LINKED ISSUE #11206: Status History dates display incorrectly in timezones ahead of UTC
## Description

Asset log dates on the Status History tab (and potentially other places that display `log_datetime`) show the wrong date for users in timezones ahead of UTC.

For example, a status update created at 11:30am NZST (UTC+12) is stored as `2026-04-09T23:30:00` (UTC). When `localisedDate` parses this with `parseISO`, it treats it as local time (no `Z` suffix), so it displays as 09/04/2026 instead of 10/04/2026.

## Steps to reproduce

1. Be in a timezone ahead of UTC (e.g. NZST, UTC+12)
2. Create or update an asset status between midnight UTC and midnight local time — in NZ (currently NZST, UTC+12) this means **before noon NZST**, since that period maps to the previous day in UTC
3. Open the asset's Status History tab
4. The log entry date shows **yesterday** instead of today

## Root cause

`log_datetime` is a `NaiveDateTime` stored in UTC. The frontend's `parseISO` (from date-fns) treats strings without timezone info as local time, not UTC.

## Recommended fix

In `StatusLogs.tsx`, append `Z` to the datetime string before passing to `localisedDate` so it's parsed as UTC:

```tsx
// Before
{localisedDate(log.logDatetime)}

// After
{localisedDate(log.logDatetime + 'Z')}
```

This is a minimal fix scoped to Status History. A broader fix could be applied to `dateInputHandler` in `DateUtils.ts` or to all `NaiveDateTime` fields returned from the server, but that would need wider testing.

## Affected area

`client/packages/coldchain/src/Equipment/DetailView/Tabs/StatusLogs.tsx`

This is a pre-existing issue, not introduced by #10845.

esmehm:
Closes #11206 

## Summary
- The Status History tab on assets shows incorrect dates for users in timezones ahead of UTC (e.g. a status update made on April 10 NZST shows as April 9)
- Root cause: `NaiveDateTime` fields from the server (e.g. `log_datetime`) have no timezone suffix, so `parseISO` treats them as local time instead of UTC
- This PR implements both a backend fix (root cause) and a frontend safety net

### Backend fix
Converts the 3 remaining GraphQL files that still returned raw `NaiveDateTime` to use `DateTime<Utc>`, matching the pattern used throughout the rest of the codebase:
- `asset_log.rs` — `log_datetime`
- `asset.rs` — `created_datetime`, `modified_datetime`
- `sync_file_reference.rs` — `created_datetime`

This ensures the server always includes a `Z` suffix on these datetime fields.

### Frontend safety net
Adds a guard in `dateInputHandler` (`DateUtils.ts`) that detects datetime strings with a time component (`T`) but no timezone suffix and appends `Z` before parsing. This protects against any future `NaiveDateTime` fields that might be missed on the backend, and covers all ~20 formatting call sites in a single change.

### Affected areas
- **coldchain**: asset status logs, asset summary, mobile log cards, temperature breach cells/modals, breach popovers, notifications
- **invoices**: inbound/outbound shipment side panels, prescription details
- **requisitions**: side bar components, RnR form side panel
- **system**: activity log list, item ledger, encounter detail/side panel, vaccination modal, name details
- **inventory**: stocktake side panel
- **programs**: encounter line chart, historic encounter data, program events
- **common**: column defaults for date/datetime columns, StatusCrumbs

## Test plan
- [ ] Set system timezone to UTC+12 (e.g. NZST)
- [ ] Create or update an asset status before noon local time (so the UTC date is the previous day)
- [ ] Open the asset's Status History tab and verify the date shows **today**, not yesterday
- [ ] Verify date-only fields (e.g. date of birth, vaccination date) still display correctly
- [ ] Verify datetimes that already have `Z` suffix (e.g. invoice created dates) still display correctly

### DIFF
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -53,7 +53,20 @@ const URL_QUERY_DATE_TIME = 'yyyy-MM-dd HH:mm';
 
 const dateInputHandler = (date: Date | string | number): Date => {
   // Assume a string is an ISO date-time string
-  if (typeof date === 'string') return parseISO(date);
+  if (typeof date === 'string') {
+    // NaiveDateTime strings from the server have no timezone suffix but are
+    // stored as UTC. Without the 'Z', parseISO treats them as local time,
+    // which shifts dates in timezones ahead of UTC. Appending 'Z' ensures
+    // they are parsed as UTC. Date-only strings (no 'T') are left as-is.
+    const tIndex = date.indexOf('T');
+    if (
+      tIndex !== -1 &&
+      !date.endsWith('Z') &&
+      !/[+-]/.test(date.substring(tIndex))
+    )
+      return parseISO(date + 'Z');
+    return parseISO(date);
+  }
   // Assume a number is a UNIX timestamp
   if (typeof date === 'number') return fromUnixTime(date);
   return date as Date;
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -57,14 +57,11 @@ const dateInputHandler = (date: Date | string | number): Date => {
     // Any dates we receive from the server without timezone information are
     // assumed to be in UTC time
     const tIndex = date.indexOf('T');
-    if (
+    const needsUtcSuffix =
       tIndex !== -1 &&
       !date.endsWith('Z') &&
-      !/[+-]/.test(date.substring(tIndex))
-    ) {
-      return parseISO(date + 'Z');
-    }
-    return parseISO(date);
+      !/[+-]/.test(date.substring(tIndex));
+    return parseISO(needsUtcSuffix ? date + 'Z' : date);
   }
   // Assume a number is a UNIX timestamp
   if (typeof date === 'number') return fromUnixTime(date);
--- a/client/packages/common/src/intl/utils/DateUtils.ts
+++ b/client/packages/common/src/intl/utils/DateUtils.ts
@@ -54,17 +54,16 @@ const URL_QUERY_DATE_TIME = 'yyyy-MM-dd HH:mm';
 const dateInputHandler = (date: Date | string | number): Date => {
   // Assume a string is an ISO date-time string
   if (typeof date === 'string') {
-    // NaiveDateTime strings from the server have no timezone suffix but are
-    // stored as UTC. Without the 'Z', parseISO treats them as local time,
-    // which shifts dates in timezones ahead of UTC. Appending 'Z' ensures
-    // they are parsed as UTC. Date-only strings (no 'T') are left as-is.
+    // Any dates we receive from the server without timezone information are
+    // assumed to be in UTC time
     const tIndex = date.indexOf('T');
     if (
       tIndex !== -1 &&
       !date.endsWith('Z') &&
       !/[+-]/.test(date.substring(tIndex))
-    )
+    ) {
       return parseISO(date + 'Z');
+    }
     return parseISO(date);
   }
   // Assume a number is a UNIX timestamp
--- a/client/packages/coldchain/src/Equipment/DetailView/Tabs/StatusLogs.tsx
+++ b/client/packages/coldchain/src/Equipment/DetailView/Tabs/StatusLogs.tsx
@@ -174,7 +174,7 @@ const StatusLog = ({
                 lineHeight: 2,
               })}
             >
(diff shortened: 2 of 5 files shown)


jmbrunskill (dismissed):
I haven't looked too deep into this but I think we normally expose the graphql from the backend something like this, which should mean it gets converted properly.

```
    pub async fn datetime(&self) -> DateTime<Utc> {
        DateTime::<Utc>::from_naive_utc_and_offset(self.row().datetime, Utc)
    }
```

Can you check if this is correct from the backend side?

esmehm (commented):
I agree that most of the codebase already does this

```rust
pub async fn datetime(&self) -> DateTime<Utc> {
    DateTime::<Utc>::from_naive_utc_and_offset(self.row().datetime, Utc)
}
```

Only 3 files still return raw `NaiveDateTime`:
- `server/graphql/asset/src/types/asset_log.rs` — `log_datetime`
- `server/graphql/asset/src/types/asset.rs` — `created_datetime`, `modified_datetime`
- `server/graphql/types/src/types/sync_file_reference.rs` — `created_datetime`

Converting these to `DateTime<Utc>` (Option C in the PR description) would fix the root cause. But I was unsure about the breaking change because the `NaiveDateTime` scalar is also used as a GraphQL **input** type (`NullableDatetimeUpdate.value`)

The frontend fix here is forward-compatible with that backend change — once those 3 files return `DateTime<Utc>` (with `Z`), the `!date.endsWith('Z')` guard in `dateInputHandler` makes the frontend fix a no-op.

But do you prefer the back end fix from the get go?

Also considering how we prevent this from reoccuring

client/packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -53,7 +53,20 @@ const URL_QUERY_DATE_TIME = 'yyyy-MM-dd HH:mm';
...
+    // stored as UTC. Without the 'Z', parseISO treats them as local time,
+    // which shifts dates in timezones ahead of UTC. Appending 'Z' ensures
+    // they are parsed as UTC. Date-only strings (no 'T') are left as-is.
+    const tIndex = date.indexOf('T');
+    if (
+      tIndex !== -1 &&
+      !date.endsWith('Z') &&
+      !/[+-]/.test(date.substring(tIndex))
+    )
+      return parseISO(date + 'Z');
```
jmbrunskill: I'd prefer to have braces here, and possibly just the one return? Just from a code readability.

I'd add the to comment something like this.

`Any dates we receive from the server without timezone information are assumed to be in UTC time` 
Just to clarify the reasoning behind it.

jmbrunskill:
> I agree that most of the codebase already does this
> 
> ```rust
> pub async fn datetime(&self) -> DateTime<Utc> {
>     DateTime::<Utc>::from_naive_utc_and_offset(self.row().datetime, Utc)
> }
> ```
> 
> Only 3 files still return raw `NaiveDateTime`:
> 
> * `server/graphql/asset/src/types/asset_log.rs` — `log_datetime`
> * `server/graphql/asset/src/types/asset.rs` — `created_datetime`, `modified_datetime`
> * `server/graphql/types/src/types/sync_file_reference.rs` — `created_datetime`
> 
> Converting these to `DateTime<Utc>` (Option C in the PR description) would fix the root cause. But I was unsure about the breaking change because the `NaiveDateTime` scalar is also used as a GraphQL **input** type (`NullableDatetimeUpdate.value`)
> 
> The frontend fix here is forward-compatible with that backend change — once those 3 files return `DateTime<Utc>` (with `Z`), the `!date.endsWith('Z')` guard in `dateInputHandler` makes the frontend fix a no-op.
> 
> But do you prefer the back end fix from the get go?
> 
> Also considering how we prevent this from reoccuring

I think I prefer the backend fix, but I agree we can do frontend as well.
It is a mild breaking change as I'm pretty sure we have no integrations using assets or sync_file_references and I see it as a fix of the API rather than breaking change. Hopefully any consumers would corectly handle the updated format.

These are both things I added and think I just missed the pattern, I think if we keep our code base consistent and always return DateTime<Utc> AI and future developers are much more likely to copy the correct pattern. That's my hope anyway.

Sorry I think I missed your options considered section originally? I think lets do both, fix the API so timezone is returned and also add the safety net?

esmehm:
@jmbrunskill have upodated the BE now too - thanks for the feedback! Totally agree others are likely to follow suit now

jmbrunskill (approved):
Thanks Esme happy with this now!

Issue #10294, opened 2026-01-21
Sync: keep Vecs of data small


Chris-Petty:
## Is your feature request related to a problem? Please describe 👀

In the context of integration, where we are processing everything in the sync_buffer which may be millions of records, we should ideally iterate through subsets of those records least we run out of memory on devices.

The integration function starts out with this:
https://github.com/msupply-foundation/open-msupply/blob/42306d03f045f4110929fb6e5c06e2be08f2950d/server/service/src/sync/synchroniser.rs#L373-L377

That method essentially gets all records in table topological order into one (giant) Vec:
https://github.com/msupply-foundation/open-msupply/blob/42306d03f045f4110929fb6e5c06e2be08f2950d/server/service/src/sync/sync_buffer.rs#L49-L85

If a buffer had say 1 million records, at roughly 1KB/record, that's a gig of RAM right there. Also lot to pull out of sqlite or pg in one go. (We could play fast and loose with our code if we had an [effective DB caching layer, perhaps with an ORM built over top](https://developer.4d.com/docs/19/ORDA/overview)... 😁)

In `translate_and_integrate_sync_records` we then iterate over them all, there are little Vecs of operations used there but the scale is fine!

## Describe the solution you'd like 🎁

Rewrite the process loop to get say 5000 records per iteration from each table in topologic order and integrate those 5000. Then the Vec is only ~5MB of data not ~1GB if following my earlier logic.

Chris-Petty:
@andreievg I think I should do this one today, using waaaaayyyy too much memory on integration of 8 million records in sync buffer:

<img width="1500" height="177" alt="Image" src="https://github.com/user-attachments/assets/e7a42619-106d-449d-90c0-6016227c00a5" />

<img width="258" height="456" alt="Image" src="https://github.com/user-attachments/assets/11061c83-bce1-4eec-b37e-1321e365142d" />

Interestingly seems kind of separate from performance on my m4 pro. When it finally got started I was getting integration speeds as low as 20/s on name_store_join (has to do a query on `name` and a query on `store`) but nearly 4000/s for requisitions.

andreievg:
Please 🙏

jmbrunskill:
There was a similar/related idea that we can actually do transactions at a per table level, not sure if that's been implemented in sync v7 but we should be able to commit say all the new names before we even read the list of invoices to commit.

andreievg:
[...]

PR #11211, opened 2026-04-10
#11050 #11109 Goods received sync translators

### LINKED ISSUE #11050: When migrating a OG store to omSupply only the Goods Receipt that have been finalised are migrated
## What went wrong? 😲
It seems that the goods Receipt that have not been finalised are not migrated.

I had 2 Goods Receipt for the same PO. One was finalised (so a supplier invoice was created automatically for it) and one was not finalised (it didn't have a supplier invoice created automatically for it)

Purchase order

Image

First Goods receipt

Image

Second Goods receipt

Image

Only the first one has been migrated

Image

ALSO the PO number related to the Goods receipt is not displayed in the "More" section. Following the example above, it should be Purchase order 2

Image

## Expected behaviour 🤔
The Goods receipt not finalised should also be migrated.

## How to Reproduce 🔨

1. Have a Store on OG that have Goods receipts that have not been finalised ie The status is "nw"
2. Migrate that store to OMS
3. See error. These Goods Receipts have not been migrated

## Your environment 🌱

- Open mSupply Version: 2.17.00-RC-03270721
- Legacy mSupply Central Server Version: v9.00RC5

- Platform: omSupply Standalone

- Database type: SQLite

### LINKED ISSUE #11109: Extra tabs doesn't appear for External Inbound shipment that have been migrated from OG to omSupply
## What went wrong? 😲
 
External Inbound shipment created from Goods receipt following a migration to OMS of a store from OG are missing the extra tabs that an External Inbound shipment should have. It has only the tab of a normal Inbound shipment

External Inbound shipment created from migration of Goods Receipts

Image

On creating a new External Inbound shipment manually

Image

## Expected behaviour 🤔
Migrated Goods receipts should show the same tabs than those created manually

## How to Reproduce 🔨

On OG log to a store that has finalised goods receipts
1. Migrate that store to an OMS site
2. Click on 'Replenishment > Inbound shipment'
3. Open that inbound shipment that was created

## Additional Info
Found while testing #11050 
## Your environment 🌱

- Open mSupply Version: 

- Platform: 2.17.00-RC-04011344

- Database type: SQLite

ria8651:
> [!WARNING]

Fixes #11050
Fixes #11109

# 👩🏻‍💻 What does this PR do?

Replaces the one-shot `import_goods_received` migration with proper sync translators for `Goods_received` and `Goods_received_line` tables. This fixes two issues:

1. **Non-finalized GRs not migrated** (#11050): Previously only GRs in the sync buffer at migration time were processed. Now they're handled by translators that run on every sync cycle, so GRs arriving after the upgrade are also processed.

2. **Missing PO link on migrated GRs** (#11109): The old migration created duplicate invoices. The new translators correctly handle both cases:
   - **Finalized GRs**: Find the existing supplier invoice (via `goods_received_ID` in sync buffer) and set `purchase_order_id`. Find existing invoice lines (via `goods_received_lines_ID`) and set `purchase_order_line_id`.
   - **Non-finalized GRs**: Create new InboundShipment invoices with lines, with PO links set from the start.

### Changes
- **New**: `goods_received.rs` sync translator (pull-only, no push)
- **New**: `goods_received_line.rs` sync translator (pull-only, no push)
- **New**: `SyncBufferRowRepository::find_by_table_and_data_like()` for efficient sync buffer lookups
- **New**: v2.17.3 migration to reintegrate GR sync buffer records (skipped if old migration already ran)
- **Removed**: `import_goods_received.rs` migration
- **Added**: Unit tests for both translators

- The translators are **pull-only** (no `change_log_type`, no push methods). GR data is never synced back to mSupply. The `is_sync_update` flag on changelogs from upserted invoices prevents re-push.
- For finalized GRs, the link from GR to supplier invoice goes through the sync buffer: transact records have `goods_received_ID` field, and trans_line records have `goods_received_lines_ID` field. We use SQL `LIKE` queries on the `data` column with JSON verification to find these efficiently.
- The v2.17.3 reintegration migration checks `migration_fragment_log` for the old `import_goods_received` entry. If found, it skips — those sites will need separate cleanup of duplicate invoices from the old migration.

# 🧪 Testing

- [ ] OG central server with a store that has both finalized and non-finalized Goods Receipts
- [ ] Migrate that store to an OMS site running this PR
- [ ] Verify finalized GRs: the existing supplier invoice should show as External Inbound Shipment with PO link and extra tabs (Financial, Currency, Delivery). No duplicate invoice should be created.
- [ ] Verify non-finalized GRs: a new InboundShipment should be created with PO link and lines
- [ ] Verify PO line links: invoice lines should have `purchase_order_line_id` set
- [ ] Verify nothing syncs back to mSupply from the GR-created records

## My setup:

Purchase order:
Screenshot 2026-04-10 at 4 49 44 pm

Two Goods Received - one finalised one not:
Screenshot 2026-04-10 at 4 50 08 pmScreenshot 2026-04-10 at 4 50 18 pm

One invoice made from the finalised GR:
Screenshot 2026-04-10 at 4 50 30 pm

## Previous migration:

Ends up with duplicate invoices from the migration and the invoice mSupply created for the finalised GR:
Screenshot 2026-04-10 at 1 18 10 pm
Also the invoice lines aren't linked to the PO lines.

## This PR:
Screenshot 2026-04-10 at 4 54 41 pm

# 📃 Documentation

Need some kind of information on what is migrated when and what do do if the v2.17 migration has already ran...


server/service/src/sync/test/test_data/goods_received_line.rs
```diff
@@ -0,0 +1,68 @@
...
+            action: SyncAction::Upsert,
+            ..Default::default()
+        }],
+        ..Default::default()
+    });
+    record
+}
+
+pub(crate) fn test_pull_upsert_records() -> Vec<TestSyncIncomingRecord> {
+    vec![gr_line_non_finalised_pull_record()]
```
ria8651: 🟡 **Should fix — Missing test for finalized GR line path**

Only the non-finalized case is tested here. The finalized path (finding an existing invoice line via sync buffer lookup and updating its `purchase_order_line_id`) has no test coverage. This is the more complex and error-prone path — would be good to add a `gr_line_finalised_pull_record()` test case similar to how `goods_received.rs` tests both paths.

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,233 @@
...
+            .find_one_by_record_id(&data.goods_received_ID)?
+        {
+            Some(row) => row,
+            None => {
+                return Ok(PullTranslateResult::Ignored(format!(
+                    "parent goods_received {} not found in sync_buffer",
+                    data.goods_received_ID
+                )))
+            }
+        };
```
ria8651, line 141: 🟡 **Should fix — No table_name verification on sync buffer lookup**

The parent GR is looked up by `record_id` alone. While `record_id` is the sync_buffer PK (so unique), the code then parses the result as `GoodsReceivedStatus` without verifying `table_name == "Goods_received"`. If a record from a different table happened to share the same ID (unlikely with UUIDs but not impossible), parsing would silently extract a wrong/default status.

Suggestion:
```rust
let gr_sync_row = match SyncBufferRowRepository::new(connection)
    .find_one_by_record_id(&data.goods_received_ID)?
{
    Some(row) if row.table_name == "Goods_received" => row,
    Some(_) => {
        return Ok(PullTranslateResult::Ignored(format!(
            "sync_buffer record {} is not a Goods_received record",
            data.goods_received_ID
        )))
    }
    None => ...
};
```

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,233 @@
...
+#[derive(Deserialize)]
+struct TransLineGoodsReceivedLineId {
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_str_as_option_string")]
+    goods_received_lines_ID: Option<String>,
+}
+
+fn is_finalised(status: &str) -> bool {
+    matches!(status, "fn" | "Fin" | "Finalised")
+}
```
ria8651: 🟢 **Nit — Duplicated `is_finalised` function**

This same function exists in `goods_received.rs:76-78`. Could share it from the `goods_received` module (which is already `pub(super)`) to avoid duplication.

server/service/src/sync/translations/goods_received.rs
```diff
@@ -0,0 +1,215 @@
...
+    goods_received_ID: Option<String>,
+}
+
+/// Find the invoice ID of the supplier invoice that was created from a finalized GR,
+/// by searching transact sync_buffer records for one with goods_received_ID matching the GR's ID.
+fn find_linked_invoice_id(
+    connection: &StorageConnection,
+    goods_received_id: &str,
+) -> Result<Option<String>, anyhow::Error> {
+    let pattern = format!("%\"goods_received_ID\"%\"{goods_received_id}\"%");
```
ria8651, line 51: 🟢 **Nit — LIKE pattern with leading `%` won't use indexes**

This will cause a full scan of all `transact` records in sync_buffer. Acceptable for sync (not a hot path), but could be slow on sites with very large sync buffers. Same applies to the `trans_line` lookup in `goods_received_line.rs`.

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,233 @@
...
+            invoice_id: data.goods_received_ID,
+            item_link_id: item_link.id.clone(),
+            item_name: data.item_name,
+            item_code,
+            location_id: data.location_ID,
+            batch: data.batch_received,
+            expiry_date: data.expiry_date,
+            pack_size: data.pack_received,
+            cost_price_per_pack: data.cost_price,
+            sell_price_per_pack: data.cost_price,
```
ria8651, line 199: 🟢 **Nit** — `sell_price_per_pack` set to `cost_price`. This mirrors the old migration behavior, but sell price = cost price may not be accurate for all scenarios. Minor concern.

ria8651 (commented):
Claude review bellow:

## Review: #11050 #11109 Goods received sync translators

### Requirements Coverage

**Issue #11050 — Non-finalized GRs not migrated:**
- ✅ Non-finalized GRs now create new InboundShipment invoices via `GoodsReceivedTranslation`, running on every sync cycle instead of only at migration time.
- ✅ The v2.17.1 reintegration migration resets `integration_datetime` so existing GR sync buffer records are reprocessed by the new translators.

**Issue #11050 — PO link missing on migrated GRs:**
- ✅ `purchase_order_id` is set on all invoices created/updated from GRs.

**Issue #11109 — Extra tabs missing (because PO link missing):**
- ✅ For finalized GRs, the existing supplier invoice is found via sync buffer lookup and updated with the PO link.
- ✅ PO line links are set on invoice lines via the `GoodsReceivedLineTranslation` translator.

**Issue #11050 — Works for mSupply → 2.17 (not just 2.16 → 2.17):**
- ✅ By using sync translators instead of a migration, this handles GRs arriving via sync at any time, including direct mSupply → 2.17 scenarios.

### Changes Not in Issue
- **`SyncBufferRowRepository::find_by_table_and_data_like()`** — New repository method for LIKE-based JSON field searches in sync_buffer. Necessary infrastructure for the translator approach.
- **Reintegration migration skip logic** — Checks `migration_fragment_log` for old migration, skips if found. Prudent safety measure for sites that ran the old migration.

### Summary

The approach is sound — replacing a one-shot migration with sync translators is the right fix for both issues. The code follows existing sync translator patterns, pull dependencies are correctly declared, and the `is_sync_update` mechanism properly prevents re-push of synced records.

The main gap is **missing test coverage for the finalized GR line path**. This is the more complex code path involving sync buffer lookups and existing record mutation. I'd recommend adding a test case before merging.

The other findings are lower severity and could be addressed in follow-up. Overall close to merge-ready once the test gap is addressed.

server/repository/src/migrations/v2_17_03/mod.rs
```diff
@@ -0,0 +1,44 @@
+use super::{version::Version, Migration, MigrationFragment};
```
esmehm, line 1: File dir needs updating to 2_17_02 and related version changes in here and in package.json

server/repository/src/migrations/mod.rs
```diff
@@ -146,6 +147,7 @@ pub fn migrate(
         Box::new(v2_15_00::V2_15_00),
         Box::new(v2_16_00::V2_16_00),
         Box::new(v2_17_00::V2_17_00),
+        Box::new(v2_17_01::V2_17_01),
```
esmehm: Also 02 not 01 now

server/service/src/sync/translations/goods_received_line.rs
esmehm: I think this is valid feedback - please consider it

server/service/src/sync/test/test_data/goods_received_line.rs
esmehm: Because the finalised case is relatively simple, this is not blocking

esmehm (changes requested):
Looks good otherwise - handles the case of migrating finalised IS(E) well but does not handle the case of having IS(E) in new state when migrated and then changing to finalsied on OG and then syncing (it stays as new) - but I assume this isnt a real-life scenario as once a site moves to OMS it never goes back to OG?

ria8651:
Yea modifying the GR isn't supported by design. Though if you wanted to do what you're describing you'd have to migrate to oms, back to og then back to oms so not too realistic anyway (can't modify oms owned data on og).

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "open-msupply",
   "//": "Main version for the app, should be in semantic version format (any release candidate or test build should be separated by '-' i.e. 1.1.1-rc1 or 1.1.1-test",
-  "version": "2.17.01-develop",
+  "version": "2.17.3-develop",
```
esmehm: I think they normally have a 0 in there  `2.17.03-develop` - suspect that;'s what the conflict is

esmehm (approved):
LGTM :)

package.json
ria8651: that's ugly though XD

server/service/src/sync/translations/goods_received.rs
jmbrunskill: I wonder if this will still be too slow with sync v7 integrating all records for every store. Something to test for I guess. @Chris-Petty

server/service/src/sync/translations/goods_received.rs
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+    }
+
+    Ok(None)
+}
+
+fn map_status(status: &str) -> InvoiceStatus {
+    match status {
+        "fn" | "Fin" | "Finalised" => InvoiceStatus::Verified,
```
jmbrunskill: Where did you see Fin and Finalised?

server/service/src/sync/translations/goods_received.rs
```diff
@@ -0,0 +1,230 @@
...
+            });
+
+        let invoice = InvoiceRow {
+            id: data.id,
+            name_id: po.supplier_name_id,
+            store_id: data.store_ID,
+            user_id: data.user_id_created,
+            invoice_number: data.serial_number,
+            r#type: InvoiceType::InboundShipment,
+            status: map_status(&data.status),
```
jmbrunskill: Maybe just set this to `New` the map status thing kind of hides what's actually going on? I was worried about what statuses might get mapped to that might affect ledger, hard coding to New is the safe option here. The code's correct it's just not so obvious...

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,252 @@
...
+impl SyncTranslation for GoodsReceivedLineTranslation {
+    fn table_name(&self) -> &str {
+        "Goods_received_line"
+    }
+
+    fn pull_dependencies(&self) -> Vec<&str> {
+        vec![
+            GoodsReceivedTranslation.table_name(),
+            InvoiceLineTranslation.table_name(),
+            ItemTranslation.table_name(),
```
jmbrunskill, line 97: Pretty srure this is here transatively from invoice_line but doens't hurt

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            return Ok(PullTranslateResult::Ignored(format!(
+                "invoice_line {} already exists, skipping goods_received_line import",
+                data.id
+            )));
+        }
+
+        // Look up parent GR to check if finalized
+        let gr_sync_row = match SyncBufferRowRepository::new(connection)
+            .find_one_by_record_id(&data.goods_received_ID)?
```
jmbrunskill, line 126: With sync v7 this will need to be find latest by record_id as we'll have a history of sync_buffers. Unlikely to have multple in this case but thought worth mentioning

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,252 @@
...
+                None => Ok(PullTranslateResult::Ignored(format!(
+                    "invoice_line {invoice_line_id} not found for goods_received_line {}",
+                    data.id
+                ))),
+            };
+        }
+
+        // Non-finalized GR: create invoice line
+        let item_links =
+            ItemLinkRowRepository::new(connection).find_many_by_item_id(&data.item_ID)?;
```
jmbrunskill: I don't think you need to do this, you can just put the item_id in and skip the extra look up...

server/service/src/sync/translations/goods_received_line.rs
```diff
@@ -0,0 +1,252 @@
...
+            cost_price_per_pack: data.cost_price,
+            sell_price_per_pack: data.cost_price,
+            total_before_tax: total,
+            total_after_tax: total,
+            r#type: InvoiceLineType::StockIn,
+            number_of_packs: data.quantity_received,
+            note: data.comment,
+            volume_per_pack: data.volume_per_pack,
+            purchase_order_line_id: data.order_line_ID,
+            ..Default::default()
```
jmbrunskill: I avoid doing `..Default::default()` apart from in tests. It's ok in this case I think as I don't think we'll come back to this, but in general I think it's clearer to see what exactly is being mapped to None for example, and more importantly if we change the shape of invoice, we'd want to get the compiler to make as consider if this case needs something special. As I said for this 1 off style translator I think it's fine but wanted to mention...

jmbrunskill (approved):
I've done a read through of the code and I'm happy with the approach.
A little concerned about how the performance might be fore sync v7 which might have a huge number of records in sync_buffer. Haven't tried to think of an alternative approach but I think we're good to go with this one. For 2.17

server/service/src/sync/translations/goods_received.rs
```diff
@@ -180,22 +173,45 @@ impl SyncTranslation for GoodsReceivedTranslation {
...
             on_hold: false,
             comment: data.comment,
             their_reference: data.supplier_reference,
+            transport_reference: None,
             created_datetime,
+            allocated_datetime: None,
+            picked_datetime: None,
+            shipped_datetime: None,
+            delivered_datetime: None,
             received_datetime: data.received_date.and_then(|d| d.and_hms_opt(0, 0, 0)),
```
jmbrunskill: DANG, this should be delivered_datetime I think, recieved is the stock movement date. We're putting it in as New right so shouldn't affect the ledger...
```
               CREATE VIEW stock_movement AS
    WITH all_movements AS (
      SELECT
        invoice_line_stock_movement.id AS id,
        quantity_movement AS quantity,
        invoice_line_stock_movement.item_id AS item_id,
        invoice.store_id as store_id,
        CASE WHEN invoice.type IN (
            'OUTBOUND_SHIPMENT', 'SUPPLIER_RETURN',
            'PRESCRIPTION'
        ) THEN picked_datetime
                    WHEN invoice.type IN (
            'INBOUND_SHIPMENT', 'CUSTOMER_RETURN'
        ) THEN received_datetime
                    WHEN invoice.type IN (
            'INVENTORY_ADDITION', 'INVENTORY_REDUCTION', 'REPACK'
        ) THEN verified_datetime
            END AS datetime,
```

jmbrunskill: Easy mistake to make, and a good reason to have all the fields showing as it made me think of this!

jmbrunskill (changes requested):
Change received_datetime to delivered_datetime please

ria8651:
Damn yea, good catch.

jmbrunskill (approved):
Ok I'm happy now :)

Received date could be translated to delivered_datetime in OMS. Might be nice but happy to merge either way.

server/service/src/sync/translations/goods_received.rs
Chris-Petty: Yes this is pretty nasty haha, but possibly the best we can do at the moment? I don't know how you do this without a rewrite of our sync procedure. Because our procedure gets all of the records to integrate as `PullTranslateResult` structs, then does all the DB writes, it's not possible to query the actual `invoice` table with indexes for this even with dependency arrays known.

We'd have to do updates as we go through the sync_buffer so you could potentially query actual tables that are depended on, rather stepped based approach with potentially very large vectors. I raised an issue about this just from a RAM usage perspective in mind haha https://github.com/msupply-foundation/open-msupply/issues/10294

PR #11215, opened 2026-04-12
#11133 Fix null rounding in inbound shipment report

### LINKED ISSUE #11133: Inbound shipment forms aren't generating anymore
## What went wrong? 😲
Inbound shipment forms aren't generating anymore because of 

> DocGenerationError( "Tera rendering: Error { kind: Msg(\"Failed to render 'template.html'\"), source: Some(Error { kind: CallFilter(\"round\"),  source: Some(Error { kind: Msg(\"Filter `round` was called on an incorrect value: got `null` but expected a f64\"), source: None }) }) }", )

## Expected behaviour 🤔
Form should generate without error

## How to Reproduce 🔨

1. Go to Inbound shipment
2. click on any shipment
3. try to print report
4. see error

## Your environment 🌱

- Open mSupply Version: latest dev but it's probably in 2.17

- Platform:

- Database type:

ria8651:
Fixes #11133

# 👩🏻‍💻 What does this PR do?

Fixes inbound shipment report generation failing with a `DocGenerationError` when `costPricePerPack`, `packSize`, or pricing totals are null.

The Tera `round` filter expects a `f64` but was receiving `null` values, causing the template to fail to render. This adds `| default(value=0.0)` before all `round` calls to provide a safe fallback, and uses `{% set %}` variables to avoid operator precedence issues between arithmetic and filters.

- All `round` filter usages in the template are now guarded with `default(value=0.0)`
- The cost-per-unit calculation previously used an `{% if %}` / `{% else %}` block for null safety — this is replaced with `{% set %}` variables using `default` filters, which is more concise and consistent with the rest of the template
- `packSize` defaults to `1.0` (not `0.0`) to avoid division by zero
- The generated JSON files (`standard_forms.json`, `standard_reports.json`) are regenerated artifacts from the template change

# 🧪 Testing

- [ ] Open an inbound shipment that has lines with null `costPricePerPack` values
- [ ] Click print report — should generate without error
- [ ] Verify that null cost values display as `0.00` in the report
- [ ] Verify that lines with valid cost values still display correctly

# 📃 Documentation

- [x] **No documentation required**: bug fix which isn't a change in behaviour


roxy-dao (changes requested):
Thanks Brian~ Needs a version bump 😉 and can you bump to 2.17.2 for release since 2.17.1 was released for translations (:

standard_forms/inbound/latest/report-manifest.json
```diff
@@ -1,6 +1,6 @@
 {
   "is_custom": false,
-  "version": "2.6.6",
```
roxy-dao, line 3: Would just bump to 2.7.6 since it's a small bug fix

ria8651:
Right that makes sense, we want the update to apply to older sites.

roxy-dao (approved):
LGTM~ thanks Brian!

PR #11218, opened 2026-04-13
Add new indexes for changelog

### LINKED ISSUE #11122: v2.16.1 (13th Apr 2026)
Improve sync status reporting (reducing load on server and connections
Fix issue with item search
Fix deduplication performance (only de-dup required records)
Limit scope of the ledger fix code (reduce risk of creating ledger issues) + Reduce load checking for issues
(Possibly) reduce chance/possibility of logout on timeout
(Possibly) improve performance of views needed for dashboard and MFRC report
Review and add indexes to database as needed

## Key aims
- https://github.com/msupply-foundation/open-msupply/issues/11042
- https://github.com/msupply-foundation/open-msupply/issues/10914
- ~Backport subscriptions for sync status?~ Change log query speed up and reduced the polling time to 2s
- ~Backport item search fix~  - Just needs install on production
- [x] #11231

Dates are rough estimates for now.

## Deadlines (in NZT)
Date started: Tues 7th Apr 2026
PRs merged: Fri 10th Apr 2026
Release: Tues 14th Apr 2026

jmbrunskill:
Fixes #11122 changelog sync performance

# 👩🏻‍💻 What does this PR do?

One of the key area's we saw problems with when lots of users are connected was the postgres server was spending a lot of time trying to de-dup the changelog to display sync status.

I've benchmarked how long it use to take using a k6 script based on Brian's one here: https://github.com/msupply-foundation/open-msupply/compare/2.16.1-sl-hotfix

After a lot of headscratching I realised that the results we heavily skewed by user logins for every request, so modified it to 

No new indexes max(10 connections to postgres)  (Macbook m3)
```
    HTTP
    http_req_duration..............: avg=154.43ms min=3.44ms   med=143.95ms max=1.26s  p(90)=164.59ms p(95)=179.58ms
      { expected_response:true }...: avg=154.43ms min=3.44ms   med=143.95ms max=1.26s  p(90)=164.59ms p(95)=179.58ms
      { scenario:sync_status }.....: avg=156.71ms min=45.1ms   med=144.3ms  max=1.26s  p(90)=165.06ms p(95)=179.58ms
    http_req_failed................: 0.00%  0 out of 861
      { scenario:sync_status }.....: 0.00%  0 out of 841
    http_reqs......................: 861    9.878104/s
```

Added indexes (after extensive testing and discussion with AI):
CREATE INDEX IF NOT EXISTS index_changelog_record_id_store_id_cursor_desc
    ON changelog (record_id ASC, store_id ASC, cursor DESC);

CREATE INDEX IF NOT EXISTS index_changelog_cursor
    ON changelog (cursor DESC);

```
    HTTP
    http_req_duration..............: avg=87.96ms  min=3.19ms  med=76.03ms  max=939.62ms p(90)=97.47ms  p(95)=109.65ms
      { expected_response:true }...: avg=87.96ms  min=3.19ms  med=76.03ms  max=939.62ms p(90)=97.47ms  p(95)=109.65ms
      { scenario:sync_status }.....: avg=88.67ms  min=47.01ms med=76.31ms  max=939.62ms p(90)=97.55ms  p(95)=109.72ms
    http_req_failed................: 0.00%  0 out of 1021
      { scenario:sync_status }.....: 0.00%  0 out of 1000
    http_reqs......................: 1021   11.751873/s
```

This is from my Claude conversation...
> The important question though is whether PostgreSQL will actually push cursor >= 323700 through the DISTINCT ON  view at all. The EXPLAIN will tell you — if you still see the full 1.3M row scan after the view rewrite, then the cursor-first  index won't help via the view, and you'd be better off querying changelog directly with the dedup logic inline, where the  planner has full visibility to use the cursor range up front.

Basically it seems a least on my postgres 16 install, postgres isn't able to use the cursor range supplied in the query through into the view. The AI gave me some promising leads to fix that, but I think either doing the only fly dedup I did before, and/or try to do the inline query dedup might be worth considering.

So I think this is still going to be slow when we keep the de-duping but at least it's faster with these indexes...
Moving to the Subscription based approach for the sync statuses should make this far less critical path, only processors and sync should need to do this de-dup.

# 🧪 Testing

- [ ] test query performance on `changelog`

# 📃 Documentation

  1.
  2.

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/load_test/load-test.js
ria8651, line 1: I don't think the server folder is the right spot, maybe repurpose the load_test folder that gets created by the existing load test cli

andreievg:
@jmbrunskill do you have a benchmark for this on the database we had issues with ? What is the effect of this + postgres optimisation config in that use case ?

jmbrunskill:
> @jmbrunskill do you have a benchmark for this on the database we had issues with ? What is the effect of this + postgres optimisation config in that use case ?

Sorry @andreievg getting back to this now, I'm not 100% sure what you're looking for with benchmark in postgres config.

The issue (sorry not worded great) shows an improvement from median of 140ms per request to median of 76 ms per request. I didn't capture the exact test case and parameters. Do you want me to repeat again capturing default postgres config (before and after) and optimised config (before and after)?

I can provide the test script if you want. Currently kept it out of the repo but can share. I've had various versions of it now testing different scenarios.

jmbrunskill:
@andreievg I see these indexes as an alternative to the no more de-duping approach we'll have from sync v7. This is just a temporary improvement. Maybe I close this and instead manually add the index on the client servers that have this issue? Would you prefer that?

jmbrunskill:
Closing will wait for sync v7 to solve this problem...

PR #11219, opened 2026-04-13
#9654 PO line price rounding and label updates

### LINKED ISSUE #9654: PO Line price rounding needs adjustment
## What went wrong? 😲

The "Price per unit (before discount)" should not be rounded - due to some countries ordering tablets in ones, price can be very small, and they order a large qty - like a million - so each 1/10th of a cent is $1000 in value

Also, the "Total cost" field should have currency formating applied. This should be rounded to the currency unit's smallest value (not just formatted: actually rounded)

While I'm here: could we update the labels for PO_line.Note and PO_line.Comment to "Comment (for Supplier)" and "Note (internal)" - thanks
## Expected behaviour 🤔
As above

## Your environment 🌱

release-demo on 2025-11-01 (v2.11.00 )

esmehm:
Fixes #9654

# 👩🏻‍💻 What does this PR do?

Addresses all three points from the issue:

1. **Increased price decimal precision**: Price per pack (before and after discount) inputs now allow 6 decimal places (up from 5), using a `PRICE_DECIMAL_LIMIT` constant for clarity. We chose 6dp because at 1 million units (a realistic order size for individual tablets), the 6th decimal place represents a $1 difference in total cost — a reasonable threshold of materiality. Beyond 6dp the value per digit drops below $1 even at large volumes, and starts looking like floating-point noise to users.
2. **Total cost currency rounding**: The total cost display in the PO line modal now uses the currency's actual precision via `options.precision` from `useCurrency` (e.g. 0dp for DJF, 2dp for USD). The PO detail table footer sums unrounded line totals and rounds only the final sum, so the displayed total is representative of what's stored in the backend and flows through to inbound shipments.
3. **Label renames**: "Comment" → "Comment (for Supplier)", "Notes" → "Note (internal)" with new translation keys

image

- This PR keeps changes to visual formatting / display rounding only — server-side rounding previously included for when creating the inbound shipment has been reverted
- The client-side total cost display uses the correct currency precision via `useCurrency`
- Total cost is display-only on PO lines (never sent to server). It only becomes a stored value when creating an inbound shipment from the PO
- The broader issue of adding explicit total fields to `purchase_order_line` (so rounded line totals are persisted and transferred to the inbound shipment) has been split out to #11274

# 🧪 Testing

- [ ] Open a Purchase Order and edit a line
- [ ] Verify "Price per pack (before discount)" and "Price per pack (after discount)" accept 6 decimal places
- [ ] Verify "Total cost" rounds to the currency's smallest unit (e.g. 2dp for USD)
- [ ] Verify labels show "Comment (for Supplier)" and "Note (internal)"
- [ ] Verify the footer total in the PO detail table sums correctly (should match the sum of unrounded line costs, rounded once at the end)

Eg 
image
121.23461 + 1123.444 =  1,244.68 (2dp) shown as adding unrounded values
121.23+ 1123.44 = 1,244.67 what you would expect if adding rounded values (those shown)

# 📃 Documentation

# 📃 Reviewer Checklist

**Breaking Changes**

**Issue Review**
- [x] A follow up issue(s) have been created to cover additional requirements — #11274

**Tests Pass**
- [ ] Postgres
- [ ] SQLite
- [ ] Frontend


server/service/src/invoice/inbound_shipment/add_from_purchase_order.rs
```diff
@@ -70,12 +74,16 @@ pub fn add_from_purchase_order(
...
+            total_before_tax: round_2dp(
+                purchase_order_line.price_per_pack_after_discount
+                    * exchange_rate
+                    * number_of_packs,
+            ),
+            total_after_tax: round_2dp(
+                purchase_order_line.price_per_pack_after_discount
+                    * exchange_rate
+                    * number_of_packs,
+            ),
```
CarlosNZ: Just wondering why we need to round at all here? Why not keep the full precision and just do the rounding in the display?

CarlosNZ (commented):
Looking good. Just a couple of tiny quirks (in addition to the inline question) one of which I think we can address:

1. 
Image

There's a rounding display quirk in that the "Price per pack after discount" value should really be displayed as `0.321500` here, since there is additional precision hidden away. But this is a quirk of the `Intl.NumberFormat` methods we're using for formatting, so can't really do much about this right now. Just noting for future consideration.

2. 
Image

You've got the line total rounded to the currency precision in the modal, but in the table it's still showing the `...` for additional precision. Do you think we should be consistent here? I'm not actually sure myself, what do you think?

esmehm:
> You've got the line total rounded to the currency precision in the modal, but in the table it's still showing the `...` for additional precision. Do you think we should be consistent here? I'm not actually sure myself, what do you think?

Yes I agree that we should be consistent - have pushed another commit to change this
image
image

server/service/src/invoice/inbound_shipment/add_from_purchase_order.rs
esmehm: The issue specifically calls out `not just formatted: actually rounded` so I read that as rounding the stored value, not just the display. On the PO line itself total cost is display-only, so the only place an actual stored total exists is when the inbound shipment line is created. 
Happy to revert if display-only rounding satisfies the intent though
@craigdrown can you explain your request?

craigdrown: If we don't round the stored value we get incorrect summation. Say we are rounding to 2 DP
2 line totals 1.144 + 2.144 = invoice total 3.288
would show as 1.14 + 2.14 = 3.29 and that's... bad ;-)
(Another way to look at it: without actual rounding, you would pay 3.28 if you purchased the lines on 2 invoices (each with only one line on them), but 3.29 if you purchased them on the same invoice)

CarlosNZ: That makes sense, but it'd be good to have these "rules" all defined somewhere (e.g. "Always round line totals for all invoice types"), and should be another consideration in the ongoing "how to handle numerical precision" [discussion](https://github.com/msupply-foundation/open-msupply/pull/8907).

CarlosNZ: Also, that means we have an inconsistency here -- you're rounding the line for the generated inbound shipment from the PO, but the purchase order itself doesn't do this, it doesn't store the line total at all and just rounds in the UI, which means you'd get the problem @craigdrown described.

Do we need to add explicit total fields to `purchase_order_line` table like `invoice_line`? (Not for this PR, but if so we should make a new issue)

CarlosNZ (approved):
Thanks @esmehm. Approving, but see my latest comment about the potential inconsistency between POs and invoices -- let's remember to follow-up and make a new issue if necessary.

server/service/src/invoice/inbound_shipment/add_from_purchase_order.rs
craigdrown: @CarlosNZ - you're right: we have the same problem, and it needs fixing thanks (new issue agreed). 
I think our ability to get this stuff right is a differentiator - we learned all these lessons by making the mistakes first time we implemented in OG - good to catch it before we get burned twice.

esmehm: Ok in that case - do we agree the way forwards is that this ticket should be visual formatting only (remove the rounding when making the IS)?
And then follow up ticket will add explicit totals to purchase_order_line like invoice_line and then transfer those values when generating the IS? So summing rounded line totals would be addressed in this second ticket

CarlosNZ: Sounds good to me, @esmehm, just remember to put a link back to this comment thread in the new issue for context.

Thanks 🙏

esmehm: Thanks both! Have spat out https://github.com/msupply-foundation/open-msupply/issues/11274

Please take another look @CarlosNZ to check you're happy with the updated changes please :)

CarlosNZ (approved):
Okay, cool, so you've removed all the rounding on the assumption that there'll be a fixed d.p. value for the line totals after that new issue is created. 

LGTM, thanks 👍

